PR #320, opened 2019-06-21
Add Math.log() operator

simonbuerger:
I had need for Math.log in an Animation calculation. This code all seemed to work and _I think_ I found all the touch points, but I'm not well versed on the native side of things so please let me know if anything needs to change.


wcandillon:
top!

Just out of curiosity, what kind of animation requires `log` ? ;-)
Cool PR, in the meantime you can use this in your project: https://developer.download.nvidia.com/cg/log.html This is a little gem from NVidia that allows us to easily build complex math functions for reanimated. For instance: https://github.com/wcandillon/react-native-redash/blob/master/src/Math.ts#L44

I would be happy to add log to redash, didn't realize it could be useful ;-)

simonbuerger:
Awesome!

A loan repayment calculator ;) Two sliders for duration and payment amount with some fun math to work out one from the other (and some other variables).

PR #323, opened 2019-06-23
fix(🐛): Add TS support for animated transforms

wcandillon:
🎉Finally, we can type View transformations 😍

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -141,8 +141,10 @@ declare module 'react-native-reanimated' {
 
     export const SpringUtils: SpringUtils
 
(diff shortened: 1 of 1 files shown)


wcandillon:
🙌🏻I use https://www.npmjs.com/package/patch-package to run patched version of the typing on older reanimated releases and it works like a charm.

tpucci:
Hi 👋 @wcandillon I have trouble using this commit. It seems that `TransformStyle` is not defined.

tpucci:
I fixed it here: https://github.com/kmagiera/react-native-reanimated/pull/339

wcandillon:
@tpucci yes thks 🙌🏻

PR #311, opened 2019-06-15
fix(🐛): Fix Android bug when dealing with string animated values

wcandillon:
This PR takes @msand and @osdnk  from #176 into consideration.

fixes #300

Issue #300, opened 2019-06-03
Animated String Values on Android


wcandillon:
While string values work great on iOS, they throw errors and crash on Android. 
Below are two examples that work great on iOS but crash or throw on Android with v1.1.0.

## Example 1

```ts
import { TextInput } from "react-native";
const AnimatedTextInput = Animated.createAnimatedComponent(TextInput);

export default () => {
  return (
    <AnimatedTextInput
        underlineColorAndroid="transparent"
        editable={false}
        text={concat("hello ", "world")}
    />
  );
};
```

Error: ConcatNode cannot be cast to a number

## Example 2

```ts
import Svg, { Path } from "react-native-svg";
import Animated from "react-native-reanimated";

const AnimatedPath = Animated.createAnimatedComponent(Path);

export default () => {
  return (
    <Svg style={styles.container} viewBox="0 0 212 239">
      <AnimatedPath
        d={concat(
          "M106,1 C156,1 229.606798,153.786405 206,201 C182.393202,248.213595 31.5199889,252.039978 6,201 C-19.5199889,149.960022 56,1 106,1 Z"
        )}
      />
    </Svg>
  );
};
```

Crashes

ken0x0a:
I'm trying to implement ScrollView using reanimated to get over this issue https://github.com/kmagiera/react-native-reanimated/issues/160, and come up to this crash by following code.

```

dodicandra:
hi, I tried to make the component also from createAnimatedComponent, but I don't know all the props in that component ... I use typescript for passing props, but I also don't know the correct way

i made it like this ..
@wcandillon 

`const AnimatedFlatList: React.FC<FlatListProps<DataProps>> = Animated.createAnimatedComponent(FlatList);`

PR #176, opened 2019-01-23
Support animating Svg Path d attribute on android

msand:
https://github.com/react-native-community/react-native-svg/issues/908

```jsx
import * as React from 'react';
import { View, StyleSheet } from 'react-native';
import { Svg, Path } from 'react-native-svg';
import Animated from 'react-native-reanimated';

const { concat, multiply } = Animated;
const x = multiply(1, 2);
const y = multiply(2, 3);
const path = concat('M 0 0 L ', x, ' ', y);

const AnimatedPath = Animated.createAnimatedComponent(Path);

export default class App extends React.Component {
  render() {
    return (
      <View style={styles.container}>
        <Svg viewBox="0 0 10 10" width="100" height="100">
          <AnimatedPath d={path} stroke="black" />
        </Svg>
      </View>
    );
  }
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    backgroundColor: '#ecf0f1',
  },
});
```

PR #131, opened 2018-11-18
[Android] Fix props node

futuun:
I found this few days ago while working on #110 and I thought it might be worth fixing.
It might be useful when you want to pass different props to component based on animated value.

PR #333, opened 2019-07-04
Refactor style and props node evaluation

simonbuerger:
I'm stepping on some toes here with this PR (@wcandillon, mainly - sorry!)

Related PRs: #311 #131 #176 

Related issues: #300 

I'm not even sure if this is the most efficient approach (my Java is super rusty), but it seemed that the style node values were unnecessarily being looped over, evaluated, added to a WritableMap, then again in PropsNode they were being looped over, evaluated, type checked and then added to a new WritableMap.

It seemed really inefficient, and I couldn't think of a reason that we wouldn't just merge the style and main props together and only loop and evaluate everything once.

Please shoot me down if this will break things somewhere else.


android/src/main/java/com/swmansion/reanimated/nodes/PropsNode.java
```diff
@@ -52,49 +50,40 @@ protected Double evaluate() {
...
+        dest.putArray(key, (WritableArray) val);
       } else {
-        String key = entry.getKey();
-        if (mNodesManager.uiProps.contains(key)) {
-          hasUIProps = true;
-          mPropMap.putDouble(key, node.doubleValue());
+        if (val instanceof Double) {
+          dest.putDouble(key, (Double) val);
+        } else if (val instanceof String) {
+          dest.putString(key, (String) val);
```
simonbuerger: I took the types here from the previous StyleNode evaluation. Do we need more to support PropsNode? Almost every AnimatedNode returns Double or String. But I can think also possibly Boolean and Array too? Do any Nodes return Integers?

simonbuerger: Looking through all the other node types I can't see Integers or WritableMaps being returned. I have added Boolean, null and WritableArray

wcandillon:
@simonbuerger Huge thanks for being on top of this :)

osdnk:
@kmagiera

simonbuerger:
For reference I have manually tested this on Android in an animated loan repayment calculator I'm building for a client. It involves animated props that are boolean, string (concatted and formatted currency values passed to TextInput native text prop, and react-native-svg animated path `d` prop with a few custom bezier curves), and a bunch of style props from width, color, fontSize, opacity and a few translate X and Y transforms. No issues with any of those so far.

wcandillon:
that sounds so cool :)

android/src/main/java/com/swmansion/reanimated/nodes/StyleNode.java
```diff
@@ -21,23 +21,7 @@ public StyleNode(int nodeID, ReadableMap config, NodesManager nodesManager) {
...
-        } else if (val instanceof String) {
-          propMap.putString(entry.getKey(), (String) val);
-        } else {
-          throw new IllegalStateException("Wrong style form");
-        }
-      }
-    }
-    return propMap;
+  protected Map evaluate() {
+    return mMapping;
```
kmagiera: the code you removed it in fact necessary. Without it the input nodes evaluation for styles will only happen once. So when the styles updates as a result of running animations those changes won't be reflected on the view. You may not see the issue while testing maybe because you only rely on transforms, but if you update a prop that actually belongs to styles you should be able to see this (e.g. opacity). Also there is a chance you re-render from JS too often in which case the styles would get recreated.

simonbuerger: @kmagiera in the production app where I'm using this I am animating opacity in quite a few places, and there are also no JS re-renders on the page once it is loaded.

I've just moved the evaluation of the style input nodes to the PropsNode evaluate method

simonbuerger: @kmagiera I've just checked and actually in our use case we're animating width, opacity, fontSize and color too. All animate as expected.

kmagiera: Great to hear it is working ok. However I am not sure I follow here. Are you saying you are testing this PR on top of some other changes fro mone of the PRs you listed?

kmagiera: AH, ok, I think I understand this now. Please ignore my prev comment

android/src/main/java/com/swmansion/reanimated/nodes/ConcatNode.java
```diff
@@ -3,21 +3,36 @@
...
+  private final NumberFormat nf;
+
   private final int[] mInputIDs;
 
   public ConcatNode(int nodeID, ReadableMap config, NodesManager nodesManager) {
     super(nodeID, config, nodesManager);
     mInputIDs = Utils.processIntArray(config.getArray("input"));
+    symbols = new DecimalFormatSymbols(Locale.getDefault());
+    symbols.setDecimalSeparator('.');
+    nf = new DecimalFormat("##.###", symbols);
```
kmagiera: Can we leave this change out of this PR. It does not seem relevant IMO

simonbuerger: Sure. For some context, the original point of the PR was to get string animated values working on Android. And without this change almost all the use cases for string animated Values actually break because the animated Double values display always with a decimal point even when they are whole integers (eg 12.0 rather than 12). Without this `concat(new Value(12))` will return different results on iOS and Android - so should I rather put this particular change it in another PR then?

simonbuerger:
@kmagiera I like the idea of flattening style and props before they even reach native. Let me have a go at it

simonbuerger:
@kmagiera this is completely untested, but my first thought is that the only update required to do this is to update `AnimatedProps.js` (see an initial stab below)?

Aside from having to rip out all the traces of the separate style nodes code across native and JS of course (not a small task 😅)

```
import { findNodeHandle, StyleSheet } from 'react-native';

import AnimatedNode from './AnimatedNode';
import AnimatedEvent from './AnimatedEvent';
import { createOrReuseTransformNode } from './AnimatedTransform';

import invariant from 'fbjs/lib/invariant';
import deepEqual from 'fbjs/lib/areEqual';

function sanitizeProps(inputProps) {
  const props = {};
  for (const key in inputProps) {
    const value = inputProps[key];
    if (value instanceof AnimatedNode && !(value instanceof AnimatedEvent)) {
      props[key] = value.__nodeID;
    }
  }
  return props;
}

export function createOrReusePropsNode(props, callback, oldNode) {
  if (props.style) {
    let { style, ...rest } = props;
    style = StyleSheet.flatten(style) || {};
    if (style.transform) {
      style = {
        ...style,
        transform: createOrReuseTransformNode(
          style.transform,
          oldNode && oldNode._props.transform
        ),
      };
    }
    props = {
      ...rest,
      ...style,
    };
  }
  const config = sanitizeProps(props);
  if (oldNode && deepEqual(config, oldNode._config)) {
    return oldNode;
  }
  return new AnimatedProps(props, config, callback);
}

class AnimatedProps extends AnimatedNode {
  constructor(props, config, callback) {
    super(
      { type: 'props', props: config },
      Object.values(props).filter(n => !(n instanceof AnimatedEvent))
    );
    this._config = config;
    this._props = props;
    this._callback = callback;
    this.__attach();
  }

  _walkPropsAndGetAnimatedValues(props) {
    const updatedProps = {};
    for (const key in props) {
      const value = props[key];
      if (value instanceof AnimatedNode) {
        updatedProps[key] = value.__getValue();
      } else if (value && !Array.isArray(value) && typeof value === 'object') {
        // Support animating nested values (for example: shadowOffset.height)
        updatedProps[key] = this._walkPropsAndGetAnimatedValues(value);
      }
    }
    return updatedProps;
  }

  __onEvaluate() {
    return this._walkPropsAndGetAnimatedValues(this._props);
  }

  __detach() {
    const nativeViewTag = findNodeHandle(this._animatedView);
    invariant(
      nativeViewTag != null,
      'Unable to locate attached view in the native tree'
    );
    this._disconnectAnimatedView(nativeViewTag);
    super.__detach();
  }

  update() {
    this._callback();
  }

  setNativeView(animatedView) {
    if (this._animatedView === animatedView) {
      return;
    }
    this._animatedView = animatedView;

    const nativeViewTag = findNodeHandle(this._animatedView);
    invariant(
      nativeViewTag != null,
      'Unable to locate attached view in the native tree'
    );
    this._connectAnimatedView(nativeViewTag);
  }
}
```

kmagiera:
Nice! Looking forward to more updates

simonbuerger:
@kmagiera please review again. As suggested I have removed the StyleNode entirely and am now merging props and style props in AnimatedProps.js before they reach native

android/src/main/java/com/swmansion/reanimated/nodes/ConcatNode.java
simonbuerger: @kmagiera please confirm if I should still move this to a separate PR?

jakub-gonet:
@simonbuerger, what's the status of this PR?

jakub-gonet:
I'm closing this one due to inactivity and current Reanimated situation - we don't want to introduce big new changes to v1.

PR #335, opened 2019-07-08
Fixes transitioning API delay when delayMs is not set

tomasgcs:
After implamenting "Slow Transitions" feature an issue was introduced with Transitioning API as you can see here https://github.com/tomasgcs/react-native-reanimated/commit/4ef8b81df3f8981cb1dbd9ef0b7867165f2236c0#diff-a7ea5763fc80cde0b741dca6a5095a18R46

Previously `beginTime` variable was not set at all which was equal to 0, but after this specific line was added the value is different, and it causes animations to delay even if `delayMs` value is not set, this happens only when animation is is being executed multiple times.

Not sure why this is happening but most likely that `CACurrentMediaTime()` introduces delay even if it takes current time. In my case it is very important to run animations as fast as possible so I've sumbitted a fix here.

### DIFF
--- a/ios/Transitioning/REATransitionAnimation.m
+++ b/ios/Transitioning/REATransitionAnimation.m
@@ -42,6 +42,11 @@ CGFloat SimAnimationDragCoefficient()
 
 - (void)play
 {
+  /*
+  CACurrentMediaTime introduces some kind of delay  even if _delay is set to 0
+  it calls mach_absolute_time() which is based on the last time the device booted
+  which might cause the delay
+  */
   if (_delay > 0){
     _animation.beginTime = CACurrentMediaTime() + _delay * SimAnimationDragCoefficient();
   }
--- a/ios/Transitioning/REATransitionAnimation.m
+++ b/ios/Transitioning/REATransitionAnimation.m
@@ -42,8 +42,10 @@ CGFloat SimAnimationDragCoefficient()
 
 - (void)play
 {
+  if (_delay > 0){
+    _animation.beginTime = CACurrentMediaTime() + _delay * SimAnimationDragCoefficient();
+  }
   _animation.duration = self.duration * SimAnimationDragCoefficient();
-  _animation.beginTime = CACurrentMediaTime() + _delay * SimAnimationDragCoefficient();
   [_layer addAnimation:_animation forKey:_keyPath];
 }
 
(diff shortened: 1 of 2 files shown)


tomasgcs:
@kmagiera Hello, could you take a look at this PR?

ios/Transitioning/REATransitionAnimation.m
```diff
@@ -42,8 +42,10 @@ + (REATransitionAnimation *)transitionWithAnimation:(CAAnimation *)animation
 
 - (void)play
 {
+  if (_delay > 0){
```
kmagiera: I'm fine with merging this change in. However, I'd like to have a comment here explaining why this check has been introduced (basically what you explain in the PR description).

I also wonder what is the delay in case we use CACurrentMediaTime(). Is this like 1 frame delay (16ms) or more than that? Also if you set delay to be 60 ms is it going to start after 60ms or after 60ms + 1 frame

osdnk:
ping @tomasgcs

tomasgcs:
Apologies for delays, added a comment for the check. Also found an article about CACurrentMediaTime:
https://bendodson.com/weblog/2013/01/29/ca-current-media-time/

as Apple in apple docs

A CFTimeInterval derived by calling mach_absolute_time() and converting the result to seconds.

So I am thinking that it might be rounding issue or something like that.

tomasgcs:
@osdnk @kmagiera 

Check my latest commit an new issue was introduced with this commit 

https://github.com/kmagiera/react-native-reanimated/commit/50e695fb4e51f0b579dab1d383a2f47605859123#diff-7ab6d44707ac6cd944a490906c842dbb

backwards compatable animations stopped working.

osdnk:
Looks good! Please ping whenever you'll find it mergeable

tomasgcs:
@osdnk 

Animations are fixed here:
https://github.com/kmagiera/react-native-reanimated/pull/335/commits/4d98f4ffd0d290ed5d4072c67e606b245b1f58d1

And I already added comments for `_delay` animations issue here:
https://github.com/kmagiera/react-native-reanimated/pull/335/commits/24401b46e26d1d7c1904e8c1b2a25abde29d9c05

Apologies for not creating a separate PR for backward compatible animations fix. But I can do that if necessary.

PR #336, opened 2019-07-08
Add format node

simonbuerger:
After having to work extensively with displaying dynamic animated numbers as text without crossing the bridge (via the TextInput `text` prop, which I stumbled onto thanks to @wcandillon and react-native-redash). I wrote some pretty gnarly number formatters (adding thousand separators, currency, "padding" values etc.) using just the available mathematical operators in reanimated. They were crazy complicated and quite unwieldy (although I did get them to work).

I then thought I'd take a look to see if there was a standard number formatting approach in iOS and Android on the native side, and damned if there isn't. Both platforms allow numbers to be formatted using the unicode TR-35 spec number format patterns - https://www.unicode.org/reports/tr35/tr35-31/tr35-numbers.html#Number_Format_Patterns

Seemed really useful to me. Currencies, number padding, prefixes/suffixes/percentages/rounding with increments, localisation of separators and decimals etc.

I largely just copied concat, allowed it to support only a single value arg, and added the formatters. Happy to take suggestions here for a better API. I also wasn't sure if I should call it format or numberFormat. I figured other formatters wouldn't really be possible given what we're talking about so went with plain "format", but open to other suggestions.

So the use case would be for display purposes beyond just raw or rounded numbers (custom slider controls, in my case - but really any kind of animated custom value controls). Avoiding the bridge to render those Animated values makes some great UI possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -711,8 +711,8 @@ Format string will default to `##.###`, which is simply a standard decimal forma
 Number format patterns are as per the Unicode TR35 spec https://www.unicode.org/reports/tr35/tr35-31/tr35-numbers.html#Number_Format_Patterns
 
 Examples:
-`format(1234567, '$#,###') //$1,234,567 (localized)`
-`format(1234.1, '$#,##0.00') //$1,234.10 (localized)`
+`format(1234567, '$#,###') // $1,234,567 (localized)`
+`format(1234.1, '$#,##0.00') // $1,234.10 (localized)`
 
 ---
 ### `onChange`
--- a/README.md
+++ b/README.md
@@ -711,8 +711,8 @@ Format string will default to `##.###`, which is simply a standard decimal forma
 Number format patterns are as per the Unicode TR35 spec https://www.unicode.org/reports/tr35/tr35-31/tr35-numbers.html#Number_Format_Patterns
 
 Examples:
-`format(1234567, '$#,###') //$1,234,567`
-`format(1234.1, '$#,##0.00') //$1,234.10`
+`format(1234567, '$#,###') //$1,234,567 (localized)`
+`format(1234.1, '$#,##0.00') //$1,234.10 (localized)`
 
 ---
 ### `onChange`
--- a/README.md
+++ b/README.md
@@ -700,6 +700,17 @@ concat(nodeOrValue1, ...)
 Returns concatanation of given nodes (number or string) as string
 
 ---
+### `format`
+```js
+format(nodeOrValue, [format='##.###'])
+```
+Similar to concat. If `nodeOrValue` resolves to a number, the number is formatted according to the number formatter string passed as the second argument. A string will just be returned as is.
+
+Format string will default to `##.###`, which is simply a standard decimal format with up to 3 decimal places. (eg. 12.12, 1234, 12.134)
+
+Number format patterns are as per the Unicode TR35 spec https://www.unicode.org/reports/tr35/tr35-31/tr35-numbers.html#Number_Format_Patterns
+
+---
 ### `onChange`
 
 ```js
--- a/README.md
+++ b/README.md
@@ -710,6 +710,10 @@ Format string will default to `##.###`, which is simply a standard decimal forma
 
 Number format patterns are as per the Unicode TR35 spec https://www.unicode.org/reports/tr35/tr35-31/tr35-numbers.html#Number_Format_Patterns
 
+Examples:
+`format(1234567, '$#,###') //$1,234,567`
+`format(1234.1, '$#,##0.00') //$1,234.10`
+
 ---
 ### `onChange`
 
--- a/android/src/main/java/com/swmansion/reanimated/nodes/FormatNode.java
+++ b/android/src/main/java/com/swmansion/reanimated/nodes/FormatNode.java
@@ -1,0 +1,31 @@
+package com.swmansion.reanimated.nodes;
+
+import com.facebook.react.bridge.ReadableMap;
+import com.swmansion.reanimated.NodesManager;
+import com.swmansion.reanimated.Utils;
+import java.text.NumberFormat;
+import java.text.DecimalFormat;
+
+public class FormatNode extends Node {
+  private NumberFormat nf;
+  private final int mInputID;
+
+  public FormatNode(int nodeID, ReadableMap config, NodesManager nodesManager) {
+    super(nodeID, config, nodesManager);
+    mInputID = config.getInt("value");
+    nf = new DecimalFormat(config.getString("format"));
+  }
+
+  @Override
+  protected String evaluate() {
+    StringBuilder builder = new StringBuilder();
+    Node inputNode = mNodesManager.findNodeById(mInputID, Node.class);
+    Object value = inputNode.value();
+    if (value instanceof Double) {
+      builder.append(nf.format(value));
+    } else {
+      builder.append(value);
+    }
+    return builder.toString();
+  }
+}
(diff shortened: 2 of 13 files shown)


wcandillon:
This is amazing 🙌🏻

README.md
```diff
@@ -699,6 +699,21 @@ concat(nodeOrValue1, ...)
...
+```
+Similar to concat. If `nodeOrValue` resolves to a number, the number is formatted according to the number formatter string passed as the second argument. A string will just be returned as is.
+
+Format string will default to `##.###`, which is simply a standard decimal format with up to 3 decimal places. (eg. 12.12, 1234, 12.134)
+
+Number format patterns are as per the Unicode TR35 spec https://www.unicode.org/reports/tr35/tr35-31/tr35-numbers.html#Number_Format_Patterns
+
+Examples:
+`format(1234567, '$#,###') //$1,234,567 (localized)`
+`format(1234.1, '$#,##0.00') //$1,234.10 (localized)`
```
paldepind: Maybe add a space after the `//`? That is done elsewhere in the readme.

react-native-reanimated.d.ts
```diff
@@ -225,6 +225,10 @@ declare module 'react-native-reanimated' {
     export function concat(
       ...args: Array<Adaptable<string> | Adaptable<number>>,
     ): AnimatedNode<string>;
+    export function format(
+      value: Adaptable<number>,
+      format: Adaptable<string>
```
paldepind: The type doesn't indicate that the last argument is optional. This change should do the trick.

```diff
- format: Adaptable<string>
+ format?: Adaptable<string>
```

README.md
simonbuerger: Done

react-native-reanimated.d.ts
simonbuerger: Thanks for the review! And done

ios/Nodes/REAFormatNode.m
```diff
@@ -0,0 +1,34 @@
...
+}
+
+- (id)evaluate
+{
+  NSMutableString *result = [NSMutableString new];
+  id val = [[self.nodesManager findNodeByID:_value] value];
+  if ([val isKindOfClass:[NSNumber class]]) {
+    NSNumberFormatter *numberFormatter = [[NSNumberFormatter alloc] init];
+    [numberFormatter setPositiveFormat:_format];
+    [result appendString:[numberFormatter stringFromNumber:val]];
```
kmagiera: Small nit, but I think it'd be better to return here and in the other `if` instead of appending to an empty string.

For example:

```
if ([...]) {
  ...
  return [numberFormatter stringFromNumber:val];
}
```

ios/Nodes/REAFormatNode.m
```diff
@@ -0,0 +1,34 @@
...
+  }
+  if ([val isKindOfClass:[NSString class]]) {
```
kmagiera: val cannot be of both `NSString` and `NSNumber` types AFAIK. Please change this to `else` to clarify the intent.

Also if you use `return` as per my prev suggestion that'd also help to make it cleaner

ios/Nodes/REAFormatNode.m
```diff
@@ -0,0 +1,34 @@
...
+@implementation REAFormatNode {
+  NSNumber *_value;
+  NSString *_format;
+}
+
+- (instancetype)initWithID:(REANodeID)nodeID config:(NSDictionary<NSString *,id> *)config
+{
+  if ((self = [super initWithID:nodeID config:config])) {
+    _value = config[@"value"];
+    _format = config[@"format"];
```
kmagiera: we can perhaps precompute `NSNumberFormatter` here already instead of allocating it each time `evaluate` gets called

android/src/main/java/com/swmansion/reanimated/nodes/FormatNode.java
```diff
@@ -0,0 +1,31 @@
+package com.swmansion.reanimated.nodes;
+
+import com.facebook.react.bridge.ReadableMap;
+import com.swmansion.reanimated.NodesManager;
+import com.swmansion.reanimated.Utils;
+import java.text.NumberFormat;
+import java.text.DecimalFormat;
+
+public class FormatNode extends Node {
+  private NumberFormat nf;
```
kmagiera: can we stick to Android java naming convention (as it is being used elsewhere) here and name it `mNumberFormat` or sth that starts from lowercase `m`

android/src/main/java/com/swmansion/reanimated/nodes/FormatNode.java
```diff
@@ -0,0 +1,31 @@
...
+    nf = new DecimalFormat(config.getString("format"));
+  }
+
+  @Override
+  protected String evaluate() {
+    StringBuilder builder = new StringBuilder();
+    Node inputNode = mNodesManager.findNodeById(mInputID, Node.class);
+    Object value = inputNode.value();
+    if (value instanceof Double) {
+      builder.append(nf.format(value));
```
kmagiera: return here or assign `nf.format(val)` to some `result` variable (same commend as in the iOS code). There is no point in allocating string builder and appending only once to an empty string

kmagiera:
In high level I think it'd be better for the `format` node to support formatting style that's closer to `printf` like syntax from c++ and other languages. Both java and objc supports this syntax. The benefit is that apart from supporting number formatting we'd also get string padding and can also do more fancy number formatters in a more readable manner. WDYT?

simonbuerger:
@kmagiera that's a great idea. I'll see if I can get a workable implementation together and update this PR

kmagiera:
Sweet, thanks so much for working on this

simonbuerger:
@kmagiera hmmm, so I got this working on Android well with String.format. However in objective c I'm having a lot of trouble taking a variable length array of input values and passing them to either stringWithFormat or initWithFormat as arguments. Any thoughts on possible ways to achieve this?

kmagiera:
I think you'll need to use this method: https://www.cocoawithlove.com/2009/05/variable-argument-lists-in-cocoa.html and also check this article to learn how to create `va_list` https://www.cocoawithlove.com/2009/05/variable-argument-lists-in-cocoa.html

wcandillon:
@simonbuerger Are there still plans to work on this? It's very useful.

lundn:
Please continue with this!

simonbuerger:
@wcandillon @lundn frankly I was unable to get @kmagiera suggestion for working with variable number of arguments in Objective C to work.

Although I am not an Objective C developer by trade so someone else might have more luck.

So to get this working as suggested I would need some help

cltsang:
I studied @wcandillon 's coinbase code, and he uses different code to format [JS number](https://github.com/wcandillon/can-it-be-done-in-react-native/blob/84164a348a0b7842661dd1f45a0ea2403461a24e/season3/src/CoinbasePro/Values.tsx#L39) and [reanimated node number](https://github.com/wcandillon/can-it-be-done-in-react-native/blob/84164a348a0b7842661dd1f45a0ea2403461a24e/season3/src/CoinbasePro/Label.tsx#L34), which is not ideal and limiting.  Before this is merged, is this the only way to do it currently?

I'm wondering if there is a way to do something like using [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) to format numbers.

jakub-gonet:
@simonbuerger do you plan to finalize this PR? I can help you a bit with the ObjC if you want to.

wcandillon:
maybe with Reanimated 2, it might not be worth the effort?

jakub-gonet:
I leave the decision to the PR author, v1 still will be used in apps for some time. 
I agree that v2 makes formatting strings substantially easier.

jakub-gonet:
Closing due to no response.

PR #339, opened 2019-07-11
:bug: (react-native-reanimated.d.ts) fix TransformsStyle

tpucci:
This fixes typescript error `Cannot find name 'TransformStyle'.`.


wcandillon:
@tpucci Thank you for fixing this! I can confirm that it fixes the issue. I didn't notice this bug before because I had `skipLibCheck: true` 🙈Sorry about that!

PR #340, opened 2019-07-11
Update react-native version

wojteg1337:
I've upgraded react-native to 60.0 in library and example project, next step will be to create bindings and specs for turbo module support.


android/build.gradle
```diff
@@ -27,6 +27,7 @@ repositories {
         // https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/local-cli/templates/HelloWorld/android/build.gradle#L21
         url "$projectDir/../node_modules/react-native/android"
     }
+    google()
```
osdnk: is it needed?

android/gradle.properties
```diff
@@ -16,5 +16,3 @@ org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryErro
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
-
-android.useDeprecatedNdk=true
```
osdnk: hm?

android/src/main/java/com/swmansion/reanimated/transitions/ChangeTransition.java
```diff
@@ -3,15 +3,15 @@
...
-import android.support.transition.ChangeBounds;
-import android.support.transition.ChangeTransform;
-import android.support.transition.Transition;
-import android.support.transition.TransitionPropagation;
-import android.support.transition.TransitionValues;
 import android.view.ViewGroup;
 
+import androidx.annotation.Nullable;
+import androidx.transition.ChangeBounds;
+import androidx.transition.ChangeTransform;
```
osdnk: I wonder if we will be still able to use in old RN version

babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ['module:metro-react-native-babel-preset'],
```
osdnk: shouldn't we drop `.babelrc` then?

Example/ios/Podfile
```diff
@@ -0,0 +1,49 @@
+platform :ios, '9.0'
```
osdnk: great!

wojteg1337:
> Lovely! I'm not super super about the implication of moving to AndroidX. Could you tell me more about it? Are we going to drop support for prev version then?

Probably yes, this is breaking change, users with older versions of RN will have to use current stable version or update their projects

android/gradle.properties
wojteg1337: while building example project i noticed:

image

so decided to remove it

babel.config.js
wojteg1337: sure, I have removed `.babelrc` file

android/build.gradle
wojteg1337: nope, removed google repo

rozPierog:
What is the status of this PR @osdnk? I would love to go back to using "official" version without relying on fork

osdnk:
Can’t we use jettifier?

rozPierog:
Jetifier is _quick and dirty_ fix for library maitainers that [aren't willing to spend much time trying to get AndroidX to work properly](https://facebook.github.io/react-native/blog/2019/07/03/version-60.html#androidx-support). The truth is, [AndroidX is replacing Android Support library, and not supporting it (pun not inteded) fully you are just asking for when apis change](https://developer.android.com/topic/libraries/support-library/). Everyone back on RN <0.60 will still be able to use older versions of this package without problems. Updating RN to current version is now easier than ever with tools like [upgrade-helper](https://react-native-community.github.io/upgrade-helper/)
BTW Jetifier has reverse option that is able to change AndoridX back to Support Library so if someone is using 0.59 and really needs new version of some library they can use that to make it work.

android/src/main/java/com/swmansion/reanimated/transitions/ChangeTransition.java
hyochan: I don't think this will be compatible with the old `RN` version. However, we can manually write a doc for those who use `react-native` version below `0.60`.

kmagiera:
Thanks for submitting this PR, as I've just merged #395 which covers the migration I'm going to close this one. Thank you for your contribution regardless 🙌

PR #342, opened 2019-07-12
Remove deprecated componentWillMount

owinter86:
Refactors the create animated component removing the deprecated componentWillMount lifecycle and moving the attachProps to the constructor.

### DIFF
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -26,6 +26,11 @@ export default function createAnimatedComponent(Component) {
   class AnimatedComponent extends React.Component {
     _invokeAnimatedPropsCallbackOnMount = false;
 
+    constructor(props) {
+      super(props);
+      this._attachProps(this.props);
+    }
+  
     componentWillUnmount() {
       this._detachPropUpdater();
       this._propsAnimated && this._propsAnimated.__detach();
@@ -34,10 +39,6 @@ export default function createAnimatedComponent(Component) {
 
     setNativeProps(props) {
       this._component.setNativeProps(props);
-    }
-
-    componentWillMount() {
-      this._attachProps(this.props);
     }
 
     componentDidMount() {


osdnk:
can't we move to `UNSAFE`. That what we've done in RN core.

owinter86:
Why would we need to do it for this usage of CWM, given we have access to props in the constructor?

osdnk:
let's say:
```
const a = <Animated.View ...>
```
is invoked outside render. It will trigger constructor, but not CWM / CDM. In this case we don't want to attach props, no?

owinter86:
Im not sure I fully understand, sorry. I understand that declaring that constant wont trigger the constructor, but how would you invoke this outside a render and then use it inside a render so different props appear in the constructor and CWM.

osdnk:
```
render() {
  return (
    {this.state.visible ? A : null}
  )
}
```

owinter86:
Doesn't the constructor and CWM get called when state is visible both with the same props?

osdnk:
I think so. Let you check! Maybe I'm wrong :)

owinter86:
From what I understand, the only time the constructor and CWM are called are when this.A is returned in the render method, triggering the constructor and CWM method with the same prop values.

I am not sure how you can invoke `A = <Animated.View ...>` outside a render method, and then reuse the same instance inside a render method, invoking the lifecycle methods with new props?

The example above will call the constructor and CWM only when visible is true, and neither get called when false as this.A is never invoked.

ScreamZ:
Which release ?

suretarget:
@ScreamZ in master branch.
`yarn add https://github.com/kmagiera/react-native-reanimated.git\#master`

ScreamZ:
That's good but I don't like to rely on master using NPM, It's better to fix version :D

avasuro:
Strange thing, according to commits history PR with this fix should have been released in current version (v1.2.0), but if I used this version from npm - error is still appears, and if I look at source code in node_modules - this fix is not presented there.

ScreamZ:
@FrederickEngelhardt So ?

mattmogford:
This doesn't seem to be in 1.2.0 ...?

anastely:
I have v1.2.0 and the issue is still!

sincraianul:
@kmagiera Please release a patch version (1.2.0) on NPM that includes this fix.

rmindo:
This should be in NPM already.

IntelliJAbhishek:
@rmindo its not there

net8floz:
would love a npm fix!

chrisparton1991:
This is fixed in v1.3.0, which is available on NPM. I was still getting the warning after upgrading, but clearing the transform cache fixed that:
```
npm start -- --reset-cache
```

mattmogford:
v1.3.0
```yarn cache clean```
for yarn users

kmagiera:
Sorry for responding here so late but just noticed this change when browsing through the codebase.

We should *not* be calling `_attachProps` in the constructor as this may lead to memory leaks. Attaching props allocates resources on the native side that are then released by `ropsAnimated.__detach()` which is called in `componentWillUnmount`. React does not guarantee that the constructor call leads to the component being mounted. So if it does not get mounted we will never release resources as unmounting won't happen as well.

Instead of doing that this call should've been moved to `componentDidMount` and not into the class constructor.

PR #352, opened 2019-07-25
Fix: Support optional string style types

brmenchl:
It looks like because all style properties are optional, they're failing the conditional type in `AnimateStyle`, because `string | undefined` does not `extend string`.


kmagiera:
Thanks @brmenchl 🙌

PR #354, opened 2019-07-28
Add support for nodes as functions

chrfalch:
# Motivation

When creating multiple dynamic animations with react-native-reanimated the number of nodes that needs to be transported over the bridge is a performance challenge. One of the limitations of the current primitives offered by reanimated is the lack of function nodes. Being able to define node trees that can be reused will offer better performance. A previous attempt has been done in this pull request: https://github.com/kmagiera/react-native-reanimated/pull/42

# Implementation

The suggested implementation tries to limit changes to the current source code by adding new nodes without changing the existing implementation too much. Three new nodes have been added:

- `ParamNode`
- `CallFunctionNode`
- `FunctionNode`

A few changes to the NodeManager to be able to force reevaluation when calling functions have been added.

The implementation uses a simple `proc` function in Javascript to create function nodes. Following is an example of how a function is defined and used.

```js
// Defining the function
const mulXY = proc((x, y) => multipy(x, y));

// Using the function
const myX = new Animated.Value(12);
const myY = new Animated.Value(14);
const myResult = mulXY(myX, myY);
```

## Technical

The solution I have used in this implementation is to implement the ParamNode as a node that holds a reference to other nodes - returning the referenced node's value when evaluated and setting the referenced node's value when changed.

Calling the function node is done using the CallFuncNode that is created when calling the function (on the JS side). When the CallFuncNode is evaluated, it performs the following steps:

1) Start a new context
2) Copy current argument nodes to the param nodes (using a Stack in the param node)
3) Evaluate the function node's expression
4) Pop the current argument nodes from the param nodes
5) End context

I have also changed NodeManager's invalidate strategy so that it will force a reevaluation of nodes when we are in a function call to avoid returning the same value for multiple functions calls with different parameter references.

Using a stack in the ParamNode ensures that we can create recursive functions.

## Memory

A single proc node should be created as a global constant in javascript since they are immediately attached and will never be removed.

## Platforms

Both Android and iOS platform-specific code has been added in addition to the required JS code.

## Further plans

Rewriting functionality in reanimated called often should be considered. Moving complex spring functions to functions could also be considered but is not necessarily part of this PR.

## Testing

Convert built-in derived functionality to use proc nodes and verify that all examples work nicely. I have already done this for `interpolate` and `easings` and it seems to be working fine. Verify that no memory leaks exist.

I believe that the current implementation supports caching of function evaluation - would be happy if someone else could confirm this.


Example/test/index.js
```diff
@@ -92,6 +94,14 @@ export default class Example extends Component {
     // const transX = new Value(0);
     const clock = new Clock();
     // const twenty = new Value(20);
+    const a = param("a");
```
osdnk: Do we need these names? why?

Example/test/index.js
```diff
@@ -92,6 +94,14 @@ export default class Example extends Component {
...
+    const b = param("b");
+    const calc = proc(add(multiply(a, b), 333), a, b);
```
osdnk: Can you add some better example?

android/src/main/java/com/swmansion/reanimated/nodes/Node.java
```diff
@@ -36,7 +36,7 @@ public Node(int nodeID, @Nullable ReadableMap config, NodesManager nodesManager)
   protected abstract @Nullable Object evaluate();
 
   public final @Nullable Object value() {
-    if (mLastLoopID < mUpdateContext.updateLoopID) {
+    if (mLastLoopID < mUpdateContext.updateLoopID || mUpdateContext.contextCount > 0) {
```
osdnk: This part I'm mostly afraid of. This is tradeoff which probably is not needed. I wonder if can't we somehow bind memoization with contexts' stack. I don't want to have to decide if I want to take advantages of memoization of functional nodes. Maybe we can make a stack of memoized values as well? What do you think?

Example/test/index.js
chrfalch: No we don't need the names - it was just to make it a little more easy to debug.. ;)

Example/test/index.js
chrfalch: Yes I have another example where I have converted all easings to use `proc` nodes - should I commit that as well?

android/src/main/java/com/swmansion/reanimated/nodes/Node.java
chrfalch: Function nodes should be memoized already - but we need a way to force a function that is called recursively to update even though the `mLastLoogID < mUpdateContrext.updateLoopID`.

osdnk: But why can’t we memoized recursively called node in every context? I mean that probably we can make memoization context bound to evaluation context. 
I did it i my pr so it’s definitely possible!

chrfalch: I tried to avoid propagating the evaluation context - we really don't have any - it is all handled by the param nodes and function call nodes. In the current implementation a function call is memoized correctly (it only overrides the `evaluate` function). What I need is a way to call a function node more than one time in the same update loop. Using a counter (`contextCount`) that is increased for each new function call we enter (and decreased when the function call is done) makes this all work - when you are in a function call, not outside function calls.

android/src/main/java/com/swmansion/reanimated/nodes/CallFuncNode.java
```diff
@@ -0,0 +1,46 @@
...
+            paramNode.endContext();
+        }
+        mNodesManager.updateContext.contextCount--;
+    }
+
+    @Override
+    protected Object evaluate() {
+        beginContext();
+        Node whatNode = mNodesManager.findNodeById(mWhatNodeID, Node.class);
+        Object retVal = whatNode.value();
```
osdnk: maybe we can pass context here? And then taker advantages of memoization per context?

chrfalch: I really didn't want an external evaluation context.... To try to avoid changing too much of the code - what is the evaluation context anyways? :) What I'm using is the param node that wraps another node - and for each function call it's inner node is switched to the id of the current parameter. This way things will be correct. We only need to ensure that calling the same function twice (recursive), the function should be evaluated each time.

android/src/main/java/com/swmansion/reanimated/UpdateContext.java
```diff
@@ -7,6 +7,7 @@
 public class UpdateContext {
 
   public long updateLoopID = 0;
+  public long contextCount = 0;
   public final ArrayList<Node> updatedNodes = new ArrayList<>();
```
osdnk: Shouldn’t `updatedNode` be also context-related?

chrfalch:
Why?

chrfalch:
Thanks for the thorough explanation. I think my implementation is a little bit different than what your are describing:

1) Register a param with the `param` function. 

This is a simple node that will be considered a child of the function node.

2) Register a function node using the `proc` function with the expression to be evaluated and the defined param from the previous step. This node attaches and creates the node instantly with its children.

The JS-call for creating a proc node (the `proc` function) will return a new js-function that "invokes" the previously defined proc node. When "invoking" the function node you are basically just returning the new function call node - this will be mounted in the view hierarchy - not the function node (the call function node just has a numeric reference to the function onde).

Does this make the implementation clearer?

osdnk:
Could you please draw or describe what will happen if we will have two views relying on one on a Function which accept one input node (one param)?

osdnk:
Ohhh, so `call` is a direct child of params' inputs. Nice idea! 

I wonder, if we can somehow manually drop this function from native side if it will not be used any longer and then attach it manually if needed? What do you think about it?

Also, how about this API:
```
proc((a, b) => multiply(a,b))
```
How to type it then?

src/core/AnimatedFunction.js
```diff
@@ -0,0 +1,28 @@
+import AnimatedNode from './AnimatedNode';
+import { createAnimatedCallFunc } from './AnimatedCallFunc';
+
+class AnimatedFunction extends AnimatedNode {
```
osdnk: How to make it more "reactive"? I'm a bit afraid about this part, if it's not forced to be in global scope. Maybe we can make it a hook:
```js
useEffect(() => {
  ...
  a.attach()
  return a.detach
}, [inputs] 
)
```

chrfalch: Great that you understand the idea :) The only nodes that will stay attached are the function nodes - they should not need to be detached, should they? 

I agree with the `proc((a, b) => multiply(a,b))` API - it looks super nice! I'll try to see if I can implement that one! Don't think typing will be to hard though.

chrfalch: I think you are right with the hook idea - it should definitely be used when functions are defined inside components to avoid memory leaks.

chrfalch: `function proc(cb: (...args: any[]) => AnimatedNode<any>): AnimatedNode<any>;`

chrfalch: Then we can ditch the `param` function from the js side :)

chrfalch: I've added the change in API you suggested, and also the `useProc` hook. I'm a little bit uncertain about how we can solve attaching functions without hooks. 

I would imagine that you'd often define a function in the constructor of a component or as a global in a Javascript file and are struggling a bit with how we can ensure that these are registered correctly.

chrfalch:
Look at the latest changes ;)

src/core/AnimatedFunction.js
apiko-bondarenko: @chrfalch according to what @terrysahaidak described as his use case:

```javascript
// Consider this as a meta code, describing a use case(or desired use case)
import { TextInput } from 'reanimated'

let myFunc = proc((x, y) => add(x, y)) 

const Input = (props) => { return ( <TextInput />) } 

Input.initAnimation = () => myFunc.__attach()

// Use this in a case like: do this on a screen A,
// but before going to the screen B, you would use Input.initAnimation().
// While screen be is importing and using this Input.
```
So I kind of wrote something like this:
```javascript
const test = (x, y) => multiply(x, y)
let myStuff = createAnimatedFunction(test)
// Somewhere
myStuff.__attach()
// Elsewere
myStuff.__detach()
```
And this kind of code should probably work because `AnimatedNode` has `__attach` and `__detach`. The only difference to attaching you've done is that should not require/force us to be within React component lifecycle to run this.

Question is: will they be registered as nodes via Reanimated properly. Otherwise good job and thank you!

chrfalch:
It would be great if we could have a chat Wednesday anyways - I have some questions about updates and processing of nodes.

src/core/AnimatedFunction.js
ddzirt: @chrfalch according to what @terrysahaidak described as his use case:
```javascript
// Consider this as a meta code, describing a use case(or desired use case)
import { TextInput } from 'reanimated'

let myFunc = proc((x, y) => add(x, y)) 

const Input = (props) => { return ( <TextInput />) } 

Input.initAnimation = () => myFunc.__attach()

// Use this in a case like: write your component on a screen A(or even elsewhere),
// but before going to the screen B, you would use Input.initAnimation().
// While screen be is importing and using this Input.
```
So I kind of wrote something like this:
```javascript
const test = (x, y) => multiply(x, y)
let myStuff = createAnimatedFunction(test)
// Somewhere
myStuff.__attach()
// Elsewere
myStuff.__detach()
```
And this kind of code should probably work because AnimatedNode has `__attach` and `__detach`. The only difference to attaching you've done in `useProc` is that it should not require/force us to be within React component lifecycle to do this.

Question is: will they be registered as nodes via Reanimated properly. Otherwise good job and thank you!

chrfalch: I've added some logic to the createAnimatedFunction (use the exported `proc` and you get it automatically) that attaches automatically when outside the `useProc` hook. When called using `useProc` hook attach/detach will be run automatically for you. It is therefore not recommended to store proc nodes as class members. Make them global instead:

```js
let myFunc

class MyComp extends React.Component {
  constructor(props){
    super(props);
    if(!myFunc) {
      myFunc = proc((a,b) => add(a,b);
    }
  }
}
```

terrysahaidak: Thank you @chrfalch. What looks really promising. The only change (feature request?) I would suggest is to add an ability to `await` proc attaching. My use case is the screen with a big form of identical inputs. And if I know the user is gonna navigate to that screen. I would initialize that proc and wait until it has been initialized and only then navigate to that specific screen, or just won't show the form for the user until. So the user will exactly know the app works as expected and there are no lags. The snippet above probably won't work because we might render 5 identical components at once, so there is a big chance to create that proc several times.

terrysahaidak: Oh, yes, since proc will return something synchronously, makes sense. It's time to take a break :D

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
```diff
@@ -0,0 +1,42 @@
...
+    public ParamNode(int nodeID, ReadableMap config, NodesManager nodesManager) {
+        super(nodeID, config, nodesManager);
+        mName = config.getString("name");
+        mArgsStack = new Stack<Integer>();
+    }
+
+    @Override
+    public void setValue(Object value) {
+        Node node = mNodesManager.findNodeById(mArgsStack.peek(), Node.class);
+        if(node != null) {
```
osdnk: What's the sense of this check?

chrfalch: Done in #b4157a7

chrfalch:
[ X] Please remove example from `test.js` and create another route for it.


> Also, please leave some docs regarding you changes (can be minimal, but at least few words).

> Add `prevPosition` as an optional argument to TS and mind it in docs (include that it's not advised to use unless you're using it as a param of function).

??

> Could you also add this component which we was discussing? I'm not fan of excluding ppl who don't want to use functional components :)

I'm still a bit reluctant to adding a component for the proc node - would rather explain to people that they should be added as constants - that will create clearer code than wrapping them in code. I'm even a bit reluctant to adding the `useProc` hook :)

osdnk:
ok, so let's drop `useProc`. 

https://github.com/kmagiera/react-native-reanimated/pull/354/files#diff-bf95de5d07bde5266d3ea7fd5ae52f90R50 <- I had to extract `prevPosition` from spring

osdnk:
Also, please add error that it's not allowed to set value (not param) inside function.

chrfalch:
Will do :)

chrfalch:
- Removed useProc
- Added missing key property to spring example in test.js
- Added documentation

osdnk:
So just an example and a warn and I hope we'll finish the last iteration 🏁

chrfalch:
> So just an example and a warn and I hope we'll finish the last iteration 🏁

The warning is added to the documentation:
"It is not possible to reference nodes that are not passed as parameters."

I still think that the spring example you made is good enough for now?

osdnk:
so at least move it another route with different name (not test)

chrfalch:
> so at least move it another route with different name (not test)

My take on the proc node is that it might not be for anyone to use - it is best suited for those not just developing regular apps with some animations using reanimated - they will probably not need it (it is just needed when repeatedly creating large node trees).

It is my opinion that use of the proc node should be in the reanimated library itself (or derived libraries) to build complex expressions like springs in a controlled way. This can in turn be used by app developers - therefore I believe that it shouldn't be necessary to include any examples for the proc node. 

The spring example you added to the `test.js` file is also something that we might want to refactor into a real spring function that will be used in ex. the `Animated.spring` class before giving it away to developers?
 
If you still want us to include an example, I'll do as you suggest to copy the test you created into an example - just wanted to raise the discussion :)

terrysahaidak:
> they will probably not need it (it is just needed when repeatedly creating large node trees).

I've tried using reanimated on android several times and every time it had failed because of the performance issue. Especially when you're trying to navigate to a specific screen with animation, or doing something else when the user can notice lags. Mostly it's because of creating lots of nodes. But with the ability to use

osdnk:
Thanks for this words, @chfalch.
Yes, we probably need to discuss it, but probably none of us has time resources for this discussion in this moment. Also, I don’t have currently strong opinion and need some time for thinking. I don’t want to block this PR bc of that.
Let’s discuss in another PR after merging this one, which satisfies our cases.

satya164:
I agree with @terrysahaidak. Currently everywhere I use reanimated, I have to implement some kind of memoization to avoid perf issues, and even then they aren't completely gone. This feature will help a lot with that.

osdnk:
I agree with everyone, but I want to exclude it from scope of this PR. Let’s discuss, but I’m afraid of blocking this change with discussions over usages of already workable and usable code :)

terrysahaidak:
@osdnk Let's merge it then and we will test it out :)

ddzirt:
Well, my current use case is a onboarding image transition between images, changing image and text while staying on the same screen and on reaching end triggering another set of animations. It has to run on timeout and with gestures. So I do wish to do this with reanimated and rngh, which I am attempting right now. Still, on Android it is beyond bad. Since I have at least 20 nodes that need init and running straight afterwards init and all that has to happen right after splash animations are done.

chrfalch:
I think I might have discovered a bug in the memoization code:

```js
const interpolate = proc((a) => add(a, 10));
const setValue = proc((a) => interpolate(a));

export default class Example extends Component {
render() {
    return (
      <View>        
        <Animated.Code>
          {() => block([
            call([
              setValue(10), 
              setValue(20)
            ], 
            (args) => console.log("value", args[0], args[1]))
          ])}
        </Animated.Code>
      </View>
    );
  }
}
```

**Expected output:** 20, 30
**Actual output:** 20, 20

This is probably happening since in the evaluation the interpolate proc is called twice, and since it wraps the same node id it will return the memoized value. Should this kind of `proc` be valid, @osdnk ?

src/core/AnimatedCallFunc.js
```diff
@@ -0,0 +1,20 @@
+import AnimatedNode from './AnimatedNode';
+import { adapt } from './AnimatedBlock';
+
+class AnimatedCallFunc extends AnimatedNode {
+  _value;
```
kmagiera: What is the being used for? seems that we only set it to -1 and never read

android/src/main/java/com/swmansion/reanimated/nodes/CallFuncNode.java
```diff
@@ -0,0 +1,48 @@
...
+    mParams = Utils.processIntArray(config.getArray("params"));
+    mArgs = Utils.processIntArray(config.getArray("args"));
+  }
+
+  private void beginContext() {
+    mPreviousCallID = mNodesManager.updateContext.callID;
+    mNodesManager.updateContext.callID = mWhatNodeID;
+    for (int i = 0; i < mParams.length; i++) {
+      int paramId = mParams[i];
+      ParamNode paramNode = mNodesManager.findNodeById(paramId, ParamNode.class);
```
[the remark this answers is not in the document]
chrfalch: Done

src/core/AnimatedCallFunc.js
chrfalch: The javascript implementation has not been updated since it is not in use.

src/derived/useProc.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { createAnimatedFunction } from '../core/AnimatedFunction';
+
+function useProc(cb) {
+  if (typeof React.useEffect === 'function') {
+    const proc = React.useState(() => createAnimatedFunction(cb));
```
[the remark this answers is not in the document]
chrfalch: useProc has been removed - functions should be declared as globals to avoid any memory leaks by redefining them in components.

android/src/main/java/com/swmansion/reanimated/nodes/CallFuncNode.java
```diff
@@ -0,0 +1,48 @@
+package com.swmansion.reanimated.nodes;
+
+import com.facebook.react.bridge.ReadableMap;
+import com.swmansion.reanimated.NodesManager;
+import com.swmansion.reanimated.Utils;
+
+public class CallFuncNode extends Node {
+
+  private int mPreviousCallID;
```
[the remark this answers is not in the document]
chrfalch: They are updated when calling functions - so I believe they should stay as they are? (Or maybe my Java skills are a bit outdated?)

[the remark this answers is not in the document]
chrfalch: Looks like @osdnk is fixing this!!

PR #362, opened 2019-08-07
[iOS] Fix floor() implementation

### LINKED ISSUE #316: floor(0) returns "-1"
Thank you for great library!

Today I got an issue, so please let me share.

```ts
const value = floor(
  interpolate(progress, {
    inputRange: [0, 1],
    outputRange: [0, 0],
    extrapolate: Extrapolate.CLAMP,
  }),
)
// value === -1
const value = round(
  interpolate(progress, {
    inputRange: [0, 1],
    outputRange: [0, 0],
    extrapolate: Extrapolate.CLAMP,
  }),
)
// value === 0
```

`value` should be `Animated.Node<0>` but will be `Animated.Node<-1>` somehow.
Is this the specification? or bug?
I'm sorry if this is expected result!

I'm using Expo SDK33.
iPhoneSE version 12.1.4
Expo Client.

yarn.lock
```
react-native-reanimated "1.0.1"
```

Thank you in advance!

wcandillon:
fixes #316

### DIFF
--- a/src/derived/floor.js
+++ b/src/derived/floor.js
@@ -1,5 +1,5 @@
 import { sub, round } from '../base';
 
 export default function floor(a) {
-  return round(sub(a, 0.5 - Number.MIN_VALUE));
+  return round(sub(a, 0.5 - 1e-8));
 }
--- a/src/derived/floor.js
+++ b/src/derived/floor.js
@@ -1,6 +1,5 @@
 import { sub, round } from '../base';
 
 export default function floor(a) {
-  const b = round(a);
-  return cond(lessThan(a, b), round(sub(a, 0.5)), b);
+  return round(sub(a, 0.5 - Number.MIN_VALUE));
 }
--- a/src/derived/floor.js
+++ b/src/derived/floor.js
@@ -1,5 +1,6 @@
 import { sub, round } from '../base';
 
 export default function floor(a) {
-  return round(sub(a, 0.5));
+  const b = round(a);
+  return cond(lessThan(a, b), round(sub(a, 0.5)), b);
 }


src/derived/floor.js
```diff
@@ -1,5 +1,6 @@
 import { sub, round } from '../base';
 
 export default function floor(a) {
-  return round(sub(a, 0.5));
+  const b = round(a);
+  return cond(lessThan(a, b), round(sub(a, 0.5)), b);
```
kmagiera: I think we don't need it to be this complex. I think `round(sub(a, 0.5 - Number.MIN_VALUE))` should do.

wcandillon: This suggested solution doesn't work, what is the strategy you are trying to use?

kmagiera: So the problem here is that -0.5 gets rounded to -1 and not to 0. If we subtract fewer than 0.5 from the value we calculate floor for, then it should get rounded in the correct direction. I'd try with sth like `0.5 - 1e-8` if `MIN_VALUE` doesn't work.

wcandillon: This solution is not working either

kmagiera: I checked the following implementation:
```
function floor(a) {
  return round(sub(a, 0.5 - 1e-8));
}
```

Output:
```
IN 1 OUT 1
IN 0.9 OUT 0
IN 0.1 OUT -0
IN 0 OUT -0
IN -0.1 OUT -1
IN -0.9 OUT -1
IN -1 OUT -1
```

And appears to be working fine. If possible I'd prefer to stick to a smaller number of nodes

wcandillon:
yes it works great sorry about that! I got confused because floor(0)
returns -0 in this implementation. But I'm ok with it!

wcandillon:
@kmagiera I update the solution to your implementation

src/derived/floor.js
```diff
@@ -1,5 +1,5 @@
 import { sub, round } from '../base';
 
 export default function floor(a) {
-  return round(sub(a, 0.5));
+  return round(sub(a, 0.5 - Number.MIN_VALUE));
```
kmagiera: I meant the solution with `0.5 - 1e-8` works ok, the one with MIN_VALUE doesn't

wcandillon:
both work for me in my use case but I can revert to  -1e-8

wcandillon:
just updated it

wcandillon:
This is ready to be merged. I use this implementation of floor in many projects now.

PR #389, opened 2019-09-17
Add createAnimatedComponent to mock

PierreCapo:
### Goal

Following https://github.com/kmagiera/react-native-reanimated/issues/355 , `createAnimatedComponent` is not present in `mock.js` file and makes jest test crashes.

### Issue Example
image


kmagiera:
Makes sense, thanks for sending this patch!

PR #390, opened 2019-09-18
Added web support

EvanBacon:
# Why

- [Preview](https://twitter.com/Baconbrix/status/1173825701192929280?s=20)
- [Demo](https://reanimated.netlify.com)

# How

- Enabled JS runner
- Added JS to broken runner nodes: `CallFunc`, `Param`
- Added support for web gesture-handler
- Added expo-web to Example
- Added `format` node for web color transform

# Test Plan

- [Example](https://reanimated.netlify.com) should run
  - Transitions API is not supported yet


src/core/AnimatedReformat.js
```diff
@@ -0,0 +1,19 @@
+import AnimatedNode from './AnimatedNode';
+import { adapt } from '../core/AnimatedBlock';
+import { val } from '../val';
+
+class AnimatedReformat extends AnimatedNode {
```
kmagiera: Can you add this to readme?

src/core/InternalAnimatedValue.js
```diff
@@ -44,7 +51,10 @@ export default class InternalAnimatedValue extends AnimatedNode {
     if (this.__inputNodes && this.__inputNodes.length) {
       this.__inputNodes.forEach(val);
     }
-    return this._value + this._offset;
+    if (this._offset && typeof this.value === 'number') {
```
kmagiera: did you mean `this._value` here instead?

src/core/InternalAnimatedValue.js
```diff
@@ -13,17 +13,24 @@ function sanitizeValue(value) {
...
-      val => (this.__nodeConfig.value = val)
-    );
+    if (ReanimatedModule.getValue) {
+      ReanimatedModule.getValue(
+        this.__nodeID,
+        val => (this.__nodeConfig.value = val)
+      );
+    } else {
+      setTimeout(() => {
+        this.__nodeConfig.value = this.__getValue();
```
kmagiera: not sure if this is needed at all, if so you'd likely need to reset offset to 0 here to. Also not sure if `setTimeout` is required. But in general I think no action is needed

src/core/InternalAnimatedValue.js
```diff
@@ -31,6 +31,8 @@ export default class InternalAnimatedValue extends AnimatedNode {
     );
   }
 
+  _offset = 0;
```
kmagiera: I don't think this is needed. Looks like the `_offset` param's been left here by mistake. We never assign to it, so maybe we can clean it up from `__onEvaluate` completely?

src/core/AnimatedReformat.js
kmagiera: Now that I think about this node I'd prefer no leave it out of the PR. The only use case here is to transform color value and we need a web specific code path for instantiating it anyways. So maybe we can create a web specific color node that formats input values into `rgba(r,b,g,a)` string in `onEvaluate` instead?

kmagiera: Alternatively we can consider using concat for web implementation there:
```
concat('rgba(', r, g, b, a, ')')
```

EvanBacon: I'm going to just delete color support and reformat since I don't like it either.

src/core/InternalAnimatedValue.js
EvanBacon: This needs to be here to make the `imperative` demo work. This is set when you call `setOffset`. It must be 0 for all other cases or you get NaN

jordmccord:
@EvanBacon, fantastic job as ever! Just out of curiosity what are the blockers for getting the Transition API to work?

Issue #381, opened 2019-09-05
RN 0.61: Animated.color causes "excessive number of pending callbacks"


felippepuhle:
Hey guys, how are you?

Just updated react native version to 0.61 (rc) and now my app is crashing on some screens that use color interpolation.

We have a theme based template, so I created a function that parses hex colors using polished:

```
import Animated from 'react-native-reanimated'
import { parseToRgb } from 'polished'

const interpolateColors = (animation, { inputRange, outputRange }) => {
  const colors = outputRange.map(parseToRgb)

  const r = Animated.interpolate(animation, {
    inputRange,
    outputRange: colors.map(c => c.red),
    extrapolate: Animated.Extrapolate.CLAMP,
  })

  const g = Animated.interpolate(animation, {
    inputRange,
    outputRange: colors.map(c => c.green),
    extrapolate: Animated.Extrapolate.CLAMP,
  })

  const b = Animated.interpolate(animation, {
    inputRange,
    outputRange: colors.map(c => c.blue),
    extrapolate: Animated.Extrapolate.CLAMP,
  })

  return Animated.color(r, g, b)
```

Any tips on this?

<img width="706" alt="Screen Shot 2019-09-05 at 11 41 22 AM" src="https://user-images.githubusercontent.com/15961929/64354054-193e2e80-cfd5-11e9-8f1d-9c44d2251042.png">

Thanks!

marcorm:
I get the same error in some conditions using react-navigation 5.0.0 (unstable)

rt2zz:
I have the same on: react-native@0.61-rc.3, react-navigation@4 and latest reanimated. 

I am not sure how to debug this, but will keep looking.

Jarred-Sumner:
i wonder if it's a TurboModules API change

Jarred-Sumner:
This commit looks relevant: https://github.com/facebook/react-native/commit/1e45a41dc2d4699d4e439994e81ca5ab82d630b3. This warning was added in 0.61.0.

Do you experience this issue on devices running in release mode? Or only development mode in the simulator? I currently only see it in development mode on the simulator.

My guess is: for most libraries, this warning message is a good way to find certain types of bugs. However, for this library, a lot of callbacks is sometimes expected behavior (e.g. stuff happening after a pan gesture recognizer updates coordinates)...and everything is just slower in development mode.

felippepuhle:
You're right!

[...]

PR #398, opened 2019-09-24
Change drop procedure to rely on events instead of callbacks.

kmagiera:
Here we change the way we handled nodes dropped on the native side. Since we don't know at the point of ropping a native node if it is going to be kept on the javascript side and reused later on, we have to copy the last seen value on the native side to be used later in case the node gets recreated on the native side. Previously we've been using getValue call. But because of an excessive amount of RN callback's that this way would produce in some cases we had to switch to using events instead (RN 61 added a check for a large number of callbacks being registered).

In order to test it I recommend checking "Imperative set value / toggle visibility" example app. There we can change the value of a node that exists on native and then detach component causing the node to be dropped from native but retained in javascript. Then we can render back the component and expect the reatached node to have the updated value.

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/nodes/Node.java
+++ b/android/src/main/java/com/swmansion/reanimated/nodes/Node.java
@@ -100,6 +100,9 @@ public abstract class Node {
     markUpdated();
   }
 
+  public void onDrop() {
+  }
+
   private static void findAndUpdateNodes(Node node, Set<Node> visitedNodes, Stack<FinalNode> finalNodes) {
     if (visitedNodes.contains(node)) {
       return;
--- a/src/core/InternalAnimatedValue.js
+++ b/src/core/InternalAnimatedValue.js
@@ -1,6 +1,35 @@
 import AnimatedNode from './AnimatedNode';
 import { val } from '../val';
-import ReanimatedModule from '../ReanimatedModule';
+import ReanimatedEventEmitter from '../ReanimatedEventEmitter';
+
+const DETACHED_NODE_MAPPING = new Map();
+let DETACHED_LISTENER_SETUP = false;
+
+/**
+ * This method registers a detached nodes in a temporary map and waits for
+ * onReanimatedValueDropped event to be triggered. Once the event happens it
+ * updates the dropped node's condif with the value provided from native.
+ * This is needed for the nodes that are being reused at later times and mounted
+ * as dependencies with some other components. Before this method's been
+ * introduced we were relying on `getValue` call. However due to a large volume
+ * of RN callback's this method yielded we had to update this code to rely on
+ * events instead. Ideally we should find a way to tell if a given node is going
+ * to be reused or not (most of the nodes are not being reused). Perhaps
+ * a concept of `TempValue` would be a good enough solution?
+ */
+function updateConfigValueOnDetach(node) {
+  if (!DETACHED_LISTENER_SETUP) {
+    ReanimatedEventEmitter.addListener('onReanimatedValueDropped', data => {
+      const node = DETACHED_NODE_MAPPING.get(data.id);
+      if (node) {
+        node.__nodeConfig.value = data.value;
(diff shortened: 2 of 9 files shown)


kmagiera:
This fixes #381 but after recent change in RN that changes the console.error to warnOnce when callbacks count is exceeded I am no longer sure I want to land this change. Maybe we can explore the idea of adding "TempValue" a bit more?

android/src/main/java/com/swmansion/reanimated/nodes/Node.java
```diff
@@ -100,6 +100,9 @@ protected final void forceUpdateMemoizedValue(Object value) {
     markUpdated();
   }
 
+  public void onDrop() {
```
osdnk: I'm not happy with this approach. It's rather pointless to create a method and override it in exactly one place. If I were you, I'd rather use `instanceof` and `isKindOfClass`.
I'm rather a fan of changing the smallest possible number of files / classes.

src/core/InternalAnimatedValue.js
```diff
@@ -1,6 +1,35 @@
 import AnimatedNode from './AnimatedNode';
 import { val } from '../val';
-import ReanimatedModule from '../ReanimatedModule';
+import ReanimatedEventEmitter from '../ReanimatedEventEmitter';
+
+const DETACHED_NODE_MAPPING = new Map();
+let DETACHED_LISTENER_SETUP = false;
+
+/**
+ * This method registers a detached nodes in a temporary map and waits for
```
osdnk:
```suggestion
 * This method registers detached nodes in a temporary map and waits for
```

src/core/InternalAnimatedValue.js
```diff
@@ -1,6 +1,35 @@
...
+ * onReanimatedValueDropped event to be triggered. Once the event happens it
+ * updates the dropped node's condif with the value provided from native.
+ * This is needed for the nodes that are being reused at later times and mounted
+ * as dependencies with some other components. Before this method's been
+ * introduced we were relying on `getValue` call. However due to a large volume
```
osdnk:
```suggestion
 * introduced we were relying on `getValue` call. However, due to a large volume
```

src/core/InternalAnimatedValue.js
```diff
@@ -1,6 +1,35 @@
...
+ * of RN callback's this method yielded we had to update this code to rely on
+ * events instead. Ideally we should find a way to tell if a given node is going
+ * to be reused or not (most of the nodes are not being reused). Perhaps
```
osdnk:
```suggestion
 * to be reused or not (most of the nodes are not being reused). Perhaps,
```

src/core/InternalAnimatedValue.js
```diff
@@ -1,6 +1,35 @@
...
-import ReanimatedModule from '../ReanimatedModule';
+import ReanimatedEventEmitter from '../ReanimatedEventEmitter';
+
+const DETACHED_NODE_MAPPING = new Map();
+let DETACHED_LISTENER_SETUP = false;
+
+/**
+ * This method registers a detached nodes in a temporary map and waits for
+ * onReanimatedValueDropped event to be triggered. Once the event happens it
+ * updates the dropped node's condif with the value provided from native.
```
osdnk: condif?

Jarred-Sumner:
fwiw, I have noticed occasional crashes shortly after the error triggers. The breakpoint points to some C++ code for the JS bridge call.

Still only in development, however.

osdnk:
@Jarred-Sumner I have also seen, but probably it will fix this as well

kmagiera:
I think I'm going to close this one if favor of #403 – that other PR does not really fix the issue but should put the volume of callbacks into a reasonable range

Jarred-Sumner:
I’m on master and it does seem better!

I still hit the excessive number of callback issues occasionally, however
there are lots of animations and the error seems to mostly trigger during
Fast Refresh when the app is already slow

PR #401, opened 2019-09-25
Update Example app to use RN 0.61

kmagiera:


kmagiera:
Forgot to update ios project 🙀

PR #403, opened 2019-09-25
Optimize constant nodes creation and cleanup

kmagiera:
This PR optimizes the way we create and cleanup value nodes.

When adapting constants (e.g. when doing `interpolate({ inputRange: [0,1]})` we generate new nodes for start and end of the range) we've been creating new animated value nodes. This kind of nodes can never be updated on the native side and hence there is no need to send its final value back to JS when they get detached.

On top of there there are several constant values that are frequently used across the library and product code with reanimated. Those can only be instantiated once and reused in all the places. That list includes `0`, `1` and `-1` for now.

These two changes reduce the number of nodes instantiated by one third on `colors/index.js` example (from 88 down to 57) and reduce the number of nodes we call `getValue` for by 94% (from 88 down to 5).

### DIFF
--- a/src/core/InternalAnimatedValue.js
+++ b/src/core/InternalAnimatedValue.js
@@ -14,7 +14,7 @@ function initializeConstantValues() {
   if (CONSTANT_VALUES.size != 0) {
     return;
   }
-  [0, -1, 1].forEach(v =>
+  [0, -1, 1, -2, 2].forEach(v =>
     CONSTANT_VALUES.set(v, new InternalAnimatedValue(v, true))
   );
 }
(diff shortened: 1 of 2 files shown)


src/core/InternalAnimatedValue.js
```diff
@@ -8,22 +8,43 @@ function sanitizeValue(value) {
     : Number(value);
 }
 
+const CONSTANT_VALUES = new Map();
+
+function initializeConstantValues() {
+  if (CONSTANT_VALUES.size != 0) {
+    return;
+  }
+  [0, -1, 1].forEach(v =>
```
osdnk: [](url)
```suggestion
  [0, -1, 1, -2, 2].forEach(v =>
```

ShaMan123:
This PR seems to cause significant problems
#417

PR #237, opened 2019-04-09
Add Animated.useCode hook as an alternative to Animated.Code

jeremy-deutsch:
I'm not sure about the current stance on adding hooks to `react-native-reanimated`, but this adds an `Animated.useCode` hook with an API halfway between `useEffect` and `Animated.Code`, that serves as an alternative to `Animated.Code`. Instead of passing in a callback, the user passes in a reanimated node to execute, along with an array of dependencies. Since we use `useEffect` under the hood, any change in the dependency array also causes the node to be updated.

One piece of behavior that I'm still iffy on is what happens if a dependency array isn't passed in. The idea is that one is always passed in (the typings reflect this), but right now, not passing one in causes the node to be re-run on every render. I think this is expected enough that we can leave it as is.

Also, this is kind of obvious, but `Animated.useCode` is a no-op on a version of RN without hooks.

PR #408, opened 2019-09-26
🎣Allow to pass a callback as parameter of useCode()

### LINKED ISSUE #343: Should useCode use a callback as first argument?
I'm asking because this is the first time I see a hook with the `deps` parameter where the first parameter is not a callback. In fact, the eslint-hook plugin won't lint for exhaustive dependencies when the first argument is not a callback: https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L64

And indeed now it feels a bit strange that node expression is always "evaluated" for each render even though it should only have effects depending on the `deps` parameter. What are your thoughts on this?

wcandillon:
fixes #343 

This can enable us to save some unnecessary node creation but more importantly, it will enable us to benefit from the Exhaustive Deps eslint rule. This rule expect a callback as first parameter: https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L64
This rule is incredibly useful to avoid bugs when using hooks that have dependencies.

@osdnk what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,16 +224,18 @@ block([
 
 The `useCode` hook acts as an alternative to the `Animated.Code` component.
 ```js
-Animated.useCode(node, deps)
+Animated.useCode(() => node, deps)
 ```
 It's passed an animated node and an array of dependencies, and updates that node both when the component mounts and every time a value in that array changes. It does nothing on versions of React Native that don't support hooks (<0.59).
 ```js
 const [offset, setOffset] = React.useState(20);
 Animated.useCode(
-  set(transX1, add(_transX, offset)),
+  () => set(transX1, add(_transX, offset)),
   [offset]
 );
 ```
+
+We recommend to use `useCode()` with the `react-hooks/exhaustive-deps` [eslint rule](https://www.npmjs.com/package/eslint-plugin-react-hooks).
 
 ## Event handling with reanimated nodes
 
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -329,7 +329,7 @@ declare module 'react-native-reanimated' {
 
     // hooks
     export function useCode(
-      exec: AnimatedNode<number>,
+      exec: AnimatedNode<number> | (() => AnimatedNode<number>),
       deps: Array<any>,
     ): void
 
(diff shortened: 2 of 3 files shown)


kmagiera:
My comment on #343 

> Hm, now that I think of it I'd say it should always be a callback and never just a list of nodes. Would it make sense to make a breaking change here and change the behavior?

kmagiera:
Maybe @jeremy-deutsch who authored the hook originally in #237 have an opinion here?

kmagiera:
I'd also like for the example usage in README to be updated

jeremy-deutsch:
Thanks for tagging me! I have some unorganized thoughts:

After looking this problem over, my brain actually first jumped to an API like the following:
```js
const [offset, setOffset] = React.useState(20);
const offsetNode = useMemo(
  () => set(transX1, add(_transX, offset)),
  [offset]
);
Animated.useCode(offsetNode); // doesn't reattach if offsetNode doesn't change
```
On first glance, that seems more in line with what hooks APIs are trending towards (and makes use of the eslint plugin). It wouldn't be a breaking change, either - just make the second argument optional and add `exec` as a dependency of the attach effect. The problem is that now nodes get reattached on every render if they aren't wrapped in `useMemo`. And that basically means that users *always* have to wrap their nodes in `useMemo`.

All of this is to say that I think the callback/`useEffect`-style API is probably best, but that we might want to call `useMemo` internally to avoid re-creating nodes unnecessarily.

Also, mixed feelings on having a breaking change. I think the callback API is better enough that at the very least the non-callback API should be deprecated with a warning (if not removed altogether).

wcandillon:
Thank you for the feedback. I implemented the review comments. Just on a side note, `useMemo` doesn't guarantee that the identity would be preserved across time. I use `useMemoOne()` but there are other custom solutions available.

likern:
I think breaking change is totally Ok, if it will be major release update.
What about collect multiple ideas / breaking changes into one major release?

osdnk:
@likern 
It's not breaking for now. I think people will be encouraged to do it by warning and we will make it breaking in next major

PR #439, opened 2019-10-29
fix: mock.js NOOP to allow for "new" constructor call

ajsmth:
i'm getting errors in my tests from this mock as the `new Value(0)` call throws for arrow functions, this seems to work when i updated my mock.js in node_modules.


ajsmth:
note -- this can also be fixed w/ babel by adding `react-native-reanimated` to transformIgnorePatterns jest config

PR #445, opened 2019-10-30
Fix web support

osdnk:


Example/Animated
```diff
@@ -0,0 +1 @@
+../src
```
osdnk: This is symlink for resolving reanimated

PR #458, opened 2019-11-09
Codegen

osdnk:
# Codegen for RNR

We still struggle with the problem of a great number of nodes creating while attaching in React Navigation and Bottom Sheet libraries. I decided to move some more complicated derived nodes to the native side instead of building them from them with primitives. When I started implementing some of them, I figured out that my work is quite mechanical so I wanted to automate this.
However, creating a generic mechanism can help not only with this issue but also can be reused by other developers if we will make it reliable enough. 
ezgif com-video-to-gif (2)

Actually I realized that it's not that difficult to build native derived node:

```objc
@interface REACODEGENAbsNode : REANode
@end

@implementation REACODEGENAbsNode {
  NSArray<NSNumber *> *_input;
}

- (instancetype)initWithID:(REANodeID)nodeID config:(NSDictionary<NSString *,id> *)config
{
  if ((self = [super initWithID:nodeID config:config])) {
    _input = config[@"input"];
  }
  return self;
}

- (id)evaluate
{
  if ([[[self.nodesManager findNodeByID:_input[0]] value] doubleValue] < 0) {
    return @(-1 * [[[self.nodesManager findNodeByID:_input[0]] value] doubleValue]);
  }
  return [[self.nodesManager findNodeByID:_input[0]] value];

}
```
from 

```js
import { cond, lessThan, multiply } from '../base';

export default function abs(a) {
  return cond(lessThan(a, 0), multiply(-1, a), a);
}
```

And I did it:
```bash
yarn codegen
```

Now it supports a very limited set of features, creates lines in files checked to git but it somehow works and command above is the only thing which has to be run.
Also, there's a number of other problems which are not solved not but at least I proved that it's possible. 

TODO: 
- [ ] handle all primitives
- [ ] constructing nodes from other derived nodes
- [ ] make operators more generic
- [ ] move codegened things to files which will be eventually added to .gitignore
- [ ] the problem of unique naming of nodes
- [ ] android 
- [ ] support extra variables defined in function's body
- [ ] expose to other consumers (outside lib)
- [ ] figure out how to connect it with compiling
- [ ] OTA safety


chrfalch:
Hi @osdnk! I was thinking and playing around a bit with this idea - can this be solved by creating a node that could run a pure js function totally out of context and return it's value on evaluate, like this?

```js
const a = new Animated.Value(10);
const b = new Animated.Value(10);
const jsNode = Animated.js("function (a, b) { return a + b }", a, b);
```

This way we can write pure functions that can operate on input nodes - whatever logic we wanted would be available as long as it does not have side effects. JS could should be evaluated directly (not through the bridge) and return. This is fast and maybe gives us the room needed for customization?

I've made a test on iOS that works. Still (as usual) having some issues finding out a way to solve the same thing under Android, but I think it should be possible (trying to find a way to load either Hermes og JSC)

<https://github.com/software-mansion/react-native-reanimated/compare/master...chrfalch:Feature/JSExecNodes?expand=1>

osdnk:
Hey, thanks!
I'm not convinced of codegen tbh. Your approach looks promising :)

chrfalch:
Would you like us to provide a PR with this implementation, or should we discuss how we can solve running js outside the bridge on android and avoid using v8?

osdnk:
Sure, I think it’s awesome

satya164:
This is so awesome. Why am I seeing this now 😍

@chrfalch I'd propose a minor change to the API you proposed:

```js
const buildNode = js`
  function(a, b) {
    return a + b;
  }
`;

const jsNode = buildNode(a, 120);
```

Why tagged template literal? Because then it'll be possible to add syntax highlight to this code snippet like styled-components does for `css` tag :)

chrfalch:
You can see and test a working implementation in a separate PR (#479) - the idea of using tagged templates sounds interesting - I’m not aware of how this is done -  it it should be doable and looks good.

jakub-gonet:
This PR is a bit old, any updates?

jakub-gonet:
Closing for now

PR #148, opened 2018-12-11
fix build on android

Krosh:
This commit fix error "Could not resolve all artifacts for configuration ':react-native-reanimated:classpath'". I just up grade version and add google repo

PR #463, opened 2019-11-12
Fix Android Gradle setup

friederbluemle:
Trying to open or build the `android/` folder was not possible, due to a missing `buildscript` block:

```
ERROR: Plugin with id 'com.android.library' not found.
```

After adding the reference to the source of the `library` plugin, the following error showed up:

```
ERROR: Failed to resolve: androidx.transition:transition:1.1.0
```

This was caused by missing entries in the `repositories` block. Added the necessary references to resolve the errors.

It is now possible to individually open and build the `android/` folder in Android Studio (or build it on command line, after generating a Gradle wrapper).

Note that these changes will not affect consumer (app) projects - The Android Gradle plugin is included _conditionally_, meaning the dependency classpath _only_ gets added when the project is opened standalone. This is better than simply removing it (as seen in #148)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,12 +2,27 @@ def safeExtGet(prop, fallback) {
     rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
 }
 
+buildscript {
+    if (project == rootProject) {
+        // The Android Gradle plugin is only required when opening the android folder stand-alone.
+        // This avoids unnecessary downloads and potential conflicts when the library is included as a
+        // module dependency in an application project.
+        repositories {
+            google()
+            jcenter()
+        }
+        dependencies {
+            classpath 'com.android.tools.build:gradle:3.5.2'
+        }
+    }
+}
+
 apply plugin: 'com.android.library'
 apply plugin: 'maven'
 
 android {
     compileSdkVersion safeExtGet('compileSdkVersion', 28)
-
+    buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
     defaultConfig {
         minSdkVersion safeExtGet('minSdkVersion', 16)
         targetSdkVersion safeExtGet('targetSdkVersion', 28)
@@ -20,17 +35,22 @@ android {
 }
 
 repositories {
+    mavenLocal()
     maven {
         // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
-        // Matches the RN Hello World template
-        // https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/local-cli/templates/HelloWorld/android/build.gradle#L21
-        url "$projectDir/../node_modules/react-native/android"
+        url "$rootDir/../node_modules/react-native/android"
     }
-    mavenCentral()
+    maven {
+        // Android JSC is installed from npm
+        url "$rootDir/../node_modules/jsc-android/dist"
+    }
+    google()
+    jcenter()
 }
 
 dependencies {
-    implementation 'com.facebook.react:react-native:+'
+    //noinspection GradleDynamicVersion
+    implementation 'com.facebook.react:react-native:+'  // From node_modules
     implementation "androidx.transition:transition:1.1.0"
 }
 


friederbluemle:
@osdnk - Please take a look if you have a moment. Thanks!

SaeedZhiany:
@friederbluemle 

Thanks

I just took a look at the android project folder, there is no gradle wrapper folder, is it necessary to be added too?

friederbluemle:
A matching Gradle wrapper is generated _automatically_ by Android Studio when you open the folder (you will notice Android Studio prompting you). Since the primary use case of a library is to be included in other projects, the Gradle wrapper should _not_ be packaged with it (and also does not have to be in source control). The repo is already set up like this, so I did not add it here.

Alternatively, you can also generate a Gradle wrapper on command line by running:

```
gradle wrapper --gradle-version 5.4.1 --distribution-type all
```

SaeedZhiany:
Thanks for your information, I have seen some repositories that have such a folder. Do you think it's a good idea to submit PR on that repositories and remove the folder from the repositories?

friederbluemle:
Yes, many library project repos have the folder you mentioned in source control. For an _application_ project, the Gradle wrapper should _definitely_ be committed to source control - for a _library_ project there is no "100% correct answer" - Some people like to have it, some don't. What's more important (and less controversial) is that all Gradle wrapper files need to be _excluded from distribution_. That is done using an appropriate combination of `.npmignore`/`.gitignore` and `files` declaration in package.json.
The setup `react-native-reanimated` chooses is not to have Gradle wrapper for the library folder in source control (which is also my preference).

SaeedZhiany:
I get it, Thanks

SaeedZhiany:
I got it, Thanks

friederbluemle:
Rebased onto latest `master` branch.

osdnk:
Hi, @friederbluemle 
Why do you need this?

friederbluemle:
Hi @osdnk - The quick version is: To fix the errors you can see above. The longer version is: Without it, it's an incomplete project setup. Gradle/Android Studio does not know where to get the plugin from (when opened standalone - it will only work when opened _together_ with an app project). Why would you want to open it standalone? Quite simply, it allows the project to be compiled and verified without any extra setup and files. It will be easily possible to write **unit tests** that only operate on the library code. You will also be able to **lint** the code in isolation (catching many errors that would otherwise go unnoticed).
Note that as I said above, this will have _no_ effect on app consumer projects.

osdnk:
None of this cases is reasonable in Reanimated so far, but I’m fine with merging it if you really need

android/build.gradle
```diff
@@ -20,17 +35,22 @@ android {
...
-        // Matches the RN Hello World template
-        // https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/local-cli/templates/HelloWorld/android/build.gradle#L21
-        url "$projectDir/../node_modules/react-native/android"
+        url "$rootDir/../node_modules/react-native/android"
+    }
+    maven {
+        // Android JSC is installed from npm
+        url "$rootDir/../node_modules/jsc-android/dist"
     }
-    mavenCentral()
```
osdnk: I’m not super familiar with these configs, but are you sure it won’t affect anything?

friederbluemle: It will not affect anything. `mavenCentral` is the "old" entry which has been replaced by `jcenter` (which is a superset of `mavenCentral`).

friederbluemle:
I think being able to write standalone unit tests and being able to easily line, compile and test the code in isolation is something that _any_ repo would benefit from. I hope you agree, and since this won't have any negative effect on app projects, it would be great if you could merge. 🙏

android/build.gradle
```diff
@@ -20,17 +35,22 @@ android {
 }
 
 repositories {
+    mavenLocal()
```
osdnk: Why this?

android/build.gradle
```diff
@@ -20,17 +35,22 @@ android {
...
     maven {
         // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
-        // Matches the RN Hello World template
-        // https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/local-cli/templates/HelloWorld/android/build.gradle#L21
-        url "$projectDir/../node_modules/react-native/android"
+        url "$rootDir/../node_modules/react-native/android"
```
osdnk: And this?

friederbluemle: It's taken directly from the main `react-native` repository. I did not make any additional changes to the repos listed here. It's also the way plenty of other react native library repos are set up. If you have a different opinion, please let me know and I'll change it.

osdnk:
Btw, you mentioned some tests and linting so I assume you use them, right? Do you want to include them as well?

friederbluemle:
> Btw, you mentioned some tests and linting so I assume you use them, right? Do you want to include them as well?

Yes, good point. I do have some experimental setup (across many locally cloned RN library repos). It's not at a point for public use yet, but I'll keep this repo in mind for potential future integration. As a prerequisite, the changes in this PR are definitely required. Thanks!

osdnk:
So maybe let’s wait until you find these tests ready? :)

Issue #430, opened 2019-10-21
How to respond to events like TextInput focus/blur and Keyboard show/hide


nhuesmann:
Hello. I love the concept of reanimated, and understand how to implement some basic animations using it. However, I am stuck on how to respond to events that require communication over the bridge. I have 2 primary use cases I'm currently trying to solve for:

1. Run an animation when a TextInput is focused, run the animation in reverse when the TextInput is blurred.
2. Run an animation when the Keyboard is shown, run the animation in reverse when the Keyboard is hidden.

I've read the documentation a few times and have checked out most of the examples, I couldn't find a super clear solution for these requirements. I asked a more specific version of this question on [Stack Overflow](https://stackoverflow.com/questions/58418470/how-to-react-to-keyboard-events-in-react-native-reanimated) that includes a runnable example, but I know the implementation is wrong. I just am wondering what the recommended approach is for reacting to user-driven events other than touch/gestures (i.e. without using gesture handler lib and just hooking into events with the `event` function). Is this even a good use case for reanimated? Is it possible to achieve this and achieve 60 fps animations?

gringocl:
Check out [`redash`](https://github.com/wcandillon/react-native-redash#usetoggletoggle-boolean-duration-number-easing-easingfunction-node)

nhuesmann:
> Check out [`redash`](https://github.com/wcandillon/react-native-redash#usetoggletoggle-boolean-duration-number-easing-easingfunction-node)

I actually have, and I corresponded with @wcandillon some over this. I just wanted to know if anyone who has contributed to this library had any suggestions about the recommended way of going about this. I just feel like it goes against the model of the library, as reacting to input/keyboards state relies on tracking some state in the JS thread, when reanimated's whole core value prop comes from sending the animation config to the native thread upfront to avoid the costs of crossing the bridge. @kmagiera any resources you or any other lib contributors have would be awesome :)

euroclydon37:
As long as business logic lives in JS, I don't think there's a way to _completely_ remove the need for the bridge in relation to animations. What I've done is make use of the `useCode` hook.

```js
const [focused, setFocused] = useState(false)

useCode(
  cond(
    focused,
    [/* reanimated nodes if true */],
    [/* reanimated nodes if false */]
  ),
  [focused]
)
```

nhuesmann:
[...]

Issue #437, opened 2019-10-29
[Enhancement] proc + NativeModules 


ShaMan123:
After playing around with `proc` I came up with an idea.
I'm after interacting with a native module without going through the bridge (who isn't??).
This is mainly the goal of `Animated` and especially of `reanimated` and is **the** `react-native` turn off.

So now that `proc` is live I'm thinking of something as follows:

```ts
type NativeMethodProc = (
   moduleName: string, 
   (methodName : string) | (viewManagerCommand: number), 
   ...args: Animated.Value | Animated.Proc
) => Animated.Proc

const methodProc = nativeMethodProc('myCustomModule', 'someNativeMethod', ...(procs | values))

const myVal = new Value(0)
useCode(set(myVal, methodProc(a, b, procSuccess(a1, a2), procFailure(a1))))
```

The procs passed as args will be in charge of consuming the `callback` args (onSuccess/onFailure?) that every `ReactMethod` has. The only thing I'm thinking of are `ReadableMaps` that are passed back to the `callback` - I'm guessing there could be an option to reuse `event` for this purpose.

This approach is valid for `dispatchViewManagerCommand` as well.
It unlocks endless options (e.g. native state management), complements `createAnimatedComponent` and sounds to me pretty revolutionary.

### A simple use case:
A touch event has started.
I need the absolute layout of a view for some animation calculation.
I need to call `UIManager.measure` and then set values.
This is a major bottle neck to the UI.

### Another use case:
I need to get a value returned for a native module method during a pan gesture event (e.g. if touches are hitting an SVG path)
I'm calling the method from js, so it seems unresponsive.

### Ultimate use case:
let's say I'm dragging a view in a `ScrollView` and want to scroll. With `NativeMethodProc` I can simply add the proc to my `block` and scroll freely while panning.

ShaMan123:
#354 
@chrfalch mentioning you as you are the author of the PR - wanted to hear your thoughts

ShaMan123:
```java
package com.swmansion.reanimated.nodes;

import android.view.View;

import com.facebook.react.bridge.ReactContext;
import com.facebook.react.bridge.ReadableNativeArray;
import com.facebook.react.uimanager.NativeViewHierarchyManager;
import com.facebook.react.uimanager.UIBlock;
import com.facebook.react.uimanager.UIManagerModule;

public class ModuleFuncNode {
    interface MethodInvoker {
        public void invoke(int tag, ReadableNativeArray array);
    }

[...]

Issue #375, opened 2019-08-23
event() handler skips State.ACTIVE when used with <TapGestureHandler />


vpuonti:
Title.

Example code:

```ts

function interaction(tapState: Animated.Value<number>) {
  return cond(
    eq(tapState, State.ACTIVE),
    [debug('active', tapState)],
    [debug('inactive', tapState)]
  )
}

class Example extends React.Component {
  private tapState: Animated.Value<number> = new Value(State.UNDETERMINED)

  public render() {
    return (
      <TapGestureHandler
        onHandlerStateChange={event([
          {
            nativeEvent: {
              state: this.tapState,
            },
          },
        ])}
      >
        <Animated.View style={{ flex: 1 }}>
          <Animated.Code exec={interaction(this.tapState)} />
        </Animated.View>
      </TapGestureHandler>
    )
  }
}
```

Output to console.

```
inactive 0
inactive 2
inactive 5
inactive 2
inactive 5
inactive 2
inactive 5
inactive 2
inactive 5

```

osdnk:
In my opinion, it might be considered as an expected behavior since it's a discrete gesture and a quick gesture state's changes in UIKit might not be sent to JS.

alikazemkhanloo:
Hey, I can't handle ACTIVE state, how should I handle it?
is this issue resolved? I'm using version 1.7.0
if I use a function instead of `event()`, I can log ACTIVE state. but does this method cause performance issue calling the js thread to handle tap?

also I tried to use the state inside my code not even calling JS. but I couldn't see the result of my code in action. maybe sth is wrong with my code, but I couldn't find out anything wrong, I thought this issue might be relevant.

PR #317, opened 2019-06-19
Map node

gabrielbull:
@osdnk @kmagiera I need your help to get this working.

So the idea is to be able to animate ScrollView props like `scrollIndicatorInsets` or other such properties.

For this, I created a Map node. I don't know if this is what you guys would have in mind for such a use case so please let me know if not.

If you guys are okay with the map node, I'd need a little help making it work. I have it somewhat working right now as it will make the initial value work but will not update if I change the value.

Here is a quick example:

```jsx
import React, { useEffect, useMemo } from 'react'
import { ScrollView, View } from 'react-native'
import Animated from 'react-native-reanimated'
import AnimatedMap, {
  createOrReuseMapNode
} from 'react-native-reanimated/src/core/AnimatedMap'

const { createAnimatedComponent, Value } = Animated

const AnimatedScrollView = createAnimatedComponent(ScrollView)

export const App: React.FC<{}> = () => {
  const top = useMemo(() => {
    // This one works, the scroll indicator inset is 100
    return new Value<number>(100)
  }, [])

  useEffect(() => {
    setTimeout(() => {
      // This one does not work, the scroll indicator inset remains 100
      top.setValue(300)
    }, 2000)
  }, [])

  return (
    <AnimatedScrollView
      scrollIndicatorInsets={createOrReuseMapNode({
        top
      })}
    >
      <View style={{ height: 1000 }} />
    </AnimatedScrollView>
  )
}
```

PR #469, opened 2019-11-12
Direct Manipulation Nodes #2

ShaMan123:
continuing #462 

**EDITED**

#### Interacting with a native module without going through the bridge. #437 
1. Assume we have a node `A` that evaluates in response to an `event`.
1. Once it evaluates we need to calculate something in a native module from a third party. 
1. The result is then used for other nodes and so on.
1. This must recur on every touch.

The only way it can be done is with `call`, additional `js` and `setValue`.
1. node `A` evaluates
1. `call` evaluates -> **over the bridge** to `js`
1. calling native module from `js` -> back and forth **over the bridge**
1. `setValue` on target node -> **over the bridge** to native

It doesn't make sense to **cross the bridge 4 times for a single touch**, it's not responsive and is off with what `reanimated` is all about.

Another important motivator: `react-native` has a vast variety of third party modules. If there would be a way to use `reanimated` with them it would turn `reanimated` into a hyper module. I believe this can change stuff dramatically, making a lot of `js` redundant (even some state management). 

After a few times the need rose I decided to give it a shot, and here you have it.
Unfortunately I don't have a mac so I couldn't work on `iOS` code (not so fluent either), so apart from that all is set up.
I'm hoping someone will take it on (here to help) or for a mac :).

### In Depth
I've introduced 3 new nodes `dispatch`, `invoke` & `intercept` and forked 2 nodes `map` & `callback` from `event`.
`dispatch` dispatches a view manager command.
`invoke` calls a react method on a native module.
`intercept` intercepts a global event before being sent across the bridge.
`map` and `callback` are exposed to pass as params for `dispatch`, `invoke` and `intercept` and are based on `event` (I've extracted common logic from `event` to `map`).

The logic behind it all is as follows:
1. All communications over the bridge accept primitives (all supported by `reanimated`), `ReadableArray`, `ReadableMap`, `Callback` and `Promise`. The last four are handled by `map` and `callback` respectfully. `callback` is identical to `event` in use and functionality (not in logic of course).
1. `react-native` is basically the interface between all these native modules so, like `event`, this approach walked through an open door.
1. Using reflection to access the modules.

### Usage

The ultimate use case is **scrolling** a `ScrollView`. There wasn't a way to control scrolling with animation. 
Now there is:
```js
dispatch('RCTScrollView', 'scrollTo', tag, scrollX, scrollY, animated)
```

I've added a few use cases including this one to the example app under `Direct Manipulation Nodes`.
Give it a shot, it's great and is exactly what I was missing (and I believe that many others too).

```js
const measureView = proc((tag, cb) => cond(defined(tag), invoke('UIManager', 'measure', tag, cb)));
```

### Additional

1. I've exposed a component `DirectManipulationHelper`. It displays all available commands and methods for direct manipulation including events intercepted at runtime, with the exception of view managers that have no commands map. If the commands map is not present the only way to get the commands is in the docs/source files.

1. In addition I had to know when `Transition` has ended a pass (for layout calculations) so I've exposed a way to track the state of the transition (via event and callback) #269  
**EDIT**: now resides in the fork [android-cwd](https://github.com/ShaMan123/react-native-reanimated/tree/android-cwd). Once this PR is completed including iOS will create a new PR for this issue. In addition this branch experiments with event registry, it enables registering more than 1 event per view + js listener.

1. I think I've stumbled upon a weird bug regarding `event` registration. First of all, if I'm not mistaken `onLayout` can't fire as an animated event because it fires before the event is connected to the view. This is bad. #79 
It seems that something similar happens with `TapGestureHandler` and `RectButton`. The event doesn't fire. However as a `js` event using `setValue` all is good. #375 
See PR #524 

1. `sanitizeArgMapping` fails to handle nested proxies (opened new issue #478)

### To Do `iOS` [WIP fork](https://github.com/ShaMan123/react-native-reanimated/tree/direct-m-ios)
1. Setting up reflection for `ReactModule`, `ViewManager` and the js module event dispatcher.
1. extract logic from `event`, create `map`, `event` extends `map` (almost done).
1. map.fromEnd, map effect proxy (#430)
1. create `callback` (forked from `FunctionNode` with adjustments)
1. Consume `map | callback` into `NSArray`, `Callback` and `Promise` (see Utils in android)
1. Expose `getDirectManipulationUtil` in `ReanimatedModule`
1. Fully expose direct manipulation in example
1. `informViewOfAttachingEvent` - address common optimization that a view emits events only if there's a listener attached to the prop.
1. Useful resources:
[invoke/dispatch](https://github.com/facebook/react-native/blob/master/React/CxxModule/RCTNativeModule.h)
[RCTJSInvokerModule.h](https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Base/RCTJSInvokerModule.h)
[RCTGetModulsClasses](https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Base/RCTBridge.m#L45)
[RCTBridgeModuleNameForClass](https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Base/RCTBridge.m#L82)
[moduleForName/moduleForClass](https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Base/RCTBridge.m#L258)
[enqueueJSCall](https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Base/RCTBridge.h#L185)
[Dynamic Method Resolution](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtDynamicResolution.html#//apple_ref/doc/uid/TP40008048-CH102-SW1)
[Does Objective-C have reflection? - Stack Overflow](https://stackoverflow.com/questions/7550012/does-objective-c-have-reflection)
[Objective-C Runtime/Reflection | Apple Developer Documentation](https://developer.apple.com/documentation/objectivec/objective-c_runtime)
[Objective-C Introspection/Reflection - Stack Overflow](https://stackoverflow.com/questions/2299841/objective-c-introspection-reflection)
[Introspection, reflection and swizzling in Objective C - replace methods](http://www.mirageapps.com/whats-new/articles/introspection-reflection-and-swizzling-in-objective-c)
[dispatchViewManagerCommand](https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Modules/RCTUIManager.m#L1090)

### Next
This is quite a big PR . It made me think about more things that can be done but I have decided to limit the scope for now.
Here is what came to mind:

1. handle wrong view tag gracefully (if the view hasn't been initialized or what)
1. ~**major caveat**: support `map | event | callback` inside `proc` (currently it is only possible to pass them down).~ **EDIT**: only `event` left (see [WIP fork](https://github.com/ShaMan123/react-native-reanimated/tree/event-proc), just need to implement a bit of js and test android)
1. connect `invoke | dispatch` to view in `createAnimatedComponent`.
1. enhance `createAnimatedComponent` or, better off, add a `ReanimatedNativeModules` for common used commands and methods.
1. `map` is well designed because it is a mirror of `event`. I've seen there are a few PRs trying to create array/map nodes (#317 #110 ). This PR basically includes that effort. I think the only thing missing is a way to create new `map | event` as a combination of others. This can be achieved with the manipulation of `argMapping`. I'm thinking of a single function that accepts 2 nodes and a callback comparing them. This abstract way can fit a vast range of needs.

### Finally
This is a fantastic library
Thanks


ShaMan123:
@osdnk reopened
1. got rid of lodash, even though I think it's a bit funny :)
2. made Transitioning safe for iOS, review and tell me if you want me to remove more stuff.
3. > I do need your help figuring out reanimated's evaluation/memoization model.
I don't fully grasp how it works.
I want to check if invoke runs too many times.

4. Anything else?

Did you manage to run the example?

ShaMan123:
Am I on the right fork? Somehow my fork rebased to software-mansion and was previously on kmagiera

ShaMan123:
### iOS
[Accessing ViewManager](https://github.com/facebook/react-native/blob/620f83aae50f00e3e21987d73b2d42768ebe8bea/React/Modules/RCTUIManager.m#L1106)

In iOS are the view manager and the module the same class?

osdnk:
Hey, thanks again!
I think in order to split up this PR I want to ask you for removing all changes related to transitions. Let's focus on one thing for now and then follow up with the rest! :)

ShaMan123:
@osdnk Hey man,
1. Removed all transitioning stuff.
1. Added functionality to `proc`, so it can deal with `map`. So now this is possible, for example:
**Measuring absolute position on screen**:
```js
const successMap = proc((x, y, w, h) => map([0, 0, w, h, x, y]));
const measureView = proc((tag, cb) => cond(defined(tag), invoke('UIManager', 'measure', tag, cb)));


measureView(tag, callback(successMap(x, y, width, height)))
```
 Unfortunately I couldn't come up with a viable way to support `callback` inside `proc` due to it's async nature (resulting in empty stack errors on `ParamNode` because `endContext` was called on `CallFuncNode`). **Do you have any suggestions?**

### Question:
`map` and `callback` are very close in terms of usage.
`callback` is always an array so I've defined it as follows.
```ts
 export function map<T>(argMapping: Mapping<T>, config?: {}): Adaptable<Value>;
 export function callback<T>(...argMapping: Mapping<T>[]): AnimatedNode<number>;
```
However I'm having second thoughts because of this:
```ts
const a = proc((x, y, w, h) => map([0, 0, w, h, x, y]));
const b = callback(0, 0, w, h, x, y);

// should it be this?
const b = callback([0, 0, w, h, x, y]);
   => export function callback<T>(argMapping: Mapping<T>[]): AnimatedNode<number>;
```

What do you think?


### Besides that android is ready.

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
```diff
@@ -17,31 +19,40 @@ public ParamNode(int nodeID, ReadableMap config, NodesManager nodesManager) {
...
+  @Override
   public void beginContext(Integer ref, String prevCallID) {
     mPrevCallID = prevCallID;
     mArgsStack.push(ref);
   }
 
-
+  @Override
   public void endContext() {
     mArgsStack.pop();
```
ShaMan123: Is this vital?
If not `callback` will work in `proc`

Example/directManipulation/shuffle.js
```diff
@@ -0,0 +1,307 @@
+import React, { useState, useRef, useMemo, useEffect, useCallback } from 'react';
+import { StyleSheet, Button, findNodeHandle, UIManager, processColor, Platform, StatusBar } from 'react-native';
+import Animated, { Transitioning, Transition, TransitionState } from 'react-native-reanimated';
```
osdnk: So AFAIK this example is no longer relevant right?

Example/directManipulation/shuffle.js
```diff
@@ -0,0 +1,307 @@
...
+            <Text style={styles.text}>Drag your finger over the list</Text>
+            <RectButton
+              onPress={() => {
+                ref.current.animateNextTransition(() => console.log('animateNextTransition end'));
+                const shuffled = items.slice();
+                shuffle(shuffled);
+                setItems(shuffled);
+
+                //  an ugly workaround until `onTransitionStateChange` will be available on both platforms (hopefully!)
+                setTimeout(() => evaluate.setValue(add(evaluate, 1)), 500);
```
ShaMan123: see here

Example/directManipulation/shuffle.js
ShaMan123: Very relevant. I've changed it to re-evaluate with a timeout instead of once the transition event has been fired.

ShaMan123: See the line in the next review

Example/directManipulation/shuffle.js
ShaMan123: You can comment this line and see how everything stops working as it should

android/src/main/java/com/swmansion/reanimated/nodes/CallFuncNode.java
```diff
@@ -18,25 +18,33 @@ public CallFuncNode(int nodeID, ReadableMap config, NodesManager nodesManager) {
...
-      ParamNode paramNode = mNodesManager.findNodeById(paramId, ParamNode.class);
-      paramNode.endContext();
+      Node paramNode = mNodesManager.findNodeById(paramId, Node.class);
+      ((ContextNode) paramNode).endContext();
     }
     mNodesManager.updateContext.callID = mPreviousCallID;
   }
 
+  @Override
+  public void setValue(Object value) {
```
osdnk: Why `CallFunc` should implement `ValueManagingNode`?

android/src/main/java/com/swmansion/reanimated/nodes/InvokeNode.java
```diff
@@ -0,0 +1,33 @@
+package com.swmansion.reanimated.nodes;
+
+import com.facebook.react.bridge.ReadableMap;
+import com.swmansion.reanimated.NodesManager;
+import com.swmansion.reanimated.Utils;
+import com.swmansion.reanimated.reflection.ReanimatedAccessor;
+import com.swmansion.reanimated.reflection.ReanimatedReflectionHelper;
+
+public class InvokeNode extends Node implements ConnectedNode {
```
osdnk: Why `InvokeNode` is `ConnectedNode`?

react-native-reanimated.d.ts
```diff
@@ -363,12 +398,27 @@ declare module 'react-native-reanimated' {
   }
   export const Easing: EasingStatic;
 
+  export enum TransitionState {
```
osdnk: And this is relevant?

src/core/AnimatedMap.js
```diff
@@ -0,0 +1,124 @@
...
+
+  __attach() {
+    const deps = [...this.__inputNodes, ...this._alwaysNodes];
+    deps.forEach(n => n.__attach());
+    super.__attach();
+  }
+
+  __detach() {
+    const deps = [...this.__inputNodes, ...this._alwaysNodes];
+    super.__detach();
```
osdnk: AFAIK we're detaching `__inputNodes` in `super.__detach();` no?

src/core/AnimatedMap.js
```diff
@@ -0,0 +1,124 @@
...
+    this._alwaysNodes = alwaysNodes;
+  }
+
+  __onEvaluate() {
+    return val(this);
+  }
+
+  __attach() {
+    const deps = [...this.__inputNodes, ...this._alwaysNodes];
+    deps.forEach(n => n.__attach());
```
osdnk: don't we need to attach only `this._alwaysNodes` and then invoke `super.__attach()`?

src/createAnimatedComponent.js
```diff
@@ -15,6 +17,26 @@ function listener(data) {
   component && component._updateFromNative(data.props);
 }
 
+function getEventNode(node) {
+  if (node instanceof AnimatedEvent) {
+    return node;
+  } else if (node instanceof AnimatedCallFunc) {
+    throw new Error('events nested in procs are not yet supported');
```
osdnk: So we throw an error and then do sth after the error?

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
osdnk: yes

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
```diff
@@ -17,31 +19,40 @@ public ParamNode(int nodeID, ReadableMap config, NodesManager nodesManager) {
...
-    return val;
+    try {
+      String callID = mUpdateContext.callID;
+      mUpdateContext.callID = mPrevCallID;
+      Node node = mNodesManager.findNodeById(mArgsStack.peek(), Node.class);
+      Object val = node.value();
+      mUpdateContext.callID = callID;
+      return val;
+    } catch (EmptyStackException e) {
+      throw new JSApplicationCausedNativeException(getClass().getSimpleName() + " is trying to evaluate with no context.", e);
```
osdnk: It should never happen

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
osdnk: if you need this you can make a copy of the stack and pass it to the callback and the on resolving revert previous stack for a while?

android/src/main/java/com/swmansion/reanimated/nodes/InvokeNode.java
```diff
@@ -0,0 +1,33 @@
...
+    ReanimatedAccessor mEvalHelper;
+    protected final int[] mParams;
+    private int invocationId = 0;
```
osdnk:
```suggestion
    private int mInvocationId = 0;
```

and why you need this?

Example/android/app/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,7 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
+    <uses-permission android:name="android.permission.VIBRATE"/>
```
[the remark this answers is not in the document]
osdnk: nvm :)

android/src/main/java/com/swmansion/reanimated/nodes/InvokeNode.java
ShaMan123: I've set it up for future use in `createAnimatedComponent` just like `PropsNode`.
If you connect it:
```diff
-  dispatch('module', 'command', tag, ...args)
+  dispatch('module', 'command', ...args)
```
Haven't tested it yet though

react-native-reanimated.d.ts
ShaMan123: missed that

src/core/AnimatedMap.js
ShaMan123: Wasn't sure about that. Are we?

android/src/main/java/com/swmansion/reanimated/nodes/InvokeNode.java
osdnk: So I'd rather revert this change if it's not used neither verified :)

src/core/AnimatedMap.js
osdnk: Take a look at `AnimatedNode.js`

src/createAnimatedComponent.js
ShaMan123: for future usage.
I'm thinking about refactoring `event` to completely extends `map`

osdnk: I'd rather drop it as well. It's not good to do anything after throwing an error

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
ShaMan123: What if the `param` is a `proc`? wouldn't I need to do that for every param in the tree that is needed for the evaluation?

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
ShaMan123: It happens with callback because it's async

react-native-reanimated.d.ts
ShaMan123: commited

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
osdnk: So the error should say so

android/src/main/java/com/swmansion/reanimated/nodes/CallFuncNode.java
ShaMan123: In the case a `proc` was passed as a value that needs to be set by `callback`

src/createAnimatedComponent.js
ShaMan123: removed

android/src/main/java/com/swmansion/reanimated/nodes/CallFuncNode.java
ShaMan123: If I'm not mistaken this is important for `set` too.

android/src/main/java/com/swmansion/reanimated/nodes/InvokeNode.java
ShaMan123: 74074ce

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
ShaMan123: 5d23b28

src/createAnimatedComponent.js
ShaMan123: 8080646

src/core/AnimatedMap.js
ShaMan123: 727c164

src/core/AnimatedMap.js
ShaMan123: 727c164

react-native-reanimated.d.ts
ShaMan123: 390f0f2 19e70f3

ShaMan123:
@osdnk is it correct to say that if I have a `param` that in turn holds a `proc` and so on then to use it with `callback` would require cloning the entire node tree?

android/src/main/java/com/swmansion/reanimated/nodes/InvokeNode.java
ShaMan123: verified
1f75bc9

android/src/main/java/com/swmansion/reanimated/nodes/ParamNode.java
ShaMan123: Let's continue this [here](https://github.com/software-mansion/react-native-reanimated/pull/469#issuecomment-555013397)

osdnk:
https://github.com/ShaMan123/react-native-reanimated/pull/1

ShaMan123:
I'll look into the PR.
Rethinking 1f75bc9, wanted to hear your thoughts:
AFAIK `connectToView` can't be used with `proc` because `proc` reuses the same `dispatch`/`invoke` node. Calling `connectToView` essentially connects the `proc` to the view, which is undesirable and is the same as a memory leak. 
**Using `connectToView` = no `proc`.**
To my understanding `proc` is the future of `reanimated` so anything that doesn't support it should be considered bad practice.
Am I right?
Plus, it is very convenient to use `proc` with `dispatch`/`invoke`, much more than `connectToView`, which is a tiny, non-vital option for `dispatch`:


_Originally posted by @ShaMan123 in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjEzOTQxODM1OnYy/pull_request_review_threads/discussion_
I've set it up for future use in `createAnimatedComponent` just like `PropsNode`.
If you connect it:
```diff
-  dispatch('module', 'command', tag, ...args)
+  dispatch('module', 'command', ...args)
```


### So I vote to remove it completely.

osdnk:
Hm, I'm not super confident about your reasoning, but I'm fine with the necessity of adding a tag explicitly on each invoke :).
After having it merged we can discuss another approach (connecting to view).

ShaMan123:
All set.

osdnk:
Hey @ShaMan123,
Thanks a ton for this amazing progress. I'm really excited about these lines!
However, I see a few issues regarding examples:
1. Shuffle sometimes leaves one element  
image

2. ControlledScrollView is not really scrollable
https://streamable.com/0kynq

ShaMan123:
looking into it

ShaMan123:
https://streamable.com/or1nh
I can't reproduce..
I have 2 suggestions:
1. I think the scrollable issue is because of `waitFor` I commented out. I'll commit it back.
But shuffle... 😬
I'm not sure it's to do with direct manipulation. I can't think of how.
1. Can you try my working fork and see if it persists? [android-working-branch](https://github.com/ShaMan123/react-native-reanimated/tree/android-working-branch)

osdnk:
So I can't see an error with shuffle on `android-working-branch`

ShaMan123:
That's good.
I'm just about to commit a change.
Hopefully it'll fix it.

osdnk:
I see that streamable.com/or1nh appears only on Simulator. Cannot see an issue on device

ShaMan123:
on `master` too?
I'm developing on a device

osdnk:
yes

ShaMan123:
so it's a fake bug?

osdnk:
I mean since it's only an example I'm okay with having it merged even if it's not perfect. But it's definitely kind of drawback

ShaMan123:
I'll try debugging with simulator

osdnk:
Another thought regarding `SyncedScrollView`:
Is it possible to make overdrag synced as well? https://streamable.com/x5eqs

ShaMan123:
I saw that too.
Could it be because of `decay` still running?

osdnk:
Here it's visible: https://streamable.com/abgtu

ShaMan123:
I got confused. I meant `decay` in `ControlledScrollView`.
Yeah, I need to add a `cond` to stop scrolling or something once a touch starts

osdnk:
I'm talking about `SyncedScrollViews` and overdrag effect

osdnk:
> I got confused. I meant decay in ControlledScrollView.
> Yeah, I need to add a cond to stop scrolling or something once a touch starts

Whatever which will work :)

osdnk:
Also, please rebase onto master and use linter :)

ShaMan123:
can you eliminate the cause of the bug being `Transition`?

ShaMan123:
@osdnk I solved a few caveats. And linted.
added these features:
```
map.fromEnd([foo, bar])
debug('evaluating this is very expensive', callback(debug('some value I can debug, foo))
```

for the need to pop dynamic arrays.

Android is in a good shape.
The transition problem was solved by #483, I've updated the fork so you can test it yourself.
Overall I came across 2 major issues:
1. Thread asserting operations, not allowing evaluation on separate threads. 
  e.g. `invoke('UIManager`, `measureLayout, ...)`
1. `map` `evaluate()` is very expensive, 
  I've optimized it as far as I can with memozation and laziness so it's not bad but it's not good for sure. If you want to test it in an extreme case simply wrap `measureView` `callback` with a `debug` node in `shuffle.js`

ShaMan123:
I think it is stable enough to merge

ShaMan123:
I was looking at the map problem the wrong way. Using `WritableNativeMap` is necessary only for crossing the bridge so I'll stub the functionality.

osdnk:
Can you please let me know when you'll feel it's ready. I still see new commits and I'm uncertain about the status. 
Also, would you have time for scheduling a call regarding these changes? Can you DM me on TT or with e-mail?

ShaMan123:
Your right about the status, sorry. I'm refactoring again the map and it is very promising.
What is DM?

osdnk:
Direct message

ShaMan123:
I'm hoping to finish up in 2 hours or so. What's your email?

ShaMan123:
found it

osdnk:
micosa97@gmail.com

ShaMan123:
introduced `intercept` - a listener intercepting global events sent through the bridge (via `getJSModule()`)

Before finalizing android I have 2 more things to finalize:
1. `callback` inside `proc` - @osdnk I may need your help for this one. Made some progress
1. optimize `map` value

ShaMan123:
I think I'm done.
Checking now implementation with a third party library to vet some stuff.
The last bunch of commits made all direct manipulation nodes compatible with `proc`.

android/src/main/java/com/swmansion/reanimated/reflection/ReanimatedMapBuilder.java
```diff
@@ -147,19 +157,67 @@ public A asArray() {
...
+                collection = collection.has(key) ? collection.value(key, ReanimatedBridge.ReadableCollection.class) : null;
+            }
+
+            if (collection != null) {
+                key = path.get(path.size() - 1);
+                Object source = mMapBuilder.resolver(key).source();
+                if (source instanceof ReanimatedWritableMap) {
+                    ((ReanimatedWritableMap) source).put((String) key, value);
+                } else if (source instanceof ReanimatedWritableArray) {
+                    ((ReanimatedWritableArray) source).set((int) key, value);
```
ShaMan123: @osdnk I need your help here.
This method is supposed to help reduce map value calculation overload.
It manages to set the value on `source` at the right key. It seems that `source` is a local copy, not a ref to the original object, hence changes do not propagate, making this approach stale.

ShaMan123:
Trying to implement in third party raises difficulties with `callback` inside `proc`.
+ I haven't addressed `alwaysNode` in case of a `mapNode` proxy.

ShaMan123:
@osdnk 
## Context Propagation
### The problem
`callback` is used asynchronously, setting values on nodes the same as `map` does. Therefore, if nested inside `proc` it has to evaluate in some previously flushed context.
My solution is creating a kind of an instance of `callback` named `ContextProvider.ValueManager` that evaluates with the given context. This way the sync context begun by some `CallFuncNode` is saved by the context provider for later use. Context is then flushed, once needed `callback` restores context and sets value.
Trying to understand more complex use cases to determine if this approach is solid.

ShaMan123:
Fabric working with no problems!

Jarred-Sumner:
I would love to use this!

Another example usecase: I want to use this [color slider library](https://gizmosachin.github.io/ColorSlider/Classes/ColorSlider.html) (which is written in Swift) and have it update an `Animated.Value` when the color changes. If this was merged, then I think I could just write a small subclass of this which accepts a `red`, `green`, and `blue` `REANode` as a prop.  When the `valueChange` action runs, it would update the nodes.

ShaMan123:
@Jarred-Sumner If I understand correctly `event` better fits your needs

ShaMan123:
@osdnk edited my review on context propagation

mikedemarais:
@ShaMan123 will `direct manipulation nodes` allow us to bind reanimated to `<BlurView />` from `@react-native-community/blur`?

ShaMan123:
@mikedemarais I looked at BlurView. The only manipulation done to it is via props, meaning `createAnimatedComponent` should do the job (no commands, methods or global events (accept an error event)). Am I missing something?
What's the use case?

osdnk:
@ShaMan123 Do you find current state ready for reviewing?

ShaMan123:
I think so.
There might be a need to commit a minor change to `PropsNode` that I already tested.

ShaMan123:
I am sure structure of direct manipulation will not change

bartzy:
Just as a +1 (not sure this is completely relevant to post here) - I think this PR is essential for implementing behaviors like the one I posted here: https://stackoverflow.com/questions/59446635/collapsible-header-in-react-native-reanimated-with-multiple-tabs-in-react-native

Without controlling the scroll view from reanimated itself (by using `dispatch` as @ShaMan123` suggests), I don't think it's possible to implement this kind of UI.

Jarred-Sumner:
I think you can do it without controlling the scroll view via contentInset
+ contentOffset and then having a containing Animated.View which has a
translateY and then using diffClamp like any of the guides say on how to do
collapsing header in a scroll view

Basically the header would sit above the tab view, each of the scroll views
would update the scroll the header tracks,  and then you can interpolate on
the position change in the tab bar to reset the header to be tracking the
scroll y of the correct tab, so the header would change as you scroll
sideways without a gap. Avoid using paddingTop/paddingBottom to do content
inset, unless you’re on Android where it’s not supported 😢

the downside of this approach is all the animated nodes it’d create,
unmounting might be slow

ShaMan123:
@bartzy If I understand what you're saying you want to move from using:
```
scrollContentContainerStyle={{ paddingTop: TOTAL_HEADER_HEIGHT }}
```
I think `dispatch` is an option. 
Another approach that comes to mind is to use an animated node for `paddingTop` that is shared across all tabs. Could this work?

bartzy:
@ShaMan123 Can paddingTop have an animated value attached to it?

ShaMan123:
why not?

bartzy:
Does it cause a layout when the animated value change? and if so, is it on the JS side or still only natively?

ShaMan123:
Entirely native. Regarding layout not sure, I guess you mean printing/redrawing. It should do that.

bartzy:
@ShaMan123 I see, thanks. by layout I meant that Yoga needs to relayout views since padding is a layout property.

ShaMan123:
Don't know. Maybe you should open an issue and ask it there. With `dispatch` you can drop padding and adjust scroll offset directly.

ShaMan123:
android is stable.
ios not developed yet.
I couldn't understand your use case

xenophilicibex:
@ShaMan123 thanks for the information. I guess I'll try to find another way since I'm working on iOS. I couldn't record by screen, but I was able to replicate my problem through the simulator.

Here are two GIFs.

First one, buttery smooth animation. This is how I want the performance to be. _But it doesn't affect anything else._
Good performance

Second one, poopy animation. But this is how I want it to function. _As the hexagon expands, the tabs slide away_ (using the same Animation.Value) (Notice the JS FPS dropping to 20s)
Kapture 2020-01-16 at 14 19 18

In my app I have a number of gesture based animations like this that affect the rest of the app. And when it goes through JavaScript, the FPS tanks. I understand the purpose of Reanimated is to avoid going through the JS bridge, but right now I don't know of any other away to achieve the effect I want.

If anyone has any suggestions, I welcome them.

ShaMan123:
If you're not using native modules with reanimated this PR doesn't affect you at all, it seems to be the case.
I suggest you file a question/issue.

ShaMan123:
@osdnk why don't you merge?

ouoru:
This is looking like an amazing PR, which solves a lot of the difficulty working with scrollViews. @osdnk Is there any way to pull in resources to get this working for iOS?

ShaMan123:
7104c33 fixes #430

ouoru:
@ShaMan123 is the scrollTo use case working for iOS? I would love to test this out as soon as possible 👍 let me know when I can!

kmagiera:
Thanks a lot for your work on this @ShaMan123. This work inspired us to design Reanimated 2 API to make such cases much simpler. As you noted there is an ongoing work to bring a similar functionality to v2 in #936. We will likely reuse a lot of what you built here in that PR or in a followup set of changes. That being said, I'll be closing this PR as we won't have enough capacity to support such a big set of changes in V1 while focusing on getting V2 stable and making V2 fill the gap this PR was addressing. Thanks again and looking forward to your comments on #936

ShaMan123:
V2 looks promising, very different though.
I think you guys should maintain a clear and concise roadmap (and get to writing some guides and concepts ASAP) so the community can join in and boost your efforts.

PR #479, opened 2019-11-20
Feature: Add support for JS node

chrfalch:
## Intro
In reference to @osdnk's PR #458 about CodeGen, this is a draft showing new node that makes it possible to execute Javascript functions on node evaluation. 

## Background
The reason for this is that we are seeing too many nodes being created in Reanimated when animations are a bit complex, and we would like to move some of the overhead of transporting thousands of nodes across the bridge.

## Example

```js
const a = new Animated.Value(10);
const jsNode = Animated.js("function (a, b) { return a + b }", a, 120);
```

We can write *pure functions* that operates on input nodes - whatever logic we wanted would be available as long as it does not have any side effects. JS could should be evaluated directly (not through the bridge) and return. 

## Technical
The way it is done for now is that the NodeManager has a Javascript context (not using the bridge) that nodes can access. A new node has been added (REAJSNode) which executes a javascript function passed from React Native JS on evaluate. It supports taking other nodes as parameters as well (just like the proc node).

On iOS we're using the JSContext class, whilst on Android I was not able to find a way to execute Javascript on the currently loaded JS engine (Hermes / JSC) without going through the bridge. It seems like the functions for doing this is written in C++ and not exposed through any Java interface at the moment.

## Advanced example
This is an example showing a spring written in Javascript that results in only one node being created for the whole spring. (the spring function is actually implemented as a precomputed easing to be able to know its duration).

```js
function getSpringNode (
  duration: number,
  mass: number,
  stiffness: number,
  damping: number,
) {
  const w0 = Math.sqrt(stiffness / mass);
  const zeta = damping / (2 * Math.sqrt(stiffness * mass));
  const wd = zeta < 1 ? w0 * Math.sqrt(1 - zeta * zeta) : 0;
  const a = 1;
  const b =
    zeta < 1 ? (zeta * w0 + -initialVelocity) / wd : -initialVelocity + w0;

  return (t: Animated.Node) => {
    return Animated.js(
      `function (t) {      
      var progress = (${duration} * t) / 1000;
      if (${zeta} < 1) {
        progress =
          Math.exp(-progress * ${zeta} * ${w0}) *
          (${a} * Math.cos(${wd} * progress) + ${b} * Math.sin(${wd} * progress));
      } else {
        progress = (${a} + ${b} * progress) * Math.exp(-progress * ${w0});
      }
      if (t === 0 || t === 1) {
        return t;
      }
      return 1 - progress;      
    }`,
      t,
    );
}
```

## Discussion
Adding the JS node is a bit radical - it might introduce some issues for devs, as you will not be able to debug your functions directly (you can write tests using `eval()`). 

We also need to discuss wether we think it is correct to use the V8 JS engine on Android, or if we should ask for some help to see if it is possible to execute them outside the bridge on the currently loaded engine. Maybe @kmagiera knows something or someone that could add to this? I don't even know if all Android devices have V8 preinstalled... ;)

### DIFF
--- a/src/core/AnimatedJSFunc.js
+++ b/src/core/AnimatedJSFunc.js
@@ -1,0 +1,17 @@
+import AnimatedNode from './AnimatedNode';
+import { adapt } from './AnimatedBlock';
+
+class AnimatedJSFunc extends AnimatedNode {
+
+  constructor(code, args) {
+    super({ 
+      type: 'jsfunc', 
+      args: args.map(n => n.__nodeID),
+      code: code,
+    }, [...args]);
+  }
+}
+
+export function createAnimatedJSFunc(code, ...args) {
+  return new AnimatedJSFunc(code, args.map(p => adapt(p)));
+}
--- a/Example/test/index.js
+++ b/Example/test/index.js
@@ -167,7 +167,7 @@ export default class Example extends Component {
         {Array.from(Array(40)).map((_, i) => (
           <Animated.View
             key={i}
-            style={[styles.box, { transform: [{ translateX: this.t[i] }] }]}
+            style={[styles.box, { transform: [{ translateX: Animated.js("function(t){return t*0.25}",this.t[i] ) }] }]}
           />
         ))}
       </View>
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -52,6 +52,7 @@ dependencies {
     //noinspection GradleDynamicVersion
     implementation 'com.facebook.react:react-native:+'  // From node_modules
     implementation "androidx.transition:transition:1.1.0"
+    implementation 'com.eclipsesource.j2v8:j2v8_android:3.0.5@aar'
 }
 
 def configureReactNativePom(def pom) {
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -2,6 +2,7 @@ package com.swmansion.reanimated;
 
 import android.util.SparseArray;
 
+import com.eclipsesource.v8.V8;
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Callback;
 import com.facebook.react.bridge.GuardedRunnable;
@@ -78,6 +79,7 @@ public class NodesManager implements EventDispatcherListener {
 
   public double currentFrameTimeMs;
   public final UpdateContext updateContext;
+  public final V8 jsRuntime;
   public Set<String> uiProps = Collections.emptySet();
   public Set<String> nativeProps = Collections.emptySet();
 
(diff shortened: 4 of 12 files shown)


terrysahaidak:
Are you able to point the place where in c++ you have seen js context on Android? Maybe we can setup JNI and run that c++ code from Java.

chrfalch:
It is a bit hard to find, since it is hidden in some base classes. But I managed to load the java js executor interface classes, but they don’t have any methods in java for executing js functions. Spent a few hours looking for a way “in” without any success.

chrfalch:
Some findings on performance:

iOS can run hundreds of js evaluations or frame without any problems. Android struggles with this. Not a surprise - but we might need to pursue seeing if Hermes might help in any way if this is an issue with Google V8.

ShaMan123:
barging in with a suggestion...
1. why not use `Function.toString()` ?
2. #469 enables interacting with native modules. It's a different solution but one could expose methods via a native modules and use `invoke` node to evaluate them, or even better simply accept a node id and not use `invoke`. It doesn't enable js but it's solid and high on performance.

terrysahaidak:
@ShaMan123 this draft is an attempt to solve performance problems of Reanimated. As I understand by a second suggestion you mean implementing the same math logic in Obj-C and Java (twice write the same code) and use it. But why? Also, it breaks OTA updates :(

ShaMan123:
You understand correctly. I agree, but react-native is all about 2 cores.

src/core/AnimatedJSFunc.js
```diff
@@ -0,0 +1,17 @@
...
+
+class AnimatedJSFunc extends AnimatedNode {
+
+  constructor(code, args) {
+    super({ 
+      type: 'jsfunc', 
+      args: args.map(n => n.__nodeID),
+      code: code,
+    }, [...args]);
+  }
```
EvanBacon: wouldn't this need to implement an `__onEvaluate` as well?

chrfalch: In this draft (which will probably be deleted) only the native side was implemented to test if this was a viable way to get faster executing by reducing the number of nodes.

osdnk:
I think I might try to merge it only on iOS. I think it's possible to do it on Android, but idk how

terrysahaidak:
folks, did you see https://github.com/joltup/react-native-threads ?

chrfalch:
> I think I might try to merge it only on iOS. I think it's possible to do it on Android, but idk how

Please don't merge this one, @osdnk :) I think it will be too hard to fix on Android - and might create friction for developer.

terrysahaidak:
And as the author of react-native-threads mentioned in its readme, creating separate VMs may be really bad for memory :)

jfrolich:
As an alternative, wouldn’t it be possible to send a data structure in a single node that composed multiple operations to reduce nodes? It would also be possible to collapse static nodes using a Babel plugin. Or are there some complexities I’m missing here?

jakub-gonet:
This PR is a bit outdated, especially regarding the recent release of Reanimated2, which has support for executing JS functions on the native side out of the box.

@chrfalch, many thanks for your effort ❤️

chrfalch:
No problem, @jakub-gonet! Awesome to see how far you software-mansion-guys have come with this original idea in Reanimated2!!

Issue #480, opened 2019-11-20
Examples don't run on iOS


brobertsUPS:
Thanks for the project!
I am getting an error when running the examples:

```
/react-native-reanimated/Example/ios/Pods/Target Support Files/Pods-ReanimatedExample/Pods-ReanimatedExample.debug.xcconfig: unable to open file (in target "ReanimatedExample" in project "ReanimatedExample") (in target 'ReanimatedExample'
```

I thought this might be similar to this failure to run the examples but I get a different error than in this issue: https://github.com/software-mansion/react-native-reanimated/issues/326

I follow the instructions for the example by:

```
Run yarn to install the dependencies of library
cd to Example directory
Run yarn to install the dependencies of example
Run yarn start to start the packager with custom config
react-native run-ios
```

I ensured that I have the pre-req `react-native` installed globally as well. Not sure if I missed something but I thought that would do it.

RockinRonE:
@brobertsUPS I was able to get this to work by running `npm run start` in one tab in my terminal and then running `react-native run-ios`. I also used `npm install`. 

I also went into the `Pods` directory and ran `pod install`. I'm not sure which one was the right way, but that was my sequence of events to get it to run!

brobertsUPS:
Thanks @RockinRonE! I was able to get them running after I did the `pod install` in the `ios` directory. I will update the readme in the `Example` directory.

PR #482, opened 2019-11-22
Update docs for running examples on iOS

brobertsUPS:
closes: #480 

- Add instructions to `pod install` before `react-native run-ios`

### DIFF
--- a/Example/README.md
+++ b/Example/README.md
@@ -1,8 +1,9 @@
 ## Run the example
 
-+ Clone the repository
-+ Run `yarn` to install the dependencies of library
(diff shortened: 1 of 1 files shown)


AdamGerthel:
Perhaps the installations instructions on this page should be removed or updated: https://software-mansion.github.io/react-native-reanimated/examples.html

jakub-gonet:
That's right, probably weren't deployed, if you take a look at [source file](https://github.com/software-mansion/react-native-reanimated/blob/master/docs/pages/15.example.md) you'll see that they're up to date

Issue #495, opened 2019-12-10
-0.0 is not equal to 0.0 on Android


zamotany:
## Bug report

Animated Value of `-0.0` is not equal to Animated Value of `0.0` on Android, but it's working properly on iOS. Even when wrapping `-0.0` with `abs` the result is the same.

## Environment
- `react-native`: `0.61.2`
- `react-native-reanimated`: `1.4.0`

## Repro code

Running the following code:

```js
Animated.useCode(
    () =>
      Animated.block([
        Animated.debug(
          'ReanimatedTest: is -0.0 equal 0.0 =',
          Animated.eq(new Animated.Value(-0.0), new Animated.Value(0.0)),
        ),
        Animated.debug(
          'ReanimatedTest: is -0.0 less or equal 0.0 =',
          Animated.lessOrEq(new Animated.Value(-0.0), new Animated.Value(0.0)),
        ),
        Animated.debug(
          'ReanimatedTest: is 0.0 less or equal -0.0 =',
          Animated.lessOrEq(new Animated.Value(0.0), new Animated.Value(-0.0)),
        ),
        Animated.debug(
          'ReanimatedTest: is abs(-0.0) equal 0.0 =',
          Animated.eq(
            Animated.abs(new Animated.Value(-0.0)),
            new Animated.Value(0.0),
          ),
        ),
      ]),
    [],
  );
```
Gives:
```
ReanimatedTest: is -0.0 equal 0.0 = 0.0
ReanimatedTest: is -0.0 less or equal 0.0 = 1.0
ReanimatedTest: is 0.0 less or equal -0.0 = 1.0
ReanimatedTest: is abs(-0.0) equal 0.0 = 0.0
```

## Workaround

When comparing value which can be `-0.0` for equality, the following snippet returns correct results:

```js
lessOrEq(
  yourAnimatedValue,
  new Animated.Value(0.0),
)
```

jakub-gonet:
Hey @zamotany, thanks for reporting!

Could you write how you stumbled upon this bug? It's a bit unusual to compare doubles by eq and even more unusual to get `-0.0`. I wrote explanation below, but I'm deeply interested how did you manage to discover that :>

This happened because we compared values by `Double.equals(Double)` which respects [IEEE 754 standard](https://en.wikipedia.org/wiki/Signed_zero) and that's why `-0.0 != 0.0` on Android.

After merging #498 this should be consistent on all platforms.

zamotany:
[...]

PR #498, opened 2019-12-12
[Android] Fix EQ and NEQ operators. 

Szymon20000:
Related to: https://github.com/software-mansion/react-native-reanimated/issues/495

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/nodes/OperatorNode.java
+++ b/android/src/main/java/com/swmansion/reanimated/nodes/OperatorNode.java
@@ -189,7 +189,7 @@ public class OperatorNode extends Node {
   private static final Operator EQ = new CompOperator() {
     @Override
     public boolean eval(Double x, Double y) {
-      return x.equals(y);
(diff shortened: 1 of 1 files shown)


wcandillon:
Could the tests from #495 be added to this PR?

android/src/main/java/com/swmansion/reanimated/nodes/OperatorNode.java
```diff
@@ -213,7 +213,7 @@ public boolean eval(Double x, Double y) {
   private static final Operator NEQ = new CompOperator() {
     @Override
     public boolean eval(Double x, Double y) {
-      return !x.equals(y);
+      return !(x.doubleValue() == y.doubleValue());
```
jakub-gonet:
```suggestion
      return x.doubleValue() != y.doubleValue();
```

jakub-gonet: This is more readable IMHO

Issue #481, opened 2019-11-20
How to run examples in Web?


brobertsUPS:
Thanks for all of the work on this project!

I would like to run the examples in the browser. I read that there is now browser support and looked through this PR by @EvanBacon https://github.com/software-mansion/react-native-reanimated/pull/390. I know they are running live [here](https://reanimated.netlify.com/#/Main) but I think I am missing something in the setup.

I followed the instructions to run the examples and used the `web` script in package.json to try running on web:

- Clone the repository
- Run yarn to install the dependencies in the root of this repo
- cd to Example
- Run `yarn`
- Run `yarn web` (found in the package.json)

I get the following error when the examples open up and I navigate to any of them:

`TypeError: this._what._updateValue is not a function`

Located in:

`react-native-reanimated/Example/Animated/core/AnimatedSet.js:20`

Does anyone know how to get these running locally? 

I had a hunch that maybe we need to do an `expo install react-native-reanimated` to get the correct version but that didn't seem to change anything. Thanks for any help!

PR #499, opened 2019-12-12
Possible fix for web error

samin:
Possible fix for #481


EvanBacon:
This seems to create an endless loop

EvanBacon:
Created a fix with #532

Issue #235, opened 2019-04-08
`interpolate` not backwards compatible with [0,0] range


xaviershay:
Hello! Thanks for this library - it seems like it's going to be a big improvement over the old `Animated`.

I'm trying to upgrade an app to RN 0.59, which via [`react-navigation`](https://github.com/react-navigation/react-navigation/issues/5684) and `react-navigation-tabs` has necessitated upgrading to this library. I'm getting a whole lot of `NaN` from what I was expecting to be backwards compatible calls (from README: "Use the same Animated library API from React Native that you've been using. You don't need to change anything."). For example, the [interpolate](https://github.com/react-native-community/react-native-tab-view/releases/tag/v2.0.1) example from the release notes:

    # Old API
    > a = new OldAnimated.Value(4)
    > a._value
    4
    > a.__getValue()
    4
    > c = a.interpolate({inputRange: [0, 10], outputRange: [10, 20], extrapolate: 'clamp'})
    AnimatedInterpolation {_children: Array(0), _parent: AnimatedValue, _config: Object, _interpolation: function}
    > c._value
    undefined
    > c.__getValue()
    14

    # New API
    > b = new Animated.Value(5)
    > b._value
    5
    > b.__getValue()
    NaN
    > d = Animated.interpolate(b, {inputRange: [0, 10], outputRange: [10, 20], extrapolate: 'clamp'})
    AnimatedCond {__lastLoopID: 0, __memoizedValue: null, __children: Array(0), __nodeID: 7266, __nodeConfig: Object…}
    > d._value
    undefined
    > d.__getValue()
    NaN

(I'm not sure which of `_value` or `__getValue()` is supposed to be the same, I can't find public API clearly documented, but this does seem to be causing my issue.)

I've also noticed similar disparities in other functions, such as return value of `Animated.subtract`.

This leads to hard to track down bugs, since the `NaN` value often gets all the way into native code and gives no stack trace as to where it originated.

What is the intended scope of backwards compatibility here? Am I doing it wrong?

xaviershay:
So on further investigation, though I'm still curious about the values returned above, I'm managed to find some "working" instances of `interpolate` in my app, so it's probably something different going on. Will re-open if I can get a better repro.

xaviershay:
I think I figured out an actual incompatibility: in the old API, interpolating between `[0, 0]` would return 0, where as with new API it returns `NaN`. While the latter is "more correct", it causes an error that doesn't point to the source of the issue:

[...]

PR #502, opened 2019-12-13
Make interpolate consistent with react-native animated lib

Szymon20000:
WHY:  
resolves: https://github.com/software-mansion/react-native-reanimated/issues/235

To compute a proper value for an input, we divide in code by input range size which is incorrect for ranges that have sizes equal to zero.  This pull-request fixes this bug.

How it's implemented in Animated:
https://github.com/facebook/react-native/blob/master/Libraries/Animated/src/nodes/AnimatedInterpolation.js#L140

### DIFF
--- a/src/derived/interpolate.js
+++ b/src/derived/interpolate.js
@@ -5,6 +5,8 @@ import {
   add,
   divide,
   greaterThan,
+  lessOrEq,
+  eq,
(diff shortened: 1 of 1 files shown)


osdnk:
Can you add some PR description and comment in code?

PR #507, opened 2019-12-13
Adjust IOS CI job

jakub-gonet:
Add caching, enable IOS build on all branches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,11 +41,9 @@ matrix:
     xcode_destination: platform=iOS Simulator,OS=12.2,name=iPhone X
     env: NODE_VERSION=10.15.3
 
-    branches:
-      only:
-      - master
-
-    cache: yarn
+    cache:
+      yarn : true
+      cocoapods: true
 
     install:
     - nvm --version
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,11 +41,9 @@ matrix:
     xcode_destination: platform=iOS Simulator,OS=12.2,name=iPhone X
     env: NODE_VERSION=10.15.3
 
-    branches:
-      only:
-      - master
```
osdnk: The idea under this line was to make PR build only if PR is targeted to master. IMO it's a reasonable approach not to build every commit pushed to any branch in the repo.

Szymon20000:
We don't have Podfile in the root directory so probably this won't work. There is also a problem with the Development Pods.  However, we always do "pod install" so they shouldn't get out of sync with the cached version(development pods are linked but new files are not if we don't call "pod install").  Maybe we should cache Pods like [here](https://docs.travis-ci.com/user/caching/#arbitrary-directories)?

.travis.yml
jakub-gonet: Fixed.

osdnk:
What's the status?

osdnk:
Basically looks good. Have you actually made a benchmark and observed a boost? If yes, can you share your results?

Please, fix conflicts

jakub-gonet:
After fixing CI this PR is no longer relevant, I'm closing it.

PR #513, opened 2019-12-19
Use in AnimatedValue.setValue react method instead of evaluateOnce.

Szymon20000:
There is no need to create new nodes just to change the value of ValueNode.


osdnk:
let's wait for CI

Issue #79, opened 2018-08-20
Declarative onLayout


osdnk:
It would be a nice feature to add possibility to use `onLayout` in declarative way with `Animated.Event` and normal non-animated `View`

osdnk:
Welp, it's not really useful I realized

vaukalak:
@osdnk do you mean:
```
<Animated.View
        onLayout={event([
          { nativeEvent: { layout: { y: screenWidth } } }
        ])}
...
</Animated.View>
```
Why do you think this is not useful? I think it's way better than using callback.

osdnk:
I thought It worked. Now I see it does not

osdnk:
Hm, it works FM
```jsx
import React from 'react';
import { StyleSheet, View, Dimensions, Animated as NativeAnimated } from 'react-native';
import Animated from 'react-native-reanimated';;

const { cond, eq, add, call, set, Value, event, multiply } = Animated;

export default class Example extends React.Component {
  constructor(props) {
    super(props);
    this.val1 = new Value(100);
    this.val2 = new NativeAnimated.Value(1);
  }
  render() {
    return (
      <View style={styles.container}>
        <Animated.View
          onLayout={Animated.event([
            {
              nativeEvent: {
                layout: {
                  height: this.val1
                }
              }
            }
          ])}
          style={[
            styles.box,
            {
              backgroundColor: 'blue',
              transform: [ { translateY: this.val1 }]
            }
          ]}
        />
        <NativeAnimated.View
          onLayout={NativeAnimated.event([
            {
              nativeEvent: {
                layout: {
                  height: this.val2
                }
              }
            }
          ], {
            useNativeDriver: true
          })}
          style={[
            styles.box,
            {
              transform: [
                {
                  translateY: this.val2
                }
              ]
            }
          ]}
        />
      </View>
    );
  }
}

const CIRCLE_SIZE = 70;

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',

  },
  box: {
    backgroundColor: 'tomato',
    marginLeft: -(CIRCLE_SIZE / 2),
    marginTop: -(CIRCLE_SIZE / 2),
    width: CIRCLE_SIZE,
    height: CIRCLE_SIZE,
    borderRadius: CIRCLE_SIZE / 2,
    borderColor: '#000',
  },
});

```

osdnk:
[...]

PR #524, opened 2019-12-26
Positive Event Emitting

ShaMan123:
This small patch addresses the common performance optimization that emits events only if a prop exists or not.
Issues:  #430 #79 #375 
This is why `onLayout` was not handled by reanimated as well as more events.
See example to test changes.


ShaMan123:
closing this mess in favor of #528

PR #528, opened 2019-12-28
[v1] Patch event subscription

### LINKED ISSUE #79: Declarative onLayout

### LINKED ISSUE #375: event() handler skips State.ACTIVE when used with <TapGestureHandler />

### LINKED ISSUE #430: How to respond to events like TextInput focus/blur and Keyboard show/hide

ShaMan123:
This small patch addresses the common performance optimization that emits events only if a prop exists or not.

Fixes #430.
Fixes #79.
Fixes #375.

This is why `onLayout` was not handled by reanimated as well as more events.
See example to test changes.


src/createAnimatedComponent.js
```diff
@@ -109,8 +122,13 @@ export default function createAnimatedComponent(Component) {
         if (prop instanceof AnimatedEvent && !attached.has(prop.__nodeID)) {
           // not yet attached
           prop.attachEvent(node, key);
+          nativeUpdate[key] = true;
         }
       }
+
+      if (Object.keys(nativeUpdate).length > 0) {
+        this.setNativeProps(nativeUpdate);
```
jakub-gonet, line 130: Are we sure that `setNativeProps` is available here? Especially regarding [comment below](https://github.com/software-mansion/react-native-reanimated/pull/528/files#diff-3be87c4c6b24297654a7e0b2990508f2R134-R138).

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -398,6 +398,13 @@ private void handleEvent(Event event) {
       int viewTag = event.getViewTag();
       String key = viewTag + eventName;
       EventNode node = mEventMapping.get(key);
+      // If for some reason name resolution failed try a more generic approach
+      // converting `topEventName` to `onEventName`
+      if (node == null && eventName != null && eventName.startsWith("top")) {
+        eventName = "on" + eventName.substring(3);
+        key = viewTag + eventName;
+        node = mEventMapping.get(key);
+      }
```
jakub-gonet, line 407: Shouldn't `.resolveCustomEventName()` take care of that? If events are registered in `ModuleConstants` then they should be converted. Looks like a hacky way to do it.

ShaMan123: You are right - this is a bit hacky. This was an effort to make `TextInput` events work with reanimated if I recall correctly. I guess submitting a PR to `react-native` is the right approach. I think it was `onBlur` event that wasn't supported because it was exported as `topBlur`

src/createAnimatedComponent.js
ShaMan123: I had an app run with this. Why wouldn't it? If I recall correctly this is how event registration is done.

android/src/main/java/com/swmansion/reanimated/NodesManager.java
ShaMan123: I can revert commit 76b0dd1.

jakub-gonet: If it doesn't break anything I'd like to get this removed.

src/createAnimatedComponent.js
jakub-gonet: I will test that because `this` here should be a composite component which theoretically doesn't provide this method 😄

ShaMan123: It's exposed [here](https://github.com/software-mansion/react-native-reanimated/blob/76b0dd15e367b9dd09f0bc2d8a3a88d5d076bf18/src/createAnimatedComponent.js#L45)

android/src/main/java/com/swmansion/reanimated/NodesManager.java
ShaMan123: Done  0a9c204
FYI `TextInput` doesn't export event names properly. See [reference](https://github.com/facebook/react-native/blob/065fbe3be53e156fe4f2dbf47791be53b340096f/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java#L151).

src/createAnimatedComponent.js
jakub-gonet: Ah, wrong `this`.

I meant that `this._component.setNativeProps(props);` may fail.

Add definition of the wrapper:

```jsx
class TextWrapper extends Component {
  constructor(props) {
    super(props);
    const {text, ...rest} = props;
    this.text = text;
    this.rest = rest;
  }
  render() {
    return <Text {...this.rest}>{this.text}</Text>;
  }
}
const AnimTextWrapper = createAnimatedComponent(TextWrapper);
```

and replace one of the `Text`s by it:

```jsx
  <AnimTextWrapper
          onLayout={this._layoutEvent}
          style={[
            styles.box,
            {
              width: this._trans,
              height: this._trans,
            },
          ]}
          text={`
          sample text is getting bigger and bigger more and moar staph staph
          stophhh`}
        />
```

You can see that `_attachNativeEvents` will fail.

image


Probably the best solution is to extract logic from `_animatedPropsCallback` to function and call it `_setNativePropsOrRerender` or sth and use it instead `_setNativeProps`. Works fine on this example, don't know if it can introduce some edge cases which will cause infinite rerenders, worth checking out.

jakub-gonet:
@ShaMan123 🏓

jakub-gonet:
@ShaMan123, do you plan on working on this PR? I think it's almost ready so I can finish it if you don't have time to do that.

ShaMan123:
Currently I don't have the time so be my guest

בתאריך יום ה׳, 31 בדצמ׳ 2020, 14:49, מאת Jakub ‏<notifications@github.com>:

piaskowyk:
Time flows fast 🚀 2 years. We no longer support Reanimated v1, so we won't finish this PR. I am closing it, but if anyone wants to finish it, you are welcome - I will reopen it.

PR #531, opened 2019-12-31
Fix issue with some events not propagating correctly via Animated.event

kmagiera:
The issue has been discovered while testing onScroll listener with react-native-webview. After investigating it turned out that there are two problems:
1) WebView was expecting a listener to be assigned to the native component and only with the listener preset the events would be dispatched
2) On iOS we weren't translating event names properly. Some older component still follow "topSth" naming convention while we were solely relying on "onSth" convention for direct event.

In order to address (1) we now assign an empty listener to component properties that have Animated.event object assigned. This way native component knows there is a listener assigned and can start dispatching events that we hijack on the native side.

To fix (2) we matched the orifinal Animated implementation and adapted RCTNormalizeInputEventName utility method from RN core that normalizes event names (it actually changes all names to "topSth" convetion but at least it is now consistent across the board).

After fixing (1) the Android does not seem to need any more changes.

### DIFF
--- a/ios/REANodesManager.m
+++ b/ios/REANodesManager.m
@@ -323,7 +323,9 @@
   REANode *eventNode = _nodes[eventNodeID];
   RCTAssert([eventNode isKindOfClass:[REAEventNode class]], @"Event node is of an invalid type");
 
-  NSString *key = [NSString stringWithFormat:@"%@%@", viewTag, eventName];
+  NSString *key = [NSString stringWithFormat:@"%@%@",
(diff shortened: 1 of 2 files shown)


panda0603:
I still can't seem to get updated animated value for onScroll prop in WebView using event method. 
Animated value doesn't seem to get updated upon scroll.
May this be a separate issue?

karltaylor:
I'm having the same problem as @panda0603

PR #543, opened 2020-01-13
Add displayName to createAnimatedComponent

lindesvard:
Would be nice to add displayName for the createAnimatedComponent.


osdnk:
Thankssss

PR #547, opened 2020-01-14
[web] Added support for callID

EvanBacon:
## Why

- fixes "test" example - all boxes animate across the screen
- fixes "interpolate" example - now boxes don't jitter when you drag them
- fixes "colors" example - now colors interpolate smoothly rather than snap to highest RGB
- fixes "image viewer" example - no more jittering when the user stops panning
- fixes snapping to edges in "chat heads" example - doesn't fix the bug where other chat heads don't follow main one
- changes "different spring configs" to go from jittery circles to not moving circles (¯\\_(ツ)\_/¯)
- doesn't fix any of the interactable examples

## How

Adding more JS support for the features added by this PR https://github.com/software-mansion/react-native-reanimated/pull/354 we should be more careful in the future not to add native-only features.

## Issue:

Screen Shot 2020-01-14 at 3 46 32 PM

### DIFF
--- a/src/core/AnimatedCallFunc.js
+++ b/src/core/AnimatedCallFunc.js
@@ -1,9 +1,10 @@
-import AnimatedNode from './AnimatedNode';
+import AnimatedNode, { getCallID, setCallID } from './AnimatedNode';
 import { adapt } from './AnimatedBlock';
 import { val } from '../val';
 import invariant from 'fbjs/lib/invariant';
 
 class AnimatedCallFunc extends AnimatedNode {
+  _previousCallID;
   _what;
   _args;
   _params;
@@ -39,8 +40,11 @@ class AnimatedCallFunc extends AnimatedNode {
   }
 
   beginContext() {
+    this._previousCallID = getCallID();
+    setCallID(getCallID() + '/' + this.__nodeID);
+
     this._params.forEach((param, index) => {
-      param.beginContext(this._args[index]);
+      param.beginContext(this._args[index], this._previousCallID);
     });
   }
 
@@ -48,6 +52,7 @@ class AnimatedCallFunc extends AnimatedNode {
     this._params.forEach((param, index) => {
       param.endContext();
     });
+    setCallID(this._previousCallID);
   }
 
   __onEvaluate() {
--- a/src/core/AnimatedNode.js
+++ b/src/core/AnimatedNode.js
@@ -5,6 +5,16 @@ const UPDATED_NODES = [];
 
 let loopID = 1;
 let propUpdatesEnqueued = null;
+let nodeCount = 0;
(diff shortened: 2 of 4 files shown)


src/core/AnimatedCallFunc.js
```diff
@@ -39,15 +40,19 @@ class AnimatedCallFunc extends AnimatedNode {
...
-      param.beginContext(this._args[index]);
+      param.beginContext(this._args[index], this._previousCallID);
     });
   }
 
   endContext() {
     this._params.forEach((param, index) => {
       param.endContext();
     });
+    setCallID(this._previousCallID);
```
osdnk: just wondering if `setCallID(getCallID.split('/').slice(0, -1).join('/'))` is not enough? I know I used it in native code, but maybe it's worth refactoring

src/core/AnimatedNode.js
```diff
@@ -52,9 +62,14 @@ function runPropUpdates() {
   loopID += 1;
 }
 
-let nodeCount = 0;
-
 export default class AnimatedNode {
+
+  /* { [loop: string]: int } */
```
osdnk: nit: this comment does not tell me anything. Can you use some words if you want to explain it? If not, just drop

src/core/AnimatedValue.js
```diff
@@ -23,7 +23,7 @@ export default class AnimatedValue extends InternalAnimatedValue {
   }
   
   toString() {
-    return `AnimatedValue, id: ${super.__nodeID}`;
+    return `AnimatedValue, id: ${this.__nodeID}`;
```
osdnk: interesting. Don't we use `super.__nodeID` in more places?

EvanBacon: For some reason, using `super.__nodeID` returns `undefined` here, this fixes it.

src/core/AnimatedCallFunc.js
osdnk: I think I made it like these because keys of dictionaries in ObjC are objects not actually texts and then I followed up with Java impl in the same way.

src/core/AnimatedValue.js
osdnk: ¯\_(ツ)_/¯

osdnk: What's the reason?

src/core/AnimatedCallFunc.js
EvanBacon: I'm attempting to make the JS logic match the native logic exactly. Makes it easier to spot bugs. Would probably be worth updating it everywhere in another PR.

osdnk: sure

PR #548, opened 2020-01-15
Fix case where `new Animated.Value()` is NaN on web

EvanBacon:
In the `different spring config's` example, the value `const transX = new Value();` resolves to NaN on web, but `null` (and then a proper number) on native. Perform the `undefined -> null` conversion upfront so all platforms function as expected.

### DIFF
--- a/src/core/InternalAnimatedValue.js
+++ b/src/core/InternalAnimatedValue.js
@@ -32,6 +32,9 @@ export default class InternalAnimatedValue extends AnimatedNode {
   }
 
   constructor(value, constant = false) {
+    if (value === undefined) {
+      value = null;
+    }
     super({ type: 'value', value: sanitizeValue(value) });
     this._startingValue = this._value = value;
     this._animation = null;


osdnk:
[web] Fix case where `new Animated.Value()` is `undefined` on web?

src/core/InternalAnimatedValue.js
```diff
@@ -32,6 +32,9 @@ export default class InternalAnimatedValue extends AnimatedNode {
   }
 
   constructor(value, constant = false) {
+    if (value === undefined) {
```
osdnk: maybe?
```suggestion
    if (value == undefined) {
```

osdnk: nope, 0 is okay

EvanBacon: sure

osdnk: okay, whatever. I think NaN is handled natively in another way, so let's leave it

Issue #558, opened 2020-01-16
v1.6 Require cycle


blueberry6401:
Got require cycle in the latest version (1.6).
```node_modules/react-native-reanimated/src/core/AnimatedClock.js -> node_modules/react-native-reanimated/src/core/AnimatedValue.js -> node_modules/react-native-reanimated/src/derived/interpolate.js -> node_modules/react-native-reanimated/src/core/AnimatedFunction.js -> node_modules/react-native-reanimated/src/core/AnimatedParam.js -> node_modules/react-native-reanimated/src/core/AnimatedClock.js```

Issue #561, opened 2020-01-16
☔️ Regression in 1.6.0 / 1.5.0 


osdnk:
I have released a 1.6.0 version yesterday. It was containing a lot of small changes and updates and unfortunately brought few regressions. I want to get rid of them as soon as I can collecting feedback from the community.

Here's a list of issues I'm conscious about: 

1. https://github.com/software-mansion/react-native-reanimated/issues/551 fixed  by https://github.com/software-mansion/react-native-reanimated/issues/551
2. #559 fixed by https://github.com/software-mansion/react-native-reanimated/pull/560
3. https://github.com/software-mansion/react-native-reanimated/issues/557#issuecomment-575009400 fixed by https://github.com/software-mansion/react-native-reanimated/pull/566
4. https://github.com/software-mansion/react-native-reanimated/issues/558 fixed by #562
5. https://github.com/software-mansion/react-native-reanimated/issues/555 fixed by (#568) 
6. https://github.com/software-mansion/react-native-reanimated/issues/563 fixed by https://github.com/software-mansion/react-native-reanimated/pull/564

Also, I want to include https://github.com/software-mansion/react-native-reanimated/pull/556 in order to verify reanimated with a complex example

Jarred-Sumner:
Flagging this comment incase it was missed: https://github.com/software-mansion/react-native-reanimated/commit/bdc7b5962a8fec49230446a21c1f739c98b8effa#commitcomment-36633385

It’s probably too vague to be actionable though

osdnk:
All issues should be gone

PR #562, opened 2020-01-16
Remove cycle of requires in AnimatedParam

osdnk:
Fixes https://github.com/software-mansion/react-native-reanimated/issues/558
rel: https://github.com/software-mansion/react-native-reanimated/issues/561
Regression made with: #534 

In this PR I get rid of the cycle of requires which cause warnings by fixing one of the imports


osdnk:
Actually `AnimatedMainClock` is not used anywhere so it cannot be breaking.

osdnk:
rel https://github.com/software-mansion/react-native-reanimated/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc

Issue #557, opened 2020-01-16
Malformed calls from JS: field sizes are different.


computerjazz:
I'm the author of [react-native-draggable-flatlist](https://github.com/computerjazz/react-native-draggable-flatlist), which uses react-native-reanimated + RNGH under the hood, and users who have upgraded to the latest reanimated release are reporting the following error, which I have independently verified is new as of Reanimated 1.5.0:

```
ExceptionsManager.js:126 Malformed calls from JS: field sizes are different.

[...]

PR #566, opened 2020-01-16
Fallback to old mechanism of setting value if value is a node or string

osdnk:
The current implementation of setValue is not accepting value as a node so for now, I'd like to revert to the old mechanism in this case in order to limit a regression


osdnk:
Fixes https://github.com/software-mansion/react-native-reanimated/issues/557

janicduplessis:
Tested that it fixes the regression in my app

PR #574, opened 2020-01-19
Update react-native-reanimated.d.ts

iyegoroff:
Fix for typescript@3.7.5 compiler error `error TS1254: A 'const' initializer in an ambient context must be a string or numeric literal or literal enum reference.`


CesarLanderos:
🙏

wcandillon:
Thank you @iyegoroff. Tested it and it works great.

PR #583, opened 2020-01-23
[v1] Fix animated value not set if not attached

kruczy:
Fixes an edge case where if an `Animated.Value` is not attached yet and `setValue` is called that value will not be set since the node does not exist on the native side.

For me this situation happens during first render where some values are calculated from props inside a `useMemo`

reading the code it also makes me believe that if a value node is detached and reattached it would reset its value to the initial value? is that a bug, or feature? if this is relied upon this might be a breaking change

Let me know what you think 👍


jakub-gonet:
Hi @kruczy,

this is actually a feature — you may want to animate some component every time it's mounted or unmounted and animation should reset in that case.

jakub-gonet:
Closing due to no response. Feel free to ping me, I'll reopen it then.

PR #597, opened 2020-02-06
Update Mock to support the new import style

wcandillon:


wcandillon:
Is it possible that the CI failure is unrelated to the proposed change?

PR #599, opened 2020-02-07
Hide 'Nullable' declaration from global scope

mokoshi:
`Nullable` should be inside the namespace in order to avoid `error TS2300: Duplicate identifier` error which may occur if we already have it within our own projects.


jakub-gonet:
Moved to #685.

PR #601, opened 2020-02-10
💄Round decimal values in color()

wcandillon:


jakub-gonet:
Hey @wcandillon,
thanks for a contribution!

Could you please describe intent behind this PR?

wcandillon:
Oups sorry about that. A color can only be a integer value. If you return decimal values, you might get strange result. Maybe this should be done outside the `color` node? I don't have a strong opinon if you think that this logic should be kept outside.

jakub-gonet:
Keeping this here seems reasonable. What with alpha? Shouldn't it be also rounded,  as it's also stored with offset?

wcandillon:
alpha is already rounded at https://github.com/software-mansion/react-native-reanimated/blob/7964904c5b20e10e30e3c9665fa985c0e43dd2d9/src/derived/color.js#L25

jakub-gonet:
Ah, my bad, should've checked.

All right, merging then 🎉

PR #606, opened 2020-02-12
Transition Api Alternative (Proposal)

### LINKED ISSUE #328: Can't change transition sequence once initialized
Hi,
I've noticed that once a Transition.view is initialized, we can't update transition to new transition style. Here is an Example code to replicate the issue 

```
import React, { Component } from 'react';
import { Transitioning, Transition } from 'react-native-reanimated';

const transitionSlideIn = (
    <Transition.Sequence>
        <Transition.Out type="slide-left" />
        <Transition.Change interpolation="easeInOut" />
        <Transition.In type="slide-right" />
    </Transition.Sequence>
);

const transitionSlideOut = (
    <Transition.Sequence>
        <Transition.Out type="slide-right" />
        <Transition.Change interpolation="easeInOut" />
        <Transition.In type="slide-left" />
    </Transition.Sequence>
);
class Example extends Component {
    constructor(props) {
        super(props);
        this.state = {
            isSlideIn: true
        };
        this.ref = React.createRef();
    }

    componentDidUpdate(prevProps) {
        if (this.props.isSlideIn != prevProps.isSlideIn) {
            this.setState({ isSlideIn: isSlideIn }, () => { this.ref.current.animateNextTransition(); })
        }
    }

    render() {
        return (
            <Transitioning.View
                transition={this.state.isSlideIn ? transitionSlideIn : transitionSlideOut}
                style={{ flex: 1 }}>
                {this.props.children}
            </Transitioning.View>
        )
    }
}

export default Example;
```

adkenyon:
Originally, my goal was to fix #328. After a lot of fiddling around around with the Transition Api I landed on instead passing the desired transition config directly through `animateNextTransition` rather than using the `Transition` and `Transitioning.View` `transition` prop. While this doesn't necessarily fix #328 it does provide a working alternative.

Feedback would be great! Thoughts on other strategies to fix #328 would also be appreciated. 

```
function component() {
  const [state, setState] = React.useState(false);
  const viewRef = Reac.useRef()

  return (
    <>
      <Button onPress={() =>
         viewRef.current.animateNextTransition(
           TransitionApi.Together([
              TransitionApi.Out({ type: 'fade', durationMs: 400, interpolation: 'easeIn' }),
              TransitionApi.Change(),
              TransitionApi.In({ type: 'slide-right', durationMs: 200, interpolation: 'linear' }),
           ])
         )
        }
      />
      <Transitioning.View
        ref={viewRef}
      >
        {state ? <Thing1/> : <Thing2/>}
      </Transitioning.View>
    </>
    
  )
}

```

### DIFF
--- a/Example/src/transitions/ticket.js
+++ b/Example/src/transitions/ticket.js
@@ -1,13 +1,8 @@
 import React, { useState, useRef } from 'react';
-import { Text, View, StyleSheet, Button, StatusBar } from 'react-native';
-import { Transitioning, Transition } from 'react-native-reanimated';
+import { Text, View, Button } from 'react-native';
+import { Transitioning, TransitionApi } from 'react-native-reanimated';
 
-function shuffleArray(array) {
-  for (let i = array.length - 1; i > 0; i--) {
-    const j = Math.floor(Math.random() * (i + 1));
-    [array[i], array[j]] = [array[j], array[i]];
-  }
-}
+const { Sequence, Together, Out, In, Change } = TransitionApi;
 
 const Hour = ({ hour, min, pm }) => (
   <View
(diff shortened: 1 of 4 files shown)


adkenyon:
@kmagiera && @osdnk

djaoui:
Hi. In case my feedback helps, this added functionnality will be really useful and i have a use case for it right now that's relevant in real-world physics . When navigating forward, I want to implement one type of animation (in:slide-left) and I want a different one when navigation backward (in:slide-right). Cheers for that

adkenyon:
@djaoui Thanks for the feedback, that was the same implementation that led me here as well

djaoui:
Thanks for letting me know.

Is it already implemented ? I mean Can I use it ?

Not sure what install command to run to get the updated package (If available)?

Copied the transitioning.js from your repo but had no luck reproducing your example.

Could you help me ? Much appreciated.

Cheers !

David Jaoui

adkenyon:
Hi David,

If you're feeling bold you're welcome to use the branch on my fork with the changes. Everything as seen in the example should be working. You can install the package with either of the following

`npm install react-native-reanimated@adkenyon/react-native-reanimated#5e8ca40c2b12e8a89f68f8e24d498b9b395635e6`
or
`yarn add react-native-reanimated@adkenyon/react-native-reanimated#5e8ca40c2b12e8a89f68f8e24d498b9b395635e6`

Thanks,
Alan

djaoui:
Thanks you very much Alan, I'll give it a try.

Cheers,

David

jakub-gonet:
Hi, thanks for your PR, we really appreciate it.

We discussed Transitions internally in our team. Considering that v2 makes creating animations easier and Transitions was an experimental API for a long time, we decided to leave Transitions as they are.

This PR introduces fairly big change in API and I'm afraid we cannot maintain it.

I’m closing it, but if you have further questions, I will be happy to help.

adkenyon:
@jakub-gonet, Does this mean the Transition API is going away in v2?

jakub-gonet:
No, everything we had in v1 should be working in v2. We don't want to introduce big changes to it (like this PR) and only bug fixes. We're considering using v2 to implement some helpers which will allow making animations from simple blocks, replacing Transitions with that.

PR #607, opened 2020-02-12
Bump OSX to 11.3 on Travis

jakub-gonet:


jakub-gonet:
Tried to fix
```
/Users/travis/build/software-mansion/react-native-reanimated/Example/ios/Pods/Headers/Public/React-Core/React/RCTLinkingManager.h:10:9: fatal error: 'UIKit/UIUserActivity.h' file not found
#import <UIKit/UIUserActivity.h>
```
(https://developer.apple.com/documentation/uikit/uiactivity which is supported since SDK 13)
by bumping IOS version.

Travis doesn't support SDK 13 yet, so I'm closing this.

PR #608, opened 2020-02-12
@szymon/turbo version

osdnk:


Jarred-Sumner:
wow this is cool!

I took a swing at getting `react-native-reanimated` to work on iOS via JSI a few days ago (without turbomodules): https://github.com/Jarred-Sumner/react-native-reanimated/blob/yeet/ios/REAJSIModule.mm. It didn't really work well though because the nodes expect the views to be attached and the JSI stuff can happen before views or props are attached/updated.  I think @chrfalch's attempt was better though: https://github.com/chrfalch/react-native-reanimated/blob/jsi-methods/ios/REAJsiModule.mm

@terrysahaidak also has a branch that also uses JSI on Android: https://github.com/terrysahaidak/react-native-reanimated/blob/tr-nodes/jsi/android/ReanimatedJSI.cpp

terrysahaidak:
This is something from an another dimension!! Sooo coool!!

Is this going to run on a separate thread (maybe another instance of the Hermes) or on the same but async?

Szymon20000:
> Is this going to run on a separate thread (maybe another instance of the Hermes) or on the same but async?

Yes, It will run on UI thread and new runtime instance.

terrysahaidak:
Have you thought already about some babel-macros in order to avoid that get/set boilerplate? 

Or even also creating and releasing the SharedValues?

Issue #614, opened 2020-02-18
TypeError: Cannot read property ' __value' of undefined


GioCirque:
Testing under jest, this line throws an error calling from`react-navigation-drawer`. This test executes via the mock as expected in version `1.4.0` and then it breaks in `1.5.0` and later.

https://github.com/software-mansion/react-native-reanimated/blob/33f827311f035eaed430d82b057edc6e00781ec7/mock.js#L26

```
    TypeError: Cannot read property ' __value' of undefined

      15 |   beforeEach(async () => {
      16 |     store = mockStore({ metricsState: initialMetricsState, preferencesState: initialPreferencesState });
    > 17 |     container = mount(
         |                 ^
      18 |       <Provider store={store}>
      19 |         <RootContainer />
      20 |       </Provider>,

      at getValue (node_modules/react-native-reanimated/mock.js:26:14)
      at getValue (node_modules/react-native-reanimated/mock.js:105:11)
      at new cond (node_modules/react-navigation-drawer/lib/commonjs/views/Drawer.tsx:325:9)
      at constructClassInstance (node_modules/react-dom/cjs/react-dom.development.js:14204:18)
      at updateClassComponent (node_modules/react-dom/cjs/react-dom.development.js:18413:5)
      at beginWork$1 (node_modules/react-dom/cjs/react-dom.development.js:20186:16)
      at HTMLUnknownElement.callCallback (node_modules/react-dom/cjs/react-dom.development.js:336:14)
      at Object.invokeGuardedCallbackDev (node_modules/react-dom/cjs/react-dom.development.js:385:16)
      at invokeGuardedCallback (node_modules/react-dom/cjs/react-dom.development.js:440:31)
      at beginWork$$1 (node_modules/react-dom/cjs/react-dom.development.js:25780:7)
      at performUnitOfWork (node_modules/react-dom/cjs/react-dom.development.js:24698:12)
      at workLoopSync (node_modules/react-dom/cjs/react-dom.development.js:24671:22)
      at performSyncWorkOnRoot (node_modules/react-dom/cjs/react-dom.development.js:24270:11)
      at scheduleUpdateOnFiber (node_modules/react-dom/cjs/react-dom.development.js:23698:7)
      at updateContainer (node_modules/react-dom/cjs/react-dom.development.js:27103:3)
      at node_modules/react-dom/cjs/react-dom.development.js:27528:7
[...]

PR #622, opened 2020-02-25
fix TypeError: Cannot read property ' __value' of undefined

andrescalco:
Was facing some errors while testing with jest and found this quick fix.

### DIFF
--- a/mock.js
+++ b/mock.js
@@ -23,7 +23,7 @@ const getValue = node => {
   if (typeof node === "number") {
     return node;
   }
-  return node[" __value"];
+  return node && node[" __value"];
 };
 
 class AnimatedValue {


andrescalco:
Kudos to [adrianolsk](https://github.com/adrianolsk) on helping me with this

GioCirque:
This should resolve #614 and #622

jakub-gonet:
Isn't that treating symptoms instead of a cause? In what cases `getValue` should return `undefined`?

Thomazella:
```
Isn't that treating symptoms instead of a cause? In what cases getValue should return undefined?
```
☝️ mocks should follow the types. That's the cause of the bug I think.

balzdur:
I agree with @Thomazella.

I think the approach described here is better https://github.com/lucianomlima/react-navigation-v4-examples/pull/1

PR #624, opened 2020-02-27
Turbo version babel plugin

karolbisztyga-swmansion:


babel-plugin.js
```diff
@@ -14,19 +14,23 @@ module.exports = function ({ types: t }) {
             /\/\*[\s\S]*?\*\/|([^\\:]|^)\/\/.*$/gm, '');
           str = str.replace(/\n/ig, '');
           let newCode = ""
-          if (fun.node.type === 'FunctionExpression') {
+          let type = fun.node.type
+          if (type === 'ArrowFunctionExpression') {
+            str = ('function' + str).replace('=> {', '{')
+            type = 'FunctionDeclaration'
```
Szymon20000: That's incorrect because arrow functions don't have names. Use 'FunctionExpression' instead.

PR #625, opened 2020-02-27
Add release-it

jakub-gonet:
Adding release-it to make releases easier.
Added a custom check for changes in ios/ and android/ folders, which prevents releasing patches when minor ver. should apply. Unfortunately, release-it doesn't support hook return codes, so it crashes :/
Hopefully, the comment above check is sufficient to know what went wrong.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "test": "npm run format && npm run lint && npm run test:unit",
     "test:unit": "jest",
     "format": "prettier --write --list-different './src/**/*.js'",
-    "lint": "eslint './src/**/*.js'"
+    "lint": "eslint './src/**/*.js'",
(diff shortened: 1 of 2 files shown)


kmagiera:
nice! Can you add a description to this PR to document what is the expected output of the check hook and what happens when the check fails?

kmagiera:
^ this is just so that we know what the expected behavior was in case we find issue with it

jakub-gonet:
Ah, my bad, will do.

osdnk:
looks good. Can we merge it? How about making release notes? Is there anything blocking?

jakub-gonet:
Release notes are still made manually, so I didn't configure that. This PR can be merged as soon as it gets approval.

PR #627, opened 2020-03-02
shared boolean works

karol-bisztyga:


android/src/main/cpp/NativeProxy.cpp
```diff
@@ -135,6 +135,12 @@ jobject getChangedSharedValues(JNIEnv* env) {
         y = env->NewStringUTF(str.c_str());
         break;
       }
+      case 'B':
+      {
+        bool val = ((SharedBoolean*)(sv.get()))->value;
+        y = env->CallStaticObjectMethod(integerClass, integerValueOf, val);
```
karolbisztyga-swmansion: is it ok to use integer class here?

Common/cpp/SharedBoolean.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+        auto callback = [this](
+          jsi::Runtime &runtime,
+          const jsi::Value &thisValue,
+          const jsi::Value *arguments,
+          size_t count
+        ) -> jsi::Value {
+          (*dirty) = true;
+          //double newValue = arguments[0].asNumber();
+          bool newValue = (bool)arguments[0].asNumber();
```
Szymon20000: getBool()?

PR #642, opened 2020-03-11
Turbo version cleanup test

karol-bisztyga:


Common/cpp/NativeReanimatedModule.cpp
```diff
@@ -206,6 +206,51 @@ void NativeReanimatedModule::onEvent(std::string eventName, std::string eventAsS
...
+  //  2 - worklets
+  //  3 - appliers
+  jsi::Function fun = value.getObject(rt).asFunction(rt);
+  std::shared_ptr<jsi::Function> funPtr(new jsi::Function(std::move(fun)));
+
+  scheduler->scheduleOnUI([&rt, funPtr, this, option]() {
+    std::string ids;
+    switch(option) {
+      case 1: {
+        for(auto it = sharedValueRegistry->sharedValueMap.begin(); it != sharedValueRegistry->sharedValueMap.end(); ++it) {
```
Szymon20000: Can you use foreach loop?

Common/cpp/headers/ApplierRegistry.h
```diff
@@ -11,10 +11,10 @@
 #include <unordered_map>
 
 class ApplierRegistry {
-  std::unordered_map<int, std::shared_ptr<Applier>> renderAppliers;
-  std::map<std::string, std::unordered_map<int, std::shared_ptr<Applier>>> eventAppliers;
-  std::map<int, std::string> eventMapping;
   public:
+    std::unordered_map<int, std::shared_ptr<Applier>> renderAppliers;
```
Szymon20000: Use getters/setters instead.

Example/testComponents/CleanUpTest.js
```diff
@@ -0,0 +1,228 @@
+import React from 'react';
+import Animated, { useSharedValue, useWorklet, RegistersState } from 'react-native-reanimated';
+import { View, Text } from 'react-native';
+
+const workletBody = function(a, b, c) {
```
Szymon20000: If you don't create Worklet instance here then each Event Worklet will register separate Worklet. Is this really what you want?

src/reanimated2/RegistersState.js
```diff
@@ -0,0 +1,40 @@
+import NativeModule from './NativeReanimated';
+
+
+export default class RegistersState {
```
Szymon20000: Can you export instance instead of class?

Common/cpp/NativeReanimatedModule.cpp
karol-bisztyga: fixed 70dd9c2b456dc5edb71581e9c70aded6a060f804

Common/cpp/headers/ApplierRegistry.h
karol-bisztyga: fixed 70dd9c2b456dc5edb71581e9c70aded6a060f804

Example/testComponents/CleanUpTest.js
karol-bisztyga: yes this is it exactly

src/reanimated2/RegistersState.js
karol-bisztyga: fixed d84f2ab0bb3275696ef078668869069e14f1f7b9

ios/native/NativeProxy.mm
```diff
@@ -93,7 +93,7 @@ + (void*) getNativeReanimatedModule:(void*)jsInvokerVoidPtr
 + (NSArray<NSArray*>*)getChangedSharedValues
 {
   NSMutableArray *changed = [NSMutableArray new];
-  for(auto & sharedValue : nativeReanimatedModule->sharedValueRegistry->sharedValueMap) {
+  for(auto & sharedValue : nativeReanimatedModule->sharedValueRegistry->getSharedValueMap()) {
```
Szymon20000: Shouldn't we make the same change in NativeProxy.cpp?

karol-bisztyga: yes, fixed 688b7e4b5f7734a1a7a6bc47b735250fcd7614d1

PR #666, opened 2020-03-20
Turbo version worklet start stop

karol-bisztyga:


Common/cpp/SharedWorkletStarter.cpp
```diff
@@ -26,7 +35,93 @@ SharedWorkletStarter::~SharedWorkletStarter() {
 }
 
 jsi::Value SharedWorkletStarter::asParameter(jsi::Runtime &rt) {
-  return jsi::Value(this->id);
+  
+  class HO : public jsi::HostObject {
+    public:
+    int id;
+    SharedWorkletStarter *starter;
```
Szymon20000: Why not std::shared_ptr ?

Common/cpp/SharedWorkletStarter.cpp
```diff
@@ -26,7 +35,93 @@ SharedWorkletStarter::~SharedWorkletStarter() {
...
+            const jsi::Value *args,
+            size_t count
+        ) -> jsi::Value {
+          if (starter->unregisterListener != nullptr) {
+            (*starter->unregisterListener)();
+            starter->unregisterListener = nullptr;
+          }
+          return jsi::Value::undefined();
+        };
+        return jsi::Function::createFromHostFunction(rt, name, 1, callback);
```
Szymon20000: Start and Stop don't have any args.

Common/cpp/WorkletModule.cpp
```diff
@@ -129,4 +76,4 @@ void WorkletModule::setApplierId(int applierId) {
   this->applierId = applierId;
 }
 
-int WorkletModule::applierId = INT_MAX;
+int WorkletModule::Applier_Id = INT_MAX;
```
Szymon20000: Maybe we should move it to the ApplierRegistry? 
New_Local_Applier_Id?

Common/cpp/SharedWorkletStarter.cpp
karol-bisztyga: because `this` is passed down to the inner class

Common/cpp/WorkletModule.cpp
karol-bisztyga: fixed 957dc94

Common/cpp/SharedWorkletStarter.cpp
karol-bisztyga: fixed 957dc94

PR #675, opened 2020-03-22
fix: prevent initial flickering in browsers

jaulz:
I noticed that animated `View`s initially flicker because there is a minimal delay between when the element is rendered and when the animated styles are applied. This fix uses the current values of the animated styles for the initial rendering and thus avoids flickering.

### DIFF
--- a/src/core/AnimatedClock.js
+++ b/src/core/AnimatedClock.js
@@ -6,7 +6,7 @@ class AnimatedMainClock extends InternalAnimatedValue {
   _frameCallback;
 
   constructor() {
-    super();
+    super(Date.now());
   }
 
   __onEvaluate() {
--- a/src/core/AnimatedClock.js
+++ b/src/core/AnimatedClock.js
@@ -6,11 +6,11 @@ class AnimatedMainClock extends InternalAnimatedValue {
   _frameCallback;
 
   constructor() {
-    super({ type: 'MAIN_CLOCK' });
+    super(Date.now());
   }
 
   __onEvaluate() {
-    return +new Date();
+    return Date.now();
   }
 
   _runFrame = () => {
--- a/src/core/AnimatedClock.js
+++ b/src/core/AnimatedClock.js
@@ -6,7 +6,7 @@ class AnimatedMainClock extends InternalAnimatedValue {
   _frameCallback;
 
   constructor() {
-    super(Date.now());
+    super();
   }
 
   __onEvaluate() {
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -191,10 +191,12 @@ export default function createAnimatedComponent(Component) {
       if (c !== this._component) {
         this._component = c;
 
-        // Set initial animated styles to prevent minimal flickering
-        if (this._component && this._component.setNativeProps) {
+        // Set initial animated styles to prevent minimal flickering in browsers
+        if (Platform.OS === 'web' && this._component && this._component.setNativeProps) {
+          const initialStyle = this._getInitialStyle(StyleSheet.flatten(this.props.style))
+
           this._component.setNativeProps({
-            style: this._getInitialStyle(StyleSheet.flatten(this.props.style))
+            style: initialStyle,
           });
         }
       }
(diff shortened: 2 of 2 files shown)


jaulz:
I added a semaphore that is set to `true` as long as no animated node is attached and that indicates if also animated styles should be applied during rendering.

jfrolich:
Very nice, however I get an error "Field sizes are different" when I try out this PR.

jaulz:
@jfrolich unfortunately, I cannot reproduce your issue. Could you please create a repository to reproduce it or maybe even debug on your side? Though, I also have to admit that I was not yet able to run it on a device (something is wrong with the project setup and I need to dive deeper into it). However, in the browser it should work without flickering.

jaulz:
@jfrolich could you check my latest commit? I think previously only `AnimatedValue`s were handled but conditions etc. were ignored.

jfrolich:
Still the same result
image

jfrolich:
It looks like it crashes on interpolations, normal evaluations work well. If I call `__getValue()` on an `AnimatedNode` even if I don't include it in the style but just log it to the console it gives this error later on. Same if I call `onEvaluate`.

image

jaulz:
Hm, unfortunately, I'm not too familiar with the internals of this library. @kmagiera @osdnk could you maybe give a quick feedback what could go wrong here?

jfrolich:
I think it `__getValue` doesn't work initially on other node types than AnimatedValue. So if you add this check `value.__nodeConfig.type === "value"` in `_filterAnimatedStyle` then it doesn't crash anymore.

jaulz:
@jfrolich I think something collided between React render phase and setting native props. Could you please check my recent commit? Basically, now we just set the native props immediately as soon as we have access to the component und thus we do not interfere with the render phase.

jfrolich:
Hi, it doesn't work on other node types than a vanilla AnimatedValue (try an interpolation and it raises an error).

jaulz:
@jfrolich it seems the main clock has a wrong constructor parameter which is resolved to NaN when it is passed to the bridge:
https://github.com/software-mansion/react-native-reanimated/blob/66c0e08791d9ce577c579e84f09a0f60d1e2db28/src/core/AnimatedClock.js#L9

Though this fixes the red error screen I see some flickering afterwards on native devices. In browser though, the initial flickering is fixed and it still works as expected. I tried to dive deeper but could not find the root cause. Thus I added a platform check here:
https://github.com/software-mansion/react-native-reanimated/pull/675/files#diff-3be87c4c6b24297654a7e0b2990508f2R195

jfrolich:
Did you try interpolated values, and values resulting from nodes like `cond`?

jaulz:
Yes, I did.

jfrolich:
Ok so probably that only works on Web, to make it work on native you have to check for `instanceof AnimatedValue` and then it works. You also have to make sure that you don't return `{}` so check if the object is empty and return undefined if so. Also filter arrays for undefined values and then it should work. Only it doesn't work for interpolated values. I managed it to get it to work by manually setting an `__Custom_InitialValue` property on the node and checking or that to fix some of my UI.

jaulz:
Even interpolated values work for me natively but as I said it causes some flickering afterwards as it seems that the values are set back and forth.

jfrolich:
> Even interpolated values work for me natively but as I said it causes some flickering afterwards as it seems that the values are set back and forth.

Did you also try different reanimated operators. I don't really know where the point of failure is but it definitely raises in more complex scenarios.

jaulz:
I tested it in an app that uses `react-navigation` so I would say it also contains some more complex scenarios.

jfrolich:
> I tested it in an app that uses `react-navigation` so I would say it also contains some more complex scenarios.

It works right now because it asserts the 'web' platform. Once you remove it it will fail on native.

jaulz:
Did you also apply the patch to the clock?

jakub-gonet:
Hi @jaulz,

could you provide some small example I can test it on?

jakub-gonet:
Also, check if #1027 don't fix that by any chance.

jakub-gonet:
Closing until the author responds.

PR #676, opened 2020-03-23
Turbo version jni registry

karol-bisztyga:
This code reduces number of JNI searches to one per specific class and one per certain method

### DIFF
--- a/android/src/main/cpp/JNIRegistry.cpp
+++ b/android/src/main/cpp/JNIRegistry.cpp
@@ -16,36 +16,50 @@ std::string JNIRegistry::generateMethodKey(
 std::tuple<jclass, jmethodID> JNIRegistry::getClassAndMethod(
         std::string className,
         std::string methodName,
-        std::string methodSignature) {
-    Logger::log(std::string("[JNIRegistry] HERE Entering getClassAndMethod with args: " + className + "/" + methodName + "/" + methodSignature).c_str());
+        std::string methodSignature,
+        JNIMethodMode methodMode,
+        JNIEnv *currentEnv) {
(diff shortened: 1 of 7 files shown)


android/src/main/cpp/JNIRegistry.cpp
```diff
@@ -0,0 +1,58 @@
...
+    jclass classPtr = nullptr;
+    jmethodID methodPtr = nullptr;
+    // check if class is already stored, if not, try to obtain it
+    auto classIt = this->classes.find(className);
+    if (classIt != this->classes.end()) { // found
+        classPtr = classIt->second;
+    } else { // not found
+        classPtr = currentEnv->FindClass(className.c_str());
+        // register obtained class
+        classPtr = (jclass)currentEnv->NewGlobalRef(classPtr);
```
Szymon20000: I think we also have to delete this ref somewhere.

android/src/main/cpp/headers/JNIRegistry.h
```diff
@@ -0,0 +1,40 @@
...
+
+enum class JNIMethodMode {
+    standard_method,
+    static_method,
+};
+
+class JNIRegistry {
+    JNIEnv* env;
+    // keys are class names
+    std::unordered_map<std::string, jclass> classes;
```
Szymon20000: I think it's better to have all methods and classes that we use in arrays. It's a really important part of Reanimated.

karol-bisztyga: fixed e0fd007

android/src/main/cpp/JNIRegistry.cpp
karol-bisztyga: fixed ac5c74f

PR #679, opened 2020-03-24
Add missing fbjs dependency

satya164:


kmagiera:
@satya164 just published 1.7.1 with this update. Thanks!

PR #686, opened 2020-03-26
useAnimatedStyle

Szymon20000:


src/createAnimatedComponent.js
```diff
@@ -238,7 +240,7 @@ export default function createAnimatedComponent(Component) {
     _filterNonAnimatedProps(inputProps) {
       const props = {};
       for (const key in inputProps) {
-        const value = inputProps[key];
+        let value = inputProps[key];
```
karol-bisztyga: why such change?

Szymon20000: Reverted

Common/cpp/headers/NativeModules/NativeReanimatedModule.h
```diff
@@ -59,13 +59,13 @@ class NativeReanimatedModule : public NativeReanimatedModuleSpec {
     void render();
     void onEvent(std::string eventName, std::string eventAsString);
 
+    std::unique_ptr<jsi::Runtime> runtime;
```
Szymon20000: https://stackoverflow.com/questions/2254263/order-of-member-constructor-and-destructor-calls

PR #687, opened 2020-03-26
Update createEventObjectProxyPolyfill.js

ShaMan123:
App crashes when hermes is enabled (tested on android).
App sometimes reports an error regarding not being able to assign values to a transient object.
This seems to fix the issue


jakub-gonet:
Can you check if app crashes on some commit before #682 was merged?

ShaMan123:
> Can you check if app crashes on some commit before #682 was merged?

@jakub-gonet  Sorry for not replying

jakub-gonet:
@ShaMan123 happens to everyone, don't worry about it 😉

PR #694, opened 2020-03-31
Add global Animated object.

Szymon20000:


Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,77 @@
...
+  public:
+    Animated(std::unordered_map<std::string, jsi::Value> && props) {
+      this->props = std::move(props);
+    }
+
+    jsi::Value get(jsi::Runtime &rt, const jsi::PropNameID &name) {
+      auto propName = name.utf8(rt);
+      
+      if (props.count(propName) > 0) {
+        return jsi::Value(rt,props[propName]);
```
karol-bisztyga: You search the collection twice for the same key. I'd go with something like: `auto it = props.find(propName); if(it != props.end()) return jsi::Value(rt, *it);` (treat it as a pseudocode pls)

Szymon20000: Done

PR #696, opened 2020-04-01
Use component type in createAnimatedComponent

salockhart:
When using `createAnimatedComponent` with Typescript, it would blow away the type of the inner component. This means that JSX code completion and type inference would no longer work.

This solves this by passing back the same type that was input, so that the exported component will have the expected prop types.


jakub-gonet:
Reverting, introduced bug #719 - described in [comment](https://github.com/software-mansion/react-native-reanimated/issues/719#issuecomment-611435860)

PR #698, opened 2020-04-02
Hide leading zeroes on int number in concatNode

### LINKED ISSUE #677: `concat`  issue with Integer Value on Android
Hi, first of all, thanks for the great work.
I'm trying to use the `concat`  function to animate a text using ReText component from `react-native-redash`,
but look like the `concat` function always add ".0" to the end of the animated number node on Android .

My Code   
```
const price = new Value(44);
const MyString = concat('$', price);  
``` 

## Epcpectd Result 
MyString should be `$44`

## Result 

IOS :  `$44`
Android : `$44.0`

Note:  The problem is Not related to `Retext` Component, I tried passing the MyString value to the javascript side using `call` function and noticed the same behavior.

jakub-gonet:
## Description

Fixes #677

When using `concat` node on `Value` which is an integer, Reanimated created a string with leading zeroes on Android and string without leading zeroes on IOS.

This PR fixes this and also fixes the localization of formatted numbers.

## Reproduction Example

```jsx
import React from 'react';
import { SafeAreaView, TextInput } from 'react-native';
const { Value, concat, createAnimatedComponent } = Animated;

const price = new Value(44.5);
const AnimatedTextInput = createAnimatedComponent(TextInput);
const App = () => {
  return (
    <>
      <SafeAreaView>
        <AnimatedTextInput
          underlineColorAndroid="transparent"
          editable={false}
          {...{ text: concat('', price), style: { color: 'black' } }}
        />
      </SafeAreaView>
    </>
  );
};

export default App;
```

Pre-fix Screenshot:
image

Post-fix screenshots (localization with comma on decimals):
image

image

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/nodes/ConcatNode.java
+++ b/android/src/main/java/com/swmansion/reanimated/nodes/ConcatNode.java
@@ -8,6 +8,10 @@ import com.swmansion.reanimated.Utils;
 
 public class ConcatNode extends Node {
   private final int[] mInputIDs;
+  private final static NumberFormat formatter = NumberFormat.getInstance();
+  static {
+    formatter.setMinimumFractionDigits(0);
+  }
 
   public ConcatNode(int nodeID, ReadableMap config, NodesManager nodesManager) {
     super(nodeID, config, nodesManager);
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/reanimated/nodes/ConcatNode.java
```diff
@@ -15,9 +17,16 @@ public ConcatNode(int nodeID, ReadableMap config, NodesManager nodesManager) {
   @Override
   protected String evaluate() {
     StringBuilder builder = new StringBuilder();
+    NumberFormat formatter = NumberFormat.getInstance();
```
kmagiera, line 20: maybe avoid instantiating formatter on each eval. It seem like this could be a static object

jakub-gonet: Moved to a static variable & block

PR #700, opened 2020-04-03
Turbo version bug fix

karol-bisztyga:


Common/cpp/Tools/Applier.cpp
```diff
@@ -35,7 +35,10 @@ bool Applier::apply(jsi::Runtime &rt, std::shared_ptr<BaseWorkletModule> module)
     }
     sharedValues.push_back(sv);
   }
-  
+
+  if (this->errorHandler->getError() != nullptr && this->errorHandler->getError()->handled) {
```
Szymon20000, line 39: Why do we need it here?  It should be enough to check this in NativeProxy.

karol-bisztyga: fixed 3b7b8a91f96d85c44f230db115b2d07c79cbfcd4

karol-bisztyga: ok I checked, this code is needed for iOS, I brought it back in 
3815172
without it on iOS `PreventMultipleEventWorklets` fails in `WorkletFailureTest.js`

Szymon20000: I think we can move it to the NativeProxy.

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -78,6 +78,7 @@
   private List<OnAnimationFrame> mFrameCallbacks = new ArrayList<>();
   private ConcurrentLinkedQueue<Event> mEventQueue = new ConcurrentLinkedQueue<>();
   private boolean mWantRunUpdates;
+  private boolean preventOnAnimationFrame = false;
```
Szymon20000, line 81: If we check this in applier we do we need it here?

karol-bisztyga: It is needed here. Checking from Applier has been moved to NativeProxy on iOS, but on Android moving it to NativeProxy does not work as expected

karol-bisztyga: 933088b

PR #707, opened 2020-04-06
Fixed incorrect transform style prop typings

### LINKED ISSUE #297: Issue with AnimateStyle<> typing
While the typing supports perfectly top-level properties and style properties (kudos to that because the RN Animated API types them as any), there seems to be an issue with transform properties which are one level of nesting deeper.

For instance:

```tsx
const scale = new Value(1);
const opacity = new Value(1);
<Animated.View style={{ opacity }} /> // Work great
<Animated.View style={{ transform: [{ scale }] }} /> //  type error
```

I looked at the type definition `AnimateStyle` in `react-native-reanimated.d.ts` but this type definition is challenging for me to understand. I would love to help to solve this issue if pointed to the right direction though.

hastom:
`TransformsStyle["transform"]` is an array prop so we should infer it's element type, apply `Animated.Adaptable` and construct it back. Otherwise `Animated.Adaptable` applies to `Array` keys (`length`, `map`, `forEach`, etc).

Truly fixes #297

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -149,7 +149,7 @@ declare module 'react-native-reanimated' {
 
     export type TransformStyleTypes = TransformsStyle['transform'] extends readonly (infer T)[] ? T : never
     export type AdaptTransforms<T> = { [P in keyof T]: Animated.Adaptable<T[P]> }
-    export type AnimatedTransform = AdaptTransforms<TransformStyleTypes>
+    export type AnimatedTransform = (AdaptTransforms<TransformStyleTypes>)[]
 
     export type AnimateStyle<S extends object> = {
       [K in keyof S]: K extends 'transform' ? AnimatedTransform : (S[K] extends ReadonlyArray<any>
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -147,7 +147,9 @@ declare module 'react-native-reanimated' {
 
     export const SpringUtils: SpringUtils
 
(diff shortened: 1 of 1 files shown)


jhd-lego:
@jakub-gonet When will this be released? :)

Jomik:
Apologies, had to delete the above comment as I had used the wrong GitHub account :)
I can look at using the git version, but that doesn't seem like a stable solution :(
Is there a reason that we cannot simply release this immediately? The changes on master seem to be pretty minor.

jakub-gonet:
@Jomik, well, I wanted to accumulate more changes, but that's not a problem.

Version 1.8.0 is released on npm.

Jomik:
Thank you! 
That is much appreciated :)

wcandillon:
Thanks @hastom! That was well needed :)
Incredible how concisely you expressed it. I'm still struggling a lot with the infer keyword.

wcandillon:
@hastom vs tells me that this infers to `never`: https://www.dropbox.com/s/lrh8js7rl8jdt9p/Screenshot%202020-04-18%2021.26.11.png?dl=0
Could there be an issue with this?

wcandillon:
I'm also concern that this doesn't reflect the true data types of transform. For instance, rotate is Adaptable<string | number>, not Adaptable<string>

hastom:
@wcandillon I fixed issue with never in #722

wcandillon:
Amazing, I added a small comment to the PR.

PR #709, opened 2020-04-06
Bump examples to 0.62

jakub-gonet:
## Motivation

This PR bumps Reanimated example app to newest RN 0.62

## Description

Instead of using [diffs](https://react-native-community.github.io/upgrade-helper/?from=0.61.1&to=0.62.0) I created a new RN app (`react-native init ReanimatedExample`) and moved files from the old project.

This saved a bit of time and simplified project configuration:
- we use `react-native-reanimated` package as a local import (similar to [`react-native-screens`](https://github.com/software-mansion/react-native-screens/blob/7ab2a9d1ab63f29a440b798256e8a1377dbe31bd/Example/package.json#L14) approach) and postinstall script
- instead of tinkering with IOS workspace, we add local pod path to Podfile
- all rules from `proguard-rules.pro` was removed
- changes in android are also related to importing native module from a local path (settings.gradle) and typical config for RNGH
- all JS files were moved to `src/` folder

## Drawbacks

Unfortunately, I wasn't able to make `BottomSheetPlayground` work, so I removed it. It was pulled as a dependency from `osdnk/react-native-reanimated-bottom-sheet#reanimated-example`, so it should be easy to add it back though.

I also removed [Fingertips](https://github.com/mapbox/Fingertips), do we need it?

## Notes

We should also update Navigation in the near future to get rid of warnings of `getNode()` deprecation:
```
 Calling `getNode()` on the ref of an Animated component is no longer necessary. You can now directly use the ref instead. This method will be removed in a future release. ReactNativeFiberHostComponent
    in SafeView (at withOrientation.js:54)
    in withOrientation (at Header.tsx:738)
    in Header (created by withOrientation)
    in withOrientation (at StackViewLayout.tsx:217)
    in StackViewLayout (created by withOrientation)
    in withOrientation (at StackView.tsx:104)
    in Transitioner (at StackView.tsx:41)
    in StackView (created by Navigator)
    in Navigator (created by KeyboardAwareNavigator)
    in KeyboardAwareNavigator (created by NavigationContainer)
    in NavigationContainer (at renderApplication.js:45)
```


jakub-gonet:
Marking as WIP, because I forgot to add web version 😅

PR #712, opened 2020-04-07
fix(TypeScript): Fix return type for block()

wcandillon:
Currently the return type for `block()` is the same as the parameter type. This is a problem because the parameter type can be quite heterogeneous. For instance:

```
block([
  set(value, 1),
  not(value)
])
```
Has a return type of `Value | Node`. With this patch, it will have a return type of `Node`

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -269,9 +269,9 @@ declare module 'react-native-reanimated' {
       ifNode: Adaptable<T1>,
       elseNode?: Adaptable<T2>,
     ): AnimatedNode<T1 | T2>;
-    export function block<T>(
-      items: ReadonlyArray<Adaptable<T>>,
-    ): AnimatedNode<T>;
+    export function block<T1 extends Value = number, T2 extends Value = any>(
+      items: ReadonlyArray<Adaptable<T2>>,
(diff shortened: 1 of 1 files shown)


react-native-reanimated.d.ts
```diff
@@ -269,9 +269,9 @@ declare module 'react-native-reanimated' {
       ifNode: Adaptable<T1>,
       elseNode?: Adaptable<T2>,
     ): AnimatedNode<T1 | T2>;
-    export function block<T>(
-      items: ReadonlyArray<Adaptable<T>>,
-    ): AnimatedNode<T>;
+    export function block<T1 extends Value = number, T2 extends Value = any>(
```
jakub-gonet, line 272: `T1 extends Value = number` - I don't quite get it, can you elaborate why `number`? Wouldn't it prevent returning Values containing strings, null or undefined values?

wcandillon: Thank you for looking into this Jakub. This is the default type is no type is specified. You can look at other occurrences of `extends Value = number` in this file.

jakub-gonet: Oh, my bad, haven't used TS in a while :>

PR #720, opened 2020-04-08
Fix formatting numbers in concat node

### LINKED ISSUE #718: Latest 1.8.0 release breaks animated SVG paths on android
Due to this commit, animating svg paths will no longer work for android, as we initially pass in a path with no leading zeros, then the translation will then append the decimal place which will not work when iterpolating path. 

https://github.com/software-mansion/react-native-reanimated/commit/8a33e0e8569af03a5327d8c2ee08892aa3a1bcc2#diff-f137785d3392ec955cdfa51cd322cb53

This can be reproduced when trying to animate a path using interpolatePath with react-native-redash.

I think there is a lot of merit in having this functionality, but maybe we should introduce a new node, or a formatter node that can have the same result without breaking the previous implementation and retaining the ability to interpolate svg paths.

jakub-gonet:
## Description

It turned out that IOS doesn't take locale into consideration when formatting `NSNumber`, so this commit fixes that on android

## Changes

- Add `Locale.ENGLISH` to enforce formatting with a dot
- Add `setGroupingUsed(false)` to prevent grouping thousands (1000.0 doesn't get formatted into 1,000)

Fixes #718

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/nodes/ConcatNode.java
+++ b/android/src/main/java/com/swmansion/reanimated/nodes/ConcatNode.java
@@ -1,6 +1,7 @@
 package com.swmansion.reanimated.nodes;
 
 import java.text.NumberFormat;
+import java.util.Locale;
 
 import com.facebook.react.bridge.ReadableMap;
 import com.swmansion.reanimated.NodesManager;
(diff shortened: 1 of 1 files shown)


owinter86:
We can retain the previous commit, but add `sFormatter.setGroupingUsed(false);` which will stop the addition of commas in larger numbers

jakub-gonet:
Yeah, added it.

owinter86:
Nice, I have this as the patched commit and everything is working well.

mattijsf:
Isn't this worth a patch release? Something common to do with rotation transformations is concatting `deg` to the actual rotation value. This is currently broken in the latest release.

jakub-gonet:
@mattijsf, we need to adjust some typings in TS before that but sure, we'll probably release it at some point in the near future

PR #722, opened 2020-04-09
Fix strict mode typings

### LINKED ISSUE #719: TypeScript - Issue with transform (and other) props (1.8.0)
Hello.

I've just upgraded from 1.4.0 and I keep getting lots of TypeScript compilation errors. So for example following code:

```ts
export type AnimatedNumber = Animated.Adaptable<number>; // or Animated.Value<number>

export type AnimatedPosition = {
  x: AnimatedNumber;
  y: AnimatedNumber;
};

type Props = {
  initialPosition: AnimatedPosition;
};

const GameInteractableFlag = ({initialPosition}: Props) => {
  return (
    <Animated.View
      style={[
        styles.imageFlag,
        {
          transform: [ // these lines
            {translateX: initialPosition.x}, // and this
            {translateY: initialPosition.y}, // and this
          ], // and this
        },
      ]}>
      <FastImage source={images.flag} imageStyle={styles.imageFlag} />
    </Animated.View>
  );
};
```

is showing following error:

```ts
No overload matches this call.
  Overload 1 of 2, '(props: Readonly<AnimateProps<ViewStyle, ViewProps>>): View', gave the following error.
    Type '{ transform: ({ translateX: Animated.Adaptable<number>; } | { translateY: Animated.Adaptable<number>; })[]; }' is not assignable to type 'false | AnimateStyle<ViewStyle> | RegisteredStyle<AnimateStyle<ViewStyle>> | RecursiveArray<false | AnimateStyle<ViewStyle> | RegisteredStyle<...> | null | undefined> | readonly (false | ... 3 more ... | undefined)[] | null | undefined'.
      Types of property 'transform' are incompatible.
        Type '({ translateX: Adaptable<number>; } | { translateY: Adaptable<number>; })[]' is not assignable to type 'AnimatedTransform'.
          Type '{ translateX: Adaptable<number>; } | { translateY: Adaptable<number>; }' is not assignable to type 'never'.
            Type '{ translateX: Animated.Adaptable<number>; }' is not assignable to type 'never'.
  Overload 2 of 2, '(props: AnimateProps<ViewStyle, ViewProps>, context?: any): View', gave the following error.
    Type '{ transform: ({ translateX: Animated.Adaptable<number>; } | { translateY: Animated.Adaptable<number>; })[]; }' is not assignable to type 'false | AnimateStyle<ViewStyle> | RegisteredStyle<AnimateStyle<ViewStyle>> | RecursiveArray<false | AnimateStyle<ViewStyle> | RegisteredStyle<...> | null | undefined> | readonly (false | ... 3 more ... | undefined)[] | null | undefined'.
      Types of property 'transform' are incompatible.
        Type '({ translateX: Adaptable<number>; } | { translateY: Adaptable<number>; })[]' is not assignable to type 'AnimatedTransform'.
          Type '{ translateX: Adaptable<number>; } | { translateY: Adaptable<number>; }' is not assignable to type 'never'.
            Type '{ translateX: Animated.Adaptable<number>; }' is not assignable to type 'never'.ts(2769)
```

and also for example animated SVG path:

```ts
import {Path} from 'react-native-svg';
const AnimatedPath = Animated.createAnimatedComponent<typeof Path>(Path);

...

<AnimatedPath
      d={d}
      strokeDasharray={length}
      strokeDashoffset={interpolate(runner, { // these lines below
        inputRange: [0, 1],
        outputRange: [length, 1],
        extrapolate: Extrapolate.CLAMP,
      })}
      opacity={interpolate(runner, {.  // and these lines below
        inputRange: [0, 0.25, 1],
        outputRange: [0, 0.8, 1],
        extrapolate: Extrapolate.CLAMP,
      })}
      fill={fill}
      stroke={stroke}
      strokeWidth={strokeWidth}
    />
```

are showing following error message:

```ts
Type 'AnimatedNode<number>' is not assignable to type 'string | number | undefined'.
  Type 'AnimatedNode<number>' is not assignable to type 'number'.ts(2322)
index.d.ts(136, 3): The expected type comes from property 'strokeDashoffset' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<PathProps, any, any>> & Readonly<PathProps> & Readonly<...>'
```

Can anyone help me to make this work? Thank you so much!

hastom:
Partially fixes #719 
My bad 😞 Forgot that `transform` property on `TransformsStyle['transform']` is optional so in strict mode inference must check for `undefined`


react-native-reanimated.d.ts
```diff
@@ -147,7 +147,7 @@ declare module 'react-native-reanimated' {
 
     export const SpringUtils: SpringUtils
 
-    export type TransformStyleTypes = TransformsStyle['transform'] extends readonly (infer T)[] ? T : never
+    export type TransformStyleTypes = TransformsStyle['transform'] extends (readonly (infer T)[] | undefined) ? T : never
     export type AdaptTransforms<T> = { [P in keyof T]: Animated.Adaptable<T[P]> }
```
wcandillon, line 151: so here I would add `T[P] extends string ? string | number : number` so that rotate types work well too.

wcandillon: `export type AdaptTransforms<T> = { [P in keyof T]: Animated.Adaptable<T[P] extends string ? string | number : number> }`

hastom: According to react-native typings 
```
interface RotateTransform {
    rotate: string;
}

interface RotateXTransform {
    rotateX: string;
}

interface RotateYTransform {
    rotateY: string;
}

interface RotateZTransform {
    rotateZ: string;
}
```
Rotates accepts only strings

wcandillon:
not in reanimated

hastom:
Oh, I see. @wcandillon were you the one who write those types initially?

wcandillon:
Not initially no

react-native-reanimated.d.ts
wcandillon: not in reanimated, it accepts both string and number

hastom:
Are there any other differences in practice?

wcandillon:
Are you asking if there are benefits to use rotate as a number? Yes there are many. The main one being able to perform operations on 2d transformations such as accumulations. Another one is that support of strings as animation values in reanimated is a bit shaky. Overall it is such nice to be able to treat your 2d transformation in a homogeneous way.

Here is an example of transform accumulation (useful if you have use the rotate gesture handler) that would only work if the rotate is of number type: https://github.com/wcandillon/react-native-redash/blob/master/packages/core/src/__tests__/Matrix.test.ts#L39

hastom:
No, I mean if there any other differences in types between reanimated and "vanilla" react-native. So, may be, we should not take base transform types from react-native and just hardcode them (if they differ much)

wcandillon:
it's the only difference I know of.

hastom:
Skew transforms also use string values, so they are affected too, i guess.

wcandillon:
Thank you @hastom, you rock 🎸

PR #723, opened 2020-04-09
[v1] Fix animated component typings

hastom:
This is just a copy of `@types/react-native: 0.62.2` typings of `createAnimatedComponent`. I guess this one should fix the second part of #719


react-native-reanimated.d.ts
```diff
@@ -198,7 +200,32 @@ declare module 'react-native-reanimated' {
...
+
+    export type ComponentProps<T> = T extends React.ComponentType<infer P> | React.Component<infer P> ? P : never
+
+    export interface WithAnimatedValue<T>
+      extends ThisType<
+        T extends object
+          ? { [K in keyof T]?: WithAnimatedValue<T[K]> }
+          : T extends (infer P)[]
+          ? WithAnimatedValue<P>[]
+          : T | Value | AnimatedInterpolation
```
jakub-gonet, line 212: I think this line should be adjusted to Reanimated types, we have our own interpolation type

jakub-gonet:
Closing due to no activity.

PR #724, opened 2020-04-09
Example Liquid Swipe

Szymon20000:


Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -323,26 +329,38 @@ void NativeReanimatedModule::unregisterMapper(jsi::Runtime &rt, int id) {
...
+    this->workletModule = std::shared_ptr<BaseWorkletModule>(new WorkletModule(
+      sharedValueRegistry,
+      applierRegistry,
+      workletRegistry,
+      eventPtr,
+      this->errorHandler));
+  } else {
+      std::dynamic_pointer_cast<WorkletModule>(this->workletModule)->setEvent(eventPtr);
+  }
+  SpeedChecker::checkSpeed("Event:", [=] () {
```
karol-bisztyga, line 361: speed check in general purpose code

terrysahaidak:
It seems like I cannot build it on Android, some problems with linking yoga (again). Is this related to the PR or it's just me?

Szymon20000:
It should work now. However, you need to call "git checkout -- Example/node_modules" after yarn.

terrysahaidak:
Thanks. Yep, it's working now. So, RN currently doesn't provide a way to link TurboModule enabled libraries (at least its NDK part)?

PR #725, opened 2020-04-09
Provide useSpring.

Szymon20000:


src/reanimated2/helpingHooks/Spring.js
```diff
@@ -0,0 +1,109 @@
...
+  stiffness: 100,
+  overshootClamping: 0,
+  restDisplacementThreshold: 0.001,
+  restSpeedThreshold: 0.001,
+};
+
+const workletBody = new Worklet( 
+  function (sv, state, config) {
+    'worklet';
+    const memory = Reanimated.memory(this);
```
osdnk, line 22: are you sure about this part of API? Don't you afraid about some linters or type-checks?

Szymon20000: I think we can export {Reanimated: memory:()=>{return {};}, ...}.

Common/cpp/SharedItems/SharedWorkletStarter.cpp
```diff
@@ -180,6 +175,10 @@ void SharedWorkletStarter::willUnregister(jsi::Runtime &rt) {
   if (this->unregisterListener != nullptr) {
     (*this->unregisterListener)();
   }
+  
+  applierRegistry->unregisterApplierFromRender(applierId, rt);
+  // unregister applier
```
karol-bisztyga, line 180: I'd put comments above the lines to which they refer

PR #726, opened 2020-04-10
Fix link to transitions example in docs

osdnk:


jakub-gonet:
Could you also fix link to imagePreview example in About reanimated page?

PR #727, opened 2020-04-10
Add publishing docs with gh-pages

osdnk:
Please add access token in secrets


osdnk:
https://github.com/tsunematsu21/actions-publish-gh-pages

PR #731, opened 2020-04-14
Worklets function install

karol-bisztyga:


src/reanimated2/Hooks.js
```diff
@@ -3,14 +3,7 @@ import { useEffect, useRef, useLayoutEffect } from 'react';
 import SharedValue from './SharedValue';
 import Worklet from './Worklet';
 import WorkletEventHandler from './WorkletEventHandler';
-
-global.Reanimated = {};
```
Szymon20000, line 7: it's important. Revert this change.

PR #745, opened 2020-04-18
Fix TypeScript types for transform

wcandillon:
The current type definition doesn't seem to be doing any type checking on my side. 
If it were to work, it doesn't reflect the true accepted types for instance rotate is  `Adaptable<number | string>`.

see https://github.com/software-mansion/react-native-reanimated/pull/707


wcandillon:
Closing in favor of #722

PR #746, opened 2020-04-20
Adds option to specify diffClamp animation

zobeirhamid:
I needed to use diffClamp for two different components using the same Animation.Value. It didn't work, only one component was able to use the diffClamp Value. I did not look much into it, but I solved the issue with specifying an Animation.Value the clamp should be set to, which allowed me to have several diffClamps Value based on one Animation.Value.

### DIFF
--- a/src/derived/diffClamp.js
+++ b/src/derived/diffClamp.js
@@ -4,12 +4,14 @@ import min from './min';
 import max from './max';
 import diff from './diff';
 
-const procAcc = proc(function(a, minVal, maxVal, value) {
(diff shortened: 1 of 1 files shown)


rauchp:
Ha I was literally just working on this tonight... thanks! Hope it gets approved soon.

Matergi:
I have the same issue, i think that is a fast workaround, but worked!

jakub-gonet:
Thanks for the PR, appreciated it.

This is resolved by #1068, so I'll close this one.

Issue #452, opened 2019-11-04
ScrollTo does not work on reanimated ScrollView


jacobarvidsson:
When using scrollTo on a reanimated ScrollView ref, an error is thrown: "TypeError: node.scrollTo is not a function". I have reproduced it in a snack:

https://snack.expo.io/@jacobarvidsson/2c7078

```javascript
const App = () => {
  const scrollViewRef = useRef(null);
  useEffect(() => {
    if (scrollViewRef.current) {
      const node = scrollViewRef.current;
      if (node) {
        node.scrollTo({ x: 0, y: 0 });
      }
    }
  }, [])
    return (
      <View style={styles.container}>
        <Animated.ScrollView ref={scrollViewRef}>
          <Text>Test</Text>
        </Animated.ScrollView>
      </View>
    );
};
```

emzet93:
I'm facing the same issue. @jacobarvidsson did you find any workaround for this?

jacobarvidsson:
@emzet93 The workaround works, thanks! Obviously better if the animated component would forward the ref of the actual component, so leaving this open for now.

nhuesmann:
@emzet93 Thanks for posting your solution, it worked for me too.

yairopro:
I opened a PR that would solve this issue.

co2nut:
Have you guys tried with  renderScrollComponent inside FlatList? It seems to be not working. 

Will get error  TypeError:null is not an object( 'reffVar.current.geNode' )

![image](https://user-images.githubusercontent.com/4594000/82640604-76e6d980-9c3d-11ea-927a-d24ce91cf184.png)

amanmanhas:
Hi guys, you can solved it by using the **getNode()** method

There are two ways to solve this issue.-
1)If you are using the new recommended way to create a ref to a component since React 16.3 is to use React.createRef() then you will use in this way:-

scrollView = React.createRef();
<Animated.ScrollView
  ref={this.scrollViewRef}/>
  this.scrollViewRef.current.getNode().scrollTo({x: number, y: number, animated: true|false})

2) If you are using the previous way to create a ref before React 16.3 then you will use in this way:-

<Animated.ScrollView
  ref={c => (this.myRef = c)}/>
  this.myRef.getNode().scrollTo({
    y: 0,
    animated: true,
  });


The difference between the above two solution is , in one you can get the getNode() method inside the *current* and the other you can access it without current

PR #748, opened 2020-04-20
A ScrollView component with mixin

yairopro:
This PR is addressed to solve the issue #452.
For now the mixin added to the instance has just `scrollTo()` method, where it simply redirect the call to the original `scrollView.scrollTo`.
Adding methods, like `scrollTo` here, allows a better integration of the animated scrollview component in a `FlatList`.

### DIFF
--- a/src/Animated.js
+++ b/src/Animated.js
@@ -1,4 +1,4 @@
-import { Image, ScrollView, Text, View } from 'react-native';
+import { Image, Text, View } from 'react-native';
 import Easing from './Easing';
 import AnimatedClock from './core/AnimatedClock';
 import AnimatedValue from './core/AnimatedValue';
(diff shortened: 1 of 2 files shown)


jakub-gonet:
Hi @yairopro, thanks for the PR.

Instead of adding mixin, we decided to implement ref forwarding which is more scalable and works for any component.

yairopro:
Hi @jakub-gonet 

With a forwarded ref of the `ScrollView` instance, the user won't be able to access the animated component's instance and its methods if he needs to.

jakub-gonet:
That's right, but do you have any use cases for it? `createAnimatedComponent` is basically a HOC and doesn't expose any public function to users.

yairopro:
Ah ok.
Then the forwarding ref is a nice solution. 👍🏻

jakub-gonet:
Anyway, thanks for contributing to the Reanimated, we really appreciate it.

PR #750, opened 2020-04-21
Simplify typings for 'proc'

iyegoroff:


wcandillon:
Just tried it and it works great 🙌🏻

react-native-reanimated.d.ts
```diff
@@ -236,22 +236,22 @@ declare module 'react-native-reanimated' {
...
+    ): typeof cb;
     export function proc<P1, P2, P3, P4, P5>(
       cb: (p1: P1, p2: P2, p3: P3, p4: P4, p5: P5) => AnimatedNode<number>
-    ): (p1: P1, p2: P2, p3: P3, p4: P4, p5: P5) => AnimatedNode<number>;
+    ): typeof cb;
     export function proc<P1, P2, P3, P4, P5, P6>(
       cb: (p1: P1, p2: P2, p3: P3, p4: P4, p5: P5, p6: P6) => AnimatedNode<number>
-    ): (p1: P1, p2: P2, p3: P3, p4: P4, p5: P5, p6: P6) => AnimatedNode<number>;
+    ): typeof cb;
     export function proc <P>(
```
wcandillon, line 255: you can probably use `typeof cb` here as well

react-native-reanimated.d.ts
```diff
@@ -236,25 +236,37 @@ declare module 'react-native-reanimated' {
...
-    ): (p1: P1, p2: P2, p3: P3, p4: P4, p5: P5) => AnimatedNode<number>;
+    ): typeof cb;
     export function proc<P1, P2, P3, P4, P5, P6>(
       cb: (p1: P1, p2: P2, p3: P3, p4: P4, p5: P5, p6: P6) => AnimatedNode<number>
-    ): (p1: P1, p2: P2, p3: P3, p4: P4, p5: P5, p6: P6) => AnimatedNode<number>;
-    export function proc <P>(
-      cb: (...params: AnimatedValue<number>[]) => AnimatedNode<number>
-    ): (...params: Adaptable<number>[]) => AnimatedNode<number>;
+    ): typeof cb;
+    export function proc<P1, P2, P3, P4, P5, P6, P7>(
```
jakub-gonet, line 255: Do you think adding arity 7-10 proc functions is necessary?

iyegoroff: Yes, in some cases it is not even enough

https://github.com/software-mansion/react-native-reanimated/blob/6710cbd636954d2f30990fc90ecbb4640749d723/Example/src/test/index.js#L28-L42

jakub-gonet: To be honest we can go on and on with extending generic overloads in that case 😄 

Ten arguments may be a good compromise though

PR #757, opened 2020-04-24
Add common methods.

Szymon20000:


src/reanimated2/InstallFunctions.js
```diff
@@ -87,12 +96,63 @@ export function installFunctions(innerNativeModule) {
...
+    const diffs = values.map((it) => Math.abs(it - x));
+    const index = diffs.indexOf(Math.min(...diffs));
+    return values[index];
+  }
+
+  global.Reanimated.clamp = clamp;
+  install('Reanimated.clamp', clamp);
+
+  // interpolate
+  const interpolate = function (x, l, r, ll, rr, type) {
```
terrysahaidak, line 112: what about interpolate with more than 2 args?

```
opacity: interpolate(scrollPosition, {
  input: [0, 0, 1]
  output: [-400, 0, 100] // this is common thing to avoid overscroll
});
```

src/reanimated2/InstallFunctions.js
```diff
@@ -87,12 +96,63 @@ export function installFunctions(innerNativeModule) {
     }
     _log(stringRepresentation(data))
   })
+
+  // clamp
+  const clamp = function (x, values) {
```
terrysahaidak, line 101: by the wah, `function(x, ...values)` works just fine

src/reanimated2/InstallFunctions.js
Szymon20000: Done :)

src/reanimated2/InstallFunctions.js
terrysahaidak: well, it's not :) some problems with Hermes which doesn't like the spread to be used at all

But works just fine on JSC

PR #764, opened 2020-04-27
Change deprecated maven plugin to newer maven-publish plugin

jkadamczyk:
I changed the way of publication of maven dependencies from now deprecated `maven` plugin to the newer and suggested `maven-publish` plugin


jakub-gonet:
Should be closed in favor of #795?

PR #773, opened 2020-04-30
fix: prevent clock from being set to 0

### LINKED ISSUE #763: Clock initial value is negative
## Description
I think clock value should always be greater than the old value? However, the screenshot shows that clock didn't behave like this and even caused the frameTime to be negative. This also made the animating view flickering. I think there are some bugs inside native code?

### Screenshots
image

## Steps To Reproduce
runTiming code from https://software-mansion.github.io/react-native-reanimated/declarative-animation-api.html can reproduce this behaviour

## Package versions
react-native 0.62.0
react-native-reanimated 1.8.0
This bug actually has been a while, not just v.1.8.0

## Platform
iOS only

jaulz:
## Description

In some weird cases the clock jumps to 0 and thus it causes animations to act weird because the frameTime is negative. It also improves the performance of the evaluate function by using `Date.now`:
https://stackoverflow.com/questions/14255664/new-date-is-this-good-practice/14255732#14255732

Fixes #763.

## Changes

For some reason the value was set to 0 but it is now set to `Date.now`.

### Before
image

Afterwards it causes an (near) infinite loop:

image

### After
The animations just runs.

### DIFF
--- a/src/core/AnimatedClock.js
+++ b/src/core/AnimatedClock.js
@@ -6,15 +6,15 @@ class AnimatedMainClock extends InternalAnimatedValue {
   _frameCallback;
 
   constructor() {
-    super({ type: 'MAIN_CLOCK' });
+    super(Date.now());
   }
 
   __onEvaluate() {
-    return +new Date();
+    return Date.now();
   }
 
   _runFrame = () => {
-    this._updateValue(0);
+    this._updateValue(this.__onEvaluate());
     if (this.__children.length > 0) {
       this._frameCallback = requestAnimationFrame(this._runFrame);
     }


src/core/AnimatedClock.js
```diff
@@ -6,15 +6,15 @@ class AnimatedMainClock extends InternalAnimatedValue {
   _frameCallback;
 
   constructor() {
-    super({ type: 'MAIN_CLOCK' });
+    super(Date.now());
```
jakub-gonet, line 9: This breaks the contract of `AnimatedNode`, `type` have to be specified.

jakub-gonet:
Closing until the author responds.

Issue #181, opened 2019-02-05
Color node documentation fixes


kmagiera:
Document that the values provided for the color node should be integers (rounded to integers). In addition we should verify if the provided values are rounded and warn if the aren't as otherwise the resulting color is kind of random.

osdnk:
I'm rather into throwing error from native or rounding it inside color node

terrysahaidak:
Probably, rounding is the reason for my bug with animating the color.
Here is an example:
https://snack.expo.io/@terrysahaidak/reanimatedcolorbugexample
Let me know if I should open a separate issue for it.

(Thank you guys for such an awesome lib!)

jacobarvidsson:
I am getting this color bug as well, where instead of transitioning between two colors smoothly, it flashes with all different colors throughout the animations, just as in the snack above. I am using integers in the interpolation, am I doing something wrong or is this a bug?
`
interpolate(scale, {
      inputRange: [0.7, 1],
      outputRange: [
        Animated.color(67, 125, 239, 1),
        Animated.color(227, 232, 243, 1),
      ]
})
`

gazedash:
Same for me @jacobarvidsson

jacobarvidsson:
I realised what I did wrong, wrote a helper method for future use:
`
const hexToRgb = (hex: string) => {
  const result = colorRegex.exec(hex);
  return result
    ? {
        r: parseInt(result[1], 16),
        g: parseInt(result[2], 16),
        b: parseInt(result[3], 16),
      }
    : null;
};

const white = { r: 255, g: 255, b: 255 };

export const interpolateColors = (
  animationValue: Animated.Adaptable<number>,
  inputRange: number[],
  hexColors: string[],
) => {
  const colors = hexColors.map(hexColor => hexToRgb(hexColor) || white);
  const r = round(
    interpolate(animationValue, {
      inputRange,
      outputRange: colors.map(c => c.r),
      extrapolate: Extrapolate.CLAMP,
    }),
  );
  const g = round(
    interpolate(animationValue, {
      inputRange,
      outputRange: colors.map(c => c.g),
      extrapolate: Extrapolate.CLAMP,
    }),
  );
  const b = round(
    interpolate(animationValue, {
      inputRange,
      outputRange: colors.map(c => c.b),
      extrapolate: Extrapolate.CLAMP,
    }),
  );
  return color(r, g, b);
};
`

agrinko:
[...]

Issue #262, opened 2019-05-02
color rgb opacity bug in reanimated


nomi9995:
```
	const backgroundColor = interpolate(this.translateY, {
			inputRange: [0, position.y],
			outputRange: [Animated.color(0, 0, 0, 1), Animated.color(0, 0, 0, 0)]
		});
```
i am using black color with opacity but it show all the color during interpolate
 you can check reporducable bug on my snack
https://snack.expo.io/@nomi9995/reanimatedcolorbugexample

nomi9995:
@kmagiera can you help me?

KakarN:
I have worked with it `Animated.color()` in the past, however, I have implemented it differently than yours.

```
const backgroundColor = color(1, 1, 1, interpolate(translateY, {
                inputRange: [0, HEIGHT],
                outputRange: [1, 0],
                extrapolate: Extrapolate.CLAMP
            }))
```

Hope this helps.

nomi9995:
@KakarN thanks it is working but it should working same as core animated

CatapultJesse:
Getting the same thing when trying to use multiple colours

nomi9995:
yeah there is a bug in this library and @kmagiera should fix it

CatapultJesse:
> yeah there is a bug in this library and @kmagiera should fix it

Have a bit of respect for the thousands of lines of free code you are getting.

nomi9995:
sorry i did not mean this. i just mension. i love @kmagiera for this awesome library <3

Fantasim:
If like me you meet this non understanding thing on the lib because of the different way of doing compared to the original Animated library, I just found this code which helped me a lot to understand how it works with colors on the lib.

https://github.com/wcandillon/react-native-redash/blob/master/src/Colors.ts#L136

jeremybarbet:
Any news about that? 

It also looks like animating `tintColor` is an issue, and doesn't work as expecting. I try to look around the documentation and other issues, but didn't find anything related. Is it a known issue?

kmagiera:
Reanimated does not support string based color interpolation. There are some sample demos under Example folder demonstrating how color interpolation should be done with reanimated. Also this comment by @KakarN should work https://github.com/software-mansion/react-native-reanimated/issues/262#issuecomment-489107824

[...]

PR #777, opened 2020-05-05
feat: Add `interpolateColors` function

jtomaszewski:
## Description

Lets users animate colors with `react-native-reanimated` without the need of additional helper functions like [this one](https://github.com/software-mansion/react-native-reanimated/issues/181#issuecomment-528245794) or [this one](https://github.com/wcandillon/react-native-redash/blob/master/packages/core/src/Colors.ts#L136). 

Also will help avoid people from running into a problem "_why does this color doesn't animate?_ :(" like I did today in the morning in one of my projects. :)

Fixes https://github.com/software-mansion/react-native-reanimated/issues/181 .

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -328,6 +328,16 @@ declare module 'react-native-reanimated' {
       value: Adaptable<number>,
       config: InterpolationConfig,
     ): AnimatedNode<number>;
+    export function interpolateColors(
+      animationValue: Adaptable<number>,
+      {
+        inputRange,
+        outputRgbaRange: rgbaColors
+      }: {
+        inputRange: ReadonlyArray<Adaptable<number>>;
+        outputRgbaRange: [number, number, number, number][];
+      }
+    ): AnimatedNode<number>;
     export const max: BinaryOperator;
     export const min: BinaryOperator;
 
@@ -471,6 +481,7 @@ declare module 'react-native-reanimated' {
   export const abs: typeof Animated.abs
   export const acc: typeof Animated.acc
   export const color: typeof Animated.color
+  export const interpolateColors: typeof Animated.interpolateColors
   export const diff: typeof Animated.diff
   export const diffClamp: typeof Animated.diffClamp
   export const interpolate: typeof Animated.interpolate
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -332,10 +332,10 @@ declare module 'react-native-reanimated' {
       animationValue: Adaptable<number>,
       {
         inputRange,
-        outputRgbaRange: rgbaColors
+        outputColorRange
       }: {
         inputRange: ReadonlyArray<Adaptable<number>>;
-        outputRgbaRange: [number, number, number, number][];
+        outputColorRange: (string | number)[];
       }
     ): AnimatedNode<number>;
     export const max: BinaryOperator;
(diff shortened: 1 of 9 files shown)


jakub-gonet:
Thanks for your contribution!
This also fixes #262.

I wonder if we could include it directly in `interpolate` wrapping it (if input are strings) or leave as a helper

jtomaszewski:
> I wonder if we should include it directly in interpolate wrapping it or leave as a helper

Maybe it could be included directly in `interpolate`, but then the `interpolate` would need to be able to distinguish between a color output and non-color output? and so we'd need to edit the `Animated.color` (or create a new method for that) to create some kind of a "color class"?

I was thinking about it, but then decided: doing is definitely more work; maybe needs to be discussed first; and still, I think it can be done later on as improvement. (The helper would then still work.) 

For now I think merging this PR will be very helpful, as it'll make animating colors way easier, and also provide the docs describing how to do it. ( Currently there are none ;P )

Let me know if you want me to do some changes to the PR if anything blocks merging it, I'm happy to help.

jtomaszewski:
Also, it does fix https://github.com/software-mansion/react-native-reanimated/issues/262 I think, at least the demo - you can see it being fixed after using the helper here: https://snack.expo.io/@jack-ailo/reanimatedcolorbugexample .

jakub-gonet:
> the interpolate would need to be able to distinguish between a color output and non-color output, right? and so we'd need to edit the Animated.color (or create a new method for that) to create some kind of a "color class"?

You're right, yes. I was convinced that color is a node but it's just a few bitshifts glued together.

> Also, it does fix #262 (...)

That's also true, I mistook it for another issue with interpolating colors as strings e.g. from "#fff" to "#000" (which can also be added in some followup)

I think a minor update to docs would be awesome:
- crossreferencing `interpolateColors` in the interpolate and color node, so people can see that it exists
- adding info about [0, 255] range for colors (most users will know about that, but it won't hurt).

After those changes, this will be ready for merging.

jtomaszewski:
Updates to docs added ;)

wcandillon:
Not sure if this helps but we have implemented such functions in redash: https://github.com/wcandillon/react-native-redash/blob/master/packages/core/src/Colors.ts
maybe it can help to make the API a little bit more user friendly? For instance to pass colors as strings.

jtomaszewski:
I didn't know that react-native has such function as `processColor` built in. I was afraid that parsing the colors as strings would require including a dependency, but I can see `react-native` is enough. I'll improve it in a minute. Thanks @wcandillon !

jtomaszewski:
Actually the `interpolateColors` wasn't exported at all here. I added an example and fixed all the bugs. Now this PR is 100% merge ready IMHO.

react-native-reanimated.d.ts
```diff
@@ -328,6 +328,16 @@ declare module 'react-native-reanimated' {
       value: Adaptable<number>,
       config: InterpolationConfig,
     ): AnimatedNode<number>;
+    export function interpolateColors(
+      animationValue: Adaptable<number>,
+      {
+        inputRange,
+        outputColorRange
+      }: {
+        inputRange: ReadonlyArray<Adaptable<number>>;
```
jakub-gonet, line 337: After adding string handling this shouldn't be `ReadonlyArray<Adaptable<number | string>>;`?

wcandillon:
Just had a look at the diff and it looks good :)

ekatzenstein:
interpolateColors is not exported with the v1.9.0

jakub-gonet:
It's not, it'll be included in the next version.

PR #786, opened 2020-05-12
Add useValue hook

iamolegga:
## Description

When working with functional components this is easy to make bugs with statement, that I already saw several times from different people:

```ts
function MyComponent() {
  const value = new Value(...);
}
```

This will lead to creating new value instance for every re-render. And components, that use internal value of previous instance will be reset to the new one, and this will lead to visual bugs. So one should always keep the same instance wrapping it either in `useRef(new Value(...))` or `useMemo(() => new Value(...), [])`

## Changes

This PR adds hook, that could be used in very often simple case of creating instance of `Value` without extra wrap:

```ts
function MyComponent() {
  const value = useValue(0);
}
```

### DIFF
--- a/src/useValue.js
+++ b/src/useValue.js
@@ -2,5 +2,9 @@ import React from 'react';
 import AnimatedValue from './core/AnimatedValue';
 
 export default function useValue(initialValue) {
-  return React.useRef(new AnimatedValue(initialValue)).current;
+  const ref = React.useRef(null);
+  if (ref.current === null) {
+    ref.current = new AnimatedValue(initialValue);
+  }
+  return ref.current;
 }
--- a/src/useValue.js
+++ b/src/useValue.js
@@ -1,0 +1,6 @@
+import React from 'react';
+import AnimatedValue from './core/AnimatedValue';
+
+export default function useValue(initialValue) {
+  return React.useRef(new AnimatedValue(initialValue)).current;
+}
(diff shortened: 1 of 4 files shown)


src/useValue.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import AnimatedValue from './core/AnimatedValue';
+
+export default function useValue(initialValue) {
+  return React.useRef(new AnimatedValue(initialValue)).current;
```
jakub-gonet, line 5: When rerendering functional component `useValue` is called and we create `new AnimatedValue(initialValue)`, don't we?

WDYT about this impl?
```js
export default function useValue(initialValue) {
  const ref = React.useRef(null);
  if(ref === null){
    ref.current = new AnimatedValue(initialValue);
  }
  return ref.current
}
```

iamolegga: I'm totally fine with this 👍If creating new instance could lead to perf issues let's use this one

iamolegga: just change `if(ref === null){` to `if(ref.current === null){` 🙂

jakub-gonet: You're right, should be `ref.current`

Reanimated have some issues with initialization of nodes, so we aim to minimize nodes created whenever possible 🙂

iamolegga: done ✔️

iamolegga:
Hi, FYI, just opened [the same PR to react-native](https://github.com/facebook/react-native/pull/28887) for consistency of API

PR #804, opened 2020-05-25
Use double instead of CGFloat in iOS operators

### LINKED ISSUE #797: Math errors with modulo on iOS
## Description

When using `modulo` with certain values, the result appears to be incorrect. I haven't worked out what the pattern is for values that cause this behaviour, and the bug only appears to exist on iOS. Reanimated, Android and Web appear to behave correctly.

As an example, the operation `9999999999 % 10 === 9` in JS. In Reanimated `modulo(new Value(9999999999), 10) === 9` on web and Android. However on Reanimated iOS I get a result of `0`.

Initially I thought this might have been an issue with being above max 32bit int, but that doesn't appear to be the case as some numbers give the correct result, while some values much smaller than max int do not.

### Screenshots

Reanimated Web | Reanimated Android | Reanimated iOS
---|---|---
`9` | `9` | `0`
image | image | image

## GIF from iOS

This GIF shows some of the unusual results that are given from my [expo snack](https://snack.expo.io/@levibuzolic/reanimated-math). Values appear to be mostly incorrect when the start value is `9999999999` (or a similarly large number) but setting the value to `1234567` and then gradually increasing the value, the result appears to be correct, even when exceeding a value of `9999999999`.

The left-hand side is pure JS (`value % 10`) while the right side is the result from reanimated (`modulo(animatedValue, 10)`).

RPReplay_Final1590116531

https://snack.expo.io/@levibuzolic/reanimated-math

```js
import { ReText } from 'react-native-redash';
const { concat, Value, modulo } = Animated;
const { useRef } = React;

  const { current: animValue } = useRef(new Value(9999999999));
  return <ReText text={concat('', modulo(animValue, 10))} />;
}
```

## Package versions

For the sample app I've tested in Expo 37.0.0 (which I believe uses Reanimated 1.70)

The issue also exists in my current production app running:
- React: 16.9.0
- React Native: 0.61.5
- React Native Reanimated: 1.8.0

jakub-gonet:
## Description

Fixes #797

iOS operators implementation used `CGFloat` which is a typedef for `double` / `float` depending on OS architecture. This caused some precision errors on big numbers, so this PR fixes that by changing the type to double.

Transitions and Bezier node wasn't changed, because they're responsible for visuals and not calculations - doesn't need to be exact.

## Changes

- change type from `CGFloat` to `double` in operator macros.
- change `fmodf` to `fmod` in `modulo` node

Tested on [snack](https://snack.expo.io/@levibuzolic/reanimated-math) provided by @levibuzolic

### DIFF
--- a/ios/Nodes/REAOperatorNode.m
+++ b/ios/Nodes/REAOperatorNode.m
@@ -6,22 +6,22 @@
 typedef id (^REAOperatorBlock)(NSArray<REANode *> *inputNodes);
 
 #define REA_REDUCE(OP) ^(NSArray<REANode *> *inputNodes) { \
-CGFloat acc = [[inputNodes[0] value] doubleValue]; \
+double acc = [[inputNodes[0] value] doubleValue]; \
 for (NSUInteger i = 1; i < inputNodes.count; i++) { \
-  CGFloat a = acc, b = [[inputNodes[i] value] doubleValue]; \
+  double a = acc, b = [[inputNodes[i] value] doubleValue]; \
   acc = OP; \
 } \
 return @(acc); \
 }
 
 #define REA_SINGLE(OP) ^(NSArray<REANode *> *inputNodes) { \
-CGFloat a = [[inputNodes[0] value] doubleValue]; \
+double a = [[inputNodes[0] value] doubleValue]; \
 return @(OP); \
 }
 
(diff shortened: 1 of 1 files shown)


jakub-gonet:
Closing and reopening to retrigger checks.

levibuzolic:
Awesome. 🙇‍♂️

PR #826, opened 2020-05-29
Fix broken link on 2.0 docs home page

marcusstenbeck:
## Description

The link now goes to the first page in the docs instead of a 404.


jakub-gonet:
Thanks for the contribution, I included your change in #832 to not multiply typo fixing PRs.

PR #832, opened 2020-05-29
Fix typos & setup code in Reanimated 2 docs

terrysahaidak:
## Description

Fixes some typos and confusion in the new docs.

PR #836, opened 2020-05-29
Fixes button alignment in hero

tomekzawada:
## Description

Fixes button alignment in hero


jakub-gonet:
Included typo fix in #832.

PR #840, opened 2020-05-30
v2 types

renanmav:
## Description

This PR includes the type definitions for react-native-reanimated v2. I created these types following the documentation and doing a quick check within the source code. Tested it against some changes I made on my reanimated-v2-playground. A better test would be to use these types on example app.

### Suggestions

Increase DX for types, lint-staged isn’t configured to work for typescript files.


react-native-reanimated.d.ts
```diff
@@ -380,8 +413,56 @@ declare module 'react-native-reanimated' {
...
     ): AnimatedValue<T>;
 
+    // reanimated2 hooks
+    export function useSharedValue<T extends Value>(
+      initialValue: T
+    ): AnimatedValue<T>;
+    export function useDerivedValue<T extends Value>(
+      processor: () => T
+    ): AnimatedValue<T>;
+    export function useAnimatedStyle<T extends FlexStyle, TransformsStyle = ViewStyle>(
```
terrysahaidak, line 423: You can use any styles you want here, not just View's styles

renanmav: Yeah, that’s why `T` extends `FlexStyle` and `TransformsStyle`. The default would be `ViewStyle` if generic isn’t passed. Check react-native types:
- [TextStyle](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cbb32bd37e8c941bd802c9eff19a845b7fe65043/types/react-native/index.d.ts#L784)
- [ImageStyle](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cbb32bd37e8c941bd802c9eff19a845b7fe65043/types/react-native/index.d.ts#L3531)

react-native-reanimated.d.ts
```diff
@@ -56,6 +64,8 @@ declare module 'react-native-reanimated' {
     class AnimatedValue<T extends Value> extends AnimatedNode<T> {
       constructor(value?: T);
 
+      value: T;
```
wcandillon, line 67: Shouldn't it be a separate type?

renanmav: Sure, I’ll separate it.

renanmav:
After some local usage of these types (patch-package to enjoy the types), I saw that generics don’t work for `useAnimatedGestureHandler` and `useAnimatedStyle`, maybe there is something related to the UI Thread not being able to parse the Typescript type annotation. I also saw I’ve forgot the `interpolation` function type.

react-native-reanimated.d.ts
```diff
@@ -371,6 +381,29 @@ declare module 'react-native-reanimated' {
       config: DecayConfig,
     ): BackwardCompatibleWrapper;
 
+    // reanimated2 animations
+    export function withTiming(
+      toValue: number,
+      userConfig?: Omit<TimingConfig, 'toValue'>,
+      callback?: (isCancelled: boolean) => void,
+    ): number;
```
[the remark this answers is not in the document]
renanmav: I’m doing this to make the imperative compatible with:
1. SharedValue
2. react-native styles (ViewStyle, TextStyle, ImageStyle)
Thoughts @kmagiera?

react-native-reanimated.d.ts
```diff
@@ -371,6 +381,29 @@ declare module 'react-native-reanimated' {
...
+    export function withSpring(
+      toValue: number,
+      userConfig?: Omit<SpringConfig, 'toValue'>,
+      callback?: (isCancelled: boolean) => void,
+    ): number;
+    export function cancelAnimation<T extends Value>(
+      sharedValue: AnimatedValue<T>
```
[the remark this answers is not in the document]
renanmav: Updated it

terrysahaidak:
> After some local usage of these types (patch-package to enjoy the types), I saw that generics don’t work for `useAnimatedGestureHandler` and `useAnimatedStyle`, maybe there is something related to the UI Thread not being able to parse the Typescript type annotation.

Yeah, right now typescript annotations aren't getting removed from worklets before babel plugin processing them.

react-native-reanimated.d.ts
```diff
@@ -64,13 +63,17 @@ declare module 'react-native-reanimated' {
...
       constructor(value?: T);
 
-      value: T;
-
       setValue(value: Adaptable<T>): void;
 
       interpolate(config: InterpolationConfig): AnimatedNode<number>;
     }
 
+    class SharedValue<T extends Value> extends AnimatedNode<T> {
```
terrysahaidak, line 71: It should not extend AnimatedNode, since it has nothing to do with AniimatedNode.

For example, by extending it, you inherit setValue and initerpolate. But Shared Values doesn't have these.

terrysahaidak:
@renanmav actually, Shared Value is not a class in the code, I would just use type/interface here. There is no way to `new` it :)

Also, as I mentioned before, `T extends Value` but it' not correct since Value - `string | number | boolean`. But you can pass also an object, array, and even function here.

wcandillon:
@renanmav Thank you so much for your effort. With now type annotations supported with worklets, things are getting interesting :) I'm also thinking about an eslint plugin to check the env variables used from worklets 😅

I'm trying your definition and added a definition for `onRunUI()`

wcandillon:
@terrysahaidak how can we deal with assigning animations to shared values? Here `withTiming` return value is typed as number. But it is not a number. I'm wondering what is the most elegant to deal with this.

renanmav:
> I'm trying your definition and added a definition for `onRunUI()`

I’m not sure what `onRunUI` means. Would you mind clarify it @wcandillon?

wcandillon:
runOnUI call a worklet function from the JS thread. You can pass it parameter. You can see an example of it at https://github.com/wcandillon/can-it-be-done-in-react-native/blob/master/reanimated-2/src/Worklets/Worklets.tsx#L53

renanmav:
So, can I add your `runOnUI` definition in this PR @wcandillon?

terrysahaidak:
Here is the implementation of runOnUI which can infer params and ReturnType

```ts
function runOnUI<A extends any[], R>(fn: (...args: A) => R): (...args: Parameters<typeof fn>) => ReturnType<typeof fn> {
  return fn;
}
```

Not sure we should support return type though.

Here is without return type:

```
function runOnUI<A extends any[]>(fn: (...args: A) => void): (...args: Parameters<typeof fn>) => void {
  return fn;
}

```

wcandillon:
@terrysahaidak I fear that this type definition won't type `args` as a tuple. For instance, for `foo(string, number)`,  it will be `(string | number)[] `instead of `[string, number]`.

terrysahaidak:
Also, I have a problem with this definition of SharedValue

This code

```ts
type RawSharedValue = number | string | boolean | object | Function;
type SharedValueType = RawSharedValue | RawSharedValue[];
interface SharedValue<T extends SharedValueType> {
  value: T;
}

export function useSharedValue<T extends SharedValueType>(
  initialValue: T,
): SharedValue<T> {
  return {
    value: initialValue,
  };
}

const sv = useSharedValue(0);
```

Makes sv of type of `SharedValue<0>` but not a number. You cannot assign different value than 0 then.

terrysahaidak:
> @terrysahaidak I fear that this type definition won't type `args` as a tuple. For instance, for `foo(string, number)`, it will be `(string | number)[] `instead of `[string, number]`.

do you have a suggestion on how to still have types for the returning function and support types?

wcandillon:
@renanmav I added `processColor` to the PR :) https://github.com/renanmav/react-native-reanimated/pull/1/files#diff-c3ef4cab7a703f83e079356d6873fef5

wcandillon:
@terrysahaidak That type definition works indeed! Nice that typescript supports this now. @renanmav I updated the PR

wcandillon:
@renanmav To fix @terrysahaidak concern with value types, you can simply write: 
```
  interface SharedValue<T> {
      value: T;
    }
```

That will do the job.

react-native-reanimated.d.ts
```diff
@@ -380,8 +424,63 @@ declare module 'react-native-reanimated' {
       initialValue: T
     ): AnimatedValue<T>;
 
+    // reanimated2 functions
+    export function runOnUI<A, R>(fn: (...args: A) => R): (...args: Parameters<typeof fn>) => void;
```
terrysahaidak, line 428: It should be:

```suggestion
   export function runOnUI<A extends any[], R>(fn: (...args: A) => R): (...args: Parameters<typeof fn>) => void;
```

terrysahaidak:
I think SharedValue type should be exposed as a public API

react-native-reanimated.d.ts
```diff
@@ -380,8 +424,63 @@ declare module 'react-native-reanimated' {
...
+    export function processColor(color: number | string): number;
+                        
+    // reanimated2 hooks
+    export function useSharedValue<T extends SharedValueType>(
+      initialValue: T
+    ): SharedValue<T>;
+    export function useDerivedValue<T extends SharedValueType>(
+      processor: () => T
+    ): SharedValue<T>;
+    export function useAnimatedStyle<T extends ViewStyle, TextStyle, ImageStyle>(
```
terrysahaidak, line 438: this works way better for me

```suggestion
    export function useAnimatedStyle<T extends AnimateStyle>(
```

react-native-reanimated.d.ts
```diff
@@ -371,6 +392,29 @@ declare module 'react-native-reanimated' {
...
+      toValue: number,
+      userConfig?: Omit<TimingConfig, 'toValue'>,
+      callback?: (isCancelled: boolean) => void,
+    ): number;
+    export function withSpring(
+      toValue: number,
+      userConfig?: Omit<SpringConfig, 'toValue'>,
+      callback?: (isCancelled: boolean) => void,
+    ): number;
+    export function cancelAnimation<T extends Value>(
```
terrysahaidak, line 406:
```suggestion
    export function cancelAnimation<T extends SharedValue>(
```

terrysahaidak:
Ok, based on @satya164 suggestion, this seems to be working, but now SharedValue doesn't have a constraint:

```ts
type RawSharedValue = number | string | boolean | object | Function;
export type SharedValueType = RawSharedValue| RawSharedValue[];

type Value<T> = T extends number
  ? number
  : T extends string
  ? string
  : T extends boolean
  ? boolean
  : T;

type SharedValue<T> = {
  value: T;
};

export function useSharedValue<T extends SharedValueType>(
  initialValue: Value<T>,
): SharedValue<Value<T>> {
  return {
    value: initialValue,
  };
}

const sv = useSharedValue(1); // SharedValue<number>
```

renanmav:
> ```ts
> type Value<T> = T extends number
>   ? number
>   : T extends string
>   ? string
>   : T extends boolean
>   ? boolean
>   : T;
> ```

This feels weird.

react-native-reanimated.d.ts
renanmav: I’m not sure about this. `AnimateStyle` needs a mandatory generic, as you can see below.
https://github.com/software-mansion/react-native-reanimated/blob/fffcb9f541dd2d2009f8bb68283e81b62d16870c/react-native-reanimated.d.ts#L167

terrysahaidak:
> > ```ts
> > type Value<T> = T extends number
> >   ? number
> >   : T extends string
> >   ? string
> >   : T extends boolean
> >   ? boolean
> >   : T;
> > ```
> 
> This feels weird.

Yeah, but at least it works. The current your version won't work without explicit generic type to be used. 

How does useState work? I believe it infers the type based on value too.

terrysahaidak:
Another one by @xaviergonz

This one is way less weird, it has constraint for SharedValue, but returns never for unsupported value types.

```ts
type RawSharedValue = number | string | boolean | object | Function;
export type SharedValueType = RawSharedValue | RawSharedValue[];

type SharedValue<T extends SharedValueType> = {
  value: T;
};

// lets not use extends on the function generic, which causes it to be more specific
// but let's do it on the return type instead
export function useSharedValue2<T>(
  initialValue: T
): T extends SharedValueType ? SharedValue<T> : never {
  return {
    value: initialValue
  } as any;
}

const sv = useSharedValue2(0); // SharedValue<number>

const sv1 = useSharedValue(null); // never
```

terrysahaidak:
By the way, I just realized we should get rid of Function type in useSharedValuy, because it creates mutable, which doesn't work for functions.

So

```ts
type RawSharedValue = number | string | boolean | object
```

renanmav:
> ```
>   interface SharedValue<T> {
>       value: T;
>     }
> ```
> 
> 
> That will do the job, the same way than for `useState`.

Why @wcandillon solution wouldn’t work @terrysahaidak?

terrysahaidak:
Because SharedValue should be a public API and now we lose any validation "what can be a shared value".

Types should not just describe types but also add validation in the first place.

react-native-reanimated.d.ts
```diff
@@ -380,6 +424,68 @@ declare module 'react-native-reanimated' {
...
+
+    // reanimated2 hooks
+    // lets not use extends on the function generic, which causes it to be more specific
+    // but let's do it on the return type instead
+    export function useSharedValue<T>(
+        initialValue: T
+    ): T extends SharedValueType ? SharedValue<T> : never {
+      return {
+        value: initialValue
+      } as any;
```
renanmav, line 439: I’m not sure about this `any`

terrysahaidak: I think you can remove function implementation from type definition :)

react-native-reanimated.d.ts
```diff
@@ -371,6 +392,29 @@ declare module 'react-native-reanimated' {
...
+      toValue: number,
+      userConfig?: Omit<TimingConfig, 'toValue'>,
+      callback?: (isCancelled: boolean) => void,
+    ): number;
+    export function withSpring(
+      toValue: number,
+      userConfig?: Omit<SpringConfig, 'toValue'>,
+      callback?: (isCancelled: boolean) => void,
+    ): number;
+    export function cancelAnimation<T extends SharedValue>(
```
renanmav, line 406: Also, this generic type 'SharedValue' requires 1 type argument.

terrysahaidak: try to use SharedValue<any> here

react-native-reanimated.d.ts
renanmav: Oh, yeah, silly me

renanmav:
Shall we merge it?

terrysahaidak:
> Shall we merge it?

Do you have some big project to test it?

I recently migrated my project to TS, I won't have time to verify it's still working so would be cool if you check it out:

https://github.com/terrysahaidak/reanimated-gallery

wcandillon:
Last time I tried this PR it was in decent shape, I would be happy to take a look at it again.

renanmav:
@terrysahaidak just tested against your reanimated-gallery repo and I found out problems like this:

image

terrysahaidak:
@renanmav That's because I use my custom shared value type for vectors. So it has a different kind of SharedValue. Check out vectors.ts file.

Also, please check out if styles work. I remember I had some issues. That's the reason I used AnimatedStyle generic instead of your proposal.

Thank you for an awesome work!

renanmav:
@terrysahaidak @wcandillon I added some type checks, same as we do in DefinetelyTyped

dested:
I am running into a strange issue with useSharedValue in the current type defs. 

You can see it here: https://tinyurl.com/ybvhh7uf

If you use a string union for the type of the shared value, if you test its value, you can't **set** it to another value after. I've never really experienced this in typescript (and I have been doing typescript forever). It seems once the type of the generic is narrowed it cannot be assigned to something else, which is really strange. 

If you change `T extends SharedValueType ? SharedValue<T> : never` to simply `SharedValue<T>` the error goes away, which leads me to believe its related to typescript distributing the string union.

Combing through this thread I found @terrysahaidak's comment, which while it does not throw any error, I lose the safety of the string union.

https://tinyurl.com/y8ebmmh2


I'm not well versed enough in this framework to make a recommendation on how to change it yet :-/

wcandillon:
@dested what about such snippet: https://www.typescriptlang.org/play?#code/C4TwDgpgBASghgdwMoAs4CcIBMBqcA2ArtALxQB2hAtgEYTpQA+UAzsOgJbkDmTUNAewH4IccnwE0AVhADGwANwAoUJCioM2PEQgAVcKViINmXAWJ94yNKe3EA2gF1lEAB5gB6YFFXQTW8wgAHl0oN2AIciwWdRsAnX1IAD4oMgBvJSgoADdAgC4oXWUAX2UlADNCcnkOAXFCFgh-Mx0QsNcIqJirZrsIJIAKTKguDmAOAj6C3SUASgLewJCUjKzMYEJ0cTSc-JHyMYn8PqhSpWKlJVk6tigpVKgGpriW4iCAIjh3vneab+Z3rJ3oNfu9ZsoOOUoAMpAA6XI6VIkMiAsFQVZ3eGBB6fd7KLJwhEWFHcPHnJRAA?

dested:
> @dested what about such snippet: https://bit.ly/3dDQaOh

This works for me! It has to be the extends causing the union to be distributed over the SharedValue<T> turning it into `SharedValue<'a'>|SharedValue<'b'>|SharedValue<'c'>` instead of `SharedValue<'a'|'b'|'c'>`, which narrows differently with the if statement. 

Thanks!

terrysahaidak:
but this will case problem I described above

dested:
@terrysahaidak @wcandillon 

Alright, using a fair amount of sorcery I was able to force typescript to stop distributing the union while maintaining the extends:

https://bit.ly/3dJW5RQ

The trick was 

```typescript
export function useSharedValue2<T>(
  initialValue: T
): [T] extends [SharedValueType] ? SharedValue<T> : never {
  return {
    value: initialValue
  } as any;
}
```

Wrapping the types in an array is sufficient enough to force typescript to not distribute the union. This seems to cover the null case as well as the union case!

wcandillon:
@renanmav I'm currently reviewing the typings and this is incredible work. I noticed that the transform, rotate, rotateZ = number is not supported. You might want to add it to the testing (for instance `{ rotate: 0 }`).

wcandillon:
@renanmav With the changes suggested at https://github.com/renanmav/react-native-reanimated/pull/2, it's a green light for me 💚

wcandillon:
@renanmav Great work! about the CI failure, could it be just a matter of merging the latest master into this branch?

@terrysahaidak Does this work for you as well?

react-native-reanimated.d.ts
```diff
@@ -371,6 +392,29 @@ declare module 'react-native-reanimated' {
...
 
+    // reanimated2 animations
+    export function withTiming(
+      toValue: number,
+      userConfig?: Omit<TimingConfig, 'toValue'>,
+      callback?: (isCancelled: boolean) => void,
+    ): number;
+    export function withSpring(
+      toValue: number,
+      userConfig?: Omit<SpringConfig, 'toValue'>,
```
terrysahaidak, line 403: SpringConfig is missing `velocity: number`

wcandillon:
Thks for your review @terrysahaidak 
On top of that I would merge the proc change from https://github.com/software-mansion/react-native-reanimated/pull/890/files

```
    export function proc<A extends Adaptable[]>(
    ): typeof cb;
```

@renanmav you rock!

react-native-reanimated.d.ts
renanmav: Done in 5fd666ea7702ff1a9a7a6f10d78b0fe3aa768521

terrysahaidak:
@renanmav velocity is not required AFAIK.

wcandillon:
@renanmav Could you fix the merge conflict?
@jakub-gonet this is good to go.

wcandillon:
Is a proposal for the merge conflict: https://github.com/renanmav/react-native-reanimated/pull/3/files

tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+  "compilerOptions": {
+    "target": "es2015",
+    "module": "es2015",
+    "jsx": "react-native",
+    "noEmit": true,
+    "allowSyntheticDefaultImports": true,
+    "moduleResolution": "node",
+    "lib": ["es2015"]
+  },
+  "files": ["react-native-reanimated-tests.tsx", "react-native-reanimated.d.ts"]
```
jakub-gonet, line 11: I wonder if it's a good place for the ts tests, it requires adding more things to the devDependencies.

Maybe we could include it in the test project at some point?

/cc @Szymon20000

terrysahaidak:
The sequence helper has been added recently. Also, Loop has been renamed to repeat
https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/animations.js#L332
https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/animations.js#L371

likern:
> The sequence helper has been added recently. Also, Loop has been renamed to repeat
> https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/animations.js#L332
> https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/animations.js#L371

@renanmav  This pull request should fix this https://github.com/renanmav/react-native-reanimated/pull/4/files

tsconfig.json
Szymon20000: What's wrong with adding more to the devDeps? I

jakub-gonet: Actually nothing, but seems strange to have RNGH as devDep.

I guess we can leave it in this shape for now.

renanmav:
Shall we merge this?

jakub-gonet:
🎉

wcandillon:
Woah congrats everyone 🙌🏻 @renanmav you're our TypeScript hero 🦸🏻‍♂️

MinimogDev:
Do you guys have an eta on when these will become available in alpha release?

wcandillon:
you can use https://www.npmjs.com/package/patch-package in the meantime.

PR #842, opened 2020-05-30
Add lightbox example

terrysahaidak:
## Description

This PR implements a Lightbox example. Also, it may be used as an example of a shared element transition.

## Changes

Add Lightbox example

## Screenshots / GIFs

promo_lightbox


Example/src/LightboxExample.js
```diff
@@ -0,0 +1,259 @@
...
+  return (
+    <TouchableWithoutFeedback
+      style={containerStyle}
+      onPress={() => onPress(ref, item, opacity)}>
+      <AnimatedImage ref={ref} source={{ uri: item.uri }} style={styles} />
+    </TouchableWithoutFeedback>
+  );
+}
+
+function useSharedVector(x, y) {
```
karol-bisztyga, line 80: both `useSharedVector` and `timingConfig` are used only in `ImageTransition` so they could be declared inside that component

Example/src/LightboxExample.js
```diff
@@ -0,0 +1,259 @@
...
+    backdropOpacity.value = withTiming(1, timingConfig);
+  })();
+
+  return (
+    <View style={StyleSheet.absoluteFillObject}>
+      <Animated.View style={[styles.backdrop, backdropStyles]} />
+
+      <PanGestureHandler onGestureEvent={onPan}>
+        <Animated.View style={StyleSheet.absoluteFillObject}>
+          <AnimatedImage source={{ uri: item.uri }} style={imageStyles} />
```
karol-bisztyga, line 205: it could be done using object destructing like `const { uri } = item` and then `<AnimatedImage source={{ uri }} style={imageStyles} />`

karol-bisztyga:
@terrysahaidak could you perform a rebase to with the current master? I could merge the changes then ;)

Example/src/LightboxExample.js
terrysahaidak: I got rid of useSharedVector.

But for timingConfig, I prefer static things to stay static. We don't really need to recreate it for each render :)

Example/src/LightboxExample.js
terrysahaidak: It required another destructuring of the item which is destructured too. But yeah, changed it.

terrysahaidak:
@karol-bisztyga done

karol-bisztyga:
Thank you!

PR #851, opened 2020-06-01
Get rid of FrozenObject limitations and improve caching.

Szymon20000:
This pr changes the way in which we store FrozenObj as a jsi::Value. On iOS object created from jsi::HostObject is missing a basic JS method for instance hasOwnProperty and cannot be passed to Object.assign. In order to get rid of these limitations, we can store FrozenObj as a jsi::Object with a hidden property that represents FrozenObject.  

IMG_20200612_081159

Why there won't be any cycles?
That's because all the edges that are added points in the same direction as the edge from jsi::Object to FrozenObject.

The second part of PR "Improving caching":
WHY:
If we capture worklet (e.g. withTiming) in two other worklets then withTiming will be converted twice to FrozenObject. To avoid it we add a hidden property to withTiming with reference to the FrozenObject.

HOW:
We cannot automatically add a hidden reference to internal cpp objects because input object can be altered in the meantime. That's why we export makeCacheable function which adds hidden property "__reanimatedShouldBeCached". The function should be called only on constant objects.
Worklet functions are cached automatically cause they are constant.

There are two other hidden properties "__reanimatedSharedRef" and "__reanimatedAdaptable".
The first one keeps a reference to the internal cpp object (e.g. FrozenObject) and the second one tells us if we can save jsi::Value in adapt function.  We need the second property because we want to adaptCache only for copy instance, not the original one. 
Example:

```JS
useSharedValue(5);
```
We don't want to save jsi::Value(5) as our js representation. We createHost(rt, mutableValue) instead.

### DIFF
--- a/Common/cpp/SharedItems/Shareable.cpp
+++ b/Common/cpp/SharedItems/Shareable.cpp
@@ -4,6 +4,23 @@
 
 namespace reanimated {
 
+const char *HIDDEN_PROPERTY_NAME = "__reanimatedSharedRef";
+const char *HIDDEN_CAN_BE_ADPTED = "__reanimatedAdaptable";
+const char *HIDDEN_SHOULD_BE_CACHED = "__reanimatedShouldBeCached";
+
+void addHiddenProperty(jsi::Runtime &rt,
+                       jsi::Value &&value,
+                       jsi::Object &obj,
+                       const char *name) {
+  jsi::Object globalObject = rt.global().getPropertyAsObject(rt, "Object");
+  jsi::Function defineProperty = globalObject.getPropertyAsFunction(rt, "defineProperty");
+  jsi::String internalPropName = jsi::String::createFromUtf8(rt, name);
+  jsi::Object paramForDefineProperty(rt);
+  paramForDefineProperty.setProperty(rt, "enumerable", false);
+  paramForDefineProperty.setProperty(rt, "value", value);
+  defineProperty.call(rt, obj, internalPropName, paramForDefineProperty);
+}
+
 void ShareableValue::adaptCache(jsi::Runtime &rt, const jsi::Value &value) {
   // when adapting from host object we can assign cached value immediately such that we avoid
   // running `toJSValue` in the future when given object is accessed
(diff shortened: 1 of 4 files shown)


jgonet (dismissed):

Common/cpp/SharedItems/Shareable.cpp
```diff
@@ -114,7 +181,7 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
     case StringType:
       return jsi::Value(rt, jsi::String::createFromAscii(rt, stringValue));
     case ObjectType:
-      return createHost(rt, frozenObject);
+      return createFrozenWrapper(this, rt, frozenObject);
```
kmagiera, line 184: would it make sense to use the same approach with remote object and mutable object. They won't support enumeration now as well and it seem like we can follow a very similar pattern there too and get rid of `createHost` completely

Common/cpp/SharedItems/Shareable.cpp
```diff
@@ -4,6 +4,20 @@
 
 namespace reanimated {
 
+const std::string HIDDEN_PROPERTY_NAME = "__reanimatedSharedRef";
```
kmagiera, line 7: why can't this be a c string or a constant, we only use it as char* everywhere

Common/cpp/SharedItems/Shareable.cpp
```diff
@@ -73,6 +110,22 @@ void ShareableValue::adapt(jsi::Runtime &rt, const jsi::Value &value, ValueType
   } else {
     throw "Invalid value type";
   }
+  
+  if (value.isObject()) {
```
kmagiera, line 114: have you checked this with arrays? Array data is stored under `frozenArray` and afaik isObject will return true.

The same can be a problem with host functions

Common/cpp/SharedItems/Shareable.cpp
```diff
@@ -15,6 +29,29 @@ void ShareableValue::adaptCache(jsi::Runtime &rt, const jsi::Value &value) {
 }
 
 void ShareableValue::adapt(jsi::Runtime &rt, const jsi::Value &value, ValueType objectType) {
+  if (value.isObject()) {
+    jsi::Object object = value.asObject(rt);
+    
+    if (object.hasProperty(rt, HIDDEN_PROPERTY_NAME.c_str()) and
```
kmagiera, line 35: we don't need hasProperty check. See below where we test for existence of `__worklet` property

Common/cpp/SharedItems/Shareable.cpp
```diff
@@ -101,6 +154,20 @@ jsi::Object ShareableValue::createHost(jsi::Runtime &rt, std::shared_ptr<jsi::Ho
   return jsi::Object::createFromHostObject(rt, host);
 }
 
+jsi::Value createFrozenWrapper(ShareableValue *sv, jsi::Runtime &rt, std::shared_ptr<FrozenObject> frozenObject) {
+  jsi::Object __reanimatedSharedRef = sv->createHost(rt, frozenObject);
```
kmagiera, line 158: I think that with this approach we no longer need host objects to actually expose any properties. We only need host objects to keep references to data that we need to use when running `toJS` for the first time.

kmagiera (commented):
Added a bunch of comments

Common/cpp/SharedItems/Shareable.cpp
Szymon20000: I don't think so. We still have to control access to mutable and remote.

Common/cpp/SharedItems/Shareable.cpp
Szymon20000: I tried but with the current "adapt" approach it's hard for remoteObj and we still want to control access to mutable.

PR #853, opened 2020-06-02
V2 fix withSpring user config velocity usage

terrysahaidak:
## Description

Noticed my velocity isn't getting applied to withSpring. Then noticed it resets it on start.

This change fixes this. ChatHeads example works great with that change.

## Changes

- Fix usage of passed velocity to withSpring

### DIFF
--- a/src/reanimated2/animations.js
+++ b/src/reanimated2/animations.js
@@ -161,7 +161,6 @@ export function withSpring(toValue, userConfig, callback) {
       animation.velocity = previousAnimation.velocity || 0;
       animation.lastTimestamp = previousAnimation.lastTimestamp || now;
     } else {
-      animation.velocity = 0;
       animation.lastTimestamp = now;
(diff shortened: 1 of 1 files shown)


src/reanimated2/animations.js
```diff
@@ -161,7 +161,6 @@ export function withSpring(toValue, userConfig, callback) {
       animation.velocity = previousAnimation.velocity || 0;
```
Szymon20000, line 161: I think we also have a bug here.  Probably it should be: 
`previousAnimation.velocity || animation.velocity || 0;`

terrysahaidak: I think this makes sense

terrysahaidak, line 161:
```suggestion
      animation.velocity = previousAnimation.velocity || animation.velocity || 0;
```

PR #860, opened 2020-06-03
Fix error when style in an array is null

osdnk:
## Description

Maybe it's not obvious but the style in styles' array can also be null and it's handled by RN and the previous version of reanimated. Thus I think we shouldn't make the migration breaking in this aspect. Filtering is quite an easy trick to make it workable. 

btw, I think in the current implementation we're ignoring that styles can also be nested arrays, but it's a nit.


kmagiera:
Closing as of #923

PR #862, opened 2020-06-04
Fix problem with typescript in worklets.

Szymon20000:


plugin.js
```diff
@@ -223,50 +223,71 @@ function processWorkletFunction(t, fun) {
   );
 }
 
+function processIfWorkletNode(t, path) {
+  const fun = path;
```
jkadamczyk, line 227: if `fun` means function, can it be named `function` instead of `fun`?

jkadamczyk: `function` is a keyword, nvm 😄

PR #878, opened 2020-06-09
Updated useAnimatedGestureHandler implementation (#835)

### LINKED ISSUE #835: V2 alpha: Unable to run TapGestureHandler-based animation more than once
## Description

Unsure if this is a bug, but when using `<TapGestureHandler />` and reanimated v2 (which is amazing!), I'm unable to re-trigger my animation after it finishes. It's pretty simple - on tap start, set the scale from 1.0 to 0.9. On tap end, reset back to 1.0. It works once, then not again. Having migrated from v1 to v2, I suspect it may be an underlying issue with the clock? I didn't see anything documentation suggesting that anything other than what's below is required for triggering this kind of simple animation more than once. The basic random width example works fine.

## Code

```js
import React from 'react'
import { View } from 'react-native'
import Animated, {
  useSharedValue,
  withSpring,
  useAnimatedStyle,
  useAnimatedGestureHandler,
import { TapGestureHandler } from 'react-native-gesture-handler';

const styles = {
  container: {
    flex: 1,
    backgroundColor: 'rgba(200, 0, 0, 0.5)',
    justifyContent: 'center',
    alignItems: 'center'
  },
  box: {
    width: 140,
    height: 300,
    backgroundColor: 'rgba(200, 200, 200, 0.6)',
    borderRadius: 26
  }
}

export function ReanimatedExample() {
  const s = useSharedValue(1);

  const tapGestureHandler = useAnimatedGestureHandler({
    onStart: (event, ctx) => {
      s.value = withSpring(0.9);
    },
    onEnd: (event, ctx) => {
      s.value = withSpring(1);
    },
  });

    return {
      transform: [
        { scale: s.value },
      ],
    };
  });

  return (
      <TapGestureHandler onGestureEvent={tapGestureHandler}>
        <Animated.View style={[
          styles.box,
          animatedStyle
        ]} />
      </TapGestureHandler>
    </View>
  );
}
```

### Package versions

- React: v16.11.0
- React Native: v0.62.2
- React Native Reanimated: v2.0.0-alpha.1
- React Native Gesture Handler: v1.6.0

divykj:
## Description

Updated the gesture handler's triggering logic in useAnimatedGestureHandler hook in v2.

Fixes #835

### Notes

- Along with the mentioned issue, it should also resolve a group of similar issues that aren't posted.
- Tested it with @terrysahaidak 's reanimated-gallery, and some custom test cases (for TapGestureHandler).
- This is a conservative implementation (it also checks for the event.oldState, which might not completely necessary)
-  I could do a small refactor maybe to use less comparisons (waiting for reviewer's opinions)

Edit: tagged the wrong person :-/

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -304,36 +304,41 @@ export function useAnimatedGestureHandler(handlers) {
   return useEvent(
     event => {
       'worklet';
-      const UNDETERMINED = 0;
       const FAILED = 1;
+      const BEGAN = 2;
       const CANCELLED = 3;
       const ACTIVE = 4;
       const END = 5;
 
-      if (event.oldState === UNDETERMINED && handlers.onStart) {
+      if (event.state === BEGAN && handlers.onStart) {
         handlers.onStart(event, context);
       }
       if (event.state === ACTIVE && handlers.onActive) {
         handlers.onActive(event, context);
       }
-      if (event.oldState === ACTIVE && event.state === END && handlers.onEnd) {
+      if (
+        (event.oldState === BEGAN || event.oldState === ACTIVE) &&
+        event.state === END && handlers.onEnd
+      ) {
         handlers.onEnd(event, context);
       }
       if (
-        event.oldState === ACTIVE &&
-        event.state === FAILED &&
-        handlers.onFail
+        (event.oldState === BEGAN || event.oldState === ACTIVE) &&
+        event.state === FAILED && handlers.onFail
       ) {
         handlers.onFail(event, context);
       }
       if (
-        event.oldState === ACTIVE &&
-        event.state === CANCELLED &&
-        handlers.onCancel
+        (event.oldState === BEGAN || event.oldState === ACTIVE) &&
+        event.state === CANCELLED && handlers.onCancel
       ) {
         handlers.onCancel(event, context);
       }
-      if (event.oldState === ACTIVE && handlers.onFinish) {
+      if (
+        (event.oldState === BEGAN || event.oldState === ACTIVE) &&
+        event.state !== BEGAN && event.state !== ACTIVE &&
+        handlers.onFinish
+      ) {
         handlers.onFinish(
           event,
           context,
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -316,22 +316,13 @@ export function useAnimatedGestureHandler(handlers) {
       if (event.state === ACTIVE && handlers.onActive) {
         handlers.onActive(event, context);
       }
-      if (
-        (event.oldState === BEGAN || event.oldState === ACTIVE) &&
-        event.state === END && handlers.onEnd
-      ) {
+      if (event.oldState === ACTIVE && event.state === END && handlers.onEnd) {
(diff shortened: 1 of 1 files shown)


divykj:
Hey @terrysahaidak, could you please review this?

Thank you.

src/reanimated2/Hooks.js
```diff
@@ -304,36 +304,41 @@ export function useAnimatedGestureHandler(handlers) {
...
-      if (event.oldState === UNDETERMINED && handlers.onStart) {
+      if (event.state === BEGAN && handlers.onStart) {
         handlers.onStart(event, context);
       }
       if (event.state === ACTIVE && handlers.onActive) {
         handlers.onActive(event, context);
       }
-      if (event.oldState === ACTIVE && event.state === END && handlers.onEnd) {
+      if (
+        (event.oldState === BEGAN || event.oldState === ACTIVE) &&
```
terrysahaidak, line 320: Not sure about this. I don't see any docs regarding this in gesture handler. But I think it should check only for Active. Because onEnd should be triggered after Active.

src/reanimated2/Hooks.js
```diff
@@ -304,36 +304,41 @@ export function useAnimatedGestureHandler(handlers) {
...
+        event.state === END && handlers.onEnd
+      ) {
         handlers.onEnd(event, context);
       }
       if (
-        event.oldState === ACTIVE &&
-        event.state === FAILED &&
-        handlers.onFail
+        (event.oldState === BEGAN || event.oldState === ACTIVE) &&
```
terrysahaidak, line 326: According to this https://docs.swmansion.com/react-native-gesture-handler/docs/state.html#state-flows it should check only for Began

src/reanimated2/Hooks.js
```diff
@@ -304,36 +304,41 @@ export function useAnimatedGestureHandler(handlers) {
...
+        event.state === FAILED && handlers.onFail
       ) {
         handlers.onFail(event, context);
       }
       if (
-        event.oldState === ACTIVE &&
-        event.state === CANCELLED &&
-        handlers.onCancel
+        (event.oldState === BEGAN || event.oldState === ACTIVE) &&
```
terrysahaidak, line 332: And according again to this flow it should check for active only https://docs.swmansion.com/react-native-gesture-handler/docs/state.html#state-flows

divykj: The way I interpreted the docs was that those were some of the few possible flows, but there could be more possible flows depending on the situation. But I think you are probably right.
 
Also I mentioned this in the PR that I was just being conservative in checking the conditions, so I could get a proper review.

So i'll go ahead with working on all of your suggestions.

terrysahaidak:
Cool. Thank you. Hey @kmagiera, would be cool to hear your feedback on this.

divykj:
Hey, thank you, this is my first PR here, I was very excited for this, hope I could contribute more in the future!

PS: Just my excitement talking, ignore it.

osdnk:
Thank you so much. Is there a chance to make a release with that, @Szymon20000?

osdnk:
@Szymon20000 @karol-bisztyga

PR #882, opened 2020-06-10
Avoid unnecessary babel polyfills and fix problem with nested worklets.

Szymon20000:
The Typescript [fix](https://github.com/software-mansion/react-native-reanimated/pull/862) revealed another issue with overzealous babel polyfills. The pull-request solves two most important problems caused by  @babel/plugin-transform-object-assign and @babel/plugin-transform-parameters.

Fixes problem with nested worklets by adding 'global' to our blacklist and providing dummy __reanimatedWorkletInit implementation on UI side.

Additional info: 

1 "Object.assign" 
Babel plugin "@babel/plugin-transform-object-assign" transforms Object.assign() call into (0, _extend.default)() where _extend is a module with Object.assign polyfill. Because the plugin do that we capture _extend object and bad things happen.  This pr renames "Object" to "Object__DO_NOT_TRANSFOR" in each Object.assign call on entry and reverts this change on exit.

2 "function(...args)"
Babel plugin @babel/plugin-transform-parameters transforms spread syntax and uses arguments instead. To prevent "arguments" variable from being captured the pr add it to the blacklist.

3 Nested Worklets problem
```Js
function out(easing) {
  'worklet';
  return t => {
    'worklet';
    return 1 - easing(1 - t);
  };
}
``` 
-> 
```JS
var out = function () {
    var _f = function _f(easing) {
      return function () {
        var _f = function _f(t) {
          return 1 - easing(1 - t);
        };

        _f._closure = {
          easing: easing
        };
        _f.asString = "function(t){const{easing}=this._closure;{return 1-easing(1-t);}}";
        _f.__workletID = 615968805831595100;

        return _f;
      }();
    };

    _f._closure = {
      global: global
    };
    _f.asString = "function(easing){const{global}=this._closure;{return function(){var _f=function _f(t){return 1-easing(1-t);};_f._closure={easing:easing};_f.asString=\"function(t){const{easing}=this._closure;{return 1-easing(1-t);}}\";_f.__workletID=615968805831595100;global.__reanimatedWorkletInit(_f);return _f;}();}}";
    _f.__workletID = 571397156231203700;

    return _f;
  }();
```

As we can see global is captured.

### DIFF
--- a/Common/cpp/Tools/RuntimeDecorator.cpp
+++ b/Common/cpp/Tools/RuntimeDecorator.cpp
@@ -7,6 +7,20 @@ namespace reanimated {
 
 void RuntimeDecorator::addNativeObjects(jsi::Runtime &rt, UpdaterFunction updater, RequestFrameFunction requestFrame) {
   rt.global().setProperty(rt, "_WORKLET", jsi::Value(true));
+  
+  jsi::Object dummyGlobal(rt);
+  auto dummyFunction = [requestFrame](
+     jsi::Runtime &rt,
+     const jsi::Value &thisValue,
+     const jsi::Value *args,
+     size_t count
+     ) -> jsi::Value {
+   return jsi::Value::undefined();
+  };
+  jsi::Function __reanimatedWorkletInit = jsi::Function::createFromHostFunction(rt, jsi::PropNameID::forAscii(rt, "__reanimatedWorkletInit"), 1, dummyFunction);
+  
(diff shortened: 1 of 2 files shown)


kmagiera:
Re pt 3 I think we should transform all the worklets. In the example you provided the way `out` function can be used is should support the following usecases:
1) `out` is called on the JS side, and the result is passed to UI and expected to be a worklet
2) `out` is called on the UI side and the result is used on the JS side

If we only transform `out` and not the inner worklet, we wouldn't be able to support 1st usecase. If we only transformed the inner and not outer we wouldn't be able to support 2nd worklet.

PR #883, opened 2020-06-10
Support recursive worklets

Szymon20000:
1 Pass closure by global object instead of "this expression".   
2 Use function name in function expression.

Before:
```JS
function () { const { } = this.closure_;
```
After
``` JS
function name () { const {} = jsThis.closure_; { name(sth)}
```

Additional Info:
"jsThis" is a global's property that is changed just before function call and reverted after.
We have to restore a previous value because it may be needed later.
Example
```JS 
function worklet1(x) {
  'worklet';
  worklet2();
  // sth from closure
}
```
remember old jsThis -> set worklet1's jsThis -> call worklet1 -> remember old jsThis (worklet1's) 
-> set worklets2's jsThis -> call worklet2 -> restore old jsThis -> restore old jsThis.

Why do we need this change?
Example:
```JS
function fib(n) {
  'worklet';
  console.log("fib", n);
  if (n == 0) {
     return 1;
  } 
  if (n == 1) {
     return 1;
  }
  return fib(n-1) + fib(n-2);
}
```
We do not pass "this" value while calling fib (for instance like: fib.apply(this, [n]) ). Because of that fib will get null as "this" object and it will impossible to get console variable from this._closure.

### DIFF
--- a/Common/cpp/SharedItems/Shareable.cpp
+++ b/Common/cpp/SharedItems/Shareable.cpp
@@ -195,7 +195,19 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
                    const jsi::Value *args,
                    size_t count
                    ) mutable -> jsi::Value {
-          return funPtr->callWithThis(rt, *jsThis, args, count);
+           jsi::Value oldJSThis = rt.global().getProperty(rt, "jsThis");
+           rt.global().setProperty(rt, "jsThis", *jsThis); //set jsThis
+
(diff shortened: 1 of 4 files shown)


nicefella:
Hi, 

In the version 2 documentation, it says:

> Recursive worklets aren’t supported (however there are simple workarounds for this problem
[https://docs.swmansion.com/react-native-reanimated/docs/next/about#known-problems-and-limitations](url)

May i ask what these workarounds are?

Thank you,

PR #888, opened 2020-06-11
V2: Fix Color interpolation 🎨

wcandillon:
This example fixes color interpolation. If the shared value is a number, it is a pass-through via UpdateProps. This small change enables this example: https://www.dropbox.com/s/zy7rv95ax5w5bzp/demo.mov?dl=0 that leverages color interpolation on a View but also an SVG component.


Szymon20000:
Can you show your code? How does this change fix anything? 
We check in normalizeColor function if an input value is a number.

PR #890, opened 2020-06-12
Fix: type signature for proc()

wcandillon:
Thanks @terrysahaidak for showing me the ways of the function typing :)

I tried this patch in one of my project that uses procs and it works really well.


terrysahaidak:
This is funny. I tried this same approach and it didn't work for me back then :)

jakub-gonet:
This breaks some types, check out this (maybe a bit artificial) examples:
```js
const procc = proc((undef: undefined, value: Animated.Value<number>) =>
  cond(defined(undef), set(value, 1)),
);
```

or 

```js
const procc = proc((value: Animated.Value<number>, toSet: number) =>
  set(value, toSet),
);
```

terrysahaidak:
I'll try to reproduce it later today

terrysahaidak:
Will this work?

It seems to be for me, but my VSCode laggy for some reason right now.

```ts
type ProcArgs = (Adaptable<number> | undefined)[];
export function proc<A extends ProcArgs>(
): typeof cb;
```

image

jakub-gonet:
I think `Adaptable<any>[]` is the best solution here, could you guys test if it works in your projects?

jakub-gonet:
@wcandillon, @terrysahaidak

terrysahaidak:
yeah, seems to be. didn't know Adaptable works with undefined too.

wcandillon:
@jakub-gonet This change doesn't work for me (it types as any). However, the original proposed change works perfectly for me.

wcandillon:
@jakub-gonet I fixed the code to support @terrysahaidak use case. It's good to go now.

jakub-gonet:
@wcandillon could you rebase this branch to v1? We're using separate branches for alpha v2 code and v1 code atm - this PR should be merged into v1

wcandillon:
That makes sense. 👍🏽

PR #891, opened 2020-06-12
Use function hash as workletID instead of a random number.

Szymon20000:
WHY:
Because we want reproducible builds.

HOW:
Use a hash function from https://www.npmjs.com/package/string-hash-64.

REMOVE DEBUG CONSOLE.LOG !!!!!!!!!

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -158,12 +158,7 @@ function processWorkletFunction(t, fun) {
   const funExpression = t.functionExpression(null, clone.params, clone.body);
 
   const funString = buildWorkletString(t, fun, variables);
-<<<<<<< HEAD
-||||||| constructed merge base
-  const workletID = hash(funString);
-=======
   const workletHash = hash(funString);
->>>>>>> workletID -> workletHash
 
   const newFun = t.functionExpression(
     fun.id,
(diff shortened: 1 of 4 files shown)


plugin.js
```diff
@@ -145,11 +146,11 @@ function processWorkletFunction(t, fun) {
   const variables = Array.from(closure.values());
 
   const privateFunctionId = t.identifier('_f');
+  const workletID = hash(fun.toString());
```
kmagiera, line 149: can we use the result of `buildWorkletString` call few lines below. I think it makes more sense that we make hash based on the actual content of `asString`

Szymon20000: We can but it shouldn't matter cause buildWorkletString is the same for identical functions.

Szymon20000: Done.

plugin.js
```diff
@@ -149,11 +150,11 @@ function processWorkletFunction(t, fun) {
   // if we don't clone other modules won't process parts of newFun defined below
   // this is weird but couldn't find a better way to force transform helper to
   // process the function
-  const workletID = Math.random() * 1e18;
   const clone = t.cloneNode(fun.node);
   const funExpression = t.functionExpression(null, clone.params, clone.body);
 
   const funString = buildWorkletString(t, fun, variables);
+  const workletID = hash(funString);
```
kmagiera, line 157: can we also now rename the attribute "workletHash" to indicate that this is not an identifier?

Szymon20000: Done

kmagiera:
Approved, note that there are some conflicts to be resolved prior to merging

PR #898, opened 2020-06-16
Update mock.js

stackia:
# Description

For now it we call Animated.spring(config).start(myCallback) in tests, myCallback will not be called. This makes it difficult to test the callback code.

# Changes

Updated mocks for .timing, .decay and .spring, added setTimeout() to simulate the callback, such that we can use Jest fake timers to schedule the callback.


kmagiera:
Thanks!

PR #900, opened 2020-06-16
[Android] Improve building.

Szymon20000:
WHY:
To speed up building.

HOW:
1 Added necessary tasks to Android/build.gradle similar to tasks from react-native/ReactAndroid/build.gradle.
2 Created Application.mk 
3 Created Android.mk
4 Add important lines from ReactAndroid/src/main/jni/react/jni/Android.mk (I've tried to import/include this module but it didn't work)
5 update createNPMPackage.sh


createNPMPackage.sh
```diff
@@ -3,62 +3,27 @@ set -e
...
 mv android-npm android
 
-npm pack
+# PART II (create package)
 
-#mv android android-npm
-#mv android-temp android
+npm pack
 
+mv android android-npm
```
kmagiera, line 26: I'd prefer to avoid having that `android-npm` folder and doing this move here. Afraid of being hit one time with a problem when build fails in the middle and the folder isn't restored.

createNPMPackage.sh
```diff
@@ -3,62 +3,27 @@ set -e
...
-  cp "${d}/libreanimated.so" "${ROOT}/android/src/main/JNILibs/${d}/."
-done
-
-cd "${ROOT}"
-rm -r android/build/outputs/aar/*.aar
-
-cd  Example/android/
+# part I - I (clear old aar)
 
+rm -rf android/build/outputs/aar/*.aar
```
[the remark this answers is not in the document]
Szymon20000: Done

createNPMPackage.sh
```diff
@@ -3,62 +3,27 @@ set -e
...
-# clean so
-
-
-cd ../node_modules/react-native/ReactAndroid/build/react-ndk/all
-
-rm -rf "${ROOT}/android/src/main/JNILibs/"
+# part I (prepare android-npm)
 
+cd Example/android
+./gradlew :react-native-reanimated:assembleDebug
```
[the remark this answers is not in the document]
Szymon20000: Done

createNPMPackage.sh
Szymon20000: I don't see what could go wrong here. 
Commands are executed synchronously in our script.
If you see a better solution then I will change it.

Szymon20000: If you want we can create another directory with android-npm (renamed to android) there.  With symlinks to package.json, plugin, iOS, and other dirs/files used during publish/pack.

PR #904, opened 2020-06-17
Enable starting animations from RN thread.

kmagiera:
This change makes it possible to start animations using `withTiming` and similar methods directly on the RN JS thread.

For that to work the following changes were made:
1) On the c++ side, when updating mutable value we now need to call setter. As the value is kept on the "remote" side we run the setter on the UI thread using scheduler
2) On the JS side we need to distinguish between cases when `withTiming` and similar methods are called to determine initial style in `useAnimatedStyle` vs when they are called to be set under `.value` prop. To do that we wrap initial updater calls in `useAnimatedStyle` with a method that sets updates global variable, then read that variable and depending on its state we either return the initial value or the animation.
3) All animation definitions needed to be refactored. We now wrap all animations with `defineAnimation` call, that is responsible for doing the check decsribed in (2). On top of that `defineAnimation` wraps the animation definition such that we no longer return animation object but a worklet which generates that object. This is needed because if the animation object was constructed on the RN JS side, it'd be frozen and UI thred wouldn't be able to modify it while running animations. By passing a worklet we can initialize animation object on the UI side.

Because of the new approach for defining animations, the value setter now must consider two separate cases: when it gets animation object, or when it gets a worklet that outputs animation. We allow for those two options to be returned, otherwise we'd need to make adjustments in the useAnimatedStyle updater which are far more complex.

Note that the changes made to animation functions described in (3) are of a smaller scope that it may seem on the diff. This is because after wrapping animations in `defineAnimation` method we had to indent all the existing code. In `delay` and `loop` methods we transform variables passed to those methods to keep the changeset shorter.

#### Testing

To test this change I modified `AnimatedStyleUpdateExample` to use `withTiming`, `withSpring`, `delay` and `loop` directly in `onPress` handler.

### DIFF
--- a/Common/cpp/SharedItems/Shareable.cpp
+++ b/Common/cpp/SharedItems/Shareable.cpp
@@ -292,14 +292,15 @@ void MutableValue::set(jsi::Runtime &rt, const jsi::PropNameID &name, const jsi:
 
   if (module->isHostRuntime(rt)) {
     if (propName == "value") {
-      {
-        std::lock_guard<std::mutex> lock(readWriteMutex);
-        value = ShareableValue::adapt(rt, newValue, module);
-      }
-      module->scheduler->scheduleOnUI([this] {
-        for (auto listener : listeners) {
-          listener.second();
-        }
+      auto shareable = ShareableValue::adapt(rt, newValue, module);
+      module->scheduler->scheduleOnUI([this, shareable] {
+        jsi::Runtime &rt = *this->module->runtime.get();
+        auto setterProxy = jsi::Object::createFromHostObject(rt, std::make_shared<MutableValueSetterProxy>(shared_from_this()));
+        jsi::Value newValue = shareable->getValue(rt);
+        module->valueSetter->getValue(rt)
+          .asObject(rt)
+          .asFunction(rt)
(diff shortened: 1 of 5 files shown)


kmagiera:
Make sure to check "hide whitespace changes" when reviewing this, othrewise the changeset will be very big

Szymon20000:
Looks fine to me. However, I think we should make Mutable::get asynchronous as well. 
That will allow us to get rid of the mutex. It also will solve a problem with wrong value on RN JS side. With this pull-request we don't know what's the real value of mutable. 
Example:
```JS
const mutable = useSharedValue(5);
mutable.value = 50;
console.log(mutable.value);
```
I think we will get 5 here instead of 50. I think we can fix it in separate pull-request though.

terrysahaidak:
I think ability to read the value synchronously is one of the coolest feature Reanimated provides. Would be cool to keep it.

Szymon20000:
I understand your point of view but It also makes it slower on UI side because each set/get has to acquire the mutex.

kmagiera:
getting rid of mutexes is outside of the scope of this change. We can consider doing this but as a separate change.

Will merge this now, and we can hold the discussion about sync/async value access elsewhere

PR #908, opened 2020-06-19
Fix mappers' execution.

Szymon20000:
WHY:
We have to notify mappers about their dirtiness right away because they won't execute otherwise.
(Actually, they will but on next render).

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -249,16 +249,22 @@ function styleUpdater(viewTag, updater, state) {
   }
 }
 
-export function useAnimatedStyle(updater) {
+export function useAnimatedStyle(updater, exclude = []) {
   const viewTag = useSharedValue(-1);
 
   const initRef = useRef(null);
   if (initRef.current === null) {
     const initial = initialUpdaterRun(updater);
+
+    const toExclude = new Set(exclude);
+    const filteredInput = Object.values(updater._closure).filter(
+      (x) => !toExclude.has(x)
+    );
+
     initRef.current = {
       initial,
       remoteState: makeRemote({ last: initial }),
-      inputs: Object.values(updater._closure),
+      inputs: filteredInput,
     };
   }
   const { initial, remoteState, inputs } = initRef.current;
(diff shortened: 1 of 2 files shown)


kmagiera:
There are many unrelated JS changes that I think might be due to misconfigured prettier. Can you please revert those and look into your prettier config?

src/reanimated2/Hooks.js
```diff
@@ -249,16 +249,22 @@ function styleUpdater(viewTag, updater, state) {
...
+export function useAnimatedStyle(updater, exclude = []) {
   const viewTag = useSharedValue(-1);
 
   const initRef = useRef(null);
   if (initRef.current === null) {
     const initial = initialUpdaterRun(updater);
+
+    const toExclude = new Set(exclude);
+    const filteredInput = Object.values(updater._closure).filter(
+      (x) => !toExclude.has(x)
```
karol-bisztyga, line 261: cannot it be done by `indexOf` simply? Without creating `toExclude`

src/reanimated2/Hooks.js
```diff
@@ -278,12 +284,17 @@ export function useAnimatedStyle(updater) {
 // when you need styles to animated you should always use useAS
 export const useAnimatedProps = useAnimatedStyle;
 
-export function useDerivedValue(processor) {
+export function useDerivedValue(processor, exclude = []) {
   const initRef = useRef(null);
   if (initRef.current === null) {
+    const toExclude = new Set(exclude);
+    const filteredInput = Object.values(updater._closure).filter(
+      (x) => !toExclude.has(x)
```
karol-bisztyga, line 292: cannot it be done by `indexOf` simply? Without creating `toExclude`

kmagiera:
I've been thinking about this issue lately and am not convinced that adding ways of making updates in UAS is the right way to go.

I'd like to think of UAS as a render function in react. Therefore we should expect it not to make any updates. Not sure if this can lead to any problems but my intuition is that we should be more explicit about cases where we want to run some side-effects.

As a consequence, I think it makes the most sense to analyze the usecases where side-effects are desirable and come up with a new API method for executing those. I think currently people are doing side-effects in UAS and UDV only because there is no other place where those would run. From what I've heard people feel like those amends don't belong to UAS where they are put (by that I mean, they are unrelated to styles of a given view but only placed there because there is no other option)

PR #917, opened 2020-06-22
enable ios dev menu

karol-bisztyga:
Just a small change which enables dev menu on iOS. Useful, for instance, when in need of switching fast refresh on and off or reloading an app.


kmagiera:
We should add a note here: https://docs.swmansion.com/react-native-reanimated/docs/installation and add that to playground app as well: https://github.com/software-mansion-labs/reanimated-2-playground

Issue #921, opened 2020-06-23
v2-alpha3 cannot invoke worklet in useAnimatedStyle()


wcandillon:
This bug is shown by this example: https://github.com/software-mansion/react-native-reanimated/pull/915

PR #922, opened 2020-06-23
Fix problem with wrong closure and add possibility to blacklist other plugins.

Szymon20000:
#### WHY:
Some of the plugins that are used by metro-react-native-preset or by babel itself don't update correctly bindings properties in scopes. Because we use bindings in order to find all the missing variables to capture them we have to work with scopes that reflect a current state in AST.

Example:
``` JS
let {x, y} = obj;
``` 
is changed to 
```JS
let _obj = obj,
    x = _obj.x,
    y = _obj.y;
```
by babel-plugin-transform-destructuring.
However, in binding, we only find [x, y].

#### SOLUTION
Work on a copy of worklet tree with proper bindings:
``` JS
const copyWithCorrentBindings = Parser.parse('(' + fun.toString() + '\n)');
traverse(copyWithCorrentBindings, {...visitors}); 
```
----------------------------------------------------
#### WHY:
Most of the babel presets used by React Native developers contains plugins that polyfills mechanisms that are supported by Hermes and JSC. Additionally, some of the polyfills use external code that cannot be captured by worklets. 

Example:
Object.assign polyfill
```JS
Object.assign(a, b);
```
is changed to 
``` JS
var _extends = ...;
_extends(a, b);
```
by  @babel/plugin-transform-object-assign.

#### SOLUTION
There is something like ```manipulateOptions``` property that plugin can use to access babel configuration. Unfortunately, when a plugin is added within a preset plugin name is replaced by file descriptor name that says nothing. In order to find which plugin should be removed, we get plugin visitors from node_modules and then compare keys. In most cases (always) that step is enough. However, to be sure we additionally compare visitor string representation (but only if keys are the same).

## TESTING
Check [example](https://github.com/software-mansion/react-native-reanimated/pull/915/files)
Check also if Object.assign is visible in stylediff worklet.

## CLOSING
resolves: https://github.com/software-mansion/react-native-reanimated/issues/921
resolves: https://github.com/software-mansion/react-native-reanimated/issues/939

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -94,7 +94,7 @@ function processWorkletFunction(t, fun) {
   const outputs = new Set();
 
   // We use copy because some of the plugins don't update bindings and
-  // some even brake them!
+  // some even break them
   const astWorkletCopy = parse('\n(' + fun.toString() + '\n)');
 
   traverse(astWorkletCopy, {
--- a/plugin.js
+++ b/plugin.js
@@ -354,6 +354,8 @@ module.exports = function ({ types: t }) {
         },
       },
     },
+    // In this way we can modify babel options 
+    // https://github.com/babel/babel/blob/eea156b2cb8deecfcf82d52aa1b71ba4995c7d68/packages/babel-core/src/transformation/normalize-opts.js#L64
     manipulateOptions(opts, parserOpts) {
       const plugins = opts.plugins;
       removePluginsFromBlacklist(plugins);
--- a/plugin.js
+++ b/plugin.js
@@ -93,7 +93,8 @@ function processWorkletFunction(t, fun) {
   const closure = new Map();
   const outputs = new Set();
 
-  console.log(fun.toString());
+  // We use copy because some of the plugins don't update bindings and
+  // some even brake them!
   const astWorkletCopy = parse('\n(' + fun.toString() + '\n)');
 
   traverse(astWorkletCopy, {
@@ -129,7 +130,6 @@ function processWorkletFunction(t, fun) {
         currentScope = currentScope.parent;
       }
       closure.set(name, path.node);
-      console.log("add", name);
     },
   });
 
@@ -225,7 +225,7 @@ function processWorkletFunction(t, fun) {
       ),
       t.returnStatement(privateFunctionId),
     ])
-  ); 
+  );
 
   const replacement = t.callExpression(newFun, []);
   // we check if function needs to be assigned to variable declaration.
@@ -287,23 +287,21 @@ function processWorklets(t, path, processor) {
 }
 
 const PLUGIN_BLACKLIST_NAMES = [
- // '@babel/plugin-transform-destructuring',
+  // '@babel/plugin-transform-destructuring', do not blacklist
+  'metro-react-native-babel-preset/node_modules/@babel/plugin-transform-object-assign',
 ];
 
-const PLUGIN_BLACKLIST = PLUGIN_BLACKLIST_NAMES.map(pluginName => {
-  const blacklistedPluginObject = require(pluginName);
-
-
-  if (!blacklistedPluginObject) {
-    console.log(`No ${pluginName}!`);
-    return null;
-  } else {
-    const blacklistedPlugin = blacklistedPluginObject.default(
-      {assertVersion: (x) => true}
-    );
+const PLUGIN_BLACKLIST = PLUGIN_BLACKLIST_NAMES.map((pluginName) => {
+  try {
+    const blacklistedPluginObject = require(pluginName);
+    const blacklistedPlugin = blacklistedPluginObject.default({
+      assertVersion: (x) => true,
+    });
 
     visitors.explode(blacklistedPlugin.visitor);
-    return blacklistedPlugin;  
+    return blacklistedPlugin;
+  } catch (e) {
+    console.log(`No ${pluginName}!`);
   }
 });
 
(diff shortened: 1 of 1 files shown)


wcandillon:
Thank you @Szymon20000 🙌🏻 Should we merge https://github.com/software-mansion/react-native-reanimated/pull/915 as a test or close it maybe?

Szymon20000:
It's a good test and I think we should have it but I'm not sure if we want it as an example that people should see. It should be more like a test screen. I think we need another directory with test screens. @karol-bisztyga  What do you think?

plugin.js
```diff
@@ -281,24 +286,64 @@ function processWorklets(t, path, processor) {
...
+});
+
+function removePluginsFromBlacklist(plugins) {
+  PLUGIN_BLACKLIST.forEach((blacklistedPlugin) => {
+    if (!blacklistedPlugin) {
+      return;
+    }
+
+    let toRemove = [];
+    for (let i = 0; i < plugins.length; i++) {
```
karol-bisztyga, line 315: I think it could be done using something like `Array.filter`

plugin.js
```diff
@@ -281,24 +286,64 @@ function processWorklets(t, path, processor) {
...
+      if (
+        JSON.stringify(Object.keys(plugins[i].visitor)) !=
+        JSON.stringify(Object.keys(blacklistedPlugin.visitor))
+      ) {
+        continue;
+      }
+      let areEqual = true;
```
karol-bisztyga, line 322: Can be determined using `Array.some`

karol-bisztyga:
> It's a good test and I think we should have it but I'm not sure if we want it as an example that people should see. It should be more like a test screen. I think we need another directory with test screens. @karol-bisztyga What do you think?

Yes this is definitely a good idea, I was thinking about it too!

wcandillon:
@Szymon20000 alternatively, I can update the example as a nice example people could see. It just wanted to show the minimal reproducible example. Let me know. Thank you for much for looking into it ♥️

plugin.js
```diff
@@ -90,14 +93,11 @@ function processWorkletFunction(t, fun) {
...
+  // some even brake them!
+  const astWorkletCopy = parse('\n(' + fun.toString() + '\n)');
 
-  fun.traverse({
-    DirectiveLiteral(path) {
-      if (path.node.value === 'worklet' && path.getFunctionParent() === fun) {
-        path.parentPath.remove();
-      }
-    },
+  traverse(astWorkletCopy, {
```
kmagiera, line 100: Wouldn't it be safer to extract "outputs" on the parsed copy instead of keeping it in `fun.traverse` block? Then we can keep `fun.traverse` only to do `'worklet'` directive cleanup

plugin.js
```diff
@@ -281,24 +286,64 @@ function processWorklets(t, path, processor) {
   }
 }
 
+const PLUGIN_BLACKLIST_NAMES = [
+  // '@babel/plugin-transform-destructuring', do not blacklist
```
kmagiera, line 290: why is this here. if its not on a list, there is no point of adding it in a comment as if it "almost" was on a list

plugin.js
```diff
@@ -281,24 +286,64 @@ function processWorklets(t, path, processor) {
...
+  'metro-react-native-babel-preset/node_modules/@babel/plugin-transform-object-assign',
+];
+
+const PLUGIN_BLACKLIST = PLUGIN_BLACKLIST_NAMES.map((pluginName) => {
+  try {
+    const blacklistedPluginObject = require(pluginName);
+    const blacklistedPlugin = blacklistedPluginObject.default({
```
kmagiera, line 297: this seem to be very obscure and rely on undocumented babel features. Can we add some comments on what it does as when the babel api changes we won't be able to track down what this piece of code should do

plugin.js
```diff
@@ -281,24 +286,64 @@ function processWorklets(t, path, processor) {
...
+      assertVersion: (x) => true,
+    });
+
+    visitors.explode(blacklistedPlugin.visitor);
+    return blacklistedPlugin;
+  } catch (e) {
+    console.log(`No ${pluginName}!`);
```
kmagiera, line 304: please remove console.log statements. If this is an error let's throw here instead? Or use console.warn or anything that'd suggest you are not accidently commiting some debug/logging code

plugin.js
```diff
@@ -281,24 +286,64 @@ function processWorklets(t, path, processor) {
...
+function removePluginsFromBlacklist(plugins) {
+  PLUGIN_BLACKLIST.forEach((blacklistedPlugin) => {
+    if (!blacklistedPlugin) {
+      return;
+    }
+
+    let toRemove = [];
+    for (let i = 0; i < plugins.length; i++) {
+      if (
+        JSON.stringify(Object.keys(plugins[i].visitor)) !=
```
kmagiera, line 317: shouldn't those objects just be the same? we load plugins using require and I think there should always be one source where plugins are loaded, no?

Szymon20000:
> Nice to see a more resilient solution to typescript and related issues we were having. Left a few comments inline

I think that this is a proper solution. The previous one was wrong. Our plugin has to process worklets after all other plugins.  People can use their own plugins and we want to let them do that.

plugin.js
Szymon20000: Not true. Plugins can be loaded by several presets. Moreover, "require" returns a function, not an object.

plugin.js
```diff
@@ -90,14 +93,11 @@ function processWorkletFunction(t, fun) {
   const closure = new Map();
   const outputs = new Set();
 
-  const funScope = fun.scope;
+  // We use copy because some of the plugins don't update bindings and
+  // some even brake them!
```
kmagiera, line 97:
```suggestion
  // some even break them
```

plugin.js
```diff
@@ -308,5 +354,9 @@ module.exports = function ({ types: t }) {
         },
       },
     },
+    manipulateOptions(opts, parserOpts) {
```
kmagiera, line 357: Can we document here what this thing does? Babel documentsation is very hard to browse and I don't understand and which point the options are going to be changed. Does the chaneg apply only to the AST branches that we traverse? Or maybe it is applied on a single-file level? Or maybe it is applied to all the files? Why we don't restore the previous options?

plugin.js
kmagiera: I wish we'd used soem different approach than JSON serialization. There is nothing technically wrong with it but it feels like the plugin object should have some field that we can extract its name from? If we want to use this as a last resort there should be a proper comment explaining that plugin objects are created by babel internals and they don't carry any identifier etc...

plugin.js
Szymon20000: It impacts all files. We don't have to restore previous options because Hermes and JSC support Object.assign.

wcandillon:
Congratulations 🙌🏻

PR #927, opened 2020-06-24
[withTiming] Check if previousAnimation has been started

Szymon20000:
WHY:
When we call withTiming twice in a row then nothing will happen because the first animation won't be started and as a result, we will copy undefined values from it.


Szymon20000:
WHY:
When we call withTiming twice in a row then nothing will happen because the first animation won't be started and as a result, we will copy undefined values from it.

Issue #920, opened 2020-06-22
V2: re-use useAnimatedStyle in multiple views


xignor:
When using animated style on multiple style props, it only works on the last one. Would be great to be able to reuse it to for generic style properties:

In this example only the blue box gets the opacity style updated

```
...
  const opacityAnimatedStyle = useAnimatedStyle(() => {
    return {
      opacity: 1 // have something change this value between 0<->1 such from a pan handler translation
    };
  });
...
  return <Animated.View>
      <Animated.View style={[{height: 20, width: 20, backgroundColor: 'red'}, opacityAnimatedStyle]} />
     <Animated.View style={[{height: 20, width: 20, backgroundColor: 'blue'}, opacityAnimatedStyle]} />
  </Animated.View>;
```

karol-bisztyga:
Hey, thanks for reporting! We are aware of the problem although we think it does not appear that often and also it can be done using multiple `useAnimatedStyle` calls. We are going to take care of this in the future.
Cheers

a-eid:
@karol-bisztyga which more performant sharing the style down with children or calling `useAnimatedStyle` hook multiple times ?

karol-bisztyga:
> @karol-bisztyga which more performant sharing the style down with children or calling `useAnimatedStyle` hook multiple times ?

Sharing the style with multiple children will not work in the first place. There is no difference if you do this in one component or in the decomposed way - the problem persists anyway.

karol-bisztyga:
This feature has been added to the TODO list, thank you for contributing, closing this for now 🙏

robrechtme:
Wow I just spent way too long debugging this 😄  Could this be mentioned in the docs maybe? Like a little note or a warning?

M2Changezi:
@jakub-gonet  @karol-bisztyga  how to call  multiple useAnimatedStyle ?

jakub-gonet:
@M2Changezi Not sure if I follow. You can use multiple UAS (`useAnimatedStyle`) hooks but you can't reuse one UAS, applying it to multiple views.

VityaSchel:
How to call `useAnimatedStyle` with same argument as function? It throws `TypeError: Object.values requires that input parameter not be null or undefined`

PR #931, opened 2020-06-26
@karol/uas multiple views

karol-bisztyga:
This was created as a response to this feature request
https://github.com/software-mansion/react-native-reanimated/issues/920 

It introduces the possibility of using one outcome of `useAnimatedStyle` for a multiple views.

### DIFF
--- a/Example/src/App.js
+++ b/Example/src/App.js
@@ -23,10 +23,13 @@ import ScrollableViewExample from './ScrollableViewExample';
 import AnimatedTabBarExample from './AnimatedTabBarExample';
 import LightboxExample from './LightboxExample';
 import LiquidSwipe from './LiquidSwipe';
+import FeaturesTestsScreen from './FeaturesTestsScreen'
+// Features tests
+import AnimatedStyleMultipleViews from './FeaturesTests/AnimatedStyleMultipleViews';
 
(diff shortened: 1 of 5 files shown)


Szymon20000:
I think that this approach won't work property. That's because updating a viewTags array will not change frozenArray on UI side. It won't update on rerender.

karol-bisztyga:
> I think that this approach won't work property. That's because updating a viewTags array will not change frozenArray on UI side. It won't update on rerender. However, I think it can be done with mutable string.

The render itself is not really a problem - it works ok. But excluding views from the component and including them again results in not updating styles with current values(this is because new view tag ids are assigned but outdated are used for updating).

Szymon20000:
> > I think that this approach won't work property. That's because updating a viewTags array will not change frozenArray on UI side. It won't update on rerender. However, I think it can be done with a mutable string.
> 
> The rerender itself is not really a problem - it works ok. But excluding views from the component and including them again results in not updating styles with current values(this is because new view tag ids are assigned but outdated are used for updating).

That's what I meant. We won't be able to update FrozenArray on the second render and next ones.

kmagiera:
As per our discussion on slack I understand that we should close this one. Feel free to reopen if that wasnt the case

PR #934, opened 2020-06-30
Use forward ref in createAnimatedComponent

### LINKED ISSUE #452: ScrollTo does not work on reanimated ScrollView

### LINKED ISSUE #701: Animated.ScrollView ref RN[0.62]
## Animated.ScrollView ref RN[0.62]

0.62 for React Native brings deprecation related to getting refs from animated scroll nodes:

> Refs on an Animated component are now the internal component. The getNode call has been deprecated. (66e72bb4e0 by @yungsters)

So now instead of using 
```
ref.current._component.scrollTo(...)
```
We should use
```
ref.current.scrollTo(...)
```

It would be nice to have the same behavior while using Animated scrolls from reanimated.

jkadamczyk:
## Description

Fixes #701, fixes #452

This issue gave me motivation #701 
React Native in 0.62 changed the way to perform operations on Animated components like ScrollView.

## Changes

I tried to take a look at
https://github.com/facebook/react-native/commit/66e72bb4e00aafbcb9f450ed5db261d98f99f82a
And build something similar.
I managed to reuse some of Facebook's code.

### DIFF
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -9,6 +9,8 @@ import WorkletEventHandler from './reanimated2/WorkletEventHandler';
 
 import invariant from 'fbjs/lib/invariant';
 
+const setAndForwardRef = require('react-native/Libraries/Utilities/setAndForwardRef');
+
 const NODE_MAPPING = new Map();
 
 function listener(data) {
@@ -54,10 +56,6 @@ export default function createAnimatedComponent(Component) {
       this._detachPropUpdater();
       this._propsAnimated && this._propsAnimated.__detach();
       this._detachNativeEvents();
-    }
-
-    setNativeProps(props) {
-      this._component.setNativeProps(props);
     }
 
     componentDidMount() {
@@ -223,11 +221,33 @@ export default function createAnimatedComponent(Component) {
       this._propsAnimated && this._propsAnimated.setNativeView(this._component);
     }
 
-    _setComponentRef = c => {
-      if (c !== this._component) {
-        this._component = c;
-      }
-    };
+    // _setComponentRef = c => {
+    //   if (c !== this._component) {
+    //     this._component = c;
+    //   }
+    // };
+
+    _setComponentRef = setAndForwardRef({
+      getForwardedRef: () => this.props.forwardedRef,
+      setLocalRef: ref => {
+        if (ref !== this._component) {
+          this._component = ref;
+        }
+
+        // TODO: Delete this after React Native also deletes this deprecation helper.
+        if (ref != null && ref.getNode == null) {
+          ref.getNode = () => {
+            console.warn(
+              '%s: Calling `getNode()` on the ref of an Animated component ' +
+                'is no longer necessary. You can now directly use the ref ' +
+                'instead. This method will be removed in a future release.',
+              ref.constructor.name ?? '<<anonymous>>'
+            );
+            return ref;
+          };
+        }
+      },
+    });
 
     _filterNonAnimatedStyle(inputStyle) {
       const style = {};
@@ -292,15 +312,18 @@ export default function createAnimatedComponent(Component) {
         <Component {...props} ref={this._setComponentRef} {...platformProps} />
       );
     }
-
-    // A third party library can use getNode()
-    // to get the node reference of the decorated component
-    getNode() {
-      return this._component;
-    }
   }
 
-  AnimatedComponent.displayName = `AnimatedComponent(${Component.displayName || Component.name || 'Component'})`
+  AnimatedComponent.displayName = `AnimatedComponent(${Component.displayName ||
+    Component.name ||
+    'Component'})`;
 
-  return AnimatedComponent;
+  return React.forwardRef(function AnimatedComponentWrapper(props, ref) {
+    return (
+      <AnimatedComponent
+        {...props}
+        {...(ref == null ? null : { forwardedRef: ref })}
+      />
+    );
+  });
 }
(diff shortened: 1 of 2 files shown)


src/createAnimatedComponent.js
```diff
@@ -292,15 +306,18 @@ export default function createAnimatedComponent(Component) {
...
-  return AnimatedComponent;
+  AnimatedComponent.displayName = `AnimatedComponent(${Component.displayName ||
+    Component.name ||
+    'Component'})`;
+
+  return React.forwardRef(function AnimatedComponentWrapper(props, ref) {
+    return (
+      <AnimatedComponent
+        {...props}
+        {...(ref == null ? null : { forwardedRef: ref })}
```
jakub-gonet, line 319: If refs are only `null | T` then I think we should use `===` here

jkadamczyk: Are they enforced to be null | T in pure JS without type checking? If not I'd leave it as is, cause I found that it was this way in Facebook's RN repo. Also type coercion table tells me that it will resolve to true only in case undefined is passed. I think it might be safe. On the other hand I'm thinking if we shouldn't be more explicit. Relying on JS type coercion in equality check is something that a lot of people may interpret wrong in the future. What do you think?

jakub-gonet: Yeah, it'll catch `undefined | null` there, same goes with `L#232`

Doesn't hurt to leave it as it is, so I'm cool with the current solution

jkadamczyk:
@jakub-gonet here's a minimal repro for a scrollview with `scrollToEnd` without any `getNode()`

```
import React, { useEffect, useRef } from 'react';
import { ScrollView as BaseScrollView, Text } from 'react-native';
import createAnimatedComponent from '../../src/createAnimatedComponent';

const ScrollView = createAnimatedComponent(BaseScrollView);

export default function ScrollViewTest() {
  const svRef = useRef();

  useEffect(() => {
    setTimeout(() => {
      svRef.current.scrollToEnd({ animated: true });
    }, 1000);
  }, []);

  return (
    <ScrollView ref={svRef}>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
      <Text>Test</Text>
    </ScrollView>
  );
}
```

src/createAnimatedComponent.js
```diff
@@ -223,11 +221,27 @@ export default function createAnimatedComponent(Component) {
...
+
+        // TODO: Delete this after React Native also deletes this deprecation helper.
+        if (ref != null && ref.getNode == null) {
+          ref.getNode = () => {
+            console.warn(
+              '%s: Calling `getNode()` on the ref of an Animated component ' +
+                'is no longer necessary. You can now directly use the ref ' +
+                'instead. This method will be removed in a future release.',
+              ref.constructor.name ?? '<<anonymous>>'
+            );
```
jakub-gonet, line 239:
```suggestion
           console.warn(
              '%s: Calling %s on the ref of an Animated component ' +
                'is no longer necessary. You can now directly use the ref ' +
                'instead. This method will be removed in a future release.',
              ref.constructor.name ?? '<<anonymous>>',
              'getNode()'
            );
```

This fixes yellow box appearance

jakub-gonet:
Fixes #701, fixes #452

PR #936, opened 2020-07-01
Provide synchronous methods 'measure' and 'scrollTo'.

Szymon20000:
## Testing | Example
<details>
<summary>code</summary>

``` JS
//export default createApp(ExampleApp);

import Animated, {
  useSharedValue,
  useAnimatedStyle,
  runOnUI,
  measure,
  scrollTo,
  getTag,

export default function screen() {
  const boxRef = useRef(null);
  const scrollView = useRef(null);
  const selected = useRef(0);

  function handleClick() {
   
    const handle = getTag(boxRef.current);
    runOnUI(() => {
      'worklet';
      console.log(measure(handle));
    })();
  }

  function nextOne() {
    selected.current++;
    selected.current %= 4;
    return selected.current;
  }

  function handleClick2() {
    const handle = getTag(scrollView.current);
    const where = nextOne() * 20;
    runOnUI(() => {
      'worklet';
      scrollTo(handle, 0, where, false);
    })();
  }

  function handleClick3() {
    const handle = getTag(scrollView.current);
    const where = nextOne() * 20;
    runOnUI(() => {
      'worklet';
      scrollTo(handle, 0, where, true);
    })();
  }

  return (
    <View>
      <TouchableHighlight style={stylez.box} ref={boxRef} onPress={handleClick}>
        <Text style={{ color: 'white' }}>click me to measure</Text>
      </TouchableHighlight>
      <View style={stylez.wrapper}>
        <ScrollView style={stylez.scrollView} ref={scrollView}>
          <Text style={stylez.element}>aa</Text>
          <Text style={stylez.element}>bb</Text>
          <Text style={stylez.element}>cc</Text>
          <Text style={stylez.element}>dd</Text>
        </ScrollView>
      </View>
      <TouchableHighlight style={stylez.button} onPress={handleClick2}>
        <Text style={{ textAlign: 'center' }}>moveToThe nextOne</Text>
      </TouchableHighlight>
      <TouchableHighlight style={stylez.button} onPress={handleClick3}>
        <Text style={{ textAlign: 'center' }}>moveToThe nextOne Animated</Text>
      </TouchableHighlight>
    </View>
  );
}

const stylez = StyleSheet.create({
  box: {
    margin: 100,
    width: 100,
    height: 200,
    backgroundColor: 'black',
  },
  scrollView: {
    borderColor: 'black',
  },
  element: {
    height: 20,
    backgroundColor: 'red',
    borderColor: 'black',
    borderWidth: 1,
  },
  wrapper: {
    height: 20,
  },
  button: {
    height: 30,
    margin: 10,
    backgroundColor: 'grey',
    borderRadius: 10,
  },
});

```

</details>

## HOW
I Implemented internal methods `_scrollTo` and `_measure` and wrapped them to make it more user/linter friendly.
(`_measure` method returns {x: -1234567} if a view or root view couldn't be found.)

### iOS
Copied from [here](https://github.com/facebook/react-native/blob/ab1f54c61cb4ac330d3cfd90ec1a52980eeb9e2e/React/Modules/RCTUIManager.m#L1300)

### Android
On Android it was much more complicated cause every method that can be used to measure a view is private. To make it work I copied code from [here](https://github.com/facebook/react-native/blob/5c48c94f8c0441bc78a007f0ea0c5b2763ff6875/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java#L631).

There is no magic in scrollTo implementation, but just in case see the following links:
[iOS](https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Views/ScrollView/RCTScrollView.m#L486)
[Android-1](https://github.com/facebook/react-native/blob/v0.62.2/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollViewManager.java#L194)
[Android-2](https://github.com/facebook/react-native/tree/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/ReactAndroid/src/main/java/com/facebook/react/views/scroll)

### POSSIBILITY TO CALL ANY VIEW METHOD
I think that not all view methods can be called in a simple way. Some just have a too complicated signature and some are asynchronous.  However, we definitely can provide a way for users to call a view method if it's simple enough. It's not included in the pr though. 
### SOLVES
resolves: https://github.com/software-mansion/react-native-reanimated/issues/949

### New `useAnimatedRef` hook
A new hook `useAnimatedRef` has been created to provide a functionality of animated reference which acts as a normal reference on the javascript thread but at the same time holds a view tag of the component which it is referring to through a shared value and can be accessed from the UI.


Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -51,18 +51,27 @@ NativeReanimatedModule::NativeReanimatedModule(std::shared_ptr<CallInvoker> jsIn
                                                std::shared_ptr<Scheduler> scheduler,
                                                std::unique_ptr<jsi::Runtime> rt,
                                                std::function<void(std::function<void(double)>)> requestRender,
-                                               std::function<void(jsi::Runtime&, int, const jsi::Object&)> propUpdater):
+                                               std::function<void(jsi::Runtime&, int, const jsi::Object&)> propUpdater,
+                                               ScrollToFunction scrollToFunction,
```
kmagiera, line 55: I think it is about the time we make some sort of interface that'd provide methods that have separate platform-dependent implementation. Was something I wanted to do from the startt. But as we are getting more methods here this starts to look a bit obscure

ios/native/RENativeMethods.h
```diff
@@ -0,0 +1,11 @@
+#import <Foundation/Foundation.h>
+#include <vector>
+#include <string>
+#import <React/RCTUIManager.h>
+
+NS_ASSUME_NONNULL_BEGIN
```
kmagiera, line 6: what?

ios/native/RENativeMethods.h
```diff
@@ -0,0 +1,11 @@
+#import <Foundation/Foundation.h>
+#include <vector>
```
kmagiera, line 2: maybe use import for consistency (given this is an objc header)

Szymon20000:
> Would be nice to have some example included (or even just a sample usage in the PR description) such that we can see how this can be used. When it comes to that I think it would be much more convenient if we supported passing view refs and not view tags.

There is a simple example.  I don't see what's wrong with viewTags. How can we do it with refs only?

ShaMan123:
#469

ShaMan123:
BTW I think there's a lot of good code I wrote for android you can re-use (i.e native `measure` function, instead of simply coping it).
If I can help implement an abstraction ping me.

Example/src/ScrollToExample.js
```diff
@@ -0,0 +1,153 @@
...
+        {indices.map((i) => {
+          return <Digit digit={getDigit(number, i)} key={i} />;
+        })}
+      </View>
+    </View>
+  );
+}
+
+function Digit({ digit }) {
+  const tag = useSharedValue(-1);
```
kmagiera, line 65: I was thinking of a slightly different API that'd allow us to avoid some of the boilerplate here:

```js
const ref = useAnimatedRef();
```

Then in view we could just do:
```js
<ScrollView ref={ref}>
```

And in the worklet we could use it this way:
```js
scrollTo(ref, 0, difit.value * 200, true);
```

This would be the easiest approach in my opinion and could be implemented in the follwoing way:
`useAnimatedRef` should return a function – refs can be functions that are called when the view ref is set. When that happens the function should update a shared values kept underneath. Methods like `scrollTo` or `measure` should expect `ref` to be an object and extract the reference to the shared value out of it.

terrysahaidak: The only problem with that approach - you may need this ref as a regular ref somewhere else too.

karol-bisztyga: Here is the sample solution
https://github.com/software-mansion/react-native-reanimated/pull/1040

karol-bisztyga: This works although it might be good to consider whether there would be any more cases of using this, also as @terrysahaidak mentioned the name could be misleading.

PR #945, opened 2020-07-03
got rid of initialUpdaterRun

karol-bisztyga:
todo


kmagiera:
What is the problem with that initial run? Can we document this somewhere?

karol-bisztyga:
this PR had been created only to show some idea(thus as a draft). Closed due to becoming outdated.

PR #951, opened 2020-07-07
@karol/error handler

karol-bisztyga:
## Description

Provides a mechanism of handling errors more gracefully and in some cases prevents crashes. 

## Changes

Modified Error handler code, added `try ... catch ...` clauses and injected error handler into `NativeReanimatedModule`. Also refactored calling exception throw on Android in java with `fbjni`. 

## Example

Sample code which is better handled after those changes:

`const sv = useSharedValue(3)`
 ` const style = useAnimatedStyle(() => {`
  `sv = 33; // here`
    `console.log('invalid ' + sv) // and here`
    `return {`
      `width: withTiming(someOtherSV.value, config),`
    `};`
  `});`

### DIFF
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -195,7 +195,7 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
           return jsi::Value(rt, *funPtr);
         }
 
-        auto clb = [=](
+        auto clb = [=, &module](
                    jsi::Runtime &rt,
                    const jsi::Value &thisValue,
                    const jsi::Value *args,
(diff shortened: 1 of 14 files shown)


Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -205,11 +205,18 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
...
+             if (thisValue.isObject()) {
+               res = funPtr->callWithThis(rt, thisValue.asObject(rt), args, count);
+             } else {
+               res = funPtr->call(rt, args, count);
+             }
+           } catch(std::exception &e) {
+             std::string str = e.what();
+             module->errorHandler->setError(str.c_str());
+             if (!module->errorHandler->raise()) {
+               throw;
```
karol-bisztyga, line 218: this is probably not going to be called ever

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -250,11 +257,18 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
...
+            try {
+              returnedValue = funPtr->call(rt,
                                              static_cast<const jsi::Value*>(args),
                                              (size_t)params.size());
             
+            } catch(std::exception &e) {
+              std::string str = e.what();
+              module->errorHandler->setError(str.c_str());
+              if (!module->errorHandler->raise()) {
+                throw;
```
karol-bisztyga, line 269: this is probably not going to be called ever

PR #958, opened 2020-07-08
fix object assign plugin lookup

karol-bisztyga:


Szymon20000:
I think we also need to add `@babel/plugin-transform-object-assign` to our root package.json. In order to be sure that the plugin exists in node_modules.

PR #959, opened 2020-07-08
Simplify proc type signature

wcandillon:
@jakub-gonet @terrysahaidak  thks for your support on this!


jakub-gonet:
Linked with #890

Issue #950, opened 2020-07-06
Build Android Error: "specified for property '$3' does not exist"


adrianwijaya0394:
## Description

I am using react-native-reanimated #908  Pull Request Head. But when I run react-native run-android, build error with Error: specified for property '$3' does not exist. Why this error occured?

### Screenshots
![build failed_70](https://user-images.githubusercontent.com/65936251/86877133-6668c080-c110-11ea-8fba-2c71538d24be.png)

## Steps To Reproduce

1. run : yarn add "software-mansion/react-native-reanimated#908/head"
2. build android with: npx react-native run-android

### Expected behavior

build android passed without error.

### Actual behavior

build android error, with Error: specified for property '$3' does not exist

## Package versions

- React: 16.11.0
- React Native: 0.62.2
- React Native Reanimated: react-native-reanimated@2.0.0-alpha-3#908

likern:
@adrianwijaya0394 Have you solved this problem?
I came across the same bug when I forked repository and tried to build app from master.

adrianwijaya0394:
no, I have not solved this problem.

terrysahaidak:
Seems like Common directory should be symlinked/copied into `android/src/main`.

/cc @kmagiera

ibrahim019:
> Seems like Common directory should be symlinked/copied into `Example/android/src/main`.
> 
> /cc @kmagiera

This works

Szymon20000:
@adrianwijaya0394 We have a script for building npm package. It's not possible right now to add Reanimated 2 to a project directly from GitHub.

PR #960, opened 2020-07-08
## Description

likern:

### DIFF
--- a/src/reanimated2/Colors.js
+++ b/src/reanimated2/Colors.js
@@ -386,6 +386,42 @@ function normalizeColor(color) {
   return null;
 }
 
+function red(color) {
+  'worklet';
+  return (color >> 24) & 255;
+}
+
+function green(color) {
+  'worklet';
+  return (color >> 16) & 255;
+}
+
+function blue(color) {
+  'worklet';
+  return (color >> 8) & 255;
+}
+
+function alpha(color) {
+  'worklet';
+  return (color & 255) / 255;
+}
+
+export function destructColor(color) {
+  'worklet';
+
+  const normalizedColor = normalizeColor(color);
+  if (normalizedColor === undefined || normalizedColor === null) {
+    return undefined;
+  }
+
+  const r = red(normalizedColor);
+  const g = green(normalizedColor);
+  const b = blue(normalizedColor);
+  const a = alpha(normalizedColor);
+
+  return [r, g, b, a];
+}
+
 export default function processColor(color) {
   'worklet';
   if (color === null || color === undefined || typeof color === 'number') {
--- a/src/reanimated2/interpolation.js
+++ b/src/reanimated2/interpolation.js
@@ -1,4 +1,5 @@
 import { Extrapolate } from '../derived/interpolate';
+import { destructColor } from './Colors';
 
 function internalInterpolate(x, l, r, ll, rr, type) {
   'worklet';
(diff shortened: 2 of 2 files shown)


likern:
Unfortunately, I couldn't test these changes using modified react-native-reanimated package because I've got **"Unsatisfied Link Error"** for **reanimated.so**. I've tested using javascript playground.

If there are maintainers with complete build setup - please, verify this pull-request is working. I think it should.

photo5204307569185762798

karol-bisztyga:
Just a tiny note: you do not have to add `[DO NOT MERGE]`, creating a draft pull request is a feature designed for such cases

wcandillon:
I have these functions working well I think at https://github.com/wcandillon/react-native-redash/pull/289/files#diff-1234f550d692aff6b8f2a1952daf79ec. Is this helping?

src/reanimated2/Colors.js
```diff
@@ -386,6 +386,42 @@ function normalizeColor(color) {
...
+
+  const normalizedColor = normalizeColor(color);
+  if (normalizedColor === undefined || normalizedColor === null) {
+    return undefined;
+  }
+
+  const r = red(normalizedColor);
+  const g = green(normalizedColor);
+  const b = blue(normalizedColor);
+  const a = alpha(normalizedColor);
```
karol-bisztyga, line 420: why are all those calculations wrapped into separate functions?

likern: > why are all those calculations wrapped into separate functions?

For better code reusability and ability to extract and use color channel later, when more code for working with colors will be added.

src/reanimated2/interpolation.js
```diff
@@ -51,3 +52,24 @@ export function interpolate(x, input, output, type) {
...
 }
+
+export function interpolateColor(x, input, output, type) {
+  'worklet';
+
+  // Extrapolate.CLAMP is the default interpolation type for colors
+  type = type || Extrapolate.CLAMP;
+
+  const startColor = destructColor.apply({}, output[0]);
+  const endColor = destructColor.appy({}, output[1]);
```
karol-bisztyga, line 63: typo, I guess expected word is `apply`

src/reanimated2/interpolation.js
```diff
@@ -51,3 +52,24 @@ export function interpolate(x, input, output, type) {
...
   return internalInterpolate.apply({}, [x].concat(narrowedInput).concat(type));
 }
+
+export function interpolateColor(x, input, output, type) {
+  'worklet';
+
+  // Extrapolate.CLAMP is the default interpolation type for colors
+  type = type || Extrapolate.CLAMP;
+
+  const startColor = destructColor.apply({}, output[0]);
```
karol-bisztyga, line 62: is this called properly? Shouldn't it be `const startColor = destructColor(output[0]);`? `destructColor` (if I understand it correctly) is a function more or less like `(number) => Array<number>` so why calling `apply on it`?

src/reanimated2/interpolation.js
```diff
@@ -51,3 +52,24 @@ export function interpolate(x, input, output, type) {
...
+  const endColor = destructColor.appy({}, output[1]);
+
+  if (startColor === undefined || endColor === undefined) {
+    return undefined;
+  }
+
+  const r = interpolate.apply({}, x, input, [startColor[0], endColor[0]], type);
```
karol-bisztyga, line 69: shouldn't those args be wrapped in an array like `const r = interpolate.apply({}, [x, input, [startColor[0], endColor[0]], type]);` ?

karol-bisztyga: here and for the 3 following calls

src/reanimated2/interpolation.js
likern: I wasn't able to test this patch (I hoped react-native-reanimated have some kind of CI tests to check pull requests).
Also I wasn't able to build react-native-reanimated npm package to test my patch manually https://github.com/software-mansion/react-native-reanimated/issues/1024 https://github.com/software-mansion/react-native-reanimated/issues/950
But I was able test code in isolation manually.

likern: Looks like with `appy` misspell it doesn't work. If I would able to build react-native-reanimated npm package from master branch I'll fix misspell and add tests.

src/reanimated2/interpolation.js
jakub-gonet: @likern, any update on this?

wcandillon:
Interpolate color works well with redash worklets: https://github.com/wcandillon/react-native-redash/blob/master/src/Colors.ts#L201 feel free to copy this implementation if needed.

Issue #397, opened 2019-09-24
Native crash when using an undefined Animated.Value in a debug node


yolpsoftware:
When setting the value of an Animated.Value object to `null` or `undefined` and then using it in a `debug` statement in an `Animated.Code` block, we get a native exception that crashes the app.

Reproducible demo:

```javascript

import React, {Fragment} from 'react';
import {
  SafeAreaView,
  Text,
  TouchableOpacity,
} from 'react-native';

import Animated from 'react-native-reanimated';
const {
  set,
  block,
  debug,
} = Animated;

const node = new Animated.Value(0);

const App = () => {
  return (
    <Fragment>
      <Animated.Code>
        {() => block([debug('node value: ', node)])}
      </Animated.Code>
      <SafeAreaView>
        <TouchableOpacity
          onPress={() => node.setValue(undefined)}    // setValue(null) will throw a different exception
          style={{padding: 20}}
        >
          <Text>Click Me</Text>
        </TouchableOpacity>
      </SafeAreaView>
    </Fragment>
  );
};

export default App;
```

This error can be quite difficult to find in a complex animation, since there is no indication about what the problem is. If it is intended that an error is thrown on an undefined `Value`, then a better error message would be very useful.

The exact error is on line 21 of `REAJSCallNode.m`:

    [error][tid:main][RCTUIManager.m:1133] Exception thrown while executing UI block: *** -[__NSArrayM setObject:atIndexedSubscript:]: object cannot be nil

In the case of setting the `Value` to `null`, we get a JS exception: `Unhandled JS Exception: Cannot read property '__isProxy' of null`.

Here are the versions used:

    "react": "16.8.6",
    "react-native": "0.60.5",
    "react-native-reanimated": "1.2.0"

Hirbod:
Ping

Jarred-Sumner:
More info from debugging this a little.

I'm seeing this too. It happens when:
- The component is unmounted
- The component is detached (e.g. `display: none`)
- Fast refresh loads the component (sometimes)

If you have any animated nodes inside a `<Modal>`, it breaks.

It happens right here:
![image](https://user-images.githubusercontent.com/709451/73701132-636b2f00-469d-11ea-8d43-2998543d0ed8.png)

In `-(id) evaluate` for most (maybe all) of the nodes, this error can happen.

[...]

PR #970, opened 2020-07-10
Forbid using null in Animated.Value

adamgrzybowski:
## Description

Reanimated doesn't have any use case for holding `null` value in the `Animated.Value` besides using `defined` node to set it to some initial state, so we're relying on `undefined` being treated as that "empty" value. 
In order to prevent some obscure errors when `null` is used we introduce explicit check for creating Value with `null` - it works when instantiating new Value and using imperative `value.setValue()`. 

We don't need to change the `set` node, because it expects `double | AnimatedValue` to be used as it's second argument.

### DIFF
--- a/src/core/InternalAnimatedValue.js
+++ b/src/core/InternalAnimatedValue.js
@@ -1,6 +1,7 @@
 import AnimatedNode from './AnimatedNode';
 import { val } from '../val';
 import ReanimatedModule from '../ReanimatedModule';
+import invariant from 'fbjs/lib/invariant';
 
 function sanitizeValue(value) {
   return value === null || value === undefined || typeof value === 'string'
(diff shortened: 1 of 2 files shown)


src/core/InternalAnimatedValue.js
```diff
@@ -32,6 +33,10 @@ export default class InternalAnimatedValue extends AnimatedNode {
   }
 
   constructor(value, constant = false) {
+    invariant(
+      value !== null,
+      'you are trying to set null on animated value'
```
jakub-gonet, line 38: I think something like: "Animated.Value cannot be set to the null" would be more clearer as an error message

jakub-gonet:
Rebased to `v1` branch

yolpsoftware:
This closes #397, right?

jakub-gonet:
Yeah, thanks for mentioning

PR #973, opened 2020-07-13
Changed the signature of RCTTurboModuleManager init

SudoPlz:
- On react-native 0.63 the signature of RCTTurboModuleManager alloc] initWithBridge
changed.
https://github.com/facebook/react-native/commit/459c54c4079e1ca9c2d12bae89a935e5a4c6444e

We need to keep those docs up to date


jakub-gonet:
Hi @SudoPlz, thanks for the PR.

Unfortunately, it seems that #969 and #975 already fixed that.

SudoPlz:
Cool thanks for taking care of that

PR #975, opened 2020-07-14
Update installation guide.

Szymon20000:
Update info about:
 - useTurboModule step on Android (We no longer use TurboModuleManager on Android)
 - The constructor signature of TurboModuleManager  has changed on iOS


kmagiera:
You should only make amend in the unversioned docs. Versioned ones are automatically copied when we make a release

PR #977, opened 2020-07-14
Add new animation helper "sequence"

kmagiera:
The primary goal of this PR is to add new animation helper called "sequence".

Sequence takes a number of animations and runs them one after another. This is a convenience method and could've previously be done by starting animations with callback and then starting next one in callback, but we see there are some nice usecases sequence fits in and we'd like to include it in the API.

This PR also makes a braking change to `loop` method. We are deprecating `loop` and adding `repeat` method instead. The difference between `loop` and `repeat` is that `repeat` does not "reverse" the animation by default. This behavior, however, can be adjusted with an option. It makes more sense not to "reverse" by default, as the "reverse" behavior can be easily reproduced using `repeat` and `sequence` (enough to make a sequence that is the animation and its reverse). The second argument for this change is that the automatic reverse does not work in most of the cases and can only be applied to plain animations that can be directly modified by us (we need to change animation settings to make it animate to a different target value). For compatibility we are keeping `loop` that defaults to call `repeat` with `reverse: true` but also warns that the method is deprecated. We also removed `loop` from the documentation and added docs for `repeat` and `sequence`.

One other change I made in reverse is that I removed the workaround for broken default value of function parameter (with a corresponding todo). This seem to no longer be an issue, not sure if the actual resolution of this problem is due to some change in the plugin or to refactor that allowed for animations to be started from main JS thread.

Last minor change I made to make examples work is to fix `withTiming` reuse scheme. Now we compare `toValue` before we decide whether we reuse current timing or start a new one. This approach had a problem when the animation was already done. In that case we want to start the new animation each time. To fix that, I reset `startTime` when the animation reaches the end.

In this PR I also fixed a problem with missing 'worklet' directive in some of the Easing method. This change also triggered prettier reformatting.

# Testing

I tested this change by running a number of loop examples. One sample I tested in particular is when we implement a wobble effect (animate rotation starting from 0, then lean left a bit and then do a couple of oscillations before coming back to 0).

```
const ANGLE = 10;
rotation.value = sequence(
  withTiming(-ANGLE, { duration: 50 }),
  loop(
    sequence(
      withTiming(ANGLE, { duration: 100 }),
      withTiming(-ANGLE, { duration: 100 })
    ),
    10
  ),
  withTiming(0, { duration: 50 })
);
```


src/reanimated2/animations.js
```diff
@@ -325,47 +327,94 @@ export function delay(delayMs, _nextAnimation) {
...
+      typeof a === 'function' ? a() : a
+    );
+    const firstAnimation = animations[0];
+
+    function sequence(animation, now) {
+      const currentAnim = animations[animation.animationIndex];
+      const finished = currentAnim.animation(currentAnim, now);
+      animation.current = currentAnim.current;
+      if (finished) {
+        const finalValue = currentAnim.current;
```
karol-bisztyga, line 344: unused variable could be used in `nextAnim.start`

Issue #979, opened 2020-07-14
V2: createReanimatedModule should be running on UI thread but isn't


SudoPlz:
## Description
I'm integrating react-native-reanimated v2, and it's currently throwing runtime errors, because [this part of the code ](https://github.com/software-mansion/react-native-reanimated/blob/3aca172d3f69e0a064f869fb62c4ffc1bc09da43/ios/native/NativeProxy.mm#L82)requires being ran on the UI thread and instead it's being ran on the Javascript thread.

## Steps To Reproduce

1. Enable the Main Thread checker <img src="https://p-5JFjjq.b1.n0.cdn.getcloudapp.com/items/eDu1plry/Image%202020-07-14%20at%207.16.00%20PM.png?v=5f24bec8b831f2726098b6c9caa1afc8" />

2. Try to run the project with Turbo Modules enabled

### Expected behavior
No runtime errors
### Actual behavior
Runtime error
<img src="https://p-5JFjjq.b1.n0.cdn.getcloudapp.com/items/5zuXknv0/Image%202020-07-14%20at%207.17.33%20PM.png?v=8544298a8e0fb54cb74abc386faa645a" />

## Package versions

- React:
"react": "16.13.1",
- React Native:
"react-native": "0.63.0",
- React Native Reanimated:
"react-native-reanimated": "^2.0.0-alpha.3",

SudoPlz:
Here's a patch that fixes it:

```
diff --git a/node_modules/react-native-reanimated/ios/native/NativeProxy.mm b/node_modules/react-native-reanimated/ios/native/NativeProxy.mm
index e5e83f4..40090c8 100644
--- a/node_modules/react-native-reanimated/ios/native/NativeProxy.mm
+++ b/node_modules/react-native-reanimated/ios/native/NativeProxy.mm
@@ -77,11 +77,23 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
 }
 
 std::shared_ptr<NativeReanimatedModule> createReanimatedModule(std::shared_ptr<CallInvoker> jsInvoker) {
+
+
+
+  __block RCTBridge *bridge;
+  void (^getBridge)(void) = ^(void) {
+//      NSLog(@"on main queue!");
+      if ([[UIApplication sharedApplication].delegate respondsToSelector:@selector(bridge)]) {
[...]

PR #981, opened 2020-07-15
Simplify installation process on iOS.

Szymon20000:
## How
Initialize global bridge variable in AppDelegate.jsExecutorFactoryForBridge. It's cleared in REAModule.invalidate. 

## Testing
Checked if reload works on iOS.

## Closes
resolves: https://github.com/software-mansion/react-native-reanimated/issues/979


SudoPlz:
Wow such a fast response. Thank you!
Any idea when alpha 4 will go live?

PR #985, opened 2020-07-16
Add web/js only support

terrysahaidak:

### DIFF
--- a/Example/ios/Podfile.lock
+++ b/Example/ios/Podfile.lock
@@ -268,7 +268,7 @@ PODS:
     - React
   - RNGestureHandler (1.5.2):
     - React
-  - RNReanimated (2.0.0-alpha.3):
+  - RNReanimated (2.0.0-alpha.4):
     - DoubleConversion
     - FBLazyVector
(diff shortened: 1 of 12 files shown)


gut4:
What is status of this? Any ETA?

terrysahaidak:
I'm currently working on this in my free te so no ETA. I would appreciate any testing and bug reports here.

I can guide how to debug and turn it on.

gut4:
Guide would be nice

terrysahaidak:
> Guide would be nice

In the file called NativeReanimated.js there is native flag. It checks for availability of JSI provided native Reanimated module. You can just replace it with "false" and now you're running js version of Reanimated2.

The whole logic is executed on the same thread as everything else does.

The only big missing thing in the current implementation is topological sorting of mappers (you can refer to c++ implementation of the same algorithm).

kesha-antonov:
Will it work with react-native-web?

terrysahaidak:
Yeah, it should.

terrysahaidak:
Closing it because the work on it is currently on different branch.

PR #1000, opened 2020-07-20
Support Windows and macOS platforms

kaiguo:
## Description

Adding Windows and macOS supports.

Close https://github.com/microsoft/react-native-windows/issues/4151.
Unblock https://github.com/microsoft/react-native-windows/issues/3884.

## Changes

Following the web path to add Windows and macOS stubs. This helps us unblock the usages of react-navigation.


src/ReanimatedModule.windows.js
```diff
@@ -0,0 +1,3 @@
+import ReanimatedModuleCompat from './ReanimatedModuleCompat';
+
+export default ReanimatedModuleCompat;
```
hatched-tom, line 3: small thing but missing new line at eof unlike macos equiv

jakub-gonet:
Have you tested those changes on examples from the Example app?

kaiguo:
@jakub-gonet Many of the examples use gesture, Windows and macOS don't have react-native-gesture-handler support yet. Also the example app is targeting react-native 0.63, react-native-windows is still on 0.62 so I wasn't able to just add a Windows example app.

I did test some of the pages that don't require gesture in my separate test app, like imperative, progressBar and rotations, they worked fine. We are trying to unblock react-navigation so I tested it in drawer navigation as well, here's a gif I captured earlier : )

jakub-gonet:
Thanks for testing it.

Even if it's working partially it's better than nothing I guess

jakub-gonet:
BTW, congrats on such nice PR number 😄

kaiguo:
@jakub-gonet one more question, can we backport this to 1.x? react-navigation is still on reanimated1, we'd have to wait them to migrate to 2.x to add windows support otherwise.

jakub-gonet:
Sure, could you prepare the same PR against v1 branch? I want to release 1.10.0 tomorrow.

kaiguo:
> Sure, could you prepare the same PR against v1 branch? I want to release 1.10.0 tomorrow.

Yep, will do it later today. Thanks!

PR #1006, opened 2020-07-22
@karol/fix error handling

karol-bisztyga:
Improve error handling by adding some cases in `Hooks.js` and `animations.js` regarding lack of `.value` field when passing shared value objects.

As a little refactor changed `const char*` to `std::string` in ErrorHandler due to obeying coding best practices.

Also moved catching errors higher in the calling hierarchy, which is not a problem since when there is no error set in error handler it just rethrows.

### DIFF
--- a/src/reanimated2/animations.js
+++ b/src/reanimated2/animations.js
@@ -3,6 +3,17 @@ import { Easing } from './Easing';
 
 let IN_STYLE_UPDATER = false;
 
+const assertNumber = (value, callerName) => {
+  const valueType = typeof value
+  if (valueType !== 'number') {
+    let error = `invalid type of toValue passed to ${ callerName }, expected \`number\`, got \`${ valueType }\``;
+    if (valueType === 'object') {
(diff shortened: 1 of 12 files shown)


src/reanimated2/animations.js
```diff
@@ -33,7 +33,16 @@ export function cancelAnimation(value) {
...
-  toValue = (typeof toValue === 'object' && toValue.value) ? toValue.value : toValue
+  // check toValue
+  const valueType = typeof toValue
+  if (valueType !== 'number') {
+    let error = `invalid type of toValue passed to withTiming, expected \`number\`, got \`${ valueType }\``;
+    if (valueType === 'object') {
+      error += ', maybe you forgot to add `.value`?'
+    }
+    throw new Error(error);
+  }
```
jakub-gonet, line 44: Maybe it's worth extracting into some function?

karol-bisztyga: point well taken, fixed: 58f0e92

PR #1016, opened 2020-07-23
chore: added few good config

hosseinmd:
Listed below done.
 - commitlint [config-conventional](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional)
 - react-nativ-community eslint config
 - update eslint
 - update prettier


jakub-gonet:
Hey, thanks for PR.

I'll talk with other maintainers if we want to enforce conventional commit names. 

If yes I think adding some CLI helpers like [Commitizen](https://github.com/commitizen/cz-cli) or [Standard commit](https://github.com/stasson/standard-commit) may be a good move here.

hosseinmd:
This config is also used in react-navigation too.

Issue #949, opened 2020-07-05
How to couple two ScrollViews together?


yolpsoftware:
## Description

I want to couple two (or more) ScrollViews together, so that the scrolling on one directly influences the scroll offset of the other, and vice versa. So if you scroll one of them, the other one scrolls as well (one of them faster, and not coupled in a linear way). A scroll bar would be an example of such a behavior. If you scroll the content, the scroll bar's knob would be scrolled as well. If you move the scroll bar, the content gets scrolled too.

Can it be done in React Native Reanimated?

I thought about having a main and secondary ScrollView (once called "master/slave pattern"). A `Reanimated.Value` is passed as the `contentOffset` prop to the secondary ScrollView, and the `onScroll` event of the main ScrollView would influence this value. However, when the user scrolls the secondary ScrollView, updating the main ScrollView would lead to an infinite loop, wouldn't it?

I also thought about having just one ScrollView which invisibly overlaps the secondary ScrollViews, so only one ScrollView gets any touch events. This would work, but then I need to somehow accelerate/decelerate the scroll speed as soon as the user scrolls in the region of the secondary ScrollViews.

Is there an obvious solution to this? I'm interested in all kinds of answers, does not matter if Reanimated v1 or v2.

terrysahaidak:
so you need a real scroll view? or can it be implemented by your own?

if so, you can use decay animation to create your own scroll view.

yolpsoftware:
@terrysahaidak A real scroll view would be very nice, since it would have the same properties (deceleration rate, bouncyness etc.) as the native one. Moreover, I want to use this in a `FlatList`. So if I'm implementing my own `ScrollView`, this would mean to also implement my own `FlatList`.

Moreover, if it is possible with Reanimated, it might be a good example to learn from.

terrysahaidak:
there is work in progress for sync scrollTo calls from worklets in Reanimated2. this might help you.

could you describe your case more? why do you need flatlist? do you plan for few hundreds items to be rendered?

yolpsoftware:
Yes, in one of the lists, there are hundreds of items.

My use case involves a big tree of data, like a big mind map. There are thousands of leaves, but having them in one big list is not helpful, so we use the tree to kind of index them. We don't want any nested expanders (like in a mind map). We prefer to show each tree level as a scrollable list. The first level might fit on an iPad screen, but on a phone it must be scrolled. The second level is scrollable on every device.

[...]

PR #1019, opened 2020-07-24
Use animated reaction

karol-bisztyga:
## Description

Added a new hook - `useAnimatedReaction` which takes 2 arguments(both are worklets):
1. `prepare` - prepares data for the second argument. Value returned from this function is the only one argument passed to the following one
2. `react` - takes one argument which is a result from the previous function's invocation and just performs some specified operations

## Motivation

This hook has been added to allow performing some actions on some shared values' change. The key idea is all of the shared values included in the first worklet are the `inputs` set. Every time any of those change both worklets are triggered in the order specified above. Also the second worklet may modify any shared values excluding those used in the first worklet.

## Usage

```
useAnimatedReaction(
    () => {
        'worklet';
        return (x.value + y.value) / 2; // `data` below
    },
    (data) => {
        'worklet';
        if (sv.value < 100) {
            sv.value = data;
        }
    }
);
```

## Changes

`Hooks.js` - added `useAnimatedReaction` hook
`AnimatedReactionTest.js` - added simple test showing an example of using the new hook
`plugin.js` - added automatic workletization for the new hook

### DIFF
--- a/Example/test/AnimatedReactionTest.js
+++ b/Example/test/AnimatedReactionTest.js
@@ -1,0 +1,76 @@
+import React from 'react';
+import { Button, View, StyleSheet, Text } from 'react-native';
+import Animated, {
+  useSharedValue,
+  useAnimatedStyle,
+  withTiming,
+  useAnimatedReaction,
+} from 'react-native-reanimated';
+
+const AnimatedReactionTest = () => {
+  const x = useSharedValue(0);
+  const x2 = useSharedValue(0);
+
+  const maxX2 = 80;
+
+  useAnimatedReaction(
+    () => {
+      'worklet';
+      return x.value / 1.5;
+    },
+    data => {
+      'worklet';
+      if (x2.value < maxX2) {
+        x2.value = data;
+      }
+    }
+  );
+
+  const uas = useAnimatedStyle(() => {
+    'worklet';
+    return {
+      transform: [
+        {
+          translateX: withTiming(x.value),
+        },
+      ],
+    };
+  });
+  const uas2 = useAnimatedStyle(() => {
+    'worklet';
+    return {
+      transform: [
+        {
+          translateX: withTiming(x2.value),
+        },
+      ],
+    };
+  });
+
+  const handle = () => {
+    x.value = x.value + 25;
+  };
+
+  const styles = StyleSheet.create({
+    box: {
+      height: 25,
+      width: 50,
+      backgroundColor: 'black',
+      marginTop: 10,
+    },
+  });
+
+  return (
+    <View>
+      <Text>
+        The second box should follow the first one for the first 5 moves(then it
+        should stop)
+      </Text>
+      <Button title="Move" onPress={handle} />
+      <Animated.View style={[styles.box, uas]} />
+      <Animated.View style={[styles.box, uas2]} />
+    </View>
+  );
+};
+
+export default AnimatedReactionTest;
--- a/Example/test/TestApp.js
+++ b/Example/test/TestApp.js
@@ -11,6 +11,7 @@ import SimpleTest from './SimpleTest';
 import MeasureTest from './MeasureTest';
 import FastRefreshTest from './FastRefreshTest';
 import ScrollToTest from './scrollToTest';
+import AnimatedReactionTest from './AnimatedReactionTest';
 
 YellowBox.ignoreWarnings(['Calling `getNode()`']);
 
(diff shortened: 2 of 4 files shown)


kmagiera:
The description does not seem to be accurate. The problem we're solving is not the inability to update SV in UAS but rather the lack of an API to support some of the usecases we discussed on the group. Would rather see those examples listed here with answers of how this new API fills this gap.

Beside that I think we should automatically workletize methods provided to the new hook, the same way we do for other hooks (to avoid the necessity of writing `'worklet'` directive).

On top of that I think I understand reaction API a bit different. From what I understand the first method is not to prepare a data but mostly to tell whether the second method should or should not run. The first one is run on every input change and we want to avoid the reaction firing in certain scenarios.

likern:
> The description does not seem to be accurate. The problem we're solving is not the inability to update SV in UAS but rather the lack of an API to support some of the usecases we discussed on the group. Would rather see those examples listed here with answers of how this new API fills this gap.
> 
> Beside that I think we should automatically workletize methods provided to the new hook, the same way we do for other hooks (to avoid the necessity of writing `'worklet'` directive).
> 
> On top of that I think I understand reaction API a bit different. From what I understand the first method is not to prepare a data but mostly to tell whether the second method should or should not run. The first one is run on every input change and we want to avoid the reaction firing in certain scenarios.

@karol-bisztyga This starts looking like an old V1 API with all conditions, `proc` etc. nodes.

```ts
const sharedWidth = useSharedValue(10)
const val = useSharedValue(10)
  val.value = sharedWidth.value * 2
  return {
    width: sharedWidth.value + 1
  }
})
```

As I understand shared values and worklets have automatic reactivity, opposed to React Hooks.
When it comes to this exact use case this is very similar to React Hooks, where there is the same problem.
```ts
const [value, setValue] = useState(0);
  useEffect(() => {
    console.log(`[useEffect] whenever value is changed`);
    setValue(val => val + 1); // here should be render loop !
  }, [value])
```
this is solved by using reference with direct manipulation
```ts
const ref = useRef(0);
  useEffect(() => {
    console.log(`[useEffect] whenever value is changed`);
    ref.current = ref.current + 1; // no render loop
  }, [ref])
```

why we couldn't solve this using a similar concept, like `useSharedReference` (which disables reactivity and is just usual value)?

In general I think to be as close to React concepts as possible might eliminate a lot of problems,
because React Hooks already implemented there own patterns and solutions to solve newly aroused problems.

likern:
We can rewrite example as this
```ts

const sharedWidth = useSharedValue(10)
const counterRef = useSharedReference(0);

// No need to do this computation in useAnimatedStyle
// with forever loop if we can do that in useDerivedValue

// on first render val has default value 50
// on all other renders will have current sharedWidth.value * 2
const val = useDerivedValue(() => {
  // do not cause rerender
  // changes are synchronous and visible immidiately
  counterRef.value = counterRef.value + 1;
  // on first render
 if (counterRef.value === 1) {
    return 50 // return default value
 } else {
    return sharedWidth.value * 2;
 }
})

  // Use val.value for something usefull here
  return {
    height: val.value,
    width: sharedWidth.value + 1
  }
})
```

karol-bisztyga:
> The description does not seem to be accurate. The problem we're solving is not the inability to update SV in UAS but rather the lack of an API to support some of the usecases we discussed on the group. Would rather see those examples listed here with answers of how this new API fills this gap.
> 
> Beside that I think we should automatically workletize methods provided to the new hook, the same way we do for other hooks (to avoid the necessity of writing `'worklet'` directive).
> 
> On top of that I think I understand reaction API a bit different. From what I understand the first method is not to prepare a data but mostly to tell whether the second method should or should not run. The first one is run on every input change and we want to avoid the reaction firing in certain scenarios.

Yes I accidentaly provided misleading description update, sorry about that the main goal is in fact to react to certain changes. 

Those methods for sure will be automatically workletized.

The first method does both of those things. What it's going to be used for mostly depends on the usecase.

likern:
@software-mansion Why my comments were deleted?

karol-bisztyga:
> @software-mansion Why my comments were deleted?

I'm sorry I made a mistake providing a description and they were not accurate anymore, I just wanted to clean up. Should have posted it as a draft in the fisrt place. Thank you for contribution anyway.

yolpsoftware:
Why was this closed without a merge? `useAnimatedReaction` would be great instead of `useDerivedValue` without using the derived value :-)

terrysahaidak:
could you describe your usecase @yolpsoftware ?

yolpsoftware:
My use case is coupling several ScrollViews together (#949). So I need a worklet that runs whenever one of the scroll offsets changes (scroll offsets will be shared values, I guess).

I could use `useDerivedValue` for the worklet as in the [scrollTo docs](https://docs.swmansion.com/react-native-reanimated/docs/next/api/nativeMethods/scrollTo/), but as @karol-bisztyga pointed out in #1082, it does not seem clean to use `useDerivedValue` when we are not interested in the derived value. That was when you proposed to use `useAnimatedReaction` for this use case.

It does not seem to be clean naming when `useDerivedValue` is the only option to run worklets when a value changes. We should have a method named `useAnimatedReaction` or `runOnChange`, even if it is just a wrapper around something similar to `useDerivedValue`.

karol-bisztyga:
Reopening due to incoming requests

Szymon20000 (approved):

PR #1027, opened 2020-07-26
Fix flickering on mount

### LINKED ISSUE #219: Changing style values from regular to Animated.Value removes view for several frames
## Description
If you have an `Animated.View` with regular `width: 100` style values and want to change it to an `Animated.Value` instance `width: new Animated.Value(100)`, it removes the view completely from view hierarchy for several frames and mounts again.

## Reproducing example
https://snack.expo.io/@terrysahaidak/reanimated-removes-view-bug

## Video
https://drive.google.com/file/d/1HIfFt9vWX23lcnZukQSlw1n15-1ZWzcw/view?usp=drivesdk

## Explanation
Why you ever want to change some values from animated one to regular?
The case I was playing with is lazy value initializations. 
Since creating a lot of `Animated.Value` blocks the js thread (see #194) and I have a lot of them on the same screen, I do not initialize them until I start the animation by some event (for example, pressing the button or focusing the input).

### LINKED ISSUE #762: Animated.View jumps position on mount on all platforms
## Description
Mounting
```
<Animated.View style={{ position: "absolute", bottom: new Animated.Value(0) }} />
```
results in the view briefly appearing at the top of its parent (as if we'd set `top: 0`) before subsequently jumping down into its correct position. This results in a janky-looking UI.

I've been able to reproduce this bug on all platforms.

### Screenshots
Here's a screen capture: <https://drive.google.com/file/d/1l7UdF-nZmzLsXpnzHW3WfEL4UxAcygTL/view?usp=sharing>

## Steps To Reproduce

1. Open https://snack.expo.io/@bdrobinson/jumpy-animated.view . I can reproduce the bug on the embedded web simulator, but it also reproduces on my physical android device.
2. Press the "Trigger remount!" button a few times in quick succession.

### Expected behavior
Each time you press "Trigger remount!", nothing changes visually.

### Actual behavior
Each time you press "Trigger remount!", the green footer briefly flashes at the top of the screen, then jumps back down to the bottom.

Interestingly, if you change `bottom: new Animated.Value(0)` to `bottom: 0`, the bug goes away.

https://snack.expo.io/@bdrobinson/jumpy-animated.view

## Package versions
I'm pretty sure this is a bug that affects all platforms (Web, iOS, Android).

Verified it on Expo v37.0.0, and also a normal RN app with the following versions:
- React: 16.9.0
- React Native: 0.61.2
- React Native Reanimated: 1.8.0

Thanks very much!

jakub-gonet:
## Description

Fixes #219 
Fixes #762

Previously components were flickering when mounted due to asynchronously updating component styles.

In `createAnimatedComponent()` we filtered out styles with Animated nodes, then component rendered with default values,
then nodes were initialized on the native side and applied their styles.
This issue was mostly observed when layout styles were changed, for example when setting height like in the example below.

This fix relies on updating the value of `Value` when nodes are detached (`__detach()` in `InternalAnimatedValue`) because it updates its value on the JS side. Without that this fix would only work on the first mount (when initial values are used).

Example:
```js
import {Button, Text, View} from 'react-native';

export default function Flickering() {
  const value = Animated.useValue(80);
  const [i, setI] = React.useState(0);
  const triggerRemount = () => {
    setI(i => i + 1);
  };
  return (
    <View
      key={i}
      style={{
        height: '100%',
        backgroundColor: 'white',
        position: 'relative',
        paddingTop: 100,
      }}>
      <Animated.View
        style={{
          height: value,
          position: 'absolute',
          width: '100%',
          backgroundColor: 'green',
        }}>
        <Text style={{color: 'white', fontSize: 15}}>Some footer</Text>
      </Animated.View>
      <Button
        onPress={() => {
          triggerRemount();
        }}
        style={{marginTop: 100}}
        title="Trigger remount!"
      />
    </View>
  );
}
```

### DIFF
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -197,8 +197,12 @@ export default function createAnimatedComponent(Component) {
       const style = {};
       for (const key in inputStyle) {
         const value = inputStyle[key];
-        if (!(value instanceof AnimatedNode) && key !== 'transform') {
-          style[key] = value;
+        if (key !== 'transform') {
+          if (value instanceof AnimatedNode) {
+            style[key] = value.__getValue();
+          } else {
+            style[key] = value;
+          }
         }
       }
       return style;
@@ -216,7 +220,9 @@ export default function createAnimatedComponent(Component) {
           // alltogether. Therefore we provide a dummy callback here to allow
           // native event dispatcher to hijack events.
           props[key] = dummyListener;
-        } else if (!(value instanceof AnimatedNode)) {
+        } else if (value instanceof AnimatedNode) {
+          props[key] = value.__getValue();
+        } else {
           props[key] = value;
         }
       }


src/createAnimatedComponent.js
```diff
@@ -197,8 +197,12 @@ export default function createAnimatedComponent(Component) {
...
       for (const key in inputStyle) {
         const value = inputStyle[key];
-        if (!(value instanceof AnimatedNode) && key !== 'transform') {
-          style[key] = value;
+        if (key !== 'transform') {
+          if (value instanceof AnimatedNode) {
+            style[key] = value.__getValue();
+          } else {
+            style[key] = value;
+          }
```
jkadamczyk, line 205: You could probably use ternary operator here, but I am not sure if it would be as readable as the code above 🤔

jakub-gonet: Thought about it but this is a bit long condition for my taste and decided that plain if is more readable

Ashoat:
Is there an ETA for a release with this change?

jakub-gonet:
I'd like to merge #1029 first, then do the 10.2 release

PR #1029, opened 2020-07-27
Update proc type

### LINKED ISSUE #1021: Typescript error - Can't build - Generic type 'Adaptable' requires 1 type argument(s)
## Description

When using [material-top-tabs](https://github.com/react-navigation/react-navigation/tree/main/packages/material-top-tabs) (which uses this library) in a typescript React Native project the command "yarn tsc" fails, this error doesn't happen in [V2](https://github.com/software-mansion/react-native-reanimated/releases/tag/2.0.0-alpha.4) 

## Steps To Reproduce

1. `npx --ignore-existing react-native init MyApp --template react-native-template-typescript`
2. Install dependencies: 
  2.1. `cd MyApp`
  2.2. `yarn add @react-navigation/material-top-tabs react-native-tab-view @react-navigation/native react-native-reanimated react-native-gesture-handler react-native-screens react-native-safe-area-context @react-native-community/masked-view`
3. Update app.ts with code example:
4. Run `yarn tsc`

### Expected behavior
Build without errors

### Actual behavior
Throws error:

`node_modules/react-native-reanimated/react-native-reanimated.d.ts:237:36 - error TS2314: Generic type 'Adaptable' requires 1 type argument(s).`

```
import 'react-native-gesture-handler';
import {NavigationContainer} from '@react-navigation/native';
import {createMaterialTopTabNavigator} from '@react-navigation/material-top-tabs';
import {Text} from 'react-native';

const HomeScreen = () => <Text>Example</Text>;

const Tab = createMaterialTopTabNavigator();

const App = () => {
  return (
    <NavigationContainer>
      <Tab.Navigator>
        <Tab.Screen name="Home" component={HomeScreen} />
      </Tab.Navigator>
    </NavigationContainer>
  );
};

export default App;
```

## Package versions

- @react-native-community/masked-view: ^0.1.10
- @react-navigation/material-top-tabs: ^5.2.14
- @react-navigation/native: ^5.7.1
- react: 16.13.1
- react-native: 0.63.0
- react-native-gesture-handler: ^1.7.0
- react-native-reanimated: ^1.10.1
- react-native-safe-area-context: ^3.1.1
- react-native-screens: ^2.9.0
- react-native-tab-view: ^2.15.
- @babel/core: ^7.8.4
- @babel/runtime: ^7.8.4
- @react-native-community/eslint-config: ^1.1.0
- @types/jest: ^25.2.3
- @types/react-native: ^0.63.1
- @types/react-test-renderer: ^16.9.2
- @typescript-eslint/eslint-plugin: ^2.27.0
- @typescript-eslint/parser: ^2.27.0
- babel-jest: ^25.1.0
- eslint: ^6.5.1
- jest: ^25.1.0
- metro-react-native-babel-preset: ^0.59.0
- prettier: ^2.0.4
- react-test-renderer: 16.13.1
- typescript: ^3.8.

jakub-gonet:
## Description

Fixes #1021.

This commit fixes regression introduced in #959 (f5eaf18) by specializing a generic parameter.
 
The previous implementation was failing with error:
```
node_modules/react-native-reanimated/react-native-reanimated.d.ts:237:36 - error TS2314: Generic type 'Adaptable' requires 1 type argument(s).
```
because `Adaptable` generic type was not specialized.

References:
www.typescriptlang.org/docs/handbook/utility-types.html#parameterst
www.typescriptlang.org/docs/handbook/utility-types.html#returntypet

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -234,9 +234,9 @@ declare module 'react-native-reanimated' {
     export const neq: BinaryOperator<0 | 1>;
     export const and: MultiOperator<0 | 1>;
     export const or: MultiOperator<0 | 1>;
-    export function proc<T extends (...args: any[]) => AnimatedNode<number>>(
-      func: T
-    ): (...funcArgs: Parameters<T>) => ReturnType<T>;
+    export function proc<T extends Adaptable<Value>[]>(
+      func: (...args: T) => AnimatedNode<number>
+    ): typeof func;
     export function defined(value: Adaptable<any>): AnimatedNode<0 | 1>;
     export function not(value: Adaptable<any>): AnimatedNode<0 | 1>;
     export function set<T extends Value>(
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -234,7 +234,7 @@ declare module 'react-native-reanimated' {
     export const neq: BinaryOperator<0 | 1>;
     export const and: MultiOperator<0 | 1>;
     export const or: MultiOperator<0 | 1>;
-    export function proc<T extends Adaptable<Value>[]>(
+    export function proc<T extends (Adaptable<Value> | undefined)[]>(
       func: (...args: T) => AnimatedNode<number>
     ): typeof func;
     export function defined(value: Adaptable<any>): AnimatedNode<0 | 1>;
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -234,9 +234,9 @@ declare module 'react-native-reanimated' {
     export const neq: BinaryOperator<0 | 1>;
     export const and: MultiOperator<0 | 1>;
     export const or: MultiOperator<0 | 1>;
-    export function proc<A extends Adaptable[]>(
-      cb: (...args: A) => AnimatedNode<number>
-    ): typeof cb;
+    export function proc<T extends (...args: any[]) => AnimatedNode<number>>(
+      func: T
+    ): (...funcArgs: Parameters<T>) => ReturnType<T>;
     export function defined(value: Adaptable<any>): AnimatedNode<0 | 1>;
     export function not(value: Adaptable<any>): AnimatedNode<0 | 1>;
     export function set<T extends Value>(


wcandillon:
Isn't the fix to edit `Adaptable` to ` Adaptable<number>`?

wcandillon:
In master, the version looks correct, sorry for the confusion.

wcandillon:
I was able to reproduce the error on v1 and confirm that

```
    export function proc(
      cb: (...params: Adaptable<number>[]) => AnimatedNode<number>
    ): typeof cb;
```

fixes the issue.

jakub-gonet:
but still, `Adaptable<number>` prevents using `undefined` as a parameter.

terrysahaidak:
I had suggestion to use `number | undefined` here. It worked for me.

What types of parameters proc accepts @jakub-gonet?
Because not sure it actually accepts `any` - this may lead to some runtime errors.

wcandillon:
oh right. What about:

```
    export function proc<A>(
    ): typeof cb;
```
I'm just not a big fan of any ;-)

wcandillon:
But also it should be all possible subtypes in proc() (Clock, Adaptable<string | number | boolean>). Not sure why undefined is supported

jakub-gonet:
`undefined` is for `defined()` node support.

> oh right. What about:
>
> ```
>    export function proc<A>(
>      cb: (...args: A) => AnimatedNode<number>
>    ): typeof cb;
> ```
> I'm just not a big fan of any ;-)

I think this will prevent from using different arguments types.


`createAnimatedFunction` supports all adapted types in args, so any node should work as an argument. I wanted to be more permissive there, you can use pretty much any function that returns the animated node if I'm not mistaken.

wcandillon:
It doesn't prevent it: https://www.dropbox.com/s/7ee6wdnejo1pzou/Screenshot%202020-07-27%2019.45.44.png?dl=0. This is what triggered the change already, that TS is capable to infer the tuple. But it would be nice to restrict `A`

jakub-gonet:
@wcandillon using your type fails for me with error `A rest parameter must be of an array type.ts(2370)`

After changing it to the:
```js
 export function proc<A>(
      cb: (...args: A[]) => AnimatedNode<number>
    ): typeof cb;
```

code like:
```js
const v_n = new Value(1);
const v_s = new Value('test');
const procc = proc((v_n: Animated.Value<number>, v_s: Animated.Value<string>) => {
  return block([add(v_n, v_n)]);
});
```

fails with 
```
App.tsx:20:3 - error TS2345: Argument of type '(v_n: Animated.Value<number>, v_s: Animated.Value<string>) => Animated.Node<number>' is not assignable to parameter of type '(...args: (AnimatedValue<number> | AnimatedValue<string>)[]) => AnimatedNode<number>'.
  Types of parameters 'v_n' and 'args' are incompatible.
    Type 'AnimatedValue<number> | AnimatedValue<string>' is not assignable to type 'AnimatedValue<number>'.
      Type 'AnimatedValue<string>' is not assignable to type 'AnimatedValue<number>'.
        Type 'string' is not assignable to type 'number'.

20   (v_n: Animated.Value<number>, v_s: Animated.Value<string>) => {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Do I have old version of TS or I'm doing something wrong here?

terrysahaidak:
@jakub-gonet what type of parameters proc accepts? is there any list? I think it should just use it or we need to create a type which represents it.

jakub-gonet:
It appears that I was wrong about this `any[]` :>

Only Animated nodes can be used there.

jakub-gonet:
I tinkered with it a bit and seems like first @wcandillon suggestion works. 

@wcandillon, @terrysahaidak it'd be awesome if you guys could verify that it works correctly

terrysahaidak:
what about undefined though?

jakub-gonet:
Done :>

But I'm pretty sure not that many people used `undefined` as an arg because proc checks if it's used with the correct number of args (it should make not specified args undefined IMHO, but that's a separate PR).

BTW, I didn't use Nullable, because we don't support using `null`s as `Value`s anymore.

wcandillon:
I like it :)

PR #1030, opened 2020-07-27
Fix initialisation of NodesManager

### LINKED ISSUE #604: [Android] NullPointerException - java.util.ArrayDeque.addLast from com.swmansion.reanimated.NodesManager.startUpdatingOnAnimationFrame
Hi there,

We started seeing this error when we upgraded another library which in turn required an upgrade to Reanimated from 1.0.0 to 1.4.0:
```
java.lang.NullPointerException
    ArrayDeque.java:228 java.util.ArrayDeque.addLast
    ReactChoreographer:91 com.facebook.react.modules.core.ReactChoreographer.postFrameCallback
    NodesManager:130 com.swmansion.reanimated.NodesManager.startUpdatingOnAnimationFrame
    NodesManager:382 com.swmansion.reanimated.NodesManager.onEventDispatch
    EventDispatcher:113 com.facebook.react.uimanager.events.EventDispatcher.dispatchEvent
    UIImplementation:919 com.facebook.react.uimanager.UIImplementation.applyUpdatesRecursive
...
    UIImplementation:904 com.facebook.react.uimanager.UIImplementation.applyUpdatesRecursive
    UIImplementation:644 com.facebook.react.uimanager.UIImplementation.updateViewHierarchy
    UIImplementation:599 com.facebook.react.uimanager.UIImplementation.dispatchViewUpdates
    UIManagerModule:782 com.facebook.react.uimanager.UIManagerModule.onBatchComplete
    NativeModuleRegistry:118 com.facebook.react.bridge.NativeModuleRegistry.onBatchComplete
    CatalystInstanceImpl:164 com.facebook.react.bridge.CatalystInstanceImpl$BridgeCallback.onBatchComplete
    SourceFile:-2 com.facebook.react.bridge.queue.NativeRunnable.run
    Handler.java:789 android.os.Handler.handleCallback
    Handler.java:98 android.os.Handler.dispatchMessage
    MessageQueueThreadHandler:26 com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage
    Looper.java:164 android.os.Looper.loop
    MessageQueueThreadImpl:225 com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run
    Thread.java:764 java.lang.Thread.run
```
This implies that `NodesManager`'s `mChoreographerCallback` is null, which doesn't make sense since it gets initialized in the constructor.

Since other issues have mentioned that 1.7.0 has fixes for other NPEs, we tried upgrading to 1.7.0, but that hasn't helped. I'm still investigating, but am wondering if others have seen this problem or know a workaround?

Thanks!

jakub-gonet:
## Description

Previously, `NodesManager` registered (passing this) as an event listener before ending initialization in c-tor.

`NodesManager` is created by `ReanimatedModule` (when calling `getNodesManager()` for the first time) on [UI thread](https://github.com/software-mansion/react-native-reanimated/blob/9af276693d136b712259fe9dd0126aee6c78bd4b/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java#L77-L80).

When `onEventDispatch()` is called on non-UI thread it calls `startUpdatingOnAnimationFrame()`, which in turn uses `mChoreographerCallback`.
`mChoreographerCallback` is initialised **after** listener is registered which causes NPE in React Choreographer's `postFrameCallback()` method after trying adding callback to the queue.

Fixes #604.

References:
https://www.ibm.com/developerworks/java/library/j-jtp0618/index.html#2
https://stackoverflow.com/questions/25281301/what-exactly-are-the-dangers-of-passing-this-from-a-java-constructor

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -97,8 +97,6 @@ public class NodesManager implements EventDispatcherListener {
     updateContext = new UpdateContext();
     mUIImplementation = mUIManager.getUIImplementation();
     mCustomEventNamesResolver = mUIManager.getDirectEventNamesResolver();
-    mUIManager.getEventDispatcher().addListener(this);
-
     mEventEmitter = context.getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class);
 
     mReactChoreographer = ReactChoreographer.getInstance();
@@ -110,6 +108,12 @@ public class NodesManager implements EventDispatcherListener {
     };
 
     mNoopNode = new NoopNode(this);
+
+    // We register as event listener at the end, because we pass `this` and we haven't finished contructing an object yet.
+    // This lead to a crash described in https://github.com/software-mansion/react-native-reanimated/issues/604 which was caused by Nodes Manager being constructed on UI thread and registering for events.
+    // Events are handled in the native modules thread in the `onEventDispatch()` method.
+    // This method indirectly uses `mChoreographerCallback` which was created after event registration, creating race condition
+    mUIManager.getEventDispatcher().addListener(this);
   }
 
   public void onHostPause() {


android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
```diff
@@ -92,7 +92,7 @@ public String getName() {
 
   private NodesManager getNodesManager() {
     if (mNodesManager == null) {
-      mNodesManager = new NodesManager(getReactApplicationContext());
+      mNodesManager = NodesManager.createNodesManager(getReactApplicationContext());
```
jakub-gonet, line 95: if the node manager is used as a singleton maybe we should do it in the `createNodesManager()` method?

jkadamczyk: Yup I agree. Optimally I would make createNodesManager private,, or move that logic to constructor that will be private and have just `getInstance()` method to retrieve it and create it when it is not yet initialised

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -112,6 +113,15 @@ protected void doFrameGuarded(long frameTimeNanos) {
     mNoopNode = new NoopNode(this);
   }
 
+  public static NodesManager getInstanceWithContext(ReactContext context){
+    if(sInstance == null || sInstance.mContext != context) {
```
jakub-gonet, line 117: I'm not sure if we should recreate instance if the reference to `ReactContext` changed

PR #1033, opened 2020-07-28
Make `interpolate` `type` optional

wcandillon:
I believe that it is optional in the JS implementation.


jakub-gonet:
Could you please rebase against v1 branch?

jakub-gonet:
Oh, this is new `interpolate`, my bad.

PR #1040, opened 2020-07-29
Add useAnimatedRef

karol-bisztyga:
## Description
Create new hook `useAnimatedRef` to provide a functionality of animated reference which acts as a normal reference on the javascript thread but at the same time holds a view tag of the component which it is referring to through a shared value.

It was created to simplify code of `scrollTo` example but may be used in other cases as well.

### DIFF
--- a/Example/src/MeasureExample.js
+++ b/Example/src/MeasureExample.js
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react';
+import React, { useRef } from 'react';
 import { StyleSheet, View, Text } from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
 import Animated, {
@@ -7,9 +7,9 @@ import Animated, {
   useAnimatedGestureHandler,
   measure,
   withTiming,
-  getTag,
   Easing,
   useDerivedValue,
+  useAnimatedRef,
 } from 'react-native-reanimated';
 import { TapGestureHandler } from 'react-native-gesture-handler';
 
@@ -84,7 +84,7 @@ export default function MeasureExample() {
 }
 
 function Section({ title, children, height, contentHeight, z, show }) {
-  const [randomComRef, setRef] = useState(null);
+  const aref = useAnimatedRef();
 
   const stylez = useAnimatedStyle(() => {
     return {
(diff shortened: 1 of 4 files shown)


terrysahaidak:
Quick example which won't work with that API @kmagiera 

I want to bukd animated input. I need a ref to interact with it - focus, blue etc. But also I would want to use measure to know the exact height and width of that input for animation (on focus, for example). How can I achieve it?

kmagiera:
Yeah, I am not convinced measure is actually a good fit for being available as a sync method. I think we will keep it this way for now, but here is what I think:

In your usecase, you want the style to depend on the view dimensions. If that's the case you want the worklet to react to dimension change, and a better fit for this case is to have dimensions stored as shared values hooked to onLayout event and then capture them in the worklet. Otherwise how can we tell the size has changed if we are not subscribed to the event?

kmagiera:
If you want the ref to be used in reactjs code and by a worklet I think we can add `.current` field to it that'd just point to the value we receive in the ref callback

terrysahaidak:
In my example I need the size just before the animation happens, not on some change - then you focus the input. I can measure it using async measure, but it adds 20-60ms delay, so sync measure just before the animation start would be perfect here.

All we need here - make sure we also allow users to use it as regular ref, since that ref stores instance of the component with all the methods, not just tag or so.

For example, this could be really handy for different font sizes (when we have support for don't scalling) and material floating label input. In that case we can measure label, input itself and know where to animate them just on focus.

karol-bisztyga:
I think we can put the target component in the ref alongside with the tag(and just update the usage in `scrollTo`) as well as any other fields in the future if needed. If that solvers @terrysahaidak's case I do not see a problem. It would also be more intuitive I think.

Szymon20000:
I think it's enough to provide a ref only on JS side. In that way, we won't copy the whole view object. 
```js
export function useAnimatedRef() {
  const tag = useSharedValue(-1)
  const function refFun(comp) {
      'worklet';
      if (comp !== undefined) {
        refFun.current = comp;
        tag.value = getTag(comp);
      }
      return tag.value
    }
  }
  Object.defineProperty(refFun, 'current', {value: null, writable:true, enumerable:false});
  return refFun;
}
```

Szymon20000:
@karol-bisztyga Can you make the same change for measure function?

terrysahaidak:
> I think it's enough to provide a ref only on JS side. In that way, we won't copy the whole view object. 
> ```js
> export function useAnimatedRef() {
>   const tag = useSharedValue(-1)
>   const function refFun(comp) {
>       'worklet';
>       // enters when ref is set by attaching to a component
>       if (comp !== undefined) {
>         refFun.current = comp;
>         tag.value = getTag(comp);
>       }
>       return tag.value
>     }
>   }
>   Object.defineProperty(refFun, 'current', {value: null, writable:true, enumerable:false});
>   return refFun;
> }
> ```

This might fix it for all the cases. In your code Define property should be executed only if there is no such property, though

PR #1042, opened 2020-07-29
Handle rerender behaviour

karol-bisztyga:
## Description

This is supposed to provide a proper way to handle React Native's [Fast refresh](https://reactnative.dev/docs/fast-refresh) and also rerender of a component containing certain animations.

## Problem 🦯 

On component rerender all of the worklets which are included into that component are not rebuilt. This introduces following problems:
1. If someone changes worklet's body it will not be updated using React Native's [Fast refresh](https://reactnative.dev/docs/fast-refresh)
2. When worklet uses component's state and this state changes(causing rerender) that worklet will use outdated state value
<details>
<summary>ad. 2 example</summary>

``` JS
const Comp = () => {
  const [state, setState] = useState(0)
  const udv = useDerivedValue(() => {
    return state
  })
  return (
    <>
      <Button title="press" onPress={ () => {
        setState(state + 1)
      } } />
      <Text>Value(will not update): { udv.value }</Text>
    </>
  )
}
```
</details>

### Solution 1 👎 

As a first solution we came up with an idea of rebuilding everything on every hook call considering:
- `useSharedValue`
- `useAnimatedStyle`
- `useDerivedValue`

However after running some performance tests we noticed that `useSharedValue`'s performance has not changed that much but as for the other mentioned hooks they have slowed significantly. That led us to another solution.

### Solution 2 👍 

This is what is presented in this PR.
Shared value is reconstructed in `useSharedValue` when:
- it has not been created yet
- the hook is called with a different init value than in the previous render

`useAnimatedStyle` and `useDerivedValue` act different(but in the same manner as each other). They receive an array of dependencies like `useEffect` hook. The worklet inside is being reconstructed in following cases:
- it has not been created yet(this is first hook call)
- any provided dependencies changed(or there have not been any provided which is equal to listen to any change - for now `null` has to be passed, default is `[]`)
- worklet hash changed(which means worklet body changed)

### Changes

In `Hooks.js` three hooks has been modified as described above.
`FastRefreshTest.js` a testing screen has been added to check some of the most common use cases.

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -256,14 +256,10 @@ function styleUpdater(viewTag, updater, state) {
   }
 }
 
-export function useAnimatedStyle(updater, dependencies = []) {
+export function useAnimatedStyle(updater, dependencies) {
   const viewTag = useSharedValue(-1);
 
   const initRef = useRef(null);
-
-  if (dependencies === null) {
-    dependencies = undefined;
-  }
   /**
    * create(or recreate) the object if
    *  it has not been created yet
@@ -321,12 +317,9 @@ export function useAnimatedStyle(updater, dependencies = []) {
 // when you need styles to animated you should always use useAS
 export const useAnimatedProps = useAnimatedStyle;
 
-export function useDerivedValue(processor, dependencies = []) {
+export function useDerivedValue(processor, dependencies) {
   const initRef = useRef(null);
 
-  if (dependencies === null) {
-    dependencies = undefined;
-  }
   /**
    * create(or recreate) the object if
    *  it has not been created yet
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -261,10 +261,13 @@ export function useAnimatedStyle(updater, dependencies = []) {
 
   const initRef = useRef(null);
 
+  if (dependencies === null) {
+    dependencies = undefined;
+  }
   /**
    * create(or recreate) the object if
    *  it has not been created yet
-   *  any provided dependencies changes(or there have not been any provided which is equal to listen to any change)
+   *  any provided dependencies changed(or there have not been any provided which is equal to listen to any change)
    *  worklet hash changed(which means worklet body changed)
    */
   if (
@@ -321,10 +324,13 @@ export const useAnimatedProps = useAnimatedStyle;
 export function useDerivedValue(processor, dependencies = []) {
   const initRef = useRef(null);
 
+  if (dependencies === null) {
+    dependencies = undefined;
+  }
   /**
    * create(or recreate) the object if
    *  it has not been created yet
-   *  any provided dependencies changes(or there have not been any provided which is equal to listen to any change)
+   *  any provided dependencies changed(or there have not been any provided which is equal to listen to any change)
    *  worklet hash changed(which means worklet body changed)
    */
   if (
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -258,40 +258,38 @@ function styleUpdater(viewTag, updater, state) {
 
 export function useAnimatedStyle(updater, dependencies) {
   const viewTag = useSharedValue(-1);
-
   const initRef = useRef(null);
-  /**
-   * create(or recreate) the object if
-   *  it has not been created yet
-   *  any provided dependencies changed(or there have not been any provided which is equal to listen to any change)
-   *  worklet hash changed(which means worklet body changed)
-   */
-  if (
-    initRef.current === null ||
-    handleDependenciesChange(initRef, dependencies) ||
-    initRef.current.__workletHash !== updater.__workletHash
-  ) {
+  const inputs = Object.values(updater._closure);
+
+  // build dependencies
+  if (dependencies === undefined) {
+    dependencies = [...inputs, updater.__workletHash];
+  } else {
+    dependencies.push(updater.__workletHash);
+  }
+
+  if (initRef.current === null) {
     const initial = initialUpdaterRun(updater);
     initRef.current = {
       initial,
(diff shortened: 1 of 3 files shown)


src/reanimated2/Hooks.js
```diff
@@ -8,23 +8,29 @@ import { initialUpdaterRun } from './animations';
...
 
-export function useMapper(fun, inputs = [], outputs = []) {
+export function useMapper(fun, inputs = [], outputs = [], dependencies = []) {
   useEffect(() => {
     const mapperId = startMapper(fun, inputs, outputs);
     return () => {
       stopMapper(mapperId);
     };
-  }, []);
+  }, dependencies);
```
karol-bisztyga, line 29: not sure if this is necessary

src/reanimated2/Hooks.js
```diff
@@ -293,12 +320,28 @@ export function useAnimatedStyle(updater) {
 // when you need styles to animated you should always use useAS
 export const useAnimatedProps = useAnimatedStyle;
 
-export function useDerivedValue(processor) {
+export function useDerivedValue(processor, dependencies = []) {
   const initRef = useRef(null);
-  if (initRef.current === null) {
+
+  if (dependencies === null) {
```
karol-bisztyga, line 326: for now the default value for `dependencies` is `[]` so if this argument is not passed the worklet reconstruction mechanism is not listening on anything. If `null` is passed it is listening on any change so it does not work exactly like the other hooks such as `useEffect`. This is a temporal solution, just wanted to avoid reconstructing every worklet by default, but this needs a discusion. The same refers to `useAnimatedStyle`

terrysahaidak:
Won't recreating sharedValues on each render cause some loops? for example, if we pass it as a dependency to some useEffect?

Also, I didn't get why they should change and how they're going to change on each rerender. Will we lose reference? Or it just assigns a new value?

karol-bisztyga:
> Won't recreating sharedValues on each render cause some loops? for example, if we pass it as a dependency to some useEffect?

No because `useEffect` runs after the render and shared value change does not trigger the render and if you look at the code here so called 'recreating' the SV is in fact just reassigning the value.

> Also, I didn't get why they should change and how they're going to change on each rerender. Will we lose reference? Or it just assigns a new value?

Just reassigns, all we want is updated value from the current render.

terrysahaidak:
What about such example, will it reset the value of sharedValue because of showContent change? If so, why? I would not expect this behavior. 

```jsx
function My component() {
  const height = useSharedValue(0);
  const [showContent, setShowContent] = useState(false); 

  function expand() {
    height.value = runTiming(200, undefined, () => setShowContent(true));
  }

  return (
    < Animated.View style={styles}>
      {showContent && content}
    </Animated.View>
  )
}
````

**Comment by karol-bisztyga** [comment]
> What about such example, will it reset the value of sharedValue because of showContent change? If so, why? I would not expect this behavior.
> 
> ```js-jsx
> function MyComponent() {
>   const height = useSharedValue(0);
>   const [showContent, setShowContent] = useState(false); 
> 
>   function expand() {
>     height.value = runTiming(200, undefined, () => setShowContent(true));
>   }
> 
>   return (
>     <Animated.View style={styles}>
>       {showContent && content}
>     </Animated.View>
>   )
> }
> ```

No it will not, the `height` value will still be `200` after rerender triggered by the callback set state. It's all in `useSharedValue` hook, the value is reset if it has not been set yet or the hook has been called with different init value than last time. Now I see I made a mistake in the description, thanks for pointing out @terrysahaidak :)

**Comment by terrysahaidak** [comment]
I still think that we should stick with similar to useMemo behavior here. Otherwise it's really unclear, at least it was for me

So the proposal is:
- 1 parameter to be value/function which returns value. 
- 2 parameter to be an array of dependencies when we should re-init it. Just like all the other hooks.

**Comment by karol-bisztyga** [comment]
This PR has been updated. Some code has been moved(to get rid of `useMapper`) but acts more or less the same. The key difference here is how we build dependencies. If the dep array isn't provided, we assign the worklet's closure and its hash(which tells whether it's changed), otherwise we only take the hash.

File: src/reanimated2/Hooks.js [hunk 3]
```diff
@@ -249,25 +255,40 @@ function styleUpdater(viewTag, updater, state) {
...
-export function useAnimatedStyle(updater) {
+export function useAnimatedStyle(updater, dependencies) {
   const viewTag = useSharedValue(-1);
-
   const initRef = useRef(null);
+  const inputs = Object.values(updater._closure);
+
+  // build dependencies
+  if (dependencies === undefined) {
+    dependencies = [...inputs, updater.__workletHash];
```
Szymon20000, line 265: Why don't we just pass undefined to useEffect?

terrysahaidak: Do you want users to use [] explicitly if they want it to not recreate?

Szymon20000: Currently, you have to pass null if you don't want to recreate worklet. I just think that there is no need to add an input array to dependencies in any case. It won't work with complex objects. For instance:
```JS
useAnimatedStyle(() => {return {backgroundColor: state.color}});
```
In my option, the default behavior should be recreating always or never.

karol-bisztyga: This has been discused beyond github

karol-bisztyga:
> I still think that we should stick with similar to useMemo behavior here. Otherwise it's really unclear, at least it was for me
> 
> So the proposal is:
> 
> * 1 parameter to be value/function which returns value.
> * 2 parameter to be an array of dependencies when we should re-init it. Just like all the other hooks.

I do not see any difference between your description and the way it is implemented now ;)

src/reanimated2/Hooks.js
karol-bisztyga: I decided to leave it as it is as it could go handy in some case

PR #1048, opened 2020-07-30
Remove emoji from landing page

jakub-gonet:
## Description

It just looked bad ngl

### Before
image

### After

image


jakub-gonet:
👍

PR #1068, opened 2020-08-04
Make diffClamp non-proc'ed

### LINKED ISSUE #570: 1.7.0 regression while moving objects using PanGestureHandler
Hello there.

I've noticed some regression using latest 1.7.0 reanimated while moving objects. Using older 1.4.0 everything works just fine.

## How it should look like

good

## How it actually looks like
bad

## Reproducible Demo

Please find reproducible demo in following git repository [repo](https://github.com/sebinq/react-native-reanimated-bug)

Thanks.

### LINKED ISSUE #767: diffClamp conflict when used more than once?
## Description

When I use diffClamp to operate on a single value the results are predictable.
Once a second instance of diffClamp is employed both values become unpredictable.

## Screenshots

panTest

## Steps To Reproduce

```jsx
import React from "react";
import { Dimensions, SafeAreaView, StyleSheet, View } from "react-native";
import Animated, {
  Value,
  add,
  cond,
  diffClamp,
  eq,
  event,
  set,
import { PanGestureHandler, State } from "react-native-gesture-handler";

const SCREEN = {
  height: Dimensions.get("screen").height,
  width: Dimensions.get("screen").width,
};

const BOUNDS = {
  x: {
    min: 0,
    max: SCREEN.width - 100,
  },
  y: {
    min: 0,
    max: SCREEN.height - 100,
  },
};

const PanTest = () => {
  const state = new Value(State.UNDETERMINED);
  const translationX = new Value(0);
  const translationY = new Value(0);

  const gestureHandler = event([
    {
      nativeEvent: {
        state,
        translationX,
        translationY,
      },
    },
  ]);

  const withOffset = (
    value: Animated.Node<number>,
    gestureState: Animated.Node<number>
  ): Animated.Node<number> => {
    const offset = new Value(0);
    return cond(
      eq(gestureState, State.END),
      [set(offset, add(offset, value)), offset],
      add(offset, value)
    );
  };

  const translateX = diffClamp(
    withOffset(translationX, state),
    BOUNDS.x.min,
    BOUNDS.x.max
  );

  const translateY = diffClamp(
    withOffset(translationY, state),
    BOUNDS.y.min,
    BOUNDS.y.max
  );

  return (
    <SafeAreaView>
        <PanGestureHandler
          onHandlerStateChange={gestureHandler}
          onGestureEvent={gestureHandler}
        >
          <Animated.View
            style={[
              styles.box,
              {
                transform: [{ translateX }, { translateY }],
              },
            ]}
          />
        </PanGestureHandler>
      </View>
    </SafeAreaView>
  );
};

  container: {
    flex: 1,
  },
  box: {
    position: "absolute",
    width: 100,
    height: 100,
    backgroundColor: "red",
  },
});

export default PanTest;

```

### Expected behavior

Values should remain independent, and within the bounds specified, regardless of how many instances of `diffClamp` are used

### Actual behavior

Seemingly, multiple instances of `diffClamp` affect values in others.

## Package versions

- Expo: 37.0.0
- React: 16.9.0
- React Native: 0.61
- React Native Reanimated: 1.7.0
- React Native Gesture Handler: 1.6.0

jakub-gonet:
## Description

Fixes #570.
Fixes #767.

Creating procs of stateful functions (like diff or diffClamp) make them effectively a singleton, which means that every instance of it shares nodes (Value nodes in particular). 

In Reanimated most of derived nodes are proc'ed and stateful ones are `diff()`, `diffClamp()`, `acc()` and `onChange()`. We export those as plain JS functions, creating state out of proc'ed functions. This works because every use of them creates a new, non-shared state. When one of those functions is used in proc context, they're created once – they share state.
 
The example below shows this behavior.

## Example

```js
const {Value, block, proc, set, useCode, debug, add} = Animated;

const procStateful = proc(x => {
  const state = new Value(0.5);
  return block([set(state, add(state, x)), state]);
});

const Example = () => {
  const a = new Value(2);
  const b = new Value(4);

  const call_1 = procStateful(a);
  const call_2 = procStateful(b);

  useCode(
    () => [
      /* should print 2+0.5=2.5 */
      debug('PROC 1', call_1),
      /* should print 4+0.5=4.5, prints 4+2.5 (prev state) = 6.5 */
      debug('PROC 2', call_2),
    ],
    [],
  );

  return null;
};

export default Example;
```

### DIFF
--- a/src/derived/diffClamp.js
+++ b/src/derived/diffClamp.js
@@ -1,15 +1,11 @@
-import { cond, defined, set, add, proc, min, max } from '../base';
+import { cond, defined, set, add, min, max } from '../base';
 import AnimatedValue from '../core/InternalAnimatedValue';
 import diff from './diff';
 
-const procAcc = proc(function(a, minVal, maxVal, value) {
+export default function diffClamp(a, minVal, maxVal) {
+  const value = new AnimatedValue();
   return set(
     value,
     min(max(add(cond(defined(value), value, a), diff(a)), minVal), maxVal)
   );
-});
-
-export default function diffClamp(a, minVal, maxVal) {
-  const value = new AnimatedValue();
-  return procAcc(a, minVal, maxVal, value);
 }


jakub-gonet:
Maybe the previous description wasn't clear enough, I updated it (with simplified example).

jakub-gonet:
This doesn't fix the root cause which is sharing state between proc calls but I think this is not really fixable, because procs are created only once, globally. We could export some helper which clones procs, but IMHO it'd defeat the whole purpose of procs.

It may be worth mentioning stateless caveat in procs docs though.

PR #1084, opened 2020-08-07
Use native measure in lightbox example

terrysahaidak:
## Description

The react-native-reanimated@alpha.5 release introduced measure function in the worklet. Previously we used RN's async measure for that. Now we can measure things just before the actual animation. This change helps us to run the animation 10-25ms faster on my Pixel 3.

## Changes

- Migrate lightbox example to native measure

### DIFF
--- a/Example/src/LightboxExample.js
+++ b/Example/src/LightboxExample.js
@@ -2,7 +2,6 @@ import React, { useState, useEffect } from 'react';
 import Animated, {
   useSharedValue,
   useAnimatedStyle,
-  runOnUI,
   useAnimatedGestureHandler,
   interpolate,
   Extrapolate,
@@ -15,7 +14,7 @@ import { Dimensions, StyleSheet, View, Image, StatusBar } from 'react-native';
 import {
   ScrollView,
   PanGestureHandler,
-  TouchableWithoutFeedback,
+  TapGestureHandler,
 } from 'react-native-gesture-handler';
 import { Header } from 'react-navigation-stack';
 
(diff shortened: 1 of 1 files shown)


kmagiera:
I like the idea we can simplify this example. However, I think we don't want to expose use of `runOnUI` as a public API. This method is not documented and despite we are open to considering a public API it is still not the case. I think that for this example it makes to use an API that's been proposed in #1019. We will likely need something like this anyways, so maybe we can wait with this update till we get an API for running arbitrary code for side-effects

terrysahaidak:
Currently, runOnUI is used for two reasons:
1. To make sure both `animationProgress` withTiming and backdropOpacity withTiming are executed simultaneously. Is this going to be the same without runOnUi? Haven't tried it though.
2. To make sure we measure the view as soon as possible. The other best place to do it is `useAnimatedStyle` but we need to update all the shared values too which is side effect.

I also thought about measuring it inside the handler of the tap on image.

Example/src/LightboxExample.js
```diff
@@ -182,9 +204,15 @@ function ImageTransition({ activeImage, onClose }) {
   useEffect(() => {
     runOnUI(() => {
       'worklet';
-      animationProgress.value = withTiming(1, timingConfig, () => {
+
+      measureStuff();
+
+      // fixes flickering of the image
+      setTimeout(() => {
```
karol-bisztyga, line 211: that timeout looks ugly, wouldn't it be better to use something like `imageOpacity.value = withTiming(0, { duration: 16 });`?

terrysahaidak: I'll try so, but not sure. It might be even some bug with ordering of updates because the fix is single frame delay.

terrysahaidak:
Updated pr @kmagiera @karol-bisztyga

terrysahaidak:
Currently, there is a 70-90ms delay before an actual animation starts (this includes render of the image and executing all the worklets).
Sync measure reduced the previous one by 16-30ms.
Hope one day we could reduce the delay this even more.

karol-bisztyga (approved):

Issue #1062, opened 2020-08-03
[v2] rotate transform (withTiming) exception on Android


Johan-dutoit:
## Description

The following snippet throws an exception on Android but works on iOS.

Exception:  Invariant Violation: Transform with key of "rotateZ" must be a string: {"rotateZ":0}
Adding deg or rad to the string just crashes the app instead.

```typescript
const animatedStyle = useAnimatedStyle(() => {
    const toValue = isOpen.value ? Math.PI / 4 : 0;
    const rotateZ = withTiming(toValue, {
      duration: 100,
      easing: Easing.linear,
    });

    return {
      transform: [
        {
          rotateZ,
        },
      ],
    };
});
```

### Screenshots
![image](https://user-images.githubusercontent.com/1611055/89190460-c92f6980-d5a1-11ea-9381-df34232ed54f.png)

## Steps To Reproduce

use any of the rotate transforms on Android and apply withTiming

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.4

## Logcat

```
 Error while updating prop transform
    java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Method.invoke(Native Method)
        at com.facebook.react.uimanager.ViewManagersPropertyCache$PropSetter.updateViewProp(ViewManagersPropertyCache.java:87)
        at com.facebook.react.uimanager.ViewManagerPropertyUpdater$FallbackViewManagerSetter.setProperty(ViewManagerPropertyUpdater.java:136)
        at com.facebook.react.uimanager.ViewManagerPropertyUpdater.updateProps(ViewManagerPropertyUpdater.java:56)
        at com.facebook.react.uimanager.ViewManager.updateProperties(ViewManager.java:49)
        at com.facebook.react.uimanager.NativeViewHierarchyManager.updateProperties(NativeViewHierarchyManager.java:137)
        at com.facebook.react.uimanager.UIImplementation.synchronouslyUpdateViewOnUIThread(UIImplementation.java:291)
        at com.swmansion.reanimated.NodesManager.updateProps(NodesManager.java:473)
        at com.swmansion.reanimated.NativeProxy.updateProps(NativeProxy.java:91)
        at com.swmansion.reanimated.NativeProxy$AnimationFrameCallback.onAnimationFrame(Native Method)
        at com.swmansion.reanimated.NodesManager.onAnimationFrame(NodesManager.java:174)
[...]

PR #1087, opened 2020-08-07
remove initialUpdaterRun from useAnimatedStyle

karol-bisztyga:
## Description

Removed `initialUpdaterRun` from `useAnimatedStyle` hook.

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -255,7 +255,7 @@ export function useAnimatedStyle(updater) {
 
   const initRef = useRef(null);
   if (initRef.current === null) {
-    const initial = initialUpdaterRun(updater);
+    const initial = {};
     initRef.current = {
       initial,
       remoteState: makeRemote({ last: initial }),
@@ -451,4 +451,3 @@ export function useAnimatedRef() {
 
   return ref.current
 }
-


karol-bisztyga:
@kmagiera I have mentioned this problem before. I don't think this is the best design because it causes the callback passed to the hook to be called on the JS thread. Plus it is not necessary as the mapper launches that callback right away anyway. Apparently this caused the code from linked issue to not work(thus could be considered error prone). I don't see a scenario when getting rid of that initial updater breaks anything. If there is one, let's see it. Marked as a draft as there might be some further changes required to increase gracefulness of handling the case.

karol-bisztyga:
With that being said I understand that the author of that issue came up with some kind of workaround, however I think we should aim at handling as much ways of writing the code as possible.

kmagiera:
And example is when you have any layout related prop placed in the animated style.

Layout props cannot be immediately set on the UI thread because layout runs on a separate thread. If we skip that layout prop here the initial layout is going to be calculated as if the property wasn't there, and only after a frame of delay we'd get a proper layout updated. This will cause view flickering right after mounting.

kmagiera:
I don't see any issue linked in the description. There is just a reverse link in #1062 so I'm assuming you are talking about that issue then? If so I don't see how removing initial run fixes that problem. The problem in #1062 is that animation helpers don't yet support types other than numbers. It appears that we are able to process the rotation param internally, but this is inconsistent with the way react-native expects that parameters to be defined. I'd prefer to keep it consistent and if react-native does not allow numbers for rotation we shouldn't too

karol-bisztyga:
I understand. I still think this initial call could be removed(talking about that frame delay). As I get it we are talking about something like this:
```
const Comp = () => {
	const widthSv = useSharedValue(200)
	const uas = useAnimatedStyle=(() => {
		return {
			width: widthSV.value
		}
	})
	return (
		<Animated.View style={[{ width:10, height: 10, backgroundColor: 'red' }, uas]} />
	)
}
```
So here the problem is here will be that frame-delay between the initial value of `10` and the value of `200` which would be set later on the UI thread.

I just think the users should be aware of that and such code would be some kind of bad practice which should be avoided. So it would require specifying some pre defined default styles manually but there would be only one place where the styles from the callback would be set.

Johan-dutoit:
@kmagiera to clarify #1062 a bit more (if it helps with this or not, I don't know), is that I couldn't use the result of `withTiming` directly to do the rotation i.e. `${withTiming(45)}deg`.  This could be incorrect usage, maybe it should be `${withTiming(45).value}deg`, which I hadn't tried as the return type for `withTiming` is number.

What did eventually work, was creating a new derived value that returned the `withTiming` result i.e. `${derivedValue.value}deg`

terrysahaidak:
> @kmagiera to clarify #1062 a bit more (if it helps with this or not, I don't know), is that I couldn't use the result of `withTiming` directly to do the rotation i.e. `${withTiming(45)}deg`.  This could be incorrect usage, maybe it should be `${withTiming(45).value}deg`, which I hadn't tried as the return type for `withTiming` is number.
> 
> What did eventually work, was creating a new derived value that returned the `withTiming` result i.e. `${derivedValue.value}deg`

Both cases won't work because there will be no way to tell the core to which property the animation is assigned.

PR #1089, opened 2020-08-07
Fix events dispatching for TapGestureHandler

osdnk:
## Description

I observed that `onStart` is called only once and then is not being invoked again. Following my understanding of RNGH's state machine, `TapGestureHandler` finishes gesture with `END` state and is not coming back to `UNDETERMINED`. Also, failing can happen not only from `ACTIVE`, but also `BEGAN` state.

## Changes

- Handle case when `oldState` is `END`.
- Handle case when `oldState` is `BEGAN` and the new one is `FAILED`.

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -335,7 +335,7 @@ export function useAnimatedGestureHandler(handlers) {
       const ACTIVE = 4;
       const END = 5;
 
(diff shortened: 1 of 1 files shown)


terrysahaidak:
Seems like this is a duplication of https://github.com/software-mansion/react-native-reanimated/pull/878

osdnk:
Looks better than mine!
-- 
Pozdrawiam
Michał Osadnik

PR #1091, opened 2020-08-07
Improve tests

karol-bisztyga:
## Description

Added test for `scrollTo`
Added test-suite for `processColor`
Did a little clean up in the `test-suite`.

### DIFF
--- a/Example/test-suite/tests/Colors.js
+++ b/Example/test-suite/tests/Colors.js
@@ -1,0 +1,33 @@
+'use strict';
+import { processColor } from 'react-native-reanimated';
+
+export const name = 'Colors';
+
+export function test(t) {
+  t.describe('colors', () => {
+    // test basic colors
+    t.it('process red color', () => {
+      const padding = 2;
+      // test red
+      const redStr = processColor('red')
+        .toString(16)
+        .substring(padding + 0, padding + 2);
+      const redNum = Number.parseInt(redStr, 16);
+      t.expect(redNum).not.toBe(0);
+
+      // test green
+      const greenStr = processColor('green')
+        .toString(16)
+        .substring(padding + 2, padding + 4);
+      const greenNum = Number.parseInt(greenStr, 16);
+      t.expect(greenNum).not.toBe(0);
+
+      // test blue
+      const blueStr = processColor('blue')
+        .toString(16)
+        .substring(padding + 4, padding + 6);
+      const blueNum = Number.parseInt(blueStr, 16);
+      t.expect(blueNum).not.toBe(0);
+    });
+  });
+}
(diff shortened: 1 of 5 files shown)


Example/test-suite/tests/Colors.js
```diff
@@ -0,0 +1,33 @@
...
+  t.describe('colors', () => {
+    // test basic colors
+    t.it('process red color', () => {
+      const padding = 2;
+      // test red
+      const redStr = processColor('red')
+        .toString(16)
+        .substring(padding + 0, padding + 2);
+      const redNum = Number.parseInt(redStr, 16);
+      t.expect(redNum).not.toBe(0);
```
Szymon20000, line 16: Can we check exactly if it's the correct value?

karol-bisztyga: fixed 7a5999d

Example/test-suite/tests/Colors.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+import { processColor } from 'react-native-reanimated';
+
+export const name = 'Colors';
+
+export function test(t) {
+  t.describe('colors', () => {
+    t.it('process basic colors', () => {
+      const testData = [
+        ['red', 'ff0000'],
```
Szymon20000, line 10: Does this work? I guess the values should be different for iOS and Android.

karol-bisztyga: cdf71f0

karol-bisztyga:
The color test could be also run on UI thread

Szymon20000:
> The color test could be also run on UI thread

It would be nice to test is on both UI and JS thread. To test it on UI thread we could pass already computed values from JSThread

karol-bisztyga:
> > The color test could be also run on UI thread
> 
> It would be nice to test is on both UI and JS thread. To test it on UI thread we could pass already computed values from JSThread

It is now done on both UI and JS

Example/test-suite/tests/Colors.js
```diff
@@ -0,0 +1,54 @@
...
+        'cyan',
+        'orange',
+        'yellow',
+        'purple',
+      ];
+
+      const uiResults = {};
+      testData.forEach((color) => {
+        // run on JS thread
+        t.expect(processColor(color)).toBe(reaProcessColor(color));
```
Szymon20000, line 29: Does it make a any difference if we swap processColor with reaProcessColor?

Szymon20000 (approved):

Example/test-suite/tests/Colors.js
karol-bisztyga: Not at all. The only difference would probably be an order of how they're printed once the test fails, like `test failed: got aaa, expected bbb`

PR #1092, opened 2020-08-07
Add TS type for `scrollTo()`

wcandillon:


react-native-reanimated.d.ts
```diff
@@ -476,6 +476,7 @@ declare module 'react-native-reanimated' {
       pageX: number;
       pageY: number;
     };
+    export function scrollTo(ref: RefOject<ReactNativeScrollView>, x: number, y: number, animated: boolean);
```
jakub-gonet, line 479: I believe you have a typo here `RefOject` -> `RefObject`

wcandillon: Thks a lot, that helped me fixed another issue with it. It seems to work great now.

react-native-reanimated.d.ts
```diff
@@ -477,6 +477,8 @@ declare module 'react-native-reanimated' {
       pageY: number;
     };
 
+    export function scrollTo(ref: RefObject<ReactNativeScrollView | ScrollView>, x: number, y: number, animated: boolean);
```
jakub-gonet, line 480: Could you please specify return type as well?

```
node_modules/react-native-reanimated/react-native-reanimated.d.ts:479:21 - error TS7010: 'scrollTo', which lacks return-type annotation, implicitly has an 'any' return type.

479     export function scrollTo(ref: RefObject<ReactNativeScrollView | ScrollView>, x: number, y: number, animated: boolean);
                        ~~~~~~~~
```

react-native-reanimated.d.ts
wcandillon: done, thank you :)

Issue #1090, opened 2020-08-07
Reanimated2: exception in useAnimatedStyle


farwayer:
## Description

Exception inside `useAnimatedStyle` function while trying to call `Object.values` on `updater._closure`:
```js
 TypeError: Cannot convert undefined value to object
```

### Screenshots

![Grimshot 2020-08-07 14-56-42](https://user-images.githubusercontent.com/2343101/89643582-bfb64200-d8be-11ea-9640-f00bc305098e.png)

## Package versions

- Platform: Android (not tested on iOS)
- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.5
- Hermes: 0.5.2-rc1
- Babel config: 
```js
module.exports = {
  plugins: [
    ['@babel/plugin-proposal-decorators', {legacy: true}],
    '@babel/plugin-proposal-export-namespace-from',
    ['babel-plugin-module-resolver', {root: 'src'}],
    'react-native-reanimated/plugin',
  ],
  presets: ['module:metro-react-native-babel-preset'],
}
```

karol-bisztyga:
Hey, thanks for reporting, could you please provide a code snippet which causes the error?

farwayer:
Nothing special, just calling `useAnimatedStyle`:
```js
useAnimationStyle(() => ({
  transform: [...],
}))
```
Look like it is more environment issue.

terrysahaidak:
Did you try to run metro with reset cache?

farwayer:
Thanks! Look like the problem really was with metro cache.

terrysahaidak:
@karol-bisztyga @jakub-gonet worth adding this specific case where it throws on "`something._clousure` of undefined" to throubleshooting section in the docs

jakub-gonet:
it's only connected with cache or something else as well?

terrysahaidak:
I saw this like 5 times already. All the fixes were cache.

NateAGeek:
I know web support is still technically in a different release, but I am experiencing this in V2 with TypeScript...

terrysahaidak:
What exactly?

NateAGeek:
Sorry for the lack of details. However, I'm using React Native Web and attempting to use Reanimated 2. It appears with my current configuration I am getting the following errors.

```×
TypeError: Cannot convert undefined or null to object
useAnimatedStyle
node_modules/react-native-reanimated/src/reanimated2/Hooks.js:270
[...]

PR #1094, opened 2020-08-08
Add troubleshooting section to docs

jakub-gonet:
## Description

Added troubleshooting page to docs in order to create a list of common problems with Reanimated.

Suggested by @terrysahaidak in [this comment](https://github.com/software-mansion/react-native-reanimated/issues/1090#issuecomment-670948991)

### DIFF
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -1,0 +1,14 @@
+---
+id: troubleshooting
+title: Troubleshooting common problems
+sidebar_label: Troubleshooting
+---
+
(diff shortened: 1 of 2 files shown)


terrysahaidak:
Also lots of people actually forget to add babel plugin too.

jakub-gonet:
Could you please provide some examples of this error? I'd like to create a mapping between error message - solution and I didn't use v2 that much 😄

terrysahaidak:
I think it was this same error. Since babel plugin creates these closures. But there were more. I'll try and report back :)

jkadamczyk (approved):
Nice addition!

PR #1097, opened 2020-08-09
Decouple Easing and EasingNode in types

wcandillon:
Not sure how this happened. Easing is exported in Reanimated 1 no?


jakub-gonet:
It is, it was renamed to `EasingNode`

References:
[Types export](https://github.com/software-mansion/react-native-reanimated/blob/master/react-native-reanimated.d.ts#L542)
[Docs mention](https://docs.swmansion.com/react-native-reanimated/docs/next/migration#2-easing-renamed-to-easingnode)

wcandillon:
but in v2, Easing is still exported (same difference than interpolate and
interpolateNode) and we should still export it in the types. Should I
adjust it to make it a separate type definition maybe?

jakub-gonet:
Oh, I see now.

Seems like `Easing` is not typed right now, it was earlier?

wcandillon:
yes it was earlier

jakub-gonet:
`Easing` wasn't added in [aa0b739](https://github.com/software-mansion/react-native-reanimated/commit/aa0b7394d9f9003b5e85c711e549b98b57ed20b3#diff-c3ef4cab7a703f83e079356d6873fef5) and I renamed `Easing` to `EasingNode` in [939b5b5](https://github.com/software-mansion/react-native-reanimated/commit/939b5b5816ef2a6c2c297e941735015ba1bbfc27#).

If those types are compatible we can add another export for it.

wcandillon:
They seem compatible because `Adaptable<number>` accepts number but would you like me to make a dedicated v2 `Easing` type?

jakub-gonet:
Yeah, I think it's better to decouple v1 from v2 as much as possible and `Adaptable<T>` is a bit too broad type if you ask me 😄

wcandillon:
@jakub-gonet I think that you will be please with this new iteration, it works great in my project ;-)

PR #538, opened 2020-01-07
Support translate array syntax, enables translateZ

msand:
```jsx
    <Animated.View
      style={{
        color: 'green',
        transform: [
          {rotateY: '30deg'},
          Platform.OS === 'web'
            ? {translateZ: 100}
            : {translate: [0, 0, 100]},
        ],
      }}
    />
```

PR #1104, opened 2020-08-10
Fix non-animated properties in `transform`

jakub-gonet:
## Description

After changing original nested ternary in #538 to if-else chain double case didn't have an assignment, so it was crashing with a null value for transform style error.

## Example

```jsx
import React from 'react';
import {SafeAreaView, Text} from 'react-native';

const App = () => {
  return (
    <SafeAreaView>
      <Animated.View
        style={{
          transform: [{translateX: 100.2}],
        }}>
        <Text>Test </Text>
      </Animated.View>
    </SafeAreaView>
  );
};

export default App;
```

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/nodes/TransformNode.java
+++ b/android/src/main/java/com/swmansion/reanimated/nodes/TransformNode.java
@@ -56,7 +56,7 @@ public class TransformNode extends Node {
         } else if(type == ReadableType.Array) {
           transformConfig.value = transformConfigMap.getArray("value");
         } else {
-          transformConfigMap.getDouble("value");
+          transformConfig.value = transformConfigMap.getDouble("value");
         }
         configs.add(transformConfig);
       }


mikehardy:
Nice! Confirmed this patch works

I'm consuming it via patch-package like so

```

diff --git a/node_modules/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/nodes/TransformNode.java b/node_modules/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/nodes/TransformNode.java
index 8f6f8d1..7b653cd 100644
--- a/node_modules/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/nodes/TransformNode.java
+++ b/node_modules/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/nodes/TransformNode.java
@@ -56,7 +56,7 @@ public class TransformNode extends Node {
         } else if(type == ReadableType.Array) {
           transformConfig.value = transformConfigMap.getArray("value");
         } else {
-          transformConfigMap.getDouble("value");
+          transformConfig.value = transformConfigMap.getDouble("value");
         }
         configs.add(transformConfig);
       }

```

jakub-gonet:
Cool, I'll merge it then. Sorry for the inconvenience!

jakub-gonet:
Version 1.12.0 has been released.

PR #1113, opened 2020-08-12
Defend against nullish value

cltsang:
## Description

On v2, it crashes when called by [TabBar](https://github.com/satya164/react-native-tab-view/blob/master/src/TabBar.tsx).
I am not sure if it's caused by TabBar itself, but this simple edit prevents the crash.

### Example

A null value in backgroundColor props in style object of a component created with `Animated.createAnimatedComponent` will crash the app.

``` js
import {TextInput} from 'react-native';
import React from 'react';

export default function App(props) {
  return (
    <AnimatedTextInput
      style={{ backgroundColor: null }}
    />
  );
};
```


jakub-gonet:
Could you please give some small code example of crashing code?

src/createAnimatedComponent.js
```diff
@@ -30,7 +30,7 @@ function hasAnimatedNodes(value) {
   if (Array.isArray(value)) {
     return value.some(item => hasAnimatedNodes(item));
   }
-  if (typeof value === 'object') {
+  if (typeof value === 'object' && value) {
```
jakub-gonet, line 33: I'd suggested changing the order of them, following code uses established idiom for null checking and reads more naturally

```suggestion
  if (value && typeof value === 'object') {
```

cltsang:
> Looks good, would appreciate an example which is fixed by this change (creating any animated component and setting any property/style to `null` would be sufficient).

I edited the original post to all a code block as an example.
If that's not the preferred place to place it, please advise. I'll gladly follow.
Thanks.

PR #1130, opened 2020-08-16
Fix `overshootClamping` in `withSpring`

wcandillon:
I cannot get overshootClamping to work in v2 and it seems to be because of this condition no? I couldn't make sense of it from a logical perspective. My apologies with I missed something.

### DIFF
--- a/src/reanimated2/animations.js
+++ b/src/reanimated2/animations.js
@@ -174,7 +174,9 @@ export function withSpring(toValue, userConfig, callback) {
 
       const isOvershooting = () => {
         if (config.overshootClamping && config.stiffness !== 0) {
-          return current < toValue ? animation.current > toValue : animation.current < toValue;
+          return current < toValue
+            ? animation.current > toValue
+            : animation.current < toValue;
         } else {
           return false;
         }
--- a/src/reanimated2/animations.js
+++ b/src/reanimated2/animations.js
@@ -174,7 +174,7 @@ export function withSpring(toValue, userConfig, callback) {
 
       const isOvershooting = () => {
         if (config.overshootClamping && config.stiffness !== 0) {
-          return current > toValue;
+          return current < toValue ? animation.current > toValue : animation.current < toValue;
         } else {
           return false;
         }
--- a/src/reanimated2/animations.js
+++ b/src/reanimated2/animations.js
@@ -174,7 +174,7 @@ export function withSpring(toValue, userConfig, callback) {
 
       const isOvershooting = () => {
         if (config.overshootClamping && config.stiffness !== 0) {
-          return current < toValue ? current > toValue : current < toValue;
+          return current > toValue;
         } else {
           return false;
         }


src/reanimated2/animations.js
```diff
@@ -174,7 +174,7 @@ export function withSpring(toValue, userConfig, callback) {
 
       const isOvershooting = () => {
         if (config.overshootClamping && config.stiffness !== 0) {
-          return current < toValue ? current > toValue : current < toValue;
```
wcandillon, line 177: I read this to return false all the time, am I correct?

src/reanimated2/animations.js
```diff
@@ -174,7 +174,7 @@ export function withSpring(toValue, userConfig, callback) {
...
+          return current > toValue;
```
terrysahaidak, line 177: What if we use negative velocity? Then we might have current ad positive value but toValue might be negative one.

Then current will immediately be bigger than toValue. Check out clamp of withDecay.

wcandillon: Thank you for the explanation! Is the overshootClamping option working for you? I couldn't get it to work in an example. I still struggling to understand how line 177 work.

In my example toValue is 0, let's say current is minus -1: isOvershooting returns false (current < toValue  === true and then current > toValue  === false).

terrysahaidak: I'm not sure it's working right now, can't test right away, but I mean your change wrong here too :)

But I still can't understand how to handle the clamping without velocity. Because velocity is the only thing what can tell the direction of animation.

wcandillon: I will do more testing and maybe try to implement it the way it's done in decay and keep you posted. Thank you for your support on this.

src/reanimated2/animations.js
jakub-gonet: Yup, A < B and A >(=) B are mutually exclusive, so if `current < toValue` is false you return same expression (so false) and if it's true you return inverted expression (which is false in that case)

kmagiera:
There is in fact a mistake here. I translated it from the reanimated1 code where the code look as follows:

```js
  const isOvershooting = cond(
    and(config.overshootClamping, neq(config.stiffness, 0)),
    cond(
      lessThan(prevPosition, config.toValue),
      greaterThan(state.position, config.toValue),
      lessThan(state.position, config.toValue)
    )
  );
```

So the second and third parts of the ternary should use updated position and not `current` which in this place represents the previous position

I think what it should be there instead is this:
```js
return current < toValue ? animation.current > toValue : animation.current < toValue;
```

Mind checking the above code and updating your PR?

wcandillon:
@kmagiera I tried the proposed fix and it works great thank you :) @jakub-gonet I updated the PR accordingly.

src/reanimated2/animations.js
```diff
@@ -174,7 +174,7 @@ export function withSpring(toValue, userConfig, callback) {
 
       const isOvershooting = () => {
         if (config.overshootClamping && config.stiffness !== 0) {
-          return current < toValue ? current > toValue : current < toValue;
+          return current < toValue ? animation.current > toValue : animation.current < toValue;
```
jakub-gonet, line 177:
```suggestion
          return current < toValue
            ? animation.current > toValue
            : animation.current < toValue;
```

Could you format it accordingly?

jakub-gonet (approved):
Works great, I left a small nit.

src/reanimated2/animations.js
wcandillon: done, thks :)

jakub-gonet (approved):
🎉

PR #1141, opened 2020-08-19
[v1] Add `dependency` prop to `Animated.Code` type

jinshin1013:
## Description
This PR adds `dependencies` prop to the `Animated.Code` component prop type which was introduced in 212f80d587c9f719768797792a146fe957e289ce.

## Changes

- Added `dependencies` prop to `<Animated.Code />` component type.


jakub-gonet (approved):
Thank you!

PR #1150, opened 2020-08-23
Add callback parameter to repeat()

wcandillon:
I use this extra parameter so that when I'm "pausing" the loop, I can remember in which direction it was going. The direction is updated at the end of each iteration via callback().

```ts
repeat(
    withTiming(1 - dest.value, {
        duration: 1000,
         easing,
     }),
     -1,
      true,
      () => (dest.value = 1 - dest.value)
)
```

### DIFF
--- a/src/reanimated2/animations.js
+++ b/src/reanimated2/animations.js
@@ -394,7 +394,7 @@ export function repeat(_nextAnimation, numberOfReps = 2, reverse = false, callba
       animation.current = nextAnimation.current;
       if (finished) {
         animation.reps += 1;
-        callback();
+        callback(animation.current);
         if (numberOfReps > 0 && animation.reps >= numberOfReps) {
           return true;
         }
--- a/src/reanimated2/animations.js
+++ b/src/reanimated2/animations.js
@@ -381,7 +381,7 @@ export function sequence(..._animations) {
   });
 }
 
-export function repeat(_nextAnimation, numberOfReps = 2, reverse = false) {
+export function repeat(_nextAnimation, numberOfReps = 2, reverse = false, callback = () => {}) {
   'worklet';
   return defineAnimation(_nextAnimation, () => {
     'worklet';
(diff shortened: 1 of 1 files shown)


src/reanimated2/animations.js
```diff
@@ -394,6 +394,7 @@ export function repeat(_nextAnimation, numberOfReps = 2, reverse = false) {
       animation.current = nextAnimation.current;
       if (finished) {
         animation.reps += 1;
+        callback();
```
karol-bisztyga, line 397: I think it could be good to pass `animation.current` to the callback. Sometimes reading sv value inside the callback does not give exact results. What do you think?

wcandillon: What does sv stand for? For course, I wouldn't mind do this change.

karol-bisztyga: sv=shared value, sorry for using such shortcuts

karol-bisztyga: my point is let's say you have an example where you alternate between 100 and 200. With short periods if you'd like to use the current value somehow in the callbacks it could be wrong:
```
const mySV = useSharedValue(100) // this is gonna alternate between 100 and 200
mySV.value = repeat(/*...*/, () => { const flag = (mySV === 200); /* do something with the flag */ /* and this is going to sometimes give you a result of 200 but sometimes something close but not exact, like 198.34 */ })
```
instead
```
const mySV = useSharedValue(100) // this is gonna alternate between 100 and 200
mySV.value = repeat(/*...*/, (currentValue) => { const flag = (currentValue === 200); /* do something with the flag */ /* this is most likely going to be exact */ })
```

wcandillon: I added the parameter to the callback.

wcandillon: What does `defineAnimation` do exactly? And do we need it for every custom animations we build?

karol-bisztyga: it distinguishes style updates on JS/UI I think it should be used in order to prevent flickering

karol-bisztyga (approved):

src/reanimated2/animations.js
wcandillon: Should it be exported for people who build custom animations?

karol-bisztyga: Maybe but I would consider exposing more easy-going tool for that

Issue #1057, opened 2020-07-31
No animation end callback when using delay()


lachtos:
## Description

When using `withSpring` with `delay` like `delay(100, withSpring(scale.value, {}, finished => {}))` the callback does not get called, meanwhile `withSpring(scale.value, {}, finished => {})` works fine.

## Steps To Reproduce

1. Use `withSpring` with `delay`

### Expected behavior

On animation end callback gets called

### Actual behavior

On animation end callback does not get called

## Snack or minimal code example

```
const scale = useSharedValue(0);

const animatedStyle = useAnimatedStyle(() => {
  const scl = delay(100, withSpring(scale.value, { damping: 15 }, finished => {
   console.log(`Animation ended; finished: ${finished}`); // <-- does not log when animated with delay
  }));

  return {
    transform: [{ scale: scl }]
  };
});

useEffect(() => {
  scale.value = 1;
}, []);
```

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.4

MinimogDev:
Also happens when using `withTiming` and delay

PR #1153, opened 2020-08-24
Add callbacks for animations

karol-bisztyga:
## Description

Add callbacks for `delay`, `sequence` and `repeat`.
- for `delay` the callback of the inner animation's now going to be called after that animation ends(finished or being cancelled)
- for `sequence` there are two cases:
  - the sequence finishes - after every animation passed to it that animation's callback is being called
  - the sequence is being cancelled - for those animations which have not been called yet due to cancellation the callbacks are being called(with `finish = false`)
- in repeat, the callback passed to the inner animation will be called on every repeat. The first argument tells us whether the animation finished or was cancelled. The second one hold animation's current value(when the animation has been cancelled it is `undefined`). The callback passed to `repeat` is called in the end when animation is finished or cancelled with `finished` set accordingly.

This PR fixes [this issue](https://github.com/software-mansion/react-native-reanimated/issues/1057)

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -109,7 +109,6 @@ function runAnimations(animation, timestamp, key, result) {
       animation.timestamp = timestamp;
       if (finished) {
         animation.finished = true;
-        // TODO: also trigger callback when animation is cancelled (overwritten)
         animation.callback && animation.callback(true /* finished */);
       }
       result[key] = animation.current;
@@ -476,4 +475,3 @@ export function useAnimatedRef() {
 
   return ref.current
 }
-
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -475,3 +475,4 @@ export function useAnimatedRef() {
 
   return ref.current
 }
+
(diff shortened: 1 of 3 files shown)


src/reanimated2/Hooks.js
```diff
@@ -109,7 +109,6 @@ function runAnimations(animation, timestamp, key, result) {
       animation.timestamp = timestamp;
       if (finished) {
         animation.finished = true;
-        // TODO: also trigger callback when animation is cancelled (overwritten)
```
Szymon20000, line 112: Has it been resolved?

Szymon20000:
Looks fine to me.  I still have one question though, so left a comment.

src/reanimated2/Hooks.js
karol-bisztyga: In fact I checked all the animations and that does not work for all of them, good point, reverting

Szymon20000 (approved):

src/reanimated2/Hooks.js
karol-bisztyga: ok so I took a deeper look and it does not seem to be a case anymore. Triggering a callback on animation cancel works for:
- withTiming
- withSpring
- withDecay
- delay
- sequence
- repeat
so I think I covered all the animations for now

karol-bisztyga:
@wcandillon just wanted to let you know that if this is merged `repeat` will receive one more callback. Had to distinguish them in order to have a callback for end of whole repeat and also every step(your idea). Now it will be 5th not 4th argument of that function.

Szymon20000 (approved):

PR #1157, opened 2020-08-25
Mark `WithTimingConfig` TS type fields as optional

mczernek:
## Description

According to docs, both those values have defaults so they should be optional.

## Changes

Only changed types in types definition, no executable code is changed.

Please refer to [documentation](https://docs.swmansion.com/react-native-reanimated/docs/next/api/withTiming)


jakub-gonet (approved):
Thanks! 🎉

PR #1159, opened 2020-08-26
Remove redundant call on Android

karol-bisztyga:
## Description

`NativeProxy.prepare`(followed by `installJSIBindings`) is called twice one call after another. First call is in `NativeProxy` constructor and the other one is in `NodesManager.initWithContext` right after the `NativeProxy` object is created.

This PR removes one of the calls.


Szymon20000 (approved):

karol-bisztyga:
## Description

`NativeProxy.prepare`(followed by `installJSIBindings`) is called twice one call after another. First call is in `NativeProxy` constructor and the other one is in `NodesManager.initWithContext` right after the `NativeProxy` object is created.

This removes one of the calls.

Issue #1155, opened 2020-08-25
How to use setState in worklet


mohammadanaraki:
## Description

how to change state or same in reanimated v2 events 

## Code
````js
  const x = useSharedValue(0);
  const [state , setState] = React.useState(false);
  const gestureHandler = useAnimatedGestureHandler({
    onStart: (_, ctx) => {
      ctx.startX = x.value;
    },
    onActive: (event, ctx) => {
      x.value = ctx.startX + event.translationX;
    },
    onEnd: _ => {
      x.value = withSpring(0);
      setState(true); // <- how to use this here 
    },
  });
````
### Package versions

- React:
- React Native:
- React Native Reanimated: 2.0.0-alpha.5

**Comment by terrysahaidak** [comment]
Could you post more code? The code you provided works as expected on alpha.5.

**Comment by karol-bisztyga** [comment]
No @terrysahaidak it does not work properly, it does not read a proper value of a state on UI thread. Try with the following code and you'll see: `setState(!state); // <- how to use this here`

**Comment by terrysahaidak** [comment]
Yeah @karol-bisztyga, as well as `setState(v => !v)` I couldn't get to work. But I mean the provided code still works as expected :)

**Comment by mohammadanaraki** [comment]
> Yeah @karol-bisztyga, as well as `setState(v => !v)` I couldn't get to work. But I mean the provided code still works as expected :)

Yes ! :) 
i just made an example 
i meant can't access xD

**Comment by karol-bisztyga** [comment]
fixed in https://github.com/software-mansion/react-native-reanimated/pull/1165

**Comment by karol-bisztyga** [comment]
Fixed in https://github.com/software-mansion/react-native-reanimated/pull/1165
Now this code works as expected
```
import React from 'react';
import { TapGestureHandler } from 'react-native-gesture-handler';
import { View } from 'react-native';
import Animated, {
  useSharedValue,
  useAnimatedGestureHandler,
  withSpring,
} from 'react-native-reanimated';

export default function App() {
  const x = useSharedValue(0);
  const [state, setState] = React.useState(false);

  console.log('rerender ' + state);

  const handler = useAnimatedGestureHandler({
    onStart: (_, ctx) => {
      ctx.startX = x.value;
    },
    onActive: (event, ctx) => {
      x.value = ctx.startX + event.translationX;
    },
    onEnd: (_) => {
      x.value = withSpring(0);
      setState(!state); // <- how to use this here
    },
  });

  return (
    <View
      style={{
[...]

PR #1165, opened 2020-08-28
Fast refresh for gestures

karol-bisztyga:
## Motivation

Motivated by [this issue](https://github.com/software-mansion/react-native-reanimated/issues/1155) this PR has been created.

## Description

To keep the hooks consistent `useAnimatedGestureHandler` is about to rebuild worklets on certain changes in the same manner as `useAnimatedStyle` or `useDerivedValue`. The decision whether to rebuild or not will depend on the dependency array which may be passed to the hook or not. If it is passed it is read directly from the argument, if not, it pulls the dependencies from the worklets' closure. Also in both cases the dependency array will be decorated with the worklets' hashes which means on every worklet's code change the rebuild will be triggered.

This is in fact a continuation of the works regarding fast refresh issues.

## Changes

`Hooks.js` - build dependencies for `useAnimatedGestureHandler` and rebuild worklet in `useEvent`
`createAnimatedComponent.js` - added logic of reattaching for event worklets
`FastRefreshTest.js` - added tests for the `useAnimatedGestureHandler` hook.

---------

## Deprecated

The idea of the solution for the linked issue changed, but keeping the old one as there were some comments regarding it.

### Motivation

Motivated by [this issue](https://github.com/software-mansion/react-native-reanimated/issues/1155) this PR has been created. Not sure yet if the idea itself is appropriate and if there are real life use cases for this thus creating a draft PR.

### Description

The goal is to provide a functionality which would allow to use React's `useState` state on both JS and UI threads. Now we can set state from the UI thread and, what comes after that, trigger component's rerender from there. 

Bear in mind that this is just POC, not only this may never be merged but also the API might change.

### Changes

Added a hook in `Hooks.js` and a test to show how it works


likern:
@karol-bisztyga Am I correct it's about to notify React that something has changes and we need to rerender?

So right now it looks like it's not possible to do that? Can we just set `sharedValue.value = 10` inside worklets and read that value on JS side?

terrysahaidak:
of course we can, but that won't trigger any rerenders so no useEffect or other stuff which depends on rerenders will be executed.

karol-bisztyga:
PR updated

likern:
I think I need this patch. 
My task is to **render the same component multiple times as fast as possible** and **for each time measure component layout**.
If I try to do that using `useEffect` and `onLayout` effect there are problems:
1. `onLayout` is called after `useEffect`
2. If update component state inside `useEffect` `onLayout` might be called later.

So generally I expected that I have `useEffect -> onLayout -> useEffect` callbacks order. But if I rerun component too fast (and I need exactly that) - `onLayout` is delayed and I get `useEffect -> useEffect ->  onLayout`. So it looks like there is no order between `useEffect` and `onLayout`.

```tsx
const [index, setIndex] = useState(0);
...
function workl() {
  'worklet';
  console.log(`worklet: viewRef is ${viewRef}`);
  console.log(
    `worklet: viewRef measure is ${JSON.stringify(measure(viewRef))}`
  );
}
...
useEffect(() => {
  console.log(`useEffect: before runOnUI`);
  runOnUI(workl)();
  console.log(`useEffect: after runOnUI`);

  const nextIndex = index + 1;
  if (nextIndex <= durations.length) {
    setTimeout(() => {
      setIndex(nextIndex);
    }, 0);
  }
}, [index,);
```

Using Reanimated now I need to coordinate / sync `useEffect` and `worklet` where I measure component - so next `useEffect` / component rerender is not happening until I measure current one.

I think using `useAnimatedState` inside `worklet` is a way to go.

### Why
Why I need to rerender component as fast as possible and syncronous way?
List and List scrolling are the most important and most common tasks in pretty any mobile application.

React Native lacks any decent solution / implementation to use ScrollView / FlatList. They are pretty slow and it's seen visually.
Even though I use only optimised mobile local data - inside sqlite with indexes. Stilll rendering is not good.

The **only** way to overcome this is to use https://github.com/Flipkart/recyclerlistview. Which on first measures is super fast and uses View recycling.
But it requires function to estimate component (list cell) layout based on data.

Unfortunately it's not possible to do that in most cases. **Estimating text layout based on text string is extremely difficult task for flex layout**. It depends on text case, text font, ligatures, etc. It's only possible just to render component and measure it's layout and then to store that somewhere (in my case in sqlite database).

When to render recyclerlistview we use that layout information for infinite scrolling and view recycling.

likern:
@karol-bisztyga Hello! I tried your original patch with `useAnimatedStated`. It turns out it is the only way to achieve my goal - prerender component and measure it's layouts.

```ts
export const RunLayoutMeasure = () => {
  const TextComponent = (props: TextComponentProps) => (
    <Text>{props.message}</Text>
  );
  const propsList: TextComponentProps[] = [
    { message: '1' },
    { message: '2' },
    { message: '3' },
    { message: '4' },
    { message: '5' }
  ];

  return (
    <MeasureLayout
      propsList={propsList}
      componentRef={TextComponent}
      onFinish={(layouts) => {
        console.log(
          `measures: ${layouts}, isArray? ${Array.isArray(layouts)}, length ${
            layouts.length
          }`
        );
      }}
    />
  );
};
```

Later these layouts will be used in https://github.com/Flipkart/recyclerlistview in `LayoutProvider`. Instead of guessing type and layout I will return direct measurements.
```tsx
        this._layoutProvider = new LayoutProvider(
            index => {
                if (index % 3 === 0) {
                    return ViewTypes.FULL;
                } else if (index % 3 === 1) {
                    return ViewTypes.HALF_LEFT;
                } else {
                    return ViewTypes.HALF_RIGHT;
                }
            },
            (type, dim) => {
                // Return direct measures
                // instead of this
                switch (type) {
                    case ViewTypes.HALF_LEFT:
                        dim.width = width / 2;
                        dim.height = 160;
                        break;
                    case ViewTypes.HALF_RIGHT:
                        dim.width = width / 2;
                        dim.height = 160;
                        break;
                    case ViewTypes.FULL:
                        dim.width = width;
                        dim.height = 140;
                        break;
                    default:
                        dim.width = 0;
                        dim.height = 0;
                }
            }
        );
```

But still I have a problem. I think this is a bug. Original `useAnimatedState` **doesn't work with arrays**. I can't update / modify array and see only initial array value.

My full source code is here https://gist.github.com/likern/9ea94b38fe51cfab358afa239281e989

I tried every possible way to achieve that, but without success.
If anyone could help I will highly appreciate.

Also I noticed that **sometimes worklets are not called** from `useEffect`.

src/reanimated2/Hooks.js
```diff
@@ -345,14 +346,51 @@ export function useDerivedValue(processor, dependencies = undefined) {
...
+    });
+  } else {
+    dependencies.push(buildWorkletsHash(handlers));
+  }
+
+  if (dependencies.length !== savedDependencies.length) {
+    dependenciesDiffer = true;
+  } else {
+    for (let i = 0; i < dependencies.length; ++i) {
+      if (dependencies[i] !== savedDependencies[i]) {
```
karol-bisztyga, line 387: this is probably not a good solution for checking if dependencies changed

src/reanimated2/Hooks.js
```diff
@@ -399,19 +437,49 @@ export function useAnimatedGestureHandler(handlers) {
...
```
karol-bisztyga, line 474: this is probably not a good solution for checking if dependencies changed

src/reanimated2/Hooks.js
terrysahaidak: I would use at least Object.is for that. React itself is using similar thing afaik

karol-bisztyga: Yes, I just wanted to mark it here so I can see that it should be done in a different way

karol-bisztyga:
@likern the first idea here has been abandoned for good for a couple of reasons. I think, if you have some problems, the best way to go is opening new issues. In that case every one of those problems can be approached separately and, in this PR, I think it would be better to focus on the current solution.

src/reanimated2/Hooks.js
```diff
@@ -412,6 +446,20 @@ export function useAnimatedScrollHandler(handlers) {
   }
   const { context } = initRef.current;
 
+  // build dependencies
```
karol-bisztyga, line 449: DRY alert

Szymon20000 (approved):
Looks fine to me. As we decided on slack don't use `useEffect` and instead use [this function](https://github.com/facebook/react-native/blob/21396bb38015e266bab944b3ab1e35db64686ea5/Libraries/Renderer/implementations/ReactFabric-prod.js#L3067).

Szymon20000 (approved):

PR #1166, opened 2020-08-28
docs: rename RETurboModuleProvider

RatebSeirawan:
REATurboModuleProvider couldn't be resolved so I went into the code and didn't find any REATurboModuleProvider header file but I found RETurboModuleProvider, I got it fixed on my installation steps after I renamed to RETurboModuleProvider.


jakub-gonet:
This is correct on the master branch, [this commit](https://github.com/software-mansion/react-native-reanimated/commit/b7cd896bacaf799b3fb679a7af4dfb1dafad2a83) and [this commit](https://github.com/software-mansion/react-native-reanimated/commit/f200fa60a2dfbb0a8d394b03c3194532255bea96) changed that. Alpha 6 with those changes should be released in the next few days.

Issue #1132, opened 2020-08-17
Capturing Animated (default export from react-native-reanimated) in worklet causes error loop


kmagiera:
## Description

When Animated object (and perhaps this may work for other complex objects too) is captured in a worklet, this results in some unexpected error/crash behavior

## Steps To Reproduce

Render the app with one of the worklets capturing Animated object (see examples posted below).

### Expected behavior

Properties under Animated access from within a worklet (e.g. Animated.Extrapolate.CLAMP) should return their actual values.

### Actual behavior

The app goes into an error loop with the following stacktrace being thrown every second or so:
```
This error is located at:
    in FlatListStuff
    in RCTView (at View.js:34)
    in View (at AppContainer.js:106)
    in RCTView (at View.js:34)
    in View (at AppContainer.js:132)
    in AppContainer (at renderApplication.js:39)
[Mon Aug 17 2020 10:30:17.100]  ERROR    Error: Exception in HostFunction: <unknown>
```

## Snack or minimal code example

```js
import Animated, { useAnimatedStyle } from 'react-native-reanimated';

function Something() {
  const something = useAnimatedStyle(() => {
    return { backgroundColor: Animated.Extrapolate.CLAMP };
  });
  return <Animated.View style={something}/>
}
```

Full example as prepared by @ryanbaldwin: https://gist.githubusercontent.com/ryanbaldwin/68a5fa2c35dbfadf3d3a3bab2e96590d/raw/c2c81c1c0b4fedd6b3e6e251ecf8c39bff210c42/App.js

luixo:
Hello,

I recently added `react-native-reanimated` (via `nativewind`) and stumbled upon exact same error whenever I used `Animated.View`.
Going back from `3.7.x` to `3.6.x` helped and I took some time to bisect the exact build, [3.7.0-nightly-20231221-ea5c667ce](https://www.npmjs.com/package/react-native-reanimated/v/3.7.0-nightly-20231221-ea5c667ce) is the first one to introduce the problem.
It basically narrows down the issue to this commit: https://github.com/software-mansion/react-native-reanimated/commit/5d1c25401268e114923effde5b1583557896a71c

Let me know if I can help you reproduce the issue.

My lib versions:
```ts
{
"react-native": "0.73.5",
"expo": "^50.0.7",
}
```

Issue #1051, opened 2020-07-31
v2 Dynamic property accessors don't work


wcandillon:
## Description

I noticed that it is not possible to access the property of an animation value dynamically.
See the code example below.

```tsx
import React from "react";
import { View, StyleSheet } from "react-native";
import Animated, {
  useSharedValue,
  useAnimatedStyle,
} from "react-native-reanimated";

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: "space-evenly",
  },
  box: {
    width: 100,
    height: 100,
    backgroundColor: "cyan",
  },
});

const Test = () => {
  const a1 = "bar";
  const val = useSharedValue({
    foo: 100,
    bar: 200,
  });
  // This works
  const style1 = useAnimatedStyle(() => ({
    // eslint-disable-next-line dot-notation
    transform: [{ translateX: val.value["bar"] }],
  }));
  // This doesn't work
  const style2 = useAnimatedStyle(() => ({
    transform: [{ translateX: val.value[a1] }],
  }));
  return (
    <View style={styles.container}>
      <Animated.View style={[styles.box, style1]} />
      <Animated.View style={[styles.box, style2]} />
    </View>
  );
};

export default Test;
```

karol-bisztyga:
Looks like an issue with closure. There is an easy workaound:
```
const style2 = useAnimatedStyle(() => {
  a1; // dummy instruction to fetch the a1 value
  return {
    transform: [{ translateX: val.value[a1] }],
  }
});
```

or just define the `a1` inside the hook:
```
const style2 = useAnimatedStyle(() => {
  const a1 = 'bar';
  return {
    transform: [{ translateX: val.value[a1] }],
  }
});
```

PR #1174, opened 2020-09-01
Capture only used values and convert symbols to strings.

Szymon20000:
## Why
Some people like to capture huge objects such as `Animated`. These objects may contain symbols that weren't handled in the past. The problem with symbols can be partially solved by converting symbols to strings. (it's better than nothing). 
Symbols cannot be converted to symbols and it can be seen in the following example: 
A->C, B->C, and C is a symbol. Then  after conversion A->C != B->C.  Another reason to merge the pr is to speed up Reanimated. Converting objects can take a lot of time. By capturing used values only we significantly reduce this time.

## How
I've added `ClosureGenerator` class that's responsible for capturing used paths only. It takes a path with the name of the variable that would be fully captured in the previous implementation and iterates upwards in the babel nodes tree until there is an edge case or the current node's type is different than 'MemberExpression'.  `ClosureGenerator` maintains a Trie tree where all used references are stored. 

### Assigment operator 
If somebody wants to add a new property to the captured value then obviously we don't want to capture it.
a.b.c.d = 5;
captures a.b.c;

### Mutable  
sth.value

### a.b.c[]
Sometimes we can't tell what property the programmer wants to get. In such a case we capture only `a.b.c`

### Methods that use `this`
The problems only occur with objects that have been sent between threads. However, in Reanimated we only allow for Object|Array|String|Number|Boolean|Function types so it's enough to blacklist functions from their prototypes. The blacklist was base on https://developer.mozilla.org/pl/docs/Web/JavaScript/Referencje/Obiekty/Array/prototype.

## Problems
Update2: 
The problem below has been solved by blacklisting Array|Object|String|Number|Boolean prototype methods. It's a correct solution because in Reanimated2 we do not allow other types to be sent between threads.
Update: 
Currently, there is a problem with methods that use 'this'. For Example:
```JS
sth.map(() =>)
```

_____________
How it works:
```JS
function A() {
  'worklet';
  Animated.withSpring(0);
  Animated.a.b.c = 3;
  const x = Animated.a.b;
  const y = Animated.c.d['szymon'];
  const w = Animated.e.r[getPropName()];
  const ee = Animated.c.r[x];
  const ww = Animated.wer.ty[klocek];
}
```
->
```JS
var A = function () {
    var _f = function _f() {
      _Animated.default.withSpring(0);

      _Animated.default.a.b.c = 3;
      var x = _Animated.default.a.b;
      var y = _Animated.default.c.d['szymon'];

      var w = _Animated.default.e.r[getPropName()];

      var ee = _Animated.default.c.r[x];
      var ww = _Animated.default.wer.ty[klocek];
    };

    _f._closure = {
      Animated: {
        withSpring: _Animated.default.withSpring,
        a: {
          b: _Animated.default.a.b
        },
        c: {
          d: _Animated.default.c.d,
          r: _Animated.default.c.r
        },
        e: {
          r: _Animated.default.e.r
        },
        wer: {
          ty: _Animated.default.wer.ty
        }
      },
      getPropName: getPropName,
      klocek: klocek
    };
    _f.asString = "function A(){const{Animated,getPropName,klocek}=jsThis._closure;{Animated.withSpring(0);Animated.a.b.c=3;var x=Animated.a.b;var y=Animated.c.d['szymon'];var w=Animated.e.r[getPropName()];var ee=Animated.c.r[x];var ww=Animated.wer.ty[klocek];}}";
    _f.__workletHash = 12002772052812;

    return _f;
  }();
```

fixes: https://github.com/software-mansion/react-native-reanimated/issues/1132
fixes: https://github.com/software-mansion/react-native-reanimated/issues/1051

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -68,12 +68,14 @@ class ClosureGenerator {
       return notFound;
     }
     const purePath = [memberExpressionNode.property.name];
+    console.log('findRec dla', purePath[0]);
     const node = memberExpressionNode;
     const upAns = this.findPrefixRec(path.parentPath);
     return this.mergeAns([purePath, node], upAns);
   }
 
   findPrefix(base, babelPath) {
+    console.log('find dla', base);
     const purePath = [base];
     const node = babelPath.node;
     const upAns = this.findPrefixRec(babelPath.parentPath);
@@ -81,7 +83,9 @@ class ClosureGenerator {
   }
 
   addPath(base, babelPath) {
+    console.log('dodaje dla', base);
     const [purePath, node] = this.findPrefix(base, babelPath);
+    console.log('po findPref', purePath);
     let parent = this.trie;
     let index = -1;
     for (const current of purePath) {
@@ -90,7 +94,7 @@ class ClosureGenerator {
         continue;
       }
       if (!parent[0][current]) {
-        parent[0][current] = [null, false];
+        parent[0][current] = [{}, false];
       }
       if (index === purePath.length - 1) {
         parent[0][current] = [node, true];
@@ -100,15 +104,16 @@ class ClosureGenerator {
   }
 
   generateNodeForBase(t, current, parent) {
-    const currentNode = parent[current];
+    console.log('gen dla', current);
+    const currentNode = parent[0][current];
     if (currentNode[1]) {
       return currentNode[0];
     }
     return t.objectExpression(
-      Object.keys(currentNode).map((propertyName) =>
+      Object.keys(currentNode[0]).map((propertyName) =>
         t.objectProperty(
           t.identifier(propertyName),
-          this.generateNodeForBase(t, propertyName, currentNode[0]),
+          this.generateNodeForBase(t, propertyName, currentNode),
           false,
           true
         )
@@ -116,12 +121,14 @@ class ClosureGenerator {
     );
   }
 
-  generate(t, variables, closureVariables) {
+  generate(t, variables, names) {
+    const arrayOfKeys = [...names];
+    console.log('names', names);
     return t.objectExpression(
       variables.map((variable, index) =>
         t.objectProperty(
           t.identifier(variable.name),
-          this.generateNodeForBase(t, closureVariables[index], this.trie),
+          this.generateNodeForBase(t, arrayOfKeys[index], this.trie),
           false,
           true
         )
@@ -274,7 +281,7 @@ function processWorkletFunction(t, fun) {
             t.identifier('_closure'),
             false
           ),
-          closureGenerator.generate()
+          closureGenerator.generate(t, variables, closure.keys())
         )
       ),
       t.expressionStatement(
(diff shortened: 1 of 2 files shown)


terrysahaidak:
I don't see the "szymon" prop in the result example you provided. Is it missing or example is outdated already?

Szymon20000:
Do you mean there is no 'szymon' in closure? That's because there is the [] operator.

terrysahaidak:
Hmm, how does it work with those values? Do you handle them in that pr?

Szymon20000:
Babel adds `computed` flag to a MemberExpression node if it changes [sth] to .sth

karol-bisztyga (approved):

plugin.js
```diff
@@ -45,6 +47,162 @@ const globals = new Set([
   '_scrollTo',
 ]);
 
+const blacklistedFunctions = new Set([
+  'stopCapturing',
```
karol-bisztyga, line 51: maybe we should leave a note about this

Szymon20000: What would be a proper place for that?

plugin.js
```diff
@@ -45,6 +47,162 @@ const globals = new Set([
...
+      return notFound;
+    }
+    if (
+      memberExpressionNode.computed ||
+      memberExpressionNode.property.name === 'value' ||
+      blacklistedFunctions.has(memberExpressionNode.property.name)
+    ) {
+      // a.b[w] -> a.b.w in babel nodes
+      // a.v.value
+      // sth.map(() => )
```
jakub-gonet, line 131: is this comment necessary?

plugin.js
```diff
@@ -45,6 +47,162 @@ const globals = new Set([
...
+      return notFound;
+    }
+    if (
+      path.parent &&
+      path.parent.type === 'AssignmentExpression' &&
+      path.parent.left === path.node
+    ) {
+      /// captured.newProp = 5;
```
jakub-gonet, line 139: is this comment necessary?

plugin.js
```diff
@@ -45,6 +47,162 @@ const globals = new Set([
   '_scrollTo',
 ]);
 
+const blacklistedFunctions = new Set([
```
jakub-gonet, line 50: Can you write a comment on how this list of blacklisted was created?

plugin.js
karol-bisztyga: line 49 could be ok, like explaining the constant

plugin.js
Szymon20000: It explains the main edge case in which we don't want to capture the prototypes methods like map for example.

plugin.js
Szymon20000: All those comments are just examples of edge cases.

plugin.js
Szymon20000: Answered in pr description.

Issue #1181, opened 2020-09-02
[iOS] Frequent crashed when using react-navigation-v5 and Reanimated v2


mczernek:
## Description

When using newest available versions of react-navigation (@react-navigation/native@5.7.3 and @react-navigation/stack@5.7.0) along with reanimated v2 (2.0.0-alpha.5), i see frequent crashes when navigating to screen with animations.

## Steps To Reproduce

1. Just run this app on iOS [repro repo](https://github.com/mczernek/reanimated-navigation-crash)
2. After several (usually it happens before 30 seconds passes) navigations, app crashes

App crashes with different stack traces, but usually it seems to be somehow related to deallocation.

### Expected behavior

App should not crash

## Snack or minimal code example

[repro repo](https://github.com/mczernek/reanimated-navigation-crash)

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.5

Example stack traces:
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libobjc.A.dylib               	0x000000010f99194b objc_release + 11
1   com.apple.CoreFoundation      	0x000000010fb87b36 -[__NSArrayI dealloc] + 86
2   libobjc.A.dylib               	0x000000010f991a16 objc_object::sidetable_release(bool) + 174
3   com.apple.Foundation          	0x000000010f3b1672 empty + 58
4   com.apple.Foundation          	0x000000010f3b173b dealloc + 38
5   com.apple.Foundation          	0x000000010f3b16d9 -[NSConcreteMapTable dealloc] + 73
6   libobjc.A.dylib               	0x000000010f991a16 objc_object::sidetable_release(bool) + 174
7   org.reactjs.native.example.ReanimatedNavigationCrash	0x000000010a95f103 __destroy_helper_block_e8_32s + 35 (GCDAsyncUdpSocket.m:461)
8   libsystem_blocks.dylib        	0x00000001119f89ba _Block_release + 101
9   com.apple.CoreFoundation      	0x000000010fa2c746 -[__NSArrayM dealloc] + 406
10  libobjc.A.dylib               	0x000000010f991a16 objc_object::sidetable_release(bool) + 174
11  org.reactjs.native.example.ReanimatedNavigationCrash	0x000000010a961cbc __destroy_helper_block_e8_32s40s48s + 92 (GCDAsyncUdpSocket.m:956)
[...]

PR #1187, opened 2020-09-04
Fix Navigation crash

Szymon20000:
Fixing: https://github.com/software-mansion/react-native-reanimated/issues/1181
The problem linked above caused crashes in random places and after 1.5d of debugging I came up with an idea that there must be sth wrong with the memory management. Because of the problem could be reproduced with just one mapper and one animation it was clear for me that the root cause of the problem is in the notification of mapper dirtiness. We may want to notify the mapper which has been unregistered. In order to fix the problem, I added a code that removes listeners of unregistered mappers.

### DIFF
--- a/Common/cpp/SharedItems/MutableValue.cpp
+++ b/Common/cpp/SharedItems/MutableValue.cpp
@@ -8,8 +8,10 @@ namespace reanimated {
 void MutableValue::setValue(jsi::Runtime &rt, const jsi::Value &newValue) {
   std::lock_guard<std::mutex> lock(readWriteMutex);
   value = ShareableValue::adapt(rt, newValue, module);
-  auto notifyListeners = [this] () {
-    for (auto listener : listeners) {
+  
+  std::shared_ptr<MutableValue> thiz = shared_from_this();
+  auto notifyListeners = [thiz] () {
+    for (auto listener : thiz->listeners) {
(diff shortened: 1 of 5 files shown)


karol-bisztyga (approved):

Common/cpp/SharedItems/MutableValue.cpp
```diff
@@ -87,18 +89,18 @@ std::vector<jsi::PropNameID> MutableValue::getPropertyNames(jsi::Runtime &rt) {
...
+unsigned long int MutableValue::addListener(unsigned long id, std::function<void ()> listener) {
+  listeners[id] = listener;
   return id;
 }
 
 void MutableValue::removeListener(unsigned long listenerId) {
-  listeners.erase(std::remove_if(listeners.begin(), listeners.end(), [=](const std::pair<unsigned long, std::function<void()>>& pair) {
-    return pair.first == listenerId;
-  }), listeners.end());
+  if (listeners.count(listenerId) > 0) {
```
kmagiera, line 101: why do we need `count` here? Can't we just call `erase` regardless of whether the key exists or not?

Szymon20000: Looks like it's not necessary to check if the key is present in the map.
You are right It can be removed.

PR #1190, opened 2020-09-04
Fix Install Guide Mistake

FuzzyFade:
#import <RNReanimated/REATurboModuleProvider.h> -> #import <RNReanimated/RETurboModuleProvider.h>


jakub-gonet:
This is in fact correct but in not yet released alpha version. See #1166 for more context.

Either way, thanks for contributing.

FuzzyFade:
https://github.com/software-mansion-labs/reanimated-2-playground/blob/master/ios/Reanimated2Playground/AppDelegate.mm#L19 @jakub-gonet 那你倒是把这个给👴改了啊

jakub-gonet:
Please use English. We will update the playground after alpha 6 will be released (today, hopefully).

PR #1191, opened 2020-09-04
Android fixes

karol-bisztyga:
## Description

- make liquid swipe example work
- Fix scroll imitation example

### DIFF
--- a/Example/ios/Podfile.lock
+++ b/Example/ios/Podfile.lock
@@ -266,11 +266,11 @@ PODS:
     - React-cxxreact (= 0.63.0)
     - React-jsi (= 0.63.0)
     - ReactCommon/turbomodule/core (= 0.63.0)
-  - RNCMaskedView (0.1.7):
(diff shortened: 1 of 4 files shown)


Szymon20000 (approved):

jakub-gonet (changes requested):
CI fails.

karol-bisztyga:
> CI fails.

I noticed, repairing

jakub-gonet:
@karol-bisztyga, you probably meant 0.10.0 version and not 1.10.0, that's why it's still failing.

Szymon20000:
> @karol-bisztyga, you probably meant 0.10.0 version and not 1.10.0, that's why it's still failing.

0.1.10 actually

karol-bisztyga:
Ok so the problem persisted because I chose non existent version so yarn came up with the prompt which made it work locally but did not change package.json

Szymon20000 (approved):

jakub-gonet:
Included in #1193.

PR #1194, opened 2020-09-04
Add ability to use pinch gesture handle event type

### LINKED ISSUE #1164: PinchGestureHandler is not supported?
## Description

PinchGestureHandler is not supported?

## Code

```
const Test: FunctionComponent = () => {
  const onGestureEvent = useAnimatedGestureHandler({
    onActive: () => {
      console.info('active');
    },
  });

  return (
    <PinchGestureHandler onGestureEvent={onGestureEvent}>
      <Animated.View >
      </Animated.View>
    </PinchGestureHandler>
  );
};
```

## typescript error 

```
No overload matches this call.
  Overload 1 of 2, '(props: Readonly<PinchGestureHandlerProperties>): PinchGestureHandler', gave the following error.
    Type 'OnGestureEvent' is not assignable to type '(event: PinchGestureHandlerGestureEvent) => void'.
      Types of parameters 'event' and 'event' are incompatible.
        Type 'PinchGestureHandlerGestureEvent' is not assignable to type 'PanGestureHandlerGestureEvent'.
          Types of property 'nativeEvent' are incompatible.
            Type 'GestureHandlerGestureEventNativeEvent & PinchGestureHandlerEventExtra' is not assignable to type 'NativeEvent'.
              Type 'GestureHandlerGestureEventNativeEvent & PinchGestureHandlerEventExtra' is missing the following properties from type 'PanGestureHandlerEventExtra': x, y, absoluteX, absoluteY, and 4 more.
  Overload 2 of 2, '(props: PinchGestureHandlerProperties, context?: any): PinchGestureHandler', gave the following error.
    Type 'OnGestureEvent' is not assignable to type '(event: PinchGestureHandlerGestureEvent) => void'.
```

### Package versions

 "react": "16.13.1",
 "react-native": "0.63.2",
 "react-native-gesture-handler": "1.7.0",
 "react-native-reanimated": "2.0.0-alpha.5",

terrysahaidak:
## Description

Right now there is no way to use pinch gesture handler with `useAnimatedGestureHandler` in typescript because of wrong typing. This PR fixes it.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/1164

Now when we want to use PanGestureHandler, we can just use `useAnimatedGestureHandler`, but if you want to use typed context, you need to pass both context type and gesture handler type:

```ts
const handler = useAnimatedGestureHandler<
  PanGestureHandlerGestureEvent,
  {
    someContextProp: number;
  },
>(handlers);
```

For all the other handlers, you need to pass them:

```ts
const tapHandler = useAnimatedGestureHandler<
  TapGestureHandlerGestureEvent,
>(handlers);
```

By default, the context has a type of an empty object.

## Changes

- Typings: Add ability to pass event type to useAnimatedGestureHandler


renanmav (approved):
Nice, this PR is only for typings, right? Do you know which PR has done the implementation, I mean, parsed typescript syntax correctly?

terrysahaidak:
The TS has been working just fine starting from alpha.3 I think. 

I've been using it for a while :)

react-native-reanimated.d.ts
```diff
@@ -463,20 +465,19 @@ declare module 'react-native-reanimated' {
     export function scrollTo(ref: RefObject<ReactNativeScrollView | ScrollView>, x: number, y: number, animated: boolean): void;
 
     // gesture-handler
-    type OnGestureEvent = (event: PanGestureHandlerGestureEvent) => void;
+    type OnGestureEvent<T> = (event: T) => void;
```
jakub-gonet, line 468: I wonder if it shouldn't be `T extends GestureHandlerGestureEvent` to narrow type down. Writing this here, but applies to changes in `Handler` and `GestureHandlers` types.

terrysahaidak: Yep, it can be, it won't change anything in the current implementation, but may help in the future.

kmagiera:
Maybe we can add ts run on the tests.tsx file to commit hook too @karol-bisztyga (unrelated to this PR but writing it down so that we don't forget about it)

jakub-gonet:
@kmagiera, it does run right now, no? Or I'm missing something?
https://travis-ci.org/github/software-mansion/react-native-reanimated/jobs/727021701#L225

karol-bisztyga (approved):

karol-bisztyga:
> @kmagiera, it does run right now, no? Or I'm missing something?
> [travis-ci.org/github/software-mansion/react-native-reanimated/jobs/727021701#L225](https://travis-ci.org/github/software-mansion/react-native-reanimated/jobs/727021701#L225)

You are talking about CI and we wanted to add this to pre commit hooks

Issue #823, opened 2020-05-29
Reanimated V2, worklet API


jfrolich:
The worklet API is JavaScript code that contains redundant code (the `"worklet"` string). This means that some compile to JS languages can't produce the code because the compiler rightfully strips it away. An example is BuckleScript / ReasonML.

Would it be possible to revise this API decision, or have an alternative API? I would suggest having distinct naming for the function so the babel plugin knows it's a worklet. Like pretending the name with `worklet__` or something similar.

romanonthego:
I too find API... strange. A bit uncomfortable event.
Could it be done like `use...` pattern for hooks? like `worklet...`?
Or may be like 
```js
const animation = () => {
  //....
}

animation.worklet = true
```

terrysahaidak:
Hi @romanonthego. The only thing required in order to mark it as worklet is to use "worklet". This is just mark for babel to do its thing. The problem with .worklet - it can be defined at runtime or like something which can lead to tought times for babel to detect it. 

Could you explain why .worklet is way better than "worklet"?

jfrolich:
> Hi @romanonthego. The only thing required in order to mark it as worklet is to use "worklet". This is just mark for babel to do its thing. The problem with .worklet - it can be defined at runtime or like something which can lead to tought times for babel to detect it.
> 
> Could you explain why .worklet is way better than "worklet"?

The `"worklet"` string will be removed by every optimizing compiler or smart enough compile to JS languages (specifically it's impossible to do this in ReasonML/BuckleScript). That is the problem. It can be resolved by indicating that something is a worklet by pretending the function name, or something that cannot be stripped away by the compiler, like `worklet()` in the body or something like:

```js
let aWorkLet = worklet(() => {
  ...
})
```

terrysahaidak:
Do you use babel with ReasonML/BuckleScript projects?

jfrolich:
> Do you use babel with ReasonML/BuckleScript projects?

Yes

terrysahaidak:
Well, worklet wrapper could work then, yeah.
Just bit sure it's the best solution for now, but we'll take a look.

jfrolich:
Thanks. I know of a few companies that write apps with Reason and React-Native (including my own company). So having a way to do this in Reason would be amazing.

Ragnar-H:
To add to the API discussion here for compile-to-JS frameworks.

[...]

PR #1207, opened 2020-09-08
Add unit tests for the babel plugin

Ragnar-H:
## Description

Groundwork for addressing things like https://github.com/software-mansion/react-native-reanimated/issues/823

I wanted to make changes to the babel plugin, but I couldn't see any focused tests for the plugin itself. From the looks of thing we're relying on the test app here.

Since I don't want to break things, here are some fairly basic tests to cover the babel plugin.

If there already are tests that cover this feel free to close this out and point me in the right direction :)

## Changes

- Add Jest spec for babel plugin transformation

### DIFF
--- a/src/reanimated2/plugin.test.js
+++ b/src/reanimated2/plugin.test.js
@@ -1,0 +1,53 @@
+import plugin from '../../plugin';
+import { transform } from '@babel/core';
+import { format } from 'prettier';
+
+function runPlugin(input) {
+  return transform(input, {
+    filename: 'someFile.js',
+    plugins: [plugin],
+  });
+}
+
+function prettierOutput(input) {
+  return format(input, { parser: 'babel' });
+}
+
+describe('babel plugin', () => {
+  it('transforms ', () => {
+    const input = `
+    import Animated, { useAnimatedStyle, useSharedValue } from 'react-native-reanimated'
+    
+    function Box() {
+        const offset = useSharedValue(0);
+      
+        const animatedStyles = useAnimatedStyle(() => {
+          return {
+            transform: [{ translateX: offset.value * 255 }],
+          };
+        });
+      
+        return (
+          <>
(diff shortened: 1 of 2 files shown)


src/reanimated2/plugin.test.js
```diff
@@ -0,0 +1,53 @@
...
+import { format } from 'prettier';
+
+function runPlugin(input) {
+  return transform(input, {
+    filename: 'someFile.js',
+    plugins: [plugin],
+  });
+}
+
+function prettierOutput(input) {
```
Ragnar-H, line 12: opted for prettifying the output before snapshots for readability. Otherwise we're going to be diffing against a loooong one-liner.

Ragnar-H:
Possible improvements for this could be a more robust test setup like this: https://github.com/thepassle/babel-plugin-demo/tree/master/test

I went as simple as possible to start with since I'm not that familiar with the babel plugin and I'm not sure this is the way we want to be testing this.

src/reanimated2/plugin.test.js
jakub-gonet: `compact: false` in transform options won't help with that? It should prevent babel from removing newlines

jakub-gonet (commented):
I think this is a step in the right direction, we should definitely test some aspects of the rea2 babel plugin like variable capturing, hashing, checking what's transformed and what's not, etc.

LGTM (except that comment about compacting).

src/reanimated2/plugin.test.js
Ragnar-H: ah nice @jakub-gonet! That's the option I was looking for but couldn't find. Thanks!

Added a commit with `compact: false`

jakub-gonet (approved):

Ragnar-H:
Thanks for the quick response @jakub-gonet :bow:

PR #1209, opened 2020-09-08
Lazy imports/exports

radex:
## Description

Hi. I've been profiling the launch time of my app, and optimizing the critical path to launch the app. I found reanimated to take ~30ms to launch (to import it), but very few of its exports are actually required to launch my app.

And so I applied the technique react-native uses (explicit export with getters that lazily require sub-modules) to make the reexport lazy. I didn't apply it completely, mostly just to the biggest culprits - animations/ which create very complicated proc()s that take a lot of time to send to native.

Is it ugly? Yes. Is it worth it? Sure. After this change, reanimated takes ~10ms to import.

## Test plan

I ran automated tests in this repo + ran manual tests on [Nozbe Teams](https://nozbe.com) to check for regressions.

### DIFF
--- a/src/Animated.js
+++ b/src/Animated.js
@@ -7,15 +7,10 @@ import AnimatedCode from './core/AnimatedCode';
 import * as base from './base';
 import * as derived from './derived';
 import createAnimatedComponent from './createAnimatedComponent';
-import decay from './animations/decay';
-import timing from './animations/timing';
-import spring from './animations/spring';
-import Animation from './animations/Animation';
 import {
   addWhitelistedNativeProps,
   addWhitelistedUIProps,
 } from './ConfigHelper';
-import backwardCompatibleAnimWrapper from './animations/backwardCompatibleAnimWrapper';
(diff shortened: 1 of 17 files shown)


jakub-gonet (changes requested):
1. Formatting shouldn't be the part of this change, please keep only the exports change here.
2. Why did you update `yarn.lock`?
3. Please change the branch to `master`, we cherry-pick into v1 and not the other way around.

jkadamczyk:
Hey @radex! If you wouldn't mind applying the changes @jakub-gonet suggested, I would be happy to merge your PR after some additional testing.

I also spotted your PR in Gesture Handler https://github.com/software-mansion/react-native-gesture-handler/pull/1187

Looking forward to your changes!

jakub-gonet:
@radex 🏓

piaskowyk:
I close it because PR is inactive, but if someone wants to take care of this one I will reopen it with pleasure 🚀

PR #1211, opened 2020-09-08
Make `processColor()` available via the mock

wcandillon:
processColor is exported since alpha.6 and should be available via the mock.


jakub-gonet:
It is, thanks!

jakub-gonet (approved):

Issue #1123, opened 2020-08-14
useAnimatedStyle height blocked by main thread, transform is not.


kyle-ssg:
## Description

If I block the main thread with an intensive function, animations only seem to be non-blocking if I use transform rather than height. It was my understanding from the documentation that although some styles are more intensive to compute it would all happen on the UI thread. 

"supports animations of all native properties without generating any load on the main JavaScript thread (including layout properties like width, flex, etc.)."

However, I have a very basic scenario that seems to disprove this.

**Using transform**

<img src="http://g.recordit.co/Wwfpufe5tg.gif"/>

**Using height**

<img src="http://g.recordit.co/InWoyl57ra.gif"/>

## Steps To Reproduce

1. Add the following to a component that has a gesture handler

```
  useEffect(() => {
    setTimeout(() => {
      function mySlowFunction(baseNumber) {
        console.warn('mySlowFunction');
        let result = 0;
        for (var i = Math.pow(baseNumber, 7); i >= 0; i--) {
          result += Math.atan(i) * Math.tan(i);
        };
        console.warn('mySlowFunction');
      }

      mySlowFunction(100); // higher number => more iterations => slower

    }, 2000);
  }, []);

```

The following is blocked by the main thread 

```
  const animatedHeightStyle = useAnimatedStyle(() => ({
    height: animatedHeight.value,
  }));
```

Where as this is not 

```
  const animatedHeightStyle = useAnimatedStyle(() => ({
    transform: [{translateY:animatedHeight.value}],
  }));
```

### Expected behavior
I'd expect a blocked main thread to have no effect regardless of style in useAnimatedStyle.

### Actual behavior
Using height rather than transform causes a blocked animation.

## Snack or minimal code example

VerticalSlider in:

https://github.com/kyle-ssg/reanimated2-experiments

**The file in question**

https://github.com/kyle-ssg/reanimated2-experiments/blob/master/App/components/VerticalSlider.tsx

## Package versions

```
   "react": "16.13.1",
    "react-native": "0.63.1",
    "react-native-reanimated": "2.0.0-alpha.5",
```

terrysahaidak:
All the UI properties (which don't require layout recalculation) are updated on UI Thread. Layout properties are asynchronously updated on Main Thread - don't confuse it with JS Thread which you're blocking in this example.

I remember I tried to test similar example and it worked perfectly back then.

Might be some regression.

kyle-ssg:
[...]

PR #1215, opened 2020-09-10
Get rid of lag during direct events and improve native props updating on iOS.

Szymon20000:
## WHY
Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1123
To synchronize prop updates caused by worklets with native components (e.g. ScrollView) during direct events (e.g. scroll).

## HOW

React Native is overzealous and avoids layout operation even when it's completely safe to perform. So In this pr, we loosen the limitations put on the layout operation and perform it if UIOperationQueue is empty. Additionally, we block UI thread in order to perform both layout and mount synchronously. That's the way in which The linked issue has been solved. Another part of the pull-request is to remove lag during scrolling (and other direct events). To make it work the following things have been done:
 
- execute mappers in onEvent (in order to trigger props update in useAnimatedStyle)
- block UI thread during layout operation and call mount on UI thread right after the layout is done.
- trigger layout not only on render but also on direct events.

## Testing

I've run the example from the issue on both iOS and Android and it worked perfectly. 
To test if the lag has been removed I used this code:

<details> 
<summary> lag_scroll_code </summary>

```JS
import Animated, {
  useDerivedValue,
  useSharedValue,
  useAnimatedStyle,
  useAnimatedScrollHandler,
  interpolate,
import {
  View,
  StyleSheet,
  FlatList,
  Text,
  Dimensions,
  ScrollView,
  SafeAreaView,
} from 'react-native';
import React, { useCallback } from 'react';

const HEADER_HEIGHT = Dimensions.get('window').height * 0.31;
const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
export default function FlatListStuff(props) {
  const scrollOffset = useSharedValue(0);
  // const headerHeight = useDerivedValue(() => {
  //   return ;
  // });
    onScroll: (event) => {
      scrollOffset.value = event.contentOffset.y;
    },
  });
  const animatedHeaderStyles = useAnimatedStyle(() => {
    return {
       height: 350,
       transform: [{ translateY: -scrollOffset.value }],
       //height: 350 - scrollOffset.value,
    };
  });
  const renderItem = useCallback(
    ({ item }) => (
      <View style={styles.item}>
        <Text>{item}</Text>
      </View>
    ),
    []
  );
  return (
    <SafeAreaView style={styles.container}>
      <View style={StyleSheet.absoluteFill}  collapsable={false}>
        <View style={StyleSheet.absoluteFill} collapsable={false}>
          <AnimatedFlatList
            style={styles.list}
            contentContainerStyle={{ paddingTop: HEADER_HEIGHT + 20 }}
            data={[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}
            renderItem={renderItem}
            onScroll={scrollHandler}
            scrollEventThrottle={16}
            keyExtractor={(item, i) => i.toString()}
          />
          <Animated.View style={[styles.header, animatedHeaderStyles]} />
          <Text style={{ position: 'absolute', bottom: 50, left: 16 }}>
            {scrollOffset.value}
          </Text>
        </View>
      </View>
    </SafeAreaView>
  );
}
  header: {
    position: 'absolute',
    height: HEADER_HEIGHT,
    width: '100%',
    backgroundColor: 'purple',
    shadowColor: 'black',
    shadowOpacity: 0.3,
    shadowRadius: 15,
    shadowOffset: { height: 30 },
    borderBottomLeftRadius: 24,
    borderBottomRightRadius: 24,
  },
  list: {
    flex: 1,
  },
  item: {
    height: 50,
    borderBottomWidth: StyleSheet.hairlineWidth,
    justifyContent: 'center',
  },
  container: {
    flex: 1,
  }
});
```

</details>


karol-bisztyga (approved):

ios/REANodesManager.m
```diff
@@ -39,6 +40,12 @@ - (void)setNeedsLayout;
 
 @end
 
+@interface REANodesManager() <RCTUIManagerObserver>
+
+@property BOOL shouldInterceptMountingBlock;
```
kmagiera, line 45: There is no point in defining this as a property since you only want to use it in the class implementation. In this case you should define it in the block direcly after `@implementation` statement below (e.g., where `_wantRunUpdates` is defined)

ios/REANodesManager.m
```diff
@@ -52,6 +59,7 @@ @implementation REANodesManager
   NSMutableArray<REAOnAnimationCallback> *_onAnimationCallbacks;
   NSMutableArray<REANativeAnimationOp> *_operationsInBatch;
   REAEventHandler _eventHandler;
+  volatile void (^_mounting)(void);
```
kmagiera, line 62: why `volatile`?

ios/REANodesManager.m
```diff
@@ -174,12 +196,36 @@ - (void)performOperations
   if (_operationsInBatch.count != 0) {
     NSMutableArray<REANativeAnimationOp> *copiedOperationsQueue = _operationsInBatch;
     _operationsInBatch = [NSMutableArray new];
+    
+    __weak typeof(self) weakSelf = self;
+    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
     RCTExecuteOnUIManagerQueue(^{
+      NSMutableArray *pendingUIBlocks = [self.uiManager valueForKey:@"_pendingUIBlocks"];
+      bool canPerformLayout = (pendingUIBlocks == nil) || ([pendingUIBlocks count] == 0);
```
kmagiera, line 204: `[_pendingUIBlocks count]` will return 0 when `pendingUIBlocks` is `nil` so there is no need to have these two checks

ios/REANodesManager.m
```diff
@@ -174,12 +196,36 @@ - (void)performOperations
...
+        dispatch_semaphore_signal(semaphore);
+      }
+      
       for (int i = 0; i < copiedOperationsQueue.count; i++) {
         copiedOperationsQueue[i](self.uiManager);
       }
-      [self.uiManager setNeedsLayout];
+      
+      if (canPerformLayout) {
+        weakSelf.shouldInterceptMountingBlock = YES;
```
kmagiera, line 215: we already retain `self` (see, e.g. the line below), why adding another reference via weakSelf?

ios/REANodesManager.m
```diff
@@ -382,7 +428,13 @@ - (void)dispatchEvent:(id<RCTEvent>)event
   event.eventName];
 
   if (_eventHandler != nil) {
-    _eventHandler(eventHash, event);
+    __strong REAEventHandler eventHandler = _eventHandler;
```
kmagiera, line 431: this should not be needed. The block below already captures `self`, so you can access event handler using `self->_eventHandler`

ios/REANodesManager.m
Szymon20000: That's because we use it frequently on two threads. I used volatile because I don't want threads to cache mounting.

ios/REANodesManager.m
Szymon20000: I don't want to retain self because our block is used by UIManager so there would be a reference loop if UIManager was invalidated.

ios/REANodesManager.m
Szymon20000: However, I want to access it by a weak reference.

kmagiera: this does not matter, you can always do `weakRef->_shouldInreceptMountikngBlock`

ios/REANodesManager.m
kmagiera: Why don't we use it with `shouldInterceptMountingBlock` too?

Szymon20000: Properties are atomic by default.

ios/REANodesManager.m
Szymon20000: Dereferencing a __weak pointer is not allowed due to possible null value caused by race condition, assign it to strong variable first

wcandillon:
This is a game changer :)

SudoPlz:
I still experience that issue in `2.2.2` (on iOS). Shouldn't that be fixed for versions above 2.0.0.alpha?

piaskowyk:
@SudoPlz 
If you have still a similar problem, could you open a new issue?

PR #1218, opened 2020-09-11
Add hooks deps to TS types

terrysahaidak:
## Description

In the latest release (alpha.6) dependencies have been added to hooks in order to make sure worklets stay up to date with the rest of the react code. But they are missing in types. This PR fixes it.

## Changes

- Add hooks deps to TS types


jakub-gonet (approved):
TS code looks good but I don't know if every hook was changed, so I'll leave merge decision to the @Szymon20000 and @karol-bisztyga

Issue #1152, opened 2020-08-24
Web support for v2


kmagiera:

hosseinmd:
@terrysahaidak

karol-bisztyga:
https://github.com/software-mansion/react-native-reanimated/tree/web-support

hosseinmd:
Can we test this?

karol-bisztyga:
Yes but it does contain bugs

terrysahaidak:
Some features such as gesture handler support is missing right now

karol-bisztyga:
Web support has been added, closing.

PR #1225, opened 2020-09-14
Web support for reanimated 2

karol-bisztyga:
## Description

Adding web support for reanimated 2 to allow firing it up in the browser.

## Usage

Two new scripts have been added to Example app:
- `generate-web` - overwrites `index.js` with values proper for the web version.
- `start-web` - performs the previous script, starts a local server and runs the web app.

Basically, if you want a quick start, just launch `yarn start-web`, the generating script has been added to keep the scripts consistent.

### Cleaning cache

Unlike the other scripts(`start`, `start-test`, `start-test-suite`) for which you would use `reset-cache` to clean the cache, here all you need is `-c`(this comes from expo).
So here is the example of starting metro server with cleaning cache for the iOS/Android:
`yarn start --reset-cache`
And that's like you would use it for starting the web server:
`yarn start-web -c`

## Changes

The native code of the library has been rewritten in javascript.

Keep in mind that some packages' versions changed and those changes are very important, they make the web support work properly so do not forget to rebuild node modules, the best way to go is reinstalling the entire thing like:
`rm -rf node_modules && yarn`

https://github.com/software-mansion/react-native-reanimated/issues/1152


karol-bisztyga:
Big thank you for @terrysahaidak for a huge help in implementing this ❤️

Example/src/DragAndSnapExample.js
```diff
@@ -53,7 +53,7 @@ function DragAndSnap() {
 
   return (
     <View style={{ flex: 1, margin: 50 }}>
-      <PanGestureHandler onGestureEvent={gestureHandler}>
+      <PanGestureHandler onHandlerStateChange={gestureHandler}>
```
Szymon20000, line 56: When `onHandlerStateChange` is triggered? Will the behavior be the same?

src/reanimated2/NativeReanimated.js
```diff
@@ -1,10 +1,15 @@
...
 const InnerNativeModule =
-  global.__reanimatedModuleProxy ||
-  TurboModuleRegistry.getEnforcing('NativeReanimated');
+  Platform.OS === 'web'
+    ? false
+    : global.__reanimatedModuleProxy ||
+      TurboModuleRegistry.getEnforcing('NativeReanimated');
+
+const NativeReanimated = {
+  native: !!InnerNativeModule, // FIXME: Put !! here instead
```
terrysahaidak, line 11:
```suggestion
  native: !!InnerNativeModule,
```

src/reanimated2/js-reanimated/MapperRegistry.js
```diff
@@ -0,0 +1,116 @@
...
+  updateOrder() {
+    class Node {
+      mapper = null;
+      children = [];
+
+      constructor(mapper, parents = [], children = []) {
+        this.mapper = mapper;
+        this.children = children;
+      }
+    }
```
terrysahaidak, line 44: I think this can be moved away from this method

terrysahaidak (approved):
The only thing that can be updated now/later as a performance optimization - to move away from `forEach` to regular `for()`. Since animations are running on each frame that 1-3ms delay may be significant.

src/reanimated2/js-reanimated/MapperRegistry.js
karol-bisztyga: 5ee34fd

src/reanimated2/NativeReanimated.js
karol-bisztyga: 5ee34fd

terrysahaidak:
There are few others forEach calls in this PR which can be replaced with for as well

src/reanimated2/js-reanimated/MutableValue.js
```diff
@@ -0,0 +1,38 @@
...
+    this._setter = setter;
+  }
+
+  get value() {
+    return this._value;
+  }
+
+  set value(nextValue) {
+    // eslint-disable-next-line
+    this._setter.apply(this, [nextValue]);
```
jakub-gonet, line 19: Why this can't be just `this._setter(nextValue)`?

karol-bisztyga:
> There are few others forEach calls in this PR which can be replaced with for as well

I know 😄 I did not want to replace all of them, I will check how often they get called later, I would replace only those which affect efficiency.

src/reanimated2/Hooks.js
```diff
@@ -32,7 +40,7 @@ export function useMapper(fun, inputs = [], outputs = [], dependencies = []) {
 
 export function useEvent(handler, eventNames = [], rebuild = false) {
   const initRef = useRef(null);
-  if (initRef.current === null) {
+  if (initRef.current === null || (Platform.OS === 'web' && rebuild)) {
```
Szymon20000, line 43: Don't we want to rebuild on iOS and Android?

karol-bisztyga: it does rebuild on mobile two lines below.

src/reanimated2/js-reanimated/MutableValue.js
karol-bisztyga: Just wanted to make sure what `this` would be but in fact this straight call works as well.

karol-bisztyga: ef26915

Szymon20000 (approved):

terrysahaidak:
I will try again with gallery in an hour to confirm it works as expected

Example/src/MeasureExample.js
```diff
@@ -112,22 +112,51 @@ function Section({ title, children, height, contentHeight, z, show }) {
...
+    }
+  };
+
+  let onActiveImpl;
+  if (Platform.OS === 'web') {
+    onActiveImpl = async (_, ctx) => {
+      try {
+        applyMeasure(await asyncMeasure(animatedRef));
+      } catch (e) {
+        console.log(e);
```
jakub-gonet, line 148: What happens if measure fails on native? Shouldn't we just let it fails instead of logging the exception?

Example/src/ScrollableViewExample.js
```diff
@@ -128,10 +128,22 @@ function Box({ color }) {
...
   );
 }
+
+const styles = StyleSheet.create({
+  wrapper: {
+    height:
+      Platform.OS === 'web'
+        ? windowDimensions.height - Header.HEIGHT
+        : undefined,
+    overflow: Platform.OS === 'web' ? 'hidden' : undefined,
```
jakub-gonet, line 147: Setting `undefined` here uses a default value or just overwrite default with `undefined`?

package.json
```diff
@@ -11,7 +11,7 @@
     "release": "npm login && release-it",
     "type:check": "tsc"
   },
-  "main": "lib/commonjs/Animated",
+  "main": "src/Animated.js",
```
jakub-gonet, line 14: BTW, this is from bob config, seems like we don't use it right now. Could we reintroduce it or the way how we build android is preventing us from doing it?

src/createAnimatedComponent.js
```diff
@@ -31,7 +31,7 @@ function hasAnimatedNodes(value) {
   if (Array.isArray(value)) {
     return value.some((item) => hasAnimatedNodes(item));
   }
-  if (value && typeof value === 'object') {
+  if (typeof value === 'object') {
```
jakub-gonet, line 34: If the value is `null` then we'll get `Uncaught TypeError: Cannot convert undefined or null to object`.

Check #1113 for more info.

src/index.js
```diff
@@ -0,0 +1 @@
+Animated.js
```
jakub-gonet, line 1: What does it do?

src/reanimated2/NativeReanimated.js
```diff
@@ -1,10 +1,15 @@
-import { TurboModuleRegistry } from 'react-native';
+import { TurboModuleRegistry, Platform } from 'react-native';
+import reanimatedJS from './js-reanimated';
 
 const InnerNativeModule =
-  global.__reanimatedModuleProxy ||
-  TurboModuleRegistry.getEnforcing('NativeReanimated');
+  Platform.OS === 'web'
+    ? false
+    : global.__reanimatedModuleProxy ||
```
jakub-gonet, line 7: Could you add parentheses here? It's not clear if it's evaluated like `(A ? B : C) || D` or `A ? B : (C || D)`

Example/src/ScrollableViewExample.js
karol-bisztyga: it assigns udefined

src/reanimated2/NativeReanimated.js
karol-bisztyga: eslint reverts this anyway, could disable it of course, but this looks fine to me(not ambiguous)

src/createAnimatedComponent.js
karol-bisztyga: fixed c7ce9ee

Example/src/MeasureExample.js
karol-bisztyga: ok, we can rethrow c7ce9ee

Example/src/ScrollableViewExample.js
jakub-gonet: So we actually want an undefined here?

karol-bisztyga: it's ok

nandorojo:
Amazing work – thank you!

PR #1230, opened 2020-09-16
Add v2 mocks

### LINKED ISSUE #954: Testing error: useSharedValue is not a function while testing
## Description

useSharedValue throws an error that it's not a function when using jest and react-native-testing-library.  I've mocked react-native-reanimated by using the mock provided as part of the package.  I also notice VS Code highlights that reanimated has no exported member useSharedValue, useAnimatedStyle and withTiming, however the component still runs properly without any errors.

Does anyone know what needs to be done to mock it properly?

### Screenshots
Screen Shot 2020-07-07 at 2 48 22 PM
image

## Steps To Reproduce

1. Create the simple switch component
2. Create the mock file
3. Create the test 
4. Run the test

### Expected behavior
I expected the render test to pass since I'm successfully mocking react-native-reanimated

### Actual behavior
The render test fails due to useSharedValue not being a function. `"TypeError: (0 , _reactNativeReanimated.useSharedValue) is not a function"`

**The Test to see if the component renders**

`import React from "react";
import { render } from "react-native-testing-library";
import { Switch } from "../src/native/Switch";

describe("Switch Package", () => {
  it("renders", () => {
    render(<Switch />);
  });
});`

**Mock file which is placed in __mocks__ named react-native-reanimated.js**

`jest.mock("react-native-reanimated", () =>
  require("react-native-reanimated/mock")
);`

**The Component using Reanimated - Switch.tsx**
`import { Button } from 'react-native';
import {
  useSharedValue,
  useAnimatedStyle,
  withTiming,
  Easing,

function Switch() {
  const opacity = useSharedValue(0);

    return {
      opacity: withTiming(opacity.value, {
        duration: 200,
        easing: sing.out(Easing.cubic),
      }),
    };
  });

  return (
    <View>
      <Animated.View style={[{width: 100, height: 100, backgroundColor: 'green'}, style]} />
      <Button onPress={() => (opacity.value = 1)} title="Hey" />
    </View>
  );
}`

## Package versions

- React: 16.11.0
- React Native: .062.2
- React Native Reanimated: 2.0.0-alpha.3

### LINKED ISSUE #1195: useSharedValue is missing from mocks
## Description

I'm using `react-native-reanimated/mock` to allow testing with Jest and I get the following error:

```
TypeError: (0 , _reactNativeReanimated.useSharedValue) is not a function
```

In `setupTests.ts`
```ts
jest.mock('react-native-reanimated', () =>
  require('react-native-reanimated/mock')
)
```

and anywhere you use `useSharedValue` the tests will fail:
```ts
const offSetY = useSharedValue(-20)
```

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.5

mgambati:
## Description
Closes #1195.
Closes #954.

Adding missing mocks for V2 API.

## Changes

Add a new `src/reanimated2/mock.js` file with methods to for every API currently in reanimated2.

Also, it renames old mocks of `Easing` to `EasingNode` and reexports new v2 `Easing` in the mocks since there's no need to mock it.


src/reanimated2/mock.js
```diff
@@ -0,0 +1,44 @@
...
+  useSharedValue: (value) => ({ value }),
+  useDerivedValue: (a) => ({ value: a() }),
+  useAnimatedScrollHandler: () => NOOP,
+  useAnimatedGestureHandler: () => NOOP,
+  useAnimatedStyle: (style) => style,
+  useAnimatedRef: () => ({ current: null }),
+  useAnimatedReaction: NOOP,
+
+  withTiming: (toValue, config, cb) => {
+    cb && setTimeout(() => cb(true), config?.duration ?? 300);
```
jakub-gonet, line 16: Why only timing has a duration that isn't 0? And it shouldn't be somehow synchronized with default values in JS code?

mgambati: My initial thoughts on this are that since timing is the only one with a user-specified duration, respecting that duration may be the right choice and the expected behavior by the user.

But I not so sure about it. I can change it to 0.

jakub-gonet: @Szymon20000, @karol-bisztyga do you have any thoughts about this?

wcandillon:
Should we disable the reanimated plugin to work if in the testing environment: https://github.com/software-mansion/react-native-reanimated/blob/master/plugin.js#L417? How do you deal with it @mgambati?

wcandillon:
Nevermind, I just realized that this can done in the babel config directly 😅

mgambati:
I'm disabling it on testing. I had some strange behavior using babelrc json file, so I used the API. Just to document for people searching about it:

```js
module.exports = api => {
  const babelEnv = api.env()
 
 // ....

  const isInTesting = babelEnv === 'test'

  if (isInTesting)
    return baseConfig

  return {
    ...baseConfig,
    plugins: [...baseConfig.plugins, 'react-native-reanimated/plugin'],
  }
}

```

 I think would be nice to add a section about testing on the docs. I can work on it, what do you think?

wcandillon:
I agree, it could be useful to have it mentioned in the docs.

mym0404:
Well.. will this PR be merged at official version `2.0.0`? 
My jest configuration is broken in somethings.

src/reanimated2/mock.js
jakub-gonet: All right, seems like no one had time to look at it. For me, it should have duration 0, we don't want to artificially delay tests.

Besides, is `setTimeout` necessary? Can't we just call the callback directly?

jakub-gonet: Fixed in c6807ec.

jakub-gonet:
I resolved conflicts and changed some parts of the code, @mgambati please review it.

piaskowyk (approved):

jakub-gonet:
> and reexports new v2 Easing in the mocks since there's no need to mock it.

Just want to mention I removed `Easing` import, it's safe to reexport it _right now_ but this may change in the future if Easing will start using native API. Other things from Reanimated are mocked so having Easing implemented doesn't change much, we can't test native animations in jest anyway.

mock.js
```diff
@@ -230,7 +233,10 @@ module.exports = {
 
   ...Reanimated,
 
-  default: Reanimated,
+  default: {
+    ...Reanimated,
+    ...ReanimatedV2,
+  },
```
Andreyco, line 239: Late to the party but I'd like to ask regardless...

I see you are introducing `default` export here.

With given setup instructions, mock are not initialised properly
```
jest.mock('react-native-reanimated', () => require('react-native-reanimated/mock'));
```

Not unless I use `default` export.

```
jest.mock('react-native-reanimated', () => require('react-native-reanimated/mock').default);
```

I believe your approach with `default` export is correct, however usage docs should be updated.

According to implementation (index.js) named exports of:
```
...Reanimated
```
should be dropped, since only these name exports are specified
```
export {
  EasingNode,
  Transitioning,
  Transition,
  createTransitioningComponent,
  // classes
  AnimatedClock as Clock,
  AnimatedValue as Value,
  AnimatedNode as Node,
  // animations
  decayWrapper as decay,
  timingWrapper as timing,
  springWrapper as spring,
  SpringUtils,
  // hooks
  useValue,
};
```

I am correct or is there something I am missing?

mgambati: @Andreyco when exporting with `__esModule: true` Jest will know how to deal with default imports correctly. No change is needed.

About the V1 exports, we export Reanimated V1 APIs for backward compatibility, right?

Andreyco:
I cannot confirm your statement. Just did a test run, even with `__esModule: true`, Jest is... lost.

Seems proper mock initialisation is, indeed, following...
```js
jest.mock('react-native-reanimated', () => require('react-native-reanimated/mock').default);
```

mock.js
Andreyco: I cannot confirm your statement. Just did a test run, even with `__esModule: true`, Jest is... lost.

Seems proper mock initialisation is, indeed, following...
```js
jest.mock('react-native-reanimated', () => require('react-native-reanimated/mock').default);
```

@mgambati could I ask you to double check with your Jest setup?
My setup is:
Jest: 26.6.3
RNR: current master (https://github.com/software-mansion/react-native-reanimated.git#ba4d65e25c08493e2ad0b53d10344a2dd72bdafe)

danielpgross:
These mocks don't seem to match the new naming for `withDelay`, `withSequence`, `withRepeat` introduced [here](https://github.com/software-mansion/react-native-reanimated/commit/8ce6a881f3991f51e0f78e373ac602b6d7d76c6f). Am I missing something, or was this by mistake?

shubhamranjan312:
TypeError: _react.default.useRef.mockReturnValueOnce is not a function

PR #1236, opened 2020-09-17
Improve caching.

Szymon20000:
## WHY
Currently, when we want to use an object in two worklets then we have to covert the object twice. To avoid it this pr saves our jsi::HostObject as a hidden property (Object.defineProperty enumerable: false). Additionally, this pr introduces a different approach to representing FrozenObject and RemoteObject. JSC has problems with jsi::Objects that are created from HostObjects so instead of using them directly we expose frozen(not true with RemoteObject) jsi::Object and save jsi::HostObject as a hidden property.  

## HOW
During makeSherable operation add a hidden property to all objects converted to FrozenObject and WorkletFunction. 
`const char *HIDDEN_HOST_OBJECT_PROP = "__reanimatedHostObjectRef";`
Remove get/set from RemoteObject and return jsi::Object on UI side and frozenObj on JS side. Remove get from FrozenObject and use it only as a C++ representation. We only add a hidden property on JS side. When we convert jsi::Object on UI side in order to send it to RN we don't want to freeze it because it would be torture and it's a rare case.

## BREAKING CHANGE
From now on, all captured objects are frozen. It means if you want to pass a mutable object to the worklet then you have to copy it before.

### DIFF
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -9,6 +9,27 @@
 
 namespace reanimated {
 
+const char *HIDDEN_HOST_OBJECT_PROP = "__reanimatedHostObjectRef";
+  
+void addHiddenProperty(jsi::Runtime &rt,
+                       jsi::Value &&value,
+                       jsi::Object &obj,
+                       const char *name) {
+  jsi::Object globalObject = rt.global().getPropertyAsObject(rt, "Object");
+  jsi::Function defineProperty = globalObject.getPropertyAsFunction(rt, "defineProperty");
+  jsi::String internalPropName = jsi::String::createFromUtf8(rt, name);
+  jsi::Object paramForDefineProperty(rt);
+  paramForDefineProperty.setProperty(rt, "enumerable", false);
+  paramForDefineProperty.setProperty(rt, "value", value);
+  defineProperty.call(rt, obj, internalPropName, paramForDefineProperty);
+}
+
+void freeze(jsi::Runtime &rt, jsi::Object &obj) {
+  jsi::Object globalObject = rt.global().getPropertyAsObject(rt, "Object");
+  jsi::Function freeze = globalObject.getPropertyAsFunction(rt, "freeze");
+  freeze.call(rt, obj);
+}
+
 void ShareableValue::adaptCache(jsi::Runtime &rt, const jsi::Value &value) {
   // when adapting from host object we can assign cached value immediately such that we avoid
   // running `toJSValue` in the future when given object is accessed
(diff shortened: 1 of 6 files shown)


Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -20,6 +41,23 @@ void ShareableValue::adaptCache(jsi::Runtime &rt, const jsi::Value &value) {
 }
 
 void ShareableValue::adapt(jsi::Runtime &rt, const jsi::Value &value, ValueType objectType) {
+  bool isRNRuntime = !(module->isUIRuntime(rt));
+  if (value.isObject()) {
+    jsi::Object object = value.asObject(rt);
+    if (!(object.getProperty(rt, HIDDEN_HOST_OBJECT_PROP).isUndefined())) {
+      jsi::Object hiddenProperty = object.getProperty(rt, HIDDEN_HOST_OBJECT_PROP).asObject(rt);
```
karol-bisztyga, line 48: couldnt `object.getProperty(rt, HIDDEN_HOST_OBJECT_PROP)` be called only once?
POC:
```
auto op = object.getProperty(rt, HIDDEN_HOST_OBJECT_PROP);
if (!op.isUndefined()) {
	jsi::Object hiddenProperty = op.asObject(rt);

	//...
```
Also do we want to check against being undefined? Maybe `isObject()` should be used here instead of `isUndefined()`?

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -20,6 +41,23 @@ void ShareableValue::adaptCache(jsi::Runtime &rt, const jsi::Value &value) {
...
+      if (hiddenProperty.isHostObject<FrozenObject>(rt)) {
+        type = ValueType::ObjectType;
+        if (object.hasProperty(rt, "__worklet") && object.isFunction(rt)) {
+          type = ValueType::WorkletFunctionType;
+        }
+        frozenObject = hiddenProperty.getHostObject<FrozenObject>(rt);
+        adaptCache(rt, value);
+        return;
+      }
+    }
```
karol-bisztyga, line 58: Why not deal with the case when value is object here entirely? The flow reaches the same condition in line 77(doing nothing between here and there)

Example/ios/Podfile.lock
```diff
@@ -268,7 +268,7 @@ PODS:
     - ReactCommon/turbomodule/core (= 0.63.0)
   - RNCMaskedView (0.1.10):
     - React
-  - RNGestureHandler (1.5.2):
```
karol-bisztyga, line 271: unrelated changes and they break CI

Common/cpp/SharedItems/ShareableValue.cpp
Szymon20000: Because it can be mutable.

karol-bisztyga (approved):

Issue #1129, opened 2020-08-16
Style doesn't work for Reanimated View when in a 2nd level nesting 


ethanshar:
## Description

Passing nested style in an array doesn't work. When it's a 2 level nesting

## Steps To Reproduce

See code snippet.

### Expected behavior
Nesting a style (animated style for that matter) should work when nested in a 2nd level array.

### Actual behavior
Style doesn't take any effect and being ignored. 

## Snack or minimal code example

```
const animatedStyle = ...;

// Doesn't Work (2 level nesting)
<Reanimated.View style={[{backgroundColor: 'red'},[animatedStyle]]} />

// Works (1 level nesting)
<Reanimated.View style={[{backgroundColor: 'red'}, animatedStyle]} />
```

## Package versions
react-native-reanimated@2.0.0-alpha.5
react-native@0.62.2
react@16.11.0

terrysahaidak:
Hi. This isn't Reanimated problem. AFAIK this should not work in React Native. Use Array.flatten for this.

ethanshar:
Hi @terrysahaidak 
It does work with React Native View. 
for instance, the following examples works as expected
```
<View style={[{width: 100, height: 100}, [[[{borderWidth: 2}]]]]}/>
```

terrysahaidak:
Does it work with Animated.View as well? Reanimated's createAnimatedComponent should work similar to Animated's one.

ethanshar:
Ok, after some digging I realized it's not related to the View. 
Reanimated View can handle plain style nesting as long it is not a style created with the `useAnimatedStyle` hook. 
So, basically the problem is with the return value of `useAnimatedStyle`. 
Not sure if this can be fixed, I assume there's a reason why this hook returns something "different".

ethanshar:
Hi @karol-bisztyga @jakub-gonet 
It appears this issue has returned in version 2.0.0.

ethanshar:
Hi, 
Is there any update on this issue?

ethanshar:
Hi @piaskowyk @jakub-gonet, 
How are you? 

Can you share an update on this issue?

We love reanimated and waited long time for v2 to be official. 
We're benn already using v1 in our [components library](https://github.com/wix/react-native-ui-lib).
But we'd like to start migrating to v2 and since we're building generic components where users can pass their own style and customize them, nesting styles is a must for us, without it, it'll be really difficult to create a components infrastructure on our end.

[...]

PR #1238, opened 2020-09-17
Fix handling useAnimatedStyle in nested styles

karol-bisztyga:
## Description

This PR fixes using `useAnimatedStyle` hook when styles are in nested arrays with depth > 1.
Before something like this worked:
```
<Animated.View style={ animatedStyle } />
<Animated.View style={[ { width: 100, height: 80, backgroundColor: 'black' }, animatedStyle]} />
```
but something like this did not:
```
<Animated.View style={[ { width: 100, height: 80, backgroundColor: 'black' }, [animatedStyle]]} />
```
From now on it's possible to nest the arrays in `style` prop as deep desired, example:
```
<Animated.View
  style={[
    [
      { backgroundColor: 'green' },
      [
        { backgroundColor: 'yellow' },
        [
          { backgroundColor: 'red' },
          [
            { backgroundColor: 'blue' },
            [{ backgroundColor: 'purple' }, style],
          ],
        ],
      ],
    ],
  ]}
/>
```

This fixes https://github.com/software-mansion/react-native-reanimated/issues/1129

### DIFF
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -37,6 +37,26 @@ function hasAnimatedNodes(value) {
   return false;
 }
 
+function flattenArray(array) {
+  const resultArr = [];
+
+  const _flattenArray = (arr) => {
+    if (!Array.isArray(arr)) {
+      resultArr.push(arr);
+      return;
+    }
+    arr.forEach((item) => {
+      if (Array.isArray(item)) {
+        _flattenArray(item);
+      } else {
+        resultArr.push(item);
+      }
+    });
+  };
+  _flattenArray(array);
+  return resultArr;
+}
+
 export default function createAnimatedComponent(Component) {
   invariant(
     typeof Component !== 'function' ||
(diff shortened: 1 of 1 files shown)


src/createAnimatedComponent.js
```diff
@@ -37,6 +37,25 @@ function hasAnimatedNodes(value) {
   return false;
 }
 
+function flattenArray(array) {
```
jakub-gonet, line 40: This method can be replaced with: [`array.flat(Infinity)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)

jakub-gonet (changes requested):

src/createAnimatedComponent.js
jakub-gonet: Which doesn't preserve `undefined` values but we don't care here.

More about this [here](https://medium.com/dailyjs/flatten-array-using-array-flat-in-javascript-ee4d0b2423e5)

karol-bisztyga: fixed 8a2e33b

jakub-gonet: Seems like we don't support flat yet

https://stackoverflow.com/questions/52868353/javascript-array-flat-function-doesnt-define-in-react-native

jakub-gonet (approved):

Szymon20000 (approved):

PR #1240, opened 2020-09-17
[v1] fix Xcode 12 compatibility

radko93:
## Description

Latest Xcode 12 fails to build while without a module to depend on `React-Core` directly hence this change is necessary for all native modules on iOS. This change requires React Native 0.60.2 or newer. For more details please check: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

### DIFF
--- a/RNReanimated.podspec
+++ b/RNReanimated.podspec
@@ -19,6 +19,6 @@ Pod::Spec.new do |s|
   s.source_files = "ios/**/*.{h,m}"
   s.requires_arc = true
 
-  s.dependency "React"
+  s.dependency "React-Core"
 end
 


jakub-gonet:
That's impressive, thanks for the PR, will merge after CI passes.

jakub-gonet (approved):

jakub-gonet:
Could you do the PR against the master branch? We use v1 to cherry-pick changes from master

radko93:
Master already includes this fix- it already has `React-Core` inside podspec. So it does not make much sense.

hsavit1:
any updates?

hosseintalebi:
I think we still need this on the `v1` branch though.

jakub-gonet (approved):

PR #1259, opened 2020-09-22
Fix URL in Troubleshooting.md

mrousavy:
## Changes

Updated Troubleshooting.md docs, the braces/brackets were just messed up


jakub-gonet (approved):
My bad, `[]` are hardwired to URLs in my brain 😄 

Thanks for catching this.

PR #1264, opened 2020-09-23
add additional example for useDerivedValue

rawatnaresh:
## Description
Added example to run animation based on the props passed to the component. I guess this is a possible use case of `useDerivedValue`

### DIFF
--- a/docs/docs/api/useDerivedValue.md
+++ b/docs/docs/api/useDerivedValue.md
@@ -46,3 +46,49 @@ function App() {
   );
 }
 ```
+
+
+`useDerivedValue` is also useful when we want to run animations based on the props passed to the component.
+
+Let's say we want to increase the width of the progress bar when the shared value in the `ComponentA` is updated.
+
+
+```js {6}
+import { Button } from 'react-native';
+import { useSharedValue, useDerivedValue } from 'react-native-reanimated';
+
+function ComponentA() {
+  const progress = useSharedValue(0);
+  const width = useDerivedValue(() => {
+    return progress.value * 250;
+  });
+
+  return (
+    <View>
+      <ProgressBar width={width} />
+      <Button onPress={() => (progress.value = Math.random())} />
+    </View>
+  );
+}
+```
+```js {6}
+import { Button } from 'react-native';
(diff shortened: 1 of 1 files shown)


terrysahaidak:
Did you try to pass the prop to useAnimatedStyle directly?

And based on that prop return withTiming for the width?

docs/docs/api/useDerivedValue.md
```diff
@@ -46,3 +46,49 @@ function App() {
...
+function ComponentA() {
+  const progress = useSharedValue(0);
+  const width = useDerivedValue(() => {
+    return progress.value * 250;
+  });
+
+  return (
+    <View>
+      <ProgressBar width={width} />
+      <Button onPress={() => (progress.value = Math.random())} />
```
karol-bisztyga, line 69: title prop is missing

karol-bisztyga:
This works alright without `useDerivedValue`, plus the code you provided is not correct

```

import React from 'react';
import { Button, View, Text } from 'react-native';
import Animated, {
  useSharedValue,
  useAnimatedStyle,


function ComponentA() {
  const progress = useSharedValue(0);
  return (
    <View style={{ marginTop: 300 }}>
      <Button title="press" onPress={() => (progress.value = Math.random())} />
      <ProgressBar width={progress} />
    </View>
  );
}

function ProgressBar({ width }) {
  const widthStyle = useAnimatedStyle(() => {
    // if shared value was passed instead of derived shared value,
    // this worklet wouldn't have noticed the changes thus, no update would have happened
    return {
      width: width.value * 250,
    };
  });
  return (
    <Animated.View style={[widthStyle]}>
      <Text>I'll animate when you change width in ComponentA</Text>
    </Animated.View>
  );
}

export default ComponentA;
```

jakub-gonet:
@rawatnaresh, what's the status on this?

rawatnaresh:
@jakub-gonet, @karol-bisztyga is right. it works without `useDerivedValue`. 
initially, I tried it a couple of times without `useDerivedValue` and it didn't work so i had to use the another approach. But later I cleaned the build and installed the app again and it worked fine without `useDerivedValue`.
Thank you @karol-bisztyga for clarification 🙏

jakub-gonet:
I'll close it then. Nevertheless, thanks for your effort, really appreciate it.

PR #1265, opened 2020-09-23
Export Bezier

wcandillon:
This function that gives us Y for X on a Bezier curve via worklet functions, is useful beyond easing.


wcandillon:
Closing it since it seems to be usable via Easing.

Issue #1258, opened 2020-09-21
Add runOnJS method and validate that worklet callbacks are wrapped with it


kmagiera:

mrousavy:
Does this mean that a warning will be emitted when we accidentally call JS methods in a worklet? If so, I'm all for it. I'm never sure if everything is being run on the reanimated thread or I made a mistake and let a bubbling event handler dispatch hundreds of messages to the react thread 😅

PR #1272, opened 2020-09-24
Add runOnJS logic.

Szymon20000:
# WHY
When you call a function on UI thread you can't be sure if you call a worklet or a callback from JS thread. To make it more transparent we introduce `runOnJS` method which calls a callback asynchronously. If you call a callback without this method then an exception will be thrown.

## HOW
add a hidden property to every hostFunction named `callAsync` and call it in runOnJS.

resolves: https://github.com/software-mansion/react-native-reanimated/issues/1258

### DIFF
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -11,6 +11,12 @@ namespace reanimated {
 
 const char *HIDDEN_HOST_OBJECT_PROP = "__reanimatedHostObjectRef";
 const char *ALREADY_CONVERTED= "__alreadyConverted";
+std::string CALLBACK_ERROR = R"(
+Tried to synchronously call function from a diffrent thread.
+Solution is:
+a) if you want to synchronously execute this method, mark it as a worklet
+b) if you want to execute this method on the JS thread, wrap it using runOnJS
+)";
   
 void addHiddenProperty(jsi::Runtime &rt,
                        jsi::Value &&value,
@@ -214,7 +220,7 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
             const jsi::Value *args,
             size_t count
             ) -> jsi::Value {
-          module->errorHandler->setError("tried to synchronously call function from a diffrent thread");
+          module->errorHandler->setError(CALLBACK_ERROR);
           module->errorHandler->raise();
           return jsi::Value::undefined();
         };
@@ -256,7 +262,7 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
         };
         jsi::Function wrapperFunction = jsi::Function::createFromHostFunction(rt, jsi::PropNameID::forAscii(rt, "hostFunction"), 0, warnFunction);
         jsi::Function res = jsi::Function::createFromHostFunction(rt, jsi::PropNameID::forAscii(rt, "hostFunction"), 0, clb);
-        addHiddenProperty(rt, std::move(res), wrapperFunction, "callAsync");
+        addHiddenProperty(rt, std::move(res), wrapperFunction, "__callAsync");
         return wrapperFunction;
       }
     case ValueType::WorkletFunctionType:
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -206,7 +206,19 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
       } else {
         // function is accessed from a different runtme, we wrap function in host func that'd enqueue
         // call on an appropriate thread
+        
         auto module = this->module;
+        auto warnFunction = [module](
+            jsi::Runtime &rt,
+            const jsi::Value &thisValue,
+            const jsi::Value *args,
+            size_t count
+            ) -> jsi::Value {
+          module->errorHandler->setError("tried to synchronously call function from a diffrent thread");
+          module->errorHandler->raise();
+          return jsi::Value::undefined();
+        };
+        
         auto hostFunction = this->hostFunction;
         auto hostRuntime = this->hostRuntime;
         auto clb = [module, hostFunction, hostRuntime](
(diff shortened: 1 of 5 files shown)


Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -206,7 +206,19 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
...
         // call on an appropriate thread
+        
         auto module = this->module;
+        auto warnFunction = [module](
+            jsi::Runtime &rt,
+            const jsi::Value &thisValue,
+            const jsi::Value *args,
+            size_t count
+            ) -> jsi::Value {
+          module->errorHandler->setError("tried to synchronously call function from a diffrent thread");
```
kmagiera, line 217: We should suggest a solution here either by stating it in the error message or providing a link where they can see how to resolve the problem.

Solution is:
a) if you want to synchronously execute this method, mark it as a worklet
b) if you want to execute this method on the JS thread, wrap it using `runOnJS`

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -113,6 +113,22 @@ void RuntimeDecorator::addNativeObjects(jsi::Runtime &rt,
   jsi::Value measureFunction = jsi::Function::createFromHostFunction(rt, jsi::PropNameID::forAscii(rt, "_measure"), 1, clb4);
   rt.global().setProperty(rt, "_measure", measureFunction);
   
+  auto runOnJS = [](
```
kmagiera, line 116: do we need `runOnJS` to be defined here? I think it'd work better if it was defined in JS. This way user has to import it from reanimated, which allows us to:
1) avoid naming conflicts
2) eliminate lint warnings about method not being defined
3) allow us to provide TS typings
4) can also make it work when used from the JS thread (which may make sense because some worklets run both on JS and UI threads (I think we have it defined already)
5) will give us better flexibility on how we want to structure the way one can pass parameters

kmagiera (commented):
See my comments inline

Common/cpp/Tools/RuntimeDecorator.cpp
terrysahaidak: Should it return a promise with the result of execution?
Or the execution will always return void for now? I'm thinking about TS types.

Szymon20000: @kmagiera  It is exported. I just don't want to make another conditional statement inside `toJSValue` method.
If you want to I can check there if the function name is equal to `runOnJS`.

kmagiera: I think it still makes sense to have it in JS. There was a discussion whether we want this to return a function, or maybe pass arguments directly: `runOnJS(fun)(arg)` or `runOnJS(fun, arg)` – this would be easier for us to change if we had it in one place and not it two.

src/reanimated2/core.js
```diff
@@ -150,3 +151,19 @@ export function startMapper(mapper, inputs = [], outputs = []) {
...
+export let runOnJS = (fun) => fun;
+
+if (Platform.OS !== 'web') {
+  runOnJS = (fun) => {
+    'worklet';
+    if (!_WORKLET) { // eslint-disable-line
+      return fun;
+    }
+    if (!fun.__callAsync) {
+      throw new Error("Function couldn't be called by runOnJS");
```
kmagiera, line 164: Would prefer for error to be more descriptive than this. In this case we throw here because someone tried to us `runOnJS` on something that is not a host function. Something along these lines would give people better pointers:
"Attempting to call runOnJS with an object that is not a host function. Using runOnJS is only possible with methods that are defined on the main React-Native Javascript thread and that aren't marked as worklets"

kmagiera (commented):
Left some minor comments inline.

On top of that can you explain how are we handling console.log calls after this change? It'd seem to me that we should now start using `runOnJS` for console.logs, can we make it transparent for console.log, console.warn, etc?

In addition wanted to cc @terrysahaidak and @wcandillon on the discussion as to the final shape and name of this method. They had some suggestions about renaming this method or changing the structure such that you can pass args directly to `runOnSth` method (e.g., `runOnJS(callback, 'argument1', 2, 3)`)

src/reanimated2/core.js
```diff
@@ -150,3 +151,19 @@ export function startMapper(mapper, inputs = [], outputs = []) {
 export function stopMapper(mapperId) {
   NativeReanimated.stopMapper(mapperId);
 }
+
+export let runOnJS = (fun) => fun;
+
+if (Platform.OS !== 'web') {
```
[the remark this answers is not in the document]
Szymon20000: I think it can change in the future I would like to leave the check.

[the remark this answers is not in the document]
kmagiera: I think the fewer platform checks we make is better. This also makes an incentive in the future to add more checks like that which may lead to bigger platform-specific split.

src/reanimated2/core.js
```diff
@@ -150,3 +151,33 @@ export function startMapper(mapper, inputs = [], outputs = []) {
...
+
+const capturableConsole = console;
+runOnUI(() => {
+  'worklet';
+  const console = {
+    log: runOnJS(capturableConsole.log),
+    warn: runOnJS(capturableConsole.warn),
+    error: runOnJS(capturableConsole.error),
+  };
+  _globalSetter('console', console); // eslint-disable-line
```
kmagiera, line 182: Can we do `global.console = console` here? That's how you'd do it in regular JS so I'd prefer to use `global` here to keep it consistent. If not maybe we can enable the use of `global` instead of adding new function?

Szymon20000: That was my first approach but unfortunately, it's not working.

karol-bisztyga (approved):

karol-bisztyga:
Please add `globaSetter` on the web

terrysahaidak:
> In addition wanted to cc @terrysahaidak and @wcandillon on the discussion as to the final shape and name of this method. They had some suggestions about renaming this method or changing the structure such that you can pass args directly to `runOnSth` method (e.g., `runOnJS(callback, 'argument1', 2, 3)`)

I don't currently have a strong opinion that the signature is better here. Kinda both works.

But people used to work with redux-saga, for example, definitely used to do things like
`call(fetchThings, arg1, arg2)`.|

I didn't use sagas so I don't know.

The approach where we wrap function is good for composition but I don't think it isn't important here.

Both approaches can be type-safe with TS typings.

karol-bisztyga (changes requested):
`_globalSetter` on the web is not defined

karol-bisztyga (approved):

Issue #1253, opened 2020-09-20
useAnimatedStyle doesn't allow combining static styles


mrousavy:
## Description

The `useAnimatedStyle` hook only works when using purely animated values, so it cannot be used to merge additional static styles (e.g. from `StyleSheet.create`).

E.g.:

#### Doesn't work

```jsx
const headerStyle = useAnimatedStyle(() => [styles.header, { transform: [{ translateY: headerY.value }] }]);
// ...
return <Reanimated.View style={headerStyle} ... />
```

#### Works

```jsx
const headerStyle = useAnimatedStyle(() => ({ transform: [{ translateY: headerY.value }] }));
// ...
return <Reanimated.View style={[styles.header, headerStyle]} ... />
```

With "doesn't work" I mean that the transforms aren't applied at all. The `styles.header` values are correctly applied, but `translateY` just always stays `0`. (Even though the function is executed multiple times, checked with `console.log`!)

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.6

mrousavy:
So I've given this some more thought, and I can't wrap my head around if it's better to not allow that, or if that should be allowed.. 

To my understanding, the `useAnimatedStyle` hook is executed in the separate JS thread by reanimated everytime the `headerY` value changes, so every frame. We don't want to copy over the `styles.header` object and re-create those expensive arrays every frame, but that's what would happen, right?

If I combine it outside of the `useAnimatedStyle` hook (`<Reanimated.View style={[styles.header, animatedHeaderStyle]} ... />`), then the `styles.header` and `animatedHeaderStyle` objects are passed to the reanimated view by reference, making it a one time array allocation. Therefore it is best to write minimal code in the worklets and combine it afterwards with static stuff (styles), correct?

@kmagiera @jakub-gonet can you give me some insights if my answers are correct?

mrousavy:
Since alpha 7 there is an exception thrown when trying to combine statics styles with animated styles in the `useAnimatedStyle` hook:

![Simulator Screen Shot - iPhone 11 - 2020-09-24 at 15 23 17](https://user-images.githubusercontent.com/15199031/94150850-f00b6d00-fe79-11ea-92d6-ff1b77933096.png)

terrysahaidak:
useAnimatedStyle hook should return an object. then you should pass that object to styles. In that way, Reanimated Views can detect those and "subscribe" for changes.

[...]

PR #1276, opened 2020-09-24
Fix types for `useAnimatedStyle()`

mrousavy:
* Only allows style objects to be used, no more arrays, `false`, `undefined`, whatever

## Description

Related to #1253 

## Changes

* Added type `AnimatedStyleProp` which is essentially a stripped down version of react-native's `StyleProp` which also has animated properties inside. This new style prop disallows arrays (and `false, `undefined`, `null`) as return values for `useAnimatedStyle`


jakub-gonet:
@terrysahaidak, @wcandillon could you please check it? I'm not that familiar with RN types :/

@mrousavy, could you fix the conflict?

wcandillon:
@mrousavy ping me once the conflict is fixed and I will review it.

mrousavy:
@wcandillon 🏓

wcandillon:
if this syntax is allowed at runtime, that looks good 👍🏽

jakub-gonet:
Cool, @mrousavy could you also add some tests for this change to ts test file?

jakub-gonet:
@mrousavy 🏓

mrousavy:
@jakub-gonet @wcandillon  I've never written type-tests for TypeScript before, could you explain what you're expecting or at what file I'd have to look at?

jakub-gonet:
Yeah, sure.

We have [a TS test file](https://github.com/software-mansion/react-native-reanimated/blob/master/react-native-reanimated-tests.tsx) in the repository which just uses typings from main TS file. Please add some code that compiles and uses different types for `useAnimatedStyle` so we can verify that `tsc --check` passes.

mrousavy:
Thanks! I've added some tests, could you take a look and tell me if they're valid? `npm run type:check` ran successfully.

jakub-gonet:
Yeah, we had tests for happy paths, I like `@ts-expect-error` usage to test arguments of the wrong type, great idea.

I can merge it after CI passes (we check TS types there).
Thanks for your contribution!

jakub-gonet (approved):

PR #1280, opened 2020-09-25
[Housekeeping] Optimize Images in Example/ and docs/

mrousavy:
## Description

This is just a housekeeping PR. I've noticed we can reduce the size of the images in this repo by **22%**.

## Changes

- Lossless-compressed all images using  [ImgBot](https://imgbot.net)


jakub-gonet:
If I'm not mistaken this won't reduce repo size. Git stores every change as a new copy. This isn't a problem with text files, they can be efficiently compressed, but binary files can't and they're duplicated. This will increase repo size by 80% size of the images if merged.

PR #1287, opened 2020-09-28
Add node >=12 note to installation docs

ferrannp:
## Description

Without this, you just get a cryptic error of `Unexpected token =` that you don't know where it is coming from.

## Changes

Note in docs that reanimated plugin needs at least `node@12`.

### DIFF
--- a/docs/docs/installation.md
+++ b/docs/docs/installation.md
@@ -60,7 +60,8 @@ Add Reanimated's babel plugin to your `babel.config.js`:
   };
 ```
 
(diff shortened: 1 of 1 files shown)


jakub-gonet:
I'd also want to check if this error can be handled and make lower node versions compatible too.

wcandillon:
I concur that this should be indicated.

Szymon20000:
From now on, It works with older node versions thanks to https://github.com/software-mansion/react-native-reanimated/pull/1331.

PR #1288, opened 2020-09-28
Add web support note to the docs

karol-bisztyga:
## Description

Add web support note to the docs

### DIFF
--- a/docs/docs/web-support.md
+++ b/docs/docs/web-support.md
@@ -6,11 +6,11 @@ sidebar_label: Web Support
 
 Since
 [2.0.0-alpha.7](https://github.com/software-mansion/react-native-reanimated/releases/tag/2.0.0-alpha.7)
-has been released it is possible to launch reanimated 2 in a web browser. For that case all of the functionalities are implemented purely in javascript, hence the efficiency of the animations might drop.
+release it's possible to launch reanimated 2 in a web browser. For that case all of the functionalities are implemented purely in javascript, hence the efficiency of the animations might drop.
 
 If you use
 [playground](https://github.com/software-mansion-labs/reanimated-2-playground)
-and want to start the app in the browser, just type
+and want to start the app in the browser, just type:
 ```
 yarn web
 ```
--- a/docs/docs/web-support.md
+++ b/docs/docs/web-support.md
@@ -1,0 +1,23 @@
+---
+id: web-support
+title: Web Support
+sidebar_label: Web Support
+---
+
+Since
+[2.0.0-alpha.7](https://github.com/software-mansion/react-native-reanimated/releases/tag/2.0.0-alpha.7)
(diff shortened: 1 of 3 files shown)


docs/docs/web-support.md
```diff
@@ -0,0 +1,23 @@
+---
+id: web-support
+title: Web Support
+sidebar_label: Web Support
+---
+
+Since
```
jakub-gonet, line 7: I think we should mention Expo support too.

docs/docs/web-support.md
```diff
@@ -0,0 +1,23 @@
...
+[2.0.0-alpha.7](https://github.com/software-mansion/react-native-reanimated/releases/tag/2.0.0-alpha.7)
+has been released it is possible to launch reanimated 2 in a web browser. For that case all of the functionalities are implemented purely in javascript, hence the efficiency of the animations might drop.
```
jakub-gonet, line 9:
```suggestion
release it's' possible to launch reanimated 2 in a web browser. On the web version, all of the functionalities are implemented purely in javascript and the efficiency of the animations might drop.
```

docs/docs/web-support.md
```diff
@@ -0,0 +1,23 @@
...
+
+If you use
+[playground](https://github.com/software-mansion-labs/reanimated-2-playground)
+and want to start the app in the browser, just type
```
jakub-gonet, line 13: Missing colon.

docs/docs/web-support.md
karol-bisztyga: I'm ok with that, could be another PR though

karol-bisztyga: or do you mean that we fire up web app using expo command?

docs/docs/web-support.md
karol-bisztyga: Well I'm not sure if those changes are good:
`release it's'` - that second apostrophe look like an error, but as for those shortcuts it's ok
`On the web version` - `web` word repetition, not that it's that bad, just not a good change for me
`and` - `hence` just suits better I think

docs/docs/web-support.md
jakub-gonet: No, I meant that we should mention that Expo supports the experimental version of Rea2 in the native and web flavour

docs/docs/web-support.md
jakub-gonet: 1. > release it's' - that second apostrophe look like an error
Yeah, fixed

2. & 3. Good point, what you think about "All of the web functionalities are implemented purely in javascript which may decrease the efficiency of animations."

karol-bisztyga: Doesn't really matter to me, basically you could rewrite all of the content since every piece of information can be passed using alternative words.

jakub-gonet (approved):

PR #1298, opened 2020-09-30
Add isNaN to the globals

JonnyBurger:
Make `isNaN` available inside a worklet.

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -50,6 +50,7 @@ const globals = new Set([
   'global',
   '_measure',
   '_scrollTo',
+  'isNaN',
 ]);
 
 // leaving way to avoid deep capturing by adding 'stopCapturing' to the blacklist


jakub-gonet (commented):
What with other things defined as [globals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects)?

In particular:
- NaN
- Infinity
- arguments

and [all functions listed under "Function properties"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects).

kmagiera (approved):

kmagiera:
merging this one, it appears however that we should work on changing this logic to something more comprahensive. i.e. getting the list from globals directly via Object.getOwnPropertyNames(global) (will need to filter out some of the props as plugin runs on a node environment and exports globally things like os, net, http modules)

PR #1299, opened 2020-09-30
Animate angles and colors.

Szymon20000:
## WHY
In order to let users for:
```JS
withTiming('0deg')
withTiming('red')
anyAnimation('color')
anyAnimation('prefixNumberSuffix') // returns 'prefix' + currentAnimationValue + 'suffix'
```

## How
Decorate all animations and add check  `onStart` if we have one of the above cases if the answer is yes then properly decorate onFrame. If the previous value of shared value is color then replace onFrame with a function that simulates 3 animations (one per HSV coordinate). If the previous value matches the` PrefixNumberSuffix` pattern then save prefix and suffix in the animation and animate raw value.

### DIFF
--- a/src/reanimated2/Colors.js
+++ b/src/reanimated2/Colors.js
@@ -28,6 +28,7 @@ const uiCachedMatchers = makeRemote({});
 function getMatchers() {
   'worklet';
   const cachedMatchers = _WORKLET ? uiCachedMatchers : jsCachedMatchers;
+  console.log("typ", _WORKLET);
   if (cachedMatchers.rgb === undefined) {
     cachedMatchers.rgb = new RegExp('rgb' + call(NUMBER, NUMBER, NUMBER));
     cachedMatchers.rgba = new RegExp(
@@ -413,6 +414,7 @@ export function processColorInitially(color) {
 }
 
 export function isColor(value) {
+  'worklet'
   return processColorInitially(value) 
 }
 
--- a/src/reanimated2/Colors.js
+++ b/src/reanimated2/Colors.js
@@ -392,7 +392,7 @@ function normalizeColor(color) {
   return null;
 }
 
-export default function processColor(color) {
+export function processColorInitially(color) {
   'worklet';
   if (color === null || color === undefined || typeof color === 'number') {
     return color;
@@ -408,7 +408,28 @@ export default function processColor(color) {
     return null;
   }
 
-  normalizedColor = ((normalizedColor << 24) | (normalizedColor >>> 8)) >>> 0;
+  normalizedColor = ((normalizedColor << 24) | (normalizedColor >>> 8)) >>> 0; //argb
+  return normalizeColor;
+}
+
+export function isColor(value) {
+  return processColorInitially(value) 
+}
+
+export function to(value) {
+
+}
+
+export default function processColor(color) {
+  'worklet';
+  const normalizedColor = processColorInitially(color);
+  if (normalizedColor === null || normalizedColor === undefined) {
+    return undefined;
+  }
+
+  if (typeof normalizedColor !== 'number') {
+    return null;
+  }
 
   if (Platform.OS === 'android') {
     // Android use 32 bit *signed* integer to represent the color
--- a/src/reanimated2/Colors.js
+++ b/src/reanimated2/Colors.js
@@ -408,21 +408,21 @@ export function processColorInitially(color) {
     return null;
   }
 
-  normalizedColor = ((normalizedColor << 24) | (normalizedColor >>> 8)) >>> 0; //argb
+  normalizedColor = ((normalizedColor << 24) | (normalizedColor >>> 8)) >>> 0; // argb
   return normalizedColor;
 }
 
 export function isColor(value) {
-  'worklet'
+  'worklet';
   if (typeof value !== 'string') {
     return false;
   }
-  return processColorInitially(value) != null; 
+  return processColorInitially(value) != null;
 }
 
 export default function processColor(color) {
   'worklet';
-  const normalizedColor = processColorInitially(color);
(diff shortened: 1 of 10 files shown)


src/reanimated2/animations.js
```diff
@@ -422,11 +644,11 @@ export function withRepeat(
   return defineAnimation(_nextAnimation, () => {
     'worklet';
 
-    const nextAnimation =
+    let nextAnimation =
```
kmagiera, line 647: why?

src/reanimated2/animations.js
```diff
@@ -1,37 +1,262 @@
...
+
+    transformAnimation(animation);
+    if (previousAnimation != animation) transformAnimation(previousAnimation);
+  };
+  const prefNumberSuffOnFrame = (animation, timestamp) => {
+    transformAnimation(animation);
+
+    const res = baseOnFrame(animation, timestamp);
+
+    transformAnimation(animation);
```
kmagiera, line 188: doesn't seem to make a lot of sense to transform toValue and startValue on each frame, shouldn't it suffice to transform them on start?

src/reanimated2/animations.js
```diff
@@ -1,37 +1,262 @@
...
+  const prefNumberSuffOnStart = (animation, value, timestamp, previousAnimation) => {
+
+    const val = transform(value, animation);
+    transformAnimation(animation);
+    if (previousAnimation != animation) transformAnimation(previousAnimation);
+
+    baseOnStart(animation, val, timestamp, previousAnimation);
+
+    transformAnimation(animation);
+    if (previousAnimation != animation) transformAnimation(previousAnimation);
```
kmagiera, line 181: what's the point of transforming the previos animation? Aren't we done with it at this point?

src/reanimated2/animations.js
```diff
@@ -1,37 +1,262 @@
...
+    transformAnimation(animation);
+
+    const res = baseOnFrame(animation, timestamp);
+
+    transformAnimation(animation);
+    return res;
+  };
+
+  const tab = ['H', 'S', 'V', 'A'];
+  const colorOnStart = (animation, value, timestamp, previousAnimation) => {
```
kmagiera, line 193: would be nice to extract color-related helpers out of this file, it is already rather lengthy and adding so much stuff that is only relevant to some of the usecases does not seem to improve readability

src/reanimated2/animations.js
```diff
@@ -1,37 +1,262 @@
...
+  return handler.__prefix + value + handler.__suffix;
+}
+
+const needsToBeChanged = ['toValue', 'current', 'startValue'];
+export function transformAnimation(animation) {
+  'worklet';
+  if (!animation) {
+    return;
+  }
+  needsToBeChanged.forEach((prop) => {
```
kmagiera, line 132: I commented in the other place that it may not be worth transforming all the values on each frame. I'm not a fan of loops like this here and would prefer to just have three lines:
```
animation.toValue = transform(animation.toValue, animation);
animation.current = transform(animation.current, animation);
animation.startValue = transform(animation.startValue, animation);
```
which is also more efficient (can actually be jitted and saves some allocations). But that isn't as much important as the fact that I think we don't really need to transform all those three props every frame.

kmagiera (commented):
left a couple of comments

src/reanimated2/animations.js
Szymon20000: All animations are transformed before and after calling onStart/onFrame. `transformAnimation` function is an Involution. transformAnimation(transformAnimation(x)) = x. It helps with higher order animations like withRepeat that need string values.

src/reanimated2/Colors.js
```diff
@@ -408,7 +408,28 @@ export default function processColor(color) {
...
+  normalizedColor = ((normalizedColor << 24) | (normalizedColor >>> 8)) >>> 0; //argb
+  return normalizedColor;
+}
+
+export function isColor(value) {
+  'worklet'
+  if (typeof value !== 'string') {
+    return false;
+  }
+  return processColorInitially(value) != null;
```
karol-bisztyga, line 420: in `processColorInitially` we perform redundant type checks as here we are sure that `value` is of type `string`, the same deeper in the callstack in `normalizeColor`, plus in `normalizeColor` we perform `getMatchers` before type checks, I think that should be moved after them so in case they're not passed matchers are not obtained(I know this is not related to this PR's changes but still could be improved)

src/reanimated2/animations.js
```diff
@@ -1,37 +1,262 @@
...
+
+export function decorateAnimation(animation) {
+  'worklet';
+  if (animation.isHigherOrder) {
+    return;
+  }
+  const baseOnStart = animation.onStart;
+  const baseOnFrame = animation.onFrame;
+  const animationCopy = Object.assign({}, animation); 
+  delete animation.callback;
```
karol-bisztyga, line 170: Why the callback is being deleted here? What about callbacks? Do they work with those changes applied?

Szymon20000: Oh, nice catch. I wanted to remove it from the copy
.

src/reanimated2/animations.js
karol-bisztyga: operation on colors have been moved to `Colors.js`. Here these two functions also contain operations on animations, I'd leave them here

src/reanimated2/animations.js
karol-bisztyga: fixed ec76ca0

src/reanimated2/animations.js
karol-bisztyga: fixed ec76ca0

karol-bisztyga (approved):

PR #1300, opened 2020-09-30
Updated docs fundamentals to include imports

Redseb:
## Description

Updates docs to include imports in most examples under Fundamentals, and fixes an incorrect import under Fundamentals>Animations>Spring. The error is 
`import {
  withSpring,
  useAnimatedStyle,
  useSharedValue,
  Animated,
} from 'react-native-reanimated';`
when it should be
`import Animated, {
  withSpring,
  useAnimatedStyle,
  useSharedValue,
} from 'react-native-reanimated';`


kmagiera (approved):
thanks

PR #1301, opened 2020-10-01
Fix dead link to v1 docs

idolize:
The original link leads to a 404 page

## Description

The previous link leads to a 404: https://docs.swmansion.com/react-native-reanimated/docs/1.x.x/

## Changes

Update the link to match the page from using the dropdown on the v2 docs: https://docs.swmansion.com/react-native-reanimated/docs/


jakub-gonet:
Thanks!

jakub-gonet (approved):

PR #1305, opened 2020-10-02
Add color interpolation

karol-bisztyga:
## Description

Add color interpolation, taken from https://github.com/wcandillon/react-native-redash/blob/master/src/Colors.ts

### DIFF
--- a/src/reanimated2/Colors.js
+++ b/src/reanimated2/Colors.js
@@ -395,6 +395,144 @@ function normalizeColor(color) {
   return null;
 }
 
+export const ColorSpace = Object.freeze({
+  RGB: 'rgb',
+  HSV: 'hsv',
+});
+
+export const opacity = (c) => {
+  'worklet';
+  return ((c >> 24) & 255) / 255;
+};
+
+export const red = (c) => {
+  'worklet';
+  return (c >> 16) & 255;
+};
+
+export const green = (c) => {
+  'worklet';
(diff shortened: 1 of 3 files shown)


DmitryFt:
Your solution is not going for Android

karol-bisztyga:
Just built it locally and it worked fine. I don't think it's possible that this PR ruins build like that as changes are only made in javascript.
Also current alpha is 7.
Jakub's got a point, try to rebuild the project.

src/reanimated2/Colors.js
```diff
@@ -420,3 +423,200 @@ export default function processColor(color) {
...
+  };
+};
+
+export const hsv2color = (h, s, v) => {
+  'worklet';
+  const { r, g, b } = hsv2rgb(h, s, v);
+  return rgbaColor(r, g, b);
+};
+
+export const colorForBackground = (r, g, b) => {
```
Szymon20000, line 518: Why do we need this?

Szymon20000 (approved):
Looks fine! Left only one question.

nhannah:
Can anyone confirm they have the alpha channel working with this interpolateColor function? Without alpha I have no issue, but adding an alpha I get a rainbow of colors depending on the value. @wcandillon Have you had this issue? I see it came from your redash repo originally.

jakub-gonet:
@nhannah, could you open a new issue with some repro so we can verify it?

PR #1210, opened 2020-09-08
Support wildcard type ES6 imports

Ragnar-H:
## Description

Addresses this comment here: https://github.com/software-mansion/react-native-reanimated/issues/823#issuecomment-688721886

Widens the support for identifying the reanimated2 hooks to also support BuckleScript type imports

```
import * as Reanimated from 'react-native-reanimated'

Reanimated.useAnimatedStyle(...)
```

The tests covering the babel plugin are pretty bare-bones, so I'm happy to extend those tests, or perform any manual tests needed before this goes any further.

## Changes

- Support wildcard ES6 import style with babel plugin

PR #1310, opened 2020-10-06
Add another ways to mark function as worklet.

Szymon20000:
The pull-request adds two methods which are the new ways for telling babel that is should transform the function into a worklet. The first is `useWorkletCallback`:
```JS
useWorkletCallback(() => {...});
```
which is equivalent to
```JS
useCallback(() => {
  'worklet'
})
``` 

The second one is `createWorklet` which just lets us omit 'worklet' mark.
```JS
const worklet = createWorklet(() => {
  ...
});
```
behaves the same as
```
const worklet = () => {
   'worklet'
    ...
}

### DIFF
--- a/react-native-reanimated-tests.tsx
+++ b/react-native-reanimated-tests.tsx
@@ -22,6 +22,9 @@ import Animated, {
   repeat,
   sequence,
   withDecay,
+  useWorkletCallback,
+  createWorklet,
+  runOnUI,
 } from 'react-native-reanimated';
 
 const styles = StyleSheet.create({
@@ -359,4 +362,34 @@ function WithDecayTest() {
       <Animated.View style={[styles.box, animatedStyle]} />
     </PanGestureHandler>
   );
+}
+
+// useWorkletCallback
+function UseWorkletCallbackTest() {
+  const callback = useWorkletCallback((a: number, b: number) => {
+    return a + b;
+  });
+
+  runOnUI(() => {
+    const res = callback(1, 1);
+
+    console.log(res);
+  })();
+
+  return <Animated.View style={styles.container} />;
+}
+
+// createWorklet
+function CreateWorkletTest() {
+  const callback = createWorklet((a: number, b: number) => {
+    return a + b;
+  });
+
+  runOnUI(() => {
+    const res = callback(1, 1);
+
+    console.log(res);
+  })();
+
+  return <Animated.View style={styles.container} />;
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/reanimated2/Hooks.js [hunk 1]
```diff
@@ -594,3 +594,11 @@ export function useAnimatedReaction(prepare, react) {
     };
   }, inputs);
 }
+
+export function useWorkletCallback(fun) {
```
terrysahaidak, line 598: You're missing deps here

terrysahaidak (changes requested):

src/reanimated2/Hooks.js
Szymon20000: Oh, true.

terrysahaidak:
I'll add TS typings in the following PR.

kmagiera (approved):

react-native-reanimated.d.ts
```diff
@@ -445,6 +445,10 @@ declare module 'react-native-reanimated' {
       fn: (...args: A) => R
     ): (...args: Parameters<typeof fn>) => void;
     export function processColor(color: number | string): number;
+    export function createWorklet<A extends any[], R>(
+      fn: (...args: A) => R,
+      deps?: DependencyList
```
terrysahaidak, line 450: Just realized I put deps here too
```suggestion
      fn: (...args: A) => R
```

react-native-reanimated-tests.tsx
```diff
@@ -360,3 +363,33 @@ function WithDecayTest() {
     </PanGestureHandler>
   );
 }
+
+// useWorkletCallback
+function UseWorkletCallbackTest() {
+  const callback = useWorkletCallback((a: number, b: number) => {
+    return a + b;
+  });
```
terrysahaidak, line 371: I should've to add deps here too.

```suggestion
  }, []);
```

jfrolich:
A note to make sure that it doesn't fail with this syntax:

```es6
import * as Reanimated from 'reanimated'

Reanimated.createWorklet(() => {
  ...
})
```

terrysahaidak:
https://github.com/software-mansion/react-native-reanimated/pull/1210 This PR should fix this type of imports as well. 

cc @jakub-gonet

mrousavy:
While I'm a big fan of this idea, I think it would be a better name to shorten `useWorkletCallback` to `useWorklet`. A 'worklet' if per definition something like a callback anyways, no?

Thoughts @kmagiera @terrysahaidak @Szymon20000 ?

Issue #1284, opened 2020-09-26
iOS: Full app crash


mrousavy:
## Description

I'm receiving the following error on iOS:

```
com.facebook.react.JavaScript (17): EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
```

I've implemented a stories screen using react-native-navigation, and see this error when I quickly close the story (= unmount the screen) while (or shortly before) I run an animation with reanimated. In this case it's a `withTiming` worklet.

This is my call stack at the time of the crash:

<img src="https://user-images.githubusercontent.com/15199031/94345704-31417f80-0028-11eb-9d80-1b9b24ede548.png" width="70%">

## Steps To Reproduce

1. Create app using react-native-navigation and reanimated
2. Open and close stories like a maniac
3. At some point, close the story (and therefore unmount the screen) exactly when a worklet is being started, not sure it it's caused by a timing function, or the timing function having a callback and then unmounting the Reanimated View.

### Expected behavior

I expect the story to close and any reanimated worklets/animations/state to be disposed

### Actual behavior

The app crashes with the error `com.facebook.react.JavaScript (17): EXC_BAD_ACCESS (code=EXC_I386_GPFLT)`

![Screenshot 2020-09-26 at 18 44 27](https://user-images.githubusercontent.com/15199031/94345716-50401180-0028-11eb-9b8d-5d5a86fc03d2.png)

## Snack or minimal code example

Working on this. Have to strip out a lot of things, not sure if I can easily replicate it using a simple example...

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha7
- React Native Navigation: 7.0.0-snapshot2020

Ghya:
I have the same issue with a header, animated while scrolling
But the same animation was working 2 days ago, before react-native upgrade (0.63.0 -> 0.63.2)

JB-CHAUVIN:
Hello, same issue goes for me on RN 0.63.2 and Reanimated 2 Alpha 7.

Animated.ScrollView with scrollHandler and complex animation make the app freeze...

ajpaulingalls:
I'm also getting an iOS app freeze when regular animations run alongside reanimations.  Specifically, the main thread is stuck waiting for a semaphore on line 227 of REANodesManager.m, in the performOperations method:

```
dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
```

jpudysz:
[...]

PR #1315, opened 2020-10-06
Make StoreUser concurrent.

Szymon20000:
I believe the problems mentioned in https://github.com/software-mansion/react-native-reanimated/issues/1284 are caused by parallel read/writes operations made on hash map. We can create shared value on different threads so we should also take care of concurrent updates.


jpudysz:
+1 
After upgrade from @alpha.6 to @alpha.7 app freezes eg. on `Keyboard.close()` etc.

kmagiera (approved):

PR #1319, opened 2020-10-07
Fix event handling

Szymon20000:
Copied from https://github.com/software-mansion/react-native-reanimated/pull/1149
## Description

Previously, we were handling dispatched events like so:
```java
@Override
 public void onEventDispatch(Event event) {
   if (UiThreadUtil.isOnUiThread()) {
     handleEvent(event);
   } else {
     mEventQueue.offer(event);
     startUpdatingOnAnimationFrame();
    }
  }
```

Event handling in RN works by utilizing `EventDispatcher.dispatchEvent(event)` which
1. runs `onEventDispatch` callback on any registered listener
2. adds `event` to internal event queue
3. adds frame callback which dispatches and disposes events on JS thread

This approach introduced timing issues - RN's `EventDispatcher` dispatches events on JS thread and Reanimated handles events on UI thread. There's a possibility that `EventDispatcher` will dispose event (possibly destroying it's state in `onDispose()`) before Reanimated would have chance to handle it.
This was found after investigating [pretty popular crash in react-native-gesture-handler](https://github.com/software-mansion/react-native-gesture-handler/pull/1171).

# HOW
The pull-request adds another method `isAnyHandlerWaitingForEvent` to NativeProxy API which lets us check if an event is important (there is workletHandler listening for the event) or not. The rest part of the pr is very similar to Jakub's pr. However, there are some differences. Instead of saving copied event Object, we save: tag, eventName, and payload in the new class `CopiedEvent`.

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -60,6 +60,8 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import javax.annotation.Nullable;
+
 public class NodesManager implements EventDispatcherListener {
 
   private static final Double ZERO = Double.valueOf(0);
@@ -104,7 +106,7 @@ public class NodesManager implements EventDispatcherListener {
 
   private RCTEventEmitter mCustomEventHandler;
   private List<OnAnimationFrame> mFrameCallbacks = new ArrayList<>();
-  private ConcurrentLinkedQueue<Event> mEventQueue = new ConcurrentLinkedQueue<>();
+  private ConcurrentLinkedQueue<CopiedEvent> mEventQueue = new ConcurrentLinkedQueue<>();
   private boolean mWantRunUpdates;
 
   public double currentFrameTimeMs;
@@ -228,7 +230,8 @@ public class NodesManager implements EventDispatcherListener {
     currentFrameTimeMs = frameTimeNanos / 1000000.;
 
     while (!mEventQueue.isEmpty()) {
-      handleEvent(mEventQueue.poll());
+      CopiedEvent copiedEvent = mEventQueue.poll();
+      handleEvent(copiedEvent.getTargetTag(), copiedEvent.getEventName(), copiedEvent.getPayload());
     }
 
     if (!mFrameCallbacks.isEmpty()) {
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -445,7 +448,13 @@ public void onEventDispatch(Event event) {
...
       performOperations();
     } else {
-      mEventQueue.offer(event);
+      boolean shouldSaveEvent = false;
+      String eventName = mCustomEventNamesResolver.resolveCustomEventName(event.getEventName());
+      int viewTag = event.getViewTag();
+      String key = viewTag + eventName;
+
+      shouldSaveEvent |= (mCustomEventHandler != null && mNativeProxy.isAnyHandlerWaitingForEvent(key));
+      mEventQueue.offer(new CopiedEvent(event));
```
jakub-gonet, line 457: Copied from #1149 
> Maybe we could also pool CopiedEvent objects same way we have pools of other events. In this case this should be rather efficient as there typically won't be more than 2-3 events in use.

Szymon20000: I don't see a need for a CopiedEvent pool.

jakub-gonet: Not using the pool will strain GC, event objects will be created and removed frequently if events will be coming from other source than the UI thread.

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -228,7 +230,8 @@ private void onAnimationFrame(long frameTimeNanos) {
     currentFrameTimeMs = frameTimeNanos / 1000000.;
 
     while (!mEventQueue.isEmpty()) {
-      handleEvent(mEventQueue.poll());
+      CopiedEvent copiedEvent = mEventQueue.poll();
+      handleEvent(copiedEvent.getTargetTag(), copiedEvent.getEventName(), copiedEvent.getPayload());
```
jakub-gonet, line 234: You should call `copiedEvent.onDispose()` here to not break the Event contract.

android/src/main/java/com/swmansion/reanimated/NodesManager.java
Szymon20000: It doesn't make sense for me because we still have to create and remove events' payloads.

android/src/main/java/com/swmansion/reanimated/NodesManager.java
Szymon20000: CopiedEvent is just a plain object and doesn't have onDispose method.

jakub-gonet: I see, my bad.

android/src/main/java/com/swmansion/reanimated/NodesManager.java
jakub-gonet: So maybe @kmagiera will chime in since he suggested this change in the original PR.

karol-bisztyga (approved):

PR #1320, opened 2020-10-07
Rename `useWorkletCallback` to `useWorklet`

mrousavy:
## Description

While I'm a big fan of the hook syntax for worklets, I think it would be a better name to shorten `useWorkletCallback` to `useWorklet`. A 'worklet' if per definition something like a callback anyways, no?

## Changes

- Renamed `useWorkletCallback` to `useWorklet`

## Reference

https://github.com/software-mansion/react-native-reanimated/pull/1310#issuecomment-704898870

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -15,7 +15,7 @@ const functionHooks = new Map([
   ['useDerivedValue', [0]],
   ['useAnimatedScrollHandler', [0]],
   ['useAnimatedReaction', [0, 1]],
-  ['useWorkletCallback', [0]],
+  ['useWorklet', [0]],
   ['createWorklet', [0]],
 ]);
 
--- a/react-native-reanimated-tests.tsx
+++ b/react-native-reanimated-tests.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable */
 import React from 'react';
 import { StyleSheet, Button, View } from 'react-native';
-import { 
+import {
   PanGestureHandler,
   PinchGestureHandlerGestureEvent,
   PinchGestureHandler,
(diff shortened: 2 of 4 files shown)


likern:
> ## Description
> While I'm a big fan of the hook syntax for worklets, I think it would be a better name to shorten `useWorkletCallback` to `useWorklet`. A 'worklet' if per definition something like a callback anyways, no?
> 
> ## Changes
> * Renamed `useWorkletCallback` to `useWorklet`
> 
> ## Reference
> [#1310 (comment)](https://github.com/software-mansion/react-native-reanimated/pull/1310#issuecomment-704898870)

Agree, as TypeScript types already show input / output types - no reason to code this into hook name.

kmagiera:
I don't have any preference here, will let @terrysahaidak chime in who was the first who suggested adding this hook helper.

terrysahaidak:
I don't have any preference here, initially, I thought about it as a strong reference to useCallback since it's just the same thing. But fewer characters may be easier to type. So yeah, can be useWorklet instead of useWorkletCallback.

mrousavy:
@kmagiera @terrysahaidak is that a yes? I guess it's too late to be included in alpha 8

jakub-gonet:
Since this is the only recommended way of creating a callback I'd go with the shortened name.

karol-bisztyga:
This has been hanging for a while and still there is no decision regarding this.

We want to avoid changing API without a really good reason as it introduces confusion - now we would have to keep the old names and at the same time add the new ones to keep backward compatibility.

Therefore I'm closing this for now, thanks @mrousavy for coming up with the idea, if you feel uncomfortable with that API you can always use aliases to make the Reanimated 2 experience better for yourself ;)

PR #1323, opened 2020-10-08
Simplify the iOS installation process and turn off turbo modules for other modules than Reanimated.

Szymon20000:
This pr makes the installation of Reanimated on iOS similar to the mechanism we already used on Android. I've found a place where we can add Reanimated module to the global object (just like we do on Android). The changes let us install automatically on iOS. Another advantage of this approach is avoiding turbo versions of react-native libraries. It's important because some packages on npm are not ready yet for turbo versions. (e.g. react-navigation).

# HOW
We add the category of UIResponder class which is the parent class of `AppDelegate` which is used as a bridge delegate.
In this way, we don't need to know the name of `AppDelegate `class. As a result, we can add `jsExecutorFactoryForBridge` method to BridgeDelegate and install there reanimated module.  Additionally, we check if someone set flag `DONT_AUTOINTSTALL_REANIMATED`. If so we don't add category.


docs/docs/installation.md
```diff
@@ -109,123 +109,13 @@ You can refer [to this diff](https://github.com/software-mansion-labs/reanimated
...
 
-```objectivec {3}
-  + (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
-  {
-    RCTEnableTurboModule(YES); // <- add
-```
+> **_NOTE:_** In previous releases, the installation process was manual and required turning turbo modules on.  Because we had 
+> reported some issues regarding bugs in turbo versions of react-native libraries we decided to changed our approach and we no longer 
+> use the standard way for registering a turbo module. It let us simplify the installation process and as a result, you can safely 
+> undo all installation steps from the previous instruction [previous instruction](https://docs.swmansion.com/react-native-reanimated/docs/installation#ios).
```
kmagiera, line 117:
```suggestion
> undo all installation steps from the [previous instruction](https://docs.swmansion.com/react-native-reanimated/docs/installation#ios).  
```

kmagiera: apparently the last stable version is not available under /docs/2.0.0-alpha.7 link – so the link here will no longer point to the correct instruction when we publish next version

kmagiera (approved):
Looks good, some minor comments added inline

osdnk:
Now it will be possible to use it with a Chrome debugger. At the same time, reanimated won't be working with V8 so I guess it might be worth mentioning that it's not a bug, @Szymon20000

docs/docs/installation.md
```diff
@@ -109,123 +109,24 @@ You can refer [to this diff](https://github.com/software-mansion-labs/reanimated
...
+
+> **_NOTE:_** In previous releases, the installation process was manual and required turning turbo modules on.  Because we had 
+> reported some issues regarding bugs in turbo versions of react-native libraries we decided to changed our approach and we no longer 
+> use the standard way for registering a turbo module. It let us simplify the installation process and as a result, you can safely 
+> undo all installation steps from the [previous instruction](https://docs.swmansion.com/react-native-reanimated/docs/installation#ios).  
+
+> **_NOTE:_** If you want to turn off autoinstall on iOS please add the following compilation flag: 
+> `DONT_AUTOINTSTALL_REANIMATED`.
+> It can be done by pasting: 
+> ```JS
```
jakub-gonet, line 122:
```suggestion
> ```js
```

docs/docs/installation.md
```diff
@@ -109,123 +109,24 @@ You can refer [to this diff](https://github.com/software-mansion-labs/reanimated
...
+> post_install do |installer|
+>    installer.pods_project.targets.each do |target|
+>        target.build_configurations.each do |config|
+>            config.build_settings['OTHER_CPLUSPLUSFLAGS'] = '-DDONT_AUTOINTSTALL_REANIMATED'
+>        end
+>    end
+> end
+> ```
+> to your `Podfile`. (Don't forget to run `pod install` after that).
```
jakub-gonet, line 131:
```suggestion
> to your `Podfile`. Don't forget to run `pod install` after doing that.
```

docs/docs/installation.md
```diff
@@ -109,123 +109,24 @@ You can refer [to this diff](https://github.com/software-mansion-labs/reanimated
...
-```objectivec {3-3}
-- (std::unique_ptr<facebook::react::JSExecutorFactory>)jsExecutorFactoryForBridge:(RCTBridge *)bridge
-{
-  _turboModuleManager = [[RCTTurboModuleManager alloc] initWithBridge:bridge delegate:self];
-  ...
-```
+On iOS installation is automatic. 
+
+> **_NOTE:_** In previous releases, the installation process was manual and required turning turbo modules on.  Because we had 
+> reported some issues regarding bugs in turbo versions of react-native libraries we decided to changed our approach and we no longer
```
jakub-gonet, line 115:
```suggestion
> **_NOTE:_** In previous releases, the installation process was manual and required turning turbo modules on. Some libraries break when turbo modules are enabled so we decided to change our approach and we no longer 
```

jakub-gonet (commented):
Left some suggestions

ios/native/UIResponder+Reanimated.mm
```diff
@@ -0,0 +1,34 @@
...
+  __weak __typeof(self) weakSelf = self;
+  return std::make_unique<facebook::react::JSCExecutorFactory>([weakSelf, bridge](facebook::jsi::Runtime &runtime) {
+    if (!bridge) {
+      return;
+    }
+    __typeof(self) strongSelf = weakSelf;
+    if (strongSelf) {
+      auto reanimatedModule = reanimated::createReanimatedModule(bridge.jsCallInvoker);
+      runtime.global().setProperty(runtime,
+                                   jsi::PropNameID::forAscii(runtime, "__reanimatedModuleProxy"),
```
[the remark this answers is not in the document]
Szymon20000: I want to leave it for now. We use it in the Expo client.

PR #1331, opened 2020-10-10
fix: react-native compilation issues with node 10.x.x

vip-git:
## Description
Fixes below compilation issue
image

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -109,7 +109,9 @@ const blacklistedFunctions = new Set([
 ]);
 
 class ClosureGenerator {
-  trie = [{}, false];
+  constructor() {
+    this.trie = [{}, false];
+  }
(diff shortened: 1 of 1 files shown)


terrysahaidak:
Could you please tell us what version of NodeJS you're using?

vip-git:
Hi @terrysahaidak - Its `v10.15.1` .

Thanks

vip-git:
Hi @terrysahaidak - i tried to update to node `12.13.0` and it seems to have fixed issue - good catch.

It looks in that case `10.15.1` was in fault we can patch it by adding the variable initialisation inside the constructor that works like charm you can replicate it to confirm.

Thanks

Szymon20000 (approved):

PR #1338, opened 2020-10-12
Add deps to useAnimatedReaction

### LINKED ISSUE #1317: Re-render react function parameter when using useAnimatedReaction
## Description

How can I trigger re-render of the react function parameter when using `useAnimatedReaction` ? 

## Code

This code works, `items` will be updated

```
 const [items, setItems] = useState<number>([])
 const height = useSharedValue<number>(0)
 const prevHeight = useSharedValue<number>(0)

 useDerivedValue(() => {
        const diffHeight = height.value - prevHeight.value
        if (diffHeight !== 0) {
            items.map((p) => {
                if (p.id !== id && p.offset.value > offset.value) {
                    p.offset.value = p.offset.value + diffHeight
                }
            })
            prevHeight.value = height.value
        }
        return height.value
    }, [items])

```

`items` will always be empty here

```
    useAnimatedReaction(
        () => height.value,
        (newHeight: number) => {
            const diffHeight = newHeight - prevHeight.value
            if (diffHeight !== 0) {
                items.map((p) => {
                    if (p.id !== id && p.offset.value > offset.value)
                        p.offset.value = p.offset.value + diffHeight
                })
                prevHeight.value = newHeight
            }
        }
    )
```

I tried using useCallBack on the `react` function or passing `items` as input but it does not work either

### Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.7

terrysahaidak:
## Description

Adds deps to useAnimatedReaction hook.

Fixes #1317 

## Changes

- Add optional dependency list to useAnimatedReaction hook

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -583,7 +583,12 @@ export function useAnimatedReaction(prepare, react, dependencies) {
   const { inputs } = inputsRef.current;
 
   if (dependencies === undefined) {
-    dependencies = [...inputs, prepare.__workletHash, react.__workletHash];
+    dependencies = [
+      Object.values(prepare._closure),
+      Object.values(react._closure),
+      prepare.__workletHash,
+      react.__workletHash,
+    ];
   } else {
     dependencies.push(prepare.__workletHash, react.__workletHash);
   }
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -573,7 +573,7 @@ export function useAnimatedRef() {
  * the first worklet defines the inputs, in other words on which shared values change will it be called.
  * the second one can modify any shared values but those which are mentioned in the first worklet. Beware of that, because this may result in endless loop and high cpu usage.
  */
-export function useAnimatedReaction(prepare, react) {
+export function useAnimatedReaction(prepare, react, dependencies) {
   const inputsRef = useRef(null);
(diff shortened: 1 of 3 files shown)


src/reanimated2/Hooks.js
```diff
@@ -582,6 +582,12 @@ export function useAnimatedReaction(prepare, react) {
   }
   const { inputs } = inputsRef.current;
 
+  if (dependencies === undefined) {
+    dependencies = [...inputs, prepare.__workletHash, react.__workletHash];
```
karol-bisztyga, line 586: this is invalid, consider following code:
```
if (dependencies === undefined) {
  const depInputs = Object.values(prepare._closure).concat(Object.values(react._closure));
  dependencies = [...depInputs, prepare.__workletHash, react.__workletHash];
} else {
  dependencies.push(prepare.__workletHash, react.__workletHash);
}
```
Pulling dependencies' values from ref results in using deprecated values(state's value won't update). Plus you should also use inputs from `react`, not only from `prepare`(`input` stored in ref pulls only from `prepare`)

karol-bisztyga (changes requested):

src/reanimated2/Hooks.js
terrysahaidak: Yeah, make sense, will update code asap

karol-bisztyga (approved):

Issue #1313, opened 2020-10-06
Frame drop on spring animation start after releasing a pan gesture


terrysahaidak:
## Description

Working on the pager component I noticed a frame drop when the user releases a finger and we call onEnd handler where I call withSpring.

As I can see, the problem is in the first frame being delay because we call onEnd and run spring, but the second frame has `deltaTime = 0`, because we just started spring and there is no previous frame timestamp

https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/animations.js#L136

The naive fix for that is to pass the initial timestamp as `now - 16.66` – subtract a single frame time here:
https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/animations.js#L216

In that case, we still have 2 frames delay, but there is no value jump for 2 frames data because we interpolate value across that second frame too.

### Screenshots

Videos can be found here:
https://drive.google.com/file/d/1ab5DiiSD3BIwodQysYgIJL2EF12wYIj7/view?usp=sharing

bug.mp4 - swipe up to dismiss behaviour. You can see the 2 frame delay and view jump after it
bug-2.mp4 - pager page changes. You can see stop/jump  after finger release
bug-fixed.mp4 - that how it supposed to be. Applied fix `now - 16.66`
Запис екрана  о 15.31.37 - the slow version of that jump

## Steps To Reproduce

1. Use PanResponder
2. Move the finger a little bit "fast" (at least 1000 velocity) and translate the view
3. Call withSpring in `onEnd` with the velocity following the finger direction
4. See 2 frames delay and view jump on the 3rd frame after onEnd

### Expected behavior

It's okay to have 1 or even 2 frame delay until the animation starts. But the animation value should not "jump"

### Actual behavior

Animation value jumps

## Snack or minimal code example

The example can be found running this branch and Pager example:
https://github.com/terrysahaidak/react-native-gallery-toolkit/tree/tr-add-pager-example

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.7

PR #1343, opened 2020-10-13
Fix animations initial calculations

karol-bisztyga:
## Description

As we can see here
https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/core.js#L72
first animation invocation happens directly after calling `animation.start`. As it is wrapped in the same `step` call it uses the same `timestamp`. In the same time, we set `animation.lastTimestamp` or `animation.startTime` inside `animation.start`, which is then used for calculating current progress(and other things) of the animations. All of the following values come from `step` called on the following frames(with according `timestamp`), which results in time delta being equal to or greater than a frame length(for 60FPS that's 1000/60=16.66). That causes a frame delay in animations.

Solution is to call `animation.start` with `timestamp - FRAME_LENGTH` and call `animation.animation` with `timestamp`(so it's like `animation.start` is called in the frame before `animation.animation`).

Thanks @terrysahaidak for investigation

Fixes https://github.com/software-mansion/react-native-reanimated/issues/1313

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -7,7 +7,6 @@ import {
   makeMutable,
   makeRemote,
   requestFrame,
-  FRAME_LENGTH,
 } from './core';
 import updateProps from './UpdateProps';
 import { initialUpdaterRun } from './animations';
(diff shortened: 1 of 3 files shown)


terrysahaidak:
I think this can also be done for decay animation. And maybe for withTiming as well.

terrysahaidak:
And also I guess it exists in Reanimated 1 too,but I'm not sure.

karol-bisztyga:
> I think this can also be done for decay animation. And maybe for withTiming as well.

Yes, it now applies for all animations

terrysahaidak:
The only concern my is 1000 / 60. because actually it can be not true for some devices where fps is higher than 60. But not sure how to address it.

Issue #1285, opened 2020-09-26
Reloading Via React Native Debug Menu Causes Crash


ksitko:
## Description

Hi, I'm using Reanimated V2 with a Expo Bare workflow app. Attempting to do a full reload in either iOS sim or Device causes a crash with a `com.facebook.react.JavaScript (13): signal SIGABRT` error in JSCRuntime::~JSCRuntime. Fast refresh does not cause a similar issue.

### Screenshots
<img width="1445" alt="Screen Shot 2020-09-26 at 2 26 09 PM" src="https://user-images.githubusercontent.com/13754630/94347752-d6e2f780-0004-11eb-90bd-a87538076ae9.png">

## Steps To Reproduce

Enable Turbomodules in ejected Bare workflow expo app and follow Reanimated V2 documentation to change AppDelegate.mm, load app, refresh using RN Debug Menu.

### Expected behavior

Normal refresh.

### Actual behavior

Crash

## Snack or minimal code example

The bug is in Bare work and I don't think Snacks support v2 yet.

## Package versions

`"react-native-reanimated": "2.0.0-alpha.6"`
```
Expo CLI 3.27.10 environment info:
    System:
      OS: macOS 10.15.6
      Shell: 5.7.1 - /bin/zsh
    Binaries:
      Node: 14.4.0 - /usr/local/bin/node
      Yarn: 1.22.5 - ~/.yarn/bin/yarn
      npm: 6.14.8 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.9.3 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: iOS 14.0, DriverKit 19.0, macOS 10.15, tvOS 14.0, watchOS 7.0
    IDEs:
      Android Studio: 3.6 AI-192.7142.36.36.6392135
      Xcode: 12.0.1/12A7300 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^39.0.0 => 39.0.2 
      react: 16.13.1 => 16.13.1 
      react-dom: 16.13.1 => 16.13.1 
      react-native: 0.63.2 => 0.63.2 
      react-native-web: ~0.13.7 => 0.13.13 
    Expo Workflow: bare
```

ksitko:
Resolved by #1256

ksitko:
[...]

PR #1346, opened 2020-10-13
Fix freeze problem

Szymon20000:
## WHY
Currently, there are some cases when Reanimated freezes. It caused by a deadlock that occurs during synchronous layout. 
I've discovered that during props update on an event we can actually trigger a new event `onLayout` synchronously. It's a deadlock situation because we try to acquire lock the second time. https://github.com/facebook/react-native/blob/master/React/CoreModules/RCTEventDispatcher.mm#L124

[update] As it turned out there were several causes of the deadlock. Because StoreUser is used from different threads (UI, JS, JS-GC) we have to make it thread-safe. Locks are not enough because clearing jsi::value from the store can trigger GC on another thread or can trigger another clearing. I wasn't able to find a proper lock-free map (I considered AtomicHashMap from folly but as authors say in the readme it's not a good solution when someone erases frequently). I decided it's that we will schedule erasing jsi::Value on the UI thread. Because there is a danger that we will try to remove jsi::Value after the NativeReaniamtedModule is destroyed I also check for the case.
 

## SOLUTION
We can overwrite EventDispatcher and intercept events in `sendEvent` method without acquiring `_observers lock`.  EventDispatcher can be overwritten in `JsExecutorFactoryForBridge` method which is called before any core module is initialized. First, we ask for EventDispatcher what triggers its initialization, and second, we replace EventDispatcher by our class.

## TESTING
Here is a repro for freezing 

<details> 
<summary>
App.tsx
</summary>

```JS

import React, { useState } from "react";
import Animated, { useSharedValue, withTiming, useAnimatedStyle, Easing } from "react-native-reanimated";
import {
  View,
  StatusBar,
  Text,
  StyleSheet,
  Button,
  FlatList,
  SectionList,
  SectionListRenderItemInfo,
} from "react-native";
import FlatListElement from "./FlatListElement";

  container: {
    flex: 1,
  },
});

const DATA = [
  {
    id: "bd7acbea-c1b1-46c2-aed5-3ad53abb28ba",
    title: "First Item",
  },
  {
    id: "3ac68afc-c605-48d3-a4f8-fbd91aa97f63",
    title: "Second Item",
  },
  {
    id: "58694a0f-3da1-471f-bd96-145571e29d72",
    title: "Third Item",
  },
  {
    id: "9273d2aa-ecb4-493a-a3d4-9c7e949c9227",
    title: "Fourth Item",
  },
  {
    id: "9ff695ed-c79a-4145-9c1c-14502faad2cb",
    title: "Fifth Item",
  },
  {
    id: "a1866a00-5c2a-4edd-97d4-60ad47dc8424",
    title: "Sixth Item",
  },
  {
    id: "2df053c5-6743-4a8e-8231-1dbb734332e1",
    title: "Seventh Item",
  },
];

export default function App(props) {
  const [selected, setSelected] = useState("");

  const onPress = (id: string) => {
    if (selected !== id) {
      setSelected(id);
    } else {
      setSelected("");
    }
  };

  const renderItem = ({ item }) => (
    <FlatListElement text={item.title} onPress={() => onPress(item.id)} selected={selected === item.id} />
  );

  return (
      <FlatList data={DATA} renderItem={renderItem} keyExtractor={(item) => item.id} />
    </View>
  );
} 

```
</details>

<details>
<summary>FlatListElement.tsx</summary>

```JS
import React, {
    useContext,
    useState,
    useCallback
} from "react";
import {
    StyleSheet,
    View,
    TouchableWithoutFeedback,
    Text
} from "react-native";
import Animated, {
    useSharedValue,
    useAnimatedStyle,
    useDerivedValue,
    withSpring,
    withTiming,

    container: {
        width: "100%",
        flexDirection: "column",
        alignItems: "center",
        justifyContent: "center",
    },
    primary: {
        height: 140,
        backgroundColor: "#222222",
    },
    alt: {
        backgroundColor: "#FF4136",
        height: 230,
    },
    text: {
        fontSize: 24,
        color: "#FFFFFF",
    },
});

interface FlatListElementProps {
    text: string;
    selected: boolean;
    onPress: (value: any) => void;
}

const FlatListElement = ({
    text,
    onPress,
    selected
}: FlatListElementProps) => {
    const open = useSharedValue(selected);
    const progress = useDerivedValue(() =>
        open.value ? withSpring(1, {
            damping: 14,
            stiffness: 130
        }) : withSpring(0, {
            damping: 14,
            stiffness: 130
        })
    );
    const progressAlpha = useDerivedValue(() => (open.value ? withTiming(1) : withTiming(0)));
    const height = useSharedValue(140);
        height: height.value + 230 * progress.value + 1,
    }));

    return ( 
        <Animated.View style = {style} >
            <TouchableWithoutFeedback onPress = {onPress} >
                <View style = {[styles.primary, styles.container]} >
                    <Text style = {styles.text} > 
                        {
                            text
                        } 
                    </Text> 
                </View> 
            </TouchableWithoutFeedback> 
            {
                selected && ( 
                <TouchableWithoutFeedback onPress = {onPress} >
                    <View style = {[styles.alt, styles.container]} >
                        <Text style = {styles.text} > 
                            ALT {text} 
                        </Text> 
                    </View> 
                </TouchableWithoutFeedback>
                )
            } 
        </Animated.View>
    );
};

export default FlatListElement; 

```

</details>

## Fixes
fixes: https://github.com/software-mansion/react-native-reanimated/issues/1284
fixes: https://github.com/software-mansion/react-native-reanimated/issues/1285


JonnyBurger:
I've tried this PR as-is out, for me commenting out the line `[bridge.uiManager.observerCoordinator addObserver:self];` causes some bugs in React Navigation, like the drawer overlay now does not hide anymore.

I uncommented that line and trying this PR out, will report back if I see any issues.

ios/native/REAIOSScheduler.mm
```diff
@@ -5,17 +5,24 @@
 using namespace facebook;
 using namespace react;
 
+bool volatile can = true;
```
karol-bisztyga, line 8: unused

Example/src/App.js
```diff
@@ -1,175 +1,59 @@
...
-  Platform,
-  StyleSheet,
-  Text,
-  View,
-  YellowBox,
-} from 'react-native';
-import { RectButton, ScrollView } from 'react-native-gesture-handler';
-import { createAppContainer, createSwitchNavigator } from 'react-navigation';
-import { createStackNavigator } from 'react-navigation-stack';
+import {default as React, useEffect, useState} from 'react';
```
karol-bisztyga, line 1: just remember to revert the repro before merging to master please

karol-bisztyga (approved):

PR #1347, opened 2020-10-14
Decorate callbacks

karol-bisztyga:
## Description

`runOnJS` change made it harder to create animation callback which runs on JS thread. Previously it worked with code like this:

```
import {
  useSharedValue,
  withTiming,

const App = function({ componentId }) {
  const progress = useSharedValue(-1);
  useEffect(() => {
    progress.value = withTiming(
      10,
      null,
      (isFinished) => {
        console.log('clb', _WORKLET, isFinished);
      }
    );
  }, []);
  return <View style={{ backgroundColor: 'red', width: 100, height: 100 }} />;
};

export default App;
```

This would just print `'clb', false, true` which meant that it runs on JS thread and the argument `isFinished` is passed correctly.

Now such code throws an error

```
Tried to synchronously call function from a diffrent thread.
Solution is:
a) if you want to synchronously execute this method, mark it as a worklet
b) if you want to execute this method on the JS thread, wrap it using runOnJS
```

Wrapping the callback with `runOnJS` like this

```
runOnJS((isFinished) => {
  console.log('clb', _WORKLET, isFinished);
})
```

does not make any change(also when lambda is defined outside and just passed as a variable).
To make it work we may convert the callback into a worklet and call `runOnJS` inside like this:

```
(isFinished) => {
  'worklet'
  runOnJS(() => {
    console.log('clb', _WORKLET, isFinished);
  })(isFinished)
}
```

or like this:

```
const fun = (isFinished) => {
    console.log('clb', _WORKLET, isFinished);
  };

  useEffect(() => {
    progress.value = withTiming(
      10,
      {duration: 2000},
      createWorklet((isFinished) => {
        runOnJS(fun)(isFinished)
      }
    ));
  }, []);
```

In order to avoid such code boilerplate I decided to decorate callbacks. Those which are worklets stay as they are and those which are not are wrapped into similar call like above: they are invoked in `runOnJS` inside of the worklet.
I also improved animations test.

## Changes

`animations.js` - decorate callbacks
`SimpleTest.js` - improve animations test

### DIFF
--- a/Example/test/SimpleTest.js
+++ b/Example/test/SimpleTest.js
@@ -1,3 +1,4 @@
+/* global _WORKLET */
 import React from 'react';
 import { TextInput, Button, View } from 'react-native';
 import Animated, {
@@ -13,7 +14,8 @@ import Animated, {
   useAnimatedScrollHandler,
   withDecay,
   withDelay,
-  loop,
+  withRepeat,
+  withSequence,
 } from 'react-native-reanimated';
 import { PanGestureHandler } from 'react-native-gesture-handler';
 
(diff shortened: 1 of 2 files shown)


terrysahaidak:
Wait. Shouldn't this callback be worklet by default instead?

I've always thought for some reason it is worklet. But just realized there is no code written in babel plugin to convert it automatically.

I'm pretty sure I'm not alone here.

karol-bisztyga:
I just fixed it so it works like it used to. Whether this should be a worklet by default is another story and, of course, could be discussed

Szymon20000 (approved):

PR #1361, opened 2020-10-20
Make runOnJS more descriptive.

Szymon20000:
It's only a temporary solution. In the future, we will display a full JS stack trace.

### DIFF
--- a/Common/cpp/headers/SharedItems/ShareableValue.h
+++ b/Common/cpp/headers/SharedItems/ShareableValue.h
@@ -13,6 +13,19 @@ using namespace facebook;
 
 namespace reanimated {
 
+struct HostFunctionHandler {
+  std::shared_ptr<jsi::Function> pureFunction;
+  std::string functionName;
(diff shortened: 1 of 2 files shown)


Common/cpp/headers/SharedItems/ShareableValue.h
```diff
@@ -13,6 +13,19 @@ using namespace facebook;
 
 namespace reanimated {
 
+struct HostFunctionHandler {
```
karol-bisztyga, line 16: why even add this? It adds more complexity and uses more memory just to make an access to `functionName` easier, but that name's only used once so I don't really see a point in doing this.

kmagiera:
I understand that we throw an exception from the implementation of such method, why can't we present the exception stacktrace in this case? the trace would include the function name and be even more helpful as it'd also include previous calls that lead to the exception.

Common/cpp/headers/SharedItems/ShareableValue.h
karol-bisztyga: discussed internally, resolving

karol-bisztyga (approved):

PR #1362, opened 2020-10-20
Use createFromJsonUtf8 in processEvent

terrysahaidak:
## Description

There was TODO to migrate from inline eval. 
I'm not sure it's faster/slower right now, but I guess it will be faster.

## Changes

— use createFromJsonUtf8 instead of inline eval in processEvent

### DIFF
--- a/Common/cpp/Registries/EventHandlerRegistry.cpp
+++ b/Common/cpp/Registries/EventHandlerRegistry.cpp
@@ -30,10 +30,8 @@ void EventHandlerRegistry::processEvent(jsi::Runtime &rt, std::string eventName,
     auto lastBracketCharactedPosition = eventPayload.size() - positionToSplit - 1;
     auto eventJSON = eventPayload.substr(positionToSplit,  lastBracketCharactedPosition);
 
-    std::vector<uint8_t> eventJSONVector(eventJSON.begin(), eventJSON.end());
+    auto eventObject = jsi::Value::createFromJsonUtf8(rt, (uint8_t*)(&eventJSON[0]), eventJSON.size());
 
-    auto eventObject = jsi::Value::createFromJsonUtf8(rt, &eventJSONVector[0], eventJSON.size());
-      
     eventObject.asObject(rt).setProperty(rt, "eventName", jsi::String::createFromUtf8(rt, eventName));
     for (auto handler : handlersIt->second) {
       handler.second->process(rt, eventObject);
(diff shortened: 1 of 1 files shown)


Common/cpp/Registries/EventHandlerRegistry.cpp
```diff
@@ -22,8 +22,18 @@ void EventHandlerRegistry::unregisterEventHandler(unsigned long id) {
...
-    auto eventObject = eval(rt, ("(" + eventPayload + ")").c_str()).asObject(rt).getProperty(rt, "NativeMap");
+    // We receive here a JS Map with JSON as a value of NativeMap key
+    // { NativeMap: { "jsonProp": "json value" } }
+    // So we need to extract only JSON part
+    std::string delimimter = "NativeMap:";
+    auto positionToSplit = eventPayload.find(delimimter) + delimimter.size();
+    auto lastBracketCharactedPosition = eventPayload.size() - positionToSplit - 1;
+    auto eventJSON = eventPayload.substr(positionToSplit,  lastBracketCharactedPosition);
+
+    std::vector<uint8_t> eventJSONVector(eventJSON.begin(), eventJSON.end());
```
karol-bisztyga, line 33: Creating a vector here seems to be unnecessary. Simple conversion in the next instruction works:
`auto eventObject = jsi::Value::createFromJsonUtf8(rt, (uint8_t*)(&eventJSON[0]), eventJSON.size());`

terrysahaidak: Yep, make sense, didn't know how to convert it properly here. One more thing learned today, thanks :)

karol-bisztyga:
As for the speed whther it increases here or not - well, haven't run any tests but internally it just calls `JSON.parse`:
https://github.com/facebook/react-native/blob/master/ReactCommon/jsi/jsi/jsi.cpp#L137
So the question is if `JSON.parse` is faster/better than `eval`.
It actually could be slower but for sure is safer and recommended.

karol-bisztyga (approved):

jakub-gonet:
[CI fails](https://travis-ci.org/github/software-mansion/react-native-reanimated/jobs/737367895#L1956)
```
./../Common/cpp/Registries/EventHandlerRegistry.cpp:6:19: error: unused function 'eval' [-Werror,-Wunused-function]
static jsi::Value eval(jsi::Runtime &rt, const char *code) {
                  ^
```

terrysahaidak:
Will fix it in a bit :)

PR #1389, opened 2020-10-29
Fix call invoker for rn 62

karol-bisztyga:
## Description

Fix call invoker for iOS for RN 0.62.x


WoLewicki (approved):
Keep up the good work!

PR #1391, opened 2020-10-30
Add runOnJS type export

mrousavy:
## Description

`runOnJS` is required to dispatch functions from a worklet, but it isn't exported from the type module.

Simulator Screen Shot - iPhone 11 - 2020-10-30 at 13 19 02

## Changes

- Exported `runOnJS` using same type signature as `runOnUI`

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -443,6 +443,9 @@ declare module 'react-native-reanimated' {
     export function runOnUI<A extends any[], R>(
       fn: (...args: A) => R
     ): (...args: Parameters<typeof fn>) => void;
+    export function runOnJS<A extends any[], R>(
+      fn: (...args: A) => R
+    ): (...args: Parameters<typeof fn>) => void;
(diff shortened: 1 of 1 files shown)


jakub-gonet:
Changes from here are included in #1392.

dhirajanand014:
How to resolve this issue. My app crashes when - 

a.i want to focus an input text on finished. 

`input_search_box_translate_x.value = withSpring(isShowInputBox && 1 || width, {
            damping: 20
        }, (finished) => {
            runOnJS(focusAndBlurInputBox)(isShowInputBox, inputTextRef); <-- Guessing this is causing the app to crash.
        })`

b. Also the same should also work when i need to update the state by calling setValues state hook when on finished.

`content_translate_y.value = withSpring(isShowInputBox && 1 || height, {
            damping: 20
        }, (finished) => {
//how to get the following working?
            setSearchValues({
                ...searchValues,
                searchForPostId: post.postId
            });
        });`

PR #1392, opened 2020-10-30
Add runOnJS types

karol-bisztyga:
## Description

- add runOnJS typescript types
- fix link in docs' installation section
- `interpolateColor` - thanks to @mrousavy 
	- add TS types tests
	- add TS types
	- replace `enum` with string union

### DIFF
--- a/docs/docs/installation.md
+++ b/docs/docs/installation.md
@@ -122,7 +122,7 @@ On iOS installation is automatic.
 
 > **_NOTE:_** In previous releases, the installation process was manual and required turning turbo modules on. Some libraries break when turbo modules are enabled so we decided to change our approach and we no longer
 > use the standard way for registering a turbo module. It let us simplify the installation process and as a result, you can safely
-> undo all installation steps from the [previous instruction](https://docs.swmansion.com/react-native-reanimated/docs/installation#ios).
+> undo all installation steps from the [previous instruction](https://docs.swmansion.com/react-native-reanimated/docs/2.0.0-alpha.7/installation#ios).
 
 > **_NOTE:_** If you want to turn off autoinstall on iOS please add the following compilation flag:
 > `DONT_AUTOINSTALL_REANIMATED`.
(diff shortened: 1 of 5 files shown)


Szymon20000 (approved):

mrousavy:
I've also noticed that types for `interpolateColor` are missing, do you want me to create a PR or are you on it?

karol-bisztyga:
> I've also noticed that types for `interpolateColor` are missing, do you want me to create a PR or are you on it?

Hey, I've added types for `interpolateColor`, you can check if they're valid, thanks for pointing this out!

mrousavy:
Looks good, I think color space can be optional, no?

Also on a side note, I'd recommend not using enums. Consider the following:

```ts
import { ColorSpace } from '...';
const x = interpolateColor(y, [0, 1], ['#fff', '#000'], ColorSpace.RGB);
```
which would so much cleaner if it were written like this:

```ts
const x = interpolateColor(y, [0, 1], ['#fff', '#000'], 'rgb');
```

Same applies for the `Extrapolate` prop, I prefer writing 
```ts
{ extrapolate: 'clamp' }
``` 
to
```ts
{ extrapolate: Extrapolate.CLAMP }
```
as this way I don't need an extra import, don't need to wait for my IDE (or typescript) to show me what values `Extrapolate` has (after it has imported it), and I don't have to repeat myself. Also see: [Why TypeScript enums suck](https://blog.logrocket.com/why-typescript-enums-suck/)

This doesn't have anything to do with this PR in particular, thought I'd just point this out to consider this before reanimated v2 is initially released, as breaking changes wouldn't be a problem in an alpha. 

Thoughts? @karol-bisztyga @Szymon20000 @kmagiera @wcandillon @terrysahaidak

jakub-gonet:
@mrousavy I think this discussion should be moved into a new issue.

Agreed, string unions are way cleaner and I'm pretty sure no one wants runtime behavior of enums (e.g. enumerating possible values) but I may be wrong.

> Looks good, I think color space can be optional, no?
Definitely, we default to RGB space in JS.

karol-bisztyga:
@mrousavy thanks for the review, I updated the code

PR #1396, opened 2020-10-30
Fix example for `useAnimatedProps()`

catalinmiron:
## Description

Fixes #1396.

Fix `useAnimatedProps` example so people can copy-paste and see the result.


jakub-gonet (approved):

jakub-gonet:
Thanks!

Issue #1387, opened 2020-10-28
Updating custom props on UI thread does not work on iOS


kmagiera:
## Description

When using third party components it is no longer possible to register additional properties to be updated directly on the UI thread. With Reanimated 1 it was possible using methods like `addWhitelistedNativeProps` or `addWhitelistedUIProps`, but this is no longer the case in Reanimated 2.

I tracked this problem down to the following line:
https://github.com/software-mansion/react-native-reanimated/blob/08c93d0c6cfef9d6930b19cf98b20c7488cc08d7/ios/native/NativeProxy.mm#L87

Where we always assume that we update props for RCTView component. Hence custom properties are added to some other view types, the props aren't properly applied and are filtered out.

The correct resolution would be to figure out a way how we can avoid having the view type hardcoded like this. In Reanimated 1 we use UIManager.viewNameForReactTag – but this method is only allowed to be called on the ui manager thread which was ok in Rea 1 as we had a registration step that was run on that thread. In Rea 2 we don't have registration step and may need to try figure out where to take this information from.

kyle-ssg:
Hi @kmagiera, came across this issue which perplexed me a bit.

Does this mean the following is not updated on the UI Thread? From what I could tell it was.

<img src="http://g.recordit.co/XfsADZ7fBS.gif"/>

***[Code](https://github.com/kyle-ssg/reanimated2-experiments/blob/master/App/components/AnimatedBlur.tsx#L33)***

```
const animatedProps = useAnimatedProps(() => {
    return {
      blurAmount: interpolate(animatedValue.value, [0, 1], [0, 25]),
    };
  });
```

kmagiera:
@kyle-ssg – yes, the example you posted work because the updates are happening via JS thread. Parameters cannot be updated on UI unless added to the whitelist. Unfortunately due to a bug only RCTView instances can be updated properly on UI thread right now.

karol-bisztyga:
Should be closed automatically when closing #1409

kyle-ssg:
@karol-bisztyga, apologies for bugging you on this - but with this what will the process be for whitelisting props ? Also, is there a way I can validate that prop updates are on the UI thread?

karol-bisztyga:
No worries my friend, I'm here to help. Now whitelisting happens automatically - that means you don't have to add props you want to update on UI thread manually, they're added dynamically.
[...]

PR #1409, opened 2020-11-03
Improve UI props updating

karol-bisztyga:
## Description

The first goal of this PR was to fetch the name of the component for the prop updater what's needed on iOS. Until now we've provided `RCTView` for any component. Quoting the issue:
> Where we always assume that we update props for RCTView component. Hence custom properties are added to some other view types, the props aren't properly applied and are filtered out

The component's name is obtained in `createAnimatedComponent`, on the javascript side, from the mounted component's props.

During creating animated component out of a regular one also UI props whitelist is being update dynamically(once for every unique component name detected). Thanks to this solution we are able to update props, specific for the certain components, on UI thread without hardcoding them.

Solves https://github.com/software-mansion/react-native-reanimated/issues/1387

## Reproducing

The issue occurs when there are some custom animated components which props are being updated on UI thread.
To reproduce you can use the following code:

<details>
<summary>code</summary>

```
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  useAnimatedProps,
import React from 'react';
import { Svg, Circle, Rect, Ellipse } from 'react-native-svg';

  const ACircle = Animated.createAnimatedComponent(Circle);
  const ARect = Animated.createAnimatedComponent(Rect);
  const AEllipse = Animated.createAnimatedComponent(Ellipse);

    return {
      width: withTiming(randomWidth.value),
    };
  });

  const circleProps = useAnimatedProps(() => {
    return {
      fill: 'red',
      cx: 50,
      cy: 50,
      r: withTiming(randomWidth.value / 2),
    };
  });

  const ellipseProps = useAnimatedProps(() => {
    return {
      cx: 200,
      cy: 200,
      rx: withTiming(randomWidth.value / 4),
      ry: 200,
      stroke: 'black',
      strokeWidth: 2,
      fill: 'purple',
    };
  });

  const rectProps = useAnimatedProps(() => {
    return {
      x: withTiming(randomWidth.value / 4),
      y: 50,
      width: 50,
      height: 50,
      stroke: 'orange',
      strokeWidth: 2,
      fill: 'green',
    };
  });

  return (
    <View
      style={{
        flex: 1,
        flexDirection: 'column',
      }}>
      <Animated.View
        style={[
          style,
        ]}
      />
      <Button
        title="toggle"
        onPress={() => {
        }}
      />
      <Svg>
        <ACircle animatedProps={circleProps} />
        <ARect animatedProps={rectProps} />
        <AEllipse animatedProps={ellipseProps} />
      </Svg>
    </View>
  );
}

```

</details>

Also you have to add updated props to `UI_THREAD_PROPS_WHITELIST` in `ConfigHelper.js`, like this:

```
let UI_THREAD_PROPS_WHITELIST = {
  // ...
  rx: true,
  r: true,
}
```

If the view name in `NativeProxy.mm` is hardcoded to `RCTView` the props are not set but with proper view names they do work.

## Testing

Some cases from svg library have been tested like rect, circle or ellipse as well as a regular `RCTView`.

## Changes

Basically the main idea is placed in `createAnimatedComponent` where the view name is obtained(not directly but still) in `componentDidMount` so there's an access to a mounted component(and therefore to `viewConfig`). The code extending UI props list dynamically is placed in `CodeHelper.js`


src/createAnimatedComponent.js
```diff
@@ -216,6 +216,19 @@ export default function createAnimatedComponent(Component) {
     }
 
     _attachPropUpdater() {
+      if (Platform.OS === 'ios') {
```
kmagiera, line 219: why is this here? Seem like it'd belong better in place where we already set viewTag for reanimated2, that is a couple of lines lower in _attachAnmatedStyles

src/createAnimatedComponent.js
```diff
@@ -216,6 +216,19 @@ export default function createAnimatedComponent(Component) {
...
+        for (const key in this.props) {
+          const value = this.props[key];
+          if (key === 'style' || key === 'animatedProps') {
+            const props = Array.isArray(value) ? value : [value];
+            props.forEach((prop) => {
+              if (prop && prop.viewName && !prop.viewName.value) {
+                prop.viewName.value = this._obtainViewName();
```
kmagiera, line 226: can't we store viewName information along viewTag that is updated in `_attachAnimatedStyles` already?

src/createAnimatedComponent.js
```diff
@@ -344,6 +357,18 @@ export default function createAnimatedComponent(Component) {
       return props;
     }
 
+    _obtainViewName(forceComponent) {
```
kmagiera, line 360: What is `forceComponent` – it does not seem like we don't pass that parameter anywhere

src/createAnimatedComponent.js
```diff
@@ -344,6 +357,18 @@ export default function createAnimatedComponent(Component) {
...
+      const component = forceComponent || this._component;
+      try {
+        const viewConf = component.viewConfig || component.root.viewConfig;
+        return viewConf.uiViewClassName;
+      } catch (e) {
```
kmagiera, line 365: why we catch here? can't we just do null checks when accessing attributes? Look like the code inside try block does not do antything other than accessing some attributes

kmagiera (changes requested):
I think this is a good direction overal. However I don't understand why we need so many places where we do some special handling specific to viewName that are not needed for viewTag which is already in the codebase. Seem to be like we should be able to treat these two attributes equally, is there any issue with that I may not see?

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -58,7 +58,10 @@ void RuntimeDecorator::addNativeObjects(jsi::Runtime &rt,
       ) -> jsi::Value {
     const auto viewTag = args[0].asNumber();
     const auto params = args[1].asObject(rt);
-    updater(rt, viewTag, params);
+    // args[2] is maybeViewRef which is only for web and will be undefined here
```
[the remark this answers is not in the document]
karol-bisztyga: This means that here `arg[2]` is a view reference which is used only for web, in fact I just added this comment to make things less confusing 😄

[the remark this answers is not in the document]
karol-bisztyga: as we use `arg[0]`, `arg[1]` and then `arg[3]` I wanted to add some explanation why we skip `arg[2]`

src/createAnimatedComponent.js
karol-bisztyga: yes indeed I decided it's just clearer to look out for an error here than check every single level of accessed object explicitly

src/createAnimatedComponent.js
karol-bisztyga: fixed c1d13d9

src/createAnimatedComponent.js
karol-bisztyga: fixed c1d13d9

src/createAnimatedComponent.js
karol-bisztyga: fixed c1d13d9

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -58,7 +58,10 @@ void RuntimeDecorator::addNativeObjects(jsi::Runtime &rt,
...
+    // view name is passed only on iOS
+    const jsi::Value* maybeViewName = &args[3];
```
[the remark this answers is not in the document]
karol-bisztyga: fixed c1d13d9

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -46,7 +46,8 @@ jni::local_ref<NativeProxy::jhybriddata> NativeProxy::initHybrid(
 void NativeProxy::installJSIBindings()
 {
 
-  auto propUpdater = [this](jsi::Runtime &rt, int viewTag, const jsi::Object &props) {
+  auto propUpdater = [this](jsi::Runtime &rt, int viewTag, const jsi::Object &props, const jsi::Value& viewName) {
```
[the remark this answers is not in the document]
karol-bisztyga: fixed a0b3e8e

Common/cpp/Tools/RuntimeDecorator.cpp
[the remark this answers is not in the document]
karol-bisztyga: fixed anyways a0b3e8e

karol-bisztyga:
> I think this is a good direction overal. However I don't understand why we need so many places where we do some special handling specific to viewName that are not needed for viewTag which is already in the codebase. Seem to be like we should be able to treat these two attributes equally, is there any issue with that I may not see?

I made some changes regarding this.

Also, if someone wants to check how it works here goes a simple example based on `AnimatedStyleUpdateExample`:

<details>
<summary>code</summary>

```
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  Easing,
  useAnimatedProps,
import React from 'react';
import { Svg, Circle, Rect, Ellipse, Text } from 'react-native-svg';

  const ACircle = Animated.createAnimatedComponent(Circle);
  const ARect = Animated.createAnimatedComponent(Rect);
  const AEllipse = Animated.createAnimatedComponent(Ellipse);
  const AText = Animated.createAnimatedComponent(Text);

  const config = {
    duration: 500,
  };

    return {
    };
  });


  const props = useAnimatedProps(() => {
    return {
      fill: 'red',
      cx: 50,
      cy: 50,
      r: withTiming(randomWidth.value / 4),
    };
  });

  return (
    <View
      style={{
        flex: 1,
        flexDirection: 'column',
      }}>
      <Animated.View
        style={[
          style,
        ]}
      />
      <Button
        title="toggle"
        onPress={() => {
        }}
      />
      <Svg>
        <ACircle animatedProps={props} />
      </Svg>
      <ARect />
      <AEllipse />
      <AText />
      <Animated.Image />
      <Animated.Text />
    </View>
  );
}

```

</details>

src/createAnimatedComponent.js
jakub-gonet: If I understand this correctly you can use [optional chaining (`?.`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here

karol-bisztyga: fixed 7889b34

ios/native/NativeProxy.mm
```diff
@@ -82,9 +82,11 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
   RCTBridge *bridge = _bridge_reanimated;
   REAModule *reanimatedModule = [bridge moduleForClass:[REAModule class]];
 
-  auto propUpdater = [reanimatedModule](jsi::Runtime &rt, int viewTag, const jsi::Object &props) -> void {
+  auto propUpdater = [reanimatedModule](jsi::Runtime &rt, int viewTag, const jsi::Value& viewName, const jsi::Object &props) -> void {
+    NSString *nsViewName = [NSString stringWithCString:viewName.asString(rt).utf8(rt).c_str() encoding:[NSString defaultCStringEncoding]];
+
     NSDictionary *propsDict = convertJSIObjectToNSDictionary(rt, props);
-    [reanimatedModule.nodesManager updateProps:propsDict ofViewWithTag:[NSNumber numberWithInt:viewTag] viewName:@"RCTView"];
+    [reanimatedModule.nodesManager updateProps:propsDict ofViewWithTag:[NSNumber numberWithInt:viewTag] viewName:nsViewName];
```
kmagiera, line 89: `viewName` part does not match objc method naming convension which is that it when you read method with arguments then it creates a sort of a sentence, before we had:

update-props-of-view-with-tag-x

after your change:

update-props-of-view-with-tag-x-view-name

so clearly the last part does not fit, I'd suggest someting like:

update-props-of-view-with-tag-x-with-name-y

src/createAnimatedComponent.js
```diff
@@ -229,14 +229,25 @@ export default function createAnimatedComponent(Component) {
         : [this.props.style];
       styles = flattenArray(styles);
       const viewTag = findNodeHandle(this);
+      const viewName = this._obtainViewName();
       styles.forEach((style) => {
-        if (style && style.viewTag !== undefined) {
-          style.viewTag.value = viewTag;
+        if (style) {
+          if (style.viewTag) {
```
kmagiera, line 235: it looks like we need both viewTag and viewName all of the time. It seem it'd make more sense to just have a single value with both these items. We'd avoid additional checks here and in other places + it'd be a bit more intuitive IMO:

For example:
```
style.viewDescriptor.value = { viewTag, viewName }
```

src/createAnimatedComponent.js
```diff
@@ -344,6 +355,12 @@ export default function createAnimatedComponent(Component) {
       return props;
     }
 
+    _obtainViewName() {
```
kmagiera, line 358: obtain is a bit weird wording, it means get+hold whereas the method does simply get the name and returns it

src/reanimated2/Hooks.js
```diff
@@ -184,7 +184,7 @@ function styleDiff(oldStyle, newStyle) {
   return diff;
 }
 
-function styleUpdater(viewTag, updater, state, maybeViewRef) {
+function styleUpdater(viewTag, updater, state, maybeViewRef, viewName) {
```
kmagiera, line 187: I still think it'd be more clear if we stopped using pair viewTag and viewName as separate arguments and just rename `viewTag` to `viewDescriptor` here and in other places. If not we should at least place them next to each other

src/reanimated2/Hooks.js
```diff
@@ -268,6 +268,7 @@ export function useAnimatedStyle(updater, dependencies) {
   const initRef = useRef(null);
   const inputs = Object.values(updater._closure);
   const viewRef = useRef(null);
+  const viewName = useSharedValue(null);
```
kmagiera, line 271: same comment as before. Lets keep viewTag and viewName at least next to each other. I'd prefer if we just have a descriptor that stores both, this way we'd also avoid additional hook call here.

src/createAnimatedComponent.js
```diff
@@ -344,6 +355,12 @@ export default function createAnimatedComponent(Component) {
...
+      const viewConf =
+        this._component?.viewConfig || this._component?.root?.viewConfig;
+      return viewConf.uiViewClassName;
```
karol-bisztyga, line 361: should be `return viewConf?.uiViewClassName;`, if `viewConf` is `undefined` it will break

src/reanimated2/UpdateProps.js
```diff
@@ -23,8 +23,13 @@ const colorProps = [
 
 const ColorProperties = makeShareable(colorProps);
 
-export default function updateProps(viewTag, updates, maybeViewRef) {
+export default function updateProps(viewTag, viewName, updates, maybeViewRef) {
   'worklet';
+  if (viewName && viewName.value) {
```
kmagiera, line 28: under what condition the viewName will not be set? please at least add a comment here. IMO we should not have a fallback but rather throw but I don't know in what situations the name is not there

src/createAnimatedComponent.js
```diff
@@ -344,6 +355,12 @@ export default function createAnimatedComponent(Component) {
       return props;
     }
 
+    _obtainViewName() {
+      const viewConf =
+        this._component?.viewConfig || this._component?.root?.viewConfig;
```
kmagiera, line 360: this uses some internal structures of react-native and deserves at least a comment on what we're accessing. What is `root`? What is `viewConfig`? What may break here if that internal structure changes?

kmagiera (changes requested):
Added a bunch of new comments.

On top of the comments this PR needs to include a test plan. As we discussed over the call, we need to make sure this has been tested in release builds as the internal fields we are accessing may work differently in dev vs release (not saying they do).

I looked at the sample code you posted in a comment – this should be a part of the PR description (maybe not the code, but at least a link to it). You should also mention what is the expected behavior. The code you posted works fine w/o this change, the difference is that the props are being updated via JS (with a roundtrip to JS and back), so it may appear to work fine while we expect it to work without doing that roundtrip.

src/createAnimatedComponent.js
jakub-gonet, line 235: This also can be flattened to `if(style?.viewTag)`

src/reanimated2/UpdateProps.js
```diff
@@ -23,8 +23,13 @@ const colorProps = [
...
+    viewName = viewName.value;
+  } else {
+    viewName = 'RCTView';
+  }
```
jakub-gonet, line 32: `viewName = viewName?.value ?? 'RCTView';` should be semantically equivalent to this if. BTW, this reassign forces update, right?

jakub-gonet (dismissed):

karol-bisztyga:
> Added a bunch of new comments.
> 
> On top of the comments this PR needs to include a test plan. As we discussed over the call, we need to make sure this has been tested in release builds as the internal fields we are accessing may work differently in dev vs release (not saying they do).
> 
> I looked at the sample code you posted in a comment – this should be a part of the PR description (maybe not the code, but at least a link to it). You should also mention what is the expected behavior. The code you posted works fine w/o this change, the difference is that the props are being updated via JS (with a roundtrip to JS and back), so it may appear to work fine while we expect it to work without doing that roundtrip.

I extended the description. The tests have been run for a several components from svg library on both debug and release modes.

src/reanimated2/UpdateProps.js
karol-bisztyga: This fails and takes that default value if `_getViewName` doesn't find corresponding view name. That shouldn't happen for react native and svg components. But I thought someone could try to create animated components out of some other external libraries(like svg) which in fact could do something what svg did which is putting component data in some custom place in a component's object(the svg's `root` case).

We could throw here - it's just a matter of how we want to handle such situation. On the other hand we could also fire up some warning like `unsupported component type` and alert the user that some props may not be set properly.

src/reanimated2/UpdateProps.js
karol-bisztyga: > this reassign forces update, right?
What update? What do you mean?

src/createAnimatedComponent.js
karol-bisztyga: fixed 6609d065d654dd05e644bee42dfdb62405907c7e

ios/native/NativeProxy.mm
karol-bisztyga: fixed 6609d065d654dd05e644bee42dfdb62405907c7e

src/createAnimatedComponent.js
karol-bisztyga: fixed 6609d065d654dd05e644bee42dfdb62405907c7e

src/reanimated2/Hooks.js
karol-bisztyga: fixed b3f674c646c6d7972e8a0ed50226c8028433ac16

src/reanimated2/Hooks.js
karol-bisztyga: fixed b3f674c646c6d7972e8a0ed50226c8028433ac16

src/createAnimatedComponent.js
karol-bisztyga: fixed b3f674c646c6d7972e8a0ed50226c8028433ac16

karol-bisztyga: fixed d1fa5c479790adcddb9622b1fce6edde61cee462

src/createAnimatedComponent.js
karol-bisztyga: fixed dafa7117cea37a58342516e6aad0e03bbf6eecc8

src/reanimated2/UpdateProps.js
jakub-gonet: Resolved internally.

Szymon20000 (approved):

src/createAnimatedComponent.js
```diff
@@ -228,15 +236,31 @@ export default function createAnimatedComponent(Component) {
...
+      const hostInstance = ReactNative.findHostInstance_DEPRECATED(this);
+      // we can access view tag in the same way it's accessed here https://github.com/facebook/react/blob/e3f4eb7272d4ca0ee49f27577156b57eeb07cf73/packages/react-native-renderer/src/ReactFabric.js#L146
+      const viewTag = hostInstance._nativeTag;
+      /**
+       * RN uses viewConfig for components for storing different properties of the component(example: https://github.com/facebook/react-native/blob/master/Libraries/Components/ScrollView/ScrollViewViewConfig.js#L16).
+       * The name we're looking for is in the field named uiViewClassName.
+       */
+      const viewName = hostInstance.viewConfig.uiViewClassName;
+      // update UI props whitelist for this view
+      this._updateWhitelistProps(hostInstance);
```
kmagiera, line 249: I think it'd be better for this to be moved outside of `createAnimatedComponent` class. No reason for it to be a class member method + now we also need to expose some implementation details from ConfigHelper which we do by making a copy each time the list of props is accessed.

I think it'd be better to move the implementation of this method directly to COnfigHelper. We add a new method there that is called `adaptViewConfig(viewConfig)` that gets the whole viewConfig such that it can use the className to remember which views have been processed and which not and also access the list of attributes.

We should also use Set() for whitelistViewNamesUpdated instead of an array as we do lookups which in case of an array are linear.

kmagiera (changes requested):
Happy that you found a way to get native component instance and we no longer need to know implementation details of SVG or blur library to access the view name.

Can you also update PR name and description? I think after all the changes the current name/description is no longer relevant. Also would be great to include blurview in your testplan (that is, to make sure animating blur radius for example works on UI thread after this change)

src/createAnimatedComponent.js
```diff
@@ -228,15 +236,31 @@ export default function createAnimatedComponent(Component) {
...
-        if (style && style.viewTag !== undefined) {
-          style.viewTag.value = viewTag;
+        if (style?.viewDescriptor) {
+          style.viewDescriptor.value = { tag: viewTag, name: viewName };
         }
       });
       // attach animatedProps property
       if (this.props.animatedProps) {
-        this.props.animatedProps.viewTag.value = viewTag;
+        if (this.props.animatedProps.viewDescriptor) {
```
kmagiera, line 258: we don't need to check for that. If we do maybe just move it inside the previous if statement?

src/createAnimatedComponent.js
kmagiera: As this method has side-effects can we make sure to only call it for reanimated 2 components? Otherwise we are changing behavior for people who are not using reanimated 2 and this is possibly a breaking change.

src/ConfigHelper.js
```diff
@@ -23,6 +23,33 @@ let UI_THREAD_PROPS_WHITELIST = {
   scaleY: true,
   translateX: true,
   translateY: true,
+  /* SVG */
```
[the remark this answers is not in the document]
karol-bisztyga: outdated - closing

src/createAnimatedComponent.js
karol-bisztyga: fixed in 0cd1b52 and 8506102

src/createAnimatedComponent.js
karol-bisztyga: fixed 79b5a88

src/reanimated2/UpdateProps.js
```diff
@@ -23,8 +23,9 @@ const colorProps = [
 
 const ColorProperties = makeShareable(colorProps);
 
-export default function updateProps(viewTag, updates, maybeViewRef) {
+export default function updateProps(viewDescriptor, updates, maybeViewRef) {
   'worklet';
+  const viewName = viewDescriptor?.value?.name || 'RCTView';
```
kmagiera, line 28: under what circumstances view name cannot be set while the tag is? Maybe we can just use the same pattern here as with the tag?

src/ConfigHelper.js
```diff
@@ -125,4 +125,34 @@ export function addWhitelistedUIProps(props) {
   configureProps();
 }
 
+const whitelistViewNamesUpdated = new Set();
```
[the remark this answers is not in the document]
karol-bisztyga: fixed b279f86

src/reanimated2/UpdateProps.js
karol-bisztyga: yes, they're assigned together and we expect view name 100% to be correct, fixed 9d309f2

kmagiera (approved):
I mentioned this earlier but can you please update PR title and description as the original title is no longer relevant. Otherwise the code looks ok

Issue #1390, opened 2020-10-30
Calling measure on the element which is off the screen causing crash on Android


terrysahaidak:
## Description

If the element is off the screen (user can't see it) on iOS we can get its measurements, on Android it causing the app to crash.

## Steps To Reproduce

1. Run provided example
2. Wait 1 second. It will try to measure the element which is off the screen

### Expected behavior

Should work just like on iOS – should return values off the screen

### Actual behavior

Crashes

## Snack or minimal code example

```jsx
import React from 'react';
import { useEffect } from 'react';
import {
  FlatList,
  TapGestureHandler,
} from 'react-native-gesture-handler';
import Animated, {
  useAnimatedRef,
  useAnimatedGestureHandler,
  measure,
  runOnUI,
} from 'react-native-reanimated';

function Block({ index }) {
  const ref = useAnimatedRef();

  const handler = useAnimatedGestureHandler({
    onFinish: () => {
      console.log(measure(ref));
    },
  });

  useEffect(() => {
    if (index === 9) {
      setTimeout(
        () =>
          runOnUI(() => {
            'worklet';

            console.log(measure(ref));
          })(),
        1000,
      );
    }
  }, []);

  return (
    <TapGestureHandler onGestureEvent={handler}>
      <Animated.View
        ref={ref}
        style={{
          height: 200,
          marginBottom: 16,
          backgroundColor: 'red',
        }}
      />
    </TapGestureHandler>
  );
}

function Navigation() {
  return (
    <FlatList
      data={Array.from({ length: 10 }).map((item, index) => index)}
      renderItem={({ index }) => <Block index={index} />}
    />
  );
}
```

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.7
- NodeJS: 12.16.2

piaskowyk:
Hi!
`measure()` function works properly. The problem is with the FlatList component. The FlatList has [VirtualizedList](https://reactnative.dev/docs/virtualizedlist) on - This is optimization based on rendering only visible components on screen. Unfortunately you can’t measure an item that does not exist.

You have some simply solutions:
- use try{} catch{} to wrap problematic code
- disable [virtualization](https://reactnative.dev/docs/virtualizedlist#disablevirtualization) but this is bad solution because have negative impact on performance nad is deprecated
- wrap `<FlatList>` with `<ScrollView>` component

terrysahaidak:
[...]

PR #1421, opened 2020-11-10
Update of docs - added info about exception in measure()

piaskowyk:
## Description
Update of documentation. I added information about possible exceptions in measure() function. It is related with issue: https://github.com/software-mansion/react-native-reanimated/issues/1390

## Changes
in file `docs/docs/api/nativeMethods/measure.md`

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/docs/docs/api/nativeMethods/measure.md
+++ b/docs/docs/api/nativeMethods/measure.md
@@ -35,3 +35,7 @@ const Comp = () => {
   return <View ref={aref} />;
 };
 ```
+
+### Note
+
+You can use `measure()` only on rendered components. Well practise is wrap function call with `try{} catch{}` if there exists possibility to call function on non rendered item, for example: Non visible item on screen from FlatList.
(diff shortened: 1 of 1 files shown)


docs/docs/api/nativeMethods/measure.md
```diff
@@ -35,3 +35,7 @@ const Comp = () => {
   return <View ref={aref} />;
 };
 ```
+
+### Note
+
+You can use `measure()` only on rendered components. Well practise is wrap function call with `try{} catch{}` if there exists possibility to call function on non rendered item, for example: Non visible item on screen from FlatList.
```
karol-bisztyga, line 41: Just a couple of remarks:
**Good** practi**c**e is **to** wrap
if there **is a** possibility
function on non rendered item => function on item which is not rendered
Non visible => invisible

karol-bisztyga (approved):

PR #1431, opened 2020-11-12
Add run on js docs note

karol-bisztyga:
## Description

Add run on js docs note

### DIFF
--- a/docs/docs/api/runonJS.md
+++ b/docs/docs/api/runonJS.md
@@ -1,0 +1,83 @@
+---
+id: runOnJS
+title: runOnJS
+sidebar_label: runOnJS
+---
+
+When you call a function on UI thread you can't be sure if you call a worklet or a callback from JS thread. To make it more transparent we introduce runOnJS method which calls a callback asynchronously. If you call a callback without this method then an exception will be thrown.
+
+#### Note
+
+If you want to invoke some function from external library in `runOnJS` please wrap it into a separate function.
+
+Code like this may not work:
+
+```js
+useDerivedValue(() => {
+    runOnJS(externalLibraryFunction)(args);
+});
+```
+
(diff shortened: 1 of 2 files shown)


docs/docs/api/runonJS.md
```diff
@@ -0,0 +1,83 @@
+---
+id: runOnJS
+title: runOnJS
+sidebar_label: runOnJS
+---
+
+When you call a function on UI thread you can't be sure if you call a worklet or a callback from JS thread. To make it more transparent we introduce runOnJS method which calls a callback asynchronously. If you call a callback without this method then an exception will be thrown.
+
+#### Note
```
jakub-gonet, line 9: I think this should be a quote block (do we have any guidelines on docs structure?)

I'd add a short explanation of why we can't wrap the function directly.

jakub-gonet: Anyway, maybe we should consider wrapping functions automatically?

docs/docs/api/runonJS.md
```diff
@@ -0,0 +1,83 @@
+---
+id: runOnJS
+title: runOnJS
+sidebar_label: runOnJS
+---
+
+When you call a function on UI thread you can't be sure if you call a worklet or a callback from JS thread. To make it more transparent we introduce runOnJS method which calls a callback asynchronously. If you call a callback without this method then an exception will be thrown.
```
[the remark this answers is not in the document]
karol-bisztyga: > "method" -> function

look like repetition, we can just skip the word. The rest is ok

docs/docs/api/runonJS.md
karol-bisztyga: fixed fd4e410

jakub-gonet (approved):

Issue #1447, opened 2020-11-17
App Crashes on Pull to Refresh


Andarius:
## Description

App crashes when swiping to refresh a ScrollView

**Note**: It works with `2.0.0-alpha.8` so the bug was likely introduced by `2.0.0-alpha.9`

### Screenshots

```
com.facebook.jni.CppException Value is null, expected an Object

Error: Value is null, expected an Object 
    NativeProxy.java:-2 com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent
    NativeProxy.java:60 com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent
    RefreshEvent.java:26 com.facebook.react.views.swiperefresh.RefreshEvent.dispatch
    NodesManager.java:471 com.swmansion.reanimated.NodesManager.handleEvent
    NodesManager.java:448 com.swmansion.reanimated.NodesManager.onEventDispatch
    EventDispatcher.java:114 com.facebook.react.uimanager.events.EventDispatcher.dispatchEvent
    SwipeRefreshLayoutManager.java:147 com.facebook.react.views.swiperefresh.SwipeRefreshLayoutManager$1.onRefresh
    SwipeRefreshLayout.java:195 androidx.swiperefreshlayout.widget.SwipeRefreshLayout$1.onAnimationEnd
    CircleImageView.java:107 androidx.swiperefreshlayout.widget.CircleImageView.onAnimationEnd
    ViewGroup.java:6912 android.view.ViewGroup.finishAnimatingView
    View.java:20272 android.view.View.draw
    ViewGroup.java:4337 android.view.ViewGroup.drawChild
    ViewGroup.java:4116 android.view.ViewGroup.dispatchDraw
    View.java:20369 android.view.View.draw
    View.java:19241 android.view.View.updateDisplayListIfDirty
    ViewGroup.java:4321 android.view.ViewGroup.recreateChildDisplayList
    ViewGroup.java:4294 android.view.ViewGroup.dispatchGetDisplayList
    View.java:19201 android.view.View.updateDisplayListIfDirty
    ViewGroup.java:4321 android.view.ViewGroup.recreateChildDisplayList
    ViewGroup.java:4294 android.view.ViewGroup.dispatchGetDisplayList
    View.java:19201 android.view.View.updateDisplayListIfDirty
    ViewGroup.java:4321 android.view.ViewGroup.recreateChildDisplayList
[...]

Issue #1444, opened 2020-11-16
V2 Android crash with modal


hoangtrucit:
## Description
After install `2.0.0-alpha.9`, it makes android crash without any message

## Screen shot
![image](https://user-images.githubusercontent.com/6366655/99411844-a04f9c00-2926-11eb-9f5a-ffe8f01f9705.png)

## Snack or minimal code example
[demo](https://github.com/hoangtrucit/demo-reanimated/blob/master/App.tsx)

## Package versions

- React: 17.0.1
- React Native: 0.63.3
- React Native Reanimated: 2.0.0-alpha.9
- NodeJS: v12.13.1

kyle-ssg:
+1 modals no longer work in alpha 9

```
11-18 11:42:31.600  4696  4696 E AndroidRuntime: com.facebook.jni.CppException: Value is null, expected an Object
11-18 11:42:31.600  4696  4696 E AndroidRuntime: 
11-18 11:42:31.600  4696  4696 E AndroidRuntime: Error: Value is null, expected an Object
11-18 11:42:31.600  4696  4696 E AndroidRuntime: 	at com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent(Native Method)
11-18 11:42:31.600  4696  4696 E AndroidRuntime: 	at com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent(NativeProxy.java:60)
11-18 11:42:31.600  4696  4696 E AndroidRuntime: 	at com.facebook.react.views.modal.ShowEvent.dispatch(ShowEvent.java:29)
11-18 11:42:31.600  4696  4696 E AndroidRuntime: 	at com.swmansion.reanimated.NodesManager.handleEvent(NodesManager.java:471)
11-18 11:42:31.600  4696  4696 E AndroidRuntime: 	at com.swmansion.reanimated.NodesManager.onEventDispatch(NodesManager.java:448)
11-18 11:42:31.600  4696  4696 E AndroidRuntime: 	at com.facebook.react.uimanager.events.EventDispatcher.dispatchEvent(EventDispatcher.java:114)
11-18 11:42:31.600  4696  4696 E AndroidRuntime: 	at com.facebook.react.views.modal.ReactModalHostManager$2.onShow(ReactModalHostManager.java:121)
11-18 11:42:31.600  4696  4696 E AndroidRuntime: 	at android.app.Dialog$ListenersHandler.handleMessage(Dialog.java:1358)
[...]

PR #1457, opened 2020-11-19
Don't process event with null native map

karol-bisztyga:
## Description

We don't want to process events with the `NativeMap` field set to `null`. Even though we use the `Nullable` annotation it still causes crashes on Android in some cases.

### Reproducing code

<details>
<summary>Tested with the following code</summary>

```
import React from 'react';
import { SafeAreaView, View, Text, StatusBar, Modal } from 'react-native';

const App = () => {
  return (
    <>
      <StatusBar barStyle="dark-content" />
      <SafeAreaView>
        <Modal animationType="slide" transparent={true} visible={true}>
          <View style={{ flex: 1 }}>
            <Text>Content modal</Text>
          </View>
        </Modal>
      </SafeAreaView>
    </>
  );
};

export default App;

```

</details>

Fixes https://github.com/software-mansion/react-native-reanimated/issues/1444
Fixes https://github.com/software-mansion/react-native-reanimated/issues/1447

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/Common/cpp/Registries/EventHandlerRegistry.cpp
+++ b/Common/cpp/Registries/EventHandlerRegistry.cpp
@@ -36,6 +36,10 @@ void EventHandlerRegistry::processEvent(jsi::Runtime &rt, std::string eventName,
   auto lastBracketCharactedPosition = eventPayload.size() - positionToSplit - 1;
   auto eventJSON = eventPayload.substr(positionToSplit,  lastBracketCharactedPosition);
 
+  if (eventJSON.compare(std::string("null")) == 0) {
+    return;
+  }
+
   auto eventObject = jsi::Value::createFromJsonUtf8(rt, (uint8_t*)(&eventJSON[0]), eventJSON.size());
 
   eventObject.asObject(rt).setProperty(rt, "eventName", jsi::String::createFromUtf8(rt, eventName));


sjchmiela:
Sorry for chiming in not mentioned, but what does "process" mean in:
> We don't want to **process** events with the `NativeMap` field set to `null`.

? From the stacktraces and code it looks like there are React Native events that have `null` body, like eg. [`RefreshEvent`](https://github.com/expo/expo/blob/a4cabf30a0cac8297c6c19ba5bd1cb33da0d6f1b/android/ReactAndroid/src/main/java/com/facebook/react/views/swiperefresh/RefreshEvent.java#L26). Would *not processing* mean they won't be dispatched to JS, or only to Reanimated worklets?

To bring my concern to this example — shouldn't we do sth like `eventBody = JSON.parse(eventBodyAsString) ?? {}` by either using `NativeMap:{}` in [`NativeProxy.h`](https://github.com/expo/expo/blob/79d295e4b31fa44ebb74eaf3bceb5329d9cfbb3d/android/expoview/src/main/cpp/headers/NativeProxy.h#L54) or setting `eventObject` to an empty object if it is `null` (instead of `return`ing)?

karol-bisztyga:
Hey. Those events would be dispatched to JS, the restriction is only for reanimated. You can check this by breaking the flow as I did in this or but without condition(always `return`) and then fire up something like this:
```
<PanGestureHandler
  onHandlerStateChange={(e) => {
    console.log('test1', e.nativeEvent.state);
  }}
  onGestureEvent={(e) => {
    console.log('test2', e.nativeEvent.state);
  }}>
  <View style={{ width: 200, height: 200, backgroundColor: 'orange' }} />
</PanGestureHandler>
```
You will receive events like before but the events for reanimated will not work:
```
const handle = useAnimatedGestureHandler({
  onActive: (e,c) => {
    console.log('rea', _WORKLET, e.state)
  }
});
//...
<PanGestureHandler
  onHandlerStateChange={handle}
  onGestureEvent={handle}>
  <Animated.View style={{ width: 200, height: 200, backgroundColor: 'purple' }} />
</PanGestureHandler>
```

piaskowyk (approved):

kyle-ssg:
Hi, unless I'm missing something this does not fix https://github.com/software-mansion/react-native-reanimated/issues/1444, I've patched EventHandlerRegistry, gradlew cleaned, re-run and still see the issue

image

PR #1458, opened 2020-11-19
Fix validator labels(add icons)

karol-bisztyga:
## Description

Add icons for the issue labels' names for the issue validator.


karol-bisztyga:
I'm ok with both options. Either we merge this or close it and remove the icons.

jakub-gonet (approved):
I think we can merge it in this state, it's not a big deal to change it later ¯\\_(ツ)_/¯

PR #1466, opened 2020-11-22
💄Refine type for createAnimatedComponent()

wcandillon:


jakub-gonet (changes requested):
This won't work, unfortunately – see the discussion under #719, especially [this comment](https://github.com/software-mansion/react-native-reanimated/issues/719#issuecomment-611435860). Very similar change was introduced in #696 and reverted in #791.

wcandillon:
Thanks a lot @jakub-gonet. I update the PR. @terrysahaidak does it look good?

terrysahaidak:
Yeah, seems to be, would be cool to have some backward compatibility because people already used to do "as something" I guess and it's a breaking change.

jakub-gonet:
With current types, [CI is failing](https://travis-ci.org/github/software-mansion/react-native-reanimated/jobs/745347355#L225).

wcandillon:
@jakub-gonet it passes now

jakub-gonet (approved):
LGTM, I want to make sure this doesn't create new issues with TS, so it'd be appreciated if you guys @terrysahaidak, @wcandillon, @mrousavy could test it out in your projects.

jakub-gonet:
Also changed the title a bit, @wcandillon sorry for removing lipstick :c

mrousavy:
@jakub-gonet @wcandillon I've tried it out in my project, and it doesn't work for the [RecyclerListView](https://github.com/Flipkart/recyclerlistview) component, because that doesn't directly derive from FC/C (`ComponentType`), but rather from a custom abstraction that extends `React.Component`. (see [recyclerlistview/utils/ComponentCompact.ts](https://github.com/Flipkart/recyclerlistview/blob/master/src/utils/ComponentCompat.ts)). I wonder what's disrupting those types, I'll have a quick look and try some things out.

Other than that, it works. Happy to see some type support, even noticed I had a typo in a prop somehwere ;)

wcandillon:
@mrousavy For such scenario, would it be accepted to use  const ReanimatedRecyclerListView = Reanimated.createAnimatedComponent<any>(RecyclerListView); ?

mrousavy:
@wcandillon I'm not sure I understand, the code snippet you sent me is exactly what's causing the TS error. 🤔

jakub-gonet:
> Yeah, seems to be, would be cool to have some backward compatibility because people already used to do "as something" I guess and it's a breaking change.

Do you think that there are many people affected by this? I'm not sure how to maintain backward compatibility with types. Till now, `createAnimatedComponent()` was types as `any` so we can't really adapt to this.

Besides that, we're on alpha so we have a unique opportunity to break as many things as we like :>

But I think this won't have that big impact so we can backport it to v1 too.

jakub-gonet:
@wcandillon ping

wcandillon:
haha sorry it looks like github suckedin the `<>` my comment should have read: `const ReanimatedRecyclerListView = Reanimated.createAnimatedComponent<any>(RecyclerListView); ` As @jakub-gonet mentioned, this should be acceptable.

mrousavy:
Yes, that looks like the right workaround. Maybe I can fix this by restructuring RLV. LGTM then!

jakub-gonet:
Welp, in that case let's merge it and we'll see if anyone's setup breaks after the next release 😄

jakub-gonet:
@wcandillon I just saw #1537 and seems like we allow FC as well. Maybe this is worth fixing before we release that?

wcandillon:
that's done in #1539

maxammann:
@jakub-gonet I have a problem with this change together with react-native-svg:

```
const AnimatedPath = Animated.createAnimatedComponent<any, any>(Path)
```

Path does not have a style attribute. Therefore I can not use this like:

```
const AnimatedPath = Animated.createAnimatedComponent<undefined, PathProps>(Path)
```

Is this the intention with this PR? Or should there be a better way? I think a lot of people will use reanimated with react-native-svg

mrousavy:
@maxammann what happens when you try

```ts
const AnimatedPath = Animated.createAnimatedComponent<never, PathProps>(Path)
```

or use `unknown` instead of `never`?

maxammann:
`never` works! I'm new to typescript so I actually don't know what this does. But I'm fine with that typing!
Let's hope other people will find this solution here :)

jakub-gonet:
`unknown` is a typesafe analog to the `any` and you can assign to it every value. `never` is a bottom type for TS and you can only assign `never` to `never`, preventing usage of any object there.

PR #1467, opened 2020-11-23
[v1] Update `color` and `interpolateColors` return type to allow strings

chj-damon:
## Checklist

- [x] Updated TS types
- [x] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes


react-native-reanimated.d.ts
```diff
@@ -357,7 +357,7 @@ declare module 'react-native-reanimated' {
         inputRange: ReadonlyArray<Adaptable<number>>;
         outputColorRange: (string | number)[];
```
jakub-gonet, line 358: Could you also change this type to `ReadonlyArray<Adaptable<number | string>>`, like in [`interpolateNode`](https://github.com/software-mansion/react-native-reanimated/pull/1467/files#diff-3eb03eea3c56460a1bdcef23b6d54e490742744b667bb3ebacc1927b8f7d2b49R57)?

jakub-gonet (commented):
LGTM, left one comment.

jakub-gonet:
Also, some tests in the [TS test file](https://github.com/software-mansion/react-native-reanimated/blob/2a6c6139a7ac80b4072bb812fdb0d0cad2fee1bc/react-native-reanimated-tests.tsx) would be appreciated.

react-native-reanimated.d.ts
chj-damon: sure thing.

react-native-reanimated-tests.tsx
```diff
@@ -499,3 +501,20 @@ function interpolateNodeTest() {
...
+  const b = useValue(255);
+  const a = useValue(255);
+  return color(r, g, b, a);
+}
+
+function interpolateColorsTest() {
+  const animationValue = useValue(0);
+  const color = interpolateColors(animationValue, {
+    inputRange: [0, 1],
+    outputColorRange: ['red', 'blue'],
```
jakub-gonet, line 517: This won't work in the real code, we only return string RGB values in the format of `rgba(r,g,b,a)` (where variables are numbers) on the web.

chj-damon: then, what it would be like?

jakub-gonet: Or actually, it'll work, we use `processColor` internally. I mixed `outputColorRange` with the return type, my bad.

jakub-gonet (approved):

PR #1470, opened 2020-11-23
Shared style

### LINKED ISSUE #1263: unable to use same useAnimatedStyle to multiple views components
## Description

I've created a `useAnimatedStyle` hook and I'm passing the animated hook style result to the two view components. When I trigger the change only one of the components is getting updated.
### Screenshots
ezgif com-video-to-gif(3)

## Steps To Reproduce

1. Create a `useAnimatedStyle` hook and pass the same hook to multiple components. trigger the update and you'll see only one of them getting updated.

### Expected behavior
Styles of all the components should be updated on using the same style hook

### Actual behavior
Style of only one component is getting updated

I've to create two separate style hook, `widthStyle` and `widthStyle2` to make it work. using `widthStyle` to the two components does not work 

```javascript
import React from 'react';
import Animated, {
  useAnimatedStyle,
  useSharedValue,
import {View, Button, Text} from 'react-native';

const Test = () => {
  const width = useSharedValue(300);
  const widthStyle = useAnimatedStyle(() => {
    return {
      width: width.value,
    };
  });
  const widthStyle2 = useAnimatedStyle(() => {
    return {
      width: width.value,
    };
  });
  return (
    <View>
      <Animated.View
        style={[{backgroundColor: 'red', height: 100}, widthStyle]}>
        <Text>View 1</Text>
      </Animated.View>
      <Animated.View
        style={[{backgroundColor: 'red', marginTop: 100, height: 100}, widthStyle]}>
        <Text>View 2</Text>
      </Animated.View>
      <Button
        title="Click"
        onPress={() => {
          width.value = Math.random() * 100 + 300;
        }}
      />
    </View>
  );
};

export default Test;

```
if this is the expected behaviour then how can i use same hook to multiple components?

## Package versions

- React: "~16.11.0"
- React Native: "~0.62.2",
- React Native Reanimated: "^2.0.0-alpha.6",

piaskowyk:
## Description
I added the possibility to use one instance of `useAnimationStyle()` for many components. `AnimationStyle` is connected with view by ViewDescriptor object. I changed this to a set of `ViewDescriptors` and now `AnimationStyle` can be shared between many components. 

Fixes #1263 

## Example of code
<details>
<summary>code</summary>

```Js
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  Easing,
import { Button, ScrollView } from 'react-native';
import React from 'react';

export default function SharedAnimatedStyleUpdateExample(props) {

  const config = {
    duration: 500,
  };

    return {
    };
  });

  const renderList = () => {
    const items = [];
    for (let i = 0; i < 100; i++) {
      items.push(
        <Animated.View
          key={i}
          style={[
            { width: 100, height: 5, backgroundColor: 'black', margin: 1 },
            style,
          ]}
        />
      );
    }
    return items;
  };

  return (
    <ScrollView
      style={{
        flex: 1,
        flexDirection: 'column',
      }}>
      <Button
        title="toggle"
        onPress={() => {
        }}
      />
      {renderList()}
    </ScrollView>
  );
}
```

</details>

<details>
<summary>code2</summary>

```Js
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  Easing,

  const [counter, setCounter] = useState(0);
  const [counter2, setCounter2] = useState(0);
  const [itemList, setItemList] = useState([]);
  const [toggleState, setToggleState] = useState(false);

  const config = {
    duration: 500,
  };

    return {
    };
  });

  const staticObject = <Animated.View
    style={[
      { width: 100, height: 8, backgroundColor: 'black', margin: 1 },
      style,
    ]}
  />

  const renderItems = () => {
    let output = []
    for(let i = 0; i < counter; i++) {
      output.push(
        <Animated.View
        key={i + 'a'}
          style={[
            { width: 100, height: 8, backgroundColor: 'blue', margin: 1 },
            style,
          ]}
        />
      )
    }
    return output
  }

  return (
    <View
      style={{
        flex: 1,
        flexDirection: 'column',
      }}>
      <Button
        title="animate"
        onPress={() => {
        }}
      />
      <Button
        title="increment counter"
        onPress={() => {
          setCounter(counter + 1)
        }}
      />
      <Button
        title="add item to static lists"
        onPress={() => {
          setCounter2(counter2 + 1)
          setItemList([...itemList, <Animated.View
            key={counter2 + 'b'}
            style={[
              { width: 100, height: 8, backgroundColor: 'green', margin: 1 },
              style,
            ]}
          />])
        }}
      />
      <Button
        title="toggle state"
        onPress={() => {
          setToggleState(!toggleState)
        }}
      />
      <Animated.View
        style={[
          { width: 100, height: 8, backgroundColor: 'orange', margin: 1 },
          style,
        ]}
      />
      {toggleState && <Animated.View
        style={[
          { width: 100, height: 8, backgroundColor: 'black', margin: 1 },
          style,
        ]}
      />}
      {toggleState && staticObject}
      {renderItems()}
      {itemList}
    </View>
  );
}

```

</details>

## Change
before changes:

final result:


mrousavy:
Awesome! I've been having _huge_ performance problems with reanimated v2 since I am animating multiple views with a sort of parallax effect, I'm sure this is a big step towards solving those performance issues. 👍

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -144,6 +144,31 @@ std::shared_ptr<ShareableValue> ShareableValue::adapt(jsi::Runtime &rt, const js
...
+  }
+  sv->adapt(rt, value, valueType);
+  return sv;
+}
+
+void ShareableValue::adaptSet(jsi::Runtime &rt, std::shared_ptr<ShareableValue> &originalValue, const jsi::Value &value, ValueType objectType) {
+  auto object = value.asObject(rt);
+  auto setItem = object.getProperty(rt, "__setItem");
+  if(!setItem.isUndefined()) {
+    //temorarty, to change on std::set
```
jakub-gonet, line 164: Typo.

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -144,6 +144,31 @@ std::shared_ptr<ShareableValue> ShareableValue::adapt(jsi::Runtime &rt, const js
...
+    frozenArray.push_back(adapt(rt, setItem, module));
+  }
+  else {
+    throw "Invalid value type";
```
jakub-gonet, line 168: This could be more descriptive.

terrysahaidak:
hey @mrousavy can you share us these performance problems in an issue? perhaps it's the same problem I had recently

mrousavy:
@terrysahaidak Yeah, [here's a video demo (imgur.com)](https://i.imgur.com/8iWW7Wn.mp4). The structure of this view is the following:
I have a view that horizontally displays all my individual "story" screens, and you can "scroll" between them using a horizontal scroll gesture, kinda like an image carousel. I have implemented this using a `PanGestureHandler` instead of a `ScrollView`, since that gives me more control.

So the whole horizontal view which contains each individual story screen is translated along the X axis depending on the swipe gesture. Then, I've tried to implement some sort of cool parallax effect with the headers, by making them go into the next screen by a tiny bit, to reveal to the user what the next page has to offer. The problem here is, that I can only achieve that by translating each header individually, depending on the carousel's `translateX`. And that's causing huge performance problems, since I now have about 30 `useAnimatedStyle` hooks for 30 stories, just for the headers! In REA v1 this was no problem, since I just passed the nodes in there.

I tried to play around with virtualization and eventually came up with a way to unmount every header that's not in viewport (and used `removeSubclippedViews` for the other views to gain additional performance), but it was still terrible, even with just 5 headers rendered at once. (Even on my iPhone 11 it has terrible FPS, on an iPhone 8 it takes about 10 seconds to even open that screen)

src/reanimated2/UpdateProps.js
```diff
@@ -27,23 +27,33 @@ const ColorProperties = makeShareable(colorProps);
...
-  if (Platform.OS !== 'web') {
-    Object.keys(updates).forEach((key) => {
-      if (ColorProperties.indexOf(key) !== -1) {
-        updates[key] = processColor(updates[key]);
-      }
-    });
-  }
+    if (Platform.OS !== 'web') {
+      Object.keys(_updates).forEach((key) => {
+        if (ColorProperties.indexOf(key) !== -1) {
```
jakub-gonet, line 35:
```suggestion
        if (ColorProperties.includes(key)) {
```

src/reanimated2/js-reanimated/MutableSet.js
```diff
@@ -0,0 +1,12 @@
+export default class MutableSet {
+  __mutableSet = true;
+  value = [];
+
+  add = (item) => {
+    this.value.push(item);
```
jakub-gonet, line 6: Judging by name – if it's a set, then it shouldn't allow duplicate values? You can use [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) for it.

src/reanimated2/js-reanimated/index.web.js
```diff
@@ -5,7 +5,7 @@ const reanimatedJS = new JSReanimated();
 global._frameTimestamp = null;
 
 export const _updatePropsJS = (viewTag, viewName, updates, viewRef) => {
-  if (viewRef.current && viewRef.current._component) {
+  if (viewRef && viewRef._component) {
```
jakub-gonet, line 8:
```suggestion
  if (viewRef?._component) {
```

Example/ios/Podfile.lock
```diff
@@ -453,4 +453,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 45e9f95c2fdf00b7647da45a852dd38e0f42e128
 
-COCOAPODS: 1.9.1
+COCOAPODS: 1.10.0
```
jakub-gonet, line 456: Can we settle on one Cocoapods version? 😄

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -144,7 +148,7 @@ std::shared_ptr<ShareableValue> ShareableValue::adapt(jsi::Runtime &rt, const js
   return sv;
 }
 
-jsi::Value ShareableValue::getValue(jsi::Runtime &rt) {
+jsi::Value ShareableValue::	getValue(jsi::Runtime &rt) {
```
jakub-gonet, line 151: I think this should be left as it is.

Common/cpp/SharedItems/MutableSet.cpp
```diff
@@ -0,0 +1,103 @@
...
+  std::vector<jsi::PropNameID> result;
+  result.push_back(jsi::PropNameID::forUtf8(rt, std::string("__mutableSet")));
+  result.push_back(jsi::PropNameID::forUtf8(rt, std::string("value")));
+  result.push_back(jsi::PropNameID::forUtf8(rt, std::string("add")));
+  result.push_back(jsi::PropNameID::forUtf8(rt, std::string("clear")));
+  return result;
+}
+
+void MutableSet::init(jsi::Runtime &rt, const jsi::Value &value) {
+  if (value.isUndefined()) return;
```
jakub-gonet, line 74: Please use braces.

Common/cpp/SharedItems/MutableSet.cpp
```diff
@@ -0,0 +1,103 @@
...
+    auto object = value.asObject(rt);
+    if (object.isArray(rt)) {
+      auto array = object.asArray(rt);
+      for (size_t i = 0, size = array.size(rt); i < size; i++) {
+        items.insert(ShareableValue::adapt(rt, array.getValueAtIndex(rt, i), module));
+      }
+    }
+    else {
+      items.insert(ShareableValue::adapt(rt, value, module));
+    }
```
jakub-gonet, line 85: I'm not that experienced with JSI – do we have to use `asObject` to make `adapt` work correctly? Because this branch is pretty much the same as the not-object case.

jakub-gonet:
Can you resolve the hooks file conflict?

src/reanimated2/UpdateProps.js
piaskowyk: I'm not sure why but when I changed `indexOf` to `includes` I received exception: 
image

Common/cpp/SharedItems/MutableSet.cpp
piaskowyk: I want to detect when `value` is an array, but `value` is a instance of `jsi::Value` and this type doesn't contains `isArray()` function. I must cast `jsi::Value` to `jsi::Object` before.

src/reanimated2/UpdateProps.js
jakub-gonet: This is very weird, should work OOTB. I searched for usages of `indexOf` in case it's treated special and found none. There's nothing in Hermes issues either.

Maybe @karol-bisztyga or @Szymon20000 know why this doesn't work?

mrousavy: @jakub-gonet @piaskowyk shouldn't `includes` be also included in [the blacklist array](https://github.com/software-mansion/react-native-reanimated/blob/aa315669402dd3bcdf882197540275d075eef12c/plugin.js#L72-L123)?

jakub-gonet: Must've missed that. Yeah, adding this to the blacklist should work.

kmagiera:
Do we really need native set support for this one? Why can't we make an array of descriptors and just assign it in one go instead of making changes with every object being added.

I haven't checked the full source but it seem problematic that we use set.add and never use set.remove in this PR.

src/reanimated2/Hooks.js
```diff
@@ -37,6 +37,32 @@ export function useSharedValue(init, shouldRebuild = true) {
   return ref.current.mutable;
 }
 
+export function useMutableSet(init) {
```
karol-bisztyga, line 40: this does not seem to be a hook

src/reanimated2/Hooks.js
```diff
@@ -37,6 +37,32 @@ export function useSharedValue(init, shouldRebuild = true) {
...
+  const data = {
+    __mutableSet: true,
+    batchToInsert: [],
+    waitForInsertSync: false,
+    mapper: makeMutable(init),
```
karol-bisztyga, line 45: why is it called `mapper`?

piaskowyk:
Finally, works and ready for review, I would be grateful for any suggestions.

charles-goode:
When this is ready, will it also work for useAnimatedProps? I didn't see issues reporting that the same need for sharing is present with animated props, so I wasn't sure if that had also been considered.

mrousavy:
Hate to "bump" this, but do you guys think this will make it into 2.3.0? @piaskowyk @jakub-gonet let me know if you need any help testing this change 🙏

piaskowyk:
The good news, this PR will be release in 2.3 🎉

piaskowyk:
> When this is ready, will it also work for useAnimatedProps? I didn't see issues reporting that the same need for sharing is present with animated props, so I wasn't sure if that had also been considered.

Yes, in Reanimated 2 `useAnimatedStyle` and `useAnimatedProps` has the same implementation:
https://github.com/software-mansion/react-native-reanimated/blob/3c2bf7c6037b95c1e07b07e0cb9c524a9eea9412/src/reanimated2/Hooks.ts#L608

Szymon20000 (approved):

bobsmits:
@piaskowyk Any sight on a alpha.2 release for 2.3? I would love to get started with this.

yanush:
This is not working for me in 2.3.0-beta.3
Any suggestions?

PR #1479, opened 2020-11-26
Allow interpolateColor()'s outputRange to contain numbers

wcandillon:


react-native-reanimated.d.ts
```diff
@@ -455,7 +455,7 @@ declare module 'react-native-reanimated' {
     export function interpolateColor(
       value: number,
       inputRange: number[],
-      outputRange: string[],
+      outputRange: (string | number)[],
```
jakub-gonet, line 458: We can also add `readonly` modifier here, we don't mutate those arrays.

jakub-gonet (approved):
LGTM, thanks!

jakub-gonet:
One more thing, could you add some tests in the TS file?

mrousavy:
The `interpolateColor()` function returns `string | number`, but a View Style doesn't allow `number` for colors such as `backgroundColor`, did you also notice this? That means, this code:

```ts
const x = useAnimatedStyle(() => ({ backgroundColor: interpolateColor(...) }), []); 
```

will raise a `ts-error` since `number` is not assignable to `backgroundColor`.

wcandillon:
@mrousavy This is fixed in #1531 ?

mrousavy:
@wcandillon no, since `ColorValue` is defined as `string | unique symbol`.

wcandillon:
Using #1531 as a patch, I'm able to return `const x = useAnimatedStyle(() => ({ backgroundColor: interpolateColor(...) }), []);`

mrousavy:
@wcandillon oh, you're right I missed the conditional `extends`, my bad! Works for me 👍

wcandillon:
Let's merge it ;-)

jakub-gonet:
@wcandillon, could you add some tests to the TS test file too?

wcandillon:
@jakub-gonet done, CI should pass.

wcandillon:
@jakub-gonet the CI failure seems unrelated, is that possible?

jakub-gonet:
Sorry for the delay, let's trigger CI and see if it's working now.

PR #1482, opened 2020-11-27
Refactor of global values

piaskowyk:
## Motivation
Using many global values is not a good practise. Global values can be changed everywhere and it makes potential debugging difficult. Local scope is also better for memory usage.

## Changes
- I removed `_updatePropsJS()` from global scope and moved it to local scope.
- I changed `_globalSetter()` na `_setGlobalConsole()` because this setter was used only to add `console` to global object, [code here](https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/core.js#L234). This blocks the possibility to create global values in runtime, we have control on content of global scope now.


src/reanimated2/UpdateProps.js
```diff
@@ -1,7 +1,8 @@
-/* global _updateProps _updatePropsJS */
+/* global _updateProps */
 import { processColor } from './Colors';
 import { makeShareable } from './core';
 import { Platform } from 'react-native';
+import { _updatePropsJS } from './js-reanimated/index';
```
jakub-gonet, line 5: This will load the JS version regardless if it's needed or not. Maybe we should change [those](https://github.com/software-mansion/react-native-reanimated/pull/1482/files#diff-109d22d762ccae27889a4992c5797583a225c263244aa7dcc4d2df50415ca608R40-R41) to something along with this:

```
const updatePropsInternal = _updateProps ?? require('./js-reanimated/index'). _updatePropsJS
```
to load it on demand?

(`._updatePropsJS` instead of `?._updatePropsJS` because `_updatePropsJS` should be available at all times, so safe navigation there is not needed.)

src/reanimated2/core.js
```diff
@@ -226,10 +226,15 @@ export const runOnJS = (fun) => {
...
+  //   warn: runOnJS(capturableConsole.warn),
+  //   error: runOnJS(capturableConsole.error),
+  // });
   const console = {
     log: runOnJS(capturableConsole.log),
     warn: runOnJS(capturableConsole.warn),
     error: runOnJS(capturableConsole.error),
   };
-  _globalSetter('console', console); // eslint-disable-line
+  _setGlobalConsole(console); // eslint-disable-line
```
jakub-gonet, line 234: Why you used `eslint-disable-line` here?
if it complains about not resolving `_setGlobalConsole` then you should use `/* global _setGlobalConsole */`, like in updateProps file.

jakub-gonet (commented):
JS looks good, can't really assess the correctness of C++ and plugin code.

src/reanimated2/UpdateProps.js
karol-bisztyga: you can give it a shot but I've tried something similar some time ago and it didn't work. I think the proper solution for conditional imports is to put them in the files with certain extensions(like `.web.js`, `ios.js`, `android.js`)

reference: https://reactnative.dev/docs/platform-specific-code#platform-specific-extensions
example in our repo: https://github.com/software-mansion/react-native-reanimated/tree/master/src/reanimated2/platform-specific

I agree it should be imported only if needed 👍

karol-bisztyga (approved):

karol-bisztyga (commented):
Please wait for @jakub-gonet to approve, for me it looks alright

src/reanimated2/UpdateProps.js
jakub-gonet: This generates a lot of files if want to keep it modularized.

If it doesn't work then we can stick to the older version, this change is something we should in mind though.

jakub-gonet (approved):

PR #1483, opened 2020-11-27
Fix web app startup for Example app

karol-bisztyga:
## Description

For the new expo versions, there is no such option as `--web-only`. Therefore we decided to remove it at the same time keeping it as a fallback instruction in the scripts for backward compatibility.

This originally showed up in our playground repo as an issue: https://github.com/software-mansion-labs/reanimated-2-playground/issues/19

It's been resolved and checked there: https://github.com/software-mansion-labs/reanimated-2-playground/pull/25/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R11

Now it's going to work for older and newer expo versions.

### DIFF
--- a/Example/ios/Podfile.lock
+++ b/Example/ios/Podfile.lock
@@ -1,14 +1,14 @@
 PODS:
   - boost-for-react-native (1.63.0)
   - DoubleConversion (1.1.6)
-  - FBLazyVector (0.63.1)
-  - FBReactNativeSpec (0.63.1):
+  - FBLazyVector (0.63.0)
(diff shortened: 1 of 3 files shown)


jakub-gonet (approved):

piaskowyk:
Awesome, this change I have needed 🙌  
I think that we could add `expo-asset` dependency to package.json because `expo 3.28.5` required this package to run a web panel.

piaskowyk (approved):

PR #1491, opened 2020-11-30
Handle worklet errors on hermes

karol-bisztyga:
## Description

This PR refers to a situation in which there is a javascript error in a worklet that's being executed by jsi.
On iOS with JSC, it works as expected which means it throws a `std::exception` from jsi's `call` invocation.
Unfortunately on Android, it does not. This c++ code is being scheduled on the UI thread by the scheduler and the error that appears in `Scheduler.java` is: `com.facebook.JNI.UnknownCppException: Unknown`.
This comes with two problems:
1. Our example app crashes with `A/libc: Fatal signal 6 (SIGABRT), code -6`
This may be a problem with how we build the example app(As @tmikov suggested maybe e.g. linking problem)
2. When reanimated is installed in any other app it shows Redbox with an error message: `Unknown` - this gives almost no information on what went wrong.

The good news is that some error is thrown in c++ and can be caught. The bad news is that it's not a `std::exception` and we don't know its type so we cannot really get the proper message. Unless it changes we only can rethrow an error with a hardcoded message which indicates that something went wrong in a worklet. That's what is done in this PR.

Edit: Using the babel plugin I was able to fetch the path of the file containing the worklet in which the error occurred so it's now attached to the hardcoded error message.

## Changes

Just added catch clauses that catch anything for try blocks that include worklet execution.

I also repaired access to a module when using the `errorHandler`.

## Testing code

<details>
<summary>code</summary>

```
import { View, Text, Button } from 'react-native';

const Bug1 = () => {
  const val = {};

  runOnUI(() => {
    'worklet';
    val.f(4); // will call https://github.com/software-mansion/react-native-reanimated/blob/master/Common/cpp/SharedItems/ShareableValue.cpp#L347
  })();

  return <></>;
};

const Bug2 = () => {
  const wrklt = (x) => {
    'worklet';
    x.f();
  };

  runOnUI(() => {
    'worklet';
    wrklt(9); // will call https://github.com/software-mansion/react-native-reanimated/blob/master/Common/cpp/SharedItems/ShareableValue.cpp#L298
  })();

  return <></>;
};

  const [bug, setBug] = useState(0);

  const renderBug = () => {
    if (bug === 1) {
      return <Bug1 />;
    }
    if (bug === 2) {
      return <Bug2 />;
    }
  };

  return (
    <View>
      <Text>Pick a bug:</Text>
      <Button
        title="bug 1"
        onPress={() => {
          setBug(1);
        }}
      />
      <Button
        title="bug 2"
        onPress={() => {
          setBug(2);
        }}
      />
      {renderBug()}
    </View>
  );
}

```

</details>

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/reanimated2/__snapshots__/plugin.test.js.snap
+++ b/src/reanimated2/__snapshots__/plugin.test.js.snap
@@ -21,7 +21,7 @@ var f = function () {
   };
   _f.asString = \\"function f(){const{x,objX}=jsThis._closure;{return{res:x+objX.x};}}\\";
   _f.__workletHash = 10184269015616;
-  _f.__fileName = \\"${ process.cwd() }/jest tests fixture (4:4)\\";
+  _f.__location = \\"${ process.cwd() }/jest tests fixture (4:4)\\";
 
   global.__reanimatedWorkletInit(_f);
 
@@ -38,7 +38,7 @@ exports[`babel plugin doesn't capture globals 1`] = `
   _f._closure = {};
   _f.asString = \\"function f(){console.log(\\\\\\"test\\\\\\");}\\";
   _f.__workletHash = 6265462104213;
-  _f.__fileName = \\"${ process.cwd() }/jest tests fixture (2:6)\\";
+  _f.__location = \\"${ process.cwd() }/jest tests fixture (2:6)\\";
 
   global.__reanimatedWorkletInit(_f);
 
@@ -73,7 +73,7 @@ function Box() {
     };
     _f.asString = \\"function _f(){const{offset}=jsThis._closure;{return{transform:[{translateX:offset.value*255}]};}}\\";
     _f.__workletHash = 7114514849439;
-    _f.__fileName = \\"${ process.cwd() }/jest tests fixture (7:48)\\";
+    _f.__location = \\"${ process.cwd() }/jest tests fixture (7:48)\\";
 
     global.__reanimatedWorkletInit(_f);
 
--- a/src/reanimated2/__snapshots__/plugin.test.js.snap
+++ b/src/reanimated2/__snapshots__/plugin.test.js.snap
@@ -21,6 +21,7 @@ var f = function () {
   };
   _f.asString = \\"function f(){const{x,objX}=jsThis._closure;{return{res:x+objX.x};}}\\";
   _f.__workletHash = 10184269015616;
+  _f.__fileName = \\"${ process.cwd() }/jest tests fixture\\";
 
   global.__reanimatedWorkletInit(_f);
 
@@ -37,6 +38,7 @@ exports[`babel plugin doesn't capture globals 1`] = `
   _f._closure = {};
   _f.asString = \\"function f(){console.log(\\\\\\"test\\\\\\");}\\";
   _f.__workletHash = 6265462104213;
+  _f.__fileName = \\"${ process.cwd() }/jest tests fixture\\";
 
   global.__reanimatedWorkletInit(_f);
 
@@ -71,6 +73,7 @@ function Box() {
     };
     _f.asString = \\"function _f(){const{offset}=jsThis._closure;{return{transform:[{translateX:offset.value*255}]};}}\\";
     _f.__workletHash = 7114514849439;
+    _f.__fileName = \\"${ process.cwd() }/jest tests fixture\\";
 
     global.__reanimatedWorkletInit(_f);
 
--- a/src/reanimated2/__snapshots__/plugin.test.js.snap
+++ b/src/reanimated2/__snapshots__/plugin.test.js.snap
@@ -21,7 +21,7 @@ var f = function () {
   };
   _f.asString = \\"function f(){const{x,objX}=jsThis._closure;{return{res:x+objX.x};}}\\";
   _f.__workletHash = 10184269015616;
-  _f.__fileName = \\"${ process.cwd() }/jest tests fixture\\";
+  _f.__fileName = \\"${ process.cwd() }/jest tests fixture (4:4)\\";
 
   global.__reanimatedWorkletInit(_f);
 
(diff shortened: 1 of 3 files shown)


src/reanimated2/__snapshots__/plugin.test.js.snap
```diff
@@ -21,6 +21,7 @@ var f = function () {
   };
   _f.asString = \\"function f(){const{x,objX}=jsThis._closure;{return{res:x+objX.x};}}\\";
   _f.__workletHash = 10184269015616;
+  _f.__fileName = \\"/Users/karolbisztyga/Desktop/workspace/react-native-reanimated/jest tests fixture\\";
```
jakub-gonet, line 24: This should be mocked in tests.

piaskowyk:
Did you try to get information about exception in this way?
```
catch(...) {
    auto expPtr = std::current_exception();
    try {
        if(expPtr) std::rethrow_exception(expPtr);
    }
    catch(const std::exception& e)  {
        std::cout << e.what();
    }
}
```
https://en.cppreference.com/w/cpp/error/current_exception

karol-bisztyga:
> Did you try to get information about exception in this way?
> 
> ```
> catch(...) {
>     auto expPtr = std::current_exception();
>     try {
>         if(expPtr) std::rethrow_exception(expPtr);
>     }
>     catch(const std::exception& e)  {
>         std::cout << e.what();
>     }
> }
> ```
> 
> [en.cppreference.com/w/cpp/error/current_exception](https://en.cppreference.com/w/cpp/error/current_exception)

I didn't try it but I've read about it and as pointed out [here](https://stackoverflow.com/a/38220498)(look at the first comment on the linked answer) it still handles only `std::exception`s.

piaskowyk (approved):

src/reanimated2/__snapshots__/plugin.test.js.snap
karol-bisztyga: fixed fc142e9

piaskowyk:
I tried to check the type of this exception. I used ABI and this below code detect `facebook::jsi::JSError` exception type. I tried to catch `facebook::jsi::JSError` and `facebook::jsi::JSException` but without success unfortunately. So It's just a suggestion, maybe this information will be valuable for you.
My example code:
```
catch(...) {
  jsi::Value fileName = jsThis.getProperty(rt, "__fileName");
  std::string str = "Javascript worklet error";
  if (fileName.isString()) {
    std::string to_demangle = abi::__cxa_current_exception_type()->name();
    std::string exceptionTypeName = "";
    int status = 0;
    char * buff = __cxxabiv1::__cxa_demangle(to_demangle.c_str(), NULL, NULL, &status);
    if (!buff) exceptionTypeName = to_demangle;
    std::string demangled = buff;
    std::free(buff);
    exceptionTypeName = demangled;
    str += "\nIn file: " + fileName.asString(rt).utf8(rt) + "\nError type: " + exceptionTypeName;
  }
  module->errorHandler->setError(str);
  module->errorHandler->raise();
}
```
Result:
image

Example usage of ABI:
https://godbolt.org/z/qfPK91

karol-bisztyga:
@kmagiera thx, updated the code

We discussed this with @piaskowyk and it was tested in the following manner:
```
try {} catch(std::exception &e){} catch(...){/* type checking code */}
```
so if the error's type is read properly it's kind of weird that it didn't hit the first `catch` since `facebook::jsi::JSError` inherits from `std::exception`. The way I see it, either ABI doesn't read the type properly or it's just some kind of undefined behavior(at least repeatable).

kmagiera:
Have you tried puttin JSIexception in the catch block?

piaskowyk:
`JSIException` also inherits from `std::exception`
std::exception <- JSIException <- JSError

mrousavy:
Using the `fileName` argument in the `processWorkletFunction` essentially also allows us to show more detailer error messages, right?

So instead of `Attempting to run synchronous function on UI thread blablabla wrap it in runOnJS()` we can show the location of that misuse and possibly also a line number. If we have a line number, we can use the RedBox API to actually show a snippet, just like in normal JS errors. 🤔

karol-bisztyga:
> Using the `fileName` argument in the `processWorkletFunction` essentially also allows us to show more detailer error messages, right?
> 
> So instead of `Attempting to run synchronous function on UI thread blablabla wrap it in runOnJS()` we can show the location of that misuse and possibly also a line number. If we have a line number, we can use the RedBox API to actually show a snippet, just like in normal JS errors. 🤔

Hmm 🤔 not exactly I guess. We can find out where the worklet is defined and where it ends but still, the call of some js function would be somewhere inside it. Where exactly - that information we do not have. Pasting the whole worklet isn't the best idea I guess - those functions might be big in some cases. For now, we provide a called function's name in the error message what's helpful. We can think however of adding the worklet's location there but I'm not sure it's that necessary. We could open a new issue if this PR is merged.

Anyway thanks for your suggestion ✌️

zrebcu411 (approved):

mrousavy:
Oh, I didn't know that the JS function's name was already mentioned in the error message! Last time I had such errors was on alpha 9 I think. 👍

karol-bisztyga:
[This](https://github.com/software-mansion/react-native-reanimated/pull/1361) was introduced in alpha 8 actually. Doesn't really matter, just saying 😜

mrousavy:
@karol-bisztyga oh, that's what those `{}` should represent! This actually doesn't work with callbacks. 

This works:

```ts
  const testfunc = () => {
    console.log("yeet");
  };
  const idk = useWorkletCallback(() => {
    testfunc();
  }, []);
```

This doesn't work:

```ts
  const testfunc = useCallback(() => {
    console.log("yeet");
  }, []);
  const idk = useWorkletCallback(() => {
    testfunc();
  }, []);
```


Do you guys want me to open a separate issue for this, or will this get reworked anyways?

karol-bisztyga:
Ha! Nice catch, thanks. If you have some time, please open a new issue, this got totally beyond the current one.

Issue #1498, opened 2020-12-01
Reanimated 2 causes uimanager commands to run out of order


kmagiera:
## Description
The issue has been identified on one of our partner's app and results in the crash that's been reported independently in #1318 (redbox with message "parentNode or node is a required parameter"). Despite that old issue being resolved I believe we haven't got to the bottom of the problem at that time, so I believe it has not been fixed and, as observed in other apps, it can still happen.

The crash can happen in various places and in particular in this line in Reanimated 1 code: https://github.com/software-mansion/react-native-reanimated/blob/master/ios/REANodesManager.m#L357

After investigating I found out that the crash is due to the native commands being run in a different order than the order at which they are dispatched in JS. I.e. we dispatch 1) createNode 2) connectToView 3) disconnectFromView 4) dropNode while sometimes we may run command no (3) first which results in the crash.

Further investigation shown that the culprit is the synchronous layout mechanism that, when reanimated 1 and 2 code is mixed, can turn on for random UI-manager batches. The relevant fragment that is problematic in particular is here: https://github.com/software-mansion/react-native-reanimated/blob/master/ios/REAModule.m#L174

```objc
- (void)uiManagerWillPerformMounting:(RCTUIManager *)uiManager {
  ...
  NSArray<AnimatedOperation> *operations = _operations;
  _operations = [NSMutableArray new];

  REANodesManager *nodesManager = _nodesManager;

  [uiManager addUIBlock:^(__unused RCTUIManager *manager, __unused NSDictionary<NSNumber *, UIView *> *viewRegistry) {
    for (AnimatedOperation operation in operations) {
      operation(nodesManager);
    }
    [nodesManager operationsBatchDidComplete];
  }];
}
```

It turns out that under certain circumstances the blocks that we enqueue with `addUIBlock` can run out of order.

[...]

Issue #1318, opened 2020-10-06
Exception thrown while executing UI block: 'parentNode' is a required parameter


terrysahaidak:
## Description

While working on React Native Gallery Toolkit, I've got this error.

I'm not sure how to reproduce it in some playground, not sure what exactly causes it:

<details>

```
Exception thrown while executing UI block: 'parentNode' is a required parameter

__44-[RCTUIManager flushUIBlocksWithCompletion:]_block_invoke
    RCTUIManager.m:1192
-[REANodesManager performOperations]
-[REANodesManager onAnimationFrame:]
CA::Display::DisplayLink::dispatch_items(unsigned long long, unsigned long long, unsigned long long)
display_timer_callback(__CFMachPort*, void*, long, void*)
__CFMachPortPerform
__CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE1_PERFORM_FUNCTION__
__CFRunLoopDoSource1
__CFRunLoopRun
CFRunLoopRunSpecific
GSEventRunModal
UIApplicationMain
main
start
0x0
```

</details>

### Screenshots

![image](https://user-images.githubusercontent.com/7809008/95223233-057d8100-0802-11eb-9aa6-23c1b5cdd3c8.png)

## Steps To Reproduce

1. Clone this repo https://github.com/terrysahaidak/react-native-gallery-toolkit
2. Switch to tr-lightbox branch and run the Example project
3. Navigate to Lightbox => Shared Element Transition
4. Open/close image a few times in a row.

The error might occur after a few reloading or immediately. 

### Expected behavior

Should not throw the error.

### Actual behavior

Throws an error but still works if I just close it.

## Snack or minimal code example

1. Clone this repo https://github.com/terrysahaidak/react-native-gallery-toolkit
2. Switch to tr-lightbox branch and run the Example project

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0-alpha.7

drewandre:
Just chiming in here to say that I also started seeing this on 2.0.0-alpha.7 when attempting to run a portal-based image transition across the screen and fast-refreshing several times in a row. I'll try to find a reliable way to reproduce, but for me it seemed to only happen when the transitions were happening in rapid succession.

Orgil:
Also having exact same issue

JulianoLoren:
facing the same issue

JB-CHAUVIN:
[...]

PR #1501, opened 2020-12-03
Don't run observer callbacks during sync uimanager batch update

kmagiera:
## Description

This PR fixes the way we perform sync updates on iOS by avoiding UIManager observers from firing. This in turn fixes the problem of ui manager commands that may be executed out of order as reported in #1498. The reason behind that issue was that due to the fact we'd trigger ui manager observers, those may enqueue some commands in the ui-manager block and we'd run them along the sync ui updates that we were expecting to get after layout phase. In the sync layout call we only expect the ui commands to come from the layout calculations and not from other sources (e.g. native animated or reanimated which uses ui manager observers to enqueue additional commands).

## Changes

1) This PR wraps uimanager batchDidComplete call with additional code that temporarily removes registered ui observers and only puts a single observer instead that is then used to capture mounting block.
2) Some additional cleanup in NodesManager on iOS to extract code that deals with RCTUIManager implementation details to a category on RCTUIManager.
3) As a part of this we also make an update to Android where we trigger semaphore release earlier when sync layout is not possible to match the implementation on iOS.
4) Opt-out of synchronous updates when no reanimated 2 code is being used by adding a flag `trySynchronousBatch` which is only set for reanimated 2 code paths. Then use this flag in the update procedure and skip semaphor wait on iOS and Android when it isn't set

### Before

Under certain condition (difficult to reproduce on a small app) reanimated 1 commands would sometimes run out of order (e.g. first "connect node" and only then "create node")

### After

We no longer allow for reanimated or native animated modules to enqueue commands that are goign to be executed in the sync layout phase.

Apparently, we don't have a good repro steps for this, will test this internally on an app where we could reproduce the issue, and also reach out for help with testing to people who reported this in #1318

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -202,6 +202,9 @@ public class NodesManager implements EventDispatcherListener {
                 @Override
                 public void runGuarded() {
                   boolean shouldDispatchUpdates = UIManagerReanimatedHelper.isOperationQueueEmpty(mUIImplementation);
+                  if (!shouldDispatchUpdates) {
+                    semaphore.release();
+                  }
                   while (!copiedOperationsQueue.isEmpty()) {
                     NativeUpdateOperation op = copiedOperationsQueue.remove();
                     ReactShadowNode shadowNode = mUIImplementation.resolveShadowNode(op.mViewTag);
@@ -211,8 +214,8 @@ public class NodesManager implements EventDispatcherListener {
                   }
                   if (shouldDispatchUpdates) {
                     mUIImplementation.dispatchViewUpdates(-1); // no associated batchId
+                    semaphore.release();
                   }
-                  semaphore.release();
                 }
               });
       while (true) {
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -136,6 +136,7 @@ public NativeUpdateOperation(int viewTag, WritableMap nativeProps) {
     }
   }
   private Queue<NativeUpdateOperation> mOperationsInBatch = new LinkedList<>();
+  private boolean mTryRunBatchUpdatesSynchronously = false;
```
Szymon20000, line 139: Why do we want to opt-out of the synchronous layout when someone uses reanimated 1? Is it possible that we will set this variable to true and some of the REA 1 operations are in batch?

Szymon20000 (approved):
Looks fine to me! I only have no idea what is the purpose of `mTryRunBatchUpdatesSynchronously `.

android/src/main/java/com/swmansion/reanimated/NodesManager.java
kmagiera: this is to keep the behavior of apps that use reanimated 1 unchanged. When no reanimated 2 code is used it will always stay false and will fallback to the old update mechanism. It isn't possible for batch to be split with this as we always process the whole batch at once, just sometimes we will do it synchronously and other times we won't

PR #1502, opened 2020-12-03
Prevent executing mappers for the same value

### LINKED ISSUE #1499: Huge performance issues when using multiple worklets (`useAnimatedStyle`, `useDerivedValue`)
## Description

I've created a view which is essentially a horizontal ScrollView that can only be swiped one item at a time. Each item (= category) has a header showing the category's name. To somehow reveal to the user what the next page/previous page has to offer, I decided to show a tiny bit of the headers flowing into the screen.

<table>
<tr>
<th>Without header parallax</th>
<th>With header parallax</th>
</tr>
<tr>
<td width="50%"></td>
<td width="50%"></td>
</tr>
</table>

### Demo

<table>
<tr>
<th>Without header parallax</th>
<th>With header parallax</th>
</tr>
<tr>
<td width="50%"></td>
<td width="50%"></td>
</tr>
</table>

While you can't _really_ see the difference in a compressed, 320 pixel wide, 25 FPS GIF, it is extremely noticeable in the app, since you can swipe/scroll normally without the parallax, and once you enable the header parallax it looks like a powerpoint presentation. I'd say without parallax it's around 60 FPS, with parallax about 5 FPS.

### Attempts to solve

I've tried multiple approaches to solve those performance issues, here's what I tried:

1. Checked my re-renders. Maybe something is unnecessarily re-rendering? Nope, everything's happening on the UI thread.
2. Tried to optimize the worklets inside the Header (put all `useDerivedValues` into a single `useAnimatedStyle`) so that there's a few less worklets executing at once - didn't make a difference at all.
3. I've tried (and I'm still using) the `removeClippedSubviews` property on the container to remove all views that are offscreen, so not rendering them if they're overflowing. That requires `overflow: 'hidden'` to be set, which I did, so now only everything that's on screen should be rendered. This doesn't affect worklets though, since still every single `useDerivedValue` and `useAnimatedStyle` worklet is executing per frame on drag. For 30 Stories, that's 1 `useAnimatedStyle` and 1 `useAnimatedGestureHandler` for the "list" container, and 30 `useDerivedValue`s and 30 `useAnimatedStyle`s for the Headers. Can we ignore them if RCTView removes that view due to clipping?
4. I tried to unmount everything if it's offscreen. So only mounting current header, header to the right and header to the left. Weirdly enough, this still had terrible performance. It looked a bit better, but still absolutely unusable, like below 20 FPS.

## Steps To Reproduce

1. Create horizontal Swiper using `<PanGestureHandler>` and translate those views using the `translateX` (plus some `offsetX`)
2. Now pass each item the `translateX` value and it's index in the Swiper
3. Each item can now have a custom `useAnimatedStyle` which interpolates the current Swiper's `translateX` to parallax-like translate the header towards the screen corners so they get revealed.

### Expected behavior

I expect it to run smoothly

### Actual behavior

It runs terribly stuttery.

<details>
<summary><code>Header.tsx</code></summary>

```tsx
function StoryHeader({
  story,
  style: _style,
  indexInSwiper,
  swiperTranslateX,
  isSwipeMode,
  onPress,
  ...passThroughProps
}: StoryHeaderProps): React.ReactElement {
  const [width, setWidth] = useState(0);

  const onLayout = useCallback(
    ({
      nativeEvent: {
        layout: { width: newWidth },
      },
    }: LayoutChangeEvent) => {
      setWidth(newWidth);
    },
    [setWidth],
  );

  const thisTranslateX = useMemo(() => -(SCREEN_WIDTH * indexInSwiper), [indexInSwiper]);
  const offsetX = useDerivedValue(() => {
    if (isSwipeMode) {
      return withTiming(width / 2, {
        duration: 200,
        easing: Easing.back(1),
      });
    } else {
      return withTiming(width / 2 - HEADER_PREVIEW_OVERFLOW, {
        duration: 300,
        easing: Easing.elastic(1),
      });
    }
  }, [isSwipeMode, width]);
    const translateX = interpolate(
      swiperTranslateX.value,
      [
        thisTranslateX - SCREEN_WIDTH - SCREEN_WIDTH,
        thisTranslateX - SCREEN_WIDTH,
        thisTranslateX, //
        thisTranslateX + SCREEN_WIDTH,
        thisTranslateX + SCREEN_WIDTH + SCREEN_WIDTH,
      ],
      [
        SCREEN_WIDTH / 2,
        SCREEN_WIDTH / 2 - offsetX.value, // Swiper is at Next slide
        0, // Swiper is at current slide
        -(SCREEN_WIDTH / 2) + offsetX.value, // Swiper is at Previous slide
        -(SCREEN_WIDTH / 2),
      ],
      Extrapolate.CLAMP,
    );
    const isCurrentHeader = between(translateX, -10, 10);
    return {
      // opacity: withTiming(isSwipeMode ? (isCurrentHeader ? 1 : 0) : 1, { duration: 300 }),
      opacity: 1,
      transform: [{ translateX: translateX }, { scale: withTiming(isSwipeMode ? (isCurrentHeader ? 1.2 : 1) : 1, { duration: 150 }) }],
    };
  }, [isSwipeMode, offsetX, swiperTranslateX, thisTranslateX]);
  const style = useMemo(() => [_style, { left: SCREEN_WIDTH * indexInSwiper + (SCREEN_WIDTH / 2 - width / 2) }], [_style, indexInSwiper, width]);

  return (
    // TODO: DEBUG IF renderToHardwareTextureAndroid={true} shouldRasterizeIOS={true} ARE A GOOD IDEA
    <Reanimated.View
      onLayout={onLayout}
      style={[style, animatedStyle]}
      renderToHardwareTextureAndroid={true}
      shouldRasterizeIOS={true}
      {...passThroughProps}>
      <StoryHeaderContent story={story} onPress={onPress} />
    </Reanimated.View>
  );
}
```
</details>

## Package versions

- React: 17
- React Native: 0.64.rc1 (but also reproduced on 0.63)
- React Native Reanimated: master (but also reproduced on alpha 9 or 2.0-rc0)
- Engine: Hermes (but also reproduced on JSC)
- NodeJS: v14.15.1

I'm really not sure how I can make this more efficient or if I'm doing something wrong, so any tips would be greatly appreciated!

## EDIT

Here's everything I've noted:

1. In the header's `useAnimatedStyle` hook, that gets executed a lot (30 times for 30 different headers per on drag event), using `withTiming` (or `withSpring`) is a bad idea. Instead, extracting that into a `useSharedValue` and _imperatively_ animating that when the `isSwipeMode` or `isCurrentHeader` props changes works a lot better. I am guessing it just takes a long time to set up the `withTiming` (or `withSpring`) animation, even though the inputs haven't changed an it already is at the desired position.
2. Splitting all my dependencies up into `useDerivedValue`s doesn't seem to really do anything, it feels a tiny bit smoother I guess
3. Animations are run even if views are removed by RCTView's performance strategy: `removeClippedSubviews`. (overflow hidden)
4. I've unmounted everything except 3 headers (current, left and right), and it still had bad performance.  Even removed all other stuff, e.g. the background you're seeing in the demo. Now I know that this is a very complex animation, but with REA v1 I just passed different interpolated nodes into this and everything worked fine and smoothly, I assume "everything working as good as REA v1 or even better" is also the goal here.
5. If you want to see another demo of this, and even have a REA v1 vs REA v2 comparison, take a look at William's Jelly Scroll video (https://www.youtube.com/watch?v=Xnj6uoW2PJM). I've tried to implement this in REA v2, and it had terrible performance because of the many `useAnimatedStyle` hooks.

karol-bisztyga:
## Description

For now, if we attempt to set a shared value's value to the same value multiple times it'll go through the entire process of updating the value and also will trigger all the mappers that treat this shared value as an input.

Short pseudo code:

```
const sv = useSharedValue(50)
const uas = useAnimatedStyle(() => {
	return { width: sv.value }
})
for (let i = 0; i < 1000; ++i) {
	sv.value = 100;
}
```

The worklet in `useAnimatedStyle` is going to be triggered 1000 extra times even though the input(and the output) is going to be the same.

Closes #1499 

## Changes

Added checks for the same values in `workletValueSetter` and only conditionally update the shared value's value.

<details>
<summary>code</summary>

```
import React from 'react';
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  useDerivedValue,
  withTiming,

const useAnimation = true;

function App() {

  const udv = useDerivedValue(()=>{
    console.log('worklet run: execute UDV', sv.value);
    return sv.value;
  });

  const uas = useAnimatedStyle(() => {
    console.log('worklet run: execute UAS', udv.value);
    return {
      width: udv.value
    }
  });

  return (
    <View style={{ flex: 1, margin: 50 }}>
      <Button
        title="change SV"
        onPress={() => {
          const target = sv.value === 100 ? 200 : 100
          sv.value = useAnimation ? withTiming(target) : target;
        }}
      />
      <Button
        title="set SV to 100"
        onPress={() => {
          sv.value = useAnimation ? withTiming(100) : 100;
        }}
      />
      <Button
        title="set SV to 200"
        onPress={() => {
          sv.value = useAnimation ? withTiming(200) : 200;
        }}
      />
      <Animated.View
        style={[
          {
            width: 100,
            height: 100,
            backgroundColor: 'purple',
          },
          uas,
        ]}
      />
    </View>
  );
}

export default App;

```

</details>

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/reanimated2/core.js
+++ b/src/reanimated2/core.js
@@ -111,8 +111,18 @@ function workletValueSetter(value) {
     typeof value === 'function' ||
     (value !== null && typeof value === 'object' && value.onFrame)
   ) {
-    // animated set
     const animation = typeof value === 'function' ? value() : value;
+    // prevent setting again to the same value
+    // and triggering the mappers that treat this value as an input
+    // note that this will work for most animations
+    // this assumes that toValue === current DOES MEAN that the animation's done
+    if (
+      this._value === animation.toValue &&
+      animation.toValue === animation.current
+    ) {
+      return;
+    }
+    // animated set
     const initializeAnimation = (timestamp) => {
       animation.onStart(animation, this.value, timestamp, previousAnimation);
     };
@@ -141,6 +151,11 @@ function workletValueSetter(value) {
       requestAnimationFrame(step);
     }
   } else {
+    // prevent setting again to the same value
+    // and triggering the mappers that treat this value as an input
+    if (this._value === value) {
+      return;
+    }
     this._value = value;
   }
 }
--- a/src/reanimated2/core.js
+++ b/src/reanimated2/core.js
@@ -114,7 +114,7 @@ function workletValueSetter(value) {
     const animation = typeof value === 'function' ? value() : value;
     // prevent setting again to the same value
     // and triggering the mappers that treat this value as an input
-    // this happens when the animation's target value is set to the same value as a current one
+    // this happens when the animation's target value(stored in animation.current until animation.onStart is called) is set to the same value as a current one(this._value)
     if (this._value === animation.current) {
       return;
     }
--- a/src/reanimated2/core.js
+++ b/src/reanimated2/core.js
@@ -114,12 +114,8 @@ function workletValueSetter(value) {
     const animation = typeof value === 'function' ? value() : value;
     // prevent setting again to the same value
     // and triggering the mappers that treat this value as an input
-    // note that this will work for most animations
-    // this assumes that toValue === current DOES MEAN that the animation's done
-    if (
-      this._value === animation.toValue &&
-      animation.toValue === animation.current
-    ) {
+    // this happens when the animation's target value is set to the same value as a current one
+    if (this._value === animation.current) {
       return;
     }
     // animated set
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -295,8 +295,8 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
              }
            } catch(std::exception &e) {
              std::string str = e.what();
-             module->errorHandler->setError(str);
-             module->errorHandler->raise();
+             this->module->errorHandler->setError(str);
+             this->module->errorHandler->raise();
            }
 
            rt.global().setProperty(rt, "jsThis", oldJSThis); //clean jsThis
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -295,8 +295,8 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
              }
            } catch(std::exception &e) {
              std::string str = e.what();
-             this->module->errorHandler->setError(str);
-             this->module->errorHandler->raise();
+             module->errorHandler->setError(str);
+             module->errorHandler->raise();
            }
 
            rt.global().setProperty(rt, "jsThis", oldJSThis); //clean jsThis


zrebcu411 (approved):

src/reanimated2/core.js
```diff
@@ -111,8 +111,18 @@ function workletValueSetter(value) {
...
+    // prevent setting again to the same value
+    // and triggering the mappers that treat this value as an input
+    // note that this will work for most animations
+    // this assumes that toValue === current DOES MEAN that the animation's done
+    if (
+      this._value === animation.toValue &&
+      animation.toValue === animation.current
+    ) {
+      return;
+    }
```
piaskowyk, line 124: What is the reason why you use here && instead of || ?

karol-bisztyga: After some checking, I decided to change that condition as I recognized 2 scenarios: either `toValue` is equal to `current`(simple animations) or `toValue` is undefined and `current` points at the target value(group animations). So I think we can skip checking `toValue` and just use `current` here(as `current` points to `toValue` until `animation.onStart` is called).

karol-bisztyga: fixed 503f327, thx for pointing out

piaskowyk (approved):

nandorojo:
@karol-bisztyga this looks like a great addition. I have a question related to it.

I'm working on a `framer-motion`-like library based on Reanimated 2.

The API is like this:

```jsx
<View initial={{ opacity: 0 }} animate={{ opacity: 1 }} />
```

Real code [here](https://github.com/nandorojo/redrip/blob/master/src/redripify/use-map-animate-to-style.ts).

<details>
<summary>Pseudo-code here</summary>

```js
const useStyle = ({ animate, initial }) => {
  const hasMounted = useSharedValue(false)

  const initialSharedValue = useSharedValue(initial, true)
  const animateSharedValue = useSharedValue(animate, true)

    const finalStyle = {}

    Object.keys(animateSharedValue.value).forEach((key) => {
      if (!hasMounted.value && initial[key] != null) {
        finalStyle[key] = initialSharedValue.value[key]
      } else {
        finalStyle[key] = withSpring(animateSharedValue.value[key])
      }
    })

    return finalStyle
  })

  useEffect(() => {
    hasMounted.value = true
  }, [hasMounted])

  return style
}

// which then gets used like this:
const style = useStyle({ initial, animate })
return <Animated.View style={style} />
```
</details>

The example above will cause `useAnimatedStyle` to re-run on every render.

Should I change it to something like this?

```js
const useStyle = ({ animate, initial }) => {
  const hasMounted = useSharedValue(false)

  const initialDerivedValue = useDerivedValue(() => JSON.stringify(initial))
  const animateDerivedValue = useDerivedValue(() => JSON.stringify(animate))

    // should I access those values like this here?
    const animateStyle = JSON.parse(animateDerivedValue.value)
    const initialStyles = JSON.parsed(initialDerivedValue.value)
 
    // now use these to generate the style object...
  })
 
  return style
}
```

By stringifying, it passes shallow equality checks. I thought I'd get your thoughts on best practice, since Reanimated v2 is still a bit new to me.

Thanks so much!

karol-bisztyga:
@nandorojo hi 👋 I don't see a reason to wrap the received prop values into shared values. You can use state and prop values inside of the worklets(functions passed to reanimated 2 hooks). You may want to take a look [here](https://docs.swmansion.com/react-native-reanimated/docs/api/useAnimatedStyle#dependencies-array) and read about the dependencies array(how the hooks (in that case `useAnimatedStyle`) behave when the rerender occurs).

nandorojo:
Got it! @terrysahaidak  mentioned it might be convenient to work with SVs directly for performance, but it's also possible I misunderstood (I'm still getting the hang of v2.)

PR #1509, opened 2020-12-07
Get rid of optional chaining in plugin

karol-bisztyga:
## Description

It turned out that using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) doesn't work for older node versions(confirmed for 12.x.x) but does work for newer ones(14.14.0). As we want to make it compatible with those versions we had to get rid of that feature and replace it with an equivalent solution.

## Changes

Instead of using conditional chaining just watch out for a `TypeError` when reading from the object.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -352,16 +352,11 @@ function processWorkletFunction(t, fun, fileName) {
   const funString = buildWorkletString(t, fun, variables, functionName);
   const workletHash = hash(funString);
 
-  try {
-    // try to append location in file
-    const loc = fun.node.loc.start;
-    if (loc) {
-      const { line, column } = loc;
+  const loc = fun && fun.node && fun.node.loc && fun.node.loc.start;
+  if (loc) {
+    const { line, column } = loc;
+    if (typeof line === 'number' && typeof column === 'number') {
       fileName = `${fileName} (${line}:${column})`;
-    }
-  } catch (e) {
-    if (!(e instanceof TypeError)) {
-      throw e;
     }
   }
 
--- a/plugin.js
+++ b/plugin.js
@@ -352,10 +352,17 @@ function processWorkletFunction(t, fun, fileName) {
   const funString = buildWorkletString(t, fun, variables, functionName);
   const workletHash = hash(funString);
 
-  const loc = fun?.node?.loc?.start;
-  if (loc) {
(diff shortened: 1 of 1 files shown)


plugin.js
```diff
@@ -352,10 +352,17 @@ function processWorkletFunction(t, fun, fileName) {
...
+    // try to append location in file
+    const loc = fun.node.loc.start;
+    if (loc) {
+      const { line, column } = loc;
+      fileName = `${fileName} (${line}:${column})`;
+    }
+  } catch (e) {
+    if (!(e instanceof TypeError)) {
+      throw e;
+    }
```
jakub-gonet, line 365: I'd rather wrap in try only access to this object:
```
let loc;
try {
	const loc = fun.node.loc.start;
} catch (e) {
    if (!(e instanceof TypeError)) {
      throw e;
}
```

or use old-fashioned method:
```
const loc = fun && fun.node && fun.node.loc && fun.node.loc.start
```
which will allow removing try-catch entirely.

karol-bisztyga: I was considering all the options you mentioned. Picked the wider try catch because it just seemed to be the clearest one to me. Doesn't really matter to me, I can change it or leave it as it is.

karol-bisztyga: ok I updated the code

jakub-gonet (approved):
LGTM.

what other than the number can be `line` or `column`? An object?

zrebcu411 (approved):

piaskowyk (approved):

karol-bisztyga:
> LGTM.
> 
> what other than the number can be `line` or `column`? An object?

I prefer to use such checks for security's sake. This code handles a lot of nodes and such checks avoid unexpected behaviors for some edge cases that could occur(if the check for type `number` fails, it will just add a file name there).

jakub-gonet:
> I prefer to use such checks for security's sake.

I'd rather throw with the context if we don't anticipate things different than numbers there, it's way easier to track down and fix any issues that may show up with that approach.

karol-bisztyga:
I personally would not throw there. This is extra info only for some cases and only for debugging. The fact that this information is provided or not does not have any impact on the library's functional correctness.

PR #1515, opened 2020-12-08
fix mem leak

Szymon20000:
## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -47,7 +47,15 @@ export function useEvent(handler, eventNames = [], rebuild = false) {
   } else if (rebuild) {
     initRef.current.updateWorklet(handler);
   }
-  return initRef.current;
+
+  useEffect(() => {
+    return () => {
+      initRef.current.unregisterFromEvents();
+      initRef.current = null;
+    }
+  }, []);
+
+  return initRef;
 }
 
 function prepareAnimation(animatedProp, lastAnimation, lastValue) {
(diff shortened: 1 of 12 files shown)


src/reanimated2/Hooks.js
```diff
@@ -47,7 +47,15 @@ export function useEvent(handler, eventNames = [], rebuild = false) {
   } else if (rebuild) {
     initRef.current.updateWorklet(handler);
   }
-  return initRef.current;
+
+  useEffect(() => {
+    return () => {
+      initRef.current.unregisterFromEvents();
```
karol-bisztyga, line 53: what's the point of using ref when it's being trashed on unmount? I thought the refs are used because we want the objects to persist between mounts

src/reanimated2/Hooks.js
```diff
@@ -423,6 +431,13 @@ export function useAnimatedGestureHandler(handlers, dependencies) {
       savedDependencies: [],
     };
   }
+
+  useEffect(()=>{
+    return () => {
+      initRef.current = null;
```
karol-bisztyga, line 437: same ads above, couldn't we just quit using refs in such a case?

Szymon20000: No, we couldn't. It will be executed only when the view unmounts.

karol-bisztyga:
as #1543 has been merged, I'm closing this one

PR #1518, opened 2020-12-09
Fix useAnimatedGestureHandler for web

### LINKED ISSUE #1406: web: useAnimatedGestureHandler not working
## Description
That is working on android, but on the web the gesture event is not called

## Steps To Reproduce

1.
```js
      {
        onStart: (_, ctx: any) => {
          //notCalled
        },
        onActive: (event, ctx) => {
          //notCalled
        },
        onEnd: (_) => {
          //notCalled
        },
      },
      [],
    );
```
## Package versions

- React: 17.0.1
- React Native: 0.63.1
- React Native Reanimated: 2.0.0-alpha.8
- React Native Gesture Handler: 1.8.0
- NodeJS: 12

karol-bisztyga:
## Description

For some time we were unable to use `react-native-web` above version `0.12.3`. Along with gesture handler `1.9.0`, this has been fixed as stated [here](https://github.com/software-mansion/react-native-gesture-handler/issues/1164#issuecomment-734246498).

It's required to pass a function on the web.

Fixes #1406 

## Changes

In Hooks.js instead of using `useEvent` on the web we just return the handling function(also extract the `nativeEvent` in that case).
For example project, the package versions have been updated.

Just run the example screens(like DragAndDrop or ChatHeads)

## Checklist

- [x] Ensured that CI passes


src/reanimated2/Hooks.js
```diff
@@ -433,51 +433,55 @@ export function useAnimatedGestureHandler(handlers, dependencies) {
   );
   initRef.current.savedDependencies = dependencies;
 
+  const handler = (event) => {
+    'worklet';
+    event = Platform.OS === 'web' ? event.nativeEvent : event;
```
jakub-gonet, line 438: I'm pretty sure we should use `Platform.OS !== 'native'` here. This way we can potentially support RN for web, ios, and windows.

src/reanimated2/Hooks.js
```diff
@@ -433,51 +433,55 @@ export function useAnimatedGestureHandler(handlers, dependencies) {
...
+    ) {
+      handlers.onFinish(
+        event,
+        context,
+        event.state === CANCELLED || event.state === FAILED
+      );
+    }
+  };
+
+  if (Platform.OS === 'web') {
```
jakub-gonet, line 479: Ditto.

src/reanimated2/Hooks.js
karol-bisztyga: by iOS here you mean macOS?

karol-bisztyga: what's the difference between checking for `native` and `web`? Aren't they just opposite(I mean `Platform.OS === web` == `Platform.OS !== native`)?

jakub-gonet: > by iOS here you mean macOS?

Yeah, I meant macos.

> what's the difference between checking for native and web? Aren't they just opposite(I mean Platform.OS === web == Platform.OS !== native)?

I don't think that `Platform.OS === web` evaluates to true in windows or macOS version of RN, that's why I think we should go with "native or everything else" here.

karol-bisztyga: AFAIK we support 3 platforms for now: web, ios, and android. There haven't been any test for any other platforms that I am aware of. In this PR I attached specific behavior for the web. If you think we should handle windows and macOS then I think you should open a new issue for that as we check for OS versions in the same(or similar) way as here in multiple places in the code.

jakub-gonet: Resolved internally, we came to the conclusion that we should focus on the web support for now, so although we should use the "not native" check here, this solution is not scalable and we'll have to address this problem in the future.

We can focus on implementation for other platforms if the community expresses interest.

karol-bisztyga: +1, also `'native'` won't work for such checks, closing both discussions.

piaskowyk (approved):
Good stuff 💪

zrebcu411 (approved):

PR #1519, opened 2020-12-09
Update README.md

danieloi:
The first step was present in docs here: https://docs.swmansion.com/react-native-reanimated/docs/1.x.x/examples but absent in README. 
Spent hours figuring out what was wrong

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/Example/README.md
+++ b/Example/README.md
@@ -1,8 +1,9 @@
 ## Running the example
 
 - clone the repository
+- `yarn install` at main directory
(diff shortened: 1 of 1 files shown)


jakub-gonet (approved):
Though it was clear enough as list of steps but I agree this is clearer.

PR #1521, opened 2020-12-09
[v1] Replace setImmediate with requestAnimationFrame

### LINKED ISSUE #1520: AnimatedNode should use requestAnimationFrame instead of setImmediate for web
Using setImmediate cause animations/mutations to be done with significant delay. This can be observed especially when UI is under heavy load.

I'm not sure whether native platforms share same problem, didn't observe even on low-end Android TV boxes. React Native docs suggests requestAnimationFrame should be used though. I'll leave it open to discussions.
https://reactnative.dev/docs/timers 

I will create PR soon, let's discuss there.

DrRefactor:
## Description

requestAnimationFrame is called before frame is flushed, as soon as possible.
https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame

Fixes #1520 

## Changes

For example:

- Changed setImmediate to requestAnimationFrame as node update scheduler for web

## Screenshots / GIFs

Change can be seen clearly when using imperative Value.setValue(number) along with calling React.setState. React.setState is handled much faster in this scenario.

Background color in below example is updated via standard React state cycle, and position is set via ReanimatedNode.setValue. Both are fired in same synchronous scope (see below repo).

### Before
setImmediate

### After
requestAnimationFrame

https://github.com/DrRefactor/reanimated-issue-1520-repro

I've included instructions for running versions "before" and "after". I'm not sure this is the easiest way to do so though.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/core/AnimatedNode.js
+++ b/src/core/AnimatedNode.js
@@ -76,6 +76,9 @@ function runPropUpdates() {
   loopID += 1;
 }
 
+const scheduleUpdates =
+  Platform.OS === 'web' ? requestAnimationFrame : setImmediate;
+
 export default class AnimatedNode {
   __nodeID;
   __lastLoopID = { '': -1 };
@@ -141,7 +144,7 @@ export default class AnimatedNode {
   __markUpdated() {
     UPDATED_NODES.push(this);
     if (!propUpdatesEnqueued) {
-      propUpdatesEnqueued = setImmediate(runPropUpdates);
+      propUpdatesEnqueued = scheduleUpdates(runPropUpdates);
     }
   }
 


nandorojo:
This would be a great fix for web, thanks!

jakub-gonet (approved):
LGTM.

There's a small note in the [`requestAnimationFrame` docs](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) that can be problematic for the users:

> Be sure to always use the first argument (or some other method for getting the current time) to calculate how much the animation will progress in a frame, otherwise the animation will run faster on high refresh rate screens. Check the example below for a way to do this.

But using `requestAnimationFrame` instead of non-polyfilled `setImmediate` is better in every aspect so I think we can merge this one and fix this problem later.

karol-bisztyga (approved):

cltsang:
To whom it may concern I have to use dynamic import to get it built on Next js
i.e.

``` js
// do this
import dynamic from 'next/dynamic';
const ScrollBottomSheet = dynamic(
  () => import('react-native-scroll-bottom-sheet'),
  { ssr: false }
);
```

``` js
// instead of this
import ScrollBottomSheet from 'react-native-scroll-bottom-sheet';
```

ref: [https://stackoverflow.com/a/64261030/2437201](https://stackoverflow.com/a/64261030/2437201)

nandorojo:
Hm, I didn't realize there was no polyfill to get `requestAnimationFrame` to work with SSR. @jakub-gonet is there any way we could add a check to see if the `window` exists to support SSR, and if not, use a no-op function?

nandorojo:
@cltsang could this polyfill solve it?

https://reactjs.org/blog/2017/09/26/react-v16.0.html#javascript-environment-requirements

cltsang:
@nandorojo Yes it solves it. Thank you!
Didn't notice this react requirement before.

nandorojo:
Cool, where did you place the polyfill? `pages/_app.tsx`?

cltsang:
Yes I did.
`import 'raf/polyfill';`
at the first line.

nandorojo:
Got it. We should probably add this as a note to the web docs for Next.js/Gatsby/SSR support.

nandorojo:
Which version of reanimated is this shipped in? I'm using Expo SDK 40 + `2.0.0-rc.0`, but it looks like it still uses set immediate. As a result, my Next.js app still sees this:

Screen Shot 2021-02-05 at 3 53 43 PM

jakub-gonet:
You can see the tag associated with [the commit](https://github.com/software-mansion/react-native-reanimated/commit/39905f28ba701599b772a492edcd3c6e7a3bf34a) at the bottom of the description: this is included in rc.2. Bear in mind when you're using expo you can only upgrade libraries to version with non-native changes – Reanimated had those in rc.2 and rc.3.

PR #1536, opened 2020-12-15
Export defineAnimation()

wcandillon:
For animations defined outside this file like in redash.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


mrousavy:
If that's going to be exported we should probably also create types for that

Szymon20000 (approved):

PR #1539, opened 2020-12-16
TS: createAnimatedComponent to use ComponentClass

### LINKED ISSUE #1537: type for createAnimatedComponent param is vague (any)
## Description

When I call `Reanaimated.createAnimatedComponent()` with a functional component I get this error at runtime.

```
`createAnimatedComponent` does not support stateless functional components; use a class component instead.
```
Here's an example snack: https://snack.expo.io/@goleary/reanimated.createanimatedcompoent-type-failure
## Steps To Reproduce

call `Reanimated.createAnimatedComponen()` & pass it a functional component.

### Expected behavior

It would be awesome if the type of the param that this function accepts would help the user understand this code will fail at runtime because whatever is being passed won't work (is not a class component).

### Actual behavior
createAnimatedComponent accepts anything passed to it and fails at runtime.

https://snack.expo.io/@goleary/reanimated.createanimatedcompoent-type-failure

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 1.13.0

wcandillon:
## Description
fixes #1537

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -2,7 +2,7 @@
 // TypeScript Version: 2.8
 
 declare module 'react-native-reanimated' {
-  import { ComponentClass, ReactNode, Component, RefObject, ComponentType } from 'react';
+  import { ComponentClass, ReactNode, Component, RefObject, ComponentClass } from 'react';
   import {
     ViewProps,
     TextProps,
@@ -242,7 +242,7 @@ declare module 'react-native-reanimated' {
       getNode(): ReactNativeScrollView;
     }
     export class Code extends Component<CodeProps> {}
-    export function createAnimatedComponent<S extends object, P extends { style?: StyleProp<S>; }>(component: ComponentType<P>): ComponentType<AnimateProps<S, P>>;
+    export function createAnimatedComponent<S extends object, P extends { style?: StyleProp<S>; }>(component: ComponentClass<P>): ComponentClass<AnimateProps<S, P>>;
 
     // classes
     export {
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -2,7 +2,7 @@
 // TypeScript Version: 2.8
 
 declare module 'react-native-reanimated' {
-  import { ComponentClass, ReactNode, Component, RefObject, ComponentClass } from 'react';
+  import { ComponentClass, ReactNode, Component, RefObject } from 'react';
   import {
     ViewProps,
     TextProps,


karol-bisztyga:
will it fix #1537? If so could you add it to the description? Thx

wcandillon:
done

jakub-gonet (changes requested):
LGTM, please add some tests with `@ts-ignore-error` passing functional component to it.

react-native-reanimated.d.ts
```diff
@@ -2,7 +2,7 @@
 // TypeScript Version: 2.8
 
 declare module 'react-native-reanimated' {
-  import { ComponentClass, ReactNode, Component, RefObject, ComponentType } from 'react';
+  import { ComponentClass, ReactNode, Component, RefObject, ComponentClass } from 'react';
```
karol-bisztyga, line 5:
```suggestion
  import { ComponentClass, ReactNode, Component, RefObject } from 'react';
```

karol-bisztyga: `ComponentClass` is already there, `ComponentType ` should just be removed

karol-bisztyga (changes requested):

karol-bisztyga:
It does not allow `Animated.createAnimatedComponent(Button);`(an error is not thrown on the runtime)

karol-bisztyga:
I'm not really into TS but maybe something like this would do?
`export function createAnimatedComponent<S extends object, P extends object>(component: ComponentClass<P>): ComponentClass<AnimateProps<S, P>>;`

maxammann:
Should we maybe support react-native-svgs `style`-less components easier/better? https://github.com/software-mansion/react-native-reanimated/pull/1466#issuecomment-748863116

react-native-reanimated.d.ts
```diff
@@ -242,7 +242,7 @@ declare module 'react-native-reanimated' {
       getNode(): ReactNativeScrollView;
     }
     export class Code extends Component<CodeProps> {}
-    export function createAnimatedComponent<S extends object, P extends { style?: StyleProp<S>; }>(component: ComponentType<P>): ComponentType<AnimateProps<S, P>>;
+    export function createAnimatedComponent<S extends object, P extends { style?: StyleProp<S>; }>(component: ComponentClass<P>): ComponentClass<AnimateProps<S, P>>;
```
maxammann, line 245:
```suggestion
    export function createAnimatedComponent<S extends object, P extends { style?: StyleProp<S>; } | {}>(component: ComponentClass<P>): ComponentClass<AnimateProps<S, P>>;
```

That way we can use react-native-svgs `PathProps` without style attribute.

jakub-gonet:
@wcandillon 🏓

wcandillon:
Should be green now 💚

react-native-reanimated.d.ts
karol-bisztyga, line 245:
```suggestion
    export function createAnimatedComponent<S extends object, P extends object>(component: ComponentClass<P>): ComponentClass<AnimateProps<S, P>>;
```

karol-bisztyga: `Animated.createAnimatedComponent(Button);` still does not work. Although I haven't found any use case for this(by that I mean updating animated props of a button on UI) it just indicates that there might be situations like this failing for the valid use cases. This does not work because `ButtonProps` does not contain `style?: StyleProp<object>`. It also won't work for any valid class components which do not contain such prop(`style?: ...`).

Example:

```
interface Props {
  name: string;
  // style?: StyleProp<object>; // will work only with this
}

class MyComponent extends React.Component<Props> {}

Animated.createAnimatedComponent(MyComponent);
```

In order to handle custom components, I'd suggest to change it like I proposed or some other way that would make it work.

karol-bisztyga (changes requested):

wcandillon:
Closing in favor of #1590

Issue #1538, opened 2020-12-16
withRepeat in the type definition only has 3 arguments.  Should have 4 with the callback


0b3r:
## Description

`withRepeat` in the type definition only has 3 arguments.  Should have 4 with the callback

https://github.com/software-mansion/react-native-reanimated/blob/9a70b7afa498cffa48d124e9276ec6939dffb1f6/react-native-reanimated.d.ts#L425

### Expected behavior

Typescript should allow 4 arguments while calling `withRepeat`

### Actual behavior

Typescript error when providing 4 arguments

### steps to reproduce

Try passing 4 arguments to withRepeat in Typescript

## Package versions

2.0.0-alpha.9.2

PR #1540, opened 2020-12-16
Add callback to withRepeat type

karol-bisztyga:
## Description

Add `callback` to `withRepeat` type

Fixes https://github.com/software-mansion/react-native-reanimated/issues/1538


jakub-gonet (commented):
LGTM, could you add a callback to the `withRepeat` test?

jakub-gonet (approved):

PR #1543, opened 2020-12-17
Fix mem leaks

karol-bisztyga:
## Description

Created with @Szymon20000 
Code was taken from #1515 

The main goal is to improve the cleaning up process.

## Changes

* stop mapper - the mapper in fact was being deleted but it persisted as a leftover in the `sortedMappers`. Requesting render results in recalculating(reconstructing) the sorted mappers thus it doesn't happen anymore
* pointer to mapper function - I think it's safer and more consistent with the rest of the code to store mapper's function as a `shared_ptr`. If there's some strong argumentation for storing it as a rvalue reference, we can revert this
* `createAnimatedComponent` - since we decided to return ref object from `useEvent`, we now have to access it via `.current`
* `Hooks.js`: 
  * manually cleaning refs
  * removed redundant `useMapper`

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -104,9 +104,9 @@ export default function createAnimatedComponent(Component) {
       const viewTag = findNodeHandle(node);
 
       for (const key in this.props) {
-        const prop = this.props[key];
+        let prop = this.props[key];
         if (prop.current && prop.current instanceof WorkletEventHandler) {
-          prop = prop.current
+          prop = prop.current;
         }
         if (prop instanceof AnimatedEvent) {
           prop.attachEvent(node, key);
@@ -122,7 +122,7 @@ export default function createAnimatedComponent(Component) {
       for (const key in this.props) {
         let prop = this.props[key];
         if (prop.current && prop.current instanceof WorkletEventHandler) {
-          prop = prop.current
+          prop = prop.current;
         }
         if (prop instanceof AnimatedEvent) {
           prop.detachEvent(node, key);
@@ -139,7 +139,10 @@ export default function createAnimatedComponent(Component) {
       let viewTag;
 
       for (const key in this.props) {
-        const prop = this.props[key];
+        let prop = this.props[key];
+        if (prop.current && prop.current instanceof WorkletEventHandler) {
+          prop = prop.current;
+        }
         if (prop instanceof AnimatedEvent) {
           nextEvts.add(prop.__nodeID);
         } else if (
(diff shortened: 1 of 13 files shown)


src/reanimated2/Hooks.js
```diff
@@ -311,6 +310,12 @@ export function useAnimatedStyle(updater, dependencies) {
     };
   }, dependencies);
 
+  useEffect(() => {
+    return () => {
+      initRef.current = undefined;
```
piaskowyk, line 315: In `useAnimationStyle()` is used `maybeViewRef`. Although it is used only for web but maybe we should also clear this `ref` in `useEffect()`

karol-bisztyga: fixed 635aaec

Szymon20000 (approved):

piaskowyk (approved):

src/createAnimatedComponent.js
```diff
@@ -104,7 +104,10 @@ export default function createAnimatedComponent(Component) {
       const viewTag = findNodeHandle(node);
 
       for (const key in this.props) {
-        const prop = this.props[key];
+        let prop = this.props[key];
+        if (prop?.current && prop.current instanceof WorkletEventHandler) {
```
kmagiera, line 108: We already have an instanceof check for this type few lines below. What's the point of adding another one if we could just reuse the one that's thre and avoid making two identical checks one after another. Also there is no need to check whether prop.current is null before instanceof check as null will never match the type provided.

```
if (prop instanceof AnimatedEvent) {
  prop.attachEvent(node, key);
} else if (prop?.current instanceof WorkletEventHandler) {
  prop.current.registerForEvents(viewTag, key);
}
```

src/reanimated2/Hooks.js
```diff
@@ -311,6 +310,13 @@ export function useAnimatedStyle(updater, dependencies) {
     };
   }, dependencies);
 
+  useEffect(() => {
+    return () => {
+      initRef.current = null;
```
kmagiera, line 315: This needs some explanation. I don't see why it'd actually make any difference. When hook is unmounted the `initRef` is gone, so it will not retain reference via `current` field either.

Even if we need this behavior it'd be a lot better to have it designed in a more resilient way and have custom hook that both creates a ref and also cleans it up instead of repeating the same code over and over in five places.

src/reanimated2/Hooks.js
```diff
@@ -31,23 +31,22 @@ export function useSharedValue(init, shouldRebuild = true) {
...
   if (initRef.current === null) {
     initRef.current = new WorkletEventHandler(handler, eventNames);
   } else if (rebuild) {
     initRef.current.updateWorklet(handler);
   }
-  return initRef.current;
+
+  useEffect(() => {
+    return () => {
+      initRef.current.unregisterFromEvents();
```
kmagiera, line 44: `unregisterFromEvents` is already called in `createAimatedComponent` – you should've explained why is it needed to be added here. It seem more appropriate to control registration/unreg in one place instead of having that logic and responsibility split across different components. Currently `createAnimatedComponnt` is responsible for registering for the events and so it should also be responsible for the cleanup.

src/createAnimatedComponent.js
karol-bisztyga: this is fixed in #1556

PR #1545, opened 2020-12-18
Bump version of navigation in Example app

piaskowyk:
## Description

Bump version of navigation dependency. It's related with issue in navigation [#8951](https://github.com/react-navigation/react-navigation/issues/8951)


jakub-gonet (approved):

karol-bisztyga (commented):
this works alright, although the problem with the transparent navigation persists on android

PR #1549, opened 2020-12-19
Feature: interpolate: Extrapolate left/right

### LINKED ISSUE #1255: interpolate: Extrapolate left/right
## Description

I'd like to see extrapolating for both sides, just like in v1.

## Motivation

Currently we can only extrapolate left and right at the same time, but if I wanted to clamp left and extend identity for the right value, that wouldn't be possible. I'd have to use a workaround: Use no extrapolation and then use `Math.min` or `Math.max` for "clamping".

## Code example

A solution for this would be:

```ts
const x = interpolate(scrollX, [0, HEADER_HEIGHT, contentSize.x], [0, 0, contentSize.x + HEADER_HEIGHT], { extrapolateLeft: 'clamp', extrapolateRight: 'identity' })
```

To make it backwards compatible (and also cleaner if you don't need different extrapolations) the `interpolate` function could take both types:

```ts
type ExtrapolateParameter = Extrapolate | { extrapolateLeft?: Extrapolate, extrapolateRight?: Extrapolate } | undefined 
```

Roka20012:
## Wix Goodness Squad Event

Motivation: to see extrapolating for both sides, just like in v1.

Fixes #1255

## Changes

- Added ts types for extrapolation
- Added ExtrapolationExample
- Added Error handling for invalid config or invalid interpolation value 

## Live example

IDK which example is better currently 2 in this PR

https://user-images.githubusercontent.com/32239452/102691648-a03d0780-4216-11eb-90e7-7ef8a5f1590a.mov

---
https://user-images.githubusercontent.com/32239452/102694608-df755380-422a-11eb-839c-8872d1427807.mov

## Checklist

- [x] Updated TS types

### DIFF
--- a/Example/src/App.js
+++ b/Example/src/App.js
@@ -31,10 +31,6 @@ import LiquidSwipe from './LiquidSwipe';
 YellowBox.ignoreWarnings(['Calling `getNode()`']);
 
 const SCREENS = {
-  ExtrapolationExample: {
-    screen: ExtrapolationExample,
-    title: '🆕 Extrapolation Example',
-  },
   AnimatedStyleUpdate: {
     screen: AnimatedStyleUpdateExample,
     title: '🆕 Animated Style Update',
@@ -82,6 +78,10 @@ const SCREENS = {
   LiquidSwipe: {
     screen: LiquidSwipe,
     title: '🆕 Liquid Swipe Example',
+  },
+  ExtrapolationExample: {
+    screen: ExtrapolationExample,
+    title: '🆕 Extrapolation Example',
   },
 };
 
--- a/Example/src/App.js
+++ b/Example/src/App.js
@@ -14,6 +14,7 @@ import { createStackNavigator } from 'react-navigation-stack';
 
 import Reanimated1 from '../reanimated1/App';
 
+import ExtrapolationExample from './ExtrapolationExample';
 import AnimatedStyleUpdateExample from './AnimatedStyleUpdateExample';
 import WobbleExample from './WobbleExample';
 import DragAndSnapExample from './DragAndSnapExample';
@@ -30,6 +31,10 @@ import LiquidSwipe from './LiquidSwipe';
 YellowBox.ignoreWarnings(['Calling `getNode()`']);
 
 const SCREENS = {
+  ExtrapolationExample: {
+    screen: ExtrapolationExample,
+    title: '🆕 Extrapolation Example',
+  },
   AnimatedStyleUpdate: {
     screen: AnimatedStyleUpdateExample,
     title: '🆕 Animated Style Update',
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -2,7 +2,13 @@
 // TypeScript Version: 2.8
 
 declare module 'react-native-reanimated' {
-  import { ComponentClass, ReactNode, Component, RefObject, ComponentType } from 'react';
+  import {
+    ComponentClass,
+    ReactNode,
+    Component,
+    RefObject,
+    ComponentType,
+  } from 'react';
   import {
     ViewProps,
     TextProps,
@@ -52,10 +58,15 @@ declare module 'react-native-reanimated' {
       IDENTITY = 'identity',
     }
 
+    type ExtrapolateParameter =
+      | Extrapolate
+      | { extrapolateLeft?: Extrapolate; extrapolateRight?: Extrapolate }
+      | undefined;
+
     interface InterpolationConfig {
       inputRange: ReadonlyArray<Adaptable<number>>;
       outputRange: ReadonlyArray<Adaptable<number | string>>;
-      extrapolate?: Extrapolate;
+      extrapolate?: ExtrapolateParameter;
       extrapolateLeft?: Extrapolate;
       extrapolateRight?: Extrapolate;
     }
@@ -242,7 +253,10 @@ declare module 'react-native-reanimated' {
       getNode(): ReactNativeScrollView;
     }
     export class Code extends Component<CodeProps> {}
-    export function createAnimatedComponent<S extends object, P extends { style?: StyleProp<S>; }>(component: ComponentType<P>): ComponentType<AnimateProps<S, P>>;
+    export function createAnimatedComponent<
+      S extends object,
+      P extends { style?: StyleProp<S> }
+    >(component: ComponentType<P>): ComponentType<AnimateProps<S, P>>;
 
     // classes
     export {
@@ -421,7 +435,10 @@ declare module 'react-native-reanimated' {
     export function cancelAnimation<T extends SharedValue<SharedValueType>>(
       sharedValue: T
     ): void;
-    export function withDelay(delayMS: number, delayedAnimation: number): number;
+    export function withDelay(
+      delayMS: number,
+      delayedAnimation: number
+    ): number;
     export function withRepeat(
       animation: number,
       numberOfReps?: number,
@@ -482,8 +499,10 @@ declare module 'react-native-reanimated' {
       effects: (dependencies: D) => void,
       deps?: DependencyList
     );
-                        
-    export type AnimatedStyleProp<T extends object> = AnimateStyle<T> | RegisteredStyle<AnimateStyle<T>>;
+
+    export type AnimatedStyleProp<T extends object> =
+      | AnimateStyle<T>
+      | RegisteredStyle<AnimateStyle<T>>;
     export function useAnimatedStyle<
       T extends AnimatedStyleProp<ViewStyle | ImageStyle | TextStyle>
     >(updater: () => T, deps?: DependencyList): T;
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -2,13 +2,7 @@
 // TypeScript Version: 2.8
 
 declare module 'react-native-reanimated' {
-  import {
-    ComponentClass,
-    ReactNode,
-    Component,
-    RefObject,
-    ComponentType,
-  } from 'react';
+  import { ComponentClass, ReactNode, Component, RefObject, ComponentType } from 'react';
   import {
     ViewProps,
     TextProps,
@@ -58,10 +52,7 @@ declare module 'react-native-reanimated' {
       IDENTITY = 'identity',
     }
 
-    type ExtrapolateParameter =
-      | Extrapolate
-      | { extrapolateLeft?: Extrapolate; extrapolateRight?: Extrapolate }
-      | undefined;
+    type ExtrapolateParameter = Extrapolate | { extrapolateLeft?: Extrapolate, extrapolateRight?: Extrapolate } | undefined 
 
     interface InterpolationConfig {
       inputRange: ReadonlyArray<Adaptable<number>>;
@@ -253,10 +244,7 @@ declare module 'react-native-reanimated' {
       getNode(): ReactNativeScrollView;
     }
     export class Code extends Component<CodeProps> {}
-    export function createAnimatedComponent<
-      S extends object,
-      P extends { style?: StyleProp<S> }
-    >(component: ComponentType<P>): ComponentType<AnimateProps<S, P>>;
+    export function createAnimatedComponent<S extends object, P extends { style?: StyleProp<S>; }>(component: ComponentType<P>): ComponentType<AnimateProps<S, P>>;
 
     // classes
     export {
@@ -435,10 +423,7 @@ declare module 'react-native-reanimated' {
     export function cancelAnimation<T extends SharedValue<SharedValueType>>(
       sharedValue: T
     ): void;
-    export function withDelay(
-      delayMS: number,
-      delayedAnimation: number
-    ): number;
+    export function withDelay(delayMS: number, delayedAnimation: number): number;
     export function withRepeat(
       animation: number,
       numberOfReps?: number,
@@ -499,10 +484,8 @@ declare module 'react-native-reanimated' {
       effects: (dependencies: D) => void,
       deps?: DependencyList
     );
-
-    export type AnimatedStyleProp<T extends object> =
-      | AnimateStyle<T>
-      | RegisteredStyle<AnimateStyle<T>>;
+                        
+    export type AnimatedStyleProp<T extends object> = AnimateStyle<T> | RegisteredStyle<AnimateStyle<T>>;
     export function useAnimatedStyle<
       T extends AnimatedStyleProp<ViewStyle | ImageStyle | TextStyle>
     >(updater: () => T, deps?: DependencyList): T;
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -52,7 +52,7 @@ declare module 'react-native-reanimated' {
       IDENTITY = 'identity',
     }
 
-    type ExtrapolateParameter = Extrapolate | { extrapolateLeft?: Extrapolate, extrapolateRight?: Extrapolate } | undefined 
+    type ExtrapolateParameter = Extrapolate | { extrapolateLeft?: Extrapolate, extrapolateRight?: Extrapolate }
 
     interface InterpolationConfig {
       inputRange: ReadonlyArray<Adaptable<number>>;
--- a/src/reanimated2/interpolation.js
+++ b/src/reanimated2/interpolation.js
@@ -52,7 +52,7 @@ function validateType(type) {
   );
 
   if (
-    type?.constructor === Object &&
+    typeof type === 'object' &&
     ((Object.keys(type).length === 2 &&
       !(hasExtrapolateLeft && hasExtrapolateRight)) ||
       (Object.keys(type).length === 1 &&
--- a/src/reanimated2/interpolation.js
+++ b/src/reanimated2/interpolation.js
@@ -32,7 +32,7 @@ function isExtrapolate(value) {
 function validateType(type) {
   'worklet';
 
-  if (!type) {
+  if (!type && type !== undefined) {
     throw new Error(
       `Reanimated: config object is not valid please provide valid config, for example:
        interpolate(value, [inputRange], [outputRange], {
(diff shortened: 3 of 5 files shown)


Roka20012:
@terrysahaidak sup dude!
I think this PR need to review

Thanks again for being a maintainer on the Wix Goodness Squad Event

Cheers ✌️

terrysahaidak:
Hey, @karol-bisztyga this is yet another PR as a result of our event, could you please assign someone for review. Thanks.

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
+}
+
+function isExtrapolate(value) {
+  'worklet';
+
+  return (
+    value !== Extrapolate.EXTEND &&
+    value !== Extrapolate.CLAMP &&
+    value !== Extrapolate.IDENTITY
+  );
```
jakub-gonet, line 29: Could you please invert the logic here?

```
value === Extrapolate.EXTEND ||
value === Extrapolate.CLAMP ||
value === Extrapolate.IDENTITY
```

is way easier to read.

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
+        extrapolateLeft: 'clamp',
+        extrapolateRight: 'extend',
+      })`
+    );
+  }
+
+  if (
+    typeof type === 'object' &&
+    type.extrapolateLeft &&
+    isExtrapolate(type.extrapolateLeft)
```
jakub-gonet, line 55: Won't `isExtrapolate` return `false` for `null` / `undefined`?

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
   const coef = rr >= ll ? 1 : -1;
 
+  const config = { type, coef, val, ll, rr, x };
+
+  validateType(type);
+
   // TODO: support default values in worklets:
   // e.g. function interplate(x, input, output, type = Extrapolate.CLAMP)
-  type = type || Extrapolate.EXTEND;
+  // type = type || Extrapolate.EXTEND;
```
jakub-gonet, line 99: Why it's commented out? If we need default then it should be left and if we don't please just remove this value.

BTW, does it break backward compat?

jakub-gonet (commented):
Left some comments, but somebody from turboteam should review this.

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
+  ) {
+    throw new Error(
+      `Reanimated: not supported value for "extrapolateRight" \nSupported values: ["extend", "clamp", "identity"]\n Valid example:
+       interpolate(value, [inputRange], [outputRange], {
+        extrapolateRight: 'clamp',
+      })`
+    );
+  }
+
+  if (typeof type === 'string' && isExtrapolate(type)) {
```
karol-bisztyga, line 78: As for the typeof checks: checking for type `object` occurs a several times. I'd move those checks to the end of this function and then bring it before them so it would look like this:
```
f() {
	// ... some checks
	if (typeof type !== 'object') {
		return;
	}
	// ... all the checks that contained `typeof type !== 'object'` before
}
```

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
   const progress = (x - l) / (r - l);
   const val = ll + progress * (rr - ll);
   const coef = rr >= ll ? 1 : -1;
 
+  const config = { type, coef, val, ll, rr, x };
+
+  validateType(type);
+
   // TODO: support default values in worklets:
   // e.g. function interplate(x, input, output, type = Extrapolate.CLAMP)
```
karol-bisztyga, line 98:
```suggestion
  // e.g. function interpolate(x, input, output, type = Extrapolate.CLAMP)
```

karol-bisztyga: please fix this typo here 🙏

karol-bisztyga:
It's a solid PR, I'll be happy to merge it, just please make those couple of thing more clear and I think we're good to go 😉

src/reanimated2/interpolation.js
Roka20012: Done ✅

src/reanimated2/interpolation.js
Roka20012: Yeap it will

src/reanimated2/interpolation.js
Roka20012: > Why it's commented out? If we need default then it should be left and if we don't please just remove this value.

Removed

---

> BTW, does it break backward compat?

If you mean using it like this -> ```interpolate(translation.y.value, [0, -75], [0, -150], 'clamp')``` 

It will work like before 🙂

Example/src/App.js
```diff
@@ -30,6 +31,10 @@ import LiquidSwipe from './LiquidSwipe';
 YellowBox.ignoreWarnings(['Calling `getNode()`']);
 
 const SCREENS = {
+  ExtrapolationExample: {
```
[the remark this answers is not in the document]
Roka20012: Done ✅

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
+    value !== Extrapolate.IDENTITY
+  );
+}
+
+function validateType(type) {
+  'worklet';
+
+  if (
+    (typeof type === 'object' &&
+      (!type.extrapolateLeft &&
```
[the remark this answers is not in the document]
Roka20012: Done ✅

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
+        !type.extrapolateRight &&
+        Object.keys(type).length >= 1 &&
+        type.constructor === Object)) ||
+    (Object.keys(type).length > 2 && type.constructor === Object)
```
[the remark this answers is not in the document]
Roka20012: Done ✅

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
+    value !== Extrapolate.CLAMP &&
+    value !== Extrapolate.IDENTITY
+  );
+}
+
+function validateType(type) {
+  'worklet';
+
+  if (
+    (typeof type === 'object' &&
```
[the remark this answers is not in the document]
Roka20012: Done ✅

src/reanimated2/interpolation.js
Roka20012: Done ✅

src/reanimated2/interpolation.js
Roka20012: Done ✅

Roka20012:
@karol-bisztyga @jakub-gonet please re-check I have made some changes for this PR ✌️ ☃️

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
+function validateType(type) {
+  'worklet';
+
+  if (!type && type !== undefined) {
+    throw new Error(
+      `Reanimated: config object is not valid please provide valid config, for example:
+       interpolate(value, [inputRange], [outputRange], {
+        extrapolateLeft: 'clamp',
+        extrapolateRight: 'extend',
+      }) or interpolate(value, [inputRange], [outputRange], 'clamp)`
```
karol-bisztyga, line 41:
```suggestion
      }) or interpolate(value, [inputRange], [outputRange], 'clamp')`
```

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
+    );
+  }
+
+  if (
+    type?.constructor === Object &&
+    ((Object.keys(type).length === 2 &&
+      !('extrapolateLeft' in type && 'extrapolateRight' in type)) ||
+      (Object.keys(type).length === 1 &&
+        !('extrapolateLeft' in type || 'extrapolateRight' in type)) ||
+      Object.keys(type).length > 2)
```
karol-bisztyga, line 51: Maybe it could be done using less code?
How about:
```
type?.constructor === Object &&
    ((Object.keys(type).length >= 1 &&
      !('extrapolateLeft' in type && 'extrapolateRight' in type)) ||
    Object.keys(type).length > 2)
```
or
```
type?.constructor === Object &&
    (([1,2].includes(Object.keys(type).length) &&
      !('extrapolateLeft' in type && 'extrapolateRight' in type)) ||
    Object.keys(type).length > 2)
```

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,115 @@
...
-  // e.g. function interplate(x, input, output, type = Extrapolate.CLAMP)
-  type = type || Extrapolate.EXTEND;
+  const config = { type, coef, val, ll, rr, x };
+
+  validateType(type);
+
+  if (typeof type === 'object') {
+    return coef * val < coef * ll
+      ? getVal({ ...config, type: type.extrapolateLeft })
+      : getVal({ ...config, type: type.extrapolateRight });
```
karol-bisztyga, line 106: I don't think that ternary replaces the original conditions, can you revert this? Or am I missing something?

karol-bisztyga (changes requested):

src/reanimated2/interpolation.js
Roka20012: Done ✅

src/reanimated2/interpolation.js
Roka20012: Done ✅

src/reanimated2/interpolation.js
Roka20012: Yep, looks too much anyway I ended up with this

Your code will not work for this code

In case if one of `extrapolateLeft | extrapolateRight` field is missing (by default if one of them is missing it's `extrapolate === Extrapolate.EXTEND`)

```js
interpolate(translation.y.value, [0, -75], [0, -150], {
        extrapolateLeft: 'clamp',
})
```

karol-bisztyga: Yes, you are right, sorry, I missed one thing in those conditions, it's ok.

karol-bisztyga (approved):

karol-bisztyga:
For me, it looks OK. @jakub-gonet?

piaskowyk (approved):

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,117 @@
...
+    value === Extrapolate.EXTEND ||
+    value === Extrapolate.CLAMP ||
+    value === Extrapolate.IDENTITY
+  );
+}
+
+function validateType(type) {
+  'worklet';
+
+  if (!type && type !== undefined) {
```
jakub-gonet, line 35: Isn't that just `type === null`?

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,117 @@
...
+  if (
+    type?.constructor === Object &&
+    ((Object.keys(type).length === 2 &&
+      !('extrapolateLeft' in type && 'extrapolateRight' in type)) ||
+      (Object.keys(type).length === 1 &&
+        !('extrapolateLeft' in type || 'extrapolateRight' in type)) ||
+      Object.keys(type).length > 2)
+  ) {
+    throw new Error(
+      `Reanimated: config object is not valid please provide valid config, for example:
```
jakub-gonet, line 54:
```suggestion
      `Reanimated: config object is not valid, please provide valid config, for example:
```

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,117 @@
...
-        if (coef * val < coef * ll) {
-          return ll;
-        }
-        return rr;
-      case Extrapolate.EXTEND:
-      default:
-        return val;
+  validateType(type);
+
+  if (typeof type === 'object') {
```
jakub-gonet, line 103: Just making sure, it can be null here? (`typeof null === "object"`)

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,117 @@
...
+      `Reanimated: config object is not valid please provide valid config, for example:
+       interpolate(value, [inputRange], [outputRange], {
+        extrapolateLeft: 'clamp',
+        extrapolateRight: 'extend',
+      }) or interpolate(value, [inputRange], [outputRange], 'clamp')`
+    );
+  }
+
+  if (
+    type?.constructor === Object &&
```
jakub-gonet, line 46: 1. Safe navigation operator isn't necessary, we checked for null earlier.
    From [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor#description):
    > All objects (with the exception of objects created with Object.create(null)) will have a constructor property. 
2. Why we check for constructor property here?

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,117 @@
...
+    ((Object.keys(type).length === 2 &&
+      !('extrapolateLeft' in type && 'extrapolateRight' in type)) ||
```
jakub-gonet, line 48: Nit; `in` operator traverses prototype hierarchy, we should probably use [`type.hasOwnProperty('extrapolateLeft')`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty).

Just a nit, because we expect an object as a dictionary here.

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,117 @@
...
+          extrapolateLeft: 'clamp',
+        })`
+      );
+    }
+
+    if ('extrapolateRight' in type && !isExtrapolate(type.extrapolateRight)) {
+      throw new Error(
+        `Reanimated: not supported value for "extrapolateRight" \nSupported values: ["extend", "clamp", "identity"]\n Valid example:
+         interpolate(value, [inputRange], [outputRange], {
+          extrapolateRight: 'clamp',
```
jakub-gonet, line 76: I'd like to have this and other error messages as string constants:
```js
const EXTRAPOLATE_RIGHT_ERROR_MSG = `(...)`;
```

react-native-reanimated.d.ts
```diff
@@ -52,10 +52,12 @@ declare module 'react-native-reanimated' {
       IDENTITY = 'identity',
     }
 
+    type ExtrapolateParameter = Extrapolate | { extrapolateLeft?: Extrapolate, extrapolateRight?: Extrapolate } | undefined
```
jakub-gonet, line 55: You already handle undefined in `extrapolate?`

jakub-gonet (commented):
Left a few comments.

src/reanimated2/interpolation.js
Roka20012: It's for `false` and empty string as well or false value

src/reanimated2/interpolation.js
Roka20012: That's why I am doing `validateType` on `101` line

jakub-gonet: Makes sense, thanks for explaining.

Line 101 if line numeration changes:
`validateType(type);`

src/reanimated2/interpolation.js
jakub-gonet: Oh, right.

react-native-reanimated.d.ts
Roka20012: If you mean `interpolate(translation.y.value, [0, -75], [0, -150], undefined})`

It will work like `Extrapolate.EXTEND`

If you mean `interpolate(translation.y.value, [0, -75], [0, -150], {
        extrapolateLeft: undefined,
})`

It should throw an error

src/reanimated2/interpolation.js
Roka20012: Good catch, removed ✅

src/reanimated2/interpolation.js
Roka20012: Done ✅

src/reanimated2/interpolation.js
Roka20012: Done ✅

src/reanimated2/interpolation.js
Roka20012: Oh actually just a `!type` would be enough)

src/reanimated2/interpolation.js
```diff
@@ -1,30 +1,119 @@
...
+  }
+
+  const hasExtrapolateLeft = Object.prototype.hasOwnProperty.call(
+    type,
+    'extrapolateLeft'
+  );
+  const hasExtrapolateRight = Object.prototype.hasOwnProperty.call(
+    type,
+    'extrapolateRight'
+  );
```
jakub-gonet, line 61: Why not do it like here ⬇️?
```suggestion
  const hasExtrapolateLeft = type.hasOwnProperty('extrapolateLeft');
  
  const hasExtrapolateRight = type.hasOwnProperty('extrapolateRight');
```

jakub-gonet (approved):
LGTM, left one comment.

react-native-reanimated.d.ts
jakub-gonet: I meant that we have this interface:
```ts
    interface InterpolationConfig {
      inputRange: ReadonlyArray<Adaptable<number>>;
      outputRange: ReadonlyArray<Adaptable<number | string>>;
      extrapolate?: ExtrapolateParameter; // <- extrapolate is optional
      extrapolateLeft?: Extrapolate;
      extrapolateRight?: Extrapolate;
    }
```

and this type:
```ts
type ExtrapolateParameter = Extrapolate | { extrapolateLeft?: Extrapolate, extrapolateRight?: Extrapolate } | undefined 
```

Marking extrapolate as optional already adds `| undefined` to the value.

src/reanimated2/interpolation.js
Roka20012: Because of eslint error 😕

jakub-gonet: I assume this is caused by this rule:
https://eslint.org/docs/rules/no-prototype-builtins

We don't use `Object.create` nor we shadow prototype methods, so it was safe to ignore.

react-native-reanimated.d.ts
Roka20012: Nice catch, fixed ✅

Removed  `| undefined ` from `ExtrapolateParameter `

Roka20012:
@karol-bisztyga thanks for merging!❤️

PR #1556, opened 2020-12-22
Memory leaks

karol-bisztyga:
## Description

Here we will solve/discuss/work on issues related to memory leaks and also code changes in #1543.

Every case will be taken care of separately in terms of both the code and the description.

# Cases

### Memory leak with `makeShareable`, inner class, and a method(that has to be inside of the component) passed to its constructor.

- solution
	  - prevent applying cache for objects containing host functions(be1cb52)
	  - remove ref loop from ShareableValue(#1580)

<details>
<summary>graph</summary>

diagram

</details>

<details>
<summary>code</summary>

```
import {
  makeShareable,
import { Text } from 'react-native';
import React, { useEffect, useRef } from 'react';

export default function OneMakeShareable() {
  function sth() {}

  class Temp1 {
    constructor(fun) {
      this.fun = fun;
    }
  }

  const ctx = useRef(null);
  if (ctx.current === null) {
    const temp = new Temp1(sth);
    ctx.current = {
      t: makeShareable(temp),
    };
  }

  useEffect(() => {
    return () => {
      ctx.current = undefined;
    };
  }, []);

  return <Text>ONE</Text>;
}

```

</details>

### The same as in the first case but a worklet with a reference to the `temp` object is passed to `makeShareable` instead of a `temp` object directly.

<details>
<summary>graph</summary>

graph

</details>

<details>
<summary>code</summary>

```
const ThreeTest = () => {
  function sth() {}
  class Temp31 {
    constructor(fun) {
      this.fun = fun;
    }
  }

  const tempc = new Temp31(sth);

  const wrk = () => {
    'worklet';
    const t = tempc;
  };

  makeShareable(wrk);
  let x = makeShareable(wrk);
  x = null;
  return <Text>Three</Text>;
};
```

</details>

### Similar pattern as above but this time `useAnimatedStyle` is used instead of `makeShareable`

❗ this turned out not to be a problem. It just depends on the UI garbage collector. If it is triggered manually there are no dangling objects(it was observed that it cleans after some time).

<details>
<summary>graph</summary>

depgraph2

</details>

<details>
<summary>code</summary>

```
import { runOnJS, useAnimatedStyle } from 'react-native-reanimated';
import { Text } from 'react-native';
import React from 'react';

export default function OneMakeShareable() {
  function sth() {}

  class Temp2 {
    constructor(fun) {
      this.fun = fun;
    }
  }

  const temp = new Temp2(sth);

  useAnimatedStyle(() => {
    const x = temp;
    return {
      width: 100,
    };
  });

  return <Text>Two</Text>;
}

```

</details>

also tested with `useState`

<details>
<summary>code</summary>

```
export function FourState() {
 
 class Temp4 {
   constructor(f) {
     this.f = f;
   }
 }
 const temp = new Temp4(setState);
 
 useAnimatedStyle(() => {
    runOnJS(temp.f)(state + 1);
   return {};
 });
 
 
 return (
   <View>
     <Text>Four</Text>
   </View>
 );
}
```

</details>

# Other

- forwarded: https://github.com/software-mansion/react-native-reanimated/pull/1543/files#r546597171
	- This seems to be unnecessary as unregistering from events happen anyway on `componentWillUnmount` in `createAnimatedComponent`
- forwarded: https://github.com/software-mansion/react-native-reanimated/pull/1543/files#r546594788
	- This prevents the loop: c++ -> ref -> fiberNode(@Szymon20000) where in ref we hold for instance `setState`

### DIFF
--- a/Common/cpp/SharedItems/FrozenObject.cpp
+++ b/Common/cpp/SharedItems/FrozenObject.cpp
@@ -9,6 +9,9 @@ FrozenObject::FrozenObject(jsi::Runtime &rt, const jsi::Object &object, NativeRe
   for (size_t i = 0, count = propertyNames.size(rt); i < count; i++) {
     auto propertyName = propertyNames.getValueAtIndex(rt, i).asString(rt);
     map[propertyName.utf8(rt)] = ShareableValue::adapt(rt, object.getProperty(rt, propertyName), module);
+    if (map[propertyName.utf8(rt)]->type == ValueType::HostFunctionType) {
+      this->containsHostFunction = true;
+    }
   }
 }
 
--- a/Common/cpp/SharedItems/FrozenObject.cpp
+++ b/Common/cpp/SharedItems/FrozenObject.cpp
@@ -8,8 +8,9 @@ FrozenObject::FrozenObject(jsi::Runtime &rt, const jsi::Object &object, NativeRe
   auto propertyNames = object.getPropertyNames(rt);
   for (size_t i = 0, count = propertyNames.size(rt); i < count; i++) {
     auto propertyName = propertyNames.getValueAtIndex(rt, i).asString(rt);
-    map[propertyName.utf8(rt)] = ShareableValue::adapt(rt, object.getProperty(rt, propertyName), module);
-    if (map[propertyName.utf8(rt)]->type == ValueType::HostFunctionType) {
+    std::string nameStr = propertyName.utf8(rt);
+    map[nameStr] = ShareableValue::adapt(rt, object.getProperty(rt, propertyName), module);
+    if (map[nameStr]->type == ValueType::HostFunctionType) {
       this->containsHostFunction = true;
     }
   }
--- a/Common/cpp/SharedItems/FrozenObject.cpp
+++ b/Common/cpp/SharedItems/FrozenObject.cpp
@@ -10,9 +10,7 @@ FrozenObject::FrozenObject(jsi::Runtime &rt, const jsi::Object &object, NativeRe
     auto propertyName = propertyNames.getValueAtIndex(rt, i).asString(rt);
     std::string nameStr = propertyName.utf8(rt);
     map[nameStr] = ShareableValue::adapt(rt, object.getProperty(rt, propertyName), module);
-    if (map[nameStr]->type == ValueType::HostFunctionType) {
-      this->containsHostFunction = true;
-    }
+    this->containsHostFunction |= map[nameStr]->containsHostFunction;
   }
 }
 
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -33,7 +33,7 @@ void addHiddenProperty(jsi::Runtime &rt,
 void freeze(jsi::Runtime &rt, jsi::Object &obj) {
   jsi::Object globalObject = rt.global().getPropertyAsObject(rt, "Object");
   jsi::Function freeze = globalObject.getPropertyAsFunction(rt, "freeze");
-  // freeze.call(rt, obj);
+  freeze.call(rt, obj);
 }
 
 void ShareableValue::adaptCache(jsi::Runtime &rt, const jsi::Value &value) {
@@ -140,6 +140,18 @@ void ShareableValue::adapt(jsi::Runtime &rt, const jsi::Value &value, ValueType 
   }
 }
 
+void cleanupShareable(ShareableValue &sv) {
+  if (sv.hostFunction != nullptr) {
+    sv.hostFunction = nullptr;
+  }
+  if (sv.frozenObject != nullptr) {
+    for (auto it = sv.frozenObject->map.begin(); it != sv.frozenObject->map.end(); ++it) {
+      auto itt = *it;
+      cleanupShareable(*it->second);
+    }
+  }
+}
+
 std::shared_ptr<ShareableValue> ShareableValue::adapt(jsi::Runtime &rt, const jsi::Value &value, NativeReanimatedModule *module, ValueType valueType) {
   auto sv = std::shared_ptr<ShareableValue>(new ShareableValue(module, module->scheduler));
   sv->adapt(rt, value, valueType);
@@ -179,7 +191,7 @@ jsi::Value createFrozenWrapper(jsi::Runtime &rt, std::shared_ptr<FrozenObject> f
     addHiddenProperty(rt, std::move(__reanimatedHiddenHost), obj, HIDDEN_HOST_OBJECT_PROP);
     addHiddenProperty(rt, true, obj, ALREADY_CONVERTED);
   }
-  return obj; //freeze.call(rt, obj);
+  return freeze.call(rt, obj);
 }
 
 jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
@@ -219,7 +231,7 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
         // call on an appropriate thread
         
         auto module = this->module;
-        auto hostFunction = this->hostFunction;
+        std::weak_ptr<HostFunctionHandler> hostFunction = this->hostFunction;
         
         auto warnFunction = [module, hostFunction](
             jsi::Runtime &rt,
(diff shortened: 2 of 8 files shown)


Common/cpp/SharedItems/FrozenObject.cpp
```diff
@@ -8,7 +8,11 @@ FrozenObject::FrozenObject(jsi::Runtime &rt, const jsi::Object &object, NativeRe
   auto propertyNames = object.getPropertyNames(rt);
   for (size_t i = 0, count = propertyNames.size(rt); i < count; i++) {
     auto propertyName = propertyNames.getValueAtIndex(rt, i).asString(rt);
-    map[propertyName.utf8(rt)] = ShareableValue::adapt(rt, object.getProperty(rt, propertyName), module);
+    std::string nameStr = propertyName.utf8(rt);
+    map[nameStr] = ShareableValue::adapt(rt, object.getProperty(rt, propertyName), module);
+    if (map[nameStr]->type == ValueType::HostFunctionType) {
```
Szymon20000, line 13: It's not enough. You have to check if a property contains HostFunction. Imagine the following case: object->object->hostFunction.

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -136,6 +140,17 @@ void ShareableValue::adapt(jsi::Runtime &rt, const jsi::Value &value, ValueType
   }
 }
 
+void cleanupShareable(ShareableValue &sv) {
```
Szymon20000, line 143: The function doesn't make sense at all. If an object is the last reference to the function then the function will garbage collected even without the function. If it's not the case then you will remove the function even if it's needed somewhere else.

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -213,22 +230,24 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
         // call on an appropriate thread
         
         auto module = this->module;
-        auto hostFunction = this->hostFunction;
+        std::weak_ptr<HostFunctionHandler> hostFunction = this->hostFunction;
```
Szymon20000, line 233: It cannot be a weak_ptr because the function will be instantly deallocated. Imagine the following example: 
```JS
runOnJS(worklet)(callback_hostfunction); // 
```

Common/cpp/SharedItems/FrozenObject.cpp
Szymon20000: It also has to be done for worklets and frozenArrays.

Common/cpp/SharedItems/ShareableValue.cpp
karol-bisztyga: using worklet in ruOnJS results in an error

Common/cpp/SharedItems/FrozenObject.cpp
karol-bisztyga: > Imagine the following case: object->object->hostFunction.
If you mean that a host function is in the nested object then that nested object would be checked here. Is there any place other than [this](https://github.com/software-mansion/react-native-reanimated/blob/master/Common/cpp/SharedItems/ShareableValue.cpp#L96) where `this->hostFunction` is being assigned?
> It also has to be done for worklets and frozenArrays.
Why? What case would it cover?(not mentioning that frozen arrays aren't really frozen)

Common/cpp/SharedItems/ShareableValue.cpp
karol-bisztyga: It's not that it doesn't make sense at all, I was although sceptic about it. Well, I checked and it does introduce some problems so it's not the best solution but it solves a problem which would persist without it.

Common/cpp/SharedItems/ShareableValue.cpp
karol-bisztyga: You probably meant something like this:
```
const wrk = (clb) => {
  'worklet';
  runOnJS(clb)();
};

const hf = () => {}

runOnUI(wrk)(hf);
```

Common/cpp/SharedItems/ShareableValue.cpp
karol-bisztyga: It in fact introduces errors but covers a leak that will persist without it

Common/cpp/SharedItems/ShareableValue.cpp
Szymon20000: I meant runOnUI.

Common/cpp/SharedItems/ShareableValue.cpp
Szymon20000: It still doesn't make sense.

Common/cpp/SharedItems/FrozenObject.cpp
Szymon20000: Like I said it's not enough because the function can have a reference to the other object.

Common/cpp/SharedItems/ShareableValue.cpp
Szymon20000: Which cases do you think it covers?

karol-bisztyga: discussed internally

Common/cpp/SharedItems/ShareableValue.cpp
karol-bisztyga: discussed internally

Common/cpp/SharedItems/FrozenObject.cpp
karol-bisztyga: added checks 668fe6e

Szymon20000 (approved):

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -94,19 +94,25 @@ void ShareableValue::adapt(jsi::Runtime &rt, const jsi::Value &value, ValueType
...
         hostFunction = std::make_shared<HostFunctionHandler>(std::make_shared<jsi::Function>(object.asFunction(rt)), rt);
+        containsHostFunction = true;
       } else {
         // a worklet
         type = ValueType::WorkletFunctionType;
         frozenObject = std::make_shared<FrozenObject>(rt, object, module);
+        containsHostFunction |= frozenObject->containsHostFunction;
         if (isRNRuntime) {
-          addHiddenProperty(rt, createHost(rt, frozenObject), object, HIDDEN_HOST_OBJECT_PROP);
+          if (!containsHostFunction) {
```
piaskowyk, line 104: Can we change this to `if (isRNRuntime && !containsHostFunction)` ? - micro optimization because there can exist only one condition jump, however compiler should change this to one jump also.

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -123,8 +129,11 @@ void ShareableValue::adapt(jsi::Runtime &rt, const jsi::Value &value, ValueType
       // create frozen object based on a copy of a given object
       type = ValueType::ObjectType;
       frozenObject = std::make_shared<FrozenObject>(rt, object, module);
+      containsHostFunction |= frozenObject->containsHostFunction;
```
piaskowyk, line 132: Should we add this check also for `MutableValue`? [here](https://github.com/software-mansion/react-native-reanimated/blob/c10217c5d6024f47af7911ec82c7d3c76fbd7fd0/Common/cpp/SharedItems/ShareableValue.cpp#L75)

For example for this situation: `const fn = makeMutable(() => {temp});`
Perhaps this is unusable edge case but I just ask

piaskowyk (approved):

Common/cpp/SharedItems/ShareableValue.cpp
karol-bisztyga: fixed 93ce23b

Common/cpp/SharedItems/ShareableValue.cpp
karol-bisztyga: I've thought about that but it seems like we do not apply that optimization in such a case, do we?

piaskowyk: You are right, we didn't.

karol-bisztyga: Thanks for the review. If you find anything like this bothering you you can always open a new PR/discussion, I'm merging this for now.

PR #1558, opened 2020-12-22
Feat: platform specific code separation

ArMANIAK:
During Wix Goodness Squad hackathon I've extracted platform specific code into separated files. Check whether such decision is acceptable


karol-bisztyga:
Hey! Once again thanks for your work, it means a lot 🙌 
However, I think you kind of missed the point, which was to get rid of all those conditional statements like `if (Platform.OS === 'web')` from our code. The solution you provided does not do that.

I see you organized files in subfolders(web, android). That actually does not help. I would rather create separate files for those platforms with the proper extensions.

I think you may want to get familiar with [this](https://reactnative.dev/docs/platform-specific-code).

### possible solution

Let's say you have something like this in your code:
```
if (OS.Platform === 'web') theFunc(args);
```

In order to implement my idea you would, for instance, go for something like this:
```
import { theFunc } from './platform-specific/theFuncImpl' // here you could go for one file which would host multiple functions(a matter of design) and it would be named differently I suppose
theFunc(args)
```
and then you could create such files:
- theFuncImpl.native.js for the native
```
export function theFunc(args) {/* nothing */}
```
- theFuncImpl.js for the web
```
export function theFunc(args) {/* code for the web */}
```

jakub-gonet:
@ArMANIAK 🏓

ArMANIAK:
> import { theFunc } from './platform-specific/theFuncImpl' // here you could go for one file which would host multiple functions(a matter of design) and it would be named differently I suppose
> theFunc(args)
> ```
> 
> and then you could create such files in `./platform-specific` directory:
> 
> * `theFuncImpl.native.js` for the native
> 
> ```
> export function theFunc(args) {/* nothing */}
> ```
> 
> * `theFuncImpl.js` for the web
> 
> ```
> export function theFunc(args) {/* code for the web */}
> ```

Sorry, guys, but I still can't get how to implement it in other way. If theFunc should be executed only for specific platform there should be if-clause. I can put it in the very file for platform specific functions, but it won't be got rid. At least my level doesn't allow me to find the way.

PR #1561, opened 2020-12-23
Multithreading with worklets

karol-bisztyga:
## Description

This idea has been postponed due to low priority.
The solution here is dirty, it's more like a POC for the future if we decide to focus on this topic again.
The goal is to be able to run worklets on multiple threads different than JS and UI.

### Use cases

- downloading data from the Internet
- processing video/audio/images
- obtaining a contact list from the device
- in fact, processing any data in the background in order to avoid blocking UI or JS.

### Examples

<details>

```
import {
  View,
  Button,
  Text,
} from 'react-native';
import {
  useSharedValue,
  spawnThread,


const Child = () => {
  const sv = useSharedValue(0)

  spawnThread(() => {
    'worklet';
    sv.value = 1.1;
    console.log('here thread #1 begin', sv.value, _WORKLET);
    for (let i = 0; i < 3000000000; ++i) {}
    sv.value = 1.2;
    console.log('here thread #1 end', sv.value, _WORKLET);
  });

  spawnThread(() => {
    'worklet';
    sv.value = 2.1;
    console.log('here thread #2 begin', sv.value, _WORKLET);
    for (let i = 0; i < 3000000000; ++i) {}
    sv.value = 2.2;
    console.log('here thread #2 end', sv.value, _WORKLET);
  });

  return (
    <View>
      <Text>Child</Text>
      <Button title="check responsiveness" onPress={() => {
        console.log('UI responsive', sv.value, Math.random())
      }} />
    </View>
  );
}

  const [s, setS] = useState(0);

  return (<View>
    <Button title="change state" onPress={()=>{ setS(s+1) }} />

    { s%2 === 1 ? <Child /> : <Text>EMPTY</Text>}
  </View>)
}
```

</details>

### DIFF
--- a/Common/cpp/NativeModules/NativeReanimatedModule.cpp
+++ b/Common/cpp/NativeModules/NativeReanimatedModule.cpp
@@ -79,7 +79,7 @@ NativeReanimatedModule::NativeReanimatedModule(std::shared_ptr<CallInvoker> jsIn
     maybeRequestRender();
   };
 
-  RuntimeDecorator::addNativeObjects(*runtime,
+  RuntimeDecorator::decorateUI(*runtime,
                                      platformDepMethodsHolder.updaterFunction,
                                      requestAnimationFrame,
                                      platformDepMethodsHolder.scrollToFunction,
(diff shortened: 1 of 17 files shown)


mrousavy:
Hi @karol-bisztyga! Thanks for taking time to look into this! I'm currently writing a Camera TurboModule, and I'm planning on creating an API where the user can process frames using JS.

So for example:

```jsx
const processFrame = useWorkletCallback((frame: CameraFrame) => {
  // process frame data in this worklet, for example to detect faces, render a square ontop of a QR code, or whatever.
}, []);


return <Camera {...props} processFrame={processFrame} />;
```

The benefits of this are hopefully immediately clear:

1. The user has full control over the video frames of the Camera and doesn't have to go native for stuff like object detection.
2. The processFrame function does not block the UI or JS thread.
3. There is no bridge-spam.

I'm currently halfway through implementing the Camera base stuff (format selection, capturing videos, capturing photo, ...), but I'm already excited to have something working in the near future.

Just leaving this here to make it clear that multithreaded worklets do indeed have many use cases, and perhaps we can work together to extend the proof of concept to a fully functional use case with the Camera process frame API 👍

mrousavy:
To those interested: I've published the react-native-vision-camera library (🎉) and am now working on the Frame Processor thing. All progress is public at: https://github.com/cuvent/react-native-vision-camera/pull/2

mrousavy:
@karol-bisztyga could you rebase with master?

likern:
It's exactly what I'm looking for and dreamed of. It allows not touching native as much as possible.

I need this for video recorder app.
And it would be cool to use another thread to work with sqlite database, not polluting main and UI thread (where reanimated animations might run).

While I can't help, eagerly waiting for this feature.

PR #1567, opened 2020-12-25
Add DerivedValue to the TS types

wcandillon:


jakub-gonet (approved):
LGTM.

mym0404:
Is this really needed?
In some use case, we may want to declare a value that is changed when the shared value changed and mutable itself.

wcandillon:
I would say so but please share your use-case with us.

PR #1575, opened 2020-12-28
Mention props config in the useAnimatedProps

### LINKED ISSUE #545: addWhitelistedNativeProps  addWhitelistedUIProps are missing from documentation
addWhitelistedNativeProps is required if attempting to natively animate SVG values as discovered in this issue https://github.com/software-mansion/react-native-reanimated/issues/537, I do not believe that many people know about this and its likely due to the missing documentation.

jakub-gonet:
## Description

Fixes #545.

### DIFF
--- a/docs/docs/api/useAnimatedProps.md
+++ b/docs/docs/api/useAnimatedProps.md
@@ -11,6 +11,11 @@ The provided worklet is responsible for returning a object with view properties.
 
 Only "native" properties of "native views" can be set via `useAnimatedProps`.
 The most common usecase for this hook is when we want to animate properties of some third-party native component, since most of the properties for the core React Native components are a part of the styles anyways (at least the properties for which it makes sense to be animated).
+You can use following functions to animate properties that Reanimated don't support by default:
+
+1. `addWhitelistedNativeProps()` is used to animate properties that triggers layout recalculation, you can find them [here](https://github.com/software-mansion/react-native-reanimated/blob/master/src/ConfigHelper.js#L31).
+
+2. `addWhitelistedUIProps()` is used for any other properties, current allowed props are listed [here](https://github.com/software-mansion/react-native-reanimated/blob/master/src/ConfigHelper.js#L6).
 
 In order to connect the `useAnimatedProps` hook result to a view, you need to pass it as `animatedProps` property to the `Animated` version of the component (e.g., `Animated.View`).
 The `animatedProps` property is added when a native component is wrapped with `Animated.createAnimatedComponent`.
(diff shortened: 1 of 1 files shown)


docs/docs/api/useAnimatedProps.md
```diff
@@ -11,6 +11,11 @@ The provided worklet is responsible for returning a object with view properties.
 
 Only "native" properties of "native views" can be set via `useAnimatedProps`.
 The most common usecase for this hook is when we want to animate properties of some third-party native component, since most of the properties for the core React Native components are a part of the styles anyways (at least the properties for which it makes sense to be animated).
+You can use following functions to animate properties that Reanimated don't support by default:
```
karol-bisztyga, line 14:
```suggestion
You can use the following functions to animate properties that Reanimated don't support by default:
```

docs/docs/api/useAnimatedProps.md
```diff
@@ -11,6 +11,11 @@ The provided worklet is responsible for returning a object with view properties.
...
+
+1. `addWhitelistedNativeProps()` is used to animate properties that triggers layout recalculation, you can find them [here](https://github.com/software-mansion/react-native-reanimated/blob/master/src/ConfigHelper.js#L31).
```
karol-bisztyga, line 16:
```suggestion
1. `addWhitelistedNativeProps()` is used to animate properties that trigger layout recalculation, you can find them [here](https://github.com/software-mansion/react-native-reanimated/blob/master/src/ConfigHelper.js#L31).
```

docs/docs/api/useAnimatedProps.md
```diff
@@ -11,6 +11,11 @@ The provided worklet is responsible for returning a object with view properties.
...
+
+2. `addWhitelistedUIProps()` is used for any other properties, current allowed props are listed [here](https://github.com/software-mansion/react-native-reanimated/blob/master/src/ConfigHelper.js#L6).
```
karol-bisztyga, line 18:
```suggestion
2. `addWhitelistedUIProps()` is used for any other properties, currently allowed props are listed [here](https://github.com/software-mansion/react-native-reanimated/blob/master/src/ConfigHelper.js#L6).
```

karol-bisztyga, line 18: + I'm not sure if 
> any other properties

is a good description here. I would rather go for something like "properties that are updated directly on the UI thread".

karol-bisztyga (changes requested):

docs/docs/api/useAnimatedProps.md
jakub-gonet: Fixed in cc4fe25. Thanks for the suggestions.

karol-bisztyga (approved):

PR #1577, opened 2020-12-29
Information about worklet localization to exception message

### LINKED ISSUE #1503: Error messages/alerts don't show callback function names
## Description

When errors such as a missing `runOnJS` call are thrown, the function name is printed to give the user more context to hunt down the bug. While this works for simple functions created by `const x = () => { }`, this doesn't work for other stuff such as callbacks, memos, or anything created from a return value.

PR that introduced this change: https://github.com/software-mansion/react-native-reanimated/pull/1361

### Screenshots

As you can see, this isn't really descriptive, plus it's missing the function name inside `{}`

Something like this would be much better:

## Steps To Reproduce

1. Create JS function using `useCallback`
2. Call that JS function from a worklet without using `runOnJS`

### Expected behavior

I expect the error to show the function name. This would be a great point to maybe think about reworking the errors, and instead of using the Red RN Alert, we could use the new RedBox system to even show a JS stacktrace and short snippet of the code that's causing those problems. That's how it is for normal JS errors too. 

> I believe showing a stacktrace is possible using the Hermes Runtime when manually throwing an error from in there. Then we could just call the RedBox alert API and put the Stack Trace + Filename + Line number in there and it should automatically display the small code preview window.

### Actual behavior

It throws an error without any information where it happened or what I should search for.

This works:

```ts
  const testfunc = () => {
    console.log("yeet");
  };
  const idk = useWorkletCallback(() => {
    testfunc();
  }, []);
```

This doesn't work:

```ts
  const testfunc = useCallback(() => {
    console.log("yeet");
  }, []);
  const idk = useWorkletCallback(() => {
    testfunc();
  }, []);
```

## Package versions

- React: 16.13.1
- React Native: 0.63.3
- React Native Reanimated: 2.0.0-rc.0
- NodeJS: v14.15.1

piaskowyk:
## Description

When an anonymous function is called `fn(() => {})` we can't get the name of this function. In this case the exception message isn't really helpful. I changed the exception message and also added information about worklet localization. It isn't exactly localization of this lambda/function called, but localization of the worklet where lambda/function is called.

Imrovement for PR: [#1361](https://github.com/software-mansion/react-native-reanimated/pull/1361)
Fixes #1503

### Before
Screenshot 2020-12-29 at 12 34 23

### After
Screenshot 2020-12-29 at 12 21 55

## Test code
<details>
<summary>Example of code</summary>

```js
import { useAnimatedStyle } from 'react-native-reanimated';
import React, { useCallback } from 'react';

  const testfunc = useCallback(() => {
    console.log('test');
  }, []);

  useAnimatedStyle(() => {
    testfunc();
    return {};
  });

  return (<></>);
}
```
</details>

### DIFF
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -6,18 +6,17 @@
 #include "MutableValueSetterProxy.h"
 #include "RemoteObject.h"
 #include "FrozenObject.h"
+#include <jsi/jsi.h>
 
 namespace reanimated {
 
 const char *HIDDEN_HOST_OBJECT_PROP = "__reanimatedHostObjectRef";
 const char *ALREADY_CONVERTED= "__alreadyConverted";
-std::string CALLBACK_ERROR_PREFIX = R"(
-Tried to synchronously call function {)";
-std::string CALLBACK_ERROR_SUFFIX = R"(} from a different thread.
+std::string CALLBACK_ERROR_SUFFIX = R"(
+
 Solution is:
 a) If you want to synchronously execute this method, mark it as a Worklet
-b) If you want to execute this method on the JS thread, wrap it using runOnJS
-)";
+b) If you want to execute this method on the JS thread, wrap it using runOnJS )";
   
 void addHiddenProperty(jsi::Runtime &rt,
                        jsi::Value &&value,
@@ -223,8 +222,22 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
             const jsi::Value *args,
             size_t count
             ) -> jsi::Value {
-          module->errorHandler->setError(CALLBACK_ERROR_PREFIX + hostFunction->functionName + CALLBACK_ERROR_SUFFIX);
+            
+          jsi::Value jsThis = rt.global().getProperty(rt, "jsThis");
+          std::string workletLocation = jsThis.asObject(rt).getProperty(rt, "__location").toString(rt).utf8(rt);
+          std::string exceptionMessage = "Tried to synchronously call ";
+          if(hostFunction->functionName == "") {
+            exceptionMessage += "anonymous function";
+          }
+          else {
+            exceptionMessage += "function {" + hostFunction->functionName + "}";
+          }
+          exceptionMessage += " from a different thread.\n\nOccurred in worklet location: ";
+          exceptionMessage += workletLocation;
+          exceptionMessage += CALLBACK_ERROR_SUFFIX;
+          module->errorHandler->setError(exceptionMessage);
           module->errorHandler->raise();
+          
           return jsi::Value::undefined();
         };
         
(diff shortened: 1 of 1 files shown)


Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -11,13 +11,11 @@ namespace reanimated {
 
 const char *HIDDEN_HOST_OBJECT_PROP = "__reanimatedHostObjectRef";
 const char *ALREADY_CONVERTED= "__alreadyConverted";
-std::string CALLBACK_ERROR_PREFIX = R"(
-Tried to synchronously call function {)";
-std::string CALLBACK_ERROR_SUFFIX = R"(} from a different thread.
+std::string CALLBACK_ERROR_SUFFIX = R"(
+
 Solution is:
```
jakub-gonet, line 16: For me this is kinda awkward wording, I'd remove it altogether.

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -223,8 +221,22 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
             const jsi::Value *args,
             size_t count
             ) -> jsi::Value {
-          module->errorHandler->setError(CALLBACK_ERROR_PREFIX + hostFunction->functionName + CALLBACK_ERROR_SUFFIX);
+            
+          jsi::Value jsThis = rt.global().getProperty(rt, "jsThis");
+          std::string workletLocation = jsThis.asObject(rt).getProperty(rt, "__location").toString(rt).utf8(rt);
+          std::string exceptionMessage = "Tried to synchronously call ";
+          if(hostFunction->functionName == "") {
```
jakub-gonet, line 228: You can probably use `.empty()` here.

jakub-gonet (approved):

Common/cpp/SharedItems/ShareableValue.cpp
karol-bisztyga: or `.size()==0`

Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -223,8 +221,22 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
...
+            exceptionMessage += "anonymous function";
+          }
+          else {
```
karol-bisztyga, line 231: this could be moved one line up so it would be `} else {`, just looks better

karol-bisztyga (approved):
I think we are good to go, you can consider changing what me and Kuba suggested, besides, mergeable 👍

Common/cpp/SharedItems/ShareableValue.cpp
piaskowyk: In my opinion It can be helpful, but maybe we should change “Solution is” -> “Possible solutions are“ 
It means that these solutions are just some of many. What do you think?

PR #1585, opened 2021-01-01
Use implementation of useSharedValue in mock

Andreyco:
Previous implementation did lack "memoize" feature, essentially yielding incorrect return in certain cases.
I added mocks of `useSharedValue`'s hook in order to "unmock" this hook and use its implementation in global mock.

PR contains hook test.

### DIFF
--- a/src/reanimated2/__mocks__/MutableValue.js
+++ b/src/reanimated2/__mocks__/MutableValue.js
@@ -1,0 +1,13 @@
+export default class MutableValue {
+  constructor(value) {
+    this._value = value;
+  }
+
+  get value() {
+    return this._value;
+  }
+
+  set value(nextValue) {
+    this._value = nextValue;
+  }
+}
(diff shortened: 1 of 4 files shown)


Andreyco:
cc @jakub-gonet @mgambati since you created/reviews v2 mocks lately.

src/reanimated2/__mocks__/MutableValue.js
```diff
@@ -0,0 +1,16 @@
+export default class MutableValue {
+  static MUTABLE_ID = 1;
+
+  constructor(value) {
+    this._id = MutableValue.MUTABLE_ID++;
```
jakub-gonet, line 5: Do we need ids for this? It's not used anywhere.

jakub-gonet (commented):
LGTM, left one comment.

src/reanimated2/__mocks__/MutableValue.js
Andreyco: No, we don't. Removing...

jakub-gonet (approved):

PR #1586, opened 2021-01-01
Use implementation of `useAnimatedStyle` in mock

Andreyco:
Previously, mock implementation was "identity function" which does not play well with real implementation.
I am adding hook test and exporting real hook implementation since it can be executed in test env.

### DIFF
--- a/src/reanimated2/__mocks__/NativeReanimated.js
+++ b/src/reanimated2/__mocks__/NativeReanimated.js
@@ -1,7 +1,5 @@
 import MutableValue from './MutableValue';
 
-global._setGlobalConsole = (val) => {};
-
 export default {
   installCoreFunctions: () => {},
   makeShareable: (worklet) => worklet,
--- a/src/reanimated2/core.js
+++ b/src/reanimated2/core.js
@@ -1,47 +1,8 @@
 /* global _WORKLET _getCurrentTime _frameTimestamp _eventTimestamp, _setGlobalConsole */
 
-import NativeReanimated from './NativeReanimated';
+import './globals';
 import { Platform } from 'react-native';
-
-global.__reanimatedWorkletInit = function(worklet) {
-  worklet.__worklet = true;
-};
-
-function _toArrayReanimated(object) {
-  'worklet';
-  if (Array.isArray(object)) {
-    return object;
-  }
-  if (
-    typeof Symbol !== 'undefined' &&
-    (typeof Symbol === 'function' ? Symbol.iterator : '@@iterator') in
-      Object(object)
-  )
-    return Array.from(object);
-  throw new 'Invalid attempt to spread non-iterable instance.\nIn order to be iterable, non-array objects must have a [Symbol.iterator]() method.'();
-}
-
-function _mergeObjectsReanimated() {
-  'worklet';
-  return Object.assign.apply(null, arguments);
-}
-
-global.__reanimatedWorkletInit = function(worklet) {
-  worklet.__worklet = true;
-
-  if (worklet._closure) {
-    const closure = worklet._closure;
-    Object.keys(closure).forEach((key) => {
-      if (key === '_toConsumableArray') {
-        closure[key] = _toArrayReanimated;
-      }
-
-      if (key === '_objectSpread') {
-        closure[key] = _mergeObjectsReanimated;
-      }
-    });
-  }
-};
+import NativeReanimated from './NativeReanimated';
 
 function pushFrame(frame) {
   NativeReanimated.pushFrame(frame);
@@ -55,11 +16,6 @@ export function requestFrame(frame) {
     pushFrame(frame);
   }
 }
-
-global._WORKLET = false;
-global._log = function(s) {
-  console.log(s);
-};
 
 export function runOnUI(worklet) {
   return makeShareable(worklet);
--- a/src/reanimated2/testUtils.js
+++ b/src/reanimated2/testUtils.js
@@ -1,0 +1,18 @@
+import hash from 'string-hash-64';
+
+export function createWorkletMock(fun) {
+  var _f = function _f() {
+    return fun();
+  };
+
+  var funString = fun.toString();
+
+  _f._closure = {};
+  _f.asString = funString;
+  _f.__workletHash = hash(funString);
+  _f.__location = 'jest tests fixture (1:0)';
+
+  global.__reanimatedWorkletInit(_f);
+
+  return _f;
+}
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "@babel/preset-typescript": "^7.7.4",
     "@react-native-community/bob": "^0.14.3",
     "@react-native-community/eslint-config": "^0.0.5",
+    "@testing-library/react-hooks": "^3.7.0",
     "@types/babel-types": "^7.0.9",
     "@types/babel__core": "^7.1.12",
     "@types/babel__generator": "^7.6.2",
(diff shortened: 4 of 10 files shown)


Andreyco:
cc @jakub-gonet @mgambati since you created/reviews v2 mocks lately.

I am not sure how you feel about adding `@testing-library/react-hooks` dependency. If you are good with it, I will update #1585 and make use of this addition there as well. Seems like usage will only grow.

src/reanimated2/__mocks__/NativeReanimated.js
```diff
@@ -2,7 +2,20 @@ global._setGlobalConsole = (val) => {};
...
+    var _f = function _f() {
+      return worklet();
+    };
+
+    _f._closure = {};
+    _f.asString = worklet.toString();
+    _f.__workletHash = Math.floor(Math.random() * 10000000000);
+    _f.__location = 'jest tests fixture (1:0)';
+
+    global.__reanimatedWorkletInit(_f);
```
jakub-gonet, line 15: Can we rely on this function being defined? AFAIK it comes from the plugin and some folks disable it in test env.

src/reanimated2/__mocks__/NativeReanimated.js
```diff
@@ -2,7 +2,20 @@ global._setGlobalConsole = (val) => {};
...
   installCoreFunctions: () => {},
-  makeShareable: (worklet) => worklet,
+  makeShareable: (worklet) => {
+    var _f = function _f() {
+      return worklet();
+    };
+
+    _f._closure = {};
+    _f.asString = worklet.toString();
+    _f.__workletHash = Math.floor(Math.random() * 10000000000);
```
jakub-gonet, line 12: 1. Why such a number?
2. We definitely don't want to call `Math.random()` here, `__workletHash` should have at least a deterministic and, preferably, a static value.

jakub-gonet (commented):
I'm fine with additional devDependency, it doesn't do any harm.

Left two comments.

src/reanimated2/__mocks__/NativeReanimated.js
Andreyco: Indeed, it's not available. I am mocking it in host project. 
What is your proposal? Provide a mock?

src/reanimated2/__mocks__/NativeReanimated.js
Andreyco: 1. Based on my observation, babel transformation results in following structure. `__workletHash` is 14 digits long number. Did not dig deep enough, instead I decided to generate random number on each run. 
2. Agree, long term solution is to generate stable hash for passed function. Anything on your mind? Is it even important for mocking? Not sure.

jakub-gonet: AFAIK `__workletHash` just needs to be unique and the same worklet (looks like a hash is built based on arguments and function content as string) should have the same hash. I'm pretty sure we can use an increasing id for that or hash of `_f.toString()` to be correct.

src/reanimated2/__mocks__/NativeReanimated.js
jakub-gonet: I think we can mock it like in [`core.js`](https://github.com/software-mansion/react-native-reanimated/blob/3781d823b597a7afc8005a6ff96f3755e122610e/src/reanimated2/core.js#L6-L8) or just make it noop.

Andreyco: alright. This will not break if folks disable it in test env.

src/reanimated2/__mocks__/NativeReanimated.js
```diff
@@ -1,7 +1,5 @@
 import MutableValue from './MutableValue';
 
-global._setGlobalConsole = (val) => {};
-
```
Andreyco, line 4: Moved into `globals.js`

src/reanimated2/core.js
```diff
@@ -56,11 +17,6 @@ export function requestFrame(frame) {
   }
 }
 
-global._WORKLET = false;
-global._log = function(s) {
-  console.log(s);
-};
-
```
Andreyco, line 63: All "globals" moved into `globals.js`. These are available in test env and don't need to be mocked.

src/reanimated2/testUtils.js
```diff
@@ -0,0 +1,18 @@
...
+
+  _f._closure = {};
+  _f.asString = funString;
+  _f.__workletHash = hash(funString);
+  _f.__location = 'jest tests fixture (1:0)';
+
+  global.__reanimatedWorkletInit(_f);
+
+  return _f;
+}
```
Andreyco, line 18: I noticed there is `createWorklet` function (that is listed among other function to be processed by babel plugin) so I decided to create test stub for this one.

In longer term, there are two options:
- enable babel transform in tests (are there any limitations against?)
- if babel tranform for tests is not an option, split `Hooks.js` and other functions (that are transformet by the plugin) into separate files. Then create hook mock that will create worklet mock and pass it to original hook implementation.
eg:
```js
function useWhatever(callback) {
   const worklet = createWorkletMock(callback)
   return useWhateverOriginal(worklet)
}
```

src/reanimated2/__mocks__/NativeReanimated.js
Andreyco: After review, `__reanimatedWorkletInit` is indeed available in test env (since `core.js` was loaded every time) and is basically noop. However, I moved such globals into dedicated file (if there is any need to mock them in the future)

src/reanimated2/__mocks__/NativeReanimated.js
Andreyco: Created a test util for generating "worklet" with stable hash. Same way as plugin does.

Andreyco:
@jakub-gonet left some comments...

src/reanimated2/testUtils.js
jakub-gonet: > enable babel transform in tests (are there any limitations against?)

Depends if you're talking about Reanimated2 internal tests or enabling transforms by every user of reanimated. AFAIK they're disabled by default because transforming node_modules take time.

I think @karol-bisztyga or @Szymon20000 can decide on this one.

src/reanimated2/core.js
jakub-gonet: I'm not familiar with plugin, @karol-bisztyga is this safe to do? (I'm mostly not sure if _WORKLET is used per module or not)

src/reanimated2/testUtils.js
Andreyco: > Depends if you're talking about Reanimated2 internal tests or enabling transforms by every user of reanimated. 

Enabling transform for internal tests.

> AFAIK they're disabled by default because transforming node_modules take time.

Was under impression node_modules is excluded from transformation run.

src/reanimated2/core.js
Andreyco: It was redeclare agains, several lines bellow. I believe this served as test mock and is a leftover, somehow.

piaskowyk:
Hey @Andreyco 
Thanks for your commitment 🙌  , but unfortunately I must close this PR because we integrated the Jest test for Reanimated last time. You can check this here: https://github.com/software-mansion/react-native-reanimated/pull/1731. I think that this PR can resolve your problems. We use web implementation of Reanimated as test mocks, well mocks should work as native implementation now. It going to appear in the next release (in few days). If you will have any problems with the new test integration let know us in discussion or report issue.

PR #1590, opened 2021-01-04
Support animatedProps and fix createAnimatedComponent typing

wcandillon:


react-native-reanimated.d.ts
```diff
@@ -238,8 +238,10 @@ declare module 'react-native-reanimated' {
       getNode(): ReactNativeScrollView;
     }
     export class Code extends Component<CodeProps> {}
-    export function createAnimatedComponent<S extends object, P extends { style?: StyleProp<S>; }>(component: ComponentType<P>): ComponentType<AnimateProps<S, P>>;
-
+    export function createAnimatedComponent<C extends ComponentClass, S extends object>(
+      component: C
+    ): ComponentType<AnimateProps<S, ComponentProps<C>>>;
```
jakub-gonet, line 243: 1. Why do we have a generic parameter for component here? I'm not quite sure how it works.
2. `S extends object` – if we want to allow only dict-like values we should use `S extends Record<string, unknown>`. We can change it in the AnimateProps generic arg too.

jakub-gonet (commented):
Is #1539 superseded by this PR? Could you reply to the comments there?

wcandillon:
#1539 could still be merged in the meantime

react-native-reanimated.d.ts
wcandillon: Yes I would really like to fix this too. Agree for 2. I'm also wondering if it would be possible to rewrite AnimatedProps to be in the Shape AnimatedProps<ProType> only. That would solve all our problems.

wcandillon: The problem with Record is that it doesn't accepts interfaces. I would like to remove S from AnimateProps but that seems a bit tricky.

jakub-gonet: > The problem with Record is that it doesn't accepts interfaces.

Could you give an example?

wcandillon:
@jakub-gonet you might like this new iteration ;-)

react-native-reanimated.d.ts
wcandillon: an example is in useAnimatedGestureHandler where TContext needs to be using the type keyword and not the interface one.

karol-bisztyga (approved):
good job 🙌

wcandillon:
Should we merge it?

wcandillon:
I'm also tempted to add this https://github.com/wcandillon/react-native-reanimated/pull/6 (cc @JonnyBurger). I'm not sure if there is a better way to do it.

react-native-reanimated.d.ts
```diff
@@ -210,15 +210,12 @@ declare module 'react-native-reanimated' {
     };
 
     export type AnimateProps<
-      S extends object,
-      P extends {
-        style?: StyleProp<S>;
-      }
+      P extends object
```
jakub-gonet, line 213: Wouldn't `Record<string, unknown>` fit better here?

jakub-gonet (commented):
Must've missed the update on this one, sorry. LGTM, left one comment.

jakub-gonet:
> I'm also tempted to add this wcandillon#6 (cc @JonnyBurger). I'm not sure if there is a better way to do it.

If they're required in the component then we shouldn't make it optional. Why you want to change that?

wcandillon:
> > I'm also tempted to add this [wcandillon#6](https://github.com/wcandillon/react-native-reanimated/pull/6) (cc @JonnyBurger). I'm not sure if there is a better way to do it.
> 
> If they're required in the component then we shouldn't make it optional. Why do you want to change that?

As shown in the test case, if the prop is mandatory, we have no way to check if it's there as it might be set via animatedProps

react-native-reanimated.d.ts
wcandillon: Props typed via interfaces doesn't extend Record.

jakub-gonet: I see.

react-native-reanimated.d.ts
```diff
@@ -210,15 +210,12 @@ declare module 'react-native-reanimated' {
...
     > = {
       [K in keyof P]: K extends 'style'
-        ? StyleProp<AnimateStyle<S>>
+        ? StyleProp<AnimateStyle<P[K]>>
         : P[K] | AnimatedNode<P[K]>;
-    };
+    } & { animatedProps?: AnimateProps<P> };
```
jakub-gonet, line 218: BTW, `AnimateProps<P>` here shouldn't be `Omit<AnimateProps<P>, "animatedProps">`?

wcandillon: Could you update the test for that? I'm not sure what you mean.

jakub-gonet: I meant that in the current form we can pass `{animatedProps: {animatedProps: {}}}` if I'm not mistaken.

wcandillon:
I would be more confortable to have a test for it.

wcandillon:
I added a test case that showcases another improvement from this PR

jakub-gonet (approved):

Issue #1488, opened 2020-11-29
RGBA opacity animation on v2 is not working properly


beqramo:
## Description

Hi, I want background color animation with opacity from ``#ffffff00`` to ``#ffffff66``, but it acts weird, during animation it shows really different colors. 

### Screenshots

<img src="https://user-images.githubusercontent.com/40486471/100538604-2d112880-324a-11eb-934d-444cf2da8ad0.gif" width="50%">

```js 
const background = useAnimatedStyle(() => {
    return {
      backgroundColor: interpolateColor(
        animation.value,
        [0, 1],
        ['#ffffff00', '#ffffff66'],
        'RGB',
      ),
    };
  });

```
## Package versions

- React:16.13.1"
- React Native:0.63.1
- React Native Reanimated:2.0.0-alpha.9"
- NodeJS:14.10.1

PR #1594, opened 2021-01-05
Fix interpolate of alpha channel

piaskowyk:
## Description

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1488
Fix interpolate of alpha channel. Alpha value is normalized to [0-1] float value by parser and in the next step this value is mapped to [0-255] value but mapping was without usage of round() and alpha value was still float instead of integer.

## Example code
<details>

```js
import React from 'react';
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  withTiming,
  Easing,
  interpolateColor,

  const color = useSharedValue(0)
    return {
        color.value,
        [0, 1],
        ['rgba(255,0,0,0.1)', 'rgba(0,255,0,0.9)'],
      ),
    };
  });

  return (
    <View>
      <Animated.View style={[{ height: 100, width: 100 }, animatedStyles]} />
      <Button onPress={() => {
        color.value = withTiming((color.value + 1) % 2, {duration: 300, easing: Easing.linear})
      }} title="click" />
    </View>
  );
}
```

</details>

## Changes

Call Math.round() on aplha channel value.

### Before
Screen Recording 2021-01-05 at 16 18 29

### After
Screen Recording 2021-01-05 at 16 18 49

### DIFF
--- a/src/reanimated2/Colors.js
+++ b/src/reanimated2/Colors.js
@@ -424,7 +424,7 @@ export const rgbaColor = (r, g, b, alpha = 1) => {
   if (Platform.OS === 'web' || !_WORKLET) {
     return `rgba(${r}, ${g}, ${b}, ${alpha})`;
   }
-  const a = alpha * 255;
+  const a = Math.round(alpha * 255);
   const c =
     a * (1 << 24) +
     Math.round(r) * (1 << 16) +


mrousavy:
Awesome, thanks! 🎉

jakub-gonet (approved):

jakub-gonet:
Could you also include code used to test this PR for sake of completeness?

piaskowyk:
Sure, I forgot about this. I have added example code.

mrousavy:
@piaskowyk @jakub-gonet while this works great for iOS, this doesn't work on Android!

mrousavy:
My bad, it actually does work, i just had to clear the metro cache!

PR #1596, opened 2021-01-06
Don't alllow numbers as transform value type

wcandillon:


react-native-reanimated-tests.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable */
 import React, { useState } from 'react';
-import { StyleSheet, Button, View } from 'react-native';
+import { StyleSheet, Button, View, ViewStyle } from 'react-native';
```
Andreyco, line 3: `ViewStyle` is not used in this file.

wcandillon: thks, done

jakub-gonet (approved):

react-native-reanimated.d.ts
```diff
@@ -197,13 +197,13 @@ declare module 'react-native-reanimated' {
             ? ReadonlyArray<AnimateStyle<S[K][0]>>
             : S[K] extends object
             ? AnimateStyle<S[K]> // allow `number` where `string` normally is to support colors
```
jakub-gonet, line 199: This comment is stale I believe.

jakub-gonet:
@wcandillon 
I updated the branch and removed the stale commit.

PR #1598, opened 2021-01-07
Add runOnJS, withDelay mock

mym0404:
### runOnJS

I added `runOnJS` mock for V2.
The usage of `runOnJS` function is switching the running context of a JS function. 
```js
runOnJS(externalLibraryFunction)(args);
```
So, we can just return the function directly. 

### withDelay

I added empty mock function.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/src/reanimated2/mock.js
+++ b/src/reanimated2/mock.js
@@ -25,8 +25,8 @@ const ReanimatedV2 = {
     cb && setTimeout(() => cb(true), 0);
     return 0;
   },
-  withDelay: () => {
-    return 0;
+  withDelay: (_, animationValue) => {
+    return animationValue;
   },
   cancelAnimation: NOOP,
   delay: (_, b) => b,
--- a/src/reanimated2/mock.js
+++ b/src/reanimated2/mock.js
@@ -25,6 +25,9 @@ const ReanimatedV2 = {
     cb && setTimeout(() => cb(true), 0);
     return 0;
   },
+  withDelay: () => {
+    return 0;
+  },
   cancelAnimation: NOOP,
   delay: (_, b) => b,
   sequence: () => 0,
--- a/src/reanimated2/mock.js
+++ b/src/reanimated2/mock.js
@@ -54,6 +54,8 @@ const ReanimatedV2 = {
     out: ID,
     inOut: ID,
   },
+
+  runOnJS: (fn) => fn,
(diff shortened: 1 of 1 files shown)


src/reanimated2/mock.js
```diff
@@ -25,6 +25,9 @@ const ReanimatedV2 = {
     cb && setTimeout(() => cb(true), 0);
     return 0;
   },
+  withDelay: () => {
+    return 0;
+  },
```
piaskowyk, line 30: The definition of `withDelay` is `export function withDelay(delayMs, _nextAnimation)` so I think that it should be:
```js
withDelay: (delayMs, cb) => {
  cb && setTimeout(() => cb(true), delayMs);
  return 0;
},
```

jakub-gonet: We definitely don't want to delay those methods and I think every timeout in this file can be replaced with just `cb && cb(true)`

mym0404: @jakub-gonet thankx

mym0404: @piaskowyk @jakub-gonet 
But the second argument of `withDelay` is an animation object not callback.
In docs, this is used like `withDelay(1000, withTiming(70));`.

I think the `withDelay` mock should be like
```js
  withDelay: (_, animationValue) => {
    return animationValue;
  },
```

src/reanimated2/mock.js
```diff
@@ -25,6 +25,9 @@ const ReanimatedV2 = {
     cb && setTimeout(() => cb(true), 0);
     return 0;
   },
+  withDelay: (_, animationValue) => {
+    return animationValue;
+  },
   cancelAnimation: NOOP,
   delay: (_, b) => b,
```
jakub-gonet, line 32: Could you also remove this one? It's an old name for `withDelay`, your implementation should replace it.
Same with sequence and repeat, but I think the implementation of those should be changed as well, so we can do it in the follow-up (but you can fix it here if you want to).

piaskowyk:
Thanks, @mym0404 for your commitment! 🙌
I merged your changes in this PR: https://github.com/software-mansion/react-native-reanimated/pull/1621

PR #1601, opened 2021-01-07
Fix no host instance case

### LINKED ISSUE #1569: createAnimatedComponent with 'Stop' from 'react-native-svg' throws Error
## Description
If you try to do createAnimatedComponent on a Stop component from 'react-native-svg' an error appears.
It works fine if you use Animated.createAnimatedComponent from 'react-native' but than I can not use 'animatedProps' 

## Steps To Reproduce

1. Create an Svg Component like this and try it out:
```
import Svg, { Defs, LinearGradient, Stop } from 'react-native-svg';

import React from 'react';

const AnimatedStop = Animated.createAnimatedComponent(Stop);

const TestComponent = () => (
  <View>
    <Svg
      height="20"
      width="20"
      viewBox="0 0 20 20"
      xmlns="http://www.w3.org/2000/svg"
      preserveAspectRatio="none"
    >
      <Defs>
        <LinearGradient
          id="animatedmood"
          x1={20 / 2}
          y1="0"
          x2={20 / 2}
          y2={20}
          gradientUnits="userSpaceOnUse"
        >
          <AnimatedStop />
          <AnimatedStop />
        </LinearGradient>
      </Defs>
    </Svg>
  </View>
);

export default TestComponent;

```

### Expected behavior
Should just work, like with original react-native

### Actual behavior
Than following error shows up: 
```
TypeError: null is not an object (evaluating 'hostInstance._nativeTag')
```

## Package versions

- React: "16.13.1"
- React Native: "0.63.4"
- React Native Reanimated: "2.0.0-rc.0"
- NodeJS: "v12.14.1"

karol-bisztyga:
## Description

This fixes the case when we create an animated component out of a component that does not render anything(render function returns `null`).

An example would be [svg Stop](https://github.com/react-native-svg/react-native-svg/blob/develop/src/elements/Stop.tsx)

Fixes #1569 

<details>
<summary>code example</summary>

```
import Svg, { Defs, Stop, RadialGradient, ClipPath, G, Circle, Ellipse, Rect, Polygon } from 'react-native-svg';

import React from 'react';
import { Text } from 'react-native';

const AnimatedStop = Animated.createAnimatedComponent(Stop);

const Test = () => (
  <Svg height="100" width="100">
    <Defs>
      <RadialGradient
        id="grad"
        cx="50%"
        cy="50%"
        rx="50%"
        ry="50%"
        fx="50%"
        fy="50%"
        gradientUnits="userSpaceOnUse">
        <AnimatedStop offset="0%" stopColor="#ff0" stopOpacity="1" />
        <AnimatedStop offset="100%" stopColor="#00f" stopOpacity="1" />
      </RadialGradient>
      <ClipPath id="clip">
        <G scale="0.9" x="10">
          <Circle cx="30" cy="30" r="20" />
          <Ellipse cx="60" cy="70" rx="20" ry="10" />
          <Rect x="65" y="15" width="30" height="30" />
          <Polygon points="20,60 20,80 50,70" />
          <Text
            x="50"
            y="30"
            fontSize="32"
            fonWeight="bold"
            textAnchor="middle"
            scale="1.2">
            Q
          </Text>
        </G>
      </ClipPath>
    </Defs>
    <Rect
      x="0"
      y="0"
      width="100"
      height="100"
      fill="url(#grad)"
      clipPath="url(#clip)"
    />
  </Svg>
);

export default Test;

```

</details>

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -250,21 +250,18 @@ export default function createAnimatedComponent(Component) {
         viewTag = findNodeHandle(this);
         viewName = null;
       } else {
+        // hostInstance can be null for a component that doesn't render anything (render function returns null). Example: svg Stop: https://github.com/react-native-svg/react-native-svg/blob/develop/src/elements/Stop.tsx
         const hostInstance = RNRenderer.findHostInstance_DEPRECATED(this);
-        if (hostInstance === null) {
-          // in a case when there is no view tag for this component we want to cut further operations that use view tag and name. This happens when a component doesn't render anything (render function returns null). Example: svg Stop: https://github.com/react-native-svg/react-native-svg/blob/develop/src/elements/Stop.tsx
-          return;
-        }
         // we can access view tag in the same way it's accessed here https://github.com/facebook/react/blob/e3f4eb7272d4ca0ee49f27577156b57eeb07cf73/packages/react-native-renderer/src/ReactFabric.js#L146
-        viewTag = hostInstance._nativeTag;
+        viewTag = hostInstance?._nativeTag;
         /**
          * RN uses viewConfig for components for storing different properties of the component(example: https://github.com/facebook/react-native/blob/master/Libraries/Components/ScrollView/ScrollViewViewConfig.js#L16).
          * The name we're looking for is in the field named uiViewClassName.
          */
-        viewName = hostInstance.viewConfig?.uiViewClassName;
+        viewName = hostInstance?.viewConfig?.uiViewClassName;
         // update UI props whitelist for this view
         if (this._hasReanimated2Props(styles)) {
-          adaptViewConfig(hostInstance.viewConfig);
+          adaptViewConfig(hostInstance?.viewConfig);
         }
       }
 
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -260,8 +260,8 @@ export default function createAnimatedComponent(Component) {
          */
         viewName = hostInstance?.viewConfig?.uiViewClassName;
         // update UI props whitelist for this view
-        if (this._hasReanimated2Props(styles)) {
-          adaptViewConfig(hostInstance?.viewConfig);
+        if (hostInstance && this._hasReanimated2Props(styles)) {
+          adaptViewConfig(hostInstance.viewConfig);
         }
       }
 
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -251,6 +251,10 @@ export default function createAnimatedComponent(Component) {
         viewName = null;
       } else {
         const hostInstance = RNRenderer.findHostInstance_DEPRECATED(this);
+        if (hostInstance === null) {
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):

src/createAnimatedComponent.js
```diff
@@ -250,17 +250,18 @@ export default function createAnimatedComponent(Component) {
...
         /**
          * RN uses viewConfig for components for storing different properties of the component(example: https://github.com/facebook/react-native/blob/master/Libraries/Components/ScrollView/ScrollViewViewConfig.js#L16).
          * The name we're looking for is in the field named uiViewClassName.
          */
-        viewName = hostInstance.viewConfig?.uiViewClassName;
+        viewName = hostInstance?.viewConfig?.uiViewClassName;
         // update UI props whitelist for this view
         if (this._hasReanimated2Props(styles)) {
-          adaptViewConfig(hostInstance.viewConfig);
+          adaptViewConfig(hostInstance?.viewConfig);
```
jakub-gonet, line 264: [This will throw](https://github.com/software-mansion/react-native-reanimated/blob/6990128ef50399004692dce589a88df65bd76418/src/ConfigHelper.js#L134-L135) if you pass an `undefined` here.

src/createAnimatedComponent.js
```diff
@@ -250,17 +250,18 @@ export default function createAnimatedComponent(Component) {
         viewTag = findNodeHandle(this);
         viewName = null;
       } else {
+        // hostInstance can be null for a component that doesn't render anything (render function returns null). Example: svg Stop: https://github.com/react-native-svg/react-native-svg/blob/develop/src/elements/Stop.tsx
         const hostInstance = RNRenderer.findHostInstance_DEPRECATED(this);
         // we can access view tag in the same way it's accessed here https://github.com/facebook/react/blob/e3f4eb7272d4ca0ee49f27577156b57eeb07cf73/packages/react-native-renderer/src/ReactFabric.js#L146
-        viewTag = hostInstance._nativeTag;
+        viewTag = hostInstance?._nativeTag;
```
jakub-gonet, line 256: Note: this will make `viewDescriptor.tag = undefined`. I'm not experienced enough with Rea2 to say anything so I'm just mentioning it.

src/createAnimatedComponent.js
karol-bisztyga: Yes I know but I don't see an option for it to be undefined

jakub-gonet: Then why we have `?.` here?

karol-bisztyga: this executes only if you attach UAS result to it but maybe there could be a check here to make it safer just in case

karol-bisztyga: hostInstance can be null but not undefined, but the whole expression might be undefined, I misunderstood, sorry

karol-bisztyga: fixed 9492c29

src/createAnimatedComponent.js
karol-bisztyga: I know about that, it is handled by the checks + view descriptor AFAIK only plays a role when you use an animated style which doesn't really make sense when nothing is rendered

PR #1602, opened 2021-01-07
Check the lengths when comparing deps.

Szymon20000:
## Description

We have to check if someone added a new dependency.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -417,7 +417,7 @@ function areDependenciesEqual(nextDeps, prevDeps) {
   var objectIs = typeof Object.is === 'function' ? Object.is : is;
 
   function areHookInputsEqual(nextDeps, prevDeps) {
-    if (prevDeps === null || (prevDeps.length !== nextDeps.length)) return !1;
+    if (!nextDeps || !prevDeps || (prevDeps.length !== nextDeps.length)) return !1;
     for (var i = 0; i < prevDeps.length && i < nextDeps.length; i++)
       if (!objectIs(nextDeps[i], prevDeps[i])) return !1;
     return !0;
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -417,7 +417,7 @@ function areDependenciesEqual(nextDeps, prevDeps) {
   var objectIs = typeof Object.is === 'function' ? Object.is : is;
 
   function areHookInputsEqual(nextDeps, prevDeps) {
-    if (prevDeps === null) return !1;
+    if (prevDeps === null || (prevDeps.length !== nextDeps.length)) return !1;
     for (var i = 0; i < prevDeps.length && i < nextDeps.length; i++)
       if (!objectIs(nextDeps[i], prevDeps[i])) return !1;
     return !0;


src/reanimated2/Hooks.js
```diff
@@ -417,7 +417,7 @@ function areDependenciesEqual(nextDeps, prevDeps) {
   var objectIs = typeof Object.is === 'function' ? Object.is : is;
 
   function areHookInputsEqual(nextDeps, prevDeps) {
-    if (prevDeps === null) return !1;
+    if (prevDeps === null || (prevDeps.length !== nextDeps.length)) return !1;
```
jakub-gonet, line 420: Unrelated, why we have `!1` here and `!0` a little below?

karol-bisztyga, line 420:
```suggestion
    if (!nextDeps || !prevDeps || (prevDeps.length !== nextDeps.length)) return !1;
```
Also, there's only a check for `prevDeps === null`, we don't check `nextDeps`(`null.length` throws) and we don't check them for `undefined. I suggested the changes above(implicit cast for boolean is ok since `!![] === true`).

karol-bisztyga: Also I think this check here https://github.com/software-mansion/react-native-reanimated/pull/1602/files#diff-e38a918776f5250bde86ec20338b6ac8f31dabe7533c7f85fa5756d5a90cac95R396 should be converted into `if (!dependencies)`, I checked and passing `null` as dependencies causes an error, some people might actually do that

karol-bisztyga (changes requested):

src/reanimated2/Hooks.js
karol-bisztyga: As for the first comment, seems like the problem is in fact in `buildDependencies`. I don't see how either of args passed to `areHookInputsEqual` could be something other than an array. So checking for anything but the length there seems to be redundant.

karol-bisztyga: what do you mean
> Unrelated

?

jakub-gonet: I mean, this is not part of the changes here. I was just curious why we don't use plain booleans here.

karol-bisztyga: I misunderstood you again today 😅 I thought you were telling the code is unrelated but I see you meant the comment was.

Szymon20000: https://github.com/facebook/react-native/blob/master/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js#L10006 React native is not checking if the nextDeps are null.

karol-bisztyga: They do for instance [here](https://github.com/facebook/react-native/blob/master/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js#L10862), [here](https://github.com/facebook/react-native/blob/master/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js#L11006) or [here](https://github.com/facebook/react-native/blob/master/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js#L11034). With that said, I don't think we can rely on something like that as we build dependencies ourselves and invoke this function in our code, our way, so it kind of makes a difference.

karol-bisztyga (approved):

PR #1605, opened 2021-01-08
Handle null deps in hooks

karol-bisztyga:
## Description

We don't handle the case when `null` is passed as dependencies in either `useAnimatedGestureHandler` or `useAnimatedScrollHandler`. Although in types it's not an option we have to consider that not all of the users use TypeScript.

Automatic casting to boolean works fine here as `!![] === true`.

<details>

```
    onActive: (event, ctx) => {
        // ...
    },
  }, null);
```

</details>

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -393,7 +393,7 @@ function buildWorkletsHash(handlers) {
 
 // builds dependencies array for gesture handlers
 function buildDependencies(dependencies, handlers) {
-  if (dependencies === undefined) {
+  if (!dependencies) {
     dependencies = Object.keys(handlers).map((handlerKey) => {
       const handler = handlers[handlerKey];
       return {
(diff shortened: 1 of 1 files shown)


src/reanimated2/Hooks.js
```diff
@@ -417,7 +417,8 @@ function areDependenciesEqual(nextDeps, prevDeps) {
   var objectIs = typeof Object.is === 'function' ? Object.is : is;
 
   function areHookInputsEqual(nextDeps, prevDeps) {
-    if (!nextDeps || !prevDeps || (prevDeps.length !== nextDeps.length)) return !1;
+    if (!nextDeps || !prevDeps || prevDeps.length !== nextDeps.length)
+      return !1;
```
jakub-gonet, line 421: Can we use {} brackets here and in `for, if` below? Also, could you change those `!1` & `!0`?

karol-bisztyga: d65eea1

jakub-gonet: Thanks!

jakub-gonet (approved):

Issue #1593, opened 2021-01-05
useAnimatedProps with SVG transform crash


dpyeates:
## Description

I am using an SVG G component and I am attempting to translate it's content in the y axis using useAnimatedProps based upon a shared value. I can't get useAnimatedProps to successfully set the SVG translation. I may be doing something wrong in the implementation.

### Expected behavior
The SVG group is translated to the new y location

### Actual behavior
I get the below error and then a full crash.
```
2021-01-06 09:45:06.886599+0000 Reanimated2Playground[52387:2756775] [native] Error setting property 'transform' of RNSVGGroup with tag #5: JSON value 'translate(0 -100)' of type NSString cannot be converted to a CATransform3D. Did you pass something other than an array?
2021-01-06 09:45:06.952167+0000 Reanimated2Playground[52387:2756775] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Application tried to present modally a view controller <UIViewController: 0x7fb611420b20> that is already being presented by <UIViewController: 0x7fb62160d160>.'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff20420af6 __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007fff20177e78 objc_exception_throw + 48
	2   UIKitCore                           0x00007fff23f70e65 -[UIViewController _presentViewController:withAnimationController:completion:] + 6016
	3   UIKitCore                           0x00007fff23f71876 __63-[UIViewController _presentViewController:animated:completion:]_block_invoke + 98
	4   UIKitCore                           0x00007fff23f71bac -[UIViewController _performCoordinatedPresentOrDismiss:animated:] + 519
	5   UIKitCore                           0x00007fff23f717d4 -[UIViewController _presentViewController:animated:completion:] + 179
	6   UIKitCore                           0x00007fff23f71c62 -[UIViewController presentViewController:animated:completion:] + 155
	7   Reanimated2Playground               0x000000010ff9eb35 -[RCTRedBoxWindow showErrorMessage:withStack:isUpdate:errorCookie:] + 1381
	8   Reanimated2Playground               0x000000010ffa39f4 __67-[RCTRedBox showErrorMessage:withParsedStack:isUpdate:errorCookie:]_block_invoke + 1012
	9   libdispatch.dylib                   0x00000001138427ec _dispatch_call_block_and_release + 12
[...]

Issue #1398, opened 2020-10-30
useAnimatedProps with react-native-svg transform


nabati:
## Description

When trying to use an array value property from the return of `useAnimatedProps` with an animated component, an error is thrown. 

### Screenshots
![image](https://user-images.githubusercontent.com/1499050/97725090-0b4b4700-1ace-11eb-99c1-07816f144b3a.png)

## Steps To Reproduce

1. Repro here: https://github.com/software-mansion-labs/reanimated-2-playground/compare/master...nabati:pn/animated-props-transform-array?expand=1

### Expected behavior
That it would animate the transform property of AnimatedCircle.
```
transform: [1, 0, 0, 1, randomWidth.value * 100, 0]
```
should be translating the circle along the x axis.

### Actual behavior
- Error, see screenshot.

## Snack or minimal code example

Check either repo above or 

```
import Animated, {
  useSharedValue,
  useAnimatedProps,
} from 'react-native-reanimated';
import {Button} from 'react-native';
import React from 'react';

import {Svg, Circle} from 'react-native-svg';
import {SafeAreaView, View} from 'react-native';

const AnimatedCircle = Animated.createAnimatedComponent(Circle);

export default function AnimatedStyleUpdateExample(props) {
  const randomWidth = useSharedValue(10);

  const animatedProps = useAnimatedProps(() => {
    return {
      transform: [1, 0, 0, 1, randomWidth.value * 100, 0],
    };
  });

  return (
    <SafeAreaView style={{flex: 1}}>
      <View>
        <Button
          title="toggle"
          onPress={() => {
            randomWidth.value = Math.random() * 25;
          }}
          style={{backgroundColor: 'red'}}
        />
        <Svg
          viewBow="0 0 200 200"
          style={{borderColor: 'red', borderWidth: 2}}
          width={200}
          height={200}>
          <AnimatedCircle
            animatedProps={animatedProps}
            transform={[1, 0, 0, 1, -100, 0]}
            r={50}
            cx={100}
            cy={100}
            fill="black"
          />
        </Svg>
      </View>
    </SafeAreaView>
  );
}
```

## Package versions

- React: 16.13.1
- React Native: 0.63.1
- React Native Reanimated: 2.0.0 alpha 7
- react-native-svg: 12.1.0

Issue #1360, opened 2020-10-20
AnimatedStyle.js assumptions on properties


wcandillon:
There are couple of assumptions made by AnimatedStyle.js that don't match with react-native-svg.

1. the transform property in SVG is not an array. For instance: 
```jsx
<Path
   transform={{
          rotation: 60,
          originX: 249.5,
          originY: 266.663,
        }}
/>
```

Trying to animate this property would cause an error because transform is an object instead of an array

2. AnimatedStyle makes some color processing for backgroundColor for instance but there are many svg properties that accepts colors as values, this can lead to inconsistent behaviour/processing/assumptions with color animations across component.

Let me know if I can help with this issue.

PR #1606, opened 2021-01-08
Use adapters for animated props

karol-bisztyga:
## Description

The goal is to handle a situation when props are set differently directly underneath, on the basic components than on a higher layer of components. Let's say we have a library that exports component A. Inside of this component, we render some other components like View, Text, etc. Now we may want to name some of the props differently for some reason. Then, when we want to update animated props it will not work unless we convert them into the proper names/formats.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/1398 - in b28038d
Fixes https://github.com/software-mansion/react-native-reanimated/issues/1593 - in 99c1aa0
Fixes https://github.com/software-mansion/react-native-reanimated/issues/1360

These changes allow users to add their custom adapters for only one use ad hoc(just call `createAnimatedPropAdapter` in your app) but also there's a place(`PropAdapters.js`) where there are predefined adapters that can be reused. We are open to receiving PRs with any cases of third-party libraries ✌️ I realize I haven't covered all of the possible variants but the world outside of the window is big - there are many external libraries and here I provided a tool for handling such situations.

## Changes

- added `createAnimatedPropAdapter` + types + types tests
- added two predefined prop adapters: for SVG and TextInput
- added test for adapters
- updated docs
- `ConfigHelper.js` - assert anything changed before calling `configureProps`
- `Hooks.js` - add adapters as the third arg to UAS, also had to allow passing `null` as dependencies since it is no longer the last argument of the hook

todo

<details>
<summary>code example</summary>

```js
import Animated, {
  useSharedValue,
  useAnimatedProps,
  createAnimatedPropAdapter,
  TextInputAdapter,
  SVGAdapter,
import { Button, SafeAreaView, View, TextInput, Text } from 'react-native';
import React from 'react';
import { Svg, Path } from 'react-native-svg';

class Hello extends React.Component {
  render() {
    return <Text style={{ fontSize: this.props.helloSize }}>Hello</Text>;
  }
}

const AnimatedHello = Animated.createAnimatedComponent(Hello);

const helloAdapter = createAnimatedPropAdapter(
  (props) => {
    if (Object.keys(props).includes('helloSize')) {
      props.fontSize = props.helloSize;
      delete props.helloSize;
    }
  },
  ['fontSize']
);

export default function Test() {
  const sv = useSharedValue(10);

  const tiProps = useAnimatedProps(
    () => {
      return {
        value: `rnd: ${sv.value}`,
      };
    },
    null,
    [TextInputAdapter, SVGAdapter]
  );

  const animatedProps = useAnimatedProps(
    () => {
      return {
        transform: [1, 0, 0, 1, sv.value, 0],
      };
    },
    null,
    [TextInputAdapter, SVGAdapter]
  );

  const helloProps = useAnimatedProps(
    () => {
      return {
        helloSize: sv.value,
      };
    },
    null,
    helloAdapter
  );

  return (
    <SafeAreaView style={{ flex: 1 }}>
      <View>
        <Button
          title="toggle"
          onPress={() => {
            sv.value = Math.random() * 100;
          }}
          style={{ backgroundColor: 'red' }}
        />
        <Svg
          viewBow="0 0 400 400"
          style={{ borderColor: 'red', borderWidth: 2 }}
          width={400}
          height={400}>
          <AnimatedPath
            d="M 40 60 A 10 10 0 0 0 60 60"
            stroke="black"
            transform={{
              rotation: 100,
              translateX: 200,
            }}
            animatedProps={animatedProps}
          />
        </Svg>
        <AnimatedTextInput value="unknown" animatedProps={tiProps} />
        <AnimatedHello animatedProps={helloProps} />
      </View>
    </SafeAreaView>
  );
}

```

</details>

## Checklist

- [x] Updated TS types
- [x] Added TS types tests
- [x] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -12,6 +12,7 @@ const parse = require('@babel/parser').parse;
 const functionHooks = new Map([
   ['useAnimatedStyle', [0]],
   ['useAnimatedProps', [0]],
+  ['useAnimatedPropAdapter', [0]],
   ['useDerivedValue', [0]],
   ['useAnimatedScrollHandler', [0]],
   ['useAnimatedReaction', [0, 1]],
--- a/plugin.js
+++ b/plugin.js
@@ -12,7 +12,7 @@ const parse = require('@babel/parser').parse;
 const functionHooks = new Map([
   ['useAnimatedStyle', [0]],
   ['useAnimatedProps', [0]],
-  ['useAnimatedPropAdapter', [0]],
+  ['createAnimatedPropAdapter', [0]],
   ['useDerivedValue', [0]],
   ['useAnimatedScrollHandler', [0]],
   ['useAnimatedReaction', [0, 1]],
--- a/plugin.js
+++ b/plugin.js
@@ -7,9 +7,9 @@ const traverse = require('@babel/traverse').default;
 const parse = require('@babel/parser').parse;
 
 /**
- * holds a map of hooks names as keys and array of argument indexes which are worklets(starting from 0)
+ * holds a map of function names as keys and array of argument indexes as values which should be automatically workletized(they have to be functions)(starting from 0)
  */
-const functionHooks = new Map([
+const functionArgsToWorkletize = new Map([
   ['useAnimatedStyle', [0]],
   ['useAnimatedProps', [0]],
   ['createAnimatedPropAdapter', [0]],
@@ -490,7 +490,7 @@ function processWorklets(t, path, fileName) {
       );
     }
   } else {
-    const indexes = functionHooks.get(name);
+    const indexes = functionArgsToWorkletize.get(name);
     if (Array.isArray(indexes)) {
       indexes.forEach((index) => {
         processWorkletFunction(t, path.get(`arguments.${index}`), fileName);
--- a/react-native-reanimated-tests.tsx
+++ b/react-native-reanimated-tests.tsx
@@ -31,7 +31,7 @@ import Animated, {
   useValue,
   color,
   interpolateColors,
-  useAnimatedPropAdapter,
+  createAnimatedPropAdapter,
   useAnimatedProps,
 } from 'react-native-reanimated';
 
@@ -574,9 +574,9 @@ function interpolateColorsTest() {
 
 // update props
 function updatePropsTest() {
-  const adapter1 = useAnimatedPropAdapter((props) => {}, []);
-  const adapter2 = useAnimatedPropAdapter((props) => {}, ['prop1', 'prop2']);
-  const adapter3 = useAnimatedPropAdapter(() => {});
+  const adapter1 = createAnimatedPropAdapter((props) => {}, []);
+  const adapter2 = createAnimatedPropAdapter((props) => {}, ['prop1', 'prop2']);
+  const adapter3 = createAnimatedPropAdapter(() => {});
 
   // useAnimatedStyle(() => ({}), undefined, [adapter1, adapter2, adapter3]); // works only for useAnimatedProps
 
--- a/react-native-reanimated-tests.tsx
+++ b/react-native-reanimated-tests.tsx
@@ -74,7 +74,9 @@ function SharedValueTest() {
   const translate3 = useSharedValue(0, false);
 
   const sharedBool = useSharedValue<boolean>(false);
-  if (sharedBool.value) sharedBool.value = false;
+  if (sharedBool.value) {
+    sharedBool.value = false;
+  }
 
   return <Animated.View style={styles.container} />;
 }
@@ -578,7 +580,8 @@ function updatePropsTest() {
   const adapter2 = createAnimatedPropAdapter((props) => {}, ['prop1', 'prop2']);
   const adapter3 = createAnimatedPropAdapter(() => {});
 
-  // useAnimatedStyle(() => ({}), undefined, [adapter1, adapter2, adapter3]); // works only for useAnimatedProps
+  // @ts-expect-error works only for useAnimatedProps
+  useAnimatedStyle(() => ({}), undefined, [adapter1, adapter2, adapter3]);
 
   useAnimatedProps(
     () => ({}),
--- a/react-native-reanimated-tests.tsx
+++ b/react-native-reanimated-tests.tsx
@@ -31,6 +31,8 @@ import Animated, {
   useValue,
   color,
   interpolateColors,
+  useAnimatedPropAdapter,
+  useAnimatedProps,
 } from 'react-native-reanimated';
 
 const styles = StyleSheet.create({
@@ -72,8 +74,7 @@ function SharedValueTest() {
   const translate3 = useSharedValue(0, false);
 
   const sharedBool = useSharedValue<boolean>(false);
-  if(sharedBool.value)
-    sharedBool.value = false;
+  if (sharedBool.value) sharedBool.value = false;
 
   return <Animated.View style={styles.container} />;
 }
@@ -569,4 +570,25 @@ function interpolateColorsTest() {
     outputColorRange: ['red', 'blue'],
   });
   return color;
+}
+
+// update props
+function updatePropsTest() {
+  const adapter1 = useAnimatedPropAdapter((props) => {}, []);
+  const adapter2 = useAnimatedPropAdapter((props) => {}, ['prop1', 'prop2']);
+  const adapter3 = useAnimatedPropAdapter(() => {});
+
+  // useAnimatedStyle(() => ({}), undefined, [adapter1, adapter2, adapter3]); // works only for useAnimatedProps
+
+  useAnimatedProps(
+    () => ({}),
+    null,
+    adapter1
+  );
+
+  useAnimatedProps(
+    () => ({}),
+    null,
+    [adapter2, adapter3]
+  );
 }
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -447,11 +447,11 @@ declare module 'react-native-reanimated' {
       fn: (...args: A) => R
     ): (...args: Parameters<typeof fn>) => void;
     
-    type PropsAdapterFunctionType = (props: object) => void;
+    type PropsAdapterFunction = (props: Record<string, unknown>) => void;
     export function createAnimatedPropAdapter(
-      adapter: PropsAdapterFunctionType,
-      nativeProps?: Array<string>
-    ): PropsAdapterFunctionType;
+      adapter: PropsAdapterFunction,
+      nativeProps?: string[]
+    ): PropsAdapterFunction;
 
     export function processColor(color: number | string): number;
     export function createWorklet<A extends any[], R>(
@@ -495,7 +495,7 @@ declare module 'react-native-reanimated' {
     export function useAnimatedProps<T extends {}>(
       updater: () => T,
       deps?: DependencyList | null,
-      adapters?: PropsAdapterFunctionType | Array<PropsAdapterFunctionType> | null
+      adapters?: PropsAdapterFunction | PropsAdapterFunction[] | null
     ): T;
     export function useAnimatedGestureHandler<
       T extends GestureHandlerGestureEvent = PanGestureHandlerGestureEvent,
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -446,6 +446,13 @@ declare module 'react-native-reanimated' {
     export function runOnJS<A extends any[], R>(
       fn: (...args: A) => R
     ): (...args: Parameters<typeof fn>) => void;
+    
+    type PropsAdapterFunctionType = (props: object) => void;
+    export function useAnimatedPropAdapter(
+      adapter: PropsAdapterFunctionType,
+      nativeProps?: Array<string>
+    ): PropsAdapterFunctionType;
+
     export function processColor(color: number | string): number;
     export function createWorklet<A extends any[], R>(
       fn: (...args: A) => R
@@ -487,7 +494,8 @@ declare module 'react-native-reanimated' {
     >(updater: () => T, deps?: DependencyList): T;
     export function useAnimatedProps<T extends {}>(
       updater: () => T,
-      deps?: DependencyList
+      deps?: DependencyList | null,
+      adapters?: PropsAdapterFunctionType | Array<PropsAdapterFunctionType> | null
     ): T;
     export function useAnimatedGestureHandler<
       T extends GestureHandlerGestureEvent = PanGestureHandlerGestureEvent,
@@ -730,6 +738,7 @@ declare module 'react-native-reanimated' {
   export const SpringUtils: typeof Animated.SpringUtils;
   export const runOnUI: typeof Animated.runOnUI;
   export const runOnJS: typeof Animated.runOnJS;
+  export const useAnimatedPropAdapter: typeof Animated.useAnimatedPropAdapter;
   export const processColor: typeof Animated.processColor;
   export const makeMutable: typeof Animated.makeMutable;
   export const useValue: typeof Animated.useValue;
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -491,7 +491,7 @@ declare module 'react-native-reanimated' {
     export type AnimatedStyleProp<T extends object> = AnimateStyle<T> | RegisteredStyle<AnimateStyle<T>>;
     export function useAnimatedStyle<
       T extends AnimatedStyleProp<ViewStyle | ImageStyle | TextStyle>
-    >(updater: () => T, deps?: DependencyList): T;
+    >(updater: () => T, deps?: DependencyList | null): T;
     export function useAnimatedProps<T extends {}>(
       updater: () => T,
       deps?: DependencyList | null,
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -448,7 +448,7 @@ declare module 'react-native-reanimated' {
     ): (...args: Parameters<typeof fn>) => void;
     
     type PropsAdapterFunctionType = (props: object) => void;
-    export function useAnimatedPropAdapter(
+    export function createAnimatedPropAdapter(
       adapter: PropsAdapterFunctionType,
       nativeProps?: Array<string>
     ): PropsAdapterFunctionType;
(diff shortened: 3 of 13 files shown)


src/reanimated2/Hooks.js
```diff
@@ -270,16 +271,19 @@ function styleUpdater(viewDescriptor, updater, state, maybeViewRef) {
...
-export function useAnimatedStyle(updater, dependencies) {
+export function useAnimatedStyle(updater, dependencies, adapters) {
   const viewDescriptor = useSharedValue({ tag: -1, name: null }, false);
   const initRef = useRef(null);
   const inputs = Object.values(updater._closure);
   const viewRef = useRef(null);
+  // todo: should it be possible to attach multiple adapters?
+  adapters = !adapters || Array.isArray(adapters) ? adapters : [adapters];
 
+  // todo: add adapters to dependencies?
```
Szymon20000, line 286: We need to add adapters to the deps in order to take them into account during a fast refresh.

src/reanimated2/Hooks.js
```diff
@@ -343,6 +353,16 @@ export function useAnimatedStyle(updater, dependencies) {
 // when you need styles to animated you should always use useAS
 export const useAnimatedProps = useAnimatedStyle;
 
+export function useAnimatedPropAdapter(adapter, nativeProps) {
```
Szymon20000, line 356: If it's not a hoot why it is in the file with hooks?

Szymon20000:
Can you update the example?

src/reanimated2/Hooks.js
karol-bisztyga: fixed f5a7d01

src/reanimated2/Hooks.js
karol-bisztyga: yes, did it here 773657a

karol-bisztyga:
> Can you update the example?

done

plugin.js
```diff
@@ -12,6 +12,7 @@ const parse = require('@babel/parser').parse;
 const functionHooks = new Map([
```
Szymon20000, line 12: I think it's time to change the name.

Szymon20000 (approved):
Looks fine to me!

react-native-reanimated.d.ts
```diff
@@ -446,6 +446,13 @@ declare module 'react-native-reanimated' {
     export function runOnJS<A extends any[], R>(
       fn: (...args: A) => R
     ): (...args: Parameters<typeof fn>) => void;
+    
+    type PropsAdapterFunctionType = (props: object) => void;
```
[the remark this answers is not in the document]
jakub-gonet: Also, please rename it to `PropsAdapterFunction`, we already know it's a type.

src/reanimated2/Hooks.js
```diff
@@ -270,22 +270,25 @@ function styleUpdater(viewDescriptor, updater, state, maybeViewRef) {
...
   const viewDescriptor = useSharedValue({ tag: -1, name: null }, false);
   const initRef = useRef(null);
   const inputs = Object.values(updater._closure);
   const viewRef = useRef(null);
+  adapters = !adapters || Array.isArray(adapters) ? adapters : [adapters];
+  const adaptersHash = adapters ? buildWorkletsHash(adapters) : null;
 
   // build dependencies
-  if (dependencies === undefined) {
+  if (!dependencies) {
```
jakub-gonet, line 286: `dependencies == null` is more specific check, `!dependencies` will be true for any falsy value. Not sure if those are allowed in deps, but I think it's better to have stricker checks.

plugin.js
karol-bisztyga: done e96101f

src/reanimated2/Hooks.js
```diff
@@ -270,22 +270,25 @@ function styleUpdater(viewDescriptor, updater, state, maybeViewRef) {
...
 }
 
-export function useAnimatedStyle(updater, dependencies) {
+export function useAnimatedStyle(updater, dependencies, adapters) {
   const viewDescriptor = useSharedValue({ tag: -1, name: null }, false);
   const initRef = useRef(null);
   const inputs = Object.values(updater._closure);
   const viewRef = useRef(null);
+  adapters = !adapters || Array.isArray(adapters) ? adapters : [adapters];
+  const adaptersHash = adapters ? buildWorkletsHash(adapters) : null;
```
jakub-gonet, line 283:
```suggestion
  const adaptersHash = adapters ?? buildWorkletsHash(adapters);
```

More concise.

src/reanimated2/Hooks.js
```diff
@@ -270,22 +270,25 @@ function styleUpdater(viewDescriptor, updater, state, maybeViewRef) {
...
   }
 }
 
-export function useAnimatedStyle(updater, dependencies) {
+export function useAnimatedStyle(updater, dependencies, adapters) {
   const viewDescriptor = useSharedValue({ tag: -1, name: null }, false);
   const initRef = useRef(null);
   const inputs = Object.values(updater._closure);
   const viewRef = useRef(null);
+  adapters = !adapters || Array.isArray(adapters) ? adapters : [adapters];
```
jakub-gonet, line 282:
```suggestion
  adapters = adapters == null || Array.isArray(adapters) ? adapters : [adapters];
```

Wraps falsy values, leaves `undefined` & `null`.

piaskowyk (approved):

src/reanimated2/Hooks.js
```diff
@@ -417,7 +426,8 @@ function areDependenciesEqual(nextDeps, prevDeps) {
   var objectIs = typeof Object.is === 'function' ? Object.is : is;
 
   function areHookInputsEqual(nextDeps, prevDeps) {
-    if (!nextDeps || !prevDeps || (prevDeps.length !== nextDeps.length)) return !1;
+    if (!nextDeps || !prevDeps || prevDeps.length !== nextDeps.length)
+      return !1;
```
jakub-gonet, line 430: Braces, also !1.

src/reanimated2/PropAdapters.js
```diff
@@ -0,0 +1,36 @@
...
+  // transform - taken from https://github.com/react-native-svg/react-native-svg/blob/develop/src/elements/Shape.tsx#L200
+  if (keys.includes('transform') && Array.isArray(props.transform)) {
+    // should be an array of 6 elements
+    if (props.transform.length !== 6) {
+      throw new Error(
+        `invalid transform length of ${props.transform.length}, should be 6`
+      );
+    }
+    const transform = props.transform;
+    const x = props.x || 0;
```
jakub-gonet, line 15: `||` or `??` should be here?

src/reanimated2/PropAdapters.js
```diff
@@ -0,0 +1,36 @@
...
+    const y = props.y || 0;
+    props.transform = [
+      { translateX: transform[0] * x + transform[2] * y + transform[4] },
+      { translateX: transform[1] * x + transform[3] * y + transform[5] },
+    ];
+  }
+  // todo: other cases
```
jakub-gonet, line 22: TODO.

src/reanimated2/core.js
```diff
@@ -234,6 +235,16 @@ export const runOnJS = (fun) => {
   }
 };
 
+export function createAnimatedPropAdapter(adapter, nativeProps) {
+  const nativePropsToAdd = {};
+  nativeProps &&
+    nativeProps.forEach((prop) => {
```
jakub-gonet, line 241: `nativeProps?.forEach`

src/reanimated2/core.js
```diff
@@ -234,6 +235,16 @@ export const runOnJS = (fun) => {
   }
 };
 
+export function createAnimatedPropAdapter(adapter, nativeProps) {
```
jakub-gonet, line 238: Why do we need `adapter` here? It's not used and just returned.

src/createAnimatedComponent.js
```diff
@@ -229,7 +229,9 @@ export default function createAnimatedComponent(Component) {
     }
 
     _updateFromNative(props) {
-      this._component.setNativeProps(props);
+      if (this._component.setNativeProps) {
+        this._component.setNativeProps(props);
+      }
```
jakub-gonet, line 234: `this._component.setNativeProps?.(props)`

src/ConfigHelper.js
```diff
@@ -113,16 +113,22 @@ function configureProps() {
...
+  }
 }
 
 export function addWhitelistedUIProps(props) {
+  const oldSize = Object.keys(UI_THREAD_PROPS_WHITELIST).length;
   UI_THREAD_PROPS_WHITELIST = { ...UI_THREAD_PROPS_WHITELIST, ...props };
-  configureProps();
+  if (oldSize !== Object.keys(UI_THREAD_PROPS_WHITELIST).length) {
+    configureProps();
+  }
```
jakub-gonet, line 131: I'm assuming this is an optimization to not run the update when `UI_THREAD_PROPS_WHITELIST ` is unchanged?

What if we change `propertyName: true` to `propertyName: false`? It won't change the length.

react-native-reanimated.d.ts
```diff
@@ -484,10 +491,11 @@ declare module 'react-native-reanimated' {
     export type AnimatedStyleProp<T extends object> = AnimateStyle<T> | RegisteredStyle<AnimateStyle<T>>;
     export function useAnimatedStyle<
       T extends AnimatedStyleProp<ViewStyle | ImageStyle | TextStyle>
-    >(updater: () => T, deps?: DependencyList): T;
+    >(updater: () => T, deps?: DependencyList | null): T;
     export function useAnimatedProps<T extends {}>(
       updater: () => T,
-      deps?: DependencyList
+      deps?: DependencyList | null,
+      adapters?: PropsAdapterFunctionType | Array<PropsAdapterFunctionType> | null
```
jakub-gonet, line 498: Please use `PropsAdapterFunctionType[]` instead of `Array<PropsAdapterFunctionType>`

src/reanimated2/Hooks.js
karol-bisztyga: I think this is alright here, type is `any[] | null | undefined`. Since this is js we could add a check for an array 🤔

karol-bisztyga: But, well, in other places it's also assumed that `dependencies` is an array or undefined. So I'd leave it here as it is, it can be changed in some other PR as, from the point where I'm standing, it goes beyond this one.

src/reanimated2/Hooks.js
karol-bisztyga: I don't think this is a good change. It makes it confusing. Adapters' hash is something different than adapters.

src/reanimated2/Hooks.js
karol-bisztyga: I don't think falsy values should be wrapped. I think it's ok.

src/reanimated2/PropAdapters.js
karol-bisztyga: yes, `??` is better here

src/reanimated2/core.js
karol-bisztyga: That's how it's designed, I wanted to create one place where adapter and native props would be passed, plus, look at the plugin, it does automatic workletization of the adapter.

src/ConfigHelper.js
karol-bisztyga: Yes, I know it will not. Look at the `configureProps` function - only keys are considered.

src/reanimated2/Hooks.js
jakub-gonet: That's fair point.

src/reanimated2/PropAdapters.js
jakub-gonet: I see, resolving then.

react-native-reanimated-tests.tsx
```diff
@@ -72,8 +74,7 @@ function SharedValueTest() {
   const translate3 = useSharedValue(0, false);
 
   const sharedBool = useSharedValue<boolean>(false);
-  if(sharedBool.value)
-    sharedBool.value = false;
+  if (sharedBool.value) sharedBool.value = false;
```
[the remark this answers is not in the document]
karol-bisztyga: done a44f3d0

src/ConfigHelper.js
jakub-gonet: You're right. I checked native implementation and we use only keys. I think we should replace an object with a set in some follow-up PR, so it's clearer what it does.

react-native-reanimated-tests.tsx
```diff
@@ -570,3 +571,24 @@ function interpolateColorsTest() {
...
   return color;
 }
+
+// update props
+function updatePropsTest() {
+  const adapter1 = createAnimatedPropAdapter((props) => {}, []);
+  const adapter2 = createAnimatedPropAdapter((props) => {}, ['prop1', 'prop2']);
+  const adapter3 = createAnimatedPropAdapter(() => {});
+
+  // useAnimatedStyle(() => ({}), undefined, [adapter1, adapter2, adapter3]); // works only for useAnimatedProps
```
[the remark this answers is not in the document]
karol-bisztyga: done a44f3d0

react-native-reanimated.d.ts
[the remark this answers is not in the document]
karol-bisztyga: done a44f3d0

react-native-reanimated.d.ts
```diff
@@ -446,6 +446,13 @@ declare module 'react-native-reanimated' {
     export function runOnJS<A extends any[], R>(
       fn: (...args: A) => R
     ): (...args: Parameters<typeof fn>) => void;
+    
+    type PropsAdapterFunctionType = (props: object) => void;
+    export function createAnimatedPropAdapter(
+      adapter: PropsAdapterFunctionType,
+      nativeProps?: Array<string>
```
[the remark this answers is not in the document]
karol-bisztyga: done a44f3d0

src/reanimated2/PropAdapters.js
karol-bisztyga: done a44f3d0

src/reanimated2/core.js
karol-bisztyga: done a44f3d0

src/createAnimatedComponent.js
karol-bisztyga: done a44f3d0

react-native-reanimated.d.ts
karol-bisztyga: done a44f3d0

src/reanimated2/Hooks.js
karol-bisztyga: rebased

src/createAnimatedComponent.js
```diff
@@ -229,7 +229,8 @@ export default function createAnimatedComponent(Component) {
     }
 
     _updateFromNative(props) {
-      this._component.setNativeProps(props);
+      // eslint-disable-next-line
```
jakub-gonet, line 232: Could you disable a specific warning and write a comment about why it's ignored?

src/reanimated2/PropAdapters.js
```diff
@@ -0,0 +1,46 @@
+import { createAnimatedPropAdapter } from './core';
+
+export const SVGAdapter = createAnimatedPropAdapter((props) => {
+  'worklet';
+  const keys = Object.keys(props);
+  // transform
+  if (keys.includes('transform')) {
+    if (Array.isArray(props.transform)) {
+      // case of array with 6 values => https://github.com/react-native-svg/react-native-svg/blob/develop/src/elements/Shape.tsx#L200
```
jakub-gonet, line 9: This link can change, we should use links pinned to comment (clicking `y` key in GitHub UI changes link to commit one).

```suggestion
      // case of array with 6 values => https://github.com/react-native-svg/react-native-svg/blob/b2e2c355204ff4b10973d3afce1495f7e4167ff7/src/elements/Shape.tsx#L200
```

src/reanimated2/PropAdapters.js
```diff
@@ -0,0 +1,46 @@
...
+      const transform = props.transform;
+      const x = props.x ?? 0;
+      const y = props.y ?? 0;
+      props.transform = [
+        { translateX: transform[0] * x + transform[2] * y + transform[4] },
+        { translateX: transform[1] * x + transform[3] * y + transform[5] },
+      ];
+    } else if (typeof props.transform === 'string') {
+      // case of string 'translate(translateX translateY)'
+      // todo: handle other cases of transform string like here https://github.com/react-native-svg/react-native-svg/blob/develop/src/lib/extract/extractTransform.ts#L184
```
jakub-gonet, line 24:
```suggestion
      // case of string 'translate(translateX translateY)'
      // todo: handle other cases of transform string like here https://github.com/react-native-svg/react-native-svg/blob/b2e2c355204ff4b10973d3afce1495f7e4167ff7/src/lib/extract/extractTransform.ts#L184
```

src/reanimated2/PropAdapters.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+  // todo: other props
+});
+
+export const TextInputAdapter = createAnimatedPropAdapter(
+  (props) => {
+    'worklet';
+    const keys = Object.keys(props);
+    // convert text to value like RN does here: https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L878
```
jakub-gonet, line 39:
```suggestion
    // convert text to value like RN does here: https://github.com/facebook/react-native/blob/5803c72982efbd8e1bc64dbc7622cbd923e1729b/Libraries/Components/TextInput/TextInput.js#L878
```

src/reanimated2/core.js
```diff
@@ -234,6 +235,16 @@ export const runOnJS = (fun) => {
...
+  const nativePropsToAdd = {};
+  // eslint-disable-next-line
+  nativeProps?.forEach((prop) => {
```
jakub-gonet, line 241: Same as above, could you write a short comment on why this is ignored?

jakub-gonet (approved):
LGTM!

src/reanimated2/PropAdapters.js
karol-bisztyga: good catch, THX!

src/reanimated2/core.js
karol-bisztyga: fixed 65f7cfb

src/createAnimatedComponent.js
karol-bisztyga: fixed 65f7cfb

src/reanimated2/PropAdapters.js
karol-bisztyga: fixed e0bc1d5

src/reanimated2/PropAdapters.js
karol-bisztyga: fixed e0bc1d5

src/reanimated2/PropAdapters.js
karol-bisztyga: fixed e0bc1d5

jakub-gonet (approved):

Issue #1597, opened 2021-01-06
Value is being reset when using withDecay and PanGestureHandler


slaviqueue:
## Description

I'm trying to use scrollable view from examples directory. But it looks like when using `withDecay` with PanGestureHandler, decay animation is being reset, when starting scrolling while animation is in progress.

### Screenshots
Here's the video:

https://user-images.githubusercontent.com/20744231/103790419-d9b0a800-5049-11eb-9a88-f4d160f14f1b.mp4

## Steps To Reproduce

1. Start scrolling
2. End scrolling
3. While decaying, start scrolling again
4. `translateY` will be reset to position, from which `withDecay` animation was started

### Expected behavior
Position is not being reset, scrolling begins from current position.

### Actual behavior
Position is being reset, scrolling starts from the wrong position.

## Snack or minimal code example

https://github.com/software-mansion/react-native-reanimated/blob/master/Example/src/ScrollableViewExample.js

## Package versions

- React: 16.13.1
- React Native: 0.63.3
- React Native Reanimated: ^2.0.0-alpha.9.2
- NodeJS: 12.16.1

piaskowyk:
Hey @salockhart 
Unfortunately I can't reproduce this issue 😞 . I need more details. Could you try to run this with a newer version of Reanimated for example: 2.0.0-rc.1 and tell me what will happen? And also could you tell me which android version it is?
My teammate will check this also, maybe he can reproduce it.

slaviqueue:
@piaskowyk unfortunately, updating reanimated to 2.0.0-rc.1 did not help. I use Xiaomi Mi A1 with Android 9.

slaviqueue:
@piaskowyk thank you! :)

PR #1607, opened 2021-01-08
Fix stop condition for scroll animation in Example

piaskowyk:
## Description

The list in scroll Example app is inertness, and when list is moved and someone taps on screen and doesn't move by finger it calls only onStart() event from GestureHandler, and this sets the actual position but the list is still moving. If in the next step someone moves their finger then the list jumps to this position and starts animation of scroll from this position like in issue:

Fixes https://github.com/software-mansion/react-native-reanimated/issues/1597

## Changes
I just set the actual position of the animation to stop inertness of list.


karol-bisztyga (approved):
GJ 🙌

PR #1615, opened 2021-01-11
NodesManager.onEventDispatch(): ignore mNativeProxy when null

### LINKED ISSUE #1614: NullPointerException in NodesManager.onEventDispatch()
## Description

I'm seeing crashes in production on Android with `reanimated-2.0.0-rc.0` where NodesManager.onEventDispatch() calls a method on a null `mNativeProxy`.

Sample backtrace:
https://sentry.io/share/issue/1ae44073c9ca48879a5d9f28fbe0b785/

It looks like mNativeProxy is null in the following code (NodesManager.java around line 467):

```
      shouldSaveEvent |= (mCustomEventHandler != null && mNativeProxy.isAnyHandlerWaitingForEvent(key));
```

## Steps To Reproduce

Unknown, sorry. It seems to happen most often when the app comes into the foreground after being in the background for a while. It's showing up ~30 times per day in our Google Play metrics.

## Expected behavior

App does not crash.

## Actual behavior

App crashes, see stack trace above.

## Package versions

- React: 16.13.1
- React Native: 0.63.4
- React Native Reanimated: 2.0.0-rc.0
- NodeJS: (Hermes)

danmaas:
## Description

Fixes #1614 (maybe?)

Sorry, I have no idea if this is the right fix, or just a band-aid that hides problems elsewhere.

## Changes

Just skip calling .isAnyHandlerWaitingForEvent() when mNativeProxy is null

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -464,7 +464,7 @@ public class NodesManager implements EventDispatcherListener {
       int viewTag = event.getViewTag();
       String key = viewTag + eventName;
 
-      shouldSaveEvent |= (mCustomEventHandler != null && mNativeProxy.isAnyHandlerWaitingForEvent(key));
+      shouldSaveEvent |= (mCustomEventHandler != null && mNativeProxy != null && mNativeProxy.isAnyHandlerWaitingForEvent(key));
       if (shouldSaveEvent) {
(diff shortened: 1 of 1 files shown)


karol-bisztyga (approved):

jakub-gonet:
@Szymon20000 looked into it and it seems like this crash was a race between `NodesManager` registering to events and `nativeProxy` (which handles events) creation. We assume checking for null as you did is sufficient.

Issue #1613, opened 2021-01-11
not able to assign percent values to width


bharat-tiwari:
## Description

the way sometimes we set width of some elements in RN stylesheets as percent value - { width: '100%' }, this seem to be  not supported in reanimated2 when used with typescript 

 
### Screenshots

![1](https://user-images.githubusercontent.com/13016554/104148944-f9bdcf80-5399-11eb-99cd-376eac45493d.png)

![2](https://user-images.githubusercontent.com/13016554/104148950-017d7400-539a-11eb-9995-51dc3ab3228a.png)

## Steps To Reproduce
as can be seen in above screenshot

### Expected behavior
string values should be accepted for width prop type in useAnimation 

### Actual behavior
string values are not currently accepted for width prop type in useAnimation 

## Package versions
- React: "16.13.1"
- React Native: "0.63.4"
- React Native Reanimated: "2.0.0-alpha.9.2"
- NodeJS: 12.14.0

abumubarak-dev:
@bharat-tiwari  you the % inside the withTiming instead of the useSharedValue
E.g 
<img width="493" alt="Screenshot 2024-06-15 at 18 42 55" src="https://github.com/software-mansion/react-native-reanimated/assets/47022889/c89c4103-4a42-4708-856d-cf6ff30cc4a1">

PR #1617, opened 2021-01-11
Add support for animate string value with %

piaskowyk:
## Description

Add support for animate string value with %
Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1613

## Example code
<details>

```js
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  Easing,
import React from 'react';

  const randomWidth = useSharedValue('10%');
  
  const config = {
    duration: 500,
  };

    return {
    };
  });

  return (
    <View>
      <Animated.View
        style={[
          { width: '100%', height: 80, backgroundColor: 'black' },
          style,
        ]}
      />
      <Button
        title="toggle"
        onPress={() => {
          randomWidth.value = Math.random() * 100 + '%' ;
        }}
      />
    </View>
  );
}

```

</details>

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -403,12 +403,12 @@ declare module 'react-native-reanimated' {
       easing?: EasingFunction;
     }
     export function withTiming(
-      toValue: number | string | Exclude<ColorValue, OpaqueColorValue>,  // string as a color value like `"rgba(20,20,20,0)"`
+      toValue: number | Exclude<ColorValue, OpaqueColorValue>,  // string as a color value like `"rgba(20,20,20,0)"`
       userConfig?: WithTimingConfig,
       callback?: (isFinished: boolean) => void
     ): number;
     export function withSpring(
-      toValue: number |string | Exclude<ColorValue, OpaqueColorValue>, // string as a color value like `"rgba(20,20,20,0)"`
+      toValue: number | Exclude<ColorValue, OpaqueColorValue>, // string as a color value like `"rgba(20,20,20,0)"`
       userConfig?: WithSpringConfig,
       callback?: (isFinished: boolean) => void
     ): number;
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -403,12 +403,12 @@ declare module 'react-native-reanimated' {
       easing?: EasingFunction;
     }
     export function withTiming(
-      toValue: number | Exclude<ColorValue, OpaqueColorValue>,  // string as a color value like `"rgba(20,20,20,0)"`
+      toValue: number | string | Exclude<ColorValue, OpaqueColorValue>,  // string as a color value like `"rgba(20,20,20,0)"`
       userConfig?: WithTimingConfig,
       callback?: (isFinished: boolean) => void
     ): number;
     export function withSpring(
-      toValue: number | Exclude<ColorValue, OpaqueColorValue>, // string as a color value like `"rgba(20,20,20,0)"`
+      toValue: number |string | Exclude<ColorValue, OpaqueColorValue>, // string as a color value like `"rgba(20,20,20,0)"`
       userConfig?: WithSpringConfig,
       callback?: (isFinished: boolean) => void
     ): number;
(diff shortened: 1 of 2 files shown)


src/reanimated2/animations.js
```diff
@@ -48,6 +48,34 @@ export function transformAnimation(animation) {
...
+  }
+  animation.toValue = animation.toValue + unit;
+  animation.current = animation.current + unit;
+  animation.startValue = animation.startValue + unit;
+}
+
+export function getUnit(value) {
+  'worklet';
+  for (const unit of ['%', 'pt']) {
+    if (value.endsWith(unit)) return unit;
```
jakub-gonet, line 74: Missing braces.

src/reanimated2/animations.js
```diff
@@ -82,6 +110,21 @@ export function decorateAnimation(animation) {
     return res;
   };
 
+  const withUnitOnStart = (animation, value, timestamp, previousAnimation) => {
+    const val = parseFloat(value);
+    removeAnimationUnit(animation);
+    if (previousAnimation !== animation) removeAnimationUnit(previousAnimation);
```
jakub-gonet, line 116: Missing braces.

jakub-gonet (commented):
Could you add some examples, update TS types if necessary, and update docs?

src/reanimated2/animations.js
```diff
@@ -48,6 +48,36 @@ export function transformAnimation(animation) {
...
+    return;
+  }
+  animation.toValue = animation.toValue + unit;
+  animation.current = animation.current + unit;
+  animation.startValue = animation.startValue + unit;
+}
+
+export function getUnit(value) {
+  'worklet';
+  for (const unit of ['%', 'pt']) {
```
karol-bisztyga, line 73: maybe it would be good to define this array somewhere as, for instance, `ALLOWED_UNITS`?

karol-bisztyga (commented):
Looks alright.
Could you please add an example where the `'pt'` unit occurs?
Maybe @Szymon20000 could do a code review here?

src/reanimated2/animations.js
piaskowyk: `pt` works only for web. RN supports only `%` as unit, but RN for web supports [`%`, `pt`, `px`, `em`, `rem`, `vh`, `vw`] So what do you think, add all of this as ALLOWED_UNITS or limit support only for `%` for all platforms?

karol-bisztyga: In such a case I would go for conditional contents of allowed units.
Something like this maybe:
```
const ALLOWED_UNITS = Platform.OS === 'web' ? ['%', 'pt' ...] : ['%'];
```

react-native-reanimated.d.ts
```diff
@@ -403,12 +403,12 @@ declare module 'react-native-reanimated' {
       easing?: EasingFunction;
     }
     export function withTiming(
-      toValue: number | Exclude<ColorValue, OpaqueColorValue>,  // string as a color value like `"rgba(20,20,20,0)"`
+      toValue: number | string | Exclude<ColorValue, OpaqueColorValue>,  // string as a color value like `"rgba(20,20,20,0)"`
```
jakub-gonet, line 406: `Exclude<ColorValue, OpaqueColorValue>` already contains `string`

src/reanimated2/animations.js
```diff
@@ -21,7 +21,7 @@ export function transform(value, handler) {
   if (typeof value === 'string') {
     // toInt
     // TODO handle color
-    const match = value.match(/([A-Za-z]*)([-\d.]*)([A-Za-z]*)/);
+    const match = value.match(/([A-Za-z]*)([-\d.]*)([A-Za-z%]*)/);
```
jakub-gonet, line 24: This regex is wrong, `([-\d.]*)` allows parsing `-1-...1` as a number.
I think `(-?\d*\.?\d*)` is slightly better, it matches the optional minus sign, and optional dot between numbers (see attached screenshot).
image

jakub-gonet:
Please don't force push into branches that have a pull request and reviews. It removes old commits, so you can't check history at the time of writing comments.

karol-bisztyga (approved):

jakub-gonet (approved):

Issue #1588, opened 2021-01-03
Multiple withSpring sequences do not work.


nandorojo:
## Description

I want to do a sequence of spring animations. This works fine, until I do a sequence of springs for multiple transforms.

**This works:** (only `scale`)

```js
useAnimatedStyle(() => ({
  transform: [
    {
      scale: withSequence(withSpring(0), withSpring(1), withSpring(2))
    }
  ]
}))
```

This works, too (`withTiming` instead of `withSpring`):

```js
useAnimatedStyle(() => ({
  transform: [
    {
      scale: withSequence(withTiming(0), withTiming(1), withTiming(2))
    },
    {
      translateY: withSequence(withTiming(0), withTiming(10), withTiming(50))
    }
  ]
}))
```

**This does not work:** (both `scale` and `transformY`)

```js
useAnimatedStyle(() => ({
  transform: [
    {
      scale: withSequence(withSpring(0), withSpring(1), withSpring(2))
    },
    {
      translateY: withSequence(withSpring(0), withSpring(10), withSpring(50))
    }
  ]
}))
```

### Screenshots

<img width="395" alt="Screen Shot 2021-01-03 at 2 58 09 PM" src="https://user-images.githubusercontent.com/13172299/103489790-1bd0b400-4dd4-11eb-9805-1c3f9b8ae235.png">

Here's a video: https://imgur.com/a/reMX2cq

## Steps To Reproduce

1. Install `react-native-reanimated@2.0.0-rc.0`
2. Use the code sample below
3. Try creating a sequence with multiple transforms that both use `withSpring`. To see it working, switch `withSpring` to `withTiming`.

### Expected behavior

The springs should happen just fine in a sequence.

### Actual behavior

I get an error saying `undefined is not an object (evaluating currentAnim.onFrame`)

**If you switch `withSpring` -> `withTiming`, it works.**

## Snack or minimal code example

Doesn't work on expo snack, but here's a minimal example:

```js
import React from 'react'
import { StyleSheet } from 'react-native'
import Animated, { useAnimatedStyle, withSequence, withSpring } from 'react-native-reanimated'

export default function SequenceBug() {
  const animatedStyle = useAnimatedStyle(() => ({
    // if you change every instance of withSpring to withTiming, it works.
    transform: [
      {
        scale: withSequence(withSpring(0), withSpring(1), withSpring(2))
      },
      {
        translateY: withSequence(withSpring(0), withSpring(10), withSpring(50))
      }
[...]

PR #1618, opened 2021-01-12
do not execute finished animation

Szymon20000:
## Description

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1588

There is a case In `runAnimations` function (used by `useAnimatedStyle`) in which we have an array of animations.
Of course, we have to wait until the last animation from the array is done but we cannot execute finished animations. 
To prevent that I've added an additional check.

## Changes

Added check if the animation has already finished.

```JS
import React from 'react'
import {
  StyleSheet
} from 'react-native'
import Animated, {
  useAnimatedStyle,
  withSequence,
  withSpring
  // withTiming
} from 'react-native-reanimated'

// const withSpring = withTiming;

export default function SequenceBug() {
  const animatedStyle = useAnimatedStyle(() => ({
    // if you change every instance of withSpring to withTiming, it works.
    transform: [ {
         scale: withSequence(withSpring(0), withSpring(0.5), withSpring(1), withSpring(1.5))
     },
      {
        translateY: withSequence(withSpring(0), withSpring(5), withSpring(10), withSpring(15))
      } 
    ]
  }))
  return <Animated.View style = {
    [style.item, animatedStyle]
  }
  />
}

const style = StyleSheet.create({
  item: {
    margin: 100,
    width: 300,
    height: 300,
    backgroundColor: 'blue',
    alignSelf: 'center',
    borderRadius: 999
  }
})
```
## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -117,15 +117,20 @@ function runAnimations(animation, timestamp, key, result) {
       });
       return allFinished;
     } else if (typeof animation === 'object' && animation.onFrame) {
-      if (animation.callStart) {
-        animation.callStart(timestamp);
-        animation.callStart = null;
-      }
-      const finished = animation.onFrame(animation, timestamp);
-      animation.timestamp = timestamp;
-      if (finished) {
-        animation.finished = true;
-        animation.callback && animation.callback(true /* finished */);
+      let finished = false;
+      if (animation.finished) {
+        finished = true;
+      } else {
+        if (animation.callStart) {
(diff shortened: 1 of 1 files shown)


karol-bisztyga (approved):

karol-bisztyga:
Good catch, it could be done shorter
```
let finished = true;
if (!animation.finished) {
```
but I think it's mergeable

nandorojo:
Good work!

Issue #1610, opened 2021-01-08
[ANDROID] Easing Exp flickers on withTiming


gorhom:
## Description

Hi there 👋

While working on Reanimated v2, i notice a bug/flicker in the timing animation on Android when using `Easing.out(Easing.exp)`. It is more noticeable if animation duration is lower than 1000.

The value will flick negatively before it start animating.

### Screenshots

https://user-images.githubusercontent.com/4061838/104075341-c7359a80-5212-11eb-9143-7ecdc046f3f1.mp4

here i implemented a simple drag and drop `box` with `Pan Gesture Handler`, the `box` will become `red` when the `animatedPosition` flicks.

## Steps To Reproduce

1. run the reproducible sample code
2. drag the box and drop it multiple times

### Expected behavior

- Animated value should not flick on Android.

### Actual behavior

- Animated value flicks on Android.

## Snack or minimal code example

```tsx
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  useAnimatedReaction,
  cancelAnimation,
  useWorkletCallback,
  Easing,
  useAnimatedGestureHandler,
} from 'react-native-reanimated';
import { View, StyleSheet, Platform } from 'react-native';
import React from 'react';
import { PanGestureHandler } from 'react-native-gesture-handler';

export default function AnimatedStyleUpdateExample(props) {
  const animatedPosition = useSharedValue(0);
  const animatedLinePosition = useSharedValue(0);

  const config = {
    duration: 1000,
    easing: Easing.out(Easing.exp),
  };

  const animateToPoint = useWorkletCallback((destinationPoint) => {
    animatedPosition.value = withTiming(destinationPoint, config);
  }, []);

  const gestureHandler = useAnimatedGestureHandler(
    {
      onStart: (_ , context) => {
        context.currentPosition = animatedPosition.value;
        cancelAnimation(animatedPosition);
      },
      onActive: ({ translationY }, context) => {
        animatedLinePosition.value = context.currentPosition + translationY;
        animatedPosition.value = context.currentPosition + translationY;
      },
      onEnd: ({ state }) => {
        animateToPoint(0);
      },
    },
    [animateToPoint]
  );

  const boxAnimatedStyle = useAnimatedStyle(() => {
    return {
      backgroundColor: animatedPosition.value - animatedLinePosition.value > 0 ? 'red' : 'black',
      transform: [
        {
          translateY: animatedPosition.value,
        },
      ],
    };
[...]

PR #1620, opened 2021-01-13
Use targetTimestamp instead of timestamp

Szymon20000:
## Description

Related issue: https://github.com/software-mansion/react-native-reanimated/issues/1610

As it turned out CADisplayLink.timestamp is the time of the last frame, not the upcoming one. 

**Context from the [docs](https://developer.apple.com/documentation/quartzcore/cadisplaylink?language=objc):** ```The target can read the display link’s timestamp property to retrieve the time that the previous frame was displayed. ... To calculate the actual frame duration, use targetTimestamp - timestamp. You can use actual frame duration in your application to calculate the frame rate of the display, the approximate time that the next frame will be displayed, and to adjust the drawing behavior so that the next frame is prepared in time to be displayed.```

**Why it is a problem?**
We use ```CACurrentMediaTime()``` to compute timestamp for events what may create a situation when animation started during the event uses timestamp later than used in onFrame. The reason for that is that we use CADisplayLink.timestamp as a timestamp during the render events.

## Changes

In NativeProxy: timestamp -> targetTimestamp

```JS
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  useAnimatedReaction,
  cancelAnimation,
  useWorkletCallback,
  Easing,
  useAnimatedGestureHandler,
import { View, StyleSheet, Platform } from 'react-native';
import React from 'react';
import { PanGestureHandler } from 'react-native-gesture-handler';

  const animatedPosition = useSharedValue(0);
  const animatedLinePosition = useSharedValue(0);

  const config = {
    duration: 1000,
    easing: Easing.out(Easing.exp),
  };

  const animateToPoint = useWorkletCallback((destinationPoint) => {
    animatedPosition.value = withTiming(destinationPoint, config);
  }, []);

    {
      onStart: (_ , context) => {
        context.currentPosition = animatedPosition.value;
        cancelAnimation(animatedPosition);
      },
      onActive: ({ translationY }, context) => {
        animatedLinePosition.value = context.currentPosition + translationY;
        animatedPosition.value = context.currentPosition + translationY;
      },
      onEnd: ({ state }) => {
        animateToPoint(0);
      },
    },
    [animateToPoint]
  );

  const boxAnimatedStyle = useAnimatedStyle(() => {
    return {
      backgroundColor: animatedPosition.value - animatedLinePosition.value > 0 ? 'red' : 'black',
      transform: [
        {
          translateY: animatedPosition.value,
        },
      ],
    };
  });

  const lineAnimatedStyle = useAnimatedStyle(() => {
    return {
      transform: [
        {
          translateY: animatedLinePosition.value,
        },
      ],
    };
  });

  useAnimatedReaction(
    () => animatedPosition.value - animatedLinePosition.value,
    (value) => {
      if(value > 0){
        console.log(Platform.OS, 'Offset', value)
      }
    },
    []
  );

  return (
    <View
      style={styles.container}>
        <Animated.View style={[styles.box, boxAnimatedStyle]} />
      </PanGestureHandler>
      <Animated.View style={[styles.line, lineAnimatedStyle]} />
    </View>
  );
}

  container: {
    flex: 1,
    backgroundColor: '#444'
  },
  box: {
    position: 'absolute',
    bottom: 0,
    left: 0,
    right: 0,
    height: 400,
    backgroundColor: 'black',
  },
  line: {
    position: 'absolute',
    bottom: 400,
    left: 0,
    right: 0,
    height: 1,
    backgroundColor: 'white',
  },
});
```

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/ios/native/NativeProxy.mm
+++ b/ios/native/NativeProxy.mm
@@ -9,6 +9,7 @@
 #import <folly/json.h>
 #import <React/RCTFollyConvert.h>
 #import <React/RCTUIManager.h>
+#import <iostream>
 
 namespace reanimated {
 
@@ -113,7 +114,7 @@ std::shared_ptr<NativeReanimatedModule> createReanimatedModule(std::shared_ptr<C
 
   auto requestRender = [reanimatedModule, &module](std::function<void(double)> onRender, jsi::Runtime &rt) {
     [reanimatedModule.nodesManager postOnAnimation:^(CADisplayLink *displayLink) {
-      double frameTimestamp = displayLink.timestamp * 1000.0;
+      double frameTimestamp = displayLink.targetTimestamp * 1000;
       rt.global().setProperty(rt, "_frameTimestamp", frameTimestamp);
       onRender(frameTimestamp);
       rt.global().setProperty(rt, "_frameTimestamp", jsi::Value::undefined());
--- a/ios/native/NativeProxy.mm
+++ b/ios/native/NativeProxy.mm
@@ -9,7 +9,6 @@
 #import <folly/json.h>
 #import <React/RCTFollyConvert.h>
 #import <React/RCTUIManager.h>
-#import <iostream>
 
 namespace reanimated {
 
@@ -157,4 +156,3 @@ module = std::make_shared<NativeReanimatedModule>(jsInvoker,
 }
 
 }
-


ios/native/NativeProxy.mm
```diff
@@ -9,6 +9,7 @@
 #import <folly/json.h>
 #import <React/RCTFollyConvert.h>
 #import <React/RCTUIManager.h>
+#import <iostream>
```
jakub-gonet, line 12: Debugging artifact?

Szymon20000: 😅

karol-bisztyga, line 12:
```suggestion
```

karol-bisztyga (changes requested):

karol-bisztyga (approved):

PR #1625, opened 2021-01-14
Add `text` to the prop whitelist (for react-native-animateable-text)

JonnyBurger:
Hey!

We've been making a library called https://www.npmjs.com/package/react-native-animateable-text specifically for tieing in with Reanimated to allow to pass text as an Animated Value.

Since Alpha 9, the library does not work anymore, since `text` is not a whitelisted prop.
I would like to continue the library and try out the newer Reanimated versions, so I was wondering if it's okay to add `text` to the whitelist?

Also I am wondering if there is some way to do this in my library itself so not everybody has to send PRs to Reanimated. Thanks for considering!

### DIFF
--- a/src/ConfigHelper.js
+++ b/src/ConfigHelper.js
@@ -101,6 +101,8 @@ let NATIVE_THREAD_PROPS_WHITELIST = {
   writingDirection: true,
   /* text color */
   color: true,
+  // For react-native-animateable-text
+  text: true,
   tintColor: true,
   shadowColor: true,
 };


jakub-gonet:
You can use `addWhitelistedNativeProps()` or `addWhitelistedUIProps()` to configure it on your end. Check out [this doc page.](https://docs.swmansion.com/react-native-reanimated/docs/next/api/useAnimatedProps).

JonnyBurger:
This is exactly what I was looking for! Thank you!

wcandillon:
In redash we also use `addWhitelistedNativeProps ` for the text property. However the text property is returned by `hostInstance.viewConfig.validAttributes`. Looking at the code from `adaptViewConfig` I would expect the prop to be added automatically. Why is this not the case?

wcandillon:
Nevermind, I swapped `addWhitelistedNativeProps` and `addWhitelistedUIProps` (I assumed that UI in the name meant UI Manager). Any insights on why this property cannot be animated directly?

wcandillon:
After further investigation, the propAdapter was needed:

```
  (props) => {
    if (Object.keys(props).includes("value")) {
      props.text = props.value;
      delete props.value;
    }
  },
  ["value"]
);
```

I'm not clear why using `addWhitelistedUIProps()` does the trick too?

PR #1645, opened 2021-01-18
Add docs for native functions

karol-bisztyga:
## Description

Add docs for native functions - `scrollTo` and `measure`.

### DIFF
--- a/docs/docs/api/nativeMethods/measure.md
+++ b/docs/docs/api/nativeMethods/measure.md
@@ -4,11 +4,28 @@ title: measure
 sidebar_label: measure
 ---
 
-> Due to time constraints we weren't able to finish this page.
+Determines the location on screen, width, and height of the given view. Note that these measurements are not available until after the rendering has been completed in native. If you need the measurements as soon as possible, consider using `onLayout` instead.
+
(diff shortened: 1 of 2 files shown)


docs/docs/api/nativeMethods/measure.md
```diff
@@ -4,12 +4,29 @@ title: measure
 sidebar_label: measure
 ---
 
-> Due to time constraints we weren't able to finish this page.
+Determines the location on screen, width, and height of the given view. Note that these measurements are not available until after the rendering has been completed in native. If you need the measurements as soon as possible, consider using `onLayout` instead.
```
piaskowyk, line 7: Could we explain what is `onLayout` or add a link to documentation?

docs/docs/api/nativeMethods/measure.md
```diff
@@ -4,12 +4,29 @@ title: measure
...
+
+This function is implemented on native platforms only. On the web, it's sufficient to use a standard version of the `measure` which is available on most of the default components provided by React Native(it's [here](https://github.com/facebook/react-native/blob/65975dd28de0a7b8b8c4eef6479bf7eee5fcfb93/Libraries/Renderer/shims/ReactNativeTypes.js#L105)). In such a case it should be invoked in the following way(note it's asynchronous so if you want to make it synchronous you should use `Promise`):
+
+```javascript
+const aref = useAnimatedRef();
```
piaskowyk, line 12: Shouldn't there be `animatioinRef` instead of `aref`? because in the promise you use `animationRef` object

karol-bisztyga: fixed 0a0e54d, I switched to `aref` because it's used in other places like that, but the important thing is it's consistent now. THX for the catch :+1:

docs/docs/api/nativeMethods/measure.md
karol-bisztyga: fixed 787d616

piaskowyk (approved):

PR #1647, opened 2021-01-19
Add docs for events

karol-bisztyga:
## Description

Add docs for events


mrousavy:
You're using the `onGestureEvent` for `TapGestureHandler`s - I thought that's not recommended for discrete handlers? 

> "Keep in mind that onGestureEvent is only generated in continuous gesture handlers and shouldn't be used in the TapGestureHandler and other discrete handlers." ([source](https://docs.swmansion.com/react-native-gesture-handler/docs/about-handlers#discrete-vs-continuous))

kmagiera:
Nice catch @mrousavy — this recommendation is not applicable to reanimated2 because useAnimatedGestureHandler provides a shorthand of registering for both gesture event and gesture state changes under a single name. We should perhaps update gesture handler docs to account for that and maybe mention here that it is safe to use onGestureHandler for all handlers with reanimated2

Szymon20000 (approved):

PR #1658, opened 2021-01-25
Standarize platform-specific code

karol-bisztyga:
## Description

This PR is a response to [this idea](https://github.com/software-mansion/react-native-reanimated/discussions/1523). The goal is to gather all the platform-specific code in one place and get rid of multiple, sometimes indirect, checks. Also, if we ever want to support some other platforms that are not supported now, it will be easier to implement.

Until now, we've been using check like these:
- `if (Platform.OS !== 'ios') { ... }`
- `if (Platform.OS !== 'android') { ... }`
- `if (Platform.OS !== 'web') { ... }`

Also, in several places, we relied on certain variables' values, which was neither clear nor intuitive. A good way to make it better would be to replace all those 'variables' checks into the 'platform' ones. But I decided we could take one step further into this putting all that code together in one place instead of having it in many different places in the codebase.

## Changes

Moved all the code with platform checks to three files(`.web.js`, `.ios.js`, `.android.js`) with different versions of that code. Most of the native code is common for android and ios, therefore there's also one additional file mutual for both platforms(`PlatformSpecificNative.js`).

## Checklist

- [ ] Ensured that CI passes


nandorojo:
One suggestion is to use `.js`, `.ios.js`, and `.android.js`, getting rid of the `.web.js` for any files that don't require web-only code. 

Since most of the web files in question are pure JS, leaving out the `.web` will make these files easier to adapt to other platforms in the future.

piaskowyk:
I totally agree that this approach is more readable but unfortunately, we can't merge it because it will break Jest support. Jest mocks `Platform.OS` as 'ios' but we need to use web implementation. 😕

PR #1659, opened 2021-01-25
Add useAnimatedScrollHandler example

karol-bisztyga:
## Description

I noticed we don't have any examples of `useAnimatedScrollHandler` in our Example app.

The code is taken from https://github.com/karol-bisztyga/reanimated-2-playground/blob/707eea1e7edde74d8c23abe8a12f5604dfd29ab8/blog_apps/scroll_events/ScrollEvents.js

### DIFF
--- a/Example/src/AnimatedScrollExample.js
+++ b/Example/src/AnimatedScrollExample.js
@@ -1,0 +1,279 @@
+import React from 'react';
+import { View, StyleSheet, Dimensions, Image, Text } from 'react-native';
+import Animated, {
+  useSharedValue,
+  useAnimatedStyle,
+  useAnimatedScrollHandler,
+  useAnimatedRef,
+  scrollTo,
+  useAnimatedGestureHandler,
+  Extrapolate,
+  interpolate,
+} from 'react-native-reanimated';
+import { PanGestureHandler } from 'react-native-gesture-handler';
+
+let data = [
+  { artist: 'Nirvana', song: 'Smells Like Teen Spirit' },
+  { artist: 'John Lennon', song: 'Imagine' },
+  { artist: 'U2', song: 'One' },
+  { artist: 'Michael Jackson', song: 'Billie Jean' },
+  { artist: 'Queen', song: 'Bohemian Rhapsody' },
+  { artist: 'The Beatles', song: 'Hey Jude' },
+  { artist: 'Bob Dylan', song: 'Like A Rolling Stone' },
+  { artist: 'Rolling Stones', song: "I Can't Get No Satisfaction" },
+  { artist: 'Sex Pistols', song: 'God Save The Queen' },
+  { artist: "Guns N' Roses", song: "Sweet Child O'Mine" },
+  { artist: 'The Clash', song: 'London Calling' },
+  { artist: 'The Kinks', song: 'Waterloo Sunset' },
+  { artist: 'The Eagles', song: 'Hotel California' },
+  { artist: 'Elton John', song: 'Your Song' },
+  { artist: 'Led Zeppelin', song: 'Stairway To Heaven' },
+  { artist: 'Chubby Checker', song: 'The Twist' },
+  { artist: 'Oasis', song: 'Live Forever' },
+  { artist: 'Whitney Houston', song: 'I Will Always Love You' },
+  { artist: 'David Bowie', song: 'Life On Mars?' },
+  { artist: 'Elvis Presley', song: 'Heartbreak Hotel' },
+  { artist: 'Judy Garland', song: 'Over The Rainbow' },
+  { artist: 'Marvin Gaye', song: "What's Goin' On" },
+  { artist: 'Bruce Springsteen', song: 'Born To Run' },
+  { artist: 'The Ronettes', song: 'Be My Baby' },
+  { artist: 'Radiohead', song: 'Creep' },
+  { artist: 'Simon & Garfunkel', song: 'Bridge Over Troubled Water' },
+  { artist: 'Aretha Franklin', song: 'Respect' },
+  { artist: 'Sly And The Family Stone', song: 'Family Affair' },
+  { artist: 'ABBA', song: 'Dancing Queen' },
+  { artist: 'The Beach Boys', song: 'Good Vibrations' },
+  { artist: 'Jimi Hendrix', song: 'Purple Haze' },
+  { artist: 'The Beatles', song: 'Yesterday' },
+  { artist: 'Chuck Berry', song: 'Jonny B Good' },
+  { artist: 'Bob Marley', song: 'No Woman No Cry' },
+  { artist: 'Jeff Buckley', song: 'Hallelujah' },
+  { artist: 'The Police', song: 'Every Breath You Take' },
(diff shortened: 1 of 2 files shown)


Example/src/AnimatedScrollExample.js
```diff
@@ -0,0 +1,279 @@
...
+    for (let i = 0; i < data.length; ++i) {
+      const distance = Math.abs(i * itemTotalSize - offset);
+      if (minDistance === undefined) {
+        minDistance = distance;
+      } else {
+        if (distance < minDistance) {
+          minDistance = distance;
+          minDistanceIndex = i;
+        }
+      }
```
jakub-gonet, line 106: `if (minDistance === undefined)` is redundant, the default value for `minDistanceIndex` is zero and we go into that branch only once, in the first iteration.

I'd go with something like:
```js
let minDistance;
let minDistanceIndex = 0;
data.forEach((_, i) => {
	const distance = Math.abs(i * itemTotalSize - offset);
	if (minDistance === undefined || distance < minDistance) {
	      minDistance = distance;
	      minDistanceIndex = i;
	}
});
```
or even:
```js
let minDistance = data[0];
let minDistanceIndex = 0;
data.forEach((_, i) => {
	const distance = Math.abs(i * itemTotalSize - offset);
	if (distance < minDistance) {
	      minDistance = distance;
	      minDistanceIndex = i;
	}
});
```
since we know that data is static and contains at least one element

Example/src/AnimatedScrollExample.js
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    scrollTo(animatedRef, minDistanceIndex * itemTotalSize, 0, true);
+  };
+
+  const scrollHandler = useAnimatedScrollHandler({
+    onScroll: (e, ctx) => {
+      position.value = e.contentOffset.x;
+    },
```
jakub-gonet, line 115: Nitpick: we don't use `ctx` here so it should read `_ctx`

Example/src/AnimatedScrollExample.js
```diff
@@ -0,0 +1,279 @@
...
+        changeVector.x * toCenterV.y - changeVector.y * toCenterV.x;
+      if (crossProd === 0) {
+        return;
+      }
+      const dist = Math.sqrt(changeVector.x ** 2 + changeVector.y ** 2);
+      // up or down
+      const sign = crossProd < 0 ? -1 : 1;
+      const arr = [0, itemTotalSize * (data.length - 1)];
+      position.value = interpolate(
+        position.value + sign * dist * 5,
```
jakub-gonet, line 155: A magic number: `5`.

Example/src/AnimatedScrollExample.js
```diff
@@ -0,0 +1,279 @@
...
+      const toCenterV = {
+        x: BIG_BALL_SIZE / 2 - lastPoz.x,
+        y: BIG_BALL_SIZE / 2 - lastPoz.y,
+      };
+      const crossProd =
+        changeVector.x * toCenterV.y - changeVector.y * toCenterV.x;
+      if (crossProd === 0) {
+        return;
+      }
+      const dist = Math.sqrt(changeVector.x ** 2 + changeVector.y ** 2);
```
jakub-gonet, line 150: Browsing `Math` module docs I found [`Math.hypot()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot), maybe it'll be useful here?
```js
const dist = Math.hypot(changeVector.x, changeVector.y);
```

Example/src/AnimatedScrollExample.js
```diff
@@ -0,0 +1,279 @@
...
+  { artist: 'The Righteous Brothers', song: "You've Lost That Lovin' Feeling" },
+  { artist: 'The Who', song: 'My Generation' },
+  { artist: 'Martha Reeves and the Vandellas', song: 'Dancing In The Street' },
+  { artist: 'Prince', song: 'When Doves Cry' },
+  { artist: 'Sam Cooke', song: 'A Change Is Gonna Come' },
+  { artist: 'Ike and Tina Turner', song: 'River Deep Mountain High' },
+  { artist: 'The Emotions', song: 'Best Of My Love' },
+];
+
+data = data.slice(0, 10);
```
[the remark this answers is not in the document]
karol-bisztyga: fixed cut songs, remove link, refactor

Example/src/AnimatedScrollExample.js
```diff
@@ -0,0 +1,279 @@
...
+const IPOD_MARGIN = 20;
+const SCREEN_WIDTH =
+  Dimensions.get('window').width - IPOD_MARGIN * 2 - SCROLL_MARGIN * 2;
+const BIG_BALL_SIZE = 200;
+const BIG_BALL_MARGIN = 0;
+const SMALL_BALL_SIZE = 50;
+const INNER_BALL_SIZE =
+  BIG_BALL_SIZE - SMALL_BALL_SIZE * 2 - BIG_BALL_MARGIN * 2;
+const DEFAULT_COVER_URI =
+  'https://e7.pngegg.com/pngimages/950/513/png-clipart-eighth-note-musical-note-stem-notes-music-download-graphic-arts.png';
```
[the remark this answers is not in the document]
karol-bisztyga: fixed e7524fc

Example/src/AnimatedScrollExample.js
karol-bisztyga: fixed e7524fc
btw I implemented the first solution you posted, the second one is invalid, here `let minDistance = data[0];` you are assigning an object to `minDistance` - distance should be calculated.

I also left the `for` loop as it is, `forEach` doesn't really change anything(maybe readability for some people but it's not crucial).

Example/src/AnimatedScrollExample.js
karol-bisztyga: actually, I think this is a matter of a coding convention, in some other files in our repo I spotted just `_` but as it is the last argument in that call we can equally just skip it there what's done here: 6578cd1

Example/src/AnimatedScrollExample.js
```diff
@@ -0,0 +1,279 @@
...
+const ITEM_SIZE = {
+  size: 250,
+  margin: 70,
+};
+const SCROLL_MARGIN = 20;
+const IPOD_MARGIN = 20;
+const SCREEN_WIDTH =
+  Dimensions.get('window').width - IPOD_MARGIN * 2 - SCROLL_MARGIN * 2;
+const BIG_BALL_SIZE = 200;
+const BIG_BALL_MARGIN = 0;
```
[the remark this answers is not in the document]
karol-bisztyga: fixed 375f2f7

Example/src/AnimatedScrollExample.js
jakub-gonet: Ah, you're right. Sure, `forEach` is looking more declarative but `for` loop will do equally good.

Example/src/AnimatedScrollExample.js
karol-bisztyga: implemented b6cd607

jakub-gonet (approved):

PR #1660, opened 2021-01-25
Shared value rerender behavior

### LINKED ISSUE #1655: Update useSharedValue and others not to reinitialize on re-render
This issue is about making a change in `useSharedValue` and `useEvent` to avoid reinitializing the value when rerender with a different init value happens.

## Description

The logic responsible for overriding shared value has been added in #1042 as a mean of handling fast reloads. The ideas seemed to be that when fast reload happens, we may provide a different init value for the hook and hence the shared value should also update. However, the solution which relied on checking whether the init value is different than the previously used init value also triggers when component is rerendered with a different init. In such a case we'd reset shared value in many cases causing unexpected updates or stopping the currently running animation.

This problem has been reported in #1293 and addressed in #1351 by adding an additional parameter that can be used to prevent "rebuild". However, this does not seem like a good way of resolving the problem. It seem like the default behavior is actually to never rebuild unless we have a refresh. This is exactly how hook like `useState` works. It appears to differentiate between cases when it needs to update because of refresh and prevents state resets when new init value is provided but there was no refresh.

I think the best way to go with this is bascially rely on `useState` mechanism when creating shared value. This should give us exact same behavior to `useState`. Here is one way `useSharedValue` can be implemented:

```js
export function useSharedValue(init) {
  return useState(() => makeMutable(init))[0];
}
```

(I haven't tested this implementation but you get the idea)

## Steps To Reproduce

1. Make component that takes some prop and use that prop as init for useSharedValue 
2. Make that component re-render with different values provided to the prop

### Expected behavior

The shared value should never update after the first time it is initialized. It should update however if we change the init value in the source code and perform fast refresh.

### Actual behavior

See that the shared value updates matching the newly provided init.

Don't have a snack example for this. I'm sorry dear issue checker bot.

## Package versions

- React: 16
- React Native: 0.63
- React Native Reanimated: 2.0.0-rc.2
- NodeJS: idk

karol-bisztyga:
## Description

This is yet one more change to the logic of the `useSharedValue` hook and it comes as a modification of #1042. We decided to bring the rerender behavior of the `useSharedValue` closer to React's `useState` - every time a rerender happens(a state is changed), a shared value is not being rebuilt and it keeps the same value.
 
Fixes #1655 

#### Note

With that change, when having something like this `useSharedValue(x)` and changing the code that the hook receives a different value `useSharedValue(y)` it's not going to work with a fast refresh. This shared value will not have an updated value. Still, we decided to implement it that way as it's more important how it acts in the release mode than in the debug mode.

## Example code

```
import { useSharedValue } from 'react-native-reanimated';

  const [state, setState] = useState(10);
  const [stateCopy, setStateCopy] = useState(state);
  const sv = useSharedValue(state);

  setImmediate(() => {
    console.log(
      'render',
      state,
      sv.value,
      stateCopy
    );
  })

  return (
    <View>
      <Button
        title="set state"
        onPress={() => {
          setState(state + 1);
        }}
      />
    </View>
  );
}
```

## Checklist

- [x] Updated TS types
- [x] Added TS types tests
- [x] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -19,19 +19,16 @@ import { Platform } from 'react-native';
 export function useSharedValue(init) {
   const ref = useRef(null);
   if (ref.current === null) {
-    ref.current = {
-      mutable: makeMutable(init),
-      last: init,
-    };
+    ref.current = makeMutable(init);
   }
 
   useEffect(() => {
     return () => {
-      cancelAnimation(ref.current.mutable);
+      cancelAnimation(ref.current);
     };
   }, []);
 
-  return ref.current.mutable;
+  return ref.current;
 }
 
 export function useEvent(handler, eventNames = [], rebuild = false) {
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -16,16 +16,13 @@ import { getTag } from './NativeMethods';
 import NativeReanimated from './NativeReanimated';
 import { Platform } from 'react-native';
 
-export function useSharedValue(init, shouldRebuild = true) {
+export function useSharedValue(init) {
   const ref = useRef(null);
   if (ref.current === null) {
     ref.current = {
       mutable: makeMutable(init),
       last: init,
     };
-  } else if (init !== ref.current.last && shouldRebuild) {
-    ref.current.last = init;
-    ref.current.mutable.value = init;
   }
 
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/reanimated2/Hooks.js
```diff
@@ -16,16 +16,13 @@ import { getTag } from './NativeMethods';
 import NativeReanimated from './NativeReanimated';
 import { Platform } from 'react-native';
 
-export function useSharedValue(init, shouldRebuild = true) {
+export function useSharedValue(init) {
   const ref = useRef(null);
   if (ref.current === null) {
     ref.current = {
       mutable: makeMutable(init),
       last: init,
```
kmagiera, line 24: I don't think `last` is needed. We can perhaps just do: `ref.cuurent = makeMutable(init)`

karol-bisztyga: fixed a4e7f6e

PR #1667, opened 2021-01-27
Support RN64 for android

piaskowyk:
## Description
Added support for React Native v0.64.0.-rc.2 for android devices. This PR is based on: https://github.com/osdnk/react-native-reanimated/tree/64 - thanks @osdnk 🙌

I compiled with `android-ndk-r19c`

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1680


jakubdybczak (approved):

kaszperro (approved):
LGTM

tomekzaw:
Congratz

tomekzaw:
Good job!

BluValor (approved):

android/build.gradle
```diff
@@ -431,7 +431,9 @@ task cleanNdkLibs(type: Exec) {
 task packageNdkLibs(dependsOn: buildReanimated, type: Copy) {
     from("$buildDir/reanimated-ndk/all")
     include("**/libreanimated.so")
-    include("**/libturbomodulejsijni.so")
+    if(System.getenv('REACT_NATIVE_TARGET_VERSION').toInteger() < 64) {
```
Szymon20000, line 434: You can get RN version from `reactProperties ` as we already do with Folly version. `VERSION_NAME` property is what you need here. I think the environment variable is problematic because we have it only in script.

createNPMPackage.sh
```diff
@@ -7,8 +7,24 @@ ROOT=$(pwd)
 # unset CI so all archs are built
 unset CI
 
+rm -rf ./android/src/main/jniLibs
```
Szymon20000, line 10: I think it's time to make the script more generic. Can you make an array with RN versions and just iterate over them?

Szymon20000 (changes requested):

Szymon20000:
Have you checked if RN 63 and 62 are still working with this pr?

piaskowyk:
Yes, I tested this PR with RN 62, 63, 64, and works without any problems.

createNPMPackage.sh
piaskowyk: It's a good idea 👍 I'll do it.

createNPMPackage.sh
```diff
@@ -4,59 +4,34 @@ set -x
...
-gradle :assembleDebug
-cd $ROOT
-
-rm -rf android-npm/react-native-reanimated-63.aar
-cp android/build/outputs/aar/*.aar android-npm/react-native-reanimated-63.aar
+versions=("0.64.0-rc.2" "0.63.3" "0.62.2 --dev")
+version_name=("64" "63" "62")
 
-# PART II (clean)
+for index in {0..2}
```
jakub-gonet, line 12:
```bash
length="${#versions[@]}"

for ((i=0; i < $length; i++ )); do echo $i; done
```

I think explicitly setting length and using for loop is more readable.

jakubowiczish (approved):

android/src/main/JNI/Android.mk
```diff
@@ -37,8 +37,10 @@ ifeq ($(NATIVE_DEBUG), true)
     cmd-strip :=
 endif
 
-LOCAL_MODULE := turbomodulejsijni
-include $(BUILD_SHARED_LIBRARY)
+ifeq ($(shell test $(REACT_NATIVE_TARGET_VERSION) -lt 64; echo $$?),0)
```
Szymon20000, line 40: Do not use shell here. The variable is already available.

android/src/main/JNI/Application.mk
```diff
@@ -5,7 +5,11 @@ APP_ABI := all
 endif
 
 APP_BUILD_SCRIPT := Android.mk
-APP_PLATFORM := android-18
+ifeq ($(shell test $(REACT_NATIVE_TARGET_VERSION) -lt 64; echo $$?),0)
```
Szymon20000, line 8: Do not use shell here.

android/src/main/JNI/react/Android.mk
```diff
@@ -71,7 +71,12 @@ $(call import-module,jsi)
 $(call import-module,jsiexecutor)
 $(call import-module,callinvoker)
 $(call import-module,hermes)
-
+ifeq ($(shell test $(REACT_NATIVE_TARGET_VERSION) -ge 64; echo $$?),0)
```
Szymon20000, line 74: Do not use shell here.

Szymon20000 (changes requested):

android/src/main/JNI/Android.mk
jakubdybczak: I'm not aware of any boolean env variable defining if we should make old or new style of build. Are you suggesting creating one?

Szymon20000 (approved):

ajpaulingalls:
Any ETA on a new release with support for RN64?

karol-bisztyga:
next few days

PR #1683, opened 2021-02-02
Improve of decay animation style

piaskowyk:
## Description

I hanged a curve style to more [react like](https://github.com/facebook/react-native/blob/6e6443afd04a847ef23fb6254a84e48c70b45896/Libraries/Animated/animations/DecayAnimation.js#L94-L97) and add velocity factor to modify the unit of velocity. The default unit of velocity in decay is pixel per second but this is problematic when someone wants to animate value not related to pixels for example opacity `[0, 1]` or progress bar state `[0, 1]`. In this case, someone can use `velocityFactor` property with value `< 1` to modify the velocity of change to more matched for the required domain.

<details>
<summary>Code</summary>

```js
import Animated, {
  useSharedValue,
  withTiming,
  withDecay,
  useAnimatedRef,
  useAnimatedGestureHandler,
  measure,
  useAnimatedStyle,
import {
  PanGestureHandler,
} from 'react-native-gesture-handler';
import React from 'react';

function ProgressBarControl() {
  const progress = useSharedValue(0);
  const boxRef = useAnimatedRef();

  const reset = () => {
    progress.value = withTiming(0.3);
  };

  const barStyle = useAnimatedStyle(() => {
    return {
      width: `${progress.value * 100}%`,
    };
  });

    onStart: (_, ctx) => {
      ctx.startingProgress = progress.value;
    },
    onActive: (e, ctx) => {
      const { width } = measure(boxRef);
      const progressTranslation = e.translationX / width;
      progress.value = ctx.startingProgress + progressTranslation;
    },
    onEnd: (e) => {
      progress.value = withDecay({
        velocity: e.velocityX,
        clamp: [0,1],
        velocityFactor: 0.001, // try it without velocityFactor also
        deceleration: 0.998
      });
    },
  });

  return (
    <>
      <View ref={boxRef} style={styles.barContainer}>
          <Animated.View style={[styles.bar, barStyle]} />
        </PanGestureHandler>
      </View>
      <Button onPress={reset} title="start over" />
    </>
  );
}

export default function Main(props) {
  return (
      <ProgressBarControl />
    </View>
  );
}

  container: {
    marginTop: 150,
    marginHorizontal: 20,
    height: 120,
  },
  barContainer: {
    height: 50,
    borderColor: '#001A72',
    borderWidth: 1,
    borderRadius: 5,
    overflow: 'hidden',
  },
  bar: {
    backgroundColor: '#001A72',
    height: 50,
  },
});

```

</details>

- [x] Updated TS types
- [x] Updated documentation
- [x] Ensured that CI passes


src/reanimated2/animations.js
```diff
@@ -354,55 +354,83 @@ export function withDecay(userConfig, callback) {
...
-        }
-
+      if (Math.abs(v) < VELOCITY_EPS) {
         return true;
       }
     }
 
+    function validConfig() {
+      if (config.clamp) {
+        if (Array.isArray(config.clamp)) {
```
jakub-gonet, line 406: `if(config.clamp && Array.isArray(config.clamp))` or just `Array.isArray(config.clamp)` since it can handle falsy values without problems.

src/reanimated2/animations.js
```diff
@@ -354,55 +354,83 @@ export function withDecay(userConfig, callback) {
...
+          if (config.clamp.length !== 2) {
+            console.error(
+              '[reanimated] clampy array must contain 2 items but is given ' +
```
jakub-gonet, line 409: AFAIK we don't use `[reanimated]` tag anywhere else.
Also, I think it should be `clamp array`. You can use string interpolation here as well.

src/reanimated2/animations.js
```diff
@@ -354,55 +354,83 @@ export function withDecay(userConfig, callback) {
...
-        if (toValueIsReached !== null) {
-          animation.current = toValueIsReached;
-        }
-
+      if (Math.abs(v) < VELOCITY_EPS) {
         return true;
       }
     }
 
+    function validConfig() {
```
jakub-gonet, line 404: TBH I'd rewrite this like so:

```js
if (!Array.isArray(config.clamp)) {
	throw new Error("not an array");
}
if (config.clamp.length !== 2) {
	throw new Error("invalid number of arguments")
}
if (config.velocityFactor <= 0) {
	throw new Error("invalid velocity");
}
```
Since we actually want to developer to fix their code instead of showing errors in the console.
Inverting conditions makes it a little more readable IMHO.

jakub-gonet: Also, this should be named `validateConfig()` instead.

src/reanimated2/animations.js
```diff
@@ -354,55 +354,83 @@ export function withDecay(userConfig, callback) {
     'worklet';
     const config = {
       deceleration: 0.998,
+      velocityFactor: 1,
```
jakub-gonet, line 357: Remember to add TS types and add docs for it if necessary.

src/reanimated2/animations.js
piaskowyk: Not exactly because `clamp` is an optional argument and just `Array.isArray(config.clamp)` changes the meaning of `else` condition.

src/reanimated2/animations.js
piaskowyk: 👍

src/reanimated2/animations.js
piaskowyk: Yes, I agree that exceptions can be more readable but there is a problem. This is `worklet` and if I throw an exception in `worklet` function the user receives RedBox with `"Javascript worklet error in file reanimated2/core.js"` instead of my error message. IMO this information is useless and confusing, so that's why I use `console.error` instead of an exception.

We should resolve the problem with exceptions but this is the case for the next PR.

src/reanimated2/animations.js
jakub-gonet: The first version should work then: `if(config.clamp && Array.isArray(config.clamp))`

src/reanimated2/animations.js
jakub-gonet: Ah, I see. Makes sense to leave `console.error` in that case.

Szymon20000 (approved):

PR #1688, opened 2021-02-04
Bump rn to 0.64.0-rc.2 in Example

Szymon20000:
## Description

Bump  RN in Example to 0.64.0-rc.2.

## Changes

I went through https://react-native-community.github.io/upgrade-helper/?from=0.63.4&to=0.64.0-rc.2

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/Example/android/gradle.properties
+++ b/Example/android/gradle.properties
@@ -20,4 +20,3 @@ android.useAndroidX=true
 android.enableJetifier=true
 # Version of flipper SDK to use with React Native
 FLIPPER_VERSION=0.37.0
-ANDROID_NDK_VERSION=20.1.5948944
--- a/Example/android/gradle.properties
+++ b/Example/android/gradle.properties
@@ -20,3 +20,4 @@ android.useAndroidX=true
 android.enableJetifier=true
 # Version of flipper SDK to use with React Native
 FLIPPER_VERSION=0.37.0
+ANDROID_NDK_VERSION=20.1.5948944
--- a/Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
+++ b/Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
@@ -306,7 +306,7 @@
 			);
 			inputPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-ReanimatedExampleTests/Pods-ReanimatedExampleTests-resources.sh",
-				"${PODS_CONFIGURATION_BUILD_DIR}/React-Core.common-ARTHeaders/AccessibilityResources.bundle",
+				"${PODS_CONFIGURATION_BUILD_DIR}/React-Core/AccessibilityResources.bundle",
 			);
 			name = "[CP] Copy Pods Resources";
 			outputPaths = (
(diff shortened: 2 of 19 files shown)


Example/android/gradle.properties
```diff
@@ -19,4 +19,5 @@
 android.useAndroidX=true
 android.enableJetifier=true
 # Version of flipper SDK to use with React Native
-FLIPPER_VERSION=0.37.0
\ No newline at end of file
+FLIPPER_VERSION=0.37.0
+ANDROID_NDK_VERSION=20.1.5948944
```
piaskowyk, line 23: We have the same configuration variable in two places, here and [here](https://github.com/software-mansion/react-native-reanimated/blob/f7b140a9c379262d58f7051aef3bde658a987879/android/gradle.properties#L22). Is it possible to set NDK version in one place?

Maybe we should add to docs info about the required NDK version, what do you think?

Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
```diff
@@ -583,6 +583,7 @@
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 ";
```
piaskowyk, line 586: I'm not sure it changes something 😄  but do you sure that it should be  `"arm64 "` instead of `"arm64"` (without space)

Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
```diff
@@ -625,6 +626,7 @@
 				COPY_PHASE_STRIP = YES;
 				ENABLE_NS_ASSERTIONS = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 ";
```
piaskowyk, line 629: like above

piaskowyk (approved):

RNReanimated.podspec
```diff
@@ -7,10 +7,17 @@ reactVersion = '0.0.0'
...
-  reactVersion = '0.63.2'
+  reactVersion = '0.64.0'
 end
 
 rnVersion = reactVersion.split('.')[1]
+
+folly_prefix = ""
+if rnVersion.to_i >= 64
+  folly_prefix = "RCT-"
+end
```
jakub-gonet, line 18: I think it's slightly more readable by using ternary:
```ruby
folly_prefix = rnVersion.to_i >= 64 ? "RCT-" : ""
```

Also, a small comment on why this is needed would be nice to have.

Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
Szymon20000: It's not sth I did personally. I think it's a result of `pod install` cmd.

jakub-gonet (commented):
Have you tested if the web version still works?

Example/android/gradle.properties
Szymon20000: It's not used right now. I will remove it.

jakub-gonet:
CI is failing.

karol-bisztyga (approved):

PR #1694, opened 2021-02-05
Interpolation: replace spread operator with Object.assign

karol-bisztyga:
## Description

In interpolation code, we want to use `Object.assign` instead of a spread operator in worklets in order to avoid errors.


Szymon20000 (approved):

jakub-gonet (approved):
Looks good, remember to write a small description and change the title.

PR #1695, opened 2021-02-05
Bump version to rc.3

karol-bisztyga:
## Description

Bump version to rc.3


jakub-gonet (approved):
Do we need to bump docs too?

PR #1697, opened 2021-02-05
Make properties in animatedProps optional

mrousavy:
## Description

Makes `AnimateProps` partial, so that no required props have to be set twice.

## Changes

* Updated `AnimateProps<P>` to replace `P` with `Partial<P>`

1. Try to create an Animated component which has required props (e.g. `device` on camera)
2. Pass animatedProps 

Notice how animatedProps requires you to specify `device` since it is required. This PR fixes this so everything is optional.

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


jakub-gonet:
This was discussed [here](https://github.com/software-mansion/react-native-reanimated/pull/1590#issuecomment-760892280). Could you please write a small example with before / after code for context?

mrousavy:
@jakub-gonet 

## Camera Type Decls

```tsx
type CameraDevice = {
  // ...
}
export type CameraProps = {
  device: CameraDevice;
  zoom?: number;
}

export class Camera extends React.PureComponent<CameraProps, CameraState> {
  // ...
}

const ReanimatedCamera = Reanimated.createAnimatedComponent(Camera);
Reanimated.addWhitelistedNativeProps({
  zoom: true,
});
```

## Before

```tsx
function App() {
  const zoom = useSharedValue(0);

  // it's expensive to get the camera device
  const device = useMemo(() => null, []);

  const cameraAnimatedProps = useAnimatedProps<CameraProps>(
    () => ({
// Type error because it also expects the required `device` prop:
// Type '{ zoom: number; }' is not assignable to type 'CameraProps'.
//  Type '{ zoom: number; }' is not assignable to type 'Omit<CameraFormatProps, "preset" | "fps" | "hdr" | "lowLightBoost" | "colorSpace" | "format"> & { preset?: never; fps?: number; hdr?: boolean; lowLightBoost?: boolean; colorSpace?: ColorSpace; format?: Readonly<...>; } & ... 5 more ... & ViewProps'.
//    Type '{ zoom: number; }' is missing the following properties from type '{ scannableCodes: CodeType[]; onCodeScanned: (codes: Readonly<{ code?: string; type: CodeType; bounds: { minX: number; minY: number; maxX: number; maxY: number; width: number; height: number; }; }>[]) => void; }': scannableCodes, onCodeScannedts(2322)
      zoom: zoom.value
    }),
    [zoom]
  );

  return (
    <ReanimatedCamera
      style={StyleSheet.absoluteFill}
      device={device}
      animatedProps={cameraAnimatedProps} />
  );
}
```

## After

```tsx
function App() {
  const zoom = useSharedValue(0);

  // it's expensive to get the camera device
  const device = useMemo(() => null, []);

  const cameraAnimatedProps = useAnimatedProps<CameraProps>(
    () => ({
// No Type error because AnimateProps<P> now makes everything optional (Partial<T>).
      zoom: zoom.value
    }),
    [zoom]
  );

  return (
    <ReanimatedCamera
      style={StyleSheet.absoluteFill}
      device={device}
      animatedProps={cameraAnimatedProps} />
  );
}
```

mrousavy:
The problem here is that `useAnimatedProps` simply mirrors the props from the Camera. This means, it also expects that I set the `device`, `hdr`, `fps` and all those other props I can set on the `<Camera` component - which I shouldn't. I only want to animate a single prop, `zoom`.

It would probably be easier to pick out the Props from the component instead, so e.g.:

```ts
const p = useAnimatedProps<CameraProps>(() => ({ ... }), []);
```

becomes:

```ts
const p = useAnimatedProps<Camera>(() => ({ ... }), []);
```

since a component doesn't always export it's Props type/interface. wdyt @jakub-gonet @wcandillon ?

jakub-gonet:
Thanks for the great example. I started thinking if we can detect if a property is set in the props or in the animated props, so we can get the best of two worlds – if the prop is needed and we want it to be animated then we should specify it only in the `animatedProps` (now types forces to write it in two places). I'm didn't use Reanimated 2 that much, can we do that in the JS version?

For now, using Partial is good enough.

jakub-gonet (approved):
If you can, please add some tests similar to the example you've provided. It also needs merge conflicts to be resolved. Other than that, looks great, thanks!

wcandillon:
Would it be possible to update the test file as well?

mrousavy:
Sorry for the delay, added tests. Should work fine now

🏓 @wcandillon @jakub-gonet

mrousavy:
@jakub-gonet could you merge if everything looks fine to you?

wcandillon:
I like it

jakub-gonet (approved):
Sure, sorry for the delay.

PR #1704, opened 2021-02-09
Install global bindings for executor factory

mrousavy:
## Description

After upgrading to RN 0.64 and enabling Hermes, I noticed that my console.logs were gone.

That's the only thing I saw in metro:

After a long journey of debugging (see https://github.com/facebook/metro/issues/635#issuecomment-775939213) I finally found the weird side-effect our custom executor factory creation was having.

This is our `jsExecutorFactoryForBridge` func's return:

```objc
  return std::make_unique<ExecutorFactory>([weakSelf, bridge](facebook::jsi::Runtime &runtime) {
    if (!bridge) {
      return;
    }
    __typeof(self) strongSelf = weakSelf;
    if (strongSelf) {
      #if RNVERSION < 63
        auto callInvoker = std::make_shared<react::BridgeJSCallInvoker>(bridge.reactInstance);
        auto reanimatedModule = reanimated::createReanimatedModule(callInvoker);
      #else
        auto reanimatedModule = reanimated::createReanimatedModule(bridge.jsCallInvoker);
      #endif
      runtime.global().setProperty(runtime,
                                   jsi::PropNameID::forAscii(runtime, "__reanimatedModuleProxy"),
                                   jsi::Object::createFromHostObject(runtime, reanimatedModule)
      );
    }
  });
```

This is from rn-tester:

```objc
  __weak __typeof(self) weakSelf = self;
#if RCT_USE_HERMES
  return std::make_unique<facebook::react::HermesExecutorFactory>(
#else
  return std::make_unique<facebook::react::JSCExecutorFactory>(
#endif
      facebook::react::RCTJSIExecutorRuntimeInstaller([weakSelf, bridge](facebook::jsi::Runtime &runtime) {
        if (!bridge) {
          return;
        }
        __typeof(self) strongSelf = weakSelf;
        if (strongSelf) {
          facebook::react::RuntimeExecutor syncRuntimeExecutor =
              [&](std::function<void(facebook::jsi::Runtime & runtime_)> &&callback) { callback(runtime); };
          [strongSelf->_turboModuleManager installJSBindingWithRuntimeExecutor:syncRuntimeExecutor];
        }
      }));
```

As you can see, we're missing `RCTJSIExecutorRuntimeInstaller` which is responsible for binding nativePerformanceNow, console, etc.

## Changes

* Correctly install bindings using `RCTJSIExecutorRuntimeInstaller` for the JSI Executor.

### Before

no logs in console (no `global.console`, no `global.nativePerformanceNow`)

### After

[...]

PR #1707, opened 2021-02-10
Add console.debug to worklets

mrousavy:
## Description

Until now, the console object in worklets only allowed pseudo-sync calls with `log`, `info`, `warn` and `error`. I was using `debug` a lot, so that was always a bit weird to encapsulate it in `runOnJS`

## Changes

- Adds `console.debug` to worklet console

```jsx
const x = useAnimatedStyle(() => {
  console.debug('works')
  return { opacity: opacity.value };
}, []);
```

### DIFF
--- a/src/reanimated2/core.js
+++ b/src/reanimated2/core.js
@@ -262,6 +262,7 @@ const capturableConsole = console;
 runOnUI(() => {
   'worklet';
   const console = {
+    debug: runOnJS(capturableConsole.debug),
     log: runOnJS(capturableConsole.log),
     warn: runOnJS(capturableConsole.warn),
     error: runOnJS(capturableConsole.error),


mrousavy:
I believe we _should_ be able to entirely drop this JS logic for capturing, redirecting and setting the console to global - we _should_ be able to do it with the func from React I used in my other PR: https://github.com/software-mansion/react-native-reanimated/pull/1704

That way the console gets set at the same time the runtime is created, and doesn't need to dispatch calls to JS every time you log something. (don't quote me on the last part tho)

@Szymon20000 @karol-bisztyga

piaskowyk:
> I believe we _should_ be able to entirely drop this JS logic for capturing, redirecting and setting the console to global - we _should_ be able to do it with the func from React I used in my other PR: #1704
> 
> That way the console gets set at the same time the runtime is created, and doesn't need to dispatch calls to JS every time you log something. (don't quote me on the last part tho)
> 
Good point, I'll try it.

piaskowyk (approved):

PR #1719, opened 2021-02-14
Typescript: createAnimatedComponent() add missing style prop type

### LINKED ISSUE #1717: Typescript: style prop added by createAnimatedComponent() is flagged
## Description

Typescript is flagging the `style` prop on eg `<AnimatedCircle>` as nonexistent when in fact it works at runtime (and wasn't flagged by Reanimated v1 types)

## Steps To Reproduce

```
import {Circle} from 'react-native-svg'

const AnimatedCircle = Animated.createAnimatedComponent(Circle)

type AnimatedCircleStyleProp = (typeof AnimatedCircle)['style']
```

The above gets flagged by Typescript as `Property 'style' does not exist on type 'ComponentType<AnimateProps<CircleProps>>'.`

Looking at the types for Reanimated, this is because `AnimateProps` only includes `style` prop if the wrapped component's type includes a `style` prop. And I guess `<Circle>` doesn't actually have a `style` prop. But clearly Reanimated's `createAnimatedComponent()` wrapper adds support for a `style` prop since passing an animated-value `style` prop still works at runtime in Reanimated v2. And I'm not sure what the types were for Reanimated v1 but it did not flag this as a type error

### Expected behavior

No Typescript error, `style` prop on `<AnimatedCircle>` should be typed appropriately (`StyleProp<AnimateStyle<...>>`?)

### Actual behavior

Typescript flags a type error

## Package versions

- React: 16.13.1
- React Native: 0.63.3
- React Native Reanimated: 2.0.0-rc.3
- NodeJS: 14.15.3
- Typescript: 3.9.7

helixbass:
## Description

Fixes #1717 

## Changes

Updates Typescript type declaration for `createAnimatedComponent()` to reflect the fact that the returned wrapper component always exposes a `style` prop even if the wrapped component doesn't have a `style` prop

Added test cases

## Checklist

- [x] Updated TS types
- [x] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -244,10 +244,28 @@ declare module 'react-native-reanimated' {
 
     export function createAnimatedComponent<P extends object>(
       component: ComponentClass<P>
-    ): ComponentClass<AnimateProps<P>>;
+    ): ComponentClass<
+      AnimateProps<
+        P &
+          ('style' extends keyof P
+            ? {}
+            : {
+                style?: StyleProp<ViewStyle>
+              })
+      >
+    >;
     export function createAnimatedComponent<P extends object>(
       component: FunctionComponent<P>
-    ): FunctionComponent<AnimateProps<P>>;
+    ): FunctionComponent<
+      AnimateProps<
+        P &
+          ('style' extends keyof P
+            ? {}
+            : {
+                style?: StyleProp<ViewStyle>
+              })
+      >
+    >;
 
     // classes
     export {
--- a/react-native-reanimated-tests.tsx
+++ b/react-native-reanimated-tests.tsx
@@ -121,6 +121,64 @@ function CreateAnimatedFlatList() {
   )
 }
 
+function CreateAnimatedComponentAddsStylePropTest() {
+  const animatedStyle = useAnimatedStyle(() => ({ width: 100 }));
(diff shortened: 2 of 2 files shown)


react-native-reanimated.d.ts
```diff
@@ -244,10 +244,28 @@ declare module 'react-native-reanimated' {
...
     export function createAnimatedComponent<P extends object>(
       component: ComponentClass<P>
-    ): ComponentClass<AnimateProps<P>>;
+    ): ComponentClass<
+      AnimateProps<
+        P &
+          ('style' extends keyof P
+            ? {}
+            : {
+                style?: StyleProp<ViewStyle>
```
helixbass, line 253: I don't actually know what Reanimated is doing under the hood but per #1717 it works to add an animated `style` prop to eg an `<AnimatedCircle>` (where `<Circle>` does not have a `style` prop). So I guessed here that the `style` prop exposed by Reanimated is compatible with `ViewStyle`

piaskowyk:
Hey, thanks for your commitment and I appreciate your work but I need to close this PR 😕  because we currently rewriting the whole reanimated to TypeScript so we won't need types defined in `react-native-reanimated.d.ts`

Issue #1703, opened 2021-02-09
Android: `scrollTo` with FlatList breaks as soon as refresh control is added


andreialecu:
## Description

Using `scrollTo` with a normal FlatList seems to work great, just until it gets a RefreshControl.

This only seems to occur on Android. On iOS everything works fine.

## Steps To Reproduce

```tsx

const AnimatedFlatList = Animated.createAnimatedComponent(FlatList)

<AnimatedFlatList
      ref={ref}
      ...
      onRefresh={() => {}} // comment out and it scrolls
      refreshing={true}    //
    />

scrollTo(ref, ...) // on android it only scrolls when refresh control is missing

```

### Expected behavior

Programmatic scrolling should work.

### Actual behavior

Nothing happens.

## Snack or minimal code example

See above.

## Package versions

- React Native: via Expo 40.0.1
- React Native Reanimated: 2.0.0-rc.0

Only occurs on Android

PedroBern:
Same issue here on `rc.0`

andreialecu:
Note that ScrollView works fine. Only FlatList is affected.

PR #1742, opened 2021-02-22
Fix Flatlist scroll with refresh

piaskowyk:
## Description

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1703

If FlatList contains `onRefresh` property then `ReactScrollView` is wrapped by `ReactSwipeRefreshLayout` who doesn't have `smoothScrollTo()` method.

ReactSwipeRefreshLayout support only vertical scroll - "[The SwipeRefreshLayout should be used whenever the user can refresh the contents of a view via a vertical swipe gesture.](https://developer.android.com/reference/androidx/swiperefreshlayout/widget/SwipeRefreshLayout)"

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NativeMethodsHelper.java
+++ b/android/src/main/java/com/swmansion/reanimated/NativeMethodsHelper.java
@@ -5,11 +5,13 @@ import android.graphics.RectF;
 import android.util.Log;
 import android.view.View;
 import android.view.ViewParent;
+import android.view.animation.AnimationUtils;
 
 import com.facebook.react.uimanager.PixelUtil;
 import com.facebook.react.uimanager.RootViewUtil;
 import com.facebook.react.views.scroll.ReactHorizontalScrollView;
 import com.facebook.react.views.scroll.ReactScrollView;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/reanimated/NativeMethodsHelper.java
```diff
@@ -64,6 +72,15 @@ public static void scrollTo(View view, double argX, double argY, boolean animate
 
   }
 
+  private static ReactScrollView findScrollView(ReactSwipeRefreshLayout view) {
+    for(int i = 0; i < view.getChildCount(); i++) {
+      if(view.getChildAt(i) instanceof ReactScrollView) {
```
Szymon20000, line 77: Is it possible to have HorizontalScrollView here?

piaskowyk: ReactSwipeRefreshLayout support only vertical scroll - "[The SwipeRefreshLayout should be used whenever the user can refresh the contents of a view via a vertical swipe gesture.](https://developer.android.com/reference/androidx/swiperefreshlayout/widget/SwipeRefreshLayout)"

Szymon20000 (approved):

PR #1757, opened 2021-02-26
Notify about renaming Easing to EasingNode [sol1]

Szymon20000:
## Description

The pr is to notify people migrating from version 1.x about renaming Easing to EasingNode.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


src/reanimated2/Easing.js
```diff
@@ -252,3 +253,32 @@ export const Easing = {
   out,
   inOut,
 };
+
+function createChecker(worklet) {
+  function checkIfReaOne() {
+    'worklet'
+    if (arguments && (!_WORKLET)) {
+      for (let i = 0; i < arguments.length; i++) {
+        const arg = arguments[i];
```
jakub-gonet, line 262: If this is transpiled then you can use `for(arg of arguments)`

src/reanimated2/Easing.js
```diff
@@ -252,3 +253,32 @@ export const Easing = {
...
+        if (arg && arg.__nodeID) {
+          throw new Error(
+            'Reanimated: Easing from V1 has been renamed to EasingNode.'
+          );
+        }
+      }
+    }
+    const res = worklet.apply(this, arguments);
```
jakub-gonet, line 270: and here `worklet(..arguments)`

src/reanimated2/Easing.js
```diff
@@ -252,3 +253,32 @@ export const Easing = {
...
+    if (res && (typeof res === 'function') && (!_WORKLET)) {
```
jakub-gonet, line 271: if `res` is undefined or null it won't have typeof "function". But maybe we want to be explicit here.

PR #1761, opened 2021-02-26
Move plugin check

piaskowyk:
## Description

Added possibility to use Reanimated2 without added plugin when someone uses only Reanimated1 API.

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -311,7 +311,21 @@ function styleUpdater(
   }
 }
 
+const checkPluginState = () => {
+  if (
+    !(() => {
+      'worklet';
+    }).__workletHash &&
+    !process.env.JEST_WORKER_ID
+  ) {
+    throw new Error(
+      "Reanimated 2 failed to create a worklet, maybe you forgot to add Reanimated's babel plugin?"
+    );
+  }
+};
+
 export function useAnimatedStyle(updater, dependencies, adapters) {
+  checkPluginState();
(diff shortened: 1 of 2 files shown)


src/reanimated2/Hooks.js
```diff
@@ -311,7 +311,21 @@ function styleUpdater(
...
+    !process.env.JEST_WORKER_ID
+  ) {
+    throw new Error(
+      "Reanimated 2 failed to create a worklet, maybe you forgot to add Reanimated's babel plugin?"
+    );
+  }
+};
+
 export function useAnimatedStyle(updater, dependencies, adapters) {
+  checkPluginState();
```
jakub-gonet, line 328: This check should be only here or in other hooks as well?

src/reanimated2/core.js
```diff
@@ -8,16 +8,11 @@ global.__reanimatedWorkletInit = function (worklet) {
...
-  !process.env.JEST_WORKER_ID
-) {
-  throw new Error(
-    "Reanimated 2 failed to create a worklet, maybe you forgot to add Reanimated's babel plugin?"
-  );
+if (!global._setGlobalConsole) {
+  // it can happen when Reanimated plugin wasn't added, but the user uses the only API from version 1
+  global._setGlobalConsole = () => {
+    // noop
+  };
```
jakub-gonet, line 15: Idea: maybe we should check things that use `_setGlobalConsole` and make sure they're handling undefined correctly instead of setting default value?

Szymon20000 (approved):

PR #1768, opened 2021-02-27
Use `"` instead of `<` for `JSIStoreValueUser.h` include

mrousavy:
## Description

`JSIStoreValueUser.h` is a local file, use `"` instead of `<` for importing user headers.

This compiled fine before since CocoaPods allows flattening system headers, but once I try to import some REA headers in another library, it fails to compile:

Screenshot 2021-02-27 at 21 38 40


jakub-gonet (approved):
LGTM.

piaskowyk (approved):

PR #1788, opened 2021-03-05
Update docs to v2 stable

piaskowyk:
## Description

New documentation version for 2.0.0


docs/src/theme/Navbar/index.js
```diff
@@ -226,7 +226,7 @@ function Navbar() {
                   [styles.hideLogoText]: isSearchBarExpanded,
                 })}>
                 {title}
-                <span className="badge badge--info">V2 alpha</span>
+                <span className="badge badge--info">V2 stable</span>
```
jakub-gonet, line 229: I think we should remove this badge, it stays when you're browsing v1 docs. This confuses people.

docs/versions.json
```diff
@@ -1,4 +1,5 @@
 [
+  "2.0.0",
   "2.0.0-rc.2",
   "2.0.0-alpha.9",
```
jakub-gonet, line 4: I think we can remove 2.0.0-alpha.7, 2.0.0-alpha.8, 2.0.0-alpha.9 docs.

jakub-gonet (approved):

PR #1790, opened 2021-03-05
Expose native worklet API

mrousavy:
## Description

This PR exposes the native "worklet" API by lifting the "manage-a-runtime" logic out of `NativeReanimatedModule` so that other native modules can implement this for their own use case. This "manage-a-runtime" logic is implementable by the `RuntimeManager` base class.

### Advantages

* Other native libraries can make use of Reanimated's workletization functionality to spawn new JS threads. Example: My camera library: [react-native-vision-camera](https://github.com/cuvent/react-native-vision-camera): Creating "frame processors" to run QR Code scanning, AI, Realtime Video Chats, etc
* Splitting the SharedValue parts from the `NativeReanimatedModule`, making everything a bit more "decoupled"

### Related

* Makes use of a few parts of #1561
* Is required for my frame processors: https://github.com/cuvent/react-native-vision-camera/pull/2

## Approach

I will be starting from [this location in my VisionCamera project](https://github.com/cuvent/react-native-vision-camera/blob/76abf0400d299dbe85065810b9b4e098d88e9fc2/ios/Frame%20Processor/FrameProcessorDelegate.mm#L52-L66), I am trying to create a worklet here.

> "Creating a worklet" = Copy the function from the default React JS Runtime to the newly created one for the frame processors, and copy over all variables from "outside".

So `ShareableValue::adapt` is my starting point, from there on I will start to remove all `NativeReanimatedModule` references and will try to decouple it as much as possible.

## Solution

The current solution works fine and nothing has changed for Reanimated. No side-effects whatsoever (afaik).

But I am not 100% happy with one thing: `RuntimeManager` has a Scheduler, ValueSetter and ErrorHandler which are all properties that are really specific to Reanimated's `NativeReanimatedModule`. Ideally other native modules shouldn't be using those 3 things, since ValueSetter is specific for mirroring styles from native to JS (afaik), ErrorHandler just displays the errors in worklets (afaik) and scheduler is responsible for dispatching from Worklet -> JS and vice versa. Ideally I want to also abstract those away with the following solution:
* `errorHandler` is never needed, the default React LogBox should always be used for any errors in worklets and should even display source frames. For any other calls, just throw an `std::exception` (or `jsi::JSError`)
* `valueSetter` is not needed at all, this is only needed for Reanimated. Would be better to remove that from `RuntimeManager` too.
* `scheduler` is needed to use `runOnJS` I believe, no idea if we really need that. It's also really hard to create an instance of the `Scheduler` on Android...

@Szymon20000 I would like your opinion on those things. I guess we can merge this PR since no side effects are introduced and it's now possible for other libraries to hook into this logic and concentrate on those 3 things in another PR. Ideally it should be really simple to create worklets, but now it requires you to create an ErrorHandler and Scheduler.

#### Edit

As for the `scheduler`, I haven't tried it yet but what happens if I do the followng:

```js
const sv = useSharedValue(0)
const frameProcessor = useFrameProcessor(() => {
  'worklet'
  sv.value = 5
}, [])
// the above frameProcessor is created in my custom jsi::Runtime and workletized with ShareableValue::adapt
```

`sv.value = 5` will be executed in my custom `jsi::Runtime` from the Frame Processor, so it will not be updated on the UI thread. Would this still work? Is this like updating the value on the JS thread?


mrousavy:
cc @Szymon20000 @karol-bisztyga @piaskowyk

mrousavy:
it compiles 🎉

mrousavy:
I've refactored the code so that other libraries can inherit [`RuntimeManager.h`](https://github.com/software-mansion/react-native-reanimated/blob/c4d3ad3974350f8bca7068883854ddac4d7bd3e0/Common/cpp/headers/SharedItems/RuntimeManager.h) which provides the following properties:

* `runtime`: An individual runtime, effectively a secondary JS thread/process, similar to web workers - or how REA does it for the UI thread
* `valueSetter`: A `ShareableValue` (workletized JS function) which is responsible for setting a value in JS (afaik used to reflect native changes in the `useShareableValue` hook, correct me if I'm wrong) (nullable!) I'm not sure if we even need to provide this for other libraries, since it's pretty REA specific, let me know what you think @ software mansion
* `errorHandler`: Handles errors. Same thing as with `valueSetter`, not sure if that should be provided - if not we have to refactor a bit more of the code that depends on the errorHandler being set (let me know what you think @ software mansion)
* `scheduler`: Schedules code on the UI thread using platform specific code, or on the JS thread using the JSCallInvoker. Same as above, not sure if we need to provide that. let me know what you think @ software mansion

> those 4 properties are just extracted from `NativeReanimatedModule.h`

While everything compiles correctly (and uses the same semantics, so everything REA related works as expected), I'm not sure what's missing here because when I call `ShareableValue::adapt` in my VisionCamera (see [this line](https://github.com/cuvent/react-native-vision-camera/pull/2/files#diff-37ee9d9e0f73b03e48d03199b3a914e88d8b2f2714ce345e05146ef1269ac172R80)), it still gives me a "BAD_ACCESS" error and crashes the app. It does successfully run through the big `if` ([these lines here](https://github.com/software-mansion/react-native-reanimated/blob/c4d3ad3974350f8bca7068883854ddac4d7bd3e0/Common/cpp/SharedItems/ShareableValue.cpp#L76-L154)), but after a few runs it crashes. I am not exactly sure where and when, but I noticed that it at least gets to [this line](https://github.com/software-mansion/react-native-reanimated/blob/c4d3ad3974350f8bca7068883854ddac4d7bd3e0/Common/cpp/SharedItems/ShareableValue.cpp#L109).

I'd appreciate any feedback from you guys, maybe you can spot an error in my code - or let me know if there's any additional work required for this to work.

cc @Szymon20000 @karol-bisztyga @piaskowyk @terrysahaidak @kmagiera

mrousavy:
Managed to get it working successfully:

https://user-images.githubusercontent.com/15199031/110769322-559cbb80-8258-11eb-9b8f-a5c9eaf18afa.mov


This means, that this PR is technically ready. I'd like you guys to answer my questions in the comment above on how we're going to proceed with this, especially with the valueSetter errorHandler and scheduler, but it _works_ now. 🔥

mrousavy:
As a side note, I noticed that you guys did a lot of 

```cpp
std::shared_value<T> someValue(new T(...));
```

instead of

```cpp
std::shared_value<T> someValue = std::make_shared<T>(...);
```

is there a specific reason for that?

Might be a micro-optimization, but `make_shared` is faster than `new` as it only has to allocate memory once (and `new` has to do 1x for T, 1x for control-block)

mrousavy:
I made a multithreading library by using the changes in this PR: https://github.com/mrousavy/react-native-multithreading

mrousavy:
I noticed that I can't simply link the Android Reanimated native code against my library because it is pre-built using an .aar. Since react-native is starting to use more and more C++ code (Rearchitecture, TurboModules), and with the release of RN 0.64 even requires the user to install the NDK ([see template's build.gradle](https://github.com/facebook/react-native/blob/3ff7e86b0ff12823e5d89a0d3a7cceb79e0b25b0/template/android/build.gradle#L9)), I think it wouldn't be too far off to not use a prebuilt .aar library for reanimated, but instead ship the .h/.cpp sources and let the user compile it themselves.
This would allow my [VisionCamera](https://github.com/cuvent/react-native-vision-camera), my [Multithreading library](https://github.com/mrousavy/react-native-multithreading) and potentially other libraries to use native Reanimated code to allow "workletization" - also it's way easier to patch something in the lib. Plus it requires less effort for you to release a new npm version. What do you guys think?

mrousavy:
Closing in favor of https://github.com/software-mansion/react-native-reanimated/pull/1861

PR #1803, opened 2021-03-07
[v1] Replace global.Expo.Constants with NativeModules check

### LINKED ISSUE #1612: AnimatedDebug is using depricated global.Expo.Constants object (SDK40)
## Description
After upgrading to expo SDK40, I started getting this warning:
```
Your project is accessing the following APIs from a deprecated global rather than a module import: Constants (expo-constants).

The global "__expo" and "Expo" objects will be removed in SDK 41. Learn more about how to fix this warning: https://expo.fyi/deprecated-globals
```

After a bit of investigation I found that it is caused the use of `global.Expo.Constants.appOwnership`

## Steps To Reproduce
just use expo SDK 40 and react-native-reanimated

### Expected behavior
No warning is  caused

### Actual behavior
image

## Package versions
1.13.0

Could some one fix this and create PR?

brentvatne:
## Description

In Expo SDK 40 we deprecated the global Expo object. In SDK 41, we are deleting it, so we need to change this.

Fixes #1612

## Changes

Instead of using the global Expo object, we check for the `executionEnvironment` on the `ExponentConstants` native module, if it is available. This is the same thing we do in @react-native-async-storage/async-storage. In the future, we could potentially move to try/catch around requiring expo-constants, but this feature is only available on react-native@>=63 so I don't think we should start depending on it in libraries yet.

This is a bit tricky for react-native-reanimated authors to test in the Expo Go context. Outside of Expo Go, you can use AnimatedDebug to ensure that this does not cause any new problems outside of Expo Go (although I think this is easy to determine from reading the code).

## Checklist

- [n/a] Updated TS types
- [n/a] Added TS types tests
- [n/a] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/core/AnimatedDebug.js
+++ b/src/core/AnimatedDebug.js
@@ -1,4 +1,5 @@
 import invariant from 'fbjs/lib/invariant';
+import { NativeModules } from 'react-native';
 import { val } from '../val';
 import { adapt, createAnimatedBlock as block } from './AnimatedBlock';
 import { createAnimatedCall as call } from './AnimatedCall';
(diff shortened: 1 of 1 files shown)


jakub-gonet (approved):
Looks good, thanks!

Issue #1808, opened 2021-03-08
useAnimatedProps don't use a current SharedValue state


nandorojo:
## Description

If you use `animatedProps` on web, it only uses the initial value. Any changes to shared values does nothing.

### Screenshots

https://www.loom.com/share/dc2aa06097b34de99b9182d059cadc95

Here's a video. Notice that, if I change the initial shared value and refresh the page, it works as expected. However, if I mutate the shared value in a function, `useAnimatedProps` still returns the initial value.

In my case, I'm trying to enable/disable pointer events for the bottom button.

## Steps To Reproduce
 
1. Create reanimated project
2. Add `useAnimatedProps` to a component
3. Run `yarn web`

### Expected behavior

Mutating a shared value (in the example, `pointerEvents.value`) should result in the `animatedProps` updating on web.

### Actual behavior

The initial value returned by `useAnimatedProps` never updates.

## Snack or minimal code example

I created a minimal reproduction [here](https://github.com/nandorojo/reanimated-UAP-web-bug).

<details>
  <summary><b>Code here</b></summary>

```jsx
import * as React from "react";
import { StyleSheet, View, Button } from "react-native";
import Animated, {
  useAnimatedProps,
  useSharedValue,
} from "react-native-reanimated";

// these both do nothing for web
Animated.addWhitelistedNativeProps({
  pointerEvents: true,
});
Animated.addWhitelistedUIProps({
  pointerEvents: true,
});

export default function App() {
  const pointerEvents = useSharedValue("auto");

  const bottomProps = useAnimatedProps(() => ({
    pointerEvents: pointerEvents.value,
  }));

  const toggle = () => {
    if (pointerEvents.value === "none") {
      pointerEvents.value = "auto";
    } else {
      pointerEvents.value = "none";
    }
    console.log("pointer events:", pointerEvents.value);
  };

  return (
    <View style={styles.container}>
      <View style={styles.half}>
        <Button color="black" title="Toggle Pointer Events" onPress={toggle} />
      </View>
      <Animated.View
        style={[styles.half, styles.bottom]}
        animatedProps={bottomProps}
      >
        <Button
          color="black"
          title="Press me?"
          onPress={() => alert("has pointer events!")}
        />
      </Animated.View>
    </View>
  );
}

const styles = StyleSheet.create({
  container: { flex: 1 },
  half: {
    flex: 1,
    backgroundColor: "hotpink",
    justifyContent: "center",
    alignItems: "center",
  },
[...]

PR #1819, opened 2021-03-11
Set missing viewRef for animatedProps

piaskowyk:
## Description

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1808

If someone uses `animatedProps` instead of `style`, we didn't set `viewRef` and without properly setting `viewRef` we weren't able to update any web component property.

<details>
<summary>code</summary>

```js
import * as React from "react";
import { StyleSheet, View, Button } from "react-native";
import Animated, {
  useAnimatedProps,
  useSharedValue,

Animated.addWhitelistedNativeProps({
  pointerEvents: true,
});
Animated.addWhitelistedUIProps({
  pointerEvents: true,
});

  const pointerEvents = useSharedValue("auto");

  const bottomProps = useAnimatedProps(() => ({
    pointerEvents: pointerEvents.value,
  }));

  const toggle = () => {
    if (pointerEvents.value === "none") {
      pointerEvents.value = "auto";
    } else {
      pointerEvents.value = "none";
    }
    console.log("pointer events:", pointerEvents.value);
  };

  return (
      <View style={styles.half}>
        <Button color="black" title="Toggle Pointer Events" onPress={toggle} />
      </View>
      <Animated.View
        style={[styles.half, styles.bottom]}
        animatedProps={bottomProps}
      >
        <Button
          color="black"
          title="Press me?"
          onPress={() => alert("has pointer events!")}
        />
      </Animated.View>
    </View>
  );
}

  container: { flex: 1 },
  half: {
    flex: 1,
    backgroundColor: "hotpink",
    justifyContent: "center",
    alignItems: "center",
  },
  bottom: {
    backgroundColor: "green",
  },
});
```

</details>

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -396,6 +396,9 @@ export default function createAnimatedComponent(Component) {
         } else if (key === 'animatedProps') {
           Object.keys(value.initial).forEach((key) => {
             props[key] = value.initial[key];
+            if (value.viewRef?.current === null) {
+              value.viewRef.current = this;
+            }
           });
         } else if (value instanceof AnimatedEvent) {
           // we cannot filter out event listeners completely as some components
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -396,7 +396,7 @@ export default function createAnimatedComponent(Component) {
         } else if (key === 'animatedProps') {
           Object.keys(value.initial).forEach((key) => {
             props[key] = value.initial[key];
-            if (value.viewRef?.current === null) {
+            if (value.viewRef.current === null) {
               value.viewRef.current = this;
             }
           });


src/createAnimatedComponent.js
```diff
@@ -396,6 +396,9 @@ export default function createAnimatedComponent(Component) {
         } else if (key === 'animatedProps') {
           Object.keys(value.initial).forEach((key) => {
             props[key] = value.initial[key];
+            if (value.viewRef?.current === null) {
```
jakub-gonet, line 399: `value.viewRef` is optional in `animatedProps`? I'm asking because `style` prop uses `style.viewRef.current === null`

piaskowyk: Right, in `createAnimatedComponent` the `viewRef` should always exists, but in updateProps is [optional](https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/Hooks.js#L449).

jakub-gonet (approved):

PR #1820, opened 2021-03-11
Perf/use make shared

mrousavy:
## Description

This PR is based on https://github.com/software-mansion/react-native-reanimated/pull/1790, so that has to be merged first. I have de-coupled it in case you don't want to replace new with make_shared/make_unique for various reasons.

This PR replaces all `new` calls with `make_shared` / `make_unique`, since that is a bit faster (only has to allocate memory once, for T and for control block).

## Changes

- Replace all `new` with `make_shared` / `make_unique`


mrousavy:
Closing in favor of https://github.com/software-mansion/react-native-reanimated/pull/1859.

PR #1832, opened 2021-03-16
Show proper error for transform in UAS

karol-bisztyga:
## Description

We want to handle a situation when a user forgets `.value` using `transform` in `useAnimatedStyle`:

```
const sv = useSharedValue(0);
  return {
    transform: [{translateX: sv}],// should be sv.value
    width: sv,// should be sv.value
  };
});
```

For now, in such a case an uninformative error is being thrown.

### DIFF
--- a/src/reanimated2/Hooks.js
+++ b/src/reanimated2/Hooks.js
@@ -493,11 +493,26 @@ export function useAnimatedStyle(updater, dependencies, adapters) {
 
   // check for invalid usage of shared values in returned object
   let wrongKey;
-  const isError = Object.keys(initial).some((key) => {
-    const element = initial[key];
+  const isObjectInvalid = (element, key) => {
     const result = typeof element === 'object' && element.value !== undefined;
     if (result) {
       wrongKey = key;
+    }
+    return result;
+  };
+  const isError = Object.keys(initial).some((key) => {
+    const element = initial[key];
+    let result = false;
+    if (key === 'transform') {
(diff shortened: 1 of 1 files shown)


src/reanimated2/Hooks.js
```diff
@@ -493,13 +493,28 @@ export function useAnimatedStyle(updater, dependencies, adapters) {
...
     const result = typeof element === 'object' && element.value !== undefined;
     if (result) {
       wrongKey = key;
     }
     return result;
+  };
+  const isError = Object.keys(initial).some((key) => {
+    const element = initial[key];
+    let result = false;
+    if (key === 'transform') {
```
karol-bisztyga, line 506: is this the only situation like that? Or should we handle more cases like this in a more generic way?

src/reanimated2/Hooks.js
```diff
@@ -493,13 +493,28 @@ export function useAnimatedStyle(updater, dependencies, adapters) {
...
+      for (const transformItem of element) {
+        const item = Object.values(transformItem)[0];
+        result = isObjectInvalid(item, key);
```
karol-bisztyga, line 509: I know naming could be better(`item`), was also thinking about an inversion here(`isObjectValid` might be more readable)

karol-bisztyga:
I noticed it doesn't play along well with the fast refresh, maybe it's worth looking into it?

piaskowyk (approved):
I tested it and everything works 💪

PR #1833, opened 2021-03-16
RuntimeDecorator: Correctly set global

mrousavy:
## Description

This is a follow-up PR of https://github.com/software-mansion/react-native-reanimated/pull/1790. Until now, `global` in Worklets was set to a dummy object which just contained `__reanimatedWorkletInit`. This PR changes this, so that all other global properties that are being set in the `RuntimeDecorator::decorateRuntime` func are also in the `global` object.

This means, `global._LABEL` returns the correct label, `global._WORKLET` returns `true`, etc.

I have decided to make this a follow-up PR, because I am not sure if this has side-effects. I haven't noticed any, but if anyone else does, please let me know. cc @Szymon20000 @karol-bisztyga @piaskowyk 

## Changes

* Assign `global` to the hidden global object so you can correctly use `global.` code. (e.g. `global._log`, `global._WORKLET`, `global._LABEL`, ...)
* Add `global._chronoNow`, a function that returns chrono's time_since_epoch (polyfill for `performance.now()`)


mrousavy:
Closed in favor of https://github.com/software-mansion/react-native-reanimated/pull/1862.

PR #1837, opened 2021-03-16
Build Android from source

mrousavy:
## Description

This PR is a follow-up of https://github.com/software-mansion/react-native-reanimated/pull/1790.

With native worklet APIs being exposed, it became possible for other native libraries to use the Reanimated "workletization" API. This works great for iOS as you just have to include the `<RNReanimated/ShareableValue.h>` header, but since the Android code is being pre-compiled into an `.aar` file, it is not possible to include and link the reanimated code in your library without maintaining a full reanimated source code copy inside of your library's code (at least as far as I'm aware)

> I have spent a few hours trying to import the react-native-reanimated aar in [my multithreading library's `CMakeLists.txt` file](https://github.com/mrousavy/react-native-multithreading/blob/master/android/CMakeLists.txt) without success, but if anyone reading this is aware of a way to do that, please let me know.

## Changes

- Removed everything related to building a `.aar` package
- Replaced `createNPMPackage` with a simple `npm pack`

## Benefits

* ✅ Other libraries can easily include reanimated headers in their source code, and it will be linked once the user builds their app. This means, it's possible to use the "workletization" API inside of other libraries.
* ✅ Users can easily patch the reanimated source code if there is a critical issue and don't have to wait for a new release
* ✅ Users can more easily edit reanimated source code and see problems, possibly leading to more PRs
* ✅ It's easier for you guys to build the NPM package (just npm publish instead of gradle cleaning and building for every version since 0.62)

With the new architecture and turbomodules becoming more mainstream, all android apps on RN 0.64 and above require the NDK tools to be installed anyways ([see the template's build.gradle](https://github.com/facebook/react-native/blob/3ff7e86b0ff12823e5d89a0d3a7cceb79e0b25b0/template/android/build.gradle#L9)), so this means users can simply build their project without any additional change (assuming they're on RN 0.64+)

I'm of course always open for other ideas around this problem (not being able to import reanimated headers/code in other libraries), this is just the solution I came up with.


mrousavy:
Closing because SWM doesn't want to let users build from source.

Issue #1815, opened 2021-03-09
Problems reloading Android app while using Reanimated v2


nelsonprsousa:
## Description

When I reload Android using `R + R`, everything's okay.
When I reload Android using `Cmd + M -> Reload`, the app crashes.
When I reload Android using the key `R` in metro console, the app crashes.

### Screenshots

https://user-images.githubusercontent.com/8364345/110551740-ff335e00-812d-11eb-9f87-c9ffee59e2ab.mov

## Steps To Reproduce

1. Clone this repo: https://github.com/nelsonprsousa/rnn-testing/tree/test-rea-v2
2. Use `test-rea-v2` branch
3. yarn android

### Expected behavior
The app should reload with Cmd + M -> Reload and when we send the "R" key signal in metro console.
iOS works as expected on all the scenarios.

### Actual behavior
Android app crashes on reload.

## Snack or minimal code example
I've made a new react-native application, added react-native-navigation and reanimated v2.
Then I added the following code:

```javascript
import React from 'react';
import {SafeAreaView, StyleSheet, View, Text} from 'react-native';
import Animated, {useAnimatedStyle} from 'react-native-reanimated';

const App: () => React$Node = () => {
  const animated = useAnimatedStyle(() => {
    return {
      transform: [{scale: 1}],
    };
  });

  return (
    <SafeAreaView>
      <Animated.ScrollView
        contentInsetAdjustmentBehavior="automatic"
        style={{backgroundColor: 'gray'}}>
        <View>
          <Text style={styles.footer}>Hello Reanimated v2!</Text>
        </View>
      </Animated.ScrollView>
    </SafeAreaView>
  );
};

const styles = StyleSheet.create({
  scrollView: {
    backgroundColor: 'white',
  },
  footer: {
    color: 'black',
    fontSize: 12,
    fontWeight: '600',
    padding: 4,
    paddingRight: 12,
    height: 1500,
  },
});

export default App;
```

Please note that `animated` const isn't used. If we comment that const (that's not being used), everything works as expected. Also tried to really use it and got the same problem, so I am publishing this way for simplicity.

## Package versions

- React: 16.13.1
- React Native: 0.63.4
- React Native Reanimated: 2.0.0
- React Native Navigation: 7.11.3
- NodeJS: 14.6.0

nelsonprsousa:
I was able to debug a little bit this problem.

Here's what I found using logcat:
[...]

Issue #1727, opened 2021-02-17
ANR:  Input dispatching timed out 


Andarius:
## Description

App crashes on Android in production with this error

### Screenshots

```
ANR:  Input dispatching timed out (Waiting to send non-key event because the touched window has not finished processing certain input events that were delivered to it over 500.0ms ago.  Wait queue length: 2.  Wait queue head age: 5564.1ms.)
        at com.swmansion.reanimated.Scheduler.triggerUI(Scheduler.java:-2)
        at com.swmansion.reanimated.Scheduler.access$000(Scheduler.java:7)
        at com.swmansion.reanimated.Scheduler$1.run(Scheduler.java:17)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7711)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:516)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:950)
```

Full stack trace [here](https://gist.github.com/Andarius/3fb3b2e1418fa2f27503592669c23d69)

## Steps To Reproduce

Let the app run for a few minutes

### Expected behavior

No crash

### Actual behavior

Crash

## Snack or minimal code example

Unknown, Sorry not to be helpful here...

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: ^2.0.0-rc.3

Szymon20000:
Looks like our method `triggerUI` is called from two different threads at the same time. We only call it through `RNcontext.runOnUiQueueThread()` method.
```Java
mContext.runOnUiQueueThread(()->{triggerUI});
```
It would be really helpful if you could answer our questions:

Do you have multiple instances of react-native in your project? 
Do you use headless JS?
Do you use any library for background tasks?

Andarius:
Hey @Szymon20000 

**Do you have multiple instances of react-native in your project?**
No
**Do you use headless JS?**
No
**Do you use any library for background tasks?**
Maybe bugsnag (the reporting lib) otherwise no

[...]

PR #1839, opened 2021-03-17
Fix full reload on Android

Szymon20000:
## Description
fixes: https://github.com/software-mansion/react-native-reanimated/issues/1815
There can be a situation when reanimated has been destroyed but there is still a runnable on UI queue that wants to trigger our render method. In order to avoid the scenario, I've added a condition in scheduler.schedulerOnUI that checks if react native instance has been destroyed.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NativeProxy.java
+++ b/android/src/main/java/com/swmansion/reanimated/NativeProxy.java
@@ -73,10 +73,13 @@ public class NativeProxy {
   private final HybridData mHybridData;
   private NodesManager mNodesManager;
   private final WeakReference<ReactApplicationContext> mContext;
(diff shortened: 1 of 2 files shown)


Andarius:
@Szymon20000, will this fix https://github.com/software-mansion/react-native-reanimated/issues/1727#issuecomment-799289664 ?

piaskowyk (approved):

Szymon20000:
@Andarius Probably

nelsonprsousa:
Thanks for your time fixing this issue 🙏

PR #1842, opened 2021-03-17
JSC for Android

piaskowyk:
## Description

Added support for JSC runtime on Android

## Checklist

- [x] Ensured that CI passes


Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -58,7 +58,7 @@ std::vector<std::shared_ptr<MutableValue>> extractMutablesFromArray(jsi::Runtime
 
 NativeReanimatedModule::NativeReanimatedModule(std::shared_ptr<CallInvoker> jsInvoker,
                                                std::shared_ptr<Scheduler> scheduler,
-                                               std::unique_ptr<jsi::Runtime> rt,
+                                               std::unique_ptr<jsi::Runtime>& rt,
```
kmagiera, line 61: this does not seem like a good idea, if the point of the pointer is to be referenced in a number of places we should use shared_ptr and not unique_ptr -> unique is for usecases where there is only one "holder"

Szymon20000 (approved):
Looks fine to me.

Yonom:
@piaskowyk do you have an ETA on when this will land?

piaskowyk:
@Yonom We released it in the latest version [2.1.0](https://github.com/software-mansion/react-native-reanimated/releases/tag/2.1.0).

zachgibson:
Does this PR remove the need to use Hermes on Android?

piaskowyk:
@zachgibson Yes, you can choose JSC or Hermes runtime for Android

SaeedZhiany:
@piaskowyk 

What about iOS, can also we use this library without Hermes enabled for iOS? if yes what is the minimum version that can be used?

Arkan4ik:
@SaeedZhiany I am using "react-native-reanimated": "^2.3.1", and "react-native": "0.64.2", without hermes enabled and all works fine

HSReact:
> @SaeedZhiany I am using "react-native-reanimated": "^2.3.1", and "react-native": "0.64.2", without hermes enabled and all work fine

Hi @Arkan4ik how is the performance of reanimated 2 without hermes? is it better than reanimated 1? Thanks for any help

Issue #1758, opened 2021-02-26
Value is undefined, expected an Object


enesozturk:
## Description

Hi, I am trying to pass list of objects that includes fields as `function` type to a component. But when I try to pass shared value list to a React state to render, it can not call functions. When I try to press Touchable, it gives the error.

*Value is undefined, expected an Object*.

I am not sure if it is because of wrong usage or a bug. Thats why sorry for that if I am wrong place.

### Screenshots

<img width="387" alt="Screen Shot 2021-02-26 at 11 01 12 AM" src="https://user-images.githubusercontent.com/19428358/109273843-c937d500-7823-11eb-9d2b-c0ddfd83fed4.png">

<img width="389" alt="Screen Shot 2021-02-26 at 11 41 41 AM" src="https://user-images.githubusercontent.com/19428358/109276904-a60f2480-7827-11eb-9992-310bcbd8e9a8.png">

## Steps To Reproduce

1. Provide a shared list that has list of objects has function as field
2. Move shared list to a state
3. Try rendering list and using onPress handlers of the objects

### Expected behavior

Render React.ReactNode and call handler functions passed with Shared Value

### Actual behavior

I does not render and gives above error when I use the onPress function

## Snack or minimal code example

Here how I pass functions;
```js
export type MenuItemProps = {
  title: string;
  onPress: () => void;
};

const list = useSharedValue<MenuItemProps[]>([
    {
      text: 'Action 1',
      onPress: () => {
        console.log('Action 1');
      },
    },
    {
      text: 'Action 2',
      onPress: () => {
        console.log('Action 2');
      },
    },
]);

<List items={items} />

```

And List component sets shared value to state to be able to render:

`List.tsx`
```js
const MenuList = ({ items }: { items: Animated.SharedValue<MenuItemProps[]> }) => {
  // I set my list to React state to be able to render
  const [itemList, setItemList] = React.useState<MenuItemProps[]>([]);

  const setter = (items: MenuItemProps[]) => {
    setItemList(items);
  };

  useAnimatedReaction(
    () => items.value,
    items => {
      runOnJS(setter)(items);
    },
    [items]
  );

  return (
    <>
      {itemList.map((item: MenuItemProps) => {
        return (
          <TouchableOpacity onPress={item.onPress}>
            <Text>{item.text}</Text>
          </TouchableOpacity>
        );
      })}
    </>
  );
};
```

## Package versions

- React: 16.13.1
- React Native: 0.63.4
[...]

PR #1844, opened 2021-03-18
Fix Host Function

Szymon20000:
## Description

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1758
Currently, when we send callback function to UI thread and then back to RN then we will get function wrapper that checks is somebody uses runOnJS. In order to solve the problem, I've added logic that attaches additional info to a function wrapper that allows to get original callback.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -104,12 +104,10 @@ void ShareableValue::adapt(jsi::Runtime &rt, const jsi::Value &value, ValueType 
         if (!primalFunction.isUndefined()) {
             jsi::Object handlerAsObject = primalFunction.asObject(rt);
             std::shared_ptr<HostFunctionHandler> handler = handlerAsObject.getHostObject<HostFunctionHandler>(rt);
-            valueContainer = std::make_unique<HostFunctionWrapper>(handler, &rt);
+            valueContainer = std::make_unique<HostFunctionWrapper>(handler);
         } else {
             valueContainer = std::make_unique<HostFunctionWrapper>(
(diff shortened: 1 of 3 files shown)


Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -239,8 +250,10 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
     }
     case ValueType::HostFunctionType: {
       auto hostFunctionWrapper = ValueWrapper::asHostFunctionWrapper(valueContainer);
-      if (hostFunctionWrapper->hostRuntime == &rt) {
+      auto hostRuntime = hostFunctionWrapper->value->hostRuntime;
```
piaskowyk, line 253: I think we can use here `auto &` instead of `auto` to avoid unnecessary copying of objects.

piaskowyk (approved):

enesozturk:
You made my day 🔥 Looking forward to examples and new release.

enesozturk:
Hi, is this pr included in v2.0.1? Looks like the problem is still continue.

enesozturk:
Hi, I am following issue for my package. It is solved with v2.1.0 actually. When I try it on Expo project with same version (v2.1.0), problem still exist. Not sure if I open an issue in Expo. Wanted to let you know.

colinux:
I found Expo 41 ships reanimated v2.0.1 native modules, but v2.1.0 on JS side
https://github.com/expo/expo/commit/d783eeb1681de085dcf6a433ae7b41b2d238e481 

Not sure why, maybe because the reanimated release was a little late during expo 41 release, so they don't took the risk to ship with it. Anyway I don't think there will be a solution until expo 42 in managed workflow…

Issue #1838, opened 2021-03-17
TurboModuleRegistry.getEnforcing(...) - HELP!!!


aqilhex:
## Description
I can't find any solution to fix this error caused by react-native-reanimated@2.0.0, Downgraded to v1.x.x , Clearing cache, Enabling jetifier did not help. Someone help me please, I need it to work with react-navigation/drawer.

- React: 17.0.1
- React Native: 0.64.0
- React Native Reanimated: 2.0.0
- NodeJS: v12.16.3

mosoakinyemi:
I had this same issue for iOS, I solved it simply by:
1. Deleting the app on the simulator 
2. Running `npx react-native run-ios`

4yki:
Same for me with:
- React Native: 0.64.0
- React Native Reanimated: 2.0.0

Invariant Violation: TurboModuleRegistry.getEnforcing(...): 'NativeReanimated' could not be found. Verify that a module by this name is registered in the native binary., js engine: hermes

teamzz111:
Same here

 Invariant Violation: TurboModuleRegistry.getEnforcing(...): 'NativeReanimated' could not be found. Verify that a module by this name is registered in the native binary., js engine: hermes

boonook:
Same for me with

remove @react-navigation/drawer  normal
use @react-navigation/drawer error

melero2000:
Same issue for me:

- React Native: 0.64.0
- React Native Reanimated: 2.0.0
- react-navigation/drawer: 5.12.4

jakub-gonet:
Duplicate of #1825.

Issue #1825, opened 2021-03-13
[2.0.0] Invariant Violation: TurboModuleRegistry.getEnforcing(...): 'NativeReanimated' could not be found. Verify that a module by this name is registered in the native binary.


Waqas-Jani:
Hey, I'm facing this issue in 2.x.x version,When importing @react-navigation/drawer. but perviouse version 1.x.x working fine.
React-native 0.63.4 

- React: 16.13.1
- React Native: 0.63.4
- React Native Reanimated: 2.0.0
- NodeJS: 14.15.5

PritishSawant:
Yup i am facing same issue as well on RN v 0.64

braincore:
Bit of a long shot, but I'm also getting this error when using reanimated from an iOS share extension.

piaskowyk:
@Waqas-Jani 
Did you use `npm` or `yarn`? (you should use `yarn`) Did you add Reanimated plugin?
#### How to add plugin
check your babel.config.js file if contains following line:
```
module.exports = {
  ...
  plugins: ['react-native-reanimated/plugin'],
  ...
};
```
more detailed instruction in our docs: https://docs.swmansion.com/react-native-reanimated/docs/installation#babel-plugin

preetgaggu:
I have solved this problem. Move on to **React Navigation v6** . V6 is mostly same as V5. if not solved than down grade **react-native-reanimated** to version 1. Don't forget to start react native project with `npx react-native start --reset-cache`

abdullahkhan70:
> I have solved this problem. Move on to **React Navigation v6** . V6 is mostly same as V5. if not solved than down grade **react-native-reanimated** to version 1. Don't forget to start react native project with `npx react-native start --reset-cache`

Well, I am facing this same problem from the android side. My iOS is working fine, but my Android side giving me this error. Do you know anything about how can I sort out this problem?
I am using react-native-reanimated:2.0.0 and also react-native: 0.62.0.
Basically, I developed this app using react-native:0.63.2 and react-native-reanimated:2.0.0-rc.3. Is there any way to sort out this problem?

raultunduc1995:
```
React-Native: 0.62.0
React-Navigation: ^5
Reanimate: ^2.0.0
```

This message is shown in the Xcode's logs when running the app on iOS, but it works perfectly on Android.
I don't know what I am doing wrong.
I have followed all the steps described in the documentation.
Please help!

preetgaggu:
[...]

Issue #1798, opened 2021-03-06
Importing createDrawerNavigator throws 'NativeReanimated' error.


ModusPwnens1337:
### Description
After following the installation instructions for [React Navigation](https://reactnavigation.org/docs/getting-started/) and then [Drawer Navigation](https://reactnavigation.org/docs/drawer-based-navigation/), I am getting a NativeReanimated error.

I have reanimated in both my package-lock.json:
```    
"react-native-reanimated": {
      "version": "2.0.0",
      "resolved": "https://registry.npmjs.org/react-native-reanimated/-/react-native-reanimated-2.0.0.tgz",
      "integrity": "sha512-kIrdBoXSky7DQ62SOgosgimKM+Lt+SzAaM+ovVpCLBcwUK2aYRfLxa9ffgvKjeH9/n7dONlwEMjbKssGkuyq2Q==",
      "requires": {
        "@babel/plugin-transform-object-assign": "^7.10.4",
        "fbjs": "^3.0.0",
        "mockdate": "^3.0.2",
        "string-hash-64": "^1.0.3"
      }
```
and in my package.json:
```  
"dependencies": {
    "@react-native-community/masked-view": "^0.1.10",
    "@react-navigation/drawer": "^5.12.4",
    "@react-navigation/native": "^5.9.3",
    "@react-navigation/stack": "^5.14.3",
    "react": "16.13.1",
    "react-native": "0.63.4",
    "react-native-gesture-handler": "^1.10.3",
    "react-native-reanimated": "^2.0.0",
    "react-native-safe-area-context": "^3.1.9",
    "react-native-screens": "^2.18.1"
  }
  ```

It was installed vis the React Navigation instructions as follows:
`npm install react-native-reanimated react-native-gesture-handler react-native-screens react-native-safe-area-context @react-native-community/masked-view`

This is the error I'm receiving in Metro and node after running **npx react-native run-android:**
[Sat Mar 06 2021 10:25:12.380]  **BUNDLE  ./index.js**

[Sat Mar 06 2021 10:25:15.192]  **ERROR    Invariant Violation: TurboModuleRegistry.getEnforcing(...): 'NativeReanimated' could not be found. Verify that a module by this name is registered in the native binary.**
[Sat Mar 06 2021 10:25:15.196]  **ERROR    Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication)**
[...]

PR #1845, opened 2021-03-18
Let users use v1 without additional steps

Szymon20000:
Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1838
Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1825
Fixes: https://github.com/software-mansion/react-native-reanimated/issues/1798
We forgot To wrap some initialization steps of rea 2 into a condition that checks if the native side of reanimated 2 is available. Because of that people that want to use only v1 have to go through installation steps on Android. 

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


piaskowyk (approved):
I tested and is OK. 
Maybe we should change little the warn message https://github.com/software-mansion/react-native-reanimated/blob/4ff075eb8c59581642de1aded30db58af7163b8f/src/reanimated2/core.js#L25 to
`Reanimated 2 failed to create a worklet, maybe you forgot to add Reanimated's babel plugin? If you want to use only Reanimated v1, you can ignore this message.`

src/reanimated2/core.js
```diff
@@ -15,14 +15,13 @@ if (global._setGlobalConsole === undefined) {
...
+
 export const checkPluginState = () => {
-  if (
-    !(() => {
-      'worklet';
-    }).__workletHash &&
-    !process.env.JEST_WORKER_ID
-  ) {
-    throw new Error(
+  if (!testWorklet.__workletHash && !process.env.JEST_WORKER_ID) {
```
satya164, line 23: Unrelated to this PR, but is it possible to expose this in a public API? My use case is to do feature detection regarding if I can use the Reanimated 2 APIs and then use the new implementation if possible.

src/reanimated2/core.js
```diff
@@ -277,15 +268,23 @@ export function createAnimatedPropAdapter(adapter, nativeProps) {
...
-  const console = {
-    debug: runOnJS(capturableConsole.debug),
-    log: runOnJS(capturableConsole.log),
-    warn: runOnJS(capturableConsole.warn),
-    error: runOnJS(capturableConsole.error),
-    info: runOnJS(capturableConsole.info),
-  };
-  _setGlobalConsole(console);
-})();
+if (!NativeReanimated.useOnlyV1) {
```
satya164, line 271: Unrelated to this PR, but is it possible to expose this in a public API? My use case is to do feature detection regarding if I can use the Reanimated 2 APIs and then use the new implementation if possible.

piaskowyk: @Szymon20000 What do you think?

Szymon20000: Sure

PR #1859, opened 2021-03-23
Restructure code

mrousavy:
Part 1 of 3

* Re-indents a few wrongly indented code-pieces
* Uses `std::make_shared`/`std::make_unique` instead of `new` (it's a bit faster, and cleaner to read imo)
* Restructures `UIResponder+Reanimated.mm` to be a bit more future proof and easier to read (dummy macro, include checking, ...)

### DIFF
--- a/ios/native/UIResponder+Reanimated.mm
+++ b/ios/native/UIResponder+Reanimated.mm
@@ -7,19 +7,12 @@
 #import <React/RCTCxxBridgeDelegate.h>
 #import <RNReanimated/REAEventDispatcher.h>
 
-#if __has_include(<React/RCTJSIExecutorRuntimeInstaller.h>)
+#if RNVERSION >= 64
 #import <React/RCTJSIExecutorRuntimeInstaller.h>
-#else
-#define RCTJSIExecutorRuntimeInstaller(x) x
 #endif
 
-#if __has_include(<ReactCommon/CallInvoker.h>)
-#import <ReactCommon/CallInvoker.h>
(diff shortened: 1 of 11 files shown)


ios/native/UIResponder+Reanimated.mm
```diff
@@ -7,12 +7,19 @@
 #import <React/RCTCxxBridgeDelegate.h>
 #import <RNReanimated/REAEventDispatcher.h>
 
-#if RNVERSION >= 64
```
Szymon20000, line 10: Why do we need this change. Isn't RNVERSION clearer?

mrousavy: I think by checking if the JS-CallInvoker file can be included we stay future proof - just in case anything changes in the future which requires some additional RNVERSION checks that get really complicated. It takes me longer to read that `RNVERSION >= ...` or `RNVERSION < ...` than just `if has include` (which sounds more like proper english)

piaskowyk: The current approach is more informative in my opinion. This informs in a clear way why is possible these header files can don't exist, and which header file is related to which the react-native version.

mrousavy: ai ai captain, will revert in a few mins

mrousavy: reverted those changes. 👍 @piaskowyk @Szymon20000

Szymon20000 (approved):
Thanks!

PR #1860, opened 2021-03-23
Restructure `RuntimeDecorator`

mrousavy:
Part 2 of 3

* Restructures `RuntimeDecorator` to use the term "Worklet" instead of "UI", and "React Runtime" instead of "Host Runtime".
* Move `isHostRuntime` and `isUIRuntime` from `NativeReanimatedModule` to `RuntimeDecorator` (and make static)

### DIFF
--- a/Common/cpp/SharedItems/MutableValue.cpp
+++ b/Common/cpp/SharedItems/MutableValue.cpp
@@ -16,7 +16,7 @@ void MutableValue::setValue(jsi::Runtime &rt, const jsi::Value &newValue) {
     }
   };
   
-  if (module->isUIRuntime(rt)) {
+  if (RuntimeDecorator::isWorkletRuntime(rt)) {
     notifyListeners();
   } else {
     module->scheduler->scheduleOnUI([notifyListeners] {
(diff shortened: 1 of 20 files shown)


Common/cpp/SharedItems/MutableValue.cpp
```diff
@@ -16,7 +16,7 @@ void MutableValue::setValue(jsi::Runtime &rt, const jsi::Value &newValue) {
     }
   };
   
-  if (module->isUIRuntime(rt)) {
+  if (RuntimeDecorator::isWorkletRuntime(rt)) {
```
Szymon20000, line 19: I don't think that RuntimeDecorator should have such methods. Maybe it will fit better with RuntimeManager?

mrousavy: RuntimeDecorator is the one responsible for marking a runtime a "worklet runtime", why shouldn't it also be exposing a method that can determine this?

Szymon20000: You have convinced me.

piaskowyk (approved):

PR #1861, opened 2021-03-23
Extract Worklet API to `RuntimeManager.h`

mrousavy:
Part 3 of 3

* Lift `valueSetter`, `runtime`, `errorHandler`, `scheduler` and `workletsCache` out of `NativeReanimatedModule`, into `RuntimeDecorator.h`
* Let `NativeReanimatedModule` extend `RuntimeDecorator` (all worklet API consumers should do that)
* Let every API func accept `RuntimeDecorator` instead of `NativeReanimatedModule` to make it more general


mrousavy:
Note that I still haven't tested if it would work to use a SharedValue in a separate worklet runtime (that is not the REA UI runtime).

So e.g.:

```jsx
const sv = useSharedValue(1)

const as = useAnimatedStyle(() => ({ opacity: sv.value }), [sv]);
const fp = useFrameProcessor(() => {
  sv.value = Math.random() // <-- no idea if that works.
}, [sv]);

// ...
```

If it doesn't work, this will require an additional PR in the future.

mrousavy:
@Szymon20000 let me know if you need anything.

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -61,21 +61,18 @@ NativeReanimatedModule::NativeReanimatedModule(std::shared_ptr<CallInvoker> jsIn
...
                                                std::function<jsi::Value(jsi::Runtime &, const int, const jsi::String &)> propObtainer,
-                                               PlatformDepMethodsHolder platformDepMethodsHolder) : NativeReanimatedModuleSpec(jsInvoker),
-                                                  runtime(std::move(rt)),
-                                                  mapperRegistry(std::make_shared<MapperRegistry>()),
-                                                  eventHandlerRegistry(std::make_shared<EventHandlerRegistry>()),
+                                               PlatformDepMethodsHolder platformDepMethodsHolder) :
+                                                  NativeReanimatedModuleSpec(jsInvoker),
+                                                  RuntimeManager(std::move(rt), errorHandler, scheduler),
+                                                  mapperRegistry(new MapperRegistry()),
+                                                  eventHandlerRegistry(new EventHandlerRegistry()),
```
piaskowyk, line 68: Why you use here `new` instead of `make_shared`? I prefer `make_shared` because is safer and allocating memory only once.

piaskowyk (commented):
Generally, It looks ok to me👌 
It is waiting for @Szymon20000 review.

Common/cpp/NativeModules/NativeReanimatedModule.cpp
mrousavy: Fixed in https://github.com/software-mansion/react-native-reanimated/pull/1861/commits/b638ad7bdcfbb53ce5535234c5fb41139709f89e

Common/cpp/headers/SharedItems/RuntimeManager.h
```diff
@@ -0,0 +1,54 @@
+//
+//  RuntimeManager.h
+//  RNReanimated
+//
+//  Created by Marc Rousavy on 05.03.21.
+//  Copyright © 2021 Facebook. All rights reserved.
```
Szymon20000, line 6: Please, remove the entire comment.

mrousavy: done.

PR #1862, opened 2021-03-23
Correctly set global

mrousavy:
## Description

This is a follow-up PR of https://github.com/software-mansion/react-native-reanimated/pull/1861. Until now, `global` in Worklets was set to a dummy object which just contained `__reanimatedWorkletInit`. This PR changes this, so that all other global properties that are being set in the `RuntimeDecorator::decorateRuntime` func are also in the `global` object.

This means, `global._LABEL` returns the correct label, `global._WORKLET` returns `true`, etc. and `Object.keys(global)` returns all the correct keys.

## Changes

* Assign `global` to the hidden global object so you can correctly use `global.` code. (e.g. `global._log`, `global._WORKLET`, `global._LABEL`, ...)
* Add `global._chronoNow`, a function that returns chrono's time_since_epoch (polyfill for `performance.now()`)


piaskowyk (approved):

tomekzaw:
Closing this PR since it looks like all the changes from PR have already been covered with these:

* #2679
* #3143

Thanks!

PR #1871, opened 2021-03-25
Add `useFrameProcessor` and `spawnThread` to plugin whitelist

mrousavy:
## Description

This PR adds two new entries in the plugin whitelist: `useFrameProcessor` and `spawnThread`. This allows for the plugin to transform those functions into worklets when the user uses them in my libraries, and has no effect on the reanimated library.

* `useFrameProcessor`: For [VisionCamera](https://github.com/cuvent/react-native-vision-camera)
* `spawnThread`: For [my multithreading library](https://github.com/mrousavy/react-native-multithreading)

### Right place?

It is a bit awkward to add such entries to an _animation library_, but there is no other solution except letting users write `'worklet'` into every function themselves. What do you guys think about moving the worklet API out of reanimated, into a separate repo that you guys + the community maintains? So that **software-mansion/react-native-reanimated** depends on **software-mansion/react-native-worklets** for example.
This has the benefit of making the reanimated library cleaner ("lean core") and possibly introduce more contributions to the Worklet core API as more people might be interested in using it.

Again, it just doesn't _feel_ like a native API that extracts JS functions as worklets and moves them into separate threaded jsi::Runtimes belongs into an _animation library_. Of course it's not my project so it's your decision.

## Changes

- Add `useFrameProcessor` to plugin.js worklet func whitelist
- Add `spawnThread` to plugin.js worklet func whitelist

In my libraries, use:

```js
useFrameProcessor(() => {
  // ...
}, []);
```

and see how no `'worklet'` directive is needed as the REA plugin does this automatically.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes


heroic:
Wouldn't it be better to rather somehow create a injector, so people can add their own functions in their apps too?

mrousavy:
@heroic I already did that here: https://github.com/software-mansion/react-native-reanimated/commit/11250a027fbea0abc719a2af2ffcd0322e145395

That's only for global native host functions though. Not sure if I need to add the same for normal functions, since you just need to add the `'worklet'` directive to get it working.

PR #1872, opened 2021-03-26
Type Easings and Colors

jakub-gonet:
## Description

Add some types to Easing, Colors, Bezier.

### DIFF
--- a/src/reanimated2/Colors.ts
+++ b/src/reanimated2/Colors.ts
@@ -29,7 +29,7 @@ interface HSV {
 const NUMBER = '[-+]?\\d*\\.?\\d+';
 const PERCENTAGE = NUMBER + '%';
 
-function call(...args: unknown[]) {
+function call(...args: unknown[]): string {
   'worklet';
   return '\\(\\s*(' + args.join(')\\s*,\\s*(') + ')\\s*\\)';
 }
--- a/src/reanimated2/Colors.ts
+++ b/src/reanimated2/Colors.ts
@@ -7,18 +7,29 @@
  */
 
 /* eslint no-bitwise: 0 */
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-nocheck
 import { Platform } from 'react-native';
 import { makeRemote, makeShareable, isConfigured } from './core';
 import { interpolate } from './interpolation';
+// @ts-ignore JS file
 import { Extrapolate } from '../reanimated1/derived';
+
+type RBGObj = {
+  r: number;
+  g: number;
+  b: number;
+};
+
+type HSVObj = {
+  h: number;
+  s: number;
+  v: number;
+};
 
 // var INTEGER = '[-+]?\\d+';
 const NUMBER = '[-+]?\\d*\\.?\\d+';
 const PERCENTAGE = NUMBER + '%';
 
-function call(...args) {
+function call(...args: unknown[]) {
   'worklet';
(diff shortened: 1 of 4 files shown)


src/reanimated2/Colors.ts
```diff
@@ -7,18 +7,29 @@
...
+  r: number;
+  g: number;
+  b: number;
+};
+
+type HSVObj = {
+  h: number;
+  s: number;
+  v: number;
+};
```
mrousavy, line 26: Is there a specific reason for making those things `type`? `interface`s have a few advantages over types, IIRC that's for performance and decl merging.

so,

```ts
interface RGB {
  r: number;
  g: number;
  b: number;
}
```


is always better unless you have a specific reason

jakub-gonet: Mosty muscle memory after I stumbled upon this issue: https://github.com/microsoft/TypeScript/issues/15300

This code can easily be swapped for interfaces, thanks for noticing.

jakub-gonet: Fixed in c2dfc60.

src/reanimated2/Colors.ts
```diff
@@ -7,18 +7,29 @@
...
+  s: number;
+  v: number;
+}
+
 // var INTEGER = '[-+]?\\d+';
 const NUMBER = '[-+]?\\d*\\.?\\d+';
 const PERCENTAGE = NUMBER + '%';
 
-function call(...args) {
+function call(...args: unknown[]) {
```
piaskowyk, line 32: It seems the return type is missing. `: string`

src/reanimated2/Colors.ts
```diff
@@ -493,12 +528,16 @@ function RGBtoHSV(r, g, b) {
  * 0 <= h, s, v <= 1
  * returns 0 <= r, g, b <= 255
  */
-function HSVtoRGB(h, s, v) {
+function HSVtoRGB(hsv: HSV): RBG;
+function HSVtoRGB(h: number, s: number, v: number): RBG;
+function HSVtoRGB(h: any, s?: any, v?: any) {
```
piaskowyk, line 533: missing type `: RGB`

src/reanimated2/Colors.ts
```diff
@@ -559,15 +604,15 @@ export function processColorInitially(color) {
...
+export function isColor(value: unknown): boolean {
   'worklet';
   if (typeof value !== 'string') {
     return false;
   }
   return processColorInitially(value) != null;
 }
 
-export function processColor(color) {
+export function processColor(color: unknown): number | null | undefined {
```
piaskowyk, line 615: Why color is as `:unknown` type?

src/reanimated2/Colors.ts
jakub-gonet: Thanks, I must've missed that.

src/reanimated2/Colors.ts
jakub-gonet: It uses `processColorInitially` which uses `normalizeColor`, both of them can accept arbitrary arguments (they're returning nulls when they cannot process the arguments). I think typings should reflect that.

Right now, we type this function as:
`export function processColor(color: number | string): number;`
so we can also simplify those methods by using `color: number | string`, casting it internally to unknown (to support JS users too), and e.g. throw when input is invalid. What do you think?

src/reanimated2/Colors.ts
jakub-gonet: Fixed in e8d1243a.

piaskowyk (approved):

PR #1883, opened 2021-03-31
Fix Threading issues (SV access from different Thread)

mrousavy:
## Description

There are a few functions that check if you're currently on the REA UI Runtime by using `RuntimeDecorator::isWorkletRuntime`. This might actually lead to threading issues if you're calling those functions from another worklet runtime that is not from Reanimated (e.g. VisionCamera Frame Processors, Multithreading, ...) - so this PR fixes this. A few of those functions should actually check if they are specifically on the UI thread, Worklet thread is not enough.

## Changes

- Add `RuntimeDecorator::isUIRuntime` - a function that specifically checks if the given `jsi::Runtime` is the REA UI Runtime as opposed to any Worklet Runtime
- Replace a few `isWorkletRuntime` calls with `isUIRuntime`

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/Common/cpp/SharedItems/MutableValue.cpp
+++ b/Common/cpp/SharedItems/MutableValue.cpp
@@ -65,7 +65,6 @@ void MutableValue::set(jsi::Runtime &rt, const jsi::PropNameID &name, const jsi:
           .callWithThis(rt, setterProxy, newValue);
       });
     }
-    return;
   }
 
 }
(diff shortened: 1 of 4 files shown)


mrousavy:
Okay so I've managed to verify that with the current REA codebase, assigning or reading a SharedValue in a worklet (on a separate runtime, such as in react-native-multithreading or react-native-vision-camera frame processors) crashes because of a threading issue.

This was not an issue before since no other library spawns "worklet" runtimes.

This PR fixes this by using `isUIRuntime` where we need the UI thread, and `isWorkletRuntime` where we need any runtime that supports running "worklets", and `isReactRuntime` for explicitly checking if it is the default React-JS runtime.

No Reanimated behaviour has changed, this only affects libraries such as rn-multithreading or react-native-vision-camera 👍 

cc @Szymon20000 could you take a look at this when you have some free time? ❤️ 🙏

Szymon20000 (approved):

PR #1891, opened 2021-03-31
fix: ssr support (#1890)

### LINKED ISSUE #1890: [2.0.0] Server-side rendering breaks
## Description

Reanimated v2 stable breaks server side rendering, because it uses `window.performance` in the constructor of `JSReanimated`.

https://github.com/software-mansion/react-native-reanimated/blob/1e80f83df79c32c314c12ae9360e39af9b5978ed/src/reanimated2/js-reanimated/JSReanimated.ts#L18

### Expected behavior

Reanimated shouldn't call `window.performance` in the constructor. This is the solution:

```diff
  constructor() {
    this.timeProvider = process.env.JEST_WORKER_ID
      ? { now: () => Date.now() }
-     : window.performance;
+     : { now: () => window.performance.now() };
  }
```

This way, `window.performance` only gets called when `JSReanimated.getTimestamp()` is called.

Create a next.js app, install reanimated v2. 

Let me know if this is necessary. It can take a bit of work to set up a repro for this (Next.js project with Reanimated is a lot of config), and the solution is pretty clear. Note that this is a regression from `2.0.0-rc.0`, which only called `window.performance` inside of `getTimestamp()`. I'm using the solution above with `patch-package`, and it works as expected.

See the [diff here](https://github.com/software-mansion/react-native-reanimated/compare/2.0.0-rc.1...2.0.1#diff-72fe2cb1e232789234c97ae2436d8096d352b0daca81a228c14730d0063e456c).

## Package versions
 
- React Native: Expo SDK 41
- React Native Reanimated: 2.0.0
- NodeJS: n/a
- Xcode: n/a
- Java & Gradle: n/a

## Affected platforms

- [ ] Android
- [ ] iOS
- [x] Web

nandorojo:
## Description

Fix #1890

## Changes

Don't call `window` in the constructor of JS Reanimated, since this isn't available on the server. 

One-line change, which reverts back to what was already there in `2.0.0-rc.0`. Let me know if this is needed, but given that it's simply a regression patch, I figure it isn't.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/js-reanimated/JSReanimated.ts
+++ b/src/reanimated2/js-reanimated/JSReanimated.ts
@@ -13,9 +13,10 @@ export default class JSReanimated {
   timeProvider = null;
 
   constructor() {
-    this.timeProvider = process.env.JEST_WORKER_ID
-      ? { now: () => Date.now() }
-      : window.performance;
+    this.timeProvider = {
+      now: () =>
+        process.env.JEST_WORKER_ID ? Date.now() : window.performance.now(),
+    };
   }
 
   pushFrame(frame) {
--- a/src/reanimated2/js-reanimated/JSReanimated.ts
+++ b/src/reanimated2/js-reanimated/JSReanimated.ts
@@ -10,13 +10,15 @@ export default class JSReanimated {
   _renderRequested = false;
   _mapperRegistry = new MapperRegistry(this);
   _frames = [];
-  timeProvider = null;
+  timeProvider = {};
 
   constructor() {
-    this.timeProvider = {
-      now: () =>
-        process.env.JEST_WORKER_ID ? Date.now() : window.performance.now(),
-    };
+    if(process.env.JEST_WORKER_ID) {
+      this.timeProvider.now = () => Date.now();
+    }
+    else {
+      this.timeProvider.now = () => window.performance.now();
+    }
   }
 
   pushFrame(frame) {


src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -13,9 +13,10 @@ export default class JSReanimated {
   timeProvider = null;
 
   constructor() {
-    this.timeProvider = process.env.JEST_WORKER_ID
-      ? { now: () => Date.now() }
-      : window.performance;
+    this.timeProvider = {
+      now: () =>
+        process.env.JEST_WORKER_ID ? Date.now() : window.performance.now(),
+    };
```
piaskowyk, line 19: `timeProvider.now()` is called in every frame on the web implementation, so could you change this to a little bit faster version without `if` condition? Something like this:
```js
this.timeProvider = {};
if(process.env.JEST_WORKER_ID) {
  this.timeProvider.now = () => Date.now();
}
else {
  this.timeProvider.now = () => window.performance.now();
}
```

piaskowyk (changes requested):

piaskowyk (approved):

src/reanimated2/js-reanimated/JSReanimated.ts
nandorojo: Yeah that makes more sense, I went with a ternary to stay consistent. Thanks for taking this on and making the fix!

PR #1924, opened 2021-04-08
Added information about webpack configuration

piaskowyk:
## Description

Added information about webpack configuration.

image


piaskowyk:
I wonder if giving the whole file is too much information, but in fact, the whole file is quite important. 🤔

jakub-gonet (approved):
Looks alright, I'd add some info about limitations and differences between web and native implementation.

piaskowyk:
Unfortunately, I don't have enough time at this moment.

piaskowyk:
Good point, but I will add this information next time. Unfortunately, I don't have enough time at this moment.

PR #1935, opened 2021-04-14
Added support for `@quilted/react-testing`

piaskowyk:
## Description

Added support for `@quilted/react-testing` - used by Shopify.
`@testing-library/react-hooks` uses `render()`
`@quilted/react-testing` uses `mount()`
Both methods `render()` and `mount()` returns objects with different structure.

## Example code

```js
import Animated, { useSharedValue, useAnimatedStyle, withTiming, } from '../src/';
import { withReanimatedTimer, advanceAnimationByTime, } from '../src/reanimated2/jestUtils';
import { mount } from '@quilted/react-testing';

const PinCircle = ({filled = false}) => {
  const animatedValue = useSharedValue(0);
  useEffect(() => {
    animatedValue.value = withTiming(filled ? 0 : 1, {duration: 250});
  }, [animatedValue, filled]);

    const scale = animatedValue.value;
    return { transform: [{scale}] };
  });

  return (
    <Animated.View
      testID="PinCircle/InnerCircle"
      style={animatedStyle}
    />
  );
}

it('filled prop is true', () => {
  withReanimatedTimer(() => {
    const wrapper = mount(<PinCircle filled={true} />);
    advanceAnimationByTime(250);
    const InnerView = wrapper.find(Animated.View, {testID: 'PinCircle/InnerCircle'});
    expect(InnerView).toHaveAnimatedStyle({transform: [{scale: 0}]});
  });
});

it('filled prop is false', () => {
  withReanimatedTimer(() => {
    const wrapper = mount(<PinCircle filled={false} />);
    advanceAnimationByTime(250);
    const InnerViewAfter = wrapper.find(Animated.View, {testID: 'PinCircle/InnerCircle'});
    expect(InnerViewAfter).toHaveAnimatedStyle({transform: [{scale: 1}]});
  });
});

```

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/__tests__/Animation.test.js
+++ b/__tests__/Animation.test.js
@@ -96,8 +96,8 @@ describe('Tests of animations', () => {
       expect(view).toHaveAnimatedStyle(style);
 
       fireEvent.press(button);
-      advanceAnimationByTime(260);
-      style.width = 46.08; // value of component width after 260ms of animation
+      advanceAnimationByTime(250);
+      style.width = 78.22; // value of component width after 250ms of animation
       expect(view).toHaveAnimatedStyle(style);
     });
   });
@@ -109,9 +109,9 @@ describe('Tests of animations', () => {
       const button = getByTestId('button');
 
       fireEvent.press(button);
-      advanceAnimationByFrame(10);
-      // value of component width after 10 frames of animation
-      expect(view).toHaveAnimatedStyle({ width: 16.588799999999996 });
+      advanceAnimationByFrame(20);
+      // value of component width after 20 frames of animation
+      expect(view).toHaveAnimatedStyle({ width: 99.68 });
     });
   });
 
@@ -124,8 +124,8 @@ describe('Tests of animations', () => {
       const button = getByTestId('button');
 
       fireEvent.press(button);
-      advanceAnimationByTime(260);
-      style.width = 46.08; // value of component width after 260ms of animation
+      advanceAnimationByTime(250);
+      style.width = 78.22; // value of component width after 250ms of animation
       expect(view).toHaveAnimatedStyle(style, true);
     });
   });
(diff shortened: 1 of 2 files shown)


src/reanimated2/jestUtils.ts
```diff
@@ -6,34 +6,67 @@ let config = {
...
+  const styleObject = received.props.style;
+  let currentStyle = {};
+  if (Array.isArray(styleObject)) {
+    received.props.style.forEach((style) => {
+      if (isAnimatedStyle(style)) {
+        currentStyle = Object.assign(
+          {},
+          currentStyle,
+          getAnimatedStyleFromObject(style)
+        );
```
jakub-gonet, line 27: Why we use `Object.assign` here? This is Jest so we should be able to use spread too.

src/reanimated2/jestUtils.ts
```diff
@@ -6,34 +6,67 @@ let config = {
...
+
+const checkEqual = (currentStyle, expectStyle) => {
+  if (Array.isArray(expectStyle)) {
+    if (expectStyle.length !== currentStyle.length) return false;
+    for (let i = 0; i < currentStyle.length; i++) {
+      if (!checkEqual(currentStyle[i], expectStyle[i])) {
+        return false;
+      }
+    }
+  } else if (typeof currentStyle === 'object') {
```
jakub-gonet, line 53: Can `currentStyle` be null? Its typeof is also object, so I'm making sure here that we won't throw when comparing.

__tests__/Animation.test.js
```diff
@@ -97,7 +97,7 @@ describe('Tests of animations', () => {
 
       fireEvent.press(button);
       advanceAnimationByTime(260);
-      style.width = 46.08; // value of component width after 260ms of animation
+      style.width = 56.5688; // value of component width after 260ms of animation
```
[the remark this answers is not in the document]
piaskowyk: In last version I computed time in this way: `MockDate.set(new Date(Date.now() + frameTime));`
frameTime = 16.666, but MockDate.set() lossed 0.666 ms in every frame (rounding to integer). In additional I noticed that first frame was started with time delta = 0. In effect I had delay (16.666 + 0.666 * frameIndex)ms

Szymon20000 (approved):

PR #1944, opened 2021-04-19
Make mockdate a dev dependency

Andreyco:
`mockdate` is really a dev dependency, no reason to install it in host project.


Szymon20000:
Mockdate library is required for unite testing.

Issue #1955, opened 2021-04-22
Weird error occurs rarely


heroic:
## Description

On very rare occasions, we're getting a crash report in reanimated.

### Expected behavior

Should not crash

### Actual behavior & steps to reproduce

```
java.lang.RuntimeException: java.lang.ClassNotFoundException: com.swmansion.reanimated.AndroidErrorHandler
        at com.facebook.react.bridge.DefaultNativeModuleCallExceptionHandler.handleException(DefaultNativeModuleCallExceptionHandler.java:20)
        at com.facebook.react.devsupport.DisabledDevSupportManager.handleException(DisabledDevSupportManager.java:164)
        at com.facebook.react.bridge.CatalystInstanceImpl.onNativeException(CatalystInstanceImpl.java:721)
        at com.facebook.react.bridge.CatalystInstanceImpl.access$1300(CatalystInstanceImpl.java:47)
        at com.facebook.react.bridge.CatalystInstanceImpl$NativeExceptionHandler.handleException(CatalystInstanceImpl.java:739)
        at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:29)
        at android.os.Looper.loop(Looper.java:237)
        at android.app.ActivityThread.main(ActivityThread.java:7814)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1075)

Caused by: java.lang.ClassNotFoundException: com.swmansion.reanimated.AndroidErrorHandler
        at com.swmansion.reanimated.Scheduler.triggerUI(Scheduler.java:-2)
        at com.swmansion.reanimated.Scheduler.access$100(Scheduler.java:9)
        at com.swmansion.reanimated.Scheduler$1.run(Scheduler.java:21)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
        at android.os.Looper.loop(Looper.java:237)
        at android.app.ActivityThread.main(ActivityThread.java:7814)
        at java.lang.reflect.Method.invoke(Method.java:-2)
[...]

Issue #1923, opened 2021-04-08
Reanimated 2.1.0 makes the Hermes Debugger in Flipper crash


andreialecu:
## Description

Been troubleshooting an issue with the app crashing when using the Hermes Debugger:

https://github.com/facebook/hermes/issues/491

I noticed in the device logs a warning about `libjsc.so` not being found, but since I was using Hermes that error didn't make much sense. (a similar error report is here: https://github.com/facebook/react-native/issues/25537)

I noticed 2.1.0 added support for JSC. So I reverted to 2.0.1 to see if the warning above was produced by reanimated.

Reverting to 2.0.1 got rid of the `libjsc.so` warning, and the Hermes Debugger no longer crashes.

### Expected behavior

Hermes Debugger should not crash.

### Actual behavior & steps to reproduce

Hermes Debugger crashes with RN 0.64 and reanimated 2.1.0 on an Android real device.

A native stack trace can be found in: https://github.com/facebook/hermes/issues/491

## Package versions

- React Native: 0.64
- React Native Reanimated: 2.1.0

## Affected platforms

- [x] Android
- [ ] iOS
- [ ] Web

mutablestudio:
I'm also dealing with this now and can't get a local AS build to run without crashing. I setup a pure RN sample project and added reanimated2 plus hermes 0.7.2, plus some of the standard node_modules. I added rea2 imports to the home screen and rendered an `Animated.View`. The app will run without any crash (just loading one home screen) on same device that crashes with my own RN app w rea2 (Pixel 3, Anrdoid 11). Sample app is using RN 64.0.

So not sure that this is a rea2 issue, but def seems to be Hermes related. Rolling back now to older Hermes versions to see if that helps.

andreialecu:
I mentioned in the initial post that Reanimated 2.0.1 works. This is related to #1842

piaskowyk:
I know the reason and working on a solution.

rplankenhorn:
@piaskowyk Any update on this?

oleksandr-dziuban:
Also waiting this fix, thanks a lot...

piaskowyk:
Hey, It will be fixed by: https://github.com/software-mansion/react-native-reanimated/pull/2047 We want to release a new version tomorrow.

MaxToyberman:
Hi ! 
@piaskowyk flipper works, but changing to debug in developer menu crashes the app.

jameskennethrobinson:
[...]

PR #1960, opened 2021-04-23
Fix runtime creation

piaskowyk:
## Description

In https://github.com/software-mansion/react-native-reanimated/pull/1842 we use runtime factory from React Native. This factory can decorate runtime by chrome debugger - [code here](https://github.com/facebook/react-native/blob/dc80b2dcb52fadec6a573a9dd1824393f8c29fdc/ReactCommon/hermes/executor/HermesExecutorFactory.cpp#L165-L166).
```c++
facebook::hermes::inspector::chrome::enableDebugging(std::move(adapter), "Hermes React Native");
```
This line registers runtime in the Metro server with the name `Hermes React Native`. When we use the same runtime factory to create reanimated runtime, we override React Native runtime in the Metro server because both runtimes have the same name. This happens just with Hermes runtime, so I changed the place where we create runtime and I moved the creation of Hermes runtime to C++ part. Now Metro server can see both runtimes. 

Output form Metro server http://localhost:8081/json
<details>
<summary>json data</summary>

```json
[
  {
    "id": "0-2",
    "description": "com.swmansion.reanimated.example",
    "title": "Reanimated Runtime",
    "faviconUrl": "https://reactjs.org/favicon.ico",
    "devtoolsFrontendUrl": "chrome-devtools://devtools/bundled/inspector.html?experiments=true&v8only=true&ws=%5B%3A%3A%5D%3A8081%2Finspector%2Fdebug%3Fdevice%3D0%26page%3D2",
    "type": "node",
    "webSocketDebuggerUrl": "ws://[::]:8081/inspector/debug?device=0&page=2",
    "vm": "Hermes"
  },
  {
    "id": "0-1",
    "description": "com.swmansion.reanimated.example",
    "title": "Hermes React Native",
    "faviconUrl": "https://reactjs.org/favicon.ico",
    "devtoolsFrontendUrl": "chrome-devtools://devtools/bundled/inspector.html?experiments=true&v8only=true&ws=%5B%3A%3A%5D%3A8081%2Finspector%2Fdebug%3Fdevice%3D0%26page%3D1",
    "type": "node",
    "webSocketDebuggerUrl": "ws://[::]:8081/inspector/debug?device=0&page=1",
    "vm": "Hermes"
  },
  {
    "id": "0--1",
    "description": "don't use",
    "title": "React Native Experimental (Improved Chrome Reloads)",
    "faviconUrl": "https://reactjs.org/favicon.ico",
    "devtoolsFrontendUrl": "chrome-devtools://devtools/bundled/inspector.html?experiments=true&v8only=true&ws=%5B%3A%3A%5D%3A8081%2Finspector%2Fdebug%3Fdevice%3D0%26page%3D-1",
    "type": "node",
    "webSocketDebuggerUrl": "ws://[::]:8081/inspector/debug?device=0&page=-1",
    "vm": "don't use"
  }
]
```

</details>

I decided to use `getDefaultJSExecutorFactory()` [method from React Native](https://github.com/facebook/react-native/blob/dc80b2dcb52fadec6a573a9dd1824393f8c29fdc/ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManagerBuilder.java#L300), to avoid the compilation of Reanimated AAR's files for every JS engine. This a little tricky because I need access to the private method unfortunately but thanks to this we still need just 3 AARs. In additionally this PR should automatically add support for V8 runtime because the library with the V8 engine for React Native replacing the body of `getDefaultJSExecutorFactory()` method to method returning just V8 runtime - [code here](https://github.com/Kudo/react-native-v8/blob/master/src/patches/build_with_v8.patch#L70-L109).
This PR is also the first big step to make Reanimated debuggable by Flipper or chrome debugger.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/1955
Fixes https://github.com/software-mansion/react-native-reanimated/issues/1923

Update about V8:
I can start app with V8 runtime and use V8 as Reanimated runtime, but I found the issue in `callWithThis()` - this not set `jsThis` properly [here](https://github.com/software-mansion/react-native-reanimated/blob/%40piaskowyk/flipper/Common/cpp/SharedItems/MutableValue.cpp#L46), in effect ProxySetter and SharedValue don't work. This looks like bug in [V8 implementation](https://github.com/Kudo/react-native-v8).

## Checklist

- [x] Ensured that CI passes


android/src/main/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -73,63 +78,88 @@ public void receiveTouches(String eventName, WritableArray touches, WritableArra
...
+          .getDeclaredMethod(
+            "getDefaultJSExecutorFactory",
+            String.class,
+            String.class,
+            Context.class
+          );
+        getDefaultJSExecutorFactory.setAccessible(true);
+        factory = (JavaScriptExecutorFactory) getDefaultJSExecutorFactory.invoke(
+          reactInstanceManagerBuilder,
+          "Reaniamted",
```
andreialecu, line 142: Seems there's a typo here and on the next line

rplankenhorn:
@piaskowyk I tried pulling this PR down locally and running it but it didn't seem to fix the issue on Android.  I still am getting a crash when connecting the Hermes Debugger in Flipper.

android/src/main/JNI/Android.mk
```diff
@@ -21,11 +21,12 @@ LOCAL_C_INCLUDES += $(LOCAL_PATH)/../Common/cpp/headers/Registries
...
 LOCAL_C_INCLUDES += $(LOCAL_PATH)/../Common/cpp/headers/SpecTools
 LOCAL_C_INCLUDES += $(LOCAL_PATH)/../Common/cpp/headers/Tools
+LOCAL_C_INCLUDES += $(HERMES_ENGINE)/android/include
 
 LOCAL_CFLAGS += -DONANDROID -fexceptions -frtti
 
-LOCAL_STATIC_LIBRARIES := libjsi callinvokerholder
-LOCAL_SHARED_LIBRARIES := libfolly_json libfbjni libreactnativejni
+LOCAL_STATIC_LIBRARIES := libjsi callinvokerholder libhermes-inspector
+LOCAL_SHARED_LIBRARIES := libhermes libfolly_json libfbjni libreactnativejni
```
Kudo, line 29: just curious that if adding libhermes here, will mark libhermes.so as NEEDED in libreanimated.so again?
For JSC apps to load libreanimated.so, there should be errors as libhermes.so is not packed into apk.

btw thank you @piaskowyk for the awesome work trying to decouple hermes in reanimated 👍

Szymon20000: I think it won't work because RN doesn't provide libhermes.so if you choose to use jsc.

android/CMakeLists.txt
```diff
@@ -136,6 +144,7 @@ if(${FOR_HERMES})
             ${PACKAGE_NAME}
             ${LOG_LIB}
             ${HERMES_LIB}
+            ${HERMES_INSPECTOR_LIB}
```
Szymon20000, line 147: I believe Hermes inspector is available only in debug build so we cannot merge this in the current shape.

piaskowyk:
I want to back to this in the future.

Kudo:
@piaskowyk i fixed the v8 `callWithThis` issue [you reported](https://github.com/Kudo/react-native-v8/issues/64#issuecomment-834213313). can you help to revisit the possibility of adding v8 support again when you get a chance?

here are some of my proposals:

1. add v8 support [as my fork](https://github.com/Kudo/react-native-reanimated/commit/4b069950c4b75b836350b1803485819939ce632b) and also ship v8 aar inside reanimated package.
2. if reanimated can support user to build from source, for people want to use v8 + reanimated, they should build reanimated from source.
3. i think your #1842 approach is still good for jsc and v8. maybe you can use `makeHermesRuntime` for hermes engine only and runtime factory approach as a generic solution for both jsc and v8.
4. propose a `RuntimeFactory` from react-native core, so that we don't need differentiate hermes, jsc, and v8.

jzxchiang1:
Just wanted to add that @Kudo has been extremely helpful, and I've tried his `callWithThis` fix and it finally makes reanimated v2 work for Android v8.

This would be huge for all of us v8 users. At a minimum, bundling the v8 AAR inside reanimated would be a great first step.

PR #1976, opened 2021-04-29
[v1] Name exported animation functions

marklawlor:
## Description

When used with NextJS, this library produces this warning on compile

```
warn  - ./node_modules/react-native-reanimated/lib/reanimated1/animations/timing.js
Anonymous function declarations cause Fast Refresh to not preserve local component state.
Please add a name to your function, for example:

Before
export default function () { /* ... */ }

After
export default function Named() { /* ... */ }

A codemod is available to fix the most common cases: https://nextjs.link/codemod-ndc


```
## Changes

Named the default exported function

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated1/animations/decay.js
+++ b/src/reanimated1/animations/decay.js
@@ -45,8 +45,10 @@ const procDecay = proc(
     decay(clock, { time, velocity, position, finished }, { deceleration })
 );
 
-export default (
+export default function decayAnimation(
   clock,
   { time, velocity, position, finished },
   { deceleration }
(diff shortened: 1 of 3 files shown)


nandorojo:
The spring and decay files also need to have their default exports named. Also, there's an extra space after `timing`.

marklawlor:
Good catch, updated PR.

jakub-gonet (approved):
Thanks!

PR #1987, opened 2021-05-03
Add sideEffects: false to package.json

marklawlor:
## Changes

Add sideEffects: false to package.json

## Description

When used with react-native-web this library struggles with Webpack tree-shaking

For example, this code adds 61kB to the bundle

```
import { useSharedValue } from 'react-native-reanimated';

function SomeComponent() {
  const sharedVal = useSharedValue(0);
  return (
    <Button
      onPress={() => (sharedVal.value = Math.random())}
      title="Randomize"
    />
  );
}
```

By adding `sideEffects: false` to the package.json it provides webpack an optimisation hint that it can tree-shake at the module level. You can read more about this here https://webpack.js.org/guides/tree-shaking/

This reduces the react-native-reanimated's bundle size to 35kB 

 **Before adding sideEffects**

_home_mark_Hundredpoints_core_packages_ui-framework-example_ next_analyze_client html

**After adding sideEffects**

_home_mark_Hundredpoints_core_packages_ui-framework-example_ next_analyze_client html (1)

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "main": "lib/Animated.js",
   "module": "lib/Animated",
   "react-native": "src/Animated",
+  "sideEffects": false,
   "source": "src/Animated",
   "types": "react-native-reanimated.d.ts",
   "files": [


marklawlor:
Note to the reviewer: I haven't tested whether the entire codebase is side-effect free.

Let me know if any files have side effects or cannot work in isolation and I'll add them to the exception list

jakub-gonet:
For Reanimated v1 there are definitely proc'ed things that can't be added (and AnimatedNode as standalone thing).
This is an incomplete list but those should be added to exceptions:
- https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated1/core/AnimatedNode.js
- https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated1/derived/diff.js
- https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated1/derived/color.js

marklawlor:
@jakub-gonet I don't believe those files are needed, as their sideEffects are at the statement level. Webpack only uses the package.json sideEffects flag for tree-shaking at the module level. https://webpack.js.org/guides/tree-shaking/#clarifying-tree-shaking-and-sideeffects

An example of module level side effects is you cannot import import X unless you import Y first (or the import X would fail if Y was removed). In which case Y must have side effects. An example of Y imports would be polyfills and modules that creates a store on variables outside the module scope (eg global / window)

jakub-gonet:
From docs you linked:
>A "side effect" is defined as code that performs a special behavior when imported, other than exposing one or more exports. An example of this are polyfills, which affect the global scope and usually do not provide an export.

> It says ("sideEffects" property): "If no direct export from a module flagged with no-sideEffects is used, the bundler can skip evaluating the module for side effects.".

In Reanimated we register certain things in the native side on import (e.g. when using `proc`) as well as modifying `global` in v2. Could you elaborate more on "statement level" side effects? I don't see any difference between those and "module level" side effects since all code belongs to some module.

jakub-gonet:
@marklawlor 🏓

marklawlor:
@jakub-gonet Hey sorry, I'm going to close this PR as we stopped using this library and unfortunately don't have the capacity to PR the fixes we need. 

In regards to your question: A module is an entire file. If you have a function that modifies a variable inside the same file that is not causing a side effect at the module level as both the function and the variable are in the same module. So like you pointed out this is more for the files that add to `global`. At least that is my understanding of it

Unfortunately even with the sideEffect flag, webpack seems to de-optimise our bundles and they triple in sizes. We tried using [webpacks's analyzer](https://webpack.github.io/analyse/) to unravel the dependency graph but ran out of capacity. We saw some success in simply breaking up the larger files into smaller chunks. Eg instead of hooks.js, putting each hook in a separate file.

@nandorojo Pinging you as you may want to continue on from here.

nandorojo:
Framer motion drastically reduced its bundle size in a similar endeavor, which brought Moti's bundle size down like 95%. I believe they set sideEffects: false and used some sort of roll up solution for tree shaking.

As far as I know, a side effect is defined as something that is not self contained. If a file can be imported and doesn't edit any global JS variables, etc., then it's side effect free. 

As far as reanimated goes, I think it breaks RNW's tree shaking, because it uses `Animated.View`, as opposed to just exporting a variable named `AnimatedView`. It's my understanding that the entire `Animated` variable gets imported right now on web, but I’m honestly not sure. 

Webpack's tree shaking seems really hard to optimize for in my experience. I always make sure to export individual variables, and to never re-export variables like `Animated.View`. That said, I’m not an expert at this. I've just kind of followed what I've seen works best, but I don't really know why it works.

jakub-gonet:
> In regards to your question: A module is an entire file. If you have a function that modifies a variable inside the same file that is not causing a side effect at the module level as both the function and the variable are in the same module. So like you pointed out this is more for the files that add to global. At least that is my understanding of it

So I understand it a bit differently - if you're modifying some variable in module A and you import A in module B that depends in some way on that variable then you're relying on the side effect of importing this module and this cannot be tree-shaken. Additionally, in [my comment](https://github.com/software-mansion/react-native-reanimated/pull/1987#issuecomment-831116637) above I mentioned some modules that can't be marked as side-effects free. When importing those modules we register procs on the native side. This needs to be performed exactly once, that's why we can't allow Webpack to tree-shake those modules (even if they're not used anywhere else).

> As far as reanimated goes, I think it breaks RNW's tree shaking because it uses Animated.View, as opposed to just exporting a variable named AnimatedView. It's my understanding that the entire Animated variable gets imported right now on the web, but I’m honestly not sure.

Yes, but we can't do anything with it if we want to keep it backward compatible.

To expand on tree-shaking: Webpack can't optimize huge exported objects because even if you're using only property A it can use other properties underneath. Webpack can't also traverse the graph of used properties since those can be added at runtime.

nandorojo:
https://github.com/software-mansion/react-native-reanimated/blob/master/src/reanimated2/core.ts

I believe that these edits of `global` are example of side effects. `expo-module-scripts` uses `.fx.ts` files to optimize for those with side effects. expo-module-scripts is also optimized for tree shaking and low bundle size (I use it for Dripsy).

> Yes, but we can't do anything with it if we want to keep it backward compatible.

Could you make a separate file of imports, without introducing a breaking change? Something like this:

```jsx
import { createAnimatedComponent } from 'react-native-reanimated/modules'
```

This way, current users won't have to change their imports, but we could have more optimized exports as an option as well.

PR #1988, opened 2021-05-03
Add adaptor to createAnimatedComponent()

wcandillon:
Here we can use an adaptor via setNativeProps to animate component outside the React lifecycle.
The goal is to still be able to use the benefit of the Animated API even thought there are no performance benefits.

A common example is to animate OpenGL node (bare webgl, gl-react, or react-three-fiber) on the JS thread:

```tsx
import { Node } from "gl-react";

// In gl-react, updating the gl API outside React is done via setDrawProps()
// In threejs, we would have a call on a prop by prop basis.
// In webgl, we would use the gl object directly.
const AnimatedNode = Animated.createAnimatedComponent(Node, {
  setNativeProps: (ref, props) => {
    ref.setDrawProps(props);
  },
});
```

## Checklist

- [x] Updated TS types
- [x] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -251,11 +251,11 @@ declare module 'react-native-reanimated' {
     }
     export function createAnimatedComponent<P extends object>(
       component: ComponentClass<P>,
-      options: Options<P>
+      options?: Options<P>
     ): ComponentClass<AnimateProps<P>>;
     export function createAnimatedComponent<P extends object>(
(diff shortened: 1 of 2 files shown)


wcandillon:
Here is an example of using such an API with WebGL: https://gist.github.com/wcandillon/897fc6d3ef22940c952ff1bce8c38268

Issue #1826, opened 2021-03-14
V2.0.0-rc.0 different thread error when using hooks provided in reanimated v2


haibert:
### issue: 
Getting the following error anytime  a reanimated hook is used within the useAnimatedStyle hook. 

![Simulator Screen Shot - iPhone 12 - 2021-03-14 at 11 35 20](https://user-images.githubusercontent.com/31780749/111080202-9c89eb80-84ba-11eb-977a-05ba98de117c.png)

### Description
I am using reanimted v2.0.0-rc.0 because thats the version that runs with expo currently. The entire app crashes when the component mounts if there is a reanimated hook such as withTiming nested within the useAnimatedStyle hook.

Tell us what's happening here.
If I nest a reanimated hook within another reanimated hook I get the error. Other wise there is no crash. 

### Screenshots
<img width="679" alt="Screen Shot 2021-03-14 at 11 54 27 AM" src="https://user-images.githubusercontent.com/31780749/111080529-21293980-84bc-11eb-8195-f364c2d35c6a.png">

<img width="478" alt="Screen Shot 2021-03-14 at 11 55 23 AM" src="https://user-images.githubusercontent.com/31780749/111080560-3900bd80-84bc-11eb-9468-f05ac62430ce.png">

### Steps to reproduce
1. install reanimated V2.0.0-rc.0
2. just launch the app and try out the code provided below. 
3. Im also using react navigation 5 I'm not sure if that matters.

### Expected behavior
Expected behavior is for there to be no crashes and the animation to carry out. 

### Actual behavior
App crashes right when the component mounts. 

## Snack or minimal code example

import { View, StyleSheet, TouchableWithoutFeedback } from 'react-native'
import React from 'react'
import Animated, {
    useSharedValue,
    useAnimatedStyle,
    withTiming,
} from 'react-native-reanimated'

const Transitions = () => {
    const animation = useSharedValue(0)

    const animationStyle = useAnimatedStyle(() => {
        return {
            transform: [
                {
                    translateX: withTiming(
                        animation.value,
                        {
                            duration: 1000,
                        },
                        () => {
                            animation.value = 0
                        }
                    ),
                },
            ],
        }
    })

    return (
        <View style={styles.container}>
            <TouchableWithoutFeedback
                onPress={() => {
                    animation.value = 200
                }}
            >
                <Animated.View style={[styles.box, animationStyle]} />
            </TouchableWithoutFeedback>
        </View>
    )
}

[...]

PR #2034, opened 2021-05-18
Fix hardcrash caused by #1826

thomas-shares:
## Description
This is a fix for issue https://github.com/software-mansion/react-native-reanimated/issues/1826 and  https://github.com/GetStream/stream-chat-react-native/issues/640 

Issue 1826 described the solution for this problem and the symptoms I was seeing are described in issue 640.

## Changes
Make copy of objects with spread operator as per the description in issue 1826
### Before
Before I would get a hard crash of my App.
### After
After the fix my app works again

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/Hooks.ts
+++ b/src/reanimated2/Hooks.ts
@@ -306,7 +306,7 @@ function styleUpdater(
 
   // calculate diff
   const diff = styleDiff(oldValues, newValues);
-  state.last = Object.assign({}, oldValues, newValues);
+  state.last = { ...oldValues, ...newValues };
 
   if (Object.keys(diff).length !== 0) {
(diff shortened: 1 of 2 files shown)


piaskowyk:
Hey @thomas-shares 👋
Thanks for your commitment and report the issue to us - this is really helpful  🙌  We found the source of the problem and fixed it here: https://github.com/software-mansion/react-native-reanimated/pull/2161 It happened because babel was changing Object.assign() to own implementation and this function wasn't a worklet.

thomas-shares:
Hi @piaskowyk,
Your welcome and I love the work you all do on this project. Thank you for all your hard work!!!

PR #2047, opened 2021-05-26
Fix runtime creation

piaskowyk:
## Description

Replacement for PR https://github.com/software-mansion/react-native-reanimated/pull/1960

## Checklist

- [x] Ensured that CI passes


android/src/main/cpp/NativeProxy.cpp
```diff
@@ -6,6 +6,8 @@
 #include <react/jni/ReadableNativeArray.h>
 #include <react/jni/ReadableNativeMap.h>
 #include <jsi/JSIDynamic.h>
+#import <hermes/hermes.h>
+#import <jsi/JSCRuntime.h>
```
Szymon20000, line 10: Doesn't it have to be in the if?

Szymon20000 (approved):

PR #2054, opened 2021-05-27
Bump to 2.2.0

Szymon20000:


jakub-gonet (approved):
You can also run pod install to bump version in the example app

PR #2058, opened 2021-05-27
Layout Reanimation

Szymon20000:
## Description

Introduce the very first version of Layout Animations in Reanimated

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -268,16 +268,16 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
             size_t count
             ) -> jsi::Value {
 
-          jsi::Value jsThis = rt.global().getProperty(rt, "jsThis");
-          std::string workletLocation = jsThis.asObject(rt).getProperty(rt, "__location").toString(rt).utf8(rt);
+         // jsi::Value jsThis = rt.global().getProperty(rt, "jsThis");
+         // std::string workletLocation = jsThis.asObject(rt).getProperty(rt, "__location").toString(rt).utf8(rt);
```
piaskowyk, line 272: Was it removed temporarily or permanently?

Common/cpp/headers/LayoutAnimations/LayoutAnimationsProxy.h
```diff
@@ -0,0 +1,33 @@
+#ifndef LayoutAnimationsProxy_h
+#define LayoutAnimationsProxy_h
```
piaskowyk, line 2: Why you use `#ifdef` instead of `#pragma once`? I noticed this in a couple of places.

Example/src/LayoutReanimation/SpringLayoutAnimation.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+
+/*
+import React, { useState } from 'react';
+import { View, Text, Button, StyleSheet } from 'react-native';
+import Animated, { AnimatedRoot, withTiming, withSpring, withRepeat, slideAppearingAnimation } from 'react-native-reanimated';
+import { transform } from '@babel/core';
+
+function Box({label, state}) {
+  let sz = 100;
```
piaskowyk, line 61: Is it needed or we can remove it?

piaskowyk (approved):

Common/cpp/SharedItems/ShareableValue.cpp
Szymon20000: I don't remember it perfectly, but there was a case in which host function was called and jsThis hadn't been
 set.

Example/src/LayoutReanimation/SpringLayoutAnimation.tsx
Szymon20000: We can remove it

PR #2063, opened 2021-05-28
[Test] test cpplint workflow

mrousavy:
don't merge, don't review.

### DIFF
--- a/.github/workflows/validate-cpp.yml
+++ b/.github/workflows/validate-cpp.yml
@@ -25,7 +25,7 @@ jobs:
       - uses: reviewdog/action-cpplint@master
         with:
           github_token: ${{ secrets.github_token }}
-          reporter: github-pr-review
+          reporter: github-pr-check
           flags: --linelength=230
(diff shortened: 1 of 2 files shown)


jakub-gonet:
[Did it work as intended?](https://github.com/software-mansion/react-native-reanimated/pull/2063/checks?check_run_id=2693561090#step:4:1)

mrousavy:
hmm, it should still be able to report.... 🤔 

Screenshot 2021-05-28 at 12 57 23

mrousavy:
Okay yeah no the GitHub token from reanimated cannot be used since it is being ran on a fork, not the original repo. I believe you can allow running GitHub actions on forks, there must be a setting either in the software-mansion organisation or the reanimated repo.

jakub-gonet:
I'd rather not allow running GH Actions on [untrusted PRs](https://www.infoq.com/news/2021/04/GitHub-actions-cryptomining/) but I think we can enable it if it's possible to only allow running on PRs from previous contributors.

PR #1879, opened 2021-03-30
Performance optimization

piaskowyk:
## Benchmarks

<details>
<summary>bezier - before 11 FPS, after 52 FPS</summary>

```js
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  Easing
} from 'react-native-reanimated';
import { View, Button } from 'react-native';
import React from 'react';

function AnimatedStyleUpdateExample(): React.ReactElement {
  const randomWidth = useSharedValue(10);

  const config = {
    duration: 500,
    easing: Easing.bezier(0.5, 0.01, 0, 1),
  };

  const itemCount = 100;

  const components = [];
  for(let i = 0; i < itemCount; i++) {
    components.push(
      <Animated.View key={i}
        style={[
          { width: 100, height: 5, backgroundColor: 'black', margin: 1 },
          useAnimatedStyle(() => {
            return {
              width: withTiming(randomWidth.value, config),
            };
          }),
        ]}
      />
    )
  }

  return (
    <View
      style={{
        flex: 1,
        flexDirection: 'column',
      }}>
      <Button
        title="toggle"
        onPress={() => { randomWidth.value = Math.random() * 350; }}
      />
      { components }
    </View>
  );
}
export default AnimatedStyleUpdateExample;
```
</details>

<details>
<summary>withTiming rotate - before 35 FPS, after 60 FPS</summary>

```js
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  withRepeat,
} from 'react-native-reanimated';
import { View, Button } from 'react-native';
import React from 'react';

const dummy = new Array(100).fill(1);
function AnimatedStyleUpdateExample() {
  return (
    <View style={{ flex: 1, flexDirection: "row", flexWrap: "wrap" }}>
      {dummy.map((_, i) => {

        const val = useSharedValue(0);
        const DEG = Math.PI * 2 * 10;
        val.value = withRepeat(
          withTiming(DEG, { duration: 5000 }),
          -1, true
        );

        const style = useAnimatedStyle(() => {
          'worklet'
          return {
            transform: [
               { rotate: val.value + 'deg' },
            ],
          };
        });

        return (
          <Animated.View
            key={i}
            style={[
              { width: 50, height: 50, borderWidth: 1 },
              style,
            ]}
          />
        );
      })}
    </View>
  );
}

export default AnimatedStyleUpdateExample;
```
</details>

<details>
[...]

PR #2096, opened 2021-06-02
Fix, replace for..of on for loop

NikitaDudin:
## Description
Application crash when using transform or color properties on a style object. Replacing for..of loop on for loop solves it problem.

### Issue
- #2120
- #2125

### DIFF
--- a/src/reanimated2/Hooks.ts
+++ b/src/reanimated2/Hooks.ts
@@ -162,7 +162,7 @@ function runAnimations(animation, timestamp, key, result, animationsActive) {
 function isAnimated(prop) {
   'worklet';
   if (Array.isArray(prop)) {
-    for (let i = 0; i < prop.length; i += 1) {
+    for (let i = 0; i < prop.length; ++i) {
       const item = prop[i];
       for (const key in item) {
         if (item[key].onFrame !== undefined) {
--- a/src/reanimated2/Hooks.ts
+++ b/src/reanimated2/Hooks.ts
@@ -163,7 +163,7 @@ function isAnimated(prop) {
   'worklet';
   if (Array.isArray(prop)) {
     for (let i = 0; i < prop.length; i += 1) {
-      const item = prop[i]
+      const item = prop[i];
       for (const key in item) {
         if (item[key].onFrame !== undefined) {
           return true;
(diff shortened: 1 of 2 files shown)


piaskowyk:
Could you provide an example code to reproduce this issue?

NikitaDudin:
@piaskowyk https://github.com/software-mansion/react-native-reanimated/pull/1879#issuecomment-853412859

mrousavy:
Looks like the JS runtime (or the babel plugin) doesn't whitelist the `iterator` as a native prop. `for..of` uses `[Symbol.iterator]`, no?

src/reanimated2/Hooks.ts
```diff
@@ -162,7 +162,8 @@ function runAnimations(animation, timestamp, key, result, animationsActive) {
 function isAnimated(prop) {
   'worklet';
   if (Array.isArray(prop)) {
-    for (const item of prop) {
+    for (let i = 0; i < prop.length; i += 1) {
+      const item = prop[i]
```
terrysahaidak, line 166:
```suggestion
      const item = prop[i];
```

piaskowyk (changes requested):
Thanks for PR 🙌
Could you just change `i += 1` to `++i`?

piaskowyk:
@NikitaDudin which version of babel and Hermes do you use?

NikitaDudin:
@piaskowyk hermes: 0.7.2 babel: 7.9.0

NikitaDudin:
> Thanks for PR 🙌
> Could you just change `i += 1` to `++i`?

Are you sure you should change to the prefix form `++i`?  May be `i++`?

piaskowyk:
> > Thanks for PR 🙌
> > Could you just change `i += 1` to `++i`?
> 
> Are you sure me should change to the prefix form `++i`? May be `i++`?

Yes, `++i` is a little faster than `i++` and `i++` is faster than `i += 1` 😅. Sure it can depend on the runtime version, but general idea is that. I think that every above expression is the same readable, but if we can choose faster - why not - what do you think? 🤔

piaskowyk:
I generally don't pick on things like that because this doesn't make a big difference, but I optimized it last time and I want to make it as faster as possible. 😅

piaskowyk (approved):

piaskowyk:
@NikitaDudin good job, thanks 💪

PR #2097, opened 2021-06-03
Remove `@next` from docs

JonnyBurger:
Now that Reanimated 2 is stable, the `@next` suffix can be removed 🎉

### DIFF
--- a/docs/docs/installation.md
+++ b/docs/docs/installation.md
@@ -33,10 +33,10 @@ Please note that Reanimated 2 doesn't support remote debugging, only Flipper can
 
 :::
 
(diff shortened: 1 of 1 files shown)


piaskowyk:
@JonnyBurger thanks for the commitment 💪  but we want to release an alpha version with a couple of new features really soon 🚀

JonnyBurger:
Okay 😁  👀

piaskowyk:
Here we go 🚀 
https://github.com/software-mansion/react-native-reanimated/releases/tag/2.3.0-alpha.1

PR #2111, opened 2021-06-07
Fix problem with margin in default animations

piaskowyk:
## Description

In default animations was a problem when the component has a margin.

Before:
image

After:
image


src/reanimated2/layoutReanimation/defaultAnimations/Slide.ts
```diff
@@ -117,14 +119,16 @@ export class SlideInUp extends BaseAnimationBuilder {
     const [animation, config] = this.getAnimationAndConfig();
     const delay = this.delayV;
 
-    return (values) => {
+    return () => {
       'worklet';
       return {
         animations: {
-          originY: delayFunction(delay, animation(0, config)),
```
Szymon20000, line 124: I think there is a problem. We should go towards values.originY instead of 0.

Szymon20000 (approved):

Issue #1947, opened 2021-04-19
[iOS] Crash when scrolling and use scroll event to animate header component : *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Mouting block is expected to not be set'


ahanusek:
## Description

App crash on iOS when use Animated.ScrollView and some interpolation based on scroll event from that list. 

```
Application Specific Information:
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Mouting block is expected to not be set'
terminating with uncaught exception of type NSException
abort() called
CoreSimulator 732.18.6 - Device: iPhone 12 (********-****-****-****-************) - Runtime: iOS 14.4 (18D46) - DeviceType: iPhone 12
Application Specific Backtrace 1:
0 CoreFoundation 0x00007fff20421af6 __exceptionPreprocess + 242
1 libobjc.A.dylib 0x00007fff20177e78 objc_exception_throw + 48
2 CoreFoundation 0x00007fff2042191f +[NSException raise:format:] + 0
3 Foundation 0x00007fff20771633 -[NSAssertionHandler handleFailureInFunction:file:lineNumber:description:] + 166
4 mobile 0x000000010c9f076b -[REANodesManager uiManager:performMountingWithBlock:] + 267
5 mobile 0x000000010cbbf8a1 -[RCTUIManagerObserverCoordinator uiManager:performMountingWithBlock:] + 561
6 mobile 0x000000010cbb920d -[RCTUIManager flushUIBlocksWithCompletion:] + 701
7 mobile 0x000000010cbb8ec9 -[RCTUIManager _layoutAndMount] + 905
8 mobile 0x000000010cbb8b34 -[RCTUIManager batchDidComplete] + 36
9 mobile 0x000000010c9efc09 -[RCTUIManager(SyncUpdates) runSyncUIUpdatesWithObserver:] + 265
10 mobile 0x000000010c9f0c90 __36-[REANodesManager performOperations]_block_invoke + 480
11 mobile 0x000000010cbc02e0 __RCTExecuteOnUIManagerQueue_block_invoke + 32
12 libdispatch.dylib 0x00007fff2010532f _dispatch_call_block_and_release + 12
13 libdispatch.dylib 0x00007fff20106508 _dispatch_client_callout + 8
14 libdispatch.dylib 0x00007fff2010c3f9 _dispatch_lane_serial_drain + 715
[...]

PR #2122, opened 2021-06-09
Add Airbnb-like header example with reanimated v2

valentinoConti:
## Description

I researched everywhere for a way to create an animation on a header, where:
- If I scroll down, the header starts hiding along the scroll until it get completely hidden.
- If I start scrolling up, the header starts to show along the scroll until its visible.

Basically I was trying to get what its shown on [this video](https://www.youtube.com/watch?v=JPx8IlfYQ-c) or on other websites, but using the version 2 of reanimated instead of the version 1.

I tried to use the diffClamp methods along with the interpolateNode but I was having a lot of problems to get that effect working and I was not able to found an example for this with reanimated v2 anywhere on the internet.

I created my own algorithm to achieve that effect and it's working amazing, even better than the ones that I found on internet using v1 because with this, you can activate the 'bounces' prop of the FlatList/ScrollView and use the refresh control and footer component without a problem. so I wanted to contribute with this in case somebody wants to create the same animation for a header.

## Screenshots / GIFs

Simulator Screen Recording - iPhone 12 - 2021-06-08 at 23 40 20

### DIFF
--- a/Example/src/AirbnbHeaderExample.tsx
+++ b/Example/src/AirbnbHeaderExample.tsx
@@ -1,0 +1,170 @@
+import React, { useRef } from 'react';
+import {
+  View,
+  StyleSheet,
+  Text,
+  FlatList,
+  NativeScrollEvent,
+} from 'react-native';
+import Animated, {
+  useSharedValue,
+  useAnimatedStyle,
+  useAnimatedScrollHandler,
+  Extrapolate,
+  interpolate,
+  withTiming,
+} from 'react-native-reanimated';
+
+// This function is copy-pastable and it's the responsible of the 'airbnb' header effect.
+const airbnbScrollHandler = (
+  ev: NativeScrollEvent,
+  animatedValueRef: Animated.SharedValue<number>,
+  a: Animated.SharedValue<number>,
+  MAX_HEIGHT: number
+) => {
+  'worklet';
+  const { y } = ev.contentOffset;
+  const diff = y - a.value;
+  const newAnimatedValue = animatedValueRef.value + diff;
+
+  if (y < ev.contentSize.height - ev.layoutMeasurement.height) {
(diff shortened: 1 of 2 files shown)


piaskowyk:
Before merge, we need to resolve this issue: https://github.com/software-mansion/react-native-reanimated/issues/2285

valentinoConti:
> Before merge, we need to resolve this issue: #2285

Fixed!

I'm pretty sure that at the time I created this PR that wasn't happening, the first thing I did was going to the code were that crash happens (on **ios/REANodesManager.m**) and commented it, and it started working fine without crashing:

```
- (BOOL)uiManager:(RCTUIManager *)manager performMountingWithBlock:(RCTUIManagerMountingBlock)block {
  RCTAssert(_mounting == nil, @"Mouting block is expected to not be set");
  _mounting = block;
  return YES;
}
```

But, because I know nothing about native code I did a research to see if I found something else to do, I found this open issue https://github.com/software-mansion/react-native-reanimated/issues/1947 with same problem and used the same fix that they say on the comments, wrap the interpolate on a withTiming with only 10 of duration, works perfectly fine now, but I don't really know if that is the way to go of if there is some issue on the native code used on the reanimated repo

Cheers!

piaskowyk:
Related issue - https://github.com/software-mansion/react-native-reanimated/pull/2681

jwajgelt:
Hi! I really like your example and after the recent changes introduced to shareable values it seems to work great on iOS on the `main` branch.
However, due to differences in how scroll events are handled on iOS and Android, this example doesn't seem to work correctly there, and I couldn't find a way to adapt it to have the same result on both platforms.

If you're still interested in getting this example merged, and have an idea how to fix it on Android, feel free to open a new PR.

Linuhusainnk:
Any examples similar in v3

Issue #2092, opened 2021-06-02
Occasional EXC_BAD_ACCESS on iOS


rpander93:
## Description

We occasionally see iOS users crashing with EXC_BAD_ACCESS KERN_INVALID_ADDRESS. It doesnt happen often and I've not been able to reproduce it myself. It does however happen persistently, with about 5 crashes on average per day for the last months.

We have a spinner component which is essentially an icon with a repeating timed interval from 0 to 360 degrees. It seems to sometimes cause the crash. We use it on a lot of screens. For some of the crashes, users visited many different screens before crashing. For other users, the crash happens already on the 1st or 2nd screen.

Here's a stack trace from Crashlytics:
```
Crashed: com.apple.main-thread
0  AppName                          0x100b1e7e8 std::__1::__hash_iterator<std::__1::__hash_node<std::__1::__hash_value_type<int, std::__1::vector<std::__1::shared_ptr<facebook::jsi::Value>, std::__1::allocator<std::__1::shared_ptr<facebook::jsi::Value> > > >, void*>*> std::__1::__hash_table<std::__1::__hash_value_type<int, std::__1::vector<std::__1::shared_ptr<facebook::jsi::Value>, std::__1::allocator<std::__1::shared_ptr<facebook::jsi::Value> > > >, std::__1::__unordered_map_hasher<int, std::__1::__hash_value_type<int, std::__1::vector<std::__1::shared_ptr<facebook::jsi::Value>, std::__1::allocator<std::__1::shared_ptr<facebook::jsi::Value> > > >, std::__1::hash<int>, true>, std::__1::__unordered_map_equal<int, std::__1::__hash_value_type<int, std::__1::vector<std::__1::shared_ptr<facebook::jsi::Value>, std::__1::allocator<std::__1::shared_ptr<facebook::jsi::Value> > > >, std::__1::equal_to<int>, true>, std::__1::allocator<std::__1::__hash_value_type<int, std::__1::vector<std::__1::shared_ptr<facebook::jsi::Value>, std::__1::allocator<std::__1::shared_ptr<facebook::jsi::Value> > > > > >::find<int>(int const&) + 87 (__hash_table:87)
1  AppName                          0x100b1ed20 std::__1::__function::__func<reanimated::StoreUser::~StoreUser()::$_0, std::__1::allocator<reanimated::StoreUser::~StoreUser()::$_0>, void ()>::operator()() + 30 (JSIStoreValueUser.cpp:30)
[...]

PR #2147, opened 2021-06-23
Fix ios crash

Szymon20000:
## Description

fixes: https://github.com/software-mansion/react-native-reanimated/issues/2092
As you can see in the stack trace from the issue static object are deallocated before StoreUser destructor is called. 
It seems like the problem only occurs when a user closes an app. However, it's still a serious problem as crashing apps are treated differently on iOS. In order to get rid of the problem, I store old static members as one object which is kept as shared_ptr instance field. 

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/Common/cpp/Tools/JSIStoreValueUser.cpp
+++ b/Common/cpp/Tools/JSIStoreValueUser.cpp
@@ -7,7 +7,7 @@ std::shared_ptr<StaticStoreUser> StoreUser::staticStoreUserData = std::make_shar
 std::weak_ptr<jsi::Value> StoreUser::getWeakRef(jsi::Runtime &rt) {
   const std::lock_guard<std::recursive_mutex> lock(storeUserData->storeMutex);
   if (storeUserData->store.count(identifier) == 0) {
-      storeUserData->store[identifier] = std::vector<std::shared_ptr<jsi::Value>>();
+    storeUserData->store[identifier] = std::vector<std::shared_ptr<jsi::Value>>();
   }
   std::shared_ptr<jsi::Value> sv = std::make_shared<jsi::Value>(rt, jsi::Value::undefined());
   storeUserData->store[identifier].push_back(sv);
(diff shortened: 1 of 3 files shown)


Common/cpp/Tools/JSIStoreValueUser.cpp
```diff
@@ -2,42 +2,41 @@
...
-std::unordered_map<int, std::vector<std::shared_ptr<jsi::Value>>> StoreUser::store;
+std::shared_ptr<StaticStoreUser> StoreUser::staticStoreUserData = std::make_shared<StaticStoreUser>();
 
 std::weak_ptr<jsi::Value> StoreUser::getWeakRef(jsi::Runtime &rt) {
-  const std::lock_guard<std::recursive_mutex> lock(storeMutex);
-  if (StoreUser::store.count(identifier) == 0) {
-    StoreUser::store[identifier] = std::vector<std::shared_ptr<jsi::Value>>();
+  const std::lock_guard<std::recursive_mutex> lock(storeUserData->storeMutex);
+  if (storeUserData->store.count(identifier) == 0) {
+      storeUserData->store[identifier] = std::vector<std::shared_ptr<jsi::Value>>();
```
WoLewicki, line 10:
```suggestion
    storeUserData->store[identifier] = std::vector<std::shared_ptr<jsi::Value>>();
```

Common/cpp/headers/Tools/JSIStoreValueUser.h
```diff
@@ -13,11 +13,16 @@ using namespace facebook;
 
 namespace reanimated {
 
+struct StaticStoreUser {
+    std::atomic<int> ctr;
```
WoLewicki, line 17: Shouldn't it have 2 spaces?

piaskowyk (approved):

PR #2153, opened 2021-06-25
Handling JSError on iOS

piaskowyk:
## Description

Now on iOS is possible to catch JSError thrown from the worklet.

## Example
```js
const a = () => {
  'worklet';
  throw new Error('exception_message')
}

  'worklet'
  if(_WORKLET) {
    try {
      a();
    } 
    catch(e) {
      console.log("SUCCESS!", e.message)
    }
  }

  return {
    width: withTiming(sv.value),
  };
});
```

### DIFF
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -39,12 +39,6 @@ void freeze(jsi::Runtime &rt, jsi::Object &obj) {
   freeze.call(rt, obj);
 }
 
-void on_sigabrt (int signum)
-{
-//    signal (signum, SIG_DFL);
-//    longjmp (env, 1);
-}
-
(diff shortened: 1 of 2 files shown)


Common/cpp/SharedItems/ShareableValue.cpp
```diff
@@ -433,4 +440,14 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
   throw "convert error";
 }
 
+std::string ShareableValue::demangleExceptionName(std::string toDemangle) {
+  int status = 0;
+  char * buff = __cxxabiv1::__cxa_demangle(toDemangle.c_str(), nullptr, nullptr, &status);
+  if (!buff)
+    return toDemangle;
```
jakub-gonet, line 447: Missing brackets.

Szymon20000 (approved):

Issue #2142, opened 2021-06-17
Problem with spread operator.


Szymon20000:
## Snack or minimal code example
https://github.com/wojtus7/reanimated-story-cards

## Package versions

- React Native:
- React Native Reanimated:
- NodeJS:
- Xcode:
- Java & Gradle:

## Affected platforms

- [ ] Android
- [ ] iOS
- [ ] Web

PR #2161, opened 2021-06-30
Fix problems with polyfills

piaskowyk:
## Description

Babel was changing some construction to non-workletized functions inside the worklet function. For example, `Object.assign()` was sometimes changed to babel implementation - this implementation wasn't worklet. That was the cause of the worklet runtime error.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/2142

Known problem:
In a scenario when we have nested worklets the inner one won't be workletized with asString property of the outer one. It may cause problems in the future when sending functions between several threads will be more common. However, I don't think we have such examples in the reanimated world right now.


Szymon20000:
I see just one problem with our solution. In a scenario when we have nested worklets the inner one won't be workletized with asString property of the outer one. It may cause problems in the future when sending functions between several threads will be more common. However, I don't think we have such examples in the reanimated world right now. I'm fine with merging in the current state. Can you add a note in the pr description about that problem?

Szymon20000 (approved):

PR #2182, opened 2021-07-06
TypeScript for animations

piaskowyk:
## Description

Refactor and transfter of animations module to TypeScript.


Szymon20000 (approved):

src/reanimated2/animation/decay.ts
```diff
@@ -0,0 +1,134 @@
...
+import {
+  Animation,
+  AnimationCallback,
+  PrimitiveValue,
+  Timestamp,
+} from './commonTypes';
+import { Platform } from 'react-native';
+
+interface DecayConfig {
+  [key: string]: any;
```
jakub-gonet, line 11: Why we allow unknown properties here?

src/reanimated2/animation/decay.ts
```diff
@@ -0,0 +1,134 @@
...
+        ) {
+          animation.current = config.clamp[1];
+          return true;
+        }
+      }
+
+      if (Math.abs(v) < VELOCITY_EPS) {
+        return true;
+      }
+      return false;
```
jakub-gonet, line 87: `return Math.abs(v) < VELOCITY_EPS`

src/reanimated2/animation/decay.ts
```diff
@@ -0,0 +1,134 @@
...
+          if (config.clamp.length !== 2) {
+            console.error(
+              `clamp array must contain 2 items but is given ${config.clamp.length}`
+            );
+          }
+        } else {
+          console.error(
+            `config.clamp must be an array but is ${typeof config.clamp}`
+          );
+        }
```
jakub-gonet, line 102: Maybe not in scope of this PR but we could invert those ifs to flatten this a bit:
```js
if (config.clamp) {
	if (!Array.isArray(config.clamp)){
		throw //typeof error
	}
	if(config.clamp.length !== 2){
		throw //length error
	}
}
```

src/reanimated2/animation/decay.ts
```diff
@@ -0,0 +1,134 @@
...
+    return {
+      onFrame: decay,
+      onStart,
+      callback,
+      velocity: config.velocity || 0,
+      initialVelocity: 0,
+      current: 0,
+      lastTimestamp: 0,
+      startTimestamp: 0,
+    } as DecayAnimation;
```
jakub-gonet, line 132: Why we have to cast it? Return type of the function should be typed at the function signature.

src/reanimated2/animation/decay.ts
```diff
@@ -0,0 +1,134 @@
...
+      animation.startTimestamp = now;
+      animation.initialVelocity = config.velocity;
+      validateConfig();
+    }
+
+    return {
+      onFrame: decay,
+      onStart,
+      callback,
+      velocity: config.velocity || 0,
```
jakub-gonet, line 127:
```suggestion
      velocity: config.velocity ?? 0,
```

src/reanimated2/animation/delay.ts
```diff
@@ -0,0 +1,95 @@
...
+      callback,
+      previousAnimation: null,
+      startTime: 0,
+      started: false,
+    } as DelayAnimation;
+  });
+}
+
+/* Deprecated section, kept for backward compatibility. Will be removed soon */
+export function delay(
```
jakub-gonet, line 88:
```js
/**
 * @deprecated Kept for backward compatibility. Will be removed soon.
 */
```

jakub-gonet: IDEs highlight JSDoc comments so we may use them.

src/reanimated2/animation/index.ts
```diff
@@ -0,0 +1,9 @@
+export * from './commonTypes';
+export * from './util';
+export * from './timing';
+export * from './spring';
+export * from './decay';
+export * from './delay';
+export * from './repeat';
+export * from './sequence';
+export * from './styleAnimation';
```
jakub-gonet, line 9: I'm not the fan of this kind of imports. We started to export more types to share them between modules and those shouldn't be used by end users. IMO, the better approach would be to explicitly list things we want to export so we have more control of it.

src/reanimated2/animation/util.ts
```diff
@@ -0,0 +1,238 @@
...
+    } else if (typeof value === 'string') {
+      prefNumberSuffOnStart(animation, value, timestamp, previousAnimation);
+      animation.onFrame = prefNumberSuffOnFrame;
+      return;
+    }
+    baseOnStart(animation, value, timestamp, previousAnimation);
+  };
+}
+
+type ConditionalyStarting<
```
jakub-gonet, line 185: I'm not fan of this name. Also, it has a typo.

src/reanimated2/commonTypes.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export interface StyleProps extends ViewStyle, TextStyle {
+  originX?: number;
+  originY?: number;
+  [key: string]: any;
+}
+
+export interface AnimatedStyle
+  extends Record<string, Animation<AnimationObject>> {
+  [key: string]: any;
```
jakub-gonet, line 43: We have the type for AnimatedStyle IIRC.

src/reanimated2/layoutReanimation/index.ts
```diff
@@ -1,3 +1,4 @@
-export * from './AnimatedRoot';
+require('./AnimatedRoot');
+export * from './AnimatedLayout';
```
jakub-gonet, line 2: Why we do that here?

src/reanimated2/animation/util.ts
```diff
@@ -182,7 +182,7 @@ export function decorateAnimation<
   };
 }
 
-type ConditionalyStarting<
+type AniamtionToDecoration<
```
jakub-gonet, line 185: You have a typo here.

PR #2183, opened 2021-07-06
Remove dependency on `fbjs`

newyankeecodeshop:
## Description

React Native has removed the dependency on `fbjs` for several releases. The library is for sharing JS code between Facebook projects, and is not intended to be used by the broader open-source community.

React Native replaced the implementation for `invariant` a couple years ago, so I used the same replacement:
https://github.com/facebook/react-native/commit/4148976a83ab96029de1b26b515ff95d3cb58c10

## Changes

- Use `invariant` instead of `fbjs/lib/invariant`
- Use `lodash.isequal` instead of `fbjs/lib/areEqual`

All tests pass

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,8 @@
   "homepage": "https://github.com/software-mansion/react-native-reanimated#readme",
   "dependencies": {
     "@babel/plugin-transform-object-assign": "^7.10.4",
-    "fbjs": "^3.0.0",
+    "invariant": "^2.2.4",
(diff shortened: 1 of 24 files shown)


jakub-gonet:
Thanks! 🍉 🍉 🍉

PR #2195, opened 2021-07-12
Introducing Keyframe-like animation definition schema

jmysliv:
## Description

To simplify the way how complex animations are defined, this PR introduces a keyframe like animation definition schema. It is inspired by the animation definition schema used in CSS and react-native-animatable. In the result user will be able to define animation using Keyframe:
```js
const enteringAnimation = new Keyframe({
    0: {
      transform: [{ scale: 0.4 }, { rotate: '45deg' }],
    },
    30: {
      transform: [{ scale: 1.5 }, { rotate: '-90deg' }],
    },
    100: {
      transform: [{ scale: 1 }, { rotate: '0deg' }],
      easing: Easing.quad,
    },
  })
```
After creating Keyframe object the user can pass it as a prop to `Animated.View` component (i.ex. as entering animation).

## Changes

- Added Keyframe animation builder

## Screenshots / GIFs

https://user-images.githubusercontent.com/48885911/125433861-622f8cec-d145-477d-a014-0b6bbff674f9.mov

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [x] Updated documentation
- [ ] Ensured that CI passes


src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -0,0 +1,182 @@
...
+        duration: duration,
+        value: value,
+        easing: easing,
+      });
+    };
+    animationKeyPoints
+      .sort((a: string, b: string) => parseInt(a) - parseInt(b))
+      .forEach((keyPoint: string, index: number) => {
+        if (parseInt(keyPoint) < 0 || parseInt(keyPoint) > 100)
+          throw Error('Keyframe should be in between range 0 - 100.');
```
piaskowyk, line 75: Could you change it to this form??
```
if (...) {
    // ...
}
```

piaskowyk (approved):
Solid PR 🚀

docs/docs/api/LayoutAnimations/ExitAnimations.md
```diff
@@ -6,7 +6,8 @@ sidebar_label: Exiting Animations
 
 In React Native during unmounting of components from the hierarchy of views, it just disappears in the next frame. However you can beautify this process using `Exiting Animations`. Reanimated make a pretty animation of disappearing of component for you.
 #### How it is possible? 
-Reanimated listen on changes in tree of views and if detect that some of component should disappear in next frame, It replaces this process with exiting animation. It is easy and fast. You can use predefined animations - examples below or you can define your own custom animation.
+Reanimated listen on changes in tree of views and if detect that some of component should disappear in next frame, It replaces this process with exiting animation. It is easy and fast. You can use predefined animations - examples below or you can define your own custom animation. If you want to create more complex exiting animation you can use [Keyframes](keyframeAnimations).
```
jakub-gonet, line 9: Why only "exiting animation"? Doesn't it support entering animation too?

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
```diff
@@ -0,0 +1,179 @@
...
+
+```js
+import { Keyframe } from 'react-native-reanimated';
+
+const keyframe = new Keyframe({
+    0: {
+      transform: [ { rotate: '0deg' }],
+    },
+    100: {
+      transform: [ { rotate: '45deg' }],
```
jakub-gonet, line 30: 🌵 Unnecessary whitespace in transform array.

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
```diff
@@ -0,0 +1,179 @@
...
+    },
+  }
+```
+
+### 3. Add middle points
+
+Between edge points, you can define middle points in which you want your object to have certain style properties.
+Remember that for every keyframe you should provide the same set of style properties!
```
jakub-gonet, line 38: What happens if you omit them? I think better UX would be to preserve them with previous ones so using keyframes like:
```js
const keyframe = new Keyframe({
    0: {
      transform: [{ rotate: '0deg' }],
    },
    45: {
        transform: [{ transformX: 20 }]
    },
    100: {
      transform: [{ rotate: '45deg' }],
    },
  }
```

would keep rotation at 0 degrees on 45 key and add `transformX` that will be present on the last frame.

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
jakub-gonet: Actually in most of the examples whitespace is inconsistent. Could you fix that, please?

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
```diff
@@ -0,0 +1,179 @@
...
+    <AnimatedComponent exiting={keyframe} >
+```
+### 2. Customize the animation
+```js
+    <AnimatedComponent exiting={keyframe.duration(3000).delay(200)} >
+```
+### 3. Make sure that your animated component is under an AnimatedLayout. If it's not then add AnimatedLayout somewhere above the component.
+```js
+    <AnimatedLayout> // +
+        <View> sth </View>
```
jakub-gonet, line 92: Please use `<Text />` if you want to show the rendering component with text. Adding text between any other component is invalid in RN and will error.

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
```diff
@@ -0,0 +1,179 @@
...
+        <View> 
+            <AnimatedComponent exiting={keyframe.duration(3000).delay(200)} >
+        </View>
```
jakub-gonet, line 95: What's the role of the empty `<View />` here?

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
```diff
@@ -0,0 +1,179 @@
...
+    </AnimatedLayout> // +
+```
+
+## Available modifiers
+The order of Modifiers doesn't matter.
```
jakub-gonet, line 100:
```suggestion
The order of modifiers doesn't matter.
```

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
```diff
@@ -0,0 +1,179 @@
...
+
+### duration
+default: 300
+How long the animation should last
```
jakub-gonet, line 104:
```suggestion
How long the animation should last.
```

react-native-reanimated.d.ts
```diff
@@ -636,6 +636,11 @@ declare module 'react-native-reanimated' {
     }): void;
     export function addWhitelistedUIProps(props: { [key: string]: true }): void;
 
+    export class Keyframe {
+      constructor(definitions: Map<number, KeyFrameProps[]>);
```
jakub-gonet, line 640: `KeyFrameProps` is defined in `commonTypes.ts` and I believe we don't include it in the d.ts so this will probably produce an error.

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -0,0 +1,199 @@
...
+export interface KeyFramePoint {
+  duration: number;
+  value: number | string;
+  easing?: EasingFn;
+}
+export interface ParsedKeyFramesDefinition {
+  initialValues: StyleProps;
+  keyframes: {
+    [key: string]: KeyFramePoint[];
+  };
```
jakub-gonet, line 20: This can be replaced with `Record<string, KeyFramePoint[]>`.

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -0,0 +1,199 @@
...
+import { withDelay, withSequence, withTiming } from '../../animations';
+import {
+  AnimationFunction,
+  EntryExitAnimationBuild,
+  IEntryExitAnimationBuilder,
+  KeyFrameProps,
+  StyleProps,
+} from './commonTypes';
+
+export interface KeyFramePoint {
```
jakub-gonet, line 11: Naming is inconsistent: this is named `KeyFramePoint` and class is named `Keyframe`. I personally prefer `Keyframe` prefix.

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -0,0 +1,199 @@
...
+}
+export class Keyframe implements IEntryExitAnimationBuilder {
+  durationV?: number;
+  delayV?: number;
+  definitions: KeyFramesDefinition;
+
+  /*
+    Keyframe definition should be passed in the constructor as the map
+    which keys are between range 0 - 100 (%) and correspond to the point in the animation progress.
+  */
```
jakub-gonet, line 34: This information wasn't in docs and it's pretty important IMHO.

jakub-gonet: The comment about the linear easing function is important as well.

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -0,0 +1,199 @@
...
+  constructor(definitions: KeyFramesDefinition) {
+    this.definitions = definitions;
+  }
+
+  private parseDefinitions(): ParsedKeyFramesDefinition {
+    /* 
+        for each style property contains an array with all its key points:
```
jakub-gonet, line 41: I don't understand this comment, should it read "Each style property contain..."?

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -0,0 +1,199 @@
...
+    const animationKeyPoints: Array<string> = Array.from(
+      Object.keys(this.definitions)
+    );
+    let previousKeyPoint = 0;
+
+    /* 
+       Because one of the assumptions is that each keyframe should contains
+       the same set of properties ( to make animation more consistent )
+       new entries in the parsedKeyFrames map can be added only in the first iteration
+       ( flag addIfNotExists is then set to true ).
```
jakub-gonet, line 60:
```suggestion
       Because one of the assumptions is that each keyframe should contain
       the same set of properties to make animation more consistent,
       new entries in the parsedKeyFrames map can be added only in the first iteration
       (flag addIfNotExists is then set to true).
```

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -0,0 +1,199 @@
...
+    */
+    const addKeyPoint = (
+      key: string,
+      duration: number,
+      value: string | number,
+      easing: EasingFn,
+      addIfNotExists: boolean
+    ) => {
```
jakub-gonet, line 68: I'd go with named arguments (i.e. an object) if we have 5 parameters here.

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -0,0 +1,199 @@
...
+              });
+            });
+          } else {
+            propsInKeyframe = propsInKeyframe + 1;
+            addKeyPoint(
+              key,
+              animationDuration,
+              keyframe[key],
+              easing,
+              index === 0
```
jakub-gonet, line 117: This function call and the one above it are really similar. You can create a partially applied function or override props in a shared object if you decide to refactor it that way.

Example:
```js
const addKeyPointWith = (key, styleProps) => addKeyPoint(
  key,
  animationDuration,
  styleProps,
  easing,
  index === 0)

// or
const baseArgs = {...} // really bad name though
addKeyPoint({...baseArgs, key, styleProp: keyframe[key]})
```

jakub-gonet (commented):
It isn't part of the 2.3.0-alpha.1 release so adding docs to that may be unnecessary.

docs/docs/api/LayoutAnimations/ExitAnimations.md
jmysliv: Yes, and in `EntryAnimations.md` I wrote "entering animation". However, to make it more clear, I'm gonna change it to "complex animation" in both cases.

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
jmysliv: It requires some changes in the parsing logic. I think preserving values with previous ones could be done quite easily. However, it will be harder to start animate new property in the middle of animation but I'm gonna think about it.

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
jmysliv: You're right, I'm gonna add it.

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
jakub-gonet: Starting from a default value for a property is the one way to go here. It's easy for most props by using zero values with few exceptions like color or opacity. I think this is quite intuitive behavior. You'll need to traverse values first and merge a zero keyframe with props that are specified later on.

\cc @piaskowyk

jmysliv: I had a similar idea. What if we require to pass all used props to zero keyframe as it contains initial values? That way we don't have to handle those edge cases with color, opacity, etc.

piaskowyk: I like this idea! We forgot about this before, but this approach is much better and more intuitive for users. Could we also add from-to syntax?
```js
const keyframe = new Keyframe({
  from: {
    transform: [{ rotate: '0deg' }],
  },
  to: {
    transform: [{ rotate: '45deg' }],
  },
});
```

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
```diff
@@ -0,0 +1,179 @@
...
+### 2. Customize the animation
+```js
+    <AnimatedComponent exiting={keyframe.duration(3000).delay(200)} >
+```
+### 3. Make sure that your animated component is under an AnimatedLayout. If it's not then add AnimatedLayout somewhere above the component.
+```js
+    <AnimatedLayout> // +
+        <View> sth </View>
+        <View> 
+            <AnimatedComponent exiting={keyframe.duration(3000).delay(200)} >
```
piaskowyk: Could you add the missing slash before closing the bracket? `<AnimatedComponent />` This is in more than just here but I don't want to create too many comments.

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
jakub-gonet: > What if we require to pass all used props to zero keyframe as it contains initial values? 

This is a great solution and we can relax this restriction in the future if we want to.

docs/docs/api/LayoutAnimations/KeyframeAnimations.md
```diff
@@ -32,10 +32,30 @@ const keyframe = new Keyframe({
   }
 ```
 
+Instead of using '0' and '100', you can define edge points using `from` and `to` keywords. The result will be the same.
```
jakub-gonet: To be consistent with the numbers above.

```suggestion
Instead of using `0` and `100`, you can define edge points using `from` and `to` keywords. The result will be the same.
```

jmysliv:
After consultation with @piaskowyk we've changed the approach a little bit. When a property is missing from the keyframe, instead of copying its value from the previous keyframe, we decided to increase the duration of the animation to the value in the next keyframe. So in such case:
```js
const animation = new Keyframe({
  from: {
    originX: 50,
    transform: [{ rotate: '45deg' }],
  },
  30: {
    transform: [{ rotate: '-90deg' }],
  },
  100: {
    originX: 0,
    transform: [{ rotate: '0deg' }],
    easing: Easing.quad,
  },
});
```
Animation to `originX: 0` will start from the beginning and not after 30% of the animation duration like in the previous approach. 

\cc @jakub-gonet

jakubdybczak (approved):

jakub-gonet:
This looks like a better approach. If I understand it correctly, do you try to "fill the gaps" between two occurrences of the given property and interpolate it between two points?

What happens if we have code like this?
```js
const animation = new Keyframe({
  0: {
    originX: 50,   // <- this property is being interpolated
    transform: [{ rotate: '45deg' }],
    easing: Easing.linear,
  },
  30: {
    transform: [{ rotate: '-90deg' }],
    easing: Easing.bezier,
  },
  100: {
    originX: 0,
    transform: [{ rotate: '0deg' }],
    easing: Easing.quad,
  },
});
```

that is if we have different easings between those points? Do every property follows the same easing and `originX` will be approaching `0` from `50` first linearly and from 30% of animation by bezier curve?

On a side note, I'm not sure about mixing `from`/`to` with numbers. We may treat it as an alias or a simplified syntax. I'm leaning towards the latter but alias gives more freedom in defining animations. What do you think?

jmysliv:
@jakub-gonet 
> This looks like a better approach. If I understand it correctly, do you try to "fill the gaps" between two occurrences of the given property and interpolate it between two points?

Yes, exactly.

> What happens if we have code like this?
> 
> ```js
> const animation = new Keyframe({
>   0: {
>     originX: 50,   // <- this property is being interpolated
>     transform: [{ rotate: '45deg' }],
>     easing: Easing.linear,
>   },
>   30: {
>     transform: [{ rotate: '-90deg' }],
>     easing: Easing.bezier,
>   },
>   100: {
>     originX: 0,
>     transform: [{ rotate: '0deg' }],
>     easing: Easing.quad,
>   },
> });
> ```
> 
> that is if we have different easings between those points? Do every property follows the same easing and `originX` will be approaching `0` from `50` first linearly and from 30% of animation by bezier curve?

No, currently property follows only this easing which is defined in the same point. So, because `originX: 0` is defined with `Easing.quad` it will be approaching `0` from `50` with such easing.

> On a side note, I'm not sure about mixing `from`/`to` with numbers. We may treat it as an alias or a simplified syntax. I'm leaning towards the latter but alias gives more freedom in defining animations. What do you think?

@piaskowyk what do you think?

Szymon20000 (approved):

piaskowyk:
New edge case:
```js
const animation = new Keyframe({
  from: {
    transform: [{rotate: '45deg'}, {scaleX: 1}, {rotate: '0deg'}],
  },
  to: {
    transform: [{rotate: '-90deg'}, {scaleX: 0.2}, {rotate: '80deg'}],
  }
});
```

piaskowyk (approved):

jakub-gonet (approved):

PR #2197, opened 2021-07-12
Basic chrome debugger support

piaskowyk:
#### Problem
Reanimated uses JSI and this makes impossible remote debugging by chrome inspector. 
#### Solution
If someone uses chrome inspector debugging, Reanimated replaces JSI bindings with pure JS implementation from the web version. This allows using chrome debugger in an application whose contains reanimated.
#### Known limitation
- reanimated 1
#### TODO:
- [x] gesture handler
- [x] fast reload
- [x] test ios
- [x] test web
- [x] test jest


Szymon20000 (approved):

nandorojo:
This is awesome. Could this PR be used to add remote debugging in general with JSI? Would be a great addition if so.

PR #2202, opened 2021-07-20
Fixed filtering nested animated styles.

### LINKED ISSUE #2188: [Android][hermes] Crashes when animated styles are nested two levels deep
## Description

My android app crashes on Hermes when `useAnimatedStyle` values are nested two levels deep.

Example:

```jsx
  // some style code...
});

// crashes on hermes - TypeError: property is not configurable
<Animated.View style={[[style]]} />

// does not crash 
<Animated.View style={[style]} />
```

Of course, our use case is not as simple as above and thus not straightforward for us to avoid nesting styles two or more levels deep.

### Expected behavior

I expect that it is possible to use nested styles (include multiple levels of nesting) the same way it works elsewhere in react native.

- Init repository: `npx crna -y`
- Configure Android to use Hermes
- Use the following code in App.js:

```js
import React from 'react';
import Animated, { useAnimatedStyle } from 'react-native-reanimated';
    return {};
  });

  return <Animated.View style={[[style]]} />;
}
```

- Run the app via `yarn android`

❌ NOK: The app crashes with the following stack trace.

```
ERROR    TypeError: property is not configurable

This error is located at:
    in RCTView (at View.js:34)
    in View (at createAnimatedComponent.js:448)
    in AnimatedComponent(View) (at createAnimatedComponent.js:459)
    in ForwardRef(AnimatedComponentWrapper) (at App.js:9)
    in App (created by ExpoRoot)
    in ExpoRoot (at renderApplication.js:45)
    in RCTView (at View.js:34)
    in View (at AppContainer.js:106)
    in RCTView (at View.js:34)
    in View (at AppContainer.js:132)
    in AppContainer (at renderApplication.js:39), js engine: hermes
```

https://github.com/Yonom/reanimated-bug-report
 
## Package versions

- React Native: 0.63.4 (also not working on 0.64)
- NodeJS: v16.4.2
- Xcode: 12.5
- Java & Gradle: AdoptOpenJDK 8, Gradle 4.1.0

## Affected platforms

- [x] Android
- [ ] iOS
- [ ] Web

jmysliv:
## Description
`flattenArray` function was not used during filtering non-animated props, so when the style property in `Animated.View` was a nested array, the app was crashing. I've changed that so know nested array can be passed as a style property.
Fixes #2188

### DIFF
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -279,10 +279,7 @@ export default function createAnimatedComponent(Component, options = {}) {
     }
 
     _attachAnimatedStyles() {
-      let styles = Array.isArray(this.props.style)
-        ? this.props.style
-        : [this.props.style];
-      styles = flattenArray(styles);
+      const styles = flattenArray(this.props.style);
       this._styles = styles;
       let viewTag, viewName;
       if (Platform.OS === 'web') {
@@ -453,7 +450,7 @@ export default function createAnimatedComponent(Component, options = {}) {
       for (const key in inputProps) {
         const value = inputProps[key];
         if (key === 'style') {
-          const styles = Array.isArray(value) ? value : [value];
+          const styles = flattenArray(value);
           const processedStyle = styles.map((style) => {
             if (style && style.viewDescriptors) {
               // this is how we recognize styles returned by useAnimatedStyle
--- a/src/createAnimatedComponent.js
+++ b/src/createAnimatedComponent.js
@@ -46,7 +46,7 @@ function hasAnimatedNodes(value) {
 
 function flattenArray(array) {
   if (!Array.isArray(array)) {
-    return array;
+    return [array];
   }
   const resultArr = [];
 


piaskowyk (approved):
🚀

thespacemanatee:
Seems to be broken again in 2.2.4

renchap:
@jmysliv I can confirm what @thespacemanatee is seeing, this crashes again on Android (not iOS) if `style` is an array, with `TypeError: property is not configurable` on 2.2.4.

Should I open a new issue for this?

markdalgleish:
I ran into the same issue, but it looks like this fix landed in [v2.3.0-alpha.2](https://github.com/software-mansion/react-native-reanimated/releases/tag/2.3.0-alpha.2) — not available in latest stable version yet (v2.2.4 at time of writing).

piaskowyk:
@markdalgleish  Probably in next week, we will release 2.3.0 as stable

Issue #1081, opened 2020-08-06
Is it possible to bind to custom events?


robertgonzales:
I'm working on an Android app that receives events from a custom hardware integration. I have some java code that sends these events over the react-native bridge so that react-native-reanimated can play animations from these events. However, receiving these events in the JS thread is slowing down the UI. I’d like to use react-native-reanimated to bind to these events without having to impede the JS thread. Basically I'm looking for a way to bind to these custom events just like I would for gestures or scrolling events.

I did some searching for examples of react-native-reanimated integrating with custom events but I couldn’t find anything. Is this even possible? Are there any good examples of this or source code that would be a good reference?

cmnstmntmn:
@robertgonzales yes, have a [read](https://reactnative.dev/docs/native-modules-android#beyond-toasts)

jakub-gonet:
Reanimated doesn't keep any list of RN events so if you're dispatching events via RN interfaces and methods it should be supported.

If you have further questions feel free to ask.

robertgonzales:
@jakub-gonet @cmnstmntmn thanks! Yeah I'll reference react-native-gesture-handler. 

I'll maybe post some code snippets here if I get everything working right so anyone else with a similar question has a clear example.

Issue #1229, opened 2020-09-15
useEvent hook not working for custom UI modules


mrbrentkelly:
## Description

The [useEvent](https://github.com/software-mansion/react-native-reanimated/blob/20ae770f73f65023278e7081ebe29d342a08c6e1/src/reanimated2/Hooks.js#L33) hook from Reanimated v2 doesn't work for events on a custom Android and iOS UI modules that use `RCTEventEmitter`. 

If events are emitted using the `EventDispatcher` from `UIManagerModule` then the `useEvent` hook works just fine.

**Works**

```java
EventDispatcher eventDispatcher = context
   .getNativeModule(UIManagerModule.class)
   .getEventDispatcher();

SomeCustomEvent event = SomeCustomEvent.obtain(/*params*/);

eventDispatcher.dispatchEvent(event);
```

**Doesn't work**

```java
context.getJSModule(RCTEventEmitter.class).receiveEvent(getId(), "SomeEvent", data);
```

The `RCTEventEmitter` approach is probably what most custom native modules use (for example [react-native-video](https://github.com/react-native-community/react-native-video/blob/master/android/src/main/java/com/brentvatne/react/ReactVideoView.java#L166)) since that's what the [docs](https://reactnative.dev/docs/native-components-android#events) say to use. So I think this might become a common gotcha if not fixed or at least documented.

> Note, this issue exists on iOS as well.

## Steps To Reproduce

1. Create new RN project and install reanimated2
2. Create a Native UI module on Android that emits events using RCTEventEmitter
3. Consume events on the JS layer using the `useEvent` hook.

### Expected behavior

Events triggered using RCTEventEmitter should be picked up by the `useEvent` hook

### Actual behavior

The worklet inside the `useEvent` hook is not triggered.

## Snack or minimal code example

I've created a new RN project that repro's the issue:

https://github.com/mrbrentkelly/reanimated2-event-bug

<img src="https://user-images.githubusercontent.com/5814579/93265805-0b52ea00-f7a1-11ea-829c-417767312d95.gif" width="200" />

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Reanimated: 2.0.0

jakub-gonet:
Sorry for the late response.

Are those events direct (sent to the `View`s)? AFAIK we can only handle that type of events.

mrbrentkelly:
Yes these are direct view events (although I would expect RE2 to work with bubbling view events also). 

[...]

PR #2207, opened 2021-07-21
Custom animated event handlers API

mateusz1913:
## Description

To enable listening to custom native events (was mentioned first in #1081) in similar way as for gestures and scroll events, this exposes two low-level hooks that can be used to create custom event handler hooks similar to `useAnimatedGestureHandler` and `useAnimatedScrollHandler`

## Changes

- Create `useHandler` hook which creates shared context and "should rebuilt worklets" value
- Expose `useHandler` and `useEvent` hooks (added TS types)
- Add example
- Add docs sections for "custom events" guide and hooks api

Run example app on Android & iOS, look for "Custom Handler Example - Pager"

## Checklist

- [X] Updated TS types
- [ ] Added TS types tests
- [X] Updated documentation
- [ ] Ensured that CI passes

Let me know, if there is sth more I need to improve, every feedback is appreciated!


terrysahaidak:
Does it really work? I mean page view really can call it from UI Thread and the worklet will be executed on UI Thread in the next frame?

I remember the very last time I checked this I couldn't understand how and when libraries call animated events and how exactly Reanimated handles them.

I would appreciate an explanation of how it works in general. Also, some confirmation it works within that single frame delay. Thank you.

Example/ios/Podfile.lock
```diff
@@ -312,9 +314,8 @@ PODS:
     - React-RCTVibration
     - ReactCommon/turbomodule/core
     - Yoga
-  - RNScreens (3.4.0):
+  - RNScreens (2.17.1):
```
jakub-gonet, line 317: Why do we downgrade screens here?

Example/src/CustomHandler/AnimatedText.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const animatedProps = useAnimatedProps(() => {
+    return { text: text.value } as unknown as TextInputProps;
+  });
+
+  return <AnimatedTextInput
+    underlineColorAndroid="transparent"
+    editable={false}
+    value={text.value}
+    style={[styles.text, style]}
+    {...{animatedProps}}
```
jakub-gonet, line 22: Can't we use `{...animatedProps}`?

docs/docs/api/useHandler.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Returns
+
+The hook returns a context that will be reused by event handlers and value that indicates whether worklets should be rebuilt.
+
+## Example
+
+```js
+function useAnimatedPagerScrollHandler(handlers, dependencies) {
+  const { context, dependenciesDiffer } = useHandler(handlers, dependencies);
```
jakub-gonet, line 44: I think second key should be named `doDependenciesDiffer` to indicate this is a boolean. Right now its not obvious if this is a function that diffs dependencies or a boolean.

mrousavy:
@terrysahaidak I believe that comes down to which thread the events get called on, no? REA just hooks into the RN event dispatching system, and as long as those events are called on the UI thread there will not be a single frame delay in that. If it's called on any native module thread though, that's a different story and requires thread hopping (not sure how that's handled)

Example/ios/Podfile.lock
mateusz1913: I only run standard `npx pod-install` and on that branch in example's `package.json` there is downgraded version of screens (which I didn't touch)

Example/src/CustomHandler/AnimatedText.tsx
mateusz1913: `{...animatedProps}` didn't work, so I changed to `animatedProps={animatedProps}` to be consistent with other examples

mateusz1913:
I did a rebase and addressed comments

piaskowyk:
Hi @mateusz1913 Thanks for your PR! 🙌  We'll review and test it before the next release.

mateusz1913:
Hi @piaskowyk, I did a rebase and resolved merge conflicts to match current src files architecture after TypeScript changes

mrbrentkelly:
Do we know if this solves https://github.com/software-mansion/react-native-reanimated/issues/1229?

mateusz1913:
@mrbrentkelly if nothing has changed (with worklet event handling logic) since last months, then I think the answer is "no".

PR #2209, opened 2021-07-22
Override libfbjni.so

piaskowyk:
## Description

I added `libfbjni.so` (in version 0.2.2 with dynamic loading of libc++.so) to reanimated aar package. Reanimated will automatically change application configuration and add Gradle option:
```
android {
  ...
  packagingOptions {
    pickFirst 'lib/**/libfbjni.so'
  }
  ...
}
```
To disable this behavior user just need to add to `app/gradle.build` the following config:
```
project.ext.reanimated = [
    enablePackagingOptions: false,
]
```

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -94,6 +94,9 @@ if(!System.getenv('IS_TRAVIS')) {
 }
 
 android {
+    buildFeatures {
+        prefab true
+    }
     compileSdkVersion safeExtGet('compileSdkVersion', 29)
     buildToolsVersion safeExtGet('buildToolsVersion', '29.0.2')
     defaultConfig {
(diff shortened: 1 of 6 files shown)


android/build.gradle
```diff
@@ -131,7 +134,7 @@ android {
     packagingOptions {
         println "Native libs debug enabled: ${debugNativeLibraries}"
         doNotStrip debugNativeLibraries ? "**/**/*.so" : ''
-        excludes = ["**/libc++_shared.so"]
+        excludes = REACT_VERSION < 65 ? ["**/libc++_shared.so"] : ["**/libc++_shared.so", "**/libfbjni.so"]
```
Szymon20000, line 137: It would be nice to also automatically exclude 'libfbjni.so' from app. Can you check if we can add SO to the exclude list of another module?

Szymon20000 (approved):

LinusU:
Seems like this change is causing EAS (Expo) builds to fail: https://github.com/expo/eas-cli/issues/607

PR #2210, opened 2021-07-22
Fixed float with scientific notation to string conversion

### LINKED ISSUE #2170: withSpring rotation with damping = 0 creates an unexpected animation
## Description

I've used the following code to create a "moving bubbles" animation similar to apple edit apps edit screen
```
  let animatedStyle = useAnimatedStyle(() => {
    shouldShow.value ? cancelAnimation(shouldShow) : null;
    let props = {
      overshootClamping: false,
      stiffness: 100,
      mass: 0.2,
      damping: 0,
    };
    return {
      transform: [
        {
          scale: shouldShow.value
            ? withSequence(withSpring(1.1), withSpring(1))
            : withSpring(1),
        },
        {
          rotate: shouldShow.value
            ? withSpring('1deg', props)
            : withSpring('0deg'),
        },
      ],
    };
  }, [shouldShow]);
```

### Expected behavior

The animation starts correctly

After about a minute, the view starts spinning uncontrollably 

https://snack.expo.io/v8sXgmW8j (press on black square and observe for about a minute)

## Package versions

- React Native: 0.64
- NodeJS: 16.0.0
- Xcode: 13
- Java & Gradle: N/A

## Affected platforms

- [ ] Android
- [X] iOS
- [ ] Web

jmysliv:
## Description

During animations, property value is often converted between string and float. When the animation value gets too small or too big, it converts to string with scientific notation. It is then interpreted as a value with suffix `e` and is converted back to float without scientific notation which results in the unexpected and rapid change of the animation value. 

Screenshot 2021-07-22 at 13 01 24

I've changed float to string conversion, so it is now converted without scientific notation.

Fixes #2170 

## Changes
- changed animation value float to string conversion

### DIFF
--- a/src/reanimated2/animations.ts
+++ b/src/reanimated2/animations.ts
@@ -17,6 +17,7 @@ export function initialUpdaterRun(updater) {
 
 /*
   converts float to string and removes the scientific notation
+  https://stackoverflow.com/questions/1685680/how-to-avoid-scientific-notation-for-large-numbers-in-javascript/46545519#46545519
 */
 function parseString(num) {
   'worklet';
--- a/src/reanimated2/animations.ts
+++ b/src/reanimated2/animations.ts
@@ -15,6 +15,32 @@ export function initialUpdaterRun(updater) {
   return result;
 }
 
+/*
+  converts float to string and removes the scientific notation
+*/
+function parseString(num) {
+  'worklet';
+  let numStr = num.toString();
+  if (Math.abs(num) < 1.0) {
+    const e = parseInt(num.toString().split('e-')[1]);
+    if (e) {
+      const negative = num < 0;
+      if (negative) num *= -1;
+      num *= Math.pow(10, e - 1);
+      numStr = '0.' + new Array(e).join('0') + num.toString().substring(2);
+      if (negative) numStr = '-' + numStr;
+    }
+  } else {
+    const e = parseInt(num.toString().split('+')[1]);
+    if (e > 20) {
+      e -= 20;
+      num /= Math.pow(10, e);
+      numStr = num.toString() + new Array(e + 1).join('0');
+    }
+  }
+  return numStr;
+}
+
 export function transform(value, handler) {
   'worklet';
   if (value === undefined) {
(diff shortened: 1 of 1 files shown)


mrousavy:
Does `toLocaleString` require Intl? If so, that function will be incredibly slow on Hermes.

piaskowyk:
@mrousavy good point! @jmysliv we need to check the performance of this solution.

jmysliv:
@mrousavy @piaskowyk I'll check the performance of `toLocaleString` conversion and custom conversion and I'll let you know what are the results.

jakub-gonet:
I think we should Intl.NumberFormat object before doing the conversion, as described [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#performance).

And if still this method will be slow we may use [this snippet from SO](https://stackoverflow.com/a/46545519) (with proper atribution of course).

jmysliv:
I've tested all three options in 100k iteration, custom conversion similar to a snippet from SO is about 100 times faster than `toLocaleString` method and about 4 times faster than `format` method when we create Intl.NumberFormat object earlier.

mrousavy:
@jmysliv did you also compare performance of Hermes to JSC?

mrousavy:
Why is the value even converted to a string in the first place? @piaskowyk

jmysliv:
@mrousavy there is some issue with intl support on Hermes (https://github.com/facebook/hermes/issues/23), so if we don't want to use any workaround I think the best and fastest option will be using [manual conversion](https://stackoverflow.com/questions/1685680/how-to-avoid-scientific-notation-for-large-numbers-in-javascript/46545519#46545519).

jakub-gonet:
@jmysliv Please add SO link if you didn't write that function by yourself:
- https://meta.stackoverflow.com/a/253622
- https://stackoverflow.blog/2009/06/25/attribution-required/

jakub-gonet (approved):

mrousavy:
Somehow I have a feeling that this code is not the fastest (especially because of the array allocation + joining) and this is a very hot spot in the codebase...

jakub-gonet:
> Somehow I have a feeling that this code is not the fastest (especially because of the array allocation + joining) and this is a very hot spot in the codebase...

And string creation if we're on that. But good point, maybe we should do some benchmarking.

jmysliv:
@mrousavy I've tested performance before and after this change and I didn't encounter any significant frame drop. 

> Somehow I have a feeling that this code is not the fastest (especially because of the array allocation + joining) and this is a very hot spot in the codebase...

I can instead of array join perform string concatenation, which as far as I know is faster. I'm also open to any other suggestions or ideas that may speed up this code.

piaskowyk (changes requested):
We need just use this https://github.com/software-mansion/react-native-reanimated/pull/2210 and change the regex rule.

jmysliv:
I'm closing this because #2238 will fix this issue without a performance drop.

PR #2216, opened 2021-07-24
Fixed crash on iOS when using AnimatedLayout and React Navigation Native

### LINKED ISSUE #2124: Animated Layout not working with react navigation (2.3.0-alpha.1)
i try run app with react navigation then get this error. but still working without react navigation

image

### Expected behavior
Run app without error

## Package versions

- React Native: 0.64.1
- React Native Reanimated: 2.3.0-alpha.1
- React Navigation Native: 5.9.4
- React Navigation Stack: 5.14.5
- Xcode: 12.4

## Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

iPinkiPanda:
## Description

On iOS when using React Navigation Native and AnimatedLayout the app would crash with an 
`NSArrayM insertObject:atIndex:] object cannot be nil` Exception
This was caused by the Snapshoter being unable to build a complete hierarchy of views as it wouldn't be able to navigate past certain viewControllers.
This fix uses `responder` instead of `superview` for its ability to know about views connected to viewControllers (including navigation controllers etc). Note that we now have to filter out UIViewControllers from the returned hierarchy as they were not included when only using `superiew`

Fixes #2124 

## Changes

- Use nextResponder instead of superView to traverse past UIViewControllers.
- Filter out UIViewControllers and subclasses from pathToWindow array.

### DIFF
--- a/ios/LayoutReanimation/REASnapshooter.m
+++ b/ios/LayoutReanimation/REASnapshooter.m
@@ -28,10 +28,12 @@ int Id = 1e9;
   
   if ([view isKindOfClass:[REAAnimationRootView class]]) {
     NSMutableArray * pathToWindow = [NSMutableArray new];
-    UIView *current = view;
+    UIResponder *current = view;
     do {
-      [pathToWindow addObject:current];
-      current = current.superview;
+      if (![current isKindOfClass:[UIViewController class]]) {
+        [pathToWindow addObject:current];
+      }
+      current = current.nextResponder;
     } while (current != windowView);
     values[@"pathToWindow"] = pathToWindow;
   }


radko93:
Might be related to #2145?

hannojg:
I've just tested the patch, using
```
"react-native-reanimated": "~2.3.0-alpha.1",
"@react-navigation/native": "^6.0.0-next.17",
"@react-navigation/stack": "^6.0.0-next.29",
"@react-navigation/core": "^6.0.0-next.17",
```

and the native stack navigator with react navigation. The issue is still happening!

hannojg:
I fixed it by adding a `nil` check to the while condition:

```
} while (current != windowView && current != nil);
```

(not pretending to know whats going on, just guessing 😅 )

hannojg:
Just spotted another issue with AnimatedLayout and react navigation native:

We have in our app an authentication flow setup, as outlined here: https://reactnavigation.org/docs/auth-flow
On iOS when the user logs in the page doesn't transition to the authenticated stack, but the app just turns black. When I remove the `LayoutAnimation` from the sign in screen everything works again.
Tell me if I should create a new reproduction/issue for this.

happyfloat:
I got the same issue, my screen turns black after transitioning back. I'm using `react-native-screens` too

happyfloat:
I just noticed, that "swipe to go back" does work, only hitting the back button will result in a black screen.

carlos-kelly:
I can also confirm that the fix in this PR with the updated `while` statement from @hannojg also fixes the crashes with LayoutAnimation and React Navigation.

wcandillon:
I can confirm @carlos-kelly's comment. This patch works in addition of @hannojg small change.

piaskowyk:
Hi, @iPinkiPanda thanks for your commitment! 🙌  Unfortunately, I need to close this PR for two reasons:
- https://github.com/software-mansion/react-native-reanimated/tree/%40szymon/remove_animated_layout
- we have already fixed it in the above PR
- we are rewriting the whole module of layout animation to add support for react-native-navigations and react-native-screens because this is a harder issue (also above PR)

PR #2219, opened 2021-07-26
fix: Cache `StoreUser` (`JSIStoreValueUser`) on a per-`RuntimeManager`-basis

mrousavy:
## Description

Until now, the `JSIStoreValueUser` class was used to cache `jsi::Value`s. This worked fine for Reanimated since there was only one separate Runtime apart from the React-JS runtime, but once another library tries to use the workletization code, this will break since `JSIStoreValueUser` is accessed from multiple runtimes. Instead, it should only be used on a per-runtime basis, so I moved the static code to instance code, and added a Shared Pointer to `RuntimeManager`.

This fixes SIGABRT crashes for when trying to use a captured variable (such as `console`) in multiple runtimes (REA, VisionCamera, react-native-multithreading, ...):

image

The above crash happened because a `jsi::Value` (`console`) cached inside `StoreUser` was accessed from the VisionCamera runtime, but the value was initially created on the REA runtime.

## Changes

* Refactored `JSIStoreValueUser` to be an instance variable of `RuntimeManager` instead of a static one.
* Clear store on `RuntimeManager` dealloc (dtor)
* Fix a `isWorkletRuntime` check which should instead strictly compare if the runtimes are exactly same

1. Create a normal function on the React-JS Thread
2. Verify that you can call that function inside of a REA UI Worklet (`useAnimatedStyle`, or `runOnUI`) using `runOnJS`
3. Verify that you can call that same function inside of a VisionCamera FP Worklet (`useFrameProcessor`) using `runOnJS`

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/Common/cpp/NativeModules/NativeReanimatedModule.cpp
+++ b/Common/cpp/NativeModules/NativeReanimatedModule.cpp
@@ -63,7 +63,7 @@ NativeReanimatedModule::NativeReanimatedModule(std::shared_ptr<CallInvoker> jsIn
                                                std::shared_ptr<ErrorHandler> errorHandler,
                                                std::function<jsi::Value(jsi::Runtime &, const int, const jsi::String &)> propObtainer,
                                                std::shared_ptr<LayoutAnimationsProxy> layoutAnimationsProxy,
-                                               PlatformDepMethodsHolder platformDepMethodsHolder) : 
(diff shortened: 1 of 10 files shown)


mrousavy:
cc @Szymon20000 @piaskowyk

piaskowyk (approved):

PR #2220, opened 2021-07-26
NodesManager: synchronize every call to `mAnimatedNodes`

mrousavy:
## Description

I've been experiencing fatal errors like these: 

image

where the app doesn't work anymore, and I believe this has to do with trying to access nodes simultaneously. This PR synchronizes all calls to `mAnimatedNodes` to prevent race conditions (duplicate calls to `createNode`) and fix the error above.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -284,11 +284,13 @@ public class NodesManager implements EventDispatcherListener {
    * iOS behavior when the app won't just crash.
    */
   public Object getNodeValue(int nodeID) {
-    Node node = mAnimatedNodes.get(nodeID);
-    if (node != null) {
-      return node.value();
+    synchronized (mAnimatedNodes) {
+      Node node = mAnimatedNodes.get(nodeID);
+      if (node != null) {
+        return node.value();
+      }
+      return ZERO;
     }
-    return ZERO;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


mrousavy:
Okay I've been trying to test the change on Android for 2 hours now but I simply can't manage to get my patched version running locally.
1. Reanimated is not distributed as source on Android, so I cannot test my change locally in node_modules. I tried to move the `android/` folder from my fork to my `node_modules/` , but that didn't work and failed to build because the relative paths are wrong and it cannot find the libraries in `CMakeLists.txt`
2. I tried to edit the files in my fork and then build a new npm package which I can then install locally (with the prebuilt .aar), but that didn't work because of gradle build issues. (`ndk.dir` not found, `compileSdkVersion` not set, `'maven'` plugin not found, and a ton of more errors)
so I gave up on that now, I can take a look at that again when Reanimated is distributed as source (which I highly recommend!) - but maybe this PR is the solution to the problem :shrug:

jakub-gonet:
> I believe this has to do with trying to access nodes simultaneously

I don't understand from what threads do you access nodes data that synchronization became a problem. The original idea was that only UI thread can manipulate nodes and I don't think this changed.
Historically, error "Animated node with ID _x_ already exists" show up only when you're initializing RN runtime two times (by executing bundle more than one time). This is an error in initialization and has nothing to do with synchronization. More specifically, we have [a global node counter](https://github.com/software-mansion/react-native-reanimated/blob/285b64a1c7e7c77884bdcf3e4ddead3d46d410d6/src/reanimated1/core/AnimatedNode.js#L8) in JS side. The native side only takes [id as a parameter](https://github.com/software-mansion/react-native-reanimated/blob/285b64a1c7e7c77884bdcf3e4ddead3d46d410d6/android/src/main/java/com/swmansion/reanimated/NodesManager.java#L315).

Also, why only Android is synchronized in this PR? iOS impl technically can suffer from [the same problem](https://github.com/software-mansion/react-native-reanimated/blob/285b64a1c7e7c77884bdcf3e4ddead3d46d410d6/ios/REANodesManager.m#L301).

Additionally, please keep in mind that synchronization is costly in CPU & time terms and hiding every node operation behind it will have a significant runtime cost.

mrousavy:
You're right, this is probably an unrelated change. Still trying to figure out how I got that error, but I believe it has something to do with the usage of `runOnJS` in my worklets. (VisionCamera worklets)

I'll investigate more and report back if I find any fixes

rodgomesc:
@mrousavy  what was the problem?

PR #2223, opened 2021-07-27
Simplify manual installation process.

Szymon20000:
## Description

If somebody has to turn off autoinstallation, they have to copy-paste our implementation of `jsExecutorFactoryForBridge` and update that code whenever we change something in the library.  To make it a bit more flexible I want to add another layer of abstraction so that our code is hidden behind the `REAJSIExecutorRuntimeInstaller` function.
## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/Example/ios/Podfile.lock
+++ b/Example/ios/Podfile.lock
@@ -9,10 +9,6 @@ PODS:
     - React-Core (= 0.64.1)
     - React-jsi (= 0.64.1)
     - ReactCommon/turbomodule/core (= 0.64.1)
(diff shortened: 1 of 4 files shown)


tomekzaw (approved):
That's exactly what we needed 👏

piaskowyk (approved):

jakub-gonet:
What about documenting this change? 😞

PR #2238, opened 2021-07-30
Prefix recognition on animation start

### LINKED ISSUE #2170: withSpring rotation with damping = 0 creates an unexpected animation

jmysliv:
## Description
Currently on every frame prefix and suffix recognition is performed which is unnecessary. I moved that logic to executes only once on animation start. I'm also changed the regex rule to match scientific notation.

Fixes #2170

## Changes
- moved prefix recognition to on animation start
- added recognizing scientific notation
- added strippedValue to animation object.


piaskowyk (changes requested):
We can do it faster with the additional field `strippedValue`.

piaskowyk (approved):

PR #2254, opened 2021-08-04
Fixed android crashing when useAnimatedStyles returns an object with nested properties.

### LINKED ISSUE #2040: Android crashing when useAnimatedStyles returns an object with nested properties.
## Description

Crashing when useAnimatedStyles returns nested properties, on Android.

The resulting returned animated style does not need to be used.

### Expected behavior

No crashes.

Crashes app completely.

``` javascript
const shadowStyles = useAnimatedStyle(() => ({
  shadowOffset: {
      height: 0,
      width: 0,
  },
}));
```

https://snack.expo.io/@dart200/useanimatedstyles-bug

## Package versions

- React Native: 0.63.3
- Also seen with Expo SDK 41

## Affected platforms

- [x] Android
- [ ] iOS
- [ ] Web

jmysliv:
## Description
During converting props to map, nested props are converted to `ReadableNativeMap`. When they are later added to the props `WritableNativeMap` by Nodes Manager error occurs, as WritableNativeMap implementation requires map to be WritableNativeMap.
```java
  @Override
  public void putMap(@NonNull String key, @Nullable ReadableMap value) {
    Assertions.assertCondition(
        value == null || value instanceof WritableNativeMap, "Illegal type provided");
    putNativeMap(key, (WritableNativeMap) value);
  }
```
The result was that the app crashed whenever there were nested props provided in `useAnimatedStyle` or `useAnimatedProps`.
Fixes #2040.

## Changes
- Changed `ReadableNativeMap` to `WritableNativeMap` during props converting

### DIFF
--- a/android/src/main/cpp/JNIHelper.cpp
+++ b/android/src/main/cpp/JNIHelper.cpp
@@ -52,7 +52,7 @@ jni::local_ref<JNIHelper::PropsMap> JNIHelper::ConvertToPropsMap(jsi::Runtime &r
       }
       else
       {
-        map->put(key, ReadableNativeMap::newObjectCxxArgs(jsi::dynamicFromValue(rt, value)));
+        map->put(key, WritableNativeMap::newObjectCxxArgs(jsi::dynamicFromValue(rt, value)));
       }
     }
   }


piaskowyk:
The problem has resolved already with this PR: https://github.com/software-mansion/react-native-reanimated/pull/2257

PR #2257, opened 2021-08-05
Support for nested objects in animations

jmysliv:
## Description
Currently, users can only animate style properties that are nested one-level deep or transform properties. Animating styles like shadowOffset's width were not supported as there were nested two levels deep. I've changed the useAnimatedStyle hook and LayoutAnimations to support nested objects.

## Changes
- changed `useAnimatedStyle` hook to support nested objects
- changed `withStyleAnimations` to support nested objects

## Screenshots / GIFs

Here is an example of how we can animate shadowOffset, thanks to this change.

https://user-images.githubusercontent.com/48885911/128477188-bc3f0e60-0e12-4546-80b3-14ea7868a4d7.mov


src/reanimated2/Hooks.ts
```diff
@@ -163,17 +163,14 @@ function runAnimations(animation, timestamp, key, result, animationsActive) {
...
-          return true;
-        }
-      }
-    }
-    return false;
+    return prop.some((item) => isAnimated(item));
+  } else if (prop?.onFrame !== undefined) {
+    return true;
+  } else if (typeof prop === 'object') {
+    return Object.values(prop).some((value) => isAnimated(value));
```
terrysahaidak, line 170: There is a reason why for loops were used - they are just faster. And since we run that code on each frame - it's better to use the fastest code possible.

src/reanimated2/Hooks.ts
```diff
@@ -444,6 +441,30 @@ const canApplyOptimalisation = (upadterFn) => {
...
+    prop.value !== undefined
+  ) {
+    throw new Error(
+      `invalid value passed to \`${currentKey}\`, maybe you forgot to use \`.value\`?`
+    );
+  } else if (prop === 'object') {
+    // if it's a nested object, run validation for all its props
+    Object.keys(prop).forEach((key) => {
+      checkSharedValueUsage(prop[key], key);
+    });
```
terrysahaidak, line 459:
```suggestion
    for (const key of Object.keys(prop)) {
      checkSharedValueUsage(prop[key], key);
    }
```

src/reanimated2/Hooks.ts
```diff
@@ -444,6 +441,30 @@ const canApplyOptimalisation = (upadterFn) => {
...
+  } else if (Array.isArray(prop)) {
+    // if it's an array (i.ex. transform) validate all its elements
+    prop.forEach((element) => {
+      checkSharedValueUsage(element, currentKey);
+    });
```
terrysahaidak, line 464:
```suggestion
    for (const element of prop) {
      checkSharedValueUsage(element, currentKey);
    }
```

src/reanimated2/animation/styleAnimation.ts
```diff
@@ -33,43 +87,40 @@ export function withStyleAnimation(
...
-            // @ts-ignore: eslint-disable-line
-            animation.current.transform[i][type] = currentAnimation.current;
-          }
+      const onFrameRecursive = (
+        currentStyleAnimation: NestedObjectValues<AnimationObject>,
+        currentPath: PrimitiveValue[]
+      ): void => {
+        if (Array.isArray(currentStyleAnimation)) {
+          currentStyleAnimation.forEach((element, index) =>
+            onFrameRecursive(element, [...currentPath, index])
```
terrysahaidak, line 96: I guess in order to make sure it works perfectly for all the envs and just as fast as possible `Array.concat` can be used here
```suggestion
            onFrameRecursive(element, currentPath.concat(index))
```

terrysahaidak (commented):
I just mentioned few places, but overall I think you got a point: 
For loops are just faster. Like old c-like for loops are even faster but new `for of` loops are getting there. But still, they are by lots of benchmarks are 2-4x faster than array methods.

Also, I would not use unnecessary spread operators since the concat method exists. It is 100% as fast as it can get. And spread operator is just getting there.

jmysliv:
@terrysahaidak my mistake, I'm gonna change it.

Szymon20000 (approved):

dozsolti:
It will be nice to have it in the `useAnimatedProps` too

PR #2302, opened 2021-08-20
[iOS] Remove AnimatedLayout

piaskowyk:
## Description
Removed AnimatedLayout, now is possible to use layout animation without using AnimatedLayout wrapper. We overrode default UIManager behavior, it gives us new possibilities.


ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -61,226 +56,140 @@ - (void)invalidate
...
 {
+  if(_cleaningScheduled) {
+    return;
+  }
+  _cleaningScheduled = true;
+  
   __weak REAAnimationsManager *weakSelf = self;
   dispatch_async(dispatch_get_main_queue(), ^{
+    self->_cleaningScheduled = false;
     if (weakSelf == nil) {
```
Szymon20000, line 86: That check should be at the beginning of the block.

Szymon20000 (approved):
Looks fine to me! 🚀

Issue #2292, opened 2021-08-18
Reanimated alpha version will not work on Android with productFlavors configured


chakrihacker:
## Description

The problem occurs from this file https://github.com/software-mansion/react-native-reanimated/blob/master/android-npm/build.gradle from line 63 to 72.

When product flavors are configured the task name changes from the default `mergeDebugNativeLibs` to something like `mergeDevelopmentDebugNativeLibs`.

My opinion is to let developers add few properties to the app's build.gradle (which can be provided at installation instructions)

```
replaceSoTaskDebug.dependsOn(
    project.getTasks().getByPath("${appName}:mergeDebugNativeLibs"),
    project.getTasks().getByPath("${appName}:stripDebugDebugSymbols")
)
project.getTasks().getByPath("${appName}:packageDebug").dependsOn(replaceSoTaskDebug)

replaceSoTaskRelease.dependsOn(
    project.getTasks().getByPath("${appName}:mergeReleaseNativeLibs"),
    project.getTasks().getByPath("${appName}:stripReleaseDebugSymbols")
)
project.getTasks().getByPath("${appName}:packageRelease").dependsOn(replaceSoTaskRelease)
```

### Expected behavior

Build Android app without issues

### Actual behavior & steps to reproduce

Gradle build failed at the above place

## Snack or minimal code example

Can produce if necessary

## Package versions

- React Native: 0.64
- React Native Reanimated: 2.3.0-aplha-2
- NodeJS: 16
- Java & Gradle: 1.8 and 6.7

## Affected platforms

- [x] Android
- [ ] iOS
- [ ] Web

chakrihacker:
I did some debugging and came near to the solution, I hope this helps

```
def getCurrentFlavor() {
    Gradle gradle = getGradle()
    String  tskReqStr = gradle.getStartParameter().getTaskRequests().toString()
    Pattern pattern

    if( tskReqStr.contains( "assemble" ) )
        pattern = Pattern.compile("assemble(\\w+)(Release|Debug)")
    else
        pattern = Pattern.compile("generate(\\w+)(Release|Debug)")

    Matcher matcher = pattern.matcher( tskReqStr )

    if( matcher.find() )
        return matcher.group(1)
    else
    {
        println "NO MATCH FOUND"
        return ""
    }
}
```

We can get the current flavor from this function

And this can be used this way

```
def flavorString = getCurrentFlavor()
println('flavorString: ' + flavorString)

[...]

Issue #2245, opened 2021-07-30
android build fails react-native < 0.65 + 2.3.0-alpha.2 + android build flavors


mikehardy:
## Description

The < 65 version of the android build is not android build flavor aware

### Expected behavior

Building with flavors should work?

### Actual behavior & steps to reproduce

Building with flavors fails because the tasks depended on in the build.gradle for this module are not flavor aware, so they don't exist

## Snack or minimal code example

https://github.com/software-mansion/react-native-reanimated/blob/73c620a0a6fc7b8aa0ed9e2a0df1ff55c7f1ffe2/android-npm/build.gradle#L62-L72

An example of making them work correctly is to replace every 'Release' or 'Debug' with 'DevRelease' or 'DevDebug' in my project (I have a 'dev' flavor, and it results in tasks with 'Dev' sliced in the task names)

I think it needs some magic like an application variants loop and then set the replace task to depend on the merge/strip of each variants https://stackoverflow.com/a/63922131/9910298

## Package versions

- React Native: 0.64.2
- React Native Reanimated: 2.3.0-alpha.2
- NodeJS: n/a
- Xcode: n/a
- Java & Gradle: JDK11 + Gradle 7.1.1 + AGP7.0.0

## Affected platforms

- [x] Android
- [ ] iOS
- [ ] Web

mikehardy:
Here's an example of a repo that has a similar need, at the very top of the loop on variants they get the name, then at the bottom they do a dependsOn of a dynamic task name based on variant name. The middle will not be interesting here, but it illustrates the principle

https://github.com/oblador/react-native-vector-icons/blob/master/fonts.gradle

mikehardy:
@piaskowyk no intention to pressure (it's open source of course!) just wanted to note that 0.65 is getting very close to release, and this package is still alpha for 0.65 compatibility
I had a thought where maybe you release a major version as-is (even knowing this issue exists and is not fixed yet) with a big warning that android build variants are currently not supported in the new major version pending a fix, then when this is fixed it could land as a patch and with a changelog edit saying it's good to go now

Seems odd but even as an affected party (my app uses build variants) as long as it was semver major with a note, I think that's acceptable - it's what semver majors are for.

The warning could be "This release is intended for people upgrading to react-native 0.65. If you are using build variants in Android and are still on react-native 0.64, you should stay on reanimated 2.2.x until a fix for #2245 is contributed. Thanks")

andresribeiro:
[...]

PR #2319, opened 2021-08-24
Support for gradle flavors

piaskowyk:
## Description

Added support for Gradle flavor. Solution based on https://stackoverflow.com/a/34679933/7276742

Fixes
- https://github.com/software-mansion/react-native-reanimated/issues/2245
- https://github.com/software-mansion/react-native-reanimated/issues/2292

### DIFF
--- a/android-npm/build.gradle
+++ b/android-npm/build.gradle
@@ -1,4 +1,6 @@
 import groovy.json.JsonSlurper
+import java.util.regex.Matcher
+import java.util.regex.Pattern
 configurations.maybeCreate("default")
 
 def inputFile = new File(projectDir, '../../react-native/package.json')
@@ -32,6 +34,25 @@ abstract class replaceSoTask extends DefaultTask {
     }
 }
 
+def getCurrentFlavor() {
+    Gradle gradle = getGradle()
+    String taskRequestName = gradle.getStartParameter().getTaskRequests().toString()
+    Pattern pattern
+
+    if(taskRequestName.contains("assemble")) {
+        pattern = Pattern.compile("assemble(\\w+)(Release|Debug)")
+    }
+    else {
+        pattern = Pattern.compile("generate(\\w+)(Release|Debug)")
+    }
+    Matcher matcher = pattern.matcher(taskRequestName)
+
+    if(matcher.find()) {
+        return matcher.group(1).toLowerCase()
+    }
+    return ""
+}
+
 def replaceSoTaskDebug
 def replaceSoTaskRelease
 if(Integer.parseInt(minor) < 65) {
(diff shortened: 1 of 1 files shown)


android-npm/build.gradle
```diff
@@ -60,16 +82,16 @@ rootProject.getSubprojects().forEach({project ->
                 project.getProperties().android.packagingOptions.pickFirst("lib/**/libfbjni.so")
 
                 replaceSoTaskDebug.dependsOn(
-                    project.getTasks().getByPath("${appName}:mergeDebugNativeLibs"),
-                    project.getTasks().getByPath("${appName}:stripDebugDebugSymbols")
+                    project.getTasks().getByPath("${appName}:merge${flavorString}DebugNativeLibs"),
```
Szymon20000, line 85: What about build variant does it also has an impact on the task name?

Szymon20000: I didn't understand a build variant term properly. It turned out that it's just a combination of build type and a flavor.

Szymon20000 (approved):

mikehardy (commented):
Looks great @piaskowyk - thanks - sorry I couldn't PR it myself!

piaskowyk:
@mikehardy your comments were really helpful! Great cooperation for me 😄

ilyausorov:
@piaskowyk @Szymon20000 hey guys, I've been trying to get through this issue of installing react-native-reanimated@2.2.2 with RN 63.4 and multiple productFlavors configured and this solution doesn't seem to be returning the product flavor to properly finish the gradle sync process.

What I am seeing is that the value of `taskRequestName` in `getCurrentFlavor()` comes back as `[DefaultTaskExecutionRequest{args=[],projectPath='null'}]` and therefore the flavor returns as an empty string and the gradle sync process fails.

I've had to add a patch to hard code my product flavor (in each respective branch) so that it successfully builds 
`def flavorString = 'Dev' // getCurrentFlavor()`

See screenshot from Android Studio logs where I added a few println's:

image

Any ideas why I'm not getting any value for the flavor?

mikehardy:
This isn't going to be the most helpful, but with react-native 0.65.1 I don't see a problem, it appears to work for me

ilyausorov:
@mikehardy that’s correct. If you look at the code, the get flavor code only runs if the react native version is below 65

dark1451:
@ilyausorov 

I had the same experience.

The link's answer succeeds in getting the correct build value.

If I directly modify gradle like this, it is not tracked in git, so it is unsettling.
Or need to downgrade the reanimated version or upgrade the react-native version...

I hope it merges into that solution.

react-native-cli: 2.0.1
react-native: 0.64.2

https://github.com/software-mansion/react-native-reanimated/issues/2292#issuecomment-901100248

mikehardy:
@dark1451 

> If I directly modify gradle like this, it is not tracked in git, so it is unsettling.


Use patch-package: https://github.com/ds300/patch-package - 6,286 stars for a reason, it's a vital tool - cheers

PR #2324, opened 2021-08-26
Fixes before release

jmysliv:
## Description

Many fixes before release. PR including fixes for react-native-screens: https://github.com/software-mansion/react-native-reanimated/pull/2340

Authors: @Szymon20000 @piaskowyk @jmysliv

PR #2333, opened 2021-08-30
fix: Package build failing - RNScreens not found

lukebars:
## Description

Currently, package build is failing due to react-native-screens not being found. Looks like it was added by accident, because package.json doesn't include it.

## Changes

- I've removed react-native-screens from module's `build.gradle` as it fixes the build stage.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes


piaskowyk:
Hey, @lukebars thanks for your commitment! 🙌 
We need this dependency, but you are right we missed adding screens to dependencies. I fix it here - https://github.com/software-mansion/react-native-reanimated/pull/2324

lukebars:
Good! Thanks for the clarification :)

PR #2377, opened 2021-09-06
fix: Fix `RNScreens/RNSScreen.h` import if react-native-screens is not installed

mrousavy:
## Description

When not having react-native-screens installed, react-native-reanimated does not build in alpha.3 (or current master).

There is an import: 

```objc
#import <RNScreens/RNSScreen.h>
```

which does not work since RNScreens does not exist. This PR fixes this issue by conditionally importing RNScreens.

## Changes

- Conditionally import RNScreens depending on whether it's installed or not
- Conditionally use RNSScreen class depending on whether it's available or not

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/ios/LayoutReanimation/REAUIManager.mm
+++ b/ios/LayoutReanimation/REAUIManager.mm
@@ -8,7 +8,10 @@
 #import "RCTLayoutAnimationGroup.h"
 #import "REAIOSScheduler.h"
 #include "Scheduler.h"
+
+#if __has_include(<RNScreens/RNSScreen.h>)
 #import <RNScreens/RNSScreen.h>
+#endif
 
 @interface RCTUIManager(REA)
 - (void)_manageChildren:(NSNumber *)containerTag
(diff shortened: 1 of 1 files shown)


mrousavy:
btw that file has mixed `include` and `import` directives

piaskowyk:
@mrousavy Did you run `yarn` again? We have added screens as dependencies in `package.json`
https://github.com/software-mansion/react-native-reanimated/blob/master/package.json#L72

mrousavy:
oh so it's an intentional require? why?

piaskowyk:
We want to use Reanimated layout animation to animate screens transition (entry/exit). For example here: https://github.com/software-mansion/react-native-reanimated/blob/master/Example/test/ModalInNavigation.tsx This isn't perfect right now, but we work on it still.

mrousavy:
I see, weird that react-native-screens is not being installed for me though... It is in my [yarn.lock file](https://github.com/mrousavy/react-native-vision-camera/blob/e9bbedb6c9693a38974cf4cdd617a1faab292e34/example/yarn.lock#L4416-L4421) but not in the [Podfile.lock](https://github.com/mrousavy/react-native-vision-camera/blob/feat/frame-processor-fps-auto/example/ios/Podfile.lock), even deleted the files and re-generated them...

piaskowyk:
I checked this again. You are right. If screens are installed as a dependency, they aren't added as a new Pod. Then is required to run `yarn add react-native-screens && cd ios && pod install`

mrousavy:
Can't you add `RNScreens` to the podspec dependencies?

piaskowyk (approved):

piaskowyk:
Thanks!

JayBee007:
Great!  I have been copy-pasting the changes directly to node_modules 👍

PR #2399, opened 2021-09-09
Fixed ios build

jmysliv:
## Description
On ios build didn't work because of `<AndroidScheduler.h>` include. I made this include conditional, so it will be included only on android.

## Changes
* Added conditional include
* Added triggering build check when files inside common directory are edited


piaskowyk (approved):
🚀

PR #2400, opened 2021-09-09
Convert coords

jmysliv:
## Description
This pr introduces the `convertCoords` function which enables to convert absolute coordinates to coordinates relative to the given view.

## Changes
* Added `convertCoords` function
* Updated docs

### DIFF
--- a/docs/docs/api/miscellaneous/convertCoords.md
+++ b/docs/docs/api/miscellaneous/convertCoords.md
@@ -1,0 +1,44 @@
+---
+id: convertCoords
+title: convertCoords
+sidebar_label: convertCoords
+---
+
+Determines the location on the screen, relative to the given view. It might be useful, when there are only absolute coordinates available and you need coordinates relative to the parent.
+
+### Arguments
+
+#### animatedRef
+
+The product of [`useAnimatedRef`](../hooks/useAnimatedRef) is a Reanimated's extension of a standard React's ref(delivers view tag on the UI thread). This ref should be passed as a prop to view relative to which we want to know coordinates.
+
+#### x
+
+Absolute `x` coordinate.
+
+#### y
+
+Absolute `y` coordinate
+
+### Returns
+
+Object which contains relative coordinates
+
+- `x`
+- `y`
+
+### Example
+
+```js
+const Comp = () => {
+  const aref = useAnimatedRef();
+  // ...
+
+  useDerivedValue(() => {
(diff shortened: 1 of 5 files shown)


jakub-gonet (commented):
It doesn't convert cords - it _measures_ and then converts cords relative to passed arguments. Can we avoid passing absolute x & y or not really?

docs/docs/api/miscellaneous/convertCoords.md
```diff
@@ -0,0 +1,44 @@
...
+  // ...
+
+  useDerivedValue(() => {
+    const coords = convertCoords(aref, x, y);
+    // ...
+  });
+
+  return <View ref={aref} />;
+};
+```
```
piaskowyk, line 44: This example doesn't really show how to use it IMO. It doesn't show how to get `x` and `y`.

src/reanimated2/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+  x: number,
+  y: number
+): ComponentCoords {
+  'worklet';
+  const parentCoords = measure(parentRef);
+  return {
+    x: x - parentCoords.x,
+    y: y - parentCoords.y,
+  };
+}
```
piaskowyk, line 20: If I understand correctly. if someone wants to use convertCoords they need to call measure on current component and then pass `x` and `y`? I don't really like this solution. Maybe a more ergonomic solution is passing two refs (current and parent) and do a measure inside the function.

src/reanimated2/utils.ts
```diff
@@ -0,0 +1,20 @@
+import { RefObject } from 'react';
+import { measure } from '.';
+
+export interface ComponentCoords {
+  x: number;
+  y: number;
+}
+
+export function convertCoords<T>(
```
piaskowyk, line 9: I think that `getRelativeCoords` is a better name for this function because is more descriptive.

piaskowyk (changes requested):
The current solution requires two calls of `useAnimatedRef` and two calls of `measure`. I think we can do it better. For example:
We can get the only ref to the current component, call the new version of measureWithParent - new implementation in JSI and calculate cords on the native side. What do you think?

jmysliv:
`x` and `y` are passed to function because we can get them from gesture handler events. That is the main use case for this function to work with gesture handler.

piaskowyk (approved):

PR #2402, opened 2021-09-09
GitHub action for testing rn versions with reanimated

jmysliv:
## Description
This pr introduces GitHub action that will test reanimated version from a master branch with apps with different react-native versions. This will make releasing new reanimated versions faster and easier, cause we would know earlier if it is working with all react-native versions we want to support.


happyfloat:
Nice work. I tested it with RN 66, and something wasn't bundling correctly, since they have the new boost library instead of boost-for-react-native

piaskowyk:
@happyfloat We haven't been supporting RN66, but of course we will add it.

nomi9995:
@piaskowyk when you have a plan to add a new boost version and support RN66 #2434

.github/workflows/rn-version-check.yml
```diff
@@ -0,0 +1,62 @@
+name: Build npm package
```
piaskowyk, line 1:
```suggestion
name: Test build on different react-native versions
```

Issue #2352, opened 2021-09-01
`Object.keys` reverses order when `useAnimatedStyle` updates


nandorojo:
## Description

If you use a JS dictionary inside of `useAnimatedStyle`, the order of `Object.keys` gets reversed for no apparent reason.

### Expected behavior

The order of `Object.keys(object)` should always be the same for the same object.

### Actual behavior & steps to reproduce

1. Create an object in the JS context with two values. 
2. Inside of `useAnimatedStyle`, print `Object.keys(object)`.
2. Notice that the order flips the second time it's logged.

## Snack or minimal code example

<img width="1918" alt="Screen Shot 2021-08-31 at 9 20 21 PM" src="https://user-images.githubusercontent.com/13172299/131596309-7821fd4a-6549-4dc5-bdad-4ef30a9ad0da.png">

Look at the `console.log` result above ^

Reproduced on this snack: https://snack.expo.dev/@beatgig/talking-banana

```tsx
import * as React from 'react';
import Animated, {
  useSharedValue,
  useAnimatedStyle,
} from 'react-native-reanimated';

export default function App() {
  const animate = {
    translateY: 0,
    opacity: 1,
  }; 

  const style = useAnimatedStyle(() => {
    const keys = Object.keys(animate);

    console.log('[keys]', keys);
    
    return {};
  });
 

  return <></>;
}
```

## Package versions

- React Native: `0.63.4`
- React Native Reanimated: `~2.2.0`
- NodeJS: `14`
- Xcode: N/A
- Java & Gradle: N/A

## Affected platforms

- [x] Android
- [x] iOS
- [ ] Web

It works correctly on web.

This has been the cause of many, many bugs for `moti`, since it uses this API for transforms:

```tsx
<MotiView animate={{ translateY: 10, translateX: 0 }} />
```

nandorojo:
If the object is created _inside_ of `useAnimatedStyle`, it works fine. So how can I ensure that it works properly when passing to `useAnimatedStyle`?

nandorojo:
It seems that wrapping the object with `useDerivedValue` provides stable order for the keys.

```diff
import * as React from 'react';
import Animated, {
  useSharedValue,
  useAnimatedStyle,
} from 'react-native-reanimated';

export default function App() {
  const animate = {
    translateY: 0,
    opacity: 1,
  }; 

+ const object = useDerivedValue(() => animate)

  const style = useAnimatedStyle(() => {
[...]

PR #2429, opened 2021-09-15
Set order for keys in object inside a worklet

piaskowyk:
## Description

We used `unordered_map` to store object keys. It may cause different results of `Object.keys()` called on JS or UI thread. I added `vector` of keys to store the right order of keys and keep constant complexity. Instead of this solution I also can use just `std::map` to store objects in the right order but `map` logarithmic complexity.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/2352

## Reproduce
```js
import { useAnimatedStyle } from 'react-native-reanimated';

  const obj = {
    a: 0,
    b: 1,
  };
    console.log("is worklet:", _WORKLET + ",", "keys:", Object.keys(obj));
    return {};
  });
  return <View></View>;
}
```

### Before
```
is worklet: false, keys: ["a", "b"]
is worklet: true, keys: ["b", "a"]
```

### After
```
is worklet: false, keys: ["a", "b"]
is worklet: true, keys: ["a", "b"]
```

### DIFF
--- a/Common/cpp/SharedItems/FrozenObject.cpp
+++ b/Common/cpp/SharedItems/FrozenObject.cpp
@@ -10,8 +10,11 @@ FrozenObject::FrozenObject(
     const jsi::Object &object,
     RuntimeManager *runtimeManager) {
   auto propertyNames = object.getPropertyNames(rt);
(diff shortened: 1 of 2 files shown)


Szymon20000 (approved):

nandorojo:
Thank you!

PR #2433, opened 2021-09-17
Auto-workletize React Native Gesture Handler callback functions

tomekzaw:
## Description

This PR adds support for detecting patterns in JavaScript code which are specific to [React Native Gesture Handler](https://github.com/software-mansion/react-native-gesture-handler), like `Gesture.Tap().onEnd(...)`, so that they can be automatically converted into worklets even if they are not marked with `'worklet';` directive.

The general pattern is `Gesture.Foo()[*].bar(fun)`, where:
* `Foo` is a gesture object, like `Tap`, `Pan`, `Pinch`, etc.
* `[*]` is any number of chained builder method calls (e.g. `numberOfTaps(2)` or `onStart(...)`)
* `bar` is any builder method name, e.g. `onEnd`
* `fun` is a function declaration, function expression or arrow function expression

The following callback functions will be auto-workletized:

```ts
import { Gesture } from 'react-native-gesture-handler';

const foo = Gesture.Tap()
  .numberOfTaps(2)
  .onBegan(() => {
    // from now on, you may skip the 'worklet'; directive here...
    console.log('onBegan');
  })
  .onStart((_event) => {
    // ...and here...
    console.log('onStart');
  })
  .onEnd((_event, _success) => {
    // ...and here too!
    console.log('onEnd');
  });
```

Please note that the following callback functions will not be auto-workletized:

```ts
somethingNotRelatedToGestures.onEnd(( /* ... */ ) => { /* ... */ });
```

```ts
const tap = Gesture.Tap();
tap.onEnd((_event, success) => { /* ... */ });
```

```ts
const fun = (_event, success) => { /* ... */ };
Gesture.Tap().onEnd(fun);
```

```ts
import { Gesture as RNGHGesture } from 'react-native-gesture-handler';
RNGHGesture.Tap().onEnd((_event, success) => { /* ... */ });
```

In all the above cases, you still need to specify the `'worklet';` directive explicitly.

## Changes

- Added `processIfGestureHandlerEventCallbackFunctionNode` function in `babel.config.js`
- Added snapshot tests for functions that should and should not be workletized

```sh
yarn jest
```

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/__tests__/__snapshots__/plugin.test.js.snap
+++ b/__tests__/__snapshots__/plugin.test.js.snap
@@ -100,7 +100,9 @@ function Box() {
 `;
 
 exports[`babel plugin workletizes gesture object callbacks 1`] = `
-"var singleTap = Gesture.Tap().onEnd(function () {
+"var _reactNativeGestureHandler = require(\\"react-native-gesture-handler\\");
+
+var singleTap = _reactNativeGestureHandler.Gesture.Tap().onEnd(function () {
   var _f = function _f(_event, success) {
     if (success) {
       console.log('single tap!');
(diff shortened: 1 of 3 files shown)


piaskowyk (approved):
🚀

Issue #2393, opened 2021-09-08
[2.3.0-alpha.3] Trying to remove a view index above child count 0


0nlyDev:
## Description

Getting when using the package in conjunction with react-native-google-places-autocomplete.
1. It happens when you type in and then delete (maybe just delete if you already preload the field) with some text
> Trying to remove a view index above child count 0
![Capture](https://user-images.githubusercontent.com/89726447/132598143-f8f717d6-cc6d-40a9-abf8-ed2d892e4b32.PNG)
2. The text input sometimes becomes unclickable

### Expected behavior
To not throw error and to be able to click everywhere on the screen

### Actual behavior & steps to reproduce
Install "react-native": "0.65.1", "react-native-google-places-autocomplete": "^2.4.1", "react-native-reanimated": "^2.3.0-alpha.3",
Then run the example code from bellow, start deleting the default text from the text input field and you should get the error. 

## Snack or minimal code example
```
import React from 'react';
import { GooglePlacesAutocomplete } from 'react-native-google-places-autocomplete';

const GooglePlacesInput = () => {
  return (
    <GooglePlacesAutocomplete
      placeholder='Search'
      onPress={(data, details = null) => {
        // 'details' is provided when fetchDetails = true
        console.log(data, details);
      }}
      query={{
        key: 'YOUR API KEY',
        language: 'en',
      }}
      requestUrl={{
        useOnPlatform: 'web', // or "all"
        url:
          'https://cors-anywhere.herokuapp.com/https://maps.googleapis.com/maps/api', // or any proxy server that hits https://maps.googleapis.com/maps/api
        headers: {
          Authorization: `an auth token`, // if required for your proxy
        },
      }}
    />
  );
};

export default GooglePlacesInput;
```

## Package versions
`"scripts": {
    "android": "react-native run-android",
    "ios": "react-native run-ios",
    "start": "react-native start",
    "test": "jest",
    "lint": "eslint ."
  },
  "dependencies": {
    "@react-native-masked-view/masked-view": "^0.2.6",
    "@react-navigation/drawer": "^6.1.4",
    "@react-navigation/native": "^6.0.2",
    "@react-navigation/native-stack": "^6.1.0",
    "react": "17.0.2",
    "react-native": "0.65.1",
    "react-native-geolocation-service": "^5.3.0-beta.2",
    "react-native-gesture-handler": "^1.10.3",
    "react-native-google-places-autocomplete": "^2.4.1",
[...]

Issue #2394, opened 2021-09-08
[2.3.0-alpha.3] ViewManager for tag 245 could not be found.


0nlyDev:
## Description

Getting the error when using the package in conjunction with react-native-google-places-autocomplete. Might be related to https://github.com/software-mansion/react-native-reanimated/issues/2393
![Capture](https://user-images.githubusercontent.com/89726447/132599386-50908d5f-2f6f-4f05-a03b-733a3889094b.PNG)

### Expected behavior
To be able to click on the results without throwing an error
### Actual behavior & steps to reproduce
Install "react-native": "0.65.1", "react-native-google-places-autocomplete": "^2.4.1", "react-native-reanimated": "^2.3.0-alpha.3",
Then run the example code from bellow, add a letter to the current text to see some results, and click on some (they might not be clickable so keep trying results near the end)

## Snack or minimal code example
```
import React from 'react';
import { GooglePlacesAutocomplete } from 'react-native-google-places-autocomplete';

const GooglePlacesInput = () => {
  return (
    <GooglePlacesAutocomplete
      placeholder='Search'
      onPress={(data, details = null) => {
        // 'details' is provided when fetchDetails = true
        console.log(data, details);
      }}
      query={{
        key: 'YOUR API KEY',
        language: 'en',
      }}
      requestUrl={{
        useOnPlatform: 'web', // or "all"
        url:
          'https://cors-anywhere.herokuapp.com/https://maps.googleapis.com/maps/api', // or any proxy server that hits https://maps.googleapis.com/maps/api
        headers: {
          Authorization: `an auth token`, // if required for your proxy
        },
      }}
    />
  );
};

export default GooglePlacesInput;

```

## Package versions
[...]

PR #2437, opened 2021-09-20
Clear `mPendingDeletionsForTag`

piaskowyk:
## Description

React native stores some tags as pending to delete in a separate array. To keep a consistency of states we clear `mPendingDeletionsForTag` before we call `manageChildren`.
We also resolve an issue with the unknown view manager for the tag. React `dropView` method removes only children of views, we added removing parent view also.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/2393, https://github.com/software-mansion/react-native-reanimated/issues/2394

Authors: @Szymon20000 @piaskowyk

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
@@ -208,6 +208,7 @@ public class ReanimatedNativeHierarchyManager extends NativeViewHierarchyManager
                     @Override
                     public void run() {
                         toBeRemovedChildren.remove(view);
+                        viewGroupManager.removeView(viewGroup, view);
                     } // It's far from optimal but let's leave it as it is for now
                 });
             }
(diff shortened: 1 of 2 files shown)


Szymon20000 (approved):
Perfect!

tonydatw:
@piaskowyk I've updated to latest stable beta and, even though the error message is no longer shown, it still causes the same issue with google places autocomplete. When i press on a row location field, nothing happens.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -197,10 +208,18 @@ public synchronized void manageChildren(int tag, @Nullable int[] indicesToRemove
...
                     } // It's far from optimal but let's leave it as it is for now
                 });
             }
         }
+
+        // mPendingDeletionsForTag is modify by React
+        Set<Integer> pendingTags = mPendingDeletionsForTag.get(tag);
+        if (pendingTags != null) {
+            pendingTags.clear();
+        }
```
mikehardy, line 221: For some reason, in android release builds on my app (or debug builds if I configure minifyEnabled true for debug buildType) this crashes on this line (I carefully deobfuscated the stack trace, and it's definitely crashing here - no idea why minify triggers it when it's not the classic ClassNotFoundException or NoClassDef etc that proguard would normally throw)

I believe that since mPendingDeletionsForTag is initialized to null, and only set to a valid HashMap in one of the constructors but not the other, I must be using the other constructor in my app.

Then when this line is hit, I get a null pointer exception.

I think this whole chunk could be wrapped in an `if (mPendingDeletionsForTag != null) { /* ... */ }`

Issue #2432, opened 2021-09-16
[2.3.0-alpha.3] Installing react-native-reanimated changes the way the elements on screen are rendered when updated


SimpleCreations:
## Description

When installing react-native-reanimated 2.3.0-alpha.3 in a clean project, there is a visual difference in the way the elements update on screen. When the elements re-render, they can be briefly seen overlapping each other.
The issue in not present in react-native-reanimated 2.2.2.

Video demonstration:

https://user-images.githubusercontent.com/12449725/133646790-690f7efc-f8df-4486-95db-a1edffdb8ad9.mp4

Here's a frame from the video where the issue is clearly visible:

![Screen Shot 2021-09-16 at 18 59 24](https://user-images.githubusercontent.com/12449725/133646862-0752fca3-d8ac-4852-9153-9cb0e87760af.png)

Demo from react-native-reanimated 2.2.2, where the issue is not present:

https://user-images.githubusercontent.com/12449725/133647186-55dc774f-1aea-4420-91cd-3456007638fb.mp4

The bug is easier to reproduce on older/slower devices. However it can be seen on a modern simulator as well, but it is more rare and requires faster inputs.

### Expected behavior

Rendering behavior does not change when installing react-native-reanimated 2.3.0-alpha.3 or when upgrading from 2.2.2.

### Actual behavior & steps to reproduce

Rendering behavior changes.

## Snack or minimal code example

Full repro: https://gist.github.com/SimpleCreations/6e37390ffa796dac39433637419a5947

## Package versions

- React Native: 0.64.1
- React Native Reanimated: 2.3.0-alpha.3
- NodeJS: 16.6.1
- Xcode: 12.5.1
- Java & Gradle: -

## Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

PR #2438, opened 2021-09-20
Fix problem with overlapping views

Szymon20000:
## Description

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/2432
Currently, we schedule leftover elements removal to avoid traversing the whole view tree every time onViewRemoval is called. However, with the current approach, it's possible that we remove elements on the next frame that should be removed instantly. To fix the problem I added an additional synchronous call of the removedLeftovers method as the last operation of a UI operation block.


piaskowyk (approved):
🚀

PR #2462, opened 2021-09-28
Support for RN66

piaskowyk:
## Description

Added support for `react-native@0.66`. I temporarily removed binary for RN62 from the package, but in the future, I will add an option to build reanimated from sources.

### DIFF
--- a/Example/android/app/build.gradle
+++ b/Example/android/app/build.gradle
@@ -78,6 +78,11 @@ project.ext.react = [
 ]
 def enableHermes = project.ext.react.get("enableHermes", false);
 
+/**
+ * Architectures to build native code for in debug.
+ */
+def nativeArchitectures = project.getProperties().get("reactNativeDebugArchitectures")
+
 apply from: "../../node_modules/react-native/react.gradle"
 
 /**
@@ -123,6 +128,11 @@ android {
         debug {
             packagingOptions {
                 doNotStrip "**/**/*.so"
+            }
+            if (nativeArchitectures) {
+                ndk {
+                    abiFilters nativeArchitectures.split(',')
+                }
             }
         }
         release {
--- a/Example/android/build.gradle
+++ b/Example/android/build.gradle
@@ -6,6 +6,7 @@ buildscript {
         minSdkVersion = 21
         compileSdkVersion = 30
         targetSdkVersion = 30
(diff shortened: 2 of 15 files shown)


jmysliv (approved):

mikehardy:
:rocket: 

Just to note if you publish a new beta, I've got an app I can use to test-integrate it vs react-native 0.66.0-rc.4 / reanimated 2.3.0-beta.`n+1`

piaskowyk:
Thanks! 😍

piaskowyk:
I am building a 2.3.0-beta.2 package at this moment 😄

mikehardy:
I just tested it, and initially I had a problem on Android but it was my own fault. My lesson is: read the instructions! 
Of course, right?

https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation#android

It seems to work great for me on react-native 0.66.0-rc.4

----

image

https://github.com/software-mansion/react-native-reanimated/blob/437cb56ab7c567ab3ef0a63482bdfb06e03f2e5f/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java#L122

the error message makes it clear mScheduler is null, implying setScheduler was not called:

https://github.com/software-mansion/react-native-reanimated/blob/437cb56ab7c567ab3ef0a63482bdfb06e03f2e5f/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java#L60-L62

...implying initWithContext was not called since any call to it would result in a non-null WeakReference:

https://github.com/software-mansion/react-native-reanimated/blob/437cb56ab7c567ab3ef0a63482bdfb06e03f2e5f/android/src/main/java/com/swmansion/reanimated/NodesManager.java#L140-L143

...implying somehow getJSIModules was never called, as it would definitely call initWithContext:

https://github.com/software-mansion/react-native-reanimated/blob/437cb56ab7c567ab3ef0a63482bdfb06e03f2e5f/android/src/main/java/com/swmansion/reanimated/ReanimatedJSIModulePackage.java#L16-L21

That one has no callsite references that GitHub can infer, so it is not called by this module internally.

That means it must be something about registration? Indeed, I'd missed this chunk of instructions:

https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation#android

PR #2464, opened 2021-09-28
fix: Add `RNScreens` to podspec dependency

mrousavy:
## Description

Instead of disabling `RNScreens` support, we can force the `RNScreens` native dependency by adding it to the REA podspec.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


mrousavy:
cc @piaskowyk that's what I meant - will test if this works in a sec

mrousavy:
nevermind, I don't think that'll work.

PR #2486, opened 2021-10-05
Fixed removing native views without ViewManager

piaskowyk:
## Description

Some native Views don't have their own ViewManager. We had a check for this but this check was wrong. We did:
```java
ViewManager vm = nativeViewHierarchyManager.resolveViewManager(tag);
if (vm != null) {
    // code ...
}
```
But if it is unable to find view manager for tag `resolver` throws an exception instead of return null. I replace this check with `try {} catch() {}`

## Checklist

- [ ] Ensured that CI passes


codedbystability:
issue resolved? if yes, can you help me to integrate it ? i have no such a file called 
ReanimatedNativeHierarchyManager.java

PR #2488, opened 2021-10-05
fix: Avoid duplicate C++ symbols by using `libjsi.so` shared library

mrousavy:
## Description

By compiling `jsi.cpp`, we will have duplicate C++ symbols in the user's app because `jsi.cpp` is compiled multiple times. (once for RN, once for REA, and maybe also for other third party libs)

This PR fixes this by re-using the shared/dynamic C++ library `libjsi.so`.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/CMakeLists.txt
+++ b/android/CMakeLists.txt
@@ -29,8 +29,6 @@ add_library(
         ${sources_registries}
         ${sources_android}
         ${source_tools}
-        ${NODE_MODULES_DIR}/react-native/ReactCommon/jsi/jsi/jsi.cpp
-        ${NODE_MODULES_DIR}/react-native/ReactCommon/jsi/jsi/JSIDynamic.cpp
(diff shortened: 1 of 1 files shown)


piaskowyk:
At first sight, everything looks great 🚀 I just need to check building scripts for older react's versions and if everything is ok I will merge it. Thanks!

mrousavy:
oh yeah I think libjsi.so is only available on RN 0.66. We might want to add fallback behaviour on older RN versions. But I'm not sure, I'd had to test myself too.

piaskowyk:
I need to create a new PR because I can't edit this PR but I had to add backward compatibility.
https://github.com/software-mansion/react-native-reanimated/pull/2530
Thanks for the help! 🚀

PR #2497, opened 2021-10-06
Fix precommit hook & reformat code

kacperkapusciak:
## Description

This PR fixes the pre-commit hook and reformats code. From now on pre-commit hook formats `c++`, `obj-c`, `java` & `typescript` code. Linting Java code is not possible with lint-staged + spotless and running this on commit has bad a developer experience.

## Changes

- Bump release-it from 13.1.1 to 13.7.2
- Bump husky from 4.2.5 to 7.0.0
- Bump lint-staged from 10.2.11 to 11.2.0
- Use Spotless framework for formatting Java code
- Renamed and unified script names in `package.json`

Change code (add a comment or variable), add too much whitespace (spaces, tabs, newlines), and then stage & commit your changes.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/.github/workflows/ios-build.yml
+++ b/.github/workflows/ios-build.yml
@@ -25,6 +25,8 @@ jobs:
         with:
           node-version: 14
           cache: 'yarn'
+      - name: Install Reanimated node dependencies
+        run: yarn
       - name: Install node dependencies
         working-directory: ${{ env.WORKING_DIRECTORY }}
         run: yarn


WoLewicki (approved):

tomekzaw (changes requested):
1. ~~`format:ios` should also format `*.mm` files~~ (done)
2. ~~`find ios/ -iname *.h -o -iname *.m -o -iname *.cpp` seems to list only `*.h` files
a) probably `\(` and `\)` are missing
b) for zsh, I needed to replace `*.h` with `"*.h"` etc. (in bash works fine)~~
(but `yarn format:ios` works fine)
3. these commands fail quietly if `clang-format` is not installed (that's also my case), there should be at least a warning (similar as for `cpplint`)

PR #2499, opened 2021-10-06
Screens header

piaskowyk:
## Description
This PR provides proper handling of header animation during layout animation. Native stack requires an additional step - calling `applyLayoutUpdate` on updating component.

## Test case 
https://github.com/software-mansion/react-native-reanimated/blob/2.3.0-beta.3/Example/test/ModalInNavigationNativeStack.tsx


android/src/main/java/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -190,11 +190,11 @@ public synchronized void updateLayout(
     ViewManager parentViewManager = this.resolveViewManager(parentTag);
     String parentViewManagerName = parentViewManager.getName();
     View container = resolveView(parentTag);
-    if (container != null
-        && parentViewManagerName.equals("RNSScreenContainer")
-        && this.mReaLayoutAnimator != null) {
-      this.mReaLayoutAnimator.applyLayoutUpdate(
-          viewToUpdate, 0, 0, container.getWidth(), container.getHeight());
+    if (container != null 
+        && (parentViewManagerName.equals("RNSScreenContainer") || parentViewManagerName.equals("RNSScreenStack"))
```
WoLewicki, line 194: I believe we checked that you can change checking parent VM to just checking if the child is `RNSScreen` since we want this to work both for `native-stack` and other navigators: 
```suggestion
        && `viewManagerName.equals("RNSScreen")`
```

WoLewicki (commented):
Please check if applying the shorter version works and I believe we can safely merge it then :tada:

android/src/main/java/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
piaskowyk: Yes, you are right. I tested it and everything works 🚀

kmagiera (approved):

WoLewicki (approved):

PR #2511, opened 2021-10-08
Add support for animating transform matrices

kmagiera:
## Description

This change adds support for animation transform matrices. Before, it was only possible to animate attributes that are either numbers or ones representing colors (e.g. `rgba(r,g,b,a)` or `#eae`). In react-native it is possible to not only provide transform as a list of operations like `translate` or `scale` but also by sending the full affine transform matrix (4x4 matrix represented as an array of 16 elements). This change makes it possible for reanimated to recognize such matrix and animate its transforms accordingly.

## Changes

 - added a way to recognize values of type array in utils.ts
 - override animation.onStart and animation.onFrame for the array type
 - copy animation to each individual element of an array and perform these separately (this isn't likely the most optimal approach as we are spanning 16 animation instances for a single matrix but this is a relatively rare usecase and this solution should be ok for the time being)
 - add new test app that uses inline transforms and also matrix transform to showcase the changes

See newly added example `Example/test/TestApp.js` that demonstrate that changes made to matrix transform can be animated.

## Checklist

- [ ] Updated TS types
- [ ] Ensured that CI passes

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -36,6 +36,18 @@ declare module 'react-native-reanimated' {
     GestureHandlerGestureEvent,
     PanGestureHandlerGestureEvent,
   } from 'react-native-gesture-handler';
+
+  export {
+    Animation,
+    TimingAnimation,
+    SpringAnimation,
+    DecayAnimation,
+    DelayAnimation,
+    RepeatAnimation,
+    SequenceAnimation,
+    StyleLayoutAnimation,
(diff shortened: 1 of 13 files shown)


piaskowyk:
Example usage of types:
```js
const array = [1, 2, 3];
const matrix = useSharedValue<Array<number> | Animation<TimingAnimation>>(array);
matrix.value = withTiming(array);
const animation = withTiming(array);
```

Issue #2504, opened 2021-10-07
Crash: `undefined is not an object (evaluating 'Platform.OS')`


SimpleCreations:
## Description

I've received a crash report from an iOS 12.5.4 device:

```
facebook::jsi::JSError: undefined is not an object (evaluating 'Platform.OS')

getTimestamp@/Users/admin/[redacted]/[redacted]-rn-app/node_modules/react-native-reanimated/src/reanimated2/core.ts (178:7):1:107
[native code]
_f@/Users/admin/[redacted]/[redacted]-rn-app/node_modules/react-native-reanimated/src/reanimated2/animation/util.ts (208:17):1:89
[native code]
workletValueSetter@/Users/admin/[redacted]/[redacted]-rn-app/node_modules/react-native-reanimated/src/reanimated2/core.ts (186:0):1:349
[native code]
[native code]
workletValueSetter@/Users/admin/[redacted]/[redacted]-rn-app/node_modules/react-native-reanimated/src/reanimated2/core.ts (186:0):1:622
[native code]

0  [redacted]        facebook::react::JSIExecutor::defaultTimeoutInvoker(std::__1::function<void ()> const&, std::__1::function<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > ()>) ([redacted])
1  [redacted]        facebook::react::JSIExecutor::defaultTimeoutInvoker(std::__1::function<void ()> const&, std::__1::function<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > ()>) ([redacted])
2  [redacted]        folly::Expected<long long, folly::ConversionCode> folly::detail::str_to_integral<long long>(folly::Range<char const*>*) ([redacted])
3  [redacted]        folly::Expected<long long, folly::ConversionCode> folly::detail::str_to_integral<long long>(folly::Range<char const*>*) ([redacted])
4  libdispatch.dylib __dispatch_call_block_and_release
5  libdispatch.dylib __dispatch_client_callout
6  libdispatch.dylib __dispatch_main_queue_callback_4CF$VARIANT$mp
7  CoreFoundation    ___CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
8  CoreFoundation    ___CFRunLoopRun
9  CoreFoundation    _CFRunLoopRunSpecific
10 GraphicsServices  _GSEventRunModal
11 UIKitCore         _UIApplicationMain
[...]

Issue #2500, opened 2021-10-06
Runtime corrupt after fast refresh


piaskowyk:
## Description
// to do - I will do it 🙌
![image](https://user-images.githubusercontent.com/36106620/136230900-ad6407c3-b439-44be-bebf-e4e944d0bc8e.png)

## Snack or minimal code example

```js
import {NavigationContainer} from '@react-navigation/native';
import {createNativeStackNavigator} from 'react-native-screens/native-stack';
import React, {useEffect} from 'react';
import {Button, Text, View, Dimensions, StyleSheet} from 'react-native';
import Animated, { interpolate, interpolateColor, useAnimatedReaction, useAnimatedStyle, useSharedValue, withRepeat, withSequence, withTiming } from 'react-native-reanimated';

const Stack = createNativeStackNavigator();

function HomeScreen2({navigation}) {
  return (
    <View style={{flex: 1, alignItems: 'center', justifyContent: 'center'}}>
      <Text>Home Screen</Text>
      <Button title="navigate" onPress={() => navigation.navigate('Next')} />
    </View>
  );
}

const NextScreen = ({navigation}) => {
  return (
    <View>
      <Text 
      // style={{marginTop:200}}
      >Some Text</Text>
      <Button title="sss" onPress={() => navigation.push('Next')} />
    </View>
  );
};

const AnimatedText = Animated.createAnimatedComponent(Text);

const {width} = Dimensions.get("window");

const HomeScreen = ({ navigation }) => {
  const textAnimationValue = useSharedValue(0);

  // start animation on mount
  useEffect(() => {
    textAnimationValue.value = withRepeat(
      withSequence(withTiming(1, { duration: 3000 }), withTiming(0, { duration: 3000 })),
    -1)
  }, []);

  const animContainerStyle = useAnimatedStyle(() => ({
    width: '100%',
  }), [textAnimationValue])

  const animatedTextStyle = useAnimatedStyle(() => ({
    transform: [
      { translateX: interpolate(textAnimationValue.value, [0, 1], [0, width]) }
    ],
    color: interpolateColor(textAnimationValue.value, [0, 1], ["red", "green"])
  }), [textAnimationValue])

  useAnimatedReaction(() => textAnimationValue.value === 1,
  (isFinished) => {
    if (isFinished) {
      console.log('One animation rnd dosne!');
    }
  }, [textAnimationValue])

  const onPress = () => {
    navigation.push("Home");
  }

  return (
[...]

Issue #2485, opened 2021-10-05
[iOS, 2.3.0-beta.2] Crash when fast reloading


hannojg:
## Description

I was about to create a reproduction for another issue when I came across another bug, where the app crashes when making changes and the app fast reloads.

### Expected behavior

The app doesn't crash when doing a fast reload.

### Actual behavior & steps to reproduce

The iOS app crashes when making a change in JS, which causes a fast reload.

## Snack or minimal code example

I am not sure what the cause of this issue is. But I have a reproduction available here:
https://github.com/hannojg/rea-ios-reproduction

1. Clone the repo
2. Run yarn
3. Pod install
4. Open the project in Xcode and run the app
5. Make a change to the text in App.tsx for example

Here are some "specialities" about the environment (trying to replicate our company's app environment to reproduce another issue):
- It uses a custom expo client (https://docs.expo.dev/clients/getting-started/)
- It uses react navigation with native stack and native screens (but that doesn't appear to be a cause for the issue, the issue also happens if you revert the commit where I added react-naviogation)

## Package versions

- React Native: 0.65.1
- React Native Reanimated: 2.3.0-beta.2
- NodeJS: 14
- Xcode: 12.5.1
- Java & Gradle: irrelevant

## Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

hannojg:
_Note:_ Intristringly this seems to be the bug we are also experiencing in our company's app. There the crash just happens immediately when starting the app, not just when fast reloading. However the stack trace is the same.

Jackmekiss:
I have the same issue

PR #2512, opened 2021-10-08
Call triggerUI from main thread

piaskowyk:
## Description

We need to apply component changes (registration of mappers etc) before starting of animation, that's why we added `self.flushUiOperations();` but we can't call this method inside `onBatchComplete` because this call can be executed on a different thread than the main thread. So we moved this call to `onAnimationStart` on the main thread.

Fixes:
- https://github.com/software-mansion/react-native-reanimated/issues/2500 
- https://github.com/software-mansion/react-native-reanimated/issues/2504 
- https://github.com/software-mansion/react-native-reanimated/issues/2485

### DIFF
--- a/ios/LayoutReanimation/REAUIManager.mm
+++ b/ios/LayoutReanimation/REAUIManager.mm
@@ -339,9 +339,4 @@ std::weak_ptr<reanimated::Scheduler> _scheduler;
   [viewRegistry removeObjectForKey:view.reactTag];
 }
 
-- (void)batchDidComplete
-{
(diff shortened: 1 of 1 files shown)


mjmasn:
Hi @piaskowyk, just fyi I'm still seeing the installCoreFunctions issue on React Native 0.66 with this patch applied with `npx patch-package` on top of 2.3.0-beta.2

hannojg:
I just tested the patch and the issue I reported, namely #2485, seems to be fixed by this!

PR #2519, opened 2021-10-11
Add global method that allows manipulation of states in Gesture Handler

j-piasecki:
## Description

This PR adds a new `_setGestureState` method to the global object which will be used by the `react-native-gesture-handler`. In order to accomplish synchronous communication between modules I added shared interfaces between libraries (`GestureHandlerStateManager.java` and `RNGestureHandlerStateManager.h`). On startup RNGH module is tried to be loaded and then casted to the correct interface.
I did not update the documentation since this method is intended to be used internally.

## Changes

- Added global method `_setGestureState`
- Added loading of RNGH module and shared interfaces
- Added `Custom` gesture to the relevant Babel plugin list
- Changed `onBegan` to `onBegin` in the Babel plugin

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


src/reanimated2/NativeMethods.ts
```diff
@@ -38,3 +38,11 @@ export function scrollTo(animatedRef, x, y, animated) {
   const viewTag = animatedRef();
   _scrollTo(viewTag, x, y, animated);
 }
+
+export function setGestureState(handlerTag: number, newState: number): void {
+  'worklet';
+  if (!_WORKLET && !isChromeDebugger()) {
+    return;
```
kmagiera, line 45: Maybe show a warning here?

kmagiera (approved):
Looks good. Added just one minor comment

src/reanimated2/NativeMethods.ts
piaskowyk:
```suggestion
    console.warn("[Reanimated] You can not use setGestureState in non-worklet function.");
```

piaskowyk (approved):

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -207,6 +208,20 @@ void RuntimeDecorator::decorateUIRuntime(
...
       rt, jsi::PropNameID::forAscii(rt, "_stopObservingProgress"), 0, clb8);
   rt.global().setProperty(rt, "_stopObservingProgress", _stopObservingProgress);
+  
+  auto clb9 = [setGestureState](
+                  jsi::Runtime &rt,
+                  const jsi::Value &thisValue,
+                  const jsi::Value *args,
+                  size_t count) -> jsi::Value {
+    int handlerTag = (int)args[0].asNumber();
+    int newState = (int)args[1].asNumber();
```
piaskowyk, line 218:
```suggestion
    int handlerTag = static_cast<int>(args[0].asNumber());
    int newState = static_cast<int>(args[1].asNumber());
```

PR #2530, opened 2021-10-12
Fix/fix duplicate c++ symbols jsi

piaskowyk:
## Description

By compiling `jsi.cpp`, we will have duplicate C++ symbols in the user's app because `jsi.cpp` is compiled multiple times. (once for RN, once for REA, and maybe also for other third party libs)

This PR fixes this by re-using the shared/dynamic C++ library `libjsi.so`.

Based on: https://github.com/software-mansion/react-native-reanimated/pull/2488
Authors: @mrousavy @piaskowyk


mrousavy:
btw the build.gradle + CMakeLists is a bit messy already, I think we can do a few cleanups here and there to reduce it's complexity

PR #2532, opened 2021-10-13
Added support for shareable jsi host objects and host functions

chrfalch:
## Description

When implementing jsi host objects or host functions that wraps native objects/funcs it would be nice to be able to share these between javascript runtimes.

This commit adds support for marking jsi host objects/host functions as shareable and handle them in a controlled manner.

It will always be the implementer's responsibility to ensure thread safety when using objects/funcs in different runtimes.

To mark a host object or host function as shareable, either call the ShareableValue::markHostAsShareable on the object/func, or make sure the host object returns true when the `get` method is called for the property ShareableValue::HOST_SHAREABLE.

An example of such a use case can be seen here:

image

The `paint` object is constructed by calling a jsi-based host function and returns a host object that points to the underlying cpp object representing the paint object in the native world. The object can be accessed across threads so Reanimated can just copy the underlying cpp pointer and return it as a new host object when the worklet runtime asks for the object.

In the host object implementation we can look for access to the shareable host property and return true to indicate to Reanimated that this object can be freely copied between the two runtimes:

image

If we want to share a host function we can mark it with the shareable host property to accomplish the same thing for the function:

image

## Changes

- Added two new ValueTypes to support shareable host objects/functions.
- Added value wrappers for the new ValueTypes
- Added conversion from jsi::Value to the two new value wrappers
- Added conversion to jsi:Value from value wrappers.
- Added methods for making a hostobject/hostfunction shareable 

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -313,9 +313,11 @@ jsi::Value ShareableValue::toJSValue(jsi::Runtime &rt) {
     case ValueType::DirectHostFunctionType: {
       auto directFunctionWrapper =
           ValueWrapper::asDirectHostFunctionWrapper(valueContainer);
-      jsi::HostFunctionType func = *directFunctionWrapper->value;
       return jsi::Function::createFromHostFunction(
-          rt, jsi::PropNameID::forAscii(rt, "directHostFunction"), 0, func);
+          rt,
+          jsi::PropNameID::forAscii(rt, "directHostFunction"),
+          0,
+          *directFunctionWrapper->value);
     }
     case ValueType::HostFunctionType: {
       auto hostFunctionWrapper =
--- a/Common/cpp/SharedItems/ShareableValue.cpp
+++ b/Common/cpp/SharedItems/ShareableValue.cpp
@@ -13,6 +13,8 @@ const char *HIDDEN_HOST_OBJECT_PROP = "__reanimatedHostObjectRef";
 const char *ALREADY_CONVERTED = "__alreadyConverted";
 const char *CALL_ASYNC = "__callAsync";
 const char *PRIMAL_FUNCTION = "__primalFunction";
+const char *ShareableValue::HOST_SHAREABLE = "__hostShareable";
+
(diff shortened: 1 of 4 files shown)


mrousavy:
I see that it makes sense to add the hidden property to a function to be able to determine whether it can be shared or not, but I think it would be cleaner (from API user's perspective) to use inheritance.

So instead of implementing `HostObject::get` to return true if a specific string is accessed, you can just inherit from e.g. `reanimated::ShareableHostObject`, and that would type-match (`dynamic_cast`) in the `ShareableValue::adapt` function.

chrfalch:
> I see that it makes sense to add the hidden property to a function to be able to determine whether it can be shared or not, but I think it would be cleaner (from API user's perspective) to use inheritance.
> 
> So instead of implementing `HostObject::get` to return true if a specific string is accessed, you can just inherit from e.g. `reanimated::ShareableHostObject`, and that would type-match (`dynamic_cast`) in the `ShareableValue::adapt` function.

I agree that inheritance would be a better way of marking objects as Shareable - but how should we do this for host functions?

The other problem with the inheritance idea is that you add a hard dependency on a specific version of Reanimated. The idea here is that the consumer library (using this functionality) can make itself shareable with a simple property and would still work with an earlier version of REA. There are so many people out there having different versions of Reanimated.

In our case we offer Skia running on the JS Thread (without worklets) and then the possibility to add worklet support if you are willing to upgrade to the latest version of REA.

mrousavy:
> hard dependency on a specific version of Reanimated

you do that anyways when using the `ShareableValue::HOST_SHAREABLE` value

> would still work with an earlier version of REA

well then they'd have to type out the string manually... 🤷‍♂️

piaskowyk:
Ok, sure! Just let me know when it will be ready

jwajgelt:
After the `Shareable`s rewrite recently merged into v3, this would also need to be rewritten from scratch.
If you're still interested in this, feel free to open a new PR with an updated implementation.

PR #2564, opened 2021-10-21
fix: Android build.gradle doesn't work with flavors

### LINKED ISSUE #2412: Build on android fails for version 2.2.1
## Description

when running `cd android && ./gradlew clean && cd .. && react-native run-android --variant=devDebug --appIdSuffix=dev` I get the following error:

```
FAILURE: Build failed with an exception.

* Where:
Build file 'my-app/node_modules/react-native-reanimated/android/build.gradle' line: 85

* What went wrong:
A problem occurred evaluating project ':react-native-reanimated'.
> Task with path ':app:mergeDebugNativeLibs' not found in project ':app'.
```

[see in diff](https://github.com/software-mansion/react-native-reanimated/compare/2.2.0...2.2.1#diff-d260c332176b5c72008900aef1116a89a873451c7388a808c8d2d58be0865d38R85)

### Expected behavior

Should not crash 

Update from 2.2.0 to 2.2.1

## Package versions

- React Native:  `"react-native": "0.64.2",`
- React Native Reanimated: `"react-native-reanimated": "^2.2.1",`
- NodeJS: `v16.8.0`
- Xcode: N/A
- Java & Gradle: N/A

## Affected platforms

- [x] Android
- [ ] iOS
- [ ] Web

jtoce:
Fixes #2412

## Description

Fixed a build problem for Android users who have build flavors/variants:
```
FAILURE: Build failed with an exception.
* Where:
Build file '/Users/vagrant/git/node_modules/react-native-reanimated/android/build.gradle' line: 70
* What went wrong:
A problem occurred evaluating project ':react-native-reanimated'.
> Task with path ':app:mergeDebugNativeLibs' not found in project ':app'.
```
## Changes

Inside the function for `getCurrentFlavor()` the regex only matches against certain tasks:
```
Pattern pattern = Pattern.compile("(assemble|install|generate)(\\w+)(Release|Debug)")
```
Added `bundle` to this to address the comment from @wbercx. 

Note that if this match fails, the current flavor is returned as empty string. However, if you have no flavors, this also returns an empty string. This presents a problem below in the logic because it can't determine between the two. Modified this to instead return `NOT-FOUND` which is not a valid Android product flavor.

Since there are valid tasks (e.g. clean) that don't contain the current flavor, we don't want to execute the `merge` and `strip` tasks. So modified the if statement to ensure we only execute if there is a match. Again this is only a problem if you have flavors, since otherwise it can't the difference between an empty string match and non-match.

```
final NOTFOUND = "NOT-FOUND"
if(!NOTFOUND.equals(getCurrentFlavor()) && (!projectProperties.get("reanimated")
        || (projectProperties.get("reanimated") && projectProperties.get("reanimated").get("enablePackagingOptions")))
) {
```

See https://developer.android.com/studio/build/build-variants

Repro Steps:
1. Create Android build variants in `app/build.gradle` (e.g. demo, full)
2. Run `./gradlew clean`
(Alternatively, go to Android Studio > File > Sync Project with Gradle Files)

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android-npm/build.gradle
+++ b/android-npm/build.gradle
@@ -31,7 +31,7 @@ abstract class replaceSoTask extends DefaultTask {
 def getCurrentFlavor() {
     Gradle gradle = getGradle()
     String taskRequestName = gradle.getStartParameter().getTaskRequests().toString()
-    Pattern pattern = Pattern.compile("(assemble|bundle|install|generate)(\\w+)(Release|Debug)")
+    Pattern pattern = Pattern.compile("(assemble|bundle|install|generate)(\\w*)(Release|Debug)")
     Matcher matcher = pattern.matcher(taskRequestName)
 
     if(matcher.find()) {


saintpepsicola:
I still have this issue on 2.2.4 :(

rawatnaresh:
Not able to clean the build, is it because of this? I've flavours in my project
Screen Shot 2021-11-02 at 09 07 53

pke:
Yes it is related. Had this too.

Nagato717:
Fixed on 2.2.4
Using `yarn react-native-reanimated@2.2.4` to install. `@next` doesn't seem to have it yet.

ULazdins:
Hi!

It seems like I'm experiencing an error that might be related to these changes. The error appeared recently and the same command was working before and it seems that `react-native-reanimated` might have caused that.

`./gradlew bundleRelease` fails with an error:
```
FAILURE: Build failed with an exception.

* Where:
Build file '/Users/ula/Repos/lolo/node_modules/react-native-reanimated/android/build.gradle' line: 71

* What went wrong:
A problem occurred evaluating project ':react-native-reanimated'.
> Task with path ':app:mergeDebugNativeLibs' not found in project ':app'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

The error can be circumvented by explicitly using the flavor name in the bundle command, like  `./gradlew bundleStageRelease` of  `./gradlew bundleProdRelease` (or whatever your flavor names are).

sami-99:
> Hi!
> 
> It seems like I'm experiencing an error that might be related to these changes. The error appeared recently and the same command was working before and it seems that `react-native-reanimated` might have caused that.
> 
> `./gradlew bundleRelease` fails with an error:
> 
> ```
> FAILURE: Build failed with an exception.
> 
> * Where:
> Build file '/Users/ula/Repos/lolo/node_modules/react-native-reanimated/android/build.gradle' line: 71
> 
> * What went wrong:
> A problem occurred evaluating project ':react-native-reanimated'.
> > Task with path ':app:mergeDebugNativeLibs' not found in project ':app'.
> 
> * Try:
> Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
> 
> * Get more help at https://help.gradle.org
> ```
> 
> The error can be circumvented by explicitly using the flavor name in the bundle command, like `./gradlew bundleStageRelease` of `./gradlew bundleProdRelease` (or whatever your flavor names are).

same issue for me. I'm using react native 0.64.2 react native reanimated 2.2.4.

sergeymorkovkin:
Confirm: this is still reproducing for anything above 2.2.0. Patch was integrated.

PR #2565, opened 2021-10-22
Remove _bridge_renimated, fix #2555

### LINKED ISSUE #2555: RNGH+REA breaks occasionally due to a race condition
## Description

Occasionally, `react-native-gesture-handler`'s gesture recognizers will stop working in an app. RNGH itself will work correctly, as will REA, however the link between the two can break causing all gestures to stop working.

I've been unsuccessfully hunting this bug for months, since it's very, very hard to replicate in debug mode.

Today, I accidentally caught it in Simulator, and I'm quite sure I found the root cause, although I'm still unable to replicate this on demand.

The proximate cause is that RNGH tries to dispatch an event, hits this: https://github.com/software-mansion/react-native-reanimated/blob/e646dce7b85204cddeea4ef3e1ab971a85830562/ios/REAEventDispatcher.m#L10 and `_bridge_reanimated` is `nil`, so the call becomes a no-op.

After analysis of how `_bridge_reanimated` is created, I can think of only one possible cause: a race condition between creation and nullification of the `_bridge_reanimated` global.

It is created here: https://github.com/software-mansion/react-native-reanimated/blob/008ee9322f3d2ed3ffef1996cbab50039fb72a3d/ios/native/REAInitializer.mm#L38 and destroyed here: https://github.com/software-mansion/react-native-reanimated/blob/e646dce7b85204cddeea4ef3e1ab971a85830562/ios/REAModule.m#L20

This, however is a programming mistake, because: https://github.com/facebook/react-native/blob/d9689b90a4023efaa8c35824d63e8fe0c692b9fb/React/Base/RCTBridge.m#L258-L263 - due to async nature of invalidation, most of the time, invalidation will happen before new setup, but if the stars align in just the wrong way… it won't ;) I've already been personally bitten hard by this quirk of RN invalidation in https://github.com/Nozbe/watermelondb ;) 

As you can see, this bug can only possibly be triggered on bridge reload. So in most RN apps, it's impossible for it to occur in production, and is easily mitigated by a ⌘R in development. However, this affects my app in particular because it can reload its bridge in some situations in production.

### Expected behavior

⬆️

⬆️

n/a

## Package versions

2.2.x, 2.3x

## Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

radex:
## Description

Fixes #2555. 

## Changes

Perhaps I'm missing something, but it seems to me that `_bridge_renimated` is a vestigial organ that no longer serves any practical purpose. Its use in REAEventDispatcher is unnecessary, since we already assign RCTEventDispatcher.bridge with the same value, and its only other use in createReanimatedModule is a single call where we can pass bridge as an argument.

Since bridge is now passed as an explicit dependency, a race condition is no longer possible.

I've realized that I can reproduce #2555 quite easily by repeatedly ⌘Ring in the simulator in quick succession. I've easily managed to get to the point where REAEventDispatcher has a null `_bridge_reanimated`, but `[self bridge]` is correct, confirming that the fix works.

## Screenshots / GIFs

n/a

Reload the app in simulator repeatedly and see if gestures break as per #2555.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/ios/REAEventDispatcher.m
+++ b/ios/REAEventDispatcher.m
@@ -7,7 +7,7 @@
 
 - (void)sendEvent:(id<RCTEvent>)event
 {
-  [[_bridge_reanimated moduleForName:@"ReanimatedModule"] eventDispatcherWillDispatchEvent:event];
+  [[[self bridge] moduleForName:@"ReanimatedModule"] eventDispatcherWillDispatchEvent:event];
   [super sendEvent:event];
 }
 
--- a/ios/REAModule.h
+++ b/ios/REAModule.h
@@ -7,8 +7,6 @@
 
 #import "REAValueNode.h"
 
-extern RCTBridge *_bridge_reanimated;
-
 @interface REAModule : RCTEventEmitter <RCTBridgeModule, RCTEventDispatcherObserver, RCTUIManagerObserver>
 
 @property (nonatomic, readonly) REANodesManager *nodesManager;
--- a/ios/REAModule.m
+++ b/ios/REAModule.m
@@ -6,8 +6,6 @@
 
 typedef void (^AnimatedOperation)(REANodesManager *nodesManager);
 
-RCTBridge *_bridge_reanimated = nil;
-
 @implementation REAModule {
   NSMutableArray<AnimatedOperation> *_operations;
   REATransitionManager *_transitionManager;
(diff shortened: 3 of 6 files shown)


piaskowyk (approved):
Thanks, good catch! 🚀

PR #2566, opened 2021-10-22
Add `useLayoutAnimation` hook

mrousavy:
## Description

Adds a new hook: `useLayoutAnimation`.

In the docs, Layout Animations are always created _without types_ and are _not memoized_ (problem in function components, long re-renders).
For example:

```ts

function CardView() {
  const exiting = (startingValues) => {
                   // ^ startingValues is any
    'worklet';
    // ^ user has to put 'worklet' here

    const animations = {
      originX: withTiming(width, { duration: 3000 }),
      opacity: withTiming(0.5, { duration: 2000 }),
    };
    const initialValues = {
      originX: startingValues.originX,
      opacity: 1,
    };
    // return type is missing, but that's another thing: it isn't typed.
  }


  return (
    <Animated.View
      style={[styles.animatedView]} exiting={exiting} >
      <Text> Card Example </Text>
    </Animated.View>
  );
}
```

With `useLayoutAnimation` this can be avoided.

```ts
const animation = useLayoutAnimation((startingValues) => {
  return {
    animations: {},
    initialValues: {},
  }
}
```

Benefits:
* we don't need to use `'worklet'` anymore
* parameter (`startingValues`) is typed
* return value is typed
* Function is memoized for faster re-renders

## Checklist

- [x] Updated TS types
- [x] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/hook/Hooks.ts
+++ b/src/reanimated2/hook/Hooks.ts
@@ -1,4 +1,4 @@
-import { useCallback, useMemo } from 'react';
+import { useCallback, useCallback } from 'react';
 import { DependencyList } from './commonTypes';
 import { useAnimatedStyle } from './useAnimatedStyle';
 import type {} from '../../'
@@ -15,8 +15,8 @@ export function useWorkletCallback<A extends unknown[], R>(
 }
 
 /**
- * `useLayoutAnimation` just returns the memoized function, so it's just a typed helper with useMemo.
+ * `useLayoutAnimation` just returns the memoized function, so it's just a typed helper with useCallback.
  */
-export const useLayoutAnimation = useMemo
+export const useLayoutAnimation = useCallback
 
 export { useEvent, useHandler } from './utils';


nandorojo:
> Because that hook has been added to babel plugin.

Is there a way for consuming libraries to add their own functions to this? For example, `MotiPressable` exposes a number of custom hooks / functions that need to be workletized.

> Btw, @mrousavy, there is useWorkletCallback hook in Reanimated.

Is this public or only internal?

src/reanimated2/hook/Hooks.ts
```diff
@@ -13,4 +14,9 @@ export function useWorkletCallback<A extends unknown[], R>(
   return useCallback(fun, deps ?? []);
 }
 
+/**
+ * `useLayoutAnimation` just returns the memoized function, so it's just a typed helper with useMemo.
+ */
+export const useLayoutAnimation = useMemo
```
nandorojo, line 20: Shouldn't this be `useCallback`, since it's receiving an argument?

mrousavy: Right, thanks!

jwajgelt:
This PR has become a bit stale, and we don't see a clear use for including this hook in `reanimated`.
As such, we'll close this thread for now.

PR #2580, opened 2021-10-27
Operations order

### LINKED ISSUE #2345: Animated ScrollView contentOffset yields wrong animated value on mount inconsistantly.
## Description

When using `contentInset` and `contentOffset`, 

```
contentInset={{ top: 100 }}
contentOffset={{ y: -100, x: 0 }}
```

with useAnimatedScrollHandler,

```
  const scroll = useSharedValue(0);

  const scrollHandler = useAnimatedScrollHandler(({ contentOffset: { y } }) => {
    scroll.value = y;
  });
```

in combination with an animated style,

```
const translationY = useDerivedValue(() => {
    const startScroll = Platform.OS === 'ios' ? -100 : 0;
    const endScroll = Platform.OS === 'ios' ? 0 : 100;

    return interpolate(
      scroll.value,
      [startScroll, endScroll],
      [0, -100],
      Extrapolate.CLAMP,
    );
  }, [scroll]);

  const animatedHeaderStyle = useAnimatedStyle(
    () => ({
      transform: [
        {
          translateY: translationY.value,
        },
      ],
    }),
    [translationY],
 );
```

the animated style returns the scroll position at `0` sometimes on mount, despite the scroll updating to the correct position. Moving the scrollview corrects this and the translation is corrected.

This can be worked around by adding a hacky fix, but there still is a delay at times where the view is in the wrong (`0`) position for a second.

```
 style={[
          animatedHeaderStyle,
          {
            transform: [
              {
                translateY: translationY.value, // Fix for ScrollView offset and reanimated issue on mount
              },
            ],
          },
        ]}
```

### Expected behavior

When using `contentInset` and `contentOffset` (on Android paddingTop is used instead of contentInset for the same effect), the initial scroll position is calculated based on the contentOffset.

Create a scrollview and animated view as described and apply the provided styles. Open the ios simulator, the animated view will appear in the incorrect location inconsistently on mount.

https://snack.expo.dev/@nhannah/reanimated-2-scroll-offset-issue

## Package versions

- React Native: 0.64.0
- NodeJS: 16.8.0

## Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

piaskowyk:
## Description

Fixes #2345.

In some cases is possible to call `updateProps` on the not yet mounted component. These updates were overridden by React Layout props. I detect this situation and save new props to buffer, and schedule updates of these props after mounting of component.

<details>
<summary>Code to reproduce</summary>

```js
import { View, StyleSheet, Platform } from 'react-native';
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  useAnimatedScrollHandler,

export default function Repro() {
  const translationY = useSharedValue(200);

  const scrollHandler = useAnimatedScrollHandler(({ contentOffset: { y } }) => {
    translationY.value = 200 + y + 20;
  });

  const animatedHeaderStyle = useAnimatedStyle(() => {
    return {
      position: 'absolute',
      backgroundColor: 'red',
      height: 200,
      top: 0,
      left: 0,
      right: 0,
      paddingTop: 100,
      width: translationY.value,
    };
  }, [translationY]);

  console.log("Render", "----------------------------------")
  
  return (
      <Animated.ScrollView
        contentInset={{ top: 200 }}
        contentOffset={{
          y: -200,
          x: 0,
        }}
        contentContainerStyle={styles.contentContainer}
        style={styles.container}
        onScroll={scrollHandler}
        scrollEventThrottle={16}>
      </Animated.ScrollView>
      <Animated.View style={animatedHeaderStyle}>
      </Animated.View>
    </View>
  );
}

  container: {
    flex: 1,
  },
  contentContainer: {
    paddingTop: Platform.OS === 'android' ? 200 : undefined,
  },
  blueSpace: {
    height: 500,
    backgroundColor: 'blue',
  },
  middleLine: {
    height: 10,
    backgroundColor: 'green',
  },
});
```
</details>


ios/REANodesManager.m
```diff
@@ -552,4 +586,34 @@ - (NSString *)obtainProp:(nonnull NSNumber *)viewTag propName:(nonnull NSString
...
+  }
+
+  NSMutableArray *flushedTags = [NSMutableArray new];
+  for (NSNumber *tag in _componentUpdateBuffer) {
+    ComponentUpdate *componentUpdate = _componentUpdateBuffer[tag];
+    if (componentUpdate == Nil) {
+      return;
+    }
+    __weak typeof(self) weakSelf = self;
+    RCTExecuteOnMainQueue(^void() {
```
kmagiera, line 602: It'd be better if we could get the whole flushing to happen on the UI thread. Now we enqueue each individual command. It also seem like you just add everything to flushedTags – so what's the point of it? Why not just clear _componentUpdateBuffer instead of removing tags from flushedTags?

On top of that it looks like you modify _componentUpdateBuffer from two thread w/o any synchronization. I think that moving flush logic to ui thread would help with this as well

kmagiera: I wonder if there is other way than using `uiManagerDidPerformMounting` such that we can get our code to run in the same UI loop. Enqueueing here would only result in changes being reflected on the next frame, no?

piaskowyk: I packed all updates in one UI block and I moved all modifications of _componentUpdateBuffer to the UI thread.

ios/REAModule.m
```diff
@@ -156,6 +156,7 @@ - (void)addOperationBlock:(AnimatedOperation)operation
 
 - (void)uiManagerWillPerformMounting:(RCTUIManager *)uiManager
 {
+  [_nodesManager tryToFlushUpdateBuffer];
```
kmagiera, line 159: I still think it'd be better if we could run these updates on UI thread w/o using RCTExecuteOnMainQueue which would postpone these updates until the next frame

ios/REANodesManager.m
```diff
@@ -505,6 +520,25 @@ - (void)updateProps:(nonnull NSDictionary *)props
       ofViewWithTag:(nonnull NSNumber *)viewTag
            withName:(nonnull NSString *)viewName
 {
+  if (_viewRegistry[viewTag].superview == nil) {
```
kmagiera, line 523: I think it is possible that we reach this stage and component is already mounted but still has entries in snapshot list (e.g. we run updates before and after mount + because we enqueue flushing with RCTExecuteOnMainQueue it may actually arrive later). As a consequence we may run older updates after these new ones. It'd be safer to check here if _componentUpdateBuffer contain updates

ios/REANodesManager.m
```diff
@@ -552,4 +585,31 @@ - (NSString *)obtainProp:(nonnull NSNumber *)viewTag propName:(nonnull NSString
   return result;
 }
 
+- (void)tryToFlushUpdateBuffer
+{
+  if (_componentUpdateBuffer.count == 0) {
```
kmagiera, line 590: I think this var is modified from ui thread and is being read here from possibly any thread

ios/REANodesManager.m
```diff
@@ -505,6 +520,24 @@ - (void)updateProps:(nonnull NSDictionary *)props
       ofViewWithTag:(nonnull NSNumber *)viewTag
            withName:(nonnull NSString *)viewName
 {
+  ComponentUpdate *lastSnapshot = _componentUpdateBuffer[viewTag];
+  if (_viewRegistry[viewTag].superview == nil || lastSnapshot != nil) {
```
kmagiera, line 524: It'd be good to maybe extract this to a method and call it sth like "isNativeViewFullyMounted" such that when doing this check it is easier to tell what exactly we are testing

ios/REANodesManager.m
```diff
@@ -552,4 +585,31 @@ - (NSString *)obtainProp:(nonnull NSNumber *)viewTag propName:(nonnull NSString
...
 }
 
+- (void)tryToFlushUpdateBuffer
+{
+  if (_componentUpdateBuffer.count == 0) {
+    return;
+  }
+
+  __weak typeof(self) weakSelf = self;
+  RCTExecuteOnMainQueue(^void() {
```
kmagiera, line 595: As discussed offline, we need to make sure the updates are flushed in the same UI loop in which the component is mounted. There are two problems here:
1) testing for _componentUpdateBuffer is not thread safe as the prop is not synchronized and being updated from other threads
2) AFAIK RCTExecuteOnMainQueue will schedule execution for the following frame and not for the current one. I think you may try using addUIBlock from UIManager, but maybe there are also other ways

piaskowyk: 1. I added synchronization for `_componentUpdateBuffer`.

2. I used the method `addUIBlock` and according to https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/React/Modules/RCTUIManager.m#L1171 even if we use it inside of`uiManagerWillPerformMounting`, the updates should be applied during the same frame.

Hirbod:
Is anything holding back this PR?

piaskowyk:
@Hirbod I am just waiting for the review.

ios/REANodesManager.m
```diff
@@ -553,4 +595,36 @@ - (NSString *)obtainProp:(nonnull NSNumber *)viewTag propName:(nonnull NSString
   return result;
 }
 
+- (void)tryToFlushUpdateBuffer
```
kmagiera, line 598: I'd rename it to `maybeFlushUpdateBuffer` as `try` suggest that we are ok with it failing

ios/REANodesManager.m
```diff
@@ -553,4 +595,36 @@ - (NSString *)obtainProp:(nonnull NSNumber *)viewTag propName:(nonnull NSString
...
+{
+  [_componentUpdateBufferLock lock];
```
kmagiera, line 600: What would be helpful is if we add assertions to methods that'd indicate on which thread given method is allowed to be executed. (e.g., `RCTAssertMainQueue`)

kmagiera (commented):
If I understand this correctly we don't actually want to synchronize the access to _componentUpdateBuffer here. The only thing we want is to make some atomic flag that'd allow us to indicate there is something in that buffer such that these updates can be picked up when batch is finished processing. For that purpose it'd be safer and easier to use atomic operations and only synchronize that part. Note that the code that writes to the buffer is going to be executed more often than the code that only needs to know if there was anything added to the buffer. Moreover, if the latter reads that there is something but we already flushed it, it isn't going to have any side effects and the only issue is that we may sometimes add UI block that does nothing.

kmagiera (approved):
As discussed, can we rename _isComponentUpdateBufferEmtpy to carry some meaning about what does setting this flag would mean. Also make that meaning "reverse", like for example "shouldRunUpdatesAfterFLush"

Latropos:
Fixes https://github.com/software-mansion/react-native-reanimated/issues/2900 as well

PR #2591, opened 2021-11-01
slide in up wrong init value fix

ArekChr:
## Description

SlideInUp behaves the same as SlideInDown

## Changes

Changed initial value for originY to slide in up, opposite as in SlideInDown


piaskowyk (approved):
Thanks! 🚀

PR #2595, opened 2021-11-02
Propose Expo integration

Kudo:
## Description

Previously, to support reanimated in Expo SDK, we had reanimated builtin in our template even user don't use it. In newer SDK, we plan to support native modules as opt-in without extra setup; we proposed https://github.com/expo/expo/pull/14883 and this pr is for reanimated integration.

## Changes

- Add `expo-module.config.json` for expo autolinking
- Add `android/expo` files dedicated for expo integration
- Just add single line to existing build.gradle for the integration.

We will check the gradle dependencies to see if there are `expo` project and only enable the integration then.
  
## Test Expo integration

[demo repo](https://github.com/Kudo/ExpoAndroidWrapper) which based on expo sdk43 bare template with additional changes:
1. adapts https://github.com/expo/expo/pull/14883 patch because it doesn't publish yet.
2. remove `ReanimatedJSIModulePackage` from MainApplication.java
3. add this pr's patch by patch-package.

## Test classic RN app without Expo
makes sure not to break existing functionalities

```sh
$ npx react-native init RN066 --version 0.66
$ yarn add react-native-reanimated
# patch this pr's change into node_modules/react-native-reanimated
$ ./gradlew :app:assembleDebug
```

## Checklist

- n/a Updated TS types
- n/a Added TS types tests
- n/a Added unit / integration tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/expo-module.config.json
+++ b/expo-module.config.json
@@ -1,0 +1,4 @@
+{
+  "name": "react-native-reanimated",
+  "platforms": ["android"]
+}
--- a/android-npm/build.gradle
+++ b/android-npm/build.gradle
@@ -107,3 +107,5 @@ while (!aar.exists()) {
 }
 
 artifacts.add("default", aar)
+
+apply from: './expo/linking.gradle'
--- a/android-npm/expo/linking.gradle
+++ b/android-npm/expo/linking.gradle
@@ -1,0 +1,34 @@
+def isExpoLinked = false
+rootProject.getSubprojects().forEach({ project ->
+    if (project.plugins.hasPlugin("com.android.application")) {
+        if (project.configurations.implementation.getDependencies().find { dep -> dep.name == "expo" }) {
+            isExpoLinked = true
+        }
+    }
+})
+
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
+if (isExpoLinked) {
+    apply plugin: 'com.android.library'
+
+    android {
+        compileSdkVersion safeExtGet('compileSdkVersion', 30)
+
+        sourceSets {
+            main.manifest.srcFile 'src/AndroidManifest.xml'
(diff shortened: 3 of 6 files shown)


expo-module.config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "react-native-reanimated",
+  "platforms": ["android"]
```
tsapeta, line 3: This enables the library to be autolinked by our scripts — are you sure we want to do this? Since it may cause side effects, I'd recommend to remove it. I guess we will need to introduce another field to disable autolinking 🤔 Also `name` is unnecessary as far as I remember

Kudo: we use expo autolinking because we need the generated `ExpoModulesPackageList.java` to get the list of handlers: https://github.com/expo/expo/blob/de6c8d2f8535df5c8a9620440daa30b9dafb30b2/packages/expo/android/src/main/java/expo/modules/ReactNativeHostWrapper.kt#L21-L22

it looks like gradle can handle duplicated dependencies so expo autolinking can coexist with rn-cli autolinking. but i'm fine if you have other thought than it.

piaskowyk (approved):

Kudo:
@piaskowyk super thanks for your help to land this pr. i'm sorry i'm working on some improvement. by leverage the java 8 default interface implementation, we can remove the unused override methods. this will reduce the future maintenance cost if our interfaces changed. i'll create a pr for this about one hour later.

zhiqingchen:
hi, @Kudo , What does this mean?
```
Just add single line to existing build.gradle for the integration.
```

Kudo:
> Just add single line to existing build.gradle for the integration.

https://github.com/software-mansion/react-native-reanimated/pull/2595/files#diff-d260c332176b5c72008900aef1116a89a873451c7388a808c8d2d58be0865d38R111

PR #2600, opened 2021-11-03
Detach old animated styles and props

tomekzaw:
## Description

This pull request fixes a bug appearing when using animated styles (or props) conditionally. The root cause is that animated styles (or props) are not properly detached.

| Before | After |
|-|-|
| <video src="https://user-images.githubusercontent.com/20516055/140052148-f0584f2d-eeb0-45b8-af2f-dcfa12164f67.mp4" /> | <video src="https://user-images.githubusercontent.com/20516055/140052808-d3f566fd-4ac3-4d0a-908f-e9113d1c2691.mp4" /> |
| <video src="https://user-images.githubusercontent.com/20516055/140053185-d1a5a4d4-66a8-41e8-b0fe-7601248bd638.mp4" /> | <video src="https://user-images.githubusercontent.com/20516055/140052901-d5afada0-b782-43e3-ae38-752806091d05.mp4" /> |

## Changes

- Added missing calls to detach animated styles and props

<details>
<summary>AnimatedStylesExample.tsx</summary>

```tsx
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withTiming,

import React from 'react';

  const x = useSharedValue(0);

  const style1 = useAnimatedStyle(() => {
    return { marginLeft: 250 * x.value };
  });

  const style2 = useAnimatedStyle(() => {
    return { marginTop: 250 * x.value };
  });

  const handleToggleSharedValue = () => {
    x.value = withTiming(1 - x.value, { duration: 1000 });
  };

  const handleToggleState = () => {
    setState((state) => state + 1);
  };

  return (
    <>
      <View style={{ paddingVertical: 50 }}>
        <Button onPress={handleToggleSharedValue} title="Toggle shared value" />
        <Button onPress={handleToggleState} title="Toggle state" />
      </View>
      <Animated.View
        style={[
          {
            width: 100,
            height: 100,
            backgroundColor: state % 2 === 0 ? 'red' : 'blue',
          },
          state % 2 === 0 ? style1 : style2,
        ]}
      />
    </>
  );
}
```
</details>

<details>
<summary>AnimatedPropsExample.tsx</summary>

```tsx
import Animated, {
  useAnimatedProps,
  useSharedValue,
  withTiming,
import Svg, { Path } from 'react-native-svg';

import React from 'react';

export default function AnimatedPropsExample() {

  const radius = useSharedValue(50);

  const animatedProps1 = useAnimatedProps(() => {
    // draw a circle
    const path = `
    M 100, 100
    m -${radius.value}, 0
    a ${radius.value},${radius.value} 0 1,0 ${radius.value * 2},0
    a ${radius.value},${radius.value} 0 1,0 ${-radius.value * 2},0
    `;
    return {
      d: path,
    };
  });

  const animatedProps2 = useAnimatedProps(() => {
    // draw a circle
    const path = `
    M 200, 100
    m -${radius.value}, 0
    a ${radius.value},${radius.value} 0 1,0 ${radius.value * 2},0
    a ${radius.value},${radius.value} 0 1,0 ${-radius.value * 2},0
    `;
    return {
      d: path,
    };
  });

  const handleToggleSharedValue = () => {
    radius.value = withTiming(50 - radius.value, { duration: 1000 });
  };

  const handleToggleState = () => {
    setState((state) => state + 1);
  };

  return (
    <>
      <View style={{ paddingVertical: 50 }}>
        <Button onPress={handleToggleSharedValue} title="Toggle shared value" />
        <Button onPress={handleToggleState} title="Toggle state" />
      </View>
      <Svg>
        <AnimatedPath
          animatedProps={state % 2 === 0 ? animatedProps1 : animatedProps2}
          fill={state % 2 === 0 ? 'red' : 'green'}
        />
      </Svg>
    </>
  );
}
```
</details>

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -88,6 +88,10 @@ function flattenArray<T>(array: NestedArray<T>): T[] {
   return resultArr;
 }
 
+function onlyAnimatedStyles(styles: StyleProps[]) {
+  return styles.filter((style) => style?.viewDescriptors);
+}
+
 const has = <K extends string>(
   key: K,
   x: unknown
@@ -167,6 +171,7 @@ export default function createAnimatedComponent(
   > {
     _invokeAnimatedPropsCallbackOnMount = false;
     _styles: StyleProps[] | null = null;
+    _animatedProps?: Partial<AnimatedComponentProps<AnimatedProps>>;
     _viewTag = -1;
     _isFirstRender = true;
     animatedStyle: { value: StyleProps } = { value: {} };
@@ -255,9 +260,7 @@ export default function createAnimatedComponent(
         }
       } else if (this._viewTag !== -1 && this._styles !== null) {
         for (const style of this._styles) {
-          if (style?.viewDescriptors) {
-            style.viewDescriptors.remove(this._viewTag);
-          }
+          style.viewDescriptors.remove(this._viewTag);
         }
         if (this.props.animatedProps?.viewDescriptors) {
           this.props.animatedProps.viewDescriptors.remove(this._viewTag);
@@ -387,8 +390,15 @@ export default function createAnimatedComponent(
     }
 
     _attachAnimatedStyles() {
-      const styles = flattenArray<StyleProps>(this.props.style ?? []);
+      const styles = this.props.style
+        ? onlyAnimatedStyles(flattenArray<StyleProps>(this.props.style))
+        : [];
+      const prevStyles = this._styles;
       this._styles = styles;
+
+      const prevAnimatedProps = this._animatedProps;
+      this._animatedProps = this.props.animatedProps;
+
       let viewTag: number | null;
       let viewName: string | null;
       if (Platform.OS === 'web') {
@@ -412,7 +422,7 @@ export default function createAnimatedComponent(
         // update UI props whitelist for this view
         if (
           hostInstance &&
-          this._hasReanimated2Props(styles) &&
+          (this.props.animatedProps?.viewDescriptors || styles.length) &&
           hostInstance.viewConfig
         ) {
           adaptViewConfig(hostInstance.viewConfig);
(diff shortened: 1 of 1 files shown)


src/createAnimatedComponent.tsx
```diff
@@ -412,32 +422,51 @@ export default function createAnimatedComponent(
         // update UI props whitelist for this view
         if (
           hostInstance &&
-          this._hasReanimated2Props(styles) &&
+          (this.props.animatedProps?.viewDescriptors || styles.length) &&
```
kmagiera, line 425: I'd extract this to a variable defined before the if statement to make it clear what this part of the condition means:

```
const hasReanimated2Props = this.props.animatedProps?.viewDescriptors || styles.length;

if (hasReanimated2Props && hostInstance?.viewConfig) {
```

src/createAnimatedComponent.tsx
```diff
@@ -412,32 +422,51 @@ export default function createAnimatedComponent(
...
-              ...this.animatedStyle.value,
-              ...style.initial.value,
-            };
-            style.animatedStyle.current = this.animatedStyle;
+      // remove old styles
+      if (prevStyles) {
+        const hasOneSameStyle =
+          styles.length === 1 &&
+          prevStyles.length === 1 &&
+          styles[0] === prevStyles[0]; // optimization
```
kmagiera, line 438: document that in most of the cases views have only a single animated style and it remains unchanged (it is what useAnimatedStyle returns)

Also it'd be worth checking if we can actually use equality checks to compare useAnimatedStyle outputs

src/createAnimatedComponent.tsx
tomekzaw: Done in f6b8560835b25f1b3463a3bce66b2265060bc3ac

src/createAnimatedComponent.tsx
tomekzaw: > Also it'd be worth checking if we can actually use equality checks to compare useAnimatedStyle outputs

Indeed, equality check didn't work as expected. Fixed in 21f456f1e41d5a94b82e95376fb26957864c36e7. For now, comparing `style.viewsRef` should be enough, as this prop is assigned only once (so the reference is the same). In the future, we may introduce some auto-incremented ID field to `useAnimatedStyle` output.

piaskowyk (approved):
🚀

mrousavy:
It looks like this PR has introduced a performance issue, especially noticeable on older iPhones when animating a lot of values (e.g. in a Graph).
I'll investigate further.

tomekzaw:
@mrousavy Thanks for letting us know. It seems like `_attachAnimatedStyles` is called only during render so it should not affect the performance of animations. Can you please share more details with us? In particular, please clarify if the performance issue is related to animated styles or animated props. A repro case would be really appreciated as well.

mrousavy:
Yes, I'm investigating it! How do you guys profile performance in REA? systrace isn't very helpful on Android. Do you use the Xcode instruments on iOS?

mrousavy:
I can definitely _feel_ a regression in performance after upgrading from beta 3 to beta 4, not sure if this PR is the cause of that. What's your discord #? I can send you the file in DM.

tomekzaw:
@mrousavy I don't think we have any particular setup for profiling. We use Reanimated heavily in numerous apps that we develop here at Software Mansion so we constantly keep an eye on its performance. Other than that, there are some heavy-duty examples (see https://twitter.com/kzzzf/status/1421104290085576710).

> I can definitely feel a regression in performance after upgrading from beta 3 to beta 4, not sure if this PR is the cause of that.

Can you please undo the changes from this PR locally and check if you can still feel the regression?

PR #2601, opened 2021-11-04
Fix TS generate

piaskowyk:
## Description

Just simple TS fix 🤷‍♂️


tomekzaw (approved):

jakub-gonet (commented):
You probably want `!` operator instead of casting here.

Issue #2554, opened 2021-10-20
Exception thrown while executing UI block: Attempted to remove non-existent subview


nomi9995:
<img src="https://user-images.githubusercontent.com/36044436/138065953-6d9d1085-462c-4cf9-99c3-e9630c848345.png" width="300" />

## Description
I am using [react-native-shimmer](https://github.com/oblador/react-native-shimmer). when I update `react-native@0.66.1` and `react-native-reanimated@2.3.0-alpha.3` then I am receiving the below error 
```
Exception thrown while executing UI block: Attempted to remove non-existent subview
```
at

https://github.com/software-mansion/react-native-reanimated/blob/ce76e66aaf47c26bb1ff818d72a50b55e092ca05/ios/LayoutReanimation/REAAnimationsManager.m#L101

because it has conflict with [react-native-shimmer](https://github.com/oblador/react-native-shimmer)

[RNShimmeringView.m#L28](https://github.com/oblador/react-native-shimmer/blob/620a9e899df40e1bef38127d4ef61384e1a195c5/ios/RNShimmeringView.m#L28)

### Expected behavior
Should not throw the error.

### Actual behavior & Steps To Reproduce
Repro: https://github.com/nomi9995/ReproNonExistentsubview
Throws an error but still works if I just close it.

just installed  `Package versions` which I added below. 
then below code will reproduce this error
```
import React, {useState} from 'react';
import {SafeAreaView, StyleSheet, Text, View, Button} from 'react-native';
import Shimmer from 'react-native-shimmer';

const App = () => {
  const [isVisible, setIsVisible] = useState<boolean>(true);

  return (
    <SafeAreaView style={styles.container}>
      {isVisible && (
        <Shimmer>
          <Text>Loading...</Text>
        </Shimmer>
      )}
      <Button
        onPress={() => setIsVisible(!isVisible)}
        title={isVisible ? 'Hide' : 'Visible'}
      />
    </SafeAreaView>
  );
};

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
});

export default App;
```

## Package versions

- React Native: ^v0.66.1
- React Native Reanimated: ^2.3.0-alpha.3
- react-native-shimmer: ^0.6.0

## Affected platforms

- [ ] Android
- [X] iOS
- [ ] Web

piaskowyk:
[...]

PR #2624, opened 2021-11-12
Handle removing exception

piaskowyk:
## Description

I am not convinced to fix it by reanimated side. IMO this assertion is not necessary and handling it this way looks like dirty hacks for edge cases. I ask them to remove this assertion. in spite of it, I prepared a fix.

Handle Assert exception during disappearing of a component. Caused by https://github.com/oblador/react-native-shimmer/blob/620a9e899df40e1bef38127d4ef61384e1a195c5/ios/RNShimmeringView.m#L26

Fixes https://github.com/software-mansion/react-native-reanimated/issues/2554

Related: https://github.com/oblador/react-native-shimmer/pull/33


kmagiera (approved):
I support this change. I think it does not make sense to catch unchecked exceptions thrown by other libraries in reanimated. Especially that in this particular case the assertion does not really assert against anything and the method could proceed as normal w/o it. I think what would make sense for cases like that is what we discussed offline, that is an ability to disable exiting animations in subtrees

PR #2646, opened 2021-11-19
[Web] Fix 2589: `null` in `useDerivedValue` 

### LINKED ISSUE #2589: [Web] Can't use `null` inside of `useDerivedValue`
## Description

If you pass `null` as a variable from the JS context to `useDerivedValue`, it breaks on web.

### Expected behavior

`null` should work as a variable in `useDerivedValue`.

1. Create a variable pointing to `null` in the JS context.
2. Inside of `useDerivedValue`, access that variable.
3. Run it on Web

Snack: https://snack.expo.dev/@nandorojo/derived-value-web-issue

Screen Shot 2021-10-31 at 2 31 05 PM

Code:

```tsx
import React from 'react'
import { useDerivedValue } from 'react-native-reanimated'
 
export default function HelloWorld() {
  const broken = null
  useDerivedValue(() => {
    const d = broken
  })
  return <></>
} 
```

To fix it, turn `broken` into 0. Or, comment it out:

Screen Shot 2021-10-31 at 2 31 49 PM

## Package versions

- React Native: 0.64.2
- NodeJS:
- Xcode:
- Java & Gradle:

## Affected platforms

- [ ] Android
- [ ] iOS
- [x] Web

## Added Context

It works fine on iOS:

image

kkafar:
## Description

Fixes #2589

* https://github.com/software-mansion/react-native-reanimated/issues/2589

## Changes

[`extractMutableFromArray` function](https://github.com/software-mansion/react-native-reanimated/blob/12093cbe04d978b2ef619531755ef7d472242cd9/src/reanimated2/js-reanimated/Mapper.ts#L40) now returns empty array when `array` is `null | undefined`

Use this code to see that it now works on web:

```tsx
import React from 'react'
import { useDerivedValue } from 'react-native-reanimated'
 
export default function HelloWorld() {
  const broken = null
  useDerivedValue(() => {
    const d = broken
  })
  return (<Text>Test</Text>)
}
```

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/src/reanimated2/js-reanimated/Mapper.ts
+++ b/src/reanimated2/js-reanimated/Mapper.ts
@@ -43,7 +43,9 @@ export default class Mapper<T> {
     const res: MutableValue<T>[] = [];
 
     function extractMutables(value: NestedObjectValues<MutableValue<T>>) {
-      if (value instanceof MutableValue) {
+      if (value == null) {
+        // do nothing
+      } else if (value instanceof MutableValue) {
         res.push(value);
       } else if (Array.isArray(value)) {
         value.forEach((v) => extractMutables(v));
--- a/src/reanimated2/js-reanimated/Mapper.ts
+++ b/src/reanimated2/js-reanimated/Mapper.ts
@@ -44,7 +44,7 @@ export default class Mapper<T> {
 
     function extractMutables(value: NestedObjectValues<MutableValue<T>>) {
       if (value == null) {
-        // do nothing
+        return;
       } else if (value instanceof MutableValue) {
         res.push(value);
       } else if (Array.isArray(value)) {


src/reanimated2/js-reanimated/Mapper.ts
```diff
@@ -43,7 +43,9 @@ export default class Mapper<T> {
     const res: MutableValue<T>[] = [];
 
     function extractMutables(value: NestedObjectValues<MutableValue<T>>) {
-      if (value instanceof MutableValue) {
+      if (value == null) {
+        // do nothing
```
piaskowyk, line 47:
```suggestion
        return;
```

piaskowyk (changes requested):

piaskowyk (approved):

PR #2651, opened 2021-11-22
Disable by default layout animation

piaskowyk:
## Description

I added the possibility to disable layout animations. By default layout animations are disabled, but you don't need to turn it on manually, because these will be enabled if you will use the `entering`/`exiting`/`layout` prop in an animated component. You also have the possibility to disable/enable it manually.

## Example
```js
import { enableLayoutAnimations } from 'react-native-reanimated';

enableLayoutAnimations(false);
```

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -17,13 +17,21 @@ import './reanimated2/layoutReanimation/LayoutAnimationRepository';
 import invariant from 'invariant';
 import { adaptViewConfig } from './ConfigHelper';
 import { RNRenderer } from './reanimated2/platform-specific/RNRenderer';
-import { makeMutable, runOnUI, setEnableFeatures } from './reanimated2/core';
+import {
+  makeMutable,
+  runOnUI,
+  enableLayoutAnimations,
+} from './reanimated2/core';
 import {
   DefaultEntering,
   DefaultExiting,
   DefaultLayout,
 } from './reanimated2/layoutReanimation/defaultAnimations/Default';
-import { isJest, isChromeDebugger } from './reanimated2/PlatformChecker';
+import {
+  isJest,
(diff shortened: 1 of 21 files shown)


src/reanimated2/core.ts
```diff
@@ -385,3 +385,36 @@ if (!NativeReanimatedModule.useOnlyV1) {
...
+      const userFlag = userFeaturesConfig[key];
+      if (userFlag !== false) {
+        filteredOptions[key] = option[key];
+      }
+    }
+  }
+  if (!filteredOptions) {
+    return;
+  }
+  NativeReanimatedModule.setEnableFeatures(filteredOptions);
```
piaskowyk, line 419: You need to add implementation for web

piaskowyk: done 👍

piaskowyk: thanks

src/createAnimatedComponent.tsx
```diff
@@ -496,6 +496,7 @@ export default function createAnimatedComponent(
           (this.props.layout || this.props.entering || this.props.exiting) &&
           tag != null
         ) {
+          setEnableFeatures({ layoutAnimation: true }, false);
```
piaskowyk, line 499: It can break Chrome Debugger support. Check it, please!

kmagiera (approved):
This looks ok, however, I think it'd be easier to add just one flag specifically for turning on layout animation as opposed to designing more generic solution to it.

If instead of `setEnableFeatures` we'd just have `enableLayoutAnimations` function we could:
1) not really care about web implementation too much. Layout Animations does not work on web so we could just no-op there
2) we now call `setEnableFeatures` on every component render. If it was just setting a single boolean variable we'd avoid additional complexity of merging and filtering dicts

PR #2668, opened 2021-11-29
Fix detaching old animated props

### LINKED ISSUE #2667: [2.3.0-beta.4] Regression introduced in createAnimatedComponent from #2600
## Description

I've noticed a regression when upgrading from 2.3.0-beta3 to 2.3.0-beta4.
We have a "count up" text animation where we use [ReText](https://github.com/wcandillon/react-native-redash/blob/master/src/ReText.tsx) from redash. Here the `text` prop, which is a native prop of the `TextInput`, is being animated.

### Expected behavior

This is how the animation looks like in 2.3.0-beta3:

https://user-images.githubusercontent.com/16821682/143778598-870051af-ce41-45c4-a066-023cac0a21e0.mov

This is how it looks in 2.3.0-beta4, you can see that it doesn't have the "count up" effect. 

https://user-images.githubusercontent.com/16821682/143778613-abcd8527-14e9-4756-addb-db96916fe736.mov

I was able to pin down this issue to the following PR: #2600 
Here, when uncommenting these lines https://github.com/software-mansion/react-native-reanimated/blob/6131585594c5f0ecf7e7832f8fb53d56210db1c8/src/createAnimatedComponent.tsx#L483-L486 everything works again. I am not too much into the reanimated code base, and just removing these lines doesn't seem like a valid fix. Maybe @tomekzaw can have another look at this?

Unfortunately, or interestingly, I wasn't able to reproduce that behaviour easily in a clean environment. If its necessary I can try again to provide a reproduction, please let me know.

The code roughly looks like this:

```jsx
function CountUp({ value }) {
  const animatedValue = useSharedValue(value);
  const text = useDerivedValue(() => `${animatedValue.value}`, [animatedValue]);

  useEffect(() => {
    animatedValue.value = withTiming(value);
  }, [value]);

  return (<ReText text={text} />)
}
```

(note: Maybe #2666 is the same issue and can be used as reproduction)

## Package versions

- React Native: 0.65.2
- React Native Reanimated: 2.3.0-beta.4
- NodeJS: 14
- Xcode: 13
- Java & Gradle: Java 8, Gradle 4, but not important as the issue is in JS land

## Affected platforms

- [x] Android
- [x] iOS
- [ ] Web -> not tested, maybe affected too

tomekzaw:
## Description

Fixes #2667.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -106,6 +106,8 @@ function isSameAnimatedStyle(style1: StyleProps, style2: StyleProps): boolean {
   return style1.viewsRef === style2.viewsRef;
 }
 
+const isSameAnimatedProps = isSameAnimatedStyle;
+
 const has = <K extends string>(
   key: K,
   x: unknown
@@ -481,7 +483,10 @@ export default function createAnimatedComponent(
       });
 
       // detach old animatedProps
-      if (prevAnimatedProps && prevAnimatedProps !== this.props.animatedProps) {
+      if (
+        prevAnimatedProps &&
+        !isSameAnimatedProps(prevAnimatedProps, this.props.animatedProps)
+      ) {
         prevAnimatedProps.viewDescriptors!.remove(viewTag as number);
       }
 


mrousavy (approved):
I noticed a Path (`d` prop) not animating with the latest release, and this PR fixes it. Nice job.

piaskowyk (approved):

Issue #2664, opened 2021-11-26
Layout transition is not working


thecoorum:
## Description
When using `LayoutAnimations` entering and exiting are working but not the layout transitions

### Media
![2021-11-26 18 09 51](https://user-images.githubusercontent.com/32096016/143607833-d23bceef-b93f-46de-b478-3b68f1dd724d.gif)

### Expected behavior
Pills are waiting till the deleted one disappear and then moving to an empty place

### Actual behavior & steps to reproduce
Pills are not waiting till the deleted one will disappear. Delay modifier is also ignored

## Snack or minimal code example
```
    <Animated.View
      entering={FadeInRight.duration(120).springify().mass(0.3)}
      exiting={FadeOut.duration(100).springify().mass(0.3)}
      layout={CurvedTransition.duration(100).delay(120)}
    >
      ...
    </Animated.View>
```

## Package versions
- React Native: 0.65.1
- React Native Reanimated: 2.3.0-beta.3 / 2.3.0-beta.4
- NodeJS: 12.13.0
- Xcode: 13.1

## Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

piaskowyk:
Hey, I can reproduce your issue :/ Could you share your code? I will attach my example, maybe it will be helpful for you.

https://user-images.githubusercontent.com/36106620/144048779-206dcab2-c6ff-4b4f-aa2d-a8c76d1be2f7.mov

```js
import Animated, {
  FadeInRight,
  FadeOut,
  CurvedTransition
} from 'react-native-reanimated';
import { View, Button } from 'react-native';
import React, { useState } from 'react';

function AnimatedStyleUpdateExample(): React.ReactElement {
  const [components, setComponents] =  useState<any>([]);

  return (
    <View
      style={{ marginTop: 50 }}>
      <View style={{display: 'flex', flexDirection: 'row'}}>
        {components}
      </View>
      <Button
        title="toggle"
        onPress={() => {
          setComponents([...components,
            <Animated.View
              key={Math.random()}
              style={{width: 50, height: 50, backgroundColor: 'black', margin: 5}}
              entering={FadeInRight.duration(120).springify().mass(0.3)}
              exiting={FadeOut.duration(100).springify().mass(0.3)}
              layout={CurvedTransition.duration(100).delay(120)}
            ></Animated.View>
          ]);
        }}
      />
      <Button
        title="remove"
        onPress={() => {
          setComponents(components.slice(1));
[...]

Issue #2424, opened 2021-09-14
[2.3.0-alpha.3] Layout animation does not work inside `FlatList`


SimpleCreations:
## Description

Participant list example when using `FlatList`:

https://user-images.githubusercontent.com/12449725/133289806-321b720f-6494-4034-a451-6b1abb29d094.mp4

Participant list example when using `ScrollView`:

https://user-images.githubusercontent.com/12449725/133290050-a4015ded-de8c-47ed-a1e2-fb4c48ea828a.mp4

### Expected behavior

When using `FlatList`, `layout={Layout.springify()}` causes the list elements to slide into their new positions with a spring animation.

### Actual behavior & steps to reproduce

When using `FlatList`, `layout={Layout.springify()}` has no effect.

## Snack or minimal code example

I've modified this file in the example app:

https://github.com/software-mansion/react-native-reanimated/blob/master/Example/src/LayoutReanimation/AnimatedList.tsx

```diff
@@ -3,9 +3,10 @@ import {
   Button,
   View,
   Text,
-  ScrollView,
   TextInput,
   StyleSheet,
+  FlatList,
+  ListRenderItem,
 } from 'react-native';
 import Animated, {
   Layout,
@@ -90,17 +91,20 @@ export default function AnimatedListExample(): React.ReactElement {
     );
   };
 
+  const renderItem: ListRenderItem<EventParticipant> = ({item}) => (
+    <Participant
+        name={item.name}
+        onRemove={() => removeParticipant(item.id)}
+    />
+  );
+
   return (
     <View style={[styles.listView]}>
-      <ScrollView style={[{ width: '100%' }]}>
-        {participantList.map((participant) => (
-          <Participant
-            key={participant.id}
-            name={participant.name}
-            onRemove={() => removeParticipant(participant.id)}
-          />
-        ))}
-      </ScrollView>
+      <FlatList
+          style={[{ width: '100%' }]}
+          data={participantList}
+          renderItem={renderItem}
+          keyExtractor={({id}) => id} />
 
       <View style={[styles.bottomRow]}>
         <View style={[styles.textInput]}>
```

## Package versions

- React Native: 0.65.1
- React Native Reanimated: 2.3.0-alpha.3
- NodeJS: 16.6.1
- Xcode: 12.5.1
- Java & Gradle: -

## Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

Szymon20000:
[...]

PR #2674, opened 2021-12-01
Layout animation for FlatList

piaskowyk:
## Description

The default implementation of `FlatList` wraps every item by custom component. This component hasn't got children's properties - in effect wrapping component hasn't the layout property with `layout` animation. It means the animation of the layout will not run because the "layout event" will be consumed by the wrapping component instead of his child.

## How to use it

```js
<Animated.FlatList // use Animated.FlatList instead of FlatList
  itemLayoutAnimation={Layout.springify()} // this animation will be apply as layout property for every list item
  data={itemsData} // as always
  renderItem={renderItem} // as always
/>
```

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/2664 https://github.com/software-mansion/react-native-reanimated/issues/2424

Inspired by: @nicholash747 's https://github.com/software-mansion/react-native-reanimated/issues/2424#issuecomment-962084108 - thanks! 🚀

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -26,6 +26,7 @@ declare module 'react-native-reanimated' {
     Text as ReactNativeText,
     Image as ReactNativeImage,
     ScrollView as ReactNativeScrollView,
+    FlatList as ReactNativeFlatList,
     NativeScrollEvent,
     NativeSyntheticEvent,
     ColorValue,
@@ -258,6 +259,7 @@ declare module 'react-native-reanimated' {
     export class Code extends Component<CodeProps> {}
     export class FlatList extends Component<AnimateProps<FlatList>> {
       itemLayoutAnimation: ILayoutAnimationBuilder;
+      getNode(): ReactNativeFlatList;
     }
 
     type Options<P> = {
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -256,6 +256,9 @@ declare module 'react-native-reanimated' {
       getNode(): ReactNativeScrollView;
     }
     export class Code extends Component<CodeProps> {}
+    export class FlatList extends Component<AnimateProps<FlatList>> {
+      itemLayoutAnimation: ILayoutAnimationBuilder;
+    }
 
     type Options<P> = {
       setNativeProps: (ref: any, props: P) => void;
--- a/src/reanimated2/component/FlatList.tsx
+++ b/src/reanimated2/component/FlatList.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import { FlatList, FlatListProps, LayoutChangeEvent } from 'react-native';
+import WrappedComponents from './WrappedComponents';
+import createAnimatedComponent from '../../createAnimatedComponent';
+import { ILayoutAnimationBuilder } from '../layoutReanimation/animationBuilder/commonTypes';
+
+const AnimatedFlatList = createAnimatedComponent(FlatList as any);
+
+const createCellRenderer = (itemLayoutAnimation?: ILayoutAnimationBuilder) => {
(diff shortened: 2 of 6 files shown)


react-native-reanimated.d.ts
```diff
@@ -256,6 +256,9 @@ declare module 'react-native-reanimated' {
       getNode(): ReactNativeScrollView;
     }
     export class Code extends Component<CodeProps> {}
+    export class FlatList extends Component<AnimateProps<FlatList>> {
```
WoLewicki, line 259: other components have also `getNode` option. Is it not needed here?

src/reanimated2/component/FlatList.tsx
```diff
@@ -0,0 +1,47 @@
...
+  itemLayoutAnimation,
+  ...restProps
+}) => {
+  const cellRenderer = React.useMemo(
+    () => createCellRenderer(itemLayoutAnimation),
+    []
+  );
+  return (
+    <AnimatedFlatList
+      {...(restProps as any)}
```
WoLewicki, line 41: Why do the types don't work correctly here?

WoLewicki (commented):
I added 2 comments, if they do not introduce any new changes, I think it is ready to be merged :tada:

src/reanimated2/component/FlatList.tsx
mrousavy: Because `FlatList` is generic. @jakub-gonet and I already had a hard time trying to get this working, I'd be surprised if there was a simple solution 😅

jakub-gonet: We had to retype FlatList basically. Checkout https://github.com/software-mansion/react-native-gesture-handler/blob/6cc5e4b8f4523d9e22b5d7aef1b15d3c9dad9cf4/src/components/GestureComponents.tsx#L75-L90 and https://github.com/software-mansion/react-native-gesture-handler/commit/9c7c9115f5baaa8025f1241c068a80f7884a648d.

src/reanimated2/component/FlatList.tsx
```diff
@@ -0,0 +1,44 @@
...
+  itemLayoutAnimation?: ILayoutAnimationBuilder;
+}
+
+type ReanimatedFlatListFC<T = any> = React.FC<ReanimatedFlatlistProps<T>>;
+
+const ReanimatedFlatlist: ReanimatedFlatListFC = ({
+  itemLayoutAnimation,
+  ...restProps
+}) => {
+  const cellRenderer = React.useMemo(
```
kmagiera, line 35: any reason to use memo here instead of just a simple method? Seems like updating layout animation won't have any effect given memo has empty array as dependencies now

kmagiera (approved):
Left one comment inline. Looks good otherwise

braincore:
@piaskowyk When will there be a new beta release that includes this? I'm having trouble updating my Podfile to use this commit pre-release.

piaskowyk:
This PR is released today with stable 2.3.0

react-native-reanimated.d.ts
```diff
@@ -256,6 +257,10 @@ declare module 'react-native-reanimated' {
...
```
Swiftwork, line 260: @piaskowyk this type definition is missing the generic type mentioned in issue [2769](https://github.com/software-mansion/react-native-reanimated/issues/2769). Should probably be defined as:

```ts
export class FlatList<T = any> extends Component<AnimateProps<FlatListProps<T>>> {
  itemLayoutAnimation: ILayoutAnimationBuilder;
  getNode(): ReactNativeFlatList;
}
```

CallumHemsley:
Is there any requirements to use this? Do children need to be `Animated.View` for example? @piaskowyk

hadnet:
Is this available in Reanimated 3-rc?

PR #2680, opened 2021-12-02
Use reflection instead of common interface

j-piasecki:
## Description

Removed common interface between Reanimated and Gesture Handler: `GestureHandlerStateManager` that was causing compilation error in some cases. Instead of it, the communication between modules has been changed to use reflection.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


j-piasecki:
Closing in favor of https://github.com/software-mansion/react-native-gesture-handler/pull/1747

PR #2681, opened 2021-12-02
Clean mounting block

piaskowyk:
## Description
//TODO

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/ios/REANodesManager.m
+++ b/ios/REANodesManager.m
@@ -262,8 +262,9 @@
     }
 
     if (_mounting) {
-      _mounting();
+      volatile void (^_mountingPtr)(void) = _mounting;
       _mounting = nil;
+      _mountingPtr();
     }
   }
   _wantRunUpdates = NO;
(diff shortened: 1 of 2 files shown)


samkline (approved):
I can confirm this fixes an easily reproducible crash in our app. Thanks for fixing!

nmeibergen:
This does seem to avoid the app from crashing, however, the animation will not function as expected. For the following case I experience flickering:

```
const TestComponent = () => {
    const scrollValue = useSharedValue(0);
        onScroll: (event) => {
            scrollValue.value = event.contentOffset.y;
        }
    })

        return {
            // Just add some margin at the top (between 0 and 100)
            marginTop: Math.min(Math.max(-scrollValue.value, 0), 100)
        }
    })

    return <>
        <Animated.View style={{ backgroundColor: '#ff00000', ...style }}>
            <Text>This is a test</Text>
        </Animated.View>
        <Animated.ScrollView
            scrollEventThrottle={1}
            onScroll={scrollHandler}
        >
            <View style={{backgroundColor: '#000000', height: 200}}>
                <Text>DRAG ME DOWN PLEASE</Text>
            </View>
        </Animated.ScrollView>
    </>
}
```

When dragging down on the scrollview (just as you would when refreshing a list), you will note it continuously snaps back to the top. 

I assume this to be related or should this be considered as a separate thread?

Issue #2714, opened 2021-12-08
window.performance.now is not a function


yusufyildirim:
## Description

Reanimated 2.3.0 is broken on the web.
<img width="418" alt="Screen Shot 2021-12-08 at 23 49 37" src="https://user-images.githubusercontent.com/22980987/145284538-3309f679-541c-430f-a047-7e6a278213bf.png">

The culprit seems like the small change made in this PR: https://github.com/software-mansion/react-native-reanimated/pull/2679

This line to be exact: https://github.com/software-mansion/react-native-reanimated/pull/2679/files#diff-78a1300914cdf4318497f5e74a0a44c4f5174ef2567339edeb0d646c1ddbfaf0R386

global._chronoNow is not available on the web, it basically sets window.performance.now to undefined. Adding an if statement should be enough. I'll create a PR

### Actual behavior & steps to reproduce
Open a web project using reanimated 2.3.0 it should be enough

## Package versions
- React Native: 0.64
- React Native Reanimated: 2.3.0

## Affected platforms

- [ ] Android
- [ ] iOS
- [x] Web

PR #2679, opened 2021-12-02
feat: Add `performance.now()` to Worklets

mrousavy:
## Description

Adds support for measuring performance/elapsed time in millisecond precision using Chrono with `performance.now()` in Worklets.

## Changes

- Inject `_chronoNow` func in global runtime object

## Test code and steps to reproduce

```ts
runOnUI(() => {
  'worklet'
  const start = performance.now()
  for (let i = 0; i < 1000000; i++) {}
  const end = performance.now()
  console.log(`Loop took ${end - start} ms!`)
})()
```

## Checklist

- [ ] Included code example that can be used to test this change
- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Added unit / integration tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

## Related Issues

* https://github.com/facebook/react-native/pull/32695

PR #2715, opened 2021-12-08
fix: web performance.now is undefined

yusufyildirim:
## Description
The web was broken in 2.3.0 due to a small change that was introduced in [a previous PR](https://github.com/software-mansion/react-native-reanimated/pull/2679).

Basically, we were setting the global.performance.now to global._chronoNow but it's not available for the web and it makes window.performance.now undefined. This PR fixes that by making that change native only.

Fixes: #2714 

# Error Screenshot
Screen Shot 2021-12-08 at 23 49 37

You don't need anything special to reproduce it. Opening a web app using reanimated 2.3.0 should be enough to see the error message

### DIFF
--- a/src/reanimated2/core.ts
+++ b/src/reanimated2/core.ts
@@ -383,9 +383,11 @@ if (!NativeReanimatedModule.useOnlyV1) {
         info: runOnJS(capturableConsole.info),
       };
       _setGlobalConsole(console);
-      global.performance = {
-        now: global._chronoNow,
-      };
+      if (!runOnJS(shouldBeUseWeb)) {
+        global.performance = {
+          now: global._chronoNow,
+        };
+      }
     })();
 }
 


src/reanimated2/core.ts
```diff
@@ -383,9 +383,11 @@ if (!NativeReanimatedModule.useOnlyV1) {
         info: runOnJS(capturableConsole.info),
       };
       _setGlobalConsole(console);
-      global.performance = {
-        now: global._chronoNow,
-      };
+      if (!runOnJS(shouldBeUseWeb)) {
```
tomekzaw, line 386:
```suggestion
      if (!runOnJS(shouldBeUseWeb)()) {
```

yusufyildirim: 💯

yusufyildirim: 💯. Also, `shouldBeUseWeb` probably is not the right function to use here. Looks like @piaskowyk assigned it to himself. Probably he'll solve this

yusufyildirim:
Closing this, @piaskowyk created another PR for that 🎉 https://github.com/software-mansion/react-native-reanimated/pull/2718

Issue #2290, opened 2021-08-17
Incompatibility with react-native 0.65.0


alexco2:
## Description

After upgrading to react-native version 0.65.0 with reanimated 2.2.0 I got the following error:
```
Execution failed for task ':app:mergeDebugAssets'.
> Could not resolve all files for configuration ':app:debugRuntimeClasspath'.
   > Failed to transform react-native-reanimated-65-hermes.aar (project :react-native-reanimated) 
to match attributes {artifactType=android-assets}.
      > Execution failed for JetifyTransform: C:\Users\alexa\Documents\Programmieren\TheRightChoice\node_modules\react-native-reanimated\android\react-native-reanimated-65-hermes.aar.
         > Transform's input file does not exist: C:\Users\alexa\Documents\Programmieren\TheRightChoice\node_modules\react-native-reanimated\android\react-native-reanimated-65-hermes.aar. (See https://issuetracker.google.com/issues/158753935)
```
I tried react-native 0.65.0 with reanimated version 2.3.0-alpha.2 and then got a new error on startup, which is the following:

```
WARN  `new NativeEventEmitter()` was called with a non-null argument without the required `addListener` method.
 WARN  `new NativeEventEmitter()` was called with a non-null argument without the required `removeListeners` method.
 WARN  `new NativeEventEmitter()` was called with a non-null argument without the required `addListener` method.
 WARN  `new NativeEventEmitter()` was called with a non-null argument without the required `removeListeners` method.
 LOG  Running "TheRightChoice" with {"rootTag":1}
 WARN  `new NativeEventEmitter()` was called with a non-null argument without the required `addListener` method.
    at HomeScreen (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.therightchoice&modulesOnly=false&runModule=true:111643:83)
    at StaticContainer (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.therightchoice&modulesOnly=false&runModule=true:170859:17)
    at EnsureSingleNavigator (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.therightchoice&modulesOnly=false&runModule=true:166638:24)
    at SceneView (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.therightchoice&modulesOnly=false&runModule=true:170743:22)
    at RCTView
    at View
[...]

Issue #2246, opened 2021-08-01
Crash with 2.3.0 animating svg's


alexco2:
## Description

After installing the alpha.2 Version (coming from version 2.2.4) the app crashes in development without error message at the start. With crashlytics and in production mode, I can get the following error message.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.swmansion.reanimated.layoutReanimation.f.j()' on a null object reference
       at com.swmansion.reanimated.layoutReanimation.g.r(:11)
       at com.facebook.react.uimanager.z0$m.execute(:14)
       at com.facebook.react.uimanager.z0$a.run(:135)
       at com.facebook.react.uimanager.z0.U(:53)
       at com.facebook.react.uimanager.z0.s()
       at com.facebook.react.uimanager.z0$j.c(:31)
       at com.facebook.react.uimanager.f.a()
       at com.facebook.react.modules.core.g$d.a(:46)
       at com.facebook.react.modules.core.a$a$a.doFrame(:2)
       at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1061)
       at android.view.Choreographer.doCallbacks(Choreographer.java:845)
       at android.view.Choreographer.doFrame(Choreographer.java:775)
       at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:1048)
       at android.os.Handler.handleCallback(Handler.java:938)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loopOnce(Looper.java:226)
       at android.os.Looper.loop(Looper.java:313)
       at android.app.ActivityThread.main(ActivityThread.java:8582)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:563)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1133)

```
### Expected behavior

No crash

### Actual behavior & steps to reproduce

```
import React, {useEffect} from 'react';
import {View, StyleSheet} from 'react-native';
import Animated, {
  useSharedValue,
  useAnimatedProps,
  withRepeat,
  withTiming,
} from 'react-native-reanimated';
import {Svg, Circle} from 'react-native-svg';

const AnimatedCircle = Animated.createAnimatedComponent(Circle);

export default function App() {
  const Animation = useSharedValue(1);

  useEffect(() => {
[...]

Issue #2009, opened 2021-05-10
interpolateColor on android is not working


Daavidaviid:
## Description

I found a case where it seems like interpolateColor is not working on android when I try to animate the fill prop of a Path (react-native-svg) with `useAnimatedProps`. Link to a repo to reproduce the issue : https://github.com/Daavidaviid/android-issue-fill-reanimated

The issue seems to arise when the derived value is not consumed directly but after a loading period.

| Screenshot 1        | Screenshot 2           |
| ------------- |:-------------:|
| <img width="343" alt="Screen Shot 2021-05-10 at 07 13 05" src="https://user-images.githubusercontent.com/32457364/117643954-59b55e80-b15f-11eb-8213-e28e11f95bb9.png">     | <img width="343" alt="Screen Shot 2021-05-10 at 07 13 20" src="https://user-images.githubusercontent.com/32457364/117643960-5b7f2200-b15f-11eb-9251-8090a919af35.png"> |

### Expected behavior

Possible to animate fill property with reanimated and scroll animation

### Actual behavior & steps to reproduce

Not working and getting the warning message :  
```
"-11545781" is not a valid color or brush 
```

## Snack or minimal code example

https://github.com/Daavidaviid/android-issue-fill-reanimated

## Package versions

- React Native: 0.64
- React Native Reanimated: 2.1.0
- NodeJS:
- Xcode:
- Java & Gradle:

## Affected platforms

- [X] Android
- [ ] iOS
- [ ] Web

Daavidaviid:
I updated the example to make it clearer

divykj:
I'm facing the same issue!

jmysliv:
Hi @Daavidaviid! I tested it on reanimated `2.3.0-alpha.2` and interpolate color was working fine with scrolling. The value that is returned by `interpolateColor` is also a valid color number on Android, so this warning is probably an issue with `react-native-svg`. If you still will be encountering some problems, let me know and I'll reopen this issue.

Issue #1909, opened 2021-04-05
animate svg fill color


Stevemoretz:
## Description
fill color doesn't get animated!

When the color changes in animation nothing happens.
### Expected behavior
The color should be animated.
### Actual behavior & steps to reproduce
```js
    const animatedProps = useAnimatedProps(()=>{
        const fill = progress.value ? colorNumber2hex(interpolateColor(progress.value,
            [0,1],
            ['rgba(255,255,255,1)','rgba(0,148,68,1)'],
            'RGB')) : 'white'
        console.log(fill);
        return {d : d,fill : fill};
    });
```
I log fill and I get colors :

```
#acdcc2
#95d2b1
#6ac192
#54b782
#2fa866
#1da059
#009444
#009444
#009444
#009444
#009444
#009444
#009444
#009444
#009444
#009444
#009444
#009444
#009444
#009444
#29a562
#3dae71
#66bf8f
#7dc9a0
#b8e1cb
#daf0e4
#f8fcfa
#ffffff
```
## Package versions

- React Native: expo sdk 40.0.1

## Affected platforms

- [x] iOS

Stevemoretz:
Fixed it with using rgb 😐

```js
return {d : d,fill : hex2RgbA(fill)};
```

Stevemoretz:
Actually let it be open maybe it's a bug?
Or it's supposed to take only rgb and not hex?

Daavidaviid:
I'm having a similar issue. `interpolateColor` returns -1 and other negative values on Android and it's causing issues and I get error messages like `"-14935012" is not a valid color or brush`

mlecoq:
I have a similar issue with `marker`

```
  const rectangleAnimatedProps = useAnimatedProps(() => {
    return {
      ...(currentItem.value?.type === "rectangle"
        ? currentItem.value.data
        : { x: 0, y: 0, width: 0, height: 0 }),
      stroke: currentItem.value?.type === "rectangle" ? "rgb(0,0,0)" : "none",
      fill: "none",
      strokeWidth: 2,
      marker:
        currentItem.value?.type === "rectangle" ? "url(#selection)" : undefined,
    };
  });
```

With a dynamic marker I want to make visible selected svg

I have also noticed that this way to do it does not work at all (no properties are updated)
```
const rectangleAnimatedPropsBis = useAnimatedProps(() => {
    return currentItem.value?.type === "rectangle"
      ? {
          ...currentItem.value.data,
[...]

PR #2723, opened 2021-12-10
Prevent RGBA decimals

furkanb:
## Description

Having decimals on [RGBA](https://www.w3.org/wiki/CSS3/Color/RGBA) values when using **`interpolateColor`** makes things problematic, especially while working with SVGs. react-native-svg throws [`console.warn(`"${color}" is not a valid color or brush`);`](https://github.com/react-native-svg/react-native-svg/blob/fa9a25d36b6f863d3148bdd5dd1ac76e91339402/src/lib/extract/extractBrush.ts#L43) whenever the RGBA value cannot be converted to int32.

The error was partially seen on #2290, #2246, #2009, #1909 

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/Colors.ts
+++ b/src/reanimated2/Colors.ts
@@ -671,9 +671,9 @@ const interpolateColorsRGB = (
   colors: InterpolateCacheRGBA
 ) => {
   'worklet';
-  const r = interpolate(value, inputRange, colors.r, Extrapolate.CLAMP);
-  const g = interpolate(value, inputRange, colors.g, Extrapolate.CLAMP);
-  const b = interpolate(value, inputRange, colors.b, Extrapolate.CLAMP);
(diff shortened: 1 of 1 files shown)


furkanb:
Thank you, I guess you're right @alexco2, after removing non-animating values from useAnimatedProps I am not able to reproduce the issue, at least on a simple gist from scratch. And I didn't have any issues on 2.2.0 either.

In my case, some Rect components which used `interpolateColors` were not rendered at all on 2.3.0 (it was fine on 2.2.0) and if I rounded the values beforehand the components were rendered correctly (the values are already passed to [another function](https://github.com/software-mansion/react-native-reanimated/blob/d0faff2f9dd3f09720e4e4735e81c5b88d533007/src/reanimated2/Colors.ts#L448) to be rounded before being returned). I'm closing this as I couldn't reproduce the issue from scratch.

PR #2725, opened 2021-12-10
Add reanimated to proguard rules.

ahmetbicer:
After upgrading to 2.3.0 i was getting instant crashes on release variants. 

```
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.swmansion.reanimated.layoutReanimation.f.j()' on a null object reference
```

Adding reanimated to proguard rules fixes the issue.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [X] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/docs/docs/fundamentals/installation.md
+++ b/docs/docs/fundamentals/installation.md
@@ -127,6 +127,7 @@ You can refer [to this diff](https://github.com/software-mansion-labs/reanimated
 If you're using Proguard, make sure to add rule preventing it from optimizing Turbomodule classes:
 
 ```
+-keep class com.swmansion.reanimated.** { *; }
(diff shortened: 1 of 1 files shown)


andresribeiro:
I confirm that after upgrading from version 2.2.4 to 2.3.1 it was crashing on Android when launching. I applied the rule in proguard and the crash stopped

nassim99:
i confirm too that this change has fixed the crashing

piaskowyk (approved):
Thanks, this is useful! 🚀

swrobel:
How does the website get updated to reflect this change? It still doesn't appear there...

heroic:
Should these be included in the package's proguard config, instead of asking the user to add these?

PR #2732, opened 2021-12-12
shared values are either not cancellable or only interruptible once

andrekovac:
Fix docs which claims that "launching animation this way can be done both from the UI and from the React-Native JS thread".

**Attention**: This solution has the side-effect that the animation now is not cancellable via `cancelAnimation` anymore. So I think this is a bug in `react-native-reanimated` (see #2733 ) which should be addressed before updating the docs. When it's fixed I can update this PR accordingly.

## Description

[Shared Value Docs](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/shared-values/) falsely claim that 

>launching animation this way can be done both from the UI and from the React-Native JS thread

This PR fixes the example code and the comment.

This PR addresses bug report #2733 

## Changes

Updated `shared-values.md ` docs

Use example from the shared values docs:

```js
import Animated, { withSpring } from 'react-native-reanimated';

function Box() {
  const offset = useSharedValue(0);

    return {
      transform: [{ translateX: offset.value * 255 }],
    };
  });

  return (
    <>
      <Animated.View style={[styles.box, animatedStyles]} />
      <Button
        onPress={() => {
          offset.value = withSpring(Math.random());
        }}
        title="Move"
      />
    </>
  );
}
```

With reanimated `2.3.0` the transition won't be interruptible.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [x] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/docs/docs/fundamentals/shared-values.md
+++ b/docs/docs/fundamentals/shared-values.md
@@ -147,11 +147,12 @@ someSharedValue.value = withTiming(50);
 ```
 
 In the above code the `offset` Shared Value instead of being set to `50` immediately, will transition from the current value to `50` using time-based animation.
-Of course, launching animation this way can be done both from the UI and from the React-Native JS thread.
+
 Below is a complete code example which is the modified version of the example from the previous section.
 Here, instead of updating `offset` value immediately, we perform an animated transition with a timing curve.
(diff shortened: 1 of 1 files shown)


andrekovac:
I made a small but crucial adaptation to the code from the docs which makes my PR wrong.

PR #2743, opened 2021-12-14
Hoist worklet function parent variable declaration to prevent referen…

### LINKED ISSUE #2742: Reanimated babel plugin transforms JS to crashing form if const/let declarations are preserved
## Description

Reanimated2 babel plugin causes JS to crash when using e.g. useAnimatedRef

If const/let declarations are not transpiled to vars, output JS is crashing due to referencing parent function's variable from withing its declaration - inside internal `_closure` property.

Let me show an example of hook/useAnimatedRef.ts compilation:
 
original:
```
export function useAnimatedRef<T extends Component>(): RefObjectFunction<T> {
  const tag = useSharedValue<number | null>(-1);
  const ref = useRef<RefObjectFunction<T>>();

  if (!ref.current) {
    const fun: RefObjectFunction<T> = <RefObjectFunction<T>>((component) => {
      'worklet';
      if (component) {
        tag.value = getTag(component);
        fun.current = component;
      }
      return tag.value;
    });

    Object.defineProperty(fun, 'current', {
      value: null,
      writable: true,
      enumerable: false,
    });
    ref.current = fun;
  }

  return ref.current;
}
```

compiled:
```
export function useAnimatedRef() {
  const tag = useSharedValue(-1);
  const ref = useRef();

  if (!ref.current) {
    const fun = function () {
      const _f = function (component) {
        if (component) {
          tag.value = getTag(component);
          fun.current = component;
        }

        return tag.value;
      };

      _f._closure = {
        tag,
        getTag,
        fun // <--------------------------- this crashes!
      };
      _f.asString = "function _f(component){const{tag,getTag,fun}=jsThis._closure;{if(component){tag.value=getTag(component);fun.current=component;}return tag.value;}}";
      _f.__workletHash = 2837072662922;
      _f.__location = "/Users/pawelwieckowski/Projects/com/reanimated2-plugin-crash-repro/node_modules/react-native-reanimated/src/reanimated2/hook/useAnimatedRef.ts (15:61)";

      return _f;
    }();

    Object.defineProperty(fun, 'current', {
      value: null,
      writable: true,
      enumerable: false
    });
    ref.current = fun;
  }

  return ref.current;
}
```

### Expected behavior
No crash, safe references in `_closure` with support of consts and lets.

Fast track:
1. Go to https://github.com/DrRefactor/reanimated2-plugin-crash-repro/commits/main/useAnimatedRef.compiled.js
1. See commits before & after fix https://github.com/DrRefactor/react-native-reanimated/commit/9212505ef426ea8a8c1ed73fe891530ac6381f8d is applied

Full repro:
1. Clone https://github.com/DrRefactor/reanimated2-plugin-crash-repro
1. `yarn`
1. `git checkout bd87ede7d2f1e344f75c8a7dc678b5ac88b2501d` (pre-fix version)
1. `yarn compile`
1. open compiled `useAnimatedRef.compiled.js`, see incorrect JS code 
1. `git checkout main`
1. `yarn compile`
1. open compiled js file

https://github.com/DrRefactor/reanimated2-plugin-crash-repro

## Package versions

- React Native: *
- NodeJS: *
- Xcode: *
- Java & Gradle: *

## Affected platforms

- [x] Android (not sure if it's reliably possible to configure babel this way, assuming - yes)
- [x] iOS (not sure if it's reliably possible to configure babel this way, assuming - yes)
- [x] Web

I will provide PR shortly, let's discuss there on whether my approach is good or not

DrRefactor:
…cing uninitialized variable which causes crash if declaration kind is const or let

## Description

Please see related #2742 issue. This PR aims to fix it.

## Changes

I aimed to hoist problematic declaration that was causing crash. Assumed that it'd be easiest to do it by changing parent variable declaration kind to `var`, considering parent variable declaration always exists (if it doesn't - reanimated adds it, see https://github.com/software-mansion/react-native-reanimated/blob/main/plugin.js#L546

Alternatively, this could be done by separating definition and declaration `let fun; fun=(...)`, but that seems to be an overkill and more or less simulates `var`'s work.

## Screenshots / GIFs

Here you can add screenshots / GIFs documenting your change.

You can add before / after section if you're changing some behavior.

### Before
image

### After
image

see issue #2742 

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -547,11 +547,34 @@ function processWorkletFunction(t, fun, fileName) {
     t.isScopable(fun.parent) || t.isExportNamedDeclaration(fun.parent);
   fun.replaceWith(
     fun.node.id && needDeclaration
-      ? t.variableDeclaration('const', [
+      ? t.variableDeclaration('var', [
           t.variableDeclarator(fun.node.id, replacement),
         ])
       : replacement
   );
+
+  // replace nearest parent declaration with var to prevent crash
+  // privateFunctionId's _closure property will include it
+  // causing reference to uninitialized const/let variable
+  //
+  // example:
+  // const fun = function () {
+  //   const _f = function (component) {};
+  //   _f._closure = {
+  //     fun
+  //   };
+  //   return _f;
+  // }();
+  //
+  // in above scenario, replacing `fun`'s variable declaration with `var` solves crash through hoisting  
+  let maybeDeclaration = fun.parentPath;
+  while (maybeDeclaration != null) {
+    if (t.isVariableDeclaration(maybeDeclaration.node)) {
+      maybeDeclaration.node.kind = 'var';
+      break;
+    }
+    maybeDeclaration = maybeDeclaration.parentPath;
+  }
 }
 
 function processWorkletObjectMethod(t, path, fileName) {


DrRefactor:
@piaskowyk could you take a look? this problem might break many RN build systems

Thanks in advance!

piaskowyk:
@DrRefactor I have tested it and looks ok, but I need to discuss these changes with my coworker to be sure, please give me a couple of days

piaskowyk:
Hey @DrRefactor 👋 The last time we made many changes in the plugin and `worklet` internals. Could you check if your problem still exists in the newest version of Reanimated 3 ?

piaskowyk:
Thanks for your commitment and PR but I think that these changes are no longer relevant to the current form of the plugin.

Issue #2760, opened 2021-12-17
global.performance.now is not a function


michaelknoch:
## Description
I run into the following type error when trying to attach a chrome debugger to the iOS simulator or physical device.
`global.performance.now is not a function`

<img width="495" alt="Screenshot 2021-12-17 at 12 48 34" src="https://user-images.githubusercontent.com/5617793/146540830-7ce62ed2-830b-43b6-9bb6-859921554412.png">

### Expected behavior
The type error should not appear and chrome debugging should be possible

### Actual behavior & steps to reproduce
So far I was not able to reproduce it in a minimal code example (still working on it), but I can fix it in my app by commenting out every import of `react-native-reanimated`

## Snack or minimal code example
tbd soon

## Package versions

- React Native: 0.66.0 (also seen on 0.66.4)
- React Native Reanimated: 2.3.0
- NodeJS: v16.13.1
- Xcode: 13.1
- Chrome, Brave, react-native debugger with latest versions

## Affected platforms

- [x] Android
- [x] iOS
- [ ] Web

enesozturk:
Same here with `v2.3.1` in Expo app. Trying to use it with Storybook RN. I have a basic stack: 

```js
<Stack.Navigator>
  <Stack.Screen name="Home" component={TabNavigator} />
  <Stack.Screen name="Storybook" component={Story} />
</Stack.Navigator>
```

When I navigate to `Storybook` page, it is ok but when press go back. I got the issue.

<img width="451" alt="Screen Shot 2021-12-17 at 21 18 50" src="https://user-images.githubusercontent.com/19428358/146590070-37f640b3-b562-442d-9a19-6e8c08386b37.png">

Packages:
```json
{
  "expo": "^44.0.0",
  "react-native-reanimated": "~2.3.1",
}
```

Should I share any other details?

arealmaas:
Is this going to be released in a version soon?

boles-tradepending:
I think this issue has regressed with this line change https://github.com/software-mansion/react-native-reanimated/commit/9c441ac8050e7d389563b6281a533666cf70cc02#diff-78a1300914cdf4318497f5e74a0a44c4f5174ef2567339edeb0d646c1ddbfaf0L379

kennethlynne:
If you're on 2.3.1 just use https://www.npmjs.com/package/patch-package

`patches/react-native-reanimated+2.3.1.patch`
```
[...]

PR #2761, opened 2021-12-17
do not mock global.performance when connected to chrome debugger

michaelknoch:
## Description

Description and motivation for this PR.
`global._chronoNow` doesn't seem to be defined when connected to chrome debugger but `global.performance` is defined already. So I think there is no need to mock in that case.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/2760

## Changes
do not mock `global.performance` when connected to chrome debugger (same behaviour as for web)

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/core.ts
+++ b/src/reanimated2/core.ts
@@ -1,7 +1,7 @@
 /* global _WORKLET _getCurrentTime _frameTimestamp _eventTimestamp, _setGlobalConsole */
 import NativeReanimatedModule from './NativeReanimated';
 import { Platform } from 'react-native';
-import { nativeShouldBeMock, shouldBeUseWeb, isWeb } from './PlatformChecker';
+import { nativeShouldBeMock, shouldBeUseWeb, isWeb, isChromeDebugger } from './PlatformChecker';
 import {
   BasicWorkletFunction,
   WorkletFunction,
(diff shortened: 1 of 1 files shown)


src/reanimated2/core.ts
```diff
@@ -371,7 +371,7 @@ if (!NativeReanimatedModule.useOnlyV1) {
       : (workletValueSetterJS as <T>(value: T) => void)
   );
 
-  if (!isWeb() && isConfigured()) {
+  if (!isWeb() && !isChromeDebugger() && isConfigured()) {
```
michaelknoch, line 374: i don't know much about the console mocking here, we could also just mock global.performance if really needed:

```js
if (!global.performance && global._chronoNow) {
        global.performance = {
            now: global._chronoNow,
        }
};
```

michaelknoch: even without chrome debugger `global.performance` and `global._chronoNow` seem both to be undefined. should we remove the mocke in general? I can't see any purpose for it

michaelknoch:
would be great to get your opinion on this @piaskowyk, because as far as I can see you recently worked in that area

piaskowyk:
Hey @michaelknoch 👋
You are right, I am glad that you fix it! Thanks!

piaskowyk (approved):
🚀

PR #2762, opened 2021-12-17
Remove rn-screens from deps

ku8ar:
## Description

I noticed that in our project yarn is downloading react-native-screens in two different versions. The culprit turned out to be this library

## Changes

I removed react-native-screens from package.json.

According to a quick diff, Android RNScreens was used here:
https://github.com/software-mansion/react-native-reanimated/pull/2340/files
And then removed here:
https://github.com/software-mansion/react-native-reanimated/commit/2938606b58568139e0e3af6890c824fc129e6a9d

In iOS rn-screens is used here:
https://github.com/software-mansion/react-native-reanimated/blob/df873f0090acb14bbd993577a4bfb096f5ff8a49/ios/LayoutReanimation/REAUIManager.mm
But thanks to @mrousavy here is also an optional import:
https://github.com/software-mansion/react-native-reanimated/pull/2377

So currently rn-screens is only used in Example, which has its own package.json

rn-screens library takes up over 8MB, so node_modules will be lighter with this change.

## Screenshots / GIFs

Zrzut ekranu 2021-12-17 o 19 46 54

https://giant.gfycat.com/RawScaryCowbird.webm

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,6 @@
     "invariant": "^2.2.4",
     "lodash.isequal": "^4.5.0",
     "mockdate": "^3.0.2",
-    "react-native-screens": "^3.4.0",
     "string-hash-64": "^1.0.3"
   },
   "peerDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -9045,13 +9045,6 @@ react-native-gesture-handler@^1.6.1:
     invariant "^2.2.4"
     prop-types "^15.7.2"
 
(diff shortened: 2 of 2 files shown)


piaskowyk:
Thanks for your PR! We prepared a new PR based on your idea - https://github.com/software-mansion/react-native-reanimated/pull/3149

PR #2771, opened 2021-12-20
fix: Only inject `global.performance.now` if not already defined

mrousavy:
Alternative solution to #2761, related: #2760


mrousavy:
cc @michaelknoch

mrousavy:
sorry for not dropping a detailed description - but basically with this approach the outer if isn't too confusing, and it also provides the possibility of having `global.performance` natively injected in the future (assuming that's where RN is heading, with JSI) cc @piaskowyk

piaskowyk (approved):
Reasonable changes 🚀

nandorojo:
This fixed remote debugging for me, thanks!

daxaxelrod:
Holy macaroni salad thanks a ton. Redux for flipper still needs work and it's awesome to get react-native-debugger back

Issue #2752, opened 2021-12-15
[2.3] Toggling `withTiming` on/off doesn't animate


nandorojo:
## Description

This code never animates:

```ts
const style = useAnimatedStyle(() => ({
  width: isActive.value ? 100 : withTiming(300)
}))
```

Instead, it always snaps directly between 100 and 300, without an animation.

### Expected behavior

When active, this should snap to `100`. When inactive, it should do a timing animation to `300`. This worked on 2.2.

### Actual behavior & steps to reproduce

The animation snaps to 300 when inactive. It doesn't animate.

This solves it:

```ts
const style = useAnimatedStyle(() => ({
  width: withTiming(isActive.value ? 100 : 300, { duration: isActive.value ? 0 : 250 })
}))
```

But it shouldn't be necessary.

## Snack or minimal code example

<details>
<summary>
View Code
</summary>

```tsx
import React from 'react'
import { Pressable, StyleSheet } from 'react-native'
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withTiming,
} from 'react-native-reanimated'

export default function HelloWorld() {
  const pressed = useSharedValue(false)
  const style = useAnimatedStyle(() => {
    return {
      // this never animates
      width: pressed.value ? 100 : withTiming(300),
      // this does animate
      // width: withTiming(pressed.value ? 100 : 300, { duration: pressed.value ? 0 : 250 })
    }
  })

  return (
    <Pressable
      onPressIn={() => (pressed.value = true)}
      style={styles.container}
      onPressOut={() => (pressed.value = false)}
    >
      <Animated.View style={[styles.shape, style]} />
    </Pressable>
  )
}

const styles = StyleSheet.create({
  shape: {
    justifyContent: 'center',
    height: 250,
    width: 250,
    borderRadius: 25,
    marginRight: 10,
    backgroundColor: 'black',
  },
  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    flexDirection: 'row',
    backgroundColor: '#9c1aff',
  },
})
```

</details>

## Video

https://www.loom.com/share/42aa2a930b0f440692cc6d1e2d86375b

## Package versions

- React Native: 0.63.4
- React Native Reanimated: 2.3.2
- NodeJS:
- Xcode:
- Java & Gradle:

## Affected platforms

- [ ] Android
- [x] iOS
- [x] Web

piaskowyk:
[...]

PR #2776, opened 2021-12-22
Compute style changes before update

piaskowyk:
## Description

When I was doing performance optimizations, I removed function `styleDiff`, this function computes differences between old style and new style and returns only differences. In effect, we updated only changed style properties. When I removed the call of styleDiff (because was expensive) there was possible to call `layout()` too many times, even if someone doesn't change the layout property (layout is also expensive) but there was another problem because we need an update synchronous and asynchronous updates in the same `updateProps` call. So we decided to restore `styleDiff` but with faster implementation. We get rid of comparison by `JSON.stringify()`, the shall computation should be enough in this way.

## Benchmark
<details>
<summary>code</summary>

```js
const objA = {
  a: 1,
  b: 2,
  c: {},
  d: [],
  e: { a: 1 },
  f: [1, 2, 3],
  g: [{a: 1}, {a: 2}, {a: 2}],
  h: 'abc',
  i: 'abc1',
  a1: 1,
  a2: 1,
  a3: 1,
  a4: 1,
}
const objB = {
  a: 1,
  b: 22,
  c: {},
  d: [],
  e: { a: 11 },
  f: [11, 22, 33],
  g: [{a: 11}, {a: 2}, {a: 2}],
  h: 'abc',
  i: 'abc2',
}
const iteration = 99999;

for(let i = 0; i < 100; i++) {
  let a = performance.now();
}

const t1_start = performance.now();
for(let i = 0; i < iteration; i++) {
  styleDiff_old(objA, objB);
}
const t1_end = performance.now();

const t2_start = performance.now();
for(let i = 0; i < iteration; i++) {
  styleDiff(objA, objB);
}
const t2_end = performance.now();

console.log(
  'old', Math.round(t1_end - t1_start), 'ms',
  '|',
  'new', Math.round(t2_end - t2_start), 'ms'
)
```
</details>

Results:
```
old 919 ms | new 282 ms
old 870 ms | new 283 ms
old 901 ms | new 273 ms
old 878 ms | new 306 ms
old 868 ms | new 274 ms
old 860 ms | new 266 ms
old 907 ms | new 309 ms
old 912 ms | new 334 ms
old 922 ms | new 291 ms
old 867 ms | new 261 ms
old 880 ms | new 261 ms
```

In addition, this
```js
state.last = Object.assign({}, oldValues, newValues);
```
fixes https://github.com/software-mansion/react-native-reanimated/issues/2752

### DIFF
--- a/src/reanimated2/hook/useAnimatedStyle.ts
+++ b/src/reanimated2/hook/useAnimatedStyle.ts
@@ -451,10 +451,10 @@ export function useAnimatedStyle<T extends AnimatedStyle>(
 
   useEffect(() => {
     let fun;
-    let upadterFn = updater;
+    let updaterFn = updater;
     let optimalization = updater.__optimalization;
     if (adapters) {
-      upadterFn = () => {
+      updaterFn = () => {
         'worklet';
         const newValues = updater();
         adaptersArray.forEach((adapter) => {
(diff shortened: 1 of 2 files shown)


src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -463,9 +468,21 @@ export function useAnimatedStyle<T extends AnimatedStyle>(
...
           'worklet';
-          const style = upadterFn();
-          parseColors(style);
-          return style;
+          const newValues = upadterFn();
+          const oldValues = remoteState.last;
+          const diff = styleDiff(oldValues, newValues);
+          remoteState.last = Object.assign({}, oldValues, newValues);
+          parseColors(diff);
+          return diff as T;
```
tomekzaw, line 476: Perhaps we could call `styleDiff<T>(...)` instead

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -463,9 +468,21 @@ export function useAnimatedStyle<T extends AnimatedStyle>(
       if (hasColorProps(upadterFn())) {
         upadterFn = () => {
```
tomekzaw, line 469: Let's rename it to `updaterFn`, but perhaps in a separate PR.

tomekzaw (changes requested):

src/reanimated2/hook/utils.ts
```diff
@@ -201,12 +201,12 @@ export function isAnimated(prop: NestedObjectValues<AnimationObject>): boolean {
...
 
-export function styleDiff(
+export function styleDiff<T extends AnimatedStyle>(
   oldStyle: AnimatedStyle,
   newStyle: AnimatedStyle
-): AnimatedStyle {
+): T {
   'worklet';
-  const diff: AnimatedStyle = {};
+  const diff: any = {};
```
tomekzaw, line 209: Could be something like:
```suggestion
  const diff: Partial<T> = {};
```

piaskowyk: Fixed with [8da3ad0](https://github.com/software-mansion/react-native-reanimated/pull/2776/commits/8da3ad0094bf0c86ce697f92993462278a90c163)

tomekzaw (approved):
🚀

Kerumen:
@piaskowyk Do you plan to release a new version with this patch included? Thanks!

PR #2799, opened 2021-12-31
feat: added withInitialValues modifier

gorhom:
## Description

This pr introduce a new layout animation modifier `withInitialValues` to sets initial values for pre-defined animations.

This is needed for third-party libraries to override animation props and add initial values - like `originX` & `originY`.

## Changes

- feat: added `withInitialValues` modifier to `ComplexAnimationBuilder`.
- docs: added the mention for the `withInitialValues` modifier.

Tested the example `DefaultAnimations.tsx >> FadeIn`, where i added a background color as an initial value, and it works

```tsx
FadeIn.withInitialValues({ backgroundColor: 'red' })
``` 

## Checklist

- [x] Updated TS types
- [x] Added TS types tests
- [x] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/docs/docs/api/LayoutAnimations/EntryAnimations.md
+++ b/docs/docs/api/LayoutAnimations/EntryAnimations.md
@@ -55,6 +55,7 @@ Simple animation based on changing of opacity.
 * `restDisplacementThreshold` default: 0.001
 * `restSpeedThreshold` default: 0.001
 * `withCallback` callback that will fire after the entry animation ends
+* `withInitialValues` sets the initial values
 * `randomDelay` randomize delay of the animation between 0 and provided delay ( or 1000ms if delay not provided )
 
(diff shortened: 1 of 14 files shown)


piaskowyk (commented):
Nice improvements! 
I think that `...initialValues` should be used at the end of an object instead of at the beginning because custom `initialValues` should override default values. What do you think?
```js
initialValues: {
  // other properties
  ...initialValues,
},
```

gorhom:
@piaskowyk you are right , i am going to update it this week.

piaskowyk:
Thanks! I will merge it with pleasure ❤️

piaskowyk (approved):
Good job 🚀

piaskowyk (approved):

Issue #2758, opened 2021-12-17
Layout animation leave orphaned component in Expo


dpnick:
## Description

I found a strange behavior while trying to use react-native-reanimated layout animation with the latest Expo SDK (44). Component with layout animation seems to leave orphaned component at re-render. (related [expo issue](https://github.com/expo/expo/issues/15600)) 

I have a simple conditional rendering, when the component mount it should display a spinner for 3000ms and then a list of (transparent for demo purpose) items. However, on the first render after the 3000ms the spinner AND the list are both visible at the same time.

If I reload the project everything works as expect. But as soon as I completely restart the app, this behavior happen.

I was able to reproduce this behavior in a fresh new app, I will include the steps bellow.

https://user-images.githubusercontent.com/23317360/146521672-a5c679cf-eff2-4ba8-baee-a99346778f37.mov

### Expected behavior

The layout animation works as expected (and are, actually, awesome). But it seems to introduce a bug with conditional component rendering. In my case, after 3000ms, the spinner is supposed to be hidden and replaced by the list. 

### Actual behavior & steps to reproduce

At component mount a spinner is supposed to be visible during 3000ms. After this time, the spinner should be replaced by a scrollview.

Check my minimal code bellow.

## Snack or minimal code example

// update to expo 44 (currently beta) to be able to use `react-native-reanimated` > 2
```
$ EXPO_BETA=1 expo init --npm
$ # Choose "blank (TypeScript)"
$ expo install react-native-reanimated react-native-screens
```

babel.config.js
```
module.exports = function (api) {
  api.cache(true);
  return {
    presets: ['babel-preset-expo'],
    plugins: ['react-native-reanimated/plugin'],
  };
};
```

App.tsx
```
import { useEffect, useState } from 'react';
import { ActivityIndicator, StyleSheet, Text, View } from 'react-native';
import Animated, {
  FadeInUp,
  FadeOutDown,
  Layout,
} from 'react-native-reanimated';

const AnimatedView = Animated.createAnimatedComponent(View);

export default function App() {
  const [isLoading, setIsLoading] = useState<boolean>(true);

  useEffect(() => {
    const timeout = setTimeout(() => setIsLoading(false), 3000);
    return () => {
      clearTimeout(timeout);
    };
  }, []);

  return (
    <View style={styles.container}>
      {isLoading ? (
        <ActivityIndicator size='large' color='red' />
      ) : (
        <AnimatedView
          entering={FadeInUp}
[...]

PR #2802, opened 2022-01-04
Changed default ViewState

piaskowyk:
## Description

The problem occurs since #2651
Why?
Layout animations are disabled by default until someone call layout animation. Then layout animation is enabled. The problem is when someone uses layout animation in components that don't appear in the first render. It breaks the previous assumption of layout animation - "every component is sent to `onViewCreate` function". In effect, we tried to remove the view that wasn't registered in `_viewForTag` register.

There was also another problem because uses this:
```c++
ViewState state = [_states[tag] intValue];
```
without checking if `_states[tag]` is `nil`. Then `nil` was converted to `0` and then to the first value from `ViewState` (`Appearing`), but this is wrong because the default state should be `Inactive` because not every component has animation.

I added similar changes for Android also.

## Reproduction
```js
import React from 'react';
import { ActivityIndicator, StyleSheet, Text, View, findNodeHandle } from 'react-native';
import Animated, { FadeOutDown } from 'react-native-reanimated';

  const [isLoading, setIsLoading] = useState<boolean>(true);
  useEffect(() => {
    const timeout = setTimeout(() => setIsLoading(false), 500);
    return () => {
      clearTimeout(timeout);
    };
  }, []);

  return (
      { isLoading ? 
          <ActivityIndicator size='large' color='red' />
        : 
          <Animated.View exiting={FadeOutDown} style={styles.scrollContainer}>
            {[...Array(10).keys()].map((value) => (
              <View style={styles.row} key={value}>
                <Text>row n°{value}</Text>
              </View>
            ))}
          </Animated.View>
      }
    </View>
  );
}

  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center',
  },
  scrollContainer: {
    flex: 1,
    width: '100%',
  },
  row: {
    flexDirection: 'row',
    backgroundColor: 'transparent',
    height: 50,
  },
});
```

Fixes https://github.com/software-mansion/react-native-reanimated/issues/2758

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
@@ -61,10 +61,10 @@ public class AnimationsManager implements ViewHierarchyObserver {
   }
 
   public enum ViewState {
+    Inactive,
     Appearing,
     Disappearing,
     Layout,
-    Inactive,
     ToRemove;
   }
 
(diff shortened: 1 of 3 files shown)


piaskowyk:
I need to check it on Android also.

tomekzaw (approved):
Looks good to me!

alexnaiman:
When will this be published? 😄

skeie:
Any idea when this will be released? :)

Issue #2633, opened 2021-11-16
Doc:Installing versions outadated


tomByrer:
https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation#installing-the-package

```bash
yarn add react-native-reanimated@next
```

I think the `@next` could be dropped, since v2 is the new standard version?

---

> * select latest build and download react-native-reanimated-2.0.0-alpha.tgz artifact
> * run tar zxvf react-native-reanimated-2.0.0-alpha.tgz.zip to unpack zip (or unpack it manually)
> * run yarn add file:react-native-reanimated-2.0.0-*.tgz to install the package

This is quoted from the [v2.3.0-b1 doc](https://github.com/software-mansion/react-native-reanimated/blob/master/docs/versioned_docs/version-2.3.0-beta.1/fundamentals/installation.md), but earlier versions have same old `2.0.0-alpha` versions.

Manual changes are straightforward, but easily forgoton.  Maybe either just swap "2.0.0-alpha" for "[VERSION]" to let people figure out what to type in, make a MDX component, or some grep-type automation would be useful....

PR #2818, opened 2022-01-10
Removed @next

piaskowyk:
## Description

After the stable release, we no longer need to use `@next` during the installation

Fixes https://github.com/software-mansion/react-native-reanimated/issues/2633


tomekzaw:
Here's one left:
https://github.com/software-mansion/react-native-reanimated/blob/83e8e4f0236579082b857ea1a67bfa1ea5ffe1e1/docs/versioned_docs/version-2.2.0/installation.md?plain=1#L39

tomekzaw (approved):

PR #2842, opened 2022-01-14
Disable sample profiler for UI hermes JS runtime

### LINKED ISSUE #2143: Crash: Prevents using React Native Sampling Profiler on Hermes/Android
## Description

Using the Sampling Profiler with Hermes on Android makes the app crash as soon as the profiling is stopped.

There is a comment by a React Native maintainer in https://github.com/facebook/hermes/issues/491#issuecomment-816121451 which mentions this is caused by reanimated's threading.

> Thanks for reporting this. It seems like reanimated works by scheduling Hermes runtimes on multiple threads. I'm not familiar with how reanimated is set up, but our sampling profiler does not currently automatically support moving a given runtime between threads, so it is likely that the sampling profiler crash is a result of that.

> We would need more information about how reanimated handles the runtime to get a sense for what the problem might be with the debugger.

I'm trying to debug some Android performance issues and Reanimated seems to be preventing me from it.

It seems that the React Native team may be willing to help with this, if someone could answer the query above.

### Expected behavior

Shouldn't crash.

Crashes

I think it happens by default as soon as reanimated is used. 

On Android: Open the developer menu and enable sampling profiler, after a few seconds, disable it.

## Package versions

- React Native: 0.64

## Affected platforms

- [x] Android
- [ ] iOS
- [ ] Web

janicduplessis:
## Description

The hermes profiler expects the runtime to be running on the same thread as it was created on. Since the UI JS runtime is created from the JS thread it seems to be the cause of the issue. As a workaround it is possible to disable the profiler for the UI runtime.

Fixes #2143

## Changes

Disable profiler for Hermes JS runtime on Android.

## Screenshots / GIFs

N/A

On android open the dev menu and select "Enable Sampling Profiler". This will cause the app to crash before this fix, after it will properly record a sample trace.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/cpp/NativeProxy.cpp
+++ b/android/src/main/cpp/NativeProxy.cpp
@@ -122,8 +122,10 @@ void NativeProxy::installJSIBindings() {
   };
 
 #if FOR_HERMES
+  auto config = ::hermes::vm::RuntimeConfig::Builder()
+          .withEnableSampleProfiling(false);
   std::shared_ptr<jsi::Runtime> animatedRuntime =
-      facebook::hermes::makeHermesRuntime();
+      facebook::hermes::makeHermesRuntime(config);
 #else
   std::shared_ptr<jsi::Runtime> animatedRuntime =
       facebook::jsc::makeJSCRuntime();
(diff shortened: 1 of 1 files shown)


henrymoulton:
@piaskowyk anything needed to get this merged in/reviewed? Seems this issue goes back quite a while https://github.com/facebook/hermes/issues/491

piaskowyk:
@janicduplessis Thanks!

piaskowyk (approved):

PR #2853, opened 2022-01-18
Fix "Illegal type provided" crash

### LINKED ISSUE #2722: JNI DETECTED ERROR IN APPLICATION: JNI GetObjectRefType called with pending exception java.lang.RuntimeException: java.lang.AssertionError: Illegal type provided
## Description

The worklets in Reanimated v2 seems to be really unstable in Android. Every change in the worklet code crashes the app and the app needs to be restarted like 5 times (crashes 5 times). And if you have an error in the worklet (e.g. calling any standard javascript function like parseInt, Math.round etc) it crashes in the native code with always the same error. 

If I move the code from onEnd worklet to a normal function and call it with runOnJS it either does nothing or the app crashes. Is runOnJS even supposed to work?

```
 java_vm_ext.cc:578] JNI DETECTED ERROR IN APPLICATION: JNI GetObjectRefType called with pending exception java.lang.RuntimeException: java.lang.AssertionError: Illegal type provided
 java_vm_ext.cc:578]   at void com.swmansion.reanimated.AndroidErrorHandler.raise(java.lang.String) (AndroidErrorHandler.java:6)
 java_vm_ext.cc:578]   at void com.swmansion.reanimated.Scheduler.triggerUI() (Scheduler.java:-2)
 java_vm_ext.cc:578]   at void com.swmansion.reanimated.Scheduler$1.run() (Scheduler.java:24)
```

### Expected behavior

Crashes would be captured and a stack trace printed to the console. Or a proper error in the native logs. And when restarting the app after fixing the code (or not changing it at all when no bug in the code) the app would work in immediately not after 5 to 10 restarts (some caching issue?). 

Unhandled crashes and no proper recovery.

I'm using the pan gesture handler worklets with Samsung Galaxy A51 5G (SM-A516B).

## Package versions

- React Native: 0.66.3
- React Native Gesture Handler: 2.1.0
- NodeJS: 14.18.1
- Java & Gradle: OpenJDK 64-Bit Server VM (build 14.0.2+12-46, mixed mode, sharing) & 7.3

## Affected platforms

- [X] Android
- [ ] iOS
- [ ] Web

janicduplessis:
## Description

Sometimes RN passes ReadableMap/Array instead of WritableMap/Array when updateProps is called. Currently adding a ReadableMap/Array to a WritableMap/Array is not supported (I opened a PR to allow it since there's no reason it can't be allowed https://github.com/facebook/react-native/pull/32910). In the meantime we have to check if it's not Writable and copy before adding it.

Fixes #2722

## Screenshots / GIFs

N/A

I'm not sure exactly what the minimum repro is, I could 100% repro the crash in an app, but wasn't able to isolate. The app makes heavy usage of a bottom sheet lib that uses reanimated. I was able to repro a few times in the example app with the repro in the linked issue, but not consistently.

In the app where I could repro consistently I verified that it no longer crashes after this change.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -11,7 +11,9 @@ import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.bridge.ReadableType;
 import com.facebook.react.bridge.UiThreadUtil;
+import com.facebook.react.bridge.WritableArray;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.modules.core.DeviceEventManagerModule;
 import com.facebook.react.modules.core.ReactChoreographer;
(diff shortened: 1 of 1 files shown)


justblender:
This PR seems to be working for me as well on RN 0.67.0 and does indeed fix #2722 🚀

@piaskowyk, could you take a look at this?

Omelyan:
@janicduplessis Here's a minimum repro: #2905

piaskowyk:
To resolve it I prepared PR to `react-native` - https://github.com/facebook/react-native/pull/32796 but it will be available until `react-native@0.68.0`

piaskowyk (approved):
🚀

joy-betterhalf:
@piaskowyk We are facing many ANR which could be fixed by this change so any idea on a next release cycle?

PR #2868, opened 2022-01-22
Added useAnimatedSensor()

piaskowyk:
## Description

Added API to create an animation based on sensor device data in a very simple way.

### Example code

```js
function AnimatedStyleUpdateExample() {

  const animatedSensor = useAnimatedSensor(SensorType.ROTATION, { interval: 10 }); // <- initialization
    const yaw = Math.abs(animatedSensor.sensor.value.yaw);
    const pitch = Math.abs(animatedSensor.sensor.value.pitch);
    return {
      height: withTiming(yaw * 200 + 20, { duration: 100 }), // <- usage
      width: withTiming(pitch * 200 + 20, { duration: 100 }), // <- usage
    };
  });

  return (
    <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
      <Animated.View
        style={[
          style,
        ]}
      />
    </View>
  );
}
```

https://user-images.githubusercontent.com/36106620/158634922-eaad656e-c837-44d5-8d51-8e7fa27c5a16.mp4


mrousavy:
This is a really cool feature, but do you really think this should be part of the Reanimated repo? I would imagine this could be an optional additional dependency a user could install, and it would interact with the Reanimated API under the hood - this is the same concept as the "Lean Core" efforts in React Native Core. (my personal preference of course)

tomekzaw:
> This is a really cool feature, but do you really think this should be part of the Reanimated repo? I would imagine this could be an optional additional dependency a user could install, and it would interact with the Reanimated API under the hood - this is the same concept as the "Lean Core" efforts in React Native Core. (my personal preference of course)

Hello and thanks for your valuable suggestion. Actually we've faced the same dillema internally and honestly it's a non-trivial decision. And obviously, this usually is a matter of personal preference. Here's some context:

Ideally, we imagine this feature to be a separate library in order to keep things modular and organized. It's a new feature that makes a good use of Reanimated internals such as shared values and mappers but certainly belongs to some higher layer of abstraction. Also, we haven't measured the impact of this feature on the overall size of the library and the JS bundle yet, but it appears to be rather a small increase.

However, having two libraries to maintain always causes more difficulties than having just one, not only for developers, but also for the users. Moving this feature to another dependency would strongly affect the whole process of development, testing, version compatibility, reporting, bug fixes etc. And we haven't even mentioned installation and initialization yet. It simply creates a series of problems we don't necessarily want to get into at the moment 😄 

Taking into account all the engineering aspects behind it, we'll probably keep it within react-native-reanimated as for now. In fact, this is the natural way of how projects evolve. Once they get big enough, then it's a good time do to some modularization, similarly as in React Native core. Ideally, we'd like to deliver a similar opt-in mechanism for other features (e.g. layout animations) but we'd like to address this issue holistically once we come up with a proper solution for that 😄

android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensorContainer.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public void unregisterSensor(int sensorId) {
+        ReanimatedSensor sensor = sensors.get(sensorId);
+        if (sensor != null) {
+            sensor.cancel();
+            sensors.remove(sensorId);
+        }
+        else {
+            Log.e("Reanimated", "You try to reject non existing sensor listener.");
```
tomekzaw, line 37:
```suggestion
            Log.e("Reanimated", "Tried to unregister nonexistent sensor");
```

android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensorContainer.java
```diff
@@ -0,0 +1,40 @@
...
+    private final HashMap<Integer, ReanimatedSensor> sensors = new HashMap<>();
+
+    public ReanimatedSensorContainer(WeakReference<ReactApplicationContext> reactContext) {
+        this.reactContext = reactContext;
+    }
+
+    public int registerSensor(ReanimatedSensorType sensorType, int interval, NativeProxy.SensorSetter setter) {
+        ReanimatedSensor sensor = new ReanimatedSensor(reactContext, sensorType, interval, setter);
+        int sensorId = -1;
+        if(sensor.initialize()) {
```
tomekzaw, line 22:
```suggestion
        if (sensor.initialize()) {
```

android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensorType.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public static ReanimatedSensorType getInstanceById(int typeId) {
+        switch (typeId) {
+            case 1: return ReanimatedSensorType.ACCELEROMETER;
+            case 2: return ReanimatedSensorType.GYROSCOPE;
+            case 3: return ReanimatedSensorType.GRAVITY;
+            case 4: return ReanimatedSensorType.MAGNETIC_FIELD;
+            case 5: return ReanimatedSensorType.ROTATION_VECTOR;
+        }
+        return ReanimatedSensorType.ACCELEROMETER;
```
tomekzaw, line 28: Maybe it would be better to throw exception here?

ios/sensor/ReanimatedSensorContainer.m
```diff
@@ -0,0 +1,43 @@
...
+
+@implementation ReanimatedSensorContainer
+
+- (instancetype)init
+{
+  self = [super init];
+  _sensors = [[NSMutableDictionary alloc] init];
+  if (_nextSensorId == nil) {
+    _nextSensorId = @0;
+  }
```
tomekzaw, line 16: Why can't we just assign `_nextSensorId = @(0)` like in Java?

src/reanimated2/hook/useAnimatedSensor.ts
```diff
@@ -0,0 +1,91 @@
+import { useEffect, useRef } from 'react';
+import { makeMutable } from '../core';
+import NativeReanimated from '../NativeReanimated';
+import { SensorValue3D, SensorValueRotation } from '../commonTypes';
+
+export enum SensorType {
+  ACCELEROMETER = 1,
+  GYROSCOPE = 2,
+  GRAVITY = 3,
+  MAGNETIC = 4,
```
tomekzaw, line 10:
```suggestion
  MAGNETIC_FIELD = 4,
```
BTW why do we have to define this enum also in `react-native-reanimated.d.ts`?

tomekzaw (changes requested):
I've left some comments in the code. Really looking forward to trying out this feature!

src/reanimated2/hook/useAnimatedSensor.ts
piaskowyk: I didn't find a way to import enum to `d.ts` :/

Example/ios/ReanimatedExample/Info.plist
```diff
@@ -51,7 +53,5 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
-	<key>CADisableMinimumFrameDurationOnPhone</key>
-	<true/>
```
tomekzaw, line 55: Can we bring this back? I think this is required for 120 fps.

Common/cpp/headers/AnimatedSensor/AnimatedSensorModule.h
```diff
@@ -0,0 +1,41 @@
...
+enum SensorType {
+  ACCELEROMETER = 1,
+  GYROSCOPE = 2,
+  GRAVITY = 3,
+  MAGNETIC_FIELD = 4,
+  ROTATION_VECTOR = 5,
+};
+
+class AnimatedSensorModule {
+  std::unordered_set<int> sensorsIds;
```
tomekzaw, line 22:
```suggestion
  std::unordered_set<int> sensorsIds_;
```

tomekzaw (approved):
:shipit: 👏

nandorojo:
To follow up on @mrousavy's suggestion, I've solved this in moti by having separate exports like `moti/interactions` and `moti/skeleton`, similar to `react-native-screens/native-stack`. 

For the case of Reanimated, you could do `react-native-reanimated/sensor`, which would essentially tree shake it and its dependencies from consumers.

PR #2882, opened 2022-01-25
Debug native code in Cmake

piaskowyk:
## Description

Added configuration to make reanimated's c++ part debuggable by the debugger in Android Studio.

To enable debugging you need set `NATIVE_DEBUG_ON=TRUE` in `android/local.properties`


tomekzaw (approved):
Perhaps there is a CMake-specific flag to enable gcc's flag `-g`, but the current solution seems to be good enough.

Issue #2757, opened 2021-12-17
[iOS release] cause random crash with react-native-video


Deodes:
When use react-native-reanimated with react-native-video at launch, app can crash with:

`libc++abi: terminating with uncaught exception of type folly::json::print_error: folly::toJson: JSON object value was an INF`

### Expected behavior

App not randomly crash

### Actual behavior & steps to reproduce

Install react-native-reanimated(v2 or more) and react-native-video and call: `this.videoPlayerRef.seek(...)`

## Package versions

- React Native: 0.65.1
- React Native Reanimated: 2.2.2, 2.3.1 (maybe more)
- NodeJS: 17.2.0
- Xcode: 13.1

## Affected platforms

- [X] iOS

piaskowyk:
Hey, could you provide a repo to reproduce your issue?

Deodes:
@piaskowyk it is quite difficult to reproduce because it is a large and commercial project. I will try to create a new one with the same versions of dependencies and share with you as soon as possible

Deodes:
@piaskowyk after update react-native to 0.66.4 and react-native-reanimated to 3.2.1 the error has become more informative: 

`libc++abi: terminating with uncaught exception of type folly::json::print_error: folly::toJson: JSON object value was an INF when serializing value at "playableDuration"`

jakub-gonet:
Does it crash every time you call `videoPlayerRef.seek` or only when you use it inside the worklets? We need a repro and more information to solve that issue, unfortunately.

tomekzaw:
Duplicate of #1776, working on a fix right now.

Issue #1776, opened 2021-03-01
[iOS] Causes folly::toJson Exception when used in conjunction with react-native-ivs


kyle-ssg:
## Description

I'm using react-native-aws-ivs-player-view to play video, for some reason as soon as I install this library it causes a knock on effect where the app crashes due to folly trying to parse some json that has NaN.

If I uninstall react-native-reanimated from the project, the app runs fine without this crash. 

I fixed the issue causing this for the particular library [here](https://github.com/perrystreetsoftware/react-native-aws-ivs-player-view/pull/8) . I do appreciate that this is a knock-on effect due perhaps bad Native iOS code, however some information as to why this package could throw up these crashes would be good if it isn't fixable.

```
libc++abi.dylib: terminating with uncaught exception of type folly::json::parse_error: folly::toJson: JSON object value was a NaN or INF
```

### Screenshots
![image](https://user-images.githubusercontent.com/8608314/109545232-8181a980-7ac0-11eb-9e17-83f05c3d06dc.png)

### Expected behavior
react-native-reanimated does not cause knock on folly exceptions for other libraries, or there is explanation why this could happen.
### Actual behavior
Installing react-native-reanimated caused the mentioned exception

## Snack or minimal code example

Install react-native-reanimated alongside https://github.com/perrystreetsoftware/react-native-aws-ivs-player-view/tree/main/example

## Package versions

    "react": "16.13.1",
    "react-native": "0.63.4",
    "react-native-reanimated": "2.0.0-rc.3"

mrbrentkelly:
I'm getting the exact same issue in a custom built Video Player native module.

```
libc++abi.dylib: terminating with uncaught exception of type folly::json::parse_error: folly::toJson: JSON object value was a NaN or INF
terminating with uncaught exception of type folly::json::parse_error: folly::toJson: JSON object value was a NaN or INF
```

Deodes:
The same thing with [react-native-video](https://github.com/react-native-video/react-native-video) and reanimated..

tomekzaw:
Hello! Can you please verify if this is still an issue on Reanimated 2.3.1?

tomekzaw:
Here's the only occurrence of `folly::toJson` in our codebase:

https://github.com/software-mansion/react-native-reanimated/blob/fc837335e017e53473123d4aaedf3502cc0d901c/ios/native/NativeProxy.mm#L249

[...]

PR #2896, opened 2022-01-28
Fix crash when handling event containing `NaN` or `INF` values for iOS

tomekzaw:
## Description

Fixes:
- #1776
- #2757

See https://github.com/software-mansion/react-native-reanimated/issues/1776#issuecomment-1024215820 for details

## Changes

- Fix crash when handling event containing `NaN` or `INF` values

1. Set breakpoint in `NativeProxy.mm:259` to make sure that there is no regression in handling valid Reanimated events.
2. Send an event with `NaN` or `INF` value and set breakpoint in `catch` block to make sure the exception is caught.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/ios/native/NativeProxy.mm
+++ b/ios/native/NativeProxy.mm
@@ -250,7 +250,7 @@ std::shared_ptr<NativeReanimatedModule> createReanimatedModule(
     std::string eventAsString;
     try {
       eventAsString = folly::toJson(convertIdToFollyDynamic([event arguments][2]));
-    } catch (folly::json::parse_error) {
+    } catch (folly::json::parse_error &) {
       // Events from other libraries may contain NaN or INF values which cannot be represented in JSON.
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):

michaelknoch (approved):

michaelknoch:
btw, the same crash happens on android in this line https://github.com/software-mansion/react-native-reanimated/blob/main/android/src/main/cpp/headers/NativeProxy.h#L63. can we also try/catch it?

michaelknoch:
btw, the same crash happens on android in this line https://github.com/software-mansion/react-native-reanimated/blob/main/android/src/main/cpp/headers/NativeProxy.h#L63. can we also try/catch it? @tomekzaw

piaskowyk (approved):

Issue #2814, opened 2022-01-07
java.lang.RuntimeException 'folly::toJson: JSON object value was a NaN when serializing value at "velocity"->"y"'


michaelknoch:
## Description
We see a bunch of these crashes in production, unfortunately I was not able to reproduce it yet in a local dev environment. From the error tracking it also seems to happen when interacting with unrelated ScrollViews. Do you have any idea how to fix or workaround this? I'm happy to contribute, but so far I'm still trying to understand whats wrong.

### Expected behavior
It should not crash

### Actual behavior & steps to reproduce
It creashes with `folly::toJson: JSON object value was a NaN when serializing value at "velocity"->"y"`

```
java.lang.RuntimeException: folly::toJson: JSON object value was a NaN when serializing value at "velocity"->"y"
        at com.facebook.react.bridge.NativeMap.toString(NativeMap.java:-2)
        at com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent(NativeProxy.java:-2)
        at com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent(NativeProxy.java:56)
        at com.facebook.react.uimanager.events.Event.dispatch(Event.java:180)
        at com.swmansion.reanimated.NodesManager.handleEvent(NodesManager.java:515)
        at com.swmansion.reanimated.NodesManager.onEventDispatch(NodesManager.java:489)
        at com.facebook.react.uimanager.events.EventDispatcherImpl.dispatchEvent(EventDispatcherImpl.java:116)
        at com.facebook.react.views.scroll.ReactScrollViewHelper.emitScrollEvent(ReactScrollViewHelper.java:88)
        at com.facebook.react.views.scroll.ReactScrollViewHelper.emitScrollEndDragEvent(ReactScrollViewHelper.java:57)
        at com.facebook.react.views.scroll.ReactScrollView.onTouchEvent(ReactScrollView.java:362)
        at android.view.View.dispatchTouchEvent(View.java:15199)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3914)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:3578)
        at android.widget.ScrollView.dispatchTouchEvent(ScrollView.java:868)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3920)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:3594)
[...]

Issue #2662, opened 2021-11-26
Runtime crash: JSON object value was a NaN when serializing value at "focalX"


vtoupet:
## Description
 The following crash occurs sporadically in production build.

```
java.lang.RuntimeException: folly::toJson: JSON object value was a NaN when serializing value at "focalX"
    at com.facebook.react.bridge.NativeMap.toString(NativeMap.java)
    at com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent(NativeProxy.java)
    at com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent(NativeProxy.java:55)
    at com.swmansion.gesturehandler.react.RNGestureHandlerStateChangeEvent.dispatch(RNGestureHandlerStateChangeEvent.java:80)
    at com.swmansion.reanimated.NodesManager.handleEvent(NodesManager.java:485)
    at com.swmansion.reanimated.NodesManager.onEventDispatch(NodesManager.java:462)
    at com.facebook.react.uimanager.events.EventDispatcherImpl.dispatchEvent(EventDispatcherImpl.java:116)
    at com.swmansion.gesturehandler.react.RNGestureHandlerModule.onStateChange(RNGestureHandlerModule.java:691)
    at com.swmansion.gesturehandler.react.RNGestureHandlerModule.access$300(RNGestureHandlerModule.java:42)
    at com.swmansion.gesturehandler.react.RNGestureHandlerModule$1.onStateChange(RNGestureHandlerModule.java:436)
    at com.swmansion.gesturehandler.GestureHandler.dispatchStateChange(GestureHandler.java:81)
    at com.swmansion.gesturehandler.GestureHandlerOrchestrator.onHandlerStateChange(GestureHandlerOrchestrator.java:200)
    at com.swmansion.gesturehandler.GestureHandler.moveToState(GestureHandler.java:350)
    at com.swmansion.gesturehandler.GestureHandler.cancel(GestureHandler.java:444)
    at com.swmansion.gesturehandler.GestureHandlerOrchestrator.makeActive(GestureHandlerOrchestrator.java:223)
    at com.swmansion.gesturehandler.GestureHandlerOrchestrator.tryActivate(GestureHandlerOrchestrator.java:159)
    at com.swmansion.gesturehandler.GestureHandlerOrchestrator.onHandlerStateChange(GestureHandlerOrchestrator.java:194)
[...]

PR #2901, opened 2022-01-28
Fix crash when handling event containing NaN or INF values for android

michaelknoch:
## Description
Applies the iOS fix in https://github.com/software-mansion/react-native-reanimated/pull/2896 for android.

Fixes:
- #2662
- #2814
- #2914

## Changes
Fix crash when handling event containing NaN or INF values

tested by manually inserting NaN and INF values in receiveEvent.java

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/cpp/headers/NativeProxy.h
+++ b/android/src/main/cpp/headers/NativeProxy.h
@@ -59,16 +59,16 @@ class EventHandler : public HybridClass<EventHandler> {
       jni::alias_ref<JString> eventKey,
       jni::alias_ref<react::WritableMap> event) {
     std::string eventAsString = "{NativeMap:null}";
-    try {
-      if (event != nullptr) {
+    if (event != nullptr) {
+      try {
         eventAsString = event->toString();
+      } catch (std::exception) {
+        // Events from other libraries may contain NaN or INF values which
+        // cannot be represented in JSON. See
+        // https://github.com/software-mansion/react-native-reanimated/issues/1776
+        // for details.
+        return;
       }
-    } catch (const std::exception &) {
-      // Events from other libraries may contain NaN or INF values which cannot
-      // be represented in JSON. See
-      // https://github.com/software-mansion/react-native-reanimated/issues/1776
-      // for details.
-      return;
     }
     handler_(eventKey->toString(), eventAsString);
   }
--- a/android/src/main/cpp/headers/NativeProxy.h
+++ b/android/src/main/cpp/headers/NativeProxy.h
@@ -62,7 +62,7 @@ class EventHandler : public HybridClass<EventHandler> {
     if (event != nullptr) {
       try {
         eventAsString = event->toString();
-      } catch (std::exception) {
+      } catch (folly::json::parse_error &) {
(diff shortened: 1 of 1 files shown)


michaelknoch:
cc @tomekzaw @piaskowyk

android/src/main/cpp/headers/NativeProxy.h
```diff
@@ -59,8 +59,16 @@ class EventHandler : public HybridClass<EventHandler> {
       jni::alias_ref<JString> eventKey,
       jni::alias_ref<react::WritableMap> event) {
     std::string eventAsString = "{NativeMap:null}";
-    if (event != nullptr) {
-      eventAsString = event->toString();
+    try {
+      if (event != nullptr) {
```
tomekzaw, line 63: I'd suggest moving `try/catch` block into `if`.

android/src/main/cpp/headers/NativeProxy.h
```diff
@@ -59,8 +59,16 @@ class EventHandler : public HybridClass<EventHandler> {
...
+        eventAsString = event->toString();
+      }
+    } catch (const std::exception &) {
```
tomekzaw, line 66:
```suggestion
    } catch (folly::json::parse_error &) {
```

tomekzaw (changes requested):
Thanks for the PR! Indeed, we should handle this on Android analogously as in #2896. BTW Here's the proof that `WritableMap::toString` also uses `folly::toJson` under the hood: https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/ReactAndroid/src/main/jni/react/jni/NativeMap.cpp#L19

I left two suggestions in the comments. Once they are resolved, we are ready to merge I think 😄

android/src/main/cpp/headers/NativeProxy.h
```diff
@@ -60,7 +60,15 @@ class EventHandler : public HybridClass<EventHandler> {
       jni::alias_ref<react::WritableMap> event) {
     std::string eventAsString = "{NativeMap:null}";
     if (event != nullptr) {
-      eventAsString = event->toString();
+      try {
+        eventAsString = event->toString();
+      } catch (std::exception) {
```
michaelknoch, line 65: apparently `catch(folly::json::parse_error)` doesn't seem to catch the error @tomekzaw

michaelknoch:
thx @tomekzaw , I addressed your feedback

tomekzaw (approved):
Looks good to me!

michaelknoch:
you think we can get this (and the ios fix) into 2.4.0 release @tomekzaw

tomekzaw:
> you think we can get this (and the ios fix) into 2.4.0 release @tomekzaw

Yes, definitely. These two PRs resolve one of the most mysterious bugs hiding over there. Just waiting to get it double-reviewed (probably tomorrow).

kmagiera (approved):

piaskowyk (approved):

PR #2919, opened 2022-02-02
Check platform better

piaskowyk:
## Description

We need to mock the native module by JS implementation for web, jest, and chrome debugger. We mock it on the condition `nativeShouldBeMock` - `isJest() || isChromeDebugger()` i check it to more proper `shouldBeUseWeb` - `isJest() || isChromeDebugger() || isWeb()`.
It worked because the web had implemented `global.nativeCallSyncHook`, but I am not sure that it will work for every edge case.


kkafar (approved):
👍🏼

Issue #2924, opened 2022-02-02
Circular Require in upcoming version 2.4.0, which would probably cause core methods 'undefined' in worklet.


yiheyang:
## Description

Circular Require in upcoming version 3.4.0, which would probably cause core methods 'undefined' in worklet.

withXXX in worklet function would throw undefined is not a function near "defineAnimation", which is in util.ts(affected in circular requrie)

I compare the working tree vs 2.3.1 then found the reason: 

"import { useSharedValue } from './hook/useSharedValue';" causing all these problems.

## Snack or minimal code example

no need.

## Package versions

| name                                 | version |
| ------------------------------------ | ------- |
| react-native                         |         |
| react-native-reanimated              |         |
| NodeJS                               |         |
| Xcode                                |         |
| Java                                 |         |
| Gradle                               |         |
| expo                                 |         |

## Affected platforms

- [x] Android
- [x] iOS
- [ ] Web

![image](https://user-images.githubusercontent.com/8202550/152225602-05e338b9-f7f4-4b88-b7d3-2f8b54b34ca5.png)

Latropos:
Thank you for submitting the issue! We have solved the problem a long time ago (and we've added `circular_dependency_check` to prevent any similar issues in the future).

PR #2925, opened 2022-02-02
fix circular require in Colors.ts and make better web compatibility

yiheyang:
## Description

1. fix circular require in Colors.ts (#2924).

2. make better web compatibility:

`_WORKLET` might cause an undefined ReferenceError under some web-based build target without a global window like wechat miniprogram, so only using `_WORKLET` to detect whether the platform is RN is not proper.

## Changes

1. removed the useless `useInterpolate` function (which also has type error that would interrupt tsc) and `useSharedValue` is no longer cause the circular require problem.

2. adding `Platform.OS === "web"` test before directly testing `_WORKLET` to prevent the undefined ReferenceError under some web-based build target without a global `window`.

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/Colors.ts
+++ b/src/reanimated2/Colors.ts
@@ -13,7 +13,6 @@ import { interpolate } from './interpolation';
 // @ts-ignore JS file
 import { Extrapolate } from '../reanimated1/derived';
 import { SharedValue } from './commonTypes';
-import { useSharedValue } from './hook/useSharedValue';
 
 interface RGB {
   r: number;
@@ -54,9 +53,8 @@ type Matchers = {
 };
 function getMatchers(): Matchers {
   'worklet';
-  const cachedMatchers: Matchers = _WORKLET
-    ? uiCachedMatchers
-    : jsCachedMatchers;
+  const cachedMatchers: Matchers =
+    Platform.OS !== 'web' && _WORKLET ? uiCachedMatchers : jsCachedMatchers;
   if (cachedMatchers.rgb === undefined) {
     cachedMatchers.rgb = new RegExp('rgb' + call(NUMBER, NUMBER, NUMBER));
     cachedMatchers.rgba = new RegExp(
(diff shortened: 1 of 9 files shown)


piaskowyk:
Hey @yiheyang 👋
Thanks a lot for your PR I really appreciate your commitment! I prepared another one based on your changes - https://github.com/software-mansion/react-native-reanimated/pull/2934 because I found more than one similar problem in our code. If you don't mind I close this PR.

src/reanimated2/Colors.ts
```diff
@@ -772,19 +771,6 @@ export interface InterpolateConfig {
   cache: SharedValue<InterpolateRGB | InterpolateHSV | null>;
 }
 
-export function useInterpolateConfig(
```
piaskowyk, line 775: Actually, this function is used :/ Is exported as a public API.

src/reanimated2/animation/decay.ts
```diff
@@ -46,7 +46,7 @@ export function withDecay(
     'worklet';
     const config: DefaultDecayConfig = {
       deceleration: 0.998,
-      velocityFactor: Platform.OS !== 'web' ? 1 : 1000,
+      velocityFactor: !isWeb() ? 1 : 1000,
```
piaskowyk, line 49: Does it work? I should throw an exception because the `isWeb` function doesn't worklet.

yiheyang:
Thanks for the reply.

I have been actually working several entire days(even just now) with reanimated and trying to understand the code structure and make it fully compatible with a cross platform packer, Taro. I found the code still have many ts type problems and the building toolchains are not enough optimized.

A very big problem would be styleProps behaviors are different from browser and RN environment, as well as the dimension unit.

Still, I made it work fine with most basic animations on Taro.

Layout animation on web would be another huge challenge. I'm afraid there has to be an API usage change for implementing exiting animations and layout animations on web (more tags are required).

src/reanimated2/animation/decay.ts
yiheyang: not fully tested, but AFAIK, it seems hostFunction is also runnable in worklet functions?

piaskowyk:
You are right, the TypeScript in reanimated is problematic because Reanimated was written in plain JS and some time ago we tried to translate it to TypeScript, but there is many dynamic fields so many types are ugly.

piaskowyk:
We want to add support for layout animation for the web but I am not able to say specifically when.

yiheyang:
> We want to add support for layout animation for the web but I am not able to say specifically when.

I'll try to add support for web soon. (with ugly TS  x_x ...)

I've finished the entering part, however the layout / exiting part would be changed into state change based trigger (currently reanimated is implemented by native layout / view update, which cannot be done in DOM).

piaskowyk:
Good luck! Let me know when you will have PoC, I would love to test it 😀

PR #2931, opened 2022-02-03
Handle "Slow animations" iOS simulator option

jakub-gonet:
## Description

Reanimated animations now handle the "Slow animations" option in iOS simulator.

## Changes

`frameTimestamp` in `requestRender` is calculated using drag coefficient if the slow mode is enabled.

## Screenshots / GIFs

Started with slow animations on, disabled halfway through animation.

https://user-images.githubusercontent.com/12465392/152365949-6989db2d-0cc0-4e32-8590-fe43a71c3b70.mp4

### DIFF
--- a/ios/native/NativeProxy.mm
+++ b/ios/native/NativeProxy.mm
@@ -1,3 +1,4 @@
+#import <dlfcn.h>
 
 #import <React/RCTFollyConvert.h>
 #import <React/RCTUIManager.h>
@@ -26,6 +27,32 @@ namespace reanimated {
 
 using namespace facebook;
 using namespace react;
+
+static CGFloat SimAnimationDragCoefficient(void)
+{
+  static float (*UIAnimationDragCoefficient)(void) = NULL;
+#if TARGET_IPHONE_SIMULATOR
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    UIAnimationDragCoefficient = (float (*)(void))dlsym(RTLD_DEFAULT, "UIAnimationDragCoefficient");
+  });
+#endif
+  return UIAnimationDragCoefficient ? UIAnimationDragCoefficient() : 1.f;
+}
+
+static CFTimeInterval calculateTimestampWithSlowAnimations(CFTimeInterval currentTimestamp)
+{
+  static CFTimeInterval dragCoefChangedTimestamp = CACurrentMediaTime();
+
+  const CGFloat dragCoef = SimAnimationDragCoefficient();
+
+  const bool areSlowAnimationsEnabled = dragCoef != 1.f;
+  if (areSlowAnimationsEnabled) {
+    return (dragCoefChangedTimestamp + (currentTimestamp - dragCoefChangedTimestamp) / dragCoef);
+  } else {
+    return currentTimestamp;
+  }
+}
 
 // COPIED FROM RCTTurboModule.mm
 static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &value);
@@ -149,7 +176,7 @@ std::shared_ptr<NativeReanimatedModule> createReanimatedModule(
 
   auto requestRender = [reanimatedModule, &module](std::function<void(double)> onRender, jsi::Runtime &rt) {
     [reanimatedModule.nodesManager postOnAnimation:^(CADisplayLink *displayLink) {
-      double frameTimestamp = displayLink.targetTimestamp * 1000;
+      double frameTimestamp = calculateTimestampWithSlowAnimations(displayLink.targetTimestamp) * 1000;
       jsi::Object global = rt.global();
       jsi::String frameTimestampName = jsi::String::createFromAscii(rt, "_frameTimestamp");
       global.setProperty(rt, frameTimestampName, frameTimestamp);
@@ -158,7 +185,7 @@ std::shared_ptr<NativeReanimatedModule> createReanimatedModule(
     }];
   };
 
-  auto getCurrentTime = []() { return CACurrentMediaTime() * 1000; };
+  auto getCurrentTime = []() { return calculateTimestampWithSlowAnimations(CACurrentMediaTime()) * 1000; };
 
   // Layout Animations start
   REAUIManager *reaUiManagerNoCast = [bridge moduleForClass:[REAUIManager class]];


ios/native/NativeProxy.mm
```diff
@@ -1,3 +1,4 @@
+#import <dlfcn.h>
```
tomekzaw, line 1:
```suggestion
#if TARGET_IPHONE_SIMULATOR
#import <dlfcn.h>
#endif
```

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -141,8 +165,13 @@ private void setGestureState(int handlerTag, int newState) {
   }
 
   @DoNotStrip
-  private String getUpTime() {
-    return Long.toString(SystemClock.uptimeMillis());
+  private String getUptime() {
+    if (slowAnimationsEnabled) {
+      long ANIMATIONS_DRAG_FACTOR = 10;
```
tomekzaw, line 170: make it `private final static`

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -113,6 +118,25 @@ public Scheduler getScheduler() {
     return mScheduler;
   }
 
+  private void toggleSlowAnimations() {
+  }
```
tomekzaw, line 122: looks like it's not used

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -113,6 +118,25 @@ public Scheduler getScheduler() {
...
+
+  private void addDevMenuOption() {
+    ((ReactApplication) mContext.get().getApplicationContext())
+        .getReactNativeHost()
+        .getReactInstanceManager()
+        .getDevSupportManager()
```
tomekzaw, line 128: plz introduce intermediate variables

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -113,6 +118,25 @@ public Scheduler getScheduler() {
...
+        .addCustomDevOption("Toggle slow animations - Reanimated", new DevOptionHandler() {
```
tomekzaw, line 129: looks more professional if you use `–` (ndash) instead of `-` or use parentheses:
```suggestion
        .addCustomDevOption("Toggle slow animations (Reanimated)", new DevOptionHandler() {
```

tomekzaw: BTW indeed, it seems like currently there's no option to dynamically change the text. The only available method for custom dev options is `addCustomDevOption`. I mean, we could do some dirty workaround, but let's keep it clean and nice.

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
jakub-gonet: How do you want it to be split? One variable for devSupportManager, for each component or other way?

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
jakub-gonet: Fixed in [be69639](https://github.com/software-mansion/react-native-reanimated/pull/2931/commits/be696393eb60a1f2cdeb4c4664fdcc8e1bf055c5).

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
jakub-gonet: I extracted devSupportManager as a variable in [be69639](https://github.com/software-mansion/react-native-reanimated/pull/2931/commits/be696393eb60a1f2cdeb4c4664fdcc8e1bf055c5).

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
jakub-gonet: I moved toggling logic there in be69639, it allowed using the method reference syntax to make `addCustomDevOption` a bit nicer to read

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
jakub-gonet: It's inlined anyway. I added `final` for a nice touch.

ios/native/NativeProxy.mm
jakub-gonet: Fixed in be69639.

tomekzaw (approved):
Good job!

ios/native/NativeProxy.mm
```diff
@@ -27,6 +30,41 @@
 using namespace facebook;
 using namespace react;
 
+static CGFloat SimAnimationDragCoefficient(void)
```
piaskowyk, line 33: The compiler should do it by itself but to be sure you can use `static inline` for both definitions.

piaskowyk (commented):
Did you compare performance with these changes and without?

ios/native/NativeProxy.mm
jakub-gonet: In fact, `inline` is just a hint. I wanted to leave it to the compiler since it should trivially optimize this function call (preprocessor macro leaves just `return NULL ? <unreachable> : 1.f` and this function is used only in this translation unit).

jakub-gonet:
> Did you compare performance with these changes and without?
No, we don't have any perf tests written. On iOS it shouldn't change that much (on both release and debug), I'm more concerned about Android (we have an `if` that's always executed). I'd appreciate some pointers on how to test it (or maybe a plan for integrating some perf tests in the repo).

> we could use also preprocessor macros to distinguish debug builds (where slow animations feature is applicable) from release builds (where performance is key).

Not sure if I follow, are you proposing to add some facility that will guard debug code from going into release builds? Do you mean iOS, Android or both?

tomekzaw:
> Not sure if I follow, are you proposing to add some facility that will guard debug code from going into release builds? Do you mean iOS, Android or both?

Yes, it would be nice to add `#if DEBUG` guard that enables slow animations in debug builds but leaves the default (production) case untouched. For iOS, the compiler will most likely optimize it in release mode also due to `TARGET_IOS_SIMULATOR` flag being utilized, but better safe than sorry 😄 

As for Android, I'm wondering why on earth we first convert `long`  to `String` in Java code, then pass the string though JNI to finally parse it on C++ side? 🤔 Maybe it would be better to leave `getUpTime` implementation untouched and make calculations directly in `NativeProxy.cpp`? This way it could be implemented and thus optimized similarly as for iOS. But first let's find out why such a strange double conversion exists in our codebase – `git blame` is your friend. There might have been some specific issue in the past that we don't want to bring back now.

Also, React Native core team mentioned that JNI sometimes might be a bottleneck. Maybe it would be a good idea to migrate this platform-specific code to C++ and use high precision clocks, similarly to `performance.now` implementation both in React Native (https://github.com/facebook/react-native/pull/32695) and Reanimated (https://github.com/software-mansion/react-native-reanimated/pull/2679)? I know this sounds a bit like overengineering but `getCurrentTime` gets called on each frame so we have to keep it performant.

jakub-gonet:
> Yes, it would be nice to add #if DEBUG guard that enables slow animations in debug builds but leaves the default (production) case untouched. For iOS, the compiler will most likely optimize it in release mode but better safe than sorry 😄 Also, debug/release mode is a different build dimension than simulator/device target.

I will add that, then.

> As for Android, I'm wondering why on earth we first convert long to String in Java code, then pass the string though JNI to finally parse it on C++ side to double? 🤔 Maybe it would be better to leave getUpTime implementation untouched and make calculations directly in NativeProxy.cpp? This way it could be implemented and thus optimized similarly as for iOS. But first let's find out why such a strange double conversion exists in our codebase – git blame is your friend. There might have been some specific issue in the past that we don't want to bring back now.
>
> Also, React Native core team mentioned that JNI sometimes might be a bottleneck. Maybe it would be a good idea to migrate this platform-specific code to C++ and use high precision clocks, similarly to performance.now implementation both in React Native (https://github.com/facebook/react-native/pull/32695) and Reanimated (https://github.com/software-mansion/react-native-reanimated/pull/2679)? I know this sounds a bit like overengineering but getCurrentTime gets called on each frame so we have to keep it performant.

I feel those changes are outside of the scope of this PR. I can do them in the followup but I'm afraid we need some perf tests before this issue is tackled to asses if it helps or not.

piaskowyk (approved):

PR #2942, opened 2022-02-04
Resolved another circular dependencies

piaskowyk:
## Description

During tests before release, I noticed another circular dependency, so I create a new module `interpolateColor.ts` to break a cycle.


kkafar (approved):
:+1:

PR #2947, opened 2022-02-05
Add style prop to every animated component

### LINKED ISSUE #1717: Typescript: style prop added by createAnimatedComponent() is flagged

jakub-gonet:
## Description

Fixes #1717.

## Changes

- Added `style` property for every component that has animated props.

- `Added test in react-native-reanimated-tests.tsx`

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/react-native-reanimated-tests.tsx
+++ b/react-native-reanimated-tests.tsx
@@ -81,7 +81,12 @@ const AnimatedFCWithRef = Animated.createAnimatedComponent(SomeFCWithRef);
 
 function CreateAnimatedComponentTest1() {
   const animatedProps = useAnimatedProps(() => ({ fill: 'blue' }));
-  return <AnimatedPath animatedProps={animatedProps} />;
+  return (
+    <AnimatedPath
+      animatedProps={animatedProps}
+      style={{ backgroundColor: 'red' }}
+    />
+  );
 }
 
 function CreateAnimatedComponentTest2() {
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -146,10 +146,14 @@ declare module 'react-native-reanimated' {
               >;
     };
 
-    export type AnimateProps<P extends object> = {
-      [K in keyof P]: K extends 'style'
-        ? StyleProp<AnimateStyle<P[K]>>
-        : P[K] | AnimatedNode<P[K]>;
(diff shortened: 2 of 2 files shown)


tomekzaw (approved):
Thanks for fixing this!

Andarius:
Hey @jakub-gonet, is it expected to break props like `LinearGradientProps` from 'react-native-svg' ?
```ts
export type AnimatedProps = Partial<Animated.AnimateProps<LinearGradientProps>>
```
Fails with 
```
Type 'LinearGradientProps' does not satisfy the constraint 'Record<string, unknown>'.
  Index signature for type 'string' is missing in type 'LinearGradientProps'.ts(2344)
 ```

jakub-gonet:
@Andarius, it shouldn't. Do things get better if you replace `Record<string, unknown>` with `object` like it was before?

Andarius:
Yes it does. I don't get the error in previous versions

PR #2955, opened 2022-02-06
make better universal web compatibility & change color transitions to RGBA-based methods.

yiheyang:
## Description

Currently, there are several problems when we use reanimated in our taro application (a universal bundler based on expo@RN and webpack@h5/mini programs).

On the RN side, everything works just perfect.

However, when it comes to h5 & mini program, reanimated failed to work.

Such cross platform frameworks like taro would usually pack their own components library, but still highly based on React Runtime. In order to make reanimated more universal platform friendly, I implemented two major changes.

Meanwhile, there is also a problem on color transitions: the color transition behavior differs from which with CSS animations, for example, reanimated would give colors in multiple hues when we try to transit grey to blue.

PS: please ignore the package name changes, the change of src folder could be applied directly on react-native-reanimated.

## Changes

1. turn `if (!_WORKLET)` into `if (Platform.OS === 'web' || !_WORKLET)` such cases, in order to prevent an undefined ReferenceError from some environments like mini programs.
2. add a style state in Animated component prototype, which would be changed on any animation update by reanimated, and directly passed into the style prop of the wrapper component, then it would trigger the inner component to update. I guess this way might also help on the support for function components.
3. turn `HSVA` based transition calculation to `RGBA` based one.

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


nandorojo:
You might need to pull the latest main branch into yours

piaskowyk:
Hey 👋 Thanks for your PR and I appreciate that you wanted to spend your time improving Reanimated. But unfortunately, I don't see a clear goal for your changes, moreover, those changes will provide some issues on native platforms :/ I would suggest splitting those changes into multiple PR with single responsibilities to discuss specific changes easier.

PR #2977, opened 2022-02-09
Update path in d.ts after relocation

piaskowyk:
## Description

In `d.ts` was used export from `.ts` file. To make it correct we should use the direct import described here - https://stackoverflow.com/a/51114250/7276742

Thanks for the help for @terrysahaidak 🚀


jakub-gonet (approved):
LGTM, have you tested if it works correctly with the babel TS plugin? (that is if it removes exported types correctly)

piaskowyk:
Yes, I checked it, and everything seems to works.

PR #2978, opened 2022-02-09
Add macro for JSI spec exported functions

jakub-gonet:
## Description

Exported JSI functions seem to follow some name conventions. The function name prefix is quite long and easy to make typos so I added a macro for generating them.

Built and ran the example app


tomekzaw:
Keep in mind that originally this file should be codegen'ed. We commited it to the repo so the Reanimated users don't need to set up codegen which was non-trivial at the time.

jakub-gonet:
That's a valuable comment, I wasn't fully aware this should be codegen'ed. In that case, merge it or close it at your own discretion.

piaskowyk:
I agree with Tomek but I like these changes. It looks better, so I want to merge it 💪

piaskowyk (approved):

Issue #2878, opened 2022-01-25
App crash when set entering and exiting with same Transition


MasonLe2497:
## Description
Im try to create count reaction with layout animtion. Then i get this issue

### Expected behavior
App not crash and working

### Actual behavior & steps to reproduce
See my code bellow:
## Snack or minimal code example
![image](https://user-images.githubusercontent.com/43195241/150902334-3d13b21b-1d16-473a-81b2-24d63f625930.png)

```tsx
import React, {memo, useState} from 'react';
import {Button, SafeAreaView, View} from 'react-native';
import Animated, {SlideInDown} from 'react-native-reanimated';

const AppComponent = () => {
  const [count, setCount] = useState<number>(0);
  // render
  return (
    <SafeAreaView style={{flex: 1}}>
      <Button title="increment" onPress={() => setCount(v => v + 1)} />
      <Button title="decrement" onPress={() => setCount(v => v - 1)} />
      <View style={{height: 40, overflow: 'hidden'}}>
        <Animated.Text entering={SlideInDown} exiting={SlideInDown} key={count}>
          {count}
        </Animated.Text>
      </View>
    </SafeAreaView>
  );
};

const App = memo(AppComponent);
export default App;

```

## Package versions

| name                                 | version |
| ------------------------------------ | ------- |
| react-native                         |      0.66.4   |
| react-native-reanimated              |     2.3.1    |

## Affected platforms

- [ ] Android
- [x ] iOS
- [ ] Web

tomekzaw:
Hello! Thanks for reporting this crash. We're actively working on fixing it.

However, are you sure that you want to use the same Entry Animation both for `entering` and `exiting` props? It makes more sense to use any of pre-defined Exiting Animations for `exiting` prop, for example `SlideOutDown`, like here:

```diff
-<Animated.Text entering={SlideInDown} exiting={SlideInDown} key={count}>
+<Animated.Text entering={SlideInDown} exiting={SlideOutDown} key={count}>
```

BTW Here's my PR that added the reaction counter example: #2855. I hope it'll help. Actually, I had to use custom entry/exit animations due to two reasons:
1. the counter can be either increased or decreased, so we need to dynamically choose exiting animation based on the previous value
2. the default `SlideInUp`/`SlideInDown` and `SlideOutUp`/`SlideOutDown` animations animate the component from the very top/bottom of the screen instead of some constant distance

MasonLe2497:
[...]

PR #2980, opened 2022-02-09
Check for NaN

piaskowyk:
## Description

`CGPointMake(x, y)` doesn't accept `NaN` as argument. I added a check for this case.

### Example
```js
import React, { memo, useState } from 'react';
import { Button, SafeAreaView, View } from 'react-native';
import Animated, { SlideInDown } from 'react-native-reanimated';

const AppComponent = () => {
  const [count, setCount] = useState(0);
  return (
    <SafeAreaView style={{flex: 1}}>
      <Button title="increment" onPress={() => setCount(v => v + 1)} />
      <View style={{height: 40, overflow: 'hidden'}}>
        <Animated.Text exiting={SlideInDown} key={count}>
          {count}
        </Animated.Text>
      </View>
    </SafeAreaView>
  );
};

export default App;
```

Fixes https://github.com/software-mansion/react-native-reanimated/issues/2878

### DIFF
--- a/ios/LayoutReanimation/REAAnimationsManager.m
+++ b/ios/LayoutReanimation/REAAnimationsManager.m
@@ -198,9 +198,9 @@ typedef NS_ENUM(NSInteger, FrameConfigType) { EnteringFrame, ExitingFrame };
 - (double)getSafeDouble:(NSNumber *)number
 {
   double doubleValue = [number doubleValue];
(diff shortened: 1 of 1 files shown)


kkafar (approved):
Would consider naming this method in more descriptive manner `getDouble:(NSNumber *)number OrZero`, besides LGTM

Issue #2508, opened 2021-10-07
Fatal Exception: java.lang.NullPointerException (AnimationsManager.java) v2.3.0-beta.2


pronevich:
## Description

Stack trace from Crashlytics, Android 11, Redmi Note 9 Pro:

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'android.view.WindowManager android.app.Activity.getWindowManager()' on a null object reference
       at com.swmansion.reanimated.layoutReanimation.AnimationsManager.prepareDataForAnimationWorklet(AnimationsManager.java:340)
       at com.swmansion.reanimated.layoutReanimation.AnimationsManager.onViewRemoval(AnimationsManager.java:116)
       at com.swmansion.reanimated.layoutReanimation.ReaLayoutAnimator.dfs(ReanimatedNativeHierarchyManager.java:119)
       at com.swmansion.reanimated.layoutReanimation.ReaLayoutAnimator.dfs(ReanimatedNativeHierarchyManager.java:127)
       at com.swmansion.reanimated.layoutReanimation.ReaLayoutAnimator.dfs(ReanimatedNativeHierarchyManager.java:127)
       at com.swmansion.reanimated.layoutReanimation.ReaLayoutAnimator.dfs(ReanimatedNativeHierarchyManager.java:127)
       at com.swmansion.reanimated.layoutReanimation.ReaLayoutAnimator.dfs(ReanimatedNativeHierarchyManager.java:127)
       at com.swmansion.reanimated.layoutReanimation.ReaLayoutAnimator.dfs(ReanimatedNativeHierarchyManager.java:127)
       at com.swmansion.reanimated.layoutReanimation.ReaLayoutAnimator.dfs(ReanimatedNativeHierarchyManager.java:127)
       at com.swmansion.reanimated.layoutReanimation.ReaLayoutAnimator.deleteView(ReanimatedNativeHierarchyManager.java:106)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.manageChildren(NativeViewHierarchyManager.java:482)
       at com.swmansion.reanimated.layoutReanimation.ReanimatedNativeHierarchyManager.manageChildren(ReanimatedNativeHierarchyManager.java:230)
       at com.facebook.react.uimanager.UIViewOperationQueue$ManageChildrenOperation.execute(UIViewOperationQueue.java:217)
[...]

PR #2982, opened 2022-02-09
Catch & ignore IllegalViewOperationException in AnimationsManager#removeLeftovers 

kkafar:
## Description

This aims to prevent some of the crashes [reported here](https://github.com/software-mansion/react-native-reanimated/issues/2508#issuecomment-1032330992), however it is likely that it is not final solution to the problem.

## Changes

Catch & ignore IllegalViewOperationException in AnimationsManager#removeLeftovers 

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
@@ -233,8 +233,11 @@ public class AnimationsManager implements ViewHierarchyObserver {
     for (int tag : mToRemove) {
       View view = mViewForTag.get(tag);
       if (view == null) {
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):

lenghia241:
Hi, Is there any plan on releasing this fix? we are having crashes on production for users with android 10, 11, 12.

limaAniceto:
For anybody reading this you can add the dependency from commits in main:
https://docs.swmansion.com/react-native-reanimated/docs/2.2.0/installation/#using-main-branch-builds

Hope it's helpful in the meantime 
@lenghia241

PR #2987, opened 2022-02-10
react-native 0.68

piaskowyk:
## Description

Update RN to 0.68


Example/ios/ReanimatedExample/AppDelegate.m
```diff
@@ -1,54 +1,32 @@
...
+    rootView.backgroundColor = [UIColor whiteColor];
   }
-
   self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
-  const bool showTaps = true;
-  if (showTaps) {
-    MBFingerTipWindow *window = [[MBFingerTipWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
-    window.alwaysShowTouches = YES;
-    self.window = window;
-  }
```
tomekzaw, line 40: Can we bring this back (also with additional newlines before and after)?

Example/ios/ReanimatedExample/AppDelegate.m
```diff
@@ -1,54 +1,32 @@
...
 #import <React/RCTRootView.h>
-
-#import <React/RCTImageLoader.h>
-#import <React/RCTLocalAssetImageLoader.h>
-#import <React/RCTGIFImageDecoder.h>
-#import <React/RCTNetworking.h>
-#import <React/RCTHTTPRequestHandler.h>
-#import <React/RCTDataRequestHandler.h>
-#import <React/RCTFileRequestHandler.h>
-#import "MBFingerTipWindow.h"
```
tomekzaw, line 14: Can we bring this line back?

.github/workflows/android-build.yml
```diff
@@ -20,6 +20,15 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+      - name: Install NDK
```
tomekzaw, line 27: How does this affect the execution time of this job? Is it possible to cache this environment in order to optimize it?

Example/ios/Podfile
```diff
@@ -2,12 +2,15 @@ require_relative '../node_modules/react-native/scripts/react_native_pods'
...
 platform :ios, '11.0'
+install! 'cocoapods', :deterministic_uuids => false
 
 config = use_native_modules!
 use_react_native!(
   :path => config[:reactNativePath],
   # to enable hermes on iOS, change `false` to `true` and then install pods
-  :hermes_enabled => false
+  :hermes_enabled => false,
+  :fabric_enabled => false,
```
tomekzaw, line 12: I've noticed that Podfile of 0.68.0-rc.1 template uses `flags` dict here (see https://github.com/facebook/react-native/blob/v0.68.0-rc.1/template/ios/Podfile#L16). Can we copy the whole Podfile from react-native repo to eliminate all differences?

tomekzaw (changes requested):

.github/workflows/android-build.yml
piaskowyk: This adds 1m 10s to test duration. To be clear - this is not necessary, but I just wanted to provide a deterministic environment.

I found a workaround - https://stackoverflow.com/a/62173092/7276742

Example/ios/ReanimatedExample/AppDelegate.m
piaskowyk: Just `#import "MBFingerTipWindow.h"` or all of the imports?

piaskowyk: Update according to documentation `nttld/setup-ndk@v1` action caching NDK by default:
`https://github.com/nttld/setup-ndk` - "This action sets up an Android NDK environment by downloading and caching a version of the NDK and adding it to the PATH"

.github/workflows/android-build.yml
piaskowyk: Update according to documentation nttld/setup-ndk@v1 action caching NDK by default:
https://github.com/nttld/setup-ndk - "This action sets up an Android NDK environment by downloading and caching a version of the NDK and adding it to the PATH"

tomekzaw (approved):

renchap:
Would it be possible to release a new alpha version with RN 0.68 support?
Alternatively, how can I use it directly in my app to test?

Issue #2914, opened 2022-02-01
Android crash - folly::toJson: JSON object value was a NaN


cristianoccazinsp:
## Description
Random crash reported from the crash analytics tool.

```
java.lang.RuntimeException: folly::toJson: JSON object value was a NaN when serializing value at "velocity"->"y"
        at com.facebook.react.bridge.NativeMap.toString(NativeMap.java:-2)
        at com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent(NativeProxy.java:-2)
        at com.swmansion.reanimated.NativeProxy$EventHandler.receiveEvent(NativeProxy.java:56)
        at com.facebook.react.uimanager.events.Event.dispatch(Event.java:180)
        at com.swmansion.reanimated.NodesManager.handleEvent(NodesManager.java:515)
        at com.swmansion.reanimated.NodesManager.onEventDispatch(NodesManager.java:489)
        at com.facebook.react.uimanager.events.EventDispatcherImpl.dispatchEvent(EventDispatcherImpl.java:116)
        at com.facebook.react.views.scroll.ReactScrollViewHelper.emitScrollEvent(ReactScrollViewHelper.java:88)
        at com.facebook.react.views.scroll.ReactScrollViewHelper.emitScrollEndDragEvent(ReactScrollViewHelper.java:57)
        at com.facebook.react.views.scroll.ReactScrollView.onTouchEvent(ReactScrollView.java:362)
        at android.view.View.dispatchTouchEvent(View.java:14360)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3112)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2785)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
[...]

PR #2992, opened 2022-02-10
Fix crash when handling event containing NaN or INF values

michaelknoch:
## Description
This is a follow up pr of https://github.com/software-mansion/react-native-reanimated/pull/2901 which unfortunately doesn't solve [the issue](https://github.com/software-mansion/react-native-reanimated/issues/2814). I do not fully understand why but the try/catch currently has no effect. It's a bit surprising to me because as far as I can see `folly::json::parse_error` should be the same as `std::exception` but I see that the crash is still happening for us in production after updating to 2.4.1 and I can also reproduce it with the example app.

@piotrekzyla @tomekzaw 

## Changes
Make sure the try/catch actually works by catching a more generic error.

I tested the code change with the example app by manually adding a `NaN` value to the event.

Screenshot 2022-02-10 at 23 42 36

 
## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/cpp/headers/NativeProxy.h
+++ b/android/src/main/cpp/headers/NativeProxy.h
@@ -62,7 +62,7 @@ class EventHandler : public HybridClass<EventHandler> {
     if (event != nullptr) {
       try {
         eventAsString = event->toString();
-      } catch (folly::json::parse_error &) {
+      } catch (std::exception &) {
         // Events from other libraries may contain NaN or INF values which
         // cannot be represented in JSON. See
         // https://github.com/software-mansion/react-native-reanimated/issues/1776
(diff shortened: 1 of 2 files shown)


tomekzaw:
Hello! Thanks for another PR. It looks like we should also catch `folly::print_error`, my bad, didn't notice it at first. Here's the comment from re-opened issue: https://github.com/software-mansion/react-native-reanimated/issues/2914#issuecomment-1034745452

michaelknoch:
are you sure? My example is still crashing even when adding `catch (folly::json::print_error &)` but maybe I'm missing something.

Thanks for the quick reply and taking care of this @tomekzaw

michaelknoch:
I pushed [e348b33](https://github.com/software-mansion/react-native-reanimated/pull/2992/commits/e348b3338fa4b1a8e4409212f0a9bc2dcd25c3c0) which also introduces the change to iOS @tomekzaw

tomekzaw (approved):

tomekzaw:
@michaelknoch Thanks! One more review and we'll be merging this.

piaskowyk (approved):

PR #3005, opened 2022-02-16
Autoinstall Android

jakub-gonet:
## Description

This patch removes setup instructions for Android. iOS is in progress (there's some complexity with bridge init lifecycle - Reanimated replaces native RN modules).

## Changes

- Move initialization from `ReanimatedJSIModulePackage` to `ReanimatedModule` by utilizing `@ReactMethod(isBlockingSynchronousMethod = true)` annotation.

I tested it on the example app but maybe it's worth building the package and checking on the freshly generated app.


android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
```diff
@@ -102,6 +103,21 @@ public NodesManager getNodesManager() {
...
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  public void installTurboModule() {
+    // When debugging in chrome the JS context is not available.
+    // https://github.com/facebook/react-native/blob/v0.67.0-rc.6/ReactAndroid/src/main/java/com/facebook/react/modules/blob/BlobCollector.java#L25
+    Utils.isChromeDebugger = getReactApplicationContext().getJavaScriptContextHolder().get() == 0;
+
+    if (!Utils.isChromeDebugger) {
+      this.getNodesManager().initWithContext(getReactApplicationContext());
+    } else {
+      Log.w(
```
kmagiera, line 115: Seems like they use Chrome Debugger in this path, what's the point of showing a warning saying it should be ignored

src/reanimated2/NativeReanimated/NativeReanimated.ts
```diff
@@ -1,31 +1,34 @@
+import { NativeModules } from 'react-native';
 import { SharedValue } from '../commonTypes';
 import { Descriptor } from '../hook/commonTypes';
 
-const InnerNativeModule = global.__reanimatedModuleProxy;
-
 export class NativeReanimated {
   native: boolean;
-  useOnlyV1: boolean;
```
kmagiera, line 8: why is this no longer needed?

src/reanimated2/NativeReanimated/NativeReanimated.ts
```diff
@@ -1,31 +1,34 @@
...
+  private InnerNativeModule: any;
 
   constructor(native = true) {
+    if (global.__reanimatedModuleProxy === undefined) {
+      const { ReanimatedModule } = NativeModules;
+      ReanimatedModule?.installTurboModule();
+    }
+    this.InnerNativeModule = global.__reanimatedModuleProxy;
     this.native = native;
```
kmagiera, line 15: is this being used somewhere? Quick lookup seems to show that we never pass `false` for `native parameter, can you maybe mark that as something we want to remove in the future?

kmagiera (approved):
Looks good 👍  Left a couple of comments. Would be good to test jest environment with these changes as well

src/reanimated2/NativeReanimated/NativeReanimated.ts
jakub-gonet: Check `JSReanimated` c-tor, there's `super(false);` in there. Right now, `.native` is used in two places:
1. `NativeReanimatedModule.installCoreFunctions()` call
2. `requestFrame()` function

src/reanimated2/NativeReanimated/NativeReanimated.ts
jakub-gonet: This property was true only if Reanimated's turbomodule wasn't loaded.`useOnlyV1` was used:
1. To guard `NativeReanimatedModule.installCoreFunctions()` call, which setups Rea2 specific functions
2. inside `isConfigured()`, checking if Rea2 was correctly installed. `isConfigured()` is used only to print errors and install global bindings (`_setGlobalConsole`, `global.performance`).

I'll check if the web platform still works though.

src/reanimated2/NativeReanimated/NativeReanimated.ts
```diff
@@ -1,31 +1,34 @@
...
-const InnerNativeModule = global.__reanimatedModuleProxy;
-
 export class NativeReanimated {
   native: boolean;
-  useOnlyV1: boolean;
+  private InnerNativeModule: any;
 
   constructor(native = true) {
+    if (global.__reanimatedModuleProxy === undefined) {
+      const { ReanimatedModule } = NativeModules;
```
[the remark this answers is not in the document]
jakub-gonet: From my understanding, `native` is always `true` in this code path and this parameter was only introduced for sake of inheritance.

android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
jakub-gonet: `Utils.isChromeDebugger` variable checks if context isn't available. If this is the only situation when it isn't available then we can delete the warning.

\cc @piaskowyk

jakub-gonet:
I can confirm that `yarn test` passes.

android/src/main/java/com/swmansion/reanimated/ReanimatedJSIModulePackage.java
```diff
@@ -1,32 +0,0 @@
...
-import com.facebook.react.bridge.JSIModuleSpec;
-import com.facebook.react.bridge.JavaScriptContextHolder;
-import com.facebook.react.bridge.ReactApplicationContext;
-import java.util.Arrays;
-import java.util.List;
-
-public class ReanimatedJSIModulePackage implements JSIModulePackage {
-
-  @Override
-  public List<JSIModuleSpec> getJSIModules(
```
kmagiera, line 14: One thing that I think of now is that we can actually make the transition less of a pain and keep this class with empty implementation + some warning message. This way people won't get compile errors after upgrading. We can show a yellow box for them to update MainApplication.java

jakub-gonet: I added `@Deprecation` annotation and `Log.w()`. `LogBox` doesn't have a native interface so we'd need to set some global variable or emit an event which isn't ideal.

piaskowyk (approved):

renchap:
Is this supposed to work?
I am upgrading our app to the latest 0.68-rc2 RN version and I replaced the dependancy in `package.json` to use this repo's `main` branch, but I can not get it working:

```
> Task :app:compileDebugJavaWithJavac FAILED
/Users/renchap/dev/notos/packages/mobile/android/app/build/generated/rncli/src/main/java/com/facebook/react/PackageList.java:43: error: cannot find symbol
import com.swmansion.reanimated.ReanimatedPackage;
                               ^
  symbol:   class ReanimatedPackage
  location: package com.swmansion.reanimated
/Users/renchap/dev/notos/packages/mobile/android/app/build/generated/rncli/src/main/java/com/facebook/react/PackageList.java:111: error: cannot find symbol
      new ReanimatedPackage(),
          ^
  symbol:   class ReanimatedPackage
  location: class PackageList
```

I cleaned all Gradle caches and tried to reinstall multiple times, but it looks like Reanimated it not built, or not found

renchap:
Is this supposed to work?
I am upgrading our app to the latest 0.68-rc2 RN version and I replaced the dependancy in `package.json` to use this repo's `main` branch, but I can not get it working:

```
> Task :app:compileDebugJavaWithJavac FAILED
/Users/renchap/dev/notos/packages/mobile/android/app/build/generated/rncli/src/main/java/com/facebook/react/PackageList.java:43: error: cannot find symbol
import com.swmansion.reanimated.ReanimatedPackage;
                               ^
  symbol:   class ReanimatedPackage
  location: package com.swmansion.reanimated
/Users/renchap/dev/notos/packages/mobile/android/app/build/generated/rncli/src/main/java/com/facebook/react/PackageList.java:111: error: cannot find symbol
      new ReanimatedPackage(),
          ^
  symbol:   class ReanimatedPackage
  location: class PackageList
```

I cleaned all Gradle caches and tried to reinstall multiple times, but it looks like Reanimated it not built, or not found

ugurdalkiran:
@renchap i am getting the same issue. did you solve it?

renchap:
@ugurdalkiran I did solve it, but I dont remember how. Probably either:
- upgrading to a latest RN and/or reanimated version (I am running RN 0.70 and reanimated 3.0.0-rc4)j
- by re-reading the updated installation guide and ensuring I did not miss anything

PR #3030, opened 2022-02-22
Fix failing CI

piaskowyk:
## Description

https://github.com/software-mansion/react-native-reanimated/runs/5294182050?check_suite_focus=true
https://github.com/software-mansion/react-native-reanimated/runs/5294181942?check_suite_focus=true

I run java prettier and fix type conflict for `global.performance`.


tomekzaw (approved):
👍

PR #2792, opened 2021-12-28
Old implementation of bezier function

piaskowyk:
## Description

During the performance optimization, I changed the implementation of `Easing.bezier` function, now return an object instead of a plain function, so I added the old implementation in `Easing.bezierFn`

Fixes #2766

Issue #2766, opened 2021-12-18
_reactNativeReanimated.Easing.bezier is not a function _reactNativeReanimated.Easing.bezier is an instance of Object after upgrading to 2.3.1


jeyhk:
## Description

After upgrading my expo project from sdk43 to 44. 
This error came up.

_reactNativeReanimated.Easing.bezier(0.16, 1, 0.3, 1) is not a function _reactNativeReanimated.Easing.bezier(0.16, 1, 0.3, 1) is an instance of Object  

### Expected behavior
should be running just like before
### Actual behavior & steps to reproduce
breaking

## Steps to reproduce
...

    const animatedBoxProps = useAnimatedProps(
    () => ({
      stroke: interpolateColor(
        Easing.bezier(0.16, 1, 0.3, 1)(progress.value),
        [0, 1],
        [boxOutlineColor, highlightColor],
        'RGB'
      ),
      fill: interpolateColor(
        Easing.bezier(0.16, 1, 0.3, 1)(progress.value),
        [0, 1],
        ['#00000000', highlightColor],
        'RGB'
      )
    }),
    [highlightColor, boxOutlineColor]
  );        

## Package versions
-reanimated: 2.3.1

- React Native Reanimated: 2.3.1

felipevlima:
I have the same error.

Ediigas:
estou com esse error

felipevlima:
> estou com esse error

Look, what I did was to downgrade the expo sdk from version 44 to 43 and downgrade the other packages for compatibility and it started working again.

lehaidangdev:
> > estou com esse error
> 
> Look, what I did was to downgrade the expo sdk from version 44 to 43 and downgrade the other packages for compatibility and it started working again.

May I know which packages you have downgraded to? react-native-reanimated, react-native-gesture-handler, ...
Many thanks

felipevlima:
> 
Yes, in my case i need to downgrade this package:
    "expo": "43.0.0",
    "expo-linking": "2.4.2",
    "expo-status-bar": "1.1.0",
    "react-native-gesture-handler": "1.10.2",
    "react-native-reanimated": "2.2.0",
    "react-native-screens": "3.8.0",

When you start your project after downgrade expo to 43, you will see a warning talking about this packages

jeyhk:
> > 
> 
[...]

PR #3034, opened 2022-02-25
Fix bezier & bezierFn mock in ReanimatedV2

notjosh:
## Description

After https://github.com/software-mansion/react-native-reanimated/pull/1879 was implemented (and also #2792), the mocks weren't updated. This PR fixes that.

The error is the same as in https://github.com/software-mansion/react-native-reanimated/issues/2766, but only shows up when running tests.

## Changes

The mocks now match the signature of the functions.

There wasn't anywhere obvious to add a test for this - as far as I can see, we don't have any tests that use the mocks. I'm happy to add one though if you point me in the right direction!

To reproduce, it's as simple as a jest (with mocks set up) test that includes code like:

```typescript
const someEasing = Easing.bezierFn(0, 0, 0.15, 1);
```

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


piaskowyk (approved):
Thanks for the help! ❤️

Issue #2447, opened 2021-09-22
Some gradle flavor names are not supported


tobiasr:
For me (flavor "dev") the following line:

https://github.com/software-mansion/react-native-reanimated/blob/2b7b92b346e5ae5394f38f3bcd2cf2be0b3c737d/android-npm/build.gradle#L51

makes the task to be run

```
:app:mergedevDebugNativeLibs
```

while the correct task would be:

```
:app:mergeDevDebugNativeLibs
```

Is it necessary to `.toLowerCase()`? This seems to work for me:

```
return matcher.group(1)
```

piaskowyk:
Related to: https://github.com/software-mansion/react-native-reanimated/issues/2444

PR #3041, opened 2022-03-01
fix: accept case insensitive flavor

0akl3y:
## Description

This  is an addition to https://github.com/software-mansion/react-native-reanimated/issues/2447 
Since the [fix](https://github.com/software-mansion/react-native-reanimated/pull/2515) does not solve the problem entirely. The Problem only occurs with ReactNative < 0.65.

Gradle also accepts lower case flavors but in react-native-reanimated this will break the android build.
This is because the flavor is matched from the gradle invocation via pattern matching. If it is lowercase the path is not found and the build.gradle exits with an exception.

e.g

Running the following in the consuming react-native project.

```
./gradlew assembledevRelease 

```
Will work in Gradle but will run into the error 

`Task with path ':app:mergedevDebugNativeLibs' not found in project ':app'.`

while

```
./gradlew assembleDevRelease 

```
works as expected and does not run into this error.

## Fixes 
This PR solves this problem so that the flavor is dealt with case insensitively in ReactNative < 0.65 as well.
(`assembledevRelease` will just work the same way as `assembleDevRelease`)

## Changes

- Added a method to get the path to the task regardless of the flavor being upper or lower case 
in the gradle invocation.

## Steps to reproduce the error

Import react-native-reanimated in a ReactNative-Project with version < 0.65
Run a gradle task with lowercase flavor (e.g assembledevRelease)

Gradle will run into the error 

`Task with path ':app:mergedevDebugNativeLibs' not found in project ':app'.`


piaskowyk:
@0akl3y Thanks for the help! ❤️ I added some changes to make it compatible with the latest changes.

PR #3053, opened 2022-03-07
Source build: don't hardcode build dir

janicduplessis:
## Description

It is possible to change the gradle build dir so it is not safe to assume it will be under /build.

## Changes

Use the project build dir from gradle in cmake instead of assuming /build.

## Screenshots / GIFs

N/A

android/build.gradle
```groovy
allprojects {
    buildDir = "${System.getProperty("user.home")}/.gradle/builds/${rootProject.name}/${project.name}"
}
```

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


piaskowyk (approved):
Thanks! 🚀

PR #3055, opened 2022-03-07
Source build: support reactNativeArchitectures

janicduplessis:
## Description

React native support this property to allow building for specific architectures only.

## Changes

Add support for reactNativeArchitectures like in https://github.com/facebook/react-native/blob/main/ReactAndroid/build.gradle#L227.

Use source build with the latest react-native version and run react-native start-android.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -329,6 +329,11 @@ if (findProject(":app")) {
     FOR_HERMES = System.getenv("FOR_HERMES") == "True";
 }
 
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):
Also thanks! 😄

PR #3056, opened 2022-03-07
Source build: Support gradle plugin v7

janicduplessis:
## Description

The task names changed and include the architecture when using the gradle plugin v7. This is compatible with new and old task names.

## Changes

- Register task dependencies in afterEvaluate instead of whenTaskAdded
- Check for tasks containing CMake instead of buildCMake to match tasks like `configureCMakeDebug[arm64-v8a]`
- Make sure task doesn't include Clean to avoid matching clean tasks
- Keep checking for `externalNative` to support older gradle plugin versions

Use gradle plugin@7

android/build.gradle
```groovy
classpath 'com.android.tools.build:gradle:7.1.1'
```

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -695,17 +695,19 @@ task downloadNdkBuildDependencies {
 task prepareThirdPartyNdkHeaders(dependsOn:[downloadNdkBuildDependencies, prepareBoost, prepareDoubleConversion, prepareFolly, prepareGlog]) {
 }
 
-tasks.whenTaskAdded { task ->
-    if (
-        task.name.contains("externalNativeBuild")
-        || task.name.contains("buildCMake")
-    ) {
-        task.dependsOn(prepareThirdPartyNdkHeaders)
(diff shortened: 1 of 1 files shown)


emazzotta:
I have a react-native project using Gradle 7.1.4. I can work on it locally but CI is unable to build the .apk-file due to the error message described in this issue: https://github.com/software-mansion/react-native-reanimated/issues/3084

If this PR could be looked at, it would be greatly appreciated! 🙂

janicduplessis:
RN is moving to cmake so this might not be needed anymore eventually.

emazzotta:
Thanks for your reply, I guess that's fair enough and would prevent such issues going forward. 

However, what will happen to this PR in the meantime, will it just be ignored/closed? 
If so, is the switch to cmake "a quick thing" and planned soon or more of a long-term thing happening over the next months?

piaskowyk (approved):
Thanks for the help! ❤️

PR #3059, opened 2022-03-09
ConfigureProps by JSI

piaskowyk:
## Description

There was a possibility to run updateProps before an update of the allowlist of properties on the native side. It happened because an update of properties allowlist was asynchronous. I added the JSI function to make it update synchronously.

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NativeProxy.java
+++ b/android/src/main/java/com/swmansion/reanimated/NativeProxy.java
@@ -8,6 +8,7 @@ import com.facebook.proguard.annotations.DoNotStrip;
 import com.facebook.react.ReactApplication;
 import com.facebook.react.bridge.NativeModule;
 import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.ReadableNativeArray;
 import com.facebook.react.bridge.WritableArray;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.devsupport.interfaces.DevSupportManager;
@@ -19,8 +20,11 @@ import com.swmansion.reanimated.layoutReanimation.AnimationsManager;
 import com.swmansion.reanimated.layoutReanimation.LayoutAnimations;
 import com.swmansion.reanimated.layoutReanimation.NativeMethodsHolder;
 import java.lang.ref.WeakReference;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 public class NativeProxy {
 
@@ -178,6 +182,22 @@ public class NativeProxy {
   @DoNotStrip
   private float[] measure(int viewTag) {
     return mNodesManager.measure(viewTag);
+  }
+
+  @DoNotStrip
+  private void configureProps(ReadableNativeArray uiProps, ReadableNativeArray nativeProps) {
+    Set<String> nativePropsSet = convertProps(nativeProps);
+    Set<String> uiPropsSet = convertProps(uiProps);
+    mNodesManager.configureProps(nativePropsSet, uiPropsSet);
+  }
+
+  private Set<String> convertProps(ReadableNativeArray props) {
+    Set<String> propsSet = new HashSet<>();
+    ArrayList<Object> propsList = props.toArrayList();
+    for (int i = 0; i < propsList.size(); i++) {
+      propsSet.add((String) propsList.get(i));
+    }
+    return propsSet;
   }
 
   @DoNotStrip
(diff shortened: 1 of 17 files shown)


ios/native/NativeProxy.mm
```diff
@@ -217,6 +217,25 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
     }
   };
 
+  auto configurePropsFunction = [reanimatedModule](
+                                    jsi::Runtime &rt, const jsi::Value &uiProps, const jsi::Value &nativeProps) {
+    NSMutableSet *usPropsSet = [[NSMutableSet alloc] init];
```
tomekzaw, line 222:
```suggestion
    NSMutableSet *uiPropsSet = [[NSMutableSet alloc] init];
```

ios/native/NativeProxy.mm
```diff
@@ -217,6 +217,25 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
...
+    jsi::Array propsNames = uiProps.asObject(rt).asArray(rt);
+    for (int i = 0; i < propsNames.size(rt); i++) {
+      NSString *propName = @(propsNames.getValueAtIndex(rt, i).asString(rt).utf8(rt).c_str());
+      [usPropsSet addObject:propName];
```
tomekzaw, line 226:
```suggestion
      [uiPropsSet addObject:propName];
```

ios/native/NativeProxy.mm
```diff
@@ -217,6 +217,25 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
...
+    }
+
+    NSMutableSet *nativePropsSet = [[NSMutableSet alloc] init];
+    propsNames = nativeProps.asObject(rt).asArray(rt);
+    for (int i = 0; i < propsNames.size(rt); i++) {
+      NSString *propName = @(propsNames.getValueAtIndex(rt, i).asString(rt).utf8(rt).c_str());
+      [nativePropsSet addObject:propName];
+    }
```
tomekzaw, line 234: Can we move it to a separate utility function that accepts `const jsi::Value &props` and returns `NSMutableSet *` to eliminate repetition?

tomekzaw (changes requested):

ios/native/NativeProxy.mm
```diff
@@ -217,6 +217,25 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
...
+
+    [reanimatedModule.nodesManager configureProps:usPropsSet uiProps:nativePropsSet];
```
notjosh, line 236: I'm not familiar enough with this to know if it's a bug or just confusing naming, but "uiPropsSet" not going to the "uiProps" argument isn't intuitively obvious.

In the code it replaced it was the opposite:

```objc
[nodesManager configureProps:[NSSet setWithArray:nativeProps] uiProps:[NSSet setWithArray:uiProps]];
```

piaskowyk: Yes, you are right 👍 I fixed it.

android/src/main/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -180,6 +184,22 @@ private String getUptime() {
     return mNodesManager.measure(viewTag);
   }
 
+  @DoNotStrip
+  private void configureProps(ReadableNativeArray uiProps, ReadableNativeArray nativeProps) {
+    Set<String> nativePropsSet = convertProps(nativeProps);
+    Set<String> uiPropsSet = convertProps(uiProps);
+    mNodesManager.configureProps(nativePropsSet, uiPropsSet);
```
tomekzaw, line 191:
```suggestion
    Set<String> uiPropsSet = convertProps(uiProps);
    Set<String> nativePropsSet = convertProps(nativeProps);
    mNodesManager.configureProps(uiPropsSet, nativePropsSet);
```

ios/native/NativeProxy.mm
```diff
@@ -217,6 +228,13 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
     }
   };
 
+  auto configurePropsFunction = [reanimatedModule](
+                                    jsi::Runtime &rt, const jsi::Value &uiProps, const jsi::Value &nativeProps) {
+    NSSet *uiPropsSet = convertProps(rt, uiProps);
+    NSSet *nativePropsSet = convertProps(rt, nativeProps);
+    [reanimatedModule.nodesManager configureNativeProps:nativePropsSet andUiProps:uiPropsSet];
```
tomekzaw, line 235:
```suggestion
    [reanimatedModule.nodesManager configureUiProps:uiPropsSet andNativeProps:nativePropsSet];
```

tomekzaw (changes requested):
Looks good to me. The only thing is that sometimes the first argument is UI props and sometimes it's native props. Can we change it so it's in the same order everywhere?

tomekzaw (approved):
Good job! 👏

PR #3082, opened 2022-03-17
Use timeout of 100ms on semaphore in performOperations

pistonsky:
## Description

Timeout of 100ms allows to avoid freezing the app when standard Animated API animations are run in parallel with reanimated animations. This addresses #2327 and similar issues.

## Screenshots / GIFs

When the app was freezing, if you attach it to Xcode and pause execution, it always stops on the line that I changed:

Screenshot 2022-03-17 at 02 23 39

I'll try to make a reproducible demo where the freeze issue occurs, and where this PR is fixing it.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/ios/REANodesManager.m
+++ b/ios/REANodesManager.m
@@ -276,7 +276,9 @@
       [strongSelf.uiManager setNeedsLayout];
     });
     if (trySynchronously) {
-      dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
+      // Timeout of 100ms, otherwise this will freeze the app when standard Animated API animations
+      // are run in parallel to reanimated animations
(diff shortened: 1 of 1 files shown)


piaskowyk:
Hey! 👋 Thanks for your PR! Do you know what is the cause of this deadlock? I can understand this workaround fixed issue in this case but this is not the best approach to problem. imo we should find and fix the source of issue instead of patch just effect. Could you prepare reproduction of your issue?

malikzype:
@cltnschlosser Can you please highlight the link to that PR. We are facing the issue in Production with Reanimated v3.0.2 and RN 0.70.6.

cltnschlosser:
#3194

tomekzaw:
Closing this PR since the issue was fixed in #4239 and backported to v2 in #4283.

PR #3090, opened 2022-03-22
Invalidate display link instead of pausing it when invalidating the nodes manager

### LINKED ISSUE #2035: TextInput onBlur animation hot reload iOS crash
## Description

When using `TextInput.onBlur` to trigger an animation, the app crashes on hot reload.

### Expected behavior

The app should not crash and hot reload instead.

The app crashes.

```
import * as React from "react";
import { TextInput } from "react-native";
import { useSharedValue, withTiming } from "react-native-reanimated";

function TextInputTest() {
  const isFocused = useSharedValue(1);
  return <TextInput onBlur={() => (isFocused.value = withTiming(0))} value="foo" />;
}

export default TextInputTest;
```

Focusing this component and then triggering a hot reload will crash on iOS.

## Package versions

- React Native: 0.64.0
- NodeJS: 14.16.1
- Xcode: 12.5

## Affected platforms

- [x] iOS

Did not test other platforms.

### LINKED ISSUE #2899: Hot reload iOS crash due to EXC_BAD_ACCESS on rt.global()
With this issue I'd like to gather all bug reports related to EXC_BAD_ACCESS crash when hot-reloading the app on iOS:

- #2035 (confirmed)
- ...

This is how the crash looks like in its natural environment:

The crash happens because when `requestRender` lambda is executed, the JS runtime has already been deallocated (as a result of hot-reload). Most likely, this is due to some missing cleanup.

**Note:** This bug is mostly relevant in debug mode during development as the JS runtime gets deallocated when reloading the app. However, it also appears in release builds when using code-push updates.

danilobuerger:
## Description

Invalidate display link instead of pausing it when invalidating the nodes manager

Pausing the display link on invalidation will still make it call the onAnimationFrame selector one more time as its already schedule. When this happens animation blocks will be called and crash while accessing the jsi runtime thats already dealloced.

Fixes #2035
Fixes #2899

See #2035 for reproduction

### DIFF
--- a/ios/REANodesManager.m
+++ b/ios/REANodesManager.m
@@ -143,7 +143,7 @@
 - (void)invalidate
 {
   _eventHandler = nil;
-  [self stopUpdatingOnAnimationFrame];
+  [_displayLink invalidate];
 }
 
 - (void)operationsBatchDidComplete


kmagiera (approved):

piaskowyk (approved):
Thanks for the help I really appreciate that! ❤️

PR #3091, opened 2022-03-22
useAnimatedKeyboard()

piaskowyk:
## Description

// todo


hirbod:
Ohhh, super hyped about this one here :D

hirbod:
Ohhh, super hyped about this one here :D

hirbod:
I think something similar is also possible on Android (v 11+ only) using the WindowInsets
https://github.com/tolgahan-arikan/rn-android-window-inset/tree/main/example

kirillzyusko:
@hirbod yes, it's also doable on Android (not only v11+ versions, older versions are also supported, though animation is not always synchronised with the keyboard, but anyway it looks better than without animation 😅).

I'm trying to experiment with this API [here](https://github.com/kirillzyusko/react-native-keyboard-controller) and provide kind of universal solution for both platforms. Leaving a link in case if anyone is interested in it 🙂

savelichalex:
We also did universal hook, that does that :D https://github.com/tonlabs/UIKit/blob/development/casts/keyboard/src/useKeyboardBottomInset.tsx

ChildishForces:
> We also did a universal hook, that does that :D https://github.com/tonlabs/UIKit/blob/development/casts/keyboard/src/useKeyboardBottomInset.tsx

Your hook works very nicely, especially supporting `mode="interactive"`, however there's still a very noticeable delay. This kind of thing isn't present with utilities like PanResponders, so what is the theoretical bottleneck here?

savelichalex:
> > We also did a universal hook, that does that :D https://github.com/tonlabs/UIKit/blob/development/casts/keyboard/src/useKeyboardBottomInset.tsx
> 
> Your hook works very nicely, especially supporting `mode="interactive"`, however there's still a very noticeable delay on both simulator and real device. This kind of thing isn't present with utilities like PanResponders, so what is the theoretical bottleneck here?

Sorry, not sure that I understood you, can you be more specific please?

ChildishForces:
> > > We also did a universal hook, that does that :D https://github.com/tonlabs/UIKit/blob/development/casts/keyboard/src/useKeyboardBottomInset.tsx
> > 
> > 
> > Your hook works very nicely, especially supporting `mode="interactive"`, however there's still a very noticeable delay on both simulator and real device. This kind of thing isn't present with utilities like PanResponders, so what is the theoretical bottleneck here?
> 
> Sorry, not sure that I understood you, can you please be more specific?

Sure, I am referring to the slight delay in position of the offset vs the keyboard. They are just slightly out of sync. If you need me to I can provide a recording. Is it perfectly synchronised for you? Am I perhaps missing something?

savelichalex:
> > > > We also did a universal hook, that does that :D https://github.com/tonlabs/UIKit/blob/development/casts/keyboard/src/useKeyboardBottomInset.tsx
> > > 
> > > 
> > > Your hook works very nicely, especially supporting `mode="interactive"`, however there's still a very noticeable delay on both simulator and real device. This kind of thing isn't present with utilities like PanResponders, so what is the theoretical bottleneck here?
> > 
> > 
> > Sorry, not sure that I understood you, can you please be more specific?
> 
> Sure, I am referring to the slight delay in position of the offset vs the keyboard. They are just slightly out of sync. If you need me to I can provide a recording. Is it perfectly synchronised for you? Am I perhaps missing something?

I guess you're talking about iOS, since AFAIR on Android it's in sync. I don't know for sure, just have a guess that it might be because of [approximate coords gathering](https://github.com/tonlabs/UIKit/blob/development/casts/keyboard/ios/UIKitKeyboardIosFrameListener.m#L94) (since keyboard on iOS is animated with CoreAnimation, one can only get an approximate coords at the time, as the actual animation is done in a different process).

piaskowyk:
The accessor of this PR is here: https://github.com/software-mansion/react-native-reanimated/pull/3263

PR #3094, opened 2022-03-23
fix: Fix `useAnimatedSensor` return type

mrousavy:
It returns `any` now.

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -501,10 +501,19 @@ declare module 'react-native-reanimated' {
     interval: number;
   };
 
+  export type AnimatedSensor = {
+    sensor: SensorValue3D | SensorValueRotation | null;
+    unregister: () => void;
+    isAvailable: boolean;
+    config: {
+      interval: number;
+    };
+  };
+
   export function useAnimatedSensor(
     sensorType: SensorType,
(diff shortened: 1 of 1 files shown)


mrousavy:
cc @jakub-gonet @piaskowyk

react-native-reanimated.d.ts
```diff
@@ -501,10 +501,19 @@ declare module 'react-native-reanimated' {
     interval: number;
   };
 
+  export type AnimatedSensor = {
```
jeremybarbet, line 504: Any reason it's using a `type` rather than an `interface`?

react-native-reanimated.d.ts
```diff
@@ -501,10 +501,19 @@ declare module 'react-native-reanimated' {
...
+    sensor: SensorValue3D | SensorValueRotation | null;
+    unregister: () => void;
+    isAvailable: boolean;
+    config: {
```
jeremybarbet, line 508: Could be using the type defined above
```suggestion
    config: SensorConfig
```

mrousavy:
@jeremybarbet I used exactly the same type decl as the internal file already uses. Ideally we want to keep both the same or have a single source of truth for it

piaskowyk (approved):
Thanks for the help Marc ❤️

Issue #2968, opened 2022-02-08
After upgrade to 2.4.1 types for style prop (Animated.View) are incompatible with Animated.Value


ahanusek:
## Description

After upgrade from 2.2.4 to the latest version (2.4.1) I'm getting many TypeScript errors in all places where I use Animated.View with Animated.Value from Reanimated v1.

### Expected behavior

Types shouldn't throw errors to valid usage of style prop.

### Actual behavior & steps to reproduce

```
 <Animated.View
            style={{
                opacity: animation.current,
            }}
        >
        
        </Animated.View>
```

That's throw error but refactored to

```
 <Animated.View
            style={[{
                opacity: animation.current,
            }]}
        >
        
        </Animated.View>
```

works fine

## Snack or minimal code example

```
const Component = () => {
 
 const animation = useRef(new ReAnimated.Value(Dimensions.get('window').height));}
 ....
 
 <Animated.View
            style={{
                left: 0,
                top: animation.current,  <----- this line throw TS error after upgrade
                right: 0,
                bottom: trueBottom * -1,
                backgroundColor: colors.GRAY_50,
                zIndex: 99,
            }}
        >
        
         ......
        </Animated.View>

```

TS Error:

[...]

PR #3098, opened 2022-03-24
Fix output data type not matching input data

arthurgeron:
## Description

- withSequence, withRepeat, withDelay, withTiming, withSpring, interpolateColor and interpolateColors returns data in the format that's received, most use AnimatableValue (e.g. string, number or number[]). The typing on the other hand always assumes the returned type will be a number, this will stop builds on projects with `strict` enabled and trigger false typing errors.

- Includes Fixes for #2968 

## Changes
- Updated method with generics that identifies the type used and adjust the type returned to match that

### Before
image

### After
image
Example code to reproduce issue:

```typescript

import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withTiming,
  withSequence,
  interpolateColor,

export default function Component() {
  const animatedColor = useSharedValue('#FFFFFF');
  const someNumber = useSharedValue(0);

  useEffect(() => {
    animatedColor.value = withSequence(
      withTiming('#FF0000', {
        duration: 300,
      }),
      withTiming('#FFFFFF', {
        duration: 300,
      })
    );
  }, [animatedColor]);

      someNumber.value,
      [0, 10],
      ['rgba(255,255,255,0)', 'rgba(255,255,255,1)']
    ),
    color: animatedColor.value,
  }));

  return <Animated.View style={style} />;
}```. 
<!--
Please include code that can be used to test this change and short description how this example should work.
This snippet should be as minimal as possible and ready to be pasted into editor (don't exclude exports or remove "not important" parts of reproduction example)
-->

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -402,16 +402,16 @@ declare module 'react-native-reanimated' {
       value: Adaptable<number>,
       config: InterpolationConfig
     ): AnimatedNode<number>;
-    export function interpolateColors(
+    export function interpolateColors<T extends string | number>(
       animationValue: Adaptable<number>,
       {
         inputRange,
         outputColorRange,
       }: {
         inputRange: ReadonlyArray<Adaptable<number>>;
-        outputColorRange: ReadonlyArray<Adaptable<number | string>>;
+        outputColorRange: ReadonlyArray<Adaptable<T>>;
       }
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):
Excelent! Thanks a lot! ❤️

Issue #3117, opened 2022-03-30
2.5.0 version: tvOS complilation failed due to AnimatedSensor


etiennelab:
## Description

Due to new AnimatedSensor feature (the use of CoreMotion in ios files), compilation fails when running on tvOS platform (with react native fork: react-native-tvos). Is it possible in the future to exclude sensors to be able to use this package on tvOS ?

(With previous releases, only patching Podspec file to exclude RCTActionSheet was enough to be able to use this package on tvOS)

Thanks a lot for your great work.

### Expected behavior

Compile on tvOS platform

### Actual behavior & steps to reproduce

Build failed when using  react-native-tvos and reanimated (versons below).

## Package versions

| name                                 | version |
| ------------------------------------ | ------- |
| react-native                         |     react-native-tvos@0.64.2-4    |
| react-native-reanimated              |  2.5.0       |
| NodeJS                               |      14.18.0   |
| Xcode                                |      13.3   |
| Java                                 |         |
| Gradle                               |         |
| expo                                 |      Not using expo   |

## Affected platforms

- [ ] Android
- [ ] iOS
- [ ] Web
- [ x]  tvOS

tomekzaw:
Hello @etiennelab! Thanks for reporting. We will use `#ifdef`s to include sensors only on iOS as you suggested. We will also take a look on RCTActionSheet. Finally, we will add a CI job to ensure that Reanimated compiles successfully on tvOS as well.

PR #3135, opened 2022-04-04
TvOS

piaskowyk:
## Description

Added support for TvOS and CI tests for building.

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/3117


ios/sensor/ReanimatedSensorContainer.m
```diff
@@ -1,5 +1,7 @@
 #import "ReanimatedSensorContainer.h"
+#if !TARGET_OS_TV
```
tomekzaw, line 2: I believe this won't work on MacOS etc.

Maybe it's better to check `TARGET_OS_IOS` or `TARGET_IPHONE_SIMULATOR``? Obviously this has to compile for iPad, iPod etc.

Or maybe let's check if `<CoreMotion/CoreMotion.h>` header exists (similarly as for JSC/Hermes detection)?

tomekzaw (approved):
📺

PR #3143, opened 2022-04-06
Remove `__reanimatedWorkletInit` function and `__worklet` property

tomekzaw:
## Description

This PR removes `__reanimatedWorkletInit` function as well as `__worklet` property in favor of checking for existence of `__workletHash` property.

Additionally, in the past we've had some problems related to using spread operator in worklets (see #1453). However, this seems to work fine now thanks to the fact that we use `transformSync` to apply other plugins before workletization (in particular, stringification). Check out the attached test example. By the way, spread operator is allowed in `_f.asString` (I've confirmed that by running the unit tests on the main branch).

## Changes

- Removed `__reanimatedWorkletInit` function

```tsx

import React from 'react';

function worklet() {
  'worklet';

  const arr = [2, ...[3, 4]];
  console.log(1, ...arr, 5);

  const obj = { b: 2, ...{ c: 3, d: 4 } };
  console.log({ a: 1, ...obj, e: 5 });

  const func = (...args: number[]) => console.log(args);
  func(1, ...[2, 3, 4], 5);
}

export default function WorkletSpreadExample() {
  return (
      <Button title="Run worklet" onPress={() => runOnUI(worklet)()} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/__tests__/plugin.test.js
+++ b/__tests__/plugin.test.js
@@ -75,7 +75,7 @@ describe('babel plugin', () => {
     `;
 
     const { code } = runPlugin(input);
-    expect(code).not.toContain('_f.__worklet = true;');
+    expect(code).not.toContain('_f.__workletHash');
   });
 
   it('removes comments from worklets', () => {
@@ -183,7 +183,7 @@ describe('babel plugin', () => {
     `;
 
     const { code } = runPlugin(input);
-    expect(code).toContain('_f.__worklet = true;');
+    expect(code).toContain('_f.__workletHash');
     expect(code).not.toContain('\\"worklet\\";');
     expect(code).toMatchSnapshot();
   });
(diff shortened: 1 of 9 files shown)


__tests__/plugin.test.js
```diff
@@ -75,7 +75,7 @@ describe('babel plugin', () => {
     `;
 
     const { code } = runPlugin(input);
-    expect(code).not.toContain('__reanimatedWorkletInit');
+    expect(code).not.toContain('_f.__worklet = true;');
```
kmagiera, line 78: Do we need this `__worklet` prop? Looks like it only makes all the worklets generated code one line longer for no purpose. If we use it somewhere we should perhaps just check `__workletHash` prop that at least have some additional purpose

kmagiera (approved):
Looks good, lets just get rid of that __worklet prop if possible. I don't think we need it and was there for legacy purposes

__tests__/plugin.test.js
tomekzaw: Done in af8cda4443cce737d9d08ca7909626ccbb32be12

piaskowyk (approved):
🎉

PR #3148, opened 2022-04-07
Remove Reanimated v1

piaskowyk:
## Description

Drop support for Reanimated v1.


hirbod:
My honest opinion: I think it is too early. I still see plenty of popular libs using REA1.
And when this get merged soon, people will have issues, because it will be: waiting for LayoutAnimation to be fixed and prolly losing REA1 (if this gets merged before LayoutAnimation is resolved), or sticking with REA1 support but not working with LA. I hope you guys can postpone this one here.

IvanIhnatsiuk:
Completely agree with @hirbod. So far LA doesn't look very stable and a lot of libraries use RA1. I hope you will listen to our opinions 😃

piaskowyk:
We want to release Reanimated 3.0.0 (without Reanimated 1 in the package) with support for new architecture and Fabric but we will still support and release Reanimated 2.+ (with the source code of Reanimated 1)  but on another branch. 
Is it acceptable for you? I am open to the discussion.

kmagiera:
As the APi surface for Reanimated 1 is really huge we decided we won't be migrating Rea API 1 to Fabric. The new API has been out for almost two years and we don't think that porting old API to the new architecture is the most impactful thing that we can do. We decided to keep one branch of reanimated releases (possibly 2.x) to include v1 API still but not support Fabric and then the second branch (3.x) where we will continue working on Fabric compatibility.

hirbod:
> We want to release Reanimated 3.0.0 (without Reanimated 1 in the package) with support for new architecture and Fabric but we will still support and release Reanimated 2.+ (with the source code of Reanimated 1) but on another branch. Is it acceptable for you? I am open to the discussion.

I think this is a great suggestion. My first concerns have been that you guys just want to fully drop the v1/v2 mix. 

> As the APi surface for Reanimated 1 is really huge we decided we won't be migrating Rea API 1 to Fabric. The new API has been out for almost two years and we don't think that porting old API to the new architecture is the most impactful thing that we can do. We decided to keep one branch of reanimated releases (possibly 2.x) to include v1 API still but not support Fabric and then the second branch (3.x) where we will continue working on Fabric compatibility.

That is an absolute reasonable decision. I think nobody needs REA 1 to be ported to Fabric. I still think that productive usage of Fabric wont happen for most of the users anytime soon, since most of the community packages are not ported.

As long as you continue to support the v2 branch for a while and then backport new features from version 3, I think the user base will be happy. Or will branch v2 only get bugfixes and no further updates? Would be an understandable decision in any case, although probably too early, as it will simply take at least 1-2 years before Fabric is really relevant.

What will be interesting now, however, is how you want to do it with Expo. As long as an "Auto Install" is possible and Expo managed users can decide whether they want to choose v2 or v3, I think everything is fine. I personally use Expo only in conjunction with a custom dev client and would not be affected, but possibly a large part of the other users.

Thanks for taking community opinions into consideration. Love you SWM guys!

PR #3150, opened 2022-04-07
Use angle bracket imports

tomekzaw:
## Description

This PR replaces all occurrences of `#import "Foo.h"` from Reanimated in `ios/` directrory with `#import <RNReanimated/Foo.h>`.

The purpose of this PR is to facilitate versioning of Reanimated when releasing Expo SDKs.

## Changes

- Use angle bracket imports

1. Run example app on iOS

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


piaskowyk (approved):
I tried to build Reanimated in Example App and build from source on both platforms and everything works.

PR #3166, opened 2022-04-11
Load `RNGestureHandlerModule` lazily on iOS

j-piasecki:
## Description

At the time, the `RNGestureHandlerModule` is loaded when the `createReanimatedModule` method is called. This causes `RCTBridge required dispatch_sync to load RNGestureHandlerModule.` warning to appear. This PR changes it, so the module is loaded only the first time the `setGestureState` function is called.

Install `react-native-reanimated` and `react-native-gesture-handler` and check if the warning appears.

### DIFF
--- a/ios/native/NativeProxy.mm
+++ b/ios/native/NativeProxy.mm
@@ -162,8 +162,12 @@ std::shared_ptr<NativeReanimatedModule> createReanimatedModule(
     scrollTo(viewTag, uiManager, x, y, animated);
   };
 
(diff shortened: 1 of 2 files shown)


piaskowyk (approved):
👌

billnbell:
Can we get this out in a release?

sofiamoniz:
Hi ! I still get the warning :/

sofiamoniz:
Hi! I still get the warning :/

PR #3194, opened 2022-04-23
Fix deadlock in REANodesManager (removes synchronous UI updates)

### LINKED ISSUE #1123: useAnimatedStyle height blocked by main thread, transform is not.

### LINKED ISSUE #2285: [iOS] ScrollView event fatal crash
## Description

REANodeManager try to handle events from ScrollView but it causes a fatal crash. `Mouting block is expected to not be set`

Run code from PR: https://github.com/software-mansion/react-native-reanimated/pull/2122 try to move the content ScrollView.

https://user-images.githubusercontent.com/36106620/129409776-b51c2680-6cd7-4336-9423-8839de073bcd.mov

## Package versions
- React Native Reanimated: 2.3.0.alpha.2

## Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

### LINKED ISSUE #3180: Deadlock in [REANodesManager performOperations]
## Description

performOperations blocks the main queue while waiting for the UIManager queue to do work. It's possible that the UIManager queue is already blocked waiting to perform work on the main queue (See somewhat common usage of `RCTUnsafeExecuteOnMainQueueSync`). This can result in deadlock like the following:
```
Thread 0 Crashed:
0   libsystem_kernel.dylib        	0x00000001b82e7540 semaphore_wait_trap + 8
1   libdispatch.dylib             	0x00000001809ffc00 _dispatch_sema4_wait + 28 (lock.c:139)
2   libdispatch.dylib             	0x0000000180a002b8 _dispatch_semaphore_wait_slow + 132 (semaphore.c:132)
3   MyApp                        	0x0000000104795c64 -[REANodesManager performOperations] + 284 (REANodesManager.m:261)
4   MyApp                        	0x0000000104795aa4 -[REANodesManager onAnimationFrame:] + 452 (REANodesManager.m:209)
5   QuartzCore                    	0x0000000184a5e378 CA::Display::DisplayLink::dispatch_items(unsigned long long, unsigned long long, unsigned long long) + 756 (CADisplay.mm:4177)
6   QuartzCore                    	0x0000000184a64fac display_timer_callback(__CFMachPort*, void*, long, void*) + 372 (CADisplayTimer.cpp:219)
7   CoreFoundation                	0x0000000180cff440 __CFMachPortPerform + 176 (CFMachPort.c:549)
8   CoreFoundation                	0x0000000180d427d4 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE1_PERFORM_FUNCTION__ + 60 (CFRunLoop.c:1996)
9   CoreFoundation                	0x0000000180d45fe0 __CFRunLoopDoSource1 + 596 (CFRunLoop.c:2136)
10  CoreFoundation                	0x0000000180cffebc __CFRunLoopRun + 2380 (CFRunLoop.c:3172)
11  CoreFoundation                	0x0000000180d13468 CFRunLoopRunSpecific + 600 (CFRunLoop.c:3268)
12  GraphicsServices              	0x000000019c8b738c GSEventRunModal + 164 (GSEvent.c:2200)
13  UIKitCore                     	0x00000001836b65d0 -[UIApplication _run] + 1100 (UIApplication.m:3493)
14  UIKitCore                     	0x0000000183434f74 UIApplicationMain + 364 (UIApplication.m:5047)
15  MyApp                        	0x000000010421bf2c main + 68 (AnalyticsKey.swift:29)
16  dyld                          	0x0000000105e9daa4 start + 520 (dyldMain.cpp:879)
```
```
Thread 12:
0   libsystem_kernel.dylib        	0x00000001b82e81a4 __ulock_wait + 8
1   libdispatch.dylib             	0x0000000180a00064 _dlock_wait + 56 (lock.c:326)
2   libdispatch.dylib             	0x00000001809ffe08 _dispatch_thread_event_wait_slow + 56 (lock.c:366)
3   libdispatch.dylib             	0x0000000180a0ebac __DISPATCH_WAIT_FOR_QUEUE__ + 364 (lock.h:330)
4   libdispatch.dylib             	0x0000000180a0e754 _dispatch_sync_f_slow + 144 (queue.c:1762)
5   MyApp                        	0x0000000104849324 RCTUnsafeExecuteOnMainQueueSync + 120 (RCTUtils.m:277)
6   libdispatch.dylib             	0x00000001809ff670 _dispatch_client_callout + 20 (object.m:560)
7   libdispatch.dylib             	0x0000000180a00f18 _dispatch_once_callout + 32 (once.c:52)
8   MyApp                        	0x00000001048494e4 dispatch_once + 20 (once.h:84)
9   MyApp                        	0x00000001048494e4 RCTScreenSize + 68 (RCTUtils.m:343)
10  MyApp                        	0x0000000104815898 -[RCTModalHostShadowView insertReactSubview:atIndex:] + 124 (RCTModalHostViewManager.m:42)
11  MyApp                        	0x0000000104842ad0 RCTSetChildren + 280 (RCTUIManager.m:884)
12  MyApp                        	0x00000001048428e8 -[RCTUIManager setChildren:reactTags:] + 68 (RCTUIManager.m:865)
13  CoreFoundation                	0x0000000180d163a4 __invoking___ + 148
14  CoreFoundation                	0x0000000180d33b74 -[NSInvocation invoke] + 468 (NSForwarding.m:3378)
15  CoreFoundation                	0x0000000180d6a9d4 -[NSInvocation invokeWithTarget:] + 80 (NSForwarding.m:3475)
16  MyApp                        	0x000000010481b928 -[RCTModuleMethod invokeWithBridge:module:arguments:] + 460 (RCTModuleMethod.mm:584)
17  MyApp                        	0x000000010481df2c facebook::react::invokeInner(RCTBridge*, RCTModuleData*, unsigned int, folly::dynamic const&, int, (anonymous namespace)::SchedulingContext) + 540 (RCTNativeModule.mm:183)
18  MyApp                        	0x000000010481db5c operator() + 68 (RCTNativeModule.mm:104)
19  MyApp                        	0x000000010481db5c invocation function for block in facebook::react::RCTNativeModule::invoke(unsigned int, folly::dynamic&&, int) + 112 (RCTNativeModule.mm:95)
20  libdispatch.dylib             	0x00000001809fd924 _dispatch_call_block_and_release + 32 (init.c:1517)
21  libdispatch.dylib             	0x00000001809ff670 _dispatch_client_callout + 20 (object.m:560)
22  libdispatch.dylib             	0x0000000180a06df4 _dispatch_lane_serial_drain + 672 (inline_internal.h:2601)
23  libdispatch.dylib             	0x0000000180a07968 _dispatch_lane_invoke + 392 (queue.c:3937)
24  libdispatch.dylib             	0x0000000180a121b8 _dispatch_workloop_worker_thread + 656 (queue.c:6727)
25  libsystem_pthread.dylib       	0x00000001f181b0f4 _pthread_wqthread + 288 (pthread.c:2599)
26  libsystem_pthread.dylib       	0x00000001f181ae94 start_wqthread + 8
```

There is some documentation in `RCTUIManagerUtils.h` on the threading model of this system. There is a noteworthy piece here:
> This solution assumes that the code running on UIManager queue will never
*explicitly* block the Main queue via calling `RCTUnsafeExecuteOnMainQueueSync`.
> Otherwise, it can cause a deadlock.

What is happening in `performOperations` is effectively a custom implementation of `RCTUnsafeExecuteOnMainQueueSync` and so we get deadlock.

This logic was introduced with https://github.com/software-mansion/react-native-reanimated/pull/1215, I don't fully understand the issue this was fixing or why this needed to be synchronously waiting, but there atleast is a connected issue with some context.

This pull request https://github.com/software-mansion/react-native-reanimated/pull/3082 could "fix" the issue by breaking the deadlock, but I'm thinking more could be improved here instead of just timing out the deadlock. It's also not clear to me what happens when this runs asynchronously instead of synchronously.

Potentially related: https://github.com/software-mansion/react-native-reanimated/issues/2938, https://github.com/software-mansion/react-native-reanimated/issues/2327
This same logic is also on Android, which is likely the root cause of https://github.com/software-mansion/react-native-reanimated/issues/3062 and https://github.com/software-mansion/react-native-reanimated/issues/2251

### Expected behavior

Never deadlock. Generally best practice is to never block the main thread.

performOperations blocks the main queue waiting on the UIManager queue. It's possible that at this same time the UIManager queue is already blocked trying to perform work on the main queue. 

I don't have anything other than stacktraces showing the deadlock.

## Package versions

| react-native                         |   0.66.4      |
| react-native-reanimated              |      2.4.1 (with https://github.com/software-mansion/react-native-reanimated/pull/2681 applied)   |
| NodeJS                               |    16.14.2     |
| Xcode                                |     13.3.1    |

## Affected platforms

- [x] Android (maybe)
- [x] iOS

cltnschlosser:
## Heads up

This is my first attempt to dig in to reanimated, and I'm also new to react-native. So I'd appreciate extra scrutiny and explanation of any concerns.

## Description

Fixes #3180
Would fix #2285 by removing the mounting code

## Changes

Removes synchronous UI updates in REANodesManager (Added by #1215, to fix #1123). In a second commit here I added to the examples that were specified in that pull request and issue. I was unable to reproduce the issue on my device (iPhone 11 Pro). It's possible that I'm missing something or other conditions have improved the performance issues that were seen before. Would love feedback from @Szymon20000 and/or @kyle-ssg since they originally experienced and fixed #1123.

If we wanted to synchronously perform these updates in a more thread safe way, React-core would need to be updated to include something like the following. This would block the UIManager queue and execute on the main queue. Instead of the current approach of blocking the main queue to execute on the UIManager queue. Blocking the main queue is the current dangerous operation.
```objc
void RCTUnsafeExecuteOnMainQueueSyncFromUIManagerQueue(dispatch_block_t block)
{
  RCTAssertUIManagerQueue();

  if (RCTIsMainQueue()) {
    block();
  } else {
    dispatch_sync(dispatch_get_main_queue(), ^{
      pseudoUIManagerQueueFlag = YES;
      block();
      pseudoUIManagerQueueFlag = NO;
    });
  }
}
```

Included a second commit (which should likely be removed or cleaned up) with some test screens.

## Checklist

- [x] Ensured that CI passes


He1nr1chK:
Any updates on this?

He1nr1chK:
I can confirm that this PR does solve the issue for me. I have tested it by using patch-package, I just do not have write access else I would have reviewed it myself.

richardo2016x:
thx the work for this PR by @cltnschlosser, but I must say it's a pity this patch didn't resolve my problem:

my npm versions:

```json
     "react-native": "0.67.4",
     "react-native-reanimated": "~2.8.0", // it's 2.8.0 on my machine
```

I applied the patch, but I can still crash due to concurrency issue on multiple-thread, which is same as the version without patch:

image

I will subscribe the change of this PR, looking forward the progress by @cltnschlosser or anyone else, thx a lot again.

cltnschlosser:
@richardo2016x I think that’s unrelated (https://github.com/software-mansion/react-native-reanimated/issues/2327). But since you can reproduce it, can you find what mutex it throwing the exception and what the invalid argument it’s being passed is?

richardo2016x:
> @richardo2016x I think that’s unrelated (#2327). But since you can reproduce it, can you find what mutex it throwing the exception and what the invalid argument it’s being passed is?

@cltnschlosser  You're right, my screenshot in previous comment is related to #2327, one old, legacy bug. I can reproduce it on `react-native@0.67.4` in my App, but I' not sure if I could give a minimal sample to reproduce it, there're some other react-native-* libraries dependent on the `react-native-reanimated`. If I have spare time, I'd try to make one minimal sample but I'm busy on my project's work recently. And I don't know if one sample on `react-native@0.67.4` is valueable,  because `react-native` has been `0.69.1` for now, the new architecture **FARARIC** has been introduced in `0.68`. Compared to release a standalone patch version(such as 2.8.5) for `react-native-reanimated`, maybe it's better to encourage developer to upgrade to latest RN runtime and use this patch :)

Thx a lot again 👍 this patch inspired me how to debug/tuning `react-native-reanimated` for my App

tomekzaw:
Closing this PR since the issue was fixed in #4239 and backported to v2 in #4283.

PR #3222, opened 2022-05-06
Warn users to not use LayoutAnimation

hirbod:
in its current state, until the rewrite has been released. This issue is too impactful and it should be warned until its resolved.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 <img src="https://user-images.githubusercontent.com/16062886/117443145-ff868480-af37-11eb-8680-648bccf0d0ce.png" alt="React Native Reanimated by Software Mansion" width="100%">
 
+# Important notice
+
+We are currently revising LayoutAnimation and **strongly recommend** to refrain from using it until we are finished with the new version. Usage can cause very unexpected side effects that are difficult to identify and can ultimately crash the app. See the following ticket https://github.com/software-mansion/react-native-reanimated/issues/3124
+
+--
+
 ### React Native's Animated library reimplemented
 
 > Reanimated 2 is here! Check out our [documentation page](https://docs.swmansion.com/react-native-reanimated/) for more information


piaskowyk:
The problem with deallocation of objects occurs only for absolutely positioned components such as modal. Right, It is an important issue but IMO not as important to discourage the use of Layout Animation in every case. Especially since we almost fixed it.

hirbod:
I disagree on this one, since a single use is enough to crash your app. And it does also interfere with other libs like react-native-svg etc. I think a note about the pitfalls should be set. I've encountered plenty of users been stuck on "magical issues" that have been caused by this. I did not add this to trash talk anybody nor throw a bad light, you know I love what you guys are doing. I just think that as long as this issues are persistent and not fixed (also remember fixed versions of Expo), there has to be a warning.

LeviWilliams:
Agreed with @hirbod , a single usage of this (even with the patch in place) can cause plenty of issues across an application. `Especially since we almost fixed it` is not true in practice, been running this in prod for awhile with plenty of weird behaviour and crashes both iOS and Android side (with a simple implementation of the feature). 

I had to implement custom native fixes myself to patch some of them,  many developers who have not spent lots of time debugging these issues would have no clue what is happening. I love what the SM team is doing as well - you guys save us countless hours of work with these libraries. Yet not even warning users about this seems slightly careless

bfricka:
This is such a simple yet important MR. I love SW as well, but this is such an insidious issue, and until it's fixed, it's borderline unethical to knowingly leave it without any warning to users. I've been following this GitHub issues associated with this for months now, as I assumed such a substantial issue would be addressed with some urgency, but there hasn't been any movement.

This is fine. You guys are doing great work for the community, and your priorities are yours alone. But adding a warning seems like it should be a priority at this point.

shamilovtim:
Wow thank you for catching this. I had no idea this was going on.

hirbod:
The rewrite is out in reanimated v3 rc.6 - this PR is obsolete now.

shamilovtim:
What about everyone who is on V1 and V2. Shouldn't we warn them?

hirbod:
The rewrite is going to be backported to v2 - so it's obsolete.

PR #3235, opened 2022-05-12
Fix performance.now() in worklets

graszka22:
## Description

Adds performance.now on cpp side, instead of the js side. The `global` object is overwritten by reanimated, so adding functions to `global` object on js side doesn't make them visible in global scope. Adding them on cpp side makes them visible in global scope.

Add `console.log(performance.now())` to any worklet.
Before: worklet stops executing
After: worklet works and prints current time

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


piaskowyk (approved):
🚀

PR #3237, opened 2022-05-12
Update npm building script

piaskowyk:
## Description

Reanimated v3 no longer required the building of binaries so we can simplify the script.


tomekzaw (approved):
Looks good to me! 👏

PR #3263, opened 2022-05-19
Add useAnimatedKeyboard

graszka22:
## Description

This PR adds `useAnimatedKeyboard` hook

## Example of usage
```js
const translateStyle = useAnimatedStyle(() => ({
}));
```

Examples are added to example apps.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


Example/android/build.gradle
```diff
@@ -4,8 +4,8 @@ buildscript {
     ext {
         buildToolsVersion = "30.0.2"
         minSdkVersion = 21
-        compileSdkVersion = 30
-        targetSdkVersion = 30
+        compileSdkVersion = 31
+        targetSdkVersion = 31
```
piaskowyk, line 8: Why do we need to update SDK versions?

src/reanimated2/hook/useAnimatedKeyboard.ts
```diff
@@ -0,0 +1,34 @@
...
+    };
+    listenerId.current =
+      NativeReanimated.subscribeForKeyboardEvents(keyboardEventData);
+    ref.current = keyboardEventData;
+    isSubscribed.current = true;
+  }
+  useEffect(() => {
+    if (isSubscribed.current === false && ref.current !== null) {
+      // subscribe again after Fast Refresh
+      NativeReanimated.subscribeForKeyboardEvents(ref.current);
```
piaskowyk, line 25: I think that you should unregister the old listener (`listenerId.current`) and then register it again and save the new ID. You can see at useAniamtedSensor, that there is the same case.

ios/native/NativeProxy.mm
```diff
@@ -328,6 +329,22 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
   auto unregisterSensorFunction = [=](int sensorId) { [reanimatedSensorContainer unregisterSensor:sensorId]; };
   // end sensors
 
+  // keyboard events
+
+  static REAKeyboardEventObserver *kbObserver = [[REAKeyboardEventObserver alloc] init];
+  auto subscribeForKeyboardEventsFunction =
+      [](std::function<void(bool isShown, bool isAnimating, int height)> keyboardEventDataUpdater) {
+        return [kbObserver subscribeForKeyboardEvents:^(bool isShown, bool isAnimating, int height) {
+          NSLog(@"isShown: %d, isAnimating: %d, height: %d", isShown, isAnimating, height);
```
piaskowyk, line 338: Probably we want to remove this log.

ios/native/NativeProxy.mm
```diff
@@ -328,6 +329,22 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
   auto unregisterSensorFunction = [=](int sensorId) { [reanimatedSensorContainer unregisterSensor:sensorId]; };
   // end sensors
 
+  // keyboard events
+
+  static REAKeyboardEventObserver *kbObserver = [[REAKeyboardEventObserver alloc] init];
```
piaskowyk, line 334: I think that we should avoid abbreviations if it is not necessary, the `keyboardObserver` name is more informative.

piaskowyk (approved):

Issue #3265, opened 2022-05-21
reanimated 2.8.0 crashing android production app at Attempt to invoke virtual method 'java.util.Iterator java.util.HashSet.iterator()' on a null object reference


evelant:
## Description

App hard crashes sometimes in Android production builds at `AnimationsManager.java:234`. This seems to happen mostly but not always when using `react-native-restart` to restart the app at runtime. Stack:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.Iterator java.util.HashSet.iterator()' on a null object reference
        at com.swmansion.reanimated.layoutReanimation.AnimationsManager.removeLeftovers(AnimationsManager.java:234)
        at com.swmansion.reanimated.layoutReanimation.AnimationsManager.lambda$scheduleCleaning$0$com-swmansion-reanimated-layoutReanimation-AnimationsManager(AnimationsManager.java:291)
        at com.swmansion.reanimated.layoutReanimation.AnimationsManager$$ExternalSyntheticLambda0.run(Unknown:4)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
        at android.os.Looper.loop(Looper.java:237)
        at android.app.ActivityThread.main(ActivityThread.java:8167)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:496)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1100)
```

### Expected behavior
Reanimated should never cause native crashes

### Actual behavior & steps to reproduce
Unknown, use reanimated and attempt to use react-native-restart is the only semi-reliable trigger I've found but it doesn't always crash

## Snack or minimal code example
Unable, but I think it should be sufficient to place a null check at https://github.com/software-mansion/react-native-reanimated/blob/b4ee4ea9a1f246c461dd1819c6f3d48440a25756/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java#L234 

## Package versions

| name                                 | version |
| ------------------------------------ | ------- |
| react-native                         |    0.68.2     |
| react-native-reanimated              |     2.8.0    |
| NodeJS                               |         |
| Xcode                                |         |
[...]

PR #3266, opened 2022-05-21
Fix NPE in AnimationsManager during app shutdown

evelant:
## Description
Fixes https://github.com/software-mansion/react-native-reanimated/issues/3265

## Changes
Add null check in AnimationsManager.java to `removeLeftovers` since the `mToRemove` can be set to null if app is shut down (for example by react-native-restart) before the callback fires.

## Checklist

- [ N/A ] Included code example that can be used to test this change
- [ N/A ] Updated TS types
- [ N/A ] Added TS types tests
- [ N/A ] Added unit / integration tests
- [ N/A ] Updated documentation
- [ ] Ensured that CI passes


piaskowyk (approved):
Thanks for the help! 🙌

PR #3278, opened 2022-05-31
🌲 Reduce bundle size on Web by 80%

nandorojo:
## Description

Fixes #2843.

Reduces Reanimated's bundle size on Web from `85kb` to `17kb`.

## Changes

Zero user-facing changes are in this PR. It is strictly file organization changes to optimize Reanimated for tree shaking. After extensive testing, this PR reduced Reanimated's bundle size on Web by almost 80%.

In `package.json`, you'll notice that there is a new `sideEffects` array. This is a way to tell Webpack which files need to run global code rather than as standalone modules. I looked through every file that runs on Web which edits `global` at the top-level and found 2 files. Ideally, we could move all side effects into a single file rather than colocating them with other code that may go unused. But for now, this works. (Update: I moved some into their own file)

More on Webpack tree shaking and the `sideEffects` field can be found here: https://webpack.js.org/guides/tree-shaking/

## Screenshots / GIFs

I created a fresh [repo](https://github.com/nandorojo/reanimated-tree-shaking) to fix tree shaking for Reanimated. The playground tests it by building in a Next.js app (the most common framework for Web).

Prior to this PR, this imported _all_ code from Reanimated:

```ts
import Animated from 'react-native-reanimated'
```

This resulted in an absolutely massive impact on the Web bundle size, approximately 85kb, including tons of unused code (such as layout animations).

This PR reduces the Reanimated overhead down to about `17kb`. In the future, I may be able to investigate further improvements by seeing where that size is actually coming from. But this PR is a massive step forward, since it provides no changes to the user.

I wrote about my findings [here](https://github.com/nandorojo/reanimated-tree-shaking/issues/1). There are plenty of detailed screenshots there.

### Before

Massive bundle size on Web, nearly unusable.

### After

It's now very optimized for tree shaking.

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes


RazaShehryar:
We need this approved!

nandorojo:
Sounds good!

hirbod:
@natew just stumbled across your Tamagui documentation (especially the bundle size part with reanimated). 
I think this is also interesting for you. (hope you don't mind the random tagging)

nandorojo:
Interesting, I didn’t know this required an extra plugin. It worked fine in my expo app — does expo’s babel preset include this?

nandorojo:
The benefit of `export * as default` is that it maintains tree shaking, since it doesn't pair the imported variables into a single object.

The current approach of making a unified `Animated` variable is that all the code of that object gets included.

I'll see if there is an alternative approach.

nandorojo:
I'm going to try by doing this:

```ts
import * as Animated from './Animated'

export default Animated
```

I think it should still tree shake, will let you know.

piaskowyk:
I tested:
```
import * as Animated from './Animated'
export default Animated
```
and the size increased from 17.1 kB to 37.6 kB 😕

nandorojo:
Hmm yeah I found the same thing...seems like `export * as default` is the only way to tree shake that I see.

I'll keep poking around.

nandorojo:
Related: https://github.com/facebook/metro/issues/760

piaskowyk (approved):

piaskowyk:
@nandorojo could you verify if everything works for you properly?

nandorojo:
yeah! have some calls soon but I'll test it after

thanks for reviewing this closely

nandorojo:
the plug-in should be a direct dependency of reanimated, rather than a peer dependency, right? the current code looks like users will have to install the new plugin.

piaskowyk:
You're right. The plugin is required to run the application.

piaskowyk:
I think you can use your repo (https://github.com/nandorojo/reanimated-tree-shaking) to install reanimated from commit:
```
"react-native-reanimated": "github:software-mansion/react-native-reanimated#e278369025e2abb22258ed3fbd1a78484b9afccf"
```
and then you just move files from `node_modules/react-native-reanimated` to `lib/react-native-reanimated` and everything should work.

I tested it on a new blank react-native app, and on the Example app in the Reanimated repo.

nandorojo:
So I tried doing that, but it seems that there is no `react-native-reanimated/lib` file when I install from that commit. How do I generate `lib`? Is there a build script? I can't seem to find one. 

I want to be sure to test on `lib` and not on `src`, since transpilation can potentially break tree shaking.

nandorojo:
These are the steps I took to make it work in my Next.js app:

1. Install the version from this PR in my app
2. Run `npx @react-native-community/bob build` inside of `node_modules/react-native-reanimated`
  i. Is this correct? 
  ii. Change `main` to `lib/commonjs/index` in `package.json`
  iii. Change `module` to `lib/module/index` in `package.json`
3. In my `next.config.js`:
  i. I added `react-native-reanimated` to `next-transpile-modules`
  ii. I also added `experimental.esmExternals = true` to my next config, so that it would use the `module`
  iii. Maybe we could document this for Next.js users to optimize for tree shaking later

I'm curious if the step I took for 2 above is correct or not. If we're using Bob to build, then we should edit the `package.json` to point to `lib/module` and `lib/commonjs`. If we aren't using Bob – what is being used instead to generate the `lib` folder?

nandorojo:
Screen Shot 2022-06-15 at 3 59 46 PM

Okay, I fixed it by changing the `module` to ESNext in `tsconfig.json`. The output is now correct:

Screen Shot 2022-06-15 at 4 00 24 PM

I'll add that to the PR.

nandorojo:
Okay, tested & ready to merge!

I did the following since your changes:

1. Rebase `main`
2. Edit `tsconfig.json`'s `module` to support newer syntax for `export * as default`
3. Add `lib/index` as a side effect in `package.json`

piaskowyk:
It will be available in the next release - it should happen this week. Thanks for the help 🙌

shamilovtim:
How do you get your editor to work with this export/import style? I've found in the past that this completely busts IDEs and Intellisense.

nandorojo:
maybe set vs code to the newest TS version

rrsalian16:
@nandorojo  I loved moti,

I tried & getting this error 

error: Error: Unable to resolve module framer-motion from /node_modules/moti/src/core/index.ts: framer-motion could not be found within the project or in these directories:
  node_modules
  ../../../node_modules
  1 | export { default as motify } from './motify'
> 2 | export { AnimatePresence } from 'framer-motion'
    |                                  ^
  3 |
  4 | export * from './types'
  5 | export { default as useAnimationState } from './use-animator'


Screenshot 2022-07-29 at 12 46 01 AM

blackbing:
How to check to bundle size difference?

I update react-native-reaniamted@2.4.1 to 2.9.1, but I didn't see the  difference from bundle size via webpack-bundle-analyzer.

nandorojo:
I don't expect this to work on v2, since it includes v1 code that isn't optimized.

nandorojo:
i think there’s been a regression, see #3650

Issue #3209, opened 2022-05-03
Style prop with multiple useAnimatedStyle is not getting merged


intergalacticspacehighway:
## Description
We upgraded reanimated to 2.8.0 from 2.4.1 and saw a weird issue. The issue is mainly arising with passing multiple `useAnimatedStyle` as array in style prop. It doesn't happen in 2.4.1. I think it might have been introduced in 2.5.0.

Also, this happens **only**  when the Animated component is used in the react native screen's header elements as shown in video.

In the video, `animatedStyle2` is overriding `animatedStyle` instead of getting merged in the header. Also, the style gets merged on a fast refresh. So maybe it doesn't merge on the first mount 🤔 

https://user-images.githubusercontent.com/23293248/166405063-55101e88-aece-4c39-8b98-71b8bce7129e.mov

### Expected behavior
- Multiple animated styles should merge when used with react native screen's header elements

### Actual behavior & steps to reproduce
- Multiple animated styles gets overridden by last animated style when used with react native screen's header elements

## Snack or minimal code example
- Here's the [repo with minimal example](https://github.com/intergalacticspacehighway/reanimated-style-merge-repro/blob/master/App.js), App.js includes the code from the screen recording.

## Package versions

| name                                 | version |
| ------------------------------------ | ------- |
| react-native                         |      0.68.1   |
| react-native-reanimated         |    2.8.0         |
| NodeJS                               |      16.3.1   |
| Xcode                                |     13.3.1    |
| Java                                 |      11.0.13   |
| Gradle                               |      7.3.3   |
| expo                                 |      x   |

## Affected platforms

- [x] Android
- [x] iOS
- [ ] Web

Titozzz:
I can reproduce this outside of header elements

jevgeni-geimanen-bolt:
Reproduced somewhat similar in `2.4.1`. Downgrade to `2.3.3` helped.

Originally, issue appears while using [react-native-bottom-sheet](https://github.com/gorhom/react-native-bottom-sheet)

[This particular style](https://github.com/gorhom/react-native-bottom-sheet/blob/c7f2ce26fdaf525951b70b76cd857e0b63cb4865/src/components/bottomSheet/BottomSheet.tsx#L1160) does not get properly merged.
```
    const containerAnimatedStyle = useAnimatedStyle(
      () => ({
        opacity: Platform.OS === 'android' && animatedIndex.value === -1 ? 0 : 1,
        transform: [
          {
[...]

PR #3302, opened 2022-06-20
Fix animation not running on mount

### LINKED ISSUE #3296: [iOS] Animations don't always run on component initialization
## Description

In reanimated 2.8.0, animations that run on component initialization don't always animate. I can't figure out why they work sometimes and don't work other times. The strangest thing is that adding a `console.log` within `useAnimatedStyle` seems to fix the issue. Wrapping the call that sets the shared value to kick off the initial animation in a `setTimeout` also seems to fix it (most of the time).

This bug does not occur in 2.3.1.

I believe this is the underlying issue behind this `react-native-bottom-sheet` issue: https://github.com/gorhom/react-native-bottom-sheet/issues/925

### Expected behavior
Animations run on component initialization.
(Below is with a `console.log` added in `useAnimatedStyle`):

https://user-images.githubusercontent.com/6730148/173665563-8df0791a-0a05-41de-a828-d9179f013849.MP4

Create a component that has animations that run on initialization. Show/hide component. Animations run sometimes, but sometimes they just jump to final value (or partially animate).

(Without `console.log` in `useAnimatedStyle`):

https://user-images.githubusercontent.com/6730148/173665704-8c6f93eb-56e3-4cfd-88cd-7f64b1b02252.MP4

Run on iOS device to see issue: https://snack.expo.dev/@computerjazz/reanimated-init-bug

## Package versions
2.8.0

| react-native                         |   0.68.2      |
| react-native-reanimated              |     2.8.0    |
| expo                                 |   45      |

## Affected platforms

- [ ] Android
- [x] iOS
- [ ] Web

graszka22:
## Description

This PR fixes a bug where animation is not always run on component mount.

Fixes #3296 .

## Changes

Probably there is a race condition where the component is mounted and `maybeFlushUpdateBuffer` is called but then `isNotNativeViewFullyMounted:viewTag` returns that it's not yet mounted. `isNotNativeViewFullyMounted` relies on `UIView`'s `superview`. Then on the next frame it returns it's mounted (because iOS updated `superview`?) but the current code just updates the buffer and doesn't apply the changes to the view. It assumes that `maybeFlushUpdateBuffer` will be called but it was called already and won't be called when animation is running.

I added logic covering that case: if component is mounted (has `superview`) and has an update buffer, apply new props to the buffer, remove buffer for that view from `_componentUpdateBuffer` dict and run `updateProps` again with the new props.

Test on iOS on Paper architecture. Press the top button a few times, the red square sometimes doesn't move. 

```ts
import React, { useState, useEffect } from 'react';
import { Text, View, StyleSheet, TouchableOpacity } from 'react-native';
import Animated, { useAnimatedStyle, useDerivedValue, useSharedValue, withSpring} from 'react-native-reanimated'

const AnimatedLine = React.memo(() => {
  const containerWidth = useSharedValue(0)
  const pctAnim = useSharedValue(0)
  const percentage = 50

  useEffect(() => {
    pctAnim.value = percentage
  }, [percentage, pctAnim])

  const animatedPct = useDerivedValue(() => {
    return withSpring(pctAnim.value, { stiffness: 200, damping: 1000 })
  }, [pctAnim])

  const animStyle = useAnimatedStyle(() => {
    const translateX = (animatedPct.value / 100) * containerWidth.value
    return {
      transform: [{ translateX }],
    }
  }, [])

  return (
    <View style={styles.chartItem}>
      <View
        style={styles.graphics}
        onLayout={({ nativeEvent: { layout } }) => {
          containerWidth.value = layout.width
        }}
      >
        <Animated.View style={[styles.iconContainer, animStyle]}>
          <View style={{ right: 20, width: 34, height: 34, backgroundColor: 'tomato' }} />
        </Animated.View>
      </View>
    </View>
  )
})

  chartItem: {
    flexDirection: 'row',
    alignItems: 'center',
  },
  graphics: {
    flex: 1,
  },
  iconContainer: {
    zIndex: 1,
    position: 'absolute',
    top: 0,
    bottom: 0,
    justifyContent: 'center',
  },
})

  const [hide, setHide] = useState(false);

  return (
    <View style={{ flex: 1 }}>
      <TouchableOpacity
        onPress={() => setHide((prev) => !prev)}
        style={{ padding: 10, backgroundColor: 'gray', marginTop: 50 }}>
        <Text>Toggle</Text>
      </TouchableOpacity>
      {!hide && (
        <>
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
          <AnimatedLine />
        </>
      )}
    </View>
  );
}
```

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/ios/REANodesManager.mm
+++ b/ios/REANodesManager.mm
@@ -403,12 +403,17 @@ using namespace facebook::react;
       propsSnapshot.viewName = viewName;
       _componentUpdateBuffer[viewTag] = propsSnapshot;
       atomic_store(&_shouldFlushUpdateBuffer, true);
+      return;
     } else {
       NSMutableDictionary *lastProps = lastSnapshot.props;
       for (NSString *key in props) {
         [lastProps setValue:props[key] forKey:key];
       }
     }
+  }
+  if (lastSnapshot != nil) {
+    [_componentUpdateBuffer removeObjectForKey:viewTag];
+    [self updateProps:lastSnapshot.props ofViewWithTag:viewTag withName:viewName];
     return;
   }
 


Titozzz:
Doesn't work when applied on 2.8.0 (or it addresses another bug)

graszka22:
@Titozzz do you mean this bug: https://github.com/software-mansion/react-native-reanimated/issues/3209? I haven't looked deeply into that one yet but it looks to me it's a different issue. This one is about animations running on initialisation, that one is about merging styles.

Titozzz:
My understanding of my issue is that if I nest array of styles like [{width}, [useAnimatedStyle1, useAnimatedStyle2]], the first useAnimatedStyle will not be merged with the second one, but first render only, so I end up missing a style for no reason.

So it's a bit in between 😅.
My bug was 100% reproducible on iOS and Android, and if I would use console.log inside style1 it would then be correctly merged ...

Dark magic... 🧙

grantmagdanz:
Any updates on getting this in? We're running into #3296 I think and would love to get it fixed if possible.

inanc-g:
This patch file applies the same changes as the PR for v2.9.1, you can apply it using [patch-package](https://github.com/ds300/patch-package), it should fix #3296 :

```diff
diff --git a/node_modules/react-native-reanimated/ios/REANodesManager.m b/node_modules/react-native-reanimated/ios/REANodesManager.m
index f40467d..9977be0 100644
--- a/node_modules/react-native-reanimated/ios/REANodesManager.m
+++ b/node_modules/react-native-reanimated/ios/REANodesManager.m
@@ -537,12 +537,18 @@
       propsSnapshot.viewName = viewName;
       _componentUpdateBuffer[viewTag] = propsSnapshot;
       atomic_store(&_shouldFlushUpdateBuffer, true);
+      return;
     } else {
       NSMutableDictionary *lastProps = lastSnapshot.props;
       for (NSString *key in props) {
         [lastProps setValue:props[key] forKey:key];
       }
     }
+  }
+
+  if (lastSnapshot != nil) {
+    [_componentUpdateBuffer removeObjectForKey:viewTag];
+    [self updateProps:lastSnapshot.props ofViewWithTag:viewTag withName:viewName];
     return;
   }
 
```

b0n1q:
> This patch file applies the same changes as the PR for v2.9.1, you can apply it using [patch-package](https://github.com/ds300/patch-package), it should fix #3296 :
> 
> ```diff
> diff --git a/node_modules/react-native-reanimated/ios/REANodesManager.m b/node_modules/react-native-reanimated/ios/REANodesManager.m
> index f40467d..9977be0 100644
> --- a/node_modules/react-native-reanimated/ios/REANodesManager.m
> +++ b/node_modules/react-native-reanimated/ios/REANodesManager.m
> @@ -537,12 +537,18 @@
>        propsSnapshot.viewName = viewName;
>        _componentUpdateBuffer[viewTag] = propsSnapshot;
>        atomic_store(&_shouldFlushUpdateBuffer, true);
> +      return;
>      } else {
>        NSMutableDictionary *lastProps = lastSnapshot.props;
>        for (NSString *key in props) {
>          [lastProps setValue:props[key] forKey:key];
>        }
>      }
> +  }
> +
> +  if (lastSnapshot != nil) {
> +    [_componentUpdateBuffer removeObjectForKey:viewTag];
> +    [self updateProps:lastSnapshot.props ofViewWithTag:viewTag withName:viewName];
>      return;
>    }
>  
> ```

it doesn't fix the issue for me

piaskowyk:
I'm closing this because those changes are no longer relevant to the Reanimated 4 codebase :/

PR #3306, opened 2022-06-22
Allow to run a sequence of one animation

graszka22:
## Description

This PR allows to run a sequence of just one animation. It makes little sense, but this is no reason to crash the whole app.

## Changes

Removed code that throws exception when there's just one animation in a sequence.

It was added in https://github.com/software-mansion/react-native-reanimated/commit/66232c6579f429c74819af6c7ad076a6a3929d72 but removing it doesn't break the app.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


piaskowyk (approved):
Well done!

Issue #2906, opened 2022-01-31
React Navigation (Native-stack) header hide screen after reanimated layout animation run [only in Android]


jittuu:
## Description

After reanimated layout animation runs, react-navigation (native-stack) header hide screen. I'm not sure it is because of react-navigation or reanimated issue, but the app still working fine until reanimated layout animation runs, so I open issue here. 

Please check recorded GIF - which can explain error better. 😅

### Expected behavior

react-navigation header should not hide screen.

### Actual behavior & steps to reproduce

1. Go to Hello (working fine first time)
2. Go to Animated
3. Go to Hello (screen is below header now)

![untitled](https://user-images.githubusercontent.com/168965/151761891-ee2a766c-cff5-4735-8296-0ab4b7cc0e59.gif)

## Snack or minimal code example

https://github.com/jittuu/reanimated23-and-nativestack-issue

## Package versions

| name                                 | version |
| ------------------------------------ | ------- |
| react-native                         |   0.64.3      |
| react-native-reanimated              |   2.3.1      |
| NodeJS                               |     v16.13.1    |
| Xcode                                |     13    |
| Java                                 |     15    |
| Gradle                               |         |
| expo                                 |    44     |

## Affected platforms

- [x] Android
- [ ] iOS
- [ ] Web

kakajann:
I am also experiencing this only on android

sashansk3:
Same on react-native-reanimated@2.4.1

alanjhughes:
I have just run into this exact issue. Runs perfectly on iOS. Removing any entering or exiting animations fixes the problem

KrasniqiR:
I'm experiencing something similar in iOS (expo-managed app). If a layout animation container is somewhere in our render-tree (even a background screen), some views are duplicated or overlap.

sallar:
I had the same issue with `native-stack`, changed my app to use `stack` for now until this issue is fixed somehow...

husseinAbdulnabi:
also having this when using native-stack, anyone found any solution or workaround?

mathiasandresen:
same issue here

Guuz:
Can confirm that this is an issue.
As soon as Screen A has an Animated.View and no header and you open Screen B that has a header, the content is below the header. As if the header is positioned above it.

Guuz:
If anyone is interested; this is the workaround we applied:

```ts
/**
[...]

Issue #310, opened 2019-06-13
TouchableWithoutFeedback clips children components


todorone:
It does not respect `overflow` style, while React Native's internal does.

jakub-gonet:
Hey @todorone,
sorry for the late response.

Can you provide more information and ideally, a snack with minimal working example?

todorone:
Hey @jakub-gonet 
Thanks for getting into it. That's the snack - https://snack.expo.io/Hk2_bZEpH It uses `TouchableOpacity`, but i guess it's the same for all RNGH's Touchable*

Source:
```
import React from 'react'
import { View, TouchableOpacity } from 'react-native'
import { TouchableOpacity as RNGHTouchableOpacity } from 'react-native-gesture-handler'

export default class App extends React.Component {
  render() {
    return (
      <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
        <TouchableOpacity style={{ width: 100, height: 50, marginBottom: 50 }}>
          <View style={{ width: 200, height: 50, backgroundColor: 'blue' }} />
        </TouchableOpacity>
        
        <RNGHTouchableOpacity style={{ width: 100, height: 50 }}>
          <View style={{ width: 200, height: 50, backgroundColor: 'grey' }} />
        </RNGHTouchableOpacity>
      </View>
    )
  }
}
```

Result:
![Screenshot_2019-12-03-17-39-11-905_host exp exponent](https://user-images.githubusercontent.com/423701/70065600-0ba83d00-15f4-11ea-8e95-83522e1c03f1.jpg)

PR #3332, opened 2022-06-28
Layout Animations rewrite

kmagiera:
# Description

This PR includes a rewrite of Layout Animations. The aim of this rewrite is to:
- prevent current native View leaks in the native side of the implementation
- reducing the number of Native<->JS calls by keeping the animations configs on the C++ side
- overall clean up of the implementation

To be cleaned up in a later PR:
- [ ] use Enum types for layout animation types ('entering', 'exiting' and 'layout') in both JS and native parts of the implementation
- [ ] schedule layout animation config changes to UI thread instead of locking a mutex in `LayoutAnimationProxy.cpp` (common native implementation)
- [ ] try caching `AnimationsManager` in `LayoutAnimations.java` instead of calling `getNativeModule` (Android implementation)
- [ ] try removing all exiting views when `REAAnimationsManager` is invalidated (iOS implementation)


LeviWilliams:
I've never been so excited for a draft pr 🙌

arunim2405:
Hi folks, please also take https://github.com/software-mansion/react-native-reanimated/issues/310 into consideration 
Thanks a lot for all your hard work

jrhager84:
What all is needed to get this merged? I'm not super familiar with this code, but I'm certainly willing to lend my efforts in any way I can to keep this thing moving forward. :)

davidskaarup:
Is anything happening on this rewrite? We basically cannot use layout animation in its current state and I suspect many are in the same situation.
Is there any workaround we can do to be able to use layoutanimation until a rewrite/fix is out?

src/createAnimatedComponent.tsx
```diff
@@ -533,43 +535,35 @@ export default function createAnimatedComponent(
       setLocalRef: (ref) => {
         // TODO update config
         const tag = findNodeHandle(ref);
-        if (
-          (this.props.layout || this.props.entering || this.props.exiting) &&
-          tag != null
-        ) {
+        const { layout, entering, exiting } = this.props;
+        if ((layout || entering || exiting) && tag != null) {
```
fivecar, line 539: Is this rewrite potentially the right place to also fix #3662? Specifically, `entering` and `exiting` animations are only built/configured when setting the ref; they aren't being updated by `componentDidUpdate` when their values change.

Thanks for attempting this fix, @kmagiera & @piaskowyk .

kmagiera:
We recently resumed work on this one and there are some exciting updates pushed by @jwajgelt

android/src/main/cpp/LayoutAnimations.h
```diff
@@ -22,18 +27,22 @@ class LayoutAnimations : public jni::HybridClass<LayoutAnimations> {
...
-  void notifyAboutEnd(int tag, int cancelled);
+  void progressLayoutAnimation(int tag, const jsi::Value &progress);
+  void endLayoutAnimation(int tag, bool cancelled, std::string type);
 
  private:
   friend HybridBase;
   jni::global_ref<LayoutAnimations::javaobject> javaPart_;
   std::weak_ptr<jsi::Runtime> weakUIRuntime;
+  AnimationStartingBlock _animationStartingBlock;
+  HasAnimationBlock _hasAnimationBlock;
```
tomekzaw, line 45:
```suggestion
  AnimationStartingBlock animationStartingBlock_;
  HasAnimationBlock hasAnimationBlock_;
```

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -288,73 +242,131 @@ - (NSDictionary *)prepareDataForLayoutAnimatingWorklet:(NSMutableDictionary *)cu
...
-  if (_states[tag] == nil) {
-    _states[tag] = [NSNumber numberWithInt:Inactive];
-    _viewForTag[tag] = view;
+  if (view.reactTag) {
+    BOOL hasExitAnimation = _hasAnimationForTag(view.reactTag, @"exiting");
+    BOOL wantAnimateExit = hasExitAnimation;
+
+    for (UIView *subview in [view.reactSubviews copy]) {
+      if ([self removeRecursive:subview fromContainer:view withoutAnimation:(shouldRemove && !hasExitAnimation)]) {
+        wantAnimateExit = YES;
```
tomekzaw, line 293: First I thought we can `break;` the loop here but `[self removeRecursive]` has side-effects so now I think we cannot.

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -288,73 +242,131 @@ - (NSDictionary *)prepareDataForLayoutAnimatingWorklet:(NSMutableDictionary *)cu
...
+    if (wantAnimateExit) {
+      REASnapshot *before;
+      if (hasExitAnimation) {
+        before = [[REASnapshot alloc] init:view];
+      }
+      // start exit animation
+      UIView *originalSuperview = view.superview;
+      NSUInteger originalIndex = [originalSuperview.subviews indexOfObjectIdenticalTo:view];
+      [container removeReactSubview:view];
+      view.userInteractionEnabled = NO;
```
tomekzaw, line 306: Maybe let's explain why we disable user interaction here (if there's more to it than just disabling user interaction). If there's nothing more to it than just disabling user interaction, let's also add a comment so future "us" don't wonder if there's more to it than just disabling user interaction.

ios/LayoutReanimation/REAUIManager.mm
```diff
@@ -85,24 +114,23 @@ - (void)_manageChildren:(NSNumber *)containerTag
...
-    permanentlyRemovedChildren = (NSMutableArray *)[super _childrenToRemoveFromContainer:container
-                                                                               atIndices:removeAtIndices];
-    if (permanentlyRemovedChildren != nil) {
-      for (id<RCTComponent> permanentlyRemovedChild in permanentlyRemovedChildren) {
-        if (_toBeRemovedRegister[containerTag] == nil) {
-          _toBeRemovedRegister[containerTag] = [[NSMutableSet<id<RCTComponent>> alloc] init];
+    BOOL wasProxyRemovalSet = _proxyViewRemoval;
+    if (!_proxyViewRemoval) {
+      id<RCTComponent> container = registry[containerTag];
+      NSArray<id<RCTComponent>> *permanentlyRemovedChildren = [self _childrenToRemoveFromContainer:container
```
tomekzaw, line 121: Do you have some kind of formal proof that all animated views with layout animations are indeed eventually removed?

tomekzaw (changes requested):
The code quality is certainly way better than it was before, hopefully the behavior and performance also improves. To be honest, some parts of iOS code are a bit unclear to me so I'd like to suggest adding some comments here and there. Looking forward to merging this PR once it's complete. 👏

android/src/reactNativeVersionPatch/nativeHierarchyManager/latest/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -28,9 +28,8 @@
 class ReaLayoutAnimator extends LayoutAnimationController {
   private AnimationsManager mAnimationsManager = null;
   private volatile boolean mInitialized = false;
-  private ReactApplicationContext mContext;
-  private WeakReference<NativeViewHierarchyManager> mWeakNativeViewHierarchyManage =
-      new WeakReference<>(null);
+  private final ReactApplicationContext mContext;
+  private final WeakReference<NativeViewHierarchyManager> mWeakNativeViewHierarchyManage;
```
kmagiera, line 32: "Manager" 🙃

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -512,16 +356,150 @@ public void updateLayout(
...
+      view.setClickable(false);
+    }
+    return false;
+  }
+
+  private void registerExitingAncestors(View view) {
+    View parent = (View) view.getParent();
+    while (parent != null && !(parent instanceof RootView)) {
+      int tag = parent.getId();
+      Integer previousValue = mExitingSubviewCountMap.get(tag);
```
kmagiera, line 426: maybe use AtomicInteger such that we can just mutate the value instead of replacing objects in a hasmap (not something v important tho)

e.g.:

```
mExitingSubviewCountMap.putIfAbsent(tag, new AtomicInteger(0)).incrementAndGet()
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -512,16 +356,150 @@ public void updateLayout(
...
+    int tag = view.getId();
+    boolean hasExitAnimation = hasAnimationForTag(tag, "exiting");
+    boolean wantAnimateExit = hasExitAnimation;
+
+    // we might want to keep this view around
+    // because one of the (children's) children
+    // has an exiting animation
+    if (view instanceof ViewGroup) {
+      ViewGroup viewGroup = (ViewGroup) view;
+      ArrayList<View> children = new ArrayList<>();
```
kmagiera, line 385: We can just iterate over children in reverse order (from count-1 down to 0) and then copying children's array wouldn't be needed

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -512,16 +356,150 @@ public void updateLayout(
...
+      for (int i = 0; i < viewGroup.getChildCount(); i++) {
+        View child = viewGroup.getChildAt(i);
+        children.add(child);
+      }
+      for (View child : children) {
+        if (removeOrAnimateExitRecursive(child, viewGroup, shouldRemove && !hasExitAnimation)) {
```
kmagiera, line 391: I wonder if this is inline with the way this used to work. Not sure if a cascading remove was actually happening before. Is this implemented the same way on iOS? Maybe if all the children have no animation we could skip detaching from parent and only call drop method? They'll be removed from the hierarchy as a part of their parent's removal

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -288,73 +242,131 @@ - (NSDictionary *)prepareDataForLayoutAnimatingWorklet:(NSMutableDictionary *)cu
...
+        self->_hasAnimationForTag(view.reactTag, @"exiting");
+  });
+}
+
+- (void)registerExitingAncestors:(UIView *)child
+{
+  UIView *parent = child.superview;
+  while (parent != nil && ![parent isKindOfClass:[RCTRootView class]]) {
+    if (parent.reactTag != nil) {
+      _ancestorsOfExitingViews[parent.reactTag] = @([_ancestorsOfExitingViews[parent.reactTag] intValue] + 1);
```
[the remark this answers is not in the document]
kmagiera: I like the name used in the java code: "mExitingSubviewCountMap"

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -288,73 +242,131 @@ - (NSDictionary *)prepareDataForLayoutAnimatingWorklet:(NSMutableDictionary *)cu
...
+
+- (void)maybeDropAncestors:(UIView *)child
+{
+  UIView *parent = child.superview;
+  while (parent != nil && ![parent isKindOfClass:[RCTRootView class]]) {
+    UIView *view = parent;
+    parent = parent.superview;
+    if (view.reactTag != nil) {
+      int trackingCount = [_ancestorsOfExitingViews[view.reactTag] intValue] - 1;
+      if (trackingCount <= 0) {
```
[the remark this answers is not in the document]
kmagiera: should never fall below 0, so the latter

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -288,73 +242,131 @@ - (NSDictionary *)prepareDataForLayoutAnimatingWorklet:(NSMutableDictionary *)cu
...
     }
-    return;
   }
-  _states[tag] = [NSNumber numberWithInt:Disappearing];
-  NSDictionary *preparedValues = [self prepareDataForAnimatingWorklet:startValues frameConfig:ExitingFrame];
-  _startAnimationForTag(tag, @"exiting", preparedValues, @(0));
 }
 
-- (void)onViewCreate:(UIView *)view after:(REASnapshot *)after
+- (BOOL)removeRecursive:(UIView *)view fromContainer:(UIView *)container withoutAnimation:(BOOL)shouldRemove;
```
kmagiera, line 285: I'd rename `shouldRemove` to `removeImmediately` which should better match the prefix "withoutAnimation:"

kmagiera (commented):
Left a couple more comments. Won't be reviewing objc/c++ bits as I've written a big chunk of it.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
tomekzaw: Can we do the same in iOS part to avoid copying?

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -11,35 +11,100 @@ namespace reanimated {
...
-        ValueWrapper::asFrozenObject(sv->value->valueContainer);
-    this->notifyAboutProgress(tag, newValue->shallowClone(rt));
+    this->progressHandler(tag, sv->value->toJSValue(rt).asObject(rt));
   });
 }
 
-void LayoutAnimationsProxy::stopObserving(int tag, bool finished) {
+void LayoutAnimationsProxy::stopObserving(
+    int tag,
+    bool finished,
```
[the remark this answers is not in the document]
jwajgelt: As far as I can tell, the flag here is `finished` instead of `cancel`, since that's what the callbacks on animations are passed - I'm not sure being consistent with either of these two makes more sense than the other.

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.h
```diff
@@ -5,28 +5,46 @@
...
 #include <memory>
+#include <mutex>
+#include <string>
 
 namespace reanimated {
 
 using namespace facebook;
 
 class MutableValue;
+class ShareableValue;
```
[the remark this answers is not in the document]
jwajgelt: We can't - the headers for these classes internally refer to `LayoutAnimationsProxy`, so you need forward declaration in one of them.

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.h
```diff
@@ -5,28 +5,46 @@
...
+      const std::string &type,
+      const jsi::Object &values);
 
  private:
-  std::function<void(int, jsi::Object newProps)> notifyAboutProgress;
-  std::function<void(int, bool)> notifyAboutEnd;
+  std::function<void(int, jsi::Object newProps)> progressHandler;
+  std::function<void(int, bool, std::string)> endHandler;
   std::map<int, std::shared_ptr<MutableValue>> observedValues;
+  std::map<int, std::shared_ptr<ShareableValue>> viewSharedValues;
```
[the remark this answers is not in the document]
jwajgelt: I don't think there's a reason.
HashMaps will probably behave better here anyway, so I'll change these.

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -230,20 +230,52 @@ void NativeProxy::installJSIBindings(
...
+        auto rt = wrt.lock();
+        auto layoutAnimationsProxy = weakLayoutAnimationsProxy.lock();
+        if (!rt || !layoutAnimationsProxy) {
+          return;
+        }
+        jsi::Object yogaValues(*rt);
+        for (const auto &entry : *values) {
+          auto key =
+              jsi::String::createFromAscii(*rt, entry.first->toStdString());
+          auto value = stod(entry.second->toStdString());
```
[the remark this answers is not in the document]
jwajgelt: We definitely expect them to be.
Let's wrap it in a `try`/`catch` anyway, just to be sure.

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -230,20 +230,52 @@ void NativeProxy::installJSIBindings(
...
+
+        layoutAnimationsProxy->startLayoutAnimation(
+            *rt, tag, type->toStdString(), yogaValues);
+      });
+
+  layoutAnimations->cthis()->setHasAnimationBlock(
+      [weakLayoutAnimationsProxy](int tag, std::string type) {
+        auto layoutAnimationsProxy = weakLayoutAnimationsProxy.lock();
+        return layoutAnimationsProxy &&
+            layoutAnimationsProxy->hasLayoutAnimation(tag, type);
```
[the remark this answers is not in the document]
jwajgelt: I don't see how either of these options is better than what's already there? Is the logical `and` here confusing in any way?

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -11,35 +11,100 @@ namespace reanimated {
...
+    std::string type) {
```
piaskowyk, line 33:
```suggestion
    const std::string &type) {
```

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -11,35 +11,100 @@ namespace reanimated {
...
+    layoutAnimations[tag] = config;
+  }
+  viewSharedValues[tag] = viewSharedValue;
+}
+
+bool LayoutAnimationsProxy::hasLayoutAnimation(
+    int tag,
+    const std::string &type) {
+  auto lock = std::unique_lock<std::mutex>(animationsLock);
+  if (type == "entering") {
```
piaskowyk, line 63: Do we really need `type` as string? Could it be an enum type?

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -100,99 +105,48 @@ - (void)scheduleCleaning
...
-  for (NSNumber *viewTag in roots) {
-    UIView *view = _viewForTag[viewTag];
-    [self dfs:view view:view cands:_toRemove];
+  NSNumber *tag = [view reactTag];
+
+  // we'll remove this view anyway when exiting from recursion,
+  // no need to remove it in `maybeDropAncestors`
+  [_ancestorsToRemove removeObject:tag];
+
+  for (UIView *child in [[view subviews] copy]) {
```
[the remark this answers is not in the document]
jwajgelt: There is a need to copy, since changing the array (via `[view removeFromSuperview]` in the next recursion level) during iteration throws an exception.
I doubt copying here costs us much and I feel it's more readable this way than writing an explicit `for` loop over indices.

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -100,99 +105,48 @@ - (void)scheduleCleaning
...
+  if ([_exitingViews objectForKey:tag]) {
+    [_exitingViews removeObjectForKey:tag];
+    [self maybeDropAncestors:view];
   }
+
+  [view removeFromSuperview];
 }
 
-- (void)notifyAboutProgress:(NSDictionary *)newStyle tag:(NSNumber *)tag
+- (void)progressLayoutAnimationWithStyle:(NSDictionary *)newStyle forTag:(NSNumber *)tag
```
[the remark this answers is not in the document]
jwajgelt: Pretty sure it's here like this to mirror the `setNewProps` signature below.

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -11,35 +11,100 @@ namespace reanimated {
...
-void LayoutAnimationsProxy::notifyAboutCancellation(int tag) {
-  this->notifyAboutEnd(tag, false);
+void LayoutAnimationsProxy::configureAnimation(
+    int tag,
+    const std::string &type,
+    std::shared_ptr<ShareableValue> config,
+    std::shared_ptr<ShareableValue> viewSharedValue) {
+  auto lock = std::unique_lock<std::mutex>(animationsLock);
+  if (type == "entering") {
+    enteringAnimations[tag] = config;
```
piaskowyk, line 50: Do you ever clean `enteringAnimations` dict?

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -288,73 +242,131 @@ - (NSDictionary *)prepareDataForLayoutAnimatingWorklet:(NSMutableDictionary *)cu
...
-- (void)onViewRemoval:(UIView *)view before:(REASnapshot *)before
+- (BOOL)wantsHandleRemovalOfView:(UIView *)view
 {
-  NSNumber *tag = view.reactTag;
-  ViewState state = [_states[tag] intValue];
-  if (state == Disappearing || state == ToRemove || tag == nil) {
-    return;
+  return REANodeFind(view, ^(id<RCTComponent> view) {
+    return [self->_ancestorsOfExitingViews objectForKey:view.reactTag] != nil ||
+        self->_hasAnimationForTag(view.reactTag, @"exiting");
```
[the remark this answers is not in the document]
jwajgelt: For (non-mutable) `NSString` literals, Objective-C runtime is supposed to keep a cache of literals and reuse them instead of allocating, so it shouldn't make a difference here.

ios/native/NativeProxy.mm
```diff
@@ -409,7 +375,28 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
...
     }
   }];
+#else
+  // Layout Animation callbacks setup
+  [animationsManager setAnimationStartingBlock:^(
+                         NSNumber *_Nonnull tag, NSString *type, NSDictionary *_Nonnull values, NSNumber *depth) {
+    std::shared_ptr<jsi::Runtime> rt = wrt.lock();
+    if (wrt.expired()) {
+      return;
+    }
```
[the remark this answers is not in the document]
jwajgelt: Here, there should be no difference.

android/src/main/cpp/LayoutAnimations.h
```diff
@@ -22,18 +27,22 @@ class LayoutAnimations : public jni::HybridClass<LayoutAnimations> {
       int tag,
       alias_ref<JString> type,
       alias_ref<JMap<jstring, jstring>> values);
-  void removeConfigForTag(int tag);
+  bool hasAnimationForTag(int tag, std::string type);
```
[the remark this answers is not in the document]
jwajgelt: Pretty sure we can't, since it's a method exported to Java (I'm not a JNI expert though, so maybe?)

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
jwajgelt: `putIfAbsent` complains about Android version support, so I'm not sure if we can use it here.
As for storing a mutable in the map, I don't know. I'd say it's fine as is.

ios/LayoutReanimation/REAUIManager.mm
jwajgelt: @kmagiera ?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
jwajgelt: I don't think it worked like this before, but I'd say it's easier than trying to identify roots of non-animated subtrees and then removing them separately.
We can try changing this at some point in the future and seeing if it changes anything.

android/src/main/cpp/LayoutAnimations.cpp
```diff
@@ -17,70 +17,50 @@ void LayoutAnimations::setWeakUIRuntime(std::weak_ptr<jsi::Runtime> wrt) {
...
+          "endLayoutAnimation");
+  method(javaPart_.get(), tag, cancelled, removeView);
+}
+
+void LayoutAnimations::setHasAnimationBlock(
+    HasAnimationBlock hasAnimationBlock) {
+  this->hasAnimationBlock_ = hasAnimationBlock;
+}
+
+bool LayoutAnimations::hasAnimationForTag(int tag, std::string type) {
```
piaskowyk, line 62:
```suggestion
bool LayoutAnimations::hasAnimationForTag(int tag, const std::string &type) {
```

jwajgelt: This doesn't compile - this function is called from Java, and it seems `MakeNativeMethod` doesn't like functions which take references.

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
jwajgelt: It doesn't seem like we do 🤔

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
jwajgelt: Let's leave it for later, since it requires changing a lot.

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
jwajgelt: Should be good now.

android/src/main/cpp/NativeProxy.cpp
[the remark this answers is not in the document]
tomekzaw: Yeah, it's not a code golf competition. Both suggestions are more explicit and certainly easier to understand when you read the code for the first time. But let's leave it this way.

ios/LayoutReanimation/REAUIManager.mm
kmagiera: not really a formal proof but after the change there is only a single code path that adds and another one that removes the views from the "exiting views" map. Another imporant factor is that we now only store references to the "exiting views" as opposed to all the views that there are. These changes makes it much more easy for this code to reason about. We can see that exiting views are only added for the tree elements that are removed with an exit animation or their parent that are removed when the animation is running. The code path that cleans that up runs on the animation callback. So assuming the end callback always run after the exit animation has started at some point, the views will be cleared up from that map.

ios/LayoutReanimation/REAUIManager.mm
```diff
@@ -43,6 +50,11 @@ + (NSString *)moduleName
 
 - (void)setBridge:(RCTBridge *)bridge
 {
+  RCTLayoutAnimation *deletingAnimation = [[RCTLayoutAnimation alloc] initWithDuration:0.5 config:@{}];
+  _reactLayoutAnimationGroup = [[RCTLayoutAnimationGroup alloc] initWithCreatingLayoutAnimation:nil
+                                                                        updatingLayoutAnimation:nil
+                                                                        deletingLayoutAnimation:deletingAnimation
+                                                                                       callback:nil];
```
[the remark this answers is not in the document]
jwajgelt: Setting a `layoutAnimationGroup` with a `deletingLayoutAnimation` makes `RCTUIManager` call a different function when deleting views.
This allows us to only call our method for deleting views when there's a view with an `exiting` animation in the deleted subtree, by setting the `layoutAnimationGroup` to this `_reactLayoutAnimationGroup`.
The values here don't matter, it's just to set the `deletingLayoutAnimation` to _something_.
I added a comment in the code to hopefully explain what this property is for.

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -11,35 +11,117 @@ namespace reanimated {
 const long long idOffset = 1e9;
 
 LayoutAnimationsProxy::LayoutAnimationsProxy(
-    std::function<void(int, jsi::Object newProps)> _notifyAboutProgress,
-    std::function<void(int, bool)> _notifyAboutEnd)
-    : notifyAboutProgress(std::move(_notifyAboutProgress)),
-      notifyAboutEnd(std::move(_notifyAboutEnd)) {}
+    std::function<void(int, jsi::Object newProps)> progressHandler,
+    std::function<void(int, bool, bool)> endHandler,
+    std::weak_ptr<ErrorHandler> errorHandler)
```
tomekzaw, line 16:
```suggestion
    std::weak_ptr<ErrorHandler> weakErrorHandler)
```

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -11,35 +11,117 @@ namespace reanimated {
...
+    : progressHandler_(std::move(progressHandler)),
+      endHandler_(std::move(endHandler)),
+      weakErrorHandler_(errorHandler) {}
```
tomekzaw, line 19:
```suggestion
      weakErrorHandler_(weakErrorHandler) {}
```

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.h
```diff
@@ -1,32 +1,56 @@
...
+class ShareableValue;
 
 class LayoutAnimationsProxy {
  public:
   LayoutAnimationsProxy(
-      std::function<void(int, jsi::Object newProps)> _notifyAboutProgress,
-      std::function<void(int, bool)> _notifyAboutEnd);
+      std::function<void(int, jsi::Object newProps)> progressHandler,
+      std::function<void(int, bool, bool)> endHandler,
+      std::weak_ptr<ErrorHandler> errorHandler);
```
tomekzaw, line 24:
```suggestion
      std::weak_ptr<ErrorHandler> weakErrorHandler);
```

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.h
```diff
@@ -1,32 +1,56 @@
...
-  std::function<void(int, bool)> notifyAboutEnd;
-  std::map<int, std::shared_ptr<MutableValue>> observedValues;
+  std::function<void(int, jsi::Object newProps)> progressHandler_;
+  std::function<void(int, bool, bool)> endHandler_;
+  std::weak_ptr<ErrorHandler> weakErrorHandler_;
+  std::unordered_map<int, std::shared_ptr<MutableValue>> observedValues;
+  std::unordered_map<int, std::shared_ptr<ShareableValue>> viewSharedValues;
+  std::unordered_map<int, std::shared_ptr<ShareableValue>> enteringAnimations;
+  std::unordered_map<int, std::shared_ptr<ShareableValue>> exitingAnimations;
+  std::unordered_map<int, std::shared_ptr<ShareableValue>> layoutAnimations;
```
tomekzaw, line 50:
```suggestion
  std::unordered_map<int, std::shared_ptr<MutableValue>> observedValues_;
  std::unordered_map<int, std::shared_ptr<ShareableValue>> viewSharedValues_;
  std::unordered_map<int, std::shared_ptr<ShareableValue>> enteringAnimations_;
  std::unordered_map<int, std::shared_ptr<ShareableValue>> exitingAnimations_;
  std::unordered_map<int, std::shared_ptr<ShareableValue>> layoutAnimations_;
```

Common/cpp/NativeModules/NativeReanimatedModule.h
```diff
@@ -144,6 +144,15 @@ class NativeReanimatedModule : public NativeReanimatedModuleSpec,
...
       const jsi::Value &listenerId) override;
 
+  jsi::Value configureLayoutAnimation(
+      jsi::Runtime &rt,
+      const jsi::Value &viewTag,
+      const jsi::Value &type,
+      const jsi::Value &config,
+      const jsi::Value &viewSharedValue) override;
+
+  std::shared_ptr<LayoutAnimationsProxy> layoutAnimationsProxy_;
```
tomekzaw, line 154: Does this need to be public?

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -230,20 +230,64 @@ void NativeProxy::installJSIBindings(
...
+        }
+        jsi::Object yogaValues(*rt);
+        for (const auto &entry : *values) {
+          try {
+            auto key =
+                jsi::String::createFromAscii(*rt, entry.first->toStdString());
+            auto value = stod(entry.second->toStdString());
+            yogaValues.setProperty(*rt, key, value);
+          } catch (std::invalid_argument e) {
+            // failed to convert value to number
```
tomekzaw, line 268: Is it possible to use `errorHandler` here?
```suggestion
            errorHandler->setError('Failed to convert value to number');
            errorHandler->raise();
```
BTW What about setting value `0` or `null` for this key?

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -230,20 +230,64 @@ void NativeProxy::installJSIBindings(
...
+      [wrt, weakLayoutAnimationsProxy](
+          int tag,
+          alias_ref<JString> type,
+          alias_ref<JMap<jstring, jstring>> values) {
+        auto rt = wrt.lock();
+        auto layoutAnimationsProxy = weakLayoutAnimationsProxy.lock();
+        if (!rt || !layoutAnimationsProxy) {
+          return;
+        }
+        jsi::Object yogaValues(*rt);
```
tomekzaw, line 260: I know it's a detail, but in our codebase `rt` usually represents a reference to runtime (`jsi::Runtime &`), not a shared pointer (`std::shared_ptr<jsi::Runtime>`).

```suggestion
        auto runtime = wrt.lock();
        auto layoutAnimationsProxy = weakLayoutAnimationsProxy.lock();
        if (!runtime || !layoutAnimationsProxy) {
          return;
        }
        jsi::Runtime &rt = *runtime;
        jsi::Object yogaValues(rt);
```

ios/native/NativeProxy.mm
```diff
@@ -414,7 +378,32 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
...
   }];
+#else
+  // Layout Animation callbacks setup
+  [animationsManager setAnimationStartingBlock:^(
+                         NSNumber *_Nonnull tag, NSString *type, NSDictionary *_Nonnull values, NSNumber *depth) {
+    std::shared_ptr<jsi::Runtime> rt = wrt.lock();
+    if (rt == nullptr) {
+      return;
+    }
+    jsi::Object yogaValues(*rt);
```
tomekzaw, line 389: I'd prefer `jsi::Runtime &rt` here.

tomekzaw:
While testing out this PR I've encountered two issues:
1. Sometimes the app crashes while reloading due to failed assert in `SingleInstanceChecker` (maybe there's some reference loop in `REAModule`?)
2. Sometimes the following error appears (I'm able to replicate it by closing a screen before exiting animation ends):
error

ios/LayoutReanimation/REAUIManager.mm
```diff
@@ -64,6 +80,25 @@ - (void)setBridge:(RCTBridge *)bridge
   }
 }
 
+- (void)_removeChildren:(NSArray<UIView *> *)children
+          fromContainer:(UIView *)container
+          withAnimation:(RCTLayoutAnimationGroup *)animation
+{
+  if (_proxyViewRemoval) {
+    [_animationsManager removeChildren:children fromContainer:container];
+  } else if (animation != _reactLayoutAnimationGroup) {
```
tomekzaw, line 89: Can we add a comment here what this piece of code does (which problem it resolves) and why this is necessary?

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -230,19 +230,67 @@ void NativeProxy::installJSIBindings(
...
+      [wrt, weakLayoutAnimationsProxy, weakErrorHandler](
+          int tag,
+          alias_ref<JString> type,
+          alias_ref<JMap<jstring, jstring>> values) {
+        auto runtime = wrt.lock();
+        auto layoutAnimationsProxy = weakLayoutAnimationsProxy.lock();
+        if (!runtime || !layoutAnimationsProxy) {
+          return;
+        }
+        jsi::Object yogaValues(*runtime);
```
tomekzaw, line 261: I was thinking we can add `jsi::Runtime &rt = *runtime` and then simply use `rt` instead of `*runtime` everywhere just like everywhere else.

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -70,6 +70,9 @@ - (instancetype)initWithUIManager:(RCTUIManager *)uiManager
 
 - (void)invalidate
 {
+  for (NSNumber *tag in [[_exitingViews allKeys] copy]) {
+    [self endLayoutAnimnationForTag:tag cancelled:true removeView:true];
```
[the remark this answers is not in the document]
jwajgelt: 😭

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -512,16 +356,151 @@ public void updateLayout(
...
+      }
+    }
+  }
+
+  private void removeView(View view, ViewGroup parent) {
+    int tag = view.getId();
+    if (mCallbacks.containsKey(tag)) {
+      Runnable callback = mCallbacks.get(tag);
+      mCallbacks.remove(tag);
+      if (callback != null) {
```
piaskowyk, line 460: What we should to do if callback is null?

Common/cpp/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -11,35 +11,117 @@ namespace reanimated {
...
+    }
+    return;
+  }
+  jsi::Function startAnimationForTag =
+      layoutAnimationRepositoryAsValue.getObject(rt).getPropertyAsFunction(
+          rt, "startAnimationForTag");
+  startAnimationForTag.call(
+      rt,
+      jsi::Value(tag),
+      jsi::String::createFromAscii(rt, type),
```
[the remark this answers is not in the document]
tomekzaw: This should make no difference because the only strings passed here should be `"entering"`, `"layout"`, or `"exiting"`, but maybe let's use UTF-8 everywhere just to be on the safe side. Nice catch!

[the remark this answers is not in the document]
jwajgelt: ASCII is valid UTF8, so why not 😄 
Types are always ASCII-only anyway, so it shouldn't really matter here, and it won't be a problem once we move to enums here either.
I'd leave it as is.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
jwajgelt: Not run it? 🤷 
We don't insert nulls into the map, and we check the map contains key, so it shouldn't actually happen.

piaskowyk (approved):

tomekzaw (approved):
👏

kmagiera:
🚢 🇮🇹

ranaavneet:
Hey guys, Is this re-write available in 2.13.0 as well or it will only be available in version 3?

tomekzaw:
Hey, this PR was released in 3.0.0-rc.6 to get it thoughtfully tested on a smaller user base first. Ultimately we will backport these changes to 2.x as well.

shubhamdeol:
@tomekzaw 
Even with this Layout Animation rewrite [issue-2906](https://github.com/software-mansion/react-native-reanimated/issues/2906) not fixed. 
Do you have any idea what is required to fix this issue.

jaivinwylde:
Hey guys what's the progress of this PR? I see you're testing it in 3.0.0-rc.6, have all major features been implemented, and now just testing and bug fixing are left?

Also, does this PR add layout animations for the web?

tomekzaw:
> Hey guys what's the progress of this PR? I see you're testing it in 3.0.0-rc.6, have all major features been implemented, and now just testing and bug fixing are left?

Exactly, it was merged and released in 3.0.0-rc.6. We plan to release 3.0.0-stable soon, however most likely we won't backport this rewrite into 2.x branch.

> Also, does this PR add layout animations for the web?

Nope, it doesn't, but it's still on our roadmap for 2023.

sannajammeh:
Just going to mention that the docs are extremely lacking in usage of this. It still links to the V2 usage video as well.

tomekzaw:
@sannajammeh Check out new Reanimated docs (currently in beta): https://reanimated-beta-docs.swmansion.com/docs/category/layout-animations

PR #3354, opened 2022-07-04
Add inline styles

graszka22:
## Description

This PR adds support for animations in inline styles.

## Changes

- Added `withInlineStyles` HOC and applied it to component returned by `createAnimatedComponent`
- Added some data to animations created by `defineAnimation`. It's needed to call animations in `useAnimatedStyle` within  `withInlineStyles`
- Updated `withTiming`, `withSpring`, `withDecay` types
- Moved `flattenArray` to utils
- Moved `AnimatedComponentProps` to a separate file to avoid circular dep

```js
const width = useSharedValue(0);


<Animated.View style={{ width: withSpring(width) }} />
<Animated.View style={{ width: width }} />
```

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


graszka22:
I added the warning about using .value in inline styles to the babel plugin. It doesn't work in some cases like:
- it doesn't show a warning for a code like this:
```js
const style = { width: sharedValue.value };
```

- it shows a warning for a code like this:
```js
const Colors = { plus: '#0f0', minus: '#f00', value: '#ddd' };
<Animated.View style={{ color: Colors.value }} />
```

but you can't make them work with a babel plugin. 

I also added some unit tests for inline styles and the plugin. But there is a problem with unit tests for inline styles. In jest environment `_WORKLET` global is not set because it is set from cpp and in jest there is no cpp run. Inline styles rely on this variable (so if we run `withTiming` out of worklet, it returns a function not an object). So I set it myself in `jestStyleUpdater` and it works for now.

__tests__/plugin.test.js
```diff
@@ -559,4 +559,59 @@ describe('babel plugin', () => {
     const { code } = runPlugin(input);
     expect(code).toMatchSnapshot();
   });
+
+  it('shows a warning if user uses .value inside inline style, no functions', () => {
+    const input = `
+    function App(){
```
tomekzaw, line 565:
```suggestion
    function App (){
```
These code snippets need formatting, please throw them into Prettier.

docs/docs/fundamentals/animations.md
```diff
@@ -96,6 +96,50 @@ This approach is more convenient in many cases, especially when view properties
...
+
+For simple animations in `useAnimatedStyle` hook without any calculations like this:
+
+```js
+const animatedStyle = useAnimatedStyle(() => {
+  return {
+    width: withSpring(width.value),
+  };
+});
+<Animated.View style={animatedStyle} />
```
tomekzaw, line 109:
```suggestion
function SomeComponent() {
  const animatedStyle = useAnimatedStyle(() => {
    return {
      width: withSpring(width.value),
    };
  });

  return <Animated.View style={animatedStyle} />;
}
```

docs/docs/fundamentals/animations.md
```diff
@@ -96,6 +96,50 @@ This approach is more convenient in many cases, especially when view properties
...
+```js
+<Animated.View style={{ width: withSpring(width) }} />
+```
+
+You can also use shared value directly in inline style:
+
+```js
+<Animated.View style={{ width: width }} />
+```
+
```
tomekzaw, line 123: Is this syntax also supported?
```js
<Animated.View style={{ width }} />
```

docs/docs/fundamentals/animations.md
```diff
@@ -96,6 +96,50 @@ This approach is more convenient in many cases, especially when view properties
...
+The width of the view will change when `width` value changes.
+
+Note that we're not using `.value` getter in inline styles. If you use `width.value`, reanimated uses just current value of shared value. Let's look at an example:
+
+```js
+width.value = 5;
+...
```
tomekzaw, line 130:
```suggestion
// ...
```

docs/docs/fundamentals/animations.md
```diff
@@ -96,6 +96,50 @@ This approach is more convenient in many cases, especially when view properties
...
+<Animated.View style={{ width: width.value }} />
+```
+
+is equivalent to:
+
+```js
+<Animated.View style={{ width: 5 }} />
+```
+
+Passing a shared value without `.value` getter causes reanimated to watch for shared value changes.
```
tomekzaw, line 140:
```suggestion
Passing a shared value without `.value` getter causes Reanimated to watch for shared value changes.
```

plugin.js
```diff
@@ -852,6 +852,74 @@ function processWorklets(t, path, state) {
   }
 }
 
+function generateInlineStylesWarning(t, memberExpression) {
+  // change sharedvalue.value -> ((){console.warn("warning");return sharedvalue.value;})()
```
tomekzaw, line 856:
```suggestion
  // replaces `sharedvalue.value` with `((){console.warn("warning");return sharedvalue.value;})()`
```

src/reanimated2/animation/decay.ts
```diff
@@ -44,7 +44,14 @@ export function withDecay(
 ): Animation<DecayAnimation> {
   'worklet';
 
-  return defineAnimation<DecayAnimation>(0, () => {
+  const animationFunctionCall = {
+    functionName: 'withDecay',
+    // eslint-disable-next-line prefer-rest-params
+    functionArguments: [...arguments],
```
tomekzaw, line 50: Can you please add comment why we can't just pass `arguments` and we need to make a copy?

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -445,6 +450,22 @@ export function useAnimatedStyle<T extends AnimatedStyle>(
 
   dependencies.push(sharableViewDescriptors);
 
+  const areListsEqual = (a1: DependencyList, a2: DependencyList) => {
+    for (let i = 0; i < (a1?.length || 0); ++i) {
+      if (a1?.[i] !== a2?.[i]) return false;
+    }
+    return true;
+  };
```
tomekzaw, line 458: Can we define this function in top-level and also use `function areListsEqual(...) {}` syntax?

src/reanimated2/animation/sequence.ts
```diff
@@ -11,8 +11,18 @@ export function withSequence(
   ..._animations: NextAnimation<AnimationObject>[]
 ): Animation<SequenceAnimation> {
   'worklet';
+
+  const animationFunctionCall = {
+    functionName: 'withSequence',
+    // eslint-disable-next-line prefer-rest-params
+    functionArguments: [...arguments],
+    // eslint-disable-next-line prefer-rest-params
+    animatedArgumentsIndices: [...arguments].map((_, i) => i),
```
tomekzaw, line 20: Not sure if we need to copy `arguments` in this case since we call `map` afterwards.

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -445,6 +450,22 @@ export function useAnimatedStyle<T extends AnimatedStyle>(
...
+
+  // dependencies can change in length and React doesn't like that
+  // so in useEffect instead of passing just a list of dependencies we pass the list of dependencies itself as a dependency
```
kmagiera, line 461: This doesn't sound like a correct approach. Passing a list likely means there will never be referential equality (each list is perhaps compiled during render), so this will result in this hook being run at every render which misses the point of passing any dependencies in the first place

kmagiera (changes requested):
Left some comments inline, but most importantly I'd like to understand why changes in animation functions are needed.

__tests__/inlineStyles.test.js
```diff
@@ -0,0 +1,234 @@
...
+    const view = getByTestId('view');
+    const button = getByTestId('button');
+
+    expect(view).toHaveAnimatedStyle({ width: 100 });
+
+    fireEvent.press(button);
+    await waitFor(() => {
+      expect(view).toHaveAnimatedStyle({ width: 200 });
+    });
+    jest.useFakeTimers();
```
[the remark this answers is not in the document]
graszka22: Yes, because in `afterEach` callback we call `jest.runOnlyPendingTimers();` and it's going to show a long annoying warning. And `afterEach` is needed for all other tests that rely on fake timers. And this test needs real timers.

src/reanimated2/animation/decay.ts
graszka22: In all 6 places? 😅 
The problem with is that `arguments` is not an array (and typescript also agrees that it's not an array) and we need an array:
```js
> function foo() {
... arguments.map(a => console.log(a))
... }
undefined
> foo()
Uncaught TypeError: arguments.map is not a function
    at foo (REPL23:2:11)
```

src/reanimated2/animation/sequence.ts
graszka22: As I wrote above `arguments` is not an array and it has no `map` function.

src/createAnimatedComponent.tsx
```diff
@@ -528,9 +473,11 @@ export default function createAnimatedComponent(
     Component.displayName || Component.name || 'Component'
   })`;
 
+  const AnimatedComponentWithInlineStyles = withInlineStyles(AnimatedComponent);
```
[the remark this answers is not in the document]
graszka22: I think that's not the case. `createAnimatedComponent` is a function that is called once at the top level. It is:
- creating the inner `AnimatedComponent` class
- changing it's display name
- creating `AnimatedComponentWithInlineStyles ` class by wrapping the `AnimatedComponent`
- calling `React.forwardRef`
There aren't any component classes created in the render methods - all of them are created in `createAnimatedComponent` function.

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -272,7 +272,10 @@ function jestStyleUpdater(
 ): void {
   'worklet';
   const animations: AnimatedStyle = state.animations ?? {};
+  // in jest _WORKLET is never set, but we rely on it in inline styles
+  global._WORKLET = true;
```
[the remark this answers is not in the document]
graszka22: What undesirable behavior?
- `jestStyleUpdater` is called only when using jest - it does not affect anything other than jest testing
- style updater is supposed to be called on the main thread runtime, so it is supposed to have `_WORKLET` set to `true` in normal environment
- in jest it is not set because jest is only running js - there is no main thread runtime and native stuff at all
- so setting `_WORKLET` to true here is only simulating what would happen in a normal environment

And setting `_WORKLET` to true is necessary for inline styles to even work at all - `defineAnimation` works differently when `_WORKLET = true` (returns animation, instead of a factory).
 
But I do agree that this is a hack. We should be testing code that is run in normal environment, not replacing it with the code that is simulating for the testing environment.

src/reanimated2/hook/useAnimatedStyle.ts
graszka22: A few lines below there is a check:
```js
if (
    dependencies.length !== depsRef.current[0]?.length ||
    !areListsEqual(dependencies, depsRef.current)
  ) {
    depsRef.current = [dependencies];
  }
```
So if the length of dependencies changes or the dependencies itself change then we create a new list and assign it to `depsRef.current`. Then we pass `depsRef.current` as a hook dependency. If nothing changes in dependencies, we don't touch `depsRef.current` so it's the same list and hook is not run.

graszka22:
> I'd like to understand why changes in animation functions are needed
The basic idea of inline styles is that under the hood we have just `useAnimatedStyle` that is managed by us and we call the animation functions there with shared values normally.

Supposing we have something like that:
```js
style={{width: withTIming(sv)}}
```
the `withTiming` function is called here. It must return something, so that later we know that we're supposed to run the withTiming function in the internal `useAnimatedStyle` hook with `sv.value` as an argument.
Before inline styles the `withTiming` function would return a `create` function that returns animation object. It's hard to figure out what animation function should we call and with what arguments just from the animation object. It's even more complicated when animations functions are nested (`withRepeat(withTiming(...), ..)`). Also imagine we're adding another animation function. We would need to remember to update inline styles logic and that sounds bad to me.
So I extended it in `defineAnimation` with additional info. It shouldn't break anything, but I agree that's annoying. I couldn't find another solution though. Plus adding a new animation function doesn't require changing anything in inline styles, it should just work automatically.
And why do we need to call the animation functions inside `useAnimatedStyle` instead of using create functions returned already? In create functions we have `sharedValue` as `toValue` instead of `sharedValue.value` and that will probably crash.

docs/docs/fundamentals/animations.md
graszka22: Yes it is.

plugin.js
```diff
@@ -852,6 +852,74 @@ function processWorklets(t, path, state) {
...
+            [
+              t.stringLiteral(
+                "It looks like you might be using shared value's .value inside reanimated inline style. " +
+                  'This is probably not what you want to do. See documentation here: ' +
+                  'https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/animations#animations-in-inline-styles'
+              ),
+            ]
+          )
+        ),
+        t.returnStatement(memberExpression.node),
```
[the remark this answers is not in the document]
graszka22: No, we don't have one. I agree that there are legit use cases, I even described them above. I have no obvious idea though for a way to disable the warning.
- one idea would be to disable this warning globally by using option in the babel plugin. User would need to put something like this in his babel.rc:
```js
plugins: [
      ['react-native-reanimated/plugin', {"showInlineStylesWarning": false}],
],
```
- second idea would be a comment that allows to disable warning in specific places, something like this:
```js
<Animated.View style={{ /* @rea-ignore */ color: Colors.value }} />
```

__tests__/inlineStyles.test.js
```diff
@@ -0,0 +1,234 @@
+import React from 'react';
```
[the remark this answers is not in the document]
graszka22: I agree that it would be better. At first I thought that would be easy but typescript declarations in `react-native-reanimated.d.ts` and in reanimated code (`timing.ts`, `utils.ts`, `repeat.ts` and so on) are different. Tests use declarations from the code and in my opinion the types in the code don't make sense. Like why repeat animation takes repeat animation as an argument? It can be any animation.
```ts
export function withRepeat(
  _nextAnimation: NextAnimation<RepeatAnimation>,
  numberOfReps = 2,
  reverse = false,
  callback?: AnimationCallback
): Animation<RepeatAnimation>
```
I tried to update those types but it goes deeper and deeper. You have this monstrosity in `utils.ts`:
```ts
type AnimationToDecoration<T extends AnimationObject | StyleLayoutAnimation> =
  T extends StyleLayoutAnimation
    ? Record<string, unknown>
    : T extends DelayAnimation
    ? NextAnimation<DelayAnimation>
    : T extends RepeatAnimation
    ? NextAnimation<RepeatAnimation>
    : T extends SequenceAnimation
    ? NextAnimation<SequenceAnimation>
    : AnimatableValue | T;

export function defineAnimation<
  T extends AnimationObject | StyleLayoutAnimation
>(
  starting: AnimationToDecoration<T>,
  animationFunctionCall: AnimationFunctionCall,
  factory: () => T
): T {
```
`starting` doesn't have to be an animation, it may be `AnimatableValue`. `defineAnimation` can return animation, animation factory or a value depending on where is it called and it's already `// @ts-ignore`'d
Correcting those types should be done in separate PR, this one is already large and those types were wrong before inline styles.

docs/docs/fundamentals/animations.md
graszka22: After reading through the docs the `useAnimatedStyle` is introduced and explained in `shared-values.md` section. So I'd need to rewrite a large part of `shared-values.md` and `animations.md` pages. But we're going to rewrite all the docs soon anyway with different sections, right? Is it worth it rewriting it now if it's going to be rewritten soon anyway?

plugin.js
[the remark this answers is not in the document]
kmagiera: I'd go with the simplest approach (one that adds less complexity to this change)

src/withInlineStyles.tsx
```diff
@@ -0,0 +1,135 @@
...
+          animatedArgumentsIndices.includes(i) ? parseTree(arg) : arg
+        );
+      }
+    };
+
+    for (const key in functionsStyles) {
+      parseTree(functionsStyles[key]);
+    }
+
+    const animatedStyle = useAnimatedStyle(updater);
```
kmagiera, line 118: does this mean we add additional overhead to all animated components regardless of whether they have inline styles? Maybe instead of basing this on useAnimatedStyle we can use `startMapper`/`stopMapper` directly thus eliminating the need for a hook call. This way we can get rid of additional element in the component hierarchy as well as calling this hook with empty updater

kmagiera (commented):
I agree with most of the responses you posted and left a few more comments. I think I now have a better understanding of this change and also have one more idea how this can be approached differently that wanted to run by you

piaskowyk:
We will continue this feature in another PR.

PR #3380, opened 2022-07-12
Implement `withSpring` using `CASpringAnimation` from CoreAnimation

tomekzaw:
## Description

This PR provides an alternative implementation of `withSpring` animation that utilizes [`CASpringAnimation`](https://developer.apple.com/documentation/quartzcore/caspringanimation) from [CoreAnimation](https://developer.apple.com/documentation/quartzcore?language=objc) instead of calculating the position and velocity in [pure JS](https://github.com/software-mansion/react-native-reanimated/blob/e7947a98e230b6511f50360dbc148ae130e5f51f/src/reanimated2/animation/spring.ts#L61-L138).

With this approach we can not only offload the UI thread but also estimate duration required for the spring system to be considered at rest (see [`settlingDuration`](https://developer.apple.com/documentation/quartzcore/caspringanimation/1412524-settlingduration?language=objc)) and precalculate parameters for subsequent frames.

Currently, this PR covers hardware acceleration of `withSpring` animation only on iOS. We are still researching viable solutions for Android. Moving animation-related computations from pure JS to C++ possibly with frame precalculation is also considered.

**Demo:**

https://user-images.githubusercontent.com/20516055/178500830-ad164b5c-ee8c-4871-bb83-8f8ba05f07fb.mov

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/Common/cpp/NativeModules/NativeReanimatedModule.cpp
+++ b/Common/cpp/NativeModules/NativeReanimatedModule.cpp
@@ -150,6 +150,7 @@ NativeReanimatedModule::NativeReanimatedModule(
       requestAnimationFrame,
       platformDepMethodsHolder.getCurrentTime,
       platformDepMethodsHolder.createSpringAnimation,
+      platformDepMethodsHolder.createKeyframeAnimation,
       platformDepMethodsHolder.registerSensor,
       platformDepMethodsHolder.unregisterSensor,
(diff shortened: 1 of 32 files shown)


tomekzaw:
Closing this draft PR in favor of #4179 which is almost complete.

PR #3384, opened 2022-07-12
Exclude meta-inf from gradle to prevent intermediate compile conflicts

shamilovtim:
When trying to compile Detox and Reanimated from source there is an intermediate dependency conflict stemming from the META-INF folder. To resolve, excluded META-INF from Gradle packagingOptions. 

Please cherrypick onto v2

## Changes

- Exclude meta-inf from gradle to prevent intermediate compile conflicts with Detox

## Checklist

- [] Included code example that can be used to test this change
- [] Updated TS types
- [] Added TS types tests
- [] Added unit / integration tests
- [] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -256,6 +256,8 @@ android {
         println "Native libs debug enabled: ${debugNativeLibraries}"
         doNotStrip debugNativeLibraries ? "**/**/*.so" : ''
         excludes = [
+                "META-INF",
+                "META-INF/**",
                 "**/libc++_shared.so",
                 "**/libfbjni.so",
                 "**/libjsi.so",


piaskowyk:
Could you tell me more about your issue? I am not convinced that the source of the problem is Reanimated because `react-native` contains the same `META-INF` catalog inside `aar` file.

shamilovtim:
> Could you tell me more about your issue? I am not convinced that the source of the problem is Reanimated because `react-native` contains the same `META-INF` catalog inside `aar` file.

I confirmed reanimated was causing the conflict by implementing this patch locally. After patching I am able to compile test builds. 

I am experiencing the issue from both reanimated and detox when compiling them locally from source using the test workflow.

henriquechagasg (approved):

shamilovtim:
Looks like this PR has been proven out. Are you guys going to merge? @tomekzaw

tomekzaw:
Hey @shamilovtim, thanks a lot for the PR!

We are wondering if there is any other solution than excluding whole `META-INF` directory.

Also, could you please point out which commit in react-native-reanimated between 2.9.1 and 2.10.0 could be the root cause of this issue?

shamilovtim:
> Hey @shamilovtim, thanks a lot for the PR!
> 
> We are wondering if there is any other solution than excluding whole `META-INF` directory.
> 
> Also, could you please point out which commit in react-native-reanimated between 2.9.1 and 2.10.0 could be the root cause of this issue?

This has been happening for our project as soon as early versions of reanimated 2 so I don't think 2.9.1 is responsible. If the META-INF exclusion caused something bad, the project would simply not compile, correct? Seems like a pretty low risk solution.

tybeck:
This PR has also worked for me.

katanasushi:
also worked for me

kmagiera (approved):

PR #3405, opened 2022-07-22
docs: update scripts on example

gkartalis:
## Description 

Tried to run the example and didn't need to install rn cli so I updated the docs to match the steps I took.

The scripts are run via npx. 

Happy to hear your thoughts!

## Changes

- [x] Updated example scripts in the docs to use `npx react-native` over having react native cli globally installed

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [x] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/Example/README.md
+++ b/Example/README.md
@@ -7,14 +7,14 @@
 
 ### Running on iOS
 
-- `cd ios && pod install && cd ..` - to install pods
-- `react-native run-ios` - to run the app
+- `npx pod-install` - to install pods
+- `npx react-native run-ios` - to run the app
 
 ### Running on Android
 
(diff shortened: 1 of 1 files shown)


piaskowyk:
Hey! I agree that `react-native-cli` is unnecessary, but instead of `npx` we prefer `yarn`, Is it possible for you to replace `npx` with the `yarn` command?

gkartalis:
@piaskowyk but it doesn't work with `yarn`! `npx` is different than npm and yarn! 

`Npm` and `yarn` are tools that we use to install packages vs `npx` being a package used to **execute** packages without the need for them to be installed.

So the second updated command `yarn react-native run-ios` will not work now🤔 , the first one will work since you updated the package.json

PR #3418, opened 2022-07-28
Ensure global objects initialization for web

### LINKED ISSUE #3355: [Web]ReferenceError: _frameTimestamp is not defined
## Description
Get this error when upgrading `react-native-reanimated` to `v2.9.0+` on the web (`v2.8.0` works), it works fine on iOS and Android.

image

### Expected behavior 

Works on the web.

web not working when upgrading to `v2.9.0+`.

https://github.com/tamagui/starters/tree/main/next-expo-solito

## Package versions

| react-native                         |    0.68.2     |
| react-native-reanimated              |    2.9.1     |
| NodeJS                               |     16.15.1    |
| Xcode                                |     13.4.1      |
| Java                                 |         |
| Gradle                               |         |
| expo                                 |     45.0.2    |

## Affected platforms

- [ ] Android
- [ ] iOS
- [x] Web

piaskowyk:
## Description

The PR with tree shaking moved global objects initialization for web to another file and then imported it into the `index` file (`import './reanimated2/js-reanimated/global';`). However, the Webpack can lazy import modules, but we don't have any call to this module, so global objects were never initialized.

Fixes #3355

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,7 +6,7 @@ if (!webGlobalIsInitialized) {
     but we need to use `webGlobalIsInitialized` somewhere to ensure function execution, 
     in another way, the bundler can remove unused variables. 
   */
-  console.error('[Reanimated] Unable to initialize global objects for WEB.');
+  console.error('[Reanimated] Unable to initialize global objects for web.');
 }
 
 export * from './reanimated2';
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,13 @@
 // tree-shaken side effects
-import './reanimated2/js-reanimated/global';
+import webGlobalIsInitialized from './reanimated2/js-reanimated/global';
+if (!webGlobalIsInitialized) {
+  /* 
+    `webGlobalIsInitialized` should always be `true`, 
+    but we need to use `webGlobalIsInitialized` somewhere to ensure function execution, 
+    in another way, the bundler can remove unused variables. 
+  */
+  console.error('[Reanimated] Unable to initialize global objects for WEB.');
+}
 
 export * from './reanimated2';
 export * as default from './Animated';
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -1,5 +1,13 @@
 // tree-shaken side effects
-import './reanimated2/js-reanimated/global';
+import webGlobalIsInitialized from './reanimated2/js-reanimated/global';
+if (!webGlobalIsInitialized) {
+  /* 
+    `webGlobalIsInitialized` should always be `true`, 
+    but we need to use `webGlobalIsInitialized` somewhere to ensure function execution, 
+    in another way, the bundler can remove unused variables. 
+  */
+  console.error('[Reanimated] Unable to initialize global objects for WEB.');
```
tomekzaw, line 9:
```suggestion
  console.error('[Reanimated] Unable to initialize global objects for Web.');
```

src/reanimated2/js-reanimated/global.ts
```diff
@@ -1,38 +1,40 @@
...
-      "[Reanimated] You can't use `measure` with Chrome Debugger or with web version"
-    );
-    return {
-      x: 0,
-      y: 0,
-      width: 0,
-      height: 0,
-      pageX: 0,
-      pageY: 0,
+export default (() => {
```
tomekzaw, line 4: Can we name this lambda `initializeGlobalsForWeb` and call it as
```ts
export default initializeGlobalsForWeb();
```
? Also, a comment here why we need to do it would be valuable.

tomekzaw (changes requested):

tomekzaw (approved):

jer-sen:
@piaskowyk it does not solve anything since you did not applied the fix to `index.web.ts` which is the file executed on Web...

piaskowyk:
The `index.web.ts` file was added after this PR, and we rewrite this part and got rid of those changes so this PR is no longer related.

jer-sen:
I'm with react-native-reanimated@2.14.4 (last version for Expo SDK 48) and there is the PR's code in `index.ts` but not in the existing `index.web.ts`...

oleg3505:
Hello I'm also working with react-native-reanimated@2.14.4 (Expo SDK 48) and the error with `_frameTimestamp is not defined` is still here for web

piaskowyk:
Did you try Reanimated3?

PR #3428, opened 2022-08-01
Fix URL in Docusaurus config

tomekzaw:
## Description

This PR eliminates the following warning when running Docusaurus:
```
[WARNING] Docusaurus config validation warning. Field "url": the url is not supposed to contain a sub-path like '/react-native-reanimated/', please use the baseUrl field for sub-paths
```
by removing `/react-native-reanimated/` from `url` field in `docusaurus.config.js`.

> URL for your website. This can also be considered the top-level hostname. For example, `https://facebook.github.io` is the URL of https://facebook.github.io/metro/, and `https://docusaurus.io` is the URL for https://docusaurus.io/ This field is related to the [`baseUrl`](https://docusaurus.io/docs/api/docusaurus-config#baseUrl) field.

https://docusaurus.io/docs/api/docusaurus-config#url

> Base URL for your site. Can be considered as the path after the host. For example, `/metro/` is the base URL of https://facebook.github.io/metro/. For URLs that have no path, the baseUrl should be set to `/`. This field is related to the [`url`](https://docusaurus.io/docs/api/docusaurus-config#url) field. Always has both leading and trailing slash.

https://docusaurus.io/docs/api/docusaurus-config#baseUrl

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -70,7 +70,7 @@ In swizzled components look for "SWM -" string to see our modifications
 module.exports = {
   title: 'React Native Reanimated',
   tagline: "React Native's Animated library reimplemented",
-  url: 'https://docs.swmansion.com/react-native-reanimated/',
+  url: 'https://docs.swmansion.com',
   baseUrl: '/react-native-reanimated/',
   onBrokenLinks: 'throw',
   onBrokenMarkdownLinks: 'warn',


piaskowyk (approved):
👍

Issue #3333, opened 2022-06-28
TypeError: _ReanimatedModule.default.configureProps is not a function


nialldbarber:
## Description
- Started this [as a discussion](https://github.com/software-mansion/react-native-reanimated/discussions/3309) but got no hits and it could be bug (certainly something I can't figure out). 
- I am in the process of updating our app, updating Reanimated from `1.9.0` to `2.8.0`, which also includes various updates to React Native `0.67`, React Navigation 6 etc, so there could be issues there. 
- Every time I try to build the project, I encounter this error:

```
TypeError: _ReanimatedModule.default.configureProps is not a function
```

![Simulator Screen Shot - iPhone 13 - 2022-06-27 at 15 29 43](https://user-images.githubusercontent.com/20539827/176252257-72af4c9c-cffd-4853-a184-e3eb225694a1.png)

- The file in question looks to be in `NavigationOptions.js`, which for reference looks like:

```jsx
export const navigationOptions = (options: Object): Object => ({
  navigation,
}: {
  navigation: NavigationStateType,
}) => ({
  ...options,
})

export const navigationOptionsWithTitle = (options: Object): Object => {
  const initialObject = { ...options }

  return ({ route }: {route: NavigationRouteType}) => ({
    ...initialObject,
    ...titleOptions,
    title: route?.params?.title || options?.title || ''
  })
}

export const navigationOptionsWithDrawer = (options: Object): Object => ({
  navigation,
}: {
  navigation: NavigationStateType,
}) => ({
  headerRight: () => <DrawerButton navigation={navigation} />,
  ...options,
})

export const navigationOptionsWithComponentTitle = (options: Object): Object => ({
  navigation,
}: {
  navigation: NavigationStateType,
}) => ({
  headerTitle: () => <NavigationHeaderTitle navigation={navigation} />,
  ...options,
})
```

### Expected behaviour
- Updating Reanimated should ideally not cause any errors. 

### Actual behaviour & steps to reproduce
- This is a tough one as I'm not able to reproduce it. If someone has seen a similar issue then that would be helpful. 
- I'm thinking there could be some issue with our configuration of React Navigation, which for some reason isn't working on Reanimated v2. 

## Snack or minimal code example
- Unable to replicate this in a snack repo :(((( 

## Package versions
| name                                 | version |
| ------------------------------------ | ------- |
| react-native                         | `0.67.4` |
| react-native-reanimated              | `2.8.0` |
[...]

PR #3441, opened 2022-08-05
Detect two versions of Reanimated

piaskowyk:
## Description

When a project exists with more than one instance of Reanimated, two versions of Reanimated JS module try to use the same Native Module. It can occur when some of your dependency has installed Reanimated inside their own `node_modules` instead of using it as a peer dependency. The problem is, for example, when the old version of JS tries to call a removed function in a newer version of the native module or can call twice initialization of the native module and override existing data. The solution to this issue is to modify the `package.json` file, requiring only one reanimated version. [Details in the documentation.](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/troubleshooting#multiple-versions-of-reanimated-were-detected)

Fixes https://github.com/software-mansion/react-native-reanimated/issues/3333

## Solution
You can check which libraries are using Reanimated, for example, with the command:
```bash
npm why react-native-reanimated
``` 

If you use `yarn` you should add [`resolution` property](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/).
```json
"resolutions": {
  "react-native-reanimated": <Reanimated version>
}
```

If you use `npm` you should add [`overrides` property](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides).
```json
"overrides": {
  "react-native-reanimated": <Reanimated version>
}
```

After that you need to run you package manager again

```bash
yarn
```
or
```bash
npm install
```

### DIFF
--- a/docs/docs/fundamentals/troubleshooting.md
+++ b/docs/docs/fundamentals/troubleshooting.md
@@ -32,7 +32,7 @@ sure you have added it.
 
 ### Multpile versions of Reanimated was detected
 
-This error usually happens when in your project exists more than one instance of Reanimated. It can occur when some of your dependency has installed Reanimated inside your `node_modules` instead of using it as a peer dependency. In this situation cause that two different versions of Reanimated JS module try to use the same Native Module. You can resolve this problem manually by modifying your `package.json` file.
+This error usually happens when in your project exists more than one instance of Reanimated. It can occur when some of your dependency has installed Reanimated inside their own `node_modules` instead of using it as a peer dependency. In this situation cause that two different versions of Reanimated JS module try to use the same Native Module. You can resolve this problem manually by modifying your `package.json` file.
 
 You can check which libraries are using Reanimated, for example, with the command:
 ```bash
--- a/docs/docs/fundamentals/troubleshooting.md
+++ b/docs/docs/fundamentals/troubleshooting.md
@@ -29,3 +29,31 @@ Depending on how spread is used you may try one of the following alternatives:
 
 This error usually happens, when you forget to add the babel plugin in your `babel.plugin.js`. Please make
 sure you have added it.
+
+### Multpile versions of Reanimated was detected
+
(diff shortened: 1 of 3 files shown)


docs/docs/fundamentals/troubleshooting.md
```diff
@@ -28,4 +28,37 @@ Depending on how spread is used you may try one of the following alternatives:
...
+
+If you use `npm` you should add [`overrides` property](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides).
+```json
+"overrides": {
+  "react-native-reanimated": <Reanimated version>
+}
+```
+
+After that you need to run you package manager again
+
```
tomekzaw, line 57:
```suggestion

```

docs/docs/fundamentals/troubleshooting.md
```diff
@@ -28,4 +28,37 @@ Depending on how spread is used you may try one of the following alternatives:
...
+sure you have added it.
+
+### Multpile versions of Reanimated was detected
+
+This error usually happens when in your project exists more than one instance of Reanimated. It can occur when some of your dependency has installed Reanimated inside their own `node_modules` instead of using it as a peer dependency. In this situation cause that two different versions of Reanimated JS module try to use the same Native Module. You can resolve this problem manually by modifying your `package.json` file.
+
+You can check which libraries are using Reanimated, for example, with the command:
+```bash
+npm why react-native-reanimated
+```
```
tomekzaw, line 40:
```suggestion
* `yarn why react-native-reanimated`
* `npm why react-native-reanimated`
```

tomekzaw (changes requested):

docs/docs/fundamentals/troubleshooting.md
piaskowyk: I think that `npm why` is more informative than `yarn why`:
image

tomekzaw (approved):
🤯 👏

PR #3442, opened 2022-08-05
fix: missed FlatList component in mocks

stelmakhivan:
Added missed FlatList component in mocks file


piaskowyk (approved):
Thanks for your PR! 🙌

PR #3448, opened 2022-08-08
Cache iOS build artifacts in GitHub Actions

tomekzaw:
## Description

This PR significantly shortens iOS app build times on CI by caching `node_modules`, `Pods` and `DerivedData` directories from ~25 minutes for Example app and ~45 minutes for FabricExample app down to ~8 and ~12 minutes, respectively.

Additionally, on `macos-latest` (alias of `macos-11`), the error "kLSNoExecutableErr: The executable is missing" appears in approximately 80% of the builds. The error disappears when using `macos-12`. For more details, check out [this](https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md) document.

**Before:**

before

**After:**

after

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
```diff
@@ -493,7 +493,7 @@
 					"-ObjC",
 					"-lc++",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = org.reactjs.native.example.mleko.ReanimatedExample;
```
WoLewicki, line 496: @piaskowyk ❤️

piaskowyk: I was there 🥛

.github/workflows/ios-build.yml
```diff
@@ -2,38 +2,78 @@ name: Test iOS build
...
+      group: ios-${{ matrix.working-directory }}-${{ github.ref }}
     steps:
-      - name: checkout
+      - name: Check out Git repository
         uses: actions/checkout@v2
-      - name: Use Node.js 14
-        uses: actions/setup-node@v2
+
+      - name: Restore Reanimated node_modules from cache
+        uses: actions/cache@v2
```
piaskowyk, line 28: Is there a reason why we use `cache@v2` instead of `cache@v3`?

.github/workflows/ios-build.yml
```diff
@@ -2,38 +2,78 @@ name: Test iOS build
...
+        id: cache-node-modules-reanimated
+        with:
+          path: |
+            node_modules
+          key: ${{ runner.os }}-node-modules-reanimated-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-node-modules-reanimated
+
+      - name: Install Reanimated node_modules
+        run: yarn install --frozen-lockfile
```
piaskowyk, line 38: Why do you use `--frozen-lockfile` option? Does this mean CI will [fail](https://classic.yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile) when we change some dependency in `package.json`?

.github/workflows/ios-build.yml
tomekzaw: Totally not, I will change it to `cache@v3`. Good catch!

.github/workflows/ios-build.yml
tomekzaw: I've seen this flag being used in some other CI job in other repository. Assume that we made some changes in `package.json` but forgot to commit `yarn.lock`. Now, when CI installs node_modules, it will generate (update) `yarn.lock` in the CI runner filesystem. This flag means that `yarn install` will fail in such case. For my understanding, it can be used to detect differences between `package.json` and `yarn.lock`.

Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
tomekzaw: 🐮

.github/workflows/ios-build.yml
piaskowyk: Ok, I see it now, thanks!

WoLewicki (approved):
LGTM :tada:

piaskowyk (approved):
🚀

PR #3414, opened 2022-07-27
Add warning when measuring flattened view on Android

tomekzaw:
## Description

Resolves #3188. See https://github.com/software-mansion/react-native-reanimated/issues/3188#issuecomment-1196795764 for details.

## Changes

- Return array of NaNs instead of zero-length array when Android view is not found in `measure`
- Add console.warn with user-friendly message when `measure` returns NaNs

## Test code and steps to reproduce

Copy `App.tsx` and `components/Ripple.tsx` from https://snack.expo.dev/@boxedition/ripple-measure-random-values

## Checklist

- [ ] Included code example that can be used to test this change
- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Added unit / integration tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

PR #3453, opened 2022-08-08
Refactor `measure()` to handle `try` internally

naftalibeder:
## Description

Allows `measure()` to handle exceptions internally, and return `null` if the measurement cannot be made.

The impetus for this is occasional hard crashes when calling `measure`, seemingly due to a view not yet having loaded, but not fixable in the enclosing app with a null check on the ref, a `try` block, etc. - once the JS passes a value to the native method `_measure`, an invalid argument will cause a crash.

This way, the caller:
- is guaranteed that calling `measure` will not cause the enclosing app to crash (even in a `try` block)
- does not need to consider the various reasons why measurement might fail (uninitialized ref, component out of view, etc.)
- can cleanly check the return object, instead of needing to evaluate an arbitrary property on it (e.g. `if (isNaN(measured.x))`

Note: I still occasionally get a crash that reports that the native `_measure` method is not defined. I haven't been able to fix this, because it's impossible to do a falsy check - if it doesn't exist, even a conditional or `console.log` referencing `_measure` (even in a `try`) causes a hard crash. Any ideas for how to check this?

## Changes

- Return early if `viewTag` is `-1` (invalid)
- Make early returns return `null` instead of throwing
- Fix incorrect `RefObjectFunction` import path
- Fix example assuming non-null `measure` response
- Remove unneeded ignore comments
- Fix "indefined" misspelling

Simply run `FabricExample/src/MeasureExample.tsx`.

## Checklist

- [X] Updated TS types
- [ ] Added TS types tests
- [X] Updated documentation
- [X] Ensured that CI passes


naftalibeder:
Hi, just wanted to check when this might get a review. Thanks! @piaskowyk

docs/docs/api/nativeMethods/measure.md
```diff
@@ -46,13 +46,11 @@ const Comp = () => {
...
+    }
   });
 
   return <View ref={aref} />;
 };
-```
-
-### Note
-
-You can use `measure()` only on rendered components. A good practice is to wrap the function call with a `try{} catch{}` block if there is a risk of calling the function on an item which is not rendered, for example: an invisible off screen item from FlatList.
```
tomekzaw, line 58: Can we bring this note back? The fact that it can be used only on already rendered components (e.g. `FlatList` items) is important.

src/reanimated2/NativeMethods.ts
```diff
@@ -21,36 +19,42 @@ export interface MeasuredDimensions {
...
+      `[Reanimated] The view with tag ${viewTag} is not a valid argument for measure()`
     );
+    return null;
+  }
+  try {
+    const result = _measure(viewTag);
+    if (result.x === -1234567) {
+      throw new Error(
+        `The view with tag ${viewTag} returned an invalid measurement response`
+      );
```
tomekzaw, line 46: Do we really want to throw an error here? In all other places it's `console.warn` and `return null;`.

src/reanimated2/NativeMethods.ts
```diff
@@ -21,36 +19,42 @@ export interface MeasuredDimensions {
...
+    }
+    if (isNaN(result.x)) {
+      console.warn(
+        '[Reanimated] Trying to measure a component which gets view-flattened on Android. To disable view-flattening, set `collapsable={false}` on this component.'
+      );
+    }
+    return result;
+  } catch (e) {
```
tomekzaw, line 54: What exception is supposed to be caught here?

src/reanimated2/NativeMethods.ts
```diff
@@ -21,36 +19,42 @@ export interface MeasuredDimensions {
   pageY: number;
 }
 
-const isNativeIndefined = shouldBeUseWeb();
+const isNativeUndefined = shouldBeUseWeb();
```
tomekzaw, line 22: I wish we could come up with proper names here 😄

docs/docs/api/nativeMethods/measure.md
```diff
@@ -46,13 +46,11 @@ const Comp = () => {
 
   useDerivedValue(() => {
     const measured = measure(aref);
-    // ...
+    if (measured !== null) {
+      // ...
```
tomekzaw, line 50: Maybe let's add some `console.log` here that makes use of the props of `measured`.

src/reanimated2/utils.ts
```diff
@@ -14,6 +14,9 @@ export function getRelativeCoords(
 ): ComponentCoords {
   'worklet';
   const parentCoords = measure(parentRef);
+  if (!parentCoords) {
```
tomekzaw, line 17: Let's be more explicit.
```suggestion
  if (parentCoords !== null) {
```

src/reanimated2/utils.ts
```diff
@@ -14,6 +14,9 @@ export function getRelativeCoords(
...
+    return { x, y };
```
tomekzaw, line 18: Can we add a comment here why it is valid to return `{ x, y }` here?

tomekzaw (changes requested):
@naftalibeder Thanks a lot for your contribution!

I've left a few questions and suggestions in the comments.

src/reanimated2/NativeMethods.ts
piaskowyk: I also suggest using `console.warn` everywhere.

src/reanimated2/NativeMethods.ts
naftalibeder: Yup, I was trying to avoid making too many subjective decisions, but I agree!

docs/docs/api/nativeMethods/measure.md
naftalibeder: Done - I made a few changes, including updated wording (the old wording was no longer accurate), so let me know what you think.

src/reanimated2/NativeMethods.ts
naftalibeder: This doesn't actually throw outside of the main `measure()` function - it skips execution down to the `catch` block, which uses the error in a `console.warn`. 

But I see how this is less readable, so I can just make it an inline warning.

src/reanimated2/NativeMethods.ts
naftalibeder: Aside from catching the manual throw described [here](https://github.com/software-mansion/react-native-reanimated/pull/3453/files#r961108367), this is supposed to catch an error caused by measuring any invisible view, like the offscreen `FlatList` item described in the docs.

I suspect (but I'm not confident) that the `viewTag === -1` check above basically handles this, but since one of the main purposes of this PR is to avoid letting `measure()` crash the parent app, it seemed safer to keep it wrapped in a `try`.

Thoughts?

src/reanimated2/utils.ts
naftalibeder: Hah, it's not. That was a bug. Really, I think this function should just return `null` if it can't perform the measurement - I added a jsdoc comment and made some other tweaks.

src/reanimated2/utils.ts
naftalibeder: I made an equivalent change (the suggested change would actually have inverted the check), but good call. 👍

naftalibeder:
@tomekzaw I appreciate the thorough review! I think I've addressed everything, so please let me know if you have any other concerns.

docs/docs/api/nativeMethods/measure.md
tomekzaw: Sounds good, thanks!

src/reanimated2/NativeMethods.ts
```diff
@@ -21,34 +19,41 @@ export interface MeasuredDimensions {
   pageY: number;
 }
 
-const isNativeIndefined = shouldBeUseWeb();
+const isNative = shouldBeUseWeb();
```
tomekzaw, line 22: Shouldn't it be `!shouldBeUseWeb()`?

src/reanimated2/utils.ts
tomekzaw: Oops, my fault indeed. Now it's good 👍

src/reanimated2/NativeMethods.ts
tomekzaw: > the main purposes of this PR is to avoid letting measure() crash the parent app

Yeah, that's the main goal 😄 

I think we would like to avoid using `try`/`catch` blocks if possible, i.e. use `console.warn` instead of `throw new Error()` in JS code.

src/reanimated2/NativeMethods.ts
tomekzaw: Thanks!

src/reanimated2/NativeMethods.ts
tomekzaw: I've just checked the native part of the code and it looks like the native side shouldn't throw any exceptions. If some error occurs (e.g. view has not been rendered yet or it has been flattened), it returns `NaN` (which is still better than random values, see https://github.com/software-mansion/react-native-reanimated/pull/3414), or `-1234567` (which I would rather not comment). So I think we can remove the `try`/`catch` block here.

tomekzaw:
Hey @naftalibeder, I've left two more comments. Can you please address them? I'm looking forward to merging this PR once the PR is ready.

naftalibeder:
@tomekzaw Yep, sorry about that. Today was an American holiday, but I should be able to get to it first thing tomorrow morning (USA time)!

naftalibeder:
@tomekzaw This should be good to go.

Hope it's OK - I added the `measure()` return type to the list of common types that get exported, because especially given the possible `null` return, it's useful to be able to declare a typed variable in the parent app.

tomekzaw:
> Not a big deal, because the user can always force-quiet the warning.

After your changes, there are three cases where the return value is `null`. If we remove `console.warn`, the library user doesn't know what exactly went wrong (unless someone reads the docs thoroughly).

For now, the users should use `LogBox.ignoreLogs`. Once we notice an issue from someone complaining about this warning message, we will consider removing the `console.warn`. What do you think?

tomekzaw (approved):
Looks good to me! 🎉 I think we are ready to merge. Thanks again for your contribution! 🥳

naftalibeder:
I agree about keeping the warning. I can’t think of anything else for now - feel free to merge!

piaskowyk (approved):
@naftalibeder thanks for your PR! 😍

PR #3455, opened 2022-08-09
Added FrameTimings to the useFrameCallback hook

Kwasow:
## Description

This PR adds timing info to callback passed to useFrameCallback.

## Changes

New type:

```ts
type FrameInfo = {
  timestamp: number;
  duration: number | undefined;
  timeSinceFirstFrame: number;
};
```

### Before

```ts
function useFrameCallback(
  callback: () => void,
  autostart = true
): FrameCallback
```

### After

```ts
function useFrameCallback(
  callback: (frameInfo: FrameInfo) => void,
  autostart = true
): FrameCallback
```

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [x] Updated documentation
- [x] Ensured that CI passes


docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -36,9 +37,18 @@ Properties:
                     or not (`false`)
 * `callbackId: number`: a unique identifier of the callback function
 
+#### `FrameTimings: [object]`
+
+Properties:
+* `timestamp: number`: the current system time
+* `frameTime: number`: time since last frame - this value may be zero on the first frame update
+* `elapsedTime: number`: time since the callback started running
```
tomekzaw, line 45: 1. Add info about unit (in seconds or in milliseconds?) in each line for better readability.
2. What do you think about renaming `frameTime` to `timeSinceLastFrame`?
3. I'm wondering if instead of `frameTime = 0` we can have `isFirstFrame = true`.

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,34 +1,70 @@
...
         this.frameCallbackActive.forEach((key: number) => {
+          let startTime = this.frameCallbackStartTime.get(key) || 0;
+          if (startTime === 0) {
+            startTime = timestamp;
+            this.frameCallbackStartTime.set(key, timestamp);
+          }
+
+          const frameTimings: FrameTimings = {
+            timestamp: timestamp,
+            timeSinceLastFrame: timeSinceLastFrame,
```
tomekzaw, line 54:
```suggestion
            timestamp,
            timeSinceLastFrame,
```

FabricExample/src/FrameCallbackExample.tsx
```diff
@@ -15,7 +15,11 @@ export default function FrameCallbackExample() {
   const y1 = useSharedValue(0);
 
   const frameCallback1 = useFrameCallback((frameTimings: FrameTimings) => {
-    console.log('Frame info:', frameTimings);
+    if (frameTimings.timeSinceLastFrame == null) {
```
tomekzaw, line 18:
```suggestion
    if (frameTimings.timeSinceLastFrame === null) {
```

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -41,7 +41,9 @@ Properties:
 
 Properties:
 * `timestamp: number`: the current system time (in miliseconds)
-* `timeSinceLastFrame: number`: time (in miliseconds) since last frame - this value may be zero on the first frame update
+* `timeSinceLastFrame: number | null`: time (in miliseconds) since last frame - this value
```
tomekzaw, line 44: milliseconds

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -48,11 +48,12 @@ export const prepareUIRegistry = runOnUI(() => {
...
             this.frameCallbackStartTime.set(key, timestamp);
           }
+          const elapsedTime = timestamp - startTime;
 
           const frameTimings: FrameTimings = {
             timestamp,
-            timeSinceLastFrame,
-            elapsedTime: timestamp - startTime,
+            elapsedTime,
+            timeSinceLastFrame: elapsedTime === 0 ? null : timeSinceLastFrame,
```
tomekzaw, line 56: We don't want to check `elapsedTime === 0`.

Instead, we want to check `startTime === 0` (before we overwrite this).

BTW Why is the default `startTime` 0 and not `undefined`?

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -41,7 +41,9 @@ Properties:
...
+  will be null on the first frame after activation (then it should be ~16ms on 60Hz
+  or ~8ms on 120Hz displays when there is no lag)
```
tomekzaw, line 46:
```suggestion
  will be null on the first frame after activation. Starting from second frame, it should be ~16 ms on 60 Hz
  or ~8 on 120 Hz displays (when there is no lag)
```

tomekzaw (changes requested):

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -7,14 +7,15 @@ sidebar_label: useFrameCallback
...
-useFrameCallback(updater: () => void, autostart: boolean = true) => [FrameCallback]
+useFrameCallback(callback: (frameTimings: FrameTimings) => void, autostart = true): [FrameCallback]
 ```
 
 ### Arguments
 
 #### `callback` [Function]
 
-Single worklet that will be called on every frame update without any arguments.
+Single worklet function that will be called on every frame update.
```
tomekzaw, line 17:
```suggestion
A single worklet function that will be called on every frame update.
```

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -36,9 +37,18 @@ Properties:
                     or not (`false`)
 * `callbackId: number`: a unique identifier of the callback function
 
+#### `FrameTimings: [object]`
+
+Properties:
+* `timestamp: number`: the current system time (in milliseconds)
+* `timeSinceLastFrame: number | null`: time (in milliseconds) since last frame - this value
+  will be null on the first frame after activation. Starting from the second frame,
+  it should be ~16ms on 60Hz or ~8ms on 120Hz displays (when there is no lag)
```
tomekzaw, line 46:
```suggestion
  it should be ~16 ms on 60 Hz or ~8 ms on 120 Hz displays (when there is no lag)
```

react-native-reanimated.d.ts
```diff
@@ -543,8 +543,13 @@ declare module 'react-native-reanimated' {
...
     callbackId: number;
   };
+  export type FrameTimings = {
+    timestamp: number;
+    timeSinceLastFrame: number | null;
+    elapsedTime: number;
+  };
   export function useFrameCallback(
-    callback: () => void,
+    callback: (frameTimings: FrameInfo) => void,
```
tomekzaw, line 552: What is `FrameInfo`?

react-native-reanimated.d.ts
```diff
@@ -543,8 +543,13 @@ declare module 'react-native-reanimated' {
     isActive: boolean;
     callbackId: number;
   };
+  export type FrameTimings = {
```
tomekzaw, line 546: I don't like the fact that a single object has a plural name.

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -36,9 +37,18 @@ Properties:
                     or not (`false`)
 * `callbackId: number`: a unique identifier of the callback function
 
+#### `FrameTimings: [object]`
+
+Properties:
+* `timestamp: number`: the current system time (in milliseconds)
+* `timeSinceLastFrame: number | null`: time (in milliseconds) since last frame - this value
+  will be null on the first frame after activation. Starting from the second frame,
```
tomekzaw, line 45:
```suggestion
  will be `null` on the first frame after activation. Starting from the second frame,
```

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -51,7 +61,8 @@ import React from 'react';
 export default function FrameCallbackExample() {
   const x = useSharedValue(0);
 
-  const frameCallback = useFrameCallback(() => {
+  const frameCallback = useFrameCallback((frameTimigs) => {
+    console.log(frameTimings.frameTime + 'ms', 'have passed since the previous frame');
```
tomekzaw, line 65:
```suggestion
    console.log(`${frameTimings.frameTime} ms have passed since the previous frame`);
```

react-native-reanimated.d.ts
```diff
@@ -543,8 +543,13 @@ declare module 'react-native-reanimated' {
...
+    timestamp: number;
+    timeSinceLastFrame: number | null;
```
tomekzaw, line 548: Technically, it's "duration".

obraz

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,34 +1,71 @@
...
+        const timeSinceLastFrame = timestamp - this.lastFrameTimestamp;
+
         this.frameCallbackActive.forEach((key: number) => {
+          const startTime = this.frameCallbackStartTime.get(key);
+
+          const frameTimings: FrameTimings = {
+            timestamp,
+            elapsedTime: timestamp - (startTime || timestamp),
+            timeSinceLastFrame:
+              startTime === undefined ? null : timeSinceLastFrame,
```
tomekzaw, line 52:
```ts
const elapsedTime = ...;
const timeSinceLastFrame = ...;

return { ... };
```

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,34 +1,71 @@
...
   const frameCallbackRegistry: FrameCallbackRegistryUI = {
-    frameCallbackRegistry: new Map<number, () => void>(),
+    frameCallbackRegistry: new Map<
+      number,
+      (frameTimings: FrameTimings) => void
+    >(),
+    frameCallbackStartTime: new Map<number, number>(),
     frameCallbackActive: new Set<number>(),
     isFrameCallbackRunning: false,
+    lastFrameTimestamp: 0,
```
tomekzaw, line 35: Why not `undefined` or `null`? This way we can easily detect first frame. Yeah, I know, ain't nobody using Reanimated in 1970.

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -56,6 +96,7 @@ export const prepareUIRegistry = runOnUI(() => {
         this.runCallbacks();
       } else {
         this.frameCallbackActive.delete(frameCallbackId);
+        this.frameCallbackStartTime.delete(frameCallbackId);
```
tomekzaw, line 99: Maybe we should have a single `Map` with some struct instead instead of two, but that's just a suggestion.

tomekzaw (changes requested):
I've left a few minor suggestions in the comments.

react-native-reanimated.d.ts
Kwasow: Thanks! 💡

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
Kwasow: `undefned` sounds good

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -7,15 +7,15 @@ sidebar_label: useFrameCallback
...
+useFrameCallback(callback: (frameTimings: FrameTime) => void, autostart = true): [FrameCallback]
 ```
 
 ### Arguments
 
 #### `callback` [Function]
 
 A single worklet function that will be called on every frame update.
-This function recieves a [`FrameTimings`](#frametimings-object) object as na argument.
+This function recieves a [`FrameTime`](#frametime-object) object as na argument.
```
tomekzaw, line 18:
```suggestion
This function receives a [`FrameTime`](#frametime-object) object as na argument.
```

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -15,7 +15,7 @@ useFrameCallback(callback: (frameTimings: FrameTime) => void, autostart = true):
 #### `callback` [Function]
 
 A single worklet function that will be called on every frame update.
-This function recieves a [`FrameTime`](#frametime-object) object as na argument.
+This function receives a [`FrameTime`](#frametime-object) object as na argument.
```
tomekzaw, line 18:
```suggestion
This function receives a [`FrameTime`](#frametime-object) object as an argument.
```

Kwasow: This line better be perfect, after all these suggestions.

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,34 +1,73 @@
...
+          const startTime = this.frameCallbackRegistry.get(key)?.startTime;
+          const elapsedTime = timestamp - (startTime || timestamp);
+          const duration =
+            startTime === undefined ? undefined : timeSinceLastFrame;
+
+          if (startTime === undefined) {
+            const callbackDetails = this.frameCallbackRegistry.get(key)!;
+            callbackDetails.startTime = timestamp;
+
+            this.frameCallbackRegistry.set(key, callbackDetails);
```
piaskowyk, line 55: the same case, you don't need to reassign.

piaskowyk (changes requested):

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -36,9 +37,19 @@ Properties:
                     or not (`false`)
 * `callbackId: number`: a unique identifier of the callback function
 
+#### `FrameInfo: [object]`
+
+Properties:
+* `timestamp: number`: the system time (in milliseconds) when the last
+  frame was rendered
+* `duration: number | undefined`: time (in milliseconds) since last frame - this value
+  will be `undefined` on the first frame after activation. Starting from the second frame,
```
tomekzaw, line 46: I think we should use `null` here instead of `undefined`. What do you think?
```suggestion
* `timeSinceLastFrame: number | null`: time (in milliseconds) since last frame - this value
  will be `null` on the first frame after activation. Starting from the second frame,
```

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,11 +1,26 @@
...
+export type FrameInfo = {
+  timestamp: number;
+  duration: number | undefined;
+  timeSinceFirstFrame: number;
+};
+
+interface FrameCallbackRegistryUI {
+  frameCallbackRegistry: Map<number, CallbackDetails>;
   frameCallbackActive: Set<number>;
   isFrameCallbackRunning: boolean;
```
tomekzaw, line 17: Is this property necessary? It looks like we only assign `false` and never use it, but I might be wrong.

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,11 +1,26 @@
...
+
+export type FrameInfo = {
+  timestamp: number;
+  duration: number | undefined;
+  timeSinceFirstFrame: number;
+};
+
+interface FrameCallbackRegistryUI {
+  frameCallbackRegistry: Map<number, CallbackDetails>;
   frameCallbackActive: Set<number>;
```
tomekzaw, line 16:
```suggestion
  activeFrameCallbacks: Set<number>;
```

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -14,21 +29,44 @@ export const prepareUIRegistry = runOnUI(() => {
...
+    previousFrameTimestamp: undefined,
 
     runCallbacks() {
-      const loop = () => {
+      const loop = (timestamp: number) => {
+        if (this.previousFrameTimestamp === undefined) {
+          this.previousFrameTimestamp = timestamp;
+        }
+
+        const timeSinceLastFrame = timestamp - this.previousFrameTimestamp;
```
tomekzaw, line 43: We just assigned `this.previousFrameTimestamp = timestamp;` so effectively this will be `timestamp - timestamp`, right?

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -14,21 +29,44 @@ export const prepareUIRegistry = runOnUI(() => {
...
+
         this.frameCallbackActive.forEach((key: number) => {
-          const callback = this.frameCallbackRegistry.get(key);
-          callback!();
+          const callback = this.frameCallbackRegistry.get(key)!;
+
+          const startTime = callback.startTime;
```
tomekzaw, line 48: You can also use destructuring assignment:
```suggestion
          const { startTime } = callback;
```

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -14,21 +29,44 @@ export const prepareUIRegistry = runOnUI(() => {
...
+        if (this.previousFrameTimestamp === undefined) {
+          this.previousFrameTimestamp = timestamp;
+        }
+
+        const timeSinceLastFrame = timestamp - this.previousFrameTimestamp;
+
         this.frameCallbackActive.forEach((key: number) => {
-          const callback = this.frameCallbackRegistry.get(key);
-          callback!();
+          const callback = this.frameCallbackRegistry.get(key)!;
```
tomekzaw, line 46:
```suggestion
          const callbackDetails = this.frameCallbackRegistry.get(key)!;
```
OR
```suggestion
          const frameCallback = this.frameCallbackRegistry.get(key)!;
```

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -55,6 +99,9 @@ export const prepareUIRegistry = runOnUI(() => {
         this.frameCallbackActive.add(frameCallbackId);
         this.runCallbacks();
       } else {
+        const callback = this.frameCallbackRegistry.get(frameCallbackId)!;
```
tomekzaw, line 102: It looks like you're using different names for the same thing:
* `key`
* `callbackId`
* `frameCallbackId`.

You probably need to choose one.

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -36,9 +37,19 @@ Properties:
                     or not (`false`)
 * `callbackId: number`: a unique identifier of the callback function
 
+#### `FrameInfo: [object]`
+
+Properties:
+* `timestamp: number`: the system time (in milliseconds) when the last
+  frame was rendered
+* `duration: number | undefined`: time (in milliseconds) since last frame - this value
```
tomekzaw, line 45:
```suggestion
* `duration: number | undefined`: time (in milliseconds) since last frame. This value
```

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -51,7 +62,8 @@ import React from 'react';
 export default function FrameCallbackExample() {
   const x = useSharedValue(0);
 
-  const frameCallback = useFrameCallback(() => {
+  const frameCallback = useFrameCallback((frameInfo) => {
+    console.log(`${frameInfo.lastFrameDuration} ms have passed since the previous frame`);
```
tomekzaw, line 66: What about the first frame?

tomekzaw (changes requested):

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
Kwasow: We do use it, but actually `isFrameCallbackRunning == (frameCallbackActive.size > 0)` is always true, so we could get rid of it I think.

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
Kwasow: Yes, but only on the first frame. Do you think there is a better way of doing this?

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
Kwasow: I think I like `callbackDetails` more

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
Kwasow: I went with `callbackId`

piaskowyk (approved):

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -7,14 +7,15 @@ sidebar_label: useFrameCallback
...
+useFrameCallback(callback: (frameInfo: frameInfo) => void, autostart = true): [FrameCallback]
 ```
 
 ### Arguments
 
 #### `callback` [Function]
 
-Single worklet that will be called on every frame update without any arguments.
+A single worklet function that will be called on every frame update.
+This function recieves a [`FrameInfo`](#frameinfo-object) object as na argument.
```
tomekzaw, line 18:
```suggestion
This function receives a [`FrameInfo`](#frameinfo-object) object as an argument.
```

docs/docs/api/hooks/useFrameCallback.md
```diff
@@ -51,7 +62,12 @@ import React from 'react';
 export default function FrameCallbackExample() {
   const x = useSharedValue(0);
 
-  const frameCallback = useFrameCallback(() => {
+  const frameCallback = useFrameCallback((frameInfo) => {
+    if (frameInfo.timeSincePreviousFrame == null) {
```
tomekzaw, line 66:
```suggestion
    if (frameInfo.timeSincePreviousFrame === null) {
```

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,61 +1,112 @@
 import { runOnUI } from '../core';
 
-export default interface FrameCallbackRegistryUI {
-  frameCallbackRegistry: Map<number, () => void>;
-  frameCallbackActive: Set<number>;
-  isFrameCallbackRunning: boolean;
+type CallbackDetails = {
+  callback: (frameInfo: FrameInfo) => void;
+  startTime: number | undefined;
```
tomekzaw, line 5:
```suggestion
  startTime: number | null;
```

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,61 +1,112 @@
...
 export const prepareUIRegistry = runOnUI(() => {
   'worklet';
 
   const frameCallbackRegistry: FrameCallbackRegistryUI = {
-    frameCallbackRegistry: new Map<number, () => void>(),
-    frameCallbackActive: new Set<number>(),
-    isFrameCallbackRunning: false,
+    frameCallbackRegistry: new Map<number, CallbackDetails>(),
+    activeFrameCallbacks: new Set<number>(),
+    previousFrameTimestamp: undefined,
```
tomekzaw, line 33:
```suggestion
    previousFrameTimestamp: null,
```

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,61 +1,112 @@
...
+          }
         });
 
-        if (this.frameCallbackActive.size > 0) {
+        if (this.activeFrameCallbacks.size > 0) {
+          this.previousFrameTimestamp = timestamp;
           requestAnimationFrame(loop);
         } else {
-          this.isFrameCallbackRunning = false;
+          this.previousFrameTimestamp = undefined;
```
tomekzaw, line 71:
```suggestion
          this.previousFrameTimestamp = null;
```

src/reanimated2/frameCallback/FrameCallbackRegistryUI.ts
```diff
@@ -1,61 +1,112 @@
...
 
-    registerFrameCallback(callback: () => void, callbackId: number) {
-      this.frameCallbackRegistry.set(callbackId, callback);
+    registerFrameCallback(
+      callback: (frameInfo: FrameInfo) => void,
+      callbackId: number
+    ) {
+      this.frameCallbackRegistry.set(callbackId, {
+        callback: callback,
+        startTime: undefined,
```
tomekzaw, line 89:
```suggestion
        startTime: null,
```

tomekzaw (approved):
🚀

PR #3470, opened 2022-08-12
Fix useAnimatedKeyboard inset

piaskowyk:
## Description

On Paper architecture, we need to manually set the bottom inset to prevent cropping application content by the navigation bottom bar, but this is no longer required on Fabric.

Since rn70 Paper also doesn't need an additional bottom inset.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -270,6 +270,7 @@ android {
 
         buildConfigField("boolean", "IS_INTERNAL_BUILD", "false")
         buildConfigField("int", "EXOPACKAGE_FLAGS", "0")
+        buildConfigField("int", "REACT_NATIVE_VERSION", rnMinorVersion.toString())
     }
     externalNativeBuild {
         cmake {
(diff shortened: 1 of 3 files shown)


DmitriyGalyanov:
Hello

That's not the best report, but may be you will be able to investigate a bit (or ask me further if I can help)

We are using React Native 0.71.3 with new Architecture enabled and Reanimated 2.14.4

Using ```NativeReanimated.subscribeForKeyboardEvents``` causes mentioned problem (application content cropping) to occur whilst this fix is applied
Tested on:
- Samsung S20 5G, Android 12, One UI 4.0
- Samsung Galaxy S8+, Android 9, One UI 1.0
- HONOR 50 Lite, Android 11, Magic UI 4.2
- Samsung A51, Android 11, One UI 3.1

Reverting to ```int paddingBottom = insets.getInsets(WindowInsetsCompat.Type.navigationBars()).bottom;``` in ```ReanimatedKeyboardEventListener.setupWindowInsets``` fixes the issue for us

PR #3479, opened 2022-08-16
Update CI

piaskowyk:
## Description

Reanimated 3 doesn't contain binaries for android so we needed to update CI for the package building CI job also.

### DIFF
--- a/createNPMPackage.sh
+++ b/createNPMPackage.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+yarn
 yarn bob build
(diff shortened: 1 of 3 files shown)


createNPMPackage.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+yarn
```
tomekzaw, line 3:
```suggestion
yarn install --frozen-lockfile
```

tomekzaw (approved):

PR #3493, opened 2022-08-18
Unify assertion for no multiple instances of Reanimated with RNGH

kkafar:
## Description

See: https://github.com/software-mansion/react-native-gesture-handler/pull/2174

## Changes

Fixed crashing on symlinks with filtering files before executing operations on them, added property `disableMultipleInstancesCheck ` (possible to specify in `gradle.properties` of **rootProject**) which allow application to disable the feature. 

Note that this property is shared with `react-native-gesture-handler` - so if someone sets `disableMultipleInstancesCheck=true` the assertion won't run for neither `react-native-reanimated` nor `react-native-gesture-handler`. 

Works fine on `TestsExample` application in `react-native-screens` repository. 

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


tomekzaw:
We will need to exclude `.yarn` directory as fixed in this PR:
* https://github.com/software-mansion/react-native-reanimated/pull/3510

tomekzaw (approved):

piaskowyk (approved):

PR #3498, opened 2022-08-19
androidx backward compatibility

piaskowyk:
## Description

`androidx.core:core:1.7.0` requires the newer `minSdkVersion` than is used in react native in versions 66 and lower. `androidx` library is used by `useAnimatedKeyboard()`


tomekzaw:
Why can't we use 1.6.0 even with newer RN versions? Maybe there were some relevant bug fixes introduced in 1.7.0?

Besides, currently the latest stable version seems to be 1.8.0 (see [here](https://androidx.tech/artifacts/core/core/)).

tomekzaw (approved):
🚀

PR #3505, opened 2022-08-22
docs: Change yarn to npx example readme

gkartalis:
## Description

The reason for this is that `yarn react-native run-ios` doesn't work, `npx react-native run-ios` or `npx react-native run-android` will work cause npx runs the react native script while yarn doesn't!

**NPX** stands for Node Package eXecute. It is simply an NPM package runner. It allows developers to execute any Javascript Package available on the NPM registry without even installing it. NPX is installed automatically with NPM version 5.2.0 and above.

## Changes

- doc changes in example to be able to run the app

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [x] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/Example/README.md
+++ b/Example/README.md
@@ -8,12 +8,12 @@
 ### Running on iOS
 
 - `yarn pod-install` - to install pods
-- `yarn react-native run-ios` - to run the app
+- `npx react-native run-ios` - to run the app
 
 ### Running on Android
 
 - [install NDK in version 21.3.6528147 or higher](https://developer.android.com/studio/projects/install-ndk)
-- `yarn react-native run-android` - to run the app
+- `npx react-native run-android` - to run the app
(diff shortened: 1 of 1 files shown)


tomekzaw:
Hey @gkartalis, thanks for the PR.

Both `yarn react-native run-android` and `yarn react-native run-ios` work for me. Do you have some kind of error message or stack trace?

Also, what is the advantage of using `npx` here instead of `yarn`?

Besides, I think that `npx` will run the scripts from global installation of `react-native` while `yarn react-native` will run scripts from the version installed in the `node_modules` which is the preferred option.

gkartalis:
Gotcha, retried it and it works with `yarn` 🤔 thanks! Closing the pr

Issue #2994, opened 2022-02-11
TypeError: Cannot read properties of undefined (reading 'toString')


shensven:
## Description
After installed the `react-native-collapsible-tab-view` and wrote a minimal example, jest does not pass
see this commit https://github.com/shensven/Readhubn/commit/9b0877f890d9acf5542ec91964dc374307270f0c

```sh
yarn run v1.22.15
$ jest
 FAIL  __tests__/App-test.tsx (6.541 s)
  ✕ renders correctly (209 ms)

  ● renders correctly

    TypeError: Cannot read properties of undefined (reading 'toString')

      at reduce (node_modules/react-native-reanimated/lib/reanimated2/hook/utils.js:46:33)
          at Array.reduce (<anonymous>)
      at buildWorkletsHash (node_modules/react-native-reanimated/lib/reanimated2/hook/utils.js:44:36)
      at buildDependencies (node_modules/react-native-reanimated/lib/reanimated2/hook/utils.js:60:27)
      at useHandler (node_modules/react-native-reanimated/lib/reanimated2/hook/utils.js:36:20)
      at useAnimatedScrollHandler (node_modules/react-native-reanimated/lib/reanimated2/hook/useAnimatedScrollHandler.js:5:47)
      at node_modules/react-native-collapsible-tab-view/lib/commonjs/Container.tsx:284:30
      at renderWithHooks (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:6016:18)
      at updateForwardRef (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:8038:20)
      at beginWork (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:10015:16)

  console.warn
    [react-native-gesture-handler] Seems like you're using an old API with gesture components, check out new Gestures system!

      at showRngh2NoticeIfNeeded (node_modules/react-native-gesture-handler/lib/commonjs/handlers/createHandler.ts:128:5)
      at new Handler (node_modules/react-native-gesture-handler/lib/commonjs/handlers/createHandler.ts:176:9)
      at constructClassInstance (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:4022:18)
      at updateClassComponent (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:8399:5)
      at beginWork (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:9990:16)
[...]

PR #3531, opened 2022-08-31
Handle SequenceExpression in babel transform

scottmas:
## Description

Handle function invocations created via sequence (comma) expressions. E.g. `(0, _reactNativeReanimated.useAnimatedGestureHandler)({...})` This happens commonly when libraries ship code that has been transformed by Babel. For example, [@gorhom/bottom-sheet](https://github.com/gorhom/react-native-bottom-sheet/blob/f2a933274c88004357700bf728c1c3d1fde48d20/src/hooks/useGestureHandler.ts#L28) gets transformed to the above sequence expression and therefore does NOT get transformed. 

This is likely the root cause of many bug reports in this library and important downstream packages:

Fixes https://github.com/software-mansion/react-native-reanimated/issues/2994
Fixes https://github.com/gorhom/react-native-bottom-sheet/issues/819
Fixes https://github.com/gorhom/react-native-bottom-sheet/issues/732
Fixes https://github.com/gorhom/react-native-bottom-sheet/issues/1030

## Changes

I genericize the `processWorklets` function to handle SequenceExpressions. We are traversing all CallExpressions in the file in order to find arguments to the CallExpression that need to be converted to worklets. And since the worklet functions always live inside the arguments of the CallExpression, this change _should_ be safe since it is only changing how the name of the function is determined. 

Call the following code inside any component to trigger an error. The dependency array is important.
```jsx
import { useAnimatedGestureHandler } from 'react-native-reanimated'

function App(){
  (0, useAnimatedGestureHandler)({ onStart() {} }, []);
}

```

## Checklist

- [ ] Ensured that CI passes


tomekzaw:
Hey @scottmas, thanks a lot for your contribution!

Could you please also verify the implementation by adding snapshot unit tests in [\_\_tests\_\_/plugin.test.js](https://github.com/software-mansion/react-native-reanimated/blob/main/__tests__/plugin.test.js)?

You can run the tests with `yarn jest plugin.test.js` in the project root directory.

scottmas:
Sure. Do you have any "Contributing" readme somewhere? I can't find anything and it's just always a nightmare in my experience to get things up and running on a big native project like this.

tomekzaw:
Hey, unfortunately we don't (yet). However, for testing Reanimated Babel plugin you only need to clone the repository and run `yarn` in the project root directory. Then `yarn jest plugin.test.js` should work as expected.

piaskowyk (approved):

PR #3554, opened 2022-09-07
Fix plugin crashes when no options are being passed to it

cubex2:
## Description

Currently when adding the plugin to `babel.config.js` without passing any options to it, `state.opts` will be undefined causing the webpack build to fail because it can't read relativeSourceLocation of undefined.-->

## Changes

- Added check wheter plugin options are being passed before accesing them

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -450,7 +450,7 @@ function makeWorklet(t, fun, state) {
   const workletHash = hash(funString);
 
   let location = state.file.opts.filename;
-  if (state.opts.relativeSourceLocation) {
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):

tomekzaw:
@cubex2 Thanks for the PR!

piaskowyk (approved):

PR #3347, opened 2022-07-01
Support for Jest 28

piaskowyk:
## Description

Add support for Jest 28 with backward compatibility.

Based on: https://github.com/software-mansion/react-native-reanimated/pull/3217

Issue #3215, opened 2022-05-05
TypeError: expect.extend is not a function


billnbell:
Latest RN 2.8.0 "react-native-reanimated": "2.8.0".

Getting:
 FAIL  src/__tests__/ComponentRender-test.tsx
  ● Test suite failed to run

    TypeError: expect.extend is not a function

      at Object.setUpTests (node_modules/react-native-reanimated/lib/reanimated2/jestUtils.js:219:10)
      at Object.<anonymous> (jestSetupFile.js:163:62)

billnbell:
This happens on jest  upgrade from ^27.5.1  →  ^28.0.3

finalight:
> To fix this:
> 
> 1. Ensure you are calling this inside a script script that runs in `setupFilesAfterEnv` NOT `setupFiles`
> 2. Change the import in the script to `const { expect } = require('@jest/globals');` from `const expect = require('expect');` with `patch-package`
> 
> I have opened a PR to fix this here: #3217

i've tried the same thing you did, but it still shows me the same error

mayconline:
> To fix this:
> 
> 1. Ensure you are calling this inside a script script that runs in `setupFilesAfterEnv` NOT `setupFiles`
> 2. Change the import in the script to `const { expect } = require('@jest/globals');` from `const expect = require('expect');` with `patch-package`
> 
> I have opened a PR to fix this here: #3217

same error, i believe this resolve this error when merge,
link do helper: https://jestjs.io/pt-BR/docs/upgrading-to-jest28

finalight:
the PR is merged, but i still facing the same error

```
  ● Test suite failed to run

    TypeError: expect.extend is not a function

      22 | }
      23 |
    > 24 | require('react-native-reanimated/lib/reanimated2/jestUtils').setUpTests()
         |                                                              ^
      25 |
      26 | // eslint-disable-next-line no-underscore-dangle
      27 | global.__reanimatedWorkletInit = jest.fn()

      at Object.extend [as setUpTests] (node_modules/react-native-reanimated/lib/reanimated2/jestUtils.js:152:12)
      at Object.setUpTests (jest-utils/jest.setup.js:24:62)
```

Norfeldt:
Here is a repo showing how it breaks after upgrading to jest 28 https://github.com/Norfeldt/expo-46-with-jest-rtnl-reanimated

billnbell:
still erroring!!!!!!

[...]

PR #3559, opened 2022-09-10
Fix Jest 28+ `expect.extend is not a function` error

### LINKED ISSUE #3553: Won't build Jest issue
### Description

Jest still will not build right.

```
 export const setUpTests = (userConfig = {}) => {
-    const expect = require('expect');
+    const { expect } = require('@jest/globals');
     require('setimmediate');
```

You added a try / catch around `require('expect');` but this does not help me, since I have expect loaded.

### Steps to reproduce

Use SetUpTests() for Jest.

none

### Reanimated version

2.10.0

### React Native version

0.70.0

### Platforms

iOS

### JavaScript runtime

_No response_

### Workflow

_No response_

### Architecture

_No response_

### Build type

_No response_

### Device

_No response_

### Device model

_No response_

### Acknowledgements

Yes

yusufyildirim:
## Description
Reanimated is not fully compatible with Jest 28+ even after the PR https://github.com/software-mansion/react-native-reanimated/pull/3347. Using it with Jest 28+ leads to this error:

Ekran Resmi 2022-09-10 15 00 33

The reason is, the "expect" module's export approach has changed starting from Jest 28. It used to export the `expect` function directly, as can be seen [here](https://github.com/facebook/jest/blob/v27.5.1/packages/expect/src/index.ts#L441). 

This means, requiring the "expect" module like this

```ts
const expect = require('expect')
```

was providing direct access to the `expect` function alongside with its attached methods, like `extend`.

This has [changed with Jest 28](https://github.com/facebook/jest/blob/v28.0.0/packages/expect/src/index.ts#L449). `export default expect` is the new export statement, which means requiring the "expect" package returns the module object, instead of the function.

https://github.com/software-mansion/react-native-reanimated/pull/3347 PR attempts to fix this error by assuming "expect" package should throw an error with Jest 28 and it can catch that and require the "expect" from the `@jest/globals` package for Jest 28+ project but it's not the case in reality. "expect" package is still shipped with Jest 28, so, requiring it doesn't throw an error as expected in the PR.

By knowing the export difference, I checked if the value of the module is object or not after requiring it. If it's an object, it means user is on Jest 28+ codebase. The rest is the same with the previous PR. I require the "expect" function from `@jest/globals` and use it.

Tested on Jest 29 and Jest 27.5.1

Fixes #3553 and #3215.

### DIFF
--- a/src/reanimated2/jestUtils.ts
+++ b/src/reanimated2/jestUtils.ts
@@ -188,11 +188,13 @@ export const advanceAnimationByFrame = (count) => {
 };
 
 export const setUpTests = (userConfig = {}) => {
-  let expect;
-  try {
-    expect = require('expect');
-  } catch (_) {
-    // for Jest in version 28+
+  let expect = require('expect');
+
+  // Starting from Jest 28, "expect" package uses named exports instead of default export.
+  // So, requiring "expect" package doesn't give direct access to "expect" function anymore.
+  // It gives access to the module object instead.
+  // We use this info to detect if the project uses Jest 28 or higher.
+  if (typeof expect === 'object') {
     const { expect: expectModule } = require('@jest/globals');
     expect = expectModule;
   }
--- a/src/reanimated2/jestUtils.ts
+++ b/src/reanimated2/jestUtils.ts
@@ -188,15 +188,21 @@ export const advanceAnimationByFrame = (count) => {
 };
 
 export const setUpTests = (userConfig = {}) => {
-  let expect = require('expect');
-
-  // Starting from Jest 28, "expect" package uses named exports instead of default export.
-  // So, requiring "expect" package doesn't give direct access to "expect" function anymore.
-  // It gives access to the module object instead.
-  // We use this info to detect if the project uses Jest 28 or higher.
-  if (typeof expect === 'object') {
-    const { expect: expectModule } = require('@jest/globals');
+  let expect = global.expect;
+  if (expect === undefined) {
+    const expectModule = require('expect');
     expect = expectModule;
+    // Starting from Jest 28, "expect" package uses named exports instead of default export.
+    // So, requiring "expect" package doesn't give direct access to "expect" function anymore.
+    // It gives access to the module object instead.
+    // We use this info to detect if the project uses Jest 28 or higher.
+    if (typeof expect === 'object') {
+      const jestGlobals = require('@jest/globals');
+      expect = jestGlobals.expect;
+    }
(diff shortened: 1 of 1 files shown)


finalight:
can we merge this?

kesha-antonov:
Would love to see get it merged. Need to run our tests

finalight:
i tried to patch-package but i still have the same error

```
  ● Test suite failed to run

    TypeError: expect.extend is not a function
```

my codebase has a mix of javascript and typescript, is there anything else I need to patch?

finalight:
i did a further debugging

i log the value of this particular line 

```
const { expect: expectModule } = require('@jest/globals');
```

`expectModule` returns undefined in this case

and since it's undefined, assigning it to the variable `expect` gives a undefined as well

yusufyildirim:
Closed accidentally.

yusufyildirim:
> i did a further debugging
> 
> i log the value of this particular line
> 
> ```
> const { expect: expectModule } = require('@jest/globals');
> ```
> 
> `expectModule` returns undefined in this case
> 
> and since it's undefined, assigning it to the variable `expect` gives a undefined as well

You should have `@jest/globals` package if you're Jest 28+ AFAIK. How's that possible?

backlands:
I was able to test this patch in my react-native project with Jest and Jest-Native and everything is working once I also updated to the `setupFilesAfterEnv` value.

The error received by @yusufyildirim appears to occur if you have not migrated from using the `setupFile` to using `setupFilesAfterEnv` as outlined in the documentation [here](https://docs.swmansion.com/react-native-reanimated/docs/guide/testing) for Jest 28 and older.

fabrianibrahim:
@piaskowyk can you please review this PR? it would be nice to merge it soon to make it work with Jest 28+

parasharrajat:
It would be great if this can be released with 2.x

piaskowyk:
Hey 👋 We have two similar PR with a fix, this one and #3583. The #3583 seems more complex, could someone confirm if #3583 fixes your problem?

backlands:
> Hey wave We have two similar PR with a fix, this one and #3583. The #3583 seems more complex, could someone confirm if #3583 fixes your problem?

I did as requested and have tested #3583 as well as retested this PR and both work but further notes about #3583 are added to that PR for your review.

piaskowyk:
Thank you all so much for the help! 😍 I will port it to Reanimated2 branch also.

piaskowyk (approved):

PR #3563, opened 2022-09-12
Select appropriate AARs depending on build type

tomekzaw:
## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -288,7 +288,7 @@ android {
         abortOnError false
     }
     packagingOptions {
-        doNotStrip resolveBuildType() == 'Debug' ? "**/**/*.so" : ''
+        doNotStrip resolveBuildType() == 'debug' ? "**/**/*.so" : ''
         excludes = [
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -737,7 +737,7 @@ afterEvaluate {
 
 if (CLIENT_SIDE_BUILD) {
     def aarDir = "${buildDir}/outputs"
-    def aar = file("${aarDir}/android-debug.aar")
+    def aar = file("${aarDir}/android-${resolveBuildType()}.aar") // e.g. android-debug.aar
```
tomekzaw, line 740: I'm not sure what these line actually do. I suspect it's something related to building Android part of npm package. Do you think we can leave `android-debug.aar` here? Do we even have two different builds for debug and release mode?

piaskowyk (approved):

PR #3572, opened 2022-09-15
Fix Example app with JSC on Android

Kwasow:
## Description

The Example app crashes on Android if compiled with JSC instead of Hermes. Error message:

```
2022-09-15 12:34:45.701 15509-15509/com.swmansion.reanimated.example W/System.err: java.lang.UnsatisfiedLinkError: dlopen failed: library "libjsc.so" not found: needed by /data/app/~~YZvjm8gAMg_2TYxNnYOCDA==/com.swmansion.reanimated.example--9SU_U_zrfiGBnf1eSq-kA==/lib/arm64/libjscexecutor.so in namespace classloader-namespace
```

## Cause of crash

The `build.gradle` files in the Example app were outdated, which caused multiple issues. Some of the changes are:

- Added missing 64-bit architectures in some places
- Fixed enabled build abi resolution
- Fixed incorrect maven repositories definition in top-level `build.gradle`

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/Example/android/app/build.gradle
+++ b/Example/android/app/build.gradle
@@ -148,17 +148,20 @@ android {
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
         }
     }
+
     // applicationVariants are e.g. debug, release
     applicationVariants.all { variant ->
         variant.outputs.each { output ->
             // For each separate APK per architecture, set a unique version code as described here:
-            // http://tools.android.com/tech-docs/new-build-system/user-guide/apk-splits
-            def versionCodes = ["armeabi-v7a": 1, "x86": 2]
+            // https://developer.android.com/studio/build/configure-apk-splits.html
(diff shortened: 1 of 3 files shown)


billouboq:
Can we use reanimated 2 without Hermes without this patch, or is this allowing it ?

Kwasow:
Hi @billouboq, we have just discovered this issue and haven't had time to investigate it so at this time I cannot answer your question, but feel free to try running Reanimated 2 without Hermes and let us know if it works!

Kwasow:
I've investigated the issue @billouboq, and it seems that it is only an issue with our Example app, so you should be fine using other runtimes than Hermes with Reanimated.

billouboq:
Thanks you very much, will try it too !

tomekzaw:
Removed node_modules and cleared all caches but still getting this error while bundling:

Zrzut ekranu 2022-09-22 o 12 05 30

piaskowyk (approved):

PR #3583, opened 2022-09-18
fix(jestUtils): use global `expect` if available

### LINKED ISSUE #3215: TypeError: expect.extend is not a function

### LINKED ISSUE #3553: Won't build Jest issue

SimenB:
## Description

Doing `require('expect')` (or `require('@jest/globals')`) is not needed unless Jest is configured with [`injectGlobals: false`](https://jestjs.io/docs/configuration#injectglobals-boolean). So just using the global means less work.

We probably don't even need the fallback to `require('expect')` as `@jest/globals` will exist in any setup where `expect` is not a global.

Closes #3559
Fixes #3553
Fixes #3215

## Changes

Check for global `expect` before trying to load it through `node_modules`.

This now works correctly for Jest 28.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/src/reanimated2/jestUtils.ts
+++ b/src/reanimated2/jestUtils.ts
@@ -188,13 +188,14 @@ export const advanceAnimationByFrame = (count) => {
 };
 
 export const setUpTests = (userConfig = {}) => {
-  let expect;
-  try {
-    expect = require('expect');
-  } catch (_) {
-    // for Jest in version 28+
-    const { expect: expectModule } = require('@jest/globals');
-    expect = expectModule;
+  let jestExpect = typeof expect === 'undefined' ? undefined : expect;
+  if (!jestExpect) {
+    try {
+      jestExpect = require('@jest/globals').expect;
+    } catch (_) {
+      // for Jest in version < 25.5
+      jestExpect = require('expect');
+    }
   }
 
   require('setimmediate');
@@ -205,7 +206,7 @@ export const setUpTests = (userConfig = {}) => {
     ...userConfig,
   };
 
-  expect.extend({
+  jestExpect.extend({
     toHaveAnimatedStyle(received, expectedStyle, config = {}) {
       return compareStyle(received, expectedStyle, config);
     },


backlands:
@piaskowyk I tested this PR against a project attempting to upgrade to Jest v29 and the tests are all able to run through as before upgrading from an older version of Jest. This change or the change from #3559 works in this same project, though I noticed that the internal checks from GitHub Actions are passing on the other PR but not this one, specifically due to `TypeError: Cannot read property 'extend' of undefined` so this PR may be affecting something unexpected (hah).

Since it is not my project I am impartial on which solution you use as both provide the fix needed to support a wider range of Jest versions. Thanks the work on this PR @SimenB and for moving issues #3553 and #3215 forward @piaskowyk.

piaskowyk:
Thank you for the help! I appreciate so much 😍
https://github.com/software-mansion/react-native-reanimated/pull/3559 includes this PR also

PR #3588, opened 2022-09-20
fix: dependency downloads on v3

### LINKED ISSUE #3523: Why 2.10.0 need to download and build boost, glog, double-conversion, folly when newArchEnabled is set to false
### Description

Why 2.10.0 need to download and build boost, glog, double-conversion, folly when newArchEnabled is set to false?

even if newArchEnabled = true, it should use the react-native one

image

### Steps to reproduce

1. create a react native app
2. add react-native-reanimated: "^2.10.0" to package
3. yarn android

/

### Reanimated version

2.10.0

### React Native version

0.69.3

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

_No response_

### Device model

_No response_

### Acknowledgements

Yes

kkafar:
## Description

Fixes #3523 on v3

Fix for v2 comes with https://github.com/software-mansion/react-native-reanimated/pull/3602

## Changes

Added `resolve<lib>` tasks. Each of them check whether `<lib>` is already downloaded by `:ReactAndroid` project - if so - they copy archive files to `reanimated` downloads directory -> this causes `download<lib>` tasks to not run, as they're up to date.

To achieve this, setting dependency between `resolve<lib>` & `:ReactAndroid:download<lib>` was required.

I checked it on Reanimated v3:

* [x] `FabricExample` - `rm -fr node_modules && yarn`, run the build for the first time, see that the libraries are downloaded only once by `:ReactAndroid` project. In consecutive builds libraries are not downloaded.
* [x] `Example` - `rm -fr node_modules && yarn`, run the build for the first time, see that the libraries are downloaded by `:react-native-reanimated` project. In consecutive build libraries are not downloaded. 
* [x] Fresh React Native application (both archs)

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -52,21 +52,17 @@ def resolveBuildType() {
 def resolveClientSideBuild() {
     def clientSideBuild = System.getenv("CLIENT_SIDE_BUILD")
     if (clientSideBuild != null) {
-        println "Resolving CLIENT_SIDE_BUILD to ${clientSideBuild == "True"} due to env var"
         return clientSideBuild == "True"
     }
 
     if (hasProperty("clientSideBuild")) {
-        println "Resolving CLIENT_SIDE_BUILD to ${property("clientSideBuild") == "true"} due to project property"
         return property("clientSideBuild") == "true"
     }
 
     if (isDeveloperMode()) {
-        println "Resolving CLIENT_SIDE_BUILD to false due to developer mode"
         return false
     }
 
-    println "Resolving CLIENT_SIDE_BUILD to true by defaulting"
     return true
 }
 
@@ -157,11 +153,6 @@ def thirdPartyNdkDir = new File("$buildDir/third-party-ndk")
 def reactNativeThirdParty = new File("$reactNativeRootDir/ReactAndroid/src/main/jni/third-party")
 def reactNativeAndroidDownloadDir = new File("$reactNativeRootDir/ReactAndroid/build/downloads")
 
-println "reactNativeThirdParty ${reactNativeThirdParty}"
-println "customDownloadsDir ${customDownloadsDir}"
-println "downloadsDir ${downloadsDir}"
-println "thirdPartyNdkDir ${thirdPartyNdkDir}"
-
 def _stackProtectorFlag = true
 
 def CMAKE_NODE_MODULES_DIR = CLIENT_SIDE_BUILD
@@ -210,14 +201,10 @@ def reactNativeArchitectures() {
 //   - glog-0.3.5
 def dependenciesPath = System.getenv("REACT_NATIVE_DEPENDENCIES")
 
-println "dependenciesPath ${dependenciesPath}"
-
 // The Boost library is a very large download (>100MB).
 // If Boost is already present on your system, define the REACT_NATIVE_BOOST_PATH env variable
 // and the build will use that.
 def boostPath = dependenciesPath ?: System.getenv("REACT_NATIVE_BOOST_PATH")
-
-println "boostPath ${boostPath}"
 
 def follyReplaceContent = '''
   ssize_t r;
@@ -405,15 +392,11 @@ Task resolveGlog = resolveTaskFactory("resolveGlog", "glog-${GLOG_VERSION}.tar.g
 
 def reactNativeAndroidProject = findProject(":ReactAndroid")
 if (reactNativeAndroidProject != null) {
-    println "ReactAndroid project $reactNativeAndroidProject"
-    println "Before after evaluate block"
     reactNativeAndroidProject.afterEvaluate {
-        println "Inside afterEvaluate block"
         def resolveTasks = [resolveBoost, resolveGlog, resolveDoubleConversion, resolveFolly]
         resolveTasks.forEach({ t ->
             def reactAndroidDownloadTask = reactNativeAndroidProject.getTasks().findByName("download" + t.name.replace("resolve", ""))
             if (reactAndroidDownloadTask != null) {
-                println "Setting dependency between $t.name & $reactAndroidDownloadTask.name"
                 t.dependsOn(reactAndroidDownloadTask)
             }
         })
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -379,8 +379,8 @@ def resolveTaskFactory(String name, String artifactLocalName, File reactNativeAn
             }
 
             // If it is not the case we check whether it was downloaded by ReactAndroid project
-            def boostTar = Paths.get(reactNativeAndroidDownloadDir.toString(), artifactLocalName).toFile()
-            return boostTar.isFile()
+            def tarFile = Paths.get(reactNativeAndroidDownloadDir.toString(), artifactLocalName).toFile()
+            return tarFile.isFile()
         }
     }
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -394,10 +394,10 @@ def reactNativeAndroidProject = findProject(":ReactAndroid")
 if (reactNativeAndroidProject != null) {
     reactNativeAndroidProject.afterEvaluate {
         def resolveTasks = [resolveBoost, resolveGlog, resolveDoubleConversion, resolveFolly]
-        resolveTasks.forEach({ t ->
-            def reactAndroidDownloadTask = reactNativeAndroidProject.getTasks().findByName("download" + t.name.replace("resolve", ""))
+        resolveTasks.forEach({ task ->
+            def reactAndroidDownloadTask = reactNativeAndroidProject.getTasks().findByName("download" + task.name.replace("resolve", ""))
             if (reactAndroidDownloadTask != null) {
-                t.dependsOn(reactAndroidDownloadTask)
+                task.dependsOn(reactAndroidDownloadTask)
             }
         })
     }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -422,7 +422,6 @@ if (reactNativeAndroidProject != null) {
     println "Failed to find :ReactAndroid project -> explicit dependency between download tasks is not set."
 }
 
-
 task downloadBoost(dependsOn: resolveBoost, type: Download) {
     def transformedVersion = BOOST_VERSION.replace("_", ".")
     def artifactLocalName = "boost_${BOOST_VERSION}.tar.gz"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -401,8 +401,8 @@ if (reactNativeAndroidProject != null) {
             }
         })
     }
-} else {
-    println "Failed to find :ReactAndroid project -> explicit dependency between download tasks is not set. This is fine if you are using Paper architecture."
+} else if (isNewArchitectureEnabled()) {
+    throw new GradleScriptException("[Reanimated] Failed to find `:ReactAndroid` project. Explicit dependency between download tasks cannot be set.")
 }
 
 task downloadBoost(dependsOn: resolveBoost, type: Download) {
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -368,16 +366,56 @@ task createNativeDepsDirectories(dependsOn: applyJavaPatches) {
...
+            if (file("$reanimatedDownloadDir/$artifactLocalName").isFile()) {
+                return false
+            }
+
+            // If it is not the case we check whether it was downloaded by ReactAndroid project
+            def tarFile = Paths.get(reactNativeAndroidDownloadDir.toString(), artifactLocalName).toFile()
+            return tarFile.isFile()
+        }
+    }
+}
```
kkafar, line 386: Created a factory method so the code is not repeated four times. 

To alter task behaviour you can configure it by using extension block later:

```groovy
resolveBoost {
    ...

    doFirst { ... } 
}
```

android/build.gradle
```diff
@@ -368,16 +366,56 @@ task createNativeDepsDirectories(dependsOn: applyJavaPatches) {
...
+def reactNativeAndroidProject = findProject(":ReactAndroid")
+if (reactNativeAndroidProject != null) {
+    reactNativeAndroidProject.afterEvaluate {
+        def resolveTasks = [resolveBoost, resolveGlog, resolveDoubleConversion, resolveFolly]
+        resolveTasks.forEach({ t ->
+            def reactAndroidDownloadTask = reactNativeAndroidProject.getTasks().findByName("download" + t.name.replace("resolve", ""))
+            if (reactAndroidDownloadTask != null) {
+                t.dependsOn(reactAndroidDownloadTask)
+            }
+        })
```
kkafar, line 402: This must be done in `afterEvaluate` block, because otherwise finding tasks of other project might fail if its configuration is not completed yet (tasks may not be added yet).

kkafar (commented):
Some implementation notes.

android/build.gradle
```diff
@@ -368,16 +366,56 @@ task createNativeDepsDirectories(dependsOn: applyJavaPatches) {
...
+
+        onlyIf {
+            // First we check whether the file is already in our download directory
+            if (file("$reanimatedDownloadDir/$artifactLocalName").isFile()) {
+                return false
+            }
+
+            // If it is not the case we check whether it was downloaded by ReactAndroid project
+            def tarFile = Paths.get(reactNativeAndroidDownloadDir.toString(), artifactLocalName).toFile()
+            return tarFile.isFile()
```
tomekzaw, line 383: 1. Let's be more explicit with the logic here.
2. Let's unify the code in both cases.
```suggestion
            if (file("$reactNativeAndroidDownloadDir/$artifactLocalName").isFile()) {
                return true
            }
            
            return false
```

android/build.gradle
```diff
@@ -368,16 +366,56 @@ task createNativeDepsDirectories(dependsOn: applyJavaPatches) {
...
+        }
+    }
+}
+
+Task resolveBoost = resolveTaskFactory("resolveBoost", "boost_${BOOST_VERSION}.tar.gz", reactNativeAndroidDownloadDir, downloadsDir)
+Task resolveDoubleConversion = resolveTaskFactory("resolveDoubleConversion", "double-conversion-${DOUBLE_CONVERSION_VERSION}.tar.gz", reactNativeAndroidDownloadDir, downloadsDir)
+Task resolveFolly = resolveTaskFactory("resolveFolly", "folly-${FOLLY_VERSION}.tar.gz", reactNativeAndroidDownloadDir, downloadsDir)
+Task resolveGlog = resolveTaskFactory("resolveGlog", "glog-${GLOG_VERSION}.tar.gz", reactNativeAndroidDownloadDir, downloadsDir)
+
+def reactNativeAndroidProject = findProject(":ReactAndroid")
```
tomekzaw, line 393: Does `:ReactAndroid` project exist on Paper?

If not, can we wrap this with `if (isNewArchitectureEnabled()) { ... }`?

android/build.gradle
```diff
@@ -368,16 +366,56 @@ task createNativeDepsDirectories(dependsOn: applyJavaPatches) {
...
+if (reactNativeAndroidProject != null) {
+    reactNativeAndroidProject.afterEvaluate {
+        def resolveTasks = [resolveBoost, resolveGlog, resolveDoubleConversion, resolveFolly]
+        resolveTasks.forEach({ t ->
```
tomekzaw, line 397:
```suggestion
        resolveTasks.forEach({ task ->
```

android/build.gradle
```diff
@@ -368,16 +366,56 @@ task createNativeDepsDirectories(dependsOn: applyJavaPatches) {
     thirdPartyNdkDir.mkdirs()
 }
 
-task downloadBoost(dependsOn: createNativeDepsDirectories, type: Download) {
+def resolveTaskFactory(String name, String artifactLocalName, File reactNativeAndroidDownloadDir, File reanimatedDownloadDir) {
```
tomekzaw, line 369:
```suggestion
def resolveTaskFactory(String taskName, String artifactLocalName, File reactNativeAndroidDownloadDir, File reanimatedDownloadDir) {
```

android/build.gradle
```diff
@@ -368,16 +366,56 @@ task createNativeDepsDirectories(dependsOn: applyJavaPatches) {
...
+    return tasks.create(name: name.startsWith("resolve") ? name : "resolve" + name, dependsOn: createNativeDepsDirectories, type: Copy) {
```
tomekzaw, line 370: Why `name.startsWith("resolve") ? name : "resolve" + name` instead of just `name`?

android/build.gradle
```diff
@@ -368,16 +366,56 @@ task createNativeDepsDirectories(dependsOn: applyJavaPatches) {
...
+Task resolveFolly = resolveTaskFactory("resolveFolly", "folly-${FOLLY_VERSION}.tar.gz", reactNativeAndroidDownloadDir, downloadsDir)
+Task resolveGlog = resolveTaskFactory("resolveGlog", "glog-${GLOG_VERSION}.tar.gz", reactNativeAndroidDownloadDir, downloadsDir)
+
+def reactNativeAndroidProject = findProject(":ReactAndroid")
+if (reactNativeAndroidProject != null) {
+    reactNativeAndroidProject.afterEvaluate {
+        def resolveTasks = [resolveBoost, resolveGlog, resolveDoubleConversion, resolveFolly]
+        resolveTasks.forEach({ t ->
+            def reactAndroidDownloadTask = reactNativeAndroidProject.getTasks().findByName("download" + t.name.replace("resolve", ""))
+            if (reactAndroidDownloadTask != null) {
```
tomekzaw, line 399: I think we could remove this `null` check here (because these tasks should exist) but who knows what 0.71 brings us (prefabs)... 🤷

tomekzaw (changes requested):
Thanks for the detailed description. I've left a couple of suggestions in the comments, but it already looks good.

android/build.gradle
kkafar: [86ce034](https://github.com/software-mansion/react-native-reanimated/pull/3588/commits/86ce034d14d8e16e0c8de6f353144e6921b29802)

android/build.gradle
kkafar: [86ce034](https://github.com/software-mansion/react-native-reanimated/pull/3588/commits/86ce034d14d8e16e0c8de6f353144e6921b29802)

android/build.gradle
kkafar: [4049e3b](https://github.com/software-mansion/react-native-reanimated/pull/3588/commits/4049e3b404755c88eb0d9916241987437d4ac4e3)

android/build.gradle
kkafar: No, it does not exist on Paper. 

I wrote it this way on purpose, so it is "architecture neutral" -- I don't really see a reason to split the code with additional conditionals where it can be shared. Unless your main intention is to emphasise that it is architecture-specific.

android/build.gradle
```diff
@@ -368,16 +366,56 @@ task createNativeDepsDirectories(dependsOn: applyJavaPatches) {
...
+                t.dependsOn(reactAndroidDownloadTask)
+            }
+        })
+    }
+} else {
+    println "Failed to find :ReactAndroid project -> explicit dependency between download tasks is not set. This is fine if you are using Paper architecture."
```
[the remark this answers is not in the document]
kkafar: [b3d3a57](https://github.com/software-mansion/react-native-reanimated/pull/3588/commits/b3d3a57cfc0e877a87c1262a461b042f18262045)

android/build.gradle
kkafar: [56df771](https://github.com/software-mansion/react-native-reanimated/pull/3588/commits/56df7718fc50633b23bc91dc6d4ce8821990a487)

kkafar: I actually thought using `file(...)` won't work, because it is method defined on gradle's `Project` type & it resolves files relatively to current project directory, but I guess the case is we are passing absolute path there so it actually works (tested) 🤷🏻‍♂️

android/build.gradle
kkafar: I think it might be fine to throw exception if we wanted to get notified that some behaviour in React Native's internals changed. If it is the case let's make it an explicit Exception with reasonable message instead of NPE. 

Otherwise I would rather leave it this way.

kkafar: How about this: [d737b35](https://github.com/software-mansion/react-native-reanimated/pull/3588/commits/d737b35152df865270fa7d05dc0ba953e33879f9)? 

```groovy
String reactAndroidDownloadTaskName = "download" + task.name.replace("resolve", "")
def reactAndroidDownloadTask = reactNativeAndroidProject.getTasks().findByName(reactAndroidDownloadTaskName)
if (reactAndroidDownloadTask != null) {
    task.dependsOn(reactAndroidDownloadTask)
} else {
     throw new GradleScriptException("[Reanimated] Failed to find task named `$reactAndroidDownloadTaskName` in `:ReactAndroid` project." +
        " Explicit dependency between it and $task.name task can not be set.")
}
```

We could also consider just using `logger.warn` instead of throwing as it does not have to stop the build from succeeding.

android/build.gradle
```diff
@@ -154,6 +151,7 @@ def downloadsDir = customDownloadsDir ? new File(customDownloadsDir) : new File(
 def thirdPartyNdkDir = new File("$buildDir/third-party-ndk")
 
 def reactNativeThirdParty = new File("$reactNativeRootDir/ReactAndroid/src/main/jni/third-party")
+def reactNativeAndroidDownloadDir = new File("$reactNativeRootDir/ReactAndroid/build/downloads")
```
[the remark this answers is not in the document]
kkafar: It should work just fine in case the variable is set.

We also set reanimated's download directory with respect to whether `REACT_NATIVE_DOWNLOADS_DIR` env var is set or not [(link)](https://github.com/facebook/react-native/blob/596111fabf93e65c0a3880a4d6be56368a809ec5/ReactAndroid/build.gradle#L34-L35)
```groovy
def customDownloadsDir = System.getenv("REACT_NATIVE_DOWNLOADS_DIR")
def downloadsDir = customDownloadsDir ? new File(customDownloadsDir) : new File("$buildDir/downloads")
```

So in case the env variable is set, we use the same download directory as React Native, but our download tasks won't start before `:ReactAndroid`'s are over, as there is implicit (through `resolve<lib>` tasks) dependency between them --> they will terminate with `UP-TO-DATE` state.
`resolve<lib>` tasks terminate with `NO-SOURCE` state in case "from" directory does not exist (they do not throw any exception).

In case there is no `:ReactAndroid` project (Paper) it also should work fine.

android/build.gradle
tomekzaw: When it comes to Gradle scripts I think it's always better to be explicit 😄 If it's on purpose, then it's okay, but then some comment here would be valuable.

android/build.gradle
tomekzaw: I think `logger.warn` is a better choice indeed.

kkafar: [345a009](https://github.com/software-mansion/react-native-reanimated/pull/3588/commits/345a0091598dbd76dff393cd6c9d857065da331d)

PR #3589, opened 2022-09-20
Add ranges in version labels in docs

j-piasecki:
## Description

The current convention for making a new version in docs is to make a new one when there have been API changes. This is a good solution as it doesn't result in redundant versions of docs that are exactly the same but has one major flaw: it's not obvious what it means. This PR updates the labels to reflect that they cover more than one version of Reanimated.

## Changes

- Added custom labels for docs versions covering more than one minor version

### Before

Screenshot before

### After

Screenshot after

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -136,7 +136,7 @@ module.exports = {
           editUrl: undefined, // hide edit button
           versions: {
             '2.5.x': {
-              label: '2.5.x – 2.9.x',
+              label: '2.5.x – 2.10.x',
             },
             '2.3.x': {
               label: '2.3.x – 2.4.x',
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -136,10 +136,10 @@ module.exports = {
           editUrl: undefined, // hide edit button
           versions: {
             '2.5.x': {
-              label: '2.5.x-2.9.x',
+              label: '2.5.x – 2.9.x',
             },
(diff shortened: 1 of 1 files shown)


docs/docusaurus.config.js
```diff
@@ -134,6 +134,14 @@ module.exports = {
           path: 'docs',
           sidebarPath: require.resolve('./sidebars.js'),
           editUrl: undefined, // hide edit button
+          versions: {
+            '2.5.x': {
+              label: '2.5.x-2.9.x',
```
tomekzaw, line 139: Let's use `&ndash;` (–) for a better visual effect.
```suggestion
              label: '2.5.x – 2.9.x',
```

j-piasecki: Done in https://github.com/software-mansion/react-native-reanimated/pull/3589/commits/88c249acf55004907232b609067313d6a0d1462a

docs/docusaurus.config.js
```diff
@@ -134,6 +134,14 @@ module.exports = {
           path: 'docs',
           sidebarPath: require.resolve('./sidebars.js'),
           editUrl: undefined, // hide edit button
+          versions: {
+            '2.5.x': {
+              label: '2.5.x – 2.9.x',
```
piaskowyk, line 139:
```suggestion
              label: '2.5.x – 2.10.x',
```

piaskowyk (approved):
😍

PR #3598, opened 2022-09-22
Show more useful message when using `measure()` in `useAnimatedStyle()`

Kwasow:
## Description

When `measure()` was called inside `useAnimatedStyle()` a warning would appear with the message:

`[Reanimated] measure() cannot be used for web or Chrome Debugger`.

This message isn't very helpful, since we aren't running Chrome Debugger and aren't on web. This was previously shown because `useAnimatedStyle()` runs an initial update on the JS thread.

## Changes

When measure is called from the JS thread, but on native platforms, then show a message that explains why this could be happening.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [x] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/docs/docs/api/nativeMethods/measure.md
+++ b/docs/docs/api/nativeMethods/measure.md
@@ -21,14 +21,15 @@ new Promise((resolve, reject) => {
 });
 ```
 
-If you call `measure()` inside of [`useAnimatedStyle()`](../hooks/useAnimatedStyle)
+If you call `measure()` inside [`useAnimatedStyle()`](../hooks/useAnimatedStyle)
 you may get a warning that `measure()` was called from the wrong thread. This
-is safe to ignore, but it you don't want this error to appear then wrap the call
+is safe to ignore, but if you don't want this error to appear then wrap the call
 like this:
 
 ```js
 if (_WORKLET) {
   const measure = measure(animatedRef);
+  // ...
 }
 ```
 
--- a/docs/docs/api/nativeMethods/measure.md
+++ b/docs/docs/api/nativeMethods/measure.md
@@ -21,6 +21,17 @@ new Promise((resolve, reject) => {
 });
 ```
 
+If you call `measure()` inside of [`useAnimatedStyle()`](../hooks/useAnimatedStyle)
+you may get a warning that `measure()` was called from the wrong thread. This
+is safe to ignore, but it you don't want this error to appear then wrap the call
+like this:
+
+```js
+if (_WORKLET) {
+  const measure = measure(animatedRef);
+}
+```
+
 ### Arguments
 
 #### animatedRef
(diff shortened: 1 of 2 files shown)


docs/docs/api/nativeMethods/measure.md
```diff
@@ -21,6 +21,17 @@ new Promise((resolve, reject) => {
 });
 ```
 
+If you call `measure()` inside of [`useAnimatedStyle()`](../hooks/useAnimatedStyle)
+you may get a warning that `measure()` was called from the wrong thread. This
+is safe to ignore, but it you don't want this error to appear then wrap the call
```
tomekzaw, line 26:
```suggestion
is safe to ignore, but if you don't want this error to appear then wrap the call
```

docs/docs/api/nativeMethods/measure.md
```diff
@@ -21,6 +21,17 @@ new Promise((resolve, reject) => {
...
+like this:
+
+```js
+if (_WORKLET) {
+  const measure = measure(animatedRef);
```
tomekzaw, line 31:
```suggestion
  const measure = measure(animatedRef);
  // ...
```

src/reanimated2/NativeMethods.ts
```diff
@@ -17,11 +17,18 @@ export function measure(
   animatedRef: RefObjectFunction<Component>
 ): MeasuredDimensions | null {
   'worklet';
-  if (!_WORKLET || !isNative) {
+  if (!isNative) {
     console.warn(
       '[Reanimated] measure() cannot be used for web or Chrome Debugger'
```
tomekzaw, line 22:
```suggestion
      '[Reanimated] measure() cannot be used on web or when using Chrome Debugger'
```

src/reanimated2/NativeMethods.ts
```diff
@@ -17,11 +17,18 @@ export function measure(
...
     );
     return null;
+  } else if (!_WORKLET) {
+    console.warn(
+      '[Reanimated] measure() was called from the wrong thread. This may happen ' +
+        'if you called measure() in the useAnimatedStyle hook. If you want to' +
+        'prevent this warning then wrap the call with `if (_WORKLET)`.'
```
tomekzaw, line 29: 1. Instead of "wrong thread" use something like "main JS context".
2. Explain that `measure` is only available in the UI runtime and cannot be used from main JS context.
3. Technically, it's the worklet inside `useAnimatedStyle`, not the hook itself.
4. Explain that `useAnimatedStyle` first calls given worklet from main JS runtime during render.
5. Explain why someone needs to add `if (_WORKLET)` (so that `measure` is called only from the UI context).

src/reanimated2/NativeMethods.ts
```diff
@@ -17,11 +17,18 @@ export function measure(
...
+    );
+    return null;
```
tomekzaw, line 31: What about adding `function measure() { return null; }` on the main JS context?

tomekzaw, line 31:
```suggestion
    return null; // always return `null` when called from main JS runtime
```

piaskowyk (approved):

PR #3566, opened 2022-09-12
Raise error when using Reanimated 2.x with Fabric enabled

tomekzaw:
## Description

Android:

iOS:

## Checklist

- [ ] Included code example that can be used to test this change
- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Added unit / integration tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

PR #3475, opened 2022-08-13
Release 2.10.0

piaskowyk:
## Descriptio
Temporary PR

PR #3608, opened 2022-09-26
[Rea2] Restore using prebuilt binaries

piaskowyk:
## Description

We used the Gradle script from the main branch (Reanimated 3) for Reanimated 2. However, Reanimated3's Gradle script doesn't use prebuilt binaries from the Reanimated package. I restored the Gradle code from the 2.9.1 version.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -109,6 +109,100 @@ def getPlaygroundAppName() { // only for the development
     return playgroundAppName
 }
 
+def detectAAR(minor, engine) {
+    def rnMinorVersionCopy = Integer.parseInt(minor)
+    def aar = file("react-native-reanimated-${rnMinorVersionCopy}-${engine}.aar")
+
+    if (aar.exists()) {
+        println "AAR for react-native-reanimated has been found\n$aar"
+        return aar
+    } else {
+        while (!aar.exists() && rnMinorVersionCopy >= 63) {
+            rnMinorVersionCopy -= 1
+            aar = file("react-native-reanimated-${rnMinorVersionCopy}-${engine}.aar")
+        }
+
+        if (rnMinorVersionCopy < 63) {
+            println "No AAR for react-native-reanimated found. Attempting to build from source."
+        } else { // aar exists, but was build for lower react-native version
+            println "\n\n\n"
+            println "****************************************************************************************"
+            println "\n\n\n"
+            println "WARNING reanimated - no version-specific reanimated AAR for react-native version $minor found."
+            println "Falling back to AAR for react-native version $rnMinorVersionCopy."
+            println "The react-native JSI interface is not ABI-safe yet, this may result in crashes."
+            println "Please post a pull request to implement support for react-native version $minor to the reanimated repo."
+            println "Thanks!"
+            println "\n\n\n"
+            println "****************************************************************************************"
+
+            return aar
+        }
+    }
+    return null
+}
+
+def detectReanimatedConfig() {
+    def buildFromSourceConf = false
+    rootProject.getSubprojects().forEach({project ->
+        if (project.plugins.hasPlugin("com.android.application")) {
+            if (
+            project.ext.has("reanimated")
+                    && project.ext.reanimated.buildFromSource
+            ) {
+                buildFromSourceConf = true
+            }
+        }
+    })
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -174,22 +321,6 @@ def CMAKE_NODE_MODULES_DIR = CLIENT_SIDE_BUILD
...
-
-    // Detect js runtime from project setup
-    def defaultRuntimeType = "jsc";
-    def v8Project = rootProject.getSubprojects().find { project -> project.name == "react-native-v8" }
-    if (v8Project != null) {
-        return "v8"
-    }
-    def appProject = findProject(":app") ? project(":app") : null
-    return appProject?.ext?.react?.enableHermes ? "hermes" : defaultRuntimeType
-}.call()
```
kkafar, line 191: Havent tested the behaviour, but code wise it looks fine. 

One note: do we want to remove support for v8?

kkafar (commented):
LGTM 👍🏻

android/build.gradle
kkafar: Also these changes: https://github.com/software-mansion/react-native-reanimated/pull/3475

kkafar: And these: https://github.com/software-mansion/react-native-reanimated/pull/3566

android/build.gradle
```diff
@@ -737,7 +868,7 @@ afterEvaluate {
 
 if (CLIENT_SIDE_BUILD) {
```
tomekzaw, line 869: Are `CLIENT_SIDE_BUILD` and `BUILD_FROM_SOURCE` somehow related? If yes, can we use only `BUILD_FROM_SOURCE`?

android/build.gradle
```diff
@@ -737,7 +868,7 @@ afterEvaluate {
...
     def aarDir = "${buildDir}/outputs"
-    def aar = file("${aarDir}/android-debug.aar")
+    aar = file("${aarDir}/android-debug.aar")
```
tomekzaw, line 871: What about release mode? Shouldn't we use `android-release.aar` then?

android/build.gradle
```diff
@@ -109,6 +109,100 @@ def getPlaygroundAppName() { // only for the development
...
+            println "No AAR for react-native-reanimated found. Attempting to build from source."
+        } else { // aar exists, but was build for lower react-native version
+            println "\n\n\n"
+            println "****************************************************************************************"
+            println "\n\n\n"
+            println "WARNING reanimated - no version-specific reanimated AAR for react-native version $minor found."
+            println "Falling back to AAR for react-native version $rnMinorVersionCopy."
+            println "The react-native JSI interface is not ABI-safe yet, this may result in crashes."
+            println "Please post a pull request to implement support for react-native version $minor to the reanimated repo."
+            println "Thanks!"
```
tomekzaw, line 135:
```suggestion
            println "The react-native JSI interface is not ABI-safe, using incompatible versions may result in crashes."
            println "Please submit a pull request in react-native-reanimated repository to implement support for react-native 0.$minor"
            println "Thanks!"
```

android/build.gradle
```diff
@@ -109,6 +109,100 @@ def getPlaygroundAppName() { // only for the development
...
+            println "\n\n\n"
+            println "****************************************************************************************"
+
+            return aar
+        }
+    }
+    return null
+}
+
+def detectReanimatedConfig() {
```
tomekzaw, line 145: Can we rename this method to something more specific, i.e. `shouldBuildFromSource`?

android/build.gradle
```diff
@@ -109,6 +109,100 @@ def getPlaygroundAppName() { // only for the development
...
+    else if (isDeveloperMode()) {
+        // Example app
+        return true
+    }
+    else if (detectReanimatedConfig()) {
+        // on user demand
+        return true
+    }
+    else if (aar != null) {
+        // when binary exist
```
tomekzaw, line 173:
```suggestion
        // when binary exists
```

tomekzaw (changes requested):

android/build.gradle
tomekzaw: btw why `minor` and not `rnMinorVersionCopy`?

PR #3612, opened 2022-09-27
Cherry-pick reanimated v3 podspec changes to v2 to fix warnings

AlexanderEggers:
## Description

This PR fixes the following warning when running `pod install` using the latest stable v2 (2.10.0).

`[!] Can't merge user_target_xcconfig for pod targets: ["RNReanimated", "hermes-engine"]. Singular build setting CLANG_CXX_LANGUAGE_STANDARD has different values.`

The actual fix is ported from the v3 (main) branch which already fixed that issue.

## Changes

- Move CLANG_CXX_LANGUAGE_STANDARD as part of the pod target xconfig section.
   - Note: I have not updated the CLANG_CXX_LANGUAGE_STANDARD value to c++17. I am uncertain if that is something which we should do as part of this PR.
- Add compiler flag to enable the hermes debugger.

### DIFF
--- a/RNReanimated.podspec
+++ b/RNReanimated.podspec
@@ -94,11 +94,11 @@ Pod::Spec.new do |s|
 
   s.pod_target_xcconfig    = {
     "USE_HEADERMAP" => "YES",
-    "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/ReactCommon\" \"$(PODS_TARGET_SRCROOT)\" \"$(PODS_ROOT)/#{folly_prefix}Folly\" \"$(PODS_ROOT)/boost\" \"$(PODS_ROOT)/boost-for-react-native\" \"$(PODS_ROOT)/DoubleConversion\" \"$(PODS_ROOT)/Headers/Private/React-Core\" "
-  }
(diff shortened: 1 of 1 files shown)


AlexanderEggers:
This PR is not needed anymore with the release of 2.11.0.

tomekzaw:
Hey @AlexanderEggers, thanks for the PR and sorry for late response.

> This PR is not needed anymore with the release of 2.11.0.

Could you point out which change made this PR redundant?

FYI I'm still getting some warnings when installing Pods in ReanimatedExample app in our repo:

Zrzut ekranu 2022-10-20 o 12 15 53

AlexanderEggers:
@tomekzaw I'm actually not certain which part is directly fixing that but I did a test with the latest 2.x (without the local patch) and had no warnings anymore.
I can try to have a closer look on the weekend if that helps?

tomekzaw:
@AlexanderEggers thanks for your response.

> I can try to have a closer look on the weekend if that helps?

Since the original issue is gone I think it won't be necessary but thanks 😄 

I will try to fix these warnings in the Example app, maybe then we'll check if it doesn't break regular apps.

PR #3626, opened 2022-09-30
Improve globals.d.ts

tomekzaw:
## Description

This PR eliminates differences between two sections in `globals.d.ts`. Also, two types needed to be fixed.


piaskowyk (approved):
😍

PR #3646, opened 2022-10-05
Add WebExample

tomekzaw:
## Description

TL;DR This PR adds WebExample app which uses Expo for Web.

There is a problem when WebExample components use a different version or even different instance of the same version of React than Reanimated does. It works fine when using 2.9.1 or even 3.0.0-rc.3 but fails when using `"react-native-reanimated": "link:../"`.

One solution is to remove `"react": "17.0.2"` from `devDependencies` in `package.json` in the root directory. WebExample works fine, `npm ls react` shows only deduped 18.1.0 (which is correct), but running ESLint in root directory fails because it cannot find `react` module.

Other solution would to remove `"react-native-reanimated": "link:../"` and try linking module manually but after multiple attempts with `babel.config.js` I still wasn't able to achieve it.

For some reason, despite the fact that WebExample has a direct dependency on `"react": "18.0.0"`, it uses the instance from `node_modules` in the root directory. So as a workaround I've added a module resolver in `babel.config.js`.

Fast reload also works.

## Changes

- Added WebExample app

```sh
yarn
cd WebExample
yarn
yarn web
```

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


piaskowyk (approved):
🚀

PR #3648, opened 2022-10-05
[Reanimated2] Sync CI with main

piaskowyk:
## Description

Sync CI versions from the main branch with the Reanimated2 branch.


.github/workflows/android-build.yml
```diff
@@ -11,13 +11,15 @@ on:
...
   build:
     runs-on: ubuntu-latest
-    env:
-      WORKING_DIRECTORY: Example
+    strategy:
+      matrix:
+        working-directory: [Example]
     concurrency:
-      group: android-${{ github.ref }}
+      group: rea2-android-${{ matrix.working-directory }}-${{ github.ref }}
```
tomekzaw, line 22: It looks like `github.ref` is sufficient to distinguish v3 from v2 as it contains the branch name or the number of PR, which is bound to a specific version, so we can remove `rea2-` here.

.github/workflows/tv-os-build.yml
```diff
@@ -12,11 +12,11 @@ on:
   workflow_dispatch:
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: macos-12
```
tomekzaw, line 15: 1. Rename this file to `tvos-build.yml` (without dash)
2. Apply changes from main (this version is outdated)

.github/workflows/ios-build.yml
```diff
@@ -11,14 +11,15 @@ on:
   push:
     branches:
       - Reanimated2
+
```
tomekzaw, line 14: This file is also outdated (e.g. there are only 4 paths while on `main` there are 8). Please use current version from `main` branch.

tomekzaw (changes requested):
It works but some of the files are still different when compared to main.

.github/workflows/ios-build.yml
```diff
@@ -1,41 +1,71 @@
...
+      - .github/workflows/ios-build.yml
+      - RNReanimated.podspec
+      - ios/**
+      - Common/**
+      - Example/package.json
+      - Example/ios/**
   push:
     branches:
-      - Reanimated2
+      - main
```
tomekzaw, line 13:
```suggestion
      - Reanimated2
```

tomekzaw (changes requested):

tomekzaw (approved):

PR #3650, opened 2022-10-06
Add Next.js example to check bundle size on web

tomekzaw:
## Description

This PR adds NextExample app (based on https://github.com/nandorojo/reanimated-tree-shaking from @nandorojo) which can be used to check Reanimated bundle size as well as its tree-shaking capabilities when used on web (specifically, in Next.js project).

The motivation behind this PR is a significant increase (~10x) of bundle size on web when using Reanimated 3.0.0-rc.3 compared to 3.0.0-rc.2 caused by Babel plugin appending worklets with sourcemaps as comments for debugging worklet runtime with Flipper which is not necessary on web.

| Reanimated version | only Animated.View | with sample animation* |
|:-:|:-:|:-:|
| 3.0.0-rc.2 | 17.3 kB | 27.5 kB |
| 3.0.0-rc.3 | 126 kB ⚠️ | 224 kB ⚠️ |
| this PR | 17.3 kB ✅ | 27.2 kB ✅ |

\* - Animated.View, useSharedValue, useAnimatedStyle, withTiming, withRepeat

TODO:
- [ ] properly disable generating sourcemaps on web
- [x] confirm that sourcemaps are not necessary in dev mode
- [ ] `"react-native-reanimated": "link:../"` instead of `3.0.0-rc.3` to make local development easier
- [ ] check if fast refresh in core.ts works properly
- [ ] use SWC plugin
- [ ] show package size as CI output


NextExample/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "start": "next start",
+    "analyze": "ANALYZE=true next build",
+    "cache:clear": "rm -rf .expo .next node_modules/.cache"
+  },
+  "dependencies": {
+    "next": "latest",
+    "raf": "^3.4.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-native-reanimated": "3.0.0-rc.3",
```
tomekzaw, line 15: Is there any option to use `link:../` here so we can have fast refresh when we edit package files? cc @nandorojo

.github/workflows/check-bundle-size.yml
```diff
@@ -0,0 +1,46 @@
...
+      - .github/workflows/check-bundle-size.yml
+      - 'src/**'
+      - 'NextExample/**'
+      - '*'
+  push:
+    branches:
+      - main
+
+env:
+  MAX_SIZE_KB: 30
```
tomekzaw, line 14: The plan is that we will increase this value accordingly if needed.

plugin.js
```diff
@@ -282,11 +282,19 @@ function isRelease() {
   return process.env.BABEL_ENV === 'release';
 }
 
+function isProduction() {
+  return process.env.NODE_ENV === 'production'; // for web
```
tomekzaw, line 286: Is there any better way to detect production build for web?

NextExample/package.json
nandorojo: an easier method is probably to un-comment the `webpack` config resolver in `next.config.js` and point it to the right place. but `link` might also work. i assume this is using `yarn link`? i don’t think i’ve tried this before

plugin.js
nandorojo: this is correct

NextExample/app.json
```diff
@@ -0,0 +1,4 @@
+{
```
nandorojo, line 1: I don’t think this file is necessary

NextExample/babel.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
nandorojo, line 1: I wonder if we should use the reanimated SWC plug-in instead of Babel? Or perhaps do a run with and a run without it, using an env variable?

For reference: https://github.com/showtime-xyz/showtime-frontend/blob/staging/apps/next/next.config.js#L91-L99

This isn’t a requirement for this PR, but would be useful for testing.

NextExample/next.config.js
```diff
@@ -0,0 +1,39 @@
...
+module.exports = compose(
+  [withTM, withNextBundleAnalyzer, [withExpo, { projectRoot: __dirname }]],
+  /** @type {import('next').NextConfig} */
+  {
+    webpack5: true,
+    eslint: {
+      ignoreDuringBuilds: true,
+    },
+    webpack(config, _options) {
+      // config.resolve.alias = {
```
nandorojo, line 28: I think you could make this point to the locally-built folder.

plugin.js
nandorojo: that said, should these source maps still get generated for web, even in dev mode?

nandorojo:
Hey, just checking on the status of this PR :)

plugin.js
tomekzaw: Indeed, they don't need to be generated even in dev mode.

Here's a screenshot of debugger stopping on a breakpoint in a worklet without sourcemaps (there's no `sourceMappingURL` in `_f.asString`):

NextExample/babel.config.js
tomekzaw: Good idea, we can test the SWC plugin as well once it's merged (in a separate PR).

NextExample/package.json
tomekzaw: Does this approach support fast-refresh also when changing Reanimated codebase, e.g. `src/reanimated2/core.ts`?

piaskowyk:
@bartlomiejbloniarz works on the new version of this test.

PR #3652, opened 2022-10-06
Common Gradle for Reanimated 2 and 3

piaskowyk:
## Description

We often update the Gradle script. To make it simpler I prepared one common Gradle script for both versions Reanimated 2 and 3.


tomekzaw (approved):
🚀

PR #3655, opened 2022-10-07
fix: BABEL_ENV should consider using production

jinliming2:
## Changes

BABEL_ENV should consider using production

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -279,7 +279,7 @@ class ClosureGenerator {
 }
 
 function isRelease() {
-  return process.env.BABEL_ENV === 'release';
+  return ['production', 'release'].includes(process.env.BABEL_ENV);
 }
 
 function shouldGenerateSourceMap() {


tomekzaw (commented):
Thanks for the PR.

Starting from 3.0.0-rc.3, Reanimated Babel plugin adds sourcemaps to worklets. In my case, this increases bundle size on web to 225 kB (compared to 27.6 kB without sourcemaps).

In #3650 I've proposed using `process.env.NODE_ENV === 'production'` to detect production build on web and it works for me (checked on Next.js + Reanimated example).

However, with this PR, the bundle still contains sourcemaps. In my case, `production.env.BABEL_ENV` is `undefined`.

Could you please describe which case is resolved by this PR?

jinliming2:
@tomekzaw I'm using Re.Pack to build my RN app, which uses Webpack under the hood. In this case, `BABEL_ENV` was default set to 'production' when build in release mode.

tomekzaw:
Great, thanks for information! So I think we can merge this.

tomekzaw (approved):

piaskowyk (approved):
🚀

Issue #3124, opened 2022-04-01
[ios][bug] Huge issue with LayoutAnimation - Single `Animated.View` with `layout` prop is enough to break un-mounting and de-allocation of any View which has been popped/swiped-back/closed/destroyed via navigation


hirbod:
## Description

@piaskowyk @kmagiera

I have found a serious bug with LayoutAnimation. I have been trying to investigate the problem for several days, which is why videos that were in a screen or modal that I closed (goBack/pop/swipe to dismiss) still played even though everything was un-mounted and destroyed. The same is true for cameras (Green Light Indicator is no longer out).

It is sufficient to have a single `Reanimated.View` / `createAnimatedComponent` ANYWHERE in the app which uses `layout, entering or exiting` prop from Reanimated. Once a single `View` with that prop has been mounted (no matter where), no element that is in a navigation stack will be unmounted anymore, even if the layout hierarchy and lifecycle methods reports that it has been unmounted (eg cleanup function of `useEffect`). When you remove something conditionally on the screen (eg. with `useState`), un-mounting works fine. But it looks like a single use of `LayoutAnimation` is enough to break element un-mounting / de-allocation when closing Navigators, because (this is my guess) LayoutAnimation is still "watching all of them (or snapshotting?)" in order to be able to animate the layout, even when those views are in different screens. Maybe RNS is also an issue here?

I also did some tests, such as setting breakpoints at the deallocate function of expo-av. The methods were never called when `layout, entering` or `exiting` were present and mounted anywhere in the app. But if no `Reanimated.View` with `layout` was used the breakpoints triggered.

I have also created a video as well as a repro which can be tested with Expo Go.

### Expected behavior

Views should correctly un-mount and deallocate once a Navigation screen has been dismissed / popped etc, even when using `LayoutAnimation`.

### Actual behavior & steps to reproduce

To test my findings, simply grab my repro, spin it up, open Tab One and tap on the info icon at the top (headerRight) - as many times as you like. The video will start and stop as soon as the modal has been closed. But if you now switch to Tab two, which mounts a `Reanimated.View` with `layout={Layout}`, (imported from REA) and you now simply return to Tab one, you will notice that the video will continue to run as soon as you open and close the modal and it will start multiple videos until the App crashes eventually. If you remove `layout` from Tab two, you can open and close the modal 1000x and it will work fine. Please note that you need to restart the app when you add or remove `layout`, fast-refresh is not enough.

[...]

PR #3687, opened 2022-10-17
Use a weak map for storing view references

erik-kallen:
## Description

Sometimes the `notifyAboutEnd` method on `REAnimationsManager` is not called for some views. I don't know 100% exactly what causes this, but in our case it was triggered by a combination of react-navigation, react-native-screens and react-native-reanimated.

The effect for us was that whenever we pop'd the stack in react-navigation, we would leak all UIViews that were currently being displayed.

I don't know if this PR is the correct fix (for example it will still leak at least the entries in the _states dictionary), but it prevents the views from leaking, and I have not found any bad effects.

## Changes

Use a `[NSMapTable strongToWeakObjectMapTable] to store the mapping between react tags and UIView *.

### DIFF
--- a/ios/LayoutReanimation/REAAnimationsManager.m
+++ b/ios/LayoutReanimation/REAAnimationsManager.m
@@ -22,7 +22,7 @@ typedef NS_ENUM(NSInteger, FrameConfigType) { EnteringFrame, ExitingFrame };
   RCTUIManager *_uiManager;
   REAUIManager *_reaUiManager;
   NSMutableDictionary<NSNumber *, NSNumber *> *_states;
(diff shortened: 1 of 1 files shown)


hirbod:
This is a known issue
https://github.com/software-mansion/react-native-reanimated/issues/3124

And will most likely be fixed with https://github.com/software-mansion/react-native-reanimated/pull/3332

piaskowyk:
Thanks for your PR and commitment - I appreciate that 🙌 But in Reanimated 3 we have rewritten the whole logic behind Layout Animation so I think that those changes are no longer relevant.

Issue #3568, opened 2022-09-13
useAnimatedKeyboard native iOS crash on keyboardDidHide


MichaelDanielTom:
### Description

When using the `useAnimatedKeyboard` hook, there are occasional crashes when the keyboard is shown/hidden on iOS. Here is the stack objc stack and location of the error within the screenshot.

![image](https://user-images.githubusercontent.com/5520838/189823336-c7fced2c-7c3a-41d3-9309-b4204787d2f2.png)

### Steps to reproduce

I have been having a hard time reproducing it deterministically, however anecdotally it seems to occur after the device/simulator has been on for at least a few minutes. It can happen when the keyboard is closing or opening. 

### Snack or a link to a repository

n/a

### Reanimated version

2.10.0

### React Native version

0.66.0

### Platforms

iOS

### Acknowledgements

Yes

MingaudasVagonis:
This seems to happen quite frequently when navigating to a screen with `autoFocus` on one of the inputs enabled. Happens 1 out 5 times or so. 
<details>
<summary>Quite easy to reproduce with this simple component</summary>

```
import React, { FC, memo } from 'react';
import Animated, {
  useAnimatedKeyboard,
  useAnimatedStyle,
} from 'react-native-reanimated';

type KeyboardFillerProps = {
  offset?: number;
};

export const KeyboardFiller: FC<KeyboardFillerProps> = memo(({ offset }) => {
  const { height } = useAnimatedKeyboard();

  const placeholderStyle = useAnimatedStyle(
    () => ({
      height: height.value - (offset || 0),
    }),
    [offset],
  );

  return <Animated.View style={placeholderStyle} />;
});

KeyboardFiller.displayName = 'KeyboardFiller';
```

</details>

gtokman:
I'm also seeing an intermittent crash. 

@MingaudasVagonis I don't think that `autoFocus` is the reason. I'm not using that prop.

ChildishForces:
I had a similar bug, and solved it by doing the following:

```ts
const animatedStyle = useAnimatedStyle(() => ({
  transform: [{ translateY: -(keyboard?.height?.value ?? 0) })],
});
```

It appears it is possible for the value to not be a number under certain conditions. I don't know this for a fact, but doing this solved my issue based off my hunch.

PR #3704, opened 2022-10-21
Fixed incorrect value for height and state of useAnimatedKeyboard on iOS, fixed occasional crash on dismount on iOS

MichaelDanielTom:
## Description

Include Fixes for #3568 on iOS, and for incorrect initial values for both the `height` and `state` output of `useAnimatedKeyboard` that occasionally will occur on iOS 16 if the found keyboard view is not actually viewable.

### DIFF
--- a/ios/keyboardObserver/REAKeyboardEventObserver.m
+++ b/ios/keyboardObserver/REAKeyboardEventObserver.m
@@ -99,13 +99,21 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
 
   CGFloat keyboardHeight = [self computeKeyboardHeight:keyboardView];
   for (NSString *key in _listeners.allKeys) {
-    ((KeyboardEventListenerBlock)_listeners[key])(_state, keyboardHeight);
+    KeyboardEventListenerBlock listener = _listeners[key];
+    if (listener) {
+      listener(_state, keyboardHeight);
+    }
   }
 }
 
 - (CGFloat)computeKeyboardHeight:(UIView *)keyboardView
 {
-  CGFloat keyboardFrameY = [keyboardView.layer presentationLayer].frame.origin.y;
+  CALayer *presentationLayer = [keyboardView.layer presentationLayer];
+  BOOL hasSize = presentationLayer.frame.size.width && presentationLayer.frame.size.height;
+  if (!hasSize) {
+    return 0;
+  }
+  CGFloat keyboardFrameY = presentationLayer.frame.origin.y;
   CGFloat keyboardWindowH = keyboardView.window.bounds.size.height;
   CGFloat keyboardHeight = keyboardWindowH - keyboardFrameY;
   return keyboardHeight;


ios/keyboardObserver/REAKeyboardEventObserver.m
```diff
@@ -99,13 +99,21 @@ - (void)updateKeyboardFrame
...
   }
 }
 
 - (CGFloat)computeKeyboardHeight:(UIView *)keyboardView
 {
-  CGFloat keyboardFrameY = [keyboardView.layer presentationLayer].frame.origin.y;
+  CALayer *presentationLayer = [keyboardView.layer presentationLayer];
+  BOOL hasSize = presentationLayer.frame.size.width && presentationLayer.frame.size.height;
+  if (!hasSize) {
+    return 0;
```
tomekzaw, line 114: Just wondering if it's safe to return 0 here.

tomekzaw (commented):
Hey @MichaelDanielTom, thanks for the PR!

gtokman:
@piaskowyk would it be possible to review? It fixes the crash reported by numerous folks from #3568

piaskowyk:
Hey @MichaelDanielTom 👋 Thanks for the PR! 
Could you explain to me why this solves your problem? According to the code `_listeners` shouldn't contain nullable objects 🤔 I found a similar problem caused by data race [#3712](https://github.com/software-mansion/react-native-reanimated/issues/3712) I think that we should add synchronization on access to `_listeners`. What do you think? Could you check if synchronization solves your problem or could you provide a reproduction of your issue that I could run on my machine?

gtokman:
@piaskowyk I agree this may not be the cause of the crash. I was using a patch with the changes from this PR and am still experiencing the crash. 

I think we should follow a similar pattern to what was used in [React Native Keyboard Controller](https://github.com/kirillzyusko/react-native-keyboard-controller/blob/main/ios/KeyboardControllerModule.mm) package.

ios/keyboardObserver/REAKeyboardEventObserver.m
dmitriiivashko: I believe the `hasSize` checko should be moved to around [line 42](https://github.com/software-mansion/react-native-reanimated/pull/3704/files#diff-1894e21e8a2a03b928b81d9663210ce7061585dcc87a2c2712cd49b0c4d7c018R42)

I had this false positive when this function returned `0`, resulting in content being stuck behind the keyboard. After investigating, it turned out the views stack had multiple `UIInputSetHostView` instances, and only one matched the `hasSize` condition. The current code picks up the first `UIInputSetHostView`. However, it should pick the first one that has width and height. Otherwise, it just puts the first `UIInputSetHostView` into `_keyboardView`, and if it is irrelevant and has a zero size, the UI is stuck with `height` = `0`, and the only way to fix it is to re-mount the native view and hope this algo picks the correct view next time. `hasSize` here cures symptoms, but not the problem.

piaskowyk:
Thanks for everyone for the help! 🙌 This issue should be fixed by https://github.com/software-mansion/react-native-reanimated/pull/3911 PR

Issue #2824, opened 2022-01-11
Potential memory leak with arrays in shared values


vabruzzese:
## Description

I'm noticing a spike in memory usage over time when updating a shared value in `react-native-reanimated-2`. I suspect that it may be an issue in C++ when deleting array pointer references.

### Expected behavior

Memory usage should stay the same over time. I left this code running for 10 minutes and checked the memory usage through xcode and it was much higher.

![Screen Shot 2022-01-11 at 4 14 12 PM](https://user-images.githubusercontent.com/11095678/149024081-3e9650e9-1a1a-417d-a5c8-4aeab644f72a.png)

Shareable value pointers should be cleared up when re-assigning the value but instead we are seeing 1131065 shared value pointer references:

![Screen Shot 2022-01-11 at 4 20 10 PM](https://user-images.githubusercontent.com/11095678/149024244-ab0d9591-3bbf-4379-9b55-c20a886949e6.png)

### Actual behavior & steps to reproduce

Create a react-native-reanimated-2 playground that reassigns a shared value to an object with an array and set an interval to continuously update it

## Snack or minimal code example

```
import {useSharedValue} from 'react-native-reanimated';
import {View, Text} from 'react-native';
import React, {useEffect, useState} from 'react';

export default function AnimatedStyleUpdateExample(props) {
  const [obj, setObj] = useState();

  const randomObject = () => {
    const arr = [];
    for (let i = 0; i < 10000; i++) {
      arr.push({
        key: i,
        value: Math.random(),
      });
    }

    return {
      id: Math.random(),
      arr,
    };
  };

  const test = useSharedValue();

  useEffect(() => {
    const interval = setInterval(() => {
      setObj(randomObject());
    }, 500);

    return () => clearInterval(interval);
  }, []);

  useEffect(() => {
    test.value = obj;
  }, [obj, test]);

  return (
    <View style={{flex: 1, justifyContent: 'center', alignItems: 'center'}}>
      <Text>Hello World</Text>
      <Text>{test.value ? test.value.id : ''}</Text>
    </View>
  );
}
```

## Package versions

| name                                 | version |
| ------------------------------------ | ------- |
| react-native                         |   0.65.1      |
| react-native-reanimated              |       2.3.1  |
| NodeJS                               |     17.2.0    |
| Xcode                                |   13.2.1      |
| Java                                 |    |
| Gradle                               |         |
| expo                                 |         |

## Affected platforms

- [ ] Android
- [X] iOS
[...]

PR #3722, opened 2022-10-26
Shareable value rewrite

kmagiera:
## Description

This PR contains a new implementation of the "shareable value" concept – one of the lowest level key concepts behind shared values/worklets etc.

There were many reasons for attempting this rewrite, but the main objective was to address memory-related issues rooted in that core part and difficult to fix due to an inherent complexity of that bit of the codebase. Below, I list a couple of reasons providing some more details description:

1. The aforementioned memory issues can be best noted on #2824 – the root cause of that and similar issue lies in the way we reference objects between the two runtimes. Due to the fact we use JSI's host objects and direct references it turns out that a lot of objects can outlast garbage collection because, for example they are referenced on the secondary runtime where garbage collection didn't run yet. Moreover, JSI objects have their own, simplified version of GC that has to run as well. As a result, for some objects to be cleaned up properly, we needed to run GC on the main JS runtime, then on the UI runtime and once again on the main JS runtime.
2. Secondary objective for the rewrite was to address the maintainability of the codebase part being rewritten. After investigating the issue mentioned above we concluded that it is too difficult and risky to try to fix it given the architectural choices we made in the past.
3. As a side-effect of the things being rewritten we expect the library to perform better. When tracking performance issues in the past, we detected that a lot of issues are due to an excessive communication over JSI. JSI is a lot quicker compared to calling things over "the bridge", however it still incurs some performance penalty that is significant enough on lower level devices even at a level of couple hundred calls. This issue has became apparent to us at some point, but the existing architecture wouldn't allow us to address it in a easy way. As an example we can bring up the issue with worklet functions that, before this rewrite, were represented as JSI's HostFunctions. As a result, calling worklet from on the UI thread would require a JSI roundtrip, and that has became problematic as we expanded the codebase and had one more logic extracted out into smaller worklets. This rewrite has been architected to optimize the number of cross-JSI calls.
4. The new core makes it possible to implement some new and often requested functionalities. Specifically, the ability to make complex shareable objects, that is, to be able to append to a shared array or add/remove entries from a shared map as opposed to always having to copy the whole new object with the modifications applied. Adding this also helps to improve some bits internally, specifically the case of shared styles, where we have a single animated style object assigned to multiple views. Before, we'd use an immutable array that'd represent the list of so-called view descriptors. With this rewrite we can add and remove entries without a need to copy the whole descriptors array as we mount new views.

The description of this PR is also a work in progress and will update sections of this description as the work progresses.

## Changes

This PR changes a lot of things under the hood. The main change lies in a way we create and reference shareable values. As part of this change the whole "notion" of shareable values is removed from the core and implemented on top of a different abstraction. We now allow for "shareable" data to provide an initializer function which is called on the UI runtime in order to instantiate data that then will be accessed by worklets. In addition, we avoid keeping JSI object cache on c++ side, this prevents issues with them being referenced for longer than necessary and hence was a root cause of memory related issues (leaks -- but these weren't actual leaks as triggering GC several times would result in a complete cleanup).

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -109,7 +64,36 @@ NativeReanimatedModule::NativeReanimatedModule(
...
+        : extractShareableOrThrow(rt, argsValue);
+    auto jsRuntime = this->runtimeHelper->rnRuntime;
+    this->scheduler->scheduleOnJS([=] {
+      jsi::Runtime &rt = *jsRuntime;
+      auto remoteFun = shareableRemoteFun->getJSValue(rt);
+      if (shareableArgs == nullptr) {
+        // fast path for remote function w/o arguments
+        remoteFun.asObject(rt).asFunction(rt).call(rt);
+      } else {
+        auto argsAry = shareableArgs->getJSValue(rt).asObject(rt).asArray(rt);
```
tomekzaw, line 83:
```suggestion
        auto argsArray = shareableArgs->getJSValue(rt).asObject(rt).asArray(rt);
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -173,87 +160,62 @@ NativeReanimatedModule::NativeReanimatedModule(
...
-    optimalizationLvl = optimalization.asNumber();
-  }
-  auto updaterSV = ShareableValue::adapt(rt, updater, this);
-  auto viewDescriptorsSV = ShareableValue::adapt(rt, viewDescriptors, this);
-
+    const jsi::Value &worklet) {
+  auto shareableWorklet = extractShareableOrThrow(rt, worklet);
+  assert(
+      shareableWorklet->valueType == Shareable::WorkletType &&
+      "only worklets can be scheduled to run on UI");
```
tomekzaw, line 176: Is this warning meant to be for internal purposes only? We have this cool warning message "Tried to synchronously call function ..." with further instructions for the library user, is it something similar or totally different?

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -173,87 +160,62 @@ NativeReanimatedModule::NativeReanimatedModule(
...
+  } else if (value.isUndefined()) {
+    shareable = std::make_shared<ShareableScalar>();
+  } else if (value.isNull()) {
+    shareable = std::make_shared<ShareableScalar>(nullptr);
+  } else if (value.isBool()) {
+    shareable = std::make_shared<ShareableScalar>(value.getBool());
+  } else if (value.isNumber()) {
+    shareable = std::make_shared<ShareableScalar>(value.getNumber());
+  } else {
+    throw "something wacky";
```
tomekzaw, line 216: e.g. `jsi::BigInt` or `jsi::ArrayBuffer`, maybe we could also share those between the JS and UI context?

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -351,10 +313,10 @@ void NativeReanimatedModule::onEvent(
...
     eventHandlerRegistry->processEvent(*runtime, eventName, eventAsString);
 #endif
-    mapperRegistry->execute(*runtime);
-    if (mapperRegistry->needRunOnRender()) {
-      maybeRequestRender();
-    }
+    //    mapperRegistry->execute(*runtime);
+    //    if (mapperRegistry->needRunOnRender()) {
+    //      maybeRequestRender();
+    //    }
```
tomekzaw, line 319: Is this necessary?

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+jsi::Value CoreFunction::call(jsi::Runtime &rt, const jsi::Value &arg0) {
+  if (runtimeHelper->isUIRuntime(rt)) {
+    if (uiFunction == nullptr) {
+      // maybe need to initialize UI Function
+      auto codeBuffer =
+          std::make_shared<const jsi::StringBuffer>("(" + functionBody + ")");
+      uiFunction = std::make_shared<jsi::Function>(
+          rt.evaluateJavaScript(codeBuffer, "__TODO")
```
tomekzaw, line 20: TODO?

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -0,0 +1,83 @@
...
+std::shared_ptr<Shareable> extractShareableOrThrow(
+    jsi::Runtime &rt,
+    const jsi::Value &maybeShareableValue) {
+  if (maybeShareableValue.isObject()) {
+    auto object = maybeShareableValue.asObject(rt);
+    if (object.isHostObject<ShareableJSRef>(rt)) {
+      return object.getHostObject<ShareableJSRef>(rt)->value;
+    }
+  }
+  throw std::string("value is not shareable");
```
tomekzaw, line 40: Can we throw `std::runtime_error` here? They're way easier to catch.

src/reanimated2/ViewDescriptorsSet.ts
```diff
@@ -11,82 +10,37 @@ export interface ViewRefSet<T> {
...
-        data.items.push(item);
-
-        if (!data.waitForInsertSync) {
-          data.waitForInsertSync = true;
-
-          scheduleUpdates(() => {
-            data.sharableViewDescriptors.value = data.items;
-            data.waitForInsertSync = false;
-          });
+  const sharableViewDescriptors = makeMutable<Descriptor[]>([]);
```
tomekzaw, line 19: Can we rename `sharable` to `shareable` here and everywhere else?

src/reanimated2/hook/useSharedValue.ts
```diff
@@ -3,11 +3,16 @@ import { cancelAnimation } from '../animation';
 import { SharedValue } from '../commonTypes';
 import { makeMutable } from '../core';
 
-export function useSharedValue<T>(init: T): SharedValue<T> {
-  const ref = useRef<SharedValue<T>>(makeMutable(init));
+export function useSharedValue<T>(
+  init: T,
+  needSynchronousReadsFromReact = false
```
tomekzaw, line 8:
```suggestion
  needsSynchronousReadsFromReact = false
```
BTW `readableFromReact` is way shorter

src/reanimated2/commonTypes.ts
```diff
@@ -44,8 +44,16 @@ export interface AnimatedStyle
 }
 export interface SharedValue<T> {
   value: T;
+  addListener: (listenerID: number, listener: (value: T) => void) => void;
+  removeListener: (listenerID: number) => void;
+  modify: (modifier: <S extends T>(value: S) => S) => void;
 }
 
+export type ShareableRef = {
+  __thereIsNothingHereImJustMakingTypescriptHappy?: boolean;
```
tomekzaw, line 53: But why?

Example/ios/ReanimatedExample/AppDelegate.m
```diff
@@ -35,6 +35,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 - (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
 {
 #if DEBUG
+  [RCTBundleURLProvider sharedSettings].enableDev = YES;
```
tomekzaw, line 38: Why is it necessary?

Example/ios/Podfile.lock
```diff
@@ -546,11 +546,11 @@ SPEC CHECKSUMS:
...
-  RNReanimated: 611653f19a52e03650e4cff59af07e9d1dcb189f
+  RNReanimated: 5eaeaf4ca5209784506e5ce7c6e3f1eae1c342b7
   RNScreens: 4a1af06327774490d97342c00aee0c2bafb497b7
   RNSVG: ce9d996113475209013317e48b05c21ee988d42e
   Yoga: 82c9e8f652789f67d98bed5aef9d6653f71b04a9
 
 PODFILE CHECKSUM: b791cbee722f6239849c3e8800f24bf481293ccd
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.11.2
```
tomekzaw, line 556: plz upgrade cocoapods

Common/cpp/SharedItems/Shareables.h
```diff
@@ -0,0 +1,249 @@
...
+      case Shareable::UndefinedType:
+        return jsi::Value();
+      case Shareable::NullType:
+        return jsi::Value(nullptr);
+      case Shareable::BooleanType:
+        return jsi::Value(data.boolean);
+      case Shareable::NumberType:
+        return jsi::Value(data.number);
+      default:
+        throw "invalid scalar type";
```
tomekzaw, line 236:
```suggestion
        throw std::runtime_error("Unsupported scalar type");
```

Common/cpp/SharedItems/Shareables.h
```diff
@@ -0,0 +1,249 @@
...
+#include "RuntimeManager.h"
+#include "Scheduler.h"
+
+using namespace facebook;
+
+namespace reanimated {
+
+class CoreFunction;
+
+class JSRuntimeHelper {
```
tomekzaw, line 15: This name seems to be too general, is it some kind of runtime scheduler?

Common/cpp/AnimatedSensor/AnimatedSensorModule.cpp
```diff
@@ -25,41 +21,36 @@ AnimatedSensorModule::~AnimatedSensorModule() {
 
 jsi::Value AnimatedSensorModule::registerSensor(
     jsi::Runtime &rt,
-    const jsi::Value &sensorType,
+    JSRuntimeHelper *runtimeHelper,
```
tomekzaw, line 24: Is it safe to use raw pointers here?

Common/cpp/SharedItems/Shareables.h
```diff
@@ -0,0 +1,249 @@
...
+ public:
+  jsi::Runtime *uiRuntime; // UI runtime created by Reanimated
+  jsi::Runtime *rnRuntime; // React-Naitve's main JS runtime
```
tomekzaw, line 18:
```suggestion
  jsi::Runtime *rnRuntime; // React-Native's main JS runtime
```

Common/cpp/SharedItems/Shareables.h
```diff
@@ -0,0 +1,249 @@
...
+  void scheduleOnUI(std::function<void()> job) {
+    scheduler->scheduleOnUI(job);
+  }
+
+  void scheduleOnJS(std::function<void()> job) {
+    scheduler->scheduleOnJS(job);
+  }
+};
+
+// Core functions are not allowed to captrue oputside variables, otherwise the'd
```
tomekzaw, line 43:
```suggestion
// Core functions are not allowed to capture outside variables, otherwise they'd
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
kmagiera: Yes, this is an internal JSI API method that is expected to only be called by the end-user via `runOnUI` method (or some other methods that call to `runOnUI`) – the implementation of that method always converts the provided argument to "shareable" and if any other checks are necessary it'd make more sense to add them on the JS side in that runOnUI method. Otherwise this error is never expected to occur

Common/cpp/NativeModules/NativeReanimatedModule.cpp
kmagiera: these types, I'm afraid, will fall into the branch that checks for "isObject". It makes sense, however, to make this error more descriptive anyways, even though I don't know under what circumstances it can be reached.

Common/cpp/SharedItems/Shareables.h
kmagiera: it was meant to be that way. We already had a number of similar helpers and I was hoping to unify them into a single class after all the code that can go away is removed. Need to revisit this topic now and will make updates accordingly

Common/cpp/SharedItems/Shareables.h
kmagiera: this actually means it is "invalid" rather than unsupported. We only create ShareableScalar objects for supported types. This error should never been reached as the constructor is only called for NumberType, BooleanType, UndefinedType and NullType however putting default section there is necessary to satisfy compiler (there are more possible value types but neither is a scalar). I changed the error message accordingly.

Example/ios/ReanimatedExample/AppDelegate.m
kmagiera: this one got in by mistake

src/reanimated2/ViewDescriptorsSet.ts
kmagiera: I'd prefer not to as a part of this PR to avoid making it even more bloated. This name is also referenced in other places that I didn't touch (12 occurrences thorough the codebase)

Common/cpp/SharedItems/Shareables.h
```diff
@@ -0,0 +1,270 @@
...
+    // BigIntType, TODO
+    StringType,
+    ObjectType,
+    ArrayType,
+    WorkletType,
+    RemoteFunctionType,
+    HandleType,
+  };
+
+  Shareable(ValueType valueType_) : valueType(valueType_){};
```
tomekzaw, line 87: The convention (at least in React Native C++ codebase) is that we mark instance fields with trailing underscore (e.g. `foo_`) and don't mark constructor args with underscore, so effectively the initializer list items looks like this: `foo_(foo)`.

We are incrementally adopting this convention in Reanimated sources as well, especially since the Fabric migration.

Can we apply this convention in this PR? 

References:
* https://github.com/facebook/react-native/blob/089684ee5604a6a2a15a4fc03fe63c73a5de9fd0/ReactCommon/react/renderer/mounting/MountingTransaction.h#L81-L84
* https://github.com/facebook/react-native/blob/089684ee5604a6a2a15a4fc03fe63c73a5de9fd0/ReactCommon/react/renderer/mounting/MountingTransaction.cpp#L19-L22

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+ShareableObject::ShareableObject(jsi::Runtime &rt, const jsi::Object &object)
+    : RetainingShareable(rt, ObjectType) {
+  auto propertyNames = object.getPropertyNames(rt);
+  const size_t count = propertyNames.size(rt);
+  data.reserve(count);
+  for (size_t i = 0; i < count; i++) {
+    auto key = propertyNames.getValueAtIndex(rt, i).asString(rt);
+    auto value = extractShareableOrThrow(rt, object.getProperty(rt, key));
+    data.push_back(std::make_pair(key.utf8(rt), value));
```
tomekzaw, line 81:
```suggestion
    data.emplace_back(key.utf8(rt), value);
```

Common/cpp/SharedItems/Shareables.h
```diff
@@ -0,0 +1,270 @@
...
+      : Shareable(valueType), hostRuntime(&rt) {}
+
+  jsi::Value getJSValue(jsi::Runtime &rt) override final;
+};
+
+class ShareableJSRef : public jsi::HostObject {
+ public:
+  std::shared_ptr<Shareable> value;
+  ShareableJSRef(std::shared_ptr<Shareable> _value) : value(_value){};
+  virtual ~ShareableJSRef(){};
```
tomekzaw, line 110: Maybe let's add a comment why this empty virtual destructor is here?

Common/cpp/SharedItems/Shareables.h
```diff
@@ -0,0 +1,270 @@
...
+    jsi::Runtime &rt,
+    const jsi::Value &maybeShareableValue);
+
+class ShareableArray : public RetainingShareable {
+ public:
+  ShareableArray(jsi::Runtime &rt, const jsi::Array &array);
+
+  jsi::Value toJSValue(jsi::Runtime &rt) override {
+    auto ary = jsi::Array(rt, data.size());
+    for (size_t i = 0; i < data.size(); i++) {
```
tomekzaw, line 130: Not sure if `jsi::Array::size` is marked as `const` so this can be optimized away by the compiler.
```suggestion
    size_t size = data.size();
    auto ary = jsi::Array(rt, size);
    for (size_t i = 0; i < size; i++) {
```

Common/cpp/Tools/Scheduler.cpp
```diff
@@ -16,6 +17,7 @@ void Scheduler::scheduleOnJS(std::function<void()> job) {
 
 void Scheduler::triggerUI() {
   scheduledOnUI = false;
+  auto scope = jsi::Scope(*runtimeManager.lock()->runtime);
```
tomekzaw, line 20: Can we add a comment here to explain better what this line does?

Common/cpp/AnimatedSensor/AnimatedSensorModule.h
```diff
@@ -10,6 +10,8 @@ namespace reanimated {
 
 using namespace facebook;
 
+class JSRuntimeHelper;
```
tomekzaw, line 13: Instead of declaring the class, can we import a header file here?

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -0,0 +1,85 @@
...
+Shareable::~Shareable() {
+}
+
+jsi::Value RetainingShareable::getJSValue(jsi::Runtime &rt) {
+  jsi::Value value;
+  if (&rt == hostRuntime) {
+    // TODO: it is suboptimal to generate new object every time getJS is called on host runtime – the
+    // object we are generating already exists and we should possibly just grab a hold of that object
+    // and use it here instead of creating a new JS representation.
+    // As far as I understand the only caase where this can be realistically called this way is when
```
tomekzaw, line 54:
```suggestion
    // As far as I understand the only case where this can be realistically called this way is when
```

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -0,0 +1,85 @@
...
+    // a shared value is created and then accessed on the same runtime
+    return toJSValue(rt);
+  } else if (remoteValue == nullptr || (value = remoteValue->lock(rt)).isUndefined()) {
```
tomekzaw, line 57: What's going on here with `(value = remoteValue->lock(rt))` inside the conditional? I guess that it affects the return in line 61, is this intended? This may be easily overlooked, maybe let's add another if/else inside this branch?

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -0,0 +1,85 @@
...
+    value = toJSValue(rt);
+    remoteValue = std::make_shared<jsi::WeakObject>(rt, value.asObject(rt));
+  }
+  return value;
+}
+
+ShareableArray::ShareableArray(jsi::Runtime &rt, const jsi::Array &array)
+    : RetainingShareable(rt, ArrayType) {
+  const size_t count = array.size(rt);
```
tomekzaw, line 66: Maybe let's stick to `size` instead of `count` here and in line 76 (there's `argsSize` in other place).

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -30,7 +29,22 @@ void RuntimeDecorator::decorateRuntime(
...
+                 const jsi::Value *args,
+                 size_t count) -> jsi::Value {
+    auto stats = rt.instrumentation().getRecordedGCStats();
+    Logger::log(stats.c_str());
+    return jsi::Value::undefined();
+  };
+
   jsi::Object dummyGlobal(rt);
+  dummyGlobal.setProperty(rt, "gc", rt.global().getProperty(rt, "gc"));
+  dummyGlobal.setProperty(
```
piaskowyk, line 43: Is `dump` just temporary function for development only or you want to remove it before the merge of PR?

src/reanimated2/core.ts
```diff
@@ -69,16 +66,37 @@ export const isConfiguredCheck: () => void = () => {
   checkPluginState(true);
 };
 
-function pushFrame(frame: (timestamp: Timestamp) => void): void {
-  (NativeReanimatedModule as JSReanimated).pushFrame(frame);
-}
+const configurationCheckWrapper = __DEV__
```
piaskowyk, line 69: It should be `global.__DEV__` because this can break the Web version or SSR.

src/reanimated2/core.ts
```diff
@@ -157,15 +144,12 @@ if (nativeShouldBeMock()) {
...
+    return NativeReanimatedModule.getTimestamp();
   }
   return _getTimestamp();
 }
 
-function valueSetter<T extends WorkletValue>(
-  sv: WorkletValueSetterContext,
-  value: T
-): void {
+function valueSetter(sv: any, value: any): void {
```
marksyzm: And pass in a generic to the method?
```suggestion
function valueSetter<T>(sv: SharedValue<T>, value: any): void {
```

ios/native/NativeProxy.mm
```diff
@@ -378,26 +372,24 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
   // Layout Animation callbacks setup
   [animationsManager setAnimationStartingBlock:^(
                          NSNumber *_Nonnull tag, NSString *type, NSDictionary *_Nonnull values, NSNumber *depth) {
-    std::shared_ptr<jsi::Runtime> runtime = wrt.lock();
-    if (runtime == nullptr) {
-      return;
-    }
-    jsi::Object yogaValues(*runtime);
+    jsi::Runtime &rt = *wrt.lock();
```
tomekzaw, line 375: Not sure if syntax `*wrt.lock()` keeps the shared pointer alive.
```suggestion
    auto runtime = wrt.lock();
    jsi::Runtime &rt = *runtime;
```

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -378,6 +320,49 @@ void NativeProxy::installJSIBindings(
...
+          } catch (std::invalid_argument e) {
+            if (auto errorHandler = weakErrorHandler.lock()) {
+              errorHandler->setError("Failed to convert value to number");
+              errorHandler->raise();
+            }
+          }
+        }
+
+        module->layoutAnimationsManager().startLayoutAnimation(
+            *runtime, tag, type->toStdString(), yogaValues);
```
tomekzaw, line 348:
```suggestion
        module->layoutAnimationsManager().startLayoutAnimation(
            rt, tag, type->toStdString(), yogaValues);
```

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -378,6 +320,49 @@ void NativeProxy::installJSIBindings(
...
 
+  layoutAnimations->cthis()->setWeakUIRuntime(wrt);
+  std::weak_ptr<ErrorHandler> weakErrorHandler = errorHandler;
+
+  layoutAnimations->cthis()->setAnimationStartingBlock(
+      [wrt, weakErrorHandler](
+          int tag,
+          alias_ref<JString> type,
+          alias_ref<JMap<jstring, jstring>> values) {
+        auto &rt = *wrt.lock();
```
tomekzaw, line 331:
```suggestion
        auto runtime = wrt.lock();
        auto &rt = *runtime;
```

ios/native/NativeProxy.mm
```diff
@@ -244,19 +244,15 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
...
-  auto progressLayoutAnimation = [weakAnimationsManager, animatedRuntime](int tag, jsi::Object newStyle) {
-    REAAnimationsManager *animationsManager = weakAnimationsManager;
-    if (animationsManager) {
-      NSDictionary *propsDict = convertJSIObjectToNSDictionary(*animatedRuntime, newStyle);
-      [animationsManager progressLayoutAnimationWithStyle:propsDict forTag:@(tag)];
-    }
+  std::weak_ptr<jsi::Runtime> wrt = animatedRuntime;
+
+  auto progressLayoutAnimation = [=](int tag, jsi::Object newStyle) {
+    NSDictionary *propsDict = convertJSIObjectToNSDictionary(*wrt.lock(), newStyle);
```
tomekzaw, line 250: Another `*wrt.lock()` here, not sure if that's safe.

Common/cpp/Tools/SingleInstanceChecker.h
```diff
@@ -29,7 +29,7 @@ class SingleInstanceChecker {
 
   // A static field will exist separately for every class template.
   // This has to be inline for automatic initialization.
-  inline static int instanceCount_;
+  inline static volatile int instanceCount_;
```
jwajgelt, line 32: What's this `volatile` qualifier for?

src/reanimated2/globals.d.ts
```diff
@@ -17,6 +21,14 @@ declare global {
...
   const _getCurrentTime: () => number;
+  const _getTimestamp: () => number;
+  const _stopObservingProgress: (tag: number, flag: boolean) => void;
+  const _notifyAboutProgress: (tag: number, value: number) => void;
+  const _notifyAboutEnd: (
+    tag: number,
+    finished: boolean,
+    removeView: b
+  ) => void;
   const _stopObservingProgress: (
```
likesebiii, line 32: Screenshot 2022-11-21 at 14 55 00

src/reanimated2/globals.d.ts
```diff
@@ -17,6 +21,14 @@ declare global {
   const _setGlobalConsole: (console?: ReanimatedConsole) => void;
   const _log: (s: string) => void;
   const _getCurrentTime: () => number;
+  const _getTimestamp: () => number;
+  const _stopObservingProgress: (tag: number, flag: boolean) => void;
+  const _notifyAboutProgress: (tag: number, value: number) => void;
+  const _notifyAboutEnd: (
+    tag: number,
+    finished: boolean,
+    removeView: b
```
piaskowyk, line 30:
```suggestion
    removeView: boolean
```

Common/cpp/SharedItems/Shareables.h
```diff
@@ -244,21 +270,22 @@ class ShareableRemoteFunction
 class ShareableHandle : public Shareable {
  private:
   std::shared_ptr<JSRuntimeHelper> runtimeHelper_;
-  std::shared_ptr<ShareableObject> initializer_;
-  std::shared_ptr<jsi::Value> remoteValue;
+  std::unique_ptr<ShareableObject> initializer_;
+  std::unique_ptr<jsi::Value> remoteValue;
```
tomekzaw:
```suggestion
  std::unique_ptr<jsi::Value> remoteValue_;
```

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -67,30 +67,43 @@ jsi::Value RetainingShareable::getJSValue(jsi::Runtime &rt) {
     return toJSValue(rt);
   } else if (remoteValue == nullptr) {
     value = toJSValue(rt);
-    remoteValue = std::make_shared<jsi::WeakObject>(rt, value.asObject(rt));
+    remoteValue = new jsi::WeakObject(rt, value.asObject(rt));
```
tomekzaw, line 70: Can we avoid using raw pointers here?

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -67,30 +67,43 @@ jsi::Value RetainingShareable::getJSValue(jsi::Runtime &rt) {
...
+      remoteValue = new jsi::WeakObject(rt, value.asObject(rt));
     }
   }
   return value;
 }
 
+RetainingShareable::~RetainingShareable() {
+  if (!runtimeHelper_->uiRuntimeDestroyed) {
+    delete remoteValue;
+  }
```
tomekzaw, line 85: This is expected to leak on reloads, right? If yes, can we document it here in a comment?

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -42,7 +42,7 @@ NativeReanimatedModule::NativeReanimatedModule(
     PlatformDepMethodsHolder platformDepMethodsHolder)
     : NativeReanimatedModuleSpec(jsInvoker),
       RuntimeManager(rt, errorHandler, scheduler, RuntimeType::UI),
-      eventHandlerRegistry(std::make_shared<EventHandlerRegistry>()),
+      eventHandlerRegistry(new EventHandlerRegistry()),
```
tomekzaw, line 45: Can we use `std::make_unique` instead?

piaskowyk (approved):
🚀

marksyzm:
When is the release??

ranaavneet:
Is this available in rc7?

tomekzaw:
> Is this available in rc7?

Nope, but we will release 3.0.0-rc.8 soon only with this PR.

ranaavneet:
Awesome, Cant wait to test it out!

ranaavneet:
@kmagiera Do we have to pass a new parameter as true in useSharedValue(?, true) in order for this to come in effect?

ranaavneet:
After installing the rc8, i keep getting the below error

Simulator Screen Shot - iPhone 14 Pro Max - 2022-12-02 at 12 03 36

Common/cpp/SharedItems/Shareables.h
```diff
@@ -0,0 +1,448 @@
...
+
+#ifdef RCT_NEW_ARCH_ENABLED
+#include <react/renderer/uimanager/primitives.h>
+#endif
+
+#include "ReanimatedRuntime.h"
+#include "RuntimeManager.h"
+#include "Scheduler.h"
+
+#define HAS_JS_WEAK_OBJECTS JS_RUNTIME_HERMES
```
tomekzaw, line 17: Does V8 runtime also support `jsi::WeakObject`? cc @Kudo

Common/cpp/NativeModules/NativeReanimatedModule.h
```diff
@@ -144,23 +151,17 @@ class NativeReanimatedModule : public NativeReanimatedModuleSpec,
       jsi::Runtime &rt,
       const jsi::Value &listenerId) override;
 
-  jsi::Value configureLayoutAnimation(
-      jsi::Runtime &rt,
-      const jsi::Value &viewTag,
-      const jsi::Value &type,
-      const jsi::Value &config,
-      const jsi::Value &viewSharedValue) override;
+  inline LayoutAnimationsManager &layoutAnimationsManager() {
```
tomekzaw, line 154: Can we rename this method to `getLayoutAnimationsManager`?

Common/cpp/SharedItems/Shareables.h
Kudo: yes, it did: https://github.com/Kudo/react-native-v8/blob/96a1a2c8a35349967a3ad7219106a3475b85433a/src/v8runtime/V8Runtime.cpp#L982-L997

ranaavneet:
@kmagiera Do we have to pass a new parameter as true?

Example:
`const animatedValue = useSharedValue(?, true)`

in order for this to come in effect?

ranaavneet:
@piaskowyk @tomekzaw Do we have to pass a new parameter as true in order for this to come in effect?

Example:
const animatedValue = useSharedValue(0, true)

PR #3744, opened 2022-11-07
Fix search modal styling

blazejkustra:
## Description

Search modal has too big margin. It is due to the fact that html element `section` was styled in `custom.css` file that should style only the landing page, therefore serach modal was styled by mistake.

before:
image

after:
image


kacperkapusciak (approved):
Huh... Something has changed recently either in Docusaurus or in the Algolia side because it definitely worked and looked correctly in the past 🤔 

Either way thanks for the fix! 💞

PR #3747, opened 2022-11-07
Build monorepo CI

piaskowyk:
## Description

CI for testing monorepo build.


.github/workflows/build-monorepo-runner.yml
```diff
@@ -0,0 +1,47 @@
...
+      - .github/workflows/build-monorepo.yml
+      - RNReanimated.podspec
+      - android/build.gradle
+  push:
+    branches:
+      - main
+
+jobs:
+  build_android_reanimated_in_app:
+    name: Build Android with in app Reanimated
```
tomekzaw, line 14: What about "non-hoisted" instead of "in app"?

.github/workflows/build-monorepo-runner.yml
```diff
@@ -0,0 +1,47 @@
...
+    uses: ./.github/workflows/build-monorepo.yml
+    with:
+      runs_on: ubuntu-latest
+      os: Android
+      concurrency_group: build_monorepo_android_in_app_reanimated
```
tomekzaw, line 19: We need to add `-${{ github.ref }}` here so the CI jobs are not cancelled for example when running for two separate PRs.

.github/workflows/build-monorepo.yml
```diff
@@ -0,0 +1,76 @@
...
+        required: true
+        type: string
+      is_hoisted:
+        required: true
+        type: boolean
+
+jobs:
+  build:
+    name: Build monorepo with Reanimated
+    runs-on: ${{ inputs.runs_on }}
```
tomekzaw, line 21: Is it possible to use conditional here based on `inputs.platform`, like this?
```
${{ inputs.platform == 'iOS' ? 'macos-12' : 'ubuntu-latest' }}
```

.github/workflows/build-monorepo.yml
```diff
@@ -0,0 +1,76 @@
+name: Test build in monorepo
+on:
+  workflow_call:
+    inputs:
+      runs_on:
+        required: true
+        type: string
+      os:
```
tomekzaw, line 8: Can we use `platform` instead?

.github/workflows/build-monorepo.yml
```diff
@@ -0,0 +1,76 @@
...
+      cancel-in-progress: true
+    steps:
+      - name: Create monorepo
+        run: |
+          mkdir monorepo
+          cd monorepo
+          echo '{"name":"rnos-monorepo-tester","version":"1.0.0","license":"MIT","private":true,"workspaces":{"packages":["RootApp","packages/PackageApp", "AppA", "AppB"],"nohoist":["**/react","**/react-dom","**/react-native","**/react-native/**","**/react-native-codegen","**/react-native-dev-menu"]}}' > package.json
+          yarn
+          mkdir packages
+      - name: Create RootApp
```
tomekzaw, line 33: I'm wondering if we can name these apps `HoistedReanimatedApp` and `NonHoistedReanimatedApp` so that we don't need to remember if PackageApp is hoisted or not? BTW What's the better word for "non-hoisted"?

.github/workflows/build-monorepo.yml
```diff
@@ -0,0 +1,76 @@
...
+        working-directory: monorepo
+        run: npx react-native@next init RootApp
```
tomekzaw, line 35: Just thinking if we should pass `--version` here to use a specific version of React Native, but `latest` is okay for me.

.github/workflows/build-monorepo.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Create PackageApp
+        working-directory: monorepo/packages
+        run: npx react-native@next init PackageApp
+      
+      - name: Install dependencies for RootApp
+        working-directory: monorepo/RootApp
+        run: yarn add github:software-mansion/react-native-reanimated#${{ github.ref }}
```
tomekzaw, line 42: Is it possible to install app from `file:../react-native-reanimated` instead?

Also, for v2 this checks monorepo support only with building from source enabled. Maybe we should have another job that first builds the package with `createNPMPackage.sh`, then installs Reanimated from *.tgz file and finally tries to build the app?

.github/workflows/build-monorepo-runner.yml
```diff
@@ -0,0 +1,47 @@
+name: Test build in monorepo
+on:
+  pull_request:
+    paths:
+      - .github/workflows/build-monorepo.yml
```
tomekzaw, line 5: Let's also add `.github/workflows/build-monorepo-runner.yml` here.

.github/workflows/build-monorepo.yml
```diff
@@ -0,0 +1,76 @@
...
+        run: pod install
+      - name: Install Pods for PackageApp
+        if: ${{ inputs.os == 'iOS' }}
+        working-directory: monorepo/packages/PackageApp/ios
+        run: pod install
+
+      - name: Build iOS app RootApp
+        if: ${{ inputs.os == 'iOS' }}
+        working-directory: monorepo/RootApp
+        run: npx react-native run-ios
```
tomekzaw, line 63:
```suggestion
        run: yarn react-native run-ios
```

tomekzaw (changes requested):

.github/workflows/build-monorepo-runner.yml
```diff
@@ -0,0 +1,47 @@
...
+      - RNReanimated.podspec
+      - android/build.gradle
+  push:
+    branches:
+      - main
+
+jobs:
+  build_android_reanimated_in_app:
+    name: Build Android with in app Reanimated
+    uses: ./.github/workflows/build-monorepo.yml
```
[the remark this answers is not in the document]
piaskowyk: image

.github/workflows/build-monorepo.yml
piaskowyk: We need to create another PR for this feature

tomekzaw (approved):
👏

PR #3748, opened 2022-11-07
Fix V8 on Android CI build

tomekzaw:
## Description

This PR fixes the Android V8 CI that wasn't running because `env` is not available in `concurrency.group` context:

Zrzut ekranu 2022-11-7 o 13 41 37

Instead, we should use `${{ github.ref }}` to differentiate jobs for different branches/PRs.


piaskowyk (approved):

piaskowyk:
image

PR #3759, opened 2022-11-09
CI - Build Reanimated on Windows

piaskowyk:
## Description

CI to test build app with Reanimated on Windows operation system.

### DIFF
--- a/.github/workflows/build-on-windows.yml
+++ b/.github/workflows/build-on-windows.yml
@@ -1,0 +1,30 @@
+name: Test build on Windows
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/build-on-windows.yml'
+      - RNReanimated.podspec
+      - android/build.gradle
+      - package.json
+  push:
+    branches:
+      - main
+
+env:
+  REACT_NATIVE_VERSION: "0.70.5"
+
+jobs:
+  build:
+    runs-on: windows-latest
+    concurrency:
+      group: build-on-windows-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - name: Create React Native app
+        run: npx react-native init app --version ${{ env.REACT_NATIVE_VERSION }}
+      - name: Install Reanimated
+        working-directory: app
+        run: yarn add github:software-mansion/react-native-reanimated#${{ github.ref }}
+      - name: Build Android app
+        working-directory: app/android
+        run: ./gradlew assembleDebug --console=plain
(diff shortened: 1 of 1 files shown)


piaskowyk:
Current building results with different React Native versions - https://github.com/piaskowyk/reanimated-rn-version-tester/actions/runs/3431957313

.github/workflows/build-on-windows.yml
```diff
@@ -0,0 +1,34 @@
+name: Test build on Windows
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/build-on-windows.yml'
+      - RNReanimated.podspec
+      - android/build.gradle
```
tomekzaw, line 7:
```suggestion
      - 'android/**'
      - 'Common/**'
```

tomekzaw (approved):

.github/workflows/build-on-windows.yml
```diff
@@ -0,0 +1,34 @@
...
+      - package.json
+  push:
+    branches:
+      - main
+
+env:
+  REACT_NATIVE_VERSION: "0.70.5"
```
tomekzaw, line 14: What about using `latest` instead? This should work automatically for patch versions and we will know earlier if next minor version breaks something.

.github/workflows/build-on-windows.yml
```diff
@@ -0,0 +1,34 @@
...
+    concurrency:
+      group: build-on-windows-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+      - name: Create React Native app
+        run: npx react-native init app --version ${{ env.REACT_NATIVE_VERSION }}
```
tomekzaw, line 28: Nothing important, but there's a convention that we name all apps with PascalCase, i.e. `Example`, `FabricExample`, `WebExample`, so let's use `App` or `MyApp` here 😄 First I was thinking of `WindowsExample` but this is an Android app, just built in Windows OS.
```suggestion
        run: npx react-native init MyApp --version ${{ env.REACT_NATIVE_VERSION }}
```

.github/workflows/build-on-windows.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install Reanimated
```
tomekzaw, line 29: We can specify that we install the package from main branch.
```suggestion
      - name: Install Reanimated from main
```

.github/workflows/build-on-windows.yml
piaskowyk: good idea, thanks!

PR #3773, opened 2022-11-15
Update `AnimatedStyle` and `useAnimatedStyle` types

j-piasecki:
## Description

This PR is part of an ongoing effort to get rid of the `.d.ts` file and switch to types generated based on the source code.

## Changes

- Updated `AnimatedStyle` to allow primitives for certain props
- Updated `useAnimatedStyle` to not accept functions returning arrays

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -391,7 +391,8 @@ function checkSharedValueUsage(
 }
 
 export function useAnimatedStyle<T extends AnimatedStyle>(
-  updater: BasicWorkletFunction<T>,
+  // animated style cannot be an array
+  updater: BasicWorkletFunction<T extends Array<unknown> ? never : T>,
```
tjzel, line 395:
```suggestion
  updater: BasicWorkletFunction<Exclude<T, any[]>>,
```
Seems to be a better for readability.

src/reanimated2/commonTypes.ts
```diff
@@ -37,11 +37,31 @@ export interface StyleProps extends ViewStyle, TextStyle {
   [key: string]: any;
 }
 
-export interface AnimatedStyle
-  extends Record<string, Animation<AnimationObject>> {
+export interface AnimatedStyle extends Record<string, AnimationObject> {
   [key: string]: any;
-  transform?: Array<Record<string, Animation<AnimationObject>>>;
+  transform?: Array<
```
tjzel, line 42: I don't think hardcoding it like this is beneficial for us. I'd rather revert those changes and go for more elegant solution.

src/reanimated2/commonTypes.ts
```diff
@@ -37,11 +37,31 @@ export interface StyleProps extends ViewStyle, TextStyle {
   [key: string]: any;
 }
 
-export interface AnimatedStyle
-  extends Record<string, Animation<AnimationObject>> {
+export interface AnimatedStyle extends Record<string, AnimationObject> {
```
tjzel, line 40: As discussed this change has no effect and I'd rather revert it for now.

Issue #3754, opened 2022-11-08
Invalid `RNReanimated.podspec` file: no implicit conversion of nil into Array. IOS when updating from reanimated 2.8.0 to 2.12.0


FrozenPyrozen:
### Description

When update reanimated from `2.8.0` to `2.12.0` got error when installing pods, also update to reanimated `2.9.0` gives the same error:
```
`npx pod-install`
Scanning for pods...
1.11.3
> pod install

[!] Invalid `Podfile` file: 
[!] Invalid `RNReanimated.podspec` file: no implicit conversion of nil into Array.

 #  from /Users/admin/Projects/test_project/node_modules/react-native-reanimated/RNReanimated.podspec:6
 #  -------------------------------------------
 #  config = find_config()
 >  assert_no_multiple_instances(config)
 #  assert_no_reanimated2_with_new_architecture(reanimated_package_json)
 #  -------------------------------------------
.

 #  from /Users/admin/Projects/test_project/ios/Podfile:8
 #  -------------------------------------------
 #  target 'MobileWebWallet' do
 >    config = use_native_modules!
 #  
 #  -------------------------------------------
Couldn't install Pods. Updating the Pods project and trying again...
> pod install --repo-update

[!] Invalid `Podfile` file: 
[!] Invalid `RNReanimated.podspec` file: no implicit conversion of nil into Array.

 #  from /Users/admin/Projects/test_project/node_modules/react-native-reanimated/RNReanimated.podspec:6
 #  -------------------------------------------
 #  config = find_config()
 >  assert_no_multiple_instances(config)
 #  assert_no_reanimated2_with_new_architecture(reanimated_package_json)
 #  -------------------------------------------
.

 #  from /Users/admin/Projects/test_projectt/ios/Podfile:8
 #  -------------------------------------------
 #  target 'MobileWebWallet' do
 >    config = use_native_modules!
 #  
 #  -------------------------------------------
Couldn't install Pods. Updating the Pods project and trying again...
Command `pod install` failed.
└─ Cause: Invalid `Podfile` file: 
[!] Invalid `RNReanimated.podspec` file: no implicit conversion of nil into Array.

 #  from /Users/admin/Projects/test_project/node_modules/react-native-reanimated/RNReanimated.podspec:6
 #  -------------------------------------------
 #  config = find_config()
 >  assert_no_multiple_instances(config)
 #  assert_no_reanimated2_with_new_architecture(reanimated_package_json)
[...]

PR #3774, opened 2022-11-15
Fix null array reference in utilities

jfkm69:
## Description
This PR addresses the issue #3754
Invalid RNReanimated.podspec file: no implicit conversion of nil into Array. IOS when updating from reanimated 2.8.0 to 2.12.0 caused by  (https://github.com/software-mansion/react-native-reanimated/blob/05c93b02ede93950904812239fbe95366e372ab6/scripts/reanimated_utils.rb#L63) as its possible that we can have multiple instances of react-native-reanimated in the same node_modules directory (i.e. we require `react-native-reanimated` and one of our dependencies also requires it). If this happens to be the case, the condition that there are multiple react native reanimated instances is evaluated as `true`. As we attempt to evaluate the `parsed_location` its possible that `lib_instances_in_reanimated_node_modules_array` may be null if `react_native` and the `react_native_reanimated` modules are under the same `node_modules` folder (which is the case for most apps except the example app which the script was likely predicated upon).

### Fixed 
 - Fixes an issue with pod file that caused a `no implicit conversion of nil into Array` exception by initializing `lib_instances_in_reanimated_node_modules_array` to an Array.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/scripts/reanimated_utils.rb
+++ b/scripts/reanimated_utils.rb
@@ -52,6 +52,7 @@ def assert_no_multiple_instances(react_native_info)
 
   lib_instances_in_react_native_node_modules = %x[find #{react_native_info[:react_native_node_modules_dir]} -name "package.json" | grep "/react-native-reanimated/package.json"]
   lib_instances_in_react_native_node_modules_array = lib_instances_in_react_native_node_modules.split("\n")
+  lib_instances_in_reanimated_node_modules_array = Array.new
   reanimated_instances = lib_instances_in_react_native_node_modules_array.length()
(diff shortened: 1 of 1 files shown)


tomekzaw:
Hey @jfkm69, thanks for submitting this PR!

tomekzaw (approved):

edgargaza:
Would be nice to approve and merge this PR soon. This error kinda breaks the 2.12.0 and 2.13.0 releases for iOS.

Thx! 😄💘

piaskowyk (approved):

PR #3789, opened 2022-11-17
Fix Fabric crash on reload on iOS

tomekzaw:
## Description

This PR fixes iOS crash on reload of FabricExample app. Most likely the regression was caused by #3332.

## Changes

First, I observed that the crash was gone after I commented out the following line:

https://github.com/software-mansion/react-native-reanimated/blob/1bec9a069371bc27a3319d8c0a7aa31b77a98de7/ios/REAModule.mm#L159

Next, I confirmed that `removeEventListener` was indeed called in `invalidate` here:

https://github.com/software-mansion/react-native-reanimated/blob/1bec9a069371bc27a3319d8c0a7aa31b77a98de7/ios/REAModule.mm#L70

However, this method was called on the ShadowQueue thread while `addEventListener` was called on the JS thread. As this may potentially cause a race condition when accessing internal `std::vector` of EventListeners, I decided to wrap the call with `runtimeExecutor` in order to schedule the call on JS thread, like here:

https://github.com/software-mansion/react-native-reanimated/blob/1bec9a069371bc27a3319d8c0a7aa31b77a98de7/ios/REAModule.mm#L140-L148

This resolved the issue. However, when I set a breakpoint on `removeEventListener`, it was never hit, so the method was actually never called, so I decided to completely remove `removeEventListener` call. After the changes, reloading app works fine, even while `Animated.ScrollView` is being scrolled.

Launch FabricExample, open useScrollViewOffset example and reload the app a few times while scrolling.

### DIFF
--- a/ios/REAModule.mm
+++ b/ios/REAModule.mm
@@ -44,7 +44,6 @@ typedef void (^AnimatedOperation)(REANodesManager *nodesManager);
   __weak RCTSurfacePresenter *_surfacePresenter;
   std::shared_ptr<NewestShadowNodesRegistry> newestShadowNodesRegistry;
   std::weak_ptr<NativeReanimatedModule> reanimatedModule_;
-  std::shared_ptr<EventListener> eventListener_;
 #else
   NSMutableArray<AnimatedOperation> *_operations;
 #endif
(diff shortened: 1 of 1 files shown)


kmagiera (approved):
👍

piaskowyk (approved):

ansh:
Which version will this be published in?

tomekzaw:
Hey @ansh, this is already available in 3.0.0-rc.8.

ansh:
I have tried this fix, and unfortunately, it does not fix the "Crash on reload" issue. @tomekzaw 

I tried 3.0.0-rc.8.

Also introduces another bug where `jsThis` is missing variable
CleanShot 2022-12-31 at 14 02 12@2x

ansh:
Opened an issue #3895 with reproduction

PR #3795, opened 2022-11-18
Fix layout after entering animation ends

jwajgelt:
## Description

Currently, if a React render happens during an `entering` layout animation and causes the layout of the animated component to change, that component will continue it's animation and remain at the position it finished.
This changes the behaviour of `entering` animations to instead move the component to the correct place after the animation finishes.
This issue (and this change) doesn't affect screens with `layout` layout animations, since these are already combined with `entering`.

### Before
Simulator Screen Recording - iPhone 13 - 2022-11-18 at 15 29 54

### After
Simulator Screen Recording - iPhone 13 - 2022-11-18 at 15 37 39

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes

### DIFF
--- a/ios/LayoutReanimation/REAAnimationsManager.m
+++ b/ios/LayoutReanimation/REAAnimationsManager.m
@@ -29,6 +29,8 @@ static BOOL REANodeFind(id<RCTComponent> view, int (^block)(id<RCTComponent>))
 @implementation REAAnimationsManager {
   RCTUIManager *_uiManager;
   REAUIManager *_reaUiManager;
+  NSMutableSet<NSNumber *> *_enteringViews;
+  NSMutableDictionary<NSNumber *, REASnapshot *> *_enteringViewTargetValues;
   NSMutableDictionary<NSNumber *, UIView *> *_exitingViews;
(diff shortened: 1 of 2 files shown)


ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -107,8 +111,24 @@ - (UIView *)viewForTag:(NSNumber *)tag
...
+  if (view == nil) {
+    return;
+  }
+  if ([_enteringViews containsObject:tag] && !removeView) {
+    REASnapshot *target = _enteringViewTargetValues[tag];
+    if (target != nil) {
+      NSMutableDictionary *dataComponenetsByName = [_uiManager valueForKey:@"_componentDataByName"];
+      RCTComponentData *componentData = dataComponenetsByName[@"RCTView"];
+      _enteringViewTargetValues[[view reactTag]] = [[REASnapshot alloc] init:view];
+      [self setNewProps:target.values forView:view withComponentData:componentData];
```
piaskowyk, line 125: This is a code duplication with `progressLayoutAnimationWithStyle` method.

piaskowyk (approved):

PR #3805, opened 2022-11-24
Update `createAnimatedComponent` types

j-piasecki:
## Description

## Changes

- Updated `createAnimatedComponent` to be a generic function with two overloads: one for class components and one for function components
- `createAnimatedComponent` now extends the props of the received component to also accepts values handled by Reanimated, for now only `onScroll` and `style` are handled but extending it is as simple as adding a prop to the `PropsExtensions` type, thanks @jakub-gonet for help with this one

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


src/createAnimatedComponent.tsx
```diff
@@ -226,15 +237,59 @@ interface ComponentRef extends Component {
...
+function createAnimatedComponent<P>(
+  Component: FunctionComponent<P & InitialFunctionComponentProps>,
+  options?: Options<P & InitialFunctionComponentProps>
+): FunctionComponent<
+  AnimatedComponentProps<
+    ExtendProps<P, PropsExtensions> & InitialFunctionComponentProps
+  >
+>;
+function createAnimatedComponent<P>(
+  Component: ComponentType<P & InitialComponentProps>,
```
tjzel, line 278:
```suggestion
  Component: ComponentClass<P & InitialComponentProps>,
```

tjzel: We still have some errors with createAnimatedComponent, we should circle back to this PR and fix them

Issue #3806, opened 2022-11-24
global _updatePropsPaper _updatePropsFabric not getting resolved. Import Error.


shubhamdeol:
### Description

Error coming from path
/node_modules/react-native-reanimated/src/reanimated2/UpdateProps.ts

Error Message: Property '_updatePropsPaper' doesn't exist, js engine: hermes

<img width="329" alt="Screenshot 2022-11-24 at 8 12 17 PM" src="https://user-images.githubusercontent.com/11364449/203810872-b1f19fe8-3263-4306-b798-f203aa5bc670.png">

### Steps to reproduce

updating reanimated from 2.13.0 to 3.0.0-rc.6

### Snack or a link to a repository

NA

### Reanimated version

3.0.0-rc.6

### React Native version

0.68.5

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

PIXEL_5_API_31

### Acknowledgements

Yes

tomekzaw:
Hey, did you remember to re-build your app after upgrading Reanimated?

shubhamdeol:
yes, I did. 

steps I took.
- rm -rf node_modules/ && yarn cache clean && yarn install
- yarn start --reset-cache
- gradle clean
- rebuild

also looked into https://github.com/software-mansion/react-native-reanimated/issues/2502
no luck

there are others also facing same issue. 
https://stackoverflow.com/questions/73839500/referenceerror-property-updatepropspaper-doesnt-exist-js-engine-hermes

tomekzaw:
Can you please try to reproduce the issue on a fresh RN 0.68.5 app with Reanimated 3.0.0-rc.6?

shubhamdeol:
Not able to reproduce with dependences. 

`"dependencies": {
    "react": "17.0.2",
    "react-native": "0.68.5",
    "react-native-reanimated": "3.0.0-rc.6"
  },`

shubhamdeol:
My project started to work now. 
Looks like issue was with caching. I do not know what exactly fixed it.

PR #3810, opened 2022-11-24
Detect mismatch of JavaScript and native part version

tomekzaw:
## Description

This PR adds a runtime check that compares if the version of JavaScript sources matches the version of compiled binaries. The motivation for this PR was a series of issues related to a mismatch of interface of the native and JS part of Reanimated caused by upgrading react-native-reanimated when using Expo managed workflow or due to caching (e.g. #3806).

Note that `3.0.0-rc.6` cannot be directly injected as an argument because it gets treated like a floating-point number `3.0` with unexpected prefix `.0`, minus `rc.6` (whatever that means). Neither `#REANIMATED_VERSION` nor `"" ## #REANIMATED_VERSION` worked for me. First, I decided to wrap Reanimated version with double quotes immediately in build.gradle (and RNReanimated.podspec) but it failed to compile for Android on Windows (see the [CI logs](https://github.com/software-mansion/react-native-reanimated/actions/runs/3543966345/jobs/5950861983)). The ultimate workaround is to pass raw version (i.e. `-DREANIMATED_VERSION=3.0.0-rc.6`) and then use [this wild trick](https://stackoverflow.com/questions/2751870/how-exactly-does-the-double-stringize-trick-work) (analogusly like Expo does [here](https://github.com/expo/expo/blob/3329e04d849b3ce0de9f59d38d66a29455573c80/packages/expo-dev-launcher/ios/EXDevLauncherController.m#L32-L37)) to wrap it with double quotes on C++ side. Thanks @lukmccall for the codepointer.

Going a step further, we could also inject the specific version of react-native that Reanimated was compiled with and analogously compare it in runtime with the version from package.json in JavaScript. The motivation for this check is that we suspect that some part of Android crashes may be caused by ABI incompatibility that occurs when the app runs on a different patch version of react-native than was used to prebuild Reanimated AAR (note that react-native releases often include changes in native code between patch versions). However, currently, we are unable to provide binaries for all relevant patch versions of react-native (separately for 4 Android architectures as well as JSC/Hermes) due to already large size of npm package (the hard-limit on npm registry is ~120 MB and we really don't want to publish prefabs at this point yet), so we build AARs only with the latest patch version of given minor version. While this check would eliminate potential crashes, users would need to postpone upgrading react-native until the next release of Reanimated that is compatible with that patch. Instead of raising an error, we could just show a warning via `console.warn` that can be explicitly ignored with `LogBox.ignoreLogs` (so the user knows that we can no longer guarantee compatibility) but this kinda sounds like a bad PR. Also note that AARs were removed in Reanimated 3 which is always built from source on Android (like it always was on iOS) so the problem is only related to 2.x releases.

## Changes

- Added `REANIMATED_VERSION` compile flag with the version of react-native-reanimated from package.json
- Added `global._REANIMATED_VERSION_CPP` property injected by JSI
- Added `NativeReanimated.checkVersion` method that raises an error when the versions differ

| Android | iOS |
|-|-|
| Android | iOS |

1. Build Example or FabricExample app
2. Change `version` in package.json
3. Observe the red box

## Future ideas

- Detect that Expo bare workflow (Expo Go) is in use to warn user that it's not possible to use different version than the one bundled into Expo SDK.
- Call `console.warn` when react-native versions don't match


piaskowyk (approved):
Pure luksja 😎

cdci-gsoftconsulting:
same error in my side

PR #3815, opened 2022-11-28
Add "guards" on iOS to avoid leaking CADisplayLink instances

iamyellow:
## Description

`REAKeyboardEventObserver` could leak `CADisplayLink` instances spinning `updateKeyboardFrame` calls. This happens when you dismiss the keyboard with `interactive` mode. For some reason `runAnimation` is executed several times hiding the previous `displayLink` value and so leaking. This simple PR adds a "guard" to avoid spin another instance if one is active.

## Checklist

- [ ] Updated TS types
- [ ] Added TS types tests
- [ ] Updated documentation
- [ ] Ensured that CI passes


iamyellow:
Closing since, though is valid interactive dismissal is not working. I'll reopen if I find something.

PR #3821, opened 2022-11-30
Better error message when trying to call non-worklet from UI thread

kmagiera:
## Summary

After #3722 the error message presented after attempting to synchronously call React runtime function from the UI runtime has changed and became less descriptive than before. This PR adds a way for us to throw a more descriptive error in such scenario.

## Test plan

Use the following code:
```
function rnMethod() {
  console.log('from RN');
}

function uiMethod() {
  'worklet';
  rnMethod(); // this should throw
}

runOnUI(uiMethod)();
```

Before this change you'd get a generic "Object is not a function" thrown at the line where we call `rnMethod`, now the error message says "Tried to synchronously call a non-worklet function" and presents possible solution (see the diff for the whole error message)

### DIFF
--- a/Common/cpp/SharedItems/Shareables.h
+++ b/Common/cpp/SharedItems/Shareables.h
@@ -285,7 +285,14 @@ class ShareableRemoteFunction
         function_(std::move(function)) {}
   jsi::Value toJSValue(jsi::Runtime &rt) override {
     if (runtimeHelper_->isUIRuntime(rt)) {
+#ifdef DEBUG
+      return runtimeHelper_->valueUnpacker->call(
+          rt,
+          ShareableJSRef::newHostObject(rt, shared_from_this()),
+          jsi::String::createFromAscii(rt, "RemoteFunction"));
+#else
       return ShareableJSRef::newHostObject(rt, shared_from_this());
+#endif
     } else {
       return jsi::Value(rt, function_);
     }
--- a/src/reanimated2/core.ts
+++ b/src/reanimated2/core.ts
@@ -148,7 +148,7 @@ function valueUnpacker(objectToUnpack: any, category?: string): any {
 
 Possible solutions are:
   a) If you want to synchronously execute this method, mark it as a worklet
-  b) If you want to execute this method on the JS thread, wrap it using runOnJS`);
+  b) If you want to execute this function on the JS thread, wrap it using \`runOnJS\``);
     };
     fun.__remoteFunction = objectToUnpack;
     return fun;
--- a/src/reanimated2/core.ts
+++ b/src/reanimated2/core.ts
@@ -147,7 +147,7 @@ function valueUnpacker(objectToUnpack: any, category?: string): any {
       throw new Error(`Tried to synchronously call a non-worklet function
 
 Possible solutions are:
-  a) If you want to synchronously execute this method, mark it as a Worklet
+  a) If you want to synchronously execute this method, mark it as a worklet
   b) If you want to execute this method on the JS thread, wrap it using runOnJS`);
     };
     fun.__remoteFunction = objectToUnpack;
--- a/src/reanimated2/core.ts
+++ b/src/reanimated2/core.ts
@@ -114,7 +114,7 @@ export function getViewProp<T>(viewTag: string, propName: string): Promise<T> {
   });
 }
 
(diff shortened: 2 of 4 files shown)


src/reanimated2/core.ts
```diff
@@ -142,6 +142,16 @@ function valueUnpacker(objectToUnpack: any): any {
       handleCache!.set(objectToUnpack, value);
     }
     return value;
+  } else if (category === 'RemoteFunction') {
+    const fun = () => {
+      throw new Error(`Tried to synchronously call a non-worklet function
+
+Possible solutions are:
+  a) If you want to synchronously execute this method, mark it as a Worklet
```
tomekzaw, line 150:
```suggestion
  a) If you want to synchronously execute this method, mark it as a worklet
```

Common/cpp/SharedItems/Shareables.h
```diff
@@ -285,7 +285,14 @@ class ShareableRemoteFunction
         function_(std::move(function)) {}
   jsi::Value toJSValue(jsi::Runtime &rt) override {
     if (runtimeHelper_->isUIRuntime(rt)) {
+#ifdef DEBUG
```
tomekzaw, line 288: Why is it only available in debug?

src/reanimated2/core.ts
```diff
@@ -142,6 +142,16 @@ function valueUnpacker(objectToUnpack: any): any {
       handleCache!.set(objectToUnpack, value);
     }
     return value;
+  } else if (category === 'RemoteFunction') {
+    const fun = () => {
+      throw new Error(`Tried to synchronously call a non-worklet function
```
tomekzaw, line 147: Previously, this error message also contained the name of the function that was called. Is it possible to paste the function name here, at least in some cases?

src/reanimated2/threads.ts
```diff
@@ -27,6 +27,9 @@ export function runOnJS<A extends any[], R>(
   fun: ComplexWorkletFunction<A, R>
 ): (...args: A) => void {
   'worklet';
+  if (fun.__remoteFunction) {
+    fun = fun.__remoteFunction;
+  }
```
tomekzaw, line 32: Can we add some comment what happens here, what's  `fun`, what's `__remoteFunction` etc.?

src/reanimated2/core.ts
```diff
@@ -142,6 +142,16 @@ function valueUnpacker(objectToUnpack: any): any {
...
+
+Possible solutions are:
+  a) If you want to synchronously execute this method, mark it as a Worklet
+  b) If you want to execute this method on the JS thread, wrap it using runOnJS`);
```
tomekzaw, line 151:
```suggestion
  b) If you want to execute this method on the JS thread, wrap it using \`runOnJS\`.`);
```

src/reanimated2/core.ts
tomekzaw, line 147:
```suggestion
      throw new Error(`Tried to synchronously call a non-worklet function on the UI context.
```

Common/cpp/SharedItems/Shareables.h
kmagiera: It will throw "Object is not a function" in production and this way it works faster in release mode as it does not perform additional JSI call just for the sake of displaying an error message which only happens if you mess something up. I think it is the right tradeoff and I don't think it is likely that someone will only hit the error in release but never in production. I think the effort should be better spend on figuring out a solution for providing better stack traces which should give a much better visibility into the error root cause

src/reanimated2/core.ts
kmagiera: We could do that but not sure if its worth adding as part of this PR. This is already an improvement and I think it'd be better to spend time on solving issue with stack traces which should ideally pinpoint the exact line where the error happened rather than only showing the function name

Common/cpp/SharedItems/Shareables.h
tomekzaw: In general, I don't like such differences between in the behaviour of debug and release builds but this tradeoff sounds reasonable. 👍

BTW We could also analogously wrap `if (fun.__remoteFunction) { ... }` inside `if (__DEV__)` check so it can be optimized away but maybe it's not worth it.

tomekzaw (approved):

PR #3822, opened 2022-11-30
Fix animated refs on Fabric

kmagiera:
## Summary

This PR fixes crash on Fabric when using synchronous measure/scrollTo etc – this error was a regression introduced in #3722.

The issue was that did not transform shadow node wrapper ref into shareable ref. Shareable ref was necessary for the UI side to recognize and extract the shadow node wrapper that was needed for the sync calls to execute.

## Test plan

Run FabricExample, use "Measure on UI" on the "Measure Example" screen and "ScrollTo on UI" button on the "ScrollTo Example" screen.

### DIFF
--- a/src/reanimated2/hook/useAnimatedRef.ts
+++ b/src/reanimated2/hook/useAnimatedRef.ts
@@ -6,11 +6,20 @@ import { getTag } from '../NativeMethods';
 import { getShadowNodeWrapperFromHostInstance } from '../fabricUtils';
 import {
   makeShareableCloneRecursive,
+  makeShareableShadowNodeWrapper,
   registerShareableMapping,
 } from '../shareables';
 
+function getShareableShadowNodeFromComponent(
+  component: Component
+): ShadowNodeWrapper {
+  return makeShareableShadowNodeWrapper(
+    getShadowNodeWrapperFromHostInstance(component)
+  );
+}
+
 const getTagValueFunction = global._IS_FABRIC
-  ? getShadowNodeWrapperFromHostInstance
+  ? getShareableShadowNodeFromComponent
   : getTag;
 
 export function useAnimatedRef<T extends Component>(): RefObjectFunction<T> {


src/reanimated2/hook/useAnimatedRef.ts
```diff
@@ -6,11 +6,20 @@ import { getTag } from '../NativeMethods';
 import { getShadowNodeWrapperFromHostInstance } from '../fabricUtils';
 import {
   makeShareableCloneRecursive,
+  makeShareableShadowNodeWrapper,
   registerShareableMapping,
 } from '../shareables';
 
+function getShareableShadowNodeFromComponent(
+  component: Component
+): ShadowNodeWrapper {
```
tomekzaw, line 15: It looks like this function has an invalid return type. Technically, it returns a shareable, not a ShadowNodeWrapper directly.

There's no error since we explicitely define `ShadowNodeWrapper` type as `object` here:

 https://github.com/software-mansion/react-native-reanimated/blob/99573c396ca6b7e05ec28b5f2165ee7ad06c6dc5/src/reanimated2/commonTypes.ts#L174

kmagiera: I don't understand what the problem is here? `makeShareableShadowNodeWrapper` return type is the same as the argument and in this case it's `ShadowNodeWrapper`. There would be no error even if `ShadowNodeWrapper` was defined differently

tomekzaw: Indeed, `makeShareableShadowNodeWrapper` returns the argument ([proof](https://github.com/software-mansion/react-native-reanimated/blob/99573c396ca6b7e05ec28b5f2165ee7ad06c6dc5/src/reanimated2/shareables.ts#L29-L34)). Everything is okay.

tomekzaw (approved):

PR #3824, opened 2022-11-30
[iOS][Layout Animations] Remove `exiting` views from UIManager's view registry

jwajgelt:
## Summary
Currently, when views with an `exiting` animation are unmounted, any deleted in that transaction aren't removed from the UIManager's view registry (the mapping of react tags to native views).
This can cause memory leaks of removed views.

This PR restores the original behaviour of UIManager when removing views, allowing it to clean them up properly, and reattaches the views with `exiting` animations after the transaction removing the views has completed.

## Test plan

- In example app, go to "Entering and Exiting with Layout"
- Set a breakpoint somewhere in `RCTUIManager`, for example `_manageChildren`.
- Press "Toggle" to create the animated views, and note the number of entries in `_viewRegistry`.
- Disable the breakpoint, resume the app and start spamming the toggle button.
- Enable the breakpoint again, and press toggle one more time.
- The number of entries in `_viewRegistry` should be the same as before spamming toggle.

### DIFF
--- a/ios/LayoutReanimation/REAAnimationsManager.m
+++ b/ios/LayoutReanimation/REAAnimationsManager.m
@@ -279,20 +279,20 @@ static BOOL REANodeFind(id<RCTComponent> view, int (^block)(id<RCTComponent>))
   }
 }
 
-- (BOOL)removeRecursive:(UIView *)view fromContainer:(UIView *)container withoutAnimation:(BOOL)removeImmediately;
+- (BOOL)startAnimationsRecursive:(UIView *)view removingSubviewsWithoutAnimations:(BOOL)removeSubviewsWithoutAnimations;
 {
   if (!view.reactTag) {
     return NO;
   }
   BOOL hasExitAnimation = _hasAnimationForTag(view.reactTag, @"exiting") || [_exitingViews objectForKey:view.reactTag];
   BOOL hasAnimatedChildren = NO;
-  removeImmediately = removeImmediately && !hasExitAnimation;
+  removeSubviewsWithoutAnimations = removeSubviewsWithoutAnimations && !hasExitAnimation;
   NSMutableArray *toBeRemoved = [[NSMutableArray alloc] init];
 
   for (UIView *subview in [view.reactSubviews copy]) {
-    if ([self removeRecursive:subview fromContainer:view withoutAnimation:removeImmediately]) {
+    if ([self startAnimationsRecursive:subview removingSubviewsWithoutAnimations:removeSubviewsWithoutAnimations]) {
       hasAnimatedChildren = YES;
-    } else if (removeImmediately) {
+    } else if (removeSubviewsWithoutAnimations) {
       [toBeRemoved addObject:subview];
     }
   }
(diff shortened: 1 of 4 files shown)


ios/LayoutReanimation/REAUIManager.mm
```diff
@@ -102,48 +73,27 @@ - (void)_manageChildren:(NSNumber *)containerTag
...
-      for (UIView *removedChild in permanentlyRemovedChildren) {
-        if ([_animationsManager wantsHandleRemovalOfView:removedChild]) {
-          wantProxyRemoval = YES;
-          break;
-        }
-        [_animationsManager removeAnimationsFromSubtree:removedChild];
-      }
-      if (wantProxyRemoval) {
-        // set layout animation group
-        [super setNextLayoutAnimationGroup:_reactLayoutAnimationGroup];
```
kmagiera, line 134: I think we can drop this line now too: https://github.com/software-mansion/react-native-reanimated/blob/main/src/reanimated2/layoutReanimation/animationsManager.ts#L11

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -335,14 +329,29 @@ - (BOOL)removeRecursive:(UIView *)view fromContainer:(UIView *)container without
...
+  }
+  for (int i = 0; i < children.count; i++) {
+    id<RCTComponent> child = [children objectAtIndex:i];
+    if (![child isKindOfClass:[UIView class]]) {
+      continue;
+    }
+    UIView *childView = (UIView *)child;
+    NSNumber *originalIndex = [indices objectAtIndex:i];
+    if ([self startAnimationsRecursive:childView removingSubviewsWithoutAnimations:true]) {
+      [(UIView *)container insertSubview:childView atIndex:[originalIndex intValue]];
```
kmagiera, line 354: Is it guaranteed here that indices here are sorted? If not the inserting will put things in a different order or in the worst case we could even try to insert at index that's much greater than the current number of subviews

kmagiera (approved):
This loosk good. Left two minor comments, please address them before merging

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -279,20 +279,20 @@ - (void)maybeDropAncestors:(UIView *)child
   }
 }
 
-- (BOOL)removeRecursive:(UIView *)view fromContainer:(UIView *)container withoutAnimation:(BOOL)removeImmediately;
+- (BOOL)startAnimationsRecursive:(UIView *)view removingSubviewsWithoutAnimations:(BOOL)removeSubviewsWithoutAnimations;
```
tomekzaw, line 282: There's an inconsistency in the method signature - `removing` vs. `remove`. I'd like to suggest renaming both to `shouldRemoveSubviewsWithoutAnimations` ("remove" is a verb and those are commonly used for method names, while "should" indicates that it's a flag).

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -279,20 +279,22 @@ - (void)maybeDropAncestors:(UIView *)child
   }
 }
 
-- (BOOL)removeRecursive:(UIView *)view fromContainer:(UIView *)container withoutAnimation:(BOOL)removeImmediately;
+- (BOOL)startAnimationsRecursive:(UIView *)view
+    removingSubviewsWithoutAnimations:(BOOL)shouldRemoveSubviewsWithoutAnimations;
```
tomekzaw, line 283:
```suggestion
    shouldRemoveSubviewsWithoutAnimations:(BOOL)shouldRemoveSubviewsWithoutAnimations;
```

piaskowyk (approved):

nandorojo:
Hey guys, just checking if this has been released or if it will be in the next `rc`, thanks.

kmagiera:
@nandorojo This hasn't been included in a release yet, we perhaps want to include one more fix (#3849) before publishing

PR #2561, opened 2021-10-21
[Reanimated + Screens] Shared Element Transition

piaskowyk:
## Description

Shared Element Transition based on Reanimated + Screens

## API
```js
```

## TODO
- [ ] //

## Roadmap
- Animation for all styles
- Animation with navigation header
- Screen transition navigation

Related PR to Screens: https://github.com/software-mansion/react-native-screens/pull/1591

Co-authored-by: @piaskowyk @WoLewicki

PR #3827, opened 2022-12-01
Shared Element Transition

piaskowyk:
## Summary
Shared Element Transition - allows you to morph component on screen into component on another screen

## How to use it?
You just need to add the same `sharedTransitionTag` for two components and when you change the screen the shared transition animation will automatically run.

**Screen A**
```js
<View 
  sharedTransitionTag="reanimatedTransition" // <--
  style={{ width: 100, height: 200 }}
/>
```

**Screen B**
```js
<View 
  sharedTransitionTag="reanimatedTransition" // <--
  style={{ width: 200, height: 100 }}
/>
```

## How to add custom animation?
1. You need to create custom transition with Renimated API, for example:
```js
const transition = SharedTransition.custom((values) => {
  'worklet';
  return {
  };
});
```
2. You need to add custom transition as `sharedTransitionStyle` prop to your both components:
```js
<View 
  sharedTransitionTag="reanimatedTransition"
  sharedTransitionStyle={transition} // <--
  style={{ width: 200, height: 100 }}
/>
```

## Examples
**// TODO: add video**

<details>
<summary>Simple example</summary>

```js
import { NavigationContainer } from '@react-navigation/native';
import { createNativeStackNavigator } from 'react-native-screens/native-stack';

function Screen1({ navigation }) {
  return (
    <Animated.ScrollView style={{ flex: 1 }}>
      <Animated.View
        style={{ 
          width: '100%', 
          height: 100, 
          backgroundColor: 'green' 
        }}
        sharedTransitionTag="reanimatedTransition"
      />
      <Button
        onPress={() => navigation.navigate('Screen2')}
        title="go to screen2"
      />
    </Animated.ScrollView>
  );
}

function Screen2({ navigation }) {
  return (
    <View style={{ flex: 1 }}>
      <Animated.View
        style={{
          width: '100%',
          height: 200,
          marginTop: 200,
          backgroundColor: 'green'
        }}
        sharedTransitionTag="reanimatedTransition"
      />
      <Button
        title="go to screen1"
        onPress={() => navigation.navigate('Screen1')}
      />
    </View>
  );
}

export function SharedTransitionExample() {
  return (
    <NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen
          name="Screen1"
          component={Screen1}
          options={{ headerShown: false }}
        />
        <Stack.Screen
          name="Screen2"
          component={Screen2}
          options={{ headerShown: false }}
        />
      </Stack.Navigator>
    </NavigationContainer>
  );
}

```

</details>

More examples you can find here: https://github.com/software-mansion/react-native-reanimated/pull/3857

## More info
Successor of: https://github.com/software-mansion/react-native-reanimated/pull/2561


Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -293,11 +293,11 @@ void RuntimeDecorator::decorateUIRuntime(
                   const jsi::Value &thisValue,
                   const jsi::Value *args,
                   size_t count) -> jsi::Value {
-    progressLayoutAnimationFunction(args[0].asNumber(), args[1].asObject(rt));
+    progressLayoutAnimationFunction(args[0].asNumber(), args[1].asObject(rt), args[2].asBool());
```
kmagiera, line 296: If I remember correctly asBool was only added in RN 69 and we should use getBool instead

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -16,6 +17,20 @@ void LayoutAnimationsManager::configureAnimation(
     exitingAnimations_[tag] = config;
   } else if (type == "layout") {
     layoutAnimations_[tag] = config;
+  } else if (type == "sharedElementTransition") {
+    sharedTransitionAnimations_[tag] = config;
+    if (sharedTransitionGroups_.find(sharedTransitionTag) ==
```
kmagiera, line 22: use `contains` instead `find` + `end` as it more clearly describes the intent.

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -29,6 +44,9 @@ bool LayoutAnimationsManager::hasLayoutAnimation(
     return exitingAnimations_.find(tag) != exitingAnimations_.end();
   } else if (type == "layout") {
     return layoutAnimations_.find(tag) != layoutAnimations_.end();
+  } else if (type == "sharedElementTransition") {
+    return sharedTransitionAnimations_.find(tag) !=
```
kmagiera, line 48: `contains`

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -38,6 +56,22 @@ void LayoutAnimationsManager::clearLayoutAnimationConfig(int tag) {
   enteringAnimations_.erase(tag);
   exitingAnimations_.erase(tag);
   layoutAnimations_.erase(tag);
+
+  sharedTransitionAnimations_.erase(tag);
+  for (auto &[key, group] : sharedTransitionGroups_) {
+    bool isCurrentGroupMember = false;
+    int indexInGroup;
```
kmagiera, line 63: Would be shorter to use `std::remove` here, I think something like this should be enough to replace the whole body of this loop:
```
std::remove(group.begin(), group.end(), tag)
```

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -73,4 +109,23 @@ void LayoutAnimationsManager::startLayoutAnimation(
       config->getJSValue(rt));
 }
 
+int LayoutAnimationsManager::findTheOtherForSharedTransition(int tag) {
+  int theOtherTag = -1;
+  for (auto const &[key, group] : sharedTransitionGroups_) {
+    bool isCurrentGroupMember = false;
+    int indexInGroup;
```
kmagiera, line 116: same as above, use `std::remove`

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -491,12 +516,19 @@ private void cancelAnimationsInSubviews(ViewGroup view) {
...
+        }
+      }
     }
+
     try {
       return mUIManager.resolveView(tag);
     } catch (IllegalViewOperationException e) {
       // view has been removed already
-      e.printStackTrace();
+      //      e.printStackTrace();
```
kmagiera, line 531: is this change intended/necessary?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -505,8 +537,36 @@ private ViewManager resolveViewManager(int tag) {
     try {
       return mReanimatedNativeHierarchyManager.resolveViewManager(tag);
     } catch (Exception e) {
-      e.printStackTrace();
+      //      e.printStackTrace();
```
kmagiera, line 540: same here

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -505,8 +537,36 @@ private ViewManager resolveViewManager(int tag) {
...
       return null;
     }
   }
+
+  private Point convertPoint(Point fromPoint, View parentView) {
```
kmagiera, line 545: how does it convert the point? Maybe rename to sth that'd give better intuition about that, for example `convertRelativePointToScreenLocation`

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -62,25 +62,40 @@ public class Snapshot {
...
+  }
+
+  public Snapshot(View view) {
+    int[] location = new int[2];
+    view.getLocationOnScreen(location);
+    originX = location[0];
+    originY = location[1];
+    width = view.getWidth();
+    height = view.getHeight();
+    initKeys();
```
kmagiera, line 75: what's the point of initializing static fields in instance constructor. It seems like this should be initialized statically (e.g. use `static` block in this class or just assign in-place with the field definition.

src/reanimated2/NativeReanimated/NativeReanimated.ts
```diff
@@ -110,9 +110,15 @@ export class NativeReanimated {
   configureLayoutAnimation(
     viewTag: number,
     type: string,
+    sharedTransitionTag = '',
```
kmagiera, line 113: seems wrong to have a default value for a second to last parameter of a function but not for the last one. Also empty string does not seem like a great default value. It looks like we don't really need this default value at all?

src/reanimated2/core.ts
```diff
@@ -248,11 +248,13 @@ export function enableLayoutAnimations(
 export function configureLayoutAnimations(
   viewTag: number,
   type: string,
-  config: LayoutAnimationFunction | Keyframe
+  config: LayoutAnimationFunction | Keyframe,
+  sharedTransitionTag = ''
```
kmagiera, line 252: ok, there is a default value here so I don't think we need one in the NativeReanimatedModule

kmagiera (commented):
Leaving some comments for the files I reviewed so far

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
jwajgelt: Isn't that available only since C++20?
https://en.cppreference.com/w/cpp/container/map/contains

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
jwajgelt: From "Notes" in https://en.cppreference.com/w/cpp/algorithm/remove
"These algorithms cannot be used with associative containers such as [std::set](https://en.cppreference.com/w/cpp/container/set) and [std::map](https://en.cppreference.com/w/cpp/container/map)"

jwajgelt: This has to be followed by `std::erase(it_returned_from_remove, group.end())`, since `std::remove` doesn't actually remove the elements, only moves them to the end of the container.
- this should also be faster than the current approach, since you only do one swap and one `pop_back()`, instead of shifting all the elements of `group` after the removed one by one slot to the left
- this however means that the relative order of elements in `group` isn't preserved, if that's what you wanted. In that case, the code would still be cleaner by replacing the `for`-loop with a call to `std::find`

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
jwajgelt: But this doesn't remove anything? 🤔 
Still, you can replace the `for`-loop with `std::find`

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
jwajgelt: Also, in this case:
https://en.cppreference.com/w/cpp/container/map/insert
"Inserts element(s) into the container, if the container doesn't already contain an element with an equivalent key."

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,287 @@
...
+
+  protected void viewsDidRemoved() {
+    setupStaredTransitionForViews(mRemovedSharedViews, false);
+    for (View removedSharedView : mRemovedSharedViews) {
+      visitTree(removedSharedView.getId(), true);
+    }
+    mRemovedSharedViews.clear();
+  }
+
+  private void setupStaredTransitionForViews(List<View> sharedViews, boolean withNewElements) {
```
jwajgelt, line 59: Is this a typo?
```suggestion
  private void setupStartedTransitionForViews(List<View> sharedViews, boolean withNewElements) {
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,287 @@
...
+        viewTags.add(view.getId());
+      }
+    }
+    List<SharedElement> sharedElements = new ArrayList<>();
+    ReanimatedNativeHierarchyManager reanimatedNativeHierarchyManager =
+        mAnimationsManager.getReanimatedNativeHierarchyManager();
+    for (View sharedView : sharedViews) {
+      int targetViewTag =
+          mAnimationsManager
+              .getNativeMethodsHolder()
```
jwajgelt, line 92: This getter is called a lot. Maybe it's worth storing the `NativeMethodsHolder` in a field?

ios/LayoutReanimation/REAAnimationsManager.h
```diff
@@ -36,6 +41,19 @@ typedef void (^REAAnimationRemovingBlock)(NSNumber *_Nonnull tag);
 - (void)removeChildren:(NSArray<UIView *> *)children fromContainer:(UIView *)container;
 - (void)onViewCreate:(UIView *)view after:(REASnapshot *)after;
 - (void)onViewUpdate:(UIView *)view before:(REASnapshot *)before after:(REASnapshot *)after;
+- (void)setFindTheOtherForSharedTransitionBlock:
```
jwajgelt, line 44: Can we group this with the other block setters?

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -64,6 +67,7 @@ - (instancetype)initWithUIManager:(RCTUIManager *)uiManager
       [_targetKeys addObject:[NSString stringWithFormat:@"target%@", [key capitalizedString]]];
       [_currentKeys addObject:[NSString stringWithFormat:@"current%@", [key capitalizedString]]];
     }
+    _sharedTransitionManager = [[REASharedTransitionManager new] init:self];
```
jwajgelt, line 70: I think that the objc convention is for `new` to be the same as `[[x alloc] init]`, so this looks like initialising the same object twice.
Not sure it breaks anything here, but it may be better to replace the `new` here with `alloc`.

ios/LayoutReanimation/REASharedTransitionManager.h
```diff
@@ -0,0 +1,19 @@
...
+@interface REASharedTransitionManager : NSObject
+
+- (void)notifyAboutNewView:(UIView *)view;
+- (void)viewsDidLayout;
+- (void)setupSyncSharedTransitionForViews:(NSArray<UIView *> *)views;
+- (void)finishSharedAnimation:(UIView *)view;
+- (NSArray<UIView *> *)getCurrentSharedTransitionViews;
+- (void)setFindTheOtherForSharedTransitionBlock:
+    (REAFindTheOtherForSharedTransitionBlock)findTheOtherForSharedTransition;
+- (instancetype)init:(REAAnimationsManager *)animationManager;
```
jwajgelt, line 15:
```suggestion
- (instancetype)initWithAnimationsManager:(REAAnimationsManager *)animationManager;
```

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,373 @@
...
+- (NSArray *)sortViewsByTags:(NSArray *)views
+{
+  return [views sortedArrayUsingComparator:^NSComparisonResult(UIView *view1, UIView *view2) {
+    if ([view1.reactTag intValue] > [view2.reactTag intValue]) {
+      return (NSComparisonResult)NSOrderedAscending;
+    }
+    if ([view1.reactTag intValue] < [view2.reactTag intValue]) {
+      return (NSComparisonResult)NSOrderedDescending;
+    }
+    return (NSComparisonResult)NSOrderedSame;
```
jwajgelt, line 88: Why not use NSNumber's `compare:`? https://developer.apple.com/documentation/foundation/nsnumber/1413562-compare?language=objc ?

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,373 @@
...
+                                                                targetViewSnapshot:targetViewSnapshot];
+    [sharedElements addObject:sharedElement];
+  }
+  return sharedElements;
+}
+
+- (UIView *)getParentScreen:(UIView *)view
+{
+  UIView *screen = view;
+  while (![NSStringFromClass([screen class]) isEqualToString:@"RNSScreenView"] && screen.superview != nil) {
```
jwajgelt, line 153: Is this really the best way of checking if the view is `RNSScreenView`?
Why not use `isMemberOfClass:` or `isKindOfClass:`?

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,373 @@
...
+                           }
+                           return false;
+                         }];
+  [self setupSyncSharedTransitionForViews:removedView];
+  for (UIView *view in removedView) {
+    [_animationManager clearAnimationConfigForTag:view.reactTag];
+  }
+}
+
+- (void)runAsyncStaredTransition
```
jwajgelt, line 260:
```suggestion
- (void)runAsyncSharedTransition
```

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -73,4 +98,18 @@ void LayoutAnimationsManager::startLayoutAnimation(
...
+int LayoutAnimationsManager::findTheOtherForSharedTransition(int tag) {
+  int theOtherTag = -1;
+  for (auto const &[key, group] : sharedTransitionGroups_) {
+    auto position = std::find(group.begin(), group.end(), tag);
+    if (position != group.end()) {
+      int index = position - group.begin();
+      if (index > 0) {
+        return group[index - 1]; // get one before
+      }
+    }
```
jwajgelt, line 110: Nit:
```suggestion
    if (position != group.end() && position != group.begin()) {
      return *std::prev(position);
    }
```
https://en.cppreference.com/w/cpp/iterator/prev
I think it might be more readable this way, It's fine if you prefer the index approach though.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,287 @@
...
+  protected List<View> getCurrentSharedTransitionViews() {
+    return mCurrentSharedTransitionViews;
+  }
+
+  protected void viewsDidLayout() {
+    setupSharedTransitionForViews(mAddedSharedViews, true);
+    mAddedSharedViews.clear();
+  }
+
+  protected void viewsDidRemoved() {
```
jwajgelt, line 52: Maybe `viewsWereRemoved`? 🤔

android/src/reactNativeVersionPatch/nativeHierarchyManager/latest/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -339,6 +346,15 @@ public synchronized void manageChildren(
     super.manageChildren(tag, null, null, tagsToDelete);
   }
 
+  private void maybeNotifyAboutViewsRemoving(int[] tagsToDelete) {
+    if (tagsToDelete != null) {
+      AnimationsManager animationsManager = ((ReaLayoutAnimator) mReaLayoutAnimator).getAnimationsManager();
+      animationsManager.visitTreeForTags(tagsToDelete, false);
```
kmagiera, line 352: looks like it'd be better to move this logic to AnimationsManager rather than exposing implemnentation details to other classes.

kmagiera (commented):
I think that a better naming for `find-the-other` would be something along lines: "findSharedTransitionTargetForView" and then we should also use "viewTag" for the variable as opposed to just "tag" such that it doesn't get confused with the shared transition name tags

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,512 @@
...
+    [self swizzlMethod:@selector(viewDidLayoutSubviews)
+                  with:@selector(swizzled_viewDidLayoutSubviews)
+              forClass:[viewController class]];
+    [self swizzlMethod:@selector(notifyWillDisappear)
+                  with:@selector(swizzled_notifyWillDisappear)
+              forClass:[view class]];
+  });
+}
+
+- (void)swizzlMethod:(SEL)originalSelector with:(SEL)swizzledSelector forClass:(Class)originalClass
```
WoLewicki, line 237:
```suggestion
- (void)swizzleMethod:(SEL)originalSelector with:(SEL)swizzledSelector forClass:(Class)originalClass
```

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,512 @@
...
+  while (stack != nil && ![stack isKindOfClass:[RNSScreenStackView class]]) {
+    stack = stack.superview;
+  }
+
+  if ([keyPath isEqualToString:@"transitionDuration"]) {
+    // added new screen (push)
+    if (screen.superview != nil) {
+      [self runAsyncSharedTransition];
+    }
+  } else if ([keyPath isEqualToString:@"activityState"]) {
```
WoLewicki, line 290: `activityState` is ever changed in `native-stack`? I thought it is only used in other navigators under `RNSScreenContainerView`.

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,512 @@
...
+      return NO;
+    }
+  }
+  return YES;
+}
+
+- (BOOL)isScreen:(UIView *)screen onTopOfStack:(UIView *)stack
+{
+  int screenCount = stack.reactSubviews.count;
+  return screenCount > 0 && screen == stack.reactSubviews[screenCount - 1];
```
WoLewicki, line 339: I think you can use `reactSubviews.lastObject` instead of `reactSubviews[screenCount - 1]`.

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,512 @@
...
+      return YES;
+    }
+    stack = screen.reactSuperview;
+  }
+  return NO;
+}
+
+- (void)runSharedTransitionForSharedViewsOnScreen:(UIView *)screen
+{
+  NSMutableArray<UIView *> *removedView = [NSMutableArray new];
```
WoLewicki, line 360:
```suggestion
  NSMutableArray<UIView *> *removedViews = [NSMutableArray new];
```

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,512 @@
...
+  [self setupTransitionContainer];
+  [self reparentSharedViewsForCurrentTransition:_sharedElements];
+  [self startSharedTransition:_sharedElements];
+  [_addedSharedViews removeAllObjects];
+  [_sharedElements removeObjectsInArray:currentSharedElements];
+}
+
+- (void)setupTransitionContainer
+{
+  if (_isSharedTransitionActive == NO) {
```
WoLewicki, line 403:
```suggestion
  if (!_isSharedTransitionActive) {
```

WoLewicki (commented):
It gets better and better with each iteration! I left some comments.

ios/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: You're right but I'm changing it in `swizzled_notifyWillDisappear` method.

android/src/reactNativeVersionPatch/nativeHierarchyManager/latest/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -230,6 +236,27 @@ private boolean isLayoutAnimationDisabled() {
...
+    if (isLayoutAnimationDisabled()) {
+      return;
+    }
+    try {
+      ViewManager viewManager = this.resolveViewManager(tag);
+      String viewManagerName = viewManager.getName();
+      View container = resolveView(parentTag);
+      if (container != null
+          && viewManagerName.equals("RNSScreen")
+          && this.mReaLayoutAnimator != null) {
```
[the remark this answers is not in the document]
piaskowyk: We don't need `this`. I'll remove it.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,387 @@
...
+    if (viewGroupManager == null) {
+      return;
+    }
+    for (int i = 0; i < viewGroupManager.getChildCount(viewGroup); i++) {
+      View child = viewGroupManager.getChildAt(viewGroup, i);
+      visitTree(child.getId(), clearConfig);
+    }
+  }
+
+  void visitTreeAndMakeSnapshot(View view) {
```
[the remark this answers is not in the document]
piaskowyk: This is a bug, I forgot to do snapshot 🤦‍♂️

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -73,4 +98,18 @@ void LayoutAnimationsManager::startLayoutAnimation(
...
 }
 
+int LayoutAnimationsManager::findTheOtherForSharedTransition(int tag) {
+  int theOtherTag = -1;
+  for (auto const &[key, group] : sharedTransitionGroups_) {
+    auto position = std::find(group.begin(), group.end(), tag);
+    if (position != group.end()) {
+      int index = position - group.begin();
+      if (index > 0) {
+        return group[index - 1]; // get one before
```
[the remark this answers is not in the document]
piaskowyk: The top screen on the stack triggers the animation, so we need to find the sibling view registered in the past. That's why we look backward.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,287 @@
...
+
+  protected void setNativeMethods(NativeMethodsHolder nativeMethods) {
+    mNativeMethodsHolder = nativeMethods;
+  }
+
+  private void setupSharedTransitionForViews(List<View> sharedViews, boolean withNewElements) {
+    if (sharedViews.isEmpty()) {
+      return;
+    }
+    sortViewsByTags(sharedViews);
```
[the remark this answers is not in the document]
piaskowyk: I added a comment in the code to explain it:
```
All shared views during the transition have the same parent. It is problematic if parent
view and their children are in the same transition. To keep the valid order in the z-axis,
we need to sort views by tags. Parent tag is lower than children tags.
```

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -16,6 +18,11 @@ void LayoutAnimationsManager::configureAnimation(
     exitingAnimations_[tag] = config;
   } else if (type == "layout") {
     layoutAnimations_[tag] = config;
+  } else if (type == "sharedElementTransition") {
+    sharedTransitionAnimations_[tag] = config;
+    sharedTransitionGroups_.try_emplace(sharedTransitionTag);
+    auto &group = sharedTransitionGroups_[sharedTransitionTag];
+    group.push_back(tag);
```
tomekzaw, line 25:
```suggestion
    sharedTransitionGroups_[sharedTransitionTag].push_back(tag);
```

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -38,6 +47,14 @@ void LayoutAnimationsManager::clearLayoutAnimationConfig(int tag) {
   enteringAnimations_.erase(tag);
   exitingAnimations_.erase(tag);
   layoutAnimations_.erase(tag);
+
+  sharedTransitionAnimations_.erase(tag);
+  for (auto &[key, group] : sharedTransitionGroups_) {
+    auto position = std::find(group.begin(), group.end(), tag);
+    if (position != group.end()) {
+      group.erase(position);
+    }
```
tomekzaw, line 56:
```suggestion
    group.erase(tag);
```

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -73,4 +91,18 @@ void LayoutAnimationsManager::startLayoutAnimation(
...
+int LayoutAnimationsManager::findSiblingForSharedView(int tag) {
+  /*
+    The top screen on the stack triggers the animation, so we need to find
+    the sibling view registered in the past. That's why we look backward.
+   */
+  for (auto const &[key, group] : sharedTransitionGroups_) {
+    auto position = std::find(group.begin(), group.end(), tag);
+    if (position != group.end() && position != group.begin()) {
+      return *std::prev(position);
+    }
```
tomekzaw, line 103: 1. I don't fully understand what this function is supposed to do.
2. What about `key`? Is it needed?
3. It looks like this can be done easier

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -369,10 +369,12 @@ jsi::Value NativeReanimatedModule::configureLayoutAnimation(
     jsi::Runtime &rt,
     const jsi::Value &viewTag,
     const jsi::Value &type,
+    const jsi::Value &sharedTransitionTag,
     const jsi::Value &config) {
   layoutAnimationsManager_.configureAnimation(
       viewTag.asNumber(),
       type.asString(rt).utf8(rt),
+      sharedTransitionTag.asString(rt).utf8(rt),
```
tomekzaw, line 377: Does this also apply to regular layout animations?

Common/cpp/Tools/CollectionUtils.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <algorithm>
+
+namespace reanimated {
+namespace collection {
+
+template <class CollectionType, class ValueType>
+inline bool mapContains(CollectionType &collection, const ValueType &value) {
```
tomekzaw, line 9:
```suggestion
inline bool contains(CollectionType &collection, const ValueType &value) {
```

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -370,6 +372,12 @@ void NativeProxy::installJSIBindings(
             tag);
       });
 
+  layoutAnimations->cthis()->setFindSiblingForSharedView([weakModule](int tag) {
+    return weakModule.lock()
+        ->layoutAnimationsManager()
+        .findSiblingForSharedView(tag);
```
tomekzaw, line 378: `weakModule.lock()` may return `nullptr`
```suggestion
    if (auto module = weakModule.lock()) {
      return module->layoutAnimationsManager().findSiblingForSharedView(tag);
    } else {
      return -1;
    }
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -340,7 +357,8 @@ public void updateLayout(
       float xf,
       float yf,
       float widthf,
-      float heightf) {
+      float heightf,
+      boolean positionIsAbsolute) {
```
tomekzaw, line 361:
```suggestion
      boolean isPositionAbsolute) {
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -542,4 +578,32 @@ private ViewManager resolveViewManager(int tag) {
       return null;
     }
   }
+
+  private Point convertAbsoluteToParentRelative(Point fromPoint, View parentView) {
+    int[] toCord = {0, 0};
```
tomekzaw, line 583:
```suggestion
    int[] toPoint = {0, 0};
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -41,9 +41,26 @@ public class Snapshot {
   public int originY;
   public int globalOriginX;
   public int globalOriginY;
+  public int originYByParent;
```
tomekzaw, line 44: I don't like this name

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -14,22 +14,22 @@ public class Snapshot {
   public static final String HEIGHT = "height";
   public static final String ORIGIN_X = "originX";
   public static final String ORIGIN_Y = "originY";
-  public static final String GLOBAL_ORIGIN_X = "globalOriginX";
-  public static final String GLOBAL_ORIGIN_Y = "globalOriginY";
+  public static final String ABSOLUTE_ORIGIN_X = "globalOriginX";
```
tomekzaw, line 17: This rename introduces inconsistency between identifier and actual value, what's the purpose of doing so?

src/reanimated2/layoutReanimation/index.ts
```diff
@@ -2,3 +2,4 @@ import './animationsManager';
 export * from './animationBuilder';
 export * from './defaultAnimations';
 export * from './defaultTransitions';
+export * from './sharedTransition';
```
tomekzaw, line 5:
```suggestion
export * from './sharedTransitions';
```

src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
```diff
@@ -43,6 +43,7 @@ export type EntryExitAnimationFunction = (
 export type AnimationConfigFunction<T> = (targetValues: T) => LayoutAnimation;
 
 export interface LayoutAnimationsValues {
+  [key: string]: any;
```
tomekzaw, line 46: Why?

src/reanimated2/core.ts
tomekzaw: I don't like the the fact that regular layout animations will receive empty string as `sharedTransitionTag`, can we use `undefined` instead?

ios/native/NativeProxy.mm
```diff
@@ -400,6 +402,14 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
   [animationsManager setAnimationRemovingBlock:^(NSNumber *_Nonnull tag) {
     weakModule.lock()->layoutAnimationsManager().clearLayoutAnimationConfig([tag intValue]);
   }];
+
+  [animationsManager setFindSiblingForSharedViewBlock:^NSNumber *_Nullable(NSNumber *_Nonnull tag) {
+    int resultTag = weakModule.lock()->layoutAnimationsManager().findSiblingForSharedView([tag intValue]);
```
tomekzaw, line 407: `weakModule.lock()` may be null, you need a conditional here

ios/LayoutReanimation/REASnapshot.m
```diff
@@ -24,6 +24,29 @@ - (instancetype)init:(UIView *)view
...
+  _values[@"originX"] = [NSNumber numberWithDouble:absolutePosition.x - view.bounds.size.width / 2.0];
+  _values[@"originY"] = [NSNumber numberWithDouble:absolutePosition.y - view.bounds.size.height / 2.0];
+  _values[@"originYByParent"] = [NSNumber numberWithDouble:view.center.y - view.bounds.size.height / 2.0];
+  _values[@"globalOriginX"] = _values[@"originX"];
+  _values[@"globalOriginY"] = _values[@"originY"];
+  _values[@"globalOriginX"] = [NSNumber numberWithDouble:absolutePosition.x - view.bounds.size.width / 2.0];
+  _values[@"globalOriginY"] = [NSNumber numberWithDouble:absolutePosition.y - view.bounds.size.height / 2.0];
+  _values[@"windowWidth"] = [NSNumber numberWithDouble:mainWindow.bounds.size.width];
+  _values[@"windowHeight"] = [NSNumber numberWithDouble:mainWindow.bounds.size.height];
+  //  UIColor *color = view.backgroundColor;
```
tomekzaw, line 45: remove this

ios/LayoutReanimation/REASnapshot.m
```diff
@@ -24,6 +24,29 @@ - (instancetype)init:(UIView *)view
...
+  _values = [NSMutableDictionary new];
+  _values[@"width"] = [NSNumber numberWithDouble:(double)(view.bounds.size.width)];
+  _values[@"height"] = [NSNumber numberWithDouble:(double)(view.bounds.size.height)];
+  _values[@"originX"] = [NSNumber numberWithDouble:absolutePosition.x - view.bounds.size.width / 2.0];
+  _values[@"originY"] = [NSNumber numberWithDouble:absolutePosition.y - view.bounds.size.height / 2.0];
+  _values[@"originYByParent"] = [NSNumber numberWithDouble:view.center.y - view.bounds.size.height / 2.0];
+  _values[@"globalOriginX"] = _values[@"originX"];
+  _values[@"globalOriginY"] = _values[@"originY"];
+  _values[@"globalOriginX"] = [NSNumber numberWithDouble:absolutePosition.x - view.bounds.size.width / 2.0];
+  _values[@"globalOriginY"] = [NSNumber numberWithDouble:absolutePosition.y - view.bounds.size.height / 2.0];
```
tomekzaw, line 42: Why do we overwrite the same keys with different values?

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
piaskowyk: Unfortunately, It isn't possible to use `eraser` to remove item from `vector` by value.

Common/cpp/NativeModules/NativeReanimatedModule.cpp
piaskowyk: Yes, but regular layout animation doesn't use this argument. I didn't want to add another JSI function to almost the same registration process, so i added additional argument for existing one.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
piaskowyk: @kmagiera don't like `GLOBAL` prefix - https://github.com/software-mansion/react-native-reanimated/pull/3827#discussion_r1054337375 but if we want to change the string `globalOriginX` this requires a lot of renaming in JS code (types and predefined animations) which is definitely out of the scope of this PR.

src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
piaskowyk: I changed it to `[key: string]: number;`, but we need to `[key: string]` because I iterate over this object in `/src/reanimated2/layoutReanimation/sharedTransitions/index.ts`

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -16,6 +18,10 @@ void LayoutAnimationsManager::configureAnimation(
     exitingAnimations_[tag] = config;
   } else if (type == "layout") {
     layoutAnimations_[tag] = config;
+  } else if (type == "sharedElementTransition") {
+    sharedTransitionAnimations_[tag] = config;
+    sharedTransitionGroups_.try_emplace(sharedTransitionTag);
+    sharedTransitionGroups_[sharedTransitionTag].push_back(tag);
```
kmagiera, line 24: minor update would be to use the iterator returned from `emplace` in order to push the value. Sth along these lines should work:
```
map.try_emplace(tag).first->second.push_back(tag)
```
add an appropriate comment that because otherwise it is hardly readable but avoids the second map lookup

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
kmagiera: I'd suggest to call it differently. Also typically it's better to add method-level documentation prior to the definition not within the method.

I think that "sibling" doesn't capture the meaning of what we are searching here in this method. I'd call it "findPrecedingViewTagForTransition" and describe that it looks for a veiw tegistered in the same transition group (that is with the same tag) which has been added to that group directly before the one that we provide as an argument.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -183,11 +194,11 @@ public void progressLayoutAnimation(int tag, Map<String, Object> newStyle) {
     ViewManager viewManager = resolveViewManager(tag);
     ViewManager parentViewManager = resolveViewManager(parent.getId());
 
-    if (viewManager == null || parentViewManager == null) {
+    if (viewManager == null) {
```
kmagiera, line 197: why do we remove that parentViewManager check? is it ok to send null parentViewManager to setnewProps now?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
kmagiera: maybe just revert these changes. I didn't realize before that these weren't introduced in this diff. If it's outside ot the scope of this change maybe lets keep it the way it was. I don't like these names but such long diff is not a good place to refactor such things anyways.

kmagiera (commented):
not done reviewing this but need to take a break and wanted to post the comments I've collected so far.

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
jwajgelt: [unordered_map::operator[]](https://en.cppreference.com/w/cpp/container/unordered_map/operator_at) "is equivalent to `return this->try_emplace(std::move(key)).first->second;`. (since C++17)", so this can be written as simply `sharedTransitionGroups_[sharedTransitionTag].push_back(tag);`, without the `try_emplace`.

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
jwajgelt: I agree with @tomekzaw here, I think it would be better to have a map `tag -> sharedTransitionTag` and use it to find the shared transition group for the tag, instead of looping.

Also, the naming here isn't perfect as @kmagiera noted.

src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
jwajgelt: Would `[key: string]: unknown` work here instead? Provided we don't actually care about the types of values in this object.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,400 @@
...
+  protected View getTransitioningView(int tag) {
+    return mCurrentSharedTransitionViews.get(tag);
+  }
+
+  protected void viewsDidLayout() {
+    setupSharedTransitionForViews(mAddedSharedViews, true);
+    mAddedSharedViews.clear();
+  }
+
+  protected void onViewsRemoving(int[] tagsToDelete) {
```
jwajgelt, line 56:
```suggestion
  protected void onViewsRemoval(int[] tagsToDelete) {
```
or maybe
```suggestion
  protected void notifyAboutViewsRemoval(int[] tagsToDelete) {
```
I'm not really sure, I just don't like the current name. 😛

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,400 @@
...
+  protected void makeSnapshot(View view) {
+    Snapshot snapshot = new Snapshot(view);
+    View screen = findScreen(view);
+    if (screen != null) {
+      snapshot.originY -= screen.getTop();
+    }
+    mSnapshotRegistry.put(view.getId(), snapshot);
+  }
+
+  protected void visitTreeForTags(int[] viewTags, boolean clearConfig) {
```
jwajgelt, line 338: The `clearConfig` param is never used, right?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,400 @@
...
+          (ViewGroupManager) reanimatedNativeHierarchyManager.resolveViewManager(tag);
+    } catch (IllegalViewOperationException e) {
+      return;
+    }
+    if (viewGroupManager == null) {
+      return;
+    }
+    for (int i = 0; i < viewGroupManager.getChildCount(viewGroup); i++) {
+      View child = viewGroupManager.getChildAt(viewGroup, i);
+      visitTree(child.getId(), clearConfig);
```
jwajgelt, line 379: We call `getId()` on child, only to then use the tag to get back the view. Wouldn't it be better to instead take the `View` as argument?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,400 @@
...
+  protected void visitTreeForTags(int[] viewTags, boolean clearConfig) {
+    if (viewTags == null) {
+      return;
+    }
+    for (int viewTag : viewTags) {
+      visitTree(viewTag, clearConfig);
+    }
+  }
+
+  private void visitTree(int tag, boolean clearConfig) {
```
jwajgelt, line 347: I would refactor this whole function a bit.

Like I mentioned in another comment, I don't think there's any need to take the `tag` here, we can pass the `View` instead and get the tag from it.

Another thing is, the `clearConfig` param here is only checked in one `if` statement, and passed down in recursion - it would be more readable to have two functions, one for `clearConfig` case and one for `!clearConfig` case instead.
Then, you'd get a bit of boilerplate for getting `viewGroupManager` and recursing into children common between those two functions, so what you can do instead is have the one `visitTree` function which takes the `View` to start from, and a callback to call on each of its descendants in the tree instead - again I think it would be more readable, than a single function doing two different things depending on a boolean flag.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,400 @@
...
+          sharedElement.sourceViewSnapshot,
+          sharedElement.targetViewSnapshot);
+      onViewTransition(
+          sharedElement.targetView,
+          sharedElement.sourceViewSnapshot,
+          sharedElement.targetViewSnapshot);
+    }
+  }
+
+  private void onViewTransition(View view, Snapshot before, Snapshot after) {
```
jwajgelt, line 251:
```suggestion
  private void startViewTransition(View view, Snapshot before, Snapshot after) {
```
might me a better name

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,400 @@
...
+  private void sortViewsByTags(List<View> views) {
+    /*
+      All shared views during the transition have the same parent. It is problematic if parent
+      view and their children are in the same transition. To keep the valid order in the z-axis,
+      we need to sort views by tags. Parent tag is lower than children tags.
+    */
+    Collections.sort(views, (v1, v2) -> Integer.compare(v2.getId(), v1.getId()));
+  }
+
+  private List<SharedElement> getSharedElementForCurrentTransition(
```
jwajgelt, line 111:
```suggestion
  private List<SharedElement> getSharedElementsForCurrentTransition(
```
or even
```suggestion
  private List<SharedElement> sharedElementsForCurrentTransition(
```

tomekzaw: Let's stick to the convention that method names start with a verb, so the first option works for me.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,400 @@
...
+        Log.e("[Reanimated]", "Unable to recognize screen on stack.");
+      }
+    }
+  }
+
+  protected void setNativeMethods(NativeMethodsHolder nativeMethods) {
+    mNativeMethodsHolder = nativeMethods;
+  }
+
+  private void setupSharedTransitionForViews(List<View> sharedViews, boolean withNewElements) {
```
jwajgelt, line 87:
```suggestion
  private void startSharedTransitionForViews(List<View> sharedViews, boolean withNewElements) {
```
since this actually _starts_ the transition, right?

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -106,7 +110,10 @@ - (UIView *)viewForTag:(NSNumber *)tag
 {
   UIView *view = [_reaUiManager viewForReactTag:tag];
   if (view == nil) {
-    return [_exitingViews objectForKey:tag];
+    view = [_exitingViews objectForKey:tag];
+  }
+  if (view == nil) {
+    view = [_sharedTransitionManager getTransitioningView:tag];
   }
```
jwajgelt, line 117: So...
this could be made more concise, with something along the lines of
```objective-c
UIView *view;
(view = [_reaUiManager viewForReactTag:tag]) || (view = [_exitingViews objectForKey:tag]) || (view = [_sharedTransitionManager getTransitioningView:tag]);
return view;
```
but I'm not sure if that's better than the chain of `if` statements.

We'd like to do
```objective-c
return [_reaUiManager viewForReactTag:tag] || [_exitingViews objectForKey:tag] || [_sharedTransitionManager getTransitioningView:tag];
```
but the short-circuiting logical operators (`||` and `&&`) return `int` values (as per the C standard), and it's illegal to cast them back to pointers (in Objective C)

ios/LayoutReanimation/REASharedTransitionManager.h
```diff
@@ -0,0 +1,19 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface REASharedTransitionManager : NSObject
+
+- (void)notifyAboutNewView:(UIView *)view;
+- (void)viewsDidLayout;
+- (BOOL)setupSyncSharedTransitionForViews:(NSArray<UIView *> *)views;
+- (void)finishSharedAnimation:(UIView *)view;
+- (NSArray<UIView *> *)getCurrentSharedTransitionViews;
```
jwajgelt, line 12: I think `getCurrentSharedTransitionViews` is neither implemented nor used anywhere (please correct me if I'm wrong)

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,515 @@
...
+
+- (void)swizzleMethod:(SEL)originalSelector with:(SEL)swizzledSelector forClass:(Class)originalClass
+{
+  Class class = [self class];
+  Method originalMethod = class_getInstanceMethod(originalClass, originalSelector);
+  Method swizzledMethod = class_getInstanceMethod(class, swizzledSelector);
+  IMP originalImp = method_getImplementation(originalMethod);
+  IMP swizzledImp = method_getImplementation(swizzledMethod);
+  class_replaceMethod(originalClass, swizzledSelector, originalImp, method_getTypeEncoding(originalMethod));
+  class_replaceMethod(originalClass, originalSelector, swizzledImp, method_getTypeEncoding(swizzledSelector));
```
jwajgelt, line 250: Should this be
```suggestion
  class_replaceMethod(originalClass, originalSelector, swizzledImp, method_getTypeEncoding(swizzledMethod));
```
?

XCode even complains about type mismatches here.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
piaskowyk: During shared transition views are attached to our custom transition container which is just a native `View` without `ReactViewManager`. I added null-check to the method where `parentViewManager` is used, so I think that is safe.

ios/LayoutReanimation/REAAnimationsManager.h
```diff
@@ -39,6 +46,17 @@ typedef void (^REAAnimationRemovingBlock)(NSNumber *_Nonnull tag);
...
+- (NSDictionary *)prepareDataForLayoutAnimatingWorklet:(NSMutableDictionary *)currentValues
+                                          targetValues:(NSMutableDictionary *)targetValues;
+- (UIView *)viewForTag:(NSNumber *)tag;
+- (BOOL)hasAnimationForTag:(NSNumber *)tag type:(NSString *)type;
+- (void)clearAnimationConfigForTag:(NSNumber *)tag;
+- (void)startAnimationForTag:(NSNumber *)tag
+                        type:(NSString *)type
+                  yogaValues:(NSDictionary *)yogaValues
+                       depth:(NSNumber *)depth;
+- (void)visitTree:(UIView *)view block:(REATreeVisitor)block;
```
jwajgelt, line 59: Could we rename this to something like `visitTree:withBlock` or `visitTree:withVisitor:` to better match the ObjC style?

ios/LayoutReanimation/REASharedTransitionManager.h
```diff
@@ -0,0 +1,15 @@
+#import <RNReanimated/REAAnimationsManager.h>
+#import <RNReanimated/REASnapshot.h>
+
+@interface REASharedTransitionManager : NSObject
+
+- (void)notifyAboutNewView:(UIView *)view;
+- (void)viewsDidLayout;
+- (BOOL)setupSyncSharedTransitionForViews:(NSArray<UIView *> *)views;
```
jwajgelt, line 8: Can we rename this to `startSharedTransitionForViews`, like on Android?

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,518 @@
...
+  [view addObserver:self forKeyPath:@"transitionDuration" options:NSKeyValueObservingOptionNew context:nil];
+  [view addObserver:self forKeyPath:@"activityState" options:NSKeyValueObservingOptionNew context:nil];
+
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    // it replaces method for RNSScreenView class, so it can be done only once
+    UIViewController *viewController = [view valueForKey:@"controller"];
+    [self swizzleMethod:@selector(viewDidLayoutSubviews)
+                   with:@selector(swizzled_viewDidLayoutSubviews)
+               forClass:[viewController class]];
```
jwajgelt, line 237: Passing `[viewController class]` and `[view class]` here feels risky to me - what if we pass a wrong view or someone subclasses `RNSScreenView` (for some reason) and we pass an object of that class here?
I think passing the classes by their name (i.e. `[RNSScreenView class]` or whatever else needs to be passed here) here would be a better idea, even if it makes it more dependent on Screens' implementation.

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,518 @@
...
+        continue;
+      }
+    } else {
+      // is on top
+      UIView *viewTargetParentScreen = [self getScreenForView:viewTarget];
+      UIView *stackTarget = viewTargetParentScreen.reactViewController.navigationController.topViewController.view;
+      if (stackTarget != viewTargetParentScreen) {
+        continue;
+      }
+    }
```
jwajgelt, line 158: I wonder - should this be inside the `for`-loop over transitioning views? Can we start a transition between different pairs of screens at the same time, or should they always be a transition of views on screen A to views on screen B, in which case this could be done once, outside the loop?

ios/LayoutReanimation/REASnapshot.m
```diff
@@ -24,6 +24,26 @@ - (instancetype)init:(UIView *)view
   return self;
 }
 
+- (instancetype)init:(UIView *)view withParent:(UIView *)parent
```
jwajgelt, line 27: - this method doesn't set `_view = view` like `init:` does - is this intentional?
- would it be possible to rewrite this to call `self = [self init:view]` and then applying the differences between these methods, or the other way around, having `init:` call `self = [self init:view withParent:[view superview]]`? This would reduce a bit of the shared logic between these functions.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -0,0 +1,418 @@
...
+    startSharedTransitionForViews(mAddedSharedViews, true);
+    mAddedSharedViews.clear();
+  }
+
+  protected void onViewsRemoval(int[] tagsToDelete) {
+    if (tagsToDelete != null) {
+      visitTreeForTags(tagsToDelete, new SnapshotTreeVisitor());
+
+      startSharedTransitionForViews(mRemovedSharedViews, false);
+      ConfigCleanerTreeVisitor configCleanerTreeVisitor = new ConfigCleanerTreeVisitor();
```
jwajgelt, line 61:
```suggestion
      TreeVisitor configCleanerTreeVisitor = (View view) -> mNativeMethodsHolder.clearAnimationConfig(view.getId());
```
since this visitor is used only once, I think it might be more readable to just use a lambda, especially when it just calls a function.

jwajgelt: You could also rename this variable to something like `clearConfigForView` or even just `clearConfig`, since it's not really important that it's used as a visitor, but that's more of a preference than a suggestion.

src/createAnimatedComponent.tsx
```diff
@@ -133,6 +134,8 @@ export type AnimatedComponentProps<P extends Record<string, unknown>> = P & {
     | typeof BaseAnimationBuilder
     | EntryExitAnimationFunction
     | Keyframe;
+  sharedTransitionTag?: string;
+  sharedTransitionStyle?: any;
```
jwajgelt, line 138: would 
```suggestion
  sharedTransitionStyle?:
    | ILayoutAnimationBuilder
    | LayoutAnimationFunction
    | Keyframe;
```
work?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
piaskowyk: I think that Android API 21 doesn't support inline lambda expression :/

jwajgelt: Ok, if it's not supported, it's not supported 👍

jwajgelt (approved):

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -73,4 +94,22 @@ void LayoutAnimationsManager::startLayoutAnimation(
       config->getJSValue(rt));
 }
 
+/*
+  The top screen on the stack triggers the animation, so we need to find
+  the sibling view registered in the past. That's why we look backward.
+
+  We need to find view registered in the same transition group (with the
+  same transition tag) which has been added to that group directly before
+  the one that we provide as an argument.
```
kmagiera, line 103: I'd rather structure similar comments in a form of "this method does that" rather than "we need so we do"

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -586,4 +622,28 @@ private ViewManager resolveViewManager(int tag) {
       return null;
     }
   }
+
+  private Point convertAbsoluteToParentRelative(Point fromPoint, View parentView) {
```
kmagiera, line 626: should this be static?

kmagiera, line 626: convertScreenLocationToViewCoordinates

android/src/reactNativeVersionPatch/nativeHierarchyManager/latest/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -263,7 +285,13 @@ public synchronized void manageChildren(
     }
 
     // we don't want layout animations in native-stack since it is currently buggy there
+    AnimationsManager animationsManager = ((ReaLayoutAnimator) mReaLayoutAnimator).getAnimationsManager();
```
kmagiera, line 288: we do we need casting here? If the field is called `mReaLayoutAnimator` why isn't it of type `ReaLayoutAnimator`?

android/src/reactNativeVersionPatch/nativeHierarchyManager/latest/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -263,7 +285,13 @@ public synchronized void manageChildren(
...
     if (viewGroupManager.getName().equals("RNSScreenStack")) {
+      if (tagsToDelete == null) {
+        animationsManager.doSnapshotForTopScreenViews(viewGroup);
```
kmagiera, line 291: I think want to do "snapshot *of* sth" and not "snapshot *for* sth" here? no?

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -171,14 +182,15 @@ - (double)getDoubleOrZero:(NSNumber *)number
 
 - (void)setNewProps:(NSMutableDictionary *)newProps forView:(UIView *)view
 {
-  NSMutableDictionary *dataComponenetsByName = [_uiManager valueForKey:@"_componentDataByName"];
-  RCTComponentData *componentData = dataComponenetsByName[@"RCTView"];
-  [self setNewProps:newProps forView:view withComponentData:componentData];
+  NSMutableDictionary *dataComponentsByName = [_uiManager valueForKey:@"_componentDataByName"];
```
kmagiera, line 185: Could we just do it once in `setNewProps` instead of doing it twice here and in `progressLayoutAnimationWithStyle`?

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -194,16 +206,33 @@ - (void)setNewProps:(NSMutableDictionary *)newProps
     view.center = CGPointMake(view.center.x + view.bounds.size.width / 2.0 - oldWidth / 2.0, view.center.y);
     [newProps removeObjectForKey:@"width"];
   }
+
+  bool updateViewPosition = false;
```
kmagiera, line 210: `needViewPositionUpdate` or `needUpdateViewPosition`

ios/LayoutReanimation/REAAnimationsManager.h
kmagiera: I wonder if we could use `REANodeFind` instead of this? how is this one different?

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,518 @@
...
+  [_addedSharedViews removeAllObjects];
+}
+
+- (void)setupAsyncSharedTransitionForViews:(NSArray<UIView *> *)views
+{
+  NSArray *sharedViews = [self sortViewsByTags:views];
+  _sharedElements = [self getSharedElementForCurrentTransition:sharedViews withNewElements:YES];
+}
+
+- (BOOL)setupSyncSharedTransitionForViews:(NSArray<UIView *> *)views
```
kmagiera, line 75: why is "sync" part of the name?

ios/LayoutReanimation/REASharedTransitionManager.h
piaskowyk: I think that in this case, this name is better because on Android we always start Shared Transition synchronously, but on iOS, we can start it immediately or with delay (we need to wait on layout computation) so I think that different name shows well the difference between Android and iOS implementation.

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -0,0 +1,518 @@
...
+    Then we call KVO to notify Reanimated observer about the attached header.
+  */
+
+  // call original method from react-native-screens, self == RNScreen
+  [self swizzled_viewDidLayoutSubviews];
+  // get RNScreenView from RNScreen to call KVO on RNScreenView because
+  // RNScreen doesn't contains writable public property.
+  UIView *view = [self valueForKey:@"screenView"];
+  // call KVO to run runAsyncSharedTransition from Reanimated
+  [view setValue:[view valueForKey:@"transitionDuration"] forKey:@"transitionDuration"];
```
kmagiera, line 273: If we already do method swizzling why can't we directly call that method (in this case `runAsyncSharedTransition`) from here instead of adding KVO in between?

kmagiera (approved):
Left a couple minor commets. I think this one is good to merge. It'd be great to get all my comments addressed but lets do that in a follow up PR to avoid another round of testing for this big PR.

piaskowyk:
According to comment:
> Left a couple minor commets. I think this one is good to merge. It'd be great to get all my comments addressed but lets do that in a follow up PR to avoid another round of testing for this big PR.

I merge this PR and I will create another one with improvements from latest reviews.

ios/LayoutReanimation/REASharedTransitionManager.h
piaskowyk: But maybe other name will be better.

ios/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: In this place, I check if the animation should be run. The `_findPrecedingViewTagForTransition` return just the view tag register before the current shared view, but the pair of views is doesn't always valid. For example for screen stack:
2: ScreenC - has a shared view 
1: ScreenB - hasn't a shared view
0: ScreenA - has a shared view

And if we want to go back from ScreenC -> ScreenB, `_findPrecedingViewTagForTransition` returns a view tag from ScreenA but we don't want to run animation in this case.

ios/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/e0f1b6a7325627667feeb895e7501f320bc797c5

ios/LayoutReanimation/REASnapshot.m
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/e0f1b6a7325627667feeb895e7501f320bc797c5

ios/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/0608d3a2827d766efb0acb86f73b4a12b06b4b89

ios/LayoutReanimation/REAAnimationsManager.h
piaskowyk: Under the hood, visit tree is just a `REANodeFind`
```
- (void)visitTree:(UIView *)view block:(REATreeVisitor)block
{
  REANodeFind(view, block);
}
```
But `REANodeFind` is defined as `static BOOL REANodeFind` in the top level of `REAAnimationManager.m` and it wasn't available from other files. I can rewrite `REANodeFind` as a static method in `REAAnimationManager` class if you want

src/createAnimatedComponent.tsx
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/d28696f8d202447293a16ad73004c52c1b2b48d3

ios/LayoutReanimation/REAAnimationsManager.h
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/dd7bac67f17f0b015ef599a2307bdbccf9715713

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/e65b1e008ca12fcb14d9fce05e3ff30fe24381cb

android/src/reactNativeVersionPatch/nativeHierarchyManager/latest/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
piaskowyk: Resolve in: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/12860cc478ab0c9c4eedc371a75cc93d1af79a0f

android/src/reactNativeVersionPatch/nativeHierarchyManager/latest/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
piaskowyk: Resolved by: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/09400a87bd6e0878d84cdf6859298427905d6940

ios/LayoutReanimation/REAAnimationsManager.m
piaskowyk: Resolved by: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/f5df7e97ea2426bd95280d4383f140dff9b20516

ios/LayoutReanimation/REAAnimationsManager.m
piaskowyk: Resolved by: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/dd59309b559dd81f0f5d4853fd675cc1d1dc99b2

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
piaskowyk: Resolved by: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/83c74794453818abe6da4894412860e9811e5aef

ios/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: That was because swizzled method changes the context of execution, so the `self` in swizzled method is `RNSScreen` instead of `REASharedTransitionManager`. That's why I used KVO to change the context of execution back to `REASharedTransitionManager`. But just now I found much better solution. I'll add it to #4014 soon.

piaskowyk: Resolved by: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/2110fb5c5762b19cf5555a0c34696a7c8836f7d1

ios/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: I changed this name because was misleading. Resolved by: https://github.com/software-mansion/react-native-reanimated/pull/4014/commits/06ff48a085c6d2d8fe3704857b2b283444598877

PR #3833, opened 2022-12-02
Update types in layout animation builders

j-piasecki:
## Summary

The focus of this PR is on the code responsible for building layout animations. Instead of hard-coded return types in builders, it uses the implicit `this` argument to infer the type that should be returned.


tjzel (commented):
I would like to ask you for some more explanation about why do we do this in this certain way (using `this` argument for `createInstance` methods) in some time.

PR #3846, opened 2022-12-08
Route JS errors from UI runtime throught RN's LogBox module

kmagiera:
This PR changes the way we report errors in development. Previously we'd use RCTLog native module which would result in a relatively ugly red screen displaying an error. In addition the stack trace wouldn't be symbolicated so it was difficult to reason about the root cause of the problem when the crash happened on the UI runtime.

With this change we provide a symbolicated version of trace to ErroUtil module which results in the crash being displayed in the same way as it were to occur on the regular RN runtime.

## Summary

The main motivation is to provide better guidance for developers about the crashes on the UI JS runtime as well as to use a familiar UI for displaying those.

In a nutshell, the method we use relies on an extended version of "eval" for processing the javascript code when loading on the UI runtime. We now expose `evalWithSourceUrl` that beside the code also allows for assiging the source url that is then included in the traces, and `evalWithSourceMap` (only on hermes) that allows to provide JSON encoded source map that is then used by the javascript engine to symbolicate the stack trace.

For JS engines that does not support source maps (JSC), what we do instead, is that we provide the worklet hash as a part of the source url, this allows us to recognize which worklet a given stack entry is coming from and allows us to map the provided line from that worklet into a line of the whole JS bundle. In order to do so, we now generate an "error" object in the place where worklet is generated such that we can get its position in the Javascript bundle.

Below is a summary of changes this PR makes:
1) Changes in plugin focus on removing location metadata (we now use a string in a format "worklet_7263" where the number is worklet's hash), adding source maps in a form of JSON encoded string, and adding new error object to the worklet object which is used to remap worklet line number into the bundle line number
2) For the latter, we added some additional logic that replaces entries in the provided error and replaces "worklet_23746:16:2" with the bundle URL along  with the remapped line numbers
3) We now route all JS calls via a new method "guardCall" that adds a catch statement and passes the exception back to the main JS runtime where we use ErrorUtils module to trigger the default React Native's LogBox
4) We extend hermes runtime by exposing `evalWithSourceMap` method – this method is only added for debug builds and only on hermes.
5) On other runtimes we register additional global method `evalWIthSourceURL` that makes it possible to provide URLs along the code that needs to be evaluated.

## Test plan

Run method "something" from the following snippet and expect an error that should result in a redbox being desplayed. Note that the presented stack trace should have correct line numbers. See the expected result on iOS under the snippet. On Android the errors aren't as nicely formatted but should contain valid trace entries. This needs to be tested on both JSC and Hermes configurations.

```
function makeWorklet() {
  return () => {
    'worklet';
    throw new Error('Randomly crashing');
  };
}

const crashRandomly = makeWorklet();

function anotherWorklet() {
  'worklet';
  crashRandomly();
}

function something() {
  runOnUI(() => {
    'worklet';
    anotherWorklet();
  })();
}
```
simulator_screenshot_3CECE7D6-DA04-4661-93C3-C25EF4A19423

### DIFF
--- a/Common/cpp/NativeModules/NativeReanimatedModule.h
+++ b/Common/cpp/NativeModules/NativeReanimatedModule.h
@@ -50,7 +50,7 @@ class NativeReanimatedModule : public NativeReanimatedModuleSpec,
   void installCoreFunctions(
       jsi::Runtime &rt,
       const jsi::Value &callGuard,
-      const jsi::Value &workletMaker) override;
+      const jsi::Value &valueUnpacker) override;
 
   jsi::Value makeShareableClone(jsi::Runtime &rt, const jsi::Value &value)
       override;
--- a/Common/cpp/NativeModules/NativeReanimatedModule.h
+++ b/Common/cpp/NativeModules/NativeReanimatedModule.h
@@ -49,8 +49,8 @@ class NativeReanimatedModule : public NativeReanimatedModuleSpec,
 
   void installCoreFunctions(
       jsi::Runtime &rt,
-      const jsi::Value &workletMaker,
-      const jsi::Value &layoutAnimationStartFunction) override;
+      const jsi::Value &callGuard,
+      const jsi::Value &workletMaker) override;
 
   jsi::Value makeShareableClone(jsi::Runtime &rt, const jsi::Value &value)
       override;
--- a/Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
+++ b/Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
@@ -80,7 +80,7 @@ ReanimatedHermesRuntime::ReanimatedHermesRuntime(
   jsQueue->quitSynchronous();
 #endif
 
-#ifndef NDEBUG
+#ifdef DEBUG
   jsi::Value evalWithSourceMap = jsi::Function::createFromHostFunction(
       *runtime_,
       jsi::PropNameID::forAscii(*runtime_, "evalWithSourceMap"),
@@ -105,7 +105,7 @@ ReanimatedHermesRuntime::ReanimatedHermesRuntime(
       });
   runtime_->global().setProperty(
       *runtime_, "evalWithSourceMap", evalWithSourceMap);
-#endif
+#endif // DEBUG
 }
 
 ReanimatedHermesRuntime::~ReanimatedHermesRuntime() {
--- a/Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
+++ b/Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
@@ -80,6 +80,7 @@ ReanimatedHermesRuntime::ReanimatedHermesRuntime(
   jsQueue->quitSynchronous();
 #endif
 
+#ifndef NDEBUG
   jsi::Value evalWithSourceMap = jsi::Function::createFromHostFunction(
       *runtime_,
       jsi::PropNameID::forAscii(*runtime_, "evalWithSourceMap"),
@@ -104,6 +105,7 @@ ReanimatedHermesRuntime::ReanimatedHermesRuntime(
       });
   runtime_->global().setProperty(
       *runtime_, "evalWithSourceMap", evalWithSourceMap);
+#endif
 }
 
 ReanimatedHermesRuntime::~ReanimatedHermesRuntime() {
--- a/Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
+++ b/Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
@@ -79,6 +79,25 @@ ReanimatedHermesRuntime::ReanimatedHermesRuntime(
   // that the thread was indeed `quit` before
   jsQueue->quitSynchronous();
 #endif
+
+  jsi::Value evalWithSourceMap = jsi::Function::createFromHostFunction(
+      *runtime_,
+      jsi::PropNameID::forAscii(*runtime_, "evalWithSourceMap"),
+      3,
+      [](jsi::Runtime &rt,
+         const jsi::Value &thisValue,
+         const jsi::Value *args,
+         size_t count) -> jsi::Value {
+        auto codeBuffer = std::make_shared<const jsi::StringBuffer>(
+            args[0].asString(rt).utf8(rt));
+        auto sourceMapJSONBuffer = std::make_shared<const jsi::StringBuffer>(
+            args[1].asString(rt).utf8(rt));
+        return dynamic_cast<facebook::hermes::HermesRuntime &>(rt)
+             .evaluateJavaScriptWithSourceMap(
+                codeBuffer, sourceMapJSONBuffer, args[2].asString(rt).utf8(rt));
+      });
+  runtime_->global().setProperty(
+      *runtime_, "evalWithSourceMap", evalWithSourceMap);
 }
 
 ReanimatedHermesRuntime::~ReanimatedHermesRuntime() {
--- a/Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
+++ b/Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
@@ -88,13 +88,19 @@ ReanimatedHermesRuntime::ReanimatedHermesRuntime(
          const jsi::Value &thisValue,
          const jsi::Value *args,
          size_t count) -> jsi::Value {
-        auto codeBuffer = std::make_shared<const jsi::StringBuffer>(
+        auto code = std::make_shared<const jsi::StringBuffer>(
             args[0].asString(rt).utf8(rt));
-        auto sourceMapJSONBuffer = std::make_shared<const jsi::StringBuffer>(
-            args[1].asString(rt).utf8(rt));
+        std::string sourceURL;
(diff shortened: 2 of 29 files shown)


Common/cpp/NativeModules/NativeReanimatedModule.h
```diff
@@ -49,8 +49,8 @@ class NativeReanimatedModule : public NativeReanimatedModuleSpec,
 
   void installCoreFunctions(
       jsi::Runtime &rt,
-      const jsi::Value &workletMaker,
-      const jsi::Value &layoutAnimationStartFunction) override;
+      const jsi::Value &callGuard,
+      const jsi::Value &workletMaker) override;
```
tomekzaw, line 53: Here it's `workletMaker` but in `NativeReanimatedModuleSpec.h` it's `valueUnpacker`. Can we unify this?

plugin.js
```diff
@@ -4,6 +4,7 @@ const hash = require('string-hash-64');
 const traverse = require('@babel/traverse').default;
 const { transformSync } = require('@babel/core');
 const fs = require('fs');
+const convertSourceMap = require('convert-source-map');
```
tomekzaw, line 7: Do we need to add `convert-source-map` as a direct dependency in package.json?

plugin.js
```diff
@@ -384,15 +387,22 @@ function buildWorkletString(t, fun, closureVariables, name, inputMap) {
   const transformed = transformSync(code, {
     plugins: [prependClosureVariablesIfNecessary()],
     compact: !shouldGenerateSourceMap(),
-    sourceMaps: shouldGenerateSourceMap() ? 'inline' : false,
+    sourceMaps: shouldGenerateSourceMap(),
```
tomekzaw, line 390: There may be some difference between `'inline'` and `true`.

plugin.js
```diff
@@ -523,18 +533,9 @@ function makeWorklet(t, fun, state) {
...
-
-  const loc = fun && fun.node && fun.node.loc && fun.node.loc.start;
-  if (loc) {
-    const { line, column } = loc;
-    if (typeof line === 'number' && typeof column === 'number') {
-      location = `${location} (${line}:${column})`;
-    }
+  let lineOffset = 1;
+  if (closure.size > 0) {
+    lineOffset -= closure.size + 2;
```
tomekzaw, line 538: But why?

src/reanimated2/initializers.ts
```diff
@@ -0,0 +1,150 @@
...
+  'worklet';
+  try {
+    fn(...args);
+  } catch (e) {
+    if (global.ErrorUtils) {
+      global.ErrorUtils.reportFatalError(e);
+    } else {
+      throw e;
+    }
+  }
```
tomekzaw, line 17: What about using `try { ... } catch { ... }` only in development mode (`__DEV__`)?

src/reanimated2/initializers.ts
```diff
@@ -0,0 +1,150 @@
...
+  let handleCache = global.__handleCache;
+  if (workletsCache === undefined) {
+    // init
+    workletsCache = global.__workletsCache = new Map();
+    handleCache = global.__handleCache = new WeakMap();
+  }
+  if (objectToUnpack.__workletHash) {
+    let workletFun = workletsCache.get(objectToUnpack.__workletHash);
+    if (workletFun === undefined) {
+      const evalFn =
```
tomekzaw, line 32: We could also memoize `evalFn` (i.e. initialize it analogously as `workletsCache`).

src/reanimated2/initializers.ts
```diff
@@ -0,0 +1,150 @@
...
+  hash: number,
+  stackDetails: StackDetails
+) {
+  _workletStackDetails.set(hash, stackDetails);
+}
+
+function getBundleOffset(error: Error): [string, number, number] {
+  const frame = error.stack?.split('\n')?.[0];
+  if (frame) {
+    const parsedFrame = /@(.*):(\d+):(\d+)/.exec(frame);
```
tomekzaw, line 80: We will need to check these regular expressions for vulnerabilities (some time ago we had one in `Colors.ts`, see https://github.com/advisories/GHSA-2j79-8pqc-r7x6).

src/reanimated2/initializers.ts
kmagiera: don't see a clear benefit for memoizing this, this just a single if/else check that we need to perform so memoizing seem like a premature optimization

src/reanimated2/initializers.ts
kmagiera: wouldn't really be concerned about this, there are perhaps ways of making this regex run forever but that would've need to be a user code that triggers this behavior. If the user pulls in a malicious code into their app there are definitely other, easier to exploit vectors of attack than making some regex that runs in development builds only to execute in infinite loop

Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
```diff
@@ -79,6 +80,35 @@ ReanimatedHermesRuntime::ReanimatedHermesRuntime(
...
+        std::string sourceURL;
+        if (count > 1 && args[1].isString()) {
+          sourceURL = args[1].asString(rt).utf8(rt);
+        }
+        std::shared_ptr<const jsi::Buffer> sourceMap;
+        if (count > 2 && args[2].isString()) {
+          sourceMap = std::make_shared<const jsi::StringBuffer>(
+              args[2].asString(rt).utf8(rt));
+        }
+        return wrappedRuntime->evaluateJavaScriptWithSourceMap(
```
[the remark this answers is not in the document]
kmagiera: I actually only updated it recently, was doing a cast before but it turned out that on Android it doesn't perhaps due to issues with RTTI.

plugin.js
```diff
@@ -71,11 +72,13 @@ const globals = new Set([
   '_makeShareableClone',
   '_updateDataSynchronously',
   'eval',
+  'evalWithSourceMap',
```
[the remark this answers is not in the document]
kmagiera: I think we no longer need either because of the recent global object fix we can now use global.evalBlah and it doesn't need to be included on the whitelist

src/reanimated2/initializers.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+function valueUnpacker(objectToUnpack: any, category?: string): any {
+  'worklet';
+  let workletsCache = global.__workletsCache;
+  let handleCache = global.__handleCache;
+  if (workletsCache === undefined) {
+    // init
+    workletsCache = global.__workletsCache = new Map();
+    handleCache = global.__handleCache = new WeakMap();
```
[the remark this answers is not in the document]
kmagiera: I'm confused, they are already in initializers.ts – maybe you mean to move them to `initializeUIRuntime` method? This could be done but I'd rather do it in the future as it requires a few more things to be cleaned up. The main reason is that `valueUnpacker` is actually used to facilitate `runOnUI` that is used to schedule `initializeUIRuntime`. Hence this method is used before `intiializeUIRuntime` and so we won't be able to have these objects ready.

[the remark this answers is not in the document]
tomekzaw: Yeah, `initializeUIRuntime` is what I meant. Thanks for explanation

src/reanimated2/initializers.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+}
+
+function callGuard<T extends Array<any>, U>(
+  fn: (...args: T) => U,
+  ...args: T
+): void {
+  'worklet';
+  fn(...args);
+}
```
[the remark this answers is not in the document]
kmagiera: I don't think it is significant enough to worry but can change it anyways

Common/cpp/SharedItems/Shareables.h
```diff
@@ -61,27 +84,20 @@ class JSRuntimeHelper {
...
 
- public:
-  CoreFunction(JSRuntimeHelper *runtimeHelper, const jsi::Value &workletObject);
   template <typename... Args>
-  jsi::Value call(jsi::Runtime &rt, Args &&...args) {
-    return getFunction(rt)->call(rt, args...);
+  inline void runOnUIGuarded(const jsi::Value &function, Args &&...args) {
+    // We only use callGuard in debug mode, otherwise we call the provided
+    // function directly. CallGuard provides a way of capturing expeptions in
+    // Javascript and propagating them to the main React Native thread such that
```
tomekzaw, line 92:
```suggestion
    // function directly. CallGuard provides a way of capturing exceptions in
    // JavaScript and propagating them to the main React Native thread such that
```

Common/cpp/SharedItems/Shareables.h
```diff
@@ -61,27 +84,20 @@ class JSRuntimeHelper {
...
+    // they can be presented using RN's LogBox.
+#ifdef DEBUG
+    callGuard->call(*uiRuntime_, function, args...);
+#else
+    function.asObject(*uiRuntime_)
+        .asFunction(*uiRuntime_)
+        .call(*uiRuntime_, args...);
+#endif
```
tomekzaw, line 100: Most likely the compiler will optimize it right away but since we use `*uiRuntime_` here three times I think it's more readable if we first assign it to `rt`:
```suggestion
    jsi::Runtime &rt = *uiRuntime_;
#ifdef DEBUG
    callGuard->call(rt, function, args...);
#else
    function.asObject(rt).asFunction(rt).call(rt, args...);
#endif
```

tomekzaw (approved):
👏

PR #3852, opened 2022-12-09
Avoid multiple rAF calls over JSI when called on the same frame

kmagiera:
## Summary

This PR overrides requestAnimationFrame on the UI runtime such that it keeps a list of callbacks for a given frame locally and only calls into the native requestAnimationFrame once in every frame where the method gets triggered.

This results in a greatly reduced number of calls that need to happen over JSI in case where there are multiple components/values being animated at the same time. This also, apparently, fixes a crash in BokehExample on android that was due to some race condition issues in the naitve requestAnimationFrame implementation. Calling it only once in a given frame resolves that issue.

## Test plan

Run BokehExample on Android with the number of circles bumped up to 300. The app should not crash and perform well on an emulator (note that 300 will be too high for low-end devices but still the app shouldn't crash).

### DIFF
--- a/src/reanimated2/initializers.ts
+++ b/src/reanimated2/initializers.ts
@@ -1,5 +1,6 @@
 import { reportFatalErrorOnJS } from './errors';
 import NativeReanimatedModule from './NativeReanimated';
+import { isJest } from './PlatformChecker';
 import { runOnUI, runOnJS } from './threads';
 
 // callGuard is only used with debug builds
@@ -87,6 +88,8 @@ Possible solutions are:
 export function initializeUIRuntime() {
   NativeReanimatedModule.installCoreFunctions(callGuardDEV, valueUnpacker);
 
+  const IS_JEST = isJest();
+
   const capturableConsole = console;
   runOnUI(() => {
     'worklet';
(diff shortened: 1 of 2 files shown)


src/reanimated2/core.ts
```diff
@@ -213,6 +213,22 @@ if (!isWeb() && isConfigured()) {
       info: runOnJS(capturableConsole.info),
     };
     _setGlobalConsole(console);
+
+    const nativeRequestAnimationFrame = global.requestAnimationFrame;
+    let callbacks: Array<(time: number) => void> = [];
+    global.requestAnimationFrame = (
+      callback: (time: number) => void
```
tomekzaw, line 220: For the sake of consistency
```suggestion
      callback: (timestamp: number) => void
```

src/reanimated2/core.ts
```diff
@@ -213,6 +213,22 @@ if (!isWeb() && isConfigured()) {
...
+    ): number => {
+      callbacks.push(callback);
+      if (callbacks.length === 1) {
```
tomekzaw, line 223: I know this is obvious but maybe let's add a comment that explains the magic (we schedule a function that captures `callbacks` array which that is filled in subsequent calls (batching) but since the function needs to be executed only once we schedule it only when length equals 1 meaning that before pushing it was empty).

src/reanimated2/core.ts
```diff
@@ -213,6 +213,22 @@ if (!isWeb() && isConfigured()) {
...
+        nativeRequestAnimationFrame((timestamp) => {
+          const currentCallbacks = callbacks;
+          callbacks = [];
+          currentCallbacks.forEach((f) => f(timestamp));
+        });
+      }
+      return -1;
```
tomekzaw, line 230: Can we add a comment here that explains that technically `requestAnimationFrame` should return `requestID` (number) that can be passed to `cancelAnimationFrame` to cancel the callback but apparently we neither use the output of `requestAnimationFrame` throughout our codebase nor implement `cancelAnimationFrame`?

src/reanimated2/core.ts
```diff
@@ -213,6 +213,22 @@ if (!isWeb() && isConfigured()) {
       info: runOnJS(capturableConsole.info),
     };
     _setGlobalConsole(console);
+
+    const nativeRequestAnimationFrame = global.requestAnimationFrame;
+    let callbacks: Array<(time: number) => void> = [];
+    global.requestAnimationFrame = (
```
tomekzaw, line 219: I don't like the idea that here we override the JSI binding with a regular JS function. I would like to suggest two improvements:
1. Rename the JSI binding to `nativeRequestAnimationFrame` to resolve the name conflict and discourage future developers from using it. We already provide the JS function `requestAnimationFrame` which satisfies the semantics (i.e. can be called multiple number of times) so there's no reason to expose an identical API (except for batching) from the native side. We can treat `nativeRequestAnimationFrame` as Reanimated-specific implementation detail.
2. (Optionally) Add a check on native side (in debug mode only) that `nativeRequestAnimationFrame` is called no more than once per single frame to prevent using it directly in the future.

src/reanimated2/initializers.ts
```diff
@@ -109,5 +109,28 @@ export function initializeUIRuntime() {
...
+        // We schedule native requestAnimationFrame only when the first callback
+        // is added and then use it to execute all the enqueued callbacks. Once
+        // the callbacks are run, we clear the array.
+        nativeRequestAnimationFrame((timestamp) => {
+          const currentCallbacks = callbacks;
+          callbacks = [];
+          currentCallbacks.forEach((f) => f(timestamp));
+        });
+      }
+      // Reaminated currently does not support cancelling calbacks requested with
```
tomekzaw, line 129:
```suggestion
      // Reanimated currently does not support cancelling calbacks requested with
```

piaskowyk (approved):

src/reanimated2/initializers.ts
```diff
@@ -109,5 +112,32 @@ export function initializeUIRuntime() {
...
+          // We schedule native requestAnimationFrame only when the first callback
+          // is added and then use it to execute all the enqueued callbacks. Once
+          // the callbacks are run, we clear the array.
+          nativeRequestAnimationFrame((timestamp) => {
+            const currentCallbacks = callbacks;
+            callbacks = [];
+            currentCallbacks.forEach((f) => f(timestamp));
+          });
+        }
+        // Reaminated currently does not support cancelling calbacks requested with
```
tomekzaw, line 135:
```suggestion
        // Reanimated currently does not support cancelling calbacks requested with
```

PR #3864, opened 2022-12-14
Fix unnecessary calls startMapper/stopMapper in useAnimatedReaction

SergeyRozhkov:
## Summary
Hey!
If we do not specify an array of dependencies in useAnimatedReaction and prepare or react has external dependent values (._closure), then the startMapper and stopMapper methods will be called an extra time on each render, even if the external values have not changed.

## Test plan
If the functional component has such a block of code, then useEffect is called inside useAnimatedReaction on every render before these changes.
```
  const value = 1;
  useAnimatedReaction(
    () => value,
    () => {},
  );
```


kmagiera (approved):
This seems ok, would only like to test this locally prior to merging

PR #3865, opened 2022-12-14
[LayoutAnimations] Fix exiting animations inside a Native Stack Navigator

jwajgelt:
## Summary

Currently, removing a screen with `exiting` layout animations running cause views to not be properly removed from the native hierarchy, as well as leaking memory in Layout Animations' `AnimationManager`.

This is caused by the ancestors of exiting views being removed from the native hierarchy by the React-Native-Screens implementation, while our implementation assumes they're still mounted in the native hierarchy.

## Test plan

In the Example app, go to the "Nested NativeStacks with layout" screen and try toggling and navigating around while the `exiting` animation on the second screen of the nested stack is running.


android/src/reactNativeVersionPatch/nativeHierarchyManager/latest/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -276,6 +286,12 @@ public synchronized void manageChildren(
 
     // we don't want layout animations in native-stack since it is currently buggy there
     if (viewGroupManager.getName().equals("RNSScreenStack")) {
+      if (indicesToRemove != null) {
+        for (int index : indicesToRemove) {
+          View child = viewGroupManager.getChildAt(viewGroup, index);
+          ((ReaLayoutAnimator) mReaLayoutAnimator).cancelAnimationsInSubviews(child);
+        }
+      }
```
jwajgelt, line 294: The fix for Android is to cancel all the animations in the transaction that removes the Screen.
This will cause all the exiting views to unregister their ancestors before being unmounted.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -376,6 +380,13 @@ public boolean isLayoutAnimationEnabled() {
 
   private boolean removeOrAnimateExitRecursive(View view, ViewGroup parent, boolean shouldRemove) {
     int tag = view.getId();
+    ViewManager viewManager = resolveViewManager(tag);
+
+    if (viewManager != null && viewManager.getName().equals("RNSScreenStack")) {
+      cancelAnimationsRecursive(view);
+      return false;
+    }
```
jwajgelt, line 388: It's possible it's not just the screen that's being removed, but the whole stack.
In that case, we have to cancel the animations in the screen as well, which we do here.

kmagiera:
I think this approach is ok. I'm only wondering if the changes that check for RNScreen are necessary for this to work. If I understand this correctly, it seem like these are separate things, and the changes that reference screens are there to prevent exiting views from animating when the whole screen is unmounted. If that's the case I think it'd be better to split these changes into separate PRs. Also blocking exiting animations in such a case may be in conflict with the work we do on SET (cc @piaskowyk )

jwajgelt:
> I think this approach is ok. I'm only wondering if the changes that check for RNScreen are necessary for this to work. If I understand this correctly, it seem like these are separate things, and the changes that reference screens are there to prevent exiting views from animating when the whole screen is unmounted. If that's the case I think it'd be better to split these changes into separate PRs. Also blocking exiting animations in such a case may be in conflict with the work we do on SET (cc @piaskowyk )

On iOS, cancelling animations in subviews of a `RNSScreenStackView` fixes an issue when unmounting an entire stack with exiting animations inside the currently mounted screen keeping the entire stack alive for the duration of the animation (since in that case, Screens don't unmount the views themselves.)

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -376,6 +379,13 @@ public boolean isLayoutAnimationEnabled() {
 
   private boolean removeOrAnimateExitRecursive(View view, ViewGroup parent, boolean shouldRemove) {
     int tag = view.getId();
+    ViewManager viewManager = resolveViewManager(tag);
+
+    if (viewManager != null && viewManager.getName().equals("RNSScreenStack")) {
+      cancelAnimationsRecursive(view);
```
piaskowyk, line 385: Why we need to cancel animation instead of prevent animation start?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -427,6 +437,23 @@ private boolean removeOrAnimateExitRecursive(View view, ViewGroup parent, boolea
...
+    if (view != null) {
+      clearAnimationConfigRecursive(view);
+    }
+  }
+
+  public void clearAnimationConfigRecursive(View view) {
+    mNativeMethodsHolder.clearAnimationConfig(view.getId());
+    if (view instanceof ViewGroup) {
+      ViewGroup viewGroup = (ViewGroup) view;
+      for (int i = 0; i < viewGroup.getChildCount(); i++) {
```
piaskowyk, line 451: What if some screens are removed from native hierarchy? Maybe we should use React's tree instead of. (I just wondering)

ios/LayoutReanimation/REAUIManager.mm
```diff
@@ -76,9 +76,28 @@ - (void)_manageChildren:(NSNumber *)containerTag
...
   NSArray<id<RCTComponent>> *permanentlyRemovedChildren;
+  BOOL containerIsRootOfViewController = NO;
   if (isLayoutAnimationEnabled) {
     container = registry[containerTag];
     permanentlyRemovedChildren = [self _childrenToRemoveFromContainer:container atIndices:removeAtIndices];
+
+    if ([container isKindOfClass:[UIView class]]) {
+      UIViewController *controller = ((UIView *)container).reactViewController;
+      UIViewController *parentController = ((UIView *)container).superview.reactViewController;
+      containerIsRootOfViewController = controller != parentController;
```
piaskowyk, line 87: Did you try what happens in a nested stack?

piaskowyk:
Could you also resolve merge conflicts?

piaskowyk:
Could you also check how it works on native stack and js stack?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
jwajgelt: There may already be some animated components in the screen.
If you don't cancel them now, when later their animations end, while traversing the tree up, they won't be able to cleanup their ancestors beyond the screen boundary, preventing those screens from being unmounted later (they'll have a non-zero `mExitingSubviewCountMap`)

android/src/reactNativeVersionPatch/nativeHierarchyManager/latest/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -140,11 +141,13 @@ public void deleteView(final View view, final LayoutAnimationListener listener)
       } catch (IllegalViewOperationException e) {
         // (IllegalViewOperationException) == (vm == null)
         e.printStackTrace();
+cancelAnimationsInSubviews(view);
```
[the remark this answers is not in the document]
jwajgelt: For some reason, spotless doesn't run on the `ReanimatedNativeHierarchyManager.java` files 🤔

ios/LayoutReanimation/REAUIManager.mm
jwajgelt: Yes, there's an example with nested stack navigators in the Example app.

Example/package.json
```diff
@@ -27,22 +27,24 @@
...
     "expo-asset": "^8.2.0",
+    "patch-package": "^6.4.7",
     "react": "18.2.0",
     "react-dom": "^16.13.1",
     "react-native": "0.71.0-rc.5",
     "react-native-gesture-handler": "^2.8.0",
     "react-native-pager-view": "^5.4.1",
     "react-native-safe-area-context": "^4.4.1",
-    "react-native-screens": "^3.18.2",
+    "react-native-screens": "software-mansion/react-native-screens#@piaskowyk/shared-transition-element-v2",
```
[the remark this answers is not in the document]
jwajgelt: I think I just copied the deps from the SET branch, since I wanted to test on a newer `navigation` version.
This would need to be reverted to a normal `screens` version, along with the `navigation` packages' versions as well, before merging.

jwajgelt:
@piaskowyk 
> Could you also check how it works on native stack and js stack?

It does, though this shouldn't really change the behaviour with js stack.

piaskowyk (changes requested):

piaskowyk:
This is the last iteration and then we could merge it

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -286,12 +296,39 @@ - (void)registerExitingAncestors:(UIView *)child exitingSubviewsCount:(int)exiti
 - (void)maybeDropAncestors:(UIView *)child
 {
   UIView *parent = child.superview;
-  while (parent != nil && ![parent isKindOfClass:[RCTRootView class]]) {
+  NSNumber *parentTag = _exitingParentTags[child.reactTag];
+  [_exitingParentTags removeObjectForKey:child.reactTag];
+
+  UIViewController *childController = child.reactViewController;
```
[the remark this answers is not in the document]
jwajgelt: In the next line, in the loop condition, we compare it to `parent.reactViewController`.
The idea is to only traverse the tree up while we're still contained in the same `ViewController`, since they're the boundary between separate screens.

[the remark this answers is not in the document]
jwajgelt: Oh no, it's not. I looked at the wrong place.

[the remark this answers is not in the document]
jwajgelt: Yeah, we can get rid of it here.

piaskowyk (approved):

PR #3874, opened 2022-12-16
fixed jsi interface injection to allow brown field app use the library

Elitebigboss90:
## Summary

It's not necessary to extend UIResponder since in the implantation file, there is no usage of the methods from UIResponder. Instead, people will have issue with third party libraries which are handling the bridge, e.g. CodePush. RCTBridge is only require you to have NSObject and this should works.

## Test plan

Have an app and let codepush to handle the bridge initialization.

### DIFF
--- a/ios/native/UIResponder+Reanimated.h
+++ b/ios/native/UIResponder+Reanimated.h
@@ -4,7 +4,7 @@
 #ifndef RCT_NEW_ARCH_ENABLED
 #ifndef DONT_AUTOINSTALL_REANIMATED
 
-@interface UIResponder (Reanimated) <RCTCxxBridgeDelegate>
+@interface NSObject (Reanimated) <RCTCxxBridgeDelegate>
 
 @end
 
--- a/ios/native/UIResponder+Reanimated.mm
+++ b/ios/native/UIResponder+Reanimated.mm
@@ -16,7 +16,7 @@ typedef JSCExecutorFactory ExecutorFactory;
 #ifndef RCT_NEW_ARCH_ENABLED
 #ifndef DONT_AUTOINSTALL_REANIMATED
 
-@implementation UIResponder (Reanimated)
+@implementation NSObject (Reanimated)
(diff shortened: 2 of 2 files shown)


tomekzaw:
Hey, thanks for the PR but unfortunately I don't think we will merge this. Extending `UIResponder` allows us to automatically install Reanimated on iOS early enough so that the library just works. We will definitely need to revisit the topic of initialization in the near future, though.

If you want to disable Reanimated autoinstallation in a brownfield app, you can set `DONT_AUTOINSTALL_REANIMATED` flag (see [here](https://github.com/software-mansion/react-native-reanimated/blob/e5d2361bd914a4ed2d1b30a21a58e9df904ae6bb/ios/native/UIResponder%2BReanimated.mm#L17)).

Elitebigboss90:
> Hey, thanks for the PR but unfortunately I don't think we will merge this. Extending `UIResponder` allows us to automatically install Reanimated on iOS early enough so that the library just works. We will definitely need to revisit the topic of initialization in the near future, though.
> 
> If you want to disable Reanimated autoinstallation in a brownfield app, you can set `DONT_AUTOINSTALL_REANIMATED` flag (see [here](https://github.com/software-mansion/react-native-reanimated/blob/e5d2361bd914a4ed2d1b30a21a58e9df904ae6bb/ios/native/UIResponder%2BReanimated.mm#L17)).

Change UIResponder into NSObject will install the library even faster --- because UIResponder is the subclass of NSObject and since you guys are not using any method from UIResponder, or solely making things working for UIResponder is optimal. Any reason you keep this?

piaskowyk:
Hey 👋 Once again, thanks for your PR and commitment! 🙌 Since React Native 0.72 we will change the initialization process. We no longer use `UIResponder` - [link to PR with changes here](https://github.com/software-mansion/react-native-reanimated/pull/4523/files#diff-4e40021f8f74e0a1a3a73a8288da2a464191420879222663c6f46259fd48eddc). If you want to avoid your issue you should update your React Native version (the release of 0.72 should be available soon). But we don't want to change the legacy implementation because we are afraid of negative effects of this in some custom configurations.

PR #3893, opened 2022-12-30
Add user-friendly error message when worklet cannot be created due to missing Babel plugin

tomekzaw:
## Summary

This PR adds an user-friendly error message when worklet cannot be created due to missing Reanimated Babel plugin.

Before:
before

After:
after

## Test plan

Check on a fresh RN app without editing `babel.config.js`.

### DIFF
--- a/src/reanimated2/threads.ts
+++ b/src/reanimated2/threads.ts
@@ -27,7 +27,7 @@ if (__DEV__) {
   try {
     runOnUI(() => {
       'worklet';
-    })();
+    });
   } catch (e) {
     throw new Error(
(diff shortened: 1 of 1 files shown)


src/reanimated2/threads.ts
```diff
@@ -23,6 +23,18 @@ export function runOnUI<A extends any[], R>(
   };
 }
 
+if (__DEV__) {
+  try {
+    runOnUI(() => {
+      'worklet';
+    })();
```
tomekzaw, line 30: Most likely `()` here is not necessary.
```suggestion
    });
```

tomekzaw: Indeed, it wasn't. Removed in d059e4a4704fddc4287abe2453bd3cec3cd6f927

piaskowyk (approved):

PR #3899, opened 2023-01-02
Fix useAnimatedScrollHandler when using FlashList

### LINKED ISSUE #3889: 3.0.0-rc.9 , useAnimatedScrollHandler with FlashList not working,
### Description

No event emitted, 
when using Flashlist or MasonryFlashlist with Reanimated 

### Steps to reproduce

```tsx
const AnimatedFlashList = Animated.createAnimatedComponent(MasonryFlashList);

 const scrollHandler = useAnimatedScrollHandler(
    {
      onBeginDrag: () => {
        cancelAnimation(scrollY);
      },
      onScroll: event => {
        'worklet';
        scrollY.value = clamp(
          event.contentOffset.y,
          -SIZES.screenHeight,
          HEADER_PROFILE_HEIGHT - heightMenu - 2 * theme.spacing.s,
        );
      },
      onMomentumEnd: () => {
        cancelAnimation(scrollY);
      },
    },
    [],
  );
  
  <AnimatedFlashList
   onScroll={scrollHandler}
    renderItem={renderItem}
  />
```

https://gist.github.com/hamdij0maa/c9ed404672e2c70871495fd579350f09

### Reanimated version

 3.0.0-rc.9

### React Native version

0.70.6

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 14 Pro

### Acknowledgements

Yes

graszka22:
## Summary
Current `FlashList` implementation wraps ScrollView in `StickyHeaderContainer` (https://github.com/Shopify/flash-list/blob/main/src/FlashList.tsx#L317). As a result `findNodeHandle(this)` in `createAnimatedComponent` returns tag of the view in `StickyHeaderContainer` instead of ScrollView tag. Because of that event listeners aren't registered correctly.

This is an attempt to fix it for the `FlashList` - if the component is a `FlashList` we take the scrollview reference from `recyclerlistview_unsafe` (https://github.com/Shopify/flash-list/blob/218f314e63806b4fe926741ef73f8b9cd6ebc7eb/src/FlashList.tsx#L824) (it's unsafe because it may change in the future but at least it's public).

This is not a foolproof solution - FlashList implementation may change in the future and there may be other Lists implementations like this and this fix won't cover them.
 
Fixes #3889 

## Test plan
Events should be logged to console.

```js
import Animated, { useAnimatedScrollHandler } from 'react-native-reanimated';

import React from 'react';
import { FlashList, MasonryFlashList } from '@shopify/flash-list';

const AnimatedFlashList = Animated.createAnimatedComponent(FlashList);

const data = new Array(45);

export defaut function FlashListDemo() {
  const scrollHandler = useAnimatedScrollHandler(
    {
      onBeginDrag: () => {
        'worklet';
        console.log('ON BEGIN DRAG');
      },
      onScroll: () => {
        'worklet';
        console.log('ON SCROLL');
      },
      onMomentumEnd: () => {
        'worklet';
        console.log('ON END');
      },
    },
    []
  );

  const renderItem = () => {
    return <View style={styles.item} />;
  };

  return (
      <AnimatedFlashList
        onScroll={scrollHandler}
        renderItem={renderItem}
        data={data}
        estimatedItemSize={120}
      />
    </View>
  );
}

  container: {
    flex: 1,
  },
  item: {
    width: 200,
    height: 100,
    backgroundColor: 'red',
    margin: 10,
  },
});
```

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -190,7 +190,15 @@ export default function createAnimatedComponent(
     }
 
     _attachNativeEvents() {
-      const viewTag = findNodeHandle(this);
+      let viewTag = findNodeHandle(this);
+
+      const componentName = Component.displayName || Component.name;
+
+      if (componentName?.endsWith('FlashList') && this._component) {
+        // @ts-ignore it's FlashList specific: https://github.com/Shopify/flash-list/blob/218f314e63806b4fe926741ef73f8b9cd6ebc7eb/src/FlashList.tsx#L824
+        viewTag = findNodeHandle(this._component.recyclerlistview_unsafe);
+      }
+
       for (const key in this.props) {
         const prop = this.props[key];
         if (


tomekzaw:
I can confirm this PR resolves the issue both on Android and iOS (tested with Example app). Well done! 👏

tomekzaw (approved):

piaskowyk (approved):

Issue #3896, opened 2023-01-01
BUG(useAnimatedSensor): java.lang.IllegalStateException: register failed, the sensor listeners size has exceeded the maximum limit 128 


ashuvssut:
### Description

I am using reanimated 2.8.0. I tried updating to latest version but this bug is present in the latest version too
**Consider this scenario. This will help reproduce the error**
Suppose we have a `Child` component and a `Parent` Component

We register a sensor using the `useAnimatedSensor` hook in the `Child` component.
Suppose `Parent` component's state changes repeatedly. Thus if `Parent` re-renders 127 times, the `Child` also renders `127` times.
On the 128th render of `Parent`, the app crashes with an error `java.lang.IllegalStateException: register failed, the sensor listeners size has exceeded the maximum limit 128` which occured because of the Child component which use `useAnimatedSensor` hook

**What I think might be cause of the error is:-**
`Parent` compoenent re-renders causes the `Child` to re-render and register sensor listeners for the same sensor on each re-render.

**What I have tried to test my reasoning?**
I wrote a simple `useEffect` hook in the `Child` component:
```js
// Child.tsx
const registeredSensor = useAnimatedSensor(/* .... */) 
const count = useRef(0)
useEffect(() => {
  console.log("render count: ", ++count.current);
  return () => {
    console.log("unregister sensor ", count.current);
    registeredSensor.unregister() // I was just trying to unregister sensor listener on unMount using this cleanup function
  }
}) // (no deps array)
```
So I got the following Logs
```
render count: 1
unregister sensor 1
render count: 2
unregister sensor 2
...
...
render count: 126
unregister sensor 126
render count: 127
unregister sensor 127
Error: java.lang.IllegalStateException: register failed, the sensor listeners size has exceeded the maximum limit 128
...
...
...
...
```
**What workaround I did to fix this?**
I used `useAnimatedSensor` in my App.tsx component and passed the `registeredsensor` to `Child` component using `useContext` hook. `App` component contains the `Parent` component. `App` component renders only 1 time.

This fix worked for me. There were no issues

> NB: I have already tested this useAnimatedSensor in the latest reanimated v2.13.0
> In v2.13.0 I did not receive the error in Metro becuase the app just crashed immediatedly without logging the error. So I used flipper to see the crash log. I found that it was the same error as mentioned above. So, this issue exists now too.

### Steps to reproduce

See description details

### Snack or a link to a repository

[...]

PR #3900, opened 2023-01-02
Fix sensor listener leak on Android

graszka22:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/3896

In the scenario presented in the issue every time the component was mounted two `ReanimatedSensor` were created: one was just initialized and the second one added to the hashmap. Then, when unregistering, only the second one was unregistered, the first one was kept registered. After some time the max limit of registered listeners is hit. 

## Test plan

```js
import { Text, View, StyleSheet } from 'react-native';
import Constants from 'expo-constants';
import { useAnimatedSensor, SensorType } from 'react-native-reanimated';

export function BasicLayoutAnimation() {
  return (
      <Parent />
    </View>
  );
}

const Parent = () => {

  React.useEffect(() => {
    setInterval(() => setState((prevState) => ++prevState), 100);
  }, []);
  return (
    <View style={[styles.container, styles.parent]}>
      <Text>Parent</Text>
      <Child />
    </View>
  );
};

const Child = () => {
  useAnimatedSensor(SensorType.MAGNETIC_FIELD, {
    interval: 2000,
  });
  const count = React.useRef(0);
  React.useEffect(() => {
    console.log('render count: ', ++count.current);
  });
  return (
      <Text>Child render: {count.current}</Text>
    </View>
  );
};

  container: {
    flex: 1,
    justifyContent: 'center',
    paddingTop: Constants.statusBarHeight,
    backgroundColor: '#ecf0f1',
    padding: 20,
  },
  parent: {
    backgroundColor: 'red',
  },
});

```

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensorContainer.java
+++ b/android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensorContainer.java
@@ -23,7 +23,7 @@ public class ReanimatedSensorContainer {
     if (sensor.initialize()) {
       sensorId = nextSensorId;
       nextSensorId++;
-      sensors.put(sensorId, new ReanimatedSensor(reactContext, sensorType, interval, setter));
+      sensors.put(sensorId, sensor);
     }
     return sensorId;
   }


tomekzaw (approved):

ashuvssut:
will this fix be available in v2.13.1? or which package ver should I install to get this fix?

tomekzaw:
@ashuvssut We will ship this fix in 2.14.0 and 3.0.0-rc.10 next week.

piaskowyk (approved):

Issue #3689, opened 2022-10-18
Predefined animations should support device orientation changes


YaoHuiJi:
### Description

Let's take SlideInLeft as an example. When the device changes from portrait to landscape, the animation will fail because the "width" variable used by the animation is not changed to the landscape width.

I don't know a lot about TypeScript, only can read Javascript code, I think the reason for this bug is  obvious, all the Predefined animations use "[const { width, height } = Dimensions.get('window');](https://github.com/software-mansion/react-native-reanimated/blob/f806a030a87997af11c6eb5e4e3b57f68e301942/src/reanimated2/layoutReanimation/defaultAnimations/Slide.ts#L11)" to  get the screen width/height and save them as  file-scope variables, so they can't update dynamically as the orientation of the screen changes.

I think it should be replaced by "useWindowDimensions" and the bug will be fixed.

### Steps to reproduce

1. Add a Predefined animation(like SlideInLeft) to a View
2. Change your device from portrait to landscape
3. Trigger the animation again, you will find the animation start from a wrong position

### Snack or a link to a repository

no need, it is very obvious

### Reanimated version

2.10.0

### React Native version

0.70.2

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Acknowledgements

Yes

YaoHuiJi:
> 



> Hey! 👋
> 
> The issue doesn't seem to contain a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example).
> 
> Could you provide a [snack](https://snack.expo.dev/) or a link to a GitHub repository under your username that reproduces the problem?

I know it should be, but this bug is to obvious...so I don't think it's worth building a new repository.

PR #3901, opened 2023-01-02
Support device orientation change in default layout animations

graszka22:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/3689

## Test plan

Run `DefaultAnimations` example, change device orientation, run for example `SlideInLeft` animation.

### DIFF
--- a/src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
+++ b/src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
@@ -21,6 +21,8 @@ export interface EntryAnimationsValues {
   targetHeight: number;
   targetGlobalOriginX: number;
   targetGlobalOriginY: number;
+  windowWidth: number;
+  windowHeight: number;
 }
 
 export interface ExitAnimationsValues {
@@ -30,6 +32,8 @@ export interface ExitAnimationsValues {
   currentHeight: number;
   currentGlobalOriginX: number;
   currentGlobalOriginY: number;
+  windowWidth: number;
+  windowHeight: number;
 }
 
 export type EntryExitAnimationFunction = (
--- a/src/reanimated2/layoutReanimation/defaultAnimations/Bounce.ts
+++ b/src/reanimated2/layoutReanimation/defaultAnimations/Bounce.ts
@@ -3,10 +3,7 @@ import {
   IEntryExitAnimationBuilder,
 } from '../animationBuilder/commonTypes';
 import { withSequence, withTiming } from '../../animation';
-import { Dimensions } from 'react-native';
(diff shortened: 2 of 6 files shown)


tomekzaw:
I can confirm this PR fixes the issue for entering animations.

However, it doesn't fully fix the exiting animations. Consider the following scenario:
1. Launch the app in portrait mode
2. Open default layout animations demo
3. SlideOutRight gets rendered with predefined exiting animation
4. Change device orientation to landscape mode
5. Press SlideOutRight to remove this element start exiting animation

When the exiting animation is run immediately after device orientation change, the last frame where the block is still visible is this one (more than 50% of the view is visible):

bad

However, when the animation is run for the second and the third time, the last frame where the block is still visible shows less than 15% of the block:

good

Here's the full video, you can download it and watch it frame-by-frame in QuickTime:

https://user-images.githubusercontent.com/20516055/210341264-ab01f5ed-805c-442c-95ed-fd928c1c791d.mp4

I'm not sure if there's something wrong with the default layout animations demo or rather with the layout animation itself. It seems like the whole demo should be re-rendered after device orientation change but for some reason the exiting layout animation is not updated (`useWindowDimensions` should be called but probably isn't).

src/createAnimatedComponent.tsx
```diff
@@ -399,6 +408,34 @@ export default function createAnimatedComponent(
...
     }
 
+    _unsubscribeWindowDimensions() {
+      this._dimensionsSubscription?.remove();
+    }
+
+    _subscribeWindowDimensions() {
+      this._dimensionsSubscription = Dimensions.addEventListener('change', () =>
+        this._configureLayoutAnimations(this._viewTag)
+      );
```
tomekzaw, line 418: I don't think that each animated component should subscribe to device orientation changes as this may impact performance significantly. Do you think it would be possible to call this method only if the animated view has an exiting animation? I understand the limitation here is that the layout animation needs to be built (configured) on the JS runtime and we cannot modify it after the component is rendered. Maybe it would be possible to somehow inject the current width and height of the screen once the exiting animation starts?

graszka22:
It turned out window size was already injected:
on android: https://github.com/software-mansion/react-native-reanimated/blob/main/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java#L227
on ios: https://github.com/software-mansion/react-native-reanimated/blob/main/ios/LayoutReanimation/REAAnimationsManager.m#L197
I just needed to add ts types

tomekzaw (approved):
Looks good to me! 🥳

piaskowyk (approved):

PR #3902, opened 2023-01-03
Make animation classes generic

j-piasecki:
## Summary

The main focus of this PR is to make animation classes generic so that they are parametrized by the animated type instead of always working with the union of all possible types.

It also adds an overload for `useDerivedValue` which correctly handles the returned type when starting animation in the callback, i.e.:
```jsx
const indicatorPosition = useDerivedValue(() => {
  return withTiming(activeIndex.value * tabWidth + tabWidth / 2, {
    duration: 500,
  });
});
```

`indicatorPosition` now would have the type of `Readonly<SharedValue<number>>` instead of `Readonly<SharedValue<Animation<TimingAnimation<number>>>>`

TODO:
- `withDecay`, which was not generic in the old `.d.ts` file and it always returned a number
- `withSequence`


tjzel:
Too outdated sadly

j-piasecki:
image

piaskowyk:
😭😭😭😭😭😭😭

PR #3905, opened 2023-01-03
Add missing argument `oneWayReadsOnly` in react-native-reanimated.d.ts

tomekzaw:
## Summary

This PR copies the signature of `useSharedValue` from useSharedValue.ts to react-native-reanimated.d.ts so it includes the missing argument `oneWayReadsOnly`.

https://github.com/software-mansion/react-native-reanimated/blob/d3aea040f6ebacb9d29bbb1766cc9effdaeac0a7/src/reanimated2/hook/useSharedValue.ts#L6-L9

## Test plan

Check if CI is green.

### DIFF
--- a/react-native-reanimated.d.ts
+++ b/react-native-reanimated.d.ts
@@ -763,7 +763,10 @@ declare module 'react-native-reanimated' {
   type DependencyList = ReadonlyArray<any>;
 
   // reanimated2 hooks
-  export function useSharedValue<T>(initialValue: T): SharedValue<T>;
+  export function useSharedValue<T>(
+    initialValue: T,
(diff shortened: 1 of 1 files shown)


ranaavneet:
Hi @tomekzaw , is it safe to pass a second parameter `oneWayReadsOnly` as `true` in `useSharedValue` now ?. I dont see any documentation on what exactly it does and in which scenarios should we use this second parameter

tomekzaw:
@ranaavneet Currently it works but it's not a part of public API so we give no guarantees that it will work in next releases. Also this is very likely to be changed when #4300 lands.

ranaavneet:
Okay cool

PR #3906, opened 2023-01-04
Don't cast `Text` to any when creating `AnimatedText`

j-piasecki:
## Summary

The `AnimatedText` component is created by calling `createAnimatedComponent` on a React Native's `Text` but for some reason, it was being cast to `any` first, resulting in all the props on the animated component being erased.


tjzel:
This is directly linked to #3805.

Latropos:
@j-piasecki @tjzel Why don't we merge?

PR #3908, opened 2023-01-04
Update the type of `EntryExitAnimationFunction` and LayoutAnimation

j-piasecki:
This PR updates the type of `EntryExitAnimationFunction` to accept either `(EntryAnimationsValues) => LayoutAnimation` or `(ExitAnimationsValues) => LayoutAnimation` instead of `(EntryAnimationsValues | ExitAnimationsValues) => LayoutAnimation`. In the second case, a function accepting one type couldn't be assigned as the type requires it to accept a union. This is the same type as in the old `.d.ts.` file.

It also updates `LayoutAnimation.animations` type to `AnimatedStyle` instead of `StyleProps` so that it correctly accepts `with*` functions as props.


piaskowyk:
A lot of things changed till this time... 🕯️

Issue #3712, opened 2022-10-24
Data race on unsubscribeFromKeyboardEvents


chrisnojima:
### Description

Hi,

I'm looking at some data races using tsan in xcode and believe there is an issue with the `REAKeyboardEventObserver`. 

The `_listeners` dictionary (https://github.com/software-mansion/react-native-reanimated/blob/main/ios/keyboardObserver/REAKeyboardEventObserver.m#L16) 

which is touched by `unsubscribeFromKeyboardEventsFunction` (https://github.com/software-mansion/react-native-reanimated/blob/main/ios/native/NativeProxy.mm#L335)

can happen on the non-main thread. There is no synchronizing going on so it's not thread safe. 

I can reliably get unsubscribe to mutate `_listeners` from a JS thread while `updateKeyboardFrame` is happening in the main thread.
(https://github.com/software-mansion/react-native-reanimated/blob/main/ios/keyboardObserver/REAKeyboardEventObserver.m#L102)

I'm not familiar with the code enough to put forward and PR, but thought I would point it out.

### Steps to reproduce

1. Enable the thread sanitizer in xcode
2. Push a view with a keyboard that uses the `useAnimatedKeyboard` hook
3. Pop the view
4. goto 2

### Snack or a link to a repository

sorry i don't have the time now to do this

### Reanimated version

2.11.0

### React Native version

0.69.6

### Platforms

iOS

### JavaScript runtime

JSC

### Workflow

Expo bare workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

chrisnojima:
Related PR: https://github.com/software-mansion/react-native-reanimated/pull/3263

piaskowyk:
Thanks for your help investigation! This is so helpful! 🙌

Issue #3895, opened 2022-12-31
App crashes with EXC_CRASH (SIGABRT) when reloading on iOS


ansh:
### Description

A video of the crash happening
https://user-images.githubusercontent.com/8207733/210131053-51c7ff0a-d9ee-4b10-80a0-1f10c0e273b7.mp4

The error I get after the crash
<img width="372" alt="CleanShot 2022-12-31 at 14 23 42@2x" src="https://user-images.githubusercontent.com/8207733/210131069-8502f312-9462-4007-9875-f748d3de9414.png">
<img width="1112" alt="CleanShot 2022-12-31 at 14 24 00@2x" src="https://user-images.githubusercontent.com/8207733/210131057-f9c97169-28c9-44fe-9e6f-6b01dddda9fd.png">
Full error text: https://textbin.net/apakzpco91

I have tried 3.0.0-rc.9 and rc.8 and they kind of reduce the probability of issue but they do not fix it. Also, they introduce another bug shown below
<img width="300" alt="CleanShot 2022-12-31 at 14 02 12@2x" src="https://user-images.githubusercontent.com/8207733/210131183-f685c1ea-91e9-4581-8171-fa01669fcfe6.png">

### Steps to reproduce

1. Use `npx create-expo-app` to create a new Expo app
2. Install `reanimated` + `expo-dev-client`
3. Build using EAS or `npx expo prebuild` + `npx expo run:ios`
4. Open the app
5. Press the "R" key on the keyboard a couple of times
6. You will most definitely experience a crash

### Snack or a link to a repository

https://github.com/ansh/reanimated-reload-crash

### Reanimated version

2.12.0

### React Native version

0.70.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

iOS simulator

### Acknowledgements

Yes

ansh:
@EvanBacon It's weird that this crash only happens when I use `expo-dev-client` and not when I use Expo Go with the same versions of everything (repro provided above). Tagging you in case you know something about this.

sangameshsomawar:
@ansh did you get any solution? I am getting same error.

xanderdeseyn:
We are seeing the same issue here. Anyone found a solution?

sangameshsomawar:
It got resolved after running `watchman watch-del-all` and  `yarn start --reset-cache`

xanderdeseyn:
[...]

PR #3911, opened 2023-01-04
Fix crash caused by data race in useAnimatedKeyboard

graszka22:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/3568
Fixes https://github.com/software-mansion/react-native-reanimated/issues/3712
Related to https://github.com/software-mansion/react-native-reanimated/pull/3704

The crash was caused because `_listeners` could be accessed by two threads at the same time. For example it could happen that while js thread in `unsubscribeFromKeyboardEvents` was deleting a listener, `updateKeyboardFrame` was iterating over it on the main thread. Now it's synchronised, so only one thread can access `_listeners` at the same time (and `_state` too just in case)

## Test plan

You can reproduce the crash easier by mounting 1000 components that use `useAnimatedKeyboard` in AnimatedKeyboardExample (I tested on simulator). Just enter and exit the example a couple of times.
<details>
  <summary>Example</summary>
  
  ```js
import Animated, {
  useAnimatedStyle,
  useAnimatedKeyboard,
  KeyboardState,
import {
  View,
  Button,
  TextInput,
  StyleSheet,
  Keyboard,
  ScrollView,
} from 'react-native';
import React from 'react';

const BOX_SIZE = 50;

function NestedView(): React.ReactElement {
  useAnimatedKeyboard();
  return <View style={styles.nestedView} />;
}

function AnimatedStyleUpdateExample(): React.ReactElement {
  const OPENING = KeyboardState.OPENING;
    const color = keyboard.state.value === OPENING ? 'red' : 'blue';

    return {
      backgroundColor: color,
    };
  });
    return {
    };
  });
  const [shouldShowNestedView, setShouldShowNestedView] = React.useState(false);

  return (
    <ScrollView
      contentContainerStyle={styles.container}
      keyboardDismissMode="interactive"
      scrollEnabled={false}>
      <Animated.View style={[styles.box, style]} />
      <Button
        title="Toggle nested view"
        onPress={() => {
          setShouldShowNestedView(!shouldShowNestedView);
        }}
      />
      {new Array(1000).fill(null).map((_, i) => (
        <NestedView key={i} />
      ))}
      <Animated.View style={translateStyle}>
        <Button
          title="Dismiss"
          onPress={() => {
            Keyboard.dismiss();
          }}
        />
        <TextInput style={styles.textInput} autoCorrect  autoFocus/>
      </Animated.View>
    </ScrollView>
  );
}

  container: {
    flex: 1,
    flexDirection: 'column',
    alignItems: 'center',
    paddingVertical: 70,
  },
  box: { width: BOX_SIZE, height: BOX_SIZE, marginBottom: 100 },
  textInput: {
    borderColor: 'blue',
    borderStyle: 'solid',
    borderWidth: 2,
    height: 60,
    width: 200,
  },
  nestedView: {
    width: 100,
    height: 100,
    backgroundColor: '#ffff00',
  },
});

export default AnimatedStyleUpdateExample;

  ```
</details>

### DIFF
--- a/ios/keyboardObserver/REAKeyboardEventObserver.m
+++ b/ios/keyboardObserver/REAKeyboardEventObserver.m
@@ -139,44 +139,48 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
 {
   NSNumber *listenerId = [_nextListenerId copy];
   _nextListenerId = [NSNumber numberWithInt:[_nextListenerId intValue] + 1];
-  if ([_listeners count] == 0) {
-    NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
+  RCTExecuteOnMainQueue(^() {
+    if ([self->_listeners count] == 0) {
+      NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
 
-    [notificationCenter addObserver:self
-                           selector:@selector(keyboardWillHide:)
-                               name:UIKeyboardWillHideNotification
-                             object:nil];
+      [notificationCenter addObserver:self
+                             selector:@selector(keyboardWillHide:)
+                                 name:UIKeyboardWillHideNotification
+                               object:nil];
 
-    [notificationCenter addObserver:self
-                           selector:@selector(keyboardWillShow:)
-                               name:UIKeyboardWillShowNotification
-                             object:nil];
+      [notificationCenter addObserver:self
+                             selector:@selector(keyboardWillShow:)
+                                 name:UIKeyboardWillShowNotification
+                               object:nil];
 
-    [notificationCenter addObserver:self
-                           selector:@selector(keyboardDidHide:)
-                               name:UIKeyboardDidHideNotification
-                             object:nil];
+      [notificationCenter addObserver:self
+                             selector:@selector(keyboardDidHide:)
+                                 name:UIKeyboardDidHideNotification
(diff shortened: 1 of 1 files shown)


tomekzaw:
I'm just wondering if it would be possible to schedule a call on the UI thread from JS thread instead of using `@synchronize` everywhere. 

It seems like the only two methods that are called on the JS thread are `subscribeFromKeyboardEvents` and `unsubscribeFromKeyboardEvents`. The latter should be easy as it doesn't return anything. On the other hand, `subscribeFromKeyboardEvents` returns the listener id, but we can assign it while still on the JS thread and copy the integer into the block that will be executed on the UI thread.

This would mean that the registration of listeners is asynchronous, but we no longer need synchronization in notification methods (`keyboardDidShow` etc.).

graszka22:
@tomekzaw  That works too :)

tomekzaw (approved):
Looks good to me. One thing that needs to be checked before merging is whether reloading still works because using `self` in blocks directly may lead to reference cycles.

graszka22:
Reloading doesn't work anyway https://github.com/software-mansion/react-native-reanimated/issues/3786 and I'm currently trying to fix it. It crashes for animated sensors too and probably in other places (https://github.com/software-mansion/react-native-reanimated/issues/3895). We can hold off on merging this until reloading is fixed.

PR #3922, opened 2023-01-09
Fix issues with relative paths when bundling web apps

jwajgelt:
## Summary
Fixes issues when bundling web apps using reanimated due to relative path to `package.json` being wrongly resolved in prebuild reanimated modules.

### DIFF
--- a/src/reanimated2/platform-specific/checkVersion.ts
+++ b/src/reanimated2/platform-specific/checkVersion.ts
@@ -1,0 +1,28 @@
+import { version as jsVersion } from '../../../package.json';
+
+/**
+ * Checks that native and js versions of reanimated match.
+ */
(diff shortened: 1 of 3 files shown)


src/reanimated2/platform-specific/checkVersion.ts
```diff
@@ -0,0 +1,28 @@
+import { version as jsVersion } from '../../../package.json';
```
nandorojo, line 1: webpack won't like this since JSON is treated as a default import. it should be `import pkg` and then do `pkg.version` later on.

nandorojo: Disregard, just saw this is a native-only file.

PR #3930, opened 2023-01-11
Optimize the way we extract code string and in which we pass it across runtimes

kmagiera:
## Summary

This PR changes the way we extract code string and additional metadata for worklets and the way we pass it later on between runtimes such that we can save on allocating code strings that tend to be relatively big and waste precious time on allocating those strings. 

The scope of the change includes:
1. Changes in plugin code, we no longer assign code string directly to worklet function objects, instead we extract a metadata object and define in in the program context (at the main file level). The metadata includes things that are static, like the code converted to string (formerly "asString", location, source map and stack details data). We then reference that metadata object and assign it to the worklet object under `__initData` field.
2. We update "Core Functions" that need to access code string such that it fetches the string from a new location  – `__initData.code`
3. We refactor the concept of "retaining values" and only conditionally retain JS values on the UI runtime from the shareable. The purpose of this change is that we can be sure the initData is persisted on the UI runtime and not recreated each time the worklet it passed to be executed on the UI runtime. We add additional flag `shouldRetain` to the `makeShareableClone` JSI method that allows us to control this behavior. The refactoring updates the inheritance scheme and makes `ShareableObjects` etc no longer inherit `RetainingShareable` but instead we only use `RetainingShareable` when the flag is set. This change also deletes the previous behavior where we'd use weak objects to keep the reference to the JS Value as we are unsure about positive/negative impact of that code (we know it impacts GC times negatively but not sure if this time is counterbalanced by saving on the time to convert the values)

## Test plan

Plugin tests needs to be updated. Wanted to put this change out for others to test.


plugin.js
```diff
@@ -563,92 +592,41 @@ function makeWorklet(t, fun, state) {
...
-  }
-
-  if (options && options.optFlags) {
-    statements.push(
-      t.expressionStatement(
-        t.assignmentExpression(
-          '=',
-          t.memberExpression(
-            privateFunctionId,
-            t.identifier('__optimalization'),
```
tomekzaw, line 648: Since we no longer read `options.optFlags`, I think we can also remove the assignment `options.optFlags = isPossibleOptimization(transformed.ast);` along with the definition of `isPossibleOptimization`.

src/reanimated2/initializers.ts
```diff
@@ -29,36 +29,38 @@ function valueUnpacker(objectToUnpack: any, category?: string): any {
     workletsCache = global.__workletsCache = new Map();
     handleCache = global.__handleCache = new WeakMap();
   }
-  if (objectToUnpack.__workletHash) {
-    let workletFun = workletsCache.get(objectToUnpack.__workletHash);
+  const workletHash = objectToUnpack.__workletHash;
+  if (workletHash) {
```
tomekzaw, line 33: I know it's highly unlikely but this won't work properly if `workletHash` equals 0. Can we add `!== undefined` here?

src/reanimated2/shareables.ts
```diff
@@ -54,13 +59,27 @@ export function makeShareableCloneRecursive<T>(value: any): ShareableRef<T> {
...
+          if (__DEV__) {
+            registerWorkletStackDetails(
+              value.__workletHash,
+              value.__initData.stackDetails
+            );
+          }
+          // to save on transferring static __initData field of worklet structure
+          // we request shareable value to persist its UI counterpart. This means
+          // that the __initData field that contains long strings represeting the
+          // worklet code, source map, and locaiton, will always be
```
tomekzaw, line 74:
```suggestion
          // worklet code, source map, and location, will always be
```

tomekzaw:
Don't know why but this PR breaks the app when using JSC (Hermes is fine):

Zrzut ekranu 2023-01-12 o 15 09 01

kmagiera:
There was one issue with JSC related to Object.freeze use. I fixed that, updated other comments, made tests pass and think this is ready for review

tomekzaw (approved):
I've tested this PR on iOS (debug/release mode, JSC/Hermes) and it works fine ✅

Issue #3721, opened 2022-10-26
useAnimatedKeyboard doesn't update height value on input change


thecoorum:
### Description

While using `useAnimatedKeyboard` hook I noticed that on changing inputs with different keyboard types (keyboard suggestions are enabled or disabled) it doesn't update the height value which causes an incorrect styling of element

### Steps to reproduce

1. Create a page both with enabled and disabled suggestions `TextInput`'s
2. Use `useAnimatedKeyboard` hook to get keyboard height value
3. Switch between inputs and check the returned values from hook

### Snack or a link to a repository

https://github.com/thecoorum/reanimated-keyboard-issue

### Reanimated version

2.10.0

### React Native version

0.69.6

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

kbrandwijk:
I noticed the same thing switching from number-pad to regular keyboard with suggestions, height is not updated.

graszka22:
In your example you're using `useAnimatedKeyboard()` directly in your component like this:
```js
export default function App() {
  const keyboard = useAnimatedKeyboard();

  // Should update on switching between keyboard types
  console.log(keyboard.height.value);

  return (
    ...
  );
}
```
But `useAnimatedKeyboard()` is rather meant to be used for example with `useAnimatedStyle` hook, as shown in the example in the documentation: https://docs.swmansion.com/react-native-reanimated/docs/next/api/hooks/useAnimatedKeyboard#example. If you add `console.log(keyboard.height.value)` in the `useAnimatedStyle` you'll see that the height of the keyboard indeed changes.

`useAnimatedKeyboard()` does not cause a component re-render when keyboard height changes and that's why `console.log` is not called. The hook just returns a shared value with keyboard height that is updated without the re-renders. And that's a good thing, because `useAnimatedKeyboard()` tracks the height of the keyboard while it's animating. That means that when showing the keyboard the height of the keyboard can change at a rate of 60 times per second. You wouldn't want to re-render your component 60 times per second, because you'd get a slideshow instead of an animation. 

If you'd want to re-render your component anyway (I don't know your use case) when the keyboard opens/closes you can do so with `useAnimatedReaction` hook. Here is an example: https://gist.github.com/graszka22/a7933e69927fcc4f603abefe46f533b8
[...]

PR #3940, opened 2023-01-12
Fix keyboard height in useAnimatedKeyboard on language change on iOS

graszka22:
## Summary

Fixes a bug when changing keyboard language on iOS:

Bug:
https://user-images.githubusercontent.com/6280457/212095624-4df7e0b0-d3e6-4bdd-8378-ad8c077e6fb2.mp4

Fixed bug:

https://user-images.githubusercontent.com/6280457/212096575-aa36cb88-fee5-4f4c-bd37-676973834d48.mp4

Fixes https://github.com/software-mansion/react-native-reanimated/issues/3721

`useAnimatedKeyboard` reported invalid keyboard height in that case, because we invalided display link too early, in `stopAnimation` and we reported the height before suggestions bar appeared. In the next frame the suggestion bar appears and we can calculate correct keyboard height.

## Test plan

Run AnimatedKeyboardExample, bring up the keyboard and change keyboard language. The text input should be above the keyboard in correct position.

<details>
  <summary>Example for switching between numeric/regular keyboard:</summary>

```js
import React from 'react';

import { TextInput, View, StyleSheet } from 'react-native';

import { useAnimatedKeyboard, useAnimatedStyle } from 'react-native-reanimated';

  const { height } = useAnimatedKeyboard();

  useAnimatedStyle(() => {
    console.log('HEIGHT', height.value);
    return {};
  });

  return (
      <TextInput style={styles.input} keyboardType="default" />
      <TextInput style={styles.input} keyboardType="number-pad" />
    </View>
  );
}

  container: {
    flex: 1,
    backgroundColor: '#fff',
    padding: 20,
  },
  input: {
    borderBottomWidth: 1,
    borderBottomColor: '#ccc',
    marginBottom: 20,
    backgroundColor: '#ddd',
    height: 50,
  },
});
```
</details>

### DIFF
--- a/ios/keyboardObserver/REAKeyboardEventObserver.m
+++ b/ios/keyboardObserver/REAKeyboardEventObserver.m
@@ -18,6 +18,7 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
   int _windowsCount;
   UIView *_keyboardView;
   KeyboardState _state;
+  bool _shouldInvalidateDisplayLink;
 }
 
 - (instancetype)init
@@ -26,6 +27,7 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
   _listeners = [[NSMutableDictionary alloc] init];
   _nextListenerId = @0;
   _state = UNKNOWN;
+  _shouldInvalidateDisplayLink = false;
   return self;
 }
 
@@ -85,9 +87,11 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
 
 - (void)stopAnimation
 {
-  [displayLink invalidate];
-  displayLink = nil;
   [self updateKeyboardFrame];
+  // there might be a case that keyboard will change height in the next frame
+  // (for example changing keyboard language so that suggestions appear)
+  // so we invalidate display link after we handle that in the next frame
+  _shouldInvalidateDisplayLink = true;
 }
 
 - (void)updateKeyboardFrame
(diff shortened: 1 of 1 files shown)


kbrandwijk:
As mentioned in the linked issue, this happens in all cases where the suggestion bar appears/disappears, so not only language switches, but also switching from numeric to regular keyboard, or having two TextInputs with different settings for autoSuggest.

graszka22:
I tested and PR also fixes the issue with switching between numeric/regular keyboard too.

azcarraga:
@graszka22 thank you for looking into this, I've hit this as well but with when toggling emoji / text on the keyboard. I tried your fix as a patch locally, but the keyboard height is still not updating in tandem with the keyboard - it updates on the next toggle to the previous height.

I've created a snack with the repro:

https://snack.expo.dev/@aazcarraga/reanimated-keyboard-bug-repro

1) Click on the text input at the bottom of the screen, which will pull up the keyboard
2) Press the emoji button on the keyboard to toggle emoji - notice how the text input is now hidden (keyboard height not updated)
3) Press the 'ABC' button on the keyboard to toggle back - notice how the text input is now too high (keyboard height is now updated to the height it should have been for emoji)

This seems related and might be useful in validating this fix. It could be that I did not patch correctly when testing this change locally.

Thanks again for looking into this, the keyboard animation is so useful and it's hard to not depend on it everywhere now that it's available.

graszka22:
@azcarraga Your example is working for me.

Before fix:

https://user-images.githubusercontent.com/6280457/212907045-69d1468c-8dab-4397-b762-9a9e23d96a6e.mp4

After fix:

https://user-images.githubusercontent.com/6280457/212907068-291bdb7c-dd80-4610-9f73-00228aa7dc0e.mp4

If it's not working for you please describe how are you testing (simulator or real device, what device, versions of RN, Paper/Fabric architecture, make sure you patched and rebuilt correctly etc.)

azcarraga:
@graszka22 thank you so much for the quick reply and videos - that's great news (it works) and narrows down what I'm seeing to my patching. Let me dig in to my patching and report back here.

But based on your videos, clearly your fix works which is what matters - thank you!

azcarraga:
@graszka22 I can confirm it's working locally for me as well! I used `yarn patch` and then `yarn install` + delete pods + `pod install` and it all works perfect.

Thanks for your help and for this fix.

piaskowyk (approved):

PR #3945, opened 2023-01-13
Don't look in default path for libs if path is procided

reinismu:
## Summary

I have a `libglog.so` in my default path so CMAKE takes it instead of provided one.
Build failed then.

## Test plan

Put `libglog.so` in another architecture in your default path and you will see how linker fails.


tomekzaw (approved):
@reinismu Thanks for the PR! 🥳

piaskowyk (approved):

PR #3950, opened 2023-01-13
Optimize previous style calculation in `styleUpdater`

tomekzaw:
## Summary

This PR contains two little optimizations in `styleUpdater`:

1. Since we never use `oldValues` again (because references the same object as `state.last` does before we overwrite it in the very same line), we can re-use this instance instead of creating a new one.
2. Since we just calculated the difference between the new and old styles, we can overwrite `oldValues` just with `diff` instead of `newValues` because the values that are not present in `diff` are the same in `oldValues` and `newValues`.

Co-authored-by: @kmagiera

## Test plan

Just check if BokehExample works.

### DIFF
--- a/src/reanimated2/hook/useAnimatedStyle.ts
+++ b/src/reanimated2/hook/useAnimatedStyle.ts
@@ -353,7 +353,7 @@ function jestStyleUpdater(
 
   // calculate diff
   const diff = styleDiff(oldValues, newValues);
-  state.last = Object.assign({}, oldValues, newValues);
+  state.last = Object.assign(oldValues, diff);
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):

nandorojo:
Is this fix pulled into an rc for v3, or does it not apply? I'm getting weird errors with v2 and want to try v3, but not sure if it's stable enough or not.

tomekzaw:
@nandorojo This PR has been reverted both from v3 and v2 as it breaks `useAnimatedStyle` in some cases. See #3980.

PR #3970, opened 2023-01-18
Add setImmediate implementation to the UI runtime

kmagiera:
## Summary

This PR adds an implementation of web's setImmediate API to the reanimated's UI runtime. We utilize this implementation in mappers and event handlers for more accurate animation/gesture tracking as well as unblock further optimizations.

There are several things that this PR changes:
1) We add event timestamp metadata to touch/scroll/sensor/etc events. This change is in preparation of providing a true system timestamps for such events (we don't do that yet but instead call system timer in order to pass down the current time)
2) We update performance.now API to use the same reference point for the provided time as requestAnimationFrame API. Before performance.now would use a different reference point which made it impossible to use it as a starting point for animations.
3) We update event handler callbacks such that they perform a flush for rAF and immediates queues
4) We populate global.__frameTimestamp for all the code that executes within requestAnimationFrame and event callbacks. We then use it in useAnimatedStyle as a reference point for starting animations and in case it wasn't set, we use performance.now
5) We add some code in initializers.ts to define setImmediate method and to register method for flushing immediates queue
6) We introduce set-immediate-based batching for runOnUI calls on the main RN runtime. We then flush all "immediates" after the whole batch is processed on the UI runtime. 

## Test plan

Check a bunch of examples from the example app: "Use animated style", "Bokeh", "Drag and snap".
For touch based interaction verify that movement doesn't lag behind the event. The way this can be done is by screen recording drag and snap example with touch indicators on, then checking on the recorded clip in between frames whether there is a correct movement aligned with the updates of the touch indicator.
The above needs to be done for all configurations: Android/iOS/Fabric/Paper


android/src/main/cpp/NativeProxy.cpp
```diff
@@ -323,7 +303,7 @@ void NativeProxy::installJSIBindings(
   std::shared_ptr<UIManager> uiManager =
       binding->getScheduler()->getUIManager();
   module->setUIManager(uiManager);
-  module->setNewestShadowNodesRegistry(newestShadowNodesRegistry_);
+  module->setNewestShadowNodesRegistry(newezstShadowNodesRegistry_);
```
jwajgelt, line 323:
```suggestion
  module->setNewestShadowNodesRegistry(newestShadowNodesRegistry_);
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -425,20 +420,10 @@ void NativeReanimatedModule::maybeRequestRender() {
...
-  } catch (std::exception &e) {
-    std::string str = e.what();
-    this->errorHandler->setError(str);
-    this->errorHandler->raise();
-  } catch (...) {
-    std::string str = "OnRender error";
-    this->errorHandler->setError(str);
-    this->errorHandler->raise();
+  std::vector<FrameCallback> callbacks = frameCallbacks;
+  frameCallbacks.clear();
```
jwajgelt, line 428: You should be able to just move-construct `callbacks` from `frameCallbacks` here, right?
```suggestion
  std::vector<FrameCallback> callbacks = std::move(frameCallbacks);
```

kmagiera: This wasn't my code so prefer not to make this change. Is this behavior defined that the original vector gets reset when moved? Typically move constructors does not specify behavior of the moved object as normally the assumption is that the moved object is never used again. Here we still want to use the vector

jwajgelt: It should be fine.
C++ standard says that the moved-from object is left in a "valid but unspecified state" (so it should be fine to use the old vector later), and the `std::vector` [spec](https://en.cppreference.com/w/cpp/container/vector/vector) further specifies that "After the move, `other` is guaranteed to be [empty()](https://en.cppreference.com/w/cpp/container/vector/empty)." when move-constructing a vector, so it should clear it as well.

In case you don't feel safe using a vector after moving from it, I would still suggest doing something like
```suggestion
  std::vector<FrameCallback> callbacks;
  std::swap(callbacks, frameCallbacks)
```
since this is unambiguous in what it does, but still avoids copies/allocations.

jwajgelt: I'd just like to avoid copying the callbacks here, since they can potentially capture a bunch of things - in this case it's a couple of `shared_ptr`s, which means unnecessary work on the atomic reference counters inside.

src/reanimated2/mappers.ts
```diff
@@ -74,23 +74,23 @@ export function createMapperRegistry() {
...
+  function mapperRun() {
+    runRequested = false;
     if (mappers.size !== sortedMappers.length) {
       updateMappersOrder();
     }
     for (const mapper of sortedMappers) {
       if (mapper.dirty) {
         mapper.dirty = false;
-        mapper.worklet();
+        mapper.worklet(7);
```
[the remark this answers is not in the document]
kmagiera: idk, must've got here by mistake, mapper's worklets don't take arguments

src/reanimated2/threads.ts
```diff
@@ -14,11 +55,52 @@ export function runOnUI<A extends any[], R>(
...
+      // somce callbacks have been added to a batch from the previous test. To fix this
+      // we reset the batch queue when setImmediate function changes.
+      if (_lastSetImmediateFunction !== setImmediate) {
+        _lastSetImmediateFunction = setImmediate;
+        _runOnUIQueue = [];
+      }
+    }
+    _runOnUIQueue.push([worklet, args]);
+    if (_runOnUIQueue.length === 1) {
+      setImmediate(() => {
```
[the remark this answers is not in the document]
kmagiera: It is meant to the same setImmediate. We provide a compatibility layer with the exception that reanimated's version doesn't support cancelling callbacks (yet). So we aim for both setImmediate from RN runtime and UI runtimes to be compatible and have same behavior hence we use the standard naming

src/reanimated2/mappers.ts
```diff
@@ -74,23 +74,23 @@ export function createMapperRegistry() {
...
       }
     }
   }
 
   function maybeRequestUpdates() {
-    if (!frameRequested) {
-      requestAnimationFrame(mapperFrame);
-      frameRequested = true;
+    if (!runRequested) {
+      setImmediate(mapperRun);
```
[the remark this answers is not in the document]
kmagiera: `rAF` will delay the run of the mappers by one frame – the callback will run happen in the following frame. Whereas `setImmediate` will execute the callback at the end of the tasks that are currently scheduled on the run loop. The reason we schedule the run in the first place is that there is a chance multiple mappers gets updated in one frame and so we want to run the updates only once as it requires traversing dependency graph.

bglgwyng:
@kmagiera 

I posted a question https://github.com/software-mansion/react-native-reanimated/discussions/3925 a month ago and haven't gotten any answer yet.

I just saw this PR by chance and I don't know the details and context of this PR at all. However, the quotations sound quite relevant to my previous question.

> We populate global.__frameTimestamp for all the code that executes within requestAnimationFrame and event callbacks. We then use it in useAnimatedStyle as a reference point for starting animations and in case it wasn't set, we use performance.now

> We introduce set-immediate-based batching for runOnUI calls on the main RN runtime. We then flush all "immediates" after the whole batch is processed on the UI runtime.
Does this change solves the problem discussed in the discussion?

kmagiera:
This change impacts the way we execute updates on the main thread. Now they are all batches and will run in a single transaction. So I'd expect this pr to change the way your example work but the easiest way to find out would be to just try it out.

eweilow:
Hi! Can create an issue as well, but I think this PR may have broken running react-native-reanimated on the Web without a Babel plugin.

The issue we are getting is that any use of `react-native-reanimated` throws `runOnUI() can only be used on worklets` coming from this line
https://github.com/software-mansion/react-native-reanimated/pull/3970/files#diff-e2f6e662599fdf6a81975be4bbd34818651984e9076077405b709d7ae6354680R96


In the PR that adds support for running without a Babel plugin (https://github.com/software-mansion/react-native-reanimated/pull/3997), the `runOnUI` function is changed to add a `!shouldBeUseWeb()` check before testing  the `__workletHash`:

```diff
export function runOnUI<A extends any[], R>(
  worklet: ComplexWorkletFunction<A, R>
): (...args: A) => void {
-  if (__DEV__) {
+  if (__DEV__ && !shouldBeUseWeb()) {
```

From what I can tell, the same needs to be done for `runOnUIImmediately`:

```diff
export function runOnUIImmediately<A extends any[], R>(
  worklet: ComplexWorkletFunction<A, R>
): (...args: A) => void {
-  if (__DEV__) {
+  if (__DEV__ && !shouldBeUseWeb()) {
    if (worklet.__workletHash === undefined) {
```

Perhaps the error message `runOnUI() can only be used on worklets` should also be changed to `runOnUIImmediately() can only be used on worklets`?

nandorojo:
This PR also added a regression on Next.js, which utilizes `requestAnimationFrame` rather than `setImmediate`, which isn't available on the server. I opened an issue here: https://github.com/software-mansion/react-native-reanimated/issues/4140

nandorojo:
@eweilow Yeah you're right, I opened a PR for that fix here: https://github.com/software-mansion/react-native-reanimated/pull/4141

src/reanimated2/mappers.ts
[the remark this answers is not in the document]
nandorojo: One issue with this implementation is that it's historically very problematic with Next.js Next.js apps seem to break on the server when using `setimmediate`.

Could we update it to do this?

```tsx
const scheduleUpdates = typeof window !== 'undefined' ? setImmediate : requestAnimationFrame
```

This way, `setImmediate` won't get used on the server, which leads to #4140

PR #3801, opened 2022-11-22
Adds mocks for default layout animations

tonibardina:
## Description

This Pr adds missing mocks for default layout animations.

## Changes

- Adds missing reanimated2 default layout animations mocks

## Test code and steps to reproduce

```javascript
// Component.tsx
export const AnimatedComponent =
  Animated.createAnimatedComponent(ComponentToAnimate);

const Component = () => (
    <View>
       <AnimatedComponent entering={FadeIn} />
    </View>
)

// Component.test.tsx

describe(Component, () => {
  it('should render', () => {
     render(<Component/>); // Before these mocks were added this would fail (Cannot call `.delay()` or `.duration()` of undefined )
  });
});
```

## Checklist

- [X] Included code example that can be used to test this change
- [X] Updated TS types
- [x] Added TS types tests
- [x] Added unit / integration tests
- [x] Updated documentation
- [X] Ensured that CI passes

PR #3974, opened 2023-01-19
Extend the set of layout animation mocks

jalooc:
## Summary

This PR extends https://github.com/software-mansion/react-native-reanimated/pull/3801 by introducing layout animation types and modifiers that those animation types require to be fully functional.

Currently, the tests fail with something like:
```
TypeError: Cannot read properties of undefined (reading 'delay')

 > 107 |   const transition = CurvedTransition.delay(10).duration(10),
       |                                       ^
```

## Test plan

Creating a test (e.g. using React Native Testing Library) of a component containing any of the freshly added layout animation types and / or chained modifiers they support. E.g.:
```tsx
const MyComponent = () => (
  <Animated.View layout={CurvedTransition.delay(10).duration(10)} />
)

test('test the test', () => {
  const screen = render(<MyComponent />)
})
```
The above example fails before the patch from this PR is introduced and succeeds afterward.

## Question

One question to resolve is how closely the mocks should match the real api in the context of which modifiers apply to witch animation types (e.g. not all animation types support `.stiffness()`). Doing so would require more complex mocks that are also harder to maintain for the need to keeping them in sync. An alternative, simpler approach is to add all the modifiers to all the animations - I figure it's okay since these kinds of mocks are not directly consumed in the tests - they just serve enabling test-rendering components utilising layout animations without errors. The latter is the approach I went with, but lmk if you think the other one is a better fit.

### DIFF
--- a/src/reanimated2/mock.ts
+++ b/src/reanimated2/mock.ts
@@ -16,6 +16,18 @@ class BaseAnimationMock {
     return this;
   }
 
+  springify(_: number) {
+    return this;
+  }
+
+  damping(_: number) {
+    return this;
+  }
+
+  stiffness(_: number) {
+    return this;
+  }
+
   withCallback(_: (finsihed: boolean) => void) {
     return this;
   }
@@ -26,6 +38,10 @@ class BaseAnimationMock {
 
   withInitialValues() {
     return this;
+  }
+
+  build() {
+    return () => ({ initialValues: {}, animations: {} });
   }
 }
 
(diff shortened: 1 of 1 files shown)


piaskowyk:
@jalooc thanks for your PR! 🙌 When it comes to mocks, I think that they should be as simple as possible because mocks don't contain any logic. So adding modifiers to all the animations is fine for me.

piaskowyk (approved):

PR #3980, opened 2023-01-19
Optimize hot code-paths in uAS

kmagiera:
## Summary

This is another attempt at optimizing useAnimatedStyle and style updater after #3950 got reverted. In this approach we omit Object.assign completely in favor of just using newValues object. Since newValues may contain animation objects, we use one of the initial loops from the updater to extract non-animated-props such that they can be used later on instead of using a call to `filerAnimatedProps`. We also replace `styleDiff` that'd allocate new object by `shallowEqual` that only does comparison. We've been using newValues to update the values anyways and so the diff wasn't used. In addition there was a bug before where we had `if (diff)` check – this wasn't correct as in case there were no changes between the objects, the diff would be an empty object and the check would pass.

## Test plan

Tested on a bunch of different examples. Make sure to verify on examples that use animations inside uAS.


tomekzaw (changes requested):
Tested on Example app on iOS as well as FabricExample on Android. Everything seems to work just fine except for BokehExample 😢

kmagiera:
I think this error was fixed in f156b1294bccd052d0fc09252bbf3a5e30eb81ad – can I get this approved then 🙏

tomekzaw (approved):

piaskowyk (approved):

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -244,21 +248,19 @@ function styleUpdater(
         requestAnimationFrame(frame);
       }
     }
-    state.last = Object.assign({}, oldValues, newValues);
-    const style = getStyleWithoutAnimations(state.last);
-    if (style) {
-      updateProps(viewDescriptors, style, maybeViewRef);
+
+    if (hasNonAnimatedValues) {
+      updateProps(viewDescriptors, nonAnimatedNewValues, maybeViewRef);
```
Latropos, line 253: Why do we update nonAnimatedNewValues?

PR #3988, opened 2023-01-21
Fix attaching a scroll event if the scrollable component is not the root view

### LINKED ISSUE #3873: Cannot attach native event on animated component
### Description

When creating an animated component with:
```js
const AnimatedComponent = createAnimatedComponent(Component)
```
To do something like this:
```js
const Example = ({ onScroll }) => {
 const scrollHandler = useAnimatedScrollHandler({ onScroll })
 
 return (
  <AnimatedComponent onScroll={scrollHandler} />
 )
}
```
 I cannot attach a scroll event if the scrollable component is not the root view, for example, when the `Component` is:
```js
const Component = forwardRef((props, ref) => (
 <View>
   <ScrollableComponent ref={ref} {...props} />
 </View>
))
```
where, in this example, it works just fine (it also works when wrapped with a fragment):
```js
const Component = forwardRef((props, ref) => <ScrollableComponent ref={ref} {...props} />)
```

In version `2.10.0`, the scroll works in both examples, [this change](https://github.com/software-mansion/react-native-reanimated/commit/5829e95b5e452761c659cb7fb73bb98a630f86c1#diff-fdad18d3281131f92aa76c0969f84d27f2314251842763fedbc756df07b1eab7R175) might be the culprit, from my (little) understanding we used to get the component node (scrollable component) and attach the event to it but now we're attaching it to the root view. Is this on purpose or a bug?

This is useful in some situations, for example, in the App I'm working at the moment, we use an animated wrapper around an external component that has this type of structure.

### Steps to reproduce

In the repo provided, you have the two examples on screen, you can test that one gets the event callback but the other doesn't.

https://github.com/diogoquintas/reanimated-scroll-issue

### Reanimated version

3.0.0-rc.8

### React Native version

0.70.1

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 14

### Acknowledgements

Yes

graszka22:
## Summary

Fixes #3873
`getScrollableNode()` works not only on FlashLists, but also on RN scrollable components (ScrollView, Flatlist). It's brought back from [there](https://github.com/software-mansion/react-native-reanimated/commit/5829e95b5e452761c659cb7fb73bb98a630f86c1#diff-fdad18d3281131f92aa76c0969f84d27f2314251842763fedbc756df07b1eab7L234)

## Test plan

- tested on the repro from the issue (https://github.com/diogoquintas/reanimated-scroll-issue)
- tested on several examples from the example app (examples related to scroll events and events in general for example chat heads)
- tested on FlashList

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -189,15 +189,17 @@ export default function createAnimatedComponent(
       this._attachAnimatedStyles();
     }
 
+    _getEventViewRef() {
+      // Make sure to get the scrollable node for components that implement
+      // `ScrollResponder.Mixin`.
+      return this._component?.getScrollableNode
+        ? this._component.getScrollableNode()
+        : this._component;
+    }
+
     _attachNativeEvents() {
-      let viewTag = findNodeHandle(this);
-
-      const componentName = Component.displayName || Component.name;
-
-      if (componentName?.endsWith('FlashList') && this._component) {
-        // @ts-ignore it's FlashList specific: https://github.com/Shopify/flash-list/blob/218f314e63806b4fe926741ef73f8b9cd6ebc7eb/src/FlashList.tsx#L815
-        viewTag = findNodeHandle(this._component.getScrollableNode());
-      }
+      const node = this._getEventViewRef();
+      const viewTag = findNodeHandle(options?.setNativeProps ? this : node);
 
       for (const key in this.props) {
         const prop = this.props[key];


tomekzaw (approved):
Nice!

piaskowyk (approved):

PR #3993, opened 2023-01-25
Fix nested animated props on iOS

mstach60161:
## Summary

This PR fixes the [issue](https://github.com/rnmapbox/maps/issues/1200) from the [mapbox](https://github.com/rnmapbox/maps) library on iOS. When a user tried to use nested props with `useAnimatedProp`, they got parsing error.

Fix:
- Added dictionary to JSON conversion, if the prop value is a dictionary in `REANodesManager.mm`

## Test plan

- Run animation using nested props on iOS.


hirbod:
Happy to see that you've found a solution. So soon we can go ahead and remove our patches and JSON.stringify hacks :)

tyrauber:
Awesome! Thanks so much @mstach60161.

PR #3997, opened 2023-01-27
Add web support without a Babel/SWC Plugin

nandorojo:
## Description

This PR adds support for using Reanimated without a plugin on Web. 

### Before 😢

Without a Babel plugin (or SWC on Next.js), animations do nothing.

https://user-images.githubusercontent.com/13172299/214991524-0819fb9d-d72f-4a2c-b067-3adff7818a50.mp4

### After ✅

Everything now works flawlessly and reactively on Web, including `useAnimatedStyle`, `useDerivedValue`, and `useAnimatedReaction`. 

https://user-images.githubusercontent.com/13172299/214991572-c3a355c2-1a54-43d2-965b-916463b0e281.mp4

The only requirement to make this work is explicitly using dependency arrays.

## Changes

In places where the `._closure` is undefined, we set the `inputs` to equal the `dependencies` (so long as they have been explicitly set). 

This makes Reanimated work on Web, even if there is no Babel/SWC plugin set.

Lastly, I added a simple check in `Mapper` to make sure we aren't trying to get object keys of DOM elements.

Let me know the best way to do this. I tested the code in a separate repo outside of this one, and it works well on Web. Native won't be affected, since all changes are behind a `shouldBeUseWeb()` flag.

## Checklist

- [x] Updated TS types
- [ ] Added TS types tests
- [x] Updated documentation
- [x] Ensured that CI passes


nandorojo:
For some reason it looks like I created my PR from the wrong branch, since it seems like I I have a bunch of mismatched files. Will try to fix that.

nandorojo:
Fixed those issues I think. Will see why CI is failing.

nandorojo:
Fixed CI.

docs/docs/fundamentals/web-support.md
```diff
@@ -86,3 +86,60 @@ module.exports = {
   },
 };
 ```
+
+## Web without a Babel plugin
+
+As of Reanimated `2.15`, the Babel plugin (`react-native-reanimated/plugin`) is optional on Web.
```
kmagiera, line 92: I think it is worth noting here that adding dep array doesn't break behavior on mobile, hence can be used on code that is planned to be reused across mobile and web

kmagiera (approved):
Looks good, thanks for investigating and working on this. I left just one minor comment for the docs section. Also, for the test plan section it'd be great if you could put a part of the code snipped you used for testing. Specifically, I'd be important to verify this works ok when dependencies array contains objects that are not shared values.

nandorojo:
# Test Case

I added the following test case to the `WebExample`:

```sh
yarn web:plugin

# or:
DISABLE_BABEL_PLUGIN=1 yarn web
```

The `WithoutBabelTest` component uses a dependency array, and has [multiple types of dependencies](https://github.com/software-mansion/react-native-reanimated/pull/3997/files#diff-bbb183e65396e618ea50951a056e2112984c3deb586bbff6e4a3686db15275acR44): shared values, booleans, numbers, objects, etc. It also has a timer to trigger renders.

## Without Babel Plugin ✅

The top circle has no dependency array, so it doesn't work. The bottom one is the same, with an array, and it works.


https://user-images.githubusercontent.com/13172299/216391498-bd2aaa0d-8dcc-45d9-bee5-f0aecc5cfd81.mp4

## With Babel Plugin ✅

By running `yarn web`, we re-enable the Babel plugin (and the Babel cache is [automatically invalidated](https://github.com/software-mansion/react-native-reanimated/pull/3997/files#diff-13397acfefe9b2906bcd165469c1dd2b4a319fcfe2239754d8985626fd5a9547R4)).

Now, both circles work, including the one without a dependency array:


https://user-images.githubusercontent.com/13172299/216392349-4312bfb1-4caf-4137-9036-e8a4489bb0e0.mp4

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -400,7 +400,20 @@ export function useAnimatedStyle<T extends AnimatedStyle>(
...
   const viewsRef: ViewRefSet<any> = makeViewsRefSet();
   const initRef = useRef<AnimationRef>();
-  const inputs = Object.values(updater._closure ?? {});
+  let inputs = Object.values(updater._closure ?? {});
+  if (shouldBeUseWeb()) {
+    if (!inputs.length && dependencies?.length) {
+      // let web work without a Babel/SWC plugin
+      inputs = dependencies;
+    }
+    if (!inputs.length && !dependencies && !updater.__workletHash) {
```
kmagiera, line 409:
```suggestion
    if (__DEV__ && !inputs.length && !dependencies && !updater.__workletHash) {
```

kmagiera:
This is great 👏 👏 👏 👏  Thanks for adding a sample, I'm sure this will be helpful in the future.

PR #4027, opened 2023-02-03
`RuntimeDecorator` cleanup

jwajgelt:
## Summary
Reduces boilerplate around installing native JSI functions in `RuntimeDecorator`.

We need to install Host Functions into the JS runtimes that will be called by reanimated.
However, these functions need to use the JSI calling convention, which means having a signature `jsi::Value(Runtime &runtime, jsi::Value const &thisValue, jsi::Value const *args, size_t count)`, but we want to write host functions with regular C++ functions (i.e. taking a specific number of typed and named arguments) instead.
The current solution for this problem is to wrap the host functions in lambdas which take the JSI arguments and convert them to the C++ ones, for example
```C++
  auto _scrollTo = [scrollTo](
                       jsi::Runtime &rt,
                       const jsi::Value &thisValue,
                       const jsi::Value *args,
                       const size_t count) -> jsi::Value {
    int viewTag = static_cast<int>(args[0].asNumber());
    double x = args[1].asNumber();
    double y = args[2].asNumber();
    bool animated = args[3].getBool();
    scrollTo(viewTag, x, y, animated);
  };
```
However, since almost the same code has to be written for each host function, this introduces a lot of similar boilerplate functions in `RuntimeDecorator`, reducing readability.

We'd like to instead have a generic solution, which can take a regular C++ `std::function<ReturnType(ArgumentTypes...)>` and create this wrapper for us.
This PR accomplishes that by implementing a generic function `convertArgs<ArgumentTypes...>(jsi::Runtime &rt, const jsi::Value *args, const size_t count)` which walks over the `args`, calls appropriate functions to cast them to the native C++ types `ArgumentTypes...` and returns a tuple with the result.
Then, this result is used by `createJsiFunction(function)` to wrap the `function` in a lambda which takes the JSI arguments, calls `convertArgs` on them, and calls the `function` with them.

This approach has some limitations:
- the conversion to native types is done using the `T get<T>(jsi::Value const *)` function, which is defined only for a few basic types. This restricts the argument types for host functions to only these types, but can easily be extended by specializing the `get<T>(...)` for the desired type.
-  if the host function wants to access the `Runtime`, it must take `Runtime &` as its first parameter. This simplifies the conversion code (since we don't have to check for each argument if it's a value passed through JSI or the runtime), and seems like an acceptable trade-off.
- the host function cannot access the `thisValue` argument passed from JS. This isn't an issue in our use case, since the called functions are installed as globals.
- the host function cannot access `args` and `count` explicitly. As such, it's impossible to implement host functions which take a non-fixed number of arguments by checking the number of arguments passed from JS.
- the return type of the function must be either `void` (in which case `jsi::Value::undefined()`) or convertible to `jsi::Value`, in which case it is returned.

## Test plan
Check that features depending on the host functions installed in `RuntimeDecorator` (i.e. basically everything) work like before.


Common/cpp/Tools/JsiUtils.h
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// calls `get<First>` on the first argument to retrieve the native type,
+// then calls recursively on the rest of `args`
+// and returns the concatenation of results
+template <typename T, typename... Rest>
+inline std::tuple<T, Rest...>
+convertArgs(jsi::Runtime &rt, const jsi::Value *args, const size_t count) {
+  auto arg = std::tuple<T>(get<T>(rt, args));
+  auto rest = convertArgs<Rest...>(rt, std::next(args), count - 1);
```
graszka22, line 69: Do we need to calculate and pass `count` around here? From what I can read we use `sizeof...(Args) == 0` to stop recursion and we use `count` just for assertion.

graszka22 (commented):
Nice templates 👍 I have just two questions

Common/cpp/Tools/JsiUtils.h
jwajgelt: It's to make sure the correct number of arguments are passed - if you pass too many, it will trip the assert in the `sizeof...(Args) == 0` case.
That said, a better place to do it would be in `getArgsForFunction`, to simply compare `sizeof...(Args)` with the `count` there.

jwajgelt: The reason I did it this way is, I used to accept the `Runtime` argument in arbitrary positions, in which case this function would add `rt` to the tuple and pass `args` and `count` unchanged.
Now I've decided to accept the `Runtime` only in the first argument to the host function, this is never the case.

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -57,25 +70,7 @@ void RuntimeDecorator::decorateRuntime(
...
-      Logger::log("unsupported value type");
-    }
-    return jsi::Value::undefined();
-  };
-  jsi::Value log = jsi::Function::createFromHostFunction(
-      rt, jsi::PropNameID::forAscii(rt, "_log"), 1, callback);
-  rt.global().setProperty(rt, "_log", log);
+  jsi_utils::installJsiFunction(rt, "_log", logValue);
 
   auto chronoNow = [](jsi::Runtime &rt,
```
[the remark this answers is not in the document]
jwajgelt: For `chronoNow`, we also add it as a property on the `performance` object, with a different name.
At best, we could use `createHostFunction` to wrap a parameterless lambda here and assign that to `chronoNow`, but I'm not sure that would be any more readable.

For `evalWithSourceUrl`, that function checks the number of arguments it was passed from JS, and the helpers don't allow for that.
There's also little reason to do that in my opinion, since you can't exactly do much better than pass a pointer and count of elements when passing a variable number of arguments.

Common/cpp/Tools/JsiUtils.h
jwajgelt: Also, my approach here _is_ wrong - if too few arguments were passed, it would happily read uninitialized memory here.
We have to check the `count` at the start (or check that it's >0 at every step).

graszka22 (approved):

PR #4033, opened 2023-02-06
Make sensors work with different device orientations

graszka22:
## Summary

- Adds `adjustToInterfaceOrientation` option that adjusts the sensors values to current device orientation
- Adds `interfaceOrientation` shared value that contains the current interface orientation
- Adds `iosReferenceFrame` option that allows setting ios reference frame. Some reference frames doesn't work for example on iPods but give better results.

## Test plan
- `AnimatedSensorExample.tsx`
- `CubesExample.tsx` - when device faces north the cubes should rotate in the correct way on every device orientation.

### DIFF
--- a/Common/cpp/AnimatedSensor/AnimatedSensorModule.cpp
+++ b/Common/cpp/AnimatedSensor/AnimatedSensorModule.cpp
@@ -53,13 +53,15 @@ jsi::Value AnimatedSensorModule::registerSensor(
           value.setProperty(rt, "yaw", newValues[4]);
           value.setProperty(rt, "pitch", newValues[5]);
           value.setProperty(rt, "roll", newValues[6]);
-          handler.call(rt, value, orientationDegrees);
+          value.setProperty(rt, "interfaceOrientation", orientationDegrees);
+          handler.call(rt, value);
         } else {
           jsi::Object value(rt);
           value.setProperty(rt, "x", newValues[0]);
           value.setProperty(rt, "y", newValues[1]);
           value.setProperty(rt, "z", newValues[2]);
-          handler.call(rt, value, orientationDegrees);
+          value.setProperty(rt, "interfaceOrientation", orientationDegrees);
+          handler.call(rt, value);
         }
       });
   if (sensorId != -1) {
--- a/Common/cpp/AnimatedSensor/AnimatedSensorModule.cpp
+++ b/Common/cpp/AnimatedSensor/AnimatedSensorModule.cpp
@@ -22,6 +22,7 @@ jsi::Value AnimatedSensorModule::registerSensor(
     const std::shared_ptr<JSRuntimeHelper> &runtimeHelper,
     const jsi::Value &sensorTypeValue,
     const jsi::Value &interval,
+    const jsi::Value &iosReferenceFrame,
     const jsi::Value &sensorDataHandler) {
   SensorType sensorType = static_cast<SensorType>(sensorTypeValue.asNumber());
 
@@ -30,10 +31,11 @@ jsi::Value AnimatedSensorModule::registerSensor(
   int sensorId = platformRegisterSensorFunction_(
       sensorType,
       interval.asNumber(),
+      iosReferenceFrame.asNumber(),
       [sensorType,
        shareableHandler,
-       weakRuntimeHelper =
-           std::weak_ptr<JSRuntimeHelper>(runtimeHelper)](double newValues[]) {
+       weakRuntimeHelper = std::weak_ptr<JSRuntimeHelper>(runtimeHelper)](
+          double newValues[], int orientationDegrees) {
         auto runtimeHelper = weakRuntimeHelper.lock();
         if (runtimeHelper == nullptr || runtimeHelper->uiRuntimeDestroyed) {
           return;
@@ -51,13 +53,13 @@ jsi::Value AnimatedSensorModule::registerSensor(
           value.setProperty(rt, "yaw", newValues[4]);
           value.setProperty(rt, "pitch", newValues[5]);
           value.setProperty(rt, "roll", newValues[6]);
-          handler.call(rt, value);
+          handler.call(rt, value, orientationDegrees);
         } else {
           jsi::Object value(rt);
           value.setProperty(rt, "x", newValues[0]);
           value.setProperty(rt, "y", newValues[1]);
           value.setProperty(rt, "z", newValues[2]);
-          handler.call(rt, value);
+          handler.call(rt, value, orientationDegrees);
         }
       });
   if (sensorId != -1) {
--- a/Example/src/CubesExample.tsx
+++ b/Example/src/CubesExample.tsx
@@ -1,0 +1,360 @@
+import React from 'react';
+import Animated, {
+  useAnimatedStyle,
+  useAnimatedSensor,
+  SensorType,
+  IOSReferenceFrame,
+} from 'react-native-reanimated';
+import { View, Button, StyleSheet, Text } from 'react-native';
+
+function eulerToMatrix(pitch: number, roll: number, yaw: number) {
+  'worklet';
+  const s2 = Math.sin(pitch);
+  const c2 = Math.cos(pitch);
+  const s3 = Math.sin(roll);
+  const c3 = Math.cos(roll);
+  const s1 = Math.sin(yaw);
+  const c1 = Math.cos(yaw);
+
+  return [
+    c1 * c3 - s1 * s2 * s3,
+    c3 * s1 + c1 * s2 * s3,
+    -c2 * s3,
+    0,
+
+    -c2 * s1,
+    c1 * c2,
+    s2,
+    0,
+
+    c1 * s3 + c3 * s1 * s2,
+    s1 * s3 - c1 * c3 * s2,
+    c2 * c3,
+    0,
+
+    0,
+    0,
+    0,
+    1,
+  ];
+}
+
+function createIdentityMatrix() {
+  'worklet';
+  return [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1];
+}
+
+function reuseTranslate3dCommand(
+  matrixCommand: number[],
+  x: number,
+  y: number,
+  z: number
+) {
+  'worklet';
+  matrixCommand[12] = x;
+  matrixCommand[13] = y;
+  matrixCommand[14] = z;
+}
+
+function multiplyInto(out: number[], a: number[], b: number[]) {
+  'worklet';
+  const a00 = a[0];
+  const a01 = a[1];
+  const a02 = a[2];
+  const a03 = a[3];
+  const a10 = a[4];
+  const a11 = a[5];
+  const a12 = a[6];
+  const a13 = a[7];
+  const a20 = a[8];
+  const a21 = a[9];
+  const a22 = a[10];
+  const a23 = a[11];
+  const a30 = a[12];
+  const a31 = a[13];
+  const a32 = a[14];
+  const a33 = a[15];
+
+  let b0 = b[0];
+  let b1 = b[1];
+  let b2 = b[2];
+  let b3 = b[3];
+  out[0] = b0 * a00 + b1 * a10 + b2 * a20 + b3 * a30;
+  out[1] = b0 * a01 + b1 * a11 + b2 * a21 + b3 * a31;
+  out[2] = b0 * a02 + b1 * a12 + b2 * a22 + b3 * a32;
+  out[3] = b0 * a03 + b1 * a13 + b2 * a23 + b3 * a33;
+
+  b0 = b[4];
+  b1 = b[5];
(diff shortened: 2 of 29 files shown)


Example/src/CubesExample.tsx
```diff
@@ -0,0 +1,360 @@
...
+    0,
+    0,
+    0,
+    1,
+  ];
+}
+
+function eulerToQuaternion(pitch: number, roll: number, yaw: number) {
+  'worklet';
+  const c1 = Math.cos(pitch * 0.5);
```
tomekzaw, line 156: why not `/ 2`?

ios/sensor/ReanimatedSensor.m
```diff
@@ -49,7 +53,7 @@ - (bool)initializeGyroscope
                       return;
                     }
                     double data[] = {sensorData.rotationRate.x, sensorData.rotationRate.y, sensorData.rotationRate.z};
-                    self->_setter(data);
+                    self->_setter(data, [self getInterfaceOrientation]);
```
tomekzaw, line 56: How expensive is to call `[self getInterfaceOrientation]`? Should we memoize the value within one frame? Is there any way to keep it local and register some listener to change the value once it changes?

ios/sensor/ReanimatedSensor.m
```diff
@@ -138,7 +142,7 @@ - (bool)initializeOrientation
   [_motionManager setDeviceMotionUpdateInterval:_interval];
 
   [_motionManager setShowsDeviceMovementDisplay:YES];
-  [_motionManager startDeviceMotionUpdatesUsingReferenceFrame:CMAttitudeReferenceFrameXArbitraryZVertical
+  [_motionManager startDeviceMotionUpdatesUsingReferenceFrame:(1 << _referenceFrame)
```
tomekzaw, line 145: `(1 << _referenceFrame)` looks quite hacky, can you please explain it better in the comment?

Example/src/CubesExample.tsx
```diff
@@ -0,0 +1,360 @@
...
+  const a12 = a[6];
+  const a13 = a[7];
+  const a20 = a[8];
+  const a21 = a[9];
+  const a22 = a[10];
+  const a23 = a[11];
+  const a30 = a[12];
+  const a31 = a[13];
+  const a32 = a[14];
+  const a33 = a[15];
```
tomekzaw, line 76:
```suggestion
  const [a00, a01, ..., a33] = a;
```

Example/src/CubesExample.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+  let b0 = b[0];
+  let b1 = b[1];
+  let b2 = b[2];
+  let b3 = b[3];
```
tomekzaw, line 81:
```suggestion
  let [b0, b1, b2, b3] = b;
```

Example/src/CubesExample.tsx
```diff
@@ -0,0 +1,360 @@
...
+export default function CubesExample() {
+  const animatedSensor = useAnimatedSensor(SensorType.ROTATION, {
+    adjustToInterfaceOrientation: true,
+  });
+
+  const compassStyle = useAnimatedStyle(() => {
+    const yaw = animatedSensor.sensor.value.yaw;
+
+    return {
+      transform: [{ rotate: yaw + 'rad' }],
```
tomekzaw, line 288:
```suggestion
      transform: [{ rotate: `${yaw}rad` }],
```

src/reanimated2/hook/useAnimatedSensor.ts
```diff
@@ -52,18 +124,42 @@ export function useAnimatedSensor(
...
+      iosReferenceFrame: IOSReferenceFrame.XArbitraryCorrectedZVertical,
     },
+    interfaceOrientation: makeMutable<InterfaceOrientation>(0),
   });
 
   useEffect(() => {
-    ref.current.config = { interval: 'auto', ...userConfig };
+    ref.current.config = {
+      interval: 'auto',
+      adjustToInterfaceOrientation: false,
```
tomekzaw, line 136: Is there any reason that prevents us from enabling this by default?
```suggestion
      adjustToInterfaceOrientation: true,
```

src/reanimated2/hook/useAnimatedSensor.ts
```diff
@@ -52,18 +124,42 @@ export function useAnimatedSensor(
...
+      iosReferenceFrame: IOSReferenceFrame.XArbitraryCorrectedZVertical,
```
tomekzaw, line 137: Can we also introduce option `'best'` which uses the best possible sensor on iPhone/iPod?

src/reanimated2/hook/useAnimatedSensor.ts
```diff
@@ -52,18 +124,42 @@ export function useAnimatedSensor(
...
     const sensorData = ref.current.sensor!;
     const id = registerSensor(
       sensorType,
       ref.current.config.interval === 'auto' ? -1 : ref.current.config.interval,
-      (data) => {
+      ref.current.config.iosReferenceFrame,
+      (data, orientationDegrees) => {
         'worklet';
+        if (ref.current.config.adjustToInterfaceOrientation) {
+          if (sensorType === SensorType.ROTATION) {
```
tomekzaw, line 148: @piaskowyk doesn't like such conditionals inside worklets. Instead, you can move the if outside of worklet and assign the appropriate variant of the worklet.

src/reanimated2/hook/useAnimatedSensor.ts
```diff
@@ -52,18 +124,42 @@ export function useAnimatedSensor(
     isAvailable: false,
     config: {
       interval: 0,
+      adjustToInterfaceOrientation: false,
+      iosReferenceFrame: IOSReferenceFrame.XArbitraryCorrectedZVertical,
     },
+    interfaceOrientation: makeMutable<InterfaceOrientation>(0),
```
tomekzaw, line 130: Should we use `makeMutable` here or `makeShareable`?

android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensorListener.java
```diff
@@ -39,13 +61,15 @@ public void onSensorChanged(SensorEvent event) {
             quaternion[2], // qy
             quaternion[3], // qz
             quaternion[0], // qw
-            orientation[0], // yaw
-            orientation[1], // pitch
+            // make android consistent with iOS - iOS is better documented here:
```
tomekzaw, line 64:
```suggestion
            // make Android consistent with iOS, which is better documented here:
```

Example/src/CubesExample.tsx
```diff
@@ -0,0 +1,360 @@
...
+      const sideLength = 100;
+      const origin = { x: 0, y: 0, z: -sideLength / 2 };
+
+      const q0 = eulerToQuaternion(-rotation[0], -rotation[1], rotation[2]);
+
+      const q1 = [
+        -animatedSensor.sensor.value.qx,
+        animatedSensor.sensor.value.qy,
+        -animatedSensor.sensor.value.qz,
+        animatedSensor.sensor.value.qw,
```
tomekzaw, line 253: Let's assign these four values beforehand, i.e.
```js
const [qx, qy, qz, qw] = animatedSensor.sensor.value;
const q1 = [-qx, qy, -qz, qw];
```
in order to avoid calling `.value` more than once if not necessary.

Common/cpp/AnimatedSensor/AnimatedSensorModule.cpp
```diff
@@ -51,13 +53,13 @@ jsi::Value AnimatedSensorModule::registerSensor(
           value.setProperty(rt, "yaw", newValues[4]);
           value.setProperty(rt, "pitch", newValues[5]);
           value.setProperty(rt, "roll", newValues[6]);
-          handler.call(rt, value);
+          handler.call(rt, value, orientationDegrees);
```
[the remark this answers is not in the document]
graszka22: Ok, I can do this but what about the user's API? Should I make the orientation as separate shared value or put it in the `Value3D` and `ValueRotation`? I opt for separate shared value, because interface orientation is not sensor data. And if so, should I create a new ts type here just for data here?
```js
 (data, orientationDegrees) => {
        'worklet';
        if (ref.current.config.adjustToInterfaceOrientation) {
          if (sensorType === SensorType.ROTATION) {
            data = adjustRotationToInterfaceOrientation(
              data as ValueRotation,
              orientationDegrees
            );
```
will change into:
```js
(data) => {
        'worklet';
        if (ref.current.config.adjustToInterfaceOrientation) {
          if (sensorType === SensorType.ROTATION) {
            data = adjustRotationToInterfaceOrientation(
              data as ValueRotation+Orientation, //?
            );
```
now it doesn't look simpler unless it's just for optimisation.

ios/sensor/ReanimatedSensor.m
graszka22: There are two ways to use the listener:
- add listener for `UIDeviceOrientationDidChangeNotification` notification, but this is *device* orientation, not *interface* orientation. Device orientation can be changed without changing interface orientation (put phone on flat surface or turn the phone with orientation locked). Or vice-versa (have phone in landscape orientation and lock orientation, it will turn into portrait without moving the device). This approach will be bugged on the second case.
- override `viewWillTransitionToSize` method on some view controller but we don't have any view controller here so I haven't even tested that.

Memoize the value within one frame? We're calling `getInterfaceOrientation` only once per sensor.

As for how expensive this call - it looks like it's just reading a property but who knows, it's not open source 🤷‍♀️ I can try checking in the profiler if you insist.

src/reanimated2/hook/useAnimatedSensor.ts
graszka22: Yes, though I'm not sure myself. Whether we should adjust or not and how to do it depends on the use case. On the other hand this should work for most cases, so I'll change it.

src/reanimated2/hook/useAnimatedSensor.ts
graszka22: Which one is "best"? :D https://developer.apple.com/documentation/coremotion/cmattitudereferenceframe?language=objc
The first one works on iPod, the rest not, but give better results.

src/reanimated2/hook/useAnimatedSensor.ts
graszka22: Yeah, sounds like premature optimization to me.

src/reanimated2/hook/useAnimatedSensor.ts
graszka22: What's the difference between those two? Is it documented somewhere? I used it just because sensors used it already and orientation is very similar.

ios/sensor/ReanimatedSensor.m
tomekzaw: Thanks for clarifying. Since we call it once per sensor and usually there's just one sensor, we don't really need to memoize it. Let's assume that it's blazingly fast, no need to profile it.

src/reanimated2/hook/useAnimatedSensor.ts
tomekzaw: Yeah, let's make it beginner-friendly out-of-the-box.

src/reanimated2/hook/useAnimatedSensor.ts
tomekzaw: Our aim here is to provide the best possible behavior out-of-the-box. React Native developers just want a simple hook to access sensor measurements without deep-diving into CoreMotion documentation. I see that the default value here is `XArbitraryCorrectedZVertical`. The question is, is it really the best possible setting across all kinds of devices? I think you've mentioned that a sensor that works the best on iPhone is not available on iPod or something similar. The idea here is to add another enum value `IOSReferenceFrame.BestAvailable` that represents sensor A for iPods and sensor B for iPhones on the native side.

src/reanimated2/hook/useAnimatedSensor.ts
tomekzaw: Unfortunately, it's not documented as it's not in the public API. If you've copied it from a similar place, then that's okay.

Common/cpp/AnimatedSensor/AnimatedSensorModule.cpp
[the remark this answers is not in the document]
tomekzaw: Yeah, generally we try to optimize the number of shared values, but in case of `useAnimatedSensor` it doesn't really make that much difference as we expect our users to call it only once within a component. Let's ask @piaskowyk who should know better.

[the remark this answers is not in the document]
graszka22: I asked him and he agreed that we should put orientation as property in sensor data shared value

tomekzaw (approved):
@graszka22 Thank you for this PR. Well done! 🥳

PR #4038, opened 2023-02-07
[Shared Element Transition] Fix exiting animations on Android

piaskowyk:
## Summary

I shouldn't clear the layout animation config when the shared element transition doesn't start. Without this check, the exiting layout animation couldn't start because LayoutAnimationManager doesn't contain the view config.

## Test plan

Run exiting layout animation on Android.

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
@@ -49,7 +49,7 @@ public class SharedTransitionManager {
   }
 
   protected void viewsDidLayout() {
-    startSharedTransitionForViews(mAddedSharedViews, true);
+    tryStartSharedTransitionForViews(mAddedSharedViews, true);
     mAddedSharedViews.clear();
   }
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -85,19 +88,20 @@ protected void setNativeMethods(NativeMethodsHolder nativeMethods) {
     mNativeMethodsHolder = nativeMethods;
   }
 
-  private void startSharedTransitionForViews(List<View> sharedViews, boolean withNewElements) {
+  private boolean startSharedTransitionForViews(List<View> sharedViews, boolean withNewElements) {
```
tomekzaw, line 91:
```suggestion
  private boolean tryStartSharedTransitionForViews(List<View> sharedViews, boolean withNewElements) {
```
To better emphasise the fact that animation may not be started, let's prepend the name with `try` or `maybe`.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -57,7 +57,10 @@ protected void onViewsRemoval(int[] tagsToDelete) {
     if (tagsToDelete != null) {
       visitTreeForTags(tagsToDelete, new SnapshotTreeVisitor());
 
-      startSharedTransitionForViews(mRemovedSharedViews, false);
+      boolean startedAnimation = startSharedTransitionForViews(mRemovedSharedViews, false);
```
tomekzaw, line 60: Sounds good, let's consider `animationStarted` or `didStartAnimation` as well.

tomekzaw (approved):

PR #4042, opened 2023-02-08
Refactor NativeProxy on Android

graszka22:
## Summary

Refactor of `NativeProxy.java` to avoid some code duplication.

## Test plan

Run Example app and FabricExample app. Make sure things like keyboard, sensors, layout animations still work.

### DIFF
--- a/android/src/fabric/java/com/swmansion/reanimated/NativeProxy.java
+++ b/android/src/fabric/java/com/swmansion/reanimated/NativeProxy.java
@@ -66,6 +66,7 @@ public class NativeProxy extends NativeProxyCommon {
         mHybridData.resetNative();
     }
 
+    // TODO: move to NativeProxyCommon.java when layout animations are fixed on Fabric 
     public void prepareLayoutAnimations(LayoutAnimations LayoutAnimations) {
         if (Utils.isChromeDebugger) {
             Log.w("[REANIMATED]", "You can not use LayoutAnimation with enabled Chrome Debugger");
--- a/android/src/fabric/java/com/swmansion/reanimated/NativeProxy.java
+++ b/android/src/fabric/java/com/swmansion/reanimated/NativeProxy.java
@@ -66,44 +66,28 @@ public class NativeProxy extends NativeProxyCommon {
         mHybridData.resetNative();
     }
 
-    // TODO: move to NativeProxyCommon.java when layout animations are fixed on Fabric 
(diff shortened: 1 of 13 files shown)


android/src/fabric/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -1,333 +1,109 @@
...
-
-  private Set<String> convertProps(ReadableNativeArray props) {
-    Set<String> propsSet = new HashSet<>();
-    ArrayList<Object> propsList = props.toArrayList();
-    for (int i = 0; i < propsList.size(); i++) {
-      propsSet.add((String) propsList.get(i));
+    public void onCatalystInstanceDestroy() {
+        mScheduler.deactivate();
+        mHybridData.resetNative();
     }
```
tomekzaw, line 67: Is it possible to move this method to `NativeProxyCommon.java`?

android/src/fabric/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -1,333 +1,109 @@
...
+
+                    @Override
+                    public boolean hasAnimation(int tag, String type) {
+                        return false;
+                    }
+
+                    @Override
+                    public void clearAnimationConfig(int tag) {}
+                });
     }
```
tomekzaw, line 108: I think the only difference between these two methods is the object passed to `animationsManager.setNativeMethods` call.

Can we add a new method `createNativeMethodsHolder` that returns this complex object, separately for Paper and Fabric, and move `prepareLayoutAnimations` to `NativeProxyCommon.java`?

android/src/fabric/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -1,333 +1,109 @@
...
-    } else {
-      return SystemClock.uptimeMillis();
-    }
-  }
-
-  @DoNotStrip
-  private float[] measure(int viewTag) {
-    return mNodesManager.measure(viewTag);
-  }
+    public native void performOperations();
```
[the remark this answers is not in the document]
tomekzaw: Yeah you're right, let's keep it simple. Thanks!

tomekzaw (approved):
Tested on Example and FabricExample (Android). Well done @graszka22!

PR #4044, opened 2023-02-10
Fix missing dependency on `React-hermes` on older RN versions

jwajgelt:

### DIFF
--- a/RNReanimated.podspec
+++ b/RNReanimated.podspec
@@ -71,6 +71,7 @@ Pod::Spec.new do |s|
   s.dependency 'React-Core/DevSupport'
   if !config[:is_tvos_target]
     s.dependency 'React-RCTActionSheet'
+    s.dependency "React-hermes"
   end
   s.dependency 'React-RCTNetwork'
   s.dependency 'React-RCTAnimation'
@@ -84,7 +85,6 @@ Pod::Spec.new do |s|
   s.dependency 'React-jsi'
   s.dependency 'React-jsiexecutor'
   s.dependency 'React-jsinspector'
-  s.dependency "React-hermes"
   s.dependency 'Yoga'
   s.dependency 'DoubleConversion'
   s.dependency 'glog'
--- a/RNReanimated.podspec
+++ b/RNReanimated.podspec
@@ -84,6 +84,7 @@ Pod::Spec.new do |s|
   s.dependency 'React-jsi'
   s.dependency 'React-jsiexecutor'
   s.dependency 'React-jsinspector'
+  s.dependency "React-hermes"
   s.dependency 'Yoga'
   s.dependency 'DoubleConversion'
   s.dependency 'glog'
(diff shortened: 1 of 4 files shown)


RNReanimated.podspec
```diff
@@ -71,6 +71,7 @@ Pod::Spec.new do |s|
   s.dependency 'React-Core/DevSupport'
   if !config[:is_tvos_target]
     s.dependency 'React-RCTActionSheet'
+    s.dependency "React-hermes"
```
tomekzaw, line 74: What if someone uses JSC? We shouldn't declare a dependency on `React-hermes` in such case.

RNReanimated.podspec
```diff
@@ -14,6 +14,7 @@ folly_compiler_flags = folly_flags + ' ' + '-Wno-comma -Wno-shorten-64-to-32'
 boost_compiler_flags = '-Wno-documentation'
 fabric_flags = fabric_enabled ? '-DRCT_NEW_ARCH_ENABLED' : ''
 version_flag = '-DREANIMATED_VERSION=' + reanimated_package_json["version"]
+use_hermes = ENV['USE_HERMES'] == '1'
```
tomekzaw, line 17: Another option is `hermes_enabled` which would be consistent with already existing `fabric_enabled` flag.

piaskowyk (approved):
Good for me 👍 But I ask you to add in PR description some information how you tested if it works and which RN versions did you use for testing.

RNReanimated.podspec
```diff
@@ -87,6 +88,9 @@ Pod::Spec.new do |s|
   s.dependency 'Yoga'
   s.dependency 'DoubleConversion'
   s.dependency 'glog'
+  if config[:react_native_minor_version] >= 65 && use_hermes
```
tomekzaw, line 91: Can you please remind me why `config[:react_native_minor_version] >= 65` is necessary here?

jwajgelt: Becuase it doesn't build otherwise 🤷 
It can't find the `React-hermes` pod in that case (the pod was called `React-Core/Hermes` back then).

tomekzaw: What about this?
```rb
  if use_hermes
    if config[:react_native_minor_version] >= 65
      s.dependency "React-hermes"
    else
      s.dependency "React-Core/Hermes"
    end
  end
```

jwajgelt:
> Good for me 👍 But I ask you to add in PR description some information how you tested if it works and which RN versions did you use for testing.

I just run it through @tomekzaw's e2e test suite, on the whole version list.
I generally assume that if it builds and runs, there shouldn't be any changes in behaviour, but the tests pass locally for me as well.

RNReanimated.podspec
jwajgelt: It builds fine on 0.64 without this dependency, so I'd leave it as is :+1:

tomekzaw (approved):
Thanks @jwajgelt for this PR! 👏

PR #4046, opened 2023-02-12
fix `Error` Variable scope conflicts, and remove duplicate `Date`

jinliming2:
## Summary

The babel plugin transformed the code and added some helper code, it uses `new Error()` to get the call stack. this works in general unless you have a variable called `Error` in the local scope, in this case, It references the wrong Object. this PR uses `global.Error` to avoid this.

`'Date'` is duplicated in this `global` array, and I removed the second one.

## Test plan

```jsx
import { useState } from 'react';
import Animated, { useAnimatedStyle, useSharedValue } from 'react-native-reanimated';
import Error from './components/Error'; // imported an arrow function component named 'Error'

const App = () => {
  const [hasError, setHasError] = useState(false);
  const offset = useSharedValue(0);
  // `new Error()` not working here, because `Error` reference to the function component, it's a arrow function, not a function or class
  const animatedStyles = useAnimatedStyle(() => ({ transform: [{ translateX: offset.value * 255 }] }), []);

  if (hasError) return <Error />;
  return <Animated.View style={animatedStyles} />;
};
```

### DIFF
--- a/__tests__/__snapshots__/plugin.test.js.snap
+++ b/__tests__/__snapshots__/plugin.test.js.snap
@@ -10,7 +10,7 @@ var _worklet_5359970077727_init_data = {
   location: \\"${ process.cwd() }/jest tests fixture\\"
 };
 var f = function () {
-  var _e = [new Error(), -3, -20];
+  var _e = [new global.Error(), -3, -20];
   var _f = function _f() {
     return {
       res: x + objX.x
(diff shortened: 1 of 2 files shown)


tomekzaw:
In review by @tjzel

tjzel:
Hi, sorry for late comment and thanks for your PR. We are in the middle of a slight refactor of `plugin` and have some issues with correctly pointing to error source in the code on JSC. Since your PR potentially affects this aspect (might need to change column offset for the `Error` will be longer now) I will merge it after that refactor. I have already tested your change and it seems to be working fine (not breaking anything) but I would want to resolve current pointing issue before merging it.

tjzel (commented):
correct new position of opening bracket

tjzel:
I just broke something trying to add a change, hold on

tjzel (approved):

PR #4047, opened 2023-02-13
[Shared Element Transition] Disable animation on swipe back

piaskowyk:
## Summary

This PR prevents starting shared transition at screen swipe back on iOS. In the future, for this use case, we want to add a new type of shared animation controlled by transition progress.


tomekzaw (approved):

izakfilmalter:
Is there an issue to track the long term fix for this?

PR #4048, opened 2023-02-13
Call scrollTo function when scrollViewOffset changed

mstach60161:
## Summary

When position of the Scroll View is changed, `scrollViewOffset` value is updated.
In this PR we would like to add opposite - when `scrollViewOffset` value is changed, we move the Scroll View to appropriate position.

- Added listener in `useScrollViewOffset`
- Added flag to `sharedValue` object to prevent scrolling in the loop

## Test plan

Testing on example screen
- Animated scroll
- Static scroll
- Cancel animations
- Old functionality


Example/src/ScrollViewOffsetCallScrollToExample.tsx
```diff
@@ -0,0 +1,67 @@
...
+        />
+        <Button
+          title="Scroll with animation"
+          onPress={onAnimatedChangeScrollValue}
+        />
+      </View>
+      <View style={styles.divider} />
+      <Animated.ScrollView
+        ref={aref}
+        scrollEventThrottle={16}
```
kmagiera, line 38: is it necessary for the throttle parameter to be present? Typically with nativedriver / reanimated you want throttle to be set to 1 because all events are processed on the UI thread anyways

src/reanimated2/hook/useScrollViewOffset.ts
```diff
@@ -35,3 +38,31 @@ export function useScrollViewOffset(
...
+  const scrollSetter = (sv: any, newValue: any) => {
+    'worklet';
+    scrollValueSetter(sv, newValue, aref);
+  };
+  const ref = useRef<SharedValue<T>>(
+    makeMutable(init, oneWayReadsOnly, scrollSetter)
+  );
+
+  if (ref.current === null) {
+    ref.current = makeMutable(init, oneWayReadsOnly, scrollSetter);
```
kmagiera, line 56: I don't think you need to `makeMutable` here twice

src/reanimated2/scrollValueSetter.ts
```diff
@@ -0,0 +1,64 @@
+import { AnimationObject } from './commonTypes';
+import { scrollTo } from './NativeMethods';
+export { stopMapper } from './mappers';
+
+export function scrollValueSetter(sv: any, value: any, animatedRef: any): void {
```
kmagiera, line 5: This seems to be copied to some extent from the default value setter. I wonder if this can just be done using mutable's listeners?

kmagiera (commented):
One thing that I wonder (and posted short comment inline about this somewhere), is if this could be done using mutable's value listeners instead of overriding the setter. In that scenario we'd just use a standard shared value after which we schedule runOnUI call that sets up the listener (you can do SV.addListener(clb) that gets called when the value is updated). If that's possible it'd require less duplication, no modification of the core elements, and less code overall.

Secondly, I find this API a bit cumbersome: you need to create animated ref, then use that ref with one another hook, and then pass ref to Animated.ScrollView. I'd like to see a better example with this being used actually. The example you provided only updates the position but never reads from it. For this scenario one can just use synchronous scrollTo methods with animated-ref. Once we have a good example I'd be happy to provide better feedback on the API itself

src/reanimated2/scrollValueSetter.ts
mstach60161: Initially I did it with listener, but it cause a cycle or unnecessary calls.
We want to call `scrollTo` method when:
- public constructor is called. 
- there is a next iteration of the animation

With a listener approach we also call `scrollTo` when:
- the value is changed because of the previous `scrollTo`.
- the user drags `ScrollView`

The alternative approach can be adding 3rd setter to differentiate event assignment, but I think that my solution is more clear.

Maybe I don't see the better solution, if you have any suggestions, I'm happy to hear it.

src/reanimated2/hook/useScrollViewOffset.ts
```diff
@@ -15,23 +21,49 @@ const scrollEventNames = [
...
 ];
 
+const addListenerToScroll = (
+  offsetRef: ScrollSharedValue<number>,
+  animatedRef: any
+) => {
+  runOnUI(() => {
+    'worklet';
+    offsetRef.triggerScrollListener = true;
+    offsetRef.addListener(animatedRef(), (newValue: any) => {
```
kmagiera, line 31: Would be good to make sure we setup the listener only once. This code will be executed on every render for components using `useScrollViewOffset` hook. There are already some useEffect calls in that hook, so maybe we should put this initialization code there?

src/reanimated2/hook/useScrollViewOffset.ts
```diff
@@ -15,23 +21,49 @@ const scrollEventNames = [
...
+      }
+    });
+  })();
+};
+
 export function useScrollViewOffset(
   aref: RefObject<Animated.ScrollView>
 ): SharedValue<number> {
-  const offsetRef = useRef(useSharedValue(0));
+  const offsetRef: ScrollSharedValue<number> = useRef(
```
kmagiera, line 42: I know this was here before but i wonder why is useSharedValue wrapped with useRef? This does not look like a typicaly pattern and if it in fact is necessary than it'd be good to add some comment as to why is it designed this way. I think just using `useSharedValue` here should suffice

kmagiera (commented):
As for the implementation I don't have more comments at this point. I think it is ok with just one minor comment I added for a code that's been there before (maybe we could change it here too).

I do have some comments, however, related to the API design of the feature. I commented earlier that I find it a bit too cumbersome and would like to if you could try to come up with some alternative approach. Because scroll view offset will only be supported with the wrapped ScrollView (Animated.ScrollView) we can potentially extend the parameter list and add some new parameter to facilitate this feature as opposed to just relying on refs which just doesn't feel super convinient for this usecase.

PR #4053, opened 2023-02-15
fix: Fix nested Worklets not transpiling properly

mrousavy:
## Summary

The latest Babel plugin does not transpile nested worklets properly.

For example:

```ts
function outer() {
  'worklet'
  const inner = () => {
    'worklet'
    return 5
  }
  inner()
}
```

Does not work. 

This PR fixes that by reversing the order in which Babel executes (`enter` -> `exit`)

Credits to @Szymon20000 & @chrfalch for figuring this one out, just wanted to push this upstream.

### DIFF
--- a/plugin.js
+++ b/plugin.js
@@ -864,12 +864,12 @@ module.exports = function ({ types: t }) {
     },
     visitor: {
       CallExpression: {
-        enter(path, state) {
+        exit(path, state) {
           processWorklets(t, path, state);
         },
       },
       'FunctionDeclaration|FunctionExpression|ArrowFunctionExpression': {
-        enter(path, state) {
+        exit(path, state) {
           processIfWorkletNode(t, path, state);
           processIfGestureHandlerEventCallbackFunctionNode(t, path, state);
         },


chrfalch:
Could it be that you need an additional line in the example? The issue was not when we tried to run the inner worklet as a plain function, but when trying to run it as a worklet:

```ts

import React from 'react';

  const handlePress = () => {
    function outer() {
      'worklet'
      const inner = () => {
        'worklet'
        return 5
      };
      console.log(_WORKLET) // should be true
      const result = runOnUI(inner)() // <-  This is where we had an issue I think
      console.log(result) // should be 5
    }
    runOnUI(outer)()
  };

  return (
    <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
    </View>
  );
}
```

mrousavy:
@chrfalch Ah yeah, that was it.

@tomekzaw The `inner` function does not have `__initData`, so it can't be passed to a Worklet invoker (e.g. `runOnUI`, or my VisionCamera Frame Processor).

Let me know if you have any questions

mrousavy:
You can see this in the Snapshot tests, with my PR it now has the `__initData` prop:

image

I don't think this is a breaking change either way, right?

tomekzaw:
@chrfalch @mrousavy Thanks for your comments.

> The inner function does not have __initData, so it can't be passed to a Worklet invoker (e.g. runOnUI, or my VisionCamera Frame Processor).

Yeah, it makes sense. I'm not sure if we support calling `runOnUI` from inside a worklet on the UI thread, in such case it should work like an identity function (i.e. return the argument). I'm also thinking about the following workaround:
```ts
if (_WORKLET) {
  func();
} else {
  runOnUI(func)();
}
```

> You can see this in the Snapshot tests, with my PR it now transpiles the inner code properly:

Looks about right, but we'll need a separate unit test just for transforming nested worklets.

> I don't think this is a breaking change either way, right?

Nope, if it works for you that's great. We're just trying to better understand the motivation of this change and make sure it doesn't break other stuff.

mrousavy:
> Nope, if it works for you that's great. We're just trying to better understand the motivation of this change and make sure it doesn't break other stuff.

Gotcha - yes I am working on a feature in VisionCamera where you can hop from one Thread to another:


```ts
const frameProcessor = useFrameProcessor((frame) => {
  'worklet'
  console.log("thread 1 hi")
  runAsync(() => {
    'worklet'
    console.log("thread 2 hi!")
  })
}, [])
```

And in this case, the inner function doesn't have proper `__initData` / `code` so I can't dispatch it to another Thread.

---

Do you want me to create that Test to verify the inner function is transpiled correctly, or is that something you guys want to create internally? :)

tomekzaw:
@mrousavy Yeah it makes sense, thanks for the context.

> Do you want me to create that Test to verify the inner function is transpiled correctly, or is that something you guys want to create internally? :)

Would be great if you could add one simple unit test the PR, this way we can merge it more confidently.

mrousavy:
@tomekzaw Sure! Added a new test for the nested worklet, and updated all other snapshot tests since the hash changes! :)

__tests__/__snapshots__/plugin.test.js.snap
```diff
@@ -121,9 +121,9 @@ exports[`babel plugin supports SequenceExpression, with objectHook 1`] = `
 function App() {
   (0, useAnimatedGestureHandler)({
     onStart: function () {
-      var _e = [new Error(), 1, -20];
+      const _e = [new Error(), 1, -20];
```
tomekzaw, line 124: Previously we had `var` but now it's `const`, is this expected and are we sure it doesn't break anything?

__tests__/__snapshots__/plugin.test.js.snap
```diff
@@ -121,9 +121,9 @@ exports[`babel plugin supports SequenceExpression, with objectHook 1`] = `
...
 
-      var _f = function _f() {};
+      const _f = function () {};
```
tomekzaw, line 126: Previously we had `function _f() {};` but after the change the function has no name. I guess that's okay since `_f` was hard-coded in each worklet and thus gives no additional information that would help to identify them.

__tests__/__snapshots__/plugin.test.js.snap
```diff
@@ -243,7 +243,7 @@ function Box() {
     };
 
     _f._closure = {
-      offset: offset
+      offset
```
tomekzaw, line 246: Are we sure this change doesn't break anything? I'm trying to understand why sometimes there's `x: x` in `_f._closure` but sometimes just `offset` and not `offset: offset`.

__tests__/__snapshots__/plugin.test.js.snap
```diff
@@ -261,9 +261,54 @@ function Box() {
...
+  _f.__stackDetails = _e;
+  return _f;
+}();"
+`;
+
 exports[`babel plugin transforms spread operator in worklets for arrays 1`] = `
-"var _worklet_3161057533258_init_data = {
-  code: \\"function foo(){const bar=[4,5];const baz=[1,...[2,3],...bar];}\\",
+"var _worklet_8509370056349_init_data = {
+  code: \\"function foo(){var bar=[4,5];var baz=[1].concat([2,3],bar);}\\",
```
tomekzaw, line 311: When I was working on #3143 I checked that it's safe to use spread operator in the worklet code. I assume that for some reason previously spread operator in worklets wasn't getting transformed but after the changes in this PR now it is. Do we know how this change affects the performance and memory usage?

__tests__/__snapshots__/plugin.test.js.snap
```diff
@@ -571,21 +618,21 @@ var foo = function () {
 `;
 
 exports[`babel plugin workletizes object hook wrapped ArrowFunctionExpression automatically 1`] = `
-"var _worklet_1022605193782_init_data = {
-  code: \\"function anonymous(event){console.log(event);}\\",
+"var _worklet_338158776260_init_data = {
+  code: \\"function onStart(event){console.log(event);}\\",
```
tomekzaw, line 622: That's actually a really good change.

tomekzaw:
Thanks @mrousavy for the commit. Actually there's way more changes to the snapshot tests than just worklet hashes, so we'll need to verify those.

chrfalch:
> Thanks @mrousavy for the commit. Actually there's way more changes to the snapshot tests than just worklet hashes, so we'll need to verify those.

Could this be caused by running the snapshot creation with a newer version of some library? I've seen the same thing on my side sometimes. 

It doesn't look like the causes of the changes should be the source code change in the plugin - maybe we could test and run and just revert the enter -> exit change to see if this also reverts the snapshot - or if it generate a changed snapshot file?

kmagiera:
The reason why there are so many changes in snapshot just lies in that change made to the plugin code. After changing enter to exit we now scan a transpiled version of the code, hence we can see things like const being transpiled to var etc. We deliberately were using enter as some bits of the JS syntax used to be transpiled into polyfills which added dependencies on functions from babel runtimes. In that scenario we'd be calling some polyfill methods from babel runtime which obviously were not worklets. I know that there's been a lot of advancement in JS runtime compatibility since we did this some time ago and also the babel config was changed such that it would not transpile some bits of functionality, however this needs to be verified if we were to go with this change. Some examples of things that require polyfills in the past are data structures (e.g., Map, Set, etc), JS proxies, object spread. On top of that it'd be important to verify if source maps work properly after this change. It may affect this part as worklet generation now operates on a transpiled tree and not the original one.

I wonder if there is some less invasive approach to fixing the nested worklet issue? A bit concerned that this change may have some unwanted consequences especially for setups with custom babel configuraiton or on older RN versions where the config was different.

chrfalch:
That's a very good point, @kmagiera. Do you have an idea on how we could solve the inner worklet issue without making such a big change?

kmagiera:
Haven't tried that but possibly executing the plugin recursively would work? I.e. we could call traverse again from `processWorklet` method?

kmagiera:
I'm not completely against this change to be clear. Especially if any other approach is significantly more complex. Just wanted to point out things that needs to be tested/verified

chrfalch:
But you have a very good point - maybe we can look into a solution where we call processWorklet recursively as you suggest, definitely makes a lot of sense not to change the behaviour too much.

mrousavy:
tbh I don't have a lot of experience with writing babel plugins, I don't really know how I would attempt to process everything recursively. 

The PR as it is here works fine in my app with a few worklets (incl shared values, animated props, gesture handlers) - but yea the snapshot tests look different. 

How would I measure performance between `...` and `concat` here?

kmagiera:
I'm not concerned about performance between polyfilled/non-polyfilled versions. What I'm concerned about is that some polyfills may add dependencies on libraries that won't work in the worklet runtime because they call plain (non-worklet) JS functions. I know that some polyfills add dependencies on methods from babel-runtime and similar libraries.

kmagiera:
@tjzel is in a process of rewriting some bits of the plugin (main goal is to move the source code to typescript and split up the sources to smaller chunks/files to improve maintainance). I think he's looking into addressing this issue as a part of these efforts

tjzel:
@mrousavy I have done some testing and unfortunately using `exit` instead of `enter` breaks our code and as simple mock app as this one:

```TSX
import React from 'react';
import { runOnJS, runOnUI } from 'react-native-reanimated';

function sometimesImCalledOnJSAndSometimesOnUI() {
  'worklet';
  console.log('Hello from outer', !_WORKLET ? 'JS' : 'UI');
  function imCalledOnJSTooIfPossible() {
    'worklet';
    console.log('Hello from inner', !_WORKLET ? 'JS' : 'UI');
  }
  runOnUI(imCalledOnJSTooIfPossible)();
  if (!_WORKLET) {
    runOnJS(imCalledOnJSTooIfPossible)();
  }
}

function App() {
  runOnJS(sometimesImCalledOnJSAndSometimesOnUI)();
  runOnUI(sometimesImCalledOnJSAndSometimesOnUI)();
}

export default App;
```

Fails to log anything when using `exit` logic. However, I've done some digging and made this experimental change:

```diff
diff --git a/plugin/src/processIfWorkletNode.ts b/plugin/src/processIfWorkletNode.ts
--- a/plugin/src/processIfWorkletNode.ts
+++ b/plugin/src/processIfWorkletNode.ts
@@ -15,32 +18,45 @@ function processIfWorkletNode(
   >,
   state: ReanimatedPluginPass
 ): void {
+  let shouldBeProcessed = false;
   fun.traverse({
     DirectiveLiteral(path) {
       const value = path.node.value;
-      if (
-        value === 'worklet' &&
-        path.getFunctionParent() === fun &&
-        isBlockStatement(fun.node.body)
-      ) {
-        // make sure "worklet" is listed among directives for the fun
-        // this is necessary as because of some bug, babel will attempt to
-        // process replaced function if it is nested inside another function
-        const directives = fun.node.body.directives;
-        if (
-          directives &&
-          directives.length > 0 &&
-          directives.some(
-            (directive) =>
-              isDirectiveLiteral(directive.value) &&
-              directive.value.value === 'worklet'
-          )
+      if (value === 'worklet' && isBlockStatement(fun.node.body)) {
+        const parent = path.getFunctionParent();
+        if (parent === fun) {
+          // make sure "worklet" is listed among directives for the fun
+          // this is necessary as because of some bug, babel will attempt to
+          // process replaced function if it is nested inside another function
+          const directives = fun.node.body.directives;
+          if (
+            directives &&
+            directives.length > 0 &&
+            directives.some(
+              (directive) =>
+                isDirectiveLiteral(directive.value) &&
+                directive.value.value === 'worklet'
+            )
+          ) {
+            shouldBeProcessed = true; // move workletization of explicit worklet to the end
+          }
+        } else if (
+          state.opts.useOnExitLogicForWorkletNodes &&
+          (isFunctionDeclaration(parent) ||
+            isFunctionExpression(parent) ||
+            isArrowFunctionExpression(parent))
         ) {
-          processIfWorkletFunction(fun, state);
+          processIfWorkletNode(
+            parent as NodePath<
+              FunctionDeclaration | FunctionExpression | ArrowFunctionExpression
+            >,
+            state
+          ); // recursive call for nested explicit worklets
         }
       }
     },
   });
+  if (shouldBeProcessed) processIfWorkletFunction(fun, state); // process explicit worklet here
 }
```

The diff is a bit messy but to explain the change: once we find a function that has a `worklet` directive we transpile (only using `Reanimated plugin`) its contents depth first recursively (that would be an equivalent of using exit). After that's done, we transpile the root of the subtree. I think this change would be solve your issue.

This behaviour would be enabled by passing `useOnExitLogicForWorkletNodes: true` as an option for the plugin, hence it would be for deliberate use only. I must consult this solution with other team members to decide if we actually want to implement it but I think it is harmless as long as we would officialy not provide support for this behaviour. Keep in mind that you'd still might need to use a wrapper that @tomekzaw proposed in some cases.

Also this is new, TypeScript version of the plugin which had some changes. If you want to test this solution you can find it [on this branch](https://github.com/software-mansion/react-native-reanimated/tree/%40tjzel/plugin-nested-worklets).

mrousavy:
Interesting stuff, thanks!

I'll try this change :)

kmagiera:
What @tjzel posted seem like a more safer approach and it'd be my preference to go with that approach instead. It also makes sense to put that behind a config flag for now especially that we are making a number of changes to the plugin already and want to avoid potential issues to blame this change. It putting this functionality behind a config flag for at least one/two future releases works for you that'd help accelerate merging these change in

mrousavy:
Yep sounds amazing! I'll test this and come back with some feedback if I have some, thanks guys!

tjzel:
We enabled nested worklets with #4332

PR #4062, opened 2023-02-17
Add inline styles without animation functions

graszka22:
## Summary
- a rewrite of https://github.com/software-mansion/react-native-reanimated/pull/3354 
- adds inline styles without animation functions just by starting a mapper
- a mapper is started only when there are inline styles or inline styles have changed, also there is just one mapper for all styles properties for the given component
- adds a way to disable inline styles warning in plugin options
- adds support for transform property in styles, because I forgot to check that in the previous PR
- it would be nice to have tests for this but `toHaveAnimatedStyle` won't work without any modifications (probably adding another updater just for jest testing like in `useAnimatedStyles`), so maybe I'll add it later

## Test plan

Run this in ReanimatedExample, FabricExample and WebExample:
<details>
<summary>Example</summary>

```js
import Animated, {
  useSharedValue,
  withTiming,
  Easing,
import React from 'react';

function AnimatedStyleUpdateExample(): React.ReactElement {

  const config = {
    duration: 500,
    easing: Easing.bezierFn(0.5, 0.01, 0, 1),
  };

  return (
    <View
      style={{
        flex: 1,
        flexDirection: 'column',
      }}>
      <Animated.View
        style={[
          { height: 80, backgroundColor: 'black', margin: 30 },
          { width: randomWidth },
          { transform: [{ translateX: randomWidth }] },
        ]}
      />
      <Button
        title="toggle"
        onPress={() => {
          randomWidth.value = withTiming(Math.random() * 350, config);
        }}
      />
    </View>
  );
}

export default AnimatedStyleUpdateExample;
```
</details>


src/createAnimatedComponent.tsx
```diff
@@ -409,11 +472,64 @@ export default function createAnimatedComponent(
...
+        this._inlineStyles
+      );
+
+      if (hasChanged) {
+        const sharableViewDescriptors =
+          this._inlineStylesViewDescriptors.sharableViewDescriptors;
+
+        const maybeViewRef = NativeReanimatedModule.native
+          ? undefined
+          : ({ items: new Set([this]) } as ViewRefSet<any>); // see makeViewsRefSet
```
kmagiera, line 488: we can follow similar approach with view descriptors and only define them here, no?

src/createAnimatedComponent.tsx
```diff
@@ -479,6 +595,17 @@ export default function createAnimatedComponent(
                 };
               }
               return this.initialStyle;
+            } else if (hasInlineStyles(style)) {
+              const newStyle: StyleProps = {};
+              for (const [key, styleValue] of Object.entries(style)) {
+                if (
+                  styleValue.value === undefined &&
```
kmagiera, line 602: can we define a method `isSharedValue` and use it here? I was thinking about adding such method for some time and it'll be useful in other places as well.

src/createAnimatedComponent.tsx
```diff
@@ -479,6 +595,17 @@ export default function createAnimatedComponent(
...
+                  !(key === 'transform' && isInlineStyleTransform(styleValue))
+                ) {
+                  newStyle[key] = styleValue;
+                }
```
kmagiera, line 606: if it is shared value don't we still include it in `newStyle` ? Much like we do with `useAnimatedStyle` where we just take the initial value, so here maybe we should do `newStyle[key] = styleValue.value`

src/createAnimatedComponent.tsx
```diff
@@ -409,11 +472,64 @@ export default function createAnimatedComponent(
...
+
+        const updaterFunction = () => {
+          'worklet';
+          const update: StyleProps = {};
+          for (const [key, styleValue] of Object.entries(newInlineStyles)) {
+            if (key === 'transform') {
+              update[key] = styleValue.map((t: Record<string, any>) =>
+                Object.keys(t).reduce(
+                  (acc, curr) => ({ ...acc, [curr]: t[curr].value }),
+                  {}
```
kmagiera, line 498: I'd prefer to rewrite this bit using simple loops. One aspect is that object spread isn't typically very fast and another one is readability. Using `map` method is ok, but `reduce` with spread is a bit too much I believe

kmagiera (commented):
This is much better 👏 👏 

I left a couple of comments inline. I also think that we can relatively easily extend this now to also work for other non-style properties. We now use `useAnimatedProps` but since it is just an alias for `useAnimatedStyle`, making this inline styles approach work for non-style props should be very much limited to recognizing if shared value is used fo a given prop and then including it in `newInlineStyles` object (ofc we would rename this variable in such case)

src/createAnimatedComponent.tsx
```diff
@@ -409,11 +493,66 @@ export default function createAnimatedComponent(
...
+        };
+        this._inlineStyles = newInlineStyles;
+        if (this._inlineStylesMapperId) {
+          stopMapper(this._inlineStylesMapperId);
+        }
+        this._inlineStylesMapperId = null;
+        if (Object.keys(newInlineStyles).length) {
+          this._inlineStylesMapperId = startMapper(
+            updaterFunction,
+            Object.values(newInlineStyles)
```
kmagiera, line 538: it seem like if there is a transform object with only one of transform elements being a shared value, this won't actually extract that shared value. Can you verify if this works ok?

kmagiera (commented):
Looks good, left a few more comments inline

plugin.js
```diff
@@ -692,6 +692,87 @@ function processWorklets(t, path, state) {
...
+  // replaces `sharedvalue.value` with `((){console.warn("warning");return sharedvalue.value;})()`
+  return t.callExpression(
+    t.arrowFunctionExpression(
+      [],
+      t.blockStatement([
+        t.expressionStatement(
+          t.callExpression(
+            t.memberExpression(t.identifier('console'), t.identifier('warn')),
+            [
+              t.stringLiteral(
```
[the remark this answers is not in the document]
graszka22: I used the function to get the warning message but left the `console.warn` call in the plugin. That way we still get a nice arrow in Source section that shows *exactly* where is the error.

Screenshot 2023-02-21 at 16 51 03

When `console.warn` is in the function we'd get something like that:

Screenshot 2023-02-21 at 16 40 06

kmagiera (approved):
Thanks for making all the updates.Looks good!

plugin/index.js
```diff
@@ -692,6 +692,90 @@ function processWorklets(t, path, state) {
...
+        processTransformPropertyForInlineStylesWarning(t, value);
+      } else {
+        processPropertyValueForInlineStylesWarning(t, value);
+      }
+    }
+  }
+}
+
+function processInlineStylesWarning(t, path, state) {
+  if (state.opts.disableInlineStylesWarning) return;
```
kmagiera, line 759: We should only generate warning for DEV builds. In prod you won't see the warning appear anyways.

kmagiera: There is `isRelease` function used in the plugin that you can call here to check the build variation

PR #3994, opened 2023-01-26
[Shared Element Transition] Run animations for modal screens

piaskowyk:
## Summary

Added possibility to run animation during opening a modal screen.

Continuation of PR here - https://github.com/software-mansion/react-native-reanimated/pull/4065

PR #4065, opened 2023-02-17
[Shared Element Transition] Added shared transition support for modal screens

piaskowyk:
## Summary

This PR provides support for shared transition in modal screens and support for transform animations.

previous version: https://github.com/software-mansion/react-native-reanimated/pull/3994

### DIFF
--- a/android/src/main/cpp/NativeProxy.cpp
+++ b/android/src/main/cpp/NativeProxy.cpp
@@ -7,6 +7,7 @@
 #include <react/jni/ReadableNativeMap.h>
 
 #include <memory>
+#include <sstream>
 #include <string>
 
 #include "AndroidErrorHandler.h"
@@ -357,10 +358,26 @@ void NativeProxy::installJSIBindings(
         jsi::Object yogaValues(rt);
         for (const auto &entry : *values) {
           try {
-            auto key =
-                jsi::String::createFromAscii(rt, entry.first->toStdString());
-            auto value = stod(entry.second->toStdString());
-            yogaValues.setProperty(rt, key, value);
+            std::string keyString = entry.first->toStdString();
+            std::string valueString = entry.second->toStdString();
+            auto key = jsi::String::createFromAscii(rt, keyString);
+            if (keyString == "currentTransformMatrix" ||
+                keyString == "targetTransformMatrix") {
+              std::vector<float> transformMatrixList;
+              std::istringstream stringStream(valueString);
+              std::copy(
+                  std::istream_iterator<float>(stringStream),
+                  std::istream_iterator<float>(),
+                  std::back_inserter(transformMatrixList));
+              jsi::Array matrix(rt, 9);
+              for (int i = 0; i < 9; i++) {
+                matrix.setValueAtIndex(rt, i, transformMatrixList[i]);
+              }
+              yogaValues.setProperty(rt, key, matrix);
+            } else {
+              auto value = stod(valueString);
+              yogaValues.setProperty(rt, key, value);
+            }
           } catch (std::invalid_argument e) {
             if (auto errorHandler = weakErrorHandler.lock()) {
               errorHandler->setError("Failed to convert value to number");
--- a/android/src/main/cpp/NativeProxy.cpp
+++ b/android/src/main/cpp/NativeProxy.cpp
@@ -7,11 +7,11 @@
 #include <react/jni/ReadableNativeMap.h>
 
 #include <memory>
-#include <sstream>
 #include <string>
 
 #include "AndroidErrorHandler.h"
 #include "AndroidScheduler.h"
+#include "JsiUtils.h"
 #include "LayoutAnimationsManager.h"
 #include "NativeProxy.h"
 #include "PlatformDepMethodsHolder.h"
@@ -349,23 +349,8 @@ void NativeProxy::installJSIBindings(
 
   std::weak_ptr<ErrorHandler> weakErrorHandler = errorHandler;
 
-  auto convertToArray = [](jsi::Runtime &rt, const std::string &value) {
-    std::vector<float> transformMatrixList;
-    std::istringstream stringStream(value);
-    std::copy(
-        std::istream_iterator<float>(stringStream),
-        std::istream_iterator<float>(),
-        std::back_inserter(transformMatrixList));
-    assert(transformMatrixList.size() == 9);
-    jsi::Array matrix(rt, 9);
-    for (int i = 0; i < 9; i++) {
-      matrix.setValueAtIndex(rt, i, transformMatrixList[i]);
-    }
-    return matrix;
-  };
-
   layoutAnimations->cthis()->setAnimationStartingBlock(
-      [wrt, weakModule, weakErrorHandler, convertToArray](
+      [wrt, weakModule, weakErrorHandler](
           int tag,
           alias_ref<JString> type,
           alias_ref<JMap<jstring, jstring>> values) {
@@ -378,7 +363,8 @@ void NativeProxy::installJSIBindings(
             auto key = jsi::String::createFromAscii(rt, keyString);
             if (keyString == "currentTransformMatrix" ||
                 keyString == "targetTransformMatrix") {
-              jsi::Array matrix = convertToArray(rt, valueString);
+              jsi::Array matrix =
+                  jsi_utils::convertStringToArray(rt, valueString, 9);
               yogaValues.setProperty(rt, key, matrix);
             } else {
               auto value = stod(valueString);
--- a/android/src/main/cpp/NativeProxy.cpp
+++ b/android/src/main/cpp/NativeProxy.cpp
@@ -349,8 +349,23 @@ void NativeProxy::installJSIBindings(
 
   std::weak_ptr<ErrorHandler> weakErrorHandler = errorHandler;
 
+  auto convertToArray = [](jsi::Runtime &rt, const std::string &value) {
+    std::vector<float> transformMatrixList;
+    std::istringstream stringStream(value);
+    std::copy(
+        std::istream_iterator<float>(stringStream),
+        std::istream_iterator<float>(),
+        std::back_inserter(transformMatrixList));
+    assert(transformMatrixList.size() == 9);
+    jsi::Array matrix(rt, 9);
+    for (int i = 0; i < 9; i++) {
+      matrix.setValueAtIndex(rt, i, transformMatrixList[i]);
+    }
+    return matrix;
+  };
+
   layoutAnimations->cthis()->setAnimationStartingBlock(
-      [wrt, weakModule, weakErrorHandler](
+      [wrt, weakModule, weakErrorHandler, convertToArray](
           int tag,
           alias_ref<JString> type,
           alias_ref<JMap<jstring, jstring>> values) {
@@ -363,16 +378,7 @@ void NativeProxy::installJSIBindings(
             auto key = jsi::String::createFromAscii(rt, keyString);
             if (keyString == "currentTransformMatrix" ||
                 keyString == "targetTransformMatrix") {
-              std::vector<float> transformMatrixList;
-              std::istringstream stringStream(valueString);
-              std::copy(
-                  std::istream_iterator<float>(stringStream),
-                  std::istream_iterator<float>(),
-                  std::back_inserter(transformMatrixList));
-              jsi::Array matrix(rt, 9);
-              for (int i = 0; i < 9; i++) {
-                matrix.setValueAtIndex(rt, i, transformMatrixList[i]);
-              }
+              jsi::Array matrix = convertToArray(rt, valueString);
               yogaValues.setProperty(rt, key, matrix);
             } else {
               auto value = stod(valueString);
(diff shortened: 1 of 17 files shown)


android/src/main/cpp/NativeProxy.cpp
```diff
@@ -357,10 +358,26 @@ void NativeProxy::installJSIBindings(
...
+                keyString == "targetTransformMatrix") {
+              std::vector<float> transformMatrixList;
+              std::istringstream stringStream(valueString);
+              std::copy(
+                  std::istream_iterator<float>(stringStream),
+                  std::istream_iterator<float>(),
+                  std::back_inserter(transformMatrixList));
+              jsi::Array matrix(rt, 9);
+              for (int i = 0; i < 9; i++) {
+                matrix.setValueAtIndex(rt, i, transformMatrixList[i]);
```
jwajgelt, line 374: Can you add a length assertion on `transformMatrixList` here, so we don't read from after the end if we pass a wrong value?

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -357,10 +358,26 @@ void NativeProxy::installJSIBindings(
         jsi::Object yogaValues(rt);
         for (const auto &entry : *values) {
           try {
-            auto key =
-                jsi::String::createFromAscii(rt, entry.first->toStdString());
-            auto value = stod(entry.second->toStdString());
-            yogaValues.setProperty(rt, key, value);
+            std::string keyString = entry.first->toStdString();
+            std::string valueString = entry.second->toStdString();
```
jwajgelt, line 362: Why do we call `toStdString()` and then parse the strings here? Can't we access the values more directly?

android/src/main/cpp/NativeProxy.cpp
jwajgelt, line 374: I think that this block should be put in a helper function.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -30,6 +30,7 @@ public class SharedTransitionManager {
   private final Map<Integer, View> mCurrentSharedTransitionViews = new HashMap<>();
   private View mTransitionContainer;
   private final List<View> mRemovedSharedViews = new ArrayList<>();
+  private final Set<Integer> mShouldBeHidden = new HashSet<>();
```
jwajgelt, line 33: `mShouldBeHidden` sounds like a boolean flag, not like a set of view(tags).
```suggestion
  private final Set<Integer> mHiddenViewTags = new HashSet<>();
```
perhaps? Or maybe
```suggestion
  private final Set<Integer> mViewTagsToHide = new HashSet<>();
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -89,6 +95,23 @@ public Snapshot(View view) {
...
     height = view.getHeight();
+
+    View transformedView = findTransformedView(view);
+    if (transformedView != null) {
+      float[] transformMatrixArray = new float[9];
+      transformedView.getMatrix().getValues(transformMatrixArray);
+      transformMatrix = new ArrayList<>();
+      for (int i = 0; i < 9; i++) {
+        transformMatrix.add(transformMatrixArray[i]);
+      }
```
jwajgelt, line 106: You can replace this with
```suggestion
      transformMatrix = new ArrayList<>(Arrays.asList(transformMatrixArray));
```
If you're sure `transformMatrixArray`'s length is 9.
If you don't need `transformMatrix` to be an `ArrayList` (I don't think you do, it just needs to be a `List`), you may even just have
```suggestion
      transformMatrix = Arrays.asList(transformMatrixArray);
```
I'm pretty sure

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -137,4 +163,37 @@ public HashMap<String, Object> toBasicMap() {
...
+      isTransformed =
+          transformArray[0] != 1
+              || transformArray[1] != 0
+              || transformArray[2] != 0
+              || transformArray[3] != 0
+              || transformArray[4] != 1
+              || transformArray[5] != 0
+              || transformArray[6] != 0
+              || transformArray[7] != 0
+              || transformArray[8] != 1;
```
jwajgelt, line 193: You could store an identity matrix somewhere (as a private static final in this class, for example), and then use `Arrays.equals` to compare them - I think that'd be a bit more clear.

android/src/paper/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -57,12 +57,25 @@ public static NativeMethodsHolder createNativeMethodsHolder(LayoutAnimations lay
...
-            public void startAnimation(int tag, String type, HashMap<String, Float> values) {
+            public void startAnimation(int tag, String type, HashMap<String, Object> values) {
                 LayoutAnimations layoutAnimations = weakLayoutAnimations.get();
                 if (layoutAnimations != null) {
                     HashMap<String, String> preparedValues = new HashMap<>();
                     for (String key : values.keySet()) {
-                        preparedValues.put(key, values.get(key).toString());
+                      String stringValue = values.get(key).toString();
+                      if (key.endsWith("TransformMatrix")) {
+                        // transforms string: '[1, 2, 3]' -> '1, 2, 3'
```
jwajgelt, line 67:
```suggestion
                        // transforms string: '[1, 2, 3]' -> '1 2 3'
```
right?

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -22,6 +16,7 @@ @implementation REASharedTransitionManager {
   BOOL _isSharedTransitionActive;
   NSMutableArray<REASharedElement *> *_sharedElements;
   REAAnimationsManager *_animationManager;
+  NSMutableSet<NSNumber *> *_viewShouldBeHidden;
```
jwajgelt, line 19: like on Android, I think
```suggestion
  NSMutableSet<NSNumber *> *_hiddenViews;
```
or
```suggestion
  NSMutableSet<NSNumber *> *_viewsToHide;
```
would be clearer

ios/LayoutReanimation/REASnapshot.m
```diff
@@ -40,16 +41,71 @@ - (void)makeSnapshotForView:(UIView *)view useAbsolutePositionOnly:(BOOL)useAbso
...
-      _values[@"headerHeight"] = @(header.frame.size.height);
+      CGFloat headerHeight = header.frame.size.height;
+      CGFloat headerOriginY = header.frame.origin.y;
+      UIView *screen = [REAScreensHelper getScreenForView:view];
+      if ([REAScreensHelper isScreenModal:screen] && screen.superview == nil) {
+        int additionalModalOffset = 0;
+        UIView *screenWrapper = [REAScreensHelper getScreenWrapper:view];
+        int screenType = [REAScreensHelper getScreenType:screenWrapper];
+        if (screenType == 1) { // RNSScreenStackPresentationModal
+          additionalModalOffset = 69; // Default iOS modal is shifted from screen top edge by 69px
```
jwajgelt, line 52: Can we have this in a constant somewhere, instead of a magic constant?

jwajgelt (commented):
Left some comments, other than that, it looks good to me.

android/src/main/cpp/NativeProxy.cpp
piaskowyk: I don't like the current solution, but this was the easiest way to achieve it without changing a lot of codebase from layout animation. I think it should be done in another PR in the future because I don't want to increase the complexity of this PR.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
piaskowyk: Unfortunately not :/ It would works with `Float[]` but not with `float[]`

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
piaskowyk: It would work if transformArray would be Float[] instead of float[]. We can use conversion like with `transformMatrix` but this requires the usage of another one loop, and I am not sure if it will be more readable.

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -348,19 +349,41 @@ void NativeProxy::installJSIBindings(
 
   std::weak_ptr<ErrorHandler> weakErrorHandler = errorHandler;
 
+  auto convertToArray = [](jsi::Runtime &rt, const std::string &value) {
+    std::vector<float> transformMatrixList;
+    std::istringstream stringStream(value);
+    std::copy(
+        std::istream_iterator<float>(stringStream),
+        std::istream_iterator<float>(),
+        std::back_inserter(transformMatrixList));
```
tomekzaw, line 358: Can you please explain what exactly happens here? 😄

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -348,19 +349,41 @@ void NativeProxy::installJSIBindings(
...
+    jsi::Array matrix(rt, 9);
+    for (int i = 0; i < 9; i++) {
+      matrix.setValueAtIndex(rt, i, transformMatrixList[i]);
+    }
+    return matrix;
+  };
+
   layoutAnimations->cthis()->setAnimationStartingBlock(
-      [wrt, weakModule, weakErrorHandler](
+      [wrt, weakModule, weakErrorHandler, convertToArray](
```
tomekzaw, line 368: Can we convert `convertToArray` from lambda to a function (outside of `NativeProxy.cpp`, maybe some utils)?

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -348,19 +349,41 @@ void NativeProxy::installJSIBindings(
...
           int tag,
           alias_ref<JString> type,
           alias_ref<JMap<jstring, jstring>> values) {
         auto &rt = *wrt.lock();
         jsi::Object yogaValues(rt);
         for (const auto &entry : *values) {
```
tomekzaw, line 374: Is it possible to use this syntax here?
```suggestion
        for (const auto &[first, second] : *values) {
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -309,6 +322,31 @@ public void setNewProps(
         (props.get(Snapshot.HEIGHT) != null)
             ? ((Double) props.get(Snapshot.HEIGHT)).floatValue()
             : PixelUtil.toDIPFromPixel(view.getHeight());
+
+    if (props.containsKey(Snapshot.TRANSFORM_MATRIX)) {
+      float[] matrixValues = new float[9];
+      if (props.get(Snapshot.TRANSFORM_MATRIX) instanceof ReadableNativeArray) {
+        // this array comes from Java Script
```
tomekzaw, line 329:
```suggestion
        // this array comes from JavaScript
```

android/src/paper/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -57,12 +57,25 @@ public static NativeMethodsHolder createNativeMethodsHolder(LayoutAnimations lay
...
+                      String stringValue = values.get(key).toString();
+                      if (key.endsWith("TransformMatrix")) {
+                        // transforms string: '[1, 2, 3]' -> '1 2 3'
+                        // to make usage of std::istringstream in C++ easier
+                        preparedValues.put(
+                          key,
+                          stringValue
+                            .replace(",", "")
+                            .replace("[", "")
+                            .replace("]", "")
```
tomekzaw, line 74: Can we move it to a separate utility function?

android/src/main/cpp/NativeProxy.cpp
jwajgelt: It reads from the `value` string till the end, parsing it into floats, and pushing those floats to the `transformMatrixList`.

> `std::istream_iterator` is a single-pass input iterator that reads successive objects of type T from the [`std::basic_istream`](https://en.cppreference.com/w/cpp/io/basic_istream) object for which it was constructed, by calling the appropriate `operator>>`

from `std::istream_iterator` docs

android/src/main/cpp/NativeProxy.cpp
jwajgelt: I personally don't see a reason to, when both are immediately converted to something else.

android/src/main/cpp/NativeProxy.cpp
jwajgelt: :+1: for defining it as a separate function.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
jwajgelt: No, `Arrays.equals` has overloads for arrays of primitive types:
https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#equals(float[],%20float[])

jwajgelt: Unless you meant something different?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
jwajgelt: Right 👍

jwajgelt (approved):

PR #4067, opened 2023-02-17
Move plugin.js to its separate directory

tjzel:
## Summary

Moved `plugin.js` to `plugin/index.js` to effectively split it into multiple files in the future. Amended some configs in Example apps.

## Test plan

run `yarn jest`
build `Example` and `FabricExample` and briefly check if they work correctly
run `./createNPMPackage.sh` and create a test app that installs reanimated from this tarball, check if it builds properly and works


piaskowyk:
I saw that you updated plugin path for FabricExample, but what about Example, TVOSExample and WebExample?

FabricExample/package.json
```diff
@@ -35,6 +35,7 @@
     "@types/react-test-renderer": "^18.0.0",
     "babel-jest": "^29.2.1",
     "eslint": "^8.19.0",
+    "eslint-plugin-ft-flow": "^2.0.3",
```
piaskowyk, line 38: Why we need this?

piaskowyk:
What if someone uses direct import? `plugins: ['react-native-reanimated/plugin.js'],` maybe we should leave file `plugin.js` with export from `plugin/index.js`?
I'm not sure about that - let's discuss it

FabricExample/package.json
tjzel: I was sloppy and committed this change. In current commit it is removed.

tjzel:
> I see that you updated plugin path for FabricExample, but what about Example, TVOSExample and WebExample?

I updated config in FabricExample and root directory since those were the only configs that explicitly demanded `plugin.js`, not `plugin`.

> What if someone uses direct import? `plugins: ['react-native-reanimated/plugin.js'],` maybe we should leave file `plugin.js` with export from `plugin/index.js` or with proper warning? I'm not sure about that - let's discuss it

In our [docs](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation/) we instruct the user to add `plugin`, not `plugin.js`, same for Example apps. We already have a warning to include `plugin` - but we might amend it and specifically request to include `plugin` and not `plugin.js`.

tjzel:
> I see that you updated plugin path for FabricExample, but what about Example, TVOSExample and WebExample?

I updated config in FabricExample and root directory since those were the only configs that explicitly demanded `plugin.js`, not `plugin`.

> What if someone uses direct import? `plugins: ['react-native-reanimated/plugin.js'],` maybe we should leave file `plugin.js` with export from `plugin/index.js` or with proper warning? I'm not sure about that - let's discuss it

In our [docs](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation/) we instruct the user to add `plugin`, not `plugin.js`, same for Example apps. We already have a warning to include `plugin` - but we might amend it and specifically request to include `plugin` and not `plugin.js`.

tjzel:
> I see that you updated plugin path for FabricExample, but what about Example, TVOSExample and WebExample?

I updated config in FabricExample and root directory since those were the only configs that explicitly demanded `plugin.js`, not `plugin`.

> What if someone uses direct import? `plugins: ['react-native-reanimated/plugin.js'],` maybe we should leave file `plugin.js` with export from `plugin/index.js` or with proper warning? I'm not sure about that - let's discuss it

In our [docs](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation/) we instruct the user to add `plugin`, not `plugin.js`, same for Example apps. We already have a warning to include `plugin` - but we might amend it and specifically request to include `plugin` and not `plugin.js`.

tjzel:
> I see that you updated plugin path for FabricExample, but what about Example, TVOSExample and WebExample?

I updated config in FabricExample and root directory since those were the only configs that explicitly demanded `plugin.js`, not `plugin`.

> What if someone uses direct import? `plugins: ['react-native-reanimated/plugin.js'],` maybe we should leave file `plugin.js` with export from `plugin/index.js` or with proper warning? I'm not sure about that - let's discuss it

In our [docs](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation/) we instruct the user to add `plugin`, not `plugin.js`, same for Example apps. We already have a warning to include `plugin` - but we might amend it and specifically request to include `plugin` and not `plugin.js`. In my opinion we shouldn't support improperly imported plugin as this may cause issues in the future.

tjzel:
> I see that you updated plugin path for FabricExample, but what about Example, TVOSExample and WebExample?

I updated config in FabricExample and root directory since those were the only configs that explicitly demanded `plugin.js`, not `plugin`.

> What if someone uses direct import? `plugins: ['react-native-reanimated/plugin.js'],` maybe we should leave file `plugin.js` with export from `plugin/index.js` or with proper warning? I'm not sure about that - let's discuss it

In our [docs](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation/) we instruct the user to add `plugin`, not `plugin.js`, same for Example apps. We already have a warning to include `plugin` - but we might amend it and specifically request to include `plugin` and not `plugin.js`. In my opinion we shouldn't support improperly imported plugin as this may cause issues in the future.

tomekzaw:
> What if someone uses direct import? plugins: ['react-native-reanimated/plugin.js'], maybe we should leave file plugin.js with export from plugin/index.js or with proper warning?
I'm not sure about that - let's discuss it

That's a great idea, we could do it, but I think in this case the error message will be something like file not found so it will be easy for us to recommend the proper solution in case anyone submits an issue with this error.

tomekzaw (approved):

PR #4068, opened 2023-02-18
Add inline animated props

graszka22:
## Summary

Based on https://github.com/software-mansion/react-native-reanimated/pull/4062

Adds support for inline animated props like this:
```js

export default function SvgExample() {
  const sv = useSharedValue('0%');

  sv.value = withRepeat(withTiming('50%', { duration: 500 }), -1, true);

  return (
      <Svg height="200" width="200">
        <AnimatedCircle cx="50%" cy="50%" fill="lime" r={sv} />
      </Svg>
    </View>
  );
}
```
This syntax is so pretty 🤩  but there is one issue with it though. We can't differentiate between a shared value and ordinary object with `value` prop. It may happen (though I think it's rather unlikely) that user doesn't want the prop to animate but just wants to pass an object or shared value. So I'm checking if it's a whitelisted prop (user had to whitelist it anyway). If that's not enough we may add whitelist/blacklist per component in `createAnimatedComponent`. Or use `animatedProps` prop but that won't be so pretty ;_;.

Also I'm no adding a warning in babel plugin. Imo it would lead to too many false positives. Something like `<Component prop={obj.value} />` is normal.
 
## Test plan

Example above. Run on ReanimatedExample, FabricExample and something similar on WebExample.


mrousavy:
That's pretty cool - but can't you just add a `unique symbol` to the Shared Value that then distinguishes it from any other object with a `value` prop?

kmagiera:
@graszka22 I wonder if we still want to require whitelisting props in such a scenario. If you pass a shared value as prop you clearly want that prop whitelisted. So in my opinion we should be whitelisting such props automatically. If at all this is necessary as I thought that we made that process semi-automated in the past (maybe @piaskowyk would remember)

graszka22:
Now I see that the props are automatically whitelisted with `adaptViewConfig` for `useAnimatedStyle` (it's not documented though). Since now we can distinguish shared values better, I'm now doing the same with inline props.

docs/docs/fundamentals/animations.md
```diff
@@ -381,3 +381,24 @@ For this purpose Reanimated exposes a separate hook called `useAnimatedProps`.
 It works in a very similar way to `useAnimatedStyle` but instead of expecting a method that returns the animated styles, we expect the returned object to contain properties that we want to animate.
 Then, in order to hook animated props to a view, we provide the resulting object as `animatedProps` property to the "Animated" version of the view type we want to render (e.g. `Animated.View`).
 Please check the documentation of the [`useAnimatedProps`](../api/hooks/useAnimatedProps) hook for usage examples.
+
+## Shared Values in Properties
+You can also pass a shared value as a property like this:
+```js
+const AnimatedCircle = Animated.createAnimatedComponent(Circle);
+
+Animated.addWhitelistedNativeProps({ r: true });
```
kmagiera, line 390: Can we remove this line now? Also, there is similar example in the PR description

src/createAnimatedComponent.tsx
```diff
@@ -128,25 +128,36 @@ function hasInlineStyles(style: StyleProps): boolean {
   });
 }
 
-function getInlineStylesFromProps(
+function getInlinePropsFromProps(
```
kmagiera, line 131: This name is weird, maybe we can rename to `extractSharedValuesMapFromProps`

kmagiera (approved):
Looks good, just left some minor comments inline.

PR #4082, opened 2023-02-21
Fix host obj reference loop

kmagiera:
## Summary

Some recent changes (#4060 and #4024) resulted in introducing a reference loop. The reason was that we started capturing the whole module as host object due to the fact we referenced the module in `defineAnimation` method. As a result, the VM wouldn't deallocate properly and we started getting instance checker crashes on reloads.

## Test plan

Start app on iOS. Try out reloading JS several times -> no crash.


tomekzaw (approved):
🤯

PR #4086, opened 2023-02-22
[Shared Element Transition] Restore X-axis origin

piaskowyk:
## Summary

Restore the X-axis origin after the shared animation. Previously I was restoring only Y origin to handle the screen with a header, but X origin restoring is required when the screen has its transition animation.

### DIFF
--- a/ios/LayoutReanimation/REASharedTransitionManager.m
+++ b/ios/LayoutReanimation/REASharedTransitionManager.m
@@ -558,7 +558,8 @@ static REASharedTransitionManager *_sharedTransitionManager;
                                                    forTag:viewTag
                                        isSharedTransition:YES];
       float originYByParent = [viewSourcePreviousSnapshot.values[@"originYByParent"] floatValue];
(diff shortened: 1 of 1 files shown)


ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -558,7 +558,8 @@ - (void)finishSharedAnimation:(UIView *)view
                                                    forTag:viewTag
                                        isSharedTransition:YES];
       float originYByParent = [viewSourcePreviousSnapshot.values[@"originYByParent"] floatValue];
-      CGRect frame = CGRectMake(view.frame.origin.x, originYByParent, view.frame.size.width, view.frame.size.height);
+      float originXByParent = [viewSourcePreviousSnapshot.values[@"originXByParent"] floatValue];
```
tomekzaw, line 561: Can we swap these two lines?

tomekzaw (approved):

PR #4088, opened 2023-02-23
[Shared Element Transition] Examples

piaskowyk:
## Summary

Added examples with Shared Element Transition to Example app.


Example/src/SharedElementTransition/Modals.tsx
```diff
@@ -0,0 +1,169 @@
...
+    onStart: (_, ctx) => {
+      ctx.startX = translation.x.value;
+      ctx.startY = translation.y.value;
+    },
+    onActive: (event, ctx) => {
+      translation.x.value = ctx.startX + event.translationX;
+      translation.y.value = ctx.startY + event.translationY;
+    },
+    onEnd: (_) => {
+      console.log(translation.x.value, translation.y.value);
```
jwajgelt, line 118:
```suggestion
```

Example/src/SharedElementTransition/Modals.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  const animatedStyle = useAnimatedStyle(() => {
+    return {
+      transform: [
+        { translateX: translation.x.value },
+        { translateY: translation.y.value },
+        {
+          scale:
+            1 -
+            (Math.abs(translation.x.value) + Math.abs(translation.y.value)) /
```
jwajgelt, line 135: Can you make a derived value for distance and use it instead?

Example/src/SharedElementTransition/Modals.tsx
```diff
@@ -0,0 +1,169 @@
...
+              500,
+        },
+      ],
+    };
+  });
+
+  return (
+    <PanGestureHandler onGestureEvent={gestureHandler}>
```
jwajgelt, line 143: The gesture handler doesn't work on Android - can we wrap it in a GH Root?

jwajgelt (approved):
Other than the three comments, the examples look good 👍

Example/src/SharedElementTransition/CustomTransition.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import Animated, {
+  SharedTransition,
+  withSpring,
+} from 'react-native-reanimated';
+import { StackScreenProps } from '@react-navigation/stack';
+
+const Stack = createNativeStackNavigator();
+
+const transition = SharedTransition.custom((values: any) => {
```
tomekzaw, line 13: It works without `any` here.
```suggestion
const transition = SharedTransition.custom((values) => {
```

Example/src/App.tsx
```diff
@@ -60,6 +60,18 @@ import { WaterfallGridExample } from './LayoutReanimation/WaterfallGridExample';
...
+  CustomTransitionExample,
+  GalleryExample,
+  LayoutAnimationExample,
+  ManyScreensExample,
+  ManyTagsExample,
+  NestedStacksExample,
+  ModalsExample,
+  FlatListExample,
+  RestoreStateExample,
+} from './SharedElementTransition';
```
tomekzaw, line 74: Instead of having a separate file `SharedElementTransition/index.ts` just for importing/exporting, you can simply import each example individually just like we already do.

Example/src/App.tsx
```diff
@@ -72,104 +84,144 @@ if (Platform.OS === 'android') {
...
+    title: '[SET] Restore State',
+  },
   ColorInterpolation: {
     screen: ColorInterpolationExample,
     title: 'Color interpolation',
   },
   DeleteAncestorOfExiting: {
     screen: DeleteAncestorOfExiting,
-    title: '🆕 Deleting view with an exiting animation',
+    title: 'Deleting view with an exiting animation',
```
tomekzaw, line 133: Can we mark all existing examples that use Layout Animations with `[LA]` prefix as well?

Example/src/SharedElementTransition/Card.tsx
```diff
@@ -0,0 +1,113 @@
+import * as React from 'react';
+import { View, Image, TouchableNativeFeedback } from 'react-native';
+import { ParamListBase } from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { StackScreenProps } from '@react-navigation/stack';
+import Animated from 'react-native-reanimated';
+
+const photo = require('./assets/image.jpg');
+const Stack = createNativeStackNavigator();
+const AnimatedImage = Animated.createAnimatedComponent(Image);
```
tomekzaw, line 10: You can use `Animated.Image` directly:

https://github.com/software-mansion/react-native-reanimated/blob/073057f785346f1cdaa88a19680ae84f54a310a2/src/Animated.ts#L10

Example/src/SharedElementTransition/CustomTransition.tsx
```diff
@@ -0,0 +1,79 @@
...
+  'worklet';
+  return {
+    height: withSpring(values.targetHeight),
+    width: withSpring(values.targetWidth),
```
tomekzaw, line 17: Let's swap these two lines, i.e. `width` should be first.

Example/src/SharedElementTransition/Modals.tsx
piaskowyk: Yes, but I don't think this necessary

Example/src/App.tsx
```diff
@@ -73,108 +85,152 @@ if (Platform.OS === 'android') {
...
     screen: MountingUnmounting,
   },
   ReactionsCounterExample: {
     screen: ReactionsCounterExample,
-    title: '🆕 Reactions counter',
+    title: '[LA] Reactions counter',
   },
   SwipeableList: {
-    title: '🆕 Swipeable list',
+    title: 'Swipeable list',
```
tomekzaw, line 221:
```suggestion
    title: '[LA] Swipeable list',
```

tomekzaw (approved):

PR #4089, opened 2023-02-23
Enable calling scrollTo with one parameter

mstach60161:
## Summary

Extends #4048 

When we change value in `useScrollViewOffset` wa call `scrollTo`, but we don't know what's the direction of scrolling, so we would like to pass to`scrollTo` only one parameter and check on the native side.

Solution:
- Passed infinity as a second parameter in `scrollValueSetter`
- Checked vertical/horizontal in ios/android functions
- Assign 0, scrollValue to x, y depending on the direction

## Test plan

- Run updated example on android and IOS


android/src/main/java/com/swmansion/reanimated/NativeMethodsHelper.java
```diff
@@ -55,6 +55,13 @@ public static void scrollTo(View view, double argX, double argY, boolean animate
       }
     }
 
+    // When we use useScrollValueOffset hook, we don't know the direction of scrolling
+    if (Double.isInfinite(argX)) {
+      int scrollValue = Math.min(x, y);
```
tomekzaw, line 60: Since `x` is infinity, `y` is always less than `x`, so can we use `y` directly instead?

ios/native/NativeMethods.mm
```diff
@@ -41,6 +41,15 @@ void scrollTo(int scrollViewTag, RCTUIManager *uiManager, double x, double y, bo
 {
   UIView *view = [uiManager viewForReactTag:@(scrollViewTag)];
   RCTScrollView *scrollView = (RCTScrollView *)view;
+
+  // When we use useScrollValueOffset hook, we don't know the direction of scrolling
+  if (x == INFINITY) {
+    bool isHorizontal = scrollView.scrollView.contentSize.width > scrollView.frame.size.width;
```
tomekzaw, line 47: There should be an easier way to check it

PR #4090, opened 2023-02-23
Layout animations cleanup

jwajgelt:
## Summary
Clean up the layout animation rewrite from #3332

- use Enum types for layout animation types ('entering', 'exiting' and 'layout') in both JS and native parts of the implementation
- try caching `AnimationsManager` in `LayoutAnimations.java` instead of calling `getNativeModule` (Android implementation)


Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -388,7 +388,7 @@ jsi::Value NativeReanimatedModule::configureLayoutAnimation(
     const jsi::Value &config) {
   layoutAnimationsManager_.configureAnimation(
       viewTag.asNumber(),
-      type.asString(rt).utf8(rt),
+      (LayoutAnimationType)type.asNumber(),
```
tomekzaw, line 391: C-style casts are not welcome, please use `static_cast` or define a new function `parseLayoutAnimationType` that accepts `const jsi::Value &`.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/LayoutAnimations.java
```diff
@@ -9,6 +9,13 @@
 import java.util.Map;
 
 public class LayoutAnimations {
+  public static class Types {
+    static final int ENTERING = 0;
```
tomekzaw, line 13: Please make it start from 1.

src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
```diff
@@ -60,13 +60,20 @@ export interface LayoutAnimationsValues {
   windowHeight: number;
 }
 
+export enum LayoutAnimationType {
+  ENTERING = 0,
```
tomekzaw, line 64: Please make it start from 1.

ios/native/NativeProxy.mm
```diff
@@ -297,31 +298,30 @@ static CFTimeInterval calculateTimestampWithSlowAnimations(CFTimeInterval curren
...
-        yogaValues.setProperty(rt, [key UTF8String], matrix);
-      } else {
-        yogaValues.setProperty(rt, [key UTF8String], [(NSNumber *)value doubleValue]);
-      }
-    }
 
-    weakModule.lock()->layoutAnimationsManager().startLayoutAnimation(
-        rt, [tag intValue], std::string([type UTF8String]), yogaValues);
-  }];
+        weakModule.lock()->layoutAnimationsManager().startLayoutAnimation(rt, [tag intValue], type, yogaValues);
```
tomekzaw, line 320: I don't think it's safe to use `weakModule.lock()->layoutAnimationsManager()`. If the module no longer exists, `.lock()` returns `nullptr` so we have a null pointer dereference. Can you please refactor all occurrences of `weakModule.lock()->layoutAnimationsManager()` into
```cpp
if (auto module = weakModule.lock()) {
  module->layoutAnimationsManager() // ...
}
```

tomekzaw (approved):

piaskowyk (approved):

PR #4093, opened 2023-02-23
Add gesture handler RootView to Example app

piaskowyk:
## Summary

After navigation upgrade we need to add gesture handler RootView manually.


jwajgelt (approved):

Example/src/App.tsx
```diff
@@ -343,4 +344,4 @@ export const styles = StyleSheet.create({
   },
 });
 
-export default App;
+export default gestureHandlerRootHOC(App);
```
tomekzaw, line 347: Personally I think it's more readable to add `<GestureHandlerRootView>` inside `function App` since we already have `<NavigationContainer>` there but I'm also okay with this solution.

tomekzaw (approved):

PR #4096, opened 2023-02-23
Rewrite plugin to TypeScript

tjzel:
## Summary

Rewriting plugin to typescript. Plugin will compile when doing either `yarn` or `yarn plugin`, hence it will be available in npm tarball.

I think the best way to review the changes is to compare the compiled 'index.js' file with the old one (having in mind compilation options in 'tsconfig.json'), then remove it in the follow-up commit.

## Test plan

Generate bundles for `Example` and `FabricApp` for `main`.
Generate bundles for current head of this branch with freshly compiled `index.js`. Compare the bundles. There should be no differences as long as they were built in the same directory.


package.json
```diff
@@ -76,7 +76,8 @@
   "dependencies": {
     "@babel/plugin-transform-object-assign": "^7.16.7",
     "@babel/preset-typescript": "^7.16.7",
-    "convert-source-map": "^1.7.0",
+    "@types/convert-source-map": "^2.0.0",
```
tomekzaw, line 79: I think we can move this to `devDependencies`.

plugin/index.js
```diff
@@ -1,13 +1,22 @@
 'use strict';
```
tomekzaw, line 1: I don't think we should commit output files to the repo, it should be somehow generated automatically (the same applies for `index.d.ts` and `index.js.map`).

package.json
```diff
@@ -134,7 +135,7 @@
   },
   "lint-staged": {
     "*.(js|ts|tsx)": [
-      "eslint --ext '.js,.ts,.tsx' src/ --ignore-pattern src/reanimated1 --ignore-pattern react-native-reanimated.d.ts --ignore-pattern docs",
+      "eslint --ext '.js,.ts,.tsx' src/ --ignore-pattern src/reanimated1 --ignore-pattern react-native-reanimated.d.ts --ignore-pattern docs --ignore-pattern plugin",
```
tjzel, line 138: Compiling with `"module" : "commonjs"` results with code unaccepted by linter (TSC gives snake case). Simpliest override for now is to ignore `/plugin` for now.

tomekzaw, line 138: Let's add a rule for `"plugin/**/*.ts"` then.

plugin/README.md
```diff
@@ -0,0 +1 @@
+After making any changes to `index.ts` make sure to run `tsc` afterwards to get expected `index.js`!
```
tomekzaw, line 1: First, I would like this `README.md` to be analogous to the official Babel plugins from [here](https://github.com/babel/babel/tree/main/packages/babel-plugin-proposal-export-namespace-from).

Secondly, please provide a high-level overview of what the plugin does (preparing worklets including closure calculation, auto-workletization of Reanimated hooks and Gesture Handler callbacks).

Thirdly, this is a great place to gather some information for future users and developers of the plugin. In particular, let's document how to:
* configure the development environment
* run the plugin in dev environment (`cd react-native-reanimated && npx babel file.js`)
* debug the plugin using VSCode JavaScript debugger
* run tests (`cd react-native-reanimated && yarn jest plugin.jest.js` or even better: `cd plugin && yarn test`)
* check types (`cd plugin && yarn tsc --noEmit`)
* run linter (`cd plugin && yarn lint`)
* run formatter (`cd plugin && yarn format`)

Finally, please attach these scripts to the CI.

plugin/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "target": "ESNext",
+    "removeComments": true,
+    "module": "commonjs",
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "declaration": true
```
tomekzaw, line 8: Please briefly explain what each option here does so we can discuss and understand it better.

package.json
tjzel: You are right, I will move it.

plugin/README.md
tjzel: While it is a great idea, I think we should wait for new documentation for Reanimated3 with it. We could provide instructions on the website and simply link it in this readme.

jwajgelt: I agree with @tomekzaw that the readme is the correct place for this information.
The docs are meant for Reanimated users, and should instruct how to install the plugin, maybe provide some rationale why it's necessary, and that's basically it, while the information on plugin development belongs in its repo.

At least that's my 2 cents 😉

piaskowyk:
Are you sure that we have enough tests of the plugin?

tjzel:
@piaskowyk No, I don't think we have but my intention with this PR is to replicate current plugin's behaviour but in compiled TS file and then improve and hone tests since it will be easier then.

tjzel:
@jwajgelt @tomekzaw I put some content into README file, please read it and tell me whether it's what you had in mind and if you spot any errors (about how plugin works in the first place).

package.json
```diff
@@ -5,14 +5,16 @@
   "scripts": {
     "test": "yarn run format:js && yarn run lint:js && yarn run test:unit",
     "test:unit": "jest",
-    "lint": "yarn lint:js && yarn lint:cpp && yarn lint:java && yarn lint:ios && yarn lint:docs",
+    "lint": "yarn lint:js && yarn lint:plugin && yarn lint:cpp && yarn lint:java && yarn lint:ios && yarn lint:docs",
     "lint:js": "eslint --ext '.js,.ts,.tsx' src/ && yarn prettier --check src/",
+    "lint:plugin": "eslint --ext '.ts' plugin && yarn prettier --check plugin/",
```
tomekzaw, line 10: Move to `plugin/package.json` as `lint` script

package.json
```diff
@@ -134,9 +139,14 @@
   },
   "lint-staged": {
     "*.(js|ts|tsx)": [
-      "eslint --ext '.js,.ts,.tsx' src/ --ignore-pattern src/reanimated1 --ignore-pattern react-native-reanimated.d.ts --ignore-pattern docs",
+      "eslint --ext '.js,.ts,.tsx' src/ --ignore-pattern src/reanimated1 --ignore-pattern react-native-reanimated.d.ts --ignore-pattern docs --ignore-pattern plugin",
       "prettier --write"
     ],
+    "plugin/**/*.ts": [
+      "eslint",
+      "prettier --write"
```
tomekzaw, line 147: Please use yarn scripts instead of specific commands here.

plugin/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "ReanimatedPlugin",
```
tomekzaw, line 2: Do we need this name?

__tests__/plugin.test.js
```diff
@@ -136,6 +136,20 @@ describe('babel plugin', () => {
...
+  it("doesn't remove nested 'worklets'", () => {
+    const input = `
+    function foo(x){
+      'worklet';
+      function bar(x){
+        'worklet'
+        return x+2;
+      }
+      return bar(x)+1;
+    }`;
```
tomekzaw, line 148: Please format this code

.github/workflows/validate-plugin.yml
```diff
@@ -0,0 +1,38 @@
...
+          node-version: 16
+          cache: 'yarn'
+      - name: Install node dependencies
+        run: yarn
+      - name: Install plugin dependencies
+        working-directory: plugin
+        run: yarn
+      - name: Check types
+        working-directory: plugin
+        run: yarn tsc --noEmit
```
tomekzaw, line 30:
```suggestion
        run: yarn type:check
```

__tests__/plugin.test.js
```diff
@@ -136,6 +136,20 @@ describe('babel plugin', () => {
     expect(code).toMatchSnapshot();
   });
 
+  it("doesn't remove nested 'worklets'", () => {
```
tomekzaw, line 139:
```suggestion
  it("doesn't transform nested worklets", () => {
```

plugin/README.md
```diff
@@ -0,0 +1,141 @@
...
+    },
+  });
+```
+
+instead of
+
+```TypeScript
+ const scrollHandler = useAnimatedScrollHandler({
+    onScroll: (e) => {
+      `worklet`;
```
tomekzaw, line 50:
```suggestion
      'worklet';
```

plugin/README.md
```diff
@@ -0,0 +1,141 @@
+### ReanimatedPlugin is now in TypeScript.
+
+To compile it, either use `yarn` or explicitly use `yarn plugin` in the root directory or `yarn` in `plugin/`.
+
+# Why do we need this plugin?
```
tomekzaw, line 5: Use h2 or h3 for subsection titles instead (`##` or `###`)

plugin/index.ts
```diff
@@ -0,0 +1,1078 @@
...
+  let hash2 = 52711;
+
+  while (i--) {
+    const char = str.charCodeAt(i);
+    hash1 = (hash1 * 33) ^ char;
+    hash2 = (hash2 * 33) ^ char;
+  }
+
+  return (hash1 >>> 0) * 4096 + (hash2 >>> 0);
+}
```
tomekzaw, line 21:
```suggestion
}

```

plugin/index.ts
```diff
@@ -0,0 +1,1078 @@
+import * as BabelCore from '@babel/core';
+import * as BabelTypes from '@babel/types';
+import generate from '@babel/generator';
+import traverse from '@babel/traverse';
```
tomekzaw, line 4: We need to add these dependencies to `devDependencies` in `package.json` (project root).

plugin/README.md
```diff
@@ -0,0 +1,141 @@
+### ReanimatedPlugin is now in TypeScript.
```
tomekzaw, line 1: Let's convert it into a document title.
```suggestion
# Reanimated Babel plugin
```

piaskowyk (approved):
I generated bundles by plugin on main and by plugin on this PR and both bundles are the same.

PR #4101, opened 2023-02-24
Hardcode JS library version in `checkVersion`

jwajgelt:
## Summary
We hardcode the version of Reanimated here in order to compare it with the version used to build the native part of the library in runtime.

Ideally, we'd grab this version from the `package.json` file instead, but this causes issues when packaging the library due to misaligned relative paths.

### DIFF
--- a/src/reanimated2/platform-specific/checkVersion.ts
+++ b/src/reanimated2/platform-specific/checkVersion.ts
@@ -1,4 +1,4 @@
-import { version as jsVersion } from '../../../package.json';
+const jsVersion = '3.0.0-rc.10';
 
 /**
  * Checks that native and js versions of reanimated match.
--- a/src/reanimated2/platform-specific/checkVersion.ts
+++ b/src/reanimated2/platform-specific/checkVersion.ts
@@ -1,3 +1,8 @@
+/**
+ * We hardcode the version of Reanimated here in order to compare it
+ * with the version used to build the native part of the library in runtime.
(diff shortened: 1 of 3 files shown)


src/reanimated2/platform-specific/checkVersion.ts
```diff
@@ -1,4 +1,9 @@
-import { version as jsVersion } from '../../../package.json';
+/**
+ * We hardcode the version of Reanimated here in order to compare it
+ * with the version used to build the native part of the library in runtime.
+ * Remember to keep this in sync with the version declared in `package.json`
```
piaskowyk, line 4: `Remember to keep this in sync with the version declared in package.json` - This is the tricky part 😁
By my experience this is not a good practices to keep something variable hardcoded in code.

piaskowyk: I would rather prefer to generate file contains Reaniated version (version.json) during package build

piaskowyk: Or we can keep it in code but we need to add CI to test it

piaskowyk: I like the last idea 😎

jwajgelt: I know, but we've tried a ton of different things, and they just refuse to work :(
We'll put a check in the CI that they're in sync, and it should be fine.

piaskowyk (approved):

PR #4103, opened 2023-02-24
[Shared Element Transition] Added borderRadius to snapshot

piaskowyk:
## Summary

This PR adds the possibility to animate border-radius during shared element transition. This feature doesn't require any JavaScrip changes apart from adding a border-radius property to allowlist. The PR adds the view's border-radius value to the style snapshot. Applying this new property to view should work out of the box without additional changes because the style updater uses UIManager. 

## Limitations
The current implementation of border-radius snapshot supports only components which inherit the following classes
- on Android - `ReactViewGroup` or `ReactImageView` - more details why is in this comment - https://github.com/software-mansion/react-native-reanimated/pull/4103#discussion_r1232373705
- on iOS - `RCTView`

## Example
<details>

```js
import React from 'react';
import {
  View,
  Button,
} from 'react-native';
import { ParamListBase } from '@react-navigation/native';
import {
  createNativeStackNavigator,
  NativeStackScreenProps,
} from '@react-navigation/native-stack';

const photo = require('./assets/image.jpg');

function Screen1({ navigation }: NativeStackScreenProps<ParamListBase>) {
  return (
    <View style={{ flex: 1 }}>
      <Button title="go back" onPress={() => navigation.navigate('Screen2')} />
      <Animated.Image
        sharedTransitionTag={'image'}
        source={photo}
        style={{ width: '100%', height: 200 }}
      />
      <Animated.View 
        style={{ width: 100, height: 100, backgroundColor: 'red', borderRadius: 10 }} 
        sharedTransitionTag={'view'}
      />
    </View>
  );
}

function Screen2({ navigation }: any) {
  return (
    <View style={{ flex: 1 }}>
      <Button title="go back" onPress={() => navigation.navigate('Screen1')} />
      <Animated.Image
        sharedTransitionTag={'image'}
        source={photo}
        style={{ width: '100%', height: 200, borderRadius: 50 }}
      />
      <Animated.View 
        style={{ width: 100, height: 100, backgroundColor: 'red', borderRadius: 50 }} 
        sharedTransitionTag={'view'}
      />
    </View>
  );
}

  return (
    <Stack.Navigator>
      <Stack.Screen
        name="Screen1"
        component={Screen1}
        options={{ headerShown: false }}
      />
      <Stack.Screen
        name="Screen2"
        component={Screen2}
        options={{ headerShown: false }}
      />
    </Stack.Navigator>
  );
}

```

</details>

## Video
https://github.com/software-mansion/react-native-reanimated/assets/36106620/688faa2d-888f-4105-ad89-b78528d9ed98


android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -347,6 +348,14 @@ public void setNewProps(
       props.remove(Snapshot.TRANSFORM_MATRIX);
     }
 
+    if (props.containsKey(Snapshot.BORDER_RADIUS)) {
+      Double borderRadius = (Double) props.get(Snapshot.BORDER_RADIUS);
+      if (borderRadius != null) {
+        ((ReactViewBackgroundDrawable)view.getBackground()).setRadius(borderRadius.floatValue());
+      }
+      props.remove(Snapshot.TRANSFORM_MATRIX);
```
jwajgelt, line 356: Shouldn't this be
```suggestion
      props.remove(Snapshot.BORDER_RADIUS);
```
?

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -284,10 +290,6 @@ - (NSDictionary *)prepareDataForLayoutAnimatingWorklet:(NSMutableDictionary *)cu
   preparedData[@"targetGlobalOriginY"] = targetValues[@"globalOriginY"];
   preparedData[@"windowWidth"] = currentValues[@"windowWidth"];
   preparedData[@"windowHeight"] = currentValues[@"windowHeight"];
-  if (currentValues[@"transformMatrix"] != nil && targetValues[@"transformMatrix"] != nil) {
-    preparedData[@"currentTransformMatrix"] = currentValues[@"transformMatrix"];
-    preparedData[@"targetTransformMatrix"] = targetValues[@"transformMatrix"];
-  }
```
jwajgelt, line 290: Why did you move `transform` from Layout Animations to SETs? Does it not work with LAs?

acollazomayer:
@piaskowyk @jwajgelt any help needed for this?

piaskowyk:
@acollazomayer sure! 😃 Maybe you know how to get view border-radius on Android? I tried many different approaches but it always returns wrong results :/

ios/LayoutReanimation/REAAnimationsManager.m
piaskowyk: I don't want to change the legacy implementation of LayoutAnimations. Larger snapshots can negative influent on Layout Animation performance. At this moment, I want to add those new values only for SET's snapshot and test if it works properly.

fukemy:
waiting for merge

fukemy:
here is my test result:
package
```
"react-native-reanimated": "github:software-mansion/react-native-reanimated#@piaskowyk/SET-add-borderRadius-to-snapshot",
```

screen1:
```

const AnimatedImage = Animated.createAnimatedComponent(FastImage)
const renderActiveUser = (user) => {
      const avtUrl = getAvatarUrl(user.Id, user.ServiceId, masterInfo, user.AvatarVersion)
        return <TouchableOpacity
            onPress={() => {
                ...
            }}
            style={{
                flexDirection: 'row',
                alignItems: 'center',
                paddingLeft: 12,
                paddingRight: 12,
                paddingBottom: 4,
                paddingTop: 4
            }}>
                <AnimatedImage
                    sharedTransitionTag={avtUrl}
                    source={{ uri: avtUrl }}
                    style={{
                        width: 56,
                        height: 56,
                        borderRadius: 28
                    }} />
            <View style={{
                flex: 1,
                justifyContent: 'center'
            }}>
                <Text
                    ellipsizeMode='tail'
                    numberOfLines={1}
                    style={{
                        paddingLeft: 10,
                        fontSize: 16,
                        fontWeight: '500',
                        color: theme.defaultText
                    }}>{user.Name}</Text>

                <Text
                    ellipsizeMode='tail'
                    numberOfLines={1}
                    style={{
                        marginTop: 5,
                        paddingLeft: 10,
                        fontSize: 13,
                        fontWeight: '400',
                        color: theme.defaultText
                    }}>{getRoomUserRoleString(user.UserType, language)}</Text>
            </View>
        </TouchableOpacity>
    }
```


screen2:

```

const AnimatedImage = Animated.createAnimatedComponent(FastImage)
const renderHeader = () => {
        const title = userInfo?.TypeInfo?.Title
        const avatar = getAvatarUrl(userId, userServiceId, masterInfo, avatarVersion)
        return <View
            style={{
                flex: 1,
                marginTop: 20,
                // alignItems: 'center',
                flexDirection: 'row'
            }}>
            <TouchableOpacity
                onPress={() => {
                    NativeModules.LightBox.openLightBoxAnimate([avatar], 'userInfoScreenAvatar')
                }}>
                <AnimatedImage
                    sharedTransitionTag={avatar}
                    source={{ uri: avatar }}
                    nativeID="userInfoScreenAvatar"
                    style={{
                        marginLeft: 16,
                        marginRight: 16,
                        width: 120,
                        height: 120,
                        borderRadius: 60,
                        backgroundColor: 'lightgray'
                    }} />
            </TouchableOpacity>
....
```

Video:


https://github.com/software-mansion/react-native-reanimated/assets/8202437/4eaed003-b339-470f-9e45-88df6f805d89

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -442,7 +442,8 @@ protected void finishSharedAnimation(int tag) {
           if (key.equals(Snapshot.TRANSFORM_MATRIX)) {
             preparedValues.put(key, value);
           } else {
-            preparedValues.put(key, (double) PixelUtil.toDIPFromPixel((int) value));
+            float pixelsValue = value instanceof Integer ? (int) value : (Float) value;
```
tomekzaw, line 445: 1. Here it's `(Float) value` but in line 274 it was `(float) value`, which one is correct?
2. Is it guaranteed that it's always either int or float?
3. Should we make a utility function out of this?

ios/LayoutReanimation/REASnapshot.m
```diff
@@ -79,6 +79,17 @@ - (void)makeSnapshotForView:(UIView *)view useAbsolutePositionOnly:(BOOL)useAbso
       // Identity matrix is an default value
       _values[@"transformMatrix"] = @[ @(1), @(0), @(0), @(0), @(1), @(0), @(0), @(0), @(1) ];
     }
+#if defined(RCT_NEW_ARCH_ENABLED) || defined(TARGET_OS_TV)
+    _values[@"borderRadius"] = @(0);
+#else
+    RCTView *viewController = (RCTView *)view;
```
tomekzaw, line 85: Why `viewController`? It's not a `UIViewController`

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -115,6 +124,21 @@ public Snapshot(View view) {
...
+      borderRadius = ((ReactViewBackgroundDrawable) view.getBackground()).getFullBorderRadius();
+    } else if (view instanceof ReactImageView) {
+      borderRadius = getImageBorderRadius((ReactImageView) view);
+    }
+  }
+
+  private float getImageBorderRadius(ReactImageView imageView) {
+    try {
+      Field field = imageView.getClass().getDeclaredField("mBorderRadius");
+      field.setAccessible(true);
```
tomekzaw, line 137: I remember that we needed to use `accessFlags` in similar cases for some reason, can you confirm the current code is sufficient?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -115,6 +124,21 @@ public Snapshot(View view) {
     }
     originXByParent = view.getLeft();
     originYByParent = view.getTop();
+    if (view.getBackground() != null) {
+      borderRadius = ((ReactViewBackgroundDrawable) view.getBackground()).getFullBorderRadius();
```
tomekzaw, line 128: 1. Should we move this logic to `getViewBorderRadius` as well?
2. Does it make sense to check `instanceof ReactViewBackgroundDrawable` before casting?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -115,6 +124,21 @@ public Snapshot(View view) {
...
+    }
+  }
+
+  private float getImageBorderRadius(ReactImageView imageView) {
+    try {
+      Field field = imageView.getClass().getDeclaredField("mBorderRadius");
+      field.setAccessible(true);
+      return field.getFloat(imageView);
+    } catch (NullPointerException | NoSuchFieldException | IllegalAccessException e) {
+      return 0;
```
tomekzaw, line 140: It makes sense to return 0 here instead of throwing some exception but can we add a comment here that explains it explicitly?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
piaskowyk: You are right 👍 I made a util function from it. Resolved in https://github.com/software-mansion/react-native-reanimated/pull/4103/commits/5195fffb379a707cca35203988aec94b42bed76d

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4103/commits/1a43a54b73c46d1e9a2686143e4c051e75af6191

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
piaskowyk: Yes, this code is sufficient. The `accessFlags` field is required when we want to get access to static field of class - https://stackoverflow.com/questions/3301635/change-private-static-final-field-using-java-reflection/64378131#64378131

ios/LayoutReanimation/REASnapshot.m
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4103/commits/2864be8d241a385596d3aa3df9ab02fa510a676d

android/src/main/java/com/swmansion/reanimated/Utils.java
```diff
@@ -35,4 +35,15 @@ public static String simplifyStringNumbersList(String list) {
     // to make usage of std::istringstream in C++ easier
     return list.replace(",", "").replace("[", "").replace("]", "");
   }
+
+  public static float parseNumberToFloat(Object value) {
+    if (value instanceof Integer) {
+      return (int) value;
```
kmagiera, line 41: Integer has `floatValue` too, so we can follow the same technique as in the case of `Double`

android/src/main/java/com/swmansion/reanimated/Utils.java
```diff
@@ -35,4 +35,15 @@ public static String simplifyStringNumbersList(String list) {
     // to make usage of std::istringstream in C++ easier
     return list.replace(",", "").replace("[", "").replace("]", "");
   }
+
+  public static float parseNumberToFloat(Object value) {
```
kmagiera, line 39: This isn't parsing, rename this, e.g. `convertToFloat`

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -115,6 +126,28 @@ public Snapshot(View view) {
...
+    borderRadius = getBorderRadius(view);
+  }
+
+  private float getBorderRadius(View view) {
+    if (view.getBackground() != null) {
+      Drawable background = view.getBackground();
+      if (background instanceof ReactViewBackgroundDrawable) {
+        return ((ReactViewBackgroundDrawable) background).getFullBorderRadius();
+      }
+    } else if (view instanceof ReactImageView) {
```
kmagiera, line 138: I wonder if instead of relying on RN internals here and above we could just use background drawable and extract corner radii using `getOutline().getRadius()`

Not a huge fan of adding this type of dependencies, especially that it only gives us support for built-in image and view components. If we need to change visibility of fields this should be done only once per class anyways, so it'd be good to try and avoid more reflection calls than necessary

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -115,6 +126,28 @@ public Snapshot(View view) {
...
+        Field field = view.getClass().getDeclaredField("mBorderRadius");
+        field.setAccessible(true);
+        Float borderRadius = field.getFloat(view);
+        return Float.isNaN(borderRadius) ? 0 : borderRadius;
+      } catch (NullPointerException | NoSuchFieldException | IllegalAccessException ignored) {
+        // In case of non-standard view is better to not support the border animation
+        // instead of throwing exception
+      }
+    }
+    Log.w("[Reanimated]", "Unable to get border radius value for view: " + view);
```
kmagiera, line 149: not sure about this warning here. It looks like we try to unconditionally extract border radius, which means people will get this warning for unsupported view types even when not trying to animate the border

kmagiera (commented):
This PR has very little description. For example, some important details are omitted like the fact it only adds support for built-in RN components of Image and View. Also, I'm not sure if this is the only part of this change as this PR has no JS changes – are this changes not required or they are in some other PR? There are also no examples being added here

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
piaskowyk: I have tried to use `getOutline().getRadius()` but it doesn’t work. The `getRadius()` always returns `-Infinity`. 

According to Android documentation:(https://developer.android.com/reference/android/graphics/Outline#getRadius())
>Returns the rounded rect radius, if set, **or a value less than 0** if a path has been set via `setPath(android.graphics.Path)`

Unfortunately React Native set border radius via `setPath()` method.  [ReactViewBackgroundDrawable .drawRoundedBackgroundWithBorders() source code](https://github.com/facebook/react-native/blob/v0.72.0-rc.6/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewBackgroundDrawable.java#L341-L346)
```java
private void drawRoundedBackgroundWithBorders(Canvas canvas) {
  updatePath(); // <-- calls setPath()
  canvas.save();
  // Clip outer border
  canvas.clipPath(mOuterClipPathForBorderRadius, Region.Op.INTERSECT);
```

Every ReactNative component uses `ReactViewBackgroundDrawable` to set border radius, except ReactImageView.

[ReactViewGroup setBorderRadius() source code](https://github.com/facebook/react-native/blob/v0.72.0-rc.6/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java#L319-L322)
```java
public void setBorderRadius(float borderRadius) {
  ReactViewBackgroundDrawable backgroundDrawable = getOrCreateReactViewBackground();
  backgroundDrawable.setRadius(borderRadius);
}
```

[ReactImageView setBorderRadius() source code](https://github.com/facebook/react-native/blob/v0.72.0-rc.6/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/image/ReactImageView.java#L261-L266)
```java
public void setBorderRadius(float borderRadius) {
  if (!FloatUtil.floatsEqual(mBorderRadius, borderRadius)) {
    mBorderRadius = borderRadius;
    mIsDirty = true;
  }
}
```

I have tried many different approaches but this one with `ReactViewBackgroundDrawable` works best for me.

---

I have reduced the number of reflection calls in 2cc22c61e68bf99a690cbff239d216a893b4f5f3

ios/LayoutReanimation/REASnapshot.m
```diff
@@ -79,6 +79,17 @@ - (void)makeSnapshotForView:(UIView *)view useAbsolutePositionOnly:(BOOL)useAbso
       // Identity matrix is an default value
       _values[@"transformMatrix"] = @[ @(1), @(0), @(0), @(0), @(1), @(0), @(0), @(0), @(1) ];
     }
+#if defined(RCT_NEW_ARCH_ENABLED) || TARGET_OS_TV
+    _values[@"borderRadius"] = @(0);
+#else
+    RCTView *reactView = (RCTView *)view;
```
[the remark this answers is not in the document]
piaskowyk: I think we need, because the `borderRadius` property comes from `RCTView` not from `UIView`.
I changed it a bit in https://github.com/software-mansion/react-native-reanimated/pull/4103/commits/8bf37c19b44d9999cc78494b228d6c5405ec9b97

ios/LayoutReanimation/REASnapshot.m
```diff
@@ -79,6 +79,17 @@ - (void)makeSnapshotForView:(UIView *)view useAbsolutePositionOnly:(BOOL)useAbso
...
+    if ([reactView respondsToSelector:@selector(borderRadius)]) {
+      // For example `RCTTextView` doesn't have `borderRadius` selector
+      _values[@"borderRadius"] = @(reactView.borderRadius);
+    } else {
+      _values[@"borderRadius"] = @(0);
+    }
+#endif
   } else {
     _values[@"originX"] = [NSNumber numberWithDouble:view.center.x - view.bounds.size.width / 2.0];
```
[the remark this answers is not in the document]
piaskowyk: Resolved in https://github.com/software-mansion/react-native-reanimated/pull/4103/commits/e00a93e5f07d06374d0ff5d1732551b4f32e24f0

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
piaskowyk: Resolved in https://github.com/software-mansion/react-native-reanimated/pull/4103/commits/8d40ceced834f3b82b8fab7cc43c5828cabf77ea

android/src/main/java/com/swmansion/reanimated/Utils.java
piaskowyk: Resolved in https://github.com/software-mansion/react-native-reanimated/pull/4103/commits/f0db7d894e6d402e8cb639f8e7fc52b108b65d95

android/src/main/java/com/swmansion/reanimated/Utils.java
piaskowyk: Resolved in https://github.com/software-mansion/react-native-reanimated/pull/4103/commits/f0db7d894e6d402e8cb639f8e7fc52b108b65d95

android/src/main/java/com/swmansion/reanimated/Utils.java
```diff
@@ -35,4 +35,15 @@ public static String simplifyStringNumbersList(String list) {
     // to make usage of std::istringstream in C++ easier
     return list.replace(",", "").replace("[", "").replace("]", "");
   }
+
+  public static float convertToFloat(Object value) {
+    if (value instanceof Integer) {
+      return ((Integer) value).floatValue();
+    } else if (value instanceof Float) {
+      return (float) value;
```
tomekzaw, line 43: Why is this case not consistent with the rest? Shouldn't it be this way?
```suggestion
      return ((Float) value).floatValue();
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -115,6 +127,29 @@ public Snapshot(View view) {
...
+        return ((ReactViewBackgroundDrawable) background).getFullBorderRadius();
+      }
+    } else if (view instanceof ReactImageView) {
+      try {
+        if (mBorderRadiusField == null) {
+          mBorderRadiusField = ReactImageView.class.getDeclaredField("mBorderRadius");
+          mBorderRadiusField.setAccessible(true);
+        }
+        float borderRadius = mBorderRadiusField.getFloat(view);
+        return Float.isNaN(borderRadius) ? 0 : borderRadius;
```
tomekzaw, line 146:
```suggestion
        if (Float.isNaN(borderRadius)) {
          return 0;
        }
        return borderRadius;
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -270,7 +271,9 @@ public HashMap<String, Object> prepareDataForAnimationWorklet(
       keys = Snapshot.currentKeysToTransform;
     }
     for (String key : keys) {
-      preparedValues.put(key, PixelUtil.toDIPFromPixel((int) values.get(key)));
+      Object value = values.get(key);
+      float pixelsValue = Utils.convertToFloat(value);
+      preparedValues.put(key, PixelUtil.toDIPFromPixel(pixelsValue));
```
tomekzaw, line 276:
```suggestion
      float dipValue = PixelUtil.toDIPFromPixel(pixelsValue);
      preparedValues.put(key, dipValue);
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -442,7 +443,8 @@ protected void finishSharedAnimation(int tag) {
           if (key.equals(Snapshot.TRANSFORM_MATRIX)) {
             preparedValues.put(key, value);
           } else {
-            preparedValues.put(key, (double) PixelUtil.toDIPFromPixel((int) value));
+            float pixelsValue = Utils.convertToFloat(value);
+            preparedValues.put(key, (double) PixelUtil.toDIPFromPixel(pixelsValue));
```
tomekzaw, line 447:
```suggestion
            float dipValue = PixelUtil.toDIPFromPixel(pixelsValue);
            preparedValues.put(key, (double) dipValue);
```
BTW why do we need to cast it to double?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -115,6 +127,29 @@ public Snapshot(View view) {
     }
     originXByParent = view.getLeft();
     originYByParent = view.getTop();
+    borderRadius = getBorderRadius(view);
+  }
+
+  private float getBorderRadius(View view) {
+    if (view.getBackground() != null) {
+      Drawable background = view.getBackground();
+      if (background instanceof ReactViewBackgroundDrawable) {
```
tomekzaw, line 136: If this check fails, we'll effectively return 0. Do you think it's possible that `background instanceof ReactViewBackgroundDrawable` is false but `view instanceof ReactImageView` is true? In such case, we could change `else if` to `if` below.

android/src/main/java/com/swmansion/reanimated/Utils.java
piaskowyk: IDE suggest to avoid it:
image

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -115,6 +127,32 @@ public Snapshot(View view) {
     }
     originXByParent = view.getLeft();
     originYByParent = view.getTop();
+    borderRadius = getBorderRadius(view);
+  }
+
+  private float getBorderRadius(View view) {
```
kmagiera, line 133: feels like this could be a static method extracted to some utils class. It has a clear coupling with RN internals so extracting it out may allow us to provide separate implementations in the future if the internals changes

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -629,4 +628,16 @@ - (void)enableCleaningForViewTag:(NSNumber *)viewTag
   }
 }
 
+- (NSDictionary *)prepareDataForWorklet:(NSMutableDictionary *)currentValues
+                           targetValues:(NSMutableDictionary *)targetValues
+{
+  NSMutableDictionary *workletValues = [_animationManager prepareDataForLayoutAnimatingWorklet:currentValues
```
kmagiera, line 634: I wonder what would happen if we'd fill transform and border radius data for all layout animations. Would it suffice for regular layout animations to support these properties for layout update animations?

kmagiera (approved):
Looks good overall. Left some minor comments inline

ios/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: Yes, it should be enough

tomekzaw (approved):

PR #4109, opened 2023-02-26
Minor documentation fixes

Kwasow:
## Summary

I found a few minor mistakes in the docs while looking through it, so I fixed them for you :))

These include:
- fixed spelling mistakes
- changes spelling of `usecases` to `use cases` - I've given this some thought and it seems that all three (`use cases`, `use-cases` and `usecases`) spellings are used, but [Wiktionary](https://en.wiktionary.org/wiki/use_case) suggests `use case` as the most common one and the spelling `usecase` is much less popular than `use case` when searching occurrences on the web - I'll be happy to change it back if you prefer the old spelling
- improved the LayoutAnimations tutorial


tomekzaw (approved):
Thanks @Kwasow for the PR! Glad to have you in the team 🚀

PR #4111, opened 2023-02-26
Bump react-native version in V8 CI

tomekzaw:
## Summary

This PR bumps `react-native` version from 0.70.5 to 0.71.3 because of the failing CI after the successful release of `react-native-v8@2.0` which requires `react-native>=0.71.0`.

## Test plan

Just see if the V8 CI is green.


kmagiera (approved):
🐴

PR #4116, opened 2023-02-27
Cleanup NativeProxy on Android

jwajgelt:

### DIFF
--- a/android/src/main/cpp/NativeProxy.cpp
+++ b/android/src/main/cpp/NativeProxy.cpp
@@ -411,9 +411,7 @@ PlatformDepMethodsHolder NativeProxy::getPlatformDependentMethods() {
 
 #ifdef RCT_NEW_ARCH_ENABLED
   auto synchronouslyUpdateUIPropsFunction =
-      [this](jsi::Runtime &rt, Tag tag, const jsi::Value &props) {
-        this->synchronouslyUpdateUIProps(rt, tag, props);
-      };
+      bindThis(&NativeProxy::synchronouslyUpdateUIProps);
 #else
   auto configurePropsFunction = bindThis(&NativeProxy::configureProps);
 #endif
--- a/android/src/main/cpp/NativeProxy.cpp
+++ b/android/src/main/cpp/NativeProxy.cpp
@@ -180,7 +180,7 @@ void NativeProxy::registerNatives() {
 void NativeProxy::requestRender(
     std::function<void(double)> onRender,
     jsi::Runtime &rt) {
-  static auto method =
+  static const auto method =
       getJniMethod<void(AnimationFrameCallback::javaobject)>("requestRender");
   method(
       javaPart_.get(),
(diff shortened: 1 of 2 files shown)


android/src/main/cpp/NativeProxy.cpp
```diff
@@ -160,43 +140,22 @@ void NativeProxy::installJSIBindings(
         jni::make_jstring(propName.utf8(rt).c_str());
     auto result = method(javaPart_.get(), viewTag, propNameJStr);
     std::string str = result->toStdString();
-    return jsi::Value(rt, jsi::String::createFromAscii(rt, str.c_str()));
+    return {rt, jsi::String::createFromAscii(rt, str)};
```
tomekzaw, line 143: plz don't

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -334,10 +252,16 @@ void NativeProxy::installJSIBindings(
 
   std::weak_ptr<ErrorHandler> weakErrorHandler = errorHandler;
 
-  layoutAnimations->cthis()->setAnimationStartingBlock(
-      [wrt, weakModule, weakErrorHandler](
+  layoutAnimations_->cthis()->setAnimationStartingBlock(
+      [weakModule](
           int tag, int type, alias_ref<JMap<jstring, jstring>> values) {
-        auto &rt = *wrt.lock();
+        auto reaModule = weakModule.lock();
```
tomekzaw, line 258:
```suggestion
        auto module = weakModule.lock();
```

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -334,10 +252,16 @@ void NativeProxy::installJSIBindings(
...
+        if (reaModule == nullptr) {
+          return;
+        }
+        auto &rt = *reaModule->runtime;
```
tomekzaw, line 262:
```suggestion
        jsi::Runtime &rt = *reaModule->runtime;
```

android/src/main/cpp/NativeProxy.cpp
jwajgelt: honest mistake, will change this

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -584,9 +324,233 @@ int NativeProxy::subscribeForKeyboardEvents(
...
+}
+
+void NativeProxy::setupLayoutAnimations() {
+  auto weakModule =
+      std::weak_ptr<NativeReanimatedModule>(nativeReanimatedModule_);
+
+  layoutAnimations_->cthis()->setAnimationStartingBlock(
+      [weakModule](
+          int tag, int type, alias_ref<JMap<jstring, jstring>> values) {
+        auto reaModule = weakModule.lock();
```
tomekzaw, line 479: I know that `module` will be a keyword in C++20 but it's also used in various many places so for now let's be consistent and use `module` here and some time in the future we will rename it everywhere.
```suggestion
        auto module = weakModule.lock();
```

jwajgelt: I'm pretty sure C++20 shouldn't break `module` being used as an identifier anyway, but let's not get into that. 😛

piaskowyk (approved):

tomekzaw (approved):

PR #4117, opened 2023-02-27
Add shareable HostFunction

tomekzaw:
## Summary

This PR adds support for passing JSI HostFunctions to worklet runtime as a follow-up to #4024 which makes it possible for JSI HostObjects.

Previously, it was possible only to pass HostObjects:

```tsx
import { TurboModuleRegistry } from 'react-native';

  const Clipboard = TurboModuleRegistry.getEnforcing('Clipboard');
  Clipboard.setString('hello world');

  runOnUI(() => {
    'worklet';
    console.log(Clipboard);
    Clipboard.getString().then(console.log);
  })();

  return null;
}
```

This PR makes it possible to pass HostFunctions as well:

```tsx

  const turboModuleProxy = global.__turboModuleProxy;
  console.log(turboModuleProxy);

  runOnUI(() => {
    'worklet';
    console.log(turboModuleProxy.toString());
    const Clipboard = turboModuleProxy('Clipboard');
    console.log(Clipboard);
    Clipboard.setString('hello world');
    Clipboard.getString().then(console.log);
  })();

  return null;
}
```

Output:

## Caution

This is a breaking change since previously it was still possible to call such HostFunctions on the main JS runtime from worklets using `runOnJS`, however after this change it will throw the following error:

> Incompatible object passed to scheduleOnJS. It is only allowed to schedule functions defined on the React Native JS runtime this way.

The workaround is to wrap the HostFunction inside a regular JS function and pass the wrapper to the worklet, as explained in https://github.com/software-mansion/react-native-reanimated/pull/4117#issuecomment-1488820542.

Also, some HostFunctions may be unsafe to be called from two threads at once without proper synchronization, use it wisely.

Also, `ShareableHostFunction` constructor accepts `jsi::Function` but you shouldn't pass regular JS functions to it.

## Test plan

Run the attached demo in FabricExample.

### DIFF
--- a/Common/cpp/SharedItems/Shareables.h
+++ b/Common/cpp/SharedItems/Shareables.h
@@ -281,7 +281,9 @@ class ShareableHostFunction : public Shareable {
       jsi::Runtime &rt,
       jsi::Function function)
       : Shareable(HostFunctionType),
-        hostFunction_(function.getHostFunction(rt)),
+        hostFunction_(
+            (assert(function.isHostFunction(rt)),
+             function.getHostFunction(rt))),
         name_(function.getProperty(rt, "name").asString(rt).utf8(rt)),
         paramCount_(function.getProperty(rt, "length").asNumber()) {}
 
--- a/Common/cpp/SharedItems/Shareables.h
+++ b/Common/cpp/SharedItems/Shareables.h
@@ -116,6 +116,7 @@ class Shareable {
     HandleType,
     SynchronizedDataHolder,
     HostObjectType,
+    HostFunctionType,
   };
 
   explicit Shareable(ValueType valueType) : valueType_(valueType) {}
@@ -271,6 +272,22 @@ class ShareableHostObject : public Shareable {
 
  protected:
   std::shared_ptr<jsi::HostObject> hostObject_;
+};
+
+class ShareableHostFunction : public Shareable {
+ public:
+  ShareableHostFunction(
+      const std::shared_ptr<JSRuntimeHelper> &runtimeHelper,
+      jsi::Runtime &rt,
+      jsi::HostFunctionType hostFunction)
+      : Shareable(HostFunctionType), hostFunction_(hostFunction) {}
+  jsi::Value toJSValue(jsi::Runtime &rt) override {
+    return jsi::Function::createFromHostFunction(
+        rt, jsi::PropNameID::forAscii(rt, "HostFunction"), 0, hostFunction_);
+  }
+
+ protected:
+  jsi::HostFunctionType hostFunction_;
 };
 
 class ShareableWorklet : public ShareableObject {
--- a/Common/cpp/SharedItems/Shareables.h
+++ b/Common/cpp/SharedItems/Shareables.h
@@ -289,7 +289,7 @@ class ShareableHostFunction : public Shareable {
 
   jsi::Value toJSValue(jsi::Runtime &rt) override {
     return jsi::Function::createFromHostFunction(
-        rt, jsi::PropNameID::forAscii(rt, name_), paramCount_, hostFunction_);
+        rt, jsi::PropNameID::forUtf8(rt, name_), paramCount_, hostFunction_);
   }
 
  protected:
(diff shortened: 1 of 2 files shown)


Common/cpp/SharedItems/Shareables.h
```diff
@@ -273,6 +274,30 @@ class ShareableHostObject : public Shareable {
...
+        hostFunction_(
+            (assert(function.isHostFunction(rt)),
+             function.getHostFunction(rt))),
+        name_(function.getProperty(rt, "name").asString(rt).utf8(rt)),
+        paramCount_(function.getProperty(rt, "length").asNumber()) {}
+
+  jsi::Value toJSValue(jsi::Runtime &rt) override {
+    return jsi::Function::createFromHostFunction(
+        rt, jsi::PropNameID::forUtf8(rt, name_), paramCount_, hostFunction_);
+  }
```
jwajgelt, line 293: I'm not sure what the overhead from `createFromHostFunction` is, but we should probably look into caching the results from `toJSValue`.

jwajgelt (approved):
Looks good to me. 👍

Common/cpp/SharedItems/Shareables.h
tomekzaw: I guess this might be already covered by the shareable cache:
https://github.com/software-mansion/react-native-reanimated/blob/a38678d6ada894f4c42461de806a90d2a94a688a/src/reanimated2/shareables.ts#L49-L54

kmagiera (approved):
This looks good. One thing to note is that this is a breaking change, however I don't think that this can realistically affect anyone. The behavior that this diff is changing is that previously you could capture host function in a worklet and then call it using runOnJS, while after this change it will no longer be possible to call `runOnJS` with such a function as it will always be transformed to a host function on the UI runtime

tomekzaw:
> it will no longer be possible to call `runOnJS` with such a function

@kmagiera You're right, the following example works on main but throws an error after this change:

```tsx
import { runOnJS, runOnUI } from 'react-native-reanimated';

  const func = global.__blobCollectorProvider;
  console.log(_WORKLET, func.toString());

  runOnUI(() => {
    'worklet';
    console.log(_WORKLET, func.toString());
    runOnJS(func)('123');
  })();

  return null;
}
```

```
 ERROR  ReanimatedError: Exception in HostFunction: Incompatible object passed to scheduleOnJS.
It is only allowed to schedule functions defined on the React Native JS runtime this way., js engine: reanimated
```

The root cause of this error is that `scheduleOnJS` expects `ShareableRemoteFunction` but receives `ShareableHostFunction`:

https://github.com/software-mansion/react-native-reanimated/blob/62fef6f657d638de5dcc80e13f420593eee9aea7/Common/cpp/NativeModules/NativeReanimatedModule.cpp#L73-L76

As a workaround, you can wrap the HostFunction inside a regular JS function so that the wrapper gets copied into the worklet as a remote function and you can invoke the wrapper using `runOnJS`:

```diff
-  const func = global.__blobCollectorProvider;
+  const func = (...args) => global.__blobCollectorProvider(...args);
```

tomekzaw:
@kmagiera I was able to overcome this error by introducing the following change in `extractShareableOrThrow`:

```diff
 std::shared_ptr<Shareable> extractShareableOrThrow(
     jsi::Runtime &rt,
     const jsi::Value &maybeShareableValue,
     const char *errorMessage) {
   if (maybeShareableValue.isObject()) {
     auto object = maybeShareableValue.asObject(rt);
     if (object.isHostObject<ShareableJSRef>(rt)) {
       return object.getHostObject<ShareableJSRef>(rt)->value();
     }
+    if (object.isFunction(rt)) {
+      auto function = object.asFunction(rt);
+      if (function.isHostFunction(rt)) {
+        return std::make_shared<ShareableHostFunction>(rt, std::move(function));
+      }
+    }
   } else if (maybeShareableValue.isUndefined()) {
     return Shareable::undefined();
   }
   throw std::runtime_error(
       errorMessage != nullptr
           ? errorMessage
-          : "expecting the object to be of type ShareableJSRef");
+          : "expecting the object to be of type ShareableJSRef or ShareableHostFunction");
}
```

Then in `NativeReanimatedModule::scheduleOnJS` you need to use `extractShareableOrThrow` without the specialization:

```diff
-    auto shareableRemoteFun = extractShareableOrThrow<ShareableRemoteFunction>(
+    auto shareableFun = extractShareableOrThrow(
```

I'm not sure if this approach is correct and compliant with your idea on `extractShareableOrThrow`. Please let me know what you think.

PR #4119, opened 2023-02-27
Update dependecies to fix Server-Side Request Forgery

mstach60161:
## Summary

This PR fixes 
- https://github.com/software-mansion/react-native-reanimated/security/dependabot/132

by updating:
- parse-url
- and needed dependencies

## Test plan
- Tested example app


piaskowyk:
At first we need to fix all CI tests

PR #4128, opened 2023-02-28
Guard `isSharedValue` against nullish values

### LINKED ISSUE #4129: null dereference when Animated view has a null child
### Description

This check
```
export function isSharedValue<T>(value: any): value is SharedValue<T> {
  'worklet';
  return typeof value === 'object' && value._isReanimatedSharedValue === true;
}
```
https://github.com/software-mansion/react-native-reanimated/blob/a073f212e8e4e9c7c1e229d10fe5b21d3d980bef/src/reanimated2/utils.ts#L33

fails as `typeof null` is `object', then its immediately dereferenced so
```
<Animated.View>{maybeNull}</Animated.View> 
```

will crash out

### Steps to reproduce

Have an animated view around a maybe null child

N/A

### Reanimated version

3.0.0

### React Native version

0.70.7

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

gvarandas:
## Summary

Fixes #4129.

This PR expands the guard for the `isSharedValue` util to rule out nullish values (`null` and `undefined`).

This was necessary because `typeof null` returns `object` (some JS shenanigans) and the current implementation would try to access the `_isReanimatedSharedValue` property on that `null` value, causing an exception to be raised (and possibly crashing the app).

## Test plan

```ts
// before
isSharedValue(null) // raises an Exception

// after
isSharedValue(null) // returns `false`
```

### DIFF
--- a/src/reanimated2/utils.ts
+++ b/src/reanimated2/utils.ts
@@ -30,5 +30,5 @@ export function getRelativeCoords(
 
 export function isSharedValue<T>(value: any): value is SharedValue<T> {
   'worklet';
-  return typeof value === 'object' && value._isReanimatedSharedValue === true;
+  return value?._isReanimatedSharedValue === true;
 }


src/reanimated2/utils.ts
```diff
@@ -30,5 +30,5 @@ export function getRelativeCoords(
 
 export function isSharedValue<T>(value: any): value is SharedValue<T> {
   'worklet';
-  return typeof value === 'object' && value._isReanimatedSharedValue === true;
+  return value?._isReanimatedSharedValue === true ?? false;
```
LeviWilliams, line 33: The `?? false` is redundant, you can simply omit that

gvarandas: Oh, true! 👍

hamdij0maa:
any fix coming soon please @tomekzaw

LeviWilliams (approved):

piaskowyk (approved):

PR #4131, opened 2023-02-28
Guard `createAnimatedComponent` against empty style prop

computerjazz:
## Summary

Rendering an `Animated.View` with no styles or an `undefined` value in a styles array was crashing (eg. a view might have a layout animation without any styles)
fixes https://github.com/software-mansion/react-native-reanimated/issues/4139

Screen Shot 2023-02-28 at 14 02 07

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -120,6 +120,7 @@ function isInlineStyleTransform(transform: any): boolean {
 }
 
 function hasInlineStyles(style: StyleProps): boolean {
+  if (!style) return false
   return Object.keys(style).some((key) => {
     const styleValue = style[key];
     return (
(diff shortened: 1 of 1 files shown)


anhquan291:
Facing the same issue after upgrading from v2 to v3

tomekzaw:
Hey @computerjazz, thanks for the PR! Can you please run `yarn format` and commit the changes in order to fix the CI?

src/createAnimatedComponent.tsx
```diff
@@ -139,6 +140,7 @@ function extractSharedValuesMapFromProps(
     if (key === 'style') {
       const styles = flattenArray<StyleProps>(props.style ?? []);
       styles.forEach((style) => {
+        if (!style) return
```
Titozzz, line 143: We should also protect when key !== "style", for example we crashed line L154 with testID={null}

Titozzz (changes requested):
We should protect against more than just style keys (see comment above)

vanenshi:
If merged, this is going to fix the https://github.com/software-mansion/react-native-reanimated/issues/4139
I tried this on my project and it fix the problem
also this is happening on other third party libraries, like this https://github.com/gorhom/react-native-bottom-sheet/issues/1307

src/createAnimatedComponent.tsx
computerjazz: I think this case is covered by https://github.com/software-mansion/react-native-reanimated/pull/4128/files

tomekzaw:
Closing in favor of #4159

PR #4136, opened 2023-03-01
docs: update mock definition in Testing with Jest guide

yerevin:
## Summary

It is a fix for misleading information in `Testing with Jest` guide related to left Reanimatedv2 mock definition which is no longer available in Reanimatedv3.

## Test plan

I was running into error
```
  ● Test suite failed to run

    Cannot find module 'react-native-reanimated/lib/reanimated2/jestUtils' from 'tests/setupTests.ts'
```

After replacing   
```
require("react-native-reanimated/lib/reanimated2/jestUtils").setUpTests();
```
with
```
jest.mock("react-native-reanimated", () =>
  require("react-native-reanimated/mock")
);
```
I got rid of that error

PR #4143, opened 2023-03-01
Update testing.md for jest testing

TheRogue76:
## Summary
As of Reanimated 3, the code under lib is now in both `commonjs` and `module` formats. This PR simply updates the import path for those jest mocks

## Test plan
Tested out the `commonjs` mocks in a current project after upgrading to reanimated 3 and validated that tests pass


TheRogue76:
Closed in favour of https://github.com/software-mansion/react-native-reanimated/pull/4136

PR #4145, opened 2023-03-01
update the blog post link in the docs

TheRogue76:
## Summary

Updating the blog post link in the landing page from reanimated 2's announcement to reanimated 3

## Test plan
None


tomekzaw:
Thanks for the PR @TheRogue76, good catch but we intentionally left the link to the original blog post in the docs. The blog post "Introducing Reanimated 2" is a great starting point for library users whereas "Releasing Reanimated 3" is probably less informative for a beginner (it's more like how to upgrade from v2 to v3).

Issue #4139, opened 2023-03-01
[V3] TypeError: Cannot convert undefined value to object 


vanenshi:
### Description

when I am passing an array of styles in `Animated.View` and one of them is undefined, the `hasInlineStyles` function throw error.
The only thing I did was update the library to version 3.0.0

### Steps to reproduce

1. make a simple animate view with an array of styles, and make one of them `undefined` 
```jsx
const styleToOverwrite = undefined

<Animated.View style={[baseStyle, styleToOverwrite]}>
  ...
</Animated.View>
```

### Snack or a link to a repository

expo not suporting the reanimate 3.0.0

### Reanimated version

3.0.0

### React Native version

0.70.7

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

IPhone 14 

### Acknowledgements

Yes

dmahajan980:
I think this will be fixed after https://github.com/software-mansion/react-native-reanimated/pull/4128 is merged.

Jawish-a:
same issue here, using react navigation.



`ERROR  TypeError: Cannot convert undefined value to object

This error is located at:
    in AnimatedComponent(View)
    in Unknown (created by Drawer)
    in RCTView (created by View)
    in View (created by AnimatedComponent(View))
    in AnimatedComponent(View)
    in Unknown (created by PanGestureHandler)
    in PanGestureHandler (created by Drawer)
    in Drawer (created by DrawerViewBase)
    in DrawerViewBase (created by DrawerView)
    in RCTView (created by View)
    in View (created by GestureHandlerRootView)
    in GestureHandlerRootView (created by DrawerView)
    in RNCSafeAreaProvider (created by SafeAreaProvider)
    in SafeAreaProvider (created by SafeAreaInsetsContext)
    in SafeAreaProviderCompat (created by DrawerView)
    in DrawerView (created by DrawerNavigator)
    in PreventRemoveProvider (created by NavigationContent)
    in NavigationContent
    in Unknown (created by DrawerNavigator)
    in DrawerNavigator (created by MainStack)
    in MainStack (created by Main)
    in Main (created by App)
    in Suspense (created by App)
    in QueryClientProvider (created by App)
    in EnsureSingleNavigator
    in BaseNavigationContainer
    in ThemeProvider
    in NavigationContainerInner (created by App)
[...]

PR #4149, opened 2023-03-02
fix for TypeError: Cannot convert undefined value to object in create…

Jawish-a:
…AnimatedComponent

## Summary

after upgrading to Reanimated 3.0, some developer faced an issue with ` ERROR  TypeError: Cannot convert undefined value to object`
issue #4139 

## Test plan

none

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -120,7 +120,7 @@ function isInlineStyleTransform(transform: any): boolean {
 }
 
 function hasInlineStyles(style: StyleProps): boolean {
-  return Object.keys(style).some((key) => {
+  return Object.keys(style ?? {}).some((key) => {
     const styleValue = style[key];
     return (
       isSharedValue(styleValue) ||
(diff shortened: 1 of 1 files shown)


Jawish-a:
a patch for those who need at till this one gets merged or fixed by the maintaners

```
diff --git a/node_modules/react-native-reanimated/src/createAnimatedComponent.tsx b/node_modules/react-native-reanimated/src/createAnimatedComponent.tsx
index 6a94382..4835f2e 100644
--- a/node_modules/react-native-reanimated/src/createAnimatedComponent.tsx
+++ b/node_modules/react-native-reanimated/src/createAnimatedComponent.tsx
@@ -120,7 +120,7 @@ function isInlineStyleTransform(transform: any): boolean {
 }
 
 function hasInlineStyles(style: StyleProps): boolean {
-  return Object.keys(style).some((key) => {
+  return Object.keys(style ?? {}).some((key) => {
     const styleValue = style[key];
     return (
       isSharedValue(styleValue) ||
@@ -139,7 +139,7 @@ function extractSharedValuesMapFromProps(
     if (key === 'style') {
       const styles = flattenArray<StyleProps>(props.style ?? []);
       styles.forEach((style) => {
-        for (const [key, styleValue] of Object.entries(style)) {
+        for (const [key, styleValue] of Object.entries(style ?? {})) {
           if (isSharedValue(styleValue)) {
             inlineProps[key] = styleValue;
           } else if (

```

tomekzaw:
Thanks for this PR @Jawish-a. Closing in favor of #4159.

PR #4155, opened 2023-03-03
style might be undefined at this point

bmatasar:
style might be undefined at this point and `Object.keys(style)` would throw an exception

## Summary
I use a third party library and after upgrading to reanimated 3.0.0 I get the exception at line 123. 

## Test plan
I don't have a simple test case but checking that style is defined does not hurt

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -648,7 +648,7 @@ export default function createAnimatedComponent(
                 };
               }
               return this.initialStyle;
-            } else if (hasInlineStyles(style)) {
+            } else if (style && hasInlineStyles(style)) {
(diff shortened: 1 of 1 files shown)


jwajgelt:
Hi @bmatasar
Thanks for the PR! Your solution looks good to me, but I think it would be better to move the check inside the `hasInlineStyles` function instead, so it doesn't throw an exception in case it's called with an `undefined` value elsewhere.
Would you be able to make that change to your PR?

bmatasar:
I moved the check in `hasInlineStyles`. I do not know what your preferred coding style is, so I did it the plain old way :)

tomekzaw (approved):
Thanks @bmatasar for this PR!

PR #4179, opened 2023-03-07
Add support for spring duration

Latropos:
## Summary

We have decided to introduce two new config props - damping ratio and duration to make usage of `withSpring` easier.

```mermaid
  graph LR;
      A[Use<br>withSpring<br>animation] --> B;
      B{Did user provide<br>config with duration or<br>dampingRatio?} --YES-->C[Calculate spring movement using default or provided<br>mass and damping. Defaults:<br>damping 10, mass 1];
      B--NO-->D[Calculate spring movement using default or provided<br>duration and damping ratio. Defaults:<br>duration: 2, dampingRatio: 0.5];
```

### Implementation
When starting an animation with predefined duration we have to find a mass that would give us the expected duration. Thus we have to solve a simple numeric equation. We do it with bisection, since Newton method appears to be much slower in this case (it happens, when finding roots of functions that grow really fast 📈)

<details><summary>Testing that Newton's method is much slower</summary>

https://user-images.githubusercontent.com/56199675/233330812-5eb31550-f77d-4b68-8765-5131ba18c2d1.mov
</details> 

## Recordings
In all the examples below duration is set to 5000ms

<table>
<tr><td>dampingRatio=1</td><td>dampingRatio=0.5</td></tr>
<tr><td>

https://user-images.githubusercontent.com/56199675/233336875-b2666629-f3e9-403e-b629-fd44efba0dea.mov</td><td>

https://user-images.githubusercontent.com/56199675/233337127-433077d3-2017-4fde-8c1e-647c0ab52600.mov

</td></tr>
<tr><td>dampingRatio=1000</td><td>dampingRatio=0.1</td></tr>
<tr><td>

https://user-images.githubusercontent.com/56199675/233337155-a8f07c34-d65e-4161-ba47-b8fc9992482a.mov

</td><td>

https://user-images.githubusercontent.com/56199675/233337200-67b1719e-5963-4e65-b187-551cf80966cf.mov

</td></tr>
</table>

## Test plan

Run demo


src/reanimated2/animation/spring.ts
```diff
@@ -144,12 +148,35 @@ export function withSpring(
...
       animation.current = value;
+
+      let frameRatio;
+
+      if (config.duration) {
+        const c = config.damping;
+        const m = config.mass;
+        const amplitude = Number(animation.toValue) - animation.startValue;
+
+        const expectedDuration =
```
piaskowyk, line 159: I think that we should add link to paper here. Link: https://phys.libretexts.org/Bookshelves/University_Physics/Book%3A_University_Physics_(OpenStax)/Book%3A_University_Physics_I_-_Mechanics_Sound_Oscillations_and_Waves_(OpenStax)/15%3A_Oscillations/15.06%3A_Damped_Oscillations

src/reanimated2/animation/spring.ts
```diff
@@ -161,6 +188,8 @@ export function withSpring(
       current: toValue,
       callback,
       lastTimestamp: 0,
+      startValue: 0,
+      startTime: 0,
```
piaskowyk, line 192: I don't see the usage of `startTime` parameter, is it necessary?

piaskowyk (approved):
For me is great! But let's wait with merge since kmagiera opinion.

piaskowyk (changes requested):
Oh, we forgot about typescript and about adding proper types to d.ts

react-native-reanimated.d.ts
```diff
@@ -409,6 +409,7 @@ declare module 'react-native-reanimated' {
     restSpeedThreshold?: number;
     restDisplacementThreshold?: number;
     velocity?: number;
+    duration?: number;
```
Latropos, line 412: @piaskowyk Do we have to add some more types?

kmagiera:
Discussed that with @piaskowyk today but wanted to post the summary anyways;
1) We should allow for a few different configurations to work: a) duraiton + damping b) duration + stiffness c) just duration with some sensible default (preferably we should treat if someone has provided the default damping). The difference is that now we always override damping, so if user passes a damping parameter it'd get overridden and result in unexpected behavior. Note that damping is a parameter that's intuitive to use (e.g. critical damping results in no overshooting, underdamping results in the oscillator making swings and there are more swings the lower the parameter value).
2) We should always make sure that if duration is provided the animation always finishes at that specified time. Currently we use position and velocity thresholds for finishing the animation but that may result in the time being a bit off dompared to the provided duration. For the case when duration is provided we should use a different condition and only look at elapsed time (similar to what we do in timing animations)

src/reanimated2/animation/spring.ts
```diff
@@ -38,34 +55,43 @@ export function withSpring(
...
+      }
 
       const deltaTime = Math.min(now - lastTimestamp, 64);
       animation.lastTimestamp = now;
 
-      const c = config.damping;
-      const m = config.mass;
+      const c =
+        config.duration && !userConfig?.damping
+          ? animation.newDamping
```
kmagiera, line 89: would it be possible to only make this check when the animation is created instead of doing it in every frame. In that case we'd just use `animation.damping` here instead of having both `config.damping` and `animation.newDamping` and only set that `animation.damping` at the start (it can't really be changed in-flight).

Same applies to mass

src/reanimated2/animation/spring.ts
```diff
@@ -144,12 +170,49 @@ export function withSpring(
       previousAnimation: SpringAnimation
     ): void {
       animation.current = value;
+
+      let newDamping = config.duration;
+      let newMass = config.mass;
+
+      if (userConfig?.duration && userConfig?.damping && userConfig?.mass) {
+        console.warn(
+          "You've specified to all the parameters: damping, mass and duration. At least one of them must be calculated depending on the two others"
```
kmagiera, line 179:
```suggestion
          "You've specified all the parameters: damping, mass and duration. At least one of them must be calculated depending on the two others"
```

kmagiera (commented):
Is it ok that we don't take velocity into account when calculating damping/mass?

src/reanimated2/animation/spring.ts
```diff
@@ -93,14 +121,56 @@ export function withSpring(
...
+      const { toValue, firstTimestamp, lastTimestamp, current, velocity } =
+        animation;
+
+      const timeFromStart = now - firstTimestamp;
+      if (userConfig?.duration && timeFromStart > userConfig.duration) {
+        animation.current = toValue;
+
+        // clear lastTimestamp to avoid using stale value by the next spring animation that starts after this one
+        animation.lastTimestamp = 0;
+        animation.firstTimestamp = 0;
```
kmagiera, line 137: do we need to reset `firstTimestamp` here too? The comment above only mentions `lastTimestamp`

kmagiera (commented):
I wonder why we limit duration-based animation to underdamped case only (over-damped seem to have an easier mechanics). On top of that it looks like we are not handling the case of overriding an ongoing spring animation with duration. I'd suggest that you try and replace this line https://github.com/software-mansion/react-native-reanimated/blob/main/Example/src/AnimatedStyleUpdateExample.tsx#L20 with spring animation with duration – then when you tap the button while the animation is still running, in case of timing it will only change the destination but it will finish at the designated time. With spring animation it seem like the new animation will start from new point and on top of that velocity will be reset, so there will be no smooth transition between animation from one target to the other?

src/reanimated2/animation/spring.ts
```diff
@@ -144,13 +210,76 @@ export function withSpring(
...
+      let newMass = config.mass;
+
+      if (userConfig?.duration && userConfig?.damping && userConfig?.mass) {
+        console.warn(
+          "You've specified all the parameters: damping, mass and duration. At least one of them must be calculated depending on the two others"
+        );
+      }
+
+      if (userConfig?.duration) {
+        const x0 = Number(animation.toValue) - value;
```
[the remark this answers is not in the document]
Latropos: I've taken recordings of both approaches (and I don't really see any difference ;-) )
<table>
<tr><td>Current</td><td>With previous start value as value</td></tr>
<tr><td>

https://user-images.githubusercontent.com/56199675/227509874-a2f18430-168f-4a70-90f4-d8a59ead787f.mov

</td><td>

https://user-images.githubusercontent.com/56199675/227509894-655a038a-37fe-4765-99b9-0d7498c07835.mov

</td></tr>
</table>

Example/src/AnimatedStyleUpdateExample.tsx
```diff
@@ -29,14 +33,31 @@ function AnimatedStyleUpdateExample(): React.ReactElement {
...
+            marginTop: 0,
+          },
+          style2,
         ]}
       />
       <Button
         title="toggle"
         onPress={() => {
-          randomWidth.value = Math.random() * 350;
+          randomWidth.value = Math.random() * 250 + 100;
```
Latropos, line 60: I am making changes to example since I've added an extra view updating its with immediately which makes debugging much easier

Latropos: And also to avoid negative values here.

src/reanimated2/animation/spring.ts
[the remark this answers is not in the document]
Latropos: This is an example of triggering single animation multiple times:

https://user-images.githubusercontent.com/56199675/227569919-bd5bc902-9c77-45c3-ae03-52358eede629.mov

src/reanimated2/animation/spring.ts
```diff
@@ -144,13 +210,76 @@ export function withSpring(
...
+        }
+      }
+
       if (previousAnimation) {
         animation.velocity =
           previousAnimation.velocity || animation.velocity || 0;
         animation.lastTimestamp = previousAnimation.lastTimestamp || now;
+        animation.newDamping = previousAnimation.newDamping || newDamping;
+        animation.newMass = previousAnimation.newMass || newMass;
+        animation.firstTimestamp = now;
```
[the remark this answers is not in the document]
Latropos: (Thats the way we handle withTiming)

Latropos:
@kmagiera 
> velocity will be reset, so there will be no smooth transition

Yes, but I don't think that average user will be able to see it. I've decided to set it to zero, since if we modify mass to change the duration we get this formula [LINK](https://www.wolframalpha.com/input?i=solve%28ln%28t+*+sqrt%28x0*x0+%2B+v0*v0*m%29%29+%3D+%28m*c%29%2F%282000*d%29%2Cm%29). So unless $v_0$ is zero we have to use [Lambert W function](https://en.wikipedia.org/wiki/Lambert_W_function) 😢 I can try to improve it for cases in which user did not provide default damping, since if we update damping instead of mass it gets easier [[LINK]](https://www.wolframalpha.com/input?i=solve%28ln%28t+*+sqrt%28x0*x0+%2B+v0*v0*m%29%29+%3D+%28m*c%29%2F%282000*d%29%2Cd%29)

> I wonder why we limit duration-based animation to underdamped case only 

Type of the motion depends on $\zeta$, and
$$\zeta  = \frac{c}{2 \sqrt{km}}$$
It means we know $\zeta$ after we have calculated mass and damping, so

1. If `toValue` is not predefined different animations with the same config could have different types. In case of real world spring type of motion never change, no matter how you change the amplitude
2. We would have the following cycle: Damping depends on duration -> motion type depends on damping -> duration depends on motion type. So it would be much more difficult. Even if we can solve it, as you've noticed "over-damped seem to have an easier mechanics" so user can obtain such a result with some easing as well.

src/reanimated2/animation/spring.ts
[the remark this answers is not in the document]
kmagiera: Why there is this sudden jump to final value close to the end of this animation? at 0:05-0:06

kmagiera:
The ultimate goal was to have a complimentary API to this: https://developer.apple.com/documentation/uikit/uiview/1622594-animatewithduration

The key difference is exacly the lack of support for velocity and overdamped spring. As you mentioned, overdamped spring motion is very close to timing with exponential easing but that only holds for the case when we start from 0 velocity. Initial velocity is important for all the cases where you transfer from gesture-driven motion to animaiton, e.g. when you fling an item and expect it animate away. This is the most common usecase for spring animations as no other animaiton type allows for velocity to be specified (except from decay but then decay doesn't have a target value).

kmagiera:
I think it is critical that we support these settings. It is unintuitive that we quietly reset some of the provided parameter to say the least, and moreover it goes against of what users expect here. 

I can see the following ways to move forward with this:
1) We add a simple numeric solver and solve this equation numerically, this is an approach used for example here: https://github.com/framer/motion/blob/86ef8286f063cfbf2922bcfbd2e7bf9157e598fe/packages/framer-motion/src/animation/generators/spring/find.ts#L85
2) We can use approximation for Lambert W function, as far as I remember from researching it a while ago, it does have good and simple approximations however they are accurate only for specific ranges (however I remember that we were always falling within these ranges)

src/reanimated2/animation/spring.ts
```diff
@@ -93,39 +155,125 @@ export function withSpring(
...
+      const durationForMass = (mass: number) => {
+        const amplitude = Math.sqrt((mass * v0 * v0 + k * x0 * x0) / k);
+        const c = zeta * 2 * Math.sqrt(k * mass);
+        return (
+          1000 * ((-2 * mass) / c) * Math.log((threshold * 0.01) / amplitude) -
+          duration
+        );
+      };
+
+      return bisectRoot({ min: 0, max: 50, func: durationForMass });
```
Latropos, line 226: Perhaps I'll add here additional comment before merging - usually this is not the best way to find a root and sb in the future may want to improve it.

app/src/examples/AnimatedStyleUpdateExample.tsx
```diff
@@ -11,13 +10,18 @@ export default function AnimatedStyleUpdateExample(): React.ReactElement {
   const randomWidth = useSharedValue(10);
 
   const config = {
-    duration: 500,
-    easing: Easing.bezierFn(0.5, 0.01, 0, 1),
+    duration: 3000,
+    velocity: 5,
```
kmagiera, line 14: Setting some const velocity for start may be a good way to test things but this won't serve as a good example. We should revert changes here and add a better example that'd illustrate the case for spring with velocity (e.g. something similar to what we do in "Darg and snap" example

src/reanimated2/animation/spring.ts
```diff
@@ -38,44 +53,91 @@ export function withSpring(
...
-          ((config as any)[key] = userConfig[key as keyof typeof userConfig])
-      );
+      duration: 2000,
+      dampingRatio: 0.5,
+    } as const;
+
+    const config = { ...defaultConfig, ...userConfig };
+
+    // TODO this could be done only once
+    function initialCalculations(newMass = 0): {
```
kmagiera, line 71: If this only depends on the mass maybe we should run it in onStart and keep omegas + zeta parameters in the config instead of having mass there.

src/reanimated2/animation/spring.ts
```diff
@@ -93,39 +155,126 @@ export function withSpring(
...
+
+      // Bisection turns out to be much faster than Newton's method in our case
+      return bisectRoot({ min: 0, max: 50, func: durationForMass });
+    }
+
+    function spring(animation: InnerSpringAnimation, now: Timestamp): boolean {
+      const { toValue, startTimestamp, current } = animation;
+
+      const timeFromStart = now - startTimestamp;
+      if (userConfig?.duration && timeFromStart > userConfig.duration) {
```
kmagiera, line 234: should be `>=` here, no?

src/reanimated2/animation/spring.ts
```diff
@@ -93,39 +155,126 @@ export function withSpring(
...
+
+      const deltaTime = Math.min(now - lastTimestamp, 64);
+      animation.lastTimestamp = now;
+
+      const t = deltaTime / 1000;
+      const v0 = -velocity;
+      const x0 = toValue - current;
+
+      const { zeta, omega0, omega1 } = initialCalculations(animation.newMass);
+      animation.lastTimestamp = now;
```
kmagiera, line 252: we already did this few lines above

src/reanimated2/animation/spring.ts
```diff
@@ -144,13 +293,42 @@ export function withSpring(
       previousAnimation: SpringAnimation
     ): void {
       animation.current = value;
+      animation.startValue = value;
+
+      let newMass = config.mass;
+
+      if (userConfig?.duration) {
+        let x0 = Number(animation.toValue) - value;
```
kmagiera, line 301: Why we need `Number()` here? We don't do this with other animation types so I'd assume this isn't necessary here and in other places in this file

src/reanimated2/animation/spring.ts
```diff
@@ -144,13 +293,42 @@ export function withSpring(
...
+        if (Number(previousAnimation?.startValue)) {
+          // It makes our animation a bit smoother, especially if sb triggered same animation twice
+          x0 = Number(animation.toValue) - Number(previousAnimation.startValue);
```
kmagiera, line 304: I think the behavior in this case should be that we aim to finish at the time when the previous animation started. Therefore this shouldn't be necessary. We shouldn't need to keep `startValue` as a result. Instead, what we should be doing here is to use the current velocity (not the one from `config` but one from `animation` object) as well as use `startTimestamp` from previous animation in order to calculate the remaining time instead of using `duration` from the config.

kmagiera (changes requested):
Left a bunch of new comments inline. I think this is going it a right direction

src/reanimated2/animation/spring.ts
Latropos: Because we have 
```tsx

export interface AnimationObject {
// ...  
toValue: AnimatableValue;
// ...
}
```
And AnimatableValue is may be a string or array:
```tsx
export type AnimatableValue = number | string | Array<number>;
```

app/src/examples/AnimatedStyleUpdateExample.tsx
```diff
@@ -1,8 +1,7 @@
 import Animated, {
   useSharedValue,
-  withTiming,
   useAnimatedStyle,
-  Easing,
+  withSpring,
```
kmagiera, line 4: Can we please revert all the changes to this file. It was meant to be the most basic example of using animations inside uAS. We link to it in bunch of places. If we want, we can swap out timing with spring but don't make it any more sophisticated than that (i.e. don't add more components etc)

src/reanimated2/animation/spring.ts
```diff
@@ -38,44 +55,90 @@ export function withSpring(
...
-        (key) =>
-          ((config as any)[key] = userConfig[key as keyof typeof userConfig])
-      );
+      duration: 2000,
+      dampingRatio: 0.5,
+    } as const;
+
+    const config = { ...defaultConfig, ...userConfig };
+
+    function initialCalculations(mass = 0): {
```
kmagiera, line 72: The below methods all seem to not rely on the fact they are defined in the animation method scope. If that's the case we should extract them to the file scope and add `'worklet'`, or put as many things as we can in springUtils.ts

src/reanimated2/animation/springUtils.ts
```diff
@@ -0,0 +1,26 @@
...
+}: {
+  min: number;
+  max: number;
+  func: (x: number) => number;
+  maxIterations?: number;
+}) {
+  'worklet';
+  let idx = maxIterations;
+  let current = (max + min) / 2;
+  while (Math.abs(func(current)) > 0.005 && idx > 0) {
```
kmagiera, line 15: Can we turn 0.005 to a constant defined in the file scope?

src/reanimated2/animation/spring.ts
```diff
@@ -93,39 +156,123 @@ export function withSpring(
...
+              t,
+            });
+
+      const { isOvershooting, isVelocity, isDisplacement } =
+        isAnimationTerminatingCalculation(animation, config);
+
+      animation.current = newPosition;
+      animation.velocity = newVelocity;
+
+      if (isOvershooting || (isVelocity && isDisplacement)) {
```
kmagiera, line 275: I think it'd been discussed somewhere before but perhaps got lost: we want for animations with duration set to have additional stop condition such that it ends at the exact moment in time. This is needed because duration could be used to synchronize things happening on screen and we need to make sure the animation finishes exactly when requested

kmagiera (commented):
This is getting very close, just a couple more comments I added inline

src/reanimated2/animation/spring.ts
Latropos: All of them use in-scope variables, so I am not sure if it will really look better after extracting it.

src/reanimated2/animation/spring.ts
Latropos: I've been thinking you mean cases when the obtained animation is too long and we want to terminate it earlier, but it looks like you want to "animate a static string" as well. IMHO both situations are very different so I've added four options to specify `exactDuration` parameter, instead of a single boolean flag. 🤔

src/reanimated2/animation/springUtils.ts
Latropos, line 15:
```suggestion
  const ACCURACY = 0.005;
  let idx = maxIterations;
  let current = (max + min) / 2;
  while (Math.abs(func(current)) > ACCURACY && idx > 0) {
```

src/reanimated2/animation/spring.ts
```diff
@@ -38,94 +25,72 @@ export function withSpring(
...
-      Object.keys(userConfig).forEach(
-        (key) =>
-          ((config as any)[key] = userConfig[key as keyof typeof userConfig])
-      );
-    }
+      duration: 2000,
+      dampingRatio: 0.5,
+    } as const;
+
+    const config = { ...defaultConfig, ...userConfig };
```
kmagiera, line 40:
```suggestion
    const config = { ...defaultConfig, ...userConfig, useDuration: userConfig.duration || userConfig.dampingRatio };
```

src/reanimated2/animation/spring.ts
```diff
@@ -38,94 +25,72 @@ export function withSpring(
...
-      const deltaTime = Math.min(now - lastTimestamp, 64);
-      animation.lastTimestamp = now;
+      const timeFromStart = now - startTimestamp;
 
-      const c = config.damping;
-      const m = config.mass;
-      const k = config.stiffness;
+      const expectedDurationReached =
+        userConfig?.duration && timeFromStart >= userConfig.duration;
+      if (expectedDurationReached) {
```
kmagiera, line 49:
```suggestion
      if (config.useDuration && timeFromStart >= config.duration) {
```

src/reanimated2/animation/spring.ts
```diff
@@ -38,94 +25,72 @@ export function withSpring(
...
+            });
+
+      const { isOvershooting, isVelocity, isDisplacement } =
+        isAnimationTerminatingCalculation(animation, config);
+
+      animation.current = newPosition;
+      animation.velocity = newVelocity;
+
+      const springIsNotInMove =
+        isOvershooting || (isVelocity && isDisplacement);
```
kmagiera, line 92:
```suggestion
        config.useDration && (isOvershooting || (isVelocity && isDisplacement));
```

src/reanimated2/animation/spring.ts
```diff
@@ -137,20 +102,71 @@ export function withSpring(
...
     }
 
+    function isTriggeredTwice(
+      previousAnimation: SpringAnimation | undefined,
+      animation: SpringAnimation
+    ) {
+      return (
+        previousAnimation?.type === 'spring' &&
+        previousAnimation?.toValue === animation.toValue &&
+        previousAnimation?.duration === animation.duration
```
kmagiera, line 112: + dampingRatio

src/reanimated2/animation/springUtils.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+    current = (min + max) / 2;
+  }
+  return current;
+}
+
+export function initialCalculations(
+  mass = 0,
+  config: Record<keyof SpringConfig, any>,
+  userConfig?: SpringConfig
```
kmagiera, line 74:
```suggestion
  config?: SpringConfig
```

src/reanimated2/animation/spring.ts
```diff
@@ -137,20 +102,71 @@ export function withSpring(
...
-        animation.velocity =
-          previousAnimation.velocity || animation.velocity || 0;
-        animation.lastTimestamp = previousAnimation.lastTimestamp || now;
-      } else {
-        animation.lastTimestamp = now;
+      animation.startValue = value;
+
+      let mass = config.mass;
+      const triggeredTwice = isTriggeredTwice(previousAnimation, animation);
+
```
kmagiera, line 127:
```ts
if (triggeredTwice) {
  animation.mass = previousAnimation.mass
  ....
  return;
}
```

kmagiera (approved):

src/reanimated2/animation/spring.ts
Latropos: The only difference is that instead of mass I set zeta, omega1 and omega0

src/reanimated2/animation/spring.ts
```diff
@@ -38,94 +25,72 @@ export function withSpring(
...
       restSpeedThreshold: 2,
       velocity: 0,
-    };
-    if (userConfig) {
-      Object.keys(userConfig).forEach(
-        (key) =>
-          ((config as any)[key] = userConfig[key as keyof typeof userConfig])
-      );
-    }
+      duration: 2000,
```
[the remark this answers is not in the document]
Latropos: Discusses async, we added 
```
     useDuration: userConfig?.duration || userConfig?.dampingRatio,
```
below and now its safe to provide default duration

src/reanimated2/animation/springUtils.ts
```diff
@@ -0,0 +1,235 @@
...
+  config: Record<keyof SpringConfig | 'useDuration', any>
+): {
+  zeta: number;
+  omega0: number;
+  omega1: number;
+} {
+  'worklet';
+  if (config.useDuration) {
+    const { stiffness: k, dampingRatio: zeta } = config;
+
```
Latropos, line 81:
```suggestion

// omega0 and omega1 denote angular frequency and natural angular frequency, see this link for formulas: https://courses.lumenlearning.com/suny-osuniversityphysics/chapter/15-5-damped-oscillations/
```

piaskowyk (approved):

PR #4191, opened 2023-03-08
fix: Throw user-friendly error when `.__initData` is not found

mrousavy:
## Summary

When you forgot to install the Babel Plugin (or are coming from a v2 babel plugin), this is what it looks like:

Screenshot 2023-03-08 at 17 53 21

A SIGABRT.

With this change, a simple JS error is thrown that tells the user to install the Babel Plugin.

### DIFF
--- a/src/reanimated2/initializers.ts
+++ b/src/reanimated2/initializers.ts
@@ -39,6 +39,10 @@ function valueUnpacker(objectToUnpack: any, category?: string): any {
     let workletFun = workletsCache.get(workletHash);
     if (workletFun === undefined) {
       const initData = objectToUnpack.__initData;
+      if (initData == null)
+        throw new Error(
+          "The given function could not be parsed as a Worklet. It looks like the Reanimated Babel Plugin is missing, make sure reanimated's babel plugin is installed in your babel.config.js (you should have 'react-native-reanimated/plugin' listed there - also see the above link for details) \n2) Make sure you reset build cache after updating the config, run: yarn start --reset-cache"
+        );
       if (global.evalWithSourceMap) {
         // if the runtime (hermes only for now) supports loading source maps
         // we want to use the proper filename for the location as it guarantees


tjzel:
Hi! As you can see from the error log, `valueUnpacker@worklet_...` appeared meaning reanimated plugin has indeed worked and workletized something. In Reanimated2 plugin we did not append `initData` property and that seems to be the problem. It looks like somehow old version of plugin is being used with Reanimated3 and simple cache reset would be the solution. Could you provide more information about what exactly did you do that lead to this situation?

mrousavy:
@tjzel Hi, yep that's exactly what happened - I did an upgrade from REA v2 to v3, and since Babel cached the old plugin transforms the `initData` prop was not there yet.
So in this scenario, the user would see the above SIGABRT without a clear error message.

If you do a new install of REA V3 in a blank project, I think the other guards will show proper friendly messages, but for the V2 -> V3 upgrade they don't catch that.

I think this is a pretty unique interaction flow, so feel free to just discard this PR, but I think lots of people will create issues here posting this exact error without thinking about cleaning their babel cache first 😄

tjzel:
@mrousavy Thanks for clarification! I am currently working on adding plugin's version and adding a check in Reanimated for version matching so we will have this problem solved earlier in control flow, but I will be more than happy to use your error message there :)

mrousavy:
Sounds good! Feel free to reach out over Slack if you need any help with testing, been working a lot with the REA plugin lately (because of react-native-vision-camera).

Also, if you have some time could you maybe let me know what you guys want to happen over at https://github.com/software-mansion/react-native-reanimated/pull/4053? For me, that change works in my apps and is actually required to get Frame Processor's `runAsync()` going, without it, I can't dispatch other worklets in my worklet.

tjzel:
Thanks, I will in need. I still have your PR in mind and I think the functionality it would add is useful but I am still trying to determine what would the best strategy of implementing it. At the moment with plugin we are focusing on successfully migrating it to TypeScript, splitting into files and bundling it in the end. Also producing simplified dev documentation to eliminate all the ambiguities about what should plugin do and what it does accidentally. Once we have done that we will go into plugin's logic and try to improve it, including nested worklets.

tjzel:
Resolved with #4206

PR #4209, opened 2023-03-13
[Plugin] [0/9] Split plugin into separate files and add bundling option.

tjzel:
## Summary

More work on plugin refactorization. This is simply splitting the code into files without any changes to it and adding `esbuild` bundler to generate a single file.

## Test plan

Run `yarn jest` and build Example app.

PR #4210, opened 2023-03-13
[Plugin] [1/9] Remove passing a 'method holder' object and properly use TypeScript for plugin.

tjzel:
## Summary

Since we are using TypeScript for plugin now we should get rid of passing the magic `t` object as a function argument and use proper imports instead. Also we no longer import whole `BabelCore` and `BabelTypes` but just the required objects. This is a direct follow up to #4209 

## Test plan

`Yarn jest`, these changes don't touch plugin's logic.


plugin/package.json
```diff
@@ -2,13 +2,15 @@
...
+    "typescript": "^4.1.3",
+    "esbuild": "^0.17.11"
   },
   "license": "MIT",
   "readmeFilename": "README.md",
   "scripts": {
     "prepare": "yarn generate",
-    "generate": "yarn tsc",
+    "generate": "yarn tsc && yarn bundle",
+    "bundle": "yarn esbuild src/plugin.js --log-level=silent --bundle --external:@babel --external:convert-source-map --external:fs --platform=node --sourcemap=linked --outfile=index.js && rm src/*.js src/*.js.map",
```
tjzel, line 13: Added `--log-level:silent` for convenience.

plugin/src/processInlineStylesWarning.ts
```diff
@@ -0,0 +1,141 @@
...
+  memberExpression,
+  returnStatement,
+  blockStatement,
+  ObjectProperty,
+  isIdentifier,
+} from '@babel/types';
+import { isRelease } from './commonFunctions';
+import { ReanimatedPluginPass } from './commonInterfaces';
+
+function generateInlineStylesWarning(path: NodePath<MemberExpression>) {
```
tjzel, line 27: Renaming `memberExpression` to `path` was necessary to avoid conflicts and comply with the rest of plugin's naming policies.

plugin/src/makeWorklet.ts
```diff
@@ -23,7 +63,7 @@ function hash(str: string): number {
   return (hash1 >>> 0) * 4096 + (hash2 >>> 0);
 }
 
-function shouldGenerateSourceMap() {
+function shouldGenerateSourceMap(): boolean {
```
tomekzaw, line 66: We don't need to specify the return type.

tomekzaw (approved):

PR #4211, opened 2023-03-13
[Plugin] [2/9] Convert throws to asserts

tjzel:
## Summary

Change custom error throws on undefined objects to generic assert for better clarity. Direct follow up to #4210.

## Test plan

Nothing to test really.


tomekzaw (changes requested):

plugin/src/asserts.ts
```diff
@@ -0,0 +1,5 @@
+export function assertIsDefined<T>(value: T): asserts value is NonNullable<T> {
+  if (value === undefined || value === null) {
+    throw new Error(`${value} is not defined`);
```
tomekzaw, line 3: When value is `undefined` or `null`, the error message will be "undefined is not defined" or "null is not defined", respectively, which is not very helpful. Maybe we should pass the error message as the second argument?
```ts
assertIsDefined(inputMap, "inputMap is not defined");
```
Also, take a look at Node.js' [`assert`](https://www.w3schools.com/nodejs/met_assert.asp), but I'm not sure how this behaves in the production build of the plugin.

plugin/src/makeWorklet.ts
```diff
@@ -370,7 +372,7 @@ export function makeWorklet(
     functionName,
     transformed.map
   );
-  if (!funString) throw new Error("'funString' is not defined\n");
+  assertIsDefined(funString);
```
tomekzaw, line 375: In which case `funString` is not defined? Is it possible to use TypeScript to ensure that it's defined?

plugin/src/asserts.ts
tjzel: 😅 I don't know how I didn't see that coming. I was thinking about adding more sophisticated assert functions later on. I think instead of passing the whole message, it would be better to just pass the name of the variable.

tomekzaw (approved):

PR #4218, opened 2023-03-14
[Plugin] [8/9] Remove no longer relevant comments in ProcessInlineStylesWarning.

tjzel:
## Summary

Once previous PRs are merged this will be filled. Direct follow-up to #4217.

## Test plan

PR #4219, opened 2023-03-14
[Plugin] Changes to ProcessWorkletObjectMethod for better logic and adding missed function return types.

tjzel:
## Summary

Once previous PRs are merged this will be filled. Direct follow-up to #4218.

## Test plan


plugin/src/makeWorklet.ts
```diff
@@ -323,52 +101,7 @@ export function makeWorklet(
...
-          return;
-        }
-        currentScope = currentScope.parent;
-      }
-      closure.set(name, path.node);
-    },
-  });
-
-  const variables = Array.from(closure.values());
+  const variables = makeArrayFromCapturedBindings(transformed.ast, fun);
```
tomekzaw, line 104: What's the reason behind this rename from `closure`?

plugin/src/makeWorklet.ts
```diff
@@ -301,10 +79,10 @@ export function makeWorklet(
   // comment after the function that gets included here, and then the closing
   // bracket would become part of the comment thus resulting in an error, since
   // there is a missing closing bracket.
-  const code =
+  codeObject.code =
     '(' + (isObjectMethod(fun) ? 'function ' : '') + codeObject.code + '\n)';
 
-  const transformed = transformSync(code, {
+  const transformed = transformSync(codeObject.code, {
```
tomekzaw, line 85: This change affects the behavior of the plugin, what's the motivation and what are the effects?

plugin/src/makeWorklet.ts
tjzel: Previously we were doing a closure inside the function and then we were doing the array from its values as in mentioned line. I moved it to another function which returns this array so the code could be more readable.

plugin/src/makeWorklet.ts
tjzel: We don't use `codeObject.code` anywhere else so I thought's its just logical to replace it instead of making another variable, especially since later we use for example `codeObject.map`.

tomekzaw (approved):
👏

PR #4233, opened 2023-03-17
Emoji waterfall example

tomekzaw:


piaskowyk (commented):
The example looks great 😍 But why you used react context? Is it necessary?

tomekzaw:
> But why you used react context? Is it necessary?

Nope, it's not necessary, but I wanted to provide an imperative API for starting the animation along with a provider component so that you can easily drop in this effect in your app, similarly as in [react-native-fiesta](https://mateoguzmana.github.io/react-native-fiesta/docs/fiesta-context).

PR #4249, opened 2023-03-20
Fix corrupted yarn.locks

tjzel:
## Summary

Our `yarn.lock` files are corrupted and somehow they were not updating properly when using `yarn`. That caused adding multiple versions of some libraries into `node_modules` and TypeScript was going 🤪. I removed all the `yarn.lock` files and recreated them with `yarn` and then had to fix all the issues that were coming with that.

## Test plan

- `yarn jest`
- build Example and run it
- build FabricExample and run it
- build TVOSExample and run it to see that it doesn't work same as on current main
- build WebExample and run it to see that it doesn't work same as on current main


piaskowyk (approved):
Looks good to me 👍

PR #4258, opened 2023-03-22
Change typeof for prototype check to only process plain objects as mappers inputs

kmagiera:
## Summary

This PR changes the logic on when we scan nested objects provided as mappers' inputs. Before this change we'd do a `typeof` check to test whether it is an object. This check however passes on derived objects such as HostObjects, Maps, etc. Since these types of objects can have very complex structures we want to avoid scanning them recursively and only do that for plain object. We achieve this by testing its prototype and comparing with Object.prototype.

## Test plan

Run test examples in the app
Tested this on web with react-native-skia reanimated AnimateTextOnPath example where a HostObject is used inside a mapper and it was taking ages to process that object recursively.

### DIFF
--- a/src/reanimated2/mappers.ts
+++ b/src/reanimated2/mappers.ts
@@ -115,7 +115,10 @@ export function createMapperRegistry() {
       }
     } else if (inputs.addListener) {
       resultArray.push(inputs);
-    } else if (typeof inputs === 'object') {
+    } else if (inputs.__proto__ === Object.prototype) {
+      // we only extract inputs recursively from "plain" objects here, if object
+      // is of a derivative class (e.g. HostObject on web, or Map) we don't scan
+      // it recursively
       for (const element of Object.values(inputs)) {
         element && extractInputs(element, resultArray);
       }


wcandillon:
whoop whoop 🙌🏼

src/reanimated2/mappers.ts
```diff
@@ -115,7 +115,10 @@ export function createMapperRegistry() {
       }
     } else if (inputs.addListener) {
       resultArray.push(inputs);
-    } else if (typeof inputs === 'object') {
+    } else if (inputs.__proto__ === Object.prototype) {
```
piaskowyk, line 118: The `__proto__` property seems to be deprecated, maybe could we use [`Object.getPrototypeOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getPrototypeOf)?

kmagiera: sure, that's a good suggestion

piaskowyk (approved):

tomekzaw (approved):

src/reanimated2/mappers.ts
```diff
@@ -115,7 +115,10 @@ export function createMapperRegistry() {
       }
     } else if (inputs.addListener) {
       resultArray.push(inputs);
-    } else if (typeof inputs === 'object') {
+    } else if (Object.getPrototypeOf(inputs) === Object.prototype) {
```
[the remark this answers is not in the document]
kmagiera: if it was null, than the previous `else if` that I'm not chaging would already crash the app as it is accessing a `addListener` property of `inputs`

wcandillon:
is there a way to track in which version of Reanimated this will be shipped?

tomekzaw:
@wcandillon it's already released in 3.1.0

wcandillon:
I can confirm that the fix works perfectly thank you ❤️

wcandillon:
Gang, is it possible that this fix has been only shipped for Reanimated 3?
We are trying to push Reanimated as the only tool to do animation and we have an integration with Reanimated 2 as well now (working on the JS thread). But this bug appears again with Reanimated 2 is it possible that this should be shipped in v2 as well?

tomekzaw:
@wcandillon Yes, this PR fixes only v3. Also it cannot be simply backported to v2 because in Reanimated v2 mappers were implemented in C++ but in v3 they have been moved to JS code.

PR #4276, opened 2023-03-24
Fix issue with setImmediate not defined on Web

### LINKED ISSUE #4140: [v3][Web] `setImmediate` is not defined
### Description

This bug has appeared over the years many times. The solution has been to use `requestAnimationFrame` instead of `setImmediate` on Web.

The regression came from https://github.com/software-mansion/react-native-reanimated/pull/3970

See: https://github.com/software-mansion/react-native-reanimated/pull/2622 https://github.com/software-mansion/react-native-reanimated/pull/152.

### Steps to reproduce

Use Reanimated on Next.js.

https://stackblitz.com/edit/nextjs-pxczpl?file=pages%2Findex.js,next.config.js,package.json

### Reanimated version

3.0.0

### React Native version

0.71

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

kmagiera:
## Summary

This PR removes uses of setImmediate which is not a standard API across web browsers. Instead, we use queueMicrotask API which is available both in the browser world, in React Native and in Node. 

Fixes #4140

## Test plan

Run example app, test WebExample, test repro app from #4140 with `import setImmediate` removed


tomekzaw (approved):
LGTM 😎

piaskowyk (approved):
🐎

PR #4278, opened 2023-03-24
[Shared Element Transition] Progress animation

piaskowyk:
## Summary

This pull request introduces a new progress-based shared element transition animation that utilizes the progress value from `react-native-screens`. This value indicates the progress of the screen's transition.

This PR includes also two value managers - `JSCallbacksManager` and `JSConfigManager`. Why is it necessary? The `JSCallbacksManager` contains a JS callbacks to make it possible to call on the native side and receive jsi::Value as a result. I needed it to compute animation frame values. Moreover, we can use the `JSCallbacksManager` as a manager for Sensor and Keyboard callbacks - I will refactor it in a separate PR. This manager allows us to simplify `useAnimatedSensor` and `useAnimatedKeyboard` logic and remove some functions from Reanimated Module Spec. The `JSConfigManager` could be reused also, for example to replace `isLayoutAnimationEnabled` method. That's why I prepared a more generic solution instead of adding another function to the native spec.

Also, I suggest the new API for custom animations:
```js
const transition = new SharedTransition()
  .transitionDuration(3000)
  .animation((values) => {
    'worklet';
    return {
      originX: withSpring(values.targetOriginX),
      originY: withSpring(values.targetOriginY),
    };
  })
  .progressAnimation((values, progress) => {
    'worklet';
    return {
      width: progress * (values.targetWidth - values.currentWidth) + values.currentWidth,
      height: progress * (values.targetHeight - values.currentHeight) + values.currentHeight,
      originX: progress * (values.targetOriginX - values.currentOriginX) + values.currentOriginX,
      originY: progress * (values.targetOriginY - values.currentOriginY) + values.currentOriginY,
    };
  });
```


tomekzaw:
Just wanted to post these suggestions here so we don't forget it.

First, it's more convenient to use `interpolate` instead of calculating the values manually: 
```diff
 return {
-  width: progress * (values.targetWidth - values.currentWidth) + values.currentWidth,
-  height: progress * (values.targetHeight - values.currentHeight) + values.currentHeight,
-  originX: progress * (values.targetOriginX - values.currentOriginX) + values.currentOriginX,
-  originY: progress * (values.targetOriginY - values.currentOriginY) + values.currentOriginY,
+  width: interpolate(progress, [0, 1], [values.currentWidth, values.targetWidth]),
+  height: interpolate(progress, [0, 1], [values.currentHeight, values.targetHeight]),
+  originX: interpolate(progress, [0, 1], [values.currentOriginX, values.targetOriginX]),
+  originY: interpolate(progress, [0, 1], [values.currentOriginY, values.targetOriginY]),
 };
```

Secondly, it would be great if `withSpring` and other modifiers could operate on whole object and animate each property individually, this way the developer doesn't need to repeat the whole config 4 (or more) times, but this is a separate feature request.
```diff
-return {
-  width: withSpring(values.targetWidth),
-  height: withSpring(values.targetHeight),
-  originX: withSpring(values.targetOriginX),
-  originY: withSpring(values.targetOriginY),
-};
+return withSpring(values);
```

piaskowyk:
The https://github.com/software-mansion/react-native-reanimated/pull/4421 contains above changes.

PR #4280, opened 2023-03-25
Add flashImmediateRunning flag to setupSetImmediate

### LINKED ISSUE #4168: hooks randomly stop running
### Description

I'm the maintainer of [`react-native-collapsible-tab-view`](https://github.com/PedroBern/react-native-collapsible-tab-view). 

I recently released version 6.0 of the aforementioned library and bumped reanimated to version 3.0.

I noticed that with reanimated 3.0 that hooks like `useAnimatedReaction`, `useDerivedValue` and `useAnimatedStyle` randomly stop running their code. Only tested on iOS so far.

## Broken:

Notice how in the recording below, the code in `useDerivedValue` is not printing anything:

https://user-images.githubusercontent.com/697707/222921089-4d3cf34f-d0f0-4803-8f22-3a2c66c88d4b.mp4

## Working:

If I reload the app, it starts working again. One thing to note is that when this starts occuring, the entire app seems affected. Remounting the React component does not help. Reanimated seems dead unless the app is reloaded. Using the dev menu to reload does work, so I'm unsure if this is an issue in the native layer.

https://user-images.githubusercontent.com/697707/222921139-854a1f76-b2e6-41b4-bcf6-75546d93fde6.mp4

### Steps to reproduce

The issue occurs randomly, but I'm able to repro it in the example app of https://github.com/PedroBern/react-native-collapsible-tab-view

Go to `example` and run `yarn ios` then try swiping around the pages and scroll up and down until it eventually breaks. It's easiest to reproduce in the `Scrollable Tabs` example.

https://github.com/PedroBern/react-native-collapsible-tab-view

### Reanimated version

3.0.1

### React Native version

0.70.6

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

mstach60161:
## Summary

- fixes #4168 

The problem is caused by recursive `callMicrotasks` calls. 
Recursive calls appear when we have an event that trigger another event that calls `callMicrotasks` synchronously.
In result, we call the same microtasks array multiple times. 
In some cases we get the wrong order of instructions and we end up with `runRequested` set on `true` without callbacks in the queue, so we never call `queueMicrotask` again.

The idea of fix is to set flag that inform us, that we are inside `callMicrotasks` call, so we don't execute the same callbacks again

## Test plan

- Created minimal repro using to `Animated.ScrollView` components. (one trigger another) 
- Tested on example from issue


src/reanimated2/threads.ts
```diff
@@ -23,11 +24,16 @@ export function setupMicrotasks() {
...
+    if (callMicrotasksRunning) {
+      return;
+    }
+    callMicrotasksRunning = true;
     for (let index = 0; index < microtasksQueue.length; index += 1) {
       // we use classic 'for' loop because the size of the currentTasks array may change while executing some of the callbacks due to queueMicrotask calls
       microtasksQueue[index]();
     }
     microtasksQueue = [];
+    callMicrotasksRunning = false;
```
kmagiera, line 36: Can we use `try { } finally { }` to make sure this variable is always reset?

src/reanimated2/threads.ts
```diff
@@ -15,6 +15,7 @@ export function setupMicrotasks() {
   'worklet';
 
   let microtasksQueue: Array<() => void> = [];
+  let callMicrotasksRunning = false;
```
kmagiera, line 18: "running" doesn't not feel right here, can we rename to sth along the lines of `isExecutingMicrotasksQueue`

kmagiera (approved):
Left two comments inline, can you please address these prior to merging?

PR #4286, opened 2023-03-27
Amend plugin dependencies

tjzel:
## Summary

- Removal of no longer used `string-hash-64`;
- Bumping some plugin dependencies in `package.json`.

Read changes to README for some more info.
## Test plan

🚀


package.json
```diff
@@ -80,10 +80,10 @@
   "dependencies": {
     "@babel/plugin-transform-object-assign": "^7.16.7",
     "@babel/preset-typescript": "^7.16.7",
+    "@babel/core": "^7.20.0",
     "convert-source-map": "^2.0.0",
     "invariant": "^2.2.4",
-    "lodash.isequal": "^4.5.0",
-    "string-hash-64": "^1.0.3"
+    "lodash.isequal": "^4.5.0"
```
tomekzaw, line 86: #4181 completely removes `lodash.isequal`

package.json
```diff
@@ -80,32 +80,31 @@
     "@babel/plugin-transform-object-assign": "^7.16.7",
     "@babel/preset-typescript": "^7.16.7",
     "convert-source-map": "^2.0.0",
-    "invariant": "^2.2.4",
-    "string-hash-64": "^1.0.3"
+    "invariant": "^2.2.4"
```
tomekzaw, line 83: Fun fact: we use `invariant` only in one place [here](https://github.com/software-mansion/react-native-reanimated/blob/62fef6f657d638de5dcc80e13f420593eee9aea7/src/createAnimatedComponent.tsx#L246-L251) and technically we could just embed it's code as you did with `string-hash-64` but it doesn't make sense to do so because `invariant` is already a dependency of `react-native` so it will be installed anyway.

tjzel: The real reason for me to embed that code is that it didn't have `@types` package on npm 😆, but that is indeed a fun fact.

tomekzaw (approved):

PR #4300, opened 2023-03-29
Introduce executeOnUIRuntimeSync and use it to replace Sync Data Holder

kmagiera:
## Summary

This PR introduces a new concept of executing worklets on the UI runtime synchronously from the RN JS runtime.

It is implemented using runtime decorator that uses a lock to guard all the code that's being executed on that runtime. Thanks to this locking mechanism, we can now grab that lock on a different thread and safely execute some more methods without worrying that the runtime object will be accessed concurrently.

This new method makes opens up some new possibilities. The first one being the way we handle two-way shared values. The shared value rewrite introduced a concept of one-way shared values that are only accessed and modified from the ui runtime. We initially thought such values can be used in majority of the use cases, however, in practice it turned out that a lot of existing code already relies on reading from shared values on the RN thread. Despite the reads happening very infrequently (just a few times per shared value runtime), we still needed to serialize all the updates for a given shared value such that it can be read later on from the RN thread. With this change, we can leverage synchronous execution of the code on UI runtime and ask it to read the value on demand rather than synchronizing the value after each single update.


Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -217,29 +207,20 @@ void NativeReanimatedModule::scheduleOnUI(
...
-    jsi::Runtime &rt,
-    const jsi::Value &synchronizedDataHolderRef) {
-  auto dataHolder = extractShareableOrThrow<ShareableSynchronizedDataHolder>(
-      rt, synchronizedDataHolderRef);
-  return dataHolder->get(rt);
+    const jsi::Value &worklet) {
+  auto shareableWorklet = extractShareableOrThrow(rt, worklet);
+  assert(
+      shareableWorklet->valueType() == Shareable::WorkletType &&
+      "only worklets can be scheduled to run on UI");
```
tomekzaw, line 216: Can we use `extractShareableOrThrow<ShareableWorklet>` here instead?
```suggestion
  auto shareableWorklet = extractShareableOrThrow<ShareableWorklet>(rt, worklet, "only worklets can be scheduled to run on UI");
```

Common/cpp/ReanimatedRuntime/ReanimatedRuntime.cpp
```diff
@@ -52,4 +52,11 @@ std::shared_ptr<jsi::Runtime> ReanimatedRuntime::make(
 #endif
 }
 
+std::shared_ptr<ReanimatedRuntime> ReanimatedRuntime::make(
+    jsi::Runtime *rnRuntime,
```
tomekzaw, line 56: Just wondering if we can pass `jsi::Runtime &` here.

Common/cpp/ReanimatedRuntime/ReanimatedRuntime.h
```diff
@@ -9,20 +9,45 @@
...
+#include <utility>
 
 namespace reanimated {
 
 using namespace facebook;
 using namespace react;
 
-class ReanimatedRuntime {
+struct AroundLock {
+  std::recursive_mutex mutex;
```
tomekzaw, line 24: Should we mark this field as `mutable`?

src/reanimated2/mutables.ts
```diff
@@ -91,8 +75,8 @@ export function makeMutable<T>(
       }
     },
     get value() {
-      if (syncDataHolder) {
-        return NativeReanimatedModule.getDataSynchronously(syncDataHolder);
+      if (NativeReanimatedModule.native) {
+        return executeOnUIRuntimeSync(uiValueGetter)(mutable);
```
tomekzaw, line 79: Can we memoize `executeOnUIRuntimeSync(uiValueGetter)` here so that we don't need to call `NativeReanimatedModule.executeOnUIRuntimeSync` and `makeShareableCloneRecursive` on each read?

Example/src/AnimatedStyleUpdateExample.tsx
```diff
@@ -36,7 +36,13 @@ function AnimatedStyleUpdateExample(): React.ReactElement {
...
         title="toggle"
         onPress={() => {
-          randomWidth.value = Math.random() * 350;
+          const newValue = Math.random() * 350;
+          randomWidth.value = newValue;
+          console.log('HERE', randomWidth.value, newValue);
+          // const hh = doRandomStuff(randomWidth);
+          // console.log('AFTER', randomWidth.value, newValue, hh);
+          setTimeout(() => console.log('HHH', randomWidth.value), 50);
+          // setTimeout(() => doRandomStuff(randomWidth), 150);
```
tomekzaw, line 45: Can we move it to a separate example?

kmagiera: sure, it wasn't meant to be here actually, was just making some temporary changes for testing purposes and accidently commited these

src/reanimated2/mutables.ts
kmagiera: 👍

piaskowyk: fixed here: https://github.com/software-mansion/react-native-reanimated/pull/5513/commits/04b30b8fa4450770299e5813098ed842f4d9af3b

Example/src/AnimatedStyleUpdateExample.tsx
piaskowyk: fixed here: https://github.com/software-mansion/react-native-reanimated/pull/5513/commits/04b30b8fa4450770299e5813098ed842f4d9af3b

tjzel (commented):
With the changes in #5513 and #5516 it seems LGTM.

However I would like someone to explain to me how does it protect us (the locking mechanism implemented here) and what from exactly.

Not approving since it's still a draft.

ranaavneet:
Hi @tomekzaw @tjzel We currently pass `true` in `oneWayReads` param to almost every `useSharedValue` initialisation _(Although @tomekzaw already warned me that it could get deprecated in the future and thats why its not part of the public API yet)_. I have a couple of questions regarding this:

1. Does this mean that post this PR, `oneWayReads` parameter would be deprecated?
2. We started passing `oneWayReads` as `true` since it gave us performance improvements. Would this PR have performance gains as well ?
3. Do we plan to release this with 3.7.0, and if yes, what are the tentative timelines for the same?

Apologies for bothering you guys during holidays! **PS: Merry Christmas in advance! 🎄**

tjzel:
@ranaavneet Today @tomekzaw was explaining to me bits of this PR in regard to performance so here's what I know. 

1. Yes, `oneWayReads` will be deprecated (actually removed, but we might keep deprecated TS signature). This is because `oneWayReads=true` will kind-of be the default behavior of a shared value.

2. From our initial benchmarks done by @piaskowyk this PR will give you an edge over the current default shared value. However I don't think he has tested it with `oneWayReads=true` yet. I don't want to say anything hasty here since @tomekzaw noticed that with this PR there should come some extra optimizations in other bits of our code.

3. If we get enough testing I think it can go to 3.7.0. In regards to the timeline of that, my safe guess would be the end of January 2024, but I don't want to promise anything 😸.

Merry Christmas to you too!

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -161,6 +162,21 @@ void NativeReanimatedModule::scheduleOnUI(
...
+  auto shareableWorklet = extractShareableOrThrow<ShareableWorklet>(
+      rt,
+      worklet,
+      "[Reanimated] Only worklets can be executed synchronously on UI runtime.");
+  auto lock = uiWorkletRuntime_->lock();
+  jsi::Runtime &uiRuntime = uiWorkletRuntime_->getJSIRuntime();
+  auto result = uiWorkletRuntime_->runGuarded(shareableWorklet);
+  auto shareableResult = extractShareableOrThrow(uiRuntime, result);
+  lock.unlock();
+  return shareableResult->getJSValue(rt);
```
tomekzaw, line 177: Does it make sense to actually move this implementation to a new method in`WorkletRuntime` class and call it here?

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
```diff
@@ -4,15 +4,64 @@
 #include "WorkletRuntimeCollector.h"
 #include "WorkletRuntimeDecorator.h"
 
+#include <jsi/decorator.h>
+
 namespace reanimated {
 
+struct AroundLock {
+  std::recursive_mutex *mutex;
```
tomekzaw, line 12: It looks weird as a raw pointer but I guess that's fine.
```suggestion
  std::recursive_mutex *mutex_;
```

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
```diff
@@ -4,15 +4,64 @@
...
+  void before() {
+    mutex->lock();
+  }
+  void after() {
+    mutex->unlock();
+  }
```
tomekzaw, line 18:
```suggestion
  void before() const {
    mutex->lock();
  }
  void after() const {
    mutex->unlock();
  }
```

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
```diff
@@ -4,15 +4,64 @@
...
+  std::shared_ptr<jsi::Runtime> runtime_;
+
+ public:
+  explicit LockableRuntime(
+      std::shared_ptr<jsi::Runtime> &&runtime,
+      std::recursive_mutex *runtimeMutex)
+      : jsi::WithRuntimeDecorator<AroundLock>(*runtime, aroundLock_),
+        runtime_(std::move(runtime)) {
+    aroundLock_.mutex = runtimeMutex;
+  }
```
tomekzaw, line 33: Does it work this way?
```suggestion
        runtime_(std::move(runtime)), aroundLock_(runtimeMutex) {}
```

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
```diff
@@ -4,15 +4,64 @@
...
+};
+
+class LockableRuntime : public jsi::WithRuntimeDecorator<AroundLock> {
+ private:
+  AroundLock aroundLock_;
+  std::shared_ptr<jsi::Runtime> runtime_;
+
+ public:
+  explicit LockableRuntime(
+      std::shared_ptr<jsi::Runtime> &&runtime,
```
tomekzaw, line 28: Why is it a double reference?

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
```diff
@@ -4,15 +4,64 @@
...
+    jsi::Runtime &runtime,
+    const std::shared_ptr<MessageQueueThread> &jsQueue,
+    const std::string &name,
+    WorkletRuntimeType workletRuntimeType,
+    std::recursive_mutex *runtimeMutex) {
+  if (workletRuntimeType == WorkletRuntimeType::WithLocking) {
+    return std::make_shared<LockableRuntime>(
+        ReanimatedRuntime::make(runtime, jsQueue, name), runtimeMutex);
+  } else {
+    return ReanimatedRuntime::make(runtime, jsQueue, name);
```
tomekzaw, line 46:
```
auto reanimatedRuntime = ReanimatedRuntime::make(runtime, jsQueue, name);
if (...) {
  return std::make_shared<LockableRuntime>(reanimatedRuntime);
} else {
  return reanimatedRuntime;
}
```

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
```diff
@@ -4,15 +4,64 @@
...
+  }
+}
+
 WorkletRuntime::WorkletRuntime(
     jsi::Runtime &rnRuntime,
     const std::shared_ptr<MessageQueueThread> &jsQueue,
     const std::shared_ptr<JSScheduler> &jsScheduler,
     const std::string &name,
+    WorkletRuntimeType type,
```
tomekzaw, line 55:
```suggestion
    const WorkletRuntimeType type,
```

Common/cpp/ReanimatedRuntime/WorkletRuntime.h
```diff
@@ -18,6 +18,8 @@ using namespace react;
 
 namespace reanimated {
 
+enum WorkletRuntimeType { WithLocking, WithoutLocking };
```
tomekzaw, line 21: Why do we use enum instead of boolean `supportsLocking_`?

Common/cpp/ReanimatedRuntime/WorkletRuntime.h
```diff
@@ -59,9 +64,11 @@ class WorkletRuntime : public jsi::HostObject,
   std::vector<jsi::PropNameID> getPropertyNames(jsi::Runtime &rt) override;
 
  private:
+  std::recursive_mutex runtimeMutex_;
   const std::shared_ptr<jsi::Runtime> runtime_;
   const std::string name_;
   std::shared_ptr<AsyncQueue> queue_;
+  bool supportsLocking_;
```
tomekzaw, line 71:
```suggestion
  const bool supportsLocking_;
```

Common/cpp/ReanimatedRuntime/WorkletRuntime.h
```diff
@@ -59,9 +64,11 @@ class WorkletRuntime : public jsi::HostObject,
   std::vector<jsi::PropNameID> getPropertyNames(jsi::Runtime &rt) override;
 
  private:
+  std::recursive_mutex runtimeMutex_;
```
tomekzaw, line 67:
```suggestion
  mutable std::recursive_mutex runtimeMutex_;
```

src/reanimated2/mutables.ts
```diff
@@ -103,9 +84,9 @@ export function makeMutable<Value>(
...
       }
     },
-    get value() {
-      if (syncDataHolder) {
-        return NativeReanimatedModule.getDataSynchronously(syncDataHolder);
+    get value(): Value {
+      if (!SHOULD_BE_USE_WEB) {
+        return uiValueGetter(mutable);
       }
       return value;
```
tomekzaw, line 91: Let's do the same check as in `set value(newValue)`
```suggestion
      if (SHOULD_BE_USE_WEB) {
        return value;
      }
      return uiValueGetter(mutable);
```

src/reanimated2/mutables.ts
```diff
@@ -1,23 +1,21 @@
...
-import { runOnUI } from './threads';
+import { executeOnUIRuntimeSync, runOnUI } from './threads';
 import { valueSetter } from './valueSetter';
 
 const SHOULD_BE_USE_WEB = shouldBeUseWeb();
 
+const uiValueGetter = executeOnUIRuntimeSync(<T>(sv: Mutable<T>): T => {
+  'worklet';
+  return sv.value;
+});
```
tomekzaw, line 14: As far as I know, this function is called only once. Also, it's name `uiValueGetter` sounds similar to `valueSetter` which has a different semantics. I think it would be better to call `executeOnUIRuntimeSync(...)()` directly inside `get value()`, the same way as we call `runOnUI(...)()` inside `set value(newValue)`.

src/reanimated2/initializers.ts
```diff
@@ -14,13 +14,13 @@ const SHOULD_BE_USE_WEB = shouldBeUseWeb();
...
 
 // callGuard is only used with debug builds
-export function callGuardDEV<T extends Array<unknown>, U>(
-  fn: (...args: T) => U,
-  ...args: T
-): void {
+export function callGuardDEV<Args extends unknown[], ReturnValue>(
+  fn: (...args: Args) => ReturnValue,
+  ...args: Args
+): ReturnValue | void {
```
tomekzaw, line 20:
```suggestion
): ReturnValue {
```

Common/cpp/ReanimatedRuntime/WorkletRuntime.h
```diff
@@ -26,18 +28,21 @@ class WorkletRuntime : public jsi::HostObject,
       const std::shared_ptr<MessageQueueThread> &jsQueue,
       const std::shared_ptr<JSScheduler> &jsScheduler,
       const std::string &name,
+      WorkletRuntimeType type,
```
tomekzaw, line 31:
```suggestion
      const WorkletRuntimeType type,
```

Common/cpp/ReanimatedRuntime/WorkletRuntime.h
```diff
@@ -26,18 +28,21 @@ class WorkletRuntime : public jsi::HostObject,
...
       const std::string &valueUnpackerCode);
 
   jsi::Runtime &getJSIRuntime() const {
     return *runtime_;
   }
 
+  std::unique_lock<std::recursive_mutex> lock();
```
tomekzaw, line 38: Let's rename this function to `createLock`. I would assume that `lock` would actually lock something and I would look for `unlock`.

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
```diff
@@ -4,15 +4,64 @@
...
+    aroundLock_.mutex = runtimeMutex;
+  }
+};
+
+static std::shared_ptr<jsi::Runtime> makeRuntime(
+    jsi::Runtime &runtime,
+    const std::shared_ptr<MessageQueueThread> &jsQueue,
+    const std::string &name,
+    WorkletRuntimeType workletRuntimeType,
+    std::recursive_mutex *runtimeMutex) {
```
tomekzaw, line 41: Me no likes raw pointers.

tomekzaw (changes requested):
I've left some minor suggestions in the comments.

Apart from this, I have some thoughts regarding the granularity of locks. 

When we use the Reanimated runtime synchronously on the JS thread, we have one giant lock that wraps the call, which is nice.

On the other hand, when we use the Reanimated runtime on the UI thread, we call `lock` and `unlock` in `before` and `after` methods. This means that when we invoke the UI runtime using JSI from C++ code, we lock and unlock it multiple number of times.

Is it okay to have different levels of granularity in these two scenarios? Is there some way we can lock the Reanimated runtime once for the whole animation frame to avoid short-locking?

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
piaskowyk: We receive pointer to runtime mutex in `LockableRuntime` constructor, but he `std::recursive_mutex` class doesn't have a copy assignment operator or copy constructor defined, so thats why here is raw pointer.

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
piaskowyk: yup, but I need to add constructor to AroundLock

Common/cpp/ReanimatedRuntime/WorkletRuntime.h
piaskowyk: I don't know why kmagiera did it like that, but I can agree that simple boolean is better 👍

src/reanimated2/initializers.ts
piaskowyk: unfortunately we can't :/ because of `try` block inside there is possibility that function will return nothing

tomekzaw: so what's the expected return value of a worklet that throws an error on the UI runtime? should we re-throw the error in the RN runtime or what?

src/reanimated2/mutables.ts
```diff
@@ -85,10 +80,14 @@ export function makeMutable<Value>(initial: Value): Mutable<Value> {
...
-        return uiValueGetter(mutable);
+      if (SHOULD_BE_USE_WEB) {
+        return value;
       }
-      return value;
+      const uiValueGetter = executeOnUIRuntimeSync(<T>(sv: Mutable<T>): T => {
+        'worklet';
+        return sv.value;
+      });
+      return uiValueGetter(mutable);
```
tomekzaw, line 90: Just out of curiosity, does it work if we inline this function instead of assigning it to a variable? Can we omit the generic types if we do so? Should we rename `sv` to `mutable`?
```suggestion
      return executeOnUIRuntimeSync((mutable) => {
        'worklet';
        return mutable.value;
      })(mutable);
```

tomekzaw: Also, do we want to automatically workletize the function passed to `executeOnUIRuntimeSync` as we already do for `runOnUI`? Obviously, this can be added in a separate PR.

piaskowyk: 1. Absolutely, we can inline this function instead of assigning it to a variable. However, personally, I find it more readable this way 🥹. But if you prefer, we can certainly change it. We can inline it only to this form:
```js
return executeOnUIRuntimeSync((sv: Mutable<Value>) => {
    'worklet';
    return sv.value;
})(mutable);
```
2. Yes, we can remove the generic type. I forgot about it, good catch! 👍
3. If we change the name to `mutable`, there will be a conflict with line 72
image
4. auto-workletization - done ✅

piaskowyk:
>Is it okay to have different levels of granularity in these two scenarios? Is there some way we can lock the Reanimated runtime once for the whole animation frame to avoid short-locking?

Based on my understanding, in this case, we require both levels of locking. The `lock` and `unlock` in the `before` and `after` functions ensure thread-safety at the single operation level, which is necessary. However, the lock in `WorkletRuntime::executeSync` exists to synchronize the batch of two single JS instructions: `runGuarded` and `extractShareableOrThrow`.

We can further discuss this during our meeting tomorrow with kmagiera.

src/reanimated2/mutables.ts
tomekzaw: 1. I like the current version ✅ 
2. Thanks!
3. That's right, thanks for explaining!
4. Thanks!

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
```diff
@@ -4,15 +4,68 @@
...
+    mutex_->lock();
+  }
+  void after() const {
+    mutex_->unlock();
+  }
+};
+
+class LockableRuntime : public jsi::WithRuntimeDecorator<AroundLock> {
+  AroundLock aroundLock_;
+  std::shared_ptr<jsi::Runtime> runtime_;
```
tomekzaw, line 28: Why do we need to store `runtime_`? Where do we use it?

src/reanimated2/core.ts
```diff
@@ -28,7 +28,7 @@ import type {
 import { SensorContainer } from './SensorContainer';
 
 export { startMapper, stopMapper } from './mappers';
-export { runOnJS, runOnUI } from './threads';
+export { runOnJS, runOnUI, executeOnUIRuntimeSync } from './threads';
```
tomekzaw, line 31: We will need to add this function to our docs 🤯

tomekzaw (approved):

Common/cpp/ReanimatedRuntime/WorkletRuntime.cpp
piaskowyk: It seems that this is required by jsi::RuntimeDecorator. If I remove it the application will crash :/

src/reanimated2/core.ts
piaskowyk: that is true 🫣 but maybe save it for another PR

tomekzaw (approved):

PR #4316, opened 2023-03-30
Call assert in `SingleInstanceChecker` only in example apps

tomekzaw:
## Why

Currently, Reanimated crashes on some configurations due to an assert in `SingleInstanceChecker` which was meant to help us detect retain cycles and thus memory leaks in the library at an early stage but actually causes lots of troubles in client apps. This PR changes the behavior of `SingleInstanceChecker` so that it always prints the warning to the console but calls the failing `assert` only in Reanimated example apps (i.e. Example or FabricExample).

## How

This PR wraps the failing assert inside a preprocessor conditional based on the flags passed from `build.gradle` via `CMakeLists.txt` (Android) or `RNReanimated.podspec` (iOS). Additionally, it introduces `__android_log_print` instead of `std::cerr` on Android so that the warning is also visible in logcat.

Android:

android

iOS:

ios

## Test plan

1. Check if `IS_REANIMATED_EXAMPLE_APP` preprocessor definition is properly injected on both platforms:
     - Android: change implementation of `isReanimatedExampleApp` in build.gradle to always return `true`
     - iOS: assign `true` to `result[:is_reanimated_example_app]` in `reanimated_utils.rb`
     - add the following snippet in `SingleInstanceChecker.h`:
     ```cpp
     #ifdef IS_REANIMATED_EXAMPLE_APP
     #error "it works"
     #endif
    ```
2. Check if the example apps crash on reload:
    - modify the condition in `SingleInstanceChecker.h` so that it always fails
    - the app should crash on launch
3. Check if the client apps don't crash on reload but show a warning:
    - modify the condition in `SingleInstanceChecker.h` so that it always fails
    - build the package locally
    - create a new RN app and install Reanimated
    - the app shouldn't crash on launch
    - there should be a warning in Android Logcat / iOS app output console

### DIFF
--- a/Common/cpp/Tools/SingleInstanceChecker.h
+++ b/Common/cpp/Tools/SingleInstanceChecker.h
@@ -23,7 +23,9 @@ class SingleInstanceChecker {
   void assertWithMessage(bool condition, std::string message) {
     if (!condition) {
       std::cerr << message << std::endl;
-      assert(condition);
+#ifdef IS_REANIMATED_EXAMPLE_APP
+      assert(false);
+#endif
     }
   }
 
--- a/Common/cpp/Tools/SingleInstanceChecker.h
+++ b/Common/cpp/Tools/SingleInstanceChecker.h
@@ -7,6 +7,10 @@
 #include <iostream>
 #include <string>
 
+#ifdef ANDROID
+#include <android/log.h>
+#endif
+
 namespace reanimated {
 
 // This is a class that counts how many instances of a different class there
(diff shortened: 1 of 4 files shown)


piaskowyk (approved):

radex:
Can this be back ported to 2.x at the earliest opportunity? The same issue occurs there.

tomekzaw:
Hey @radex, generally we plan to drop support for v2 but I think we can make an exception for this issue. Let me know if you need any other bugfixes.

radex:
@tomekzaw Thanks for the info, this is the only major issue that I've been experiencing right now on v2. (Of course I intend to migrate to v3 eventually, but rewriting many complex REA sequences written with v1 API is going to be painful.)

tomekzaw:
@radex 2.16.0 is [out](https://github.com/software-mansion/react-native-reanimated/releases/tag/2.16.0)!

radex:
@tomekzaw Thanks!

tomekzaw:
Hey @radex, I've made a stupid mistake when publishing 2.16.0 (it doesn't contain `lib` folder and Android AARs so it always builds from source and doesn't work on web), please consider upgrading to 2.17.0.

PR #4323, opened 2023-04-03
Set right ReactAndroid version

piaskowyk:
## Summary

To fix this kind of error with the package:

We need to specific grade dependencies. The `+` sign instead of fix version breaks the package build for `rn@0.7.+` because `rn@0.72.0-rc.0` was released.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1035,7 +1035,7 @@ dependencies {
         if (isPackageBuild()) {
             implementation "com.facebook.react:react-android:" + REACT_NATIVE_VERSION
         } else {
-            implementation "com.facebook.react:react-android:0.71.5" // version substituted by RNGP
+            implementation "com.facebook.react:react-android:" // version substituted by RNGP
         }
         if (JS_RUNTIME == "hermes") {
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -1033,13 +1033,13 @@ dependencies {
 
     if (REACT_NATIVE_MINOR_VERSION >= 71) {
         if (isPackageBuild()) {
-            implementation "com.facebook.react:react-android:+"
+            implementation "com.facebook.react:react-android:" + REACT_NATIVE_VERSION
         } else {
-            implementation "com.facebook.react:react-android:" // version substituted by RNGP
+            implementation "com.facebook.react:react-android:0.71.5" // version substituted by RNGP
```
tomekzaw, line 1038: For RN 0.71+, the version should be substituted by react-native-gradle-plugin, at least when building an application. Also, leaving `implementation` without the version here should be safe as Gradle will throw an error if there's no version.

tomekzaw, line 1038:
```suggestion
            implementation "com.facebook.react:react-android" // version substituted by RNGP
```

PR #4326, opened 2023-04-03
Add pre-commit plugin .js files compiling and building

tjzel:
## Summary

If you forgor to build the plugin it will do it for you. But pls don't forgor.

## Test plan

Forgor on purpose and see angry message.


WoLewicki (approved):
forgor

.husky/pre-commit
```diff
@@ -3,3 +3,14 @@
 
 yarn lint-staged
 yarn type:check
+
+# this automatically builds plugin js files if you forgor
+! git diff HEAD --name-only | grep plugin
+if [ $? -eq 1 ]; then
```
tomekzaw, line 9: Is it recommended to use return code instead of embedding the command here directly?

.husky/pre-commit
```diff
@@ -3,3 +3,14 @@
...
+  yarn plugin
+  ! git status -u | grep -E 'plugin/build/plugin.js|plugin/build/plugin.js.map'
```
tomekzaw, line 11: 1. Why not `-e`?
2. `.` will match any character, we should either escape the dots or match just `plugin/build` (this also resolves 1.).

piaskowyk (approved):
👀

.husky/pre-commit
```diff
@@ -3,3 +3,14 @@
...
+  if [ $? -eq 1 ]; then
+    echo "\n[Git hook] You forgot to build plugin and I had to do it for you!!!\n"
```
tomekzaw, line 13: Let's make this error more serious (funny PR descriptions are still fine 😄 ).

kacperkapusciak (approved):

tomekzaw (changes requested):

.husky/pre-commit
```diff
@@ -3,3 +3,14 @@
 
 yarn lint-staged
 yarn type:check
+
+# this automatically builds plugin js files if you forgor
+! git diff HEAD --name-only | grep plugin
```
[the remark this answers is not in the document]
tjzel: `diff-index` is low level and compares mainly metadata of a file blob. `diff` is more general thus slower according to [this article](https://stackoverflow.com/questions/24197606/whats-the-difference-between-git-diff-and-gif-diff-index). I guess for our sake `diff-index` is better but actual performance gain is negligible.

.husky/pre-commit
tjzel: People have different [opinions](https://unix.stackexchange.com/questions/152966/should-i-embed-a-command-in-a-conditional-or-test-its-return-code) but I think it's more transparent in this form.

.husky/pre-commit
tjzel: actually -e is not needed anyway and I actually forgot escaping

tomekzaw (approved):

PR #4330, opened 2023-04-04
Add sensor slider example

kacperkapusciak:
🐶


tjzel (approved):
🚀

app/src/examples/VolumeExample.tsx
```diff
@@ -0,0 +1,211 @@
...
+    }
+    const ax = animatedSensor.sensor.value.x * 0.0001;
+    vx.value += ax * dt;
+    x.value += vx.value * dt;
+    x.value = Math.min(100, Math.max(0, x.value));
+
+    if (x.value === 0 || x.value === 100) {
+      vx.value = 0;
+    }
+  });
```
piaskowyk, line 69: Just wondering 🤔 Is a frame callback necessary here? Maybe the better idea would be the usage of the `useDerivedValue` or `useAnimationReaction` hook here.

kacperkapusciak: Hm 🤔 I kinda need the change in time `dt` for physics calculations

app/src/examples/VolumeExample.tsx
```diff
@@ -0,0 +1,211 @@
...
+} from 'react-native';
+import React, { useLayoutEffect } from 'react';
+import { useNavigation } from '@react-navigation/native';
+import Slider from '@react-native-community/slider';
+
+const primaryColor = '#FFD61E';
+const secondaryColor = '#f5f5f5';
+const backgroundColor = '#232736';
+const secondaryBackgroundColor = '#1B2445';
+
```
tomekzaw, line 25: plz unify lowercase/uppercase

app/src/examples/VolumeExample.tsx
```diff
@@ -0,0 +1,211 @@
...
+    return {
+      value: x.value,
+    };
+  });
+
+  const animatedTextProps = useAnimatedProps(() => {
+    return {
+      text: String(Math.floor(x.value)),
+      // Here we use any because the text prop is not available in the type
+    } as any;
```
tomekzaw, line 81: Can you please double-check if that's still valid?

app/src/examples/VolumeExample.tsx
```diff
@@ -0,0 +1,211 @@
...
+    };
+  });
+
+  return (
+    <View
+      style={{
+        flex: 1,
+        flexDirection: 'column',
+        backgroundColor,
+        padding: 20,
```
tomekzaw, line 96: Can you please static styles to `styles`?

app/src/examples/VolumeExample.tsx
```diff
@@ -0,0 +1,211 @@
...
+  },
+  header: {
+    fontSize: 24,
+    color: secondaryColor,
+    fontFamily: 'Poppins-Bold',
+  },
+  subHeader: {
+    fontSize: 24,
+    color: secondaryColor,
+    fontFamily: 'Poppins-Bold',
```
tomekzaw, line 209: Do you think if it makes sense to have `const FONT_FAMILY = 'Poppins-Bold';` and use `FONT_FAMILY` everywhere instead?

app/src/examples/VolumeExample.tsx
```diff
@@ -0,0 +1,211 @@
...
+          value={true}
+        />
+      </View>
+      <View style={styles.divider} />
+      <Text style={[styles.header, styles.margin]}>Streaming quality</Text>
+      <Text style={styles.text}>Automatic</Text>
+      <Text style={styles.text}>Low</Text>
+      <Text style={styles.text}>Normal</Text>
+      <Text style={styles.text}>High</Text>
+      <Text style={styles.text}>Very high</Text>
```
tomekzaw, line 137: plz use `map`

tomekzaw (changes requested):

app/src/examples/VolumeExample.tsx
kacperkapusciak: Yup, still errors out 👍

app/src/examples/VolumeExample.tsx
kacperkapusciak: Okay! Fixed in eade45b58

app/src/examples/VolumeExample.tsx
kacperkapusciak: oki doki. Fixed in eade45b58

app/src/examples/VolumeExample.tsx
```diff
@@ -0,0 +1,211 @@
...
+    borderRadius: 16,
+    borderWidth: 1,
+    width: 170,
+    height: 120,
+    alignSelf: 'center',
+    justifyContent: 'center',
+  },
+  value: {
+    textAlign: 'center',
+    // fontFamily: 'Poppins-Bold',
```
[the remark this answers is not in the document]
kacperkapusciak: 👍

app/src/examples/VolumeExample.tsx
kacperkapusciak: Valid point but there are only two `Poppins-Bold` and one `Poppins-Regular`. If there would be more then I'd definitely do that but I think here it's fine to leave it as is

Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
```diff
@@ -552,6 +552,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = 1;
+				DEVELOPMENT_TEAM = J5FM626PE2;
```
tomekzaw, line 555: plz revert

WebExample/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "@expo/webpack-config": "^18.0.0",
+    "@react-native-community/slider": "^4.4.2",
```
tomekzaw, line 13: according to https://docs.expo.dev/versions/latest/sdk/slider/, slider doesn't work on web so there's no point adding it as a dependency in WebExample I think

tomekzaw (changes requested):

WebExample/package.json
kacperkapusciak: Nope, that's not right. It works on the web just fine
image
image

piaskowyk (approved):
🐶

tomekzaw (approved):

PR #4331, opened 2023-04-04
Fix globals.d.ts and migrate to globalThis

tjzel:
## Summary

- Migrated from the use of NodeJS.global to globalThis which [seems to be the way now](https://stackoverflow.com/questions/68481686/type-typeof-globalthis-has-no-index-signature).
- Added omitted global variables to global declaration.
- For some reason had to rename `ErrorUtils` to `__ErrorUtils` for it to work properly.
- Fixed a type error in NativeMethods with incorrect typing for `_dispatchCommand`.

Also bumped `@types/node` version since now it won't cause TypeScript errors (it was doing that before the fix).

## Test plan

Run tests, build Example app, do `yarn tsc --noEmit` in `Example/`.


src/reanimated2/globals.d.ts
```diff
@@ -1,143 +1,96 @@
...
+  var _scrollTo: (
     viewTag: number,
     x: number,
     y: number,
     animated: boolean
   ) => void;
-  const _dispatchCommand: (
-    shadowNodeWrapper: ShadowNodeWrapper,
+  var _dispatchCommand: (
+    nodeRef: ShadowNodeWrapper | number,
```
tjzel, line 68: Please pay attention to this as I'm not 100% certain it's safe to change the name of the prop for a more logical one here.

plugin/src/commonObjects.ts
```diff
@@ -46,7 +46,7 @@ export const globals = new Set([
   '_removeShadowNodeFromRegistry',
   'RegExp',
   'Error',
-  'ErrorUtils',
+  '__ErrorUtils',
```
tomekzaw, line 49: Would be nice to explain why this change was necessary (as a comment, either here or in the source code).

src/reanimated2/globals.d.ts
```diff
@@ -1,143 +1,96 @@
+/* eslint-disable no-var, @typescript-eslint/no-explicit-any */
```
tomekzaw, line 1: Why do we need to ignore these rules? Can we ignore only specific lines?

src/reanimated2/globals.d.ts
tomekzaw: As discussed offline, `nodeRef` is a `number` on Paper and a `ShadowNodeWrapper` on Fabric, so I'm not sure if it's possible to type it properly since architecture detection happens in the runtime.

src/reanimated2/globals.d.ts
```diff
@@ -1,143 +1,96 @@
...
-  const _IS_FABRIC: boolean;
-  const _REANIMATED_VERSION_CPP: string;
-  const _REANIMATED_VERSION_BABEL_PLUGIN: string;
-  const __reanimatedModuleProxy: NativeReanimated;
-  const evalWithSourceMap: (
+  var _WORKLET: boolean;
+  var _IS_FABRIC: boolean;
+  var _REANIMATED_VERSION_CPP: string;
+  var _REANIMATED_VERSION_BABEL_PLUGIN: string;
+  var __reanimatedModuleProxy: NativeReanimated;
```
tomekzaw, line 24: We can also update these types, since all of these are `undefined` on web.
```suggestion
  var _WORKLET: boolean | undefined;
  var _IS_FABRIC: boolean | undefined;
  var _REANIMATED_VERSION_CPP: string | undefined;
  var _REANIMATED_VERSION_BABEL_PLUGIN: string | undefined;
  var __reanimatedModuleProxy: NativeReanimated | undefined;
```
Also, I'm wondering if TypeScript supports platform extensions, e.g. `.web.ts`, so we can use different types on native and web (probably not).

src/reanimated2/globals.d.ts
tjzel: I am not certain what the issue is tho. Here we declare that `nodeRef` is one of the types. I agree that if we are on Fabric passing a number as its value is illegal but I think that's still better than going for `unknown` or the like.

If it was desired we could differentiate for those nodes, for example:
```diff
-  shadowNodeWrapper: ShadowNodeWrapper,
+  paperRef: number,
+  fabricRef: ShadowNodeWrapper
```

and then check in runtime which prop should we use based on used architecture.

Still, we have to do something about it because this gives us type errors and I think it's the best approach for now.

tomekzaw: I like this idea but we'd still have to use `| undefined` in both places and somehow prevent passing undefined as both arguments.

plugin/src/commonObjects.ts
tjzel: Screenshot 2023-04-05 at 14 56 22
Screenshot 2023-04-05 at 14 58 09

The greatest mystery of our time 👀
([not really](https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/Libraries/vendor/core/ErrorUtils.js#L25))

src/reanimated2/globals.d.ts
tjzel: @tomekzaw you could've just asked

```TypeScript
  var _dispatchCommand:
    | ((
        paperRef: number,
        fabricRef: never,
        commandName: string,
        args: Array<unknown>
      ) => void)
    | ((
        paperRef: never,
        fabricRef: ShadowNodeWrapper,
        commandName: string,
        args: Array<unknown>
      ) => void);
      ```

src/reanimated2/globals.d.ts
tjzel: We have vars everywhere so ignoring every single line would be... pretty bad. I moved `no-explicit-any` to specific lines though.

src/reanimated2/globals.d.ts
tjzel: It seems to be [doable](https://stackoverflow.com/questions/44001050/platform-specific-import-component-in-react-native-with-typescript) but not the simplest.

tomekzaw: The example you've linked to seems to be a bit different - there are two implementations (Android, iOS) which have the same type, and Metro bundler can choose them based on the suffix (e.g. `.android.tsx`). In our case, the web implementation has different types than the native one and it looks like TypeScript cannot into suffixes.

tomekzaw: For now, adding `| undefined` is fine I think.

src/reanimated2/globals.d.ts
tomekzaw: Let's go for it then

tomekzaw (approved):

PR #4340, opened 2023-04-06
Add one listener per sensor config

mstach60161:
## Summary

This PR is to optimise using sensors.

Before:

Each `useAnimatedSensor` call creates a new native listener that update shared value assigned to it.

Now:

We identify sensors listeners by sensor type and config properties and the same configurations use the same shared value.
 

## Test plan

- Example app
```
import Animated, {
  useAnimatedStyle,
  useAnimatedSensor,
  SensorType,

import React from 'react';

export default function AnimatedSensorExample() {

  const sensorObject = (color: string, sensorType: any) => {
    const gravity = useAnimatedSensor(sensorType, { interval: 16 });
    console.log(gravity.sensor.value);

      //console.log(gravity.sensor.value);
      return {
        top: -gravity.sensor.value.y * 300,
        left: gravity.sensor.value.x * 200,
      };
    });


    return (
        <Animated.View style={[{...styles.box, backgroundColor: color}, animatedStyle]} />
      </View>
    )
  };

  return (
      {sensorObject('red', SensorType.GRAVITY)}
      {sensorObject('blue', SensorType.ACCELEROMETER)}
      {sensorObject('green', SensorType.GRAVITY)}
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: {
    width: 150,
    height: 150,
    backgroundColor: 'navy',
  },
});

```


src/reanimated2/SensorContainer.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    const sensor = this.nativeSensors?.get(sensorId);
+
+    if (!sensor?.isRunning()) {
+      if (!sensor?.initialize()) {
+        return -1;
+      }
+    }
+    sensorRef.current.sensor = sensor?.getSharedValue();
```
piaskowyk, line 48: This part is a bit confusing for me 🤔 
1. You create the new shared value in `useAnimatedSensor` for every `useAnimatedSensor` call.
```
  sensorType: SensorType,
): AnimatedSensor {
  const ref = useRef<AnimatedSensor>({
    sensor: initSensorData(sensorType),
```
2. Then we overriding it by previous registered sharedValue for this type of sensor.:
```
sensorRef.current.sensor = sensor?.getSharedValue();
```
Why we can't create a new shared value once in `Sensor` constructor?

src/reanimated2/SensorContainer.ts
```diff
@@ -0,0 +1,64 @@
...
+      this.nativeSensors.set(sensorId, newSensor);
+    }
+
+    const sensor = this.nativeSensors?.get(sensorId);
+
+    if (!sensor?.isRunning()) {
+      if (!sensor?.initialize()) {
+        return -1;
+      }
+    }
```
piaskowyk, line 47:
```
if (!sensor?.isRunning() && !sensor?.initialize()) {
  return -1;
}
```
Looks better

src/reanimated2/SensorContainer.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  unregisterSensor(sensorId: number) {
+    if (this.nativeSensors.has(sensorId)) {
+      const sensor = this.nativeSensors.get(sensorId);
+      sensor?.removeListener();
+
+      if (!sensor?.hasActiveListeners()) {
+        sensor?.unregister();
+        this.nativeSensors.delete(sensorId);
+      }
```
piaskowyk, line 61: Can we wrap it in `if (sensor)` instead of checking three times if it is null with `?` operator?

src/reanimated2/Sensor.ts
```diff
@@ -0,0 +1,69 @@
...
+  getSharedValue() {
+    return this.data;
+  }
+
+  unregister() {
+    NativeReanimatedModule.unregisterSensor(this.sensorId);
+    this.sensorId = -1;
+  }
+
+  addListener() {
```
piaskowyk, line 62: Could we change this name to more descriptive? The `addListener` method suggests that the method adds listener callback instead of just incrementing the counter.

piaskowyk (changes requested):
Looks fine, but we need just polish some details. I want to ask you about two another things:
- Add the example to the PR description
- Resolve conflict with the main branch

src/reanimated2/SensorContainer.ts
mstach60161: It's because we can have situation when we attach a native listener to shared value (e.g. `useAnimatedStyle`) before sensor is registered or we still have shared value from previous one -  this problem was mainly with fast refresh.

src/reanimated2/Sensor.ts
```diff
@@ -0,0 +1,75 @@
...
+    const config = this.config;
+    const sensorType = this.sensorType;
+    this.sensorId = NativeReanimatedModule.registerSensor(
+      sensorType,
+      config.interval === 'auto' ? -1 : config.interval,
+      config.iosReferenceFrame,
+      eventHandler
+    );
+    return this.sensorId !== -1;
+  }
```
piaskowyk, line 61: This part is problematic for me. Why do you need to register a new listener for every sensor when you can use the same shared value for all? We want to have one shared value and one native listener for every `useAnimatedSensor` with the same type.
Based on my tests, it seems that you never unregister the native listeners. This assertion fails on reload:
```
AnimatedSensorModule::~AnimatedSensorModule() {
  assert(sensorsIds_.empty());
}
```

piaskowyk (changes requested):

src/reanimated2/Sensor.ts
```diff
@@ -49,6 +49,8 @@ export default class Sensor<T> {
   register(
     eventHandler: ShareableRef<T> | ((data: Value3D | ValueRotation) => void)
   ) {
+    // prevent another sensor registration at the same time
+    this.sensorId = -2;
```
piaskowyk, line 53: This looks like a magic number, i think that the default value of sensorId should be null or undefined

piaskowyk (approved):

PR #4350, opened 2023-04-11
Make single instance assertion work with Gradle Configuration Cache

j-piasecki:
## Summary

Makes single instance assertion not break builds when using [Gradle Configuration Cache](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:intro).

Related: https://github.com/software-mansion/react-native-gesture-handler/pull/2453

## Test plan

Enable Gradle Configuration Cache by adding `org.gradle.unsafe.configuration-cache=true` in the `gradle.properties` file. If testing on the Example app, remove `disableMultipleInstancesCheck=true`.

Build the app with and without this PR.


piaskowyk (approved):
Well done!

PR #4352, opened 2023-04-11
Migrate away from `setNativeProps` on web

### LINKED ISSUE #4335: Web implementation doesn't work in react-native-web 0.19
### Description

`setNativeProps` method was removed in [react-native-web 0.19](https://github.com/necolas/react-native-web/releases/tag/0.19.0)

Related discussion: https://github.com/software-mansion/react-native-reanimated/discussions/2756

### Steps to reproduce

-

-

### Reanimated version

-

### React Native version

-

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

tomekzaw:
## Summary

This draft PR moves off of `setNativeProps` in favor of editing DOM nodes directly.

Closes #4335, related to #2756.

## Test plan

<details>
<summary>App.tsx</summary>

```tsx
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  cancelAnimation,
  withRepeat,
  interpolateColor,

export default function AnimatedStyleUpdateExample(): React.ReactElement {
  const randomWidth = useSharedValue(0);

  useEffect(() => {
    cancelAnimation(randomWidth);
    randomWidth.value = withRepeat(withTiming(1), -1, true);
  }, [randomWidth]);

    return {
      width: randomWidth.value * 100,
      height: randomWidth.value * 100,
        randomWidth.value,
        [0, 1],
        ['red', 'lime']
      ),
      transform: [{ rotate: `${randomWidth.value * 180}deg` }],
    };
  });

  return (
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

</details>

https://github.com/software-mansion/react-native-reanimated/assets/20516055/93b8096f-2a4d-4c61-848d-f483d612ac13


chrisedington:
Thank you for this @tomekzaw

macrozone:
thx @tomekzaw , is there an easy way to test this in moti?

would love to get this to work in nextjs!

nandorojo:
If you use `patch-package`, you can test it by editing this reanimated file in your node_modules: https://github.com/software-mansion/react-native-reanimated/pull/4352/files#diff-c91126f95c199aee12f3e0909df9a49ccb8bfd7c529a1430446df596e2c71330

Then `yarn patch-package react-native-reanimated` and `yarn next`

y120sb:
@tomekzaw Thanks!
Does this support transform?
I tried, but could not used.

src/reanimated2/js-reanimated/index.ts
```diff
@@ -47,14 +48,18 @@ export const _updatePropsJS = (
...
   component: JSReanimatedComponent,
   style: StyleProps
 ): void => {
   const previousStyle = component.previousStyle ? component.previousStyle : {};
   const currentStyle = { ...previousStyle, ...style };
   component.previousStyle = currentStyle;
-  component.setNativeProps({ style: currentStyle });
+
+  const domStyle = createReactDOMStyle(currentStyle);
+  for (const key in domStyle) {
```
alantoa, line 60: We should handle this situation specifically because `react-native-web` has already removed array transform. You can check it out [here](https://github.com/necolas/react-native-web/blob/39b94b194594adebdfd5d4a5b873ced8d387734e/packages/react-native-web/src/exports/StyleSheet/preprocess.js#L203).

Something just like this: 

```
import { createTransformValue } from 'react-native-web/dist/exports/StyleSheet/preprocess';
for (const key in domStyle) {
  if(key === 'transform') {
      (component.style).transform = createTransformValue(domStyle[key]);
  }else{
      (component.style)[key] = domStyle[key];
  }
}

```

nandorojo: does the current RNW function not account for it?

alantoa: Uunfortunately no. If you don't use `createTransformValue`, `useAnimatedStyle` will not work.

tomekzaw:
Thanks @y120sb and @alantoa, added `transform` array support in 551f37ac44e0c20f8f255a95b3bc179084494346

src/reanimated2/js-reanimated/index.ts
```diff
@@ -47,14 +49,21 @@ export const _updatePropsJS = (
...
   const previousStyle = component.previousStyle ? component.previousStyle : {};
   const currentStyle = { ...previousStyle, ...style };
   component.previousStyle = currentStyle;
-  component.setNativeProps({ style: currentStyle });
+
+  const domStyle = createReactDOMStyle(currentStyle);
+  if (domStyle.transform) {
+    domStyle.transform = createTransformValue(domStyle.transform);
+  }
+  for (const key in domStyle) {
```
alantoa, line 64: I think this approach will result in duplicate `transform` object being passed to `domStyle`, you will receive a [warning log](https://github.com/necolas/react-native-web/blob/39b94b194594adebdfd5d4a5b873ced8d387734e/packages/react-native-web/src/exports/StyleSheet/preprocess.js#L203) even if it is working well.

tomekzaw: @alantoa Hey, what exactly do you mean by "duplicated `transform` object"?

I've compared 551f37ac44e0c20f8f255a95b3bc179084494346 to your suggestion from https://github.com/software-mansion/react-native-reanimated/pull/4352#discussion_r1182346178 and it looks like in both cases we call `createReactDOMStyle` first, then at some point we call `createTransformValue` for `domStyle.transform` and finally we assign its result to `component.style.transform` (and we assign to `domStyle.transform` only once). What's exactly wrong with overwriting `domStyle.transform`, maybe I'm missing something?

src/reanimated2/js-reanimated/index.ts
```diff
@@ -31,8 +33,8 @@ export const _updatePropsJS = (
       [{}, {}]
     );
 
-    if (typeof component.setNativeProps === 'function') {
-      setNativeProps(component, rawStyles);
+    if (component.style !== undefined) {
+      updatePropsDOM(component, rawStyles);
```
piaskowyk, line 37: Did you test it on an older version of `react-native-web`? I'm just wondering about backward compatibility. I'm ok if it works only with the newest version of `react-native-web` but if it's true we need to give users information about that.

src/reanimated2/js-reanimated/index.ts
```diff
@@ -31,8 +33,8 @@ export const _updatePropsJS = (
       [{}, {}]
     );
 
-    if (typeof component.setNativeProps === 'function') {
-      setNativeProps(component, rawStyles);
+    if (component.style !== undefined) {
```
piaskowyk, line 36: I remember that the else statement (`else if (Object.keys(component.props).length > 0)`) was prepared for handling some SVG's components. Did you check the `blame` if those particular components don't contain the `style` property? Just to be sure that we don't break something.

piaskowyk (commented):
In overall looks great 😊 
I just posted some questions just to be sure.

src/reanimated2/core.ts
```diff
@@ -28,6 +28,11 @@ import { SensorContainer } from './SensorContainer';
 export { stopMapper } from './mappers';
 export { runOnJS, runOnUI } from './threads';
 
+if (isWeb()) {
+  // @ts-ignore because of react-native-gesture-handler
+  global.setImmediate = global.queueMicrotask;
```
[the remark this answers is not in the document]
tomekzaw: Removed these lines and bumped react-native-gesture-handler to 2.10.0 in 24fe1b7c604f4a466c5159956e464e33b0f2760b

src/reanimated2/js-reanimated/index.ts
tomekzaw: It seems to work fine on `react-native-web@0.18.2` as well. Technically, we could have two implementations (new and legacy), one with `updatePropsDOM` (for `>=0.19`) and one with `setNativeProps` (for `<0.19`) but let's keep things simple.

src/reanimated2/js-reanimated/index.ts
tomekzaw: Looks like the lines you've mentioned were added in #2197 (see [here](https://github.com/software-mansion/react-native-reanimated/blame/437cb56ab7c567ab3ef0a63482bdfb06e03f2e5f/src/reanimated2/js-reanimated/index.ts#L47)) so most likely it's all about basic Chrome Debugger support. As for animating SVG components, somehow you were right, fixed in ac60d65e7d8826ace33d298f797c7d7fa2aabe44

WebExample/App.tsx
```diff
@@ -1,3 +1,3 @@
-import App from '../app';
+import App from '../app/src/examples/AnimatedStyleUpdateExample';
```
tomekzaw, line 1: For some reason React Navigation doesn't work properly with react-native-web  0.19 yet, here's the error:

Zrzut ekranu 2023-05-15 o 18 12 37

src/reanimated2/js-reanimated/index.ts
alantoa: oh yeah, i misread, never mind! thank you!

tomekzaw: no problem, thanks!

src/reanimated2/js-reanimated/index.ts
tomekzaw: I've built the package and tested two examples (AnimatedStyleUpdateExample + transforms & SvgExample) across two Expo Web apps (with react-native-web 0.18.10 and 0.19.4) and everything seems to work fine so far.

WebExample/App.tsx
tomekzaw: We decided to downgrade react-native-web in WebExample to 0.18.10 for easier development. We'll upgrade to 0.19 once React Navigation supports it.

piaskowyk (approved):

chrisedington:
Thanks @tomekzaw for the efforts.

src/reanimated2/js-reanimated/index.ts
```diff
@@ -1,5 +1,13 @@
 import JSReanimated from './JSReanimated';
 import { AnimatedStyle, StyleProps } from '../commonTypes';
+import createReactDOMStyle from 'react-native-web/dist/exports/StyleSheet/compiler/createReactDOMStyle';
+
+let createTransformValue: (transform: any) => any;
+try {
```
nandorojo, line 6: I think it would be better to require this in the conditional itself, rather than at the root. I'm not sure if `try catch` requires work with Webpack and most parsers.

PR #4339, opened 2023-04-06
Add KeyboardAvoidingView

terrysahaidak:
## Summary

This PR implements React Native KeyboardAvoidingView using `useAnimatedKeyboard` hook.

The goal of this implementation is to be always perfectly in sync with `useAnimatedKeyboard` so we can animate things based on the keyboard state in the same frame.
Another benefit of this implementation is that it doesn't rely on Layout Animation which sometimes makes things really hard.

<details>
<summary>
Using Reanimated's KeyboardAvoidingView</summary>

https://user-images.githubusercontent.com/7809008/230395743-787b3b48-2032-415b-a640-aa1cfabc7ee4.mov

</details>

<details>
<summary>
Using React Native's KeyboardAvoidingView</summary>

https://user-images.githubusercontent.com/7809008/230396044-1d83c911-3a9d-4762-9460-5c580700b00d.mov

</details>

Another benefit of this implementation is that we can actually introduce more properties and mechanics that are not as confusing as the current `behavior` and `keyboardVerticalOffset` props no one knows how to properly pair.

As part of this implementation, I'm adding an example I'm implementing right now for Expensify (https://github.com/Expensify/App/pull/16356) but you can see it in other apps such as Slack and Messenger.
In chat applications, KeyboardAvoidingView usually pushes the scroll view to the top alongside input. But if you want to show a modal instead of the keyboard, for example, simulating [InputAccessoryView](https://reactnative.dev/docs/inputaccessoryview), the keyboard hides without animation, and on the next frame the modal is rendered. So KeyboardAvoidingView no longer pushes the content since the keyboard is gone by now and all the content is now covered by our modal.
What we can do in this particular case is to `translateY` the content of the scroll view by the last keyboard height when the modal is visible.

https://user-images.githubusercontent.com/7809008/230379454-dca83e4b-2df5-4db3-b7a1-dd2d06fcdd2d.mov

There is a bug left with `useAnimatedKeyboard` which results into a few weird frames in this example, since it reports at some point:
```
{
  state: 4, // closed
  height: 318
}
```

## Test plan

Run KeyboardAvoidingView example.

PR #4360, opened 2023-04-13
Refactor keyboard observer on iOS

kmagiera:
## Summary

This PR refactors the logic in REAKeyboardEventObserver to address some number of different issues:
1) no change notification when keyboard size changes due to change in language or showing suggestions bar
2) issue with keyboard height notification when opening modal (see #4339)
3) issue with non-zero height after keyboard hides completely.

Here is the list of the main thing that this refactor changes:
1) We no longer use willShow/willHide/didShow/didHide notifications for keyboard and only use willChangeFrame which is the most reliable one and triggers also when the keyboard updates
2) We no longer use a technique where the keyboard size is determined based on the frame of the keyboard view. This was problematic as that frame would sometimes reflect a one-frame delayed size which resulted in the animation not coming down to 0 or not updating at all when immediately hidden or appear (by a modal). Instead, we use UIView animation to animate a complementary view that is placed outside of the screen visible area. We use UIView animation to animate height of that invisible view and then use display link to access presentation layer in order to read up-to-date height.

There are few gatchas related to the new technique used:
1) Because we only look at frame change notification we need to properly determine the state of the keyboard. We do it based on the target height. If it is 0 then we consider the state to be either closing or closed – we select the final state based on whether there is animation running (closed state when no animation, or closing when we are animating). This prevents us from changing state to closing when keyboard is only becoming shorter (e.g. when text suggestion bar is closed). We handle opening/open state in similar way.
2) Due to the UIView animation technique we need to check presentationLayer.animationKeys to determine if the layer runs the animation. If it doesn't then we extract height from the view frame and not from the presentationLayer frame, as the presentationLayer may not have an up-to-date information (it will be delayed by one frame).

## Test plan

Use example from #4339 and animated keyboard screen from Example app. Open/close text suggestion bar to see how the app reacts when keyboard is changing height and not just opening/closing.

### DIFF
--- a/ios/keyboardObserver/REAKeyboardEventObserver.m
+++ b/ios/keyboardObserver/REAKeyboardEventObserver.m
@@ -12,13 +12,12 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
 };
 
 @implementation REAKeyboardEventObserver {
+  UIView *_measuringView;
   NSNumber *_nextListenerId;
   NSMutableDictionary *_listeners;
-  CADisplayLink *displayLink;
-  int _windowsCount;
-  UIView *_keyboardView;
+  CADisplayLink *_displayLink;
   KeyboardState _state;
-  bool _shouldInvalidateDisplayLink;
+  CGFloat _lastKeyboardHeight;
 }
 
 - (instancetype)init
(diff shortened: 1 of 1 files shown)


terrysahaidak:
This is awesome @kmagiera!
I'll play around next week with this PR on my branch that uses useAnimatedKeyboard heavily and where all these problems were discovered https://github.com/Expensify/App/pull/16356

piaskowyk (approved):

tomekzaw (approved):
1. I can confirm this PR fixes the issue with missing height update on keyboard change. Here are the logs from toggling between English and Emoji keyboard. Note that for each change, there are exactly two updates in an interval of one frame (16 ms):



2. As for the issue with opening a modal, it looks like after these changes useAnimatedKeyboard-based KeyboardAvoidingView behaves identically to the one from React Native so I guess this is also solved. Also, here are the logs from opening a modal while keyboard is open:



3. I can also confirm this PR fixes the issue with non-zero height after keyboard hides completely. Previously, I was able to reproduce it consistently using the following conditional breakpoint:





4. Because of the shape of the animation curve, sometimes (in particular, at the beginning and at the end of animation) as well as the precision of the measurement, keyboard height doesn't change in up to 5 subsequent frames. In such case (when both height and state don't change), we could skip calling listeners on the native side.

https://user-images.githubusercontent.com/20516055/232465504-5f4858f1-df7f-4fdf-a579-39be82554187.mov

5. I know this is outside of scope of this PR but why we need to have two identical listeners in `useAnimatedKeyboard.ts`?

https://github.com/software-mansion/react-native-reanimated/blob/2e47c4a5884093c3bb7024f591a50ca4959ec031/src/reanimated2/hook/useAnimatedKeyboard.ts#L25-L29

https://github.com/software-mansion/react-native-reanimated/blob/2e47c4a5884093c3bb7024f591a50ca4959ec031/src/reanimated2/hook/useAnimatedKeyboard.ts#L37-L41

6. Unfortunately, KeyboardAvoidingViewExample crashes on reload (Example app, iOS). It's not a regression as it also does crash on the main branch but it's certainly worth looking into, especially since reloading works fine on other examples. Here's the stack trace:

Zrzut ekranu 2023-04-17 o 13 24 15

PR #4361, opened 2023-04-13
Rename JSI bindings to split Paper and Fabric types

tomekzaw:
## Summary

This PR splits JSI bindings like `_measure` which have different argument types on Paper and Fabric into two separate names, e.g. `_measurePaper` and `_measureFabric`, so that we can make better use of TypeScript. Additionally, error messages have been improved to provide better developer experience across all configurations.

## Current status

| Function | Paper UI | Fabric UI | native JS | Web | Chrome Debugger | Jest | fallback |
|-|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
| `updateProps` | ✅ | ✅ | ❓ | ✅ | ✅ | ✅ | ❓ |
| `measure` | ✅ | ✅ | ⚠️ | ✅ | ⚠️ | ⚠️ | ⚠️ |
| `scrollTo` | ✅ | ✅ | 👻 | ✅ | ⚠️ | ⚠️ | ⚠️ |
| `dispatchCommand` | ⚠️ | ✅ | 👻 | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
| `setGestureState` | ✅ | ✅ | ⚠️ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
| `LayoutAnimations` | n/a | n/a | ✅ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
| `useAnimatedKeyboard` | n/a | n/a | ✅ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
| `useAnimatedSensor` | n/a | n/a | ✅ | ✅ | ❓ | ❓ | ❓ |

Legend:
* ✅ officially supported
* ⚠️ `console.warn` + no-op / return empty value
* 👻 no-op (without warning)
* ❓ needs to be checked or fixed


tjzel (approved):
Looks good to me - I'm not a fan of if-else-return-noop spam, but I understand that using a general function there that also would define UI implementations for no-ops is too much hassle for this.

piaskowyk (approved):

Issue #4354, opened 2023-04-12
"console.assert is not a function" when debugging after upgrading to 3.0.2


NiharR27:
### Description

After upgrading from  "react-native-reanimated": "^2.14.4", to "react-native-reanimated": "^3.0.2", i am unable to debug.

I get this error:
```
e: TypeError: console.assert is not a function
    at pd (event-target-shim.js:46)
    at setPassiveListener (event-target-shim.js:487)
    at EventTarget.dispatchEvent (event-target-shim.js:812)
    at WebSocket.js:251
    at EventEmitter.emit (EventEmitter.js:105)
    at MessageQueue.__callFunction (MessageQueue.js:423)
    at MessageQueue.js:113
    at MessageQueue.__guard (MessageQueue.js:367)
    at MessageQueue.callFunctionReturnFlushedQueue (MessageQueue.js:112)
    at RNDebuggerWorker.js:2
```

downgrading back to 2.14.4 fixes the issue and debugger starts working back

Found similar threat on stack overflow: 
https://stackoverflow.com/questions/75689543/starting-debugger-error-console-assert-is-not-a-function and this 

possible similar issue on the project:
https://github.com/software-mansion/react-native-reanimated/issues/3951#issuecomment-1384223013

### Steps to reproduce

1. add react-native-reanimated: 3.0.2 (make sure to do pod install)
2. Start the app
3. Open debugger if you have (or use web debugger)
4. Debug
5. See the crash mentioned above

### Snack or a link to a repository

https://stackoverflow.com/questions/75689543/starting-debugger-error-console-assert-is-not-a-function

### Reanimated version

3.0.2

### React Native version

0.70.6

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 14

### Acknowledgements

Yes

tomekzaw:
Should be fixed with #4365

Issue #4263, opened 2023-03-23
Debug Mode is not working : Failed to print error:  Cannot read properties of undefined (reading 'apply')


Akilramki:
### Description

While Debug the App

console.assert is not a function

RCTFatal
-[RCTCxxBridge handleError:]
__21-[RCTCxxBridge start]_block_invoke
invocation function for block in facebook::react::(anonymous namespace)::RCTObjcExecutor::RCTObjcExecutor(id<RCTJavaScriptExecutor>, void (NSError*) block_pointer, std::__1::shared_ptr<facebook::react::MessageQueueThread>, std::__1<facebook::react::ExecutorDelegate>)
__58-[RCTWebSocketExecutor _executeJSCall:arguments:callback:]_block_invoke
-[RCTWebSocketExecutor webSocket:didReceiveMessage:]
__33-[RCTSRWebSocket _handleMessage:]_block_invoke
_dispatch_call_block_and_release
_dispatch_client_callout
_dispatch_lane_serial_drain
_dispatch_lane_invoke
_dispatch_workloop_worker_thread
![simulator_screenshot_EF15469B-D9C0-48AD-A9EA-A15A9DE66451](https://user-images.githubusercontent.com/105835418/227169508-9fdcb8b5-105d-43f5-a99a-39a606995831.png)

_pthread_wqthread
start_wqthread

### Steps to reproduce

Error:
1.console.assert is not a function 
2.Failed to print error:  Cannot read properties of undefined (reading 'apply')

### Snack or a link to a repository

https://stackoverflow.com/questions/75689543/starting-debugger-error-console-assert-is-not-a-function

### Reanimated version

^3.0.2

### React Native version

0.71.4

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

None

### Acknowledgements
![e6j9D](https://user-images.githubusercontent.com/105835418/227170758-5a9c2178-a1d4-4c00-80f9-e0e219d7b9a8.png)

Yes

JerakRus:
https://github.com/software-mansion/react-native-reanimated/issues/3951

Latropos:
Closing as a duplicate of #3951

PR #4365, opened 2023-04-14
fix: console.assert is not defined during debug

### LINKED ISSUE #3951: React Native 0.71.0 - Failed to print error:  Cannot read property 'apply' of undefined
### Description

console object missing `assert` and other methods. This causes the debugger to break. Can no longer to make websocket connections or hot module reload while developing.

### Steps to reproduce

1. Upgrade to react-native@0.71.0
2. Install `"react-native-reanimated": "3.0.0-rc.10",`
3. Run app in ios simulator

[https://github.com/rili-live/therr-app/tree/main/TherrMobile](https://github.com/rili-live/therr-app/tree/main/TherrMobile)

### Reanimated version

3.0.0-rc.10

### React Native version

0.71.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

KiwiKilian:
## Summary

Fixes #3951 #4263 #4354

## Test plan

I can now successfully debug using React Native Debugger again. Same fix was proposed by @rililive https://github.com/software-mansion/react-native-reanimated/issues/3951#issuecomment-1384223013

### DIFF
--- a/src/reanimated2/initializers.ts
+++ b/src/reanimated2/initializers.ts
@@ -171,6 +171,7 @@ export function initializeUIRuntime() {
     // setup console
     // @ts-ignore TypeScript doesn't like that there are missing methods in console object, but we don't provide all the methods for the UI runtime console version
     global.console = {
+      assert: runOnJS(capturableConsole.assert),
       debug: runOnJS(capturableConsole.debug),
       log: runOnJS(capturableConsole.log),
       warn: runOnJS(capturableConsole.warn),


tomekzaw:
Hey @KiwiKilian, thanks for your PR! In this comment I would like to provide more context for you and possibly for other future readers.

As you may have noticed, all `console.*` functions available in the UI runtime are copies of their counterparts from the main RN JS runtime but wrapped with `runOnJS`. This is necessary so that the output from the UI runtime also appears in the Metro console. It was way simpler to implement it this way compared to injecting our own JSI bindings which would need to somehow intercept the communication channel with Metro bundler. Unfortunately, the trick with `runOnJS` makes these functions asynchronous, and since we shouldn't block the UI thread, we cannot wait for the return value, so we just schedule their execution on the JS thread without stopping the UI runtime until the call completes. This is perfectly fine for methods such as `log` or `warn` since they don't affect the further execution of the UI runtime code in any way. Also, the order of such messages is correct for each runtime individually, but not for both runtimes at once (some messages from the UI runtime may appear after those sent from the main RN JS runtime, even though they could be scheduler earlier time-wise).

The reason we haven't ported all available methods yet is that we're not sure enough if it's fine to use each of them in an asynchronous way. For instance, in many programming languages, `assert` stops the execution if the condition is not met. [The reference](https://developer.mozilla.org/en-US/docs/Web/API/console/assert) says that 

> The console.assert() method writes an error message to the console if the assertion is false. If the assertion is true, nothing happens.

Can you please verify that the following code snippet prints "1 2 3" in both cases, i.e. when used from the main RN JS runtime and from the UI runtime?

```ts
console.log(1);
console.assert(false);
console.log(2);
console.assert(false);
console.log(3);
```

Also, since `console.assert` prints a stack trace, can you please verify that `runOnJS` doesn't break anything, i.e. if the stack trace is correct and the very top frame points to the correct location inside the worklet? It would be great if you could provide a screenshot.

Thanks again for your contribution!

piaskowyk:
I tested It and it seems to work properly

KiwiKilian:
@tomekzaw thanks for your extended explanation, sorry I hadn't much time available. So thanks @piaskowyk for testing and merging!

Issue #4177, opened 2023-03-06
makeShareableCloneRecursive Call Stack Size Exceeded


SamuelScheit:
### Description

I've switched from version 2.14.4 to the latest version 3.0.1 and get the error `Maximum call stack size exceeded` in the function `makeShareableCloneRecursive` from reanimated.

<div float="left">
  <img src="https://user-images.githubusercontent.com/34555296/223177813-00d3d7cf-e082-458a-9e4c-68b8255983e6.png" width="200" />
  <img src="https://user-images.githubusercontent.com/34555296/223177859-d8c72374-4511-44cd-9e4b-5a395a2f9cc6.png" width="200" />
</div>

### Steps to reproduce

I can't seem to isolate the issue, but it happened after upgrading to the latest version.
I've updated the Pods and reinstalled the App.
What could cause this kind of issue?

### Snack or a link to a repository

-

### Reanimated version

3.0.1

### React Native version

0.71.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

iPhone 12 mini

### Acknowledgements

Yes

jb4e:
We hit this as well when attempting to upgrade from v3 RC1. Haven't had a chance to dig in yet, but we will add more details if we can figure out a minimum repro.

philipheinser:
Added a log to the shareable clone and seeing this:

```
me": null, "_threadCount": 0, "displayName": "ReactRedux"}
 LOG  key $$typeof
 LOG  element Symbol(react.context)
 LOG  key _currentValue
 LOG  element null
 LOG  key _currentValue2
 LOG  element null
 LOG  key _threadCount
 LOG  element 0
 LOG  key Provider
 LOG  element {"$$typeof": Symbol(react.provider), "_context": {"$$typeof": Symbol(react.context), "Consumer": {"$$typeof": Symbol(react.context), "_context": [Circular]}, "Provider": [Circular], "_currentRenderer": {}, "_currentRenderer2": null, "_currentValue": null, "_currentValue2": null, "_defaultValue": null, "_globalName": null, "_threadCount": 0, "displayName": "ReactRedux"}}
[...]

PR #4367, opened 2023-04-14
Fix "Maximum call stack size exceeded" error

kmagiera:
## Summary

This PR addresses a number of issues reported related to exceeding stack when converting objects to shareables (e.g. in #4177).

This issue got introduced in Reanimated 3 after #4060 where we got rid of filtering only attributes accessed in worklets to be captured by worklets.

This PR implements a solution that filters only plain objects before converting things to shareables. In all of the instances of this issue we encountered, the crash was due to the fact we were leaking some react internals (fiber nodes) into the captured variables (e.g. via React's refs or otherwise). These objects are not of a plain object type and hence we wouldn't be able to support transferring these on the UI runtime anyways. With this change we are skipping these type of objects and replace them with a "inaccessible object" which is implemented as a proxy that throws on any attempt of accessing its attributes – this way users will get an error if they do use or access such objects.

In addition we add a cyclic object detection mechanism. Currently cyclic objects cannot be converted to sharables – they weren't supported in Reanimated 2 either. We add it as a part of this change in case the object prototype check wouldn't be sufficient in some cases. In a scenario when cyclic object is captured by a worklet we will detect that and throw a more descriptive error.

## Test plan

The below code illustrates a scenario when we are capturing react internals via ref:
```js
function Mleko({ props }: any) {
    console.log(props)
    return {}
  })
  return (
    <View ref={props.ref} />
  );
}

const ref = useRef();
<Mleko props={{ ref }} />
```

Before this change the above coud would throw stack exceeded error. Now this works ok.

In addition we tested cyclic detection by capturing the following object:
```js
const one = {};
const two = {one};
one.two = two;
```

The above object would throw stack exceeded error. now it throws an error that says that user is capturing cyclic object.

Lastly, we tested accessing non plain objects:
```js
class Something() {
  constructor() { this.ggg = 7 }
}
const something = new Something();

useAnimatedStyle(() => {
  console.log(something.ggg);
  return {};
});
```


src/reanimated2/shareables.ts
```diff
@@ -37,16 +37,74 @@ export function registerShareableMapping(
...
+            'Trying to write to an object which cannot be sent to the UI runtime.'
+          );
+        },
+      }
+    );
+  },
+};
+
+const ANALYZE_CYCLIC_OBJECT_AT_DEPTH = 30;
+let CYCLIC_OBJECT_TEST: any;
```
tomekzaw, line 79: I'm not a huge fan of these two names:
1. As for `ANALYZE_CYCLIC_OBJECT_AT_DEPTH`, we analyze all objects, not only the cyclic ones; also it's more of a lower bound for the depth rather than the depth itself
2. As for `CYCLIC_OBJECT_TEST`, this name gives no information on what's exactly being stored

src/reanimated2/shareables.ts
```diff
@@ -75,8 +79,10 @@ const INACCESSIBLE_OBJECT = {
   },
 };
 
-const ANALYZE_CYCLIC_OBJECT_AT_DEPTH = 30;
-let CYCLIC_OBJECT_TEST: any;
+const DETECT_CYCLIC_OBJECT_DEPTH_THRESHOLD = 30;
+// Below variable stores object that we process in makeShareableCloneRecursive at the specified depth.
+// We use it to check if later on the function reenters with the same object
+let procesedObjectAtThresholdDepth: any;
```
tomekzaw, line 85:
```suggestion
let processedObjectAtThresholdDepth: any;
```

src/reanimated2/shareables.ts
```diff
@@ -37,16 +37,74 @@ export function registerShareableMapping(
...
     return value;
   }
+  if (depth >= ANALYZE_CYCLIC_OBJECT_AT_DEPTH) {
+    // if we reach certain recursion depth we suspect that we are dealing with a cyclic object.
+    // this type of objects is not supported and cannot be trasferred as shareable so we
+    // implement a simple detection mechanism that remembers the value at a given depth and
+    // tests whether we try re-enter this method later on with the same value. If that happens
+    // we throw an appropriate error.
+    if (depth === ANALYZE_CYCLIC_OBJECT_AT_DEPTH) {
+      CYCLIC_OBJECT_TEST = value;
```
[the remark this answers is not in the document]
piaskowyk: I think that we should clear `CYCLIC_OBJECT_TEST` like Tomek said, or use some kind of `WeakRef`

piaskowyk (approved):

quant-daddy:
Hi how do I install this version before it's released to npm?

quant-daddy:
I'm getting this error after installing the latest version from Github:

```
Error: Trying to convert a cyclic object to a shareable. This is not supported., js engine: hermes
```

simulator_screenshot_F9721F0F-771C-4C3B-AEA2-1113E3881C5C

kmagiera:
Hey @quant-daddy – would you be able to test your app with this patch: #4372 – it won't fix the issue but should give us a more meaningful stack trace that should narrow down the actual object that's causing this exception

quant-daddy:
Okay so it seems to be working after install npm git+https://github.com/computerjazz/react-native-draggable-flatlist.git#pull/462

However I'm getting typescript error if I install react-native-draggable-flatlist directly from github. Would you know how to fix that?

kmagiera:
Hi @quant-daddy can you maybe share more details as to how you use that draggable-flatlist library? I just tried an example from the readme to run it with the most recent version of reanimated and I'm not getting the crash that you mentioned

PR #4368, opened 2023-04-14
Set the opposite sensor values on android to be consistent with iOS

mstach60161:
## Summary

The sensors on Android and iOS return opposite values, which cause inconsistencies. To fix this, I modified the Android sensor listener to return opposite values.
I didn't modify rotation values, because these are the same.

## Test plan

- Ran each sensor and compared the values returned by both platforms
- Tested an example application to confirm that the behavior of the sensors is consistent across both platforms.


piaskowyk (approved):
I tested it on iOS and Android and works 🎉

PR #4370, opened 2023-04-17
Add version property to worklets for Reanimated plugin version verification

tjzel:
## Summary

This PR modifies the way we verify if there is no mismatch between versions of Reanimated and Reanimated Babel plugin (for eg. when metro cache was not reset).

- Added `__version` prop to every worklet in DEV.
- Added version checking in runtime.
- Added global variable to ensure error for the mismatch is logged only once.
- Removed old `checkPluginVersion.ts` and moved `checkCppVersion.ts` back to `checkVersion.ts`.
- Changed plugin's code and jest tests accordingly.

## Test plan

0. Run `yarn jest`.
1. Make a DEV bundle to confirm `__version` prop appears on worklets.
2. Make a production to bundle to confirm it doesn't appear there.
3. Build ExampleApp and fiddle with `jsVersion.ts` to confirm that it's working.


__tests__/plugin.test.js
```diff
@@ -18,19 +18,20 @@ describe('babel plugin', () => {
...
       }
     </script>`;
 
     const { code } = runPlugin(input, {});
     const { version: packageVersion } = require('../package.json');
-    expect(code).toContain(
-      `global._REANIMATED_VERSION_BABEL_PLUGIN = "${packageVersion}"`
-    );
-    expect(code).not.toContain('inject Reanimated Babel plugin version');
+    expect(code).toContain(`f.__version = "${packageVersion}"`);
```
tomekzaw, line 32:
```suggestion
    expect(code).toContain(`f.__version = "${packageVersion}";`);
```

__tests__/plugin.test.js
```diff
@@ -18,19 +18,20 @@ describe('babel plugin', () => {
...
+
+    process.env.REANIMATED_PLUGIN_TESTS = 'jest';
```
tomekzaw, line 34: Let's remove this line and initialize the env variable in `beforeEach` instead of `beforeAll` so that we don't need to remember about cleanup here.

plugin/src/makeWorklet.ts
```diff
@@ -235,6 +237,17 @@ export function makeWorklet(
         )
       )
     );
+    if (!isRelease() && process.env.REANIMATED_PLUGIN_TESTS !== 'jest') {
```
tomekzaw, line 240: Let's move this condition to a separate function like so:
```ts
export function isJest() {
  return process.env.REANIMATED_PLUGIN_TESTS === 'jest';
}
```
Then let's wrap this condition in another function:
```ts
export function shouldInjectWorkletVersion() {
  return !isRelease() && !isJest();
}
```
This way we get more self-descriptive code.

plugin/src/buildWorkletString.ts
```diff
@@ -117,9 +117,9 @@ function shouldGenerateSourceMap() {
     return false;
   }
 
-  if (process.env.REANIMATED_PLUGIN_TESTS === 'jest') {
-    // We want to detect this, so we can disable source maps (because they break
-    // snapshot tests with jest).
+  // We want to detect this, so we can disable source maps (because they break
+  // snapshot tests with jest).
+  if (process.env.REANIMATED_PLUGIN_TESTS !== undefined) {
```
tomekzaw, line 122:
```suggestion
  if (isJest()) {
```

src/reanimated2/globals.d.ts
```diff
@@ -22,6 +22,7 @@ declare global {
   var _REANIMATED_VERSION_CPP: string | undefined;
   var _REANIMATED_VERSION_BABEL_PLUGIN: string | undefined;
   var __reanimatedModuleProxy: NativeReanimated | undefined;
+  var __versionMismatch: boolean | undefined;
```
tomekzaw, line 25: remove dis

src/reanimated2/shareables.ts
```diff
@@ -69,6 +70,15 @@ export function makeShareableCloneRecursive<T>(
         if (value.__workletHash !== undefined) {
           // we are converting a worklet
           if (__DEV__) {
+            if (
+              // We don't want this error to be logged more than once.
+              global.__versionMismatch === undefined &&
+              value.__version !== jsVersion
+            ) {
+              global.__versionMismatch = true;
```
tomekzaw, line 78: Let's use `let didShowError = false;` in top-level instead of using `global`.

__tests__/plugin.test.js
```diff
@@ -18,19 +18,20 @@ describe('babel plugin', () => {
   });
 
   it('injects its version', () => {
+    process.env.REANIMATED_PLUGIN_TESTS = 'jest version check';
```
tomekzaw, line 21: Instead of using a different string (i.e. `'jest version check'` vs `'jest'`) let's just remove this env variable for this test:
```suggestion
    delete process.env.REANIMATED_PLUGIN_TESTS;
```
If this doesn't work, we can just assign `undefined`.

tomekzaw (changes requested):

__tests__/versionCheck.test.js
```diff
@@ -1,40 +1,68 @@
...
     checkVersion();
     expect(console.error).not.toHaveBeenCalled();
   });
+
+  it('invalidates version post-string', () => {
+    global._REANIMATED_VERSION_CPP = packageVersion + '-post';
+    checkVersion();
+    expect(console.error).toHaveBeenCalled();
+  });
 });
```
tomekzaw, line 68: Honestly, I was rather thinking of some simple Jest unit tests which just import and test `matchVersion` function just like:
```ts
it('...', () => {
  expect(matchVersion('3.0.0', '3.0.1')).toBeTrue();
  // ...
});
```

plugin/src/makeWorklet.ts
```diff
@@ -244,6 +257,16 @@ export function makeWorklet(
...
 }
 
+function shouldInjectVersion() {
+  // We don't inject version in release since cache is reset there anyway
+  // We don't want to pollute tests with current version number so we disable it
+  // for all tests (except one)
+  if (isRelease() || process.env.REANIMATED_JEST_DISABLE_VERSION === 'jest') {
+    return false;
+  }
+  return true;
```
tomekzaw, line 267:
```suggestion
  // We don't inject version in release since cache is reset there anyway
  if (isRelease()) {
    return false;
  }

  // We don't want to pollute tests with current version number so we disable it
  // for all tests (except one)
  if (process.env.REANIMATED_JEST_DISABLE_VERSION === 'jest') {
    return false;
  }

  return true;
```

src/reanimated2/shareables.ts
```diff
@@ -37,6 +38,8 @@ export function registerShareableMapping(
   _shareableCache.set(shareable, shareableRef || _shareableFlag);
 }
 
+let pluginVersionMismatch = false;
```
tomekzaw, line 41:
```suggestion
let didShowPluginVersionMismatchError = false;
```

src/reanimated2/shareables.ts
```diff
@@ -69,6 +72,15 @@ export function makeShareableCloneRecursive<T>(
         if (value.__workletHash !== undefined) {
           // we are converting a worklet
           if (__DEV__) {
+            if (
+              // We don't want this error to be logged more than once.
+              pluginVersionMismatch === false &&
```
tomekzaw, line 77:
```suggestion
              !didShowPluginVersionMismatchError &&
```

src/reanimated2/platform-specific/checkVersion.ts
```diff
@@ -1,7 +1,33 @@
-import { checkCppVersion } from './checkCppVersion';
-import { checkPluginVersion } from './checkPluginVersion';
+import { jsVersion } from './jsVersion';
 
-export function checkVersion(): void {
+export function checkVersion() {
   checkCppVersion();
-  checkPluginVersion();
+}
```
tomekzaw, line 5: Can we remove `checkVersion` function and export `checkCppVersion` function directly?

tomekzaw (changes requested):

src/reanimated2/platform-specific/checkVersion.ts
tjzel: Sure ting!

__tests__/versionCheck.test.js
tjzel: Honestly after I wrote the first test Github Copilot offered me the rest of the them for free (I only had to make minor adjustments).

tomekzaw (approved):

PR #4374, opened 2023-04-18
Update misleading createAnimatedComponent warning

kacperkapusciak:
## Summary

When writing a new documentation for `createAnimatedComponent` I've found out that this function works not only for class-based components but also for functional components with `React.forwardRef()`. 

This PR updates the warning caused by using plain functional components to guide users to use forwardRef.

## Test plan

When plain functional component is passed to `createAnimatedComponent` the app errors out:

```jsx
function MyComponent(props) {
  return <View {...props} />;
}

const MyAnimatedComponent = Animated.createAnimatedComponent(MyComponent);
```

but warning guides to use `React.forwardRef()`.

image

When a functional component is wrapped with forwardRef it works just fine:

```jsx
function WrappedView(props, ref) {
  return <View {...props} ref={ref} />;
}

const MyComponent = React.forwardRef(WrappedView);
const MyAnimatedComponent = Animated.createAnimatedComponent(MyComponent);
```

image

<details>
<summary>Ful example to copy & paste</summary>

```jsx
import React from 'react';
import {Button, StyleSheet, View} from 'react-native';
import Animated, {useSharedValue, withSpring} from 'react-native-reanimated';

function WrappedView(props, ref) {
  return <View {...props} ref={ref} />;
}

const MyComponent = React.forwardRef(WrappedView);
const MyAnimatedComponent = Animated.createAnimatedComponent(MyComponent);

  const width = useSharedValue(100);
  const handlePress = () => {
    width.value = withSpring(width.value + 50);
  };

  return (
      <MyAnimatedComponent style={{...styles.box, width}} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
  },
  box: {
    height: 100,
    backgroundColor: '#b58df1',
    borderRadius: 20,
    marginVertical: 64,
  },
});

```

</details>

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -246,8 +246,7 @@ export default function createAnimatedComponent(
   invariant(
     typeof Component !== 'function' ||
       (Component.prototype && Component.prototype.isReactComponent),
-    '`createAnimatedComponent` does not support stateless functional components; ' +
-      'use a class component instead.'
+    `Looks like you're passing a ${Component.name} functional component to 'createAnimatedComponent'.\n\n'createAnimatedComponent' relies on passing refs to work correctly. Wrap your functional component with React.forwardRef() or use a class component instead.`
   );
 
   class AnimatedComponent extends React.Component<
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -246,7 +246,7 @@ export default function createAnimatedComponent(
   invariant(
     typeof Component !== 'function' ||
       (Component.prototype && Component.prototype.isReactComponent),
-    `Looks like you're passing a ${Component.name} functional component to 'createAnimatedComponent'.\n\n'createAnimatedComponent' relies on passing refs to work correctly. Wrap your functional component with React.forwardRef() or use a class component instead.`
+    `Looks like you're passing a function component \`${Component.name}\` to \`createAnimatedComponent\` function which supports only class components. Please wrap your function component with \`React.forwardRef()\` or use a class component instead.`
   );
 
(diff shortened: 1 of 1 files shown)


src/createAnimatedComponent.tsx
```diff
@@ -246,8 +246,7 @@ export default function createAnimatedComponent(
   invariant(
     typeof Component !== 'function' ||
       (Component.prototype && Component.prototype.isReactComponent),
-    '`createAnimatedComponent` does not support stateless functional components; ' +
-      'use a class component instead.'
+    `Looks like you're passing a ${Component.name} functional component to 'createAnimatedComponent'.\n\n'createAnimatedComponent' relies on passing refs to work correctly. Wrap your functional component with React.forwardRef() or use a class component instead.`
```
tomekzaw, line 249: 1. The official name is "function components" ([source](https://legacy.reactjs.org/docs/components-and-props.html#function-and-class-components)).
2. We use backticks (`) to quote identifiers and other names in error messages.
3. Instead of complaining, why don't we just wrap the input component with `React.forwardRef` on Reanimated side if needed?
```suggestion
    `Looks like you're passing a function component \`${Component.name}\` to \`createAnimatedComponent\` function which supports only class components. Please wrap your function component with \`React.forwardRef()\` or use a class component instead.`
```

tomekzaw (approved):
Thanks for the PR!

src/createAnimatedComponent.tsx
kacperkapusciak: Sorry I don't have the capacity right now to do more research on this.

Animated API just throws a warning prompting the user to wrap function component with `forwardRef`. Maybe they had a reason to do so 🤔

PR #4377, opened 2023-04-19
Fix undefined `ErrorUtils.reportFatalError`

piaskowyk:
## Summary

The `__ErrorUtils` is available only in Reanimated Runtime, to use `reportFatalError` we need to use ReactNative internal function available in `ErrorUtils` global object.

## Test plan

```
function fn() {
  'worklet'
  throw new Error('error');
}
runOnUI(fn)()
```


tjzel (approved):
LGTM

tomekzaw (approved):

PR #4385, opened 2023-04-20
make build work on m1/m2 without rosetta

philipheinser:

### DIFF
--- a/android/CMakeLists.txt
+++ b/android/CMakeLists.txt
@@ -1,5 +1,5 @@
 project(Reanimated)
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.22.1)
 
 set (CMAKE_VERBOSE_MAKEFILE ON)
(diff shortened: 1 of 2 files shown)


piaskowyk:
I am not sure if it is a problem with the `CMake` version. It's more looks like a problem with the NDK version. With NDK version do you use?

philipheinser:
@piaskowyk 
> I am not sure if it is a problem with the `CMake` version. It's more looks like a problem with the NDK version. With NDK version do you use?

ndkVersion = "25.2.9519653"

tjzel:
Hi @philipheinser 

Could you elaborate more about what exactly are you fixing?

philipheinser:
@tjzel Without it wouldn’t build on m2 arm chip without Rosetta before

tjzel:
Since we rely on `x86` binaries when we develop `react-native-reanimated` and this is the only request to make it buildable without Rosetta we're closing this PR for now.

PR #4388, opened 2023-04-21
Add Shared element Transition CameraRoll example 

terrysahaidak:
This example mimics the previous LightBox (CameraRoll) example I implemented for Reanimated 2 but now it's much easier with SET API

Still have a few bugs though. I'll open separate issues for each one.

## Summary

The motivation behind this PR is to show how easy is to build a camera roll with SET API. It just works. 

https://user-images.githubusercontent.com/7809008/233576800-8f228ac4-e5a7-4bad-a434-d333f05e2d34.mov


Bayramito:
What if you pass inverted=true prop to your list ?

PR #4389, opened 2023-04-21
Release 2.16.0

tomekzaw:
## Summary

Release 2.16.0

### DIFF
--- a/Example/ios/Podfile.lock
+++ b/Example/ios/Podfile.lock
@@ -394,7 +394,7 @@ PODS:
     - React-Core
   - RNGestureHandler (2.9.0):
     - React-Core
-  - RNReanimated (2.15.0):
+  - RNReanimated (2.16.0):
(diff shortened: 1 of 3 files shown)


piaskowyk (approved):

sunnylqm:
seems it was not compiled (missing `lib` folder) @tomekzaw

tomekzaw:
@sunnylqm Thanks for reporting this. Indeed, for some reason 2.16.0 doesn't have `lib` folder while 2.15.0 does. That's weird because we always build the package on CI. Also, before the release, I've tested the package on Android and iOS and it worked correctly but the problem went undetected because by default, RN uses the files from `src` directory instead of `lib`.

tomekzaw:
Fixed in [2.17.0](https://github.com/software-mansion/react-native-reanimated/releases/tag/2.17.0).

PR #4397, opened 2023-04-24
Fix merge queue - Remove branch specifications

Latropos:


piaskowyk:
I tested it on the new Github org, but unfortunately it doesn't seem to work :(

PR #4401, opened 2023-04-24
Unfreeze UI objects :cold_face: -> :sunglasses:

piaskowyk:
## Summary

The goal of freezing is to freeze JS objects used on the UI thread. But we don't need and don't want to freeze objects used only on the UI thread.

## Test plan

```js
  const sv1 = useSharedValue([0]);
  const sv2 = useSharedValue({ prop: [0] });
  runOnUI(() => {
    'worklet';
    sv1.value[0]++;
    sv2.value.prop[0]++;
    console.log(sv1.value, sv2.value.prop, Object.isFrozen(sv1.value));
  })()
```

Output before:
```
 LOG  RENDER -----------------------------------
 LOG  [1] [1] false
 LOG  RENDER -----------------------------------
 LOG  [1] [1] true
 LOG  RENDER -----------------------------------
 LOG  [1] [1] true
```

Output after:
```
 LOG  RENDER -----------------------------------
 LOG  [1] [1] false
 LOG  RENDER -----------------------------------
 LOG  [2] [2] false
 LOG  RENDER -----------------------------------
 LOG  [3] [3] false
```

### DIFF
--- a/src/reanimated2/shareables.ts
+++ b/src/reanimated2/shareables.ts
@@ -230,10 +230,6 @@ export function makeShareableCloneOnUIRecursive<T>(value: T): ShareableRef<T> {
           toAdapt[key] = cloneRecursive(element);
         }
       }
-      if (__DEV__) {
-        // See the reasoning behind freezing in the other comment above.
-        Object.freeze(value);
-      }
       return _makeShareableClone(toAdapt);
     }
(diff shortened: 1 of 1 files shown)


tjzel:
🥶 🍨

kmagiera (approved):

Issue #4306, opened 2023-03-30
[Android] `<Animated.ScrollView />` is crashing on `navigate.goBack()` when using `refreshControl` prop


vozaldi:
### Description

This issue occurs after I upgraded from v2 to v3 in React Native 0.71.4.

My app keep crashing when I opened a page and pressing back button (or executing `navigation.goBack()`).

I tried to reproduce the issue using bare React Native 0.71.4 project and installed both Reanimated v3 and latest React Navigation.

I found that these props are having weird behavior:
- `Animated.ScrollView.refreshControl`
- `Animated.View.exiting`

Whenever I add `<Animated.ScrollView refreshControl={{....}} />` with children `<Animated.View exiting={...} />` the app is crashing when navigating back.

``` JSX
<Animated.ScrollView
  refreshControl={(
    <RefreshControl refreshing={refreshing} />
  )}
>
  <Animated.View
    entering={ZoomIn}
    exiting={ZoomOut}
  />
</Animated.ScrollView>
```

Tested on Android device and emulator (Android 13, SDK 33). Crashing in both debug and production mode.

### Steps to reproduce

Crash Repro:
1. Clone or download the repo.
2. Run the `yarn android`.
3. In Home page, press "Go to Page A" button.
4. In Page A, press "Check Crash" button or just navigate back using back button.
5. The app crashes.

Not Crashing:
1. In file PageA.tsx, comment one of these prop `refreshControl` or `exiting`
2. Run the app and navigation works normally.

### Snack or a link to a repository

https://github.com/vozaldi/repro-rn-reanimated-crash

### Reanimated version

3.0.2

### React Native version

0.71.4

### Platforms

Android

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

acollazomayer:
I am experiencing the same issue on android.

vozaldi:
Update: tested on iOS and everything is working fine.

vozaldi:
Update (2):

The issue disappeared when I changed the navigator from `@react-navigation/stack` to `@react-navigation/native-stack`. So the crash is only happening with `@react-navigation/stack` component.

[...]

PR #4404, opened 2023-04-25
Fix Layout Animations iOS crash on reload

tomekzaw:


piaskowyk (approved):
Well done!

Latropos:
Looks that you've fixed this issue and we can close it: https://github.com/software-mansion/react-native-reanimated/issues/4306 🚀

PR #4407, opened 2023-04-27
Fix #3771 for more use cases

### LINKED ISSUE #3771: POD install failed when the project path contains a space
### Description

We experience an issue when running pod install in one of our projects, below is the error log in our error reproduction:

```
[!] Invalid `Podfile` file: 
[!] Invalid `RNReanimated.podspec` file: different prefix: "" and "/Users/xxx/Desktop/tmp/test bug/AwesomeProject/ios/Pods".

 #  from /Users/xxx/Desktop/tmp/test bug/AwesomeProject/node_modules/react-native-reanimated/RNReanimated.podspec:5
 #  -------------------------------------------
 #  reanimated_package_json = JSON.parse(File.read(File.join(__dir__, "package.json")))
 >  config = find_config()
 #  assert_no_multiple_instances(config)
 #  -------------------------------------------
.

 #  from /Users/xxx/Desktop/tmp/test bug/AwesomeProject/ios/Podfile:8
 #  -------------------------------------------
 #  target 'AwesomeProject' do
 >    config = use_native_modules!
 #  
 #  -------------------------------------------
```

After research, we found out that it is caused by missing quotation marks in 
node_modules/react-native-reanimated/scripts/reanimated_utils.rb

Here is my quick fix which can solve the issue:

```
Original:
react_native_node_modules_dir = File.join(File.dirname(`cd #{Pod::Config.instance.installation_root.to_s} && node --print "require.resolve('react-native/package.json')"`), '..')

Change to:
react_native_node_modules_dir = File.join(File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('react-native/package.json')"`), '..')
```

### Steps to reproduce

(1) Create a folder with space (e.g. "Hello world")
(2) Create a new react native project inside the new folder (e.g. "testbug", relative path: "Hello world\testbug")
(3) Install latest react-native-reanimated 2.12.0
(4) cd ios & pod install
(5) Shows "Invalid Podfile file: ..."

n/a

### Reanimated version

2.12.0

### React Native version

0.66.4

### Platforms

Android, iOS

### JavaScript runtime

_No response_

### Workflow

_No response_

### Architecture

_No response_

### Build type

_No response_

### Device

_No response_

### Device model

_No response_

### Acknowledgements

Yes

fwon:
## Summary

When react native is integrated with Existing Apps, the ios folder is not inside RN project, but outside of RN project.  cd to "Pod::Config.instance.installation_root.to_s" cannot resolve "react-native/package.json", then react_native_node_modules_dir will not be correct. So we should change it into "Dir.pwd", which directly resolve path from this library itself, not matter where the native iOS folder is.

## Test plan

Move react native project into a existing ios project. and run "pod install".

### DIFF
--- a/scripts/reanimated_utils.rb
+++ b/scripts/reanimated_utils.rb
@@ -17,7 +17,7 @@ def find_config()
     :react_native_common_dir => nil,
   }
 
-  react_native_node_modules_dir = File.join(File.dirname(`cd "#{Dir.pwd}" && node --print "require.resolve('react-native/package.json')"`), '..')
+  react_native_node_modules_dir = File.join(File.dirname(`cd "#{__dir__}" && node --print "require.resolve('react-native/package.json')"`), '..')
   react_native_json = try_to_parse_react_native_package_json(react_native_node_modules_dir)
 
   if react_native_json == nil
--- a/scripts/reanimated_utils.rb
+++ b/scripts/reanimated_utils.rb
@@ -17,7 +17,7 @@ def find_config()
     :react_native_common_dir => nil,
   }
 
-  react_native_node_modules_dir = File.join(File.dirname(`cd "#{Pod::Config.instance.installation_root.to_s}" && node --print "require.resolve('react-native/package.json')"`), '..')
+  react_native_node_modules_dir = File.join(File.dirname(`cd "#{Dir.pwd}" && node --print "require.resolve('react-native/package.json')"`), '..')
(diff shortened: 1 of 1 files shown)


piaskowyk:
It seems like you have a rare project configuration. I think the better solution in this case, would be to provide an environment variable to set a custom path to React Native location. What do you think?

PR #4414, opened 2023-04-27
Add scrollViewOffset prop to AnimatedScrollView

mstach60161:
## Summary

This PR moves logic from `useScrollViewOffset` hook to `Animated.ScrollView` prop.

The API of `useScrollViewOffset` is a bit cumbersome, because we need to pass the same `ref` to both `ScrollView` and hook. 

Now we just need to pass variable with type `ScrollValue<number>` to `Animated.ScrollView` and when Scroll View changes its position, our value is updated

## Test plan

- Example app


src/reanimated2/component/ScrollView.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+      useEffect(() => {
+        if (aref !== null) {
+          const viewTag = findNodeHandle(
+            (aref as RefObject<Animated.ScrollView>).current
+          );
+          event.current?.registerForEvents(viewTag as number);
+        }
+      }, [(aref as RefObject<Animated.ScrollView>).current]);
+    }
```
piaskowyk, line 52: Why we can't just use `useScrollHandler`? It's look like code duplication with: https://github.com/software-mansion/react-native-reanimated/blob/%40mstach60161/add-scrollViewOffsetProp-to-AnimatedScrollView/src/reanimated2/hook/useScrollViewOffset.ts

piaskowyk (approved):
👏

PR #4415, opened 2023-04-28
expose convertToRGBA to the types

wcandillon:
Because Skia is using a different type this function is useful when interpolating colors therefore it would be useful to have it exposed through the types


piaskowyk (approved):
Thanks @wcandillon ! 🙌

PR #4421, opened 2023-05-02
[Shared Element Transition] Progress animation v2

piaskowyk:
## Summary

This pull request introduces a new progress-based shared element transition animation that utilizes the progress value from `react-native-screens`. This value indicates the progress of the screen's transition.

## Changes

1. Now it is possible to use two different types of animation:
 - standard animation (uses `withTiming` etc.)
 - progress-based animation (uses transition progress from `react-native-screens`). This is the default type of animation for `SET`, but a user can change the default type.
2. It is possible to define custom standard and progress-based animations.

## Implementation

The new implementation is based on event listeners. Events come from `rn-screens`. I added a new event to handle swipe back cancel on iOS - PR here - https://github.com/software-mansion/react-native-screens/pull/1810

The approach assumes that exists one `onTransitionProgress` listener that computes all progress animations and updates all shared views.

### Scenario 1:
Scenario assumption: The application consists of two screens - `ScreenA` and `ScreenB`, each having a shared element. When the user navigates from `ScreenA` to `ScreenB`, the following steps occur:

#### Part 1: JS, JS Thread
1. Creation of the JS component on the JS thread.
2. The `createAnimatedComponent` method is called within which the `SharedTransition.registerTransition()` method is invoked. The `SharedTransition.registerTransition()` method performs the following tasks:
    * Sends information about the view and animation to the native side (`LayoutAnimationManager.cpp`).
    * Adds information about the transition progress to `ProgressTransitionManager`.
    * Registers a listener for the `transitionProgress` event.

#### Part 2*: iOS, Native side, UI Thread
1. After mounting `ScreenB`, the `swizzled_viewDidLayoutSubviews` method is invoked, which is a swizzled version of the RNScreen's `viewDidLayoutSubviews` method.
    * This method is important because it ensures that all views on `ScreenB` have the correct layout, allowing for the creation of snapshots with accurate component styles. What is important, the screen header is attached too!
2. The `swizzled_viewDidLayoutSubviews` method calls the `LayoutAnimationManager.startAnimationForTag` method, which is a JS method that exists on the UI Thread.

#### Part 2*: Android, Native side, UI Thread
1. After mounting `ScreenB`, the `SharedTransitionManager.screenDidLayout()` method is invoked. We don't need swizzling technique here like it was on iOS. We can detect when the screen is ready and all components have the proper layout in the `ReanimatedNativeHierarchyManager.updateLayout` method.
2. The `screenDidLayout` method calls the `startAnimationForTag` method, which is a JS method that exists on the UI Thread.

#### Part 3: JS, UI Thread
1. The `LayoutAnimationManager[.cpp].startAnimationForTag` method, which is equivalent to `LayoutAnimationsManager[.js].start()`, is called.
2. The `start()` method invokes the `ProgressTransitionManager.onTransitionStart()` method responsible for saving snapshots with target style values.
4. At this point, everything is ready to react for the `RNScreen`'s `transitionProgress` event.

#### Part 4: JS, Event Listener, UI Thread
1. Whenever the `transitionProgress` event occurs, the `ProgressTransitionManager.frame()` method is executed.
2. The `frame()` method computes the current styles for each registered shared view and applies those styles using the `_notifyAboutProgress` method.
3. Upon the occurrence of the screen's `onAppear` or `onDisappear` event, the `ProgressTransitionManager.onTransitionEnd()` method is called. This method invokes the native `_notifyAboutEnd` method responsible for cleaning and reparenting views after the shared transition.

### Scenario 2:
Assumption: When the user navigates back from `ScreenB` to `ScreenA`, the flow is similar to Scenario 1, with the only difference in Part 2.

#### Part 2*: iOS
1. Before removing `ScreenB`, the `swizzled_notifyWillDisappear` method is called, which is a swizzled version of the `RNScreen`'s `notifyWillDisappear` method. This method is invoked always before the screen removal from the top of the stack. Fact that screen is removed from the stack doesn’t mean that this screen is removed from React/Native tree too. It is applicable in three different iOS navigation scenarios:
    * Calling `navigation.navigate()` invokes `notifyWillDisappear`, pops the screen from the stack, removes it from the React tree, starts the transition, and finally after transition end, removes the screen from the native tree.
    * Using the header's "go back" button invokes `notifyWillDisappear`, pops the screen from the stack and starts the transition. After transition end removes the screen from the React tree and native tree after.
    * Interactive swiping back invokes `notifyWillDisappear`, and based on user actions, cancels the swipe and performs no action or completes the swipe to remove the screen. Then, the screen is popped from the stack, removed from the React tree, and finally removed from the native tree.

#### Part 2*: Android
1. On removing `ScreenB`, the `SharedTransitionManager[.java].onViewsRemoval()` method is called, followed by starting the transition for the views on this screen. Unlike iOS, Android does not have interactive swiping. The "go back" button in the header works the same like calling `navigation.navigate()`, allowing for a simpler method to determine when to initiate a transition.

### Good to know:
* _How is this change coupled with screens implementation and why?_ The change involves registering for the `transitionProgress` event of the screen and then computing and applying new styles to the shared views within the event callback.
* _What_ is the role of `ProgressTransitionManager`?_ The `ProgressTransitionManager` is an object that exists on the UI Runtime. Its role is to store snapshots, view tags, and progress callbacks for shared views participating in the current transition. The manager also contains a `frame()` function that is used as an event callback for the `transitionProgress` event. During this event, the `ProgressTransitionManager` iterates over the stored view tags, computes new styles for each of them, and applies these styles to the views using the `_notifyAboutProgress` method.
* Clearing configuration from `LayoutAnimationManager.cpp`: The configuration is cleared in two specific cases:
    1. After the transition is finished (`finishSharedAnimation`).
    2. After the removal of a screen or stack (`onScreenRemoval`).

## Current shared transition configuration API 
```js
const transition = SharedTransition
  .duration(3000)
  .animation((values) => {
    'worklet';
    return {
      originX: withSpring(values.targetOriginX),
      originY: withSpring(values.targetOriginY),
    };
  })
  .progressAnimation((values, progress) => {
    'worklet';
    return {
      width: progress * (values.targetWidth - values.currentWidth) + values.currentWidth,
      height: progress * (values.targetHeight - values.currentHeight) + values.currentHeight,
      originX: progress * (values.targetOriginX - values.currentOriginX) + values.currentOriginX,
      originY: progress * (values.targetOriginY - values.currentOriginY) + values.currentOriginY,
    };
  })
  .defaultTransitionType(SharedTransitionType.ANIMATION);
```

This PR is based on https://github.com/software-mansion/react-native-reanimated/pull/4278 but I simplified the native part of the code responsible for worklet callback registration:
Previous:
image


Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -16,13 +16,21 @@ void LayoutAnimationsManager::configureAnimation(
   if (type == SHARED_ELEMENT_TRANSITION) {
     sharedTransitionGroups_[sharedTransitionTag].push_back(tag);
     viewTagToSharedTag_[tag] = sharedTransitionTag;
+  } else if (type == SHARED_ELEMENT_TRANSITION_PROGRESS) {
+    if (sharedTransitionTag == "secondary") {
+      sharedTransitioinProgressAnimationsLowPriotity_.insert(tag);
```
tomekzaw, line 21:
```suggestion
      sharedTransitionProgressAnimationsLowPriotity_.insert(tag);
```

app/src/examples/SharedElementTransitions/CustomTransition.tsx
```diff
@@ -12,15 +12,34 @@ import Animated, {
...
-  'worklet';
-  return {
-    width: withSpring(values.targetWidth),
-    height: withSpring(values.targetHeight),
-    originX: withSpring(values.targetOriginX),
-    originY: withSpring(values.targetOriginY),
-  };
-});
+const transition = new SharedTransition()
+  .transitionDuration(1000)
```
tomekzaw, line 16:
```suggestion
  .duration(1000)
```

app/src/examples/SharedElementTransitions/CustomTransition.tsx
```diff
@@ -12,15 +12,34 @@ import Animated, {
...
-const transition = SharedTransition.custom((values) => {
-  'worklet';
-  return {
-    width: withSpring(values.targetWidth),
-    height: withSpring(values.targetHeight),
-    originX: withSpring(values.targetOriginX),
-    originY: withSpring(values.targetOriginY),
-  };
-});
+const transition = new SharedTransition()
```
tomekzaw, line 15:
```suggestion
const transition = new SharedElementTransition()
```

app/src/examples/SharedElementTransitions/CustomTransition.tsx
```diff
@@ -12,15 +12,34 @@ import Animated, {
...
+        values.currentWidth,
+      height:
+        progress * (values.targetHeight - values.currentHeight) +
+        values.currentHeight,
+      originX:
+        progress * (values.targetOriginX - values.currentOriginX) +
+        values.currentOriginX,
+      originY:
+        progress * (values.targetOriginY - values.currentOriginY) +
+        values.currentOriginY,
```
tomekzaw, line 40: plz use interpolate

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
kacperkapusciak: sharedTransitionProgressAnimationsLowPriority

there's a typo in `priority` as well 🙊

src/reanimated2/layoutReanimation/sharedTransitions/index.ts
```diff
@@ -1,40 +1,90 @@
...
+  public registerTransition(
+    viewTag: number,
+    sharedTransitionTag: string
+  ): void {
+    const transitionAnimation = this.getTransitionAnimation();
+    const progressAnimation = this.getProgressAnimation();
+    let progressAnimationPriority = 'primary';
+    if (this._animationFactory) {
+      if (Platform.OS === 'ios') {
+        progressAnimationPriority = 'secondary';
```
tomekzaw, line 53: Instead of keeping these as strings, can we convert them to numeric enums or just use a boolean flag?

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -413,6 +494,11 @@ - (void)clearConfigForStack:(UIView *)stack
   }
 }
 
+- (void)maybeClearConfigForStackLeater:(UIView *)stack
```
tomekzaw, line 497:
```suggestion
- (void)maybeClearConfigForStackLater:(UIView *)stack
```

app/src/examples/SharedElementTransitions/Profiles.tsx
```diff
@@ -40,7 +40,7 @@ const springOptions = {
   damping: 15,
 };
 
-const transition = SharedTransition.custom((values) => {
+const transition = new SharedElementTransition().animation((values) => {
```
tomekzaw, line 43: 1. Do you think we could provide an API that omits the `new` keyword? I know we need to create and configure a new object but maybe we could provide some factory object/function.
2. What do you recommend in terms of optimization? Is it possible to implement SETs outside of components? Do we need to recommend memoization with `useMemo` or even enable it by default?
3. Technically, progress animation is also an animation. What about renaming `animation` to `targetAnimation` so we have `targetAnimation` for run-and-forget animations (e.g. `withSpring`) and `progressAnimation` for progress-based animations (e.g. `interpolate`)?

src/reanimated2/layoutReanimation/sharedTransitions/index.ts
```diff
@@ -1,40 +1,90 @@
...
+  values: SharedTransitionAnimationsValues
+) => StyleProps;
 
-export class SharedTransition implements ILayoutAnimationBuilder {
-  animationFactory: AnimationFactoryType | null = null;
-  defaultDuration = 500;
+export class SharedElementTransition {
+  private _animationFactory: AnimationFactory | null = null;
+  private _animation: SharedTransitionAnimationsFunction | null = null;
+  private _transitionDuration = 500;
```
tomekzaw, line 21: Wasn't it 1000 somewhere else?

Common/cpp/LayoutAnimations/LayoutAnimationsManager.h
```diff
@@ -48,6 +54,9 @@ class LayoutAnimationsManager {
   std::unordered_map<int, std::shared_ptr<Shareable>> layoutAnimations_;
   std::unordered_map<int, std::shared_ptr<Shareable>>
       sharedTransitionAnimations_;
+  std::unordered_map<int, std::shared_ptr<Shareable>>
+      sharedTransitioinProgressAnimations_;
+  std::unordered_set<int> sharedTransitionProgressAnimationsLowPriority_;
```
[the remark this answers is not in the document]
kmagiera: +1

android/src/main/cpp/JNIHelper.cpp
```diff
@@ -52,8 +52,47 @@ jni::local_ref<JNIHelper::PropsMap> JNIHelper::ConvertToPropsMap(
       }
     }
   }
-
   return map;
 }
 
+jsi::Object JNIHelper::convertJNIMapToJSIObject(
```
kmagiera, line 58: I don't quite understand why the conversion is part of this PR. Why do we need it with progress animation and didn't need this before? Seem to me like the progress animation shouldn't add a lot of additional complexity as this is just a single progress number that we are dealing with

ios/native/NativeProxy.mm
```diff
@@ -363,6 +364,21 @@ static CFTimeInterval calculateTimestampWithSlowAnimations(CFTimeInterval curren
     }
     return nil;
   }];
+
+  [animationsManager setComputeSharedTransitionProgressAnimationForTagBlock:^(
+                         int viewTag, double progress, NSDictionary *snapshotValues) {
+    if (auto reaModule = weakModule.lock()) {
+      if (auto runtime = wrt.lock()) {
+        jsi::Runtime &rt = *runtime;
+        jsi::Value snapshotJSValues = convertNSDictionaryToJSIObject(rt, snapshotValues);
```
kmagiera, line 373: I don't understand why snapshot is passed down from JS. Was it handled the same way before with regular shared element transitions?

kmagiera (commented):
I only have a small number of comments to the code on its own.

In general, what I'd like this PR to include is a better description of the concepts behind this change as well as some highlights of what's changed. Doesn't have to be too long, for example you want to mention things like whether the default SET configuration will now follow progress, what are the hooks into navigation (swizzling on iOS and accessing events on Android), etc

ios/native/NativeProxy.mm
piaskowyk: I got rid of this additional conversion 👍

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -123,9 +133,26 @@ std::unordered_map<int, std::shared_ptr<Shareable>>
...
+    jsi::Runtime &rt,
+    const int viewTag,
+    const double progress,
+    const jsi::Value &snapshotValues) {
+  std::shared_ptr<Shareable> config;
+  {
+    auto lock = std::unique_lock<std::mutex>(animationsMutex_);
+    config = getConfigsForType(SHARED_ELEMENT_TRANSITION_PROGRESS)[viewTag];
+  }
+  auto progressAnimation = config->getJSValue(rt).asObject(rt).asFunction(rt);
```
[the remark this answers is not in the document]
piaskowyk: Now I'm using `runtimeHelper->runOnUIGuarded()` method, so it should work properly with LogBox.

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -16,13 +16,21 @@ void LayoutAnimationsManager::configureAnimation(
   if (type == SHARED_ELEMENT_TRANSITION) {
     sharedTransitionGroups_[sharedTransitionTag].push_back(tag);
     viewTagToSharedTag_[tag] = sharedTransitionTag;
+  } else if (type == SHARED_ELEMENT_TRANSITION_PROGRESS) {
+    if (sharedTransitionTag == "secondary") {
```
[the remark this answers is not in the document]
piaskowyk: I changed this logic a bit. The main idea behind this is to provide information about the default type of animation selected by the user. I didn't want to add a new parameter to the `configureAnimation` method. So I decided to pass this information in `sharedTransitionTag` - I know, this is not a perfect solution but I believe I chose less evil 😅
https://github.com/software-mansion/react-native-reanimated/blob/4d97a4d2bb2c90e46b41d79932a968d421eebd2a/Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp#L21-L24

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -123,9 +133,26 @@ std::unordered_map<int, std::shared_ptr<Shareable>>
...
+jsi::Value
+LayoutAnimationsManager::computeSharedTransitionProgressAnimationForTag(
+    jsi::Runtime &rt,
+    const int viewTag,
+    const double progress,
+    const jsi::Value &snapshotValues) {
+  std::shared_ptr<Shareable> config;
+  {
+    auto lock = std::unique_lock<std::mutex>(animationsMutex_);
+    config = getConfigsForType(SHARED_ELEMENT_TRANSITION_PROGRESS)[viewTag];
```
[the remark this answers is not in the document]
piaskowyk: I agree with you 👍 but if you don't mind, let's do it in another PR.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -53,7 +60,34 @@ protected View getTransitioningView(int tag) {
...
   }
 
-  protected void screenDidLayout() {
+  protected void screenDidLayout(int screenViewTag) {
+    ReactContext context = mAnimationsManager.getContext();
+    EventDispatcher eventDispatcher =
+        UIManagerHelper.getEventDispatcherForReactTag(context, screenViewTag);
+    if (eventDispatcher != null && !eventDispatchersWithListener.contains(eventDispatcher)) {
+      eventDispatchersWithListener.add(eventDispatcher);
+      eventDispatcher.addListener(
```
kmagiera, line 69: I wonder if we could approach it differently utilizing the existing mechanism for rerounting events that we use for other type of events (gestures / scroll / etc). If I remember correctly we don't use reflection there to access individual fields. Also, the progress event need to be forwarded to the JS runtime ultimately which is something we already do for that other type of events. If I remember correctly, we use view tag and event name to register for such events, and it appears that we have all that information here.

piaskowyk: I tried to use the existing event mechanism, but I think that the current approach is better.

To register an event, we need the event name and the view tag of the component that emits it. So we need the screen `reactTag`. However, we can only obtain the screen tag on the UI Thread and it require a new native function. The `ReaniamtedNativeModule.registerEventHandler` method needs to be called from JavaScript Thread because  the `ReaniamtedNativeModule` isn't available on UI Thread.

There are two places where we can register for an event. The first is the mounting phase, but it's too early to obtain the view screen, so we can't use it. The other option is on transition start, but that's too late since the component is reparented to the transition container.

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -123,9 +139,27 @@ std::unordered_map<int, std::shared_ptr<Shareable>>
...
+    const double progress) {
+  std::shared_ptr<Shareable> config;
+  {
+    auto lock = std::unique_lock<std::mutex>(animationsMutex_);
+    config =
+        getConfigsForType(SHARED_ELEMENT_TRANSITION_PROGRESS)[sourceViewTag];
+  }
+  jsi::Value progressUpdater = config->getJSValue(*runtimeHelper->uiRuntime());
+  runtimeHelper->runOnUIGuarded(progressUpdater, sourceViewTag, progress);
+  runtimeHelper->runOnUIGuarded(progressUpdater, targetViewTag, progress);
```
kmagiera, line 162: Wonder why we sometimes pass source,target,progress and other times we only pass a single tag. This is perhaps not an error and certain flows are just set up this way but it feels conceptually inconsistent

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
kmagiera: From what I understand it is our code from reanimated that has these requirements. In the current approach we monitor screen layout events for all screen types. I think that we can modify the registration logic such that it allows us to monitor all type of events regardless of the tag, and only make the decision whether we want to start shared transition based on the subviews from a given screen.

From my perspective it'd feel cleaner to hardcode the name of the event rather than hardcoding private field names as we do it now

piaskowyk:
I got rid of reflection on Android and unnecessary method swizzling on iOS. I replaced it with events listeners.

android/src/paper/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -9,6 +9,7 @@
 import com.facebook.react.turbomodule.core.CallInvokerHolderImpl;
 import com.swmansion.reanimated.layoutReanimation.LayoutAnimations;
 import com.swmansion.reanimated.layoutReanimation.NativeMethodsHolder;
+import com.swmansion.reanimated.layoutReanimation.SharedTransitionManager;
```
kmagiera, line 12: this change is no longer needed

app/src/examples/SharedElementTransitions/CustomTransition.tsx
```diff
@@ -6,21 +6,41 @@ import {
...
-const transition = SharedTransition.custom((values) => {
-  'worklet';
-  return {
-    width: withSpring(values.targetWidth),
-    height: withSpring(values.targetHeight),
-    originX: withSpring(values.targetOriginX),
-    originY: withSpring(values.targetOriginY),
-  };
-});
+const transition = new SharedElementTransition()
```
kmagiera, line 16: not a big fan of this API. Btw: are we keeping things compatible, that is, is the old way of defining animation still going to work?

What I don't like about this API is that it uses a pattern that doesn't fit well here or are only partially applicable. Specifically, for builder pattern I wouldn't expect method called "animation" to setup an animation. I think for this purpose we can abandon this pattern completely and instead have something that's much simpler, e.g. a method that takes animation callback and optional progress animation callback, for example:

```
createCustomSharedTransition({ animationCallback: (values) => {}, animationWithProgressCallback: () => {})
```

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -277,6 +288,10 @@ - (NSArray *)sortViewsByTags:(NSArray *)views
       _snapshotRegistry[viewTarget.reactTag] = targetViewSnapshot;
     } else {
       targetViewSnapshot = _snapshotRegistry[viewTarget.reactTag];
+      if (targetViewSnapshot == nil) {
+        targetViewSnapshot = [[REASnapshot alloc] initWithAbsolutePosition:viewTarget];
+        NSLog(@"[Reanimated] Unable to find view style snapshot. It looks like you try to animate an unmounted view.");
```
kmagiera, line 293: is this an error? Seem like we have a recovery path here so the question is if we really need/want to signal this to the developer? If yes, it'd be better to make a warning as this type of log will likely not be noticed.

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -364,6 +385,13 @@ - (void)swizzled_notifyWillDisappear
   [_sharedTransitionManager screenRemovedFromStack:(UIView *)self];
 }
 
+- (void)swizzled_notifyDisappear
+{
+  // call original method from react-native-screens, self == RNSScreen
+  [self swizzled_notifyDisappear];
+  [_sharedTransitionManager screenDidDisappear];
```
kmagiera, line 392: Is this code still needed? I didn't see a corresponding code on Android? Maybe it'd be worth adding event to screens instead of swizzling here similarly to how you added progress cancel event?

ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -392,10 +422,15 @@ - (void)screenRemovedFromStack:(UIView *)screen
...
   } else {
     // removed stack
-    [self clearConfigForStack:stack];
+    [self maybeClearConfigForStack:stack isInteractive:[self isInteractiveScreenChange:screen]];
   }
 }
 
+- (bool)isInteractiveScreenChange:(UIView *)screen
+{
+  return [[[screen.reactViewController valueForKey:@"transitionCoordinator"] valueForKey:@"interactive"] boolValue];
```
kmagiera, line 431: I think we should be able to access this using public API. IIRC there is a way to extract this information via UINavController

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -610,4 +629,18 @@ private void enableCleaningForViewTag(int viewTag) {
       mDisableCleaningForViewTag.put(viewTag, counter - 1);
     }
   }
+
+  void orderByAnimationTypes(List<SharedElement> sharedElements) {
+    for (SharedElement sharedElement : sharedElements) {
```
kmagiera, line 634: in iOS code you clear `sharedelementsWithAnimation` and the other array before this runs. Given you append to these arrays here it makes sense to have them cleared prior to adding new elements. If this happens elsewhere I'd add it here anyways

src/reanimated2/layoutReanimation/sharedTransitions/ProgressTransitionManager.ts
```diff
@@ -0,0 +1,63 @@
+import { runOnUIImmediately } from '../../threads';
+import type { ProgressAnimation } from '../animationBuilder/commonTypes';
+
+function createProgressTransitionManager() {
+  'worklet';
+  const progressAnimations = new Map<number, ProgressAnimation>();
+  const snapshots = new Map<number, any>();
+  const currentTransition = new Set<number>();
```
kmagiera, line 8: `currentTransitions`

src/reanimated2/layoutReanimation/sharedTransitions/ProgressTransitionManager.ts
```diff
@@ -0,0 +1,63 @@
...
+          progressAnimations.delete(viewTag);
+        }
+        toRemove.clear();
+      }
+    },
+  };
+  return progressTransitionManager;
+}
+
+runOnUIImmediately(() => {
```
kmagiera, line 56: not sure how many of these we already have scattered across the codebase but I'd vote for having a single place where we run all the initialization code. If there are more than two places already than lets just ignore this for now

piaskowyk: There are more than two places in the code. I can refactor it in another pull request so that it doesn't make this pull request bigger.

ios/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: I removed it

piaskowyk:
CI fails because I install `rn-screens` from the GitHub repository but it doesn't contain types.

src/reanimated2/layoutReanimation/sharedTransitions/SharedElementTransition.ts
```diff
@@ -0,0 +1,301 @@
...
+      transitionAnimation,
+      sharedTransitionTag
+    );
+    configureLayoutAnimations(
+      viewTag,
+      LayoutAnimationType.SHARED_ELEMENT_TRANSITION_PROGRESS,
+      dummyWorklet,
+      this._defaultTransitionType
+    );
+    runOnUIImmediately(() => {
```
kmagiera, line 119: we does it need to be immediate? Do we also use immediate calls for similar purposes elsewhere?

kmagiera (commented):
Since this is a very complex change, I'd like the description to explain a bit about the design choices behind some of the changes. If there is anything you'd find unclear a year from now please document that in the description, I'd think at leas about the following aspects (we discussed some of these offline but still it'd be valuable to have them documented for the future):
1) why there is method swizzling added on iOS (and why there is no such thing in the Android code)
2) why there are changes related to ordering animations while before we didn't have these
3) how is this change coupled to screens implementation and why?
4) what is the role of ProgressTransitionManager


On top of that, as we discussed offline, we should change the API back to be compatible (no `new` there – also the argument that it's easier to useMemo does not seem to hold as typically custom transitions should be defined outside of the component)

src/reanimated2/layoutReanimation/sharedTransitions/SharedTransition.ts
```diff
@@ -0,0 +1,331 @@
...
+
+export class SharedTransition {
+  private _customAnimationFactory: AnimationFactory | null = null;
+  private _animation: SharedTransitionAnimationsFunction | null = null;
+  private _transitionDuration = 500;
+  private _customProgressAnimation?: ProgressAnimation = undefined;
+  private _progressAnimation?: ProgressAnimation = undefined;
+  private _defaultTransitionType?: SharedTransitionType = undefined;
+  private static _sharedElementCount = 0;
+  private static _eventHandler = {
```
kmagiera, line 56: There seem to be too many loosely related things covered by this class. It both implements builder API and does things like registering for events. It feels like the event logic should be a part of "manager" class above. Maybe worth refactoring or even extracting this event logic or extracting the builder part as they all take a significant part of this file

src/reanimated2/layoutReanimation/sharedTransitions/SharedTransition.ts
```diff
@@ -0,0 +1,331 @@
...
+    configureLayoutAnimations(
+      viewTag,
+      LayoutAnimationType.SHARED_ELEMENT_TRANSITION,
+      transitionAnimation,
+      sharedTransitionTag
+    );
+    configureLayoutAnimations(
+      viewTag,
+      LayoutAnimationType.SHARED_ELEMENT_TRANSITION_PROGRESS,
+      dummyWorklet,
```
kmagiera, line 118: Can you explain this in a comment? Why there is an empty method added as layout animation?

src/createAnimatedComponent.tsx
```diff
@@ -639,13 +641,13 @@ export default function createAnimatedComponent(
           }
           if (sharedTransitionTag) {
             const sharedElementTransition =
-              this.props.sharedTransitionStyle ?? DefaultSharedTransition;
-            configureLayoutAnimations(
+              this.props.sharedTransitionStyle?.getInstance() ??
+              new SharedTransition();
```
kmagiera, line 645: wouldn't it be better not to create a new empty transition every time here? Maybe just have some static field with default empty transition in SharedTransition class?

kmagiera (approved):
Accepting to unblock. I'm still hesitant about using swizzling here. Feels like something that we should be able to avoid given it isn't necessary on Android, but still this isn't really being added in this PR so there is no point in blocking this PR on that issue.

src/createAnimatedComponent.tsx
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4421/commits/5b95cfa3a40669568b028d97d8051adea2b261f0

src/reanimated2/layoutReanimation/sharedTransitions/SharedTransition.ts
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4421/commits/5ab12a598e5f2b64cfa2441e68cdbc4bffb7293f

src/reanimated2/layoutReanimation/sharedTransitions/SharedTransition.ts
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4421/commits/5ff4824408025f3e6ee6d862ee9ebbc18de98210

piaskowyk:
Hey @terrysahaidak 👋 Unfortunately this PR isn't related to https://github.com/software-mansion/react-native-reanimated/issues/4516 but on Monday I'll ask someone from the team to look at this issue

nicolascavallin:
Is there a roadmap about when it will be released? Thanks <3

PR #4433, opened 2023-05-08
Remove plugin dev files from npm package

tjzel:
## Summary

Currently development files from `plugin/` are included in npm package. This PR removes them from it.

b4:
Screenshot 2023-05-08 at 14 39 29
after:
Screenshot 2023-05-08 at 14 39 51

Since [README is always included](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#files) I renamed plugin's dev README and removed `README` from being included in `package.json`.

## Test plan

I recommend using this powerful oneliner: 
`./createNPMPackage.sh && mkdir tarball-new && mv react-native-reanimated-3.1.0.tgz tarball-new && tarball-new && tar -xf react-native-reanimated-3.1.0.tgz && ..`
to see the contents of the new package.

Also run _some_ Example App to see if Reanimated plugin from the tarball is actually working.

---
_Note_: Testing this PR took me longer than it should.

For some reason with current configuration of Example App and running it on Android (I didn't check iOS) it's surprisingly difficult to use reanimated from either tarball or unpacked tarball directory. I had to make a new app and then include Example's source code there. 

While it's not that troublesome I think we should have a more streamlined approach for using custom reanimated package location for tests with our Examples.


package.json
```diff
@@ -47,9 +47,9 @@
     "ios/",
     "RNReanimated.podspec",
     "scripts/reanimated_utils.rb",
-    "README.md",
     "mock.js",
-    "plugin/",
+    "plugin/index.js",
+    "plugin/build",
```
tomekzaw, line 52:
```suggestion
    "plugin/build/",
```

tomekzaw (approved):

PR #4438, opened 2023-05-09
Publish nightly builds on npm

tomekzaw:
## Summary

This PR adds a cron job to build and publish nightly version of react-native-reanimated every day at 23:27.

The granular npm access token is safely stored in `NODE_AUTH_TOKEN` secret.

Lessons learned:
1. In order to use secrets in called workflows, you need to implicitly pass them in workflow file.
1. If for some reason the environment variable `NODE_AUTH_TOKEN` is not set, `actions/setup-node@v3` will overwrite it with `XXXXX-XXXXX-XXXXX-XXXXX` (see [here](https://github.com/actions/setup-node/blob/a9893b0cfb0821c9c7b5fec28a6a2e6cdd5e20a4/src/authutil.ts#L55C1-L58))

### DIFF
--- a/.github/workflows/build-nightly-npm-package.yml
+++ b/.github/workflows/build-nightly-npm-package.yml
@@ -11,3 +11,5 @@ jobs:
     with:
       version: nightly
       publish_on_npm: true
+    secrets:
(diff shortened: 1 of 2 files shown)


piaskowyk (approved):
👏👏👏

PR #4384, opened 2023-04-20
Fix scrollTo on FlashList

tomekzaw:
## Summary

Fixes #4376. Analogous to #3988.

Solution: copy the code from `createAnimatedComponent.tsx` to `useAnimatedRef.ts`

https://github.com/software-mansion/react-native-reanimated/blob/87fb9c6a1fb6b8c839b65b90ad16f574f45a1294/src/createAnimatedComponent.tsx#L289-L291

Example:

<details>
<summary>App.tsx</summary>

```tsx
import Animated, {
  runOnUI,
  scrollTo,
  useAnimatedRef,
} from 'react-native-reanimated';
import {Button, StyleSheet, Switch, Text, View} from 'react-native';
import {FlashList} from '@shopify/flash-list';

import React from 'react';

declare const _WORKLET: boolean;

const AnimatedFlashList = Animated.createAnimatedComponent(FlashList);

interface Item {
  label: string;
}

const DATA: Item[] = [...Array(100)].map((_, i) => ({label: String(i)}));

function getRandomOffset() {
  'worklet';
  return Math.random() * 2000;
}

export default function ScrollToExample() {
  const [animated, setAnimated] = React.useState(true);

  const aref = useAnimatedRef<FlashList<Item>>();

  const scrollFromJS = () => {
    console.log(_WORKLET);
    aref.current?.scrollToOffset({offset: getRandomOffset(), animated});
  };

  const scrollFromUI = () => {
    runOnUI(() => {
      'worklet';
      console.log(_WORKLET);
      scrollTo(aref, 0, getRandomOffset(), animated);
    })();
  };

  return (
    <>
      <View style={styles.buttons}>
        <Switch
          value={animated}
          onValueChange={setAnimated}
          style={styles.switch}
        />
        <Button onPress={scrollFromJS} title="Scroll from JS" />
        <Button onPress={scrollFromUI} title="Scroll from UI" />
      </View>
      <AnimatedFlashList
        // @ts-ignore createAnimatedComponent seems to break generic types
        ref={aref}
        data={DATA}
        // @ts-ignore createAnimatedComponent seems to break generic types
        renderItem={({item}: {item: Item}) => (
          <Text key={item.label} style={styles.text}>
            {item.label}
          </Text>
        )}
        estimatedItemSize={200}
      />
    </>
  );
}

const styles = StyleSheet.create({
  switch: {
    marginBottom: 10,
  },
  buttons: {
    marginTop: 80,
    marginBottom: 40,
    alignItems: 'center',
[...]

PR #4445, opened 2023-05-10
Fix animatedRef on Fabric

### LINKED ISSUE #4256: 3.0.2 Android Cannot read property 'stateNode' of undefined
### Description

if an animated component has a ref provided by useAnimatedRef hook, it will produce this undefined error, even though fabric is enabled.

### Steps to reproduce

1. create react-native app by react-native-cli with latest stable version
2. enable new arch 
3. create an animate component in app.tsx and then give it a ref provide by useAnimatedRef hook

https://github.com/fine0813/reanimated-crash

### Reanimated version

3.0.2

### React Native version

0.71.1

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

_No response_

### Acknowledgements

Yes

mstach60161:
## Summary

- Fixes #4256

### Problem

If the scrollable component is not root, UI `scrollTo` doesn't work on Fabric.

#4384 - It was fixed here for Paper, but after this change app crashes on Fabric, because `getScrollableNode` in both architectures returns viewTag as a number, not component. It's working for Paper, because we are passing either component or number to `findNodeHandle` method, but on Fabric we need component ref.

### Fix
 
Instead of `getScrollableNode`, we can use `getNativeScrollRef` and get scrollable reference. 

### TODO

Unfortunately this method is not implemented yet for `Flash-list` -> https://shopify.github.io/flash-list/docs/usage#:~:text=Unsupported%20methods%3A
so for now, UI `scrollTo` on FlashList works only for Paper Architecture.

## Test plan

ScrollToExample:

- Paper, ScrollView
- Paper, FlatList
- Fabric, ScrollView
- Fabric, FlatList


src/reanimated2/hook/useAnimatedRef.ts
```diff
@@ -19,6 +20,23 @@ function getShareableShadowNodeFromComponent(
   return getShadowNodeWrapperFromHostInstance(component);
 }
 
+function getScrollableRef(component: ComponentRef): ComponentRef {
+  if (global._IS_FABRIC) {
+    if (component.getNativeScrollRef) {
+      return component.getNativeScrollRef();
+    } else {
+      console.warn(
+        'this component has no implemented `getScrollableNode` method. Please report this issue to the library maintainers.'
```
tomekzaw, line 29:
```suggestion
        '[Reanimated] ${Component.displayName || Component.name || 'Component'} has no implemented `getNativeScrollRef` method. Please report this issue to the library maintainers.'
```

src/reanimated2/hook/useAnimatedRef.ts
```diff
@@ -31,11 +49,7 @@ export function useAnimatedRef<T extends ComponentRef>(): RefObjectFunction<T> {
     const fun: RefObjectFunction<T> = <RefObjectFunction<T>>((component) => {
       // enters when ref is set by attaching to a component
       if (component) {
-        tag.value = getTagValueFunction(
-          component.getScrollableNode
-            ? component.getScrollableNode()
-            : component
-        );
+        tag.value = getTagValueFunction(getScrollableRef(component));
```
tomekzaw, line 52: Can we also use `getScrollableRef` in `useAnimatedRef.ts`?

mstach60161: I'm not sure what do you mean.

src/reanimated2/hook/useAnimatedRef.ts
```diff
@@ -19,6 +22,25 @@ function getShareableShadowNodeFromComponent(
...
+  if (global._IS_FABRIC) {
+    if (component.getNativeScrollRef) {
+      return component.getNativeScrollRef();
+    } else {
+      console.warn(
+        `[Reanimated] ${
+          component.displayName || component.name || 'Component'
+        } has no implemented 'getNativeScrollRef' method. Please report this issue to the library maintainers.`
+      );
+    }
```
piaskowyk, line 35: We need to remove this config because we can pass here reference to any object.

src/reanimated2/hook/useAnimatedRef.ts
```diff
@@ -19,6 +22,25 @@ function getShareableShadowNodeFromComponent(
   return getShadowNodeWrapperFromHostInstance(component);
 }
 
+function getScrollableRef(component: ComponentRef): ComponentRef {
```
piaskowyk, line 25: getScrollableRef -> getComponentRef

piaskowyk (commented):
We need add some comments here, and table with information about which method is used by ScrollView, FlatList and FlashList

src/reanimated2/hook/useAnimatedRef.ts
tomekzaw: `getComponentRef` is too generic I think

piaskowyk (approved):

src/reanimated2/hook/useAnimatedRef.ts
```diff
@@ -10,6 +10,9 @@ import {
 } from '../shareables';
 
 interface ComponentRef extends Component {
+  displayName?: string;
+  name?: string;
```
tomekzaw, line 14:
```suggestion
```

tomekzaw (approved):
👏

PR #4449, opened 2023-05-12
Move worklet function declarations to variable declarations

tjzel:
## Summary

Reanimated Babel plugin may break function hoisting for functions with 'worklet' directive. Consider the following code snippet:

```TypeScript
export function foo() {
  'worklet';
  return 1;
}
```

It will get transpiled to:
```JavaScript
Object.defineProperty(exports, '__esModule', {
  value: true,
});
exports.foo = void 0; // this line means that, at this time, our function is exported as undefined
var _worklet_6320373531492_init_data = {
  code: 'function foo() {\n  return 1;\n}',
  location: 'location',
  sourceMap: 'sourcemap',
};
var foo = (function () {
  var _e = [new global.Error(), 1, -27];
  var _f = function _f() {
    return 1;
  };
  _f._closure = {};
  _f.__initData = _worklet_6320373531492_init_data;
  _f.__workletHash = 6320373531492;
  _f.__stackDetails = _e;
  _f.__version = '3.1.0';
  return _f;
})();
exports.foo = foo; // this line actually defines our export, lack of this will cause issues when hoisting!
```
Last line might be missing because happening because in plugin changes function declarations to variable declarations, but the code has been migrated to commonjs before.

Keep in mind, that not only exports are the issue, depending on hoisting within the same file can be broken as well.

To solve that this PR changes all explicit worklet function declarations to the variables in the first place - in this example. 
```Diff
- export function foo() {
+ export const foo = () => {
  'worklet';
  return 1;
}
```
Let's see the result of this change while building our `lib/`, looking at `lib/commonjs/reanimated2/interpolation.js`.

<table>
<tr><td>BEFORE</td><td>

Screenshot 2023-05-12 at 16 00 52

</td></tr>
<tr><td>AFTER</td><td>

Screenshot 2023-05-12 at 16 01 28

</td></tr>
</table>

What solves our problem.

To prevent this issue from resurfacing in the future we should add some guidelines to our README to avoid these issues with hoisting.


tjzel:
Maybe it will wait for better times...

PR #4452, opened 2023-05-15
Replace console.error with throw new Error

tomekzaw:
## Summary

This PR changes the behavior of the app when version mismatch is detected. Previously, we called `console.error` in such case, but the app still continued to execute. This PR replaces `console.error` with `throw new Error` which stops the execution and shows full-screen LogBox.

| Before | After |
|:-:|:-:|
| before | after |


tjzel (approved):
But why are the versions on provided screenshots the same? 🤔

tomekzaw:
I've just changed `!ok` to `ok` for testing purposes in order to see how the error will look like.

PR #4453, opened 2023-05-15
Restore header back button after reload on web

tomekzaw:
## Summary

Currently, the header back button disappears on page reload. This PR adds a back button that navigates to the home screen on each example screen.

## Acknowledgements

Thanks @kacperkapusciak for showing me `headerLeft` prop and `@react-navigation/elements` package.

## Test plan

1. Open WebExample
2. Open Bubbles example 
3. Reload the page
4. Verify that header back button is visible


tjzel (approved):
Killer feature with dark mode 🕶️

PR #4458, opened 2023-05-15
Disable text measurement cache to fix animating text

mstach60161:
## Summary

- Disable caching text to fix text animations of Fabric.

## Test plan

- Before

https://github.com/software-mansion/react-native-reanimated/assets/56194058/47af48fe-b766-463e-b672-47f490c60e77

- After

https://github.com/software-mansion/react-native-reanimated/assets/56194058/a710d751-f1c2-46c1-b8d4-963996d20348

### DIFF
--- a/Common/cpp/Fabric/ReanimatedUIManagerBinding.cpp
+++ b/Common/cpp/Fabric/ReanimatedUIManagerBinding.cpp
@@ -5,6 +5,7 @@
 #include "NewestShadowNodesRegistry.h"
 
 #include <react/renderer/debug/SystraceSection.h>
+#include <react/renderer/core/CoreFeatures.h>
 
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


Common/cpp/Fabric/ReanimatedUIManagerBinding.cpp
```diff
@@ -37,6 +38,10 @@ void ReanimatedUIManagerBinding::createAndInstallIfNeeded(
     return std::move(uiManagerBindingPublic->eventHandler_);
   }();
 
+  // We can not cache text measurements, because we need to calculate it on each animation frame.
+  CoreFeatures::cacheNSTextStorage = false;
+  CoreFeatures::cacheLastTextMeasurement = false
```
tomekzaw, line 43: Disabling these flags turns off caching of all text components across the whole application, even though only a small fraction of those would be animated. I'm afraid we will need to think about another solution.

piaskowyk: I agree with Tomek, we can't disable the whole feature for all existing text components in an application.

mstach60161:
This issue was fixed in React Native code: https://github.com/facebook/react-native/blob/6d24ee13a47a58ff1b19c5a51f39f52c9a4a8a28/packages/react-native/ReactCommon/react/renderer/components/text/ParagraphLayoutManager.cpp#L4

PR #4459, opened 2023-05-16
Auto-workletize functions passed to runOnUI

tomekzaw:
## Summary

This PR adds `runOnUI` to the list of automatically workletized functions as suggested by @kacperkapusciak.

Previously, the following code:

```tsx
runOnUI(() => {
  console.log('Hello from the UI thread!');
})();
```

would throw the following error:

```
Error: runOnUI() can only be used on worklets, js engine: hermes
```

because the function passed to `runOnUI` needs to be marked as worklet:

```diff
 runOnUI(() => {
+  'worklet';
   console.log('Hello from the UI thread!');
 })();
```

After this change, the `'worklet';` directive can be omitted and the function will be automatically workletized by the Reanimated Babel plugin.

## Test plan

<details>
<summary>App.tsx</summary>

```tsx

import React from 'react';

  const handlePress = () => {
    runOnUI(() => {
      console.log('Hello from the UI thread!');
    })();
  };

  return (
      <Button title="Click me" onPress={handlePress} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

</details>


kacperkapusciak (approved):

tjzel:
While I agree that this change is a natural extension of what we have already done, I wonder if `runOnUI` has many common use cases. I always thought we should keep the 'worklet' directive here as necessary due to the fact that using this function has its limitations and should be used only when one has a deeper understanding of Reanimated. For example:

```TypeScript
runOnUI(() => {
  'worklet';
  console.log('Hello from the UI thread!');
  runOnUI(() => {
    'worklet';
    console.log("Hello again with 'worklet'!");
  })();
})();
```

and

```TypeScript
runOnUI(() => {
  'worklet';
  console.log('Hello from the UI thread!');
  runOnUI(() => {
    // Missing 'worklet' directive
    console.log("Hello again without 'worklet'!");
  })();
})();
```

Both will crash. If you consider this limitation unrelated to whether we should be forced to type 'worklet' or not, I'm open to approving this.

tjzel (changes requested):
How about adding one or two unit tests for this?

Issue #4347, opened 2023-04-10
Could you help me, when i use useSharedValue in debug, it show error?


memberwebm:
### Description

could you help me, when i open debug dev, it show bug 
<img width="2047" alt="Screenshot 2023-04-10 at 13 42 56" src="https://user-images.githubusercontent.com/110653846/230843186-6d6d23f8-9446-41c2-85bc-d61d9761a5ee.png">
+but when i delete useSharedValue it working , i want to use useSharedValue and debug, is there any way?

### Steps to reproduce

1.use  useSharedValue
2. open debug ios
3. bug will show

### Snack or a link to a repository

 https://stackoverflow.com/help/mcve

### Reanimated version

3.0.2

### React Native version

0.69.0

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

Latropos:
Duplicate of #3951, please try hints suggested in the conversation

Latropos:
Duplicate of https://github.com/software-mansion/react-native-reanimated/issues/3951

Latropos:
Duplicates #4198

piaskowyk:
Closed due to duplication. We want to keep discussion in the one place.

tjzel:
@memberwebm Could you check if your issue is fixed with new version of Reanimated? (taken from either GitHub or nightly build (e.g. yarn add react-native-reanimated@nightly))

Issue #4197, opened 2023-03-10
Issue 3.0 for jest


billnbell:
### Description

Issue docs wrong.

```
require('react-native-reanimated/lib/reanimated2/jestUtils').setUpTests();
```
Does not exist.

```
require('react-native-reanimated/lib/commonjs/reanimated2/jestUtils').setUpTests();
```

This is there - but I get:

```
 RUNS  src/__tests__/Helpers-test.ts
/Users/bill/theclub/TheClub/node_modules/react-native-reanimated/lib/commonjs/reanimated2/initializers.js:172
        debug: (0, _threads.runOnJS)(capturableConsole.debug),
                                    ^

TypeError: (0 , _threads.runOnJS) is not a function
    at _f (/Users/bill/theclub/TheClub/node_modules/react-native-reanimated/lib/commonjs/reanimated2/initializers.ts:160:25)
    at _f (/Users/bill/theclub/TheClub/node_modules/react-native-reanimated/lib/commonjs/reanimated2/threads.ts:133:5)
    at Timeout._onTimeout (/Users/bill/theclub/TheClub/node_modules/react-native-reanimated/lib/commonjs/reanimated2/initializers.js:155:16)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
error Command failed with exit code 1.
```

### Steps to reproduce

Just install 3,.0

### Snack or a link to a repository

nope

### Reanimated version

3.0

### React Native version

0.71.4

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

yerevin:
You can try with fix described in https://github.com/software-mansion/react-native-reanimated/pull/4136, but I'm not sure if it recommended way by creators

ajaxjiang96:
somehow this works:

```ts
// eslint-disable-next-line @typescript-eslint/no-unused-vars
import * as reanimatedJestUtils from 'react-native-reanimated/src/reanimated2/jestUtils';
```

croaker:
[...]

Issue #4185, opened 2023-03-07
Jest mocks broken with babel plugin


kylebake:
### Description

It appears that the babel plugin line is breaking some necessary mocks for jest tests in newer versions of the library. With the plugin enabled in the `babel.config.js` file during tests, various methods from `react-native-reanimated` are not found and give errors such as: `TypeError: (0 , _reactNativeReanimated.interpolate) is not a function` (note: interpolate is only one example, there are many others).

However, removing the babel plugin from the testing environment fixes this TypeError. I've linked a barebones repository to reproduce. 

### Steps to reproduce

On [linked repository](https://github.com/kylebake/ReanimatedTestBug):
1. `yarn` to install dependencies
2. `yarn test` to run test, note the failure
3. Comment out/remove the `plugins` line from `babel.config.js`
4. `yarn test` again, this time the test passes successfully

### Snack or a link to a repository

https://github.com/kylebake/ReanimatedTestBug

### Reanimated version

3.0.1

### React Native version

0.71.3

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

None

### Acknowledgements

Yes

thespacemanatee:
+1

tjzel:
@kylebake @thespacemanatee Could you check if your issue is fixed with new version of Reanimated? (taken from either GitHub or nightly build (e.g. yarn add react-native-reanimated@nightly))

danieloi:
I'm having the same issue. Hooks like useAnimatedStyle are undefined

`TypeError: (0 , _reactNativeReanimated.useAnimatedStyle) is not a function`

tjzel:
@danieloi What version of reanimated are you using? It's currently fixed in 3.3.0.

danieloi:
@tjzel 2.17.0

danieloi:
Looks like changing `node_modules/react-native-reanimated/src/reanimated2/mock.ts` from .ts to .js and removing type annotations fixed it for me 

Still investigating though.

tjzel:
This actually might be caused by jest's cache. Try to use `.ts` then run `yarn jest --clear-cache && yarn test`.

danieloi:
That didn't work

Here's the patch for v2.17.0 that worked for me

```
[...]

Issue #4175, opened 2023-03-06
undefined is not a function in _filterNonAnimatedProps


tapz:
### Description

After upgrading to react-native-reanimated v3 I started getting this:

```
TypeError: undefined is not a function
TypeError: undefined is not a function
    at _loop2 (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:186205:91)
    at _filterNonAnimatedProps (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:186214:19)
    at render (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:186221:51)
    at finishClassComponent (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:84495:43)
    at updateClassComponent (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:84463:50)
    at beginWork (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:85585:42)
    at beginWork$1 (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:88920:29)
    at performUnitOfWork (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:88433:29)
    at workLoopSync (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:88381:28)
    at renderRootSync (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.myapp.app&modulesOnly=false&runModule=true:88363:25)
    at performSyncWorkOnRoot (http:/
```

![screencap](https://user-images.githubusercontent.com/11726387/223165606-3ddf96de-17e7-47f2-b42b-4716cebed3bb.png)

Works just fine with 2.14.4.

### Steps to reproduce

1. Build Android (debug)

### Snack or a link to a repository

```javascript
import Animated, {
} from 'react-native-reanimated';

  return (
    <View style={styles.container}>
[...]

PR #4468, opened 2023-05-17
Remove commonjs generation

tjzel:
## Prelude

We have waited for it a long time. 

Here it is.

A perfect summary of weeks time of our work and research - two line fix.

## Summary

Ever since we've adopted CommonJS as target for our built JavaScript files we made it very difficult to our users to properly use Jest with Reanimated. This is due to fact that Reanimated Babel plugin is prone to break hoisting and this becomes critical in CommonJS. This issue had never surfaced before because we were building only for ES modules which can handle the way our plugin works. Therefore, after many talks we've decided to drop support for CommonJS and revert to ES modules only.

Refer to #4449 for _slightly_ more information.

This is likely to fix (I want to wait for confirmation):
- #4347,
- #4197,
- #4185,
- #4175,
- #4136,
- #4001,
- #3971.

## Test plan

- I tested it on the [original repro](https://github.com/kylebake/ReanimatedTestBug) that began this voyage. I recommend using [my fork](https://github.com/tjzel/Reanimated-jest-commonjs-issue) since that repro had some errors coming from used versions of React-Native and Jest. 
__Make sure you use Node 18.15 for it.__
- I also tested it on my university [app](https://github.com/MeowdoptMe/MeowdoptMeApp) that actively uses Jest and Reanimated and using tarball from this PR fixed Jest issues.

### DIFF
--- a/docs/docs/guide/testing.md
+++ b/docs/docs/guide/testing.md
@@ -13,7 +13,7 @@ First, make sure that your tests run with Node version 16 or newer.
 Add the following line to your `jest-setup.js` file:
 
 ```js
-require('react-native-reanimated/src/reanimated2/jestUtils').setUpTests();
+require('react-native-reanimated').setUpTests();
 ```
 
 `setUpTests()` can take optional config argument. Default config is `{ fps: 60 }`, setting framerate to 60fps.
--- a/docs/docs/guide/testing.md
+++ b/docs/docs/guide/testing.md
@@ -13,7 +13,7 @@ First, make sure that your tests run with Node version 16 or newer.
 Add the following line to your `jest-setup.js` file:
 
 ```js
-require('react-native-reanimated/lib/reanimated2/jestUtils').setUpTests();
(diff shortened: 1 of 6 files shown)


kmagiera (approved):
Left one comment, looks good otherwise

piaskowyk (approved):

kliakhovskii-brex:
Wonder if there is going to be a release soon with this fix included?

tomekzaw:
@kliakhovskii-brex you can try nightly build with `yarn add react-native-reanimated@nightly`

kliakhovskii-brex:
> @kliakhovskii-brex you can try nightly build with `yarn add react-native-reanimated@nightly`

awesome, thanks for the hint!

cflores-brex:
@tomekzaw , we've tested locally with the nightly build and can confirm this fix resolves our issues we were seeing with Jest when upgrading to v3. Wondering when the next release will be as we'd like to migrate to v3 asap. Thanks!

PR #4469, opened 2023-05-18
Rename `ref` & `aref` -> `animatedRef` in types

kacperkapusciak:
For better DX and consistency with the docs (current and the new ones) I've decided to rename the first argument of `useScrollViewOffset`, `measure`, `getRelativeCoords`, and `scrollTo` to `animatedRef`.

### Motivation

Just `ref` as an argument  indicates that a plain `const ref = React.useRef()` works with these functions leading to a confusing developer experience.

Also, `aref` is a bit cryptic and a hard thing to look up on the Internet when just starting to learn Reanimated.


piaskowyk (approved):
👏

PR #4476, opened 2023-05-18
Optimize `useAnimatedReaction` even more

tomekzaw:
## Summary

This PR aims to optimize `useAnimatedReaction` even more because why not.

In the essence, we only pass `previous` to worklet closure and when we do so, we freeze the object, so it's always `{ value: true }`, thus there's no need to store it using `useRef`.


tomekzaw:
Closing this PR because #4633 was merged.

PR #4484, opened 2023-05-22
Add typescript types for transform styles in createAnimatedComponent

Latropos:
## Summary

Importing correct types directly from react-native have many advantages. For example passing `style={{transfor:undefined}}` is fully supported by default view, so Animated. View should support it too. Once I've imported react-native styles I've found a line in code that would throw an error if undefined was passed.

|This is correct|This is underlined as error|
|---|---|
|image|image|

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -1,5 +1,10 @@
 import React, { Component, ComponentType, MutableRefObject, Ref } from 'react';
-import { findNodeHandle, Platform, StyleSheet } from 'react-native';
+import {
+  findNodeHandle,
+  Platform,
+  StyleSheet,
+  TransformsStyle,
+} from 'react-native';
 import WorkletEventHandler from './reanimated2/WorkletEventHandler';
 import setAndForwardRef from './setAndForwardRef';
 import './reanimated2/layoutReanimation/animationsManager';
@@ -115,11 +120,16 @@ const has = <K extends string>(
   return false;
 };
 
-function isInlineStyleTransform(transform: any): boolean {
-  if (!transform) {
+function isInlineStyleTransform(
+  transform: TransformsStyle['transform']
+): boolean {
+  if (transform === undefined) {
     return false;
   }
-  return transform.some((t: Record<string, any>) => hasInlineStyles(t));
+
+  return transform.some(
+    (t: NonNullable<TransformsStyle['transform']>[number]) => hasInlineStyles(t)
+  );
 }
 
 function hasInlineStyles(style: StyleProps): boolean {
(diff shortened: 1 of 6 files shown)


src/createAnimatedComponent.tsx
```diff
@@ -115,11 +120,16 @@ const has = <K extends string>(
...
+  transform: TransformsStyle['transform']
+): boolean {
+  if (transform === undefined) {
     return false;
   }
-  return transform.some((t: Record<string, any>) => hasInlineStyles(t));
+
+  return transform.some(
+    (t: NonNullable<TransformsStyle['transform']>[number]) => hasInlineStyles(t)
+  );
```
tomekzaw, line 132:
```suggestion
  return transform.some(hasInlineStyles);
```

src/createAnimatedComponent.tsx
```diff
@@ -179,14 +189,22 @@ function inlinePropsHasChanged(styles1: StyleProps, styles2: StyleProps) {
...
   for (const [key, styleValue] of Object.entries(inlineProps)) {
     if (key === 'transform') {
-      update[key] = styleValue.map((transform: Record<string, any>) => {
-        return getInlinePropsUpdate(transform);
-      });
+      update[key] = styleValue.map(
+        (transform: TransformsStyle['transform']) => {
+          return getInlinePropsUpdate(transform);
+        }
+      );
```
tomekzaw, line 207:
```suggestion
      update[key] = styleValue.map(getInlinePropsUpdate);
```

Latropos: Thanks!

src/createAnimatedComponent.tsx
```diff
@@ -115,11 +120,18 @@ const has = <K extends string>(
   return false;
 };
 
-function isInlineStyleTransform(transform: any): boolean {
-  if (!transform) {
+function isInlineStyleTransform(
+  transform: unknown
+): transform is TransformsStyle['transform'] {
```
tomekzaw, line 125: What's the point of having `transform is TransformsStyle['transform']` here? The function may contain valid a transform but without inline styles and in such case the argument won't be treated as `TransformsStyle['transform']` because the function returns `false`, am I correct?

src/createAnimatedComponent.tsx
```diff
@@ -179,14 +191,18 @@ function inlinePropsHasChanged(styles1: StyleProps, styles2: StyleProps) {
   return false;
 }
 
-function getInlinePropsUpdate(inlineProps: Record<string, any>) {
+function getInlinePropsUpdate(
+  inlineProps: TransformsStyle['transform'] | Record<string, any>
```
tomekzaw, line 195: I suspect that `TransformsStyle['transform']` satisfies `Record<string, any>`, what's the point of having an union type here?

src/createAnimatedComponent.tsx
```diff
@@ -179,14 +191,18 @@ function inlinePropsHasChanged(styles1: StyleProps, styles2: StyleProps) {
...
+) {
   'worklet';
   const update: Record<string, any> = {};
+  if (!inlineProps) {
+    return {};
```
tomekzaw, line 200: Why do we return `{}` here?

src/createAnimatedComponent.tsx
Latropos: Almost - it will return true for both inline transform and transform returned from useAnimatedStyle. However I will be working on a new feature for transformations so it should be safe to close this PR and move all this fix to the new feature

src/createAnimatedComponent.tsx
Latropos: TransformStyle is an array, so it doesn't

src/createAnimatedComponent.tsx
Latropos: We have inline styles and styles returned from a hook "useAnimatedStyles". The returned styles are wrapped inside `AnimatedStyleResult`, so you need to extract them from this object first to make `isInlineStyleTransform` return true.

PR #4486, opened 2023-05-23
Update imports for react-native-reanimated.d.ts

tjzel:
## Summary

In order to get rid of manually written .d.ts file I'm going to try and minimize it first. While trying to do that I found that imports for it are long outdated and resolved to `any`. This PR fixes them and also fixes a small issue stemming from this update in Example app.

## Test plan

😵 
```yarn tsc --noEmit --target es6 --module ESNext --jsx react-native --skipLibCheck true --allowSyntheticDefaultImports true --moduleResolution node --esModuleInterop true --strict true --forceConsistentCasingInFileNames true --resolveJsonModule true app/src/App.tsx```

### DIFF
--- a/app/src/examples/OldMeasureExample.tsx
+++ b/app/src/examples/OldMeasureExample.tsx
@@ -167,7 +167,9 @@ function SectionHeader({
   contentHeight,
   show,
 }: SectionHeaderProps) {
-  const applyMeasure = ({ height }: ReturnType<typeof measure>) => {
+  const applyMeasure = ({
+    height,
+  }: NonNullable<ReturnType<typeof measure>>) => {
     'worklet';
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: app/src/examples/OldMeasureExample.tsx [hunk 1]
```diff
@@ -195,7 +197,10 @@ function SectionHeader({
   } else {
     onActiveImpl = () => {
       'worklet';
-      applyMeasure(measure(animatedRef));
+      const measuredMeasure = measure(animatedRef);
```
tjzel, line 200: I did this on purpose 😼

tjzel:
It occurred that we wanted to do wasn't really possible in TypeScript so it must be scrapped.

PR #4487, opened 2023-05-23
Set `viewTag` only if needed

Latropos:
## Summary

This slightly improves performance.

## Test plan

Tested that app still works


src/createAnimatedComponent.tsx
```diff
@@ -293,14 +293,17 @@ export default function createAnimatedComponent(
 
     _attachNativeEvents() {
       const node = this._getEventViewRef();
-      const viewTag = findNodeHandle(options?.setNativeProps ? this : node);
+      let viewTag = null; // We set it only if needed
```
tomekzaw, line 296: Can we also do the same in `_reattachNativeEvents` which happens on each render?

piaskowyk (approved):

PR #4498, opened 2023-05-29
Typescript fixes

Latropos:
## Summary

Some tsc fixes I've made when working on anchor implementation.


src/reanimated2/ViewDescriptorsSet.ts
```diff
@@ -34,7 +34,7 @@ export function makeViewDescriptorsSet(): ViewDescriptorsSet {
       });
     },
 
-    remove: (viewTag: number) => {
+    remove: (viewTag: number | null) => {
```
tjzel, line 37: We know it's nullable but we don't anything about it. Either we should safeguard against null or actually check the code (probably extremely difficult) if it can pass null here.

tjzel (changes requested):

Latropos:
Improvements to definition of transform style were already merged

PR #4499, opened 2023-05-29
Fix sensors on android web

mstach60161:
### Problem

Android devices use a right-handed coordinate system, where the positive x-axis points to the right, the positive y-axis points up, and the positive z-axis points out of the screen.
However iOS uses a left-handed coordinate system, where the positive x-axis points to the right, the positive y-axis points down, and the positive z-axis points into the screen.

This is already fixed on android native code, but the problem still exists on android web browser (react native web).

### Fix

- Replace quaternion z, y and negate z  
- Do the same thing on Android native

## Test plan

- Web Example App


src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -158,4 +164,16 @@ export default class JSReanimated extends NativeReanimated {
         return 'AbsoluteOrientationSensor';
     }
   }
+
+  isPlatformAndroid(): boolean {
+    const userAgent = navigator.userAgent || navigator.vendor || window.opera;
+    return userAgent !== undefined && /android/i.test(userAgent);
```
tomekzaw, line 170: Can we memoize the result of this check?

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -10,6 +10,7 @@ import { WebSensor } from './WebSensor';
 export default class JSReanimated extends NativeReanimated {
   nextSensorId = 0;
   sensors = new Map<number, WebSensor>();
+  platform = Platform.UNKNOWN;
```
tomekzaw, line 13: Can we use `undefined` (meaning uninitialized) instead of `Platform.UNKNOWN`?
```suggestion
  platform = undefined;
```

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -66,11 +67,20 @@ export default class JSReanimated extends NativeReanimated {
       return -1;
     }
 
+    if (this.platform === Platform.UNKNOWN) {
+      this.detectPlatform();
+    }
```
tomekzaw, line 72: I'm thinking about using a getter here but this approach is probably faster.

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -158,4 +170,32 @@ export default class JSReanimated extends NativeReanimated {
         return 'AbsoluteOrientationSensor';
     }
   }
+
+  detectPlatform() {
+    const userAgent = navigator.userAgent || navigator.vendor || window.opera;
+    if (userAgent === undefined) {
+      return;
```
tomekzaw, line 177: Shouldn't we assign `Platform.UNKNOWN` here so that we won't call `detectPlatform` again?

tomekzaw (approved):
Tested on iOS and Android, looks like finally it's consistent. Thanks @mstach60161 for taking care of this!

Edit: I've tested `pitch`, `roll` and `yaw` between main and this branch and they are the same but I haven't checked `qx`, `qy`, `qz` and `qw` individually, is this the expected result?

Edit 2: Ready to merge.

PR #4504, opened 2023-05-30
fix: ios build error when using 'use_frameworks!' in Podfile

yupengyang:
when open 'use_frameworks!' in Podfile, build will fail because of unable to find hermes related headers. We need to declare hermes dependency explicit in podspec

### DIFF
--- a/RNReanimated.podspec
+++ b/RNReanimated.podspec
@@ -15,7 +15,7 @@ boost_compiler_flags = '-Wno-documentation'
 fabric_flags = fabric_enabled ? '-DRCT_NEW_ARCH_ENABLED' : ''
 example_flag = config[:is_reanimated_example_app] ? '-DIS_REANIMATED_EXAMPLE_APP' : ''
 version_flag = '-DREANIMATED_VERSION=' + reanimated_package_json["version"]
-using_hermes = ENV['USE_HERMES'] == nil || ENV['USE_HERMES'] == "1"
+using_hermes = ENV['USE_HERMES'] == nil || ENV['USE_HERMES'] == '1'
 
 Pod::Spec.new do |s|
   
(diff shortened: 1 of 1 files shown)


RNReanimated.podspec
```diff
@@ -15,6 +15,7 @@ boost_compiler_flags = '-Wno-documentation'
 fabric_flags = fabric_enabled ? '-DRCT_NEW_ARCH_ENABLED' : ''
 example_flag = config[:is_reanimated_example_app] ? '-DIS_REANIMATED_EXAMPLE_APP' : ''
 version_flag = '-DREANIMATED_VERSION=' + reanimated_package_json["version"]
+using_hermes = ENV['USE_HERMES'] == nil || ENV['USE_HERMES'] == "1"
```
tomekzaw, line 18: Just for consistency
```suggestion
using_hermes = ENV['USE_HERMES'] == nil || ENV['USE_HERMES'] == '1'
```

tomekzaw (approved):

davidgvf:
When merge? I hace this error and cant build ios app

Lissy93 (approved):
It'd be good to get this merged, as it resolves an error that a few of us seem to be facing after updating.

diego-antonelli:
@tomekzaw The check introduced in this PR is wrong.. It should be `using_hermes = ENV['USE_HERMES'] != nil || ENV['USE_HERMES'] == '1'`. Running version 3.3.0 now with hermes disabled failed with React-Hermes not found in podspec.

FYI @yupengyang

diego-antonelli:
or even `using_hermes = ENV['USE_HERMES'] != nil && ENV['USE_HERMES'] == '1'`

diego-antonelli:
Gonna apply a patch in our app then as we do not use hermes by default

diego-antonelli:
You guys could also rely on config[:hermes_enabled] from `use_react_native`

diego-antonelli:
For now I just forced `ENV["USE_HERMES"] = "0" in the Podfile, I also dont know if is possible to get the configs from the `use_react_native`. Anyways, thanks for the very quick response @tomekzaw

Issue #4510, opened 2023-05-31
Reanimated v 3.2.0 has broken withSpring animations


nicolascavallin:
### Description

We have created a pin input (like iOS or Android locked screen), so you input your pin and if it's wrong there is a shake animation and reset the pin.

We are using a sequence with `withTiming` and `withSpring`. 

Everything works well with reanimated 3.1.0. After update to 3.2.0

Replacing the `withSpring` animation with `withTiming` the bug doesn't appears.

### With Reanimated v 3.2.0

https://github.com/software-mansion/react-native-reanimated/assets/22961489/0c62a821-4d14-4824-848e-4b08a267027c

### with Reanimated v 3.1.0

https://github.com/software-mansion/react-native-reanimated/assets/22961489/e1236349-15a9-425a-8cd0-c844476f587e

### Steps to reproduce

1. Create a sequence animation that includes `withSpring`
2. Run it

The sequence we are using is:

```
    pinShake.value = withSequence(
      withTiming(20, { duration: 75 }),
      withTiming(-20, { duration: 75 }),
      withTiming(20, { duration: 75 }),
      withSpring(0, { damping: 50, stiffness: 500, mass: 2.8 }, () => runOnJS(resetPin)()),
    );
```

But if we just use `withSpring`(without a sequence) it's also broken.

### Snack or a link to a repository

pending

### Reanimated version

3.2.0

### React Native version

0.71.8

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

None

### Acknowledgements

Yes

Latropos:
Thank you for posting this issue! Looks that we have solved, so version 3.2.1 will surely include the fix.

PR #4512, opened 2023-05-31
Fix spring bug with extra type verification

Latropos:
## Summary

There is a need to check if velocity from previous animation is undefined.
(Hopefully such a bugs will be soon reported by typescript 😊)

<table>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/638f0fea-9671-42e6-9d05-2e6717cf3645

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/82d656f3-c342-4e8b-9d31-a4bbf4df12a8

</td></tr>
</table>
## Test plan
Tested on added to our bouncing box animation sequence from this issue https://github.com/software-mansion/react-native-reanimated/issues/4510:

```diff

    .onFinalize(() => {
+      offset.value = withSequence(
+        withTiming(20, { duration: 75 }),
+        withTiming(-20, { duration: 75 }),
+        withTiming(20, { duration: 75 }),
+        withSpring(0, { damping: 50, stiffness: 500, mass: 2.8 })
+      );
    });
```

### DIFF
--- a/src/reanimated2/animation/spring.ts
+++ b/src/reanimated2/animation/spring.ts
@@ -84,26 +84,25 @@ export function withSpring(
               t,
             });
 
-      if (!config.useDuration) {
-        const { isOvershooting, isVelocity, isDisplacement } =
-          isAnimationTerminatingCalculation(animation, config);
+      animation.current = newPosition;
+      animation.velocity = newVelocity;
 
-        animation.current = newPosition;
-        animation.velocity = newVelocity;
+      const { isOvershooting, isVelocity, isDisplacement } =
+        isAnimationTerminatingCalculation(animation, config);
 
-        const springIsNotInMove =
-          isOvershooting || (isVelocity && isDisplacement);
+      const springIsNotInMove =
+        isOvershooting || (isVelocity && isDisplacement);
 
-        if (springIsNotInMove) {
-          if (config.stiffness !== 0) {
-            animation.velocity = 0;
-            animation.current = toValue;
-          }
-          // clear lastTimestamp to avoid using stale value by the next spring animation that starts after this one
-          animation.lastTimestamp = 0;
-          return true;
+      if (!config.useDuration && springIsNotInMove) {
+        if (config.stiffness !== 0) {
+          animation.velocity = 0;
+          animation.current = toValue;
         }
+        // clear lastTimestamp to avoid using stale value by the next spring animation that starts after this one
+        animation.lastTimestamp = 0;
+        return true;
       }
+
       return false;
     }
 
@@ -138,11 +137,12 @@ export function withSpring(
           (previousAnimation?.startValue as number)
         : Number(animation.toValue) - value;
 
-      animation.velocity = previousAnimation
-        ? triggeredTwice
-          ? previousAnimation.velocity
-          : previousAnimation.velocity + config.velocity
-        : config.velocity;
+      animation.velocity =
+        (previousAnimation
+          ? triggeredTwice
+            ? previousAnimation?.velocity
+            : previousAnimation.velocity + config.velocity
+          : config.velocity) || 0;
 
       if (triggeredTwice) {
         animation.zeta = previousAnimation?.zeta || 0;


Eric-Andrade:
I need this change, please 👍

src/reanimated2/animation/spring.ts
```diff
@@ -138,11 +137,12 @@ export function withSpring(
...
-        ? triggeredTwice
-          ? previousAnimation.velocity
-          : previousAnimation.velocity + config.velocity
-        : config.velocity;
+      animation.velocity =
+        (previousAnimation
+          ? triggeredTwice
+            ? previousAnimation?.velocity
+            : previousAnimation.velocity + config.velocity
+          : config.velocity) || 0;
```
piaskowyk, line 145: Can we rewrite it to standard `if ()` syntax? because this one is complicated to read

src/reanimated2/animation/spring.ts
```diff
@@ -84,26 +84,25 @@ export function withSpring(
...
+      animation.velocity = newVelocity;
+
+      const { isOvershooting, isVelocity, isDisplacement } =
+        isAnimationTerminatingCalculation(animation, config);
+
+      const springIsNotInMove =
+        isOvershooting || (isVelocity && isDisplacement);
+
+      if (!config.useDuration && springIsNotInMove) {
+        if (config.stiffness !== 0) {
```
piaskowyk, line 97: Could you explain this condition? Why we need to reset velocity only in this case?

piaskowyk:
I have found a issue when we set stiffness on 0:

https://github.com/software-mansion/react-native-reanimated/assets/36106620/bc8a2b1c-8829-4316-89f9-dd6a38cf9bbd

### Repro
<details>
<summary>code</summary>

```js
import Animated, {
  useSharedValue,
  withTiming,
  withSpring,
  useAnimatedStyle,
  Easing,
import React from 'react';

export default function AnimatedStyleUpdateExample(): React.ReactElement {
    return {
      width: withSpring(randomWidth.value, { duration: 1000, stiffness: 0 }),
    };
  });

  return (
    <View
      style={{
        flex: 1,
        flexDirection: 'column',
      }}>
      <Animated.View
        style={[
          style,
        ]}
      />
      <Button
        title="toggle"
        onPress={() => {
          randomWidth.value = Math.random() * 300 + 50;
        }}
      />
    </View>
  );
}
```

</details>

src/reanimated2/animation/spring.ts
Latropos: I don't know but we have been doing that for years: https://github.com/software-mansion/react-native-reanimated/pull/2182/files

src/reanimated2/animation/spring.ts
```diff
@@ -112,6 +136,7 @@ export function withSpring(
       animation: SpringAnimation
     ) {
       return (
+        previousAnimation?.startTimestamp &&
```
piaskowyk, line 139: I ask just in case 😅 Do you know that, when `startTimestamp` will be equal `0` then `previousAnimation?.startTimestamp = false` ?

src/reanimated2/animation/spring.ts
```diff
@@ -56,6 +73,16 @@ export function withSpring(
         return true;
       }
 
+      if (config.isWrongConfig) {
+        // We don't animate wrong config
+        if (config.useDuration) return false;
+        else {
+          animation.current = toValue;
+          animation.lastTimestamp = 0;
+          console.log('STOP');
```
piaskowyk, line 82: we need to remove this `console.log`

Latropos, line 82:
```suggestion
```

src/reanimated2/animation/spring.ts
```diff
@@ -56,6 +73,16 @@ export function withSpring(
         return true;
       }
 
+      if (config.isWrongConfig) {
+        // We don't animate wrong config
+        if (config.useDuration) return false;
```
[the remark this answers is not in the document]
Latropos: Theoretically if user provides negative damping ration the spring should bounce more and more for ever

src/reanimated2/animation/spring.ts
Latropos: Yes I do, we have already identical assertion in `timing.ts`

src/reanimated2/animation/spring.ts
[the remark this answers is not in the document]
Latropos: Yes, so I've added conditional to instantly return 0 in such a case.

src/reanimated2/animation/spring.ts
```diff
@@ -56,6 +78,15 @@ export function withSpring(
         return true;
       }
 
+      if (config.configIsInvalid) {
+        // We don't animate wrong config
+        if (config.useDuration) return false;
```
piaskowyk, line 83: I am afraid if this condition could be a cause of infinity loop. I think that we should end the animation here, why you want to continue animation?

PR #4518, opened 2023-06-01
Fixed minor Android measure function small performance issue

michalmaka:
## Summary

I've noticed that there are some unnecessary reallocations during returning result from Android's measure function. Without any memory reservation, push_back may cause several reallocations and copy of the data.

## Test plan

None


tomekzaw:
Thanks for this PR! Would you like to apply this change on iOS as well?
https://github.com/software-mansion/react-native-reanimated/blob/663ee74925cafa4a1a802f4722466dc02cb1760f/ios/native/NativeMethods.mm#L28-L37

tomekzaw (approved):
Thanks for the PR! 👏 @michalmaka

PR #4519, opened 2023-06-01
Remove react-native-reanimated.d.ts

tjzel:
## Motivation

Since React Native Reanimated started as a pure JavaScript library at some point it was decided to migrate to TypeScript. Code base was complex enough for this task to be difficult and automatically generated type definitions weren't coherent with the API. Therefore a manually written file was required to provide better types for people that use TypeScript. However this created problems on its own as not all of the changes were put into the `d.ts` file and differences between inferred type and manual type could be very significant. After time we circled back to complete the TypeScript migration and remove this intermediate `d.ts` file.

## Summary

The goal of this PR is to remove `react-native-reanimated.d.ts` file that had all of our manually written type definitions in favor of automatically generated files stemming from source files. Since this was still a strenuous task we've decided not to fix all the internal types just yet and preserve only the necessary manually written types from the `d.ts` file. There's quite a lot of type casting due to that but this is planned to be corrected in the future as well.

Fixes: #4616, #4556 

## Test plan

I've added the following script to our `package.json`: `yarn type:check-api`. It compiles source code of our examples with the use of autogenerated type definitions. It was my main tool of performing this migration an was extremely handy.

All of our static check were useful as well - of course linting and type checking the source code.

Since our Examples don't cover all of the cases of our API usage (although I believe they cover a significant part of it) and we've kept as little of the `d.ts` file as possible it's to be expected for type errors to be reported in the future - but I think it's the better way to do it than to attempt to generate more API use cases by us.


src/reanimated2/animation/delay.ts
```diff
@@ -7,7 +7,13 @@ import {
 } from '../commonTypes';
 import { DelayAnimation } from './commonTypes';
 
-export function withDelay<T extends AnimationObject>(
+// TODO TYPESCRIPT This is a temporary type to get rid of .d.ts file.
+type withDelayType = <T extends AnimatableValue>(
+  delayMS: number,
```
piaskowyk, line 12:
```suggestion
  delayMs: number,
```

src/reanimated2/layoutReanimation/defaultAnimations/Bounce.ts
```diff
@@ -520,7 +522,7 @@ export class BounceOutRight
     const callback = this.callbackV;
     const initialValues = this.initialValues;
 
-    return (values) => {
+    return (values: EntryAnimationsValues | ExitAnimationsValues) => {
```
piaskowyk, line 525: We repeat this union many times, could you create wrapper for this union?

piaskowyk (commented):
👍🐎

src/reanimated2/helperTypes.ts
```diff
@@ -0,0 +1,107 @@
...
+} from '.';
+import type { ReanimatedKeyframe } from './layoutReanimation/animationBuilder/Keyframe';
+
+type Adaptable<T> = T | ReadonlyArray<T | ReadonlyArray<T>> | SharedValue<T>;
+
+type AdaptTransforms<T> = {
+  [P in keyof T]: Adaptable<T[P]>;
+};
+
+interface TransformsStyle {
```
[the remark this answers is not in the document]
tjzel: RN TransformsStyle has more props than our declared type but I have just made it to be

`type TransformsStyle = Pick<RNTransformsStyle, 'transform'>;` to be more clear.

src/reanimated2/threads.ts
```diff
@@ -169,7 +169,9 @@ export function runOnJS<A extends any[], R>(
   }
   return (...args) => {
     _scheduleOnJS(
+      // @ts-ignore TODO TYPESCRIPT omitted in the past
```
Latropos, line 172: You mean sth like "TODO TYPESCRIPT One of bugs ignored when introducing types"? 🤔  Could you please rewrite it?

src/reanimated2/hook/useAnimatedGestureHandler.ts
```diff
@@ -18,38 +16,43 @@ export interface GestureHandlers<T, TContext extends Context> {
...
 
-export enum EventType {
-  UNDETERMINED = 0,
-  FAILED,
-  BEGAN,
-  CANCELLED,
-  ACTIVE,
-  END,
-}
+export const EventType = {
```
Latropos, line 19: Why do we change enum to object?

src/reanimated2/component/Text.ts
```diff
@@ -1,6 +1,15 @@
...
+declare class AnimatedTextClass extends Component<AnimateProps<TextProps>> {
+  getNode(): Text;
+}
 
-export default AnimatedText;
+export const AnimatedText = createAnimatedComponent(
+  Text as any
+) as unknown as AnimatedText; // TODO TYPESCRIPT This temporary cast is to get rid of .d.ts file.
+
+export type AnimatedText = typeof AnimatedTextClass & AnimatedTextClass & Text;
```
Latropos, line 15: Why do we have `= typeof AnimatedTextClass & AnimatedTextClass`?

__typetests__/react-native-reanimated-tests.tsx
```diff
@@ -738,4 +752,68 @@ function testPartialAnimatedProps() {
   const test5 = (
     <AnimatedImage source={{ uri: 'whatever' }} animatedProps={aps} />
   );
+
+  // NativeMethods:
```
Latropos, line 756: Is this related to typescript changes?

src/reanimated2/threads.ts
tjzel: This is runOnJS issue I ran into when fixing types that's in #4572.

__typetests__/react-native-reanimated-tests.tsx
tjzel: Yes it is.

src/reanimated2/hook/useAnimatedGestureHandler.ts
tjzel: cc @j-piasecki on this one, we had a conversation about it a long while ago and from what I remember we concluded that using an object is better here, but I don't remember exact reasons of this. Change was made in #3772.

src/createAnimatedComponent.tsx
```diff
@@ -42,6 +50,7 @@ import { getShadowNodeWrapperFromRef } from './reanimated2/fabricUtils';
 import updateProps from './reanimated2/UpdateProps';
 import NativeReanimatedModule from './reanimated2/NativeReanimated';
 import { isSharedValue } from './reanimated2';
+import { AnimateProps } from './reanimated2/helperTypes';
```
Latropos, line 53: Why don't we import type here?
Can we add [eslint rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md) to always import type if possible? WDYT?

```suggestion
import type { AnimateProps } from './reanimated2/helperTypes';
```

Latropos (approved):

src/reanimated2/hook/useAnimatedGestureHandler.ts
j-piasecki: IIRC, it's related to change below, I don't think `(typeof EventType)[keyof typeof EventType]` would work on enum. Could be wrong though.

src/reanimated2/js-reanimated/index.ts
```diff
@@ -27,7 +27,7 @@ const reanimatedJS = new JSReanimated();
 global._makeShareableClone = (c) => c;
 global._scheduleOnJS = (func, args) => {
   if (args) {
-    queueMicrotask(() => func(...args));
+    queueMicrotask(() => func(args));
```
piaskowyk, line 30: Are you sure that we can remove spread operator here?

piaskowyk (commented):
1.
image
2.
image

src/createAnimatedComponent.tsx
tjzel: This is a great idea. I have done a test run and got (treating as errors):
Screenshot 2023-06-29 at 09 31 40
So in this case it should be a separate, follow-up pull request.

tjzel:
> image 2. image

1. Error comes from TypeScript inferring an array a bit higher in the code as an array of any. This fixes it:
```Diff
-  const output = [];
+  const output: JSX.Element[] = [];
```
2. Error is caused, from what I understand, imageRef being a inferred as a react-native component instead of being reanimated animated Image counterpart component. (tl;dr - improperly typed example).

I think it's better to fix these changes in separate PR, not to bloat this one more.

app/src/examples/OldMeasureExample.tsx
```diff
@@ -195,7 +198,7 @@ function SectionHeader({
   } else {
     onActiveImpl = () => {
       'worklet';
-      applyMeasure(measure(animatedRef));
+      applyMeasure(measure(animatedRef)!);
```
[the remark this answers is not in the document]
tjzel: Non-null assertion (`!` operator) is only present in TypeScript and doesn't get compiled to JavaScript code. In this example measure (I believe) returns a non-null value hence we can assert it to TS here.

If you get the following snippet:
```TypeScript
// TypeScript
declare const foo: (() => void) | null;
foo();
foo!();
```
It gets compiled to:
```JavaScript
// JavaScript
foo();
foo();
```

src/createAnimatedComponent.tsx
```diff
@@ -241,10 +250,20 @@ export interface InitialComponentProps extends Record<string, unknown> {
...
+export default function createAnimatedComponent<P extends object>(
+  component: ComponentClass<P>,
+  options?: Options<P>
+): ComponentClass<AnimateProps<P>>;
+
 export default function createAnimatedComponent(
   Component: ComponentType<InitialComponentProps>,
   options?: Options<InitialComponentProps>
-): ComponentType<AnimatedComponentProps<InitialComponentProps>> {
+): any {
```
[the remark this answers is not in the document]
tjzel: This is simply [function overloading](https://www.typescriptlang.org/docs/handbook/2/functions.html#writing-good-overloads) in TypeScript.

src/reanimated2/NativeMethods.ts
```diff
@@ -5,19 +5,18 @@ import {
...
-import { Component } from 'react';
-import { RefObjectFunction } from './hook/commonTypes';
+import { AnimatedRef } from './hook/commonTypes';
+import { Component, RefObject } from 'react';
 
 const IS_NATIVE = !shouldBeUseWeb();
 
 export let measure: (
-  animatedRef: RefObjectFunction<Component>
+  animatedRef: RefObject<Component>
```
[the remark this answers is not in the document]
tjzel: It's like this in current `d.ts`

```TypeScript
  export function measure<T extends Component>(
    animatedRef: RefObject<T>
  ): MeasuredDimensions | null;
```

Can't just change it because of TypeScript errors in our examples - although the issue is likely the improper typing in those examples. I'd like to fix that in following PRs.

src/reanimated2/layoutReanimation/defaultAnimations/Flip.ts
```diff
@@ -203,8 +216,10 @@ export class FlipInEasyY
   extends ComplexAnimationBuilder
   implements IEntryExitAnimationBuilder
 {
-  static createInstance(): FlipInEasyY {
-    return new FlipInEasyY();
+  static createInstance<T extends typeof BaseAnimationBuilder>(
+    this: T
+  ): InstanceType<T> {
+    return new FlipInEasyY() as InstanceType<T>;
```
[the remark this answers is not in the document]
tjzel: cc @j-piasecki on that

src/reanimated2/js-reanimated/index.ts
tjzel: No, this is a mistake.

src/reanimated2/hook/useAnimatedSensor.ts
```diff
@@ -70,17 +70,25 @@ function adjustVectorToInterfaceOrientation(data: Value3D) {
...
+): AnimatedSensor<ValueRotation>;
+export function useAnimatedSensor(
+  sensorType: Exclude<SensorType, SensorType.ROTATION>,
+  userConfig?: Partial<SensorConfig>
+): AnimatedSensor<Value3D>;
 export function useAnimatedSensor(
   sensorType: SensorType,
   userConfig?: Partial<SensorConfig>
-): AnimatedSensor {
+): AnimatedSensor<any> {
```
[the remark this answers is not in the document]
tjzel: This is also function overloading.

PR #4523, opened 2023-06-02
Unify iOS initialization / Add support for 0.72.0-rc.5

### LINKED ISSUE #4521: Upgrading 0.72.0-rc.5 with 3.2.0 causes runtime error with metro.
### Description

I'm experiencing an error after upgrading RN to `0.72.0-rc.5`. I tried reseting the cache with `yarn start --reset-cache` too.

```
 ERROR  Error: [Reanimated] The native part of Reanimated doesn't seem to be initialized. This could be caused by
  - not rebuilding the app after installing or upgrading Reanimated
  - trying to run Reanimated on an unsupported platform
  - running in a brownfield app without manually initializing the native library, js engine: hermes
 LOG  Running "mobile" with {"rootTag":1,"initialProps":{}}
 ERROR  Invariant Violation: "mobile" has not been registered. This can happen if:
* Metro (the local dev server) is run from the wrong folder. Check if Metro is running, stop it and restart it in the current project.
* A module failed to load due to an error and `AppRegistry.registerComponent` wasn't called., js engine: hermes
```

CleanShot 2023-06-01 at 12 01 12@2x

### Steps to reproduce

1. Create a project with reanimated 3.2.0
2. Upgrade React Native to `0.72.0-rc.5`
3. Run the app

n/a

### Reanimated version

3.2.0

### React Native version

0.72.0-rc.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

tomekzaw:
## Summary

This PR bumps react-native to 0.72.0-rc.5 and slightly changes initialization path on iOS (Paper) due to recent changes in the framework (https://github.com/facebook/react-native/pull/37523).

Fixes #4521. Continues #3005.

On iOS, Reanimated needs to overwrite two React internal modules:
* `RCTUIManager` &rarr; `REAUIManager` to intercept `manageChildren` calls in order to observe React tree changes ([here](https://github.com/software-mansion/react-native-reanimated/blob/main/ios/LayoutReanimation/REAUIManager.mm))
* `RCTEventDispatcher` &rarr; `REAEventDispatcher` to intercept events in the native code while still on the UI thread ([here](https://github.com/software-mansion/react-native-reanimated/blob/663ee74925cafa4a1a802f4722466dc02cb1760f/ios/REAEventDispatcher.m#L9-L10))

The rest of the initialization (injecting JSI bindings) can be safely done in `installTurboModule` method, as we already do on Android and iOS/Fabric.

Previously, this was done using categories (see [`UIResponder+Reanimated.mm`](https://github.com/software-mansion/react-native-reanimated/blob/663ee74925cafa4a1a802f4722466dc02cb1760f/ios/native/UIResponder%2BReanimated.mm#L19-L30)), in particular by overwriting `jsExecutorFactoryForBridge:` method which is called during initialization. However, since 0.72.0-rc.4, this method is already implemented in `RCTAppDelegate` so the trick does no longer work, making the app fail with the following error "[Reanimated] The native part of Reanimated doesn't seem to be initialized".

In this PR, I've used a category to overwrite another method `extraModulesForBridge` which swizzles the implementation of `jsExecutorFactoryForBridge` method and runs `REAInitializer` before the original call.

As suggested by @kmagiera, alternatively we could just swizzle the methods of `RCTUIManager` and `RCTEventDispatcher`.

## TODO

- [x] Restore support for 0.71 and below with #ifdefs


tomekzaw:
@MicroDroid Probably next week, before 0.72 stable is released.

MicroDroid:
If I modify my package.json to pull the package straight from this pull request, would the package work?

I have no idea how pods work, not sure if something needs to be published somewhere

tomekzaw:
@MicroDroid Yeah, you can install Reanimated from a particular commit with `yarn add software-mansion/react-native-reanimated#a6b0515c0d0d554dc584d3cda6dba4ce3bba7bce`

MicroDroid:
@tomekzaw 
image

MicroDroid:
I noticed that you, [here](https://github.com/software-mansion/react-native-reanimated/pull/4523/files#diff-2f3e36d153f0b1ce894c5dd8afad3b4119dd38054d65d7a593f60cf2762edf1aR100), added `React-RCTAppDelegate` as a dependency under some conditions, so I `rm -rf ./ios/Pods/`, `rm ./ios/Podfile.lock` (I know I shouldn't), and did a `Clean Build Folder` in Xcode, and yet the issue persists.

I do *not* use Fabric, and I am using RN 0.72.0-rc.5

tomekzaw:
@MicroDroid `cd ios && rm -rf Pods Podfile.lock build && pod install` usually works for me.

MicroDroid:
Tried exactly that now, but still no go.

I don't feel like it's a caching issue... not sure what it could be though.

MicroDroid:
I `echo $RCT_NEW_ARCH_ENABLED` and the result was ``, as expected.

MicroDroid:
More interestingly, I went ahead and:
1. Added random characters in `RNReanimated.podspec`, and `pod install` errored, which means `pod install` is picking it up
2. I removed the `if` statement wrapping the missing dependency entirely
3. `pod install`
4. `rm -rf build/`
5. ⌘B

... and still the same error

MicroDroid:
This builds:

image

tomekzaw:
@MicroDroid What happens if you run `cd ios && find . -name "*RCTAppDelegate*"`?

MicroDroid:
```
./Pods/Pods.xcodeproj/xcuserdata/yousef.xcuserdatad/xcschemes/React-RCTAppDelegate.xcscheme
./Pods/Target Support Files/React-RCTAppDelegate
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate-dummy.m
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate.debug.xcconfig
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate-umbrella.h
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate-prefix.pch
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate-Info.plist
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate.modulemap
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate.release.xcconfig
./Pods/Local Podspecs/React-RCTAppDelegate.podspec.json
```

MicroDroid:
```
./Pods/Pods.xcodeproj/xcuserdata/<user>.xcuserdatad/xcschemes/React-RCTAppDelegate.xcscheme
./Pods/Target Support Files/React-RCTAppDelegate
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate-dummy.m
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate.debug.xcconfig
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate-umbrella.h
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate-prefix.pch
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate-Info.plist
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate.modulemap
./Pods/Target Support Files/React-RCTAppDelegate/React-RCTAppDelegate.release.xcconfig
./Pods/Local Podspecs/React-RCTAppDelegate.podspec.json
```

tomekzaw:
@MicroDroid Indeed, there's no `RCTAppDelegate.h` in your output. But it works for me locally and on our CI as well.

MicroDroid:
Not sure honestly, I did `pod install` like a billion times, no clue what's happening

ios/native/RCTAppDelegate+Reanimated.mm
```diff
@@ -0,0 +1,28 @@
...
+
+@implementation RCTAppDelegate (Reanimated)
+
+- (std::unique_ptr<facebook::react::JSExecutorFactory>)swizzled_jsExecutorFactoryForBridge:(RCTBridge *)bridge
+{
+  reanimated::REAInitializer(bridge);
+  return [self swizzled_jsExecutorFactoryForBridge:bridge]; // call the original method
+}
+
++ (void)load
```
kmagiera, line 15: Is using load vs initialize a common practice here?

kmagiera (approved):
No super happy we are adding more version specific code. Let's get this one in but we should seriously consider either unifying initialization logic (it appears like we still use the same jsExecutorFactorySth method, so it looks like we should be able to do this across older versions too), or dropping support for older RN releases

ios/native/RCTAppDelegate+Reanimated.mm
tomekzaw: Honestly that's just what ChatGPT came up with.

ios/native/RCTAppDelegate+Reanimated.mm
```diff
@@ -0,0 +1,28 @@
...
+{
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    Class cls = [self class];
+    Method originalMethod = class_getInstanceMethod(cls, @selector(jsExecutorFactoryForBridge:));
+    Method swizzledMethod = class_getInstanceMethod(cls, @selector(swizzled_jsExecutorFactoryForBridge:));
+    method_exchangeImplementations(originalMethod, swizzledMethod);
```
tido64, line 22: Will this still work if the instance doesn't implement `jsExecutorFactoryForBridge:`?

I wonder if you should use something like this: https://github.com/microsoft/react-native-test-app/blob/2e5d0e64535230ed07a4ecf398387057b34b7121/ios/ReactTestApp/React%2BCompatibility.m#L7

I think you should also use a different prefix for the method name to avoid potential name clashes, e.g. `reanimated_jsExecutorFactoryForBridge`.

tomekzaw: Hey @tido64, thanks for the review! I've renamed the suffix in d018b2cc511df8165157e306a791d773bce2901d as you suggested 😄 

As for the case when `jsExecutorFactoryForBridge:` is not implemented, indeed we don't handle it properly now but it looks like since 0.72.0-rc.4 this is always implemented in `RCTAppDelegate`, at least in greenfield apps. We could add a conditional on `didAddMethod` as you proposed but I'm not sure what would happen in the line where we call `[self reanimated_jsExecutorFactoryForBridge:bridge]` then. I need some more time to learn and think about it.

Thanks for the code pointer to `RTASwizzleSelector`, would you mind if we copy it to Reanimated codebase as a utility function? We also need to swizzle some methods of ViewControllers for Shared Element Transitions. 

Finally, I'm curious to know what's the reason to use `+initialize` [here](https://github.com/microsoft/react-native-test-app/blob/2e5d0e64535230ed07a4ecf398387057b34b7121/ios/ReactTestApp/React%2BCompatibility.m#L38), I found one article that recommends to perform swizzling always in `+load` method ([link](https://nshipster.com/method-swizzling/#load-vs-initialize)).

tido64: > Thanks for the code pointer to `RTASwizzleSelector`, would you mind if we copy it to Reanimated codebase as a utility function? We also need to swizzle some methods of ViewControllers for Shared Element Transitions.

Feel free to copy it 👍

> Finally, I'm curious to know what's the reason to use `+initialize` [here](https://github.com/microsoft/react-native-test-app/blob/2e5d0e64535230ed07a4ecf398387057b34b7121/ios/ReactTestApp/React%2BCompatibility.m#L38), I found one article that recommends to perform swizzling always in `+load` method ([link](https://nshipster.com/method-swizzling/#load-vs-initialize)).

The reason behind using `+initialize` is because we only want to swizzle if the class is actually used. In our case, race condition isn't really a concern since React doesn't initialize modules on multiple threads (at least not currently). We also use `dispatch_once` here so its atomicity is ensured.

tomekzaw: > Feel free to copy it 👍

Thanks!

> The reason behind using `+initialize` is because we only want to swizzle if the class is actually used. In our case, race condition isn't really a concern since React doesn't initialize modules on multiple threads (at least not currently). We also use `dispatch_once` here so its atomicity is ensured.

That makes perfect sense, thanks!

tomekzaw: > Will this still work if the instance doesn't implement `jsExecutorFactoryForBridge:`?

@tido64 I'm still thinking about this question. Our aim is to automatically initialize Reanimated for greenfield apps; if some custom app doesn't use `RCTAppDelegate`, it needs to perform initialization on its own. It looks like since 0.72.0-rc.4 `jsExecutorFactoryForBridge:` is already implemented in RCTAppDelegate ([link](https://github.com/facebook/react-native/blob/e11396e9980761a205d5cd96d5129f5385fd074b/packages/react-native/Libraries/AppDelegate/RCTAppDelegate.mm#L143-L159)). This means that we can assume that `didAddMethod` returns `false` which reduces to the code that we currently have in this PR, or am I missing something?

tomekzaw: > As far as I can tell, Reanimated only needs to swizzle `jsExecutorFactoryForBridge:` on Paper. Is that correct? 

Yes, that's correct. This PR introduces swizzling in `RCTAppDelegate` only for React Native 0.72+:

https://github.com/software-mansion/react-native-reanimated/blob/d018b2cc511df8165157e306a791d773bce2901d/ios/native/RCTAppDelegate%2BReanimated.mm#L1

For 0.71 and older we will still use the old way which is to implement `jsExecutorFactoryForBridge:` method for `UIResponder`:

https://github.com/software-mansion/react-native-reanimated/blob/d018b2cc511df8165157e306a791d773bce2901d/ios/native/UIResponder%2BReanimated.mm#L1

tido64: Ah, I missed this part. Then it looks like you should be good for now.

piaskowyk (approved):

hiagolfMB:
Is it already available on npm?

tomekzaw:
Yes, you can install nightly build with `yarn add react-native-reanimated@nightly`.

tomekzaw:
@MicroDroid Here's the issue for the problem that you've mentioned: https://github.com/software-mansion/react-native-reanimated/issues/4553

PR #4525, opened 2023-06-02
Memoize ReanimatedModule in `sendEvent`

tomekzaw:
## Summary

This PR adds `static __weak REAModule *reaModule;` variable which stores the result of `[[self bridge] moduleForName:@"ReanimatedModule"]` so we can call it only once during initialization instead of on each event.

When `REAModule` is deallocated, all `__weak` pointers are automatically replaced with `nil` so it doesn't leak any resources as well as works fine after a reload.

### DIFF
--- a/ios/RCTEventDispatcher+Reanimated.h
+++ b/ios/RCTEventDispatcher+Reanimated.h
@@ -1,3 +1,5 @@
+#ifndef RCT_NEW_ARCH_ENABLED
+
 #import <Foundation/Foundation.h>
 #import <React/RCTEventDispatcher.h>
 
@@ -10,3 +12,5 @@ NS_ASSUME_NONNULL_BEGIN
 @end
 
 NS_ASSUME_NONNULL_END
+
+#endif // RCT_NEW_ARCH_ENABLED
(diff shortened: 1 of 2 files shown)


ios/REAEventDispatcher.m
```diff
@@ -1,12 +1,22 @@
...
 #import <React/RCTBridge+Private.h>
 #import <React/RCTDefines.h>
 
 @implementation REAEventDispatcher
 
 - (void)sendEvent:(id<RCTEvent>)event
 {
-  [[[self bridge] moduleForName:@"ReanimatedModule"] eventDispatcherWillDispatchEvent:event];
+  __weak static REAModule *reaModule;
+  __weak static RCTBridge *lastBridge;
```
piaskowyk, line 13: What about retain cycle? Can we keep refference to bridge?

piaskowyk: You are right I missed it. 👍 

Just to be sure - did you tested those changes with continues events (for example scrolling) and app reload?

tomekzaw: Nope, not yet, but I will, thanks! Honestly, I'm a bit afraid about this change, not sure if the performance gain is worth it.

tomekzaw: Reloading while scrolling works fine, tested on ScrollViewExample.tsx.

PR #4527, opened 2023-06-02
Don't use `__DEV__` inside worklets

tomekzaw:
## Summary

Using `babel-plugin-transform-define` to replace `__DEV__` with `true` breaks worklets (reported by @Titozzz).

The root cause is that the plugin tries to replace worklets' closure e.g. `{__DEV__: __DEV__}` with `{true: true}` but the expected result would be `{__DEV__: true}` (or even `{}` since `__DEV__` inside function body is also replaced and optimized away).

This PR replaces all occurrences of `__DEV__` inside worklets with `IS_DEV` which is captured from outer scope.

## Test plan

Ask @Titozzz if it works.

### DIFF
--- a/src/reanimated2/threads.ts
+++ b/src/reanimated2/threads.ts
@@ -8,6 +8,7 @@ import {
 
 const IS_JEST = isJest();
 const IS_WEB = shouldBeUseWeb();
+const IS_DEV = __DEV__;
 
 let _runOnUIQueue: Array<[ComplexWorkletFunction<any[], any>, any[]]> = [];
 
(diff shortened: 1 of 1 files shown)


kmagiera (commented):
This does not seem like a viable solution to this issue. I think the problem lies on the babel-transform-plugin-define side which does not handle cases when the identifier is used as object key. Should be relatively easy to fix it there though. If we add this change in we will have the same issue in some other place soon

tomekzaw:
Submitted an issue in babel-plugin-transform-define repository:
* https://github.com/FormidableLabs/babel-plugin-transform-define/issues/85

PR #4533, opened 2023-06-05
Added getAnimatableRef function

behenate:
## Summary

Added an option for a component to specify which other component should be animated when `createAnimatedComponent` is executed on it by adding `getAnimatableRef` function (we might want to think of a better name). 

This is useful for some 3rd party libraries which want to add support for animating their component, but it's wrapped inside of another component which is inaccessible to the user.

## Test plan

Included a simple example screen where inner `Rect` component is animated when `createAnimatedComponent` is executed on the component containing it.

https://github.com/software-mansion/react-native-reanimated/assets/31368152/96d2cd19-a4f9-48ac-bd23-f50746dc7e0b

### DIFF
--- a/app/src/examples/AnimatableRefExample.tsx
+++ b/app/src/examples/AnimatableRefExample.tsx
@@ -1,0 +1,82 @@
+import React, { useState } from 'react';
+import { View, StyleSheet, Button } from 'react-native';
+
+import { Rect, Svg } from 'react-native-svg';
+import Animated, {
+  useAnimatedProps,
+  useSharedValue,
+  withSpring,
+} from 'react-native-reanimated';
+
+type SwitchProps = {
+  y?: number;
+};
+
+class Switch extends React.Component<SwitchProps> {
+  rectRef: Rect | null | undefined;
+
+  // When an animated version of the Switch is created we want to animate the inner Rect instead of the outer Svg component.
+  getAnimatableRef() {
+    return this.rectRef;
+  }
+
+  render() {
+    return (
+      <Svg height="310" width="70">
+        <Rect
+          x="5"
+          y="5"
+          width="60"
+          height="300"
+          fill="none"
+          stroke={'black'}
+          strokeWidth={10}
+        />
+        <Rect
+          x="10"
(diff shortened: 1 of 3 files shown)


app/src/examples/AnimatableRefExample.tsx
```diff
@@ -0,0 +1,82 @@
...
+  render() {
+    return (
+      <Svg height="310" width="70">
+        <Rect
+          x="5"
+          y="5"
+          width="60"
+          height="300"
+          fill="none"
+          stroke={'black'}
```
tomekzaw, line 32:
```suggestion
          stroke="black"
```

tomekzaw (approved):
Works like a charm. Thanks for the PR! 👏 

Tested with the following example:

<details>
<summary>App.tsx</summary>

```tsx

import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withRepeat,
  withTiming,

class MyClassComponent extends React.Component {
  innerRef: View | null = null;

  getAnimatableRef() {
    return this.innerRef;
  }

  render() {
    return (
      <View style={styles.outer}>
        <View
          ref={(component) => {
            this.innerRef = component;
          }}
          style={styles.inner}
        />
      </View>
    );
  }
}

const AnimatedMyClassComponent =
  Animated.createAnimatedComponent(MyClassComponent);

  const sv = useSharedValue(0);

  useEffect(() => {
    sv.value = 0;
    sv.value = withRepeat(withTiming(1), -1, true);
  }, [sv]);

    return {
      width: sv.value * 200,
    };
  });

  return (
      <AnimatedMyClassComponent style={animatedStyle} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  outer: {
    backgroundColor: 'black',
    width: 200,
    height: 200,
  },
  inner: {
    backgroundColor: 'cyan',
    width: 100,
    height: 100,
  },
});
```

</details>

tomekzaw:
Doesn't work on web though 😢

piaskowyk (approved):

PR #4536, opened 2023-06-06
Improve transform matrix animation

Latropos:
## Summary
We used to animate transformation matrices the very same way we animate every array - move each value from initial to final position. However this may look weird in case of transformation matrices, since scale and rotation may appear to change randomly - the solution is to decompose actual transformation matrix into:

- translation
- skale
- rotation
  - rotation along X axis 
  - rotation along Y axis 
  - rotation along Z axis 
- skew (shear)

and then we animate translation, skale and skew the way we used to, but for rotation we calculate all the three angles and get matrix for particular angle. Finally we multiply all the matrices in correct order.

See the recording for comparison. The $\text{\large \color{blue}{\color{lime}u}pper}$ square is animated with improved transform matrix interpolation, the $\text{\large \color{darkorange}{\color{red}l}ower}$ one uses the old technique. 

https://github.com/software-mansion/react-native-reanimated/assets/56199675/974c983c-0948-4c75-a833-46b99e93e0d1

## Test plan

I've added a new test suite
image


fukemy:
@Latropos I have same issues with transition the transform value, after back to current screen, the view lost the transform origin, as described here: https://github.com/mateoguzmana/simple-gallery/issues/4

 Can you help?

Latropos:
Hi @fukemy, thanks for report, however we don't support this particular thing yet - view loses its transform origin and this is an expected behaviour. If you want to transform your view around a specific anyway, please use this library: https://www.npmjs.com/package/react-native-anchor-point

src/reanimated2/animation/util.ts
```diff
@@ -190,6 +202,105 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
     return finished;
   };
 
+  // START
+
```
tomekzaw, line 206:
```suggestion
```

src/reanimated2/animation/util.ts
```diff
@@ -190,6 +202,105 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
...
+  const transformationMatrixOnStart = (
+    animation: Animation<AnimationObject>,
+    value: AffiniteMatrixFlat,
+    timestamp: Timestamp,
+    previousAnimation: Animation<AnimationObject>
+  ): void => {
+    const toValue = animation.toValue as AffiniteMatrixFlat;
```
tomekzaw, line 213: Do we really need `as AffiniteMatrixFlat` here?

src/reanimated2/animation/util.ts
```diff
@@ -190,6 +202,105 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
...
+
+    animation.startMatrices = decomposeMatrixIntoMatricesAndAngles(value);
+    animation.stopMatrices = decomposeMatrixIntoMatricesAndAngles(toValue);
+
+    /** We create an animation copy to animate  single value between 0 and 100
+     **/
```
tomekzaw, line 219:
```suggestion
    // We create an animation copy to animate single value between 0 and 100
```

src/reanimated2/animation/util.ts
```diff
@@ -190,6 +202,105 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
...
+    animation[0] = Object.assign({}, animationCopy);
+    animation[0].current = 0;
+    animation[0].toValue = 100;
```
tomekzaw, line 222: I believe it's more intuitive to animate progress from 0 to 1.

src/reanimated2/animation/util.ts
```diff
@@ -190,6 +202,105 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
...
+    animation[0].onStart(
+      animation[0],
+      0,
+      timestamp,
+      previousAnimation ? previousAnimation[0] : undefined
+    );
+
+    animation.current = value;
+  };
```
tomekzaw, line 231:
```suggestion
  };

```

src/reanimated2/animation/util.ts
```diff
@@ -190,6 +202,105 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
...
+      progress: number,
+      a: AffiniteMatrix,
+      b: AffiniteMatrix
+    ) {
+      return addMatrices(a, scaleMatrix(subtractMatrices(b, a), progress));
+    }
+
+    function applyProgessToNumber(progress: number, a: number, b: number) {
+      return a + progress * (b - a);
+    }
```
tomekzaw, line 251: Can we move these functions to the top-level?

src/reanimated2/animation/util.ts
```diff
@@ -190,6 +202,105 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
...
+
+    function applyProgressToMatrix(
+      progress: number,
+      a: AffiniteMatrix,
+      b: AffiniteMatrix
+    ) {
+      return addMatrices(a, scaleMatrix(subtractMatrices(b, a), progress));
+    }
+
+    function applyProgessToNumber(progress: number, a: number, b: number) {
```
tomekzaw, line 249:
```suggestion
    function applyProgressToNumber(progress: number, a: number, b: number) {
```

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+function grahamSchmidtAlghoritm(matrix: AffiniteMatrix): {
+  rotationMatrix: AffiniteMatrix;
+  skewMatrix: AffiniteMatrix;
+} {
+  'worklet';
+  const a1 = matrix[0];
+  const a2 = matrix[1];
+  const a3 = matrix[2];
+  const a4 = matrix[3];
```
tomekzaw, line 256:
```suggestion
  const [a1, a2, a3, a4] = matrix;
```
BTW maybe we should use either 0-3 or 1-4 but consistently?

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,369 @@
...
+  scalar: number
+): T {
+  'worklet';
+
+  return linearOperationOnMatrix(_a, (x) => x * scalar);
+}
+
+export function getRotationMatrix(
+  angle: number,
+  axis: 'x' | 'y' | 'z' = 'z'
```
tomekzaw, line 167: How about `type Axis = 'x' | 'y' | 'z';`?

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,369 @@
...
+    [m[2], m[6], m[10], m[14]],
+    [m[3], m[7], m[11], m[15]],
+  ];
+}
+
+function innerProduct(a: number[], b: number[]) {
+  'worklet';
+  if (!(a.length === b.length)) {
+    console.warn(
+      `Error when evaluation inner product of ${a} and ${b}, because these arrays should have equal length`
```
tomekzaw, line 217:
```suggestion
      `Cannot calculate inner product of two vectors of different length`
```

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,369 @@
...
+  a.forEach((_, i) => {
+    a[i] = func(a[i]);
+  });
+
+  return isFlatOnStart ? (a as T) : (unflatten(a) as T);
+}
+
+function linearOperationOnMatrices<
+  T extends AffiniteMatrixFlat | AffiniteMatrix
+>(_a: T, _b: T, operation: 'add' | 'subtract'): T {
```
tomekzaw, line 117: According to the convention, underscore (`_`) at the beginning of the argument name means that the argument is unused, however, it's not the case here.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+export function unflatten(matrix: AffiniteMatrixFlat): AffiniteMatrix {
+  'worklet';
+  const result = [];
+  const matrixCopy = [...matrix];
+  while (matrixCopy.length) {
+    result.push(matrixCopy.splice(0, 4));
+  }
+  return result as AffiniteMatrix;
+}
```
tomekzaw, line 66: This function only supports 4x4 matrices, can we simply hard-code the calculations instead of using `while` loop?

tomekzaw (changes requested):

src/reanimated2/animation/transformationMatrix/__tests__/matrixUtils.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+  decomposeMatrix,
+  flatten,
+  isAffiniteMatrix,
+  isAffiniteMatrixFlat,
+  multiplyMatrices,
+  subtractMatrices,
+  unflatten,
+} from '../matrixUtils';
+
+const identityMatrix: AffiniteMatrix = [
```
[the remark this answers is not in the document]
Latropos: Yes, only when we make this explicit assignment typescript validates shape of the matrix.

src/reanimated2/animation/util.ts
Latropos: I've kept `/** **/` instead of `// `, since I've added more lines to this comment, to explain why we animate between 0 an 100 (instead of 0 and 1), is it okay?

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,369 @@
...
+  _matrix: AffiniteMatrixFlat | AffiniteMatrix
+): TransformMatrixDecomposition {
+  'worklet';
+
+  const matrix: AffiniteMatrixFlat = maybeFlattenMatrix(_matrix);
+
+  // Normalize matrix
+  const lastElement: number = matrix[15];
+  if (lastElement === 0) console.error('Invalid transform matrix!');
+  _matrix.forEach((_, index) => (matrix[index] /= lastElement));
```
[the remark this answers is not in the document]
Latropos: Because matrix is `const`, and with map I would have to reassign it.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
Latropos: I would prefer including some info about the vectors, since one of them may be NaN or [], and knowing this at the beginning could speed up debugging.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,369 @@
...
+  const b: AffiniteMatrixFlat = maybeFlattenMatrix(_b);
+
+  a.forEach((_, i) => {
+    switch (operation) {
+      case 'subtract':
+        a[i] -= b[i];
+        break;
+      case 'add':
+        a[i] += b[i];
+        break;
```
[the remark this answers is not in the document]
Latropos: I was thinking about providing to lib users, since functions for matrix manipulation could be very usefull

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,369 @@
...
+  func: (x: number) => number
+): T {
+  'worklet';
+
+  const isFlatOnStart = !!isAffiniteMatrixFlat(_a);
+  const a: AffiniteMatrixFlat = maybeFlattenMatrix(_a);
+
+  a.forEach((_, i) => {
+    a[i] = func(a[i]);
+  });
```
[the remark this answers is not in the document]
Latropos: If we use `.map` typescript will forget we have verified that our array has exactly 16 elements.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
Latropos: Sure!

src/reanimated2/animation/spring.ts
```diff
@@ -129,6 +129,7 @@ export function withSpring(
         animation.current = toValue;
         // clear lastTimestamp to avoid using stale value by the next spring animation that starts after this one
         animation.lastTimestamp = 0;
+
```
Latropos, line 132:
```suggestion
```

src/reanimated2/animation/spring.ts
```diff
@@ -140,6 +140,7 @@ export function withSpring(
       animation: SpringAnimation
     ) {
       return (
+        previousAnimation?.lastTimestamp &&
```
Latropos, line 143: This is needed to make `triggered twice` work good, once I've removed `value = 0` in utils.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,436 @@
...
+    projection(u2, a3)
+  );
+
+  const [e0, e1, e2, e3] = [u0, u1, u2, u3].map((u) =>
+    scale(u, 1 / Math.sqrt(innerProduct(u, u)))
+  );
+  // const e0 = scale(u0, 1 / Math.sqrt(innerProduct(u0, u0)));
+  // const e1 = scale(u1, 1 / Math.sqrt(innerProduct(u1, u1)));
+  // const e2 = scale(u2, 1 / Math.sqrt(innerProduct(u2, u2)));
+  // const e3 = scale(u3, 1 / Math.sqrt(innerProduct(u3, u3)));
```
Latropos, line 333:
```suggestion
```

src/reanimated2/animation/util.ts
```diff
@@ -288,6 +408,7 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
       return;
     }
     baseOnStart(animation, value, timestamp, previousAnimation);
+    value = 0;
```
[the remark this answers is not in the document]
Latropos: Since spring animation used was thinking it is the same animation triggered twice, I've updated `spring.ts` for better results.

src/reanimated2/animation/util.ts
Latropos: `Unfortunately yes, to assert that toValue has 16 elements`

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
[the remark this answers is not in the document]
tomekzaw: Oh, right. But I guess in this case we could use `as` keyword to ensure the type 😉

src/reanimated2/animation/util.ts
```diff
@@ -190,6 +202,105 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
...
+    );
+
+    animation.current = value;
+  };
+  const transformationMatrixOnFrame = (
+    animation: Animation<AnimationObject>,
+    timestamp: Timestamp
+  ): boolean => {
+    let finished = 1;
+    finished &= animation[0].onFrame(animation[0], timestamp);
```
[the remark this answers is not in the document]
tomekzaw: I think we already discussed this offline and decided to use boolean so I'm resolving this conversation.

src/reanimated2/animation/util.ts
tomekzaw: Technically that's correct, however we already use `//` for multi-line comments throughout our codebase so maybe let's stick to this convention for the purpose of consistency.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,433 @@
...
+export function decomposeMatrix(
+  unknownTypeMatrix: AffiniteMatrixFlat | AffiniteMatrix
+): TransformMatrixDecomposition {
+  'worklet';
+
+  const matrix: AffiniteMatrixFlat = maybeFlattenMatrix(unknownTypeMatrix);
+
+  // Normalize matrix
+  const lastElement: number = matrix[15];
+  if (lastElement === 0) {
```
tomekzaw, line 362:
```suggestion
  if (matrix[15] === 0) {
```

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,433 @@
...
+  ];
+  return { rotationMatrix, skewMatrix };
+}
+
+export function decomposeMatrix(
+  unknownTypeMatrix: AffiniteMatrixFlat | AffiniteMatrix
+): TransformMatrixDecomposition {
+  'worklet';
+
+  const matrix: AffiniteMatrixFlat = maybeFlattenMatrix(unknownTypeMatrix);
```
tomekzaw, line 358: I believe this should work without `: AffiniteMatrixFlat`, the type should be inferred from the return type of `maybeFlattenMatrix`.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,433 @@
...
+  ];
+}
+
+function linearOperationOnMatrix<T extends AffiniteMatrixFlat | AffiniteMatrix>(
+  _a: T,
+  func: (x: number) => number
+): T {
+  'worklet';
+
+  const isFlatOnStart = !!isAffiniteMatrixFlat(_a);
```
tomekzaw, line 178: Do we need `!!` here? It looks like `isAffiniteMatrixFlat` always returns a boolean.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,433 @@
...
+export type AffiniteMatrixFlat = FixedLengthArray<number, 16>;
+
+export type TransformMatrixDecomposition = Record<
+  | 'translationMatrix'
+  | 'scaleMatrix'
+  | 'rotationMatrix'
+  | 'skewMatrix'
+  | 'scaleMatrix',
+  AffiniteMatrix
+>;
```
tomekzaw, line 21: Why not just ... ?
```ts
export type TransformMatrixDecomposition = {
  translationMatrix: AffiniteMatrix,
  scaleMatrix: AffiniteMatrix,
  rotationMatrix: AffiniteMatrix,
  skewMatrix: AffiniteMatrix,
  scaleMatrix: AffiniteMatrix,
};
```
This way we can enforce that the object actually contains all these fields.

Latropos: It is enforced with `Record` as well, all the parameters are required. IMHO Record is more natural here since all the fields have the very same type.

src/reanimated2/animation/util.ts
```diff
@@ -191,6 +217,96 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
...
+          currentScale,
+          multiplyMatrices(skewMatrix, rotationMatrix)
+        ),
+        currentTranslation
+      )
+    );
+
+    animation.current = updated;
+
+    return !!finished;
```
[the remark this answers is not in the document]
Latropos: There is an eslint rule to prevent this, but for some reasons it doesn't work in our repo...

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
tomekzaw: OK, but let's remove second `scaleMatrix` because currently we have two of those.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,433 @@
...
+      `Cannot calculate inner product of two vectors of different length. Length of ${a} is ${a.length} and length of ${b} is ${b.length}.)`
+    );
+    return 0;
+  }
+  return a.reduce((acc, _, i) => acc + a[i] * b[i], 0);
+}
+
+function projection(u: number[], a: number[]) {
+  'worklet';
+  const scaler = innerProduct(u, a) / innerProduct(u, u);
```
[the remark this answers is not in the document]
Latropos: "scaler" was expected to mean sth used to scale, but I've changed it to just "s"

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,433 @@
...
+      ];
+  }
+}
+
+function norm(x: number, y: number, z: number) {
+  'worklet';
+  return Math.sqrt(x * x + y * y + z * z);
+}
+
+function transposeMatrix(matrix: AffiniteMatrix): AffiniteMatrix {
```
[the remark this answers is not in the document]
Latropos: The ones that are very likely to be delivered to the user as a part of library support both input types, the ones that should be only internal don't. Before starting further work on matrix-manipulation helpers I would like to discuss some conventions, for example it could be neat to prefix internal functions with "_"

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -0,0 +1,416 @@
...
+    [m[8], m[9], m[10], m[11]],
+    [m[12], m[13], m[14], m[15]],
+  ] as AffineMatrix;
+}
+
+function maybeFlattenMatrix(
+  matrix: AffineMatrix | AffineMatrixFlat
+): AffineMatrixFlat {
+  'worklet';
+  return isAffineMatrix(matrix) ? flatten(matrix) : matrix;
```
Latropos, line 70:
```suggestion
  return isAffineMatrix(matrix) ? flatten(matrix) : [...matrix];
```

tomekzaw (approved):
Looks good to me! ✅

PR #4547, opened 2023-06-07
Fix testing example

mccraveiro:
## Summary

This testing example was using `jest.advanceAnimationByTime`. As far as I understand this method doesn't exist on `jest`. `advanceAnimationByTime` was a API exported by the reanimated lib and [it was deprecated](https://github.com/software-mansion/react-native-reanimated/blob/2426f32095af417f1cfc1789479c3d2ee7c72e18/src/reanimated2/jestUtils.ts#L171). 

## Test plan

N/A


ariane92 (approved):

tomekzaw (approved):
Thanks for the PR!

PR #4561, opened 2023-06-12
Fix babel plugins' location

diego-antonelli:
## Summary
When applying custom folders or custom configuration for metro bundler it gives an error saying "Cannot find module "@babel/plugin-transform-shorthands-properties". Also its a good practice to rely on node to resolve the location of the package instead of just provide the module name. It might also be caused by `@babel/register` which in some cases cannot find the packages even though they are in the same node_modules.

Example: run metro with a custom folder and adding `extraNodeModules` to point to another folder outside of the current project.

## Test plan

Run metro with custom config and point react-native-reanimated as extraNodeModules to a specific folder outside of main project.


tomekzaw:
@diego-antonelli Thanks for the PR and full explanation of the motivation and changes, we really appreciate it!

I'm gonna pass this PR to @tjzel who's currently the codeowner of Reanimated Babel plugin. He's quite busy this week but once he's back I'm pretty sure he will take a look at it. Thanks again!

diego-antonelli:
> @diego-antonelli Thanks for the PR and full explanation of the motivation and changes, we really appreciate it!
> 
> I'm gonna pass this PR to @tjzel who's currently the codeowner of Reanimated Babel plugin. He's quite busy this week but I'm pretty sure he will take a look at it once he's back. Thanks again!
> 
> edit: Would be great if you could run `cd plugin && yarn generate` and push generated JS files as well.

@tomekzaw Just pushed the suggested changes

diego-antonelli:
> 

We have the following setup which fails saying the plugins were not found:

Our project source code is in C:/project for example
Then our metro and node_modules are in C:/Program Files/Product/bla/node_modules
Then we start our metro with: metro --config C:/project/metro.config.js
The above metro config has "extraNodeModules" pointing to C:/Program Files/Product/bla/node_modules, and this includes reanimated.

When metro runs, it just fails because `@babel/plugin-transform-x-x` cannot be found at C:/project, but when using require.resolve it makes the path to C:/Program Files/Product/bla/node_modules/lib as this is defined in the configs.

diego-antonelli:
@tjzel did you have the change to check?

tjzel:
Okay, I fiddled a bit with those 'outside' modules and your solution is working as expected. Thanks for making me learn more about metro and some of its use cases :)

tjzel (approved):

diego-antonelli:
> Okay, I fiddled a bit with those 'outside' modules and your solution is working as expected. Thanks for making me learn more about metro and some of its use cases :)

My pleasure @tjzel and thanks for your dedication in maintain the library!

PR #4572, opened 2023-06-14
Fix runOnJS on Web and make it asynchronous

tjzel:
## Summary

This PR is amending some changes from #4276 where we exchanged `setImmediates` for `queueMicrotask` for `_scheduneOnJS`.

`queueMicrotask` does not take any arguments, only a callback and the change of behaviour stemming from this was not considered in `runOnJS` implementation. This PR fixes this and also makes `runOnJS` actually call `_scheduleOnJS` on web, making it asynchronous and consistent with general behavior of `runOnJS` - before it was unreachable code on web since it would simply return the provided function that was called synchronously.

## Test plan

Run WebExample and see that now `runOnJS` is working properly. I used the following snippet:

```TypeScript
function foo() {
    console.log('hello');
  }

  function foz(str: string) {
    console.log(str);
  }

  runOnJS(foo)();
  runOnJS(foz)('bye');

  function bar() {
    console.log('hello again');
  }

  function baz(str: string) {
    console.log(str);
  }

  runOnUI(() => {
    runOnJS(bar)();
    runOnJS(baz)('bye again');
  })();
```

Before:

Screenshot 2023-06-14 at 13 45 59

After:
Screenshot 2023-06-14 at 13 45 11

### DIFF
--- a/src/reanimated2/js-reanimated/index.ts
+++ b/src/reanimated2/js-reanimated/index.ts
@@ -2,18 +2,22 @@ import JSReanimated from './JSReanimated';
 import { AnimatedStyle, StyleProps } from '../commonTypes';
 import { isWeb } from '../PlatformChecker';
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 let createReactDOMStyle: (style: any) => any;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 let createTransformValue: (transform: any) => any;
 
 if (isWeb()) {
   try {
     createReactDOMStyle =
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
       require('react-native-web/dist/exports/StyleSheet/compiler/createReactDOMStyle').default;
   } catch (e) {}
 
   try {
     // React Native Web 0.19+
     createTransformValue =
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
       require('react-native-web/dist/exports/StyleSheet/preprocess').createTransformValue;
   } catch (e) {}
 }
@@ -21,7 +25,9 @@ if (isWeb()) {
 const reanimatedJS = new JSReanimated();
 
 global._makeShareableClone = (c) => c;
-global._scheduleOnJS = queueMicrotask;
+global._scheduleOnJS = (func, args) => {
+  queueMicrotask(() => func(...args));
+};
 
 interface JSReanimatedComponent {
   previousStyle: StyleProps;
--- a/src/reanimated2/js-reanimated/index.ts
+++ b/src/reanimated2/js-reanimated/index.ts
@@ -26,7 +26,11 @@ const reanimatedJS = new JSReanimated();
 
 global._makeShareableClone = (c) => c;
 global._scheduleOnJS = (func, args) => {
-  queueMicrotask(() => func(...args));
+  if (args) {
+    queueMicrotask(() => func(...args));
+  } else {
+    queueMicrotask(func);
+  }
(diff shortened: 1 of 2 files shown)


src/reanimated2/threads.ts
```diff
@@ -164,13 +164,10 @@ export function runOnJS<A extends any[], R>(
...
     fun = fun.__remoteFunction;
   }
-  if (!_WORKLET) {
-    return fun;
-  }
   return (...args) => {
     _scheduleOnJS(
       fun,
-      args.length > 0 ? makeShareableCloneOnUIRecursive(args) : undefined
+      args.length > 0 ? makeShareableCloneOnUIRecursive(args) : []
```
tomekzaw, line 170: Instead of passing an empty array here, let's update typing of `_scheduleOnJS` function because `undefined` here means a fast path for remote functions without arguments:

https://github.com/software-mansion/react-native-reanimated/blob/d3b2b84df0d34338a5046f69a36dea7bf38480b7/Common/cpp/NativeModules/NativeReanimatedModule.cpp#L78-L87

src/reanimated2/js-reanimated/index.ts
```diff
@@ -2,26 +2,32 @@ import JSReanimated from './JSReanimated';
 import { AnimatedStyle, StyleProps } from '../commonTypes';
 import { isWeb } from '../PlatformChecker';
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
tomekzaw, line 5: I've also had similar warnings in VSCode but for some reason our CI doesn't complain about it. Can we modify our workflows so that it's checked?

src/reanimated2/threads.ts
```diff
@@ -164,13 +164,10 @@ export function runOnJS<A extends any[], R>(
     // reference to the original remote function in the `__remoteFunction` property.
     fun = fun.__remoteFunction;
   }
-  if (!_WORKLET) {
-    return fun;
-  }
```
piaskowyk, line 169: Did you try call `runOnJS` on React Thread on mobile device?

tjzel: Everything seems to be working fine (except for I had to wrap the bare use of console.log in RunOnJS from before)

src/reanimated2/js-reanimated/index.ts
tjzel: It does complain about it but considers it a warning, not an error. I don't think we want to consider it as an error since we have many of those.

src/reanimated2/threads.ts
tjzel: @tomekzaw Please check if I understood correctly.

tomekzaw: Yup, it's correct. Too bad `...args` causes a TypeError when `args` is `undefined`. We could also go with `...(args ?? [])` but I like your approach better.

src/reanimated2/threads.ts
tomekzaw: Hmm, `console.log` should be available out-of-the-box without a need to wrap it manually with `runOnJS`, can you make sure that the following piece of code gets called?

https://github.com/software-mansion/react-native-reanimated/blob/7b43e32d871860b9d65bbd89922e1dda67ea25b1/src/reanimated2/initializers.ts#L176-L183

tjzel: It's available - the issue was when I actually wanted to wrap it in `runOnJS`.

tomekzaw (approved):

PR #4576, opened 2023-06-15
Swizzle RCTEventDispatcher

tomekzaw:
## Summary

Reanimated needs to intercept events while still on the UI thread. To achieve this on iOS, currently we overwrite the instance of `RCTEventDispatcher` with an instance of `REAEventDispatcher` which extends `RCTEventDispatcher` but overrides `sentEvent` method. We need to overwrite event dispatcher as soon as possible, i.e. before native modules are initialized, so that  `RNGestureHandlerManager` holds an already overwritten instance (see [here](https://github.com/software-mansion/react-native-gesture-handler/blob/9a3b90df0df3e7f85bb06ae7946aea3be95f717c/ios/RNGestureHandlerManager.mm#L62)). Currently, we call `REAInitializer` in `jsExecutorFactoryForBridge` method. This PR removes some dirty initialization code and uses swizzling instead to achieve the same effect.

screenshot

## Test plan

1. Build Example app on iOS
2. Check scroll- and gesture-based examples


piaskowyk (approved):
👍

PR #4579, opened 2023-06-15
Fix ChoreographerCallback double execution

piaskowyk:
## Summary

This PR fixes an issue with long animations. On Android, it was possible to execute ChoreographerCallback twice during the same frame (after frame drop). In effect, existing animations were canceled due to this if-condition in `requestAnimationFrame` method:
```js
nativeRequestAnimationFrame((timestamp) => {
  if (lastNativeAnimationFrameTimestamp >= timestamp) {
    // Make sure we only execute the callbacks once for a given frame
    return;
  }
...
```
To fix it, I moved this check to the native code. When double execution happens I skip the execution of the JS callback, but I still need to schedule another choreographer callback.

## How to test it?
```java
final long[] lastTimestamp = {0};
new GuardedFrameCallback(context) {
  @Override
  protected void doFrameGuarded(long frameTimeNanos) {
    if (lastTimestamp[0] == frameTimeNanos) {
      Log.w("test");
    }
    onAnimationFrame(frameTimeNanos);
    lastTimestamp[0] = frameTimeNanos;
  }
}
```
### Before
- Open SVGExample
- Set a breakpoint in Android Studio at line `Log.w("test");`
- Wait some time until the breakpoint hit
- Pass breakpoint
- Animation will **stop** ❌

### After
- Open SVGExample
- Set a breakpoint in Android Studio at line `Log.w("test");`
- Wait some time until the breakpoint hit
- Pass breakpoint
- Animation will **keep going**  ✅

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -81,7 +81,7 @@ public class NodesManager implements EventDispatcherListener {
   private RCTEventEmitter mCustomEventHandler;
   private List<OnAnimationFrame> mFrameCallbacks = new ArrayList<>();
   private ConcurrentLinkedQueue<CopiedEvent> mEventQueue = new ConcurrentLinkedQueue<>();
-  public double currentFrameTimeMs;
+  private double lastFrameTimeMs;
   public Set<String> uiProps = Collections.emptySet();
   public Set<String> nativeProps = Collections.emptySet();
   private ReaCompatibility compatibility;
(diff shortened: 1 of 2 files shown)


src/reanimated2/initializers.ts
```diff
@@ -118,11 +116,6 @@ function setupRequestAnimationFrame() {
       // is added and then use it to execute all the enqueued callbacks. Once
       // the callbacks are run, we clear the array.
       nativeRequestAnimationFrame((timestamp) => {
-        if (lastNativeAnimationFrameTimestamp >= timestamp) {
```
kmagiera, line 121: I think it'd be safer to keep this check here. I think we don't want to guarantee that we will always call nativeRequestAnimationFrame only once in a given frame. The check on the native site doesn't give us that guarantee either, because if we enqueue it twice in one frame, there will be two callbacks in `mFrameCallbacks` list and will be triggered twice. Also, this check will be relevant to iOS because we also don't have that guarantee there.


The essence of this issue wasn't that we would get the callback triggered twice, but that callback scheduled with nativeRequestAnimationFrame in a given frame would be triggered with the same frame's timestamp. We still want to prevent executing this callback more than once in a given frame

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -234,25 +234,29 @@ public void runGuarded() {
...
 
   private void onAnimationFrame(long frameTimeNanos) {
-    currentFrameTimeMs = frameTimeNanos / 1000000.;
+    double currentFrameTimeMs = frameTimeNanos / 1000000.;
 
-    while (!mEventQueue.isEmpty()) {
-      CopiedEvent copiedEvent = mEventQueue.poll();
-      handleEvent(copiedEvent.getTargetTag(), copiedEvent.getEventName(), copiedEvent.getPayload());
-    }
+    if (currentFrameTimeMs != lastFrameTimeMs) {
```
kmagiera, line 239: It'd be more intuitive for me if we had `currentFrameTimeMs > lastFrameTimeMs` rather than just inequality check. We want to only do things if we are processing a new frame (that's later than the last processed frame). Would be also great to add a short comment about this check to make it cleaner.

kmagiera (approved):
Looks good but please take a look at my comments prior to merging.

src/reanimated2/initializers.ts
piaskowyk: Resolved in https://github.com/software-mansion/react-native-reanimated/pull/4579/commits/579335b94c7b3773cf56c58d62175938b40bde5e

android/src/main/java/com/swmansion/reanimated/NodesManager.java
piaskowyk: Resolved in: https://github.com/software-mansion/react-native-reanimated/pull/4579/commits/e67d9e1f2de8ad4a5acbcbd7da782911649f9d50

tlow92:
I'm not 100% sure if this really is related but I want to inform you nonetheless: 
I faced this issue on latest stable 3.3.0 and applied your fix. It seemed to have fixed my initial issue of canceled animations.
But I also use gorhom/bottom-sheet (uses reanimated) and when the animation for the sheet is opened during an already active animation running it seems to cancel the other animations.

After some investigating I think the behaviour is like this:
* So if sheet opening is triggered before other animations started everything is fine.
* If sheet is opened after other animations finished it's also fine.
* If sheet is opened after other animations started but not finished yet they will be canceled.

Unfortunately I can not provide a reproducible demo right now due to lack of time and my project is very complex, but maybe it helps in case there are other hints to a possible issue.

piaskowyk:
Hey @tlow92 👋 Thanks for your report! It seems like a different problem, could you open a new issue with it?

PR #4586, opened 2023-06-17
Implement `dispatchCommand` on Paper

tomekzaw:
## Summary

Previously, `dispatchCommand` was available only on Fabric. This PR implements `dispatchCommand` for Android and iOS (Paper) as well as exports the function in `react-native-reanimated.d.ts`.

Android:

https://github.com/software-mansion/react-native-reanimated/assets/20516055/7f86b832-0f5d-4e56-ac99-22abea8494a5

iOS:

https://github.com/software-mansion/react-native-reanimated/assets/20516055/ee7461bd-9994-455a-b20f-f7878abbc3cb

## Test plan

DispatchCommandExample.tsx


ios/native/NativeMethods.mm
```diff
@@ -42,6 +43,21 @@ void scrollTo(int scrollViewTag, RCTUIManager *uiManager, double x, double y, bo
...
+      [NSInvocation invocationWithMethodSignature:[uiManager methodSignatureForSelector:privateMethodSelector]];
+  [invocation setSelector:privateMethodSelector];
+  [invocation setTarget:uiManager];
+  [invocation setArgument:&viewTag atIndex:2];
+  [invocation setArgument:&commandID atIndex:3];
+  [invocation setArgument:&commandArgs atIndex:4];
+  RCTUnsafeExecuteOnUIManagerQueueSync(^{
+    [invocation invoke];
+  });
+}
```
piaskowyk, line 59: Maybe we could extend category instead of this like here - https://github.com/software-mansion/react-native-reanimated/blob/main/ios/native/NativeProxy.mm#L35-L37

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -53,6 +53,16 @@ public void scrollTo(int viewTag, double x, double y, boolean animated) {
     NativeMethodsHelper.scrollTo(view, x, y, animated);
   }
 
+  public void dispatchCommand(int viewTag, String commandId, ReadableArray commandArgs) {
+    mContext.runOnNativeModulesQueueThread(
+        new GuardedRunnable(mContext.getExceptionHandler()) {
+          @Override
+          public void runGuarded() {
+            mUIManager.dispatchCommand(viewTag, commandId, commandArgs);
```
tomekzaw, line 61: Technically, this will be run on the `mqt_native_module` thread so there may be a slight delay if we don't synchronize it properly with the UI thread. Instead, maybe we should push commands to the queue here and flush the queue in `performOperations` which is already synchronized with the UI thread?

ios/native/NativeProxy.mm
```diff
@@ -115,6 +121,19 @@ static CFTimeInterval calculateTimestampWithSlowAnimations(CFTimeInterval curren
...
   };
+
+  auto dispatchCommandFunction =
+      [uiManager](
+          jsi::Runtime &rt, const int tag, const jsi::Value &commandNameValue, const jsi::Value &argsValue) -> void {
+    NSNumber *viewTag = [NSNumber numberWithInt:tag];
+    NSString *commandID = [NSString stringWithCString:commandNameValue.asString(rt).utf8(rt).c_str()
+                                             encoding:[NSString defaultCStringEncoding]];
+    NSArray *commandArgs = convertJSIArrayToNSArray(rt, argsValue.asObject(rt).asArray(rt));
+    RCTUnsafeExecuteOnUIManagerQueueSync(^{
```
tomekzaw, line 132: Maybe use `RCTExecuteOnUIManagerQueue` instead

piaskowyk (commented):
imo `dispatchCommand` method by definition should be an asynchronous operation, so on iOS we should use `RCTExecuteOnUIManagerQueue` instead of `RCTUnsafeExecuteOnUIManagerQueueSync` I think. 

I just wondering if we should add even short information to the documentation about the possibility of using the `dispatchCommand` method from UI Thread. What do tho think?

The rest of PR looks good 👍

piaskowyk (approved):

Issue #4585, opened 2023-06-17
Changing its height while ScrollView is scrolling horizontial doesn't work porperly


zerosrat:
### Description

The scene: when scrolling horizontal ScrollView, it changes its height simultaneously.
I use `Animated` of react native lib to implement the correct effect which is illustrated as below.

![Simulator Screen Recording - iPhone 14 Pro - 2023-06-17 at 22 44 19](https://github.com/software-mansion/react-native-reanimated/assets/8179987/254d2ab7-9282-4ac1-be30-883e70e34c13)

When I use `react-native-reanimated`, Android platform works fine with same code while iOS doesn't. The wrong effect is illustrated as below.

![Simulator Screen Recording - iPhone 14 Pro - 2023-06-17 at 22 41 16](https://github.com/software-mansion/react-native-reanimated/assets/8179987/01556da9-6ae9-42cf-a25e-bc41b130f479)

### Steps to reproduce

```tsx
import React from 'react';
import {Dimensions, Text, View} from 'react-native';
import Animated, {
  useSharedValue,
  useAnimatedScrollHandler,
  useAnimatedStyle,
} from 'react-native-reanimated';

const winWidth = Dimensions.get('window').width;

export const Swiper = () => {
  const scrollOffset = useSharedValue(0);
  const animatedStyle = useAnimatedStyle(() => {
    return {
      height: (scrollOffset.value / winWidth) * 200 + 200,
    };
  });

  const scrollHandler = useAnimatedScrollHandler({
    onScroll: event => {
      scrollOffset.value = event.contentOffset.x;
    },
  });

  return (
    <Animated.View style={animatedStyle}>
      <Animated.ScrollView
        horizontal
        pagingEnabled
        onScroll={scrollHandler}
        showsHorizontalScrollIndicator={false}
        bounces={false}
        style-={{flex: 1}}>
        <View style={{width: winWidth, height: 200, backgroundColor: 'blue'}}>
          <Text>1</Text>
        </View>
        <View style={{width: winWidth, height: 400, backgroundColor: 'yellow'}}>
          <Text>2</Text>
        </View>
      </Animated.ScrollView>
    </Animated.View>
  );
};
```

### Snack or a link to a repository

https://github.com/zerosrat/react-native-interactive/blob/feature/reanimated/src/pages/interactive-screen/components/swiper.tsx

### Reanimated version

3.3.0

### React Native version

0.71.11

### Platforms

iOS

### JavaScript runtime

[...]

PR #4589, opened 2023-06-19
Set default scrollEventThrottle

piaskowyk:
## Summary

The inspiration for this changes are two issues:
- https://github.com/software-mansion/react-native-reanimated/issues/4585#issuecomment-1595921845
- https://github.com/software-mansion/react-native-reanimated/pull/4416/commits/4137d1a7d4ddba180c54582c2e1036a285c0ba3e

Users expect continuous scroll events. But the default value of [scrollEventThrottle](https://reactnative.dev/docs/scrollview#scrolleventthrottle-ios) is 0, which results in the scroll event being sent only once each time the view is scrolled. Which is easy to forget about it and in effect, it seems like a Reanimated issue. I know that this is a change of default ScrollView behavior but I believe that it would be less confusing for the users.

### DIFF
--- a/src/reanimated2/component/ScrollView.tsx
+++ b/src/reanimated2/component/ScrollView.tsx
@@ -26,6 +26,13 @@ const AnimatedScrollView: AnimatedScrollViewFC = forwardRef(
         scrollViewOffset
       );
     }
+
+    if (!restProps.scrollEventThrottle) {
+      // Set default scrollEventThrottle to 16, because user expects
+      // to have continuous scroll events
+      restProps.scrollEventThrottle = 16;
+    }
+
     return <AnimatedScrollViewComponent ref={aref} {...restProps} />;
   }
 );
--- a/src/reanimated2/component/ScrollView.tsx
+++ b/src/reanimated2/component/ScrollView.tsx
@@ -28,9 +28,9 @@ const AnimatedScrollView: AnimatedScrollViewFC = forwardRef(
     }
 
     if (!restProps.scrollEventThrottle) {
-      // Set default scrollEventThrottle to 16, because user expects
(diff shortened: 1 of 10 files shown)


src/reanimated2/component/ScrollView.tsx
```diff
@@ -26,6 +26,13 @@ const AnimatedScrollView: AnimatedScrollViewFC = forwardRef(
         scrollViewOffset
       );
     }
+
+    if (!restProps.scrollEventThrottle) {
+      // Set default scrollEventThrottle to 16, because user expects
+      // to have continuous scroll events
+      restProps.scrollEventThrottle = 16;
```
tomekzaw, line 33: What do you think about setting 8 instead of 16 here? I don't think it would make a difference for 60 fps screens but it would work out-of-the-box for 120 fps devices.
```suggestion
      restProps.scrollEventThrottle = 8;
```

tomekzaw (approved):
+1 for this change

tomekzaw:
BTW should we also remove `scrollEventThrottle={16}` in our examples?

piaskowyk:
Thanks! I applied your suggestions 👍

src/reanimated2/component/ScrollView.tsx
```diff
@@ -26,6 +26,13 @@ const AnimatedScrollView: AnimatedScrollViewFC = forwardRef(
         scrollViewOffset
       );
     }
+
+    if (!restProps.scrollEventThrottle) {
```
zerosrat, line 30: I wonder if it should check `typeof onScroll === 'function'` before setting the default value. 
Cause if user doesn't set `onScoll` prop, it is not necessary to set `scrollEventThrottle` prop
@piaskowyk

PR #4591, opened 2023-06-20
Fix "withDecay clamps position to the wrong boundary"

### LINKED ISSUE #4508: withDecay clamps position to the wrong boundary
### Description
I've noticed that `withDecay` works incorrectly while is used with `clamp` parameter. I use `withDecay` function for smooth deceleration of the dragged object in my app and expect the object to be moved back to the `clamp` range if its position exceeds boundaries specified in the `clamp` parameter.

When the object's position is greater than the upper bound, its position should be smoothly changed to equal the upper bound (the same applies to the lower bound of the `clamp` parameter). The issue can be noticed when the velocity of the object is lower than 0 while object's position is greater than the upper bound (or the velocity is greater than 0 when the object's position is lower than the lower bound). In such a case, the object is transitioned to the opposite bound specified in the `clamp` parameter (see recordings below).

#### When velocity is lower than 0 and object's position is lower than the lower bound
Everything works correctly when the object is pulled outside the specified boundaries

https://github.com/software-mansion/react-native-reanimated/assets/52978053/6ca0732c-f04c-49db-8bd7-43cbc83ea3a0

#### When velocity is greater than 0 and object's position is lower than the lower bound
The object is translated to the upper bound of the `clamp` value if it was outside the boundaries but it was pulled towards the boundaries

https://github.com/software-mansion/react-native-reanimated/assets/52978053/65f151ad-9404-432d-8b8a-ab1e73977cf1

### Additional information
I also tested if the issue was present in previous versions of Reanimated. It appears that the bug wasn't present in the `2.12.0` release (you can try it out on Snack with `2.12.0` version of Reanimated which is the latest available version in Expo [here](https://snack.expo.dev/@matipl01/reanimated-issue))

### Steps to reproduce

1. Run `yarn` to install necessary libraries
2. Run the application on a simulator or on a real device
3. Drag the blue square outside the screen (keep holding the square)
4. Move the square in the opposite direction (towards the center of the screen) and release it

https://github.com/MatiPl01/reanimated-issues

### Reanimated version

3.2.0

### React Native version

0.71.8

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

Latropos:
## Summary

Closes #4508 

### Bug description
<table>
<tr><th>BEFORE</th><th>AFTER</th></tr>
<tr><td width="50%">Animation looks good if and only if the animated square has no positive velocity (velocity towards center of screen). In other case it moves much too fast and clamp to the opposite position with no bounces</td><td>If square has some positive velocity it moves visibly faster. Square always bounces when approaching clamp limit (with rubber band effect on).</td></tr>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/3035ad50-b108-49f0-8714-791c4ae706fe

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/d6cb43d5-3e9c-4ee5-aee2-a5a2bb4394b3

</td></tr>
</table>

### Description of code change

I've removed a condition updating the derivative only when square is within predefined boundaries - since we want to keep its springified movement even if it is overshooting. Especially that we used to end animation almost immediately after exceeding limits.

Since after this change the `derivative` is always modified we should no longer compare it to zero, but use some epsilon instead.

## Test plan

I've tested it on this example: https://github.com/MatiPl01/reanimated-issues, linked in the issue.


piaskowyk (dismissed):

src/reanimated2/animation/decay.ts
```diff
@@ -38,6 +38,7 @@ export interface InnerDecayAnimation
   current: number;
 }
 
+const EPSILON = 0.01;
```
tomekzaw, line 41: Isn't it too much?

src/reanimated2/animation/decay.ts
```diff
@@ -76,14 +77,11 @@ export function withDecay(
...
-        if (current < config.clamp![0] || current > config.clamp![1]) {
-          derivative = current - config.clamp![clampIndex];
-        }
+        const derivative = current - config.clamp![clampIndex];
 
-        if (derivative !== 0) {
+        if (Math.abs(derivative) > EPSILON) {
           animation.springActive = true;
-        } else if (derivative === 0 && animation.springActive) {
+        } else if (Math.abs(derivative) <= EPSILON && animation.springActive) {
```
tomekzaw, line 84: If the first conditional fails (i.e. `Math.abs(derivative) > EPSILON` is false), this means that `Math.abs(derivative) <= EPSILON` so there's no need to check it again.
```suggestion
        } else if (animation.springActive) {
```

src/reanimated2/animation/decay.ts
piaskowyk: @Latropos I know that you discussed this with Tomek offline, but let's write the conclusion in a comment to keep the informations in one place, please 🙏

Latropos: Sure! We want to have similar precision for decay and withSpring. Also we have tested, that it works well with `const EPSILON = 0.01;` and that we animate static spring for long time if epsilon is smaller.

piaskowyk (approved):
As we discussed it last time, let's merge it, and add more "real life physic" behaviour in another PR.

PR #4593, opened 2023-06-20
Add missing null checks

tomekzaw:
## Summary

Currently, TypeScript guards against returning an object which contains `null` from `useAnimatedStyle`:

However, if we ignore this error, we will get the following runtime error:

This PR adds missing null checks to avoid runtime error in this scenario.

Co-authored-by: @kmagiera

## Test plan

<details>
<summary>App.tsx</summary>

```tsx
import {StyleSheet, View} from 'react-native';

import Animated, {useAnimatedStyle} from 'react-native-reanimated';

import React from 'react';

    return {
      width: 100,
      height: 100,
      backgroundColor: 'red',
      display: null,
    };
  });

  return (
      <Animated.View style={animatedStyle} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

</details>

### DIFF
--- a/src/reanimated2/hook/useAnimatedStyle.ts
+++ b/src/reanimated2/hook/useAnimatedStyle.ts
@@ -373,13 +373,14 @@ function checkSharedValueUsage(
     for (const element of prop) {
       checkSharedValueUsage(element, currentKey);
     }
-  } else if (typeof prop === 'object' && prop.value === undefined) {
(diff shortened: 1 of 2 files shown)


piaskowyk (approved):

tomekzaw:
> Why not `prop !== null` though?

Could be as well.

PR #4596, opened 2023-06-20
Allow setting ndk path/version

janicduplessis:
## Summary

This allows changing the ndk version used to build this.

## Test plan

Tested by using a different ndk version in an app.


tomekzaw (approved):
@janicduplessis Thanks for this PR. Looks good to me, what do you think @piaskowyk?

piaskowyk (approved):

PR #4601, opened 2023-06-21
Fix update props flush for multiple view descriptors

tomekzaw:
## Summary

This PR fixes a regression introduced in #4503.

If there was more than one view descriptor, `queueMicrotask(this.flush)` would never be called.

## Test plan

ChessboardExample.tsx

### DIFF
--- a/src/reanimated2/UpdateProps.ts
+++ b/src/reanimated2/UpdateProps.ts
@@ -100,10 +100,10 @@ const createUpdatePropsManager = global._IS_FABRIC
               shadowNodeWrapper: viewDescriptor.shadowNodeWrapper,
               updates,
             });
+            if (operations.length === 1) {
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):

efstathiosntonas:
@tomekzaw hi, I've applied this over `3.4.0-nightly-20230616-abec3fe10` and on iOS it throws this in every screen that has reanimated components. Will try `3.3.0` and let you know. 

`Assertion failed: (sizeof...(Args) == count), function getArgsForFunction, file JsiUtils.h, line 100.` on `JSIUtils.h`

tomekzaw:
@efstathiosntonas Thanks for reporting this, not really sure what could be the issue here but I'm glad that it finally worked for you! 😄

PR #3653, opened 2022-10-06
type all style props with `AnimateStyle`

johankasperi:
## Description

Style type properties (properties that extends `StyleProp<ViewStyle>`) can be defined with other property names than "style". For example `contentContainerStyle` in `FlatList`. Current implementation of `AnimateProps` defines all style type properties as `AnimatedNode<StyleProp<...>>` except the regular `style` property. The regular `style` property is being typed as `StyleProp<AnimateStyle<StyleProp<ViewStyle>>`. I think the type definition for all style type properties should act similarly.

## Changes

- Extend `AnimateStyle` so all properties ending with "Style" (for example "textStyle" and "contentContainerStyle") is being typed as `StyleProp<AnimateStyle<StyleProp<ViewStyle>>`.

## Test code and steps to reproduce

Added a test in "FabricExample/react-native-reanimated-tests.tsx" named "CreateAnimatedFlatListTest3" illustrating the problem and that shows that the proposed change fixes it.

## Checklist

- [ ] Included code example that can be used to test this change
- [x] Updated TS types
- [x] Added TS types tests
- [ ] Added unit / integration tests
- [ ] Updated documentation
- [x] Ensured that CI passes

PR #4610, opened 2023-06-22
Unify types of all 'style' related props in AnimateProps

tjzel:
## Summary

This pull requests is a simple rewrite of #3653 since we will no longer have the .d.ts file.

I've also fixed `react-native-reanimated-tests.tsx` so it would be able to resolve its imports properly.

## Test plan

`yarn type:check`, `yarn type:check-api`, also a look into new tests in `react-native-reanimated-tests.tsx`.

---

Co-authored by @johankasperi


__typetests__/react-native-reanimated-tests.tsx
```diff
@@ -57,10 +57,9 @@ import Animated, {
   StyleLayoutAnimation,
   Animation,
   // eslint-disable-next-line import/no-unresolved
-} from 'react-native-reanimated';
+} from '../src';
```
tomekzaw, line 60: I'm not a huge fan of this change, `import { ... } from 'react-native-reanimated';` also checks if there's a module called `react-native-reanimated`. Why is it necessary?

__typetests__/react-native-reanimated-tests.tsx
```diff
@@ -57,10 +57,9 @@ import Animated, {
...
 import {
   dispatchCommand,
-  getTag,
```
tomekzaw, line 63: This change is necessary because I've removed `getTag` function in #4361.

johankasperi (approved):
@tjzel Looks good! And thanks for picking up my original PR and transform it so it matches the relevant files. I definitely feel like I have gotten enough credit 😄 Thanks again!

__typetests__/react-native-reanimated-tests.tsx
tjzel: In my environment TypeScript couldn't resolve this import properly and inferred all of the imports as `any`, making the tests kind-of useless. I'd be more than happy if you could check if that's the case for you too.

While this file can check if there's a module called `react-native-reanimated`, this file isn't considered anywhere in our CIs etc., hence even if there was something wrong there we wouldn't know until we've had it manually checked. I think this file should focus on types only - and if we want, we could do much more reliable test that checks if module called `react-native-reanimated` exists.

tomekzaw (approved):

PR #4611, opened 2023-06-22
Remove unnecessary null-checks for mCustomEventNamesResolver

piaskowyk:
## Summary

The goal of this PR was to remove unused variables form `private void handleEvent(Event event)`:
```diff
private void handleEvent(Event event) {
-  // If the event has a different name in native, convert it to it's JS name.
-  String eventName = mCustomEventNamesResolver.resolveCustomEventName(event.getEventName());
-  int viewTag = event.getViewTag();
  if (mCustomEventHandler != null) {
    event.dispatch(mCustomEventHandler);
  }
}
```
But then I realized that we can avoid some unnecessary null-checks for mCustomEventHandler. The mCustomEventHandler could be a null value before initialization by void registerEventHandler(RCTEventEmitter handler) method.

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/nativeProxy/EventHandlerMock.java
+++ b/android/src/main/java/com/swmansion/reanimated/nativeProxy/EventHandlerMock.java
@@ -1,0 +1,18 @@
+package com.swmansion.reanimated.nativeProxy;
+
+import androidx.annotation.Nullable;
+import com.facebook.react.bridge.WritableArray;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.uimanager.events.RCTEventEmitter;
+
+public class EventHandlerMock implements RCTEventEmitter {
+  @Override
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/swmansion/reanimated/nativeProxy/EventHandlerMock.java
```diff
@@ -0,0 +1,18 @@
+package com.swmansion.reanimated.nativeProxy;
+
+import androidx.annotation.Nullable;
+import com.facebook.react.bridge.WritableArray;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.uimanager.events.RCTEventEmitter;
+
+public class EventHandlerMock implements RCTEventEmitter {
+  @Override
+  public void receiveEvent(int i, String s, @Nullable WritableMap writableMap) {
```
tomekzaw, line 10: Can we provide "official" argument names here (and below as well)?

tomekzaw (approved):

PR #4612, opened 2023-06-22
Unify naming of `NativeReanimatedModule` and `REAModule` instances

tomekzaw:
## Summary

This PR aims to rename `module` variable which holds an instance of `NativeReanimatedModule` because `module` is a keyword since C++20. From now on, we will use the following naming convention:

* `std::shared_ptr<NativeReanimatedModule> nativeReanimatedModule`
* `std::weak_ptr<NativeReanimatedModule> weakNativeReanimatedModule`
* `REAModule* reaModule` (iOS)


piaskowyk (approved):
Looks good to me. If CI pass, we can merge it 🚀

PR #4617, opened 2023-06-23
Fix passing remote functions as arguments to `runOnJS`

### LINKED ISSUE #4613: [Reanimated 3.3.0] runOnJs crash - "Object is not a function"
### Description

I have updated my project from 2.17.0 to 3.3.0 and I am running into an issue in the following scenario:

```typescript
const finishTransition = useCallback(
    (
      resolve: (value: void | PromiseLike<void>) => void,
    ) => {
       // ... some logic, not important to understand the issue
      resolve();
    },
    [onFinishedTransitioning]
  );

const handlePress = () => {
    new Promise(resolve => {
      scale.value = withSpring(2, {}, () => {
        runOnJS(finishTransition)(resolve);
      });
    }).then(() => {
      // ...
    });
  };
```

Calling the resolve within finishTransition using runOnJS was working fine in 2.17.0. In 3.3.0 is is throwing `TypeError: Object is not a function`. I had a closer look and the resolve parameter of the finishTransition function is logged as "{"__remoteFunction": {}}". There seems to be an incorrect translation happening.

### Steps to reproduce

1. Follow the description to create a code example similar to how I defined it. It seems to be related to passing a promise resolve function into the runOnJs function.

https://github.com/AlexanderEggers/ReanimatedIssue

### Reanimated version

3.3.0

### React Native version

0.71.11

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

tomekzaw:
## Summary

Fixes #4613.

Work in progress, need to fix types.

## Test plan

<details>
<summary>App.tsx</summary>

```tsx
import {Button, StyleSheet, View} from 'react-native';
import {runOnJS, runOnUI} from 'react-native-reanimated';

import React from 'react';

  const resolve = () => {
    console.log('OK');
  };

  const finishTransition = resolve => {
    resolve();
  };

  const handlePress = () => {
    runOnUI(() => {
      runOnJS(finishTransition)(resolve);
    })();
  };

  return (
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

</details>


src/reanimated2/shareables.ts
```diff
@@ -213,6 +213,11 @@ export function makeShareableCloneOnUIRecursive<T>(value: T): ShareableRef<T> {
   function cloneRecursive<T>(value: T): ShareableRef<T> {
     const type = typeof value;
     if ((type === 'object' || type === 'function') && value !== null) {
+      // @ts-ignore trust me
```
tjzel, line 216: Does not seem very trustoworthy 🤔

src/reanimated2/shareables.ts
```diff
@@ -237,18 +246,22 @@ export function makeShareableCloneOnUIRecursive<T>(value: T): ShareableRef<T> {
     return value;
   }
   function cloneRecursive<T>(value: T): ShareableRef<T> {
-    const type = typeof value;
-    if ((type === 'object' || type === 'function') && value !== null) {
-      let toAdapt: any;
+    if (
+      (typeof value === 'object' && value !== null) ||
+      typeof value === 'function'
```
tomekzaw, line 251: Let's restore original logic here:
```suggestion
    const type = typeof value;
    if ((type === 'object' || type === 'function') && value !== null) {
```

src/reanimated2/shareables.ts
```diff
@@ -238,17 +247,20 @@ export function makeShareableCloneOnUIRecursive<T>(value: T): ShareableRef<T> {
   }
   function cloneRecursive<T>(value: T): ShareableRef<T> {
     const type = typeof value;
-    if ((type === 'object' || type === 'function') && value !== null) {
-      let toAdapt: any;
+    if ((type === 'object' && value !== null) || type === 'function') {
+    ) {
```
tomekzaw, line 251:
```suggestion
```

tjzel (approved):
image

AlexanderEggers:
@tomekzaw @tjzel I patched these locally in my project and I am running into quite a bit of "Tried to synchronously call a non-worklet function on the UI thread." errors. That wasn't the case with the previous changes that were part of this PR.

This is my patch based on 3.4.2:

```patch
diff --git a/node_modules/react-native-reanimated/src/reanimated2/commonTypes.ts b/node_modules/react-native-reanimated/src/reanimated2/commonTypes.ts
index 5ab6085..07770a5 100644
--- a/node_modules/react-native-reanimated/src/reanimated2/commonTypes.ts
+++ b/node_modules/react-native-reanimated/src/reanimated2/commonTypes.ts
@@ -55,6 +55,12 @@ export type ShareableRef<T> = {
   __hostObjectShareableJSRef: T;
 };
 
+// In case of objects with depth or arrays of objects or arrays of arrays etc.
+// we add this utility type that makes it a SharaebleRef of the outermost type.
+export type FlatShareableRef<T> = T extends ShareableRef<infer U>
+  ? ShareableRef<U>
+  : ShareableRef<T>;
+
 export type ShareableSyncDataHolderRef<T> = {
   __hostObjectShareableJSRefSyncDataHolder: T;
 };
diff --git a/node_modules/react-native-reanimated/src/reanimated2/globals.d.ts b/node_modules/react-native-reanimated/src/reanimated2/globals.d.ts
index f3259e4..343695a 100644
--- a/node_modules/react-native-reanimated/src/reanimated2/globals.d.ts
+++ b/node_modules/react-native-reanimated/src/reanimated2/globals.d.ts
@@ -8,6 +8,7 @@ import type {
   ShareableSyncDataHolderRef,
   ShadowNodeWrapper,
   ComplexWorkletFunction,
+  FlatShareableRef,
 } from './commonTypes';
 import type { AnimatedStyle } from './helperTypes';
 import type { FrameCallbackRegistryUI } from './frameCallback/FrameCallbackRegistryUI';
@@ -35,7 +36,7 @@ declare global {
   ) => void;
   var _notifyAboutEnd: (tag: number, removeView: boolean) => void;
   var _setGestureState: (handlerTag: number, newState: number) => void;
-  var _makeShareableClone: <T>(value: T) => ShareableRef<T>;
+  var _makeShareableClone: <T>(value: T) => FlatShareableRef<T>;
   var _updateDataSynchronously: (
     dataHolder: ShareableSyncDataHolderRef<any>,
     data: ShareableRef<any>
diff --git a/node_modules/react-native-reanimated/src/reanimated2/shareables.ts b/node_modules/react-native-reanimated/src/reanimated2/shareables.ts
index 03575ab..e03f727 100644
--- a/node_modules/react-native-reanimated/src/reanimated2/shareables.ts
+++ b/node_modules/react-native-reanimated/src/reanimated2/shareables.ts
@@ -1,5 +1,9 @@
 import NativeReanimatedModule from './NativeReanimated';
-import type { ShareableRef, WorkletFunction } from './commonTypes';
+import type {
+  FlatShareableRef,
+  ShareableRef,
+  WorkletFunction,
+} from './commonTypes';
 import { shouldBeUseWeb } from './PlatformChecker';
 import { registerWorkletStackDetails } from './errors';
 import { jsVersion } from './platform-specific/jsVersion';
@@ -20,7 +24,7 @@ const _shareableFlag = Symbol('shareable flag');
 
 const MAGIC_KEY = 'REANIMATED_MAGIC_KEY';
 
-function isHostObject(value: NonNullable<object>): boolean {
+function isHostObject(value: NonNullable<object>) {
   // We could use JSI to determine whether an object is a host object, however
   // the below workaround works well and is way faster than an additional JSI call.
   // We use the fact that host objects have broken implementation of `hasOwnProperty`
@@ -247,28 +251,44 @@ function getWorkletCode(value: WorkletFunction) {
   return code;
 }
 
-export function makeShareableCloneOnUIRecursive<T>(value: T): ShareableRef<T> {
+type RemoteFunction<T> = {
+  __remoteFunction: FlatShareableRef<T>;
+};
+
+function isRemoteFunction<T>(value: object): value is RemoteFunction<T> {
+  return '__remoteFunction' in value;
+}
+
+export function makeShareableCloneOnUIRecursive<T>(
+  value: T
+): FlatShareableRef<T> {
   'worklet';
   if (USE_STUB_IMPLEMENTATION) {
     // @ts-ignore web is an interesting place where we don't run a secondary VM on the UI thread
     // see more details in the comment where USE_STUB_IMPLEMENTATION is defined.
     return value;
   }
-  function cloneRecursive<T>(value: T): ShareableRef<T> {
-    const type = typeof value;
-    if ((type === 'object' || type === 'function') && value !== null) {
-      let toAdapt: any;
+  function cloneRecursive<T>(value: T): FlatShareableRef<T> {
+    if (
+      (typeof value === 'object' && value !== null) ||
+      typeof value === 'function'
+    ) {
+      if (isRemoteFunction<T>(value)) {
+        return value.__remoteFunction;
+      }
+      if (isHostObject(value)) {
+        return value as FlatShareableRef<T>;
+      }
       if (Array.isArray(value)) {
-        toAdapt = value.map((element) => cloneRecursive(element));
-      } else if (value !== undefined) {
-        toAdapt = {};
-        for (const [key, element] of Object.entries(
-          value as Record<string, unknown>
-        )) {
-          toAdapt[key] = cloneRecursive(element);
-        }
+        return _makeShareableClone(
+          value.map(cloneRecursive)
+        ) as FlatShareableRef<T>;
+      }
+      const toAdapt: Record<string, FlatShareableRef<T>> = {};
+      for (const [key, element] of Object.entries(value)) {
+        toAdapt[key] = cloneRecursive<T>(element);
       }
-      return _makeShareableClone(toAdapt);
+      return _makeShareableClone(toAdapt) as FlatShareableRef<T>;
     }
     return _makeShareableClone(value);
   }
```

tomekzaw:
@AlexanderEggers I think I know what's the issue, can you please mark `isHostObject` with `'worklet';` directive?

AlexanderEggers:
@tomekzaw I tried that already but made no difference.

tomekzaw:
@AlexanderEggers Have you also tried marking `isRemoteFunction` as worklet?

AlexanderEggers:
@tomekzaw Adding `'worklet';` to `isRemoteFunction` and `isHostObject` solved the error for me. 
I have one remaining issue around worklets where any UI touches are blocked after using worklets and this latest above patch. Only a restart of my app is unblocking UI touches again. It could be related to a different issue (but it is new with reanimated v3) - I will open a new issue for that when I have more details unless you have an idea I can test for you?

src/reanimated2/shareables.ts
```diff
@@ -20,7 +24,7 @@ const _shareableFlag = Symbol('shareable flag');
 
 const MAGIC_KEY = 'REANIMATED_MAGIC_KEY';
 
-function isHostObject(value: NonNullable<object>): boolean {
+function isHostObject(value: NonNullable<object>) {
```
tomekzaw, line 27:
```suggestion
function isHostObject(value: NonNullable<object>) {
  'worklet';
```

src/reanimated2/shareables.ts
```diff
@@ -247,28 +251,44 @@ function getWorkletCode(value: WorkletFunction) {
   return code;
 }
 
-export function makeShareableCloneOnUIRecursive<T>(value: T): ShareableRef<T> {
+type RemoteFunction<T> = {
+  __remoteFunction: FlatShareableRef<T>;
+};
+
+function isRemoteFunction<T>(value: object): value is RemoteFunction<T> {
```
tomekzaw, line 258:
```suggestion
function isRemoteFunction<T>(value: object): value is RemoteFunction<T> {
  'worklet';
```

tjzel:
> @AlexanderEggers I think I know what's the issue, can you please mark `isHostObject` with `'worklet';` directive?
> 
> @tjzel Would be nice to include function name in the error message, for instance "Tried to synchronously call a non-worklet function `isHostObject` on the UI thread" (if possible, in a separate PR). Also, what do you think about describing this error on the brand new Troubleshooting page?

🤔 

good idea

AlexanderEggers:
@tomekzaw Do you have an ETA when these would be available in a release?

AlexanderEggers:
@tomekzaw I did a bit more testing using this implementation and unfortunately these changes are crashing in release builds (on Android and iOS) for some reasons.

`RCTFatalException: Unhandled JS Exception: undefined is not a function`

tjzel:
@AlexanderEggers I didn't get any crash in the reproduction you provided [here](https://github.com/software-mansion/react-native-reanimated/issues/4613) but I still have a clue what might be the offender. Could you check it with the changes that I now pushed to this branch?

AlexanderEggers:
@tjzel Thanks, I had a quick test and that seemed to have fixed the crash for me. I am just curious, do you have an idea why my case would crash because of that incorrect order?

tjzel:
Yes, it's because of how host objects works, or rather the `in` operator for them - it always returns true.

In debug we have this special object wrapper called `RemoteFunction` (terrible name imho) that has the actual host object in its `__remoteFunction` property. Now, if we get a host object and we check first, if it has this property, we will always get true - but this property doesn't exist and accessing it and calling it will result in a crash. 

Therefore we must firstly check if it's a host object and then if it's a RemoteFunction - we overlooked this in the first place.

AlexanderEggers:
oh, that makes sense. Thanks for taking the time to explain that

tomekzaw:
Thanks @tjzel for landing this PR.

PR #4511, opened 2023-05-31
Inject `_WORKLET` to RN runtime

tomekzaw:
## Summary

Fixes #4506 and fixes #4507.

PR #4618, opened 2023-06-23
Transform '_WORKLET' in worklets to 'true'.

tjzel:
## Summary

The goal of this PR is to add functionality of changing UI runtime checks (aka `_WORKLET`) to `true` in worklets, an extension to/alternate solution of #4511.

Notice that in the code I'm not traversing the worklet's tree to get `_WORKLET` nodes to change them to `true` - I'm just replacing substrings in worklet's code. 

The reason of that is the fact we don't have trees for worklets - only a given node and its code. We could change them into ASTs and then replace certain nodes, but this seems unnecessary according to my research - we would need to do following steps:

1. Clone the node for future worklet (we already do that).
2. Generate code for the node (we already do that).
3. Parse generated code to AST.
4. Traverse AST and replace certain nodes.
5. Generate code from AST.

Those could be some time-consuming extra steps that don't seem to be really needed, since we expect every `_WORKLET` in worklet's code to be a part of a boolean expression.

Things I also did in this PR:

- Small refactor of plugin tests - moved comments into `describe` blocks (it just makes much more sense that way).
- Moved commonly used unions of types from plugin into their own type alias.

TODO:
- [ ] Transform `_WORKLET` in non-worklets to `false` (probably, since I think I remember that was the chief issue). 

## Test plan

I've added a whole new test suite for this functionality in plugins tests. Also ran Example app to see if it works properly.


tjzel:
RIP in peace

PR #4619, opened 2023-06-23
Swizzle RCTUIManager

piaskowyk:
## Summary

Recently, we began rewriting the iOS initialization process. Tomek moved the injection of JSI bindings to `installTurboModule` (check out https://github.com/software-mansion/react-native-reanimated/pull/4523) and then used swizzling to eliminate the `REAEventDispatcher` (see https://github.com/software-mansion/react-native-reanimated/pull/4576). More context here: https://github.com/expo/expo/pull/23057#discussion_r1240070245

This PR removed UIManager overriding by swizzling the `_manageChildren` and `uiBlockWithLayoutUpdateForRootView` methods. These changes allowed us to get rid of `RCTAppDelegate+Reanimated`, `UIResponder+Reanimated`, and `REAInitializer`.

For now, we're gonna leave `REAInitializer` as deprecated for a few more releases to maintain backward compatibility. But just a heads up, we'll be removing it in the future.

## Test plan
- I tested building on different react native versions by CI https://github.com/piaskowyk/reanimated-rn-version-tester/actions/runs/5375166823
- Tested Layout Animations on Example app form main branch.


ios/REAUtils.h
```diff
@@ -5,3 +5,10 @@
     if (value == nil)                         \
       RCTLogError(errorMsg);                  \
   })
+
+@interface REAUtils : NSObject
```
kmagiera, line 9: is there a need for this to be encapsulated within a objc class? Could just be a function, no?

ios/LayoutReanimation/REASwizzledUIManager.mm
```diff
@@ -16,76 +17,65 @@
...
+                       with:@selector(reanimated_uiBlockWithLayoutUpdateForRootView:)
+                  fromClass:[self class]];
+    SEL manageChildrenOriginal = @selector
+        (_manageChildren:moveFromIndices:moveToIndices:addChildReactTags:addAtIndices:removeAtIndices:registry:);
+    SEL manageChildrenReaniamted =
+        @selector(reanimated_manageChildren:
+                            moveFromIndices:moveToIndices:addChildReactTags:addAtIndices:removeAtIndices:registry:);
+    [REAUtils swizzleMethod:manageChildrenOriginal
+                   forClass:[RCTUIManager class]
+                       with:manageChildrenReaniamted
```
kmagiera, line 58: reaniamted

ios/LayoutReanimation/REASwizzledUIManager.mm
```diff
@@ -16,76 +17,65 @@
...
-- (void)invalidate
+- (instancetype)initWithUIManager:(RCTUIManager *)uiManager
+            withAnimatioinManager:(REAAnimationsManager *)animationsManager
 {
-  [_animationsManager invalidate];
-  [super invalidate];
+  if (self = [super init]) {
+    _animationsManager = animationsManager;
+    _uiManager = uiManager;
+    _reaUIManager = self;
```
kmagiera, line 37: I don't quite understand the need for this whole REASwizzledUIManager class to exists. Also, assigning singletons like that seem a bit risky.

It seem to me that the reason it's been done this way was for the swizzled methods to be able to access _uiManager and _animationsManager fields. I think the former can just be replaced by `self` as the swizzling concept refers to swapping out implementation of certain methods. The UI manager object will be the same and the swizzled method self will refer to it.

As for the latter field, the best way IMO would be to use category on UIManager class that adds animationsManager property to it using associated object technique (https://nshipster.com/associated-objects/)

piaskowyk: 1. I've made the changes based on your suggestions 👍. I dumped the singleton and used associated objects with the category instead.
2. The only reason I created the `REASwizzledUIManager` class was to keep the swizzled methods separate from the animation manager. I didn't want to put everything into one file. But yeah, if you think it's best, I can move the `_manageChildren` and `_uiBlockWithLayoutUpdateForRootView `methods back to the `REAAnimationManager` class.

ios/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -74,6 +73,7 @@ - (instancetype)initWithUIManager:(RCTUIManager *)uiManager
       [_currentKeys addObject:[NSString stringWithFormat:@"current%@", [key capitalizedString]]];
     }
     _sharedTransitionManager = [[REASharedTransitionManager alloc] initWithAnimationsManager:self];
+    _reaSwizzledUIManager = [[REASwizzledUIManager alloc] initWithUIManager:uiManager withAnimatioinManager:self];
```
tomekzaw, line 76:
```suggestion
    _reaSwizzledUIManager = [[REASwizzledUIManager alloc] initWithUIManager:uiManager withAnimationManager:self];
```

ios/native/REAInitializer.mm
```diff
@@ -1,33 +1,16 @@
...
 {
-  REAInitializer(bridge);
-
-  const auto runtimeInstaller = [runtimeInstallerToWrap](facebook::jsi::Runtime &runtime) {
-    if (runtimeInstallerToWrap) {
-      runtimeInstallerToWrap(runtime);
-    }
-  };
-  return runtimeInstaller;
+  return runtimeInstallerToWrap;
```
tomekzaw, line 13: I'd like to keep the old version here anyway because we guarantee that `REAJSIExecutorRuntimeInstaller` returns a valid runtime installer but `runtimeInstallerToWrap` here can be `nullptr`.

```suggestion
  const auto runtimeInstaller = [runtimeInstallerToWrap](facebook::jsi::Runtime &runtime) {
    if (runtimeInstallerToWrap) {
      runtimeInstallerToWrap(runtime);
    }
  };
  return runtimeInstaller;
```

ios/LayoutReanimation/REASwizzledUIManager.mm
```diff
@@ -1,91 +1,91 @@
...
+                addChildReactTags:(NSArray<NSNumber *> *)addChildReactTags
+                     addAtIndices:(NSArray<NSNumber *> *)addAtIndices
+                  removeAtIndices:(NSArray<NSNumber *> *)removeAtIndices
+                         registry:(NSMutableDictionary<NSNumber *, id<RCTComponent>> *)registry
 {
   bool isLayoutAnimationEnabled = reanimated::FeaturesConfig::isLayoutAnimationEnabled();
   id<RCTComponent> container;
   NSArray<id<RCTComponent>> *permanentlyRemovedChildren;
   BOOL containerIsRootOfViewController = NO;
+  RCTUIManager *_self = (RCTUIManager *)self;
```
tomekzaw, line 85: Is it customary to use `_self` in this case? Maybe let's rename this field to something like `originalSelf` so that it's easier to see the difference?

tomekzaw (approved):
Well done! ❤️

PR #4621, opened 2023-06-24
Move `scheduleOnJS` to a separate method

tomekzaw:
## Summary

This PR moves the inline implementation of `scheduleOnJS` from C++ lambda body to a separate method of `NativeReanimatedModule` so that it can be re-used for different purposes 😉 

## Test plan

WorkletExample.tsx &rarr; runOnUI + runOnJS demo


piaskowyk (approved):
😉😉😉

PR #4628, opened 2023-06-26
Remove unused parameter from _startAnimationForTag method

piaskowyk:
## Summary

This PR just removes unused parameter `depth` from _startAnimationForTag on iOS.


tomekzaw (approved):
👍

Issue #4615, opened 2023-06-22
useAnimatedReaction react() callback does not return previous value correctly


M1chaelChen:
### Description

The `react()` callback in `useAnimatedReaction` does not return previous value correctly, instead it always return `null`. I believe the issue is caused by [this](https://github.com/software-mansion/react-native-reanimated/pull/4466/files) PR which was introduced in v3.2.0

### Steps to reproduce

Repro code:
```
const App = () => {
  const [state, setState] = useState(0);
  const sv1 = useSharedValue(1);
  const sv2 = useSharedValue(1);

  const derived = useAnimatedReaction(
    () => {
      return sv1.value * state;
    },
    (result, previous) => {
      console.log(result, previous);
      if (result !== previous) {
        sv2.value = result - 5;
      }
    },
  );

  return (
    <>
      <Button onPress={() => setState((prev) => (prev += 1))}>+</Button>
      <Button onPress={() => setState((prev) => (prev -= 1))}>-</Button>
    </>
  );
};
```

In the `console.log(result, previous);`, `previous` always return `null`

### Snack or a link to a repository

N/A

### Reanimated version

3.3.0

### React Native version

0.71.8

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

PR #4633, opened 2023-06-27
Replace `useRef` with `useSharedValue` in `useAnimatedReaction` to preserve value between renders

j-piasecki:
## Summary

In https://github.com/software-mansion/react-native-reanimated/pull/4466 I've replaced `useSharedValue` with `useRef` to avoid cloning saved objects on every update. This worked really well unless, as it turned out, there was a re-render in between the updates, in which case the saved value would reset back to null.

I think it's because the ref object that was captured in the closure of the worklet responsible for updating was actually a cloned object instead of the real ref, so each update would be applied to the cloned one. This PR reverts `useRef` change and instead passes `true` for `oneWayReadsOnly`, simply preventing the cloning as it was originally done in https://github.com/software-mansion/react-native-reanimated/pull/4466.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/4615

## Test plan

Check out the reproduction code from the issue above.

### DIFF
--- a/src/reanimated2/hook/useAnimatedReaction.ts
+++ b/src/reanimated2/hook/useAnimatedReaction.ts
@@ -1,7 +1,8 @@
-import { useEffect, useRef } from 'react';
+import { useEffect } from 'react';
 import { BasicWorkletFunction, WorkletFunction } from '../commonTypes';
 import { startMapper, stopMapper } from '../core';
 import { DependencyList } from './commonTypes';
+import { useSharedValue } from './useSharedValue';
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):
Looks good to me, although the previous PR also looked good to me 😢

programmrz:
Not well versed in this, but I used to be able to pass a useSharedValue to a script that was called in a runOnJS. In that script, I would mutate it's value and the new value would be available in the original script. Now that I no longer am able to do so, is this related?

PR #4635, opened 2023-06-27
Fix inconsistent header

piaskowyk:
## Summary

In https://github.com/software-mansion/react-native-reanimated/pull/4628 I removed `depth` parameter from `_startAnimationForTag` method, but I forgot about `startAnimationForTag` method - I changed implementation but I forgot about header. Unfortunately compiler didn't tell me me this :/


tjzel (approved):
👏

tomekzaw (approved):

PR #3241, opened 2022-05-13
withDecay - add rubber band effect

piaskowyk:
## Description

I added a new parameter to `withDecay` animation:

#### Before
https://user-images.githubusercontent.com/36106620/168560831-0ccb4092-50a1-491c-8f75-65b0cb0155d8.mov

#### After
https://user-images.githubusercontent.com/36106620/168560869-a277860d-b2ec-42c6-8841-3ed89785fede.mov

## How to use it
```js
withDecay({
  clamp: [10, 100],
  velocity: 200,
  rubberBandEffect: true, // <-- flag
  rubberBandFactor: 0.6, // <-- optional config
});
```

## Example

<details>
<summary>Example code</summary>

```js
import Animated, { useSharedValue, withDecay, useAnimatedStyle } from 'react-native-reanimated';
import { StyleSheet, Button, View } from 'react-native';

import React from 'react';

export default function ScrollViewExample() {
  const position = useSharedValue(0.000001);
  const style = useAnimatedStyle(() => {
    return {
      transform: [{
        translateX: position.value
      }]
    }
  });
  return (
    <Animated.View style={{marginTop: 50, display: 'flex', alignItems: 'center'}}>
      <Button title='right' onPress={() => {
        position.value = withDecay({
          clamp: [0.000001, 100],
          velocity: 200,
          rubberBandEffect: true,
        });
      }} />
      <Button title='left' onPress={() => {
        position.value = withDecay({
          clamp: [0.000001, 100],
          velocity: -200,
          rubberBandEffect: true,
        });
      }} />
      <View style={{width: 200, height: 150, backgroundColor: 'red'}}>
        <Animated.View style={[{width: 100, height: 100, backgroundColor: 'blue'}, style]} />
      </View>
    </Animated.View>
  );
}

```

</details>

PR #4638, opened 2023-06-28
Improve decay animation with rubberBandEffect

Latropos:
## Summary

Our withDecay animation doesn't work well in the following scenario:
* `rubberBandEffect` is true
* animated object is outside clamp and is speedily moving towards clamp

Such a situation may happen if and only if the animated object has more than one `withDecay` animations which are interfering with each other.

In such a case animated object may be too fast to smoothly stop exactly at the clamp position. 
We have decided that it would look better if in such a case the view is able to jump over the clamp limit.

### Example solutions
<table>
<tr><th width="33%">Never bounce</th><th width="33%">Bounce if acceleration is big</th><th width="33%">Bounce if velocity is big</th></tr>

<tr><th>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/70825b8f-c2d3-4f36-92d8-f89bf5f7a28c

</th><th>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/4b7d6cd7-daba-42ab-a6ff-5a27150b89d1

</th><th>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/dab6fe56-8994-464d-9848-c1848b80b86f

</th></tr>

<tr><td>
<ul>
<li> Current Solution. </li>
<li> Movement looks sharp if the view is moving quickly ❌</li>
<li> The view may move quickly only in case of additional animation - otherwise looks good ✅</li>
</ul>
</td><td>
<ul>
<li> Bounce if the view is speeding towards the clamp limit.</li>
<li> Looks good in most of the cases ✅</li>
<li> Should never trigger if user use only one `withDecay` animation ✅</li>
</ul>
</td><td>
<ul>
<li> Bounce depending on current velocity. If the velocity is too big the view should bee able to cross the clamp limit </li>
<li> We have to decide if "the current velocity is big", which is difficult to be done subjectively ❌</li>
<li> The smaller is the distance between the view and the clamp the smaller should be velocity limit we use to decide if the view is moving too fast to neatly stop at the clamp ❌</li>
<li> Could trigger if user use only one `withDecay` animation ❌</li>
</td></tr>
</table>

Therefore the second solution was chosen, since it seems to be the best one.

## Test plan

Tested on example from this PR: https://github.com/software-mansion/react-native-reanimated/pull/3241


piaskowyk (commented):
1️⃣ What do you mean by "should never trigger if user doesn't use `withDecay` with some more animations"? Could you explain it a bit to me?

2️⃣ Could we do something with this occasional jump in the "Bounce if velocity is big" example?

Latropos:
@piaskowyk I've updated the description to be more precise

Latropos:
This is an old PR, we should refactor Decay and fix its types first

PR #4640, opened 2023-06-28
Register for events by name

piaskowyk:
## Summary

The current event mechanism allows us to register for an event only when we know the react tag of the event emitter. This PR adds the possibility to register on events just by event name without knowing the emitter react tag. It'll be helpful for shared transition progress feature: https://github.com/software-mansion/react-native-reanimated/pull/4421

## Test plan

I tested it on iOS/Android & Paper/Fabric

<details>
<summary>code</summary>

```js
import { ParamListBase } from '@react-navigation/native';
import {
  createNativeStackNavigator,
  NativeStackScreenProps,
} from '@react-navigation/native-stack';
import Animated, { useEvent } from 'react-native-reanimated';

function Screen1({ navigation }: NativeStackScreenProps<ParamListBase>) {
  return (
    <View style={{ flex: 1 }}>
      <Animated.View
        style={{ width: 150, height: 150, backgroundColor: 'red' }}
        sharedTransitionTag="tag1"
      />
      <Button title="Screen2" onPress={() => navigation.navigate('Screen2')} />
      <Button title="Screen3" onPress={() => navigation.navigate('Screen3')} />
    </View>
  );
}

function Screen2({ navigation }: NativeStackScreenProps<ParamListBase>) {
  return (
    <View style={{ flex: 1, marginTop: 50 }}>
      <Animated.View
        style={{ width: 100, height: 100, backgroundColor: 'green' }}
        sharedTransitionTag="tag1"
      />
      <Button title="Screen1" onPress={() => navigation.navigate('Screen1')} />
      <Button title="Screen3" onPress={() => navigation.navigate('Screen3')} />
    </View>
  );
}

function Screen3({ navigation }: NativeStackScreenProps<ParamListBase>) {
  return (
    <View style={{ flex: 1 }}>
      <Animated.View
        style={{ width: 100, height: 100, backgroundColor: 'blue' }}
        sharedTransitionTag="tag1"
      />
      <Button title="Screen1" onPress={() => navigation.navigate('Screen1')} />
      <Button title="Screen2" onPress={() => navigation.navigate('Screen2')} />
    </View>
  );
}

export default function ManyScreensExample() {
  const event = useEvent((event) => {
    'worklet'
    console.log(event);
  });
  // event.current.registerForEventByName('topTransitionProgress');
  event.current.registerForEventByName('onTransitionProgress');
  return (
    <Stack.Navigator>
      <Stack.Screen
        name="Screen1"
        component={Screen1}
        options={{ headerShown: true }}
      />
      <Stack.Screen
        name="Screen2"
        component={Screen2}
        options={{ headerShown: true }}
      />
      <Stack.Screen
        name="Screen3"
        component={Screen3}
        options={{ headerShown: true }}
      />
    </Stack.Navigator>
  );
}
```
</details>

```js
const event = useEvent((event) => {
  'worklet'
  console.log(event);
});
event.current.registerForEventByName('topTransitionProgress'); // for iOS
event.current.registerForEventByName('onTransitionProgress'); // for Android
```

https://github.com/software-mansion/react-native-reanimated/assets/36106620/783e6e1d-6f78-4981-b303-d5fb4b6b6cb4

### DIFF
--- a/Common/cpp/Registries/EventHandlerRegistry.cpp
+++ b/Common/cpp/Registries/EventHandlerRegistry.cpp
@@ -37,8 +37,8 @@ void EventHandlerRegistry::processEvent(
         handlersForEvent.push_back(handler.second);
       }
     }
-    handlersIt =
-        eventMappings.find(std::to_string(emitterReactTag) + eventName);
+    auto eventHash = std::to_string(emitterReactTag) + eventName;
+    handlersIt = eventMappings.find(eventHash);
     if (handlersIt != eventMappings.end()) {
       for (auto handler : handlersIt->second) {
         handlersForEvent.push_back(handler.second);
--- a/Common/cpp/Registries/EventHandlerRegistry.cpp
+++ b/Common/cpp/Registries/EventHandlerRegistry.cpp
@@ -26,11 +26,18 @@ void EventHandlerRegistry::processEvent(
     jsi::Runtime &rt,
     double eventTimestamp,
     const std::string &eventName,
+    const int viewTag,
     const jsi::Value &eventPayload) {
   std::vector<std::shared_ptr<WorkletEventHandler>> handlersForEvent;
   {
     const std::lock_guard<std::mutex> lock(instanceMutex);
     auto handlersIt = eventMappings.find(eventName);
+    if (handlersIt != eventMappings.end()) {
+      for (auto handler : handlersIt->second) {
+        handlersForEvent.push_back(handler.second);
+      }
+    }
+    handlersIt = eventMappings.find(std::to_string(viewTag) + eventName);
     if (handlersIt != eventMappings.end()) {
       for (auto handler : handlersIt->second) {
         handlersForEvent.push_back(handler.second);
--- a/Common/cpp/Registries/EventHandlerRegistry.cpp
+++ b/Common/cpp/Registries/EventHandlerRegistry.cpp
@@ -26,7 +26,7 @@ void EventHandlerRegistry::processEvent(
     jsi::Runtime &rt,
     double eventTimestamp,
     const std::string &eventName,
-    const int viewTag,
+    const int emitterReactTag,
     const jsi::Value &eventPayload) {
   std::vector<std::shared_ptr<WorkletEventHandler>> handlersForEvent;
   {
@@ -37,7 +37,8 @@ void EventHandlerRegistry::processEvent(
         handlersForEvent.push_back(handler.second);
       }
     }
-    handlersIt = eventMappings.find(std::to_string(viewTag) + eventName);
+    handlersIt =
+        eventMappings.find(std::to_string(emitterReactTag) + eventName);
     if (handlersIt != eventMappings.end()) {
       for (auto handler : handlersIt->second) {
         handlersForEvent.push_back(handler.second);
--- a/android/src/main/cpp/NativeProxy.h
+++ b/android/src/main/cpp/NativeProxy.h
@@ -234,7 +234,7 @@ class NativeProxy : public jni::HybridClass<NativeProxy> {
 #endif
   void handleEvent(
       jni::alias_ref<JString> eventName,
-      jint targetTag,
+      jint emitterReactTag,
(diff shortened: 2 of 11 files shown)


piaskowyk:
Should we make the events prefix name consistent between iOS and Android? Currently, on Android in JS, we receive events with the prefix `on` but on iOS with the prefix `top` 🤷‍♂️

ios/REANodesManager.mm
```diff
@@ -187,6 +187,8 @@ - (nonnull instancetype)initWithModule:(REAModule *)reanimatedModule
     _operationsInBatch = [NSMutableDictionary new];
     _componentUpdateBuffer = [NSMutableDictionary new];
     _viewRegistry = [_uiManager valueForKey:@"_viewRegistry"];
+    _eventHandler = ^(id<RCTEvent> event) {
+    };
```
tomekzaw, line 191: Let's mark this block as no-op explicitly:
```suggestion
    _eventHandler = ^(id<RCTEvent> event) {
      // no-op
    };
```
If this block isn't meant to be called, we can add something like `assert(false);` inside to make sure we notice it once it starts to fail.

android/src/main/cpp/NativeProxy.h
```diff
@@ -66,8 +66,9 @@ class EventHandler : public HybridClass<EventHandler> {
 
   void receiveEvent(
       jni::alias_ref<JString> eventKey,
+      jint targetTag,
```
tomekzaw, line 69: Since we already have `emitterReactTag`, does it make sense to rename this to `targetReactTag` as well?
```suggestion
      jint targetReactTag,
```

android/src/main/java/com/swmansion/reanimated/nativeProxy/EventHandler.java
```diff
@@ -20,12 +20,13 @@ private EventHandler(HybridData hybridData) {
...
-  public void receiveEvent(int targetTag, String eventName, @Nullable WritableMap event) {
+  public void receiveEvent(int emitterReactTag, String eventName, @Nullable WritableMap event) {
     String resolvedEventName = mCustomEventNamesResolver.resolveCustomEventName(eventName);
-    receiveEvent(targetTag + resolvedEventName, event);
+    receiveEvent(resolvedEventName, emitterReactTag, event);
   }
 
-  public native void receiveEvent(String eventKey, @Nullable WritableMap event);
+  public native void receiveEvent(
+      String eventKey, int emitterReactTag, @Nullable WritableMap event);
```
tomekzaw, line 29: I think we should get rid of all occurrences of `eventKey`.
```suggestion
      String eventName, int emitterReactTag, @Nullable WritableMap event);
```

ios/REANodesManager.mm
piaskowyk: Theoretically there is a very rare case when it could be called, it someone had emitted event before the `registerEventHandler` method was called.

android/src/main/cpp/NativeProxy.h
piaskowyk: Thanks, i missed it! I think that more relevant will be `emitterReactTag` instead of `targetReactTag` here, what do you think?

Common/cpp/Registries/EventHandlerRegistry.cpp
```diff
@@ -36,6 +37,13 @@ void EventHandlerRegistry::processEvent(
         handlersForEvent.push_back(handler.second);
       }
     }
+    auto eventHash = std::to_string(emitterReactTag) + eventName;
```
tomekzaw, line 40: I'm wondering if we can use `std::make_pair(emitterReactTag, eventName)` as hash type instead of concatenation.

tomekzaw (approved):

Common/cpp/Registries/EventHandlerRegistry.cpp
piaskowyk: Great Idea! As we discussed offline, I'll do it in another PR.

PR #4642, opened 2023-06-28
WorkletEventHandler refactor

piaskowyk:
## Summary

This PR makes a few changes:
- It introduces a consistent naming pattern for the `WorkletEvenHandler` class.
- It updates the names of the class fields to be more relevant.
- It removes the class's friendship with `EventHandlerRegistry`.
- It removes the dummy definition of `EventHandlerRegistry`.
- It improves encapsulation by adding a getter function.

### DIFF
--- a/Common/cpp/Registries/EventHandlerRegistry.cpp
+++ b/Common/cpp/Registries/EventHandlerRegistry.cpp
@@ -6,7 +6,7 @@ namespace reanimated {
 void EventHandlerRegistry::registerEventHandler(
     std::shared_ptr<WorkletEventHandler> eventHandler) {
   const std::lock_guard<std::mutex> lock(instanceMutex);
-  const std::string &eventName = eventHandler->getEventName();
+  const auto &eventName = eventHandler->getEventName();
   auto handlerId = eventHandler->getHandlerId();
   eventMappings[eventName][handlerId] = eventHandler;
   eventHandlers[handlerId] = eventHandler;
@@ -16,10 +16,12 @@ void EventHandlerRegistry::unregisterEventHandler(uint64_t id) {
   const std::lock_guard<std::mutex> lock(instanceMutex);
   auto handlerIt = eventHandlers.find(id);
   if (handlerIt != eventHandlers.end()) {
-    const std::string &eventName = handlerIt->second->getEventName();
-    eventMappings[eventName].erase(id);
-    if (eventMappings[eventName].empty()) {
-      eventMappings.erase(eventName);
+    const auto &eventName = handlerIt->second->getEventName();
+    auto eventMappingIt = eventMappings.find(eventName);
+    auto &handlersMap = eventMappingIt->second;
+    handlersMap.erase(id);
+    if (handlersMap.empty()) {
+      eventMappings.erase(eventMappingIt);
     }
     eventHandlers.erase(handlerIt);
   }
--- a/Common/cpp/Tools/WorkletEventHandler.h
+++ b/Common/cpp/Tools/WorkletEventHandler.h
@@ -12,16 +12,16 @@ using namespace facebook;
 namespace reanimated {
 
 class WorkletEventHandler {
-  std::shared_ptr<JSRuntimeHelper> runtimeHelper_;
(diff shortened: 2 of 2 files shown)


Common/cpp/Registries/EventHandlerRegistry.cpp
```diff
@@ -6,17 +6,20 @@ namespace reanimated {
 void EventHandlerRegistry::registerEventHandler(
     std::shared_ptr<WorkletEventHandler> eventHandler) {
   const std::lock_guard<std::mutex> lock(instanceMutex);
-  eventMappings[eventHandler->eventName][eventHandler->id] = eventHandler;
-  eventHandlers[eventHandler->id] = eventHandler;
+  const std::string &eventName = eventHandler->getEventName();
+  auto handlerId = eventHandler->getHandlerId();
+  eventMappings[eventName][handlerId] = eventHandler;
```
tomekzaw, line 11: Does it make sense to use `std::make_pair(eventName, handlerId)` as key here to avoid double lookup?

Common/cpp/Registries/EventHandlerRegistry.cpp
```diff
@@ -6,17 +6,20 @@ namespace reanimated {
 void EventHandlerRegistry::registerEventHandler(
     std::shared_ptr<WorkletEventHandler> eventHandler) {
   const std::lock_guard<std::mutex> lock(instanceMutex);
-  eventMappings[eventHandler->eventName][eventHandler->id] = eventHandler;
-  eventHandlers[eventHandler->id] = eventHandler;
+  const std::string &eventName = eventHandler->getEventName();
```
tomekzaw, line 9: What about using `const auto &` here instead? 🤔 
```suggestion
  const auto &eventName = eventHandler->getEventName();
```

Common/cpp/Registries/EventHandlerRegistry.cpp
```diff
@@ -6,17 +6,20 @@ namespace reanimated {
...
 void EventHandlerRegistry::unregisterEventHandler(uint64_t id) {
   const std::lock_guard<std::mutex> lock(instanceMutex);
   auto handlerIt = eventHandlers.find(id);
   if (handlerIt != eventHandlers.end()) {
-    eventMappings[handlerIt->second->eventName].erase(id);
-    if (eventMappings[handlerIt->second->eventName].empty()) {
-      eventMappings.erase(handlerIt->second->eventName);
+    const std::string &eventName = handlerIt->second->getEventName();
+    eventMappings[eventName].erase(id);
+    if (eventMappings[eventName].empty()) {
```
tomekzaw, line 21: Do you think we can avoid double lookup here, i.e. call `auto mappingIt = eventMappings.find(eventName)` once?

Common/cpp/Tools/WorkletEventHandler.h
```diff
@@ -11,28 +11,25 @@ using namespace facebook;
...
-
- private:
-  std::shared_ptr<JSRuntimeHelper> _runtimeHelper;
-  uint64_t id;
-  std::string eventName;
-  jsi::Value _handlerFunction;
+  std::shared_ptr<JSRuntimeHelper> runtimeHelper_;
+  jsi::Value handlerFunction_;
+  uint64_t handlerId_;
+  std::string eventName_;
```
tomekzaw, line 18: Can we mark some of these fields as `const`?

Common/cpp/Tools/WorkletEventHandler.h
```diff
@@ -11,28 +11,25 @@ using namespace facebook;
...
 
  public:
   WorkletEventHandler(
       const std::shared_ptr<JSRuntimeHelper> &runtimeHelper,
-      uint64_t id,
+      uint64_t handlerId,
       std::string eventName,
```
tomekzaw, line 24:
```suggestion
      const std::string &eventName,
```

tomekzaw (approved):
+1 for the changes! I've left some more suggestions in the code, but the PR is already looking good!

tomekzaw (approved):

PR #4648, opened 2023-06-29
Add pendulum example

Latropos:
## Summary

This is a pendulum example with some config that can be defined with UI. This was used as an twitter example [LINK](https://twitter.com/AleksandraCynk/status/1676972530618486784?ref_src=twsrc%5Etfw)

image


app/src/examples/PendulumExample.tsx
```diff
@@ -0,0 +1,233 @@
...
+      <TextInput
+        key={fieldName}
+        value={textValue}
+        onChangeText={(s) => {
+          const parsedInput = Number.parseFloat(s);
+          if (parsedInput) {
+            setValue(parsedInput);
+          }
+          setTextValue(s);
+        }}
```
tomekzaw, line 56: Can we move it outside and memoize it?

app/src/examples/PendulumExample.tsx
```diff
@@ -0,0 +1,233 @@
...
+  );
+}
+
+export default function SpringExample(): React.ReactElement {
+  const pendulumSwing = useSharedValue(0);
+  const offset = useSharedValue({ x: 0, y: 0 });
+  const [useConfigWithDuration, setUseConfigWithDuration] = useState(true);
+
+  const [stiffnessText, setStiffnessText] = useState('100');
+  const [stiffness, setStiffness] = useState(100);
```
tomekzaw, line 71: We should only keep the state with number and parse it accordingly.

app/src/examples/PendulumExample.tsx
```diff
@@ -0,0 +1,233 @@
...
+        { rotateZ: `${pendulumSwing.value}deg` },
+        { translateY: 100 },
+      ],
+    };
+  });
+
+  const gesture = Gesture.Pan()
+    .manualActivation(true)
+    .onChange((e) => {
+      'worklet';
```
tomekzaw, line 108: There's no need to mark these callbacks as worklets since they will be auto-workletized.

app/src/examples/PendulumExample.tsx
Latropos: I'm afraid we can't, since I've moved InputField out of function SpringExample to reduce number re-renders (which make textInput loose its focus)

app/src/examples/PendulumExample.tsx
```diff
@@ -0,0 +1,259 @@
...
+        </View>
+      </GestureDetector>
+      <React.Fragment>
+        {fields.map((item) => {
+          return (
+            <InputField
+              fieldName={item[0]}
+              value={item[1]}
+              setValue={item[2]}
+              key={item[0]}
```
piaskowyk, line 172: Lets replace it with object instead of array

piaskowyk (approved):
👏👏

PR #4652, opened 2023-06-30
Autoworkletize callbacks in Layout Animations

tjzel:
## Summary

Currently we support auto-workletizing callbacks with method chaining for `react-native-gesture-handler`:

```TS
Gesture.Tap().onBegin(() => {
  // will be automatically workletized
  ...
});
```

This pull request adds similar functionality to Layout Animations and their `withCallback` method so the `worklet` directive is no longer necessary.

```TS
FadeIn.duration(400).withCallback(() => {
  // will be automatically workletized now
  ...
});
```

## Test plan

A new test suite was added to plugin's tests.


__tests__/plugin.test.ts
piaskowyk, line 1: I think we need also tests for class instance, for example - `new FadeIn().withCallback()` and for longer chaining like - `FadeIn.duration().springify().withCallback()`, `FadeIn.duration().withCallback().springify()`

plugin/src/isLayoutAnimationCallback.ts
piaskowyk, line 1: I'm just wondering if it is possible to create tests that will check if all names from `EntryExitAnimations`, `LayoutTransitions`, `BaseAnimationsChainableMethods`, and `DefaultTransitionChainableMethods` are still relevant. I mean, how to detect that someone changes the name of a class/method.

tjzel: I've tried to do it but it seems very complicated. TypeScript makes it difficult enough to import (actually require) anything outside plugin's root directory and forces us to use require, therefore we would need to use require on generated lib files 😵

piaskowyk (approved):

__tests__/plugin.test.ts
tjzel: I forgor to add () to constructor here, therefore objects with new operator actually have to be handled 👍

PR #4653, opened 2023-06-30
Fix TypeScript issue with useAnimatedProps

tjzel:
## Summary

During the migration from manual types, the proper typing of "useAnimatedProps" has proven to be quite challenging. As a result, I had to resort to casting it to its previous type, but unfortunately, I missed some places.

## Test plan

In our new docs the given [example](https://reanimated-beta-docs.swmansion.com/docs/core/useAnimatedProps#adapters-) doesn't work.
After applying these changes it's fixed.


src/reanimated2/helperTypes.ts
```diff
@@ -115,3 +116,11 @@ export type AnimateProps<P extends object> = NonStyleAnimatedProps<P> &
   };
 
 export type AnimatedProps<P extends object> = AnimateProps<P>;
+
+export type PropsAdapterFunction = (props: Record<string, unknown>) => void;
```
tomekzaw, line 120:
```suggestion
export type AnimatedPropsAdapterFunction = (props: Record<string, unknown>) => void;
```

src/reanimated2/PropAdapters.ts
```diff
@@ -1,18 +1,25 @@
...
+): AdapterWorkletFunction => {
   const nativePropsToAdd: { [key: string]: boolean } = {};
   // eslint-disable-next-line no-unused-expressions
   nativeProps?.forEach((prop) => {
     nativePropsToAdd[prop] = true;
   });
   addWhitelistedNativeProps(nativePropsToAdd);
   return adapter;
-}
+}) as createAnimatedPropAdapterType;
```
tomekzaw, line 22: Why do we need to cast the whole function instead of just casting the return value?

piaskowyk: If it is not a 'hot fix' I think we should avoid type casting

tomekzaw (approved):

PR #4654, opened 2023-06-30
Improve plugin's code quality and test organization

tjzel:
## Summary

Wrapping commonly used union types in their own type alias, moved inline code into function and reorganized tests into more 'describe' blocks for easier navigation.


tomekzaw (approved):
:prayge:

PR #4655, opened 2023-06-30
Add eslint rule to prefer `import type` syntax

tjzel:
## Summary

Added as discussed in #4519. All changes in file except for `.eslint.rc` were made automatically by `eslint --fix` and `yarn format`.


Latropos (approved):
🥳

Issue #4656, opened 2023-06-30
[3.4.0 nightly] isWeb is not a function on decay.ts


efstathiosntonas:
### Description

On yesterday’s nightly `3.4.0-nightly-20230629-1fa467e6a`  it throwed this error:

`facebook::jsi::JSError: isWeb is not a function (it is Object)

TypeError: isWeb is not a function (it is Object)
    at anonymous (JavaScript:1:103)
    at anonymous (JavaScript:1:1813)
    at apply (native)
    at runWorklet (JavaScript:1:207)
    at anonymous (JavaScript:1:886)
    at handleAndFlushAnimationFrame (JavaScript:1:147)

0 unknown file -0x4 unknown method`

I use this library to display full screen gallery: https://github.com/Flair-Dev/react-native-awesome-gallery

### Steps to reproduce

1. Install yesterday’s nightly `3.4.0-nightly-20230629-1fa467e6a`

### Snack or a link to a repository

*

### Reanimated version

3.4.0 nighlty 

### React Native version

0.72.1

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

Real device

### Device model

iPhone 13

### Acknowledgements

Yes

efstathiosntonas:
some more context, while swiping down the image I get this error:  

![Simulator Screenshot - iPhone 14 Pro Max - 2023-07-01 at 19 32 53](https://github.com/software-mansion/react-native-reanimated/assets/717975/f1e79bbd-835f-427d-b53f-cf2dfad7fc2c)

https://github.com/software-mansion/react-native-reanimated/blob/7143b1e049ba721b51ba570e2e22eb7f00672196/src/reanimated2/animation/decay.ts#L52

the error on first comment is from Bugsnag and it's misleading :(

efstathiosntonas:
after reverting this commit https://github.com/software-mansion/react-native-reanimated/commit/bbc755bd685435392cf37e4e69f0deb27d18aa37 everything is back to normal.

tomekzaw:
@efstathiosntonas Thanks a lot for the investigation! This is the change that I've introduced in #4594:

[...]

PR #4661, opened 2023-07-02
Fix isWeb is not a function

efstathiosntonas:
## Summary
`const IS_WEB = isWeb();` was accidentaly moved inside the `'worklet'` function in this [PR](https://github.com/software-mansion/react-native-reanimated/pull/4519) on `decay.ts`

This PR moves it outside the `'worket'` function.

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/4656


tomekzaw (approved):
@efstathiosntonas Thanks for this PR!

PR #4665, opened 2023-07-03
Add fallback to setImmediate for Expo Router static rendering

javascripter:
## Summary
- https://github.com/expo/router/issues/718
  - [Twitter thread](https://twitter.com/bhyoo99/status/1675302401677615104?s=20)

In Node.js environments such as when static rendering with Expo Router, requestAnimationFrame is not defined. This results in an error when using libraries that depend on react-native-reanimated like Drawer or when using react-native-reanimated directly in Expo Router with static rendering, causing a hydration mismatch due to rendering errors.

In this PR, I added a fallback to setImmediate if requestAnimationFrame is not defined in `globalThis`. I did not simply replace requestAnimationFrame with setImmediate because [`setImmediate` is slower than requestAnimationFrame](https://github.com/software-mansion/react-native-reanimated/pull/1521) in CSR environments. This issue is not applicable to static rendering environments however, so setImmediate is fine there.

## Test plan
- See [this issue](https://github.com/expo/router/issues/718#issue-1784125278)'s Minimal reproducible example


javascripter:
Not very sure how to fix failing jest tests.

satya164:
Seems like the actual issue is that `scheduleOnUI` is being called during render for SSR/static build - which results in calling `requestAnimationFrame`. Side effects and callbacks shouldn't be called during SSR/static build.

javascripter:
> Seems like the actual issue is that `scheduleOnUI` is being called during render for SSR/static build - which results in calling `requestAnimationFrame`. Side effects and callbacks shouldn't be called during SSR/static build.

@satya164 I see. If I change the code to the below tests seem to pass. Do you think the below approach seems more reasonable? If so, I will update both the code and the PR description accordingly.
```typescript
  scheduleOnUI<T>(worklet: ShareableRef<T>) {
    // In Node.js environments (like when static rendering with Expo Router)
    // requestAnimationFrame is unavailable. Do not call side effects in this case.
    if (typeof requestAnimationFrame === 'undefined') {
      return;
    }
    // @ts-ignore web implementation has still not been updated after the rewrite, this will be addressed once the web implementation updates are ready
    requestAnimationFrame(worklet);
  }

```

tjzel:
I've tried to understand what's causing the jest problem but still haven't got any solid results.

From what I concluded if we use `scheduleOnUI` (which comes from `scheduleOnUI = requestAnimationFrame`) in `scheduleOnUI` (btw. I think it's a bad idea to have the same name for those) it's causing tests to fail, but if we call directly `requestAnimationFrame` in `scheduleOnUI` it's ok. 
Probably because `scheduleOnUI` gets torn down (or torn down before) `requestAnimationFrame` and is still called? I haven't yet had a great deal of experience of solving Jest issues so I'm not exactly sure what that means.

satya164:
> @javascripter: Do you think the below approach seems more reasonable?

Imo it doesn't address the root of the problem - scheduleOnUI being called during SSR/static build

nDeavor9:
Sorry to directly comment on this.
Can you guys please let me know when will this PR be merged?
I'm facing the `ReferenceError: requestAnimationFrame is not defined at JSReanimated.scheduleOnUI` and it's blocking my release.
Hopefully it gets done soon. Thanks.

tomekzaw:
Hey @nDeavor9 and @arivanbastos, can you confirm if this PR resolves the issue?

nDeavor9:
@tomekzaw I guess so. 
As @satya164 said, the root problem might be different, but the error that's blocking from exporting is exactly 'requestAnimationFrame' being undefined.

sabyasachibose-sigfyn:
Our builds are blocked at the moment. Can we please have a resolution on this?

arivanbastos:
@tomekzaw 
I'm sorry, but I dont know how to test this PR.

bhyoo99:
Here is my patch-package file
https://gist.github.com/bhyoo99/f4954316c498f3ca261197166855e21d

✅ expo-router checked
✅ expo-router SSG build checked

bhyoo99:
Here is my patch-package file
https://gist.github.com/bhyoo99/f4954316c498f3ca261197166855e21d

✅ expo-router checked
✅ SSG web build checked

tjzel:
Hi @javascripter! I took the liberty to commit to your fork, since I got a solution for this that should both work and pass our tests. Turns out at some point in our code (but after creating `scheduleOnUI`) we substitute `requestAnimationFrame` implementation accordingly to how it should be done with React Native. Since it's happening later, `scheduleOnUI` would not have it's implementation substituted. 

Please let me know if it works for you - if it does, I think it is safe to merge (after my colleagues review it of course).

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -6,12 +6,16 @@ import {
...
 
 // In Node.js environments (like when static rendering with Expo Router)
 // requestAnimationFrame is unavailable, so we use setImmediate.
-const scheduleOnUI =
-  globalThis.requestAnimationFrame || globalThis.setImmediate;
+const scheduleOnUI = globalThis.requestAnimationFrame
+  ? isJest()
+    ? mockedRequestAnimationFrame
+    : globalThis.requestAnimationFrame
+  : setImmediate;
```
tomekzaw, line 18: 1. First, can we handle Jest first and then the rest?
```ts
const scheduleOnUI = isJest()
  ? mockedRequestAnimationFrame
  : globalThis.requestAnimationFrame || globalThis.setImmediate;
```
2. Can we add a new platform checker `isStaticWeb()`? This way we could write it this way:
```ts
const scheduleOnUI = isJest()
  ? mockedRequestAnimationFrame
  : isStaticWeb()
  ? globalThis.setImmediate
  : globalThis.requestAnimationFrame;
```

tjzel: 1. As far as I understand in this environment `useAnimatedFrame` is unavailable in Jest also and we want to use `setImmediate` there. With this implementation we would always use the mock - maybe it's desired - @javascripter could you give more insight on that?

2. This is hard, we have to discuss it tomorrow.

javascripter:
> Hi @javascripter! I took the liberty to commit to your fork, since I got a solution for this that should both work and pass our tests. Turns out at some point in our code (but after creating `scheduleOnUI`) we substitute `requestAnimationFrame` implementation accordingly to how it should be done with React Native when using Jest. Since it's happening later, `scheduleOnUI` would not have it's implementation substituted.
> 
> Please let me know if it works for you - if it does, I think it is safe to merge (after my colleagues review it of course).

@tjzel Hi. Thanks for taking a look at this! Yes, this works for me, and the commit looks good to me as well.

src/reanimated2/js-reanimated/JSReanimated.ts
javascripter: I used setImmediate in the static rendering environment simply because I tested it in my environment and it seemed to work well for me. As @satya164 said, scheduleOnUI probably doesn't need to execute in that environment so I don't have a concrete reason to prefer setImmediate over the mock in Jest. Always using the mock in Jest seems a reasonable approach to me.  @tjzel

tjzel: We actually had a quite fruitful discussion about this and we want to go a step further and improve how Reanimated behaves during Static Renders. This PR can stay as a workaround for the moment, but we will open a proper one in the near future.

kevinhylant:
Looks like there's a conflict — hoping this will get merged soon-ish :) 

Thanks for all your work here!

tomekzaw (approved):

src/reanimated2/initializers.ts
```diff
@@ -150,9 +151,7 @@ export function initializeUIRuntime() {
     // We override this setup here to make sure that callbacks get the proper timestamps
     // when executed. For non-jest environments we define requestAnimationFrame in setupRequestAnimationFrame
     // @ts-ignore TypeScript uses Node definition for rAF, setTimeout, etc which returns a Timeout object rather than a number
-    global.requestAnimationFrame = (callback: (timestamp: number) => void) => {
-      return setTimeout(() => callback(performance.now()), 0);
-    };
+    global.requestAnimationFrame = mockedRequestAnimationFrame;
```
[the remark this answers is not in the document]
tjzel: In our use cases it should be irrelevant what we assign it to according to [some docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) but I guess it's better practice to use `globalThis`.

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -19,7 +19,7 @@ import { mockedRequestAnimationFrame } from '../utils';
 // In Node.js environments (like when static rendering with Expo Router)
 // requestAnimationFrame is unavailable, so we use our mock.
 // It also has to be mocked for Jest purposes (see `initializeUIRuntime`).
-const _scheduleOnUI =
+const JSScheduleOnUI =
```
tomekzaw, line 22:
```suggestion
const requestAnimationFrameImpl =
```

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -38,7 +38,7 @@ export default class JSReanimated {
 
   scheduleOnUI<T>(worklet: ShareableRef<T>) {
     // @ts-ignore web implementation has still not been updated after the rewrite, this will be addressed once the web implementation updates are ready
-    _scheduleOnUI(worklet);
+    JSScheduleOnUI(worklet);
```
tomekzaw, line 41:
```suggestion
    requestAnimationFrameImpl(worklet);
```

tomekzaw (approved):

PR #4672, opened 2023-07-04
Add more eslint rules

Latropos:
## Summary

* Rules for whole repo:
  * '`@typescript-eslint/no-duplicate-type-constituents': 'error'` - means that types like `number|number` are not allowed
  * `eqeqeq: 'error' - you can't compare with `==`, use `===` instead
  
  * `'no-unreachable': 'error'` - no unreachable code, especially inside a function, after return statement. 
  
* Rules for examples in folder app:
  * `'react-native/no-unused-styles': 'error'`,
  * `'react-native/no-raw-text': 'off'`, - this rule is very convenient, but due to this bug [LINK](https://github.com/Intellicode/eslint-plugin-react-native/issues/270) we can't use it with CI 😠
  * `'no-inline-styles/no-inline-styles': 'error'`
  * `'react-native/no-inline-styles': 'off'`, (this rule does not allow variable styles with ternary operators, so I've used the one above)
  
  * `'react-native/no-single-element-style-arrays': 'error'`,
* Create separate .eslintrc.js config for tests with separate rules
  * `'jest/no-disabled-tests': 'warn'`,
  
  * `'jest/no-focused-tests': 'error'`,
  * `'jest/no-identical-title': 'error'`,
  * `'jest/prefer-to-have-length': 'warn'`,
  * `'jest/valid-expect': 'error'`,

## Test plan

CI check pass ✅


src/reanimated2/layoutReanimation/defaultTransitions/EntryExitTransition.ts
```diff
@@ -182,7 +182,6 @@ export class EntryExitTransition
           } else {
             return { [transformProp]: 1 } as unknown as TransformProperty;
           }
-          return value;
```
piaskowyk, line 185: But why? :o

package.json
```diff
@@ -94,6 +94,7 @@
     "react-native": "*"
   },
   "devDependencies": {
+    "-": "^0.0.1",
```
piaskowyk, line 97: what it is?

src/reanimated2/layoutReanimation/defaultTransitions/EntryExitTransition.ts
Latropos: Because of `'no-unreachable'` There is already a return statement inside if-else condition above.

package.json
Latropos: I fon't know, thanks for catching this!

PR #4677, opened 2023-07-06
Move monorepo-build CI to nightly runs

tjzel:
## Summary

Currently this CI is extremely slow and we often accidentally cancel it/have to cancel it when we want to merge more often. With this change it's only run when it's absolutely needed and on daily (nightly) basis.


tomekzaw (approved):
❤️

PR #4676, opened 2023-07-06
Add app icon and splash screen in FabricExample

jfedak:
## Summary

Previously, there was no app icon nor splash screen in FabricExample app. This PR adds those things both on iOS and Android. Due to the usage of different icon, users can distinguish ReanimatedExample app from FabricExample app.

iOS:
New app icon was placed in `Images.xcassets` directory. Splash screen uses the configuration from `LaunchScreen.storyboard`.

https://github.com/software-mansion/react-native-reanimated/assets/56112901/b5297452-5efc-4d50-a154-4a6318519fb0

Android:
Old icon was replaced with new one inside `res/mipmap`. The splash screen was implemented with 
[androidx.core:core-splashscreen](https://developer.android.com/jetpack/androidx/releases/core).

https://github.com/software-mansion/react-native-reanimated/assets/56112901/21816d81-ffbc-4500-b398-37377b06d2d2

https://github.com/software-mansion/react-native-reanimated/assets/56112901/44e8599d-06c0-460f-ada0-bfd523c61bd0

PR #4686, opened 2023-07-07
Update app icon and splash screen in Example

jfedak:
## Summary

While working with app icon and splash screen for FabricExample app (see #4676), it turned out that the approach used in Example is outdated. This PR fixed this and makes both approaches similar. 

iOS:
From now on there is only one icon image file. There is also a new file `Launch Screen.storyboard` that replaced `LaunchScreen.xib`. The appearance of the icon and the launch screen remained the same.

 

https://github.com/software-mansion/react-native-reanimated/assets/56112901/c64577be-a1e9-493a-b143-5911c6744c4e

Android:
`res/` directory has been restructured to make it similar with FabricExample. The splash screen was implemented with
[androidx.core:core-splashscreen](https://developer.android.com/jetpack/androidx/releases/core) thus the screen is present in Android 11 and lower. The appearance for newer Androids didn't change.

https://github.com/software-mansion/react-native-reanimated/assets/56112901/a1f99cf2-9f64-4cc0-9cbd-ad69bbb3c2fb

https://github.com/software-mansion/react-native-reanimated/assets/56112901/5808dd76-be93-42e1-b6fe-4d928e9211e5


tjzel (approved):
🥳

tomekzaw (approved):
Looks good to me! 👍

PR #4687, opened 2023-07-07
Add useReducedMotion hook

bartlomiejbloniarz:
## Summary

Animations can cause some people to feel unwell. In order to help developers create apps with those people in mind this PR adds a `useReducedMotion` hook that returns a boolean value informing whether the "Reduced Motion" setting is enabled. Using this hook will cause the component to reload whenever the setting changes.

https://github.com/software-mansion/react-native-reanimated/assets/56109050/26d75993-f504-4138-87cc-cde784b7539a

## Test plan
Create an animation that runs conditionally on the value returned from `useReducedMotion` and check the behavior on android and iOS. This code was used in the video:
 
```js
const rotation = useSharedValue(0)
  const shouldReduceMotion = useReducedMotion()

  const animatedStyles = useAnimatedStyle(() => ({
    transform: [{ rotate: shouldReduceMotion ? "0deg" : rotation.value + "deg"}]
  }));

  React.useEffect(() => {
      rotation.value = withRepeat(withTiming(360, {duration: 1000, easing: Easing.linear}), -1)
  }, []);

  return (
      <Animated.View style={[styles.box, animatedStyles]} />
    </View>
  );
```

### DIFF
--- a/src/reanimated2/hook/Hooks.ts
+++ b/src/reanimated2/hook/Hooks.ts
@@ -1,7 +1,8 @@
-import { useCallback } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useAnimatedStyle } from './useAnimatedStyle';
 import type { DependencyList } from './commonTypes';
 import type { useAnimatedPropsType } from '../helperTypes';
+import { AccessibilityInfo } from 'react-native';
 
 // TODO: we should make sure that when useAP is used we are not assigning styles
 // when you need styles to animated you should always use useAS
@@ -13,6 +14,32 @@ export function useWorkletCallback<A extends unknown[], R>(
   deps?: DependencyList
 ): (...args: Parameters<typeof fun>) => R {
   return useCallback(fun, deps ?? []);
+}
+
+export function useReducedMotion(): boolean {
+  const [shouldReduceMotion, setShouldReduceMotion] = useState(false);
+
+  useEffect(() => {
+    const handleChange = (isReduceMotionEnabled: boolean): void => {
+      setShouldReduceMotion(isReduceMotionEnabled);
+    };
+
+    async function init(): Promise<void> {
+      handleChange(await AccessibilityInfo.isReduceMotionEnabled());
+    }
+    init();
+
+    const reduceMotionChangedlistener = AccessibilityInfo.addEventListener(
+      'reduceMotionChanged',
+      handleChange
+    );
+
+    return (): void => {
+      reduceMotionChangedlistener.remove();
+    };
+  }, []);
+
+  return shouldReduceMotion;
 }
 
 export { useEvent, useHandler } from './utils';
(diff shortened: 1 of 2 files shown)


src/reanimated2/hook/Hooks.ts
```diff
@@ -15,4 +16,30 @@ export function useWorkletCallback<A extends unknown[], R>(
...
+    const handleChange = (isReduceMotionEnabled: boolean): void => {
+      setShouldReduceMotion(isReduceMotionEnabled);
+    };
+
+    async function init(): Promise<void> {
+      handleChange(await AccessibilityInfo.isReduceMotionEnabled());
+    }
+    init();
+
+    const reduceMotionChangedlistener = AccessibilityInfo.addEventListener(
```
tomekzaw, line 32:
```suggestion
    const listener = AccessibilityInfo.addEventListener(
```

src/reanimated2/hook/Hooks.ts
```diff
@@ -15,4 +16,30 @@ export function useWorkletCallback<A extends unknown[], R>(
...
+
+  useEffect(() => {
+    const handleChange = (isReduceMotionEnabled: boolean): void => {
+      setShouldReduceMotion(isReduceMotionEnabled);
+    };
+
+    async function init(): Promise<void> {
+      handleChange(await AccessibilityInfo.isReduceMotionEnabled());
+    }
+    init();
```
tomekzaw, line 30: Why not just the following?
```suggestion
    AccessibilityInfo.isReduceMotionEnabled().then(setShouldReduceMotion);
```

src/reanimated2/hook/Hooks.ts
```diff
@@ -15,4 +16,30 @@ export function useWorkletCallback<A extends unknown[], R>(
...
+
+    const reduceMotionChangedlistener = AccessibilityInfo.addEventListener(
+      'reduceMotionChanged',
+      handleChange
+    );
+
+    return (): void => {
+      reduceMotionChangedlistener.remove();
+    };
```
tomekzaw, line 39:
```suggestion
    return () => listener.remove();
```

src/reanimated2/hook/Hooks.ts
```diff
@@ -15,4 +16,30 @@ export function useWorkletCallback<A extends unknown[], R>(
   return useCallback(fun, deps ?? []);
 }
 
+export function useReducedMotion(): boolean {
+  const [shouldReduceMotion, setShouldReduceMotion] = useState(false);
+
+  useEffect(() => {
+    const handleChange = (isReduceMotionEnabled: boolean): void => {
+      setShouldReduceMotion(isReduceMotionEnabled);
+    };
```
tomekzaw, line 25:
```suggestion
```

src/reanimated2/hook/Hooks.ts
tjzel: probably with await here

tomekzaw (approved):
Looks good to me ✅

hubertdej:
Cool feature! 😍

piaskowyk (approved):
👏

Issue #4689, opened 2023-07-09
Output of react-native bundle containes source code loation when using with babel plugin


XHMM:
### Description

When I running ` npx react-native bundle --dev false  --platform ios --minify --bundle-output ./bundle.js --entry-file ./index.js --reset-cache` for building a release bundle, I foud that my source code location  was included in the output file like this:
<img width="1065" alt="image" src="https://github.com/software-mansion/react-native-reanimated/assets/24786522/22a39581-70db-43bd-bbcd-b9b095a6a97d">

If I removed `react-native-reanimated/plugin` from `babel.config.js` and rebuild, the output content will not include these location which was expected.

I noticed that the code from `react-native-gesture-handler` and `react-native-reanimated` will have location output, I am not very sure the issue was from metro or babel or this babel plugin

### Steps to reproduce

-

### Snack or a link to a repository

-

### Reanimated version

3.3.0

### React Native version

0.71.11

### Platforms

iOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

None

### Device model

-

### Acknowledgements

Yes

tjzel:
Hi @XHMM, thanks for this issue! Indeed this is our (Reanimated Babel Plugin's) fault and this has been like this forever - I'm surprised nobody has ever reported it. I already make a PR that fixes it since it's an easy fix.

PR #4695, opened 2023-07-10
Remove location property in worklets for production builds

tjzel:
## Summary

Currently we inject `location` property in worklets - which is there for debugging purposes - even in production build. This PR corrects this.

Fixes: #4689

## Test plan

Added a new test suite in plugin. Also ran production android Example just to be sure.


__tests__/__snapshots__/plugin.test.ts.snap
```diff
@@ -53,6 +71,21 @@ var f = function () {
 }();"
 `;
 
+exports[`babel plugin doesn't inject location for worklets in production builds 1`] = `
+"var _worklet_1852213185147_init_data = {
+  code: "function anonymous(){const x=1;}"
```
tomekzaw, line 76: I'm quite worried about the removal of `location` property, looks like we still use it here:

https://github.com/software-mansion/react-native-reanimated/blob/c6e7523541567a38dc4cb8d9c705b11e746a8cdf/src/reanimated2/initializers.ts#L47-L50

Maybe we should use `global.eval` when not in `__DEV__` mode?

piaskowyk: I agree with Tomek. 👍
Moreover, maybe we can use `location` and `sourceMap` only in debug mode. I could have a positive performance impact.

tjzel: We use it in debug only. `evalWithSourceMap` is injected only for dev builds.

tomekzaw (approved):
Thanks for optimizing this! 👍

piaskowyk (approved):

PR #4709, opened 2023-07-11
Minor adjustments to types in useAnimatedRef and related

tjzel:
## Summary

After a long research and with help from @Latropos I came to a conclusion that useAnimatedRef is properly typed, only the names of types were confusing. I changed those names and used directly `AnimatedRef` type which also was corrected accordingly since `useAnimatedRef` is supposed to hold components.

## Test plan

Just standard type tests suite.


src/reanimated2/hook/commonTypes.ts
```diff
@@ -13,12 +14,12 @@ export interface Descriptor {
...
+export interface RefObjectFunction<T extends Component> {
   current: T | null;
   (component?: T):
     | number // Paper
     | ShadowNodeWrapper // Fabric
     | HTMLElement; // web
 }
 
-export type AnimatedRef<T> = RefObjectFunction<T>;
+export type AnimatedRef<T extends Component> = RefObjectFunction<T>;
```
Latropos, line 25: Could we remove RefObjectFunction completely and have
```
export interface AnimatedRef<T extends Component> {
  current: T | null;
  (component?: T):
    | number // Paper
    | ShadowNodeWrapper // Fabric
    | HTMLElement; // web
}
```
instead?

Latropos (approved):
Looks good 🎉

src/reanimated2/hook/commonTypes.ts
tjzel: Yes we can!

PR #4710, opened 2023-07-11
Add duration support to springify

Latropos:
## Summary

Since we have added support for duration to our spring animations we should do the same thing with `.springify()` modifier.

Now you can use the following syntax:
```
        entering={SlideInLeft.delay(500).springify(5000).mass(2)}
        entering={SlideInLeft.delay(500).springify(5000)}
        entering={SlideInLeft.delay(500).springify(5000).dampingRatio(1)}
```

Although we use chaining ⛓️ instead of config to pass other spring props (like mass or stiffness) doing the same with duration would be unnatural as we specify duration of delay as a function argument. Also it could be confusing, since user could wonder if duration applies to spring or delay.

## Test plan
Tested on our [LA] Carousel exampel

https://github.com/software-mansion/react-native-reanimated/assets/56199675/fdd71b63-393a-418b-bd86-cb6ab5c72c85


tjzel (approved):
Looks good to me!
I like the for each addition.

piaskowyk (approved):
Good work! 👏

PR #4723, opened 2023-07-12
Remove ErrorHandler

tomekzaw:
## Summary

Currently, `ErrorHandler` is referenced in 11 files and has 2 separate implementations (one for Android and one for iOS) but it turns out that since Reanimated v3 there's only 1 place where we use it:

https://github.com/software-mansion/react-native-reanimated/blob/ae2b9ded42ec03589e1fccd1ecc8ad3d967d67c2/android/src/main/cpp/NativeProxy.cpp#L532-L535

This PR removes `ErrorHandler` completely from our codebase in favor of old simple `throw std::runtime_error`.

## Test plan

Check if CI is green.

### DIFF
--- a/Common/cpp/NativeModules/NativeReanimatedModule.h
+++ b/Common/cpp/NativeModules/NativeReanimatedModule.h
@@ -16,8 +16,13 @@
 #include "PlatformDepMethodsHolder.h"
 #include "RuntimeDecorator.h"
 #include "RuntimeManager.h"
-#include "Scheduler.h"
 #include "SingleInstanceChecker.h"
+
+#ifdef __APPLE__
+#include <RNReanimated/Scheduler.h>
+#else
+#include "Scheduler.h"
+#endif
 
 #ifdef RCT_NEW_ARCH_ENABLED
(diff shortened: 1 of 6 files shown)


tjzel (approved):
Works on andrewid 🎬

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -530,8 +523,7 @@ void NativeProxy::setupLayoutAnimations() {
               yogaValues.setProperty(rt, key, value);
             }
           } catch (std::invalid_argument e) {
-            errorHandler->setError("Failed to convert value to number");
-            errorHandler->raise();
+            throw std::runtime_error("Failed to convert value to number");
```
piaskowyk, line 526: Did you tried to trigger this exception in application? I'm wondering if it crash app or just display RedBox with message.

tomekzaw: Turns out it's actually shown using RedBox! 😄 

Zrzut ekranu 2023-07-12 o 22 29 43

PR #4724, opened 2023-07-12
Autocancel animation basing on toValue property instead of current, t…

### LINKED ISSUE #4721: Animations are automatically cancelled depending on current value from defineAnimation which may be unexpected
### Description

This code automatically cancels animation before onStart, if `current` is same as `_value`
```
 // prevent setting again to the same value
    // and triggering the mappers that treat this value as an input
    // this happens when the animation's target value(stored in animation.current until animation.onStart is called) is set to the same value as a current one(this._value)
    // built in animations that are not higher order(withTiming, withSpring) hold target value in .current
    if (sv._value === animation.current && !animation.isHigherOrder) {
      animation.callback && animation.callback(true);
      return;
    }
```

This may be unexpected for animations that don't have defined target in a similar manner that withSpring and withTiming do.
Imagine an animation of endless spinning wheel, where SharedValue represents current angle. Starting angle may be anything between 0 and 360, animation is cancellable only by `cancelAnimation` API:

```
export function withSpin(
  value: SharedValue<number>,
  config: SpinConfig,
  callback?: AnimationCallback,
): number {
  'worklet';

  return defineAnimation<SpinAnimation>(0, () => {
    'worklet';

    function spin(animation: InnerSpinAnimation, now: number): boolean {
      const {lastTimestamp, current} = animation;

      animation.current = current + (animation.velocity * deltaTime) / 1000;
      // start over from min when max reached
      animation.current = moduloClamp(animation.current, config.min, config.max);

      // never stop until manually cancelled
      return false;
    }

    function onStart(animation: SpinAnimation, value: number, now: Timestamp): void {
      animation.current = value;
    }
    return {
      onFrame: spin,
      onStart,
      callback: (finished, value) => {
        if (callback) {
          runOnJS(callback)(finished, value);
        }
      },
      // THIS IS PROBLEMATIC, REQUIRES HACKY SOLUTION
      current: value.value + 1,
      lastTimestamp: 0,
    } as SpinAnimation;
  });
}
```

I must use hacky `current: value.value + 1` to make it different than current value. In some animations, there is simply no defined target. withSpin is only simplified example of few animations of mine that encounter this.

I suggest we use `animation.toValue` instead of `animation.current`, for which I will prepare PR soon.

### Steps to reproduce

Use my animation above, change `current: value.value` to anything else, e.g. `0` or `value.value`.

https://snack.expo.dev/@drrefactor/reanimated-define-animation-repro

Press "unbreak" button to make animation work, press "break" to reproduce bug again. Reproduces on iOS and android, does not reproduce on web.

### Reanimated version

master

### React Native version

0.69

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

DrRefactor:
…o prevent unexpected cancels of custom animations (#4721)

## Summary

Fixes #4721 

Please see detailed description and snack in issue #4721 

## Test plan

Snack in #4721

### DIFF
--- a/src/reanimated2/valueSetter.ts
+++ b/src/reanimated2/valueSetter.ts
@@ -21,9 +21,8 @@ export function valueSetter(sv: any, value: any): void {
         : (value as AnimationObject);
     // prevent setting again to the same value
     // and triggering the mappers that treat this value as an input
-    // this happens when the animation's target value(stored in animation.current until animation.onStart is called) is set to the same value as a current one(this._value)
-    // built in animations that are not higher order(withTiming, withSpring) hold target value in .current
-    if (sv._value === animation.current && !animation.isHigherOrder) {
+    // this happens when the animation's target (animation.toValue) is set to the same value as a current one(this._value)
+    if (sv._value === animation.toValue && !animation.isHigherOrder) {
       animation.callback && animation.callback(true);
       return;
     }


omeid:
This will only prevent the callback, the underlying animation is still cancelled. See the [`cancelAnimation`](https://github.com/software-mansion/react-native-reanimated/blob/3.5.4/src/reanimated2/animation/util.ts#L494-L498) implementation.

Latropos:
Closing since we can't change this behaviour to avoid breaking apps of users of our library.

DrRefactor:
could you provide an example how would that change break animations? it seems to be a bug causing unexpected behavior, when these values "collide"

DrRefactor:
@omeid Did you look at snack provided in linked issue? Animation is not canceled when toValue differs from current value after changes from this PR

Latropos:
@DrRefactor We have deliberately decided to cancel animation in if the values collide. So you don't animate transition if START_POINT and END_POINT collide. This behaviour is important since animations may be accidentally triggered multiple times for various reasons.

Latropos:
@DrRefactor
>could you provide an example how would that change break animations? it seems to be a bug causing unexpected behavior, when these values "collide"

Since we provide function `defineAnimation` everything is possible. Imagine that sb defined the following animation - A view makes a 360 rotation 🔄 and then moves directly ➡️ to some predefined endpoint. Before the suggested change, if you trigger the animation twice nothing happens (since current value is equal to sv._value). After the suggested change, if you trigger the animation twice the animated object may make an extra rotation.

Latropos:
Yes, it is expected to be a better representation.

However when user defines custom `defineAnimation` he can do whatever he or she wants, so he may decide that `animation.current` and `animation.toValue` are not equal at all. And we would break their app.

Although you are right we can't merge it, since we never break the backwards compatibility if it is not necessary (and it is not, you can easily hack the solution).

DrRefactor:
I understand, that's the nature of delivering open source lib to community.
May I propose backward compatible fix that won't change current behavior, but allow non hacky implementations of animations like the one from snack example?

What if we added new property to animation, similar to `isHigherOrder`, that would prevent auto-canceling animations without defined target / toValue?

Latropos:
We haven't got more request for ignoring auto-canceling animations except for yours, so we suppose that this wouldn't be a commonly used feature. We wouldn't be interested in adding features that are used by a small number of people, since each of them has to be maintained and kept up to date.
Also I don't think that the approach you've presented is very "hacky". You only have to make sure that toValue and starting values differ. It doesn't seem to be much more difficult than adding an extra property.

Latropos:
cc @tomekzaw

DrRefactor:
Understood, thanks for discussion and explanation.

I've created PR after spending hours on debugging why my animation does not start, wanting to save community from going the same path.
Please consider documenting defineAnimation and this fact, it's not obvious without taking a thorough look into reanimated source code.

omeid:
> We haven't got more request for ignoring auto-canceling animations except for yours, so we suppose that this wouldn't be a commonly used feature. We wouldn't be interested in adding features that are used by a small number of people, since each of them has to be maintained and kept up to date.

We have used workarounds for this problem as well, so the issue is not that uncommon.

Latropos:
@omeid Would this change solve your case without hacking as well? Could you show us your example?

omeid:
It wouldn't change my use case no, my use case is based changing a value as it transition, based on the value of the transition at current point in time, sometimes, this results in the exact same value as the original, and instead of noop, it results in stopping the animation.

So what we do now is, read the value, compute the next value, cancel the animation, and reset the new value.

PR #4727, opened 2023-07-13
Refactor of isConfigured checks

jfedak:
## Summary

Previously, `core.ts` contained lots of redundant plugin checks. This PR fixes it by deleting `isConfigured` method since the same logic has been already implemented [here](https://github.com/software-mansion/react-native-reanimated/blob/cf3d1ca5d09a41dbf7be23df5316d39b98073a48/src/reanimated2/threads.ts#L144C1-L154C2).

## Test plan

Changes were tested with Example app on iOS, Android and Web.


src/reanimated2/Colors.ts
```diff
@@ -120,162 +120,160 @@ function parsePercentage(str: string): number {
...
-      tomato: 0xff6347ff,
-      turquoise: 0x40e0d0ff,
-      violet: 0xee82eeff,
-      wheat: 0xf5deb3ff,
-      white: 0xffffffff,
-      whitesmoke: 0xf5f5f5ff,
-      yellow: 0xffff00ff,
-      yellowgreen: 0x9acd32ff,
-    });
+const names: any = makeShareable({
```
tomekzaw, line 123: Does it work if we remove `any` here?
```suggestion
const names = makeShareable({
```

tomekzaw (approved):

Issue #4690, opened 2023-07-09
[android][release] makeShareableCloneRecursive Call Stack Size Exceeded


thespacemanatee:
### Description

I'll start by saying that I've spent many days on this and already looked at #4140, #4177, #4367 and #4475, and this is a separate issue. I also created an issue at expo/expo#23297 but I don't think it's expo's fault because I could not reproduce this on their bare template. It's worth noting that this does not occur on `react-native@0.71` or `expo@48`.

The bug, for me, only occurs on the Android `release` build or the `debug` build with the `JS Dev Mode` disabled.

![photo_2023-07-09_19-04-45](https://github.com/software-mansion/react-native-reanimated/assets/6837599/114b39bc-91ec-41d5-8cfe-6ed53bf56760)

This gave me the idea to search for all `__DEV__` flags within the installed version in my `node_modules` and flip all of them so that I can see the error in dev mode and get a nicer stacktrace. Here's what I got:

![photo_2023-07-09_19-07-21](https://github.com/software-mansion/react-native-reanimated/assets/6837599/d337f2ce-8d53-4c65-bba7-8f737ea8dcb6)

This narrows down the problem quite well, and I found that the following patch solves my issue:

```patch
diff --git a/node_modules/react-native-reanimated/src/reanimated2/shareables.ts b/node_modules/react-native-reanimated/src/reanimated2/shareables.ts
index 4f358f4..8be3e2b 100644
--- a/node_modules/react-native-reanimated/src/reanimated2/shareables.ts
+++ b/node_modules/react-native-reanimated/src/reanimated2/shareables.ts
@@ -1,6 +1,6 @@
 import NativeReanimatedModule from './NativeReanimated';
-import { ShareableRef } from './commonTypes';
 import { shouldBeUseWeb } from './PlatformChecker';
+import { ShareableRef } from './commonTypes';
 import { registerWorkletStackDetails } from './errors';
 import { jsVersion } from './platform-specific/jsVersion';
 
@@ -146,8 +146,9 @@ export function makeShareableCloneRecursive<T>(
               value.__workletHash,
               value.__stackDetails
             );
-            delete value.__stackDetails;
           }
+          // TODO: This is the fixes Maximum call stack size exceeded error
+          delete value.__stackDetails;
           // to save on transferring static __initData field of worklet structure
           // we request shareable value to persist its UI counterpart. This means
           // that the __initData field that contains long strings represeting the
```

[...]

PR #4732, opened 2023-07-13
Detected debug version of the worklet in release bundle

piaskowyk:
## Summary

Fixes: https://github.com/software-mansion/react-native-reanimated/issues/4690

It may happen when one of the user dependencies provided transpiled code with debug version of the Reanimated plugin. In effect, we have a debug version of the work in your release bundle. This might lead to unexpected issues or errors. 😢

I created issue with description how to fix this issue: https://github.com/software-mansion/react-native-reanimated/issues/4737

### DIFF
--- a/src/reanimated2/shareables.ts
+++ b/src/reanimated2/shareables.ts
@@ -150,10 +150,11 @@ export function makeShareableCloneRecursive<T>(
             );
             delete value.__stackDetails;
           } else if (value.__stackDetails) {
+            // Detected debug version of the worklet in your release bundle. This
+            // might lead to unexpected issues or errors. Probably one of user
+            // dependencies provided transpile code with debug version of the
+            // Reanimated plugin.
             delete value.__stackDetails;
-            console.warn(
-              `[Reanimated] Detected debug version of worklet in your release bundle. This might lead to unexpected issues or errors. Probably one of you dependencies provided transpiled code with debug version of Reanimated plugin.`
-            );
           }
           // to save on transferring static __initData field of worklet structure
           // we request shareable value to persist its UI counterpart. This means
--- a/src/reanimated2/shareables.ts
+++ b/src/reanimated2/shareables.ts
@@ -150,9 +150,9 @@ export function makeShareableCloneRecursive<T>(
             );
             delete value.__stackDetails;
           } else if (value.__stackDetails) {
-            // Detected debug version of the worklet in your release bundle. This
+            // Detected debug version of the worklet in release bundle. This
             // might lead to unexpected issues or errors. Probably one of user
-            // dependencies provided transpile code with debug version of the
+            // dependencies provided transpiled code with debug version of the
             // Reanimated plugin.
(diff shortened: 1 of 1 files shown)


src/reanimated2/shareables.ts
```diff
@@ -149,6 +149,12 @@ export function makeShareableCloneRecursive<T>(
               value.__stackDetails
             );
             delete value.__stackDetails;
+          } else if (value.__stackDetails) {
```
tomekzaw, line 152: I think we decided to use `__version` to determine that we have a dev bundle indeed but after some consideration `__stackDetails` indeed seems to be a better choice 👍

src/reanimated2/shareables.ts
```diff
@@ -149,6 +149,12 @@ export function makeShareableCloneRecursive<T>(
...
+            // Detected debug version of the worklet in release bundle. This
+            // might lead to unexpected issues or errors. Probably one of user
+            // dependencies provided transpiled code with debug version of the
+            // Reanimated plugin.
+            delete value.__stackDetails;
```
tomekzaw, line 157: Let's throw an error here to indicate that this configuration is not supported.
```suggestion
            throw new Error('[Reanimated] Using dev bundle in a release app build is not supported');
```

gstcarv: My app is crashing in production too due this issue https://github.com/software-mansion/react-native-reanimated/issues/4690. Is this error throw really necessary? Guess it can crash app again.

tomekzaw: We cannot guarantee that the app will work fine in release mode when using dev bundle so I guess it's better to fail immediately.

gstcarv: But there is any cases of release mode using dev bundle? How could I reproduce this scenario?

tomekzaw: It seems like https://github.com/software-mansion/react-native-reanimated/issues/4690#issuecomment-1634337256 and https://github.com/software-mansion/react-native-reanimated/issues/4690#issuecomment-1634344638 should answer your question.

src/reanimated2/shareables.ts
```diff
@@ -149,6 +149,14 @@ export function makeShareableCloneRecursive<T>(
               value.__stackDetails
             );
             delete value.__stackDetails;
+          } else if (value.__stackDetails) {
+            // Detected debug version of the worklet in release bundle. This
+            // might lead to unexpected issues or errors. Probably one of user
+            // dependencies provided transpiled code with debug version of the
+            // Reanimated plugin.
+            throw new Error(
+              '[Reanimated] Using dev bundle in a release app build is not supported.'
```
tomekzaw, line 158: We might want to give more information on how to troubleshoot this problem, i.e. clear cache or something.

piaskowyk: Sure thing 👍

tomekzaw (approved):

PR #4733, opened 2023-07-13
Mock version and source map in plugin snapshot tests

tomekzaw:
## Summary

I noticed that `version` and `sourceMap` properties are not included in plugin snapshots. This PR:

* renames environmental variables used for testing purposes
* unifies `compact` setting across app and tests and updates snapshots accordingly
* restores `version` property in snapshots but uses mock version "x.y.z" instead of real version
* moves `version` property to `__initData` to prevent allocating multiple strings (see #3930 for explanation)
* restores `sourceMap` property in `__initData` but uses mock source map for all snapshot tests
* adds one unit test that checks real `sourceMap` property
* adds double negation to fix return type of `isRelease` (previously it returned `boolean | string | ""`)


plugin/src/buildWorkletString.ts
```diff
@@ -84,7 +86,7 @@ export function buildWorkletString(
 
   const transformed = transformSync(code, {
     plugins: [prependClosureVariablesIfNecessary(closureVariables)],
-    compact: !includeSourceMap,
+    compact: true,
```
tomekzaw, line 89: It's too late to think, please somebody confirm if we can leave `true` here.

piaskowyk: If I remember correctly, this change was required for proper source mapping for Hermes debugger. When we use `compact: true` then some debugger breakpoints will be shifted by some random number of lines.

__tests__/plugin.test.ts
```diff
@@ -72,7 +72,21 @@ describe('babel plugin', () => {
...
+
+      const input = html`<script>
+        function foo() {
+          'worklet';
+          var foo = 'bar';
+        }
+      </script>`;
+
+      const { code } = runPlugin(input, { sourceMaps: true });
+      expect(code).toContain('{\\"version\\":3');
```
piaskowyk, line 89: Why do we check here if the `version` property exists instead of checking that the `sourceMap` property exists?

__tests__/plugin.test.ts
piaskowyk, line 1: Now we are testing debug version of the plugin, maybe we should also add tests for the release version of the plugin. What do you think?

tjzel: There are tests that compare selected behaviours for both environments.

plugin/src/buildWorkletString.ts
```diff
@@ -31,6 +31,8 @@ import * as fs from 'fs';
 import { isRelease } from './utils';
 import { WorkletizableFunction } from './types';
 
+const MOCK_SOURCE_MAP = 'mock source map';
```
tjzel, line 34:
```suggestion
const MOCKED_SOURCE_MAP = 'mocked source map';
```

tjzel (changes requested):

__tests__/plugin.test.ts
tomekzaw: Changed in 47e279f8d13fb32ac4c17cc99d999c19ab4b4b7f

__tests__/plugin.test.ts
piaskowyk: I can't find them. Could you send me a link to this test?

tjzel: https://github.com/software-mansion/react-native-reanimated/blob/main/__tests__/plugin.test.ts#L1386

not many of them so I guess this is a valid concern. 

We also have things like this
https://github.com/software-mansion/react-native-reanimated/blob/main/__tests__/plugin.test.ts#L75

which just need their respective counterpart.

piaskowyk: Thanks! Glad to hear that we have test to check it but I think we could add a couple next of test cases. But probably this is out of this PR so we can do it on another.

__tests__/plugin.test.ts
```diff
@@ -62,17 +62,32 @@ describe('babel plugin', () => {
...
           var foo = 'bar';
         }
       </script>`;
 
-      const { code } = runPlugin(input, {});
-      expect(code).toContain(`f.__version = "${packageVersion}";`);
+      const { code } = runPlugin(input);
+      expect(code).toContain('sourceMap: "{');
+      expect(code).toContain(
+        '\\"mappings\\":\\"AACQ,SAAAA,GAASA,CAAA,CAAG,CAEV,GAAI,CAAAA,GAAG,CAAG,KAAK,CACjB\\"'
```
tjzel: A comment on this line would be nice.

plugin/src/buildWorkletString.ts
tjzel: Actually nevermind, everywhere else we just have MOCK

__tests__/plugin.test.ts
```diff
@@ -821,7 +836,7 @@ describe('babel plugin', () => {
       </script>`;
 
       const { code } = runPlugin(input);
-      expect(code).toContain('...{b:2,c:3}');
+      expect(code).toContain('...{\\n      b: 2,\\n      c: 3\\n    }');
```
tjzel, line 839: why is this necessary? it was working previously

__tests__/plugin.test.ts
```diff
@@ -62,17 +62,32 @@ describe('babel plugin', () => {
     });
 
     it('injects its version', () => {
```
tjzel, line 64: Some of those tests could be moved to 'for debugging' section. Also we could add a test that checks for version not being present in release builds.

tjzel (changes requested):

plugin/src/buildWorkletString.ts
tomekzaw: Turns out we already effectively use `compact: true` when running an app. I've just checked it and breakpoint seem to work fine after this change.

tjzel (approved):

piaskowyk (approved):

PR #4734, opened 2023-07-14
iOS scheduler optimisation fix

michalmaka:
## Summary

I've noticed that `weak_ptr` for `runtimeManager` in iOS scheduler has executed `lock` method but returned `shared_ptr` is not used and thus deallocation immediately. In that case, if we just want to know whether `runtimeManager` `shared_ptr` is alive it's more performant to use `expired` method on `weak_ptr`.

I removed also some dead code:

- in `scheduleOnUI` there was unnecessary block of code checking if it's executed on main thread. If it is on main thread it returns from function several lines above the block and won't be executed.
- unnecessary virtual empty destructor for `REAIOSScheduler`. Base class has already virtual destructor and marking this again does not change anything. 

## Test plan

Check iOS application if works fine.

### DIFF
--- a/Common/cpp/Tools/Scheduler.h
+++ b/Common/cpp/Tools/Scheduler.h
@@ -83,7 +83,7 @@ class Scheduler {
   std::atomic<bool> scheduledOnUI{};
   Queue<std::function<void()>> uiJobs;
   std::shared_ptr<facebook::react::CallInvoker> jsCallInvoker_;
-  std::weak_ptr<RuntimeManager> runtimeManager;
+  std::weak_ptr<RuntimeManager> weakRuntimeManager;
 };
 
 } // namespace reanimated
--- a/Common/cpp/Tools/Scheduler.h
+++ b/Common/cpp/Tools/Scheduler.h
@@ -83,7 +83,7 @@ class Scheduler {
   std::atomic<bool> scheduledOnUI{};
   Queue<std::function<void()>> uiJobs;
   std::shared_ptr<facebook::react::CallInvoker> jsCallInvoker_;
-  std::weak_ptr<RuntimeManager> weakRuntimeManager;
+  std::weak_ptr<RuntimeManager> weakRuntimeManager_;
 };
 
 } // namespace reanimated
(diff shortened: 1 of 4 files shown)


ios/native/REAIOSScheduler.h
```diff
@@ -13,7 +13,6 @@ class REAIOSScheduler : public Scheduler {
  public:
   REAIOSScheduler(std::shared_ptr<CallInvoker> jsInvoker);
   void scheduleOnUI(std::function<void()> job) override;
-  virtual ~REAIOSScheduler();
```
tomekzaw, line 16: `REAIOSScheduler` extends `Scheduler` which has virtual methods so it needs to have a virtual destructor as well. We can move the implementation from .cpp to .h file if possible.

ios/native/REAIOSScheduler.mm
```diff
@@ -13,24 +13,19 @@
...
     return;
   }
 
   Scheduler::scheduleOnUI(job);
-  if ([NSThread isMainThread]) {
-    if (runtimeManager.lock()) {
-      triggerUI();
-    }
-    return;
-  }
```
tomekzaw, line 33: Indeed, this block is no longer relevant (confirmed with @kmagiera).

tomekzaw (changes requested):

ios/native/REAIOSScheduler.mm
```diff
@@ -13,24 +13,19 @@
 
 void REAIOSScheduler::scheduleOnUI(std::function<void()> job)
 {
-  if (runtimeManager.lock() == nullptr) {
+  const auto runtimeManagerExpired = runtimeManager.expired();
```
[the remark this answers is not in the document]
michalmaka: done

Common/cpp/Tools/Scheduler.h
```diff
@@ -83,7 +83,7 @@ class Scheduler {
   std::atomic<bool> scheduledOnUI{};
   Queue<std::function<void()>> uiJobs;
   std::shared_ptr<facebook::react::CallInvoker> jsCallInvoker_;
-  std::weak_ptr<RuntimeManager> runtimeManager;
+  std::weak_ptr<RuntimeManager> weakRuntimeManager;
```
tomekzaw, line 86:
```suggestion
  std::weak_ptr<RuntimeManager> weakRuntimeManager_;
```

tomekzaw (approved):
Thanks for this PR! 👏

piaskowyk (approved):

PR #4744, opened 2023-07-17
Fix modal SET progress animation on Android

piaskowyk:
## Summary

This PR adds missing cleanup and reparenting of shared views after modal close.

## Test plan

|Before|After|
|---|---|
| <video src="https://github.com/software-mansion/react-native-reanimated/assets/36106620/79852a34-44c6-48ab-8d83-b460963c154d" /> | <video src="https://github.com/software-mansion/react-native-reanimated/assets/36106620/679ff9ab-3e34-4395-9640-c0dac36be243" /> |

### DIFF
--- a/src/reanimated2/layoutReanimation/sharedTransitions/ProgressTransitionManager.ts
+++ b/src/reanimated2/layoutReanimation/sharedTransitions/ProgressTransitionManager.ts
@@ -72,8 +72,19 @@ export class ProgressTransitionManager {
           global.ProgressTransitionRegister.onTransitionEnd();
         }
       );
-      if (Platform.OS === 'ios') {
-        // required by modals
+
+      if (Platform.OS === 'android') {
+        // onFinishTransitioning event is available only on Android and
+        // is used to handle closing modals
+        eventHandler.onDisappear = registerEventHandler(
+          'onFinishTransitioning',
+          () => {
+            'worklet';
+            global.ProgressTransitionRegister.onAndroidFinishTransitioning();
+          }
+        );
+      } else if (Platform.OS === 'ios') {
+        // topDisappear event is required to handle closing modals on iOS
         eventHandler.onDisappear = registerEventHandler('topDisappear', () => {
           'worklet';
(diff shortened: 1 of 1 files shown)


src/reanimated2/layoutReanimation/sharedTransitions/ProgressTransitionManager.ts
```diff
@@ -149,6 +160,13 @@ function createProgressTransitionRegister() {
         progressAnimation!(viewTag, snapshot, progress);
       }
     },
+    onAndroidFinishTransitioning: () => {
+      if (toRemove.size === 0) {
```
tjzel, line 164: Let's make if conditional here same is in `onTransitionEnd` (toRemove.size > 0 and reverse order of clauses)

tjzel (changes requested):

src/reanimated2/layoutReanimation/sharedTransitions/ProgressTransitionManager.ts
```diff
@@ -72,8 +72,19 @@ export class ProgressTransitionManager {
...
+
+      if (Platform.OS === 'android') {
+        // onFinishTransitioning event is available only on Android and
+        // is used to handle closing modals
+        eventHandler.onDisappear = registerEventHandler(
+          'onFinishTransitioning',
+          () => {
+            'worklet';
+            global.ProgressTransitionRegister.onAndroidFinishTransitioning();
+          }
```
tomekzaw, line 84: Is it possible to pass this function directly here?
```suggestion
          global.ProgressTransitionRegister.onAndroidFinishTransitioning,
```

piaskowyk: I will check it!

piaskowyk: Unfortunately, it doesn't work because this lambda is created on the JS thread, where `global.ProgressTransitionRegister` is unavailable :/

tomekzaw (approved):

PR #4745, opened 2023-07-17
Make _WORKLET a part of public API for diagnostics and related code cleanup

tjzel:
## Summary

We officially mark `_WORKLET` as a diagnostic tool not to be relied on.

## Test plan

Just run Examples that are were using `_WORKLET` before.


tjzel:
While it's indeed used as our internal API we have it embedded in our docs (for more reference [here](https://docs.swmansion.com/react-native-reanimated/docs/api/nativeMethods/measure/)) and it's used in one of our non-technical Examples. Moreover it's used sometimes by our library consumers. If you want we could probably open a GitHub discussion or ask people on some discords whether they use it or not and then decide what to do about it.

src/reanimated2/publicGlobals.ts
```diff
@@ -0,0 +1,17 @@
...
+  // it used to be necessary in the past for some of the
+  // functionalities of react-native-reanimated to work
+  // properly but it's no longer the case and your code
+  // shouldn't depend on it, we keep it here
+  // mainly for backward compatibility reasons for our users.
+  var _WORKLET: boolean | undefined;
+
+  // This is simply an address of UI Runtime. Its type stems from
+  // the need to be able to pass it from the native side to JS
+  // when it's a 64-bit address value.
```
tomekzaw, line 15: Thanks for adding this comment!

tomekzaw (approved):
❤️

PR #4747, opened 2023-07-18
Improve detecting production bundle in Babel plugin

tomekzaw:
## Summary

This PR changes the way how Reanimated Babel plugin detects if dev or prod bundle should generated. Previously, we would only support `BABEL_ENV=production` and `BABEL_ENV=release`; all other options would fallback to dev bundle. Thanks to this change we will also support other configurations e.g. `productionRelease` or `staging`. See #4690 for details.


piaskowyk (approved):
👏

a-eid:
@tomekzaw @piaskowyk how does this work with expo, given that `NODE_ENV=test` ?

tomekzaw:
@a-eid Can you please specify where exactly `NODE_ENV` is set to `test`? Do you think we should prepare for a dev or a prod bundle in such case? cc @bartlomiejbloniarz

a-eid:
@tomekzaw  `NODE_ENV` is sent to `test` in `eas.json` so that it would load env variables from `.env.test`. in such case I think a prod bundle should be loaded.

PR #4748, opened 2023-07-18
Add a `reduceMotion` prop to animation functions

bartlomiejbloniarz:
## Summary

Animations can cause some people to feel unwell. In order to help developers create apps with those people in mind this PR adds a reduceMotion prop to animation functions that allows developers to configure whether a given animation should follow the system accessibility setting. Possible values are `system`, `always` and `never`. The prop defaults to `system` and is passed down to child animations, unless they override it. 

When reduced motion is enabled:
- `withSpring` and `withTiming` return the `toValue` immediately
- `withDecay` returns the current value immediately taking into account the clamp parameter
- `withDelay` starts the next animation immediately
- `withRepeat`:
    - If the numOfReps is infinite (<=0) or even and the animation is reversed, then the repeated animation does not start and the startValue is returned
    - otherwise the repeated animation is run once
- `withSequence` starts only the animations that have reduce motion disabled (the last animation is also started, so the sequence ends on the value it's supposed to)

This PR also adds changes to the Example App to reduce some animations unrelated to reanimated.

## Test plan
Test the behavior of the ReducedMotion Example (and other examples) with the accessibility setting on and off.


src/reanimated2/animation/util.ts
```diff
@@ -23,9 +24,14 @@ import type {
   AnimatableValueObject,
 } from '../commonTypes';
 import NativeReanimatedModule from '../NativeReanimated';
+import { isWeb } from '../PlatformChecker';
 
 let IN_STYLE_UPDATER = false;
 
+const IS_REDUCED_MOTION = isWeb()
+  ? !window.matchMedia('(prefers-reduced-motion: no-preference)').matches
+  : _REANIMATED_IS_REDUCED_MOTION ?? false;
```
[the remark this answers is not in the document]
bartlomiejbloniarz: I didn't want to call a hook outside of a react component. Maybe I should move this logic to a function in `reanimated2/utils.ts` and make both the hook and this part call it, so there will be no code duplication.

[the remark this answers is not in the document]
tomekzaw: Yup, let's do it

src/reanimated2/animation/sequence.ts
```diff
@@ -1,29 +1,56 @@
...
     _animations[0] as SequenceAnimation,
     () => {
       'worklet';
+      let reduceMotion: ReducedMotionConfig | undefined;
+
+      if (typeof _reduceMotion === 'string') {
+        reduceMotion = _reduceMotion as ReducedMotionConfig;
+      } else {
+        _animations.unshift(_reduceMotion as NextAnimation<AnimationObject>);
+      }
```
piaskowyk, line 36: I think that we need comment here to explain what is happening here

src/reanimated2/animation/util.ts
```diff
@@ -270,6 +297,17 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
     timestamp: Timestamp,
     previousAnimation: Animation<AnimationObject>
   ) => {
+    if (animation.reduceMotion === undefined) {
+      animation.reduceMotion = shouldReduceMotion('system');
+    }
+    if (animation.reduceMotion) {
+      animation.current = animation.toValue;
```
piaskowyk, line 304: Did you tested colors animations?

bartlomiejbloniarz: I tested the Animate Colors example and it worked fine

src/reanimated2/animation/util.ts
```diff
@@ -73,6 +75,15 @@ function recognizePrefixSuffix(value: string | number): RecognizedPrefixSuffix {
   }
 }
 
+export function shouldReduceMotion(config?: ReducedMotionConfig) {
```
tomekzaw, line 78: As discussed offline, it took me a while to understand what this function exactly does. Also, throughout the examples we often use `shouldReduceMotion` as a boolean. Can we think of a better name for this function?
```suggestion
export function shouldReduceMotionForConfig(config?: ReducedMotionConfig) {
```

src/reanimated2/animation/util.ts
```diff
@@ -383,6 +406,22 @@ function decorateAnimation<T extends AnimationObject | StyleLayoutAnimation>(
...
     previousAnimation: Animation<AnimationObject>
   ) => {
+    if (animation.reduceMotion === undefined) {
+      animation.reduceMotion = shouldReduceMotion('system');
+    }
+    if (animation.reduceMotion) {
+      if (animation.toValue !== undefined) {
+        animation.current = animation.toValue;
+      } else {
+        // if there is no toValue, then the base function is responsible for setting the current value
```
tomekzaw, line 416:
```suggestion
        // if there is no `toValue`, then the base function is responsible for setting the current value
```

src/reanimated2/animation/repeat.ts
```diff
@@ -84,7 +95,16 @@ export const withRepeat = function <T extends AnimationObject>(
       ): void {
         animation.startValue = value;
         animation.reps = 0;
-        nextAnimation.onStart(nextAnimation, value, now, previousAnimation);
+
+        if (nextAnimation.reduceMotion === undefined) {
+          nextAnimation.reduceMotion = animation.reduceMotion;
```
tomekzaw, line 100: Let's add a comment here that explains the logic.

src/reanimated2/animation/repeat.ts
```diff
@@ -34,7 +36,13 @@ export const withRepeat = function <T extends AnimationObject>(
           ? _nextAnimation()
           : _nextAnimation;
 
+      const isEven = reverse && (numberOfReps <= 0 || numberOfReps % 2 === 0);
```
tomekzaw, line 39:
```suggestion
      const isEvenNumberOfReps = reverse && (numberOfReps <= 0 || numberOfReps % 2 === 0);
```

app/src/examples/ReducedMotionExample.tsx
```diff
@@ -1,61 +1,236 @@
...
+) {
+  return examples.map((example, i) => {
+    return (
+      <Example key={i} animation={example.animation} text={example.text} />
     );
   });
+}
+
+function Example(props: { animation: number; text: string }) {
+  const sv = useSharedValue(initialValue);
```
tomekzaw, line 156: Let's add a newline here:
```suggestion
  const sv = useSharedValue(initialValue);

```

app/src/examples/ReducedMotionExample.tsx
```diff
@@ -1,61 +1,236 @@
...
+    title: 'Decay',
+    component: <DecayExample />,
+  },
+  {
+    title: 'Config',
+    component: mapExamples(CONFIG_EXAMPLES),
+  },
+];
+
+export default function App() {
```
tomekzaw, line 110:
```suggestion
export default function ReducedMotionExample() {
```

tomekzaw (changes requested):

src/reanimated2/animation/util.ts
```diff
@@ -73,6 +75,15 @@ function recognizePrefixSuffix(value: string | number): RecognizedPrefixSuffix {
   }
 }
 
+export function shouldReduceMotion(config?: ReducedMotionConfig) {
+  'worklet';
+  if (config === undefined) {
+    return undefined;
+  }
```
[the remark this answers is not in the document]
bartlomiejbloniarz: Since I want the child animation to inherit the setting from its parent I made this function return `undefined` when the setting is not defined. Then when the animation is created I can just assign `shouldReduceMotion(config.reduceMotion)` to `animation.reduceMotion` and the parent will only assign its setting to the child, when the child has it set to undefined. I will try to make this less confusing.

src/reanimated2/utils.ts
```diff
@@ -2,6 +2,11 @@ import type { Component } from 'react';
 import { measure } from './NativeMethods';
 import type { AnimatedRef } from './hook/commonTypes';
 import type { SharedValue } from './commonTypes';
+import { isWeb } from './PlatformChecker';
+
+export const IS_REDUCED_MOTION = isWeb()
+  ? !window.matchMedia('(prefers-reduced-motion: no-preference)').matches
+  : global._REANIMATED_IS_REDUCED_MOTION ?? false;
```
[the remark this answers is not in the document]
bartlomiejbloniarz: I think placing it there makes sense. It changes a bit the meaning of `PlatformChecker`, but I think it's the most fitting place for this function.

app/src/examples/ReducedMotionExample.tsx
```diff
@@ -1,61 +1,246 @@
...
+}
+
+function mapExamples(
+  examples: {
+    animation: number;
+    text: string;
+  }[]
+) {
+  return (
+    <View key={exampleNumber++}>
```
tomekzaw, line 154: Not really happy about this

src/reanimated2/animation/repeat.ts
```diff
@@ -34,7 +36,14 @@ export const withRepeat = function <T extends AnimationObject>(
           ? _nextAnimation()
           : _nextAnimation;
 
+      const isEvenNumberOfReps =
+        reverse && (numberOfReps <= 0 || numberOfReps % 2 === 0);
```
tomekzaw, line 40: I know that I suggested renaming this variable but actually the name doesn't represent the exact logic here. What about simply removing this variable and moving the condition to the conditional in line 43?

bartlomiejbloniarz: This value is used also in the line 106, because it is needed in both `onFrame` and `onStart`. But I can do `animation.onFrame = () => {return true;}` when the condition is met in `onStart`. Then we can inline this value and retain the same logic.

src/reanimated2/animation/sequence.ts
```diff
@@ -1,29 +1,58 @@
...
+export function withSequence<T extends AnimatableValue>(
+  _reduceMotion: ReducedMotionConfig,
+  ...animations: T[]
+): T;
+
+export function withSequence<T extends AnimatableValue>(...animations: T[]): T;
 
-export const withSequence = function (
+export function withSequence(
+  _reduceMotion: ReducedMotionConfig | NextAnimation<AnimationObject>,
```
tomekzaw, line 22: 1. Can we rename the variable so that it also mentions that this may be the first animation?
2. Does `withSequence()` (i.e. empty sequence) still work after these changes?
```suggestion
  _reduceMotionOrFirstAnimation: ReducedMotionConfig | NextAnimation<AnimationObject>,
```

tomekzaw (commented):
Left a few minor suggestions related to naming stuff but we're really close to getting this merged.

app/src/examples/SharedElementTransitions/Profiles.tsx
```diff
@@ -504,10 +505,24 @@ export default function ProfilesExample() {
...
   return (
     <Stack.Navigator screenOptions={{ headerShown: false }}>
-      <Stack.Screen name="Profiles" component={ProfilesScreen} />
-      <Stack.Screen name="Home" component={HomeScreen} />
+      <Stack.Screen
+        name="Profiles"
+        component={ProfilesScreen}
+        options={{
+          animation: shouldReduceMotion ? 'fade' : 'default',
+        }}
```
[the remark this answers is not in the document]
bartlomiejbloniarz: It does!

src/reanimated2/animation/sequence.ts
```diff
@@ -1,29 +1,58 @@
...
         return result;
       });
-      const firstAnimation = animations[0];
+
+      function nextIndex(index: number) {
+        while (
+          index < animations.length - 1 &&
+          animations[index].reduceMotion
+        ) {
+          index += 1;
```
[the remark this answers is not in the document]
bartlomiejbloniarz: It does

src/reanimated2/animation/sequence.ts
bartlomiejbloniarz: Empty sequence doesn't work, but it also didn't work before, since `animations[0]` is used. After some offline discussions I think it would be good to change it so the app doesn't crash when it's used, but also add some warning for the users.

src/reanimated2/animation/sequence.ts
```diff
@@ -1,29 +1,77 @@
...
+      _animations.unshift(
+        _reduceMotionOrFirstAnimation as NextAnimation<AnimationObject>
+      );
+    } else {
+      reduceMotion = _reduceMotionOrFirstAnimation as ReduceMotion;
+    }
+  }
+
+  if (_animations.length === 0) {
+    console.warn('No animation was provided for the sequence');
```
tomekzaw, line 38: Do you think it makes sense to include library tag in the warning message?
```suggestion
    console.warn('[Reanimated] No animation was provided for the sequence');
```

src/reanimated2/animation/sequence.ts
```diff
@@ -1,29 +1,77 @@
...
+  // the first argument is either a config or an animation
+  // this is done to allow the reduce motion config prop to be optional
+  if (_reduceMotionOrFirstAnimation) {
+    if (typeof _reduceMotionOrFirstAnimation !== 'string') {
+      _animations.unshift(
+        _reduceMotionOrFirstAnimation as NextAnimation<AnimationObject>
+      );
+    } else {
+      reduceMotion = _reduceMotionOrFirstAnimation as ReduceMotion;
+    }
```
tomekzaw, line 34: Since we know that the runtime type of `ReduceMotion` is `string`, would it make sense to swap `if` and `else` here?

tomekzaw (approved):
Overall the PR looks good to me. 👍 

I've left two minor suggestions in the comments. 💬

piaskowyk (approved):
👏

efstathiosntonas:
Hey guys, thank you for this really important prop, just a request: make this a breaking change, took me a while to figure out why animations were not working on nightly!

bartlomiejbloniarz:
Did they not work when `Reduced Motion` was enabled or was there an overall issue with animations?

efstathiosntonas:
@bartlomiejbloniarz it works as expected. I am one of "these" users who have reduced motion enabled on iOS and I was trying to figure why animations were disabled 😀

efstathiosntonas:
This setting on ios is set and forget, that's why it took me a while to find the "issue".

bartlomiejbloniarz:
@efstathiosntonas today we also merged #4884, so you can also expect layout animations to change their default behaviour on `Reduced Motion` 😅

efstathiosntonas:
@bartlomiejbloniarz is there to set `ReduceMotion.XXXX` globally? I use many npm modules that use reanimated and it would be a pita to patch them.

bartlomiejbloniarz:
@efstathiosntonas There is no `official` way to achieve that. 🫤

efstathiosntonas:
@bartlomiejbloniarz thanks for the update, I'll just patch it then

Issue #4736, opened 2023-07-14
Mismatch between JavaScript code version and Reanimated Babel plugin version


Andrerm124:
### Description

Getting the error: ` [Reanimated] Mismatch between JavaScript code version and Reanimated Babel plugin version (3.3.0 vs. undefined)` consistently when building for Android and iOS in Dev. 

I checked the babel plugin and it is correctly fetching the version from the package.json. It's also correctly assigning the __version value to **most** function/objects

I did some digging and found that when attempting to register the event handler `onGestureHandlerStateChange` (/platform-specific/core.ts:127) it throws the error due to 4 functions/objects that aren't being "workletized" via the babel plugin:
- onStart
- handlers
- _closure
- eventHandler

I've been searching around on this problem for what feels like an eternity now and I can't for the life of me figure out what's causing it.

### Steps to reproduce

1. Start metro with a cleaned cache
2. Open the app
3. Wait for the bundle to load
4. Immediately get the error

### Snack or a link to a repository

Private Repository

### Reanimated version

3.3.0

### React Native version

0.72.0

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

efstathiosntonas:
try to start the bundler with `--reset-cache` switch

Andrerm124:
@efstathiosntonas I've tried every possible way of clearing caches that I can think of. So far nothing has worked 😬

tomekzaw:
Hey @Andrerm124, thanks for reporting this issue. Can you please paste the `dependencies` list from your package.json? We suspect that some third-party library might be already distributed in a workletized form which causes the error.

Andrerm124:
@tomekzaw Hey, sorry for the delay, github notifications don't seem to be coming through 😬 
Here are the list of dependencies we're using in the app:

```"dependencies": {
  "@algolia/client-search": "4.17.1",
  "@expo/react-native-action-sheet": "4.0.1",
  "@fullstory/react-native": "1.1.2",
  "@fullstory/segment-react-native-plugin-fullstory": "1.0.1",
  "@gorhom/portal": "1.0.14",
  "@miblanchard/react-native-slider": "2.1.0",
  "@peppy-health/core": "2.23.0",
  "@react-native-async-storage/async-storage": "1.18.2",
[...]

PR #4749, opened 2023-07-18
Throw error when worklet version is unknown

tomekzaw:
## Summary

This PR adds a developer-friendly error when worklet version is undefined. Related to #4736.

### DIFF
--- a/src/reanimated2/shareables.ts
+++ b/src/reanimated2/shareables.ts
@@ -141,9 +141,13 @@ export function makeShareableCloneRecursive<T>(
           // we are converting a worklet
           if (__DEV__) {
             const babelVersion = value.__initData.version;
-            if (babelVersion !== jsVersion) {
+            if (babelVersion === undefined) {
+              throw new Error(
+                '[Reanimated] Unknown version of Reanimated Babel plugin. Using release bundle with debug build of the app is not supported.'
+              );
+            } else if (babelVersion !== jsVersion) {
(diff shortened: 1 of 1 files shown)


src/reanimated2/shareables.ts
```diff
@@ -141,9 +141,13 @@ export function makeShareableCloneRecursive<T>(
...
             const babelVersion = value.__initData.version;
-            if (babelVersion !== jsVersion) {
+            if (babelVersion === undefined) {
+              throw new Error(
+                '[Reanimated] Unknown version of Reanimated Babel plugin. Using release bundle with debug build of the app is not supported.'
+              );
+            } else if (babelVersion !== jsVersion) {
               throw new Error(`[Reanimated] Mismatch between JavaScript code version and Reanimated Babel plugin version (${jsVersion} vs. ${babelVersion}). Please clear your Metro bundler cache with \`yarn start --reset-cache\`,
-              \`npm start -- --reset-cache\` or \`expo start -c\` and run the app again.`);
+              \`npm start -- --reset-cache\` or \`expo start -c\` and run the app again. If the issue still persists, make sure that none of your dependencies contains already transformed worklets bundled with an outdated version of the Reanimated Babel plugin.`);
```
Andrerm124, line 150: I think this message should also be supplied in the above `if` statement (`babelVersion === undefined`), as older versions of the babel plugin might return `undefined` as the `babelVersion` (Which was the case for me previously)

piaskowyk (approved):

tjzel (approved):

PR #4751, opened 2023-07-18
Unify `swizzleMethod`

piaskowyk:
## Summary

This PR combines the swizzling methods in the codebase. Before, we had two `swizzleMethod` functions that had the exact same implementation.

## Test plan

Open example with SET


tomekzaw (approved):
👏

PR #4764, opened 2023-07-19
Fix typos in withSpring internal code

kacperkapusciak:
## Summary

Fixed a couple of typos in withSpring internal code

## Test plan

It's not part of the public API so it should be cool, right?


piaskowyk (approved):
🙌

PR #4767, opened 2023-07-19
[SET] Add border & progress transition example

kacperkapusciak:
## Summary

This PR adds an code example that's using the new features available in Shared Element Transitions since React Native Reanimated 3.4.0

1. Animating border radius
2. Progress transition on iOS

### iOS

https://github.com/software-mansion/react-native-reanimated/assets/39658211/c5308669-4609-4796-82aa-2951adbeb5d2

### Android

https://github.com/software-mansion/react-native-reanimated/assets/39658211/3d3532cb-1705-4a6e-8a61-33d5308e9f75


app/src/examples/index.ts
```diff
@@ -312,6 +313,11 @@ export const EXAMPLES: Record<string, Example> = {
     title: 'Volume slider & sensor',
     screen: VolumeExample,
   },
+  TransitionProgressExample: {
+    icon: '☕',
+    title: 'Progress transition example',
+    screen: ProgressTransitionExample,
```
tomekzaw, line 319: `TransitionProgress` or `ProgressTransition`?

app/src/examples/SharedElementTransitions/ProgressTransition.tsx
```diff
@@ -0,0 +1,322 @@
...
+  createNativeStackNavigator,
+  NativeStackScreenProps,
+} from '@react-navigation/native-stack';
+import Animated, {
+  FadeInDown,
+  FadeInLeft,
+  FadeInUp,
+} from 'react-native-reanimated';
+
+import { useNavigation } from '@react-navigation/native';
```
tomekzaw: never used

app/src/examples/SharedElementTransitions/ProgressTransition.tsx
```diff
@@ -0,0 +1,322 @@
...
+// Image by <a href="https://www.freepik.com/free-photo/coffee-cup-wooden-boards-arrangement_29301195.htm#&position=27&from_view=collections">Freepik</a>
+
+const windowWidth = Dimensions.get('window').width;
+
+type StackParamList = {
+  Home: undefined;
+  Details: { tag: string };
+};
+
+const data = [
```
tomekzaw, line 44:
```suggestion
const DATA = [
```

app/src/examples/SharedElementTransitions/ProgressTransition.tsx
```diff
@@ -0,0 +1,322 @@
...
+  },
+];
+
+const Stack = createNativeStackNavigator<StackParamList>();
+
+function HomeScreen({
+  navigation,
+}: NativeStackScreenProps<StackParamList, 'Home'>) {
+  return (
+    <Animated.View style={[homeStyles.background, { paddingTop: 64 }]}>
```
tomekzaw, line 74: plz don't use inline styles

app/src/examples/index.ts
kacperkapusciak: There's no consistency on that matter in the codebase

image
image

@piaskowyk to which naming should we stick to?

app/src/examples/SharedElementTransitions/ProgressTransition.tsx
kacperkapusciak: Fixed in 03154e2

app/src/examples/SharedElementTransitions/ProgressTransition.tsx
kacperkapusciak: Fixed in 03154e2

app/src/examples/index.ts
piaskowyk: `TransitionProgressEvent` - this is originally event name comes from `rn-screens`

`ProgressTransition` - this is shared element transition bases on progres value that comes from screen's `TransitionProgressEvent`

In summary we should stick with `ProgressTransition` I think.

kacperkapusciak: Alright, I'll try to keep the "progress transition" name consistent when creating docs for this in the future as well

Fixed in 9b42b9e

piaskowyk (approved):
👏
Well done!

PR #4768, opened 2023-07-19
Move `JSRuntimeHelper` to a separate file

tomekzaw:
## Summary

This PR moves `JSRuntimeHelper` class (along with `CoreFunction` class) to a separate file in order to decouple it from `Shareables.h`. Eventually, we will unify `JSRuntimeHelper` and `RuntimeManager` into a single class so this file will be deleted in the future.


tjzel (approved):
:clap:

piaskowyk (approved):

PR #4772, opened 2023-07-19
Add ts-prune

Latropos:
## Summary

* Ts-prune allows us to detect unused typescript code. 
* Advantages:
  *  We can remove exports of all the functions and interfaces exported for no reason. This makes importing with cmd+enter shortcut easier, since you are less likely to get wrong suggestions from vscode. 
  * We can see all the “forgotten” areas of code, and think about reasons to have them.

> **Note**
>I've removed the file `MapperRegistry.ts` which seems unnecessary since this PR: [#3722](https://github.com/software-mansion/react-native-reanimated/pull/3722)

## Test plan

Just CI


src/reanimated2/__mocks__/NativeReanimated.ts
```diff
@@ -6,6 +6,7 @@ const NOOP = () => {
   // noop
 };
 
+// ts-prune-ignore-next
```
tjzel, line 9: Why do we need this?

Latropos: This line is not in use, so we have to disable ts-prune for it with this directive.

Latropos: We never import `default` from this file, so it gets marked by ts-prune

Latropos: Why? It is not used, we keep it in advance, so IMHO it is fine

Latropos: We can add the whole file to list of ignored patterns as well

src/reanimated2/animation/util.ts
```diff
@@ -451,6 +447,7 @@ export function cancelAnimation<T>(sharedValue: SharedValue<T>): void {
 }
 
 // TODO it should work only if there was no animation before.
+// ts-prune-ignore-next
 export function withStartValue(
```
tjzel, line 451: Could check why do we have this piece of code?

src/reanimated2/helperTypes.ts
```diff
@@ -123,8 +123,6 @@ export type AnimateProps<P extends object> = NonStyleAnimatedProps<P> &
     animatedProps?: Partial<AnimatedPropsProp<P>>;
   };
 
-export type AnimatedProps<P extends object> = AnimateProps<P>;
```
tjzel, line 126: Please restore this type and mark it with ts-prune. It's an alias for AnimateProps since AnimateProps is a terrible type name - I will use it in the future.

src/reanimated2/hook/useReducedMotion.web.ts
```diff
@@ -1,3 +1,4 @@
+// ts-prune-ignore-next
```
tjzel, line 1: Could we add a rule so a comment is necessary when adding `ts-prune` directive like this?

src/reanimated2/fabricUtils.web.ts
```diff
@@ -1,3 +1,4 @@
+// ts-prune-ignore-next
```
tjzel, line 1: Here too

src/reanimated2/jestUtils.web.ts
```diff
@@ -1,4 +1,5 @@
 /*
  * Stubbed for web, where we don't use this file;
  */
+// ts-prune-ignore-next
```
tjzel, line 4: comment pls

src/reanimated2/js-reanimated/MapperRegistry.ts
```diff
@@ -1,6 +1,7 @@
 import type { JSReanimated } from './commonTypes';
 import type Mapper from './Mapper';
 
+// ts-prune-ignore-next
```
tjzel, line 4: Why isn't it used?

src/reanimated2/platform-specific/checkCppVersion.ts
```diff
@@ -15,6 +15,7 @@ export function checkCppVersion() {
   }
 }
 
+// ts-prune-ignore-next
```
tjzel, line 18: We are safe not to export it.

src/reanimated2/platform-specific/checkVersion.web.ts
```diff
@@ -3,4 +3,5 @@
  * Checks that native and js versions of reanimated match.
  * Stubbed for web, where this check is unnecessary.
  */
+// ts-prune-ignore-next
```
tjzel, line 6: comment

tjzel (changes requested):

src/reanimated2/js-reanimated/MapperRegistry.ts
Latropos: Since now, we use mappers from this file: `src/reanimated2/mappers.ts`

src/reanimated2/animation/util.ts
Latropos: It seems to be a legacy code, which should be cleaned up 🧹🪣

src/reanimated2/hook/useReducedMotion.web.ts
Latropos: We would have to create custom eslint rule 😞 so I don't think it's worth the effort. I've added comments to all the existing ts-prune-ignore-next occurences.

tjzel (approved):
Looks good to me, could you test it extensively with our examples before merging?

Issue #4422, opened 2023-05-03
Reanimated 3 on Android: when using the exiting animation on an Animated.View inside a Modal, the UI becomes unresponsive


kadikraman:
### Description

When using the exiting animation on an `Animated.View` inside a `Modal` , the UI becomes unresponsive. In the example you’re no longer able to press the Open Modal button. Commenting out the exiting parameter in the example the issue does not happen. Also, it’s only happening if the animated container is placed inside a Modal, and only on Android.

### Steps to reproduce

1. clone the repository https://github.com/ceceppa/reanimated-3-bug (I had to fork it for the "Snack or a link to a repository" step to make the GH issue bot happy)
2. build and run on Android
3. open the model
4. close the modal
5. open the modal again -> this will fail as the UI becomes unresponsive

### Snack or a link to a repository

https://github.com/kadikraman/reanimated-3-bug

### Reanimated version

3.0.2

### React Native version

0.71.4

### Platforms

Android

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

Debug mode

### Device

None

### Acknowledgements

Yes

ealmiladi:
I'm pretty this is related to unmounting of components. I'm seeing issues, specifically on Android, where if the exiting animation is incomplete and the component unmounts, the whole screen becomes unresponsive, as mentioned here.

gaishimo:
I ran into the same problem too.

I was using an exiting layout animation in a modal, and when I removed it, it stopped happening.

kartik-stan:
Same issue with me as well, if I use exiting animation inside "react-native's" Modal component the app becomes unresponsive

TheAdamBorek:
Yep, same issue on my side :(

justinadkins:
I just ran into this migrating some animations from using the deprecated `Transition` to layout animations. We have a component that has an `exiting` animation but the user also has the ability to close the `Modal` that the component is inside of at any point. Once closed, the entire parent of the `Modal` is unresponsive. 

Using the element inspector tool I can't even get it to select this component once in this state.

bfricka:
I've had this issue since updating to Reanimated 3 when it first came out, and it still persists now. We held off on upgrading to 3 at the time, but since it's an issue with layout animations, and layout animations are broken in 2 anyways, we upgraded and have had to disable them in Android.

I appreciate that this bug clarifies the issue only applies to exiting animations.

bfricka:
[...]

PR #4776, opened 2023-07-19
Cancel animation for modals

kadikraman:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/4422

Fixes the bug with Android freezing when Layout Animations are used in a modal.

## Test plan

1. clone the example repository https://github.com/ceceppa/reanimated-3-bug
2. apply the fix in this PR (or copy the code into your own project)
3. build and run on Android
4. open the model
5. close the modal
6. open the modal again -> without the fix in this PR, this will fail as the UI becomes unresponsive


tomekzaw:
Hey @kadikraman, thanks for submitting this PR! Looks good to me, but we'll need to test it just to be sure.

piaskowyk (approved):
🎉

PR #4779, opened 2023-07-20
Rename confusing variable

tomekzaw:
## Summary

Currently, in our code we have
```ts
const IS_WEB = shouldBeUseWeb();
```
which is confusing because there's also a function called `isWeb()`.

This PR replaces all occurrences of `!IS_WEB` with `IS_NATIVE` like we already do in a few places.


tjzel:
confusion of the highest orda

PR #4780, opened 2023-07-20
Properly type `__reanimatedModuleProxy` (and other stuff)

tomekzaw:
## Summary

The motivation behind this PR is that `__reanimatedModuleProxy` (which is injected via JSI) has the same type as `NativeReanimated` class which is a JS-only adapter that wraps JSI binding. While this allows us to save a few lines of code, it breaks the mental model of adapters having a potentially different API than wrapped instances.

This PR adds an explicit type `NativeReanimatedModule` type which is used for `__reanimatedModuleProxy`. Apart from this, I've decoupled `JSReanimated` from `NativeReanimated` by eliminating inheritance as well as refactored the constructor of `NativeReanimated` to simplify the logic behind it.

Finally, I've removed `updateDataSynchronously` method which not only was unused but also didn't work at all (it's used only from the UI runtime as `_updateDataSynchronously`).


tjzel (approved):
with a grain of 🧂

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -117,7 +120,7 @@ export default class JSReanimated extends NativeReanimated {
           2.0 * (qx * qy + qw * qz),
           qw * qw + qx * qx - qy * qy - qz * qz
         );
-        eventHandler({
+        (eventHandler as any)({
```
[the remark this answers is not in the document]
tomekzaw: Done in 77e550a

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -133,7 +136,7 @@ export default class JSReanimated extends NativeReanimated {
         let { x, y, z } = sensor;
         [x, y, z] =
           this.platform === Platform.WEB_ANDROID ? [-x, -y, -z] : [x, y, z];
-        eventHandler({ x, y, z, interfaceOrientation: 0 });
+        (eventHandler as any)({ x, y, z, interfaceOrientation: 0 });
```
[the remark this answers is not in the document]
tomekzaw: Done in 77e550a

src/reanimated2/js-reanimated/index.ts
```diff
@@ -24,7 +24,12 @@ if (isWeb()) {
 
 const reanimatedJS = new JSReanimated();
 
-global._makeShareableClone = (c) => c;
+global._makeShareableClone = () => {
+  throw new Error(
```
piaskowyk, line 28: I'm a little worried about this change because there is a possible path to call this method on the Web:

`runOnJS -> makeShareableCloneOnUIRecursive -> _makeShareableClone``

tjzel (approved):

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -222,6 +227,36 @@ export default class JSReanimated extends NativeReanimated {
...
+  }
+
+  getViewProp<T>(
+    _viewTag: string,
+    _propName: string,
+    _callback?: (result: T) => void
+  ): Promise<T> {
+    throw new Error(
+      '[Reanimated] getViewProp is not available in JSReanimated.'
+    );
```
piaskowyk, line 252: I'm wondering which approach is better: crashing the user's app or simply calling console.warn. If we call this method on the Web, it means we made a mistake in our logic. However, this method doesn't have any impact on the Web, so it could remain as a no-op. If you think that is better to throw an error I'm ok with that.

src/reanimated2/js-reanimated/index.ts
tjzel: You have this piece of code there:

```TS
export function makeShareableCloneOnUIRecursive<T>(value: T): ShareableRef<T> {
  'worklet';
  if (USE_STUB_IMPLEMENTATION) {
    // @ts-ignore web is an interesting place where we don't run a secondary VM on the UI thread
    // see more details in the comment where USE_STUB_IMPLEMENTATION is defined.
    return value;
  }
  ```
so _makeShareableClone won't be called on web

src/reanimated2/js-reanimated/JSReanimated.ts
tjzel: I think the reason behind it was what you mention - to highlight a possible a mistake in our logic and there's no better way for that than a good ol' error

src/reanimated2/js-reanimated/index.ts
tomekzaw: Yup, `USE_STUB_IMPLEMENTATION` (which is equal to `shouldBeUseWeb()` 😄) prevents `_makeShareableClone` from being called.

src/reanimated2/js-reanimated/JSReanimated.ts
tomekzaw: I think at this point we can finally afford to be more strict in terms of error reporting, especially with developer-friendly messages 😄

piaskowyk (approved):

PR #4787, opened 2023-07-21
Make publicGlobals comments TSDoc

tjzel:
## Summary

So the user might read it when it's used.

b4:
Screenshot 2023-07-21 at 13 09 42
after:
Screenshot 2023-07-21 at 13 10 25

### DIFF
--- a/src/reanimated2/publicGlobals.ts
+++ b/src/reanimated2/publicGlobals.ts
@@ -5,7 +5,7 @@ declare global {
   /**
    * This global variable is a diagnostic/development tool.
    *
-   * It is `true` on UI thread and to `false` on JS thread.
+   * It is `true` on the UI thread and `false` on the JS thread.
    *
    * It used to be necessary in the past for some of the
    * functionalities of react-native-reanimated to work
--- a/src/reanimated2/publicGlobals.ts
+++ b/src/reanimated2/publicGlobals.ts
@@ -5,7 +5,7 @@ declare global {
   /**
    * This global variable is a diagnostic/development tool.
    *
-   * It's set to TRUE on UI thread and to FALSE on JS thread.
+   * It is `true` on UI thread and to `false` on JS thread.
    *
    * It used to be necessary in the past for some of the
    * functionalities of react-native-reanimated to work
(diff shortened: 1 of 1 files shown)


src/reanimated2/publicGlobals.ts
```diff
@@ -2,16 +2,18 @@
...
+   * properly but it's no longer the case. Your code
+   * shouldn't depend on it, we keep it here
+   * mainly for backward compatibility reasons for our users.
+   */
   var _WORKLET: boolean | undefined;
 
-  // This is simply an address of UI Runtime. Its type stems from
-  // the need to be able to pass it from the native side to JS
-  // when it's a 64-bit address value.
+  /** This is simply an address of UI Runtime. Its type stems from
```
tomekzaw, line 14:
```suggestion
  /** This is simply an address of the Reanimated UI runtime. Its type stems from
```

tomekzaw (approved):
Nice idea but we need to polish the descriptions.

src/reanimated2/publicGlobals.ts
```diff
@@ -2,16 +2,22 @@
...
-  // This global variable is a diagnostic/development tool
-  // it used to be necessary in the past for some of the
-  // functionalities of react-native-reanimated to work
-  // properly but it's no longer the case and your code
-  // shouldn't depend on it, we keep it here
-  // mainly for backward compatibility reasons for our users.
+  /**
+   * This global variable is a diagnostic/development tool.
+   *
+   * It's set to TRUE on UI thread and to FALSE on JS thread.
```
tomekzaw, line 8:
```suggestion
   * It is `true` on the UI thread and `false` on the JS thread.
```

src/reanimated2/publicGlobals.ts
```diff
@@ -5,7 +5,7 @@ declare global {
   /**
    * This global variable is a diagnostic/development tool.
    *
-   * It's set to TRUE on UI thread and to FALSE on JS thread.
+   * It is `true` on UI thread and to `false` on JS thread.
```
tomekzaw, line 8:
```suggestion
   * It is `true` on the UI thread and `false` on the JS thread.
```

PR #4789, opened 2023-07-21
Rename remoteFunction

tjzel:
## Summary

Currently the name of RN thread functions sent to JS gives no information about them. This PR makes it explicit what they are and what is the property `__functionInDev`.

Name is not final and I'm happy to listen to your proposals (except for keeping the old name 🤢)

## Test plan

Yes


Common/cpp/SharedItems/Shareables.h
```diff
@@ -239,19 +239,18 @@ class ShareableWorklet : public ShareableObject {
...
-      public std::enable_shared_from_this<ShareableRemoteFunction> {
+class RNFunctionRef : public Shareable,
+                      public std::enable_shared_from_this<RNFunctionRef> {
  private:
   jsi::Function function_;
   std::shared_ptr<JSRuntimeHelper> runtimeHelper_;
 
  public:
-  ShareableRemoteFunction(
+  RNFunctionRef(
```
tomekzaw, line 249: This class derives from `Shareable` so its name needs to start with `Shareable` according to the convention (`ShareableString`, `ShareableObject` etc.).

Also, the name shouldn't mention React Native (`RN`). Indeed, currently this supports only functions from RN runtime but here's how it looks on my branch which adds support for multiple runtimes:

```cpp
class ShareableRemoteFunction
    : public Shareable,
      public std::enable_shared_from_this<ShareableRemoteFunction> {
 private:
  jsi::Runtime *runtime_;
  jsi::Function function_;

 public:
  ShareableRemoteFunction(jsi::Runtime &rt, jsi::Function &&function)
      : Shareable(RemoteFunctionType),
        runtime_(&rt),
        function_(std::move(function)) {}
  jsi::Value toJSValue(jsi::Runtime &rt) override {
    if (&rt == runtime_) {
      return jsi::Value(rt, function_);
    } else {
#ifdef DEBUG
      return getValueUnpacker(rt).call(
          rt,
          ShareableJSRef::newHostObject(rt, shared_from_this()),
          jsi::String::createFromAscii(rt, "RemoteFunction"));
#else
      return ShareableJSRef::newHostObject(rt, shared_from_this());
#endif
    }
  }
};
```

As you can see, when `toJSValue` is called with the original runtime where the function comes from, it simply returns the function itself. This will work with all runtimes, not only with the RN runtime.

Finally, I'm not a huge fan of using shorthand `Ref` instead of `Reference`. I still don't know what's the difference between `Component` and `ComponentRef` and I want to save others from this pain.

I believe `ShareableFunction` would be the best option here. Obviously, such functions can't be called but at least they can be shared between runtimes and that's what shareables are for.

src/reanimated2/commonTypes.ts
```diff
@@ -115,7 +115,11 @@ export interface NativeEvent<T> {
 export interface ComplexWorkletFunction<A extends any[], R>
   extends WorkletFunction {
   (...args: A): R;
-  __remoteFunction?: (...args: A) => R;
+  /**
+   * This prop is present only in DEV mode as a fallback for
+   * direct calls to RN thread functions on UI thread.
+   */
+  __functionInDEV?: (...args: A) => R;
```
tomekzaw, line 122: I don't like this name but I can't propose anything better at this moment.

src/reanimated2/initializers.ts
```diff
@@ -77,15 +77,15 @@ function valueUnpacker(objectToUnpack: any, category?: string): any {
       handleCache!.set(objectToUnpack, value);
     }
     return value;
-  } else if (category === 'RemoteFunction') {
+  } else if (category === 'RNFunctionRef') {
```
tomekzaw, line 80: I believe this is the only use case for `category`, do you think there's more use cases for this or can we convert it to some boolean flag?

tomekzaw (changes requested):

src/reanimated2/commonTypes.ts
tjzel: I've been through the desert on a horse with no name...

src/reanimated2/initializers.ts
tjzel: I think the idea behind it was to use in more places but eventually this is the only one. After I double check it (and I guess you too) I think we could move it to such flag.

Common/cpp/SharedItems/Shareables.h
tjzel: Thanks for being so concise in this explanation. Those are indeed very fair points and I think `ShareableFunction` in this case is a good name. Therefore I think we should propagate it to JS to signify this being the same object and rename it from `RemoteFunction` to `ShareableFunction`. I wonder what @piaskowyk thinks about it.

src/reanimated2/commonTypes.ts
piaskowyk: Maybe I am not up to date with the current codebase, but according to my knowledge `__remoteFunction` is used in Release mode too 🤔 Are you sure about this change?

Common/cpp/SharedItems/Shareables.h
piaskowyk: I understand the motivation of this PR. I am glad that you want to improve naming, because this is useful and important, but in this case, I am not a big fan of these changes 😅 

Personally, for me `ShareableRemoteFunction` means exactly what it is. Especially that `RNFunctionRef` is not true, because even now we can meet more than just 2 JS runtimes in the application - for example, if someone uses `rn-camera-vision`.

But overall I am open to discussion 😀

tjzel: I was always wondering what does Remote mean in this context. The name suggests that RemoteFunction is a very specific being but nowhere in the code we have it defined/described and in my opinion it turns out much more magical thanks to its name than it really is.

src/reanimated2/commonTypes.ts
tjzel: You are right, I misinterpreted the comment and made changes in a rush. I'll fix it.

tjzel: Yes, but it's not passing the `category` argument, therefore this if
`else if (category === 'RemoteFunction') {` will never be entered

PR #4790, opened 2023-07-21
Fix runOnJS calling _makeShareableClone on JS

tjzel:
## Summary

Currently if someone calls `runOnJS` from JS thread it will call JS implementation of `_makeShareableClone`. This PR fixes it and also gives a useful comment about the use of `console.log` in fishy places.

## Test plan

Check runOnJS/runOnUI example - it was broken before.

btw. @piaskowyk you were right 😎

### DIFF
--- a/src/reanimated2/shareables.ts
+++ b/src/reanimated2/shareables.ts
@@ -229,11 +229,14 @@ export function makeShareableCloneRecursive<T>(
   return NativeReanimatedModule.makeShareableClone(value, shouldPersistRemote);
 }
 
+// Do not try to call `console.log` inside this function as `runOnJS` will call it
+// and then you get into call stack exceeded. Use _log instead (on UI).
 export function makeShareableCloneOnUIRecursive<T>(value: T): ShareableRef<T> {
   'worklet';
-  if (USE_STUB_IMPLEMENTATION) {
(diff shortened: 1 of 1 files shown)


src/reanimated2/shareables.ts
```diff
@@ -229,11 +229,14 @@ export function makeShareableCloneRecursive<T>(
   return NativeReanimatedModule.makeShareableClone(value, shouldPersistRemote);
 }
 
+// Do not try to call `console.log` inside this function as `runOnJS` will call it
+// and then you get into call stack exceeded. Use _log instead (on UI).
 export function makeShareableCloneOnUIRecursive<T>(value: T): ShareableRef<T> {
```
tomekzaw, line 234: Not really sure if we need this comment, don't be overprotective

tomekzaw (approved):

tomekzaw:
Alternative solution: #4799

piaskowyk:
@tjzel great that you found that issue! 🙌 But if you don't mind let's merge @tomekzaw's PR: https://github.com/software-mansion/react-native-reanimated/pull/4799 because covers more cases.

tjzel:
Closed in favour of #4799 😿

PR #4792, opened 2023-07-21
DRY up build.gradle variables

joonmanji:
## Summary
not a big PR but noticing that some of the string literals are just the variable on top so i'm DRYing up the string literals.

## Test plan
build.gradle should still run correctly


tjzel:
Hi @joonmanji. Could you explain a bit what does it change and why is it needed? I'm just curious - and what do you mean by variable DRYing here?

joonmanji:
hey @tjzel. the PR is not at all necessary. just a style change. DRYing means "D(on't) R(epeat) Y(ourself)"
noticing that `def artifactLocalName = "boost_${BOOST_VERSION}.tar.gz"` is repeated in the url strings below. 
functionally everything should be the same. just a small PR.

tjzel:
Nice, I didn't know that! Everything seems alright, I'll start relevant CIs and then merge it 😸

tomekzaw (approved):

PR #4794, opened 2023-07-21
Fix worklet types [0/?]

tjzel:
## Summary

Since worklets are embedded deeply in the code (some might even say that they are the backbone of reanimated❗).
I decided to split all the necessary changes to an unknown number of PRs for gradual improvements and ease of review. 

## Test plan

Newly introduced types in `src/reanimated2/commonTypes` should only be present in the following files:
- `src/reanimated2/commonTypes.ts`,
- `src/reanimated2/threads.ts`.


src/reanimated2/PropAdapters.ts
```diff
@@ -1,5 +1,5 @@
 import { addWhitelistedNativeProps } from '../ConfigHelper';
-import type { AdapterWorkletFunction } from './commonTypes';
+import type { __AdapterWorkletFunction } from './commonTypes';
```
tomekzaw, line 2: What's the idea behind prefixing these types with double underscore? Internal-only type? Temporary type?

src/reanimated2/commonTypes.ts
```diff
@@ -94,30 +94,92 @@ export type MapperRegistry = {
...
+interface ReleaseInitData {
+  code: string;
+}
+
+interface DevInitData {
+  code: string;
+  location: string;
+  sourceMap: string;
+  version: string;
+}
```
tomekzaw, line 108:
```suggestion
}

```

src/reanimated2/commonTypes.ts
```diff
@@ -94,30 +94,92 @@ export type MapperRegistry = {
...
+export type WorkletClosure = Record<string, unknown>;
 
-export interface WorkletFunction {
-  _closure?: Context;
+interface ReleaseInitData {
+  code: string;
+}
+
+interface DevInitData {
+  code: string;
```
tomekzaw, line 104:
```suggestion
interface DevInitData extends ReleaseInitData {
```

src/reanimated2/commonTypes.ts
```diff
@@ -94,30 +94,92 @@ export type MapperRegistry = {
   stop: (mapperID: number) => void;
 };
 
-export type Context = Record<string, unknown>;
+export type WorkletClosure = Record<string, unknown>;
 
-export interface WorkletFunction {
-  _closure?: Context;
+interface ReleaseInitData {
```
tomekzaw, line 99: 1. I think we should name this type `WorkletInitData`, it sounds more complete then, not to mention it's more analogous to `WorkletClosure`.
2. I think we should put `Dev`/`Release` at the end of the name. People usually start reading from the left and `WorkletInitDataDev` looks more similar to `WorkletInitDataRelease` than `DevWorkletInitData` and `ReleaseWorkletInitData`, not to mention autocomplete.

src/reanimated2/commonTypes.ts
```diff
@@ -94,30 +94,92 @@ export type MapperRegistry = {
...
+}
+
+interface DevWorkletBase {
+  _closure: WorkletClosure;
+  __initData: DevInitData;
+  __workletHash: number;
+  __stackDetails: Error;
+}
+
+// A stands for Arguments, R stands for Return value
```
tomekzaw, line 122:
```suggestion
// `A` represents arguments, `R` represents return value
```

src/reanimated2/commonTypes.ts
```diff
@@ -94,30 +94,92 @@ export type MapperRegistry = {
...
+};
+
+export type WorkletFunction<A extends unknown[], T> =
+  | ReleaseWorkletFunction<A, T>
+  | DevWorkletFunction<A, T>;
+
+/**
+ * @deprecated
+ */
+export type __Context = Record<string, unknown>;
```
tomekzaw, line 139: But we already have `WorkletContext`, right?

src/reanimated2/commonTypes.ts
```diff
@@ -94,30 +94,92 @@ export type MapperRegistry = {
...
+}
+
+interface DevInitData {
+  code: string;
+  location: string;
+  sourceMap: string;
+  version: string;
+}
+interface ReleaseWorkletBase {
+  _closure: WorkletClosure;
```
tomekzaw, line 110: This single underscore here makes me furious, can we rename it to `__closure`?

src/reanimated2/threads.ts
```diff
@@ -111,14 +128,18 @@ export function runOnUI<A extends any[], R>(
       });
     }
   };
-}
+}) as unknown as runOnUIAPI;
 
+// Check comment on runOnUIAPI type to understand this type.
+type runOnUIImmediatelyAPI = <A extends unknown[], R>(
```
tomekzaw, line 134: Let's use `FooType` instead of `FooAPI`.
```suggestion
type runOnUIImmediatelyType = <A extends unknown[], R>(
```

src/reanimated2/commonTypes.ts
```diff
@@ -94,30 +94,92 @@ export type MapperRegistry = {
...
 
-export interface NativeEvent<T> {
-  nativeEvent: T;
-}
-export interface ComplexWorkletFunction<A extends any[], R>
-  extends WorkletFunction {
+/**
+ * @deprecated
+ */
+export interface __ComplexWorkletFunction<A extends any[], R>
```
tomekzaw, line 166: WTF is `ComplexWorkletFunction`?

tjzel: yes

src/reanimated2/PropAdapters.ts
tjzel: I renamed old, inaccurate types so now they have a double underscore. Since their names were good (most of the time) I wanted a clear distinction to make the transition easier for me.

src/reanimated2/commonTypes.ts
tjzel: I did it on purpose. Since it's only one property that would be inherited I decided to not make dependencies here yet and have a rather clear distinction between the two.

src/reanimated2/commonTypes.ts
tjzel: Yes, but I don't want to mix the new correct types with the old types. Also `__Context` is imported in a few places.

src/reanimated2/commonTypes.ts
tjzel: maybe

src/reanimated2/threads.ts
tjzel: Since this 'trick' might be used in the code multiple times I wanted to get it all under one type suffix and API is in my opinion a good one since this type actually defines how the function is perceived on the outside and how to use it. If I named it `...type` it would look just like any other TS type and might've been considered unnecessary.

src/reanimated2/commonTypes.ts
tjzel: definitely

src/reanimated2/threads.ts
```diff
@@ -9,7 +9,15 @@ import {
...
-let _runOnUIQueue: Array<[ComplexWorkletFunction<any[], any>, any[]]> = [];
+/**
+ * An array of [worklet, args] pairs.
+ * */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let _runOnUIQueue: Array<[WorkletFunction<any, unknown>, any[]]> = [];
+// TODO TYPESCRIPT
+// above line should be:
+// let _runOnUIQueue: Array<[WorkletFunction<any[], unknown>, any[]]> = [];
+// but for some reason TypeScript gives CLI error on that
```
Latropos, line 20: Give the following change at line (more or less) 114
``` diff
-    _runOnUIQueue.push([worklet, args]);
+    _runOnUIQueue.push([worklet as WorkletFunction<any, unknown>, args]);

```
And there shouldn't be any error with the valid type

src/reanimated2/commonTypes.ts
```diff
@@ -94,30 +94,93 @@ export type MapperRegistry = {
...
+interface WorkletInitDataRelease {
+  code: string;
+}
+
+interface WorkletInitDataDev {
+  code: string;
+  location: string;
+  sourceMap: string;
+  version: string;
+}
```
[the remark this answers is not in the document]
tjzel: Since this is the addition of only one prop I'd rather have them separately without extending to simplify the dependency chain. If at some point in time worklets would get more properties then we can go into extending I think.

src/reanimated2/threads.ts
tjzel: I made this a bit different (without `any` :pepega:)

Latropos (approved):

PR #4800, opened 2023-07-23
Split `Scheduler` into `JSScheduler` and `UIScheduler`

tomekzaw:
## Summary

💡 I recommend reviewing this PR commit-by-commit.

Currently, `Scheduler` class has two responsibilities &ndash; scheduling stuff on the UI thread (platform-specific, written mostly in Java/Obj-C++) and scheduling stuff on the JS thread (platform-agnostic, written in C++). Additionally, importing `Scheduler.h` doesn't work on iOS because the one from React Native takes precedence over ours. I thought it's a great moment to finally split them up.

This PR extracts JS scheduler from `Scheduler` class and sprinkles the code with some const-refs. Also, one might say `JSScheduler` doesn't make sense since we already have `CallInvoker` but this layer of abstraction comes in handy since it allows us to extend its possibilities in the future (e.g. obtain the exclusive access to the main RN runtime on the UI thread and execute the callback synchronously).

In the next episode, we'll take a look on how to improve the Android implementation of Scheduler which is barely understandable in its current shape.

### DIFF
--- a/Common/cpp/SharedItems/Shareables.h
+++ b/Common/cpp/SharedItems/Shareables.h
@@ -9,12 +9,7 @@
 #include "JSRuntimeHelper.h"
 #include "ReanimatedRuntime.h"
 #include "RuntimeManager.h"
-
-#ifdef __APPLE__
-#include <RNReanimated/Scheduler.h>
-#else
-#include "Scheduler.h"
-#endif
+#include "UIScheduler.h"
 
 using namespace facebook;
 
--- a/Common/cpp/Tools/JSScheduler.h
+++ b/Common/cpp/Tools/JSScheduler.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <ReactCommon/CallInvoker.h>
+
+#include <memory>
+
+namespace reanimated {
+
+class JSScheduler {
+ public:
+  explicit JSScheduler(
+      const std::shared_ptr<facebook::react::CallInvoker> &jsCallInvoker)
+      : jsCallInvoker_(jsCallInvoker) {}
+  void scheduleOnJS(std::function<void()> job);
+
+ protected:
+  std::shared_ptr<facebook::react::CallInvoker> jsCallInvoker_;
+};
+
+} // namespace reanimated
--- a/Common/cpp/Tools/JSScheduler.h
+++ b/Common/cpp/Tools/JSScheduler.h
@@ -14,7 +14,7 @@ class JSScheduler {
   void scheduleOnJS(std::function<void()> job);
 
  protected:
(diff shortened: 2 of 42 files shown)


Common/cpp/Tools/JSScheduler.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include <ReactCommon/CallInvoker.h>
+
+#include <memory>
+
+namespace reanimated {
+
+class JSScheduler {
+ public:
+  explicit JSScheduler(
```
tjzel, line 11: Are we sure we can forsake the setter of jsCallInvoker in favor of a constructor?

Common/cpp/SharedItems/Shareables.h
```diff
@@ -9,12 +9,7 @@
 #include "JSRuntimeHelper.h"
 #include "ReanimatedRuntime.h"
 #include "RuntimeManager.h"
-
-#ifdef __APPLE__
-#include <RNReanimated/Scheduler.h>
-#else
-#include "Scheduler.h"
-#endif
+#include "UIScheduler.h"
```
tjzel, line 12: Why was this __APPLE__ recurring in numerous places?

src/reanimated2/threads.ts
```diff
@@ -162,6 +162,10 @@ export function runOnJS<A extends any[], R>(
   fun: ComplexWorkletFunction<A, R>
 ): (...args: A) => void {
   'worklet';
+  if (!IS_NATIVE || !_WORKLET) {
```
tjzel, line 165: huh

tjzel (approved):
Everything looks good to me 👍

piaskowyk (approved):
🚀

Common/cpp/Tools/JSScheduler.h
tomekzaw: Yes, we can. Previously, we'd call `setJSCallInvoker` here (separately for Android and iOS):

https://github.com/software-mansion/react-native-reanimated/blob/80dfdf1306865a276498c0a44ee63b75f5e48978/android/src/main/cpp/NativeProxy.cpp#L87C1-L88

but after these changes we can simply call `std::make_shared<JSScheduler>(jsInvoker)` in NativeReanimatedModule.cpp.

Common/cpp/SharedItems/Shareables.h
tomekzaw: Because there's also `Scheduler.h` file in react-native ([here](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactCommon/react/renderer/scheduler/Scheduler.h)) which for some reason would take precedence over ours. The solution was to pass the full path to the header but it works only on iOS because on Android we need to use `#include "..."`.

PR #4803, opened 2023-07-24
Add missing underscore to `_closure` and correct transform matrices test location

tjzel:
## Summary

To be consistent with worklet properties we turn `_closure` into `__closure`. Also I noticed that tests for transform matrices are not in `__tests__` in root directory so to not get them dispersed I moved them there.

## Test plan

Try out examples


Common/cpp/SharedItems/JSRuntimeHelper.h
```diff
@@ -18,7 +18,7 @@ namespace reanimated {
 class JSRuntimeHelper;
 
 // Core functions are not allowed to capture outside variables, otherwise they'd
-// try to access _closure variable which is something we want to avoid for
+// try to access __closure variable which is something we want to avoid for
```
tomekzaw, line 21:
```suggestion
// try to access `__closure` variable which is something we want to avoid for
```

tomekzaw (approved):
Thanks! ❤️

PR #4804, opened 2023-07-25
add the missing BounceIn to the mock file

matt-livekindred:
## Summary

Found a mock issue when I writing a test with jest for my own react native component which used the `BounceIn.delay` animation, it will throw an error saying: `Cannot read properties of undefined (reading 'delay')`.
After the investigation, it turns out the `BounceIn` was missing in the mock file, so I opened this PR to resolve this issue.


piaskowyk (approved):
Thanks for the PR! 🙌

PR #4810, opened 2023-07-25
Moved (de)registration of ReanimatedCommitHook to its con/de-structor

michalmaka:
## Summary

`ReanimatedCommitHook` was not unregister during shutdown of iOS module so I moved responsibility of registration and deregistration to constructor and destructor of that class making use of RAII allowing me to skip unnecessary calls.

## Test plan

None

### DIFF
--- a/Common/cpp/Fabric/ReanimatedCommitHook.cpp
+++ b/Common/cpp/Fabric/ReanimatedCommitHook.cpp
@@ -8,6 +8,16 @@
 using namespace facebook::react;
 
 namespace reanimated {
+ReanimatedCommitHook::ReanimatedCommitHook(
+    const std::shared_ptr<PropsRegistry> &propsRegistry,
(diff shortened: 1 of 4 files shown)


tomekzaw (approved):
Thanks for submitting this PR! Looks good to me, I will give it a quick round of testing and will merge this soon (once CIs are green)!

piaskowyk (approved):

Common/cpp/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -8,6 +8,16 @@
 using namespace facebook::react;
 
 namespace reanimated {
```
tomekzaw, line 10:
```suggestion
namespace reanimated {

```

tomekzaw (approved):

PR #4813, opened 2023-07-25
Store last reanimated root as atomic ptr, avoid copying and locking

michalmaka:
## Summary

Let's store last reanimated root in `PropsRegistry` as atomic pointer.
This will let us get rid of additional `std::shared_ptr` copy and will be more thread-safe.

## Test plan

None


tomekzaw (approved):
Thanks for this PR! 👍

Issue #4674, opened 2023-07-05
Updating LottieView progress via useAnimatedProps broken


AlexanderEggers:
### Description

When trying to update the progress of a LottieView (lottie-react-native) via the useAnimatedProps hook, the actual update is not received by the component and nothing seems to change on the screen. That used to work in v2 without issues.

Possible workaround for a few cases but not all:

```typescript
  const animationProgressController = useSharedValue(0);

  const animatedProps = useAnimatedProps(() => ({
    progress: animationProgressController.value,
  }));
  const [progress, setProgress] = useState<number>(0);
  useDerivedValue(() => {
    runOnJS(setProgress)(animationProgressController.value);
  }, [animationProgressController]);
```

### Steps to reproduce

1. Checkout https://github.com/AlexanderEggers/ReanimatedIssue/tree/reanimated-hook-issue
2. You will notice that nothing happens on the screen
3. Update the reanimated version to the latest v2
4. Run the application again and the lottie animation progress is now correctly handled 

### Snack or a link to a repository

https://github.com/AlexanderEggers/ReanimatedIssue/tree/reanimated-hook-issue

### Reanimated version

3.3.0

### React Native version

0.71.11

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Acknowledgements

Yes

AlexanderEggers:
@tomekzaw This is the other blocking issue in my project to update to reanimated v3 that I have mentioned [here](https://github.com/software-mansion/react-native-reanimated/issues/4522#issuecomment-1621007156).

tomekzaw:
@AlexanderEggers Thanks for submitting the issue as well as providing the repro, currently I'm quite busy but I will take a look at it soon.

AlexanderEggers:
@tomekzaw Do you have any updates yet?

AlexanderEggers:
I think I found the root cause of this issue which is around the latest stable lottie-react-native version. I had a try with the latest preview version and everything seem to be working fine for me now. I believe this PR likely fixed the issue on the lottie-react-native side: https://github.com/lottie-react-native/lottie-react-native/pull/992 

@tomekzaw So my question would be why v3 is not working with the current stable and seems to be requiring https://github.com/lottie-react-native/lottie-react-native/pull/992?

LeoAnt02:
Any update on this ? @tomekzaw

Issue #4317, opened 2023-03-31
useAnimatedProps not working properly for G svg component


acollazomayer:
### Description

G component not being able to be animated when using animated Props. Initial value of animated Props work however when they change its not being updated on screen.

Here is a repo with the issue replicated. Please note that this behaviour was working before updating to the latest version of the library.
https://github.com/acollazomayer/ReanimatedSVGBug

### Steps to reproduce

1. git clone https://github.com/acollazomayer/ReanimatedSVGBug
2. cd ReanimatedSVGBug
3. yarn
4. cd ios && pod install
5. yarn react-native run-ios
6. Press on the white circle as the slider indicator
7. Slide through the circle

### Snack or a link to a repository

the library. https://github.com/acollazomayer/ReanimatedSVGBug

### Reanimated version

3.0.2

### React Native version

0.71.4

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

tomekzaw:
Hey @acollazomayer, thanks for reporting this issue. @WoLewicki who is a maintainer of react-native-svg promised to take a look at it.

hadnet:
@tomekzaw  @WoLewicki Hey guys, just wanted to give a heads up that there's an issue with the type return of useAnimatedProps, so I can just avoid opening an issue fro this. The return type of that hook is just Partial<T>, but I think it should be Partial<AnimatedProps<T>>. Only returning Partial, TypeScript throws a type error.

acollazomayer:
@WoLewicki did you had the chance to take a look at this issue? Thanks

WoLewicki:
@acollazomayer can you provide the versions of libraries on which it worked? I see you use `x` and `y` props as the animated ones, which are parsed on `render` on the JS side to create the `matrix` prop: https://github.com/software-mansion/react-native-svg/blob/d31ffdb762d34171fd934ac97b031ccf5e0b645d/src/lib/extract/extractProps.ts#L89. I am not sure if it could have ever worked with `react-native-reanimated` since it does indeed you the `render` method only on the initial render. I may be missing something though.

acollazomayer:
@WoLewicki Sure!
 "react-native-svg": "12.1.0",
 "react-native": "0.71.4",
 "react-native-reanimated": "2.14.4",

[...]

PR #4821, opened 2023-07-26
Reintroduce `onReanimatedPropsChange`

Latropos:
## Summary

Reintroduce support for jsProps. 
Closes https://github.com/software-mansion/react-native-reanimated/issues/4317 and closes https://github.com/software-mansion/react-native-reanimated/issues/4674

It turns out that jsProps are still in use by many libraries like `react-native-lottie` and `react-native-svg`.

> **Note**
> This fix does not work for fabric 😢 However Lottie animations are not implemented for Fabric anyway 😎

<table>

<tr><td>The white dot does not move ❌</td><td>The white dot moves ✅</td></tr>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/3c961690-383c-45dc-83ec-5801e13473bf

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/bf900564-cb5a-43cb-b2ef-2d21794727a1

</td></tr>
</table>

## Test plan

Tested on this example: https://github.com/acollazomayer/ReanimatedSVGBug


src/ReanimatedModule.ts
```diff
@@ -0,0 +1,64 @@
...
+  },
+  async createNode(
+    _nodeID: number,
+    _config: Record<string, unknown>
+  ): Promise<void> {
+    // noop
+  },
+  async dropNode(_nodeID: number): Promise<void> {
+    // noop
+  },
```
tomekzaw, line 30: Why do we need this method?

AlexanderEggers:
@Latropos Quick FYI, lottie-react-native v6 with fabric support was released a short while ago.

src/ReanimatedModule.ts
Latropos: You're right, I'll keep only the ones that actually are in use

piaskowyk (changes requested):
let's refactor the event registering logic into a separate file, just like we discussed offline.

piaskowyk:
Did you tested it on Android?

Latropos:
@piaskowyk Yes, it works on Andorid 💪

src/ReanimatedModule.ts
```diff
@@ -0,0 +1,34 @@
+import { NativeEventEmitter, NativeModules } from 'react-native';
+import { nativeShouldBeMock } from './reanimated2/PlatformChecker';
+import { StyleProps } from './reanimated2';
+
+export const NODE_MAPPING = new Map();
```
piaskowyk, line 5: 1. We don't want to export this outside the file. This is an internal module logic.
2. Can we rename `NODE_MAPPING` to something more descriptive?

src/ReanimatedModule.ts
piaskowyk, line 1: Can we rename file to something more descriptive?

src/createAnimatedComponent.tsx
```diff
@@ -409,6 +416,22 @@ export default function createAnimatedComponent(
...
+        ReanimatedEventEmitter.addListener('onReanimatedPropsChange', listener);
+      }
+    }
+
+    _detachPropUpdater() {
+      const viewTag = findNodeHandle(this);
+      NODE_MAPPING.delete(viewTag);
+      if (NODE_MAPPING.size === 0) {
+        ReanimatedEventEmitter.removeAllListeners('onReanimatedPropsChange');
+      }
```
piaskowyk, line 432: same as above

src/createAnimatedComponent.tsx
```diff
@@ -409,6 +416,22 @@ export default function createAnimatedComponent(
       }
     }
 
+    _attachPropUpdater() {
+      const viewTag = findNodeHandle(this);
+      NODE_MAPPING.set(viewTag, this);
+      if (NODE_MAPPING.size === 1) {
+        ReanimatedEventEmitter.addListener('onReanimatedPropsChange', listener);
+      }
```
piaskowyk, line 424: This logic should be in the file which you added.

src/ReanimatedModule.ts
```diff
@@ -0,0 +1,34 @@
...
+
+let reanimatedModule: typeof ReanimatedModuleMock;
+if (nativeShouldBeMock()) {
+  reanimatedModule = ReanimatedModuleMock;
+} else {
+  const { ReanimatedModule } = NativeModules;
+  reanimatedModule = ReanimatedModule;
+}
+
+export const ReanimatedEventEmitter = new NativeEventEmitter(reanimatedModule);
```
piaskowyk, line 34: we don't want to export `ReanimatedEventEmitter` outside the file. We want to export object which interface:
+ addOnJSPropsChangeListener(<component>)
+ removeOnJSPropsChangeListener(<viewtag>)

src/JSPropUpdater.ts
```diff
@@ -0,0 +1,62 @@
...
+    TAG_TO_COMPONENT_MAPPING.set(viewTag, animatedComponent);
+    if (TAG_TO_COMPONENT_MAPPING.size === 1) {
+      this.ReanimatedEventEmitter.addListener(
+        'onReanimatedPropsChange',
+        listener
+      );
+    }
+  }
+
+  _detachPropUpdater(animatedComponent: React.Component<unknown, unknown>) {
```
piaskowyk, line 55:
```suggestion
  public detachPropUpdater(animatedComponent: React.Component<unknown, unknown>) {
```

piaskowyk (changes requested):

src/JSPropUpdater.ts
```diff
@@ -0,0 +1,65 @@
...
+    },
+  };
+
+  private static _listener(data: ListenerData) {
+    const component = JSPropUpdater._tagToComponentMapping.get(data.viewTag);
+    component && component._updateFromNative(data.props);
+  }
+
+  constructor() {
+    let _reanimatedModule: typeof JSPropUpdater._reanimatedModuleMock;
```
piaskowyk, line 32:
```suggestion
    let reanimatedModule: typeof JSPropUpdater._reanimatedModuleMock;
```

piaskowyk (changes requested):

piaskowyk (approved):

src/createAnimatedComponent.tsx
```diff
@@ -286,6 +287,7 @@ export default function createAnimatedComponent(
     _inlinePropsMapperId: number | null = null;
     _inlineProps: StyleProps = {};
     _sharedElementTransition: SharedTransition | null = null;
+    _JSPropUpdater = new JSPropUpdater();
```
tomekzaw, line 290: Probably `_jsPropsUpdater` would be a better name here

src/JSPropUpdater.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (nativeShouldBeMock()) {
+      reanimatedModule = JSPropUpdater._reanimatedModuleMock;
+    } else {
+      reanimatedModule = NativeModules.ReanimatedModule;
+    }
+    this._reanimatedEventEmitter = new NativeEventEmitter(reanimatedModule);
+  }
+
+  public addOnJSPropsChangeListener(
+    animatedComponent: React.Component<unknown, unknown>
```
tomekzaw, line 42: Is there any TypeScript way to enforce that the component must be animated?

src/JSPropUpdater.ts
```diff
@@ -0,0 +1,65 @@
...
+      // noop
+    },
+    async removeListeners(): Promise<void> {
+      // noop
+    },
+  };
+
+  private static _listener(data: ListenerData) {
+    const component = JSPropUpdater._tagToComponentMapping.get(data.viewTag);
+    component && component._updateFromNative(data.props);
```
tomekzaw, line 28: I know this snippet of code was copied but it may be finally worth to convert it into a proper conditional.

Latropos:
Not all svg issues were solved, looks that fill color still doesn't change https://github.com/software-mansion/react-native-reanimated/issues/1909

tomekzaw:
As far as I know, it required a custom prop adapter, see `createAnimatedPropAdapter`.

PR #4823, opened 2023-07-26
Fix worklet types [1/?] + correct types in useAnimatedGestureHandler

tjzel:
## Summary

This PR focuses on making types of `useAnimatedGestureHandler` more approachable - there is a distinction between structure of events that `useAnimatedGestureHandler` receives and structure of events that RNGH claims it is receiving (I'm looking at you, `nativeEvent`).
Also:

- removed regression that happened in #4519 - events by default were inferred to `any` if generic parameter was not provided, before they were by default `PanGesture` events
- marked it as deprecated

I removed third generic parameter from the hook since in old `d.ts` is was defined as having two generic parameters. Also we don't want user to be able to actually pass something there - it was used to shorten the code a bit.

Type names are obviously not final and I'd gladly accept some better ones (if you can think of some 😼)

t.hanks. @tomekzaw for suggesting how to deal with regression

## Test plan

Standard TS test suite.


tjzel:
Superseded by #5056

PR #4826, opened 2023-07-26
Removed unnecessary lock during committing shadowTree

michalmaka:
## Summary

During researching on getting into assert on too many failed commit attempts I've noticed that there is a lock used by both `NativeReanimatedModule` and `ReanimatedCommitHook`. That lock in the first one is not necessary anymore as it doesn't guard anything from `PropsRegistry`. What's more, it may be a cause for some commit attempt to fail due to waiting two threads for each other.

## Test plan

Chessboard example worked fine


tomekzaw (approved):
Works like charm! ✅

PR #4833, opened 2023-07-27
Improved skipping commits

michalmaka:
## Summary

During investigation of race between RN and reanimated shadow tree committing I've noticed that if JS thread commits a new tree (triggering `pleaseSkipReanimatedCommit`) after starting `commit` from reanimated, the latter `commit` may cause JS commit failures.

What's more, additional calls to `shouldReanimatedSkipCommit ` result in avoiding CPU consuming copy operations on nodes and thus seems to work more performant.

I added those checks just before each operation handling and just after full commit.

Chessboard example seemed to work a bit faster that without those changes. I used also profiler and in overall CPU spends less time in ShadowTree.

NOTICE ⚠️ 
This change requires change in RN codebase. PR has been submitted https://github.com/facebook/react-native/pull/38715

## Test plan

Check performance of e.g. chessboard example.

### DIFF
--- a/Common/cpp/NativeModules/NativeReanimatedModule.cpp
+++ b/Common/cpp/NativeModules/NativeReanimatedModule.cpp
@@ -556,9 +556,11 @@ void NativeReanimatedModule::performOperations() {
 
           return newRoot;
         },
-        {.shouldYield = [this]() {
-          return propsRegistry_->shouldSkipCommit();
-        }});
+        {/* .enableStateReconciliation = */ false,
+         /* .mountSynchronously = */ true,
+         /* .shouldYield = */ [this]() {
+           return propsRegistry_->shouldSkipCommit();
+         }});
   });
 }
 
--- a/Common/cpp/NativeModules/NativeReanimatedModule.cpp
+++ b/Common/cpp/NativeModules/NativeReanimatedModule.cpp
@@ -525,7 +525,8 @@ void NativeReanimatedModule::performOperations() {
     ReanimatedCommitMarker commitMarker;
 
     shadowTree.commit(
-        [&](RootShadowNode const &oldRootShadowNode) {
+        [&](RootShadowNode const &oldRootShadowNode)
+            -> RootShadowNode::Unshared {
           auto rootNode =
               oldRootShadowNode.ShadowNode::clone(ShadowNodeFragment{});
 
@@ -534,6 +535,10 @@ void NativeReanimatedModule::performOperations() {
           for (const auto &[shadowNode, props] : copiedOperationsQueue) {
             const ShadowNodeFamily &family = shadowNode->getFamily();
             react_native_assert(family.getSurfaceId() == surfaceId_);
+
+            if (propsRegistry_->shouldSkipCommit()) {
+              return nullptr;
+            }
 
             auto newRootNode = shadowTreeCloner.cloneWithNewProps(
                 rootNode, family, RawProps(rt, *props));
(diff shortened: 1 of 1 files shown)


efstathiosntonas:
@michalmaka you're on fire! 💪 🔥

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -619,7 +619,8 @@ void NativeReanimatedModule::performOperations() {
 
   shadowTreeRegistry.visit(surfaceId_, [&](ShadowTree const &shadowTree) {
     shadowTree.commit(
-        [&](RootShadowNode const &oldRootShadowNode) {
+        [&](RootShadowNode const &oldRootShadowNode)
+            -> RootShadowNode::Unshared {
```
tomekzaw, line 623: +1 for this change 👍

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -632,6 +633,10 @@ void NativeReanimatedModule::performOperations() {
               const ShadowNodeFamily &family = shadowNode->getFamily();
               react_native_assert(family.getSurfaceId() == surfaceId_);
 
+              if (propsRegistry_->shouldSkipCommit()) {
+                return nullptr;
+              }
+
```
tomekzaw, line 639: I'm not particularly happy that we read the atomic for each animated component but I guess it's not that expensive compared to cloning the whole ShadowTree which is built from `ShadowNode::Shared`.

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -652,7 +657,9 @@ void NativeReanimatedModule::performOperations() {
 
           return newRoot;
         },
-        {/* default commit options */});
+        {.shouldYield = [this]() {
+          return propsRegistry_->shouldSkipCommit();
+        }});
```
tomekzaw, line 662: Looks like `.foo =` is a feature from C++20 but we need to support C++17, [here's](https://github.com/facebook/react-native/blob/3e828691ddbb909190f33b562fd0ba3c0f3df23a/packages/react-native/ReactCommon/react/renderer/uimanager/UIManager.cpp#L98) how it looks in react-native, maybe we should use the same technique here?
```suggestion
        {/* .shouldYield = */ [this]() {
          return propsRegistry_->shouldSkipCommit();
        }});
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
michalmaka: I guess so, do we have performance tests that can measure impact of this change?

Common/cpp/NativeModules/NativeReanimatedModule.cpp
michalmaka: ok

Common/cpp/NativeModules/NativeReanimatedModule.cpp
tomekzaw: Unfortunately we don't have setup for perftests yet.

tomekzaw (approved):

tomekzaw:
⚠️ Wait for https://github.com/facebook/react-native/pull/38715

efstathiosntonas:
@tomekzaw I've asked rn team to pick it on `0.72.4` https://github.com/reactwg/react-native-releases/discussions/83#discussioncomment-6664904

efstathiosntonas:
Hey folks, this was picked from rn for `0.72.4`. Thank you for your hard work

tomekzaw (changes requested):
As discussed internally, let's add #ifdefs for RN version not to change the behavior of 0.72.x since this PR requires things which were added in 0.72.4.

tomekzaw (approved):
Thanks for this PR!

Issue #4130, opened 2023-02-28
v3 not working on web (Unexpected token 'true')


laurens-lamberts:
### Description

Hi! First of all, thanks a lot for all your work. Really appreciated.

I just tried updating my react-native-web project to reanimated v3 (for the yummy sensors, obviously), but found that it does not work for me yet. The web-app does not load, and I find the following console error;

`Uncaught SyntaxError: Unexpected token 'true' (at bundle.js:96811:5)`

Navigating to the loaded bundle JS, I find it leads to this section of code;
```
  _f._closure = {
    USE_STUB_IMPLEMENTATION,
    true
  };
```
This is within the function `makeShareableCloneOnUIRecursive`.

I tried this in a snack, but it doesn't seem to happen there.
Please let me know if it is just me, but as the error I get is quite specific, maybe this issue reports still helps.

### Steps to reproduce

1. Run a react-native-web project with reanimated v3
2. The console error should pop up

### Snack or a link to a repository

https://snack.expo.dev/HbdeQDf9V (does NOT have the issue)

### Reanimated version

3.0.0

### React Native version

0.71.3

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

laurens-lamberts:
Sorry, I just tried this on a fresh react-native-web project on my machine, but I can't seem to reproduce this issue.
Will try to find out if there is another dependency conflicting with reanimated v3.
Closing this issue for now.

asherLZR:
We had the same issue. In our case we are using the `unstable_transformProfile: 'hermes-stable'` option with `'module:metro-react-native-babel-preset'` to opt out of transforms that Hermes supports, including property shorthand.

So then we have `react-native-reanimated/plugin` transpiling to
```
_f._closure = {
  USE_STUB_IMPLEMENTATION,
  __DEV__
};
```
Later on there is a define plugin for tree-shaking which does a string replacement and end up with the reported problem.
```
new webpack.DefinePlugin({
  __DEV__: JSON.stringify(dev),
}),
```

So far this patch looks fine - don't use property shorthand when transforming worklets. Would be great to understand if this is something we want to make default in the plugin though.
```
diff --git a/node_modules/react-native-reanimated/plugin/build/plugin.js b/node_modules/react-native-reanimated/plugin/build/plugin.js
index cfe8ab6..cf8b892 100644
[...]

PR #4837, opened 2023-07-28
fix: babel-plugin should not use property shorthands when transpiling worklets

asherLZR:
## Summary

The babel plugin transforms worklets' closure variables with property shorthand. This is somewhat problematic if the project uses common optimisations like the webpack DefinePlugin to transform these variable values later at build-time.

For example, we can end up transpiling with the DefinePlugin
```
_f._closure = {
  USE_STUB_IMPLEMENTATION,
  __DEV__
};
```
to 
```
_f._closure = {
  USE_STUB_IMPLEMENTATION,
  false
};
```

The solution here is to default to not using it. Fixes https://github.com/software-mansion/react-native-reanimated/issues/4130

## Test plan

See unit test. Tested in real project.


tomekzaw:
Hey @asherLZR, thanks for submitting this PR. This is in fact a problem with `babel-plugin-transform-define` which has just been fixed in https://github.com/FormidableLabs/babel-plugin-transform-define/pull/86 and released in 2.1.3.

asherLZR:
Hi @tomekzaw thanks for the review and the link. Forgive my limited understanding, but in our project we're using `webpack.DefinePlugin` which doesn't seem related to `babel-plugin-transform-define`.

However, I believe that it should be addressed by this PR https://github.com/webpack/webpack/pull/17231 where it isn't handling the shorthand property correctly. Closing this as a result.

tjzel:
Hi @asherLZR. We have discussed this topic in the past with @tomekzaw I'm not sure whether it actually solves all problems related to this issue. Now (potentially) you could end up with `{ true: true }` instead of `{ true }` and both are equally bad. 

While I could approve this pull request since it looks fine and on the first glance it's irrelevant to our logic, piece of code you submitted creates a node that is an `objectProperty`. Therefore I have a strong belief it shouldn't be touched by any transform of this kind - if `{ __DEV__: __DEV__ }` is not transformed then `{ __DEV__ }` shouldn't be transformed as well. It seems like an issue - or on the complete contrary, conscious decision - in `webpack.DefinePlugin`. I will reopen this pull request but I think relevant research has to be done before merging it.

tjzel:
@asherLZR 
I've fiddled a bit with Webpack and Webpack's DefinePlugin and were unable to get it to situation when using property shorthands leads to having illegal tokens in closure. Could you possibly provide some reproduction on this issue? 

I was using latest Webpack's version - could you see if the problem persists if you use it too?

asherLZR:
@tjzel
Thanks for looking into it. In our project I can confirm that from webpack [v5.84.0](https://github.com/webpack/webpack/releases/tag/v5.84.0) onwards the issue is resolved with the mentioned fix to the DefinePlugin. Do you still require a reproduction?

tjzel:
Oh, I must've missed that mentioned fix, sorry for that 😅. Therefore we are all good and we can close that!

PR #4838, opened 2023-07-28
Fix "Unexpected token "=" error"

### LINKED ISSUE #4825: error: Unexpected token "=" error
### Description

error: unexpected token "=" error. 

Screenshot_1690338548

### Steps to reproduce

1. `yarn start` 
2. run `yarn android` to build on a physical device or simulator
3. when the app starts up you'll immediately see the error 

https://github.com/CavanWagg/react-native-reanimated-error

### Reanimated version

3.4.0

### React Native version

0.69.1

### Platforms

Android

### JavaScript runtime

JSC

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

Real device

### Device model

Moto g power (Android 11)

### Acknowledgements

Yes

Latropos:
## Summary

According to this react-native issue [LINK](https://github.com/facebook/react-native/issues/31704) there is no support for Logical OR and AND assignment `||=` `&&=`, unless you use `@babel/plugin-proposal-logical-assignment-operators`. To simplify setup of our library we should remove all logical assignments (especially that according to [this issue](https://github.com/software-mansion/react-native-reanimated/issues/4825) adding babel plugins doesn't help). Closes #4825.

## Test plan
I've tested [this app](https://github.com/CavanWagg/react-native-reanimated-error) on Android and IOS. On IOS always everything works, however android app crashes without this fix, exactly as described in the issue.

### DIFF
--- a/src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
+++ b/src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
@@ -34,6 +34,7 @@ export function isAffineMatrixFlat(x: unknown): x is AffineMatrixFlat {
   );
 }
 
+// ts-prune-ignore-next This function is exported to be tested
 export function isAffineMatrix(x: unknown): x is AffineMatrix {
   'worklet';
   return (
@@ -53,6 +54,7 @@ export function flatten(matrix: AffineMatrix): AffineMatrixFlat {
   return matrix.flat() as AffineMatrixFlat;
 }
 
+// ts-prune-ignore-next This function is exported to be tested
 export function unflatten(m: AffineMatrixFlat): AffineMatrix {
   'worklet';
   return [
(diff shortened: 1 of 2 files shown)


tomekzaw (approved):
Thanks for the investigation and fixing it!

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -34,6 +34,7 @@ export function isAffineMatrixFlat(x: unknown): x is AffineMatrixFlat {
   );
 }
 
+// ts-prune-ignore-next This function is exported to be tested
```
tomekzaw, line 37: This should be moved to a separate PR

tomekzaw (changes requested):

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
Latropos: 🆗 So here: https://github.com/software-mansion/react-native-reanimated/pull/4839

tomekzaw (approved):
✅

tjzel (approved):

MichaelKic:
Is this fix getting released any time soon or is it already in the newest release?

Latropos:
@MichaelKic It is not released yet, you have to wait for next release. However you can see [here](https://github.com/software-mansion/react-native-reanimated/releases), that we tond to release a new version a bit more often than once a month, so this shouldn't take long

PR #4843, opened 2023-07-28
Add CI to check react-native nightly build

piaskowyk:
## Summary

This PR adds CI to check if Reanimated compiles with the newest version of the react-native.

## Test plan

[Check React Native nightly build](https://github.com/software-mansion/react-native-reanimated/actions/workflows/check-react-native-nightly.yml)


piaskowyk:
This is correct behavior that CI fails - Reanimated doesn't compatible with the latest nightly yet

.github/workflows/check-react-native-nightly.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: npx react-native init app --skip-install --version nightly
+      - name: Install Reanimated
+        working-directory: app
+        run: yarn add github:software-mansion/react-native-reanimated
+      - name: Install Paper Pods
+        if: ${{ matrix.react-native-architecture == 'Paper' }}
+        working-directory: app/ios
+        run: pod install
+      - name: Install Fabric Pods
+        if: ${{ matrix.react-native-architecture == 'Fabric' }}
```
tjzel, line 31: Why do we install both Paper and Fabric pods and then launch an app? Shouldn't we do it separately for each architecture?

tjzel (approved):
seems good (I hope CI will pass tonight)

PR #4844, opened 2023-07-28
Hide accidentally leaked functions to public API

tjzel:
## Summary

Removal of `react-native-reanimated.d.ts` exposed some of our internal API with `export * from ...`. This file limits the exports and (kind-of) defines public API.

Requires #4847.

## Test plan

CI


src/reanimated2/utils.ts
```diff
@@ -3,7 +3,7 @@ import { measure } from './NativeMethods';
 import type { AnimatedRef } from './hook/commonTypes';
 import type { SharedValue } from './commonTypes';
 
-export interface ComponentCoords {
+interface ComponentCoords {
```
tomekzaw, line 6: This should be exported as well since this is the return type of `getRelativeCoords` which is also exported.
```suggestion
export interface ComponentCoords {
```

src/reanimated2/layoutReanimation/defaultTransitions/LinearTransition.ts
```diff
@@ -48,4 +48,7 @@ export class LinearTransition
   };
 }
 
+/**
+ * @deprecated Please use `LinearTransition` instead.
+ */
 export const Layout = LinearTransition;
```
tomekzaw, line 54: What about `Layout.springify()`?

tomekzaw (changes requested):

src/reanimated2/core.ts
```diff
@@ -52,29 +47,7 @@ global._log = function (s: string) {
   console.log(s);
 };
 
-export function getViewProp<T>(viewTag: string, propName: string): Promise<T> {
```
tomekzaw, line 55: This function can still be used internally for testing purposes. We should either remove it completely (along with `NativeReanimatedModule.getViewProp` and native code) or leave it as it is now.

src/reanimated2/index.ts
```diff
@@ -27,5 +244,6 @@ export type {
   AnimatedStyle,
   StylesOrDefault,
 } from './helperTypes';
+export { setNativeProps } from './SetNativeProps';
```
tomekzaw, line 247: Can we move this line near `NativeMethods`?

tomekzaw (changes requested):

src/reanimated2/core.ts
tjzel: removele

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -357,34 +357,6 @@ void NativeReanimatedModule::unregisterEventHandler(
       [=] { eventHandlerRegistry->unregisterEventHandler(id); });
 }
 
-jsi::Value NativeReanimatedModule::getViewProp(
```
piaskowyk, line 360: Due to this PR - https://github.com/software-mansion/react-native-reanimated/pull/4929 we should restore this method.

tomekzaw (approved):

piaskowyk (approved):

Issue #4852, opened 2023-07-30
withTiming doesn't update progress properly when used with gesture handler


MatiPl01:
### Description

I am not sure what is the cause of the issue. I observed this behaviour when I was using the `react-native-gesture-handler` to detect the pan gesture and I was updating the shared value at the same time with the `withTiming` animation function.

Basically, there is a problem in how the `now` timestamp is calculated. I noticed that during successive frames of the animation, the timestamp alternately increases and decreases.

Here is the code snippet that I use to test it out:
```tsx
function App(): JSX.Element {
  const dummyValue = useSharedValue(0);

  const gesture = Gesture.Pan();

  useEffect(() => {
    dummyValue.value = withTiming(1, {duration: 10000});
  }, []);

  useAnimatedReaction(
    () => dummyValue.value,
    v => {
      console.log(v);
    },
  );

  return (
    <GestureHandlerRootView style={styles.root}>
      <GestureDetector gesture={gesture}>
        <View style={styles.view} />
      </GestureDetector>
    </GestureHandlerRootView>
  );
}
```
Everything works fine until I start moving the finger over the blue `View` rendered inside the `GestureDetector`.

You can see example output below:
```
...
0.26557472          
0.26546370009170783 <- lower than before
0.26688818000000003 <- greater than before
0.2666798827583936  <- lower than before
0.26805841999999996 <- greater than before
0.2678988449509465     ...
0.26923122
0.2691205866519347
0.27055368
0.27034510787882987
0.27173191999999996
0.2715724086141207
0.27291272
0.2728024888753581
...
```

To confirm that the issue is caused by the incorrect `now` timestamp value being passed to the `timing` function, you can add `console.log(now)` statement to the `timing` function in the `/src/reanimated2/animation/timing.ts` file.

### Steps to reproduce

1. clone [this](https://github.com/MatiPl01/reanimated-issues) repository
2. switch to the `time-measurement` branch: `git fetch && git switch time-measurement`
3. run the example
4. swipe the rectangle and observe output in the terminal

### Snack or a link to a repository

https://github.com/MatiPl01/reanimated-issues/tree/time-measurement

### Reanimated version

3.3.0

### React Native version

0.72.0

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

YesSkyscrapers:
Faced this issue.... withSpring or any not working too.
[...]

PR #4853, opened 2023-07-30
Fix withTiming progress fluctuation issue

MatiPl01:
## Summary

This PR fixes the #4852 issue with improperly updated current animation timestamp value. It uses a simple approach to take the maximum value of the previous timestamp and the timestamp passed as the `now` property to the `timing` function which prevents the progress from alternating.

## Test plan

Open the example from the #4852 issue and test if it works.


Latropos:
Thank you very much for your help! We will probably do this comparison inside a function calling the animation frame to fix it once for all animations: https://github.com/software-mansion/react-native-reanimated/pull/4856

Latropos:
Closing, since this fix was merged in https://github.com/software-mansion/react-native-reanimated/pull/4856#event-10206371217

PR #4857, opened 2023-07-31
_log for complex types

t0ms0n00:
## Summary

This PR expands the logic of _log function to print more complex types in JSON-like style. It adds support for the types listed below:

- JSI types
- JS Objects and Collections

## Test plan

You can use and example app named `Log test`


Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -188,4 +182,83 @@ void RuntimeDecorator::decorateRNRuntime(
...
+  } else if (value.isUndefined()) {
+    return "undefined";
+  } else if (value.isObject()) {
+    return parseComplexValue(rt, value.getObject(rt));
+  } else {
+    return "unsupported value type";
+  }
+}
+
+std::string parseComplexValue(jsi::Runtime &rt, jsi::Object const &object) {
```
tomekzaw, line 199:
```suggestion
std::string parseComplexValue(jsi::Runtime &rt, const jsi::Object &object) {
```

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -188,4 +182,83 @@ void RuntimeDecorator::decorateRNRuntime(
...
+  std::stringstream parsed;
+
+  if (object.isArray(rt)) {
+    parsed << "[";
+
+    jsi::Array arr = object.getArray(rt);
+    size_t length = arr.size(rt);
```
tomekzaw, line 206:
```suggestion
    auto arr = object.getArray(rt);
    auto length = arr.size(rt);
```

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -188,4 +182,83 @@ void RuntimeDecorator::decorateRNRuntime(
...
+    if (length > 0) {
+      parsed.seekp(-2, parsed.cur);
+    }
+    parsed << "] ";
+  } else if (object.isFunction(rt)) {
+    parsed << "[Function "
+           << object.getProperty(rt, "name").toString(rt).utf8(rt) << "]";
+  } else if (object.isHostObject(rt)) {
+    int status;
+    jsi::HostObject *hostObject = object.asHostObject(rt).get();
```
tomekzaw, line 222:
```suggestion
    jsi::HostObject *hostObject = object.asHostObject(rt).get();
    int status;
```

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -188,4 +182,83 @@ void RuntimeDecorator::decorateRNRuntime(
...
+        abi::__cxa_demangle(typeid(*hostObject).name(), NULL, NULL, &status);
+    if (status == 0) {
+      // create a plain jsi::Object
+      facebook::jsi::Object obj(rt);
+
+      // copy properties from hostObject to obj
+      for (auto &key : hostObject->getPropertyNames(rt)) {
+        facebook::jsi::Value value = hostObject->get(rt, key);
+        obj.setProperty(rt, key, value);
+      }
```
tomekzaw, line 233: Is it possible to do it without creating a new object and copying all properties?

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -188,4 +182,83 @@ void RuntimeDecorator::decorateRNRuntime(
...
+
+      parsed << "{\"class\": "
+             << "\"" << hostObjClassName << "\""
+             << ", \"props\": " << parseComplexValue(rt, obj) << "}";
+    } else {
+      parsed << "[jsi::HostObject] - error parsing class name";
```
tomekzaw, line 239:
```suggestion
      parsed << "[jsi::HostObject]";
```

Common/cpp/Tools/RuntimeDecorator.h
```diff
@@ -107,4 +107,7 @@ inline bool RuntimeDecorator::isReactRuntime(jsi::Runtime &rt) {
   return false;
 }
 
+std::string parseValue(jsi::Runtime &rt, jsi::Value const &value);
+std::string parseComplexValue(jsi::Runtime &rt, jsi::Object const &value);
```
tomekzaw, line 111: Do we really need two functions where the second one is just for complex types? Perhaps it would be better to implement it somehow similarly to `makeShareableClone` (and `makeShareableCloneRecursive`). So we would have `stringifyValue` which serves as an entry point and calls `stringifyValueRecursively` which, as the name suggests, is called recursively.

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -188,4 +182,83 @@ void RuntimeDecorator::decorateRNRuntime(
...
+  if (value.isBool() || value.isNumber()) {
+    return value.toString(rt).utf8(rt);
+  } else if (value.isString()) {
+    return "\"" + value.getString(rt).utf8(rt) + "\"";
+  } else if (value.isUndefined()) {
+    return "undefined";
+  } else if (value.isObject()) {
+    return parseComplexValue(rt, value.getObject(rt));
+  } else {
+    return "unsupported value type";
```
tomekzaw, line 195: What do you think?
```suggestion
    return "[jsi::Value]";
```

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -188,4 +182,83 @@ void RuntimeDecorator::decorateRNRuntime(
...
+
+      // copy properties from hostObject to obj
+      for (auto &key : hostObject->getPropertyNames(rt)) {
+        facebook::jsi::Value value = hostObject->get(rt, key);
+        obj.setProperty(rt, key, value);
+      }
+
+      parsed << "{\"class\": "
+             << "\"" << hostObjClassName << "\""
+             << ", \"props\": " << parseComplexValue(rt, obj) << "}";
```
tomekzaw, line 237: We will need to define what's the expected result here.

For objects that contains some props, here's what I'd like to propose:
`[HostObjectName {foo: bar}]`.

If there's no props, it should be only `[HostObjectClassName]`, e.g. `[ShadowNodeWrapper]`.

tomekzaw:
Would be great if you could scan through [jsi.h](https://github.com/facebook/hermes/blob/main/API/jsi/jsi/jsi.h) and check if there's anything more that we need handle. Currently we don't support things like symbols, BigInt or ArrayBuffer but we can already implement stringification for them (they can be easily constructed from regular JS code). Apart from this, it would be great to support objects with prototype like `Map` or `Set`, for instance like `[Object Map]` or `[Object Set]`.

t0ms0n00:
> Would be great if you could scan through [jsi.h](https://github.com/facebook/hermes/blob/main/API/jsi/jsi/jsi.h) and check if there's anything more that we need to handle. Currently we don't support things like symbols, BigInt or ArrayBuffer but we can already implement stringification for them (they can be easily constructed from regular JS code). Apart from this, it would be great to support objects with prototype like `Map` or `Set`, for instance like `[Object Map]` or `[Object Set]`.

Added handle for more types:

- jsi::BigInt
- jsi::Symbol
- jsi::ArrayBuffer
- JS Map
- JS Set
- JS Error

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -4,22 +4,16 @@
...
-      if (value.isString()) {
-        Logger::log(value.getString(rt).utf8(rt).c_str());
-      } else if (value.isNumber()) {
-        Logger::log(value.getNumber());
-      } else if (value.isUndefined()) {
-        Logger::log("undefined");
-      } else {
-        Logger::log("unsupported value type");
-      }
+      std::string serializedValue = stringifyValue(rt, value);
```
michalmaka, line 15:
```suggestion
      auto serializedValue = stringifyValue(rt, value);
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+  const jsi::Function &getPrototype =
+      rt.global()
+          .getPropertyAsObject(rt, "Object")
+          .getPropertyAsFunction(rt, "getPrototypeOf");
+
+  std::string result = getPrototype.call(rt, obj).toString(rt).utf8(rt);
+  std::string pattern = "[object " + expectedType + "]";
+
+  return result == pattern;
+}
```
michalmaka, line 17: wrap with unnamed namespace in order to make this function locally available only in this file

```suggestion
namespace {
bool checkJSCollectionType(
    jsi::Runtime &rt,
    const jsi::Object &obj,
    std::string expectedType) {
  const jsi::Function &getPrototype =
      rt.global()
          .getPropertyAsObject(rt, "Object")
          .getPropertyAsFunction(rt, "getPrototypeOf");

  std::string result = getPrototype.call(rt, obj).toString(rt).utf8(rt);
  std::string pattern = "[object " + expectedType + "]";

  return result == pattern;
}
}
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
+#include "JSISerializer.h"
+#include <cxxabi.h>
+
+bool checkJSCollectionType(
+    jsi::Runtime &rt,
+    const jsi::Object &obj,
+    std::string expectedType) {
```
michalmaka, line 7:
```suggestion
    const std::string& expectedType) {
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+  const jsi::Function &getPrototype =
+      rt.global()
+          .getPropertyAsObject(rt, "Object")
+          .getPropertyAsFunction(rt, "getPrototypeOf");
+
+  std::string result = getPrototype.call(rt, obj).toString(rt).utf8(rt);
+  std::string pattern = "[object " + expectedType + "]";
```
michalmaka, line 14:
```suggestion
  auto result = getPrototype.call(rt, obj).toString(rt).utf8(rt);
  auto pattern = "[object " + expectedType + "]";
```

Common/cpp/Tools/JSISerializer.h
```diff
@@ -0,0 +1,28 @@
...
+    jsi::Runtime &rt,
+    const jsi::Value &value);
+std::string stringifyJSIArray(jsi::Runtime &rt, const jsi::Array &arr);
+std::string stringifyJSIArrayBuffer(
+    jsi::Runtime &rt,
+    const jsi::ArrayBuffer &buf);
+std::string stringifyJSIFunction(jsi::Runtime &rt, const jsi::Function &func);
+std::string stringifyJSIHostObject(
+    jsi::Runtime &rt,
+    jsi::HostObject &hostObject);
```
michalmaka, line 22:
```suggestion
    const jsi::HostObject &hostObject);
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+  auto length = arr.size(rt);
+
+  for (size_t i = 0; i < length; i++) {
+    jsi::Value element = arr.getValueAtIndex(rt, i);
+    ss << stringifyValueRecursively(rt, element) << ", ";
+  }
+
+  if (length > 0) {
+    ss.seekp(-2, ss.cur);
+  }
```
michalmaka, line 81:
```suggestion
    ss << stringifyValueRecursively(rt, element);
    if (i != length - 1) {
      ss << ", ";
    }
  }
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+  auto name = func.getProperty(rt, "name").toString(rt).utf8(rt);
+  ss << "[Function " << name << "]";
+  return ss.str();
+}
+
+std::string reanimated::stringifyJSIHostObject(
+    jsi::Runtime &rt,
+    jsi::HostObject &hostObject) {
+  std::stringstream ss;
+  int status;
```
michalmaka, line 107:
```suggestion
  int status = -1;
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    ss << '[' << hostObjClassName << ' ';
+    if (propsCount > 0) {
+      ss << "{";
+      for (auto &key : props) {
+        facebook::jsi::Value value = hostObject.get(rt, key);
+        ss << '"' << key.utf8(rt) << '"' << ": "
+           << stringifyValueRecursively(rt, value) << ", ";
+      }
+      ss.seekp(-2, ss.cur);
```
michalmaka, line 122: Implement check if the key is the last one and in that case skip adding `, `

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+      ss << "}";
```
michalmaka, line 123: It's equivalent code as in `stringifyJSIObject`, please reuse the latter one

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+    auto pair = arr.getValueAtIndex(rt, i).asObject(rt).getArray(rt);
+    auto key = pair.getValueAtIndex(rt, 0);
+    auto value = pair.getValueAtIndex(rt, 1);
+    ss << stringifyValueRecursively(rt, key) << " => "
+       << stringifyValueRecursively(rt, value) << ", ";
+  }
+
+  if (length > 0) {
+    ss.seekp(-2, ss.cur);
+  }
```
michalmaka, line 227: replace to checking if an element is the last one

michalmaka (changes requested):

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+  ss << "] ";
+
+  return ss.str();
+}
+
+std::string reanimated::stringifyJSIArrayBuffer(
+    jsi::Runtime &rt,
+    const jsi::ArrayBuffer &buf) {
+  // consider logging size or contents
```
tomekzaw, line 90:
```suggestion
  // TODO: consider logging size or contents
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+  int status;
+  const char *hostObjClassName =
+      abi::__cxa_demangle(typeid(hostObject).name(), NULL, NULL, &status);
+  if (status == 0) {
+    auto props = hostObject.getPropertyNames(rt);
+    auto propsCount = props.size();
+
+    ss << '[' << hostObjClassName << ' ';
+    if (propsCount > 0) {
+      ss << "{";
```
tomekzaw, line 116:
```suggestion
      ss << '{';
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+  } else {
+    return "[jsi::Value]";
+  }
+}
+
+std::string reanimated::stringifyJSIArray(
+    jsi::Runtime &rt,
+    const jsi::Array &arr) {
+  std::stringstream ss;
+  ss << "[";
```
tomekzaw, line 70:
```suggestion
  ss << '[';
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+  // consider logging size or contents
+  return "[ArrayBuffer]";
+}
+
+std::string reanimated::stringifyJSIFunction(
+    jsi::Runtime &rt,
+    const jsi::Function &func) {
+  std::stringstream ss;
+  auto name = func.getProperty(rt, "name").toString(rt).utf8(rt);
+  ss << "[Function " << name << "]";
```
tomekzaw, line 99:
```suggestion
  ss << "[Function " << name << ']';
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+  return ss.str();
+}
+
+std::string reanimated::stringifyJSIObject(
+    jsi::Runtime &rt,
+    const jsi::Object &object) {
+  std::stringstream ss;
+
+  // just iterate through properties
+  ss << "{";
```
tomekzaw, line 138:
```suggestion
  ss << '{';
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+std::string reanimated::stringifyValueRecursively(
+    jsi::Runtime &rt,
+    const jsi::Value &value) {
+  if (value.isBool() || value.isNumber()) {
+    return value.toString(rt).utf8(rt);
+  } else if (value.isString()) {
+    return '"' + value.getString(rt).utf8(rt) + '"';
+  } else if (value.isSymbol()) {
+    return '[' + value.getSymbol(rt).toString(rt) + ']';
```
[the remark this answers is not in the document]
t0ms0n00: Using `value.getSymbol(..).toString(..)` will return it wrapped itself.

Common/cpp/Tools/JSISerializer.h
t0ms0n00: HostObject method `getPropertyNames` is not marked `const`, so i can't use it actually.

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,201 @@
+#include "JSISerializer.h"
+#include <cxxabi.h>
+#include <sstream>
+
+namespace {
```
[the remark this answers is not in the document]
t0ms0n00: Changed due to this review comment: https://github.com/software-mansion/react-native-reanimated/pull/4857#discussion_r1281585015

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -53,39 +51,38 @@ std::string stringifyJSIFunction(jsi::Runtime &rt, const jsi::Function &func) {
 std::string stringifyJSIHostObject(
     jsi::Runtime &rt,
     jsi::HostObject &hostObject) {
-  std::stringstream ss;
   int status = -1;
   const char *hostObjClassName =
```
tomekzaw, line 55: Do we need to call `free(hostObjClassName)`?

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -53,39 +51,38 @@ std::string stringifyJSIFunction(jsi::Runtime &rt, const jsi::Function &func) {
...
+  auto lastKey = props.back().utf8(rt);
+
+  ss << '[' << hostObjClassName << ' ';
+  if (propsCount > 0) {
+    ss << '{';
+    for (auto &key : props) {
+      auto formattedKey = key.utf8(rt);
+      facebook::jsi::Value value = hostObject.get(rt, key);
+      ss << '"' << formattedKey << '"' << ": "
+         << reanimated::stringifyJSIValue(rt, value);
```
tomekzaw, line 73: Is it possible to call `stringifyJSIValue`, i.e. without `reanimated::` everywhere?

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -53,39 +51,38 @@ std::string stringifyJSIFunction(jsi::Runtime &rt, const jsi::Function &func) {
...
+  auto props = hostObject.getPropertyNames(rt);
+  auto propsCount = props.size();
+  auto lastKey = props.back().utf8(rt);
+
+  ss << '[' << hostObjClassName << ' ';
+  if (propsCount > 0) {
+    ss << '{';
+    for (auto &key : props) {
+      auto formattedKey = key.utf8(rt);
+      facebook::jsi::Value value = hostObject.get(rt, key);
```
tomekzaw, line 71:
```suggestion
      auto value = hostObject.get(rt, key);
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    if (object.isArray(rt)) {
+      return stringifyJSIArray(rt, object.getArray(rt));
+    } else if (object.isArrayBuffer(rt)) {
+      return stringifyJSIArrayBuffer(rt, object.getArrayBuffer(rt));
+    } else if (object.isFunction(rt)) {
+      return stringifyJSIFunction(rt, object.getFunction(rt));
+    } else if (object.isHostObject(rt)) {
+      return stringifyJSIHostObject(rt, *object.asHostObject(rt).get());
+    } else if (getPrototypeName(rt, object) == "[object Map]") {
```
tomekzaw, line 187: I'm afraid this can differ between JS runtimes. Do you think there's any way to get only `"Map"`? Maybe we could compare if the prototype object is equal to `Object.getPrototypeOf(new Map())` or do the same thing as we do for `Error` below?

Common/cpp/Tools/JSISerializer.cpp
michalmaka: you're right, @t0ms0n00  please wrap it with unique_ptr with custom deleter:
```
std::unique_ptr<char, decltype(&free)> hostObjClassName(abi::__cxa_demangle(typeid(hostObject).name(), NULL, NULL, &status), &free);
```

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -4,22 +4,16 @@
...
-        Logger::log(value.getString(rt).utf8(rt).c_str());
-      } else if (value.isNumber()) {
-        Logger::log(value.getNumber());
-      } else if (value.isUndefined()) {
-        Logger::log("undefined");
-      } else {
-        Logger::log("unsupported value type");
-      }
+      auto serializedValue = stringifyJSIValue(rt, value);
+      Logger::log(serializedValue);
```
tomekzaw, line 16: Sorry for another suggestion on this line but I can't resist 😄 
```suggestion
      Logger::log(stringifyJSIValue(rt, value));
```

android/src/main/cpp/AndroidLogger.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <string>
```
tomekzaw, line 3: Let's add a newline here to separate different types of headers (we do it often in our codebase).
```suggestion
#include <string>

```

Common/cpp/Tools/JSISerializer.cpp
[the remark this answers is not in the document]
t0ms0n00: Nevermind, changed in the newest version and replaced this function with simpler check.

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -5,23 +5,53 @@
...
+    jsi::Runtime &rt,
+    const jsi::Object &object,
+    const std::string &type) {
+  return getObjectTypeName(rt, object) == type;
+}
+
+static inline bool isOneOfSupportedTypes(
+    jsi::Runtime &rt,
+    const jsi::Object &object,
+    std::vector<std::string> supportedTypes) {
```
michalmaka, line 30: Avoid copying of vector

```suggestion
    const std::vector<std::string>& supportedTypes) {
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -5,23 +5,53 @@
...
+  std::string instanceType = getObjectTypeName(rt, object);
```
michalmaka, line 31:
```suggestion
  auto instanceType = getObjectTypeName(rt, object);
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -157,19 +194,36 @@ std::string stringifyJSMap(jsi::Runtime &rt, const jsi::Object &object) {
   return ss.str();
 }
 
-std::string stringifyJSIValue(jsi::Runtime &rt, const jsi::Value &value) {
+std::string JSISerializer::stringifyRecursiveType(const jsi::Object &object) {
+  std::stringstream ss;
```
michalmaka, line 207:
```suggestion
```

Common/cpp/Tools/JSISerializer.h
```diff
@@ -2,7 +2,103 @@
...
+};
+
+const std::vector<std::string> SUPPORTED_REFLECTION_TYPES = {
+    "Reflect",
+    "Proxy",
+};
+
+class JSISerializer {
+ public:
+  explicit JSISerializer(jsi::Runtime &rt)
```
michalmaka, line 66: I would prefer having all method definitions in cpp file

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -10,6 +10,11 @@
 
 namespace reanimated {
 
+static const std::function<jsi::String(jsi::Runtime &, jsi::Value const &)>
+    stringifyValue = [](jsi::Runtime &rt, jsi::Value const &value) {
+      return jsi::String::createFromUtf8(rt, stringifyJSIValue(rt, value));
+    };
```
michalmaka, line 16: could you check if something like this will work? I guess it should and it's cleaner. Same for next function

```suggestion
static jsi::String stringifyValue(jsi::Runtime &rt, jsi::Value const &value) {
      return jsi::String::createFromUtf8(rt, stringifyJSIValue(rt, value));
};
```

Common/cpp/Tools/JSISerializer.h
```diff
@@ -2,7 +2,103 @@
 
 #include <jsi/jsi.h>
 #include <string>
+#include <vector>
 
 using namespace facebook;
 
+namespace {
+const std::vector<std::string> SUPPORTED_ERROR_TYPES = {
```
michalmaka, line 10: do we need to expose those vectors? Unnamed namespace in a header makes those objects avaialble in all files it is included so no reason for unnamed namespace here

michalmaka (changes requested):

android/src/main/cpp/AndroidLogger.cpp
```diff
@@ -27,4 +27,8 @@ void AndroidLogger::log(bool b) {
   __android_log_print(ANDROID_LOG_VERBOSE, APPNAME, "%s", b ? "true" : "false");
 }
 
+void AndroidLogger::log(const std::string &str) {
+  __android_log_print(ANDROID_LOG_VERBOSE, APPNAME, "%s", str.c_str());
```
michalmaka, line 31:
```suggestion
  log(str.c_str());
```

ios/native/REAIOSLogger.mm
```diff
@@ -26,4 +26,9 @@
   NSLog(@"%@", [NSString stringWithCString:str encoding:[NSString defaultCStringEncoding]]);
 }
 
+void REAIOSLogger::log(const std::string &str)
+{
+  NSLog(@"%@", [NSString stringWithCString:str.c_str() encoding:[NSString defaultCStringEncoding]]);
```
michalmaka, line 31:
```suggestion
  log(str.c_str());
```

ios/native/REAIOSLogger.mm
```diff
@@ -26,4 +26,9 @@
   NSLog(@"%@", [NSString stringWithCString:str encoding:[NSString defaultCStringEncoding]]);
```
michalmaka, line 26:
```suggestion
  log(str);
```

android/src/main/cpp/AndroidLogger.cpp
```diff
@@ -27,4 +27,8 @@ void AndroidLogger::log(bool b) {
   __android_log_print(ANDROID_LOG_VERBOSE, APPNAME, "%s", b ? "true" : "false");
```
michalmaka, line 27:
```suggestion
  log(b ? "true" : "false");
```

Common/cpp/Tools/JSISerializer.h
```diff
@@ -0,0 +1,100 @@
+#pragma once
+
+#include <jsi/jsi.h>
+#include <string>
+#include <vector>
+
+using namespace facebook;
+
+const std::vector<std::string> SUPPORTED_ERROR_TYPES = {
```
michalmaka, line 9: can you move those vectors to cpp file?

michalmaka (changes requested):

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -72,6 +68,7 @@ void RuntimeDecorator::decorateRuntime(
 #endif // DEBUG
 
   jsi_utils::installJsiFunction(rt, "_log", logValue);
+  jsi_utils::installJsiFunction(rt, "_stringify", stringifyValue);
```
tomekzaw, line 71: As discussed offline, let's rename this function to `_toString` so it's consistent with JavaScript naming style.
```suggestion
  jsi_utils::installJsiFunction(rt, "_toString", stringifyValue);
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,329 @@
...
+  ss << "[jsi::HostObject(" << hostObjClassName << ") ";
+  std::free(hostObjClassName);
+
+  auto props = hostObject.getPropertyNames(rt_);
+  auto propsCount = props.size();
+  auto lastKey = props.back().utf8(rt_);
+
+  if (propsCount > 0) {
+    ss << '{';
+    for (auto &key : props) {
```
tomekzaw, line 100: Will it work?
```suggestion
    for (const auto &key : props) {
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,329 @@
...
+  ss << '}';
+
+  return ss.str();
+}
+
+std::string JSISerializer::stringifyJSError(const jsi::Object &object) {
+  std::stringstream ss;
+  ss << '[' << object.getProperty(rt_, "name").toString(rt_).utf8(rt_) << ": "
+     << object.getProperty(rt_, "message").toString(rt_).utf8(rt_) << ']';
+
```
tomekzaw, line 141: Not sure if we need newline here

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,329 @@
...
+std::string JSISerializer::stringifyRegExp(const jsi::Object &object) {
+  return object.getPropertyAsFunction(rt_, "toString")
+      .callWithThis(rt_, object)
+      .toString(rt_)
+      .utf8(rt_);
+}
+
+std::string JSISerializer::stringifyJSIValueRecursively(
+    const jsi::Value &value,
+    bool topLevel) {
```
tomekzaw, line 231: What does "top level" mean in this case?
```suggestion
    bool isTopLevel) {
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,329 @@
...
+  if (value.isNull()) {
+    return "null";
+  }
+  if (value.isObject()) {
+    jsi::Object object = value.asObject(rt_);
+
+    if (this->wasVisited(object)) {
+      return stringifyRecursiveType(object);
+    }
+    this->visit(object);
```
tomekzaw, line 259: I like the simplicity of this line 😄 
```suggestion
    visit(object);
```
or
```suggestion
    markAsVisited(object);
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  ss << ']';
+
+  return ss.str();
+}
+
+std::string JSISerializer::stringifyJSIFunction(const jsi::Function &func) {
+  std::stringstream ss;
+  auto name = func.getProperty(rt_, "name").toString(rt_).utf8(rt_);
```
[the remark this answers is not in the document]
t0ms0n00: I wouldn't compile, because `||` tries to convert those values to a `bool` type, that's why I've used a ternary operator below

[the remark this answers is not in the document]
t0ms0n00: WDYT about 
```
auto name = func.getProperty(rt_, "name").toString(rt_).utf8(rt_);
name = name != "" ? name : "anonymous";
```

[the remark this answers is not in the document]
tomekzaw: Perfect!

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,340 @@
...
+  ss << ']';
+
+  return ss.str();
+}
+
+std::string JSISerializer::stringifyJSIFunction(const jsi::Function &func) {
+  std::stringstream ss;
+  auto kind = (func.isHostFunction(rt_) ? "jsi::HostFunction" : "Function");
+  auto name = func.getProperty(rt_, "name").toString(rt_).utf8(rt_);
+  name = name != "" ? name : "anonymous";
```
michalmaka, line 123:
```suggestion
  name = name.empty() ? "anonymous" : name;
```

michalmaka (approved):

src/reanimated2/globals.d.ts
```diff
@@ -28,6 +28,7 @@ declare global {
     | undefined;
   var evalWithSourceUrl: ((js: string, sourceURL: string) => any) | undefined;
   var _log: (s: string) => void;
+  var _toString: (s: object) => string;
```
tomekzaw, line 31:
```suggestion
  var _toString: (value: unknown) => string;
```

Common/cpp/Tools/RuntimeDecorator.cpp
```diff
@@ -72,6 +68,7 @@ void RuntimeDecorator::decorateRuntime(
 #endif // DEBUG
 
   jsi_utils::installJsiFunction(rt, "_log", logValue);
+  jsi_utils::installJsiFunction(rt, "_toString", stringifyValue);
```
tomekzaw, line 71:
```suggestion
  jsi_utils::installJsiFunction(rt, "_toString", toStringValue);
```

ios/native/REAIOSLogger.h
```diff
@@ -9,6 +9,7 @@ class REAIOSLogger : public LoggerInterface {
   void log(double d) override;
   void log(int i) override;
   void log(bool b) override;
+  void log(const std::string &str) override;
```
tomekzaw, line 12: Maybe let's move it directly below `(const char *str)`?

app/src/examples/LogExample.tsx
```diff
@@ -0,0 +1,134 @@
...
+      test(BigInt(123456), '123456n');
+
+      test(new ArrayBuffer(42), '[ArrayBuffer]');
+      test(new DataView(new ArrayBuffer(42)), '[DataView]');
+      test(new Promise(() => {}), '[Promise]');
+      test(new Uint8Array(), '[Uint8Array]');
+      test(new Int32Array(), '[Int32Array]');
+      test(new Date(0), 'Thu Jan 01 1970 01:00:00 GMT+0100');
+      test(new Error('foo'), '[Error: foo]');
+      test(new TypeError('bar'), '[TypeError: bar]');
```
tomekzaw, line 108:
```suggestion
      test(new TypeError('foo'), '[TypeError: foo]');
```

app/src/examples/LogExample.tsx
```diff
@@ -0,0 +1,134 @@
+import { Button, StyleSheet, Text, View } from 'react-native';
+import { runOnUI, useSharedValue } from 'react-native-reanimated';
+import React from 'react';
+
+export default function LogExample() {
+  useSharedValue(42); // force Reanimated initialization
+  // const testHostObject = __reanimatedModuleProxy; // TODO: find simpler HostObject (that would not crash the app)
+  // TODO: assert(testHostObject !== undefined);
+
```
tomekzaw, line 9: Can we remove it?
```suggestion
```

Common/cpp/Tools/JSISerializer.cpp
```diff
@@ -0,0 +1,340 @@
...
+}
+
+JSISerializer::JSISerializer(jsi::Runtime &rt)
+    : rt_(rt),
+      visitedNodes_(rt_.global()
+                        .getPropertyAsFunction(rt_, "Set")
+                        .callAsConstructor(rt_)
+                        .asObject(rt_)) {}
+
+std::string JSISerializer::stringifyObjectType(const jsi::Object &object) {
```
tomekzaw, line 95:
```suggestion
std::string JSISerializer::stringifyWithName(const jsi::Object &object) {
```

Common/cpp/Tools/JSISerializer.h
```diff
@@ -0,0 +1,45 @@
...
+
+ private:
+  std::string stringifyObjectType(const jsi::Object &object);
+  std::string stringifyJSIArray(const jsi::Array &arr);
+  std::string stringifyJSIFunction(const jsi::Function &func);
+  std::string stringifyJSIHostObject(jsi::HostObject &hostObject);
+  std::string stringifyJSIObject(const jsi::Object &object);
+  std::string stringifyJSError(const jsi::Object &object);
+  std::string stringifyJSSet(const jsi::Object &object);
+  std::string stringifyJSMap(const jsi::Object &object);
```
tomekzaw, line 25:
```suggestion
  std::string stringifyArray(const jsi::Array &arr);
  std::string stringifyFunction(const jsi::Function &func);
  std::string stringifyHostObject(jsi::HostObject &hostObject);
  std::string stringifyObject(const jsi::Object &object);
  std::string stringifyError(const jsi::Object &object);
  std::string stringifySet(const jsi::Object &object);
  std::string stringifyMap(const jsi::Object &object);
```

tomekzaw (approved):
👏 Thanks for this PR!

PR #4858, opened 2023-07-31
Update issue template

Latropos:
## Summary

This is just a proposal

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -20,23 +20,44 @@ body:
     attributes:
       value: |
         ## Required information
+
   - type: textarea
     id: description
     attributes:
       label: Description
-      description: Please provide a clear, concise and descriptive explanation of what the bug is. Include screenshots or a video if needed. Tell us what were you expecting to happen instead of what is happening now.
+      description: | 
+        Please provide a clear, concise and descriptive explanation of what the bug is. 
+        Make sure you tell us:
+        - [ ] What is the **actual** behavior
+        - [ ] What is the **expected** behavior
+      value: |
+        ### What happens? 
+
+        ___
+        <details>
+        ```
+        If you have long snippets of code or logs, place them HERE
+        ```
+        </details>
     validations:
       required: true
+    id: screenshot
+    attributes:
+      label: Screenshot/Recording
+      description: If possible, please add screenshots or recording. 🙏
 
   - type: textarea
     id: steps-to-reproduce
     attributes:
       label: Steps to reproduce
-      description: Provide a detailed list of steps that reproduce the issue.
+      description: |
+       How to reproduce your issue? 
+       > **Warn**
+       > It's a common mistake to provide instructions how to run the provided repository, without telling us what to do with the app once we run it.
       placeholder: |
-        1.
-        2.
-        3.
+        1. Clone my repo & run the app
+        2. Press the red button two times
+        3. ...
     validations:
       required: true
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -145,7 +167,7 @@ body:
       description: What device you are experiencing this problem on? Specify full device name along with the version of the operating system it's running.
       placeholder: ex. Samsung Galaxy A22 (Android 12)
 
-  - type: dropdown
+  - type: checkboxes
```
tomekzaw, line 170: There's a particular reason why we chose to use dropdown here &ndash; each checkbox inside issue description is treated like a task and shown like this in the issues list:

Zrzut ekranu 2023-08-1 o 00 00 40

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -20,23 +20,44 @@ body:
...
   - type: textarea
     id: description
     attributes:
       label: Description
-      description: Please provide a clear, concise and descriptive explanation of what the bug is. Include screenshots or a video if needed. Tell us what were you expecting to happen instead of what is happening now.
+      description: | 
+        Please provide a clear, concise and descriptive explanation of what the bug is. 
+        Make sure you tell us:
+        - [ ] What is the **actual** behavior
+        - [ ] What is the **expected** behavior
```
tomekzaw, line 32: Have you checked how does it look like in the final rendered issue template? (You can check it on some private repository).

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -20,23 +20,44 @@ body:
...
+      value: |
+        ### What happens? 
+
+        ___
+        <details>
+        ```
+        If you have long snippets of code or logs, place them HERE
+        ```
+        </details>
```
tomekzaw, line 41: Not a huge fan of using UPPERCASE letters for whole words.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -20,23 +20,44 @@ body:
...
     validations:
       required: true
+    id: screenshot
+    attributes:
+      label: Screenshot/Recording
+      description: If possible, please add screenshots or recording. 🙏
```
tomekzaw, line 47: We don't use emojis anywhere, why should we use it here?

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -20,23 +20,44 @@ body:
...
 
   - type: textarea
     id: steps-to-reproduce
     attributes:
       label: Steps to reproduce
-      description: Provide a detailed list of steps that reproduce the issue.
+      description: |
+       How to reproduce your issue? 
+       > **Warn**
```
tomekzaw, line 55: Again, some custom formatting that won't work.

tomekzaw (changes requested):

.github/ISSUE_TEMPLATE/bug-report.yml
Latropos: Yes, but github is going to fix this soon: https://github.com/orgs/community/discussions/5197#discussioncomment-5966864

.github/ISSUE_TEMPLATE/bug-report.yml
Latropos: To get some attention. I think we should try to force people to read the description. Sometimes we get bug descriptions, that consist of error log without any explanation

tjzel: I like the emoji but it should be without the trailing dot on the sentence

PR #4860, opened 2023-07-31
Add test for Reanimated mocks

tjzel:
## Summary

Adding a simple test suite that checks if Reanimated mocks are working properly (can be imported etc.)

## Test plan

Run the test.


tjzel:
Need to consider how to do it better

Issue #4802, opened 2023-07-24
Strict mode compat (web) by removing findNodeHandle


natew:
### Description

On web you get errors:

```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of AnimatedComponent(View) which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here
```

This is due to `findNodeHandle` which according to react-native-web is deprecated:

https://github.com/necolas/react-native-web/blob/ef95fc0c2b40cc0d3743df4b56485a16a8442e5f/packages/react-native-web/src/exports/findNodeHandle/index.js#L17

Would be nice to remove this usage. I can help with this with just some light pointers in terms of what thats needed for now.

### Steps to reproduce

Use any animated component on web, sorry for no repro but easy to trigger.

### Snack or a link to a repository

https://snack.expo.io/

### Reanimated version

3.3.0

### React Native version

0.72.0

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

tomekzaw:
Hey @natew, thanks for pointing out this. Someone from our team will take a look at it soon.

m-bert:
Hi @natew! [This PR](https://github.com/software-mansion/react-native-reanimated/pull/4866) should remove all calls of `findNodeHandle` on web. Could you please test it and confirm that it solves this issue?

natew:
It does, thank you a bunch, will look forward to the release with it. I'm not positive but my guess is our drop in performance/lighthouse recently on the site was due to this causing React to de-opt concurrent rendering which makes our main thread work go up quite a bit. Testing that next..

asenmitrev:
@m-bert Seems `findNodeHandle` is still used in https://github.com/software-mansion/react-native-reanimated/blob/main/src/createAnimatedComponent/JSPropUpdater.ts#L45 , as well as some other places. Is this an issue with my setup that `JSPropUpdater` is used at all?

m-bert:
Hi @asenmitrev! It is possible that something has changed and now `JSPropUpdater` is used on web. I'll look at it and get back to you.

asenmitrev:
@m-bert checked on NextExample in the PR branch and seems the warning is gone. All seems good, thank you!

apetta:
Still seem to be getting this on 3.5.4

[...]

PR #4866, opened 2023-08-02
Fix `findDOMNode` deprecated on web inside `StrictMode`

m-bert:
## Summary

When `StrictMode` is enabled, web version of reanimated shows error that `findDOMNode` is deprecated. This error comes from usage of `findNodeHandle` function. This PR removes usages of `findNodeHandle` on web version, so that error no longer appears.

This PR also introduces a lot of type casting. On web version `_component` and `viewTag` are `HTMLElements` (i.e. divs that we animate), hence I had to change types of these properties to also include `HTMLElement`. It resulted in many type casts. Casts to `number` are used mainly because code that requires `number` is not called on web. 

## Test plan

Tested on our example app.


src/createAnimatedComponent.tsx
```diff
@@ -643,7 +648,7 @@ export default function createAnimatedComponent(
             const sharedElementTransition =
               this.props.sharedTransitionStyle ?? new SharedTransition();
             sharedElementTransition.registerTransition(
```
piaskowyk, line 650: We shouldn't even try to register SharedElementTransition on web

piaskowyk (commented):
We use `findNodeHandle` also in many other places in the code. What we should do with the rest of them?

src/createAnimatedComponent.tsx
m-bert: I've changed the if condition ([3dec3e8](https://github.com/software-mansion/react-native-reanimated/pull/4866/commits/3dec3e89302ad5aa1bba9382f020d515ed95b6b5)), so it shouldn't be done on web now. 

When it comes to `findNodeHandle`, I'm aware that we use it in many other places, however I was focused on finding usages that are causing problems on web. I think that getting rid of this function should be discussed on a meeting, since it affects not only web version of reanimated, but also android and iOS.

src/createAnimatedComponent.tsx
```diff
@@ -602,14 +604,18 @@ export default function createAnimatedComponent(
...
-    _setComponentRef = setAndForwardRef<Component>({
+    _setComponentRef = setAndForwardRef<Component | HTMLElement>({
       getForwardedRef: () =>
         this.props.forwardedRef as MutableRefObject<
           Component<Record<string, unknown>, Record<string, unknown>, unknown>
         >,
       setLocalRef: (ref) => {
-        // TODO update config
-        const tag = findNodeHandle(ref);
+        // TODO update config)
```
tjzel, line 613:
```suggestion
        // TODO update config
```

tjzel (approved):
Looks good but I'd like you to add some more commentary about all the type casts done in this PR so if someone `git blame` this code in the future there would be a clear explanation.

src/createAnimatedComponent.tsx
m-bert: Done in [01d5140](https://github.com/software-mansion/react-native-reanimated/pull/4866/commits/01d5140f66d673675c05b685a311bf18fd8b5293), though I'm not sure whether provided explanation is sufficient. Could you have a look at it?

tjzel (approved):
types are solid

tjzel (approved):

src/createAnimatedComponent.tsx
```diff
@@ -311,13 +311,13 @@ export default function createAnimatedComponent(
     _getEventViewRef() {
       // Make sure to get the scrollable node for components that implement
       // `ScrollResponder.Mixin`.
-      return this._component?.getScrollableNode
-        ? this._component.getScrollableNode()
+      return (this._component as ComponentRef)?.getScrollableNode
+        ? (this._component as ComponentRef).getScrollableNode!()
```
tomekzaw, line 315: What about `.getScrollableNode?.()`?

src/createAnimatedComponent.tsx
```diff
@@ -402,25 +402,28 @@ export default function createAnimatedComponent(
...
-      const component = this._component?.getAnimatableRef
-        ? this._component.getAnimatableRef()
-        : this;
+      const component =
+        'getAnimatableRef' in this._component!
+          ? this._component.getAnimatableRef!()
+          : this;
       if (isWeb()) {
-        viewTag = findNodeHandle(component);
+        // At this point I assume, that _setComponentRef was already called and _component is set.
```
tomekzaw, line 424:
```suggestion
        // At this point I assume that `_setComponentRef` was already called and `_component` is set.
```

src/createAnimatedComponent.tsx
```diff
@@ -402,25 +402,28 @@ export default function createAnimatedComponent(
...
+        // this._component on web represents HTMLElement of our component, that's why we use casting
```
tomekzaw, line 425:
```suggestion
        // `this._component` on web represents HTMLElement of our component, that's why we use casting
```

src/createAnimatedComponent.tsx
```diff
@@ -602,14 +605,18 @@ export default function createAnimatedComponent(
...
+    _setComponentRef = setAndForwardRef<Component | HTMLElement>({
       getForwardedRef: () =>
         this.props.forwardedRef as MutableRefObject<
           Component<Record<string, unknown>, Record<string, unknown>, unknown>
         >,
       setLocalRef: (ref) => {
         // TODO update config
-        const tag = findNodeHandle(ref);
+
+        const tag = isWeb()
```
tomekzaw, line 616: Let's add `const IS_WEB = isWeb();` somewhere at the top of the file and use memoized value here:
```suggestion
        const tag = IS_WEB
```

src/createAnimatedComponent.tsx
```diff
@@ -639,11 +646,11 @@ export default function createAnimatedComponent(
               maybeBuild(exiting)
             );
           }
-          if (sharedTransitionTag) {
+          if (sharedTransitionTag && !isWeb()) {
```
tomekzaw, line 649:
```suggestion
          if (sharedTransitionTag && !IS_WEB) {
```

src/createAnimatedComponent.tsx
m-bert: If I understood correctly, you want to change it to:

```jsx
      return (this._component as ComponentRef)?.getScrollableNode
        ? (this._component as ComponentRef).getScrollableNode?.()
        : this._component;
```
In that case I think we can keep non-null assertion, since the condition checks whether `getScrollableNode` exists. It would work if we could do something like:

```jsx
      return (this._component as ComponentRef).getScrollableNode?.();
```

however in that case we lose return value from else (`:`)

piaskowyk:
Does this PR fix https://github.com/software-mansion/react-native-reanimated/issues/4802 ?

m-bert:
@piaskowyk I've checked that on our example app and it seems to fix the issue that you've mentioned. I can check that one more time if you'd like to.

src/createAnimatedComponent.tsx
tomekzaw: I just wanted to get rid of the exclamation mark in `getScrollableNode!`.

src/createAnimatedComponent.tsx
```diff
@@ -402,25 +404,28 @@ export default function createAnimatedComponent(
...
       // Component can specify ref which should be animated when animated version of the component is created.
       // Otherwise, we animate the component itself.
-      const component = this._component?.getAnimatableRef
-        ? this._component.getAnimatableRef()
-        : this;
-      if (isWeb()) {
-        viewTag = findNodeHandle(component);
+      const component =
+        'getAnimatableRef' in this._component!
+          ? this._component.getAnimatableRef!()
```
tomekzaw, line 423: Is it possible to somehow convert this snippet of code so that we don't have to use `!`? This is meant for "unsafe" operations when we are sure that the property exists (e.g. JSI bindings injected from C++ side). Can we use TypeScript to ensure that this property exists (either with ternary conditional or using `?`)?

tomekzaw (approved):
Let's merge this PR since it resolves #4802 (see https://github.com/software-mansion/react-native-reanimated/issues/4802#issuecomment-1675637834). Then, let's submit another PR that splits `_component` field with multiple types into separate fields of concrete types, one for each platform.

PR #4867, opened 2023-08-02
Bump React Native to 0.72.3

tomekzaw:
## Summary

This PR bumps react-native to 0.72.3 in Example and FabricExample apps.

Turns out `Device.flow.js` from `metro-inspector-proxy` is unused so I only updated the patch for `Device.js` and removed diff for `Device.flow.js`.

## Test plan

Check if CI is green.


piaskowyk (commented):
Looks good to me 👍 just to be sure - have you checked if it works with debugger? 
And also we should update link to metro-inspector-proxy patch in the documentation - https://docs.swmansion.com/react-native-reanimated/docs/guide/debugging#chrome-devtools

tomekzaw:
> have you checked if it works with debugger?

Yes, I did and it works fine.

> we should update link to metro-inspector-proxy patch in the documentation

That's true. Maybe we can even provide a list of patches for various versions of RN? 🤔

tomekzaw:
Updated docs in c16c3a0425dfd2f2424e69032983fd61d4c215b3 according to @piaskowyk's suggestion:

Zrzut ekranu 2023-08-3 o 11 38 13

piaskowyk (approved):
👏

PR #4868, opened 2023-08-02
Update globals in plugin

tomekzaw:
## Summary

This PR adds the names of missing standard built-in objects (e.g. `BigInt` or `WeakSet`) in Reanimated Babel plugin as well as removes redundant ones (e.g. `LayoutAnimationRepository` or `__ErrorUtils`).

Based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects.

## Test plan

Verify each name one-by-one.


plugin/src/commonObjects.ts
```diff
@@ -1,63 +1,141 @@
...
+  // 'Intl.ListFormat',
+  // 'Intl.Locale',
+  // 'Intl.NumberFormat',
+  // 'Intl.PluralRules',
+  // 'Intl.RelativeTimeFormat',
+  // 'Intl.Segmenter',
+
+  // Other stuff
   'null',
-  'UIManager',
```
tjzel, line 26: is this legacy?

plugin/src/commonObjects.ts
```diff
@@ -1,63 +1,141 @@
...
+  'AsyncGenerator',
+  'AsyncFunction',
+
+  // Reflection
+  'Reflect',
+  'Proxy',
+
+  // Internationalization
+  'Intl',
+  // 'Intl.Collator',
```
tjzel, line 97: Why is this commented out?

plugin/src/commonObjects.ts
```diff
@@ -1,63 +1,141 @@
...
   '_scheduleOnJS',
   '_makeShareableClone',
   '_updateDataSynchronously',
-  'eval',
   '_updatePropsPaper',
   '_updatePropsFabric',
   '_removeFromPropsRegistry',
-  'RegExp',
-  'Error',
-  '__ErrorUtils',
```
tjzel, line 51: Are you sure we can get rid of __ErrorUtils here?

plugin/src/commonObjects.ts
tomekzaw: It looks so, this line was introduced in "༼ つ ◕_◕ ༽つ Give Reanimated 2" commit and currently there's no occurrences of `UIManager` in any file with extension `*.ts,*.tsx,*.js`.

plugin/src/commonObjects.ts
tomekzaw: If `Intl.Collator` appears in worklet body, our plugin will capture only `Intl`, thus it doesn't make sense to allowlist `Intl.Collator`.

plugin/src/commonObjects.ts
tomekzaw: Yep, it's only used as `global.__ErrorUtils` so the plugin would capture only `global` but this is already allowlisted so it won't. I've tested it using one demo from WorkletExample.tsx which throws JS Errors on the UI runtime.

tjzel (approved):
🎬

PR #4869, opened 2023-08-02
Assert shareable type

tomekzaw:
## Summary

This PR replaces a few occurrences of `extractShareableOrThrow` with `extractShareableOrThrow<T>` which asserts the type of the shareable along with a proper error message.

### DIFF
--- a/Common/cpp/AnimatedSensor/AnimatedSensorModule.cpp
+++ b/Common/cpp/AnimatedSensor/AnimatedSensorModule.cpp
@@ -26,7 +26,8 @@ jsi::Value AnimatedSensorModule::registerSensor(
     const jsi::Value &sensorDataHandler) {
   SensorType sensorType = static_cast<SensorType>(sensorTypeValue.asNumber());
 
-  auto shareableHandler = extractShareableOrThrow(rt, sensorDataHandler);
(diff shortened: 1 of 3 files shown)


tjzel (approved):
Looks good but `extractShareableOrThrow` allows for `undefined`. How about we change the error message from `x must be y` to `defined x must be y`?

tomekzaw:
@tjzel `extractShareableOrThrow` indeed returns `Shareable::undefined()` for `jsi::Value::undefined()` but then `extractShareableOrThrow<ShareableWorklet>` will fail because `Shareable::undefined()` is a `ShareableScalar` of type `UndefinedType` so it cannot be dynamic_casted to `ShareableWorklet`.

tjzel (approved):
you're on 🔥 
better call the 🚒

tomekzaw:
Thanks to @tjzel's for motivating me to test out this change when `undefined` is passed instead of a real worklet.

It works as expected but we found another problem that the error message from C++ exception wasn't shown in RedBox.

Fixed in 64fa407472f08cc3206a9f2b1785546c9b2f70b8.

piaskowyk (approved):

PR #4879, opened 2023-08-03
[SET] Run animation only for source view

piaskowyk:
## Summary

In the previous implementation, both the source view and target shared views were animated during Shared Transition. This PR modifies the logic so that only the source view is animated now. This approach has several advantages:
- It reduces the number of views that need to be animated.
- It eliminates potential desynchronisation issues between the views.
- It fixes some cache problems related to expo-image.

## Test plan

Run example with SET from Example app.


ios/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -15,6 +15,7 @@ @implementation REASharedTransitionManager {
   NSMutableArray<UIView *> *_addedSharedViews;
   BOOL _isSharedTransitionActive;
   NSMutableArray<REASharedElement *> *_sharedElements;
```
tomekzaw, line 17: Does it still make sense to have `_sharedElements` since we now have `_sharedElementsLookup`?

piaskowyk: Yes, it does. The `_sharedElements` is an array that maintains the views in the correct order to preserve the proper z-index. The `_sharedElementsLookup` is used to speed up the search operation within the shared element set. However, the lookup table is not necessary, and we can remove it by replacing it with a simple iteration over the array. It will be just a bit slower.

tomekzaw (approved):
🚀

Issue #4631, opened 2023-06-26
Native side not loading with custom AppDelegate.swift


danielRChildrens:
### Description

I am receiving the following error after converting the AppDelegate.mm to AppDelegate.swift:

Error: [Reanimated] The native part of Reanimated doesn't seem to be initialized.

AppDelegate.swift
```
import Foundation

@UIApplicationMain
class AppDelegate: UIResponder, UIApplicationDelegate {
  var window: UIWindow?
  var bridge: RCTBridge!

  func application(_ application:UIApplication, didFinishLaunchingWithOptions launchOptions:[UIApplication.LaunchOptionsKey : Any]?) -> Bool {
    let jsCodeLocation: URL
    
    jsCodeLocation = RCTBundleURLProvider.sharedSettings().jsBundleURL(forBundleRoot: "index")

    let rootView = RCTRootView(bundleURL: jsCodeLocation, moduleName: "DigitalFrontDoor", initialProperties: nil, launchOptions: launchOptions)

    let rootViewController = UIViewController()
    rootViewController.view = rootView
    self.window = UIWindow(frame: UIScreen.main.bounds)
    self.window?.rootViewController = rootViewController
    self.window?.makeKeyAndVisible()
    
    return true
    
  }


  /// This method controls whether the `concurrentRoot`feature of React18 is turned on or off.
  ///
  /// @see: https://reactjs.org/blog/2022/03/29/react-v18.html
  /// @note: This requires to be rendering on Fabric (i.e. on the New Architecture).
  /// @return: `true` if the `concurrentRoot` feature is enabled. Otherwise, it returns `false`.
  func concurrentRootEnabled() -> Bool {
    return true
  }
}
```

Notes:
- The iOS build is completing successfully
- Android is working as expected
- I have tried removing the Pods, node_modules, build folders and reinstalling everything.
- I have tried restarting my computer
- Converting the AppDelegate to swift works without Reanimated
- I have followed this post for the AppDelegate swift migration (https://ospfranco.com/post/2021/07/31/migrate-rn-appdelegate-to-swift/)

### Steps to reproduce

I have two Branches as follows:

- Branch 1 (reanimated): Regular AppDelegate with Reanimated, In this branch everything works as expected in both platforms
- Branch 2 (swift-app-delegate): Swift AppDelegate without Reanimated, In this branch everything works as expected in both platforms

[...]

PR #4880, opened 2023-08-03
Mock REAInitializer for backward compatibility

piaskowyk:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/4631

In https://github.com/software-mansion/react-native-reanimated/pull/4619 I removed REAInitializer because it is no longer needed. However, some brownfield apps still use it. This PR adds a mock for REAInitializer to ensure backward compatibility.

## Test plan

CI


tjzel (approved):
🎬

PR #4893, opened 2023-08-07
Update `useAnimatedGestureHandler` in docs

m-bert:
## Summary

Since `2.10.0` gesture handler enables new web implementation by default. Because of that `useAnimatedGestureHandler` doesn't work unless both `onGestureEvent` and `onHandlerStateChange` props are defined. This PR updates docs entry for `useAnimatedGestureHandler` so that people will be aware of that change.

### DIFF
--- a/docs/docs/api/hooks/useAnimatedGestureHandler.md
+++ b/docs/docs/api/hooks/useAnimatedGestureHandler.md
@@ -61,6 +61,8 @@ const App = () => {
 The hook returns a handler object that can be attached to one of the gesture handler components provided by the `react-native-gesture-handler` library.
 The handler should be passed under `onGestureEvent` parameter regardless of what gesture states we are interested in processing.
 
+Note: Since version `2.10.0` of `react-native-gesture-handler`, to use this hook on web, you have to pass returned handler object to both `onGestureEvent` and `onHandlerStateChange` parameters.
+
 ## Example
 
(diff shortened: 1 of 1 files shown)


docs/docs/api/hooks/useAnimatedGestureHandler.md
```diff
@@ -61,6 +61,8 @@ const App = () => {
 The hook returns a handler object that can be attached to one of the gesture handler components provided by the `react-native-gesture-handler` library.
 The handler should be passed under `onGestureEvent` parameter regardless of what gesture states we are interested in processing.
 
+Note: Since version `2.10.0` of `react-native-gesture-handler`, to use this hook on web, you have to pass returned handler object to both `onGestureEvent` and `onHandlerStateChange` parameters.
```
tomekzaw, line 64: 1. Do you think it's worth converting this paragraph into a info block (i.e. `:::info`)?
2. Version numbers look better without monospace font.
```suggestion
:::info
Since version 2.10.0 of `react-native-gesture-handler`, to use this hook on web, you have to pass returned handler object to both `onGestureEvent` and `onHandlerStateChange` parameters.
:::
```

m-bert: Initially I wanted to put it inside `:::caution` block, but I've been told that it will look better as a plain text with "note" prefix. Now as we are discussing it, do you think it is better to put this inside `info` or `caution`? I prefer the second one since this change can actually break some of already existing apps.

PR #4894, opened 2023-08-07
Add `isWorklet` function

tomekzaw:
## Summary

This PR adds `isWorklet` function which is meant to be used internally as well as by other libraries in order to check whether the input value is a worklet (i.e. is a function with `__workletHash` property added by Reanimated Babel plugin).

Once worklet types are fixed, `isWorklet` should mark the input value as worklet, i.e. we'll need to change its signature to `export function isWorklet(value: unknown): value is WorkletType`.

EDIT from @tjzel:

In order to be able to use the new function `isWorklet` properly I had to fix a bunch of types with regard to worklet types.

## Test plan

Check that TS CI passes and all the ifs with `.__workletHash` were changed properly.


src/reanimated2/utils.ts
```diff
@@ -32,3 +32,8 @@ export function isSharedValue<T>(value: any): value is SharedValue<T> {
   'worklet';
   return value?._isReanimatedSharedValue === true;
 }
+
+export function isWorklet(value: unknown) {
+  // TODO: mark value as worklet
+  return typeof value === 'function' && '__workletHash' in value;
```
kmagiera, line 38: this implementation changes the behavior. We should only test for the value of `__workletHash` here, I think that if you take any HostFunction it will pass this check here because it is a function and `in` doesn't work properly for JSI objects and always return true

tjzel: After dealing with HostObjects I agree that `value.__workletHash` should be the only check here.

tomekzaw: @tjzel Another benefit of `isWorklet` function would be that we could assume that the provided value is of a worklet type. Feel free to perform a friendly take-over of this PR or close it.

piaskowyk (approved):

src/reanimated2/hook/useAnimatedProps.ts
```diff
@@ -1,15 +1,36 @@
...
+import type { DependencyList, UseAnimatedStyleInternal } from './commonTypes';
 import { shouldBeUseWeb } from '../PlatformChecker';
+import type { AnimatedPropsAdapterFunction } from '../commonTypes';
 
-// TODO: we should make sure that when useAP is used we are not assigning styles
-// when you need styles to animated you should always use useAS
-// TODO TYPESCRIPT This is a temporary cast to get rid of .d.ts file.
+type UseAnimatedProps = <Props extends object>(
+  updater: () => Partial<Props>,
+  deps?: DependencyList | null,
```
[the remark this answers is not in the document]
tjzel: Yes, we can unify it, good catch here.

src/reanimated2/commonTypes.ts
```diff
@@ -85,14 +91,32 @@ interface WorkletBaseRelease extends WorkletBaseCommon {
...
 
 export type WorkletFunction<
   Args extends unknown[] = unknown[],
   ReturnValue = unknown
 > = ((...args: Args) => ReturnValue) & (WorkletBaseRelease | WorkletBaseDev);
 
+export function isWorklet<
+  BuildType extends WorkletBaseDev | WorkletBaseRelease = WorkletBaseDev
+>(value: unknown): value is WorkletFunction & BuildType {
+  return !!(value as Record<string, unknown>).__workletHash;
```
tomekzaw, line 108: Let's add a comment why we use `.` instead of `in`

src/reanimated2/hook/useAnimatedProps.ts
```diff
@@ -1,15 +1,36 @@
...
-import type { DependencyList } from './commonTypes';
-import type {
-  AnimatedPropsAdapterFunction,
-  useAnimatedPropsType,
-} from '../helperTypes';
+import type { DependencyList, UseAnimatedStyleInternal } from './commonTypes';
 import { shouldBeUseWeb } from '../PlatformChecker';
+import type { AnimatedPropsAdapterFunction } from '../commonTypes';
 
-// TODO: we should make sure that when useAP is used we are not assigning styles
```
tomekzaw, line 10: restore plz

src/reanimated2/hook/useAnimatedProps.ts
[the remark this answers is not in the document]
tjzel: At the moment I will only change `deps` to `dependencies` - I have an idea how to make it better, but it should be a separate PR.

PR #4912, opened 2023-08-10
@jfedak/mac os

jfedak:


Saadnajmi:
Excited to see this!

apple/REAUIKit.h
```diff
@@ -0,0 +1,21 @@
+#if !TARGET_OS_OSX
+
+#import <UIKit/UIKit.h>
+
+#define REAUIView UIView
```
Saadnajmi, line 5: Note: We have a desire to convert all our '#define's in React Native macOS to typedefs, so as to move the check to compile time where the IDE can find more errors. If this was inspired by our repo, know that I'm probably going to change it eventually :D

.github/workflows/macos-build.yml
```diff
@@ -0,0 +1,75 @@
...
+        run: yarn install --frozen-lockfile
+
+      - name: Restore Pods from cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ${{ env.WORKING_DIRECTORY }}/macos/Pods
+            ~/Library/Caches/CocoaPods
+            ~/.cocoapods
+          key: ${{ runner.os }}-pods-${{ env.WORKING_DIRECTORY }}-${{ hashFiles(format('{0}/macos/Podfile.lock', env.WORKING_DIRECTORY)) }}
```
tomekzaw, line 59: Is it the same key as for iOS workflow?

.github/workflows/macos-build.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Install Reanimated node_modules
+        run: yarn install --frozen-lockfile
+
+      - name: Restore app node_modules from cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.WORKING_DIRECTORY }}/node_modules
+          key: ${{ runner.os }}-node-modules-${{ env.WORKING_DIRECTORY }}-${{ hashFiles(format('{0}/yarn.lock', env.WORKING_DIRECTORY)) }}
+          restore-keys: ${{ runner.os }}-node-modules-${{ env.WORKING_DIRECTORY }}-
```
tomekzaw, line 46: Is it the same key as for iOS workflow?

apple/LayoutReanimation/REAAnimationsManager.h
```diff
@@ -4,6 +4,8 @@
 #import <RNReanimated/REASnapshot.h>
 #import <React/RCTUIManager.h>
 
+#import <RNReanimated/REAUIKit.h>
```
tomekzaw, line 7: Let's move it directly below `#import <RNReanimated/REASnapshot.h>`

apple/LayoutReanimation/REAScreensHelper.m
```diff
@@ -1,12 +1,14 @@
 #import <RNReanimated/REAScreensHelper.h>
```
tomekzaw, line 2:
```suggestion
```

apple/LayoutReanimation/REASnapshot.m
```diff
@@ -120,6 +124,23 @@ - (UIView *)findTransformedView:(UIView *)view
   return nil;
 }
 
+#else // TARGET_OS_OSX [
```
tomekzaw, line 127:
```suggestion
#else // TARGET_OS_OSX
```

apple/LayoutReanimation/REASnapshot.m
```diff
@@ -120,6 +124,23 @@ - (UIView *)findTransformedView:(UIView *)view
...
+      // TODO: macOS implementation
```
tomekzaw, line 128:
```suggestion
// TODO macOS
```

apple/READisplayLink.h
```diff
@@ -0,0 +1,21 @@
+#if !TARGET_OS_OSX
+
+#import <QuartzCore/CADisplayLink.h>
+
+#define READisplayLink CADisplayLink
```
tomekzaw, line 5: Can we use `typedef` instead?

apple/REAUIKit.h
tomekzaw: @Saadnajmi Thanks for pointing this out! Do you think we can use typedefs now or should we wait for the changes to appear in react-native-macos first?

apple/keyboardObserver/REAKeyboardEventObserver.m
```diff
@@ -58,6 +63,19 @@ - (void)unsubscribeFromKeyboardEvents:(int)listenerId
 {
   NSLog(@"Keyboard handling is not supported on tvOS");
 }
+
+#elif TARGET_OS_OSX
```
tomekzaw, line 67:
```suggestion
#elif TARGET_OS_OSX

```

apple/native/NativeProxy.mm
```diff
@@ -175,8 +177,13 @@ static CFTimeInterval calculateTimestampWithSlowAnimations(CFTimeInterval curren
   auto maybeFlushUIUpdatesQueueFunction = [nodesManager]() { [nodesManager maybeFlushUIUpdatesQueue]; };
 
   auto requestRender = [nodesManager](std::function<void(double)> onRender, jsi::Runtime &rt) {
-    [nodesManager postOnAnimation:^(CADisplayLink *displayLink) {
+    [nodesManager postOnAnimation:^(READisplayLink *displayLink) {
+#if !TARGET_OS_OSX
       double frameTimestamp = calculateTimestampWithSlowAnimations(displayLink.targetTimestamp) * 1000;
+#else
+      // TODO: get targetTimestamp on macOS
+      double frameTimestamp = calculateTimestampWithSlowAnimations(displayLink.timestamp + displayLink.duration) * 1000;
```
tomekzaw, line 185: Can we split this logic so that we call `calculateTimestampWithSlowAnimations` only once outside of `#ifdef`?

apple/sensor/ReanimatedSensor.h
```diff
@@ -1,4 +1,4 @@
-#if __has_include(<CoreMotion/CoreMotion.h>)
+#if __has_include(<CoreMotion/CoreMotion.h>) && !TARGET_OS_OSX
```
tomekzaw, line 1: Do you think we could use `#if !TARGET_OS_TV && !TARGET_OS_OSX` here as well for consistency?

tomekzaw (changes requested):
1. Let's cleanup imports so that all `#import <RNReanimated/<...>.h>` are in single place and sorted alphabetically.
2. Let's mark all spots where we will need to do something for macOS support with `// TODO macOS` (analogously to `// TODO TYPESCRIPT`).

apple/REAUIKit.h
Saadnajmi: I think you want to have your own type? Otherwise your iOS code will depend on React Native macOS if I'm not mistaken?

tomekzaw (approved):
Works for me! 👏 

Zrzut ekranu 2023-08-17 o 13 00 41

MacOSExample/App.tsx
```diff
@@ -1,118 +1,3 @@
-/**
- * Sample React Native App
- * https://github.com/facebook/react-native
- *
- * @format
- */
+import BokehExample from '../app/src/examples/BokehExample';
```
piaskowyk, line 1: We could potentially use the JavaScript stack as a substitute for the unavailable native stack. By doing so, we would be able to make use of all the examples provided here. However, implementing this change is beyond the scope of the current PR. Let's do it in a separate PR to avoid mixing concerns.

piaskowyk (approved):

PR #4923, opened 2023-08-14
Fix `AnimatedStyle` for good

tjzel:
## Summary

After many attempts to hotfix `AnimatedStyle` and always finding something faulty about it I decided to rewrite it to make it closer to RN types.

Key changes here are in `helperTypes` file - I removed manual types for `AnimatedStyle` for a more generic approach. I included two utility types: `MaybeSharedValue` and `MaybeSharedValueRecursive` which take a generic argument and if it's an object - it makes its properties possibly SharedValues, if it's an array - the same for its items, and if it's a primitive - Shared Value of a primitive, if it's supported as declared in `AnimatableValue` type.

With this change, `AnimatedStyle` is just recursively processed `Style` of a given component and no more logic is required.

Consequently I fixed typing of `useAnimatedStyle`, removed type casts from there and used TS function overloading to differentiate between user usage and developer usage (look at the code).

## Test plan

I added 500 lines of type tests, many of them fail without this PR.

## Note

During work on this PR I noticed that not all viable usages of inline Animated Styles are supported - I will work on it in the future.


src/reanimated2/helperTypes.ts
```diff
@@ -8,74 +7,58 @@ This will not be easy though!
...
-    ? AnimatedTransform
-    : S[K] extends ReadonlyArray<any>
-    ? ReadonlyArray<AnimateStyle<S[K][0]>>
-    : S[K] extends object
-    ? AnimateStyle<S[K]>
-    : S[K] extends ColorValue | undefined
-    ? S[K] | number
-    : S[K] | SharedValue<AnimatableValue>;
-};
+type MaybeSharedValueRecursive<Value> = Value extends (infer InnerValue)[]
```
Latropos, line 39: It would be more readable if we show that we are trying to obtain an array element
```suggestion
type MaybeSharedValueRecursive<Value> = Value extends Array<infer ArrayElement>
```

src/reanimated2/helperTypes.ts
```diff
@@ -8,74 +7,58 @@ This will not be easy though!
...
+  ?
+      | SharedValue<InnerValue[]>
+      | (MaybeSharedValueRecursive<InnerValue> | InnerValue)[]
+  : Value extends object
+  ?
+      | SharedValue<Value>
+      | { [K in keyof Value]: MaybeSharedValueRecursive<Value[K]> | Value[K] }
+  : MaybeSharedValue<Value>;
```
Latropos, line 47: Should `InnerValue[]` and `Value` be the same type?
So we could move the common part above the conditional.
WDYT about this type:
```
type MaybeSharedValueRecursive<Value> =
  | (Value extends AnimatableValue ? SharedValue<Value> : never)
  | (Value extends Array<infer ArrayElement>
      ? (MaybeSharedValueRecursive<ArrayElement> | ArrayElement)[]
      : Value extends object
      ? {
          [K in keyof Value]: MaybeSharedValueRecursive<Value[K]> | Value[K];
        }
      : Value);
      
```

I've noticed that according to definition of `MaybeSharedValue<Value>` we can reduce it to `Value` (since we already have `SharedValue<Value>` in our intersection.

src/reanimated2/helperTypes.ts
```diff
@@ -118,22 +97,19 @@ type SharedTransitionProps = {
   sharedTransitionStyle?: SharedTransition;
 };
 
-type AnimatedPropsProp<P extends object> = NonStyleAnimatedProps<P> &
-  JustStyleAnimatedProp<P> &
-  StyleAnimatedProps<P> &
+type AnimatedPropsProp<Props extends object> = NonStyleAnimatedProps<Props> &
```
Latropos, line 100: I am afraid that `AnimatedPropsProp` may be a confusing name

src/reanimated2/commonTypes.ts
```diff
@@ -39,9 +39,9 @@ export interface StyleProps extends ViewStyle, TextStyle {
 
 export interface SharedValue<T> {
   value: T;
```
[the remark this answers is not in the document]
tjzel: This PR aims to only change `AnimatedStyle` and its behavior. After looking into TS guides from Expensify I took quite a fancy of prepending generic arguments with `T`. I will make the change here but I want to make this PR as concise as possible.

src/reanimated2/helperTypes.ts
```diff
@@ -84,20 +67,16 @@ type EntryOrExitLayoutType =
...
+    AnimatedStyle<PickStyleProps<P>[K]>
   >;
 };
 
-type JustStyleAnimatedProp<P extends object> = {
-  style?: StyleProp<AnimatedStyle<StylesOrDefault<P>>>;
-};
-
 type NonStyleAnimatedProps<P extends object> = {
   [K in keyof Omit<P, keyof PickStyleProps<P> | 'style'>]:
```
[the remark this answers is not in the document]
tjzel: There was a change to `PickStyleProps` so it's no longer the case:
Screenshot 2023-08-28 at 14 00 47

src/reanimated2/helperTypes.ts
tjzel: That's just a syntax preference here and honestly I'm leaning more towards the `[]` notation.

src/reanimated2/helperTypes.ts
tjzel: 1. AnimatableValue definitely cannot be first because of this:
Screenshot 2023-08-28 at 14 05 49

2. I was thinking about making this type more concise, but I think in it's current form it's much more transparent

Now we clearly see that you can **sharevaluelize** an array, array elements, an object, objects properties and some primitives.

src/reanimated2/helperTypes.ts
tjzel: It is confusing, I will change it once time comes for `AnimatedProps`.

src/reanimated2/commonTypes.ts
```diff
@@ -1,47 +1,16 @@
...
   originX?: number;
   originY?: number;
   [key: string]: any;
 }
 
-export interface SharedValue<T> {
-  value: T;
-  addListener: (listenerID: number, listener: (value: T) => void) => void;
+export interface SharedValue<TValue> {
+  value: TValue;
```
Latropos, line 10: I think that `ValueT` instead of `TValue` would match our current style
```suggestion
export interface SharedValue<ValueT> {
  value: ValueT;
```

src/reanimated2/helperTypes.ts
```diff
@@ -22,102 +24,69 @@ import type { ReanimatedKeyframe } from './layoutReanimation/animationBuilder/Ke
...
-};
-
-type RNTransformType = RNTransformsStyle['transform'];
+export type AnimatedTransform = MaybeSharedValueRecursive<
+  TransformsStyle['transform']
+>;
 
-export type ExtractArrayItemType<Arr> = Arr extends readonly (infer Item)[]
-  ? Item
+type MaybeSharedValue<TValue> = TValue | TValue extends AnimatableValue
```
Latropos, line 35: Please prefer `ValueT` instead of `TValue`

src/reanimated2/helperTypes.ts
```diff
@@ -22,102 +24,69 @@ import type { ReanimatedKeyframe } from './layoutReanimation/animationBuilder/Ke
...
+/*
   Style type properties (properties that extends StyleProp<ViewStyle>)
   can be defined with other property names than "style". For example `contentContainerStyle` in FlatList.
   Type definition for all style type properties should act similarly, hence we
   pick keys with 'Style' substring with the use of this utility type.
 */
-type PickStyleProps<T> = Pick<
-  T,
+type PickStyleProps<TProps> = Pick<
+  TProps,
```
Latropos, line 72: Please use `PropsT`

src/reanimated2/helperTypes.ts
```diff
@@ -22,102 +24,69 @@ import type { ReanimatedKeyframe } from './layoutReanimation/animationBuilder/Ke
...
+    [Key in keyof TProps]-?: Key extends `${string}Style` | 'style'
+      ? Key
+      : never;
+  }[keyof TProps]
 >;
 
-type StyleAnimatedProps<P extends object> = {
-  [K in keyof PickStyleProps<P>]: StyleProp<
-    AnimatedStyle<P[K] | MaybeSharedValue<P[K]>>
+type StyleProps<TProps extends object> = {
```
Latropos, line 80: IDK what is the better name, however these styles are animated, so maybe:
```suggestion
type AnimatedStyleProps<TProps extends object> = {
```

Latropos (approved):

PR #4925, opened 2023-08-15
fix: docs withSpring playground missing comma

schriker:
withSpring docs config section (https://docs.swmansion.com/react-native-reanimated/docs/animations/withSpring#config-) missing comma in the playground after `dampingRatio` and `damping `

image
image


piaskowyk (approved):
@schriker Thanks for your PR 🙌

PR #4926, opened 2023-08-16
Fix event dispatcher broken on ios with expo-dev-client

Kudo:
## Summary

@hirbod reported an issue with reanimated 3.4 + expo-dev-client on ios that `useEvent` is broken. this happens because of the setup of expo-dev-client and would like to propose and discuss solution in this pr.

### expo-dev-client setup

expo-dev-client will create multiple bridges:
  - app bridge: the target app
  - expo-dev-menu: the customized dev menu is also a react native app.
  - expo-dev-launcher: the default start ui that for user to choose a target app is also a react native app.

### the error startup flow

when clicking the app in ios springboard, i.e. not starting the app through deep linking, the default startup ui will launch expo-dev-launcher first. the initialize flow will be
  1. `RCTEventDispatcher+Reanimated +load`
  2. create a bridge for expo-dev-launcher
  3. even expo-dev-launcher doesn't use reanimated now, as long as other code in expo-dev-launcher sends an event. it will call the swizzled `RCTEventDispatcher.sendEvent:` and cache the `REAModule` to the one belong to expo-dev-launcher: https://github.com/software-mansion/react-native-reanimated/blob/f55e831b1e45566f3f28707344e983ade62021e6/ios/RCTEventDispatcher%2BReanimated.m#L11-L19
  4. on expo-dev-launcher ui, user selects an app to start. the app bridge is then created (and not deallocating the expo-dev-launcher's bridge). now we have two bridges and two `REAModule`s.
  5. when the app receives events through `useEvent`, it will send events to expo-dev-launcher's `REAModule`, which is wrong target.
 
### proposing solution

in the swizzled `RCTEventDispatcher.sendEvent:`, if the bridge instance is changed, we should update the cached `REAModule`.

## Test plan

based on the repro where @hirbod and @keith-kurak provided: https://github.com/keith-kurak/test-reanimated-dev-client-issue

To reproduce:

> 1. Build a development build on iOS simulator (eas build --profile development --platform simulator)
>
NOTE(kudo): could also use `npx expo run:ios`, but after launching the app, please stop the app and launch from ios springboard instead.

> 2. Run the repro project locally ( npx expo start ). This repro is a blank SDK 49 project + an example from the reanimated docs
> 
> 3. Try to move the red shade/ drawer. The ball should also move, but doesn't (the ball moves on reanimated 3.3.x and as a standalone build.)

### DIFF
--- a/ios/RCTEventDispatcher+Reanimated.m
+++ b/ios/RCTEventDispatcher+Reanimated.m
@@ -11,9 +11,14 @@
 - (void)reanimated_sendEvent:(id<RCTEvent>)event
 {
   // Pass the event to Reanimated
+  static __weak RCTBridge *bridge;
   static __weak REAModule *reaModule;
+  if (bridge != self.bridge) {
+    bridge = self.bridge;
+    reaModule = nil;
+  }
   if (reaModule == nil) {
-    reaModule = [[self bridge] moduleForName:@"ReanimatedModule"];
+    reaModule = [bridge moduleForName:@"ReanimatedModule"];
   }
   [reaModule eventDispatcherWillDispatchEvent:event];
 


tomekzaw:
@Kudo Thanks a lot for your investigation and the fix! Looks like I was the one who introduced this regression in https://github.com/software-mansion/react-native-reanimated/pull/4525 😢 If necessary, we could revert that change, or simply merge this PR, which one do you think is better?

tomekzaw (approved):

Kudo:
either way works for me. though the `moduleForName:` for reanimated should [be cached already](https://github.com/facebook/react-native/blob/e7f6f079cb4fd922ee579ff1eca5fb7134f076e7/packages/react-native/React/CxxBridge/RCTCxxBridge.mm#L575-L577), it is still a map search. for event frequency, it might come with some performance cost.

just depending on whether you want to optimize minor performance by having two weak pointers to store the instances. (well to be honest, i like the optimization more 😅)

tomekzaw:
> it is still a map search

That was exactly the motivation behind the original PR 😄 

> for event frequency, it might come with some performance cost

That's true, but realistically there's at most a few events on each frame so it's not a bottleneck.

> having two weak pointers to store the instances

I'm more worried about the fact that these pointers are static and thus common for all instances of `RCTEventDispatcher`. I can imagine that in some exotic cases of brownfield apps there potentially could be two instances of React and perhaps even two instances of REAModule.

> (well to be honest, i like the optimization more 😅)

Me too! So let's merge this PR but I promise, if there are more problems with this optimization in the future, we'll immediately revert to the unoptimized version 😅 

Thanks again @Kudo and @hirbod who reported the problem 🫡

hirbod:
Thanks guys for addressing this.

hirbod:
Also thanks to @kirillzyusko who help isolating the issue

Kudo:
> I'm more worried about the fact that these pointers are static and thus common for all instances of RCTEventDispatcher. I can imagine that in some exotic cases of brownfield apps there potentially could be two instances of React and perhaps even two instances of REAModule.

that's true and exactly our case. though we could also use `objc_getAssociatedObject` than static variables but the performance may be slower than just `moduleForName:` lookup.

> Me too! So let's merge this PR but I promise, if there are more problems with this optimization in the future, we'll immediately revert to the unoptimized version 😅

that's fair. thanks for reviewing this pr ❤️

hirbod:
@tomekzaw is a release planned for this week? 🥹

tomekzaw:
@hirbod We still need to merge a few critical PRs for 3.5.0, best we can do is Friday or early next week

PR #4929, opened 2023-08-16
Fix `getViewProp` function

tomekzaw:
## Summary

This PR fixes a crash when using `getViewProp` function caused by using RN runtime on the UI thread instead of the UI runtime. Additionally, it moves `propObtainer` function from `NativeReanimatedModule` constructor to `PlatformDepMethodsHolder` struct. Finally, it unifies naming by renaming `propObtainer` to `obtainProp`.

## Test plan

GetViewPropExample.tsx


piaskowyk (approved):
😍

PR #4938, opened 2023-08-18
Restore eslint rule to prefer `import type`

tjzel:
## Summary

Thanks to @bartlomiejbloniarz who noticed it, the rule was (iirc) accidentally removed in #4672.

## Test plan

🧨


piaskowyk (approved):
🎉

src/reanimated2/animation/util.ts
```diff
@@ -17,9 +17,11 @@ import {
   ReduceMotion,
 } from '../commonTypes';
 import NativeReanimatedModule from '../NativeReanimated';
-import {
+import type {
   AffineMatrixFlat,
   AffineMatrix,
+} from './transformationMatrix/matrixUtils';
+import {
```
bartlomiejbloniarz, line 24: This part is now a bit inconsistent, since the types from `commonTypes` have `type` inlined, but `matrixUtils` is split into two imports. I think this can be fixed using [import/consistent-type-specifier-style](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/consistent-type-specifier-style.md).

PR #4947, opened 2023-08-21
[SET] Fix transition restart

piaskowyk:
## Summary

This PR fixes the issue with shared element transition restart. Previously, there was a chance of the app crashing or a needed component disappearing. The purpose of this PR is to prevent app crashes and ensure a proper end state for the transition.

## Test plan

Run example `TransitionRestart.tsx` example - available in this PR


src/reanimated2/layoutReanimation/animationsManager.ts
```diff
@@ -95,7 +95,7 @@ function createLayoutAnimationManager() {
       if (!value) {
         return;
       }
-      stopObservingProgress(tag, value, true);
+      stopObservingProgress(tag, value);
```
tjzel, line 98: Previously there was `true` here passed, now we have default `false` - is that intended?

tjzel (commented):
I don't understand all the changes but it seems ok. I'd like to improve code quality for some of the places that I mentioned.

kacperkapusciak (approved):
I didn't understand anything from this PR so I've QA it instead

On `main` the app crashes when a Shared Transition is interrupted by the user or programmatically

With this PR applied the transition snaps to the final stage without crashing 

lgtm :shipit:

src/reanimated2/layoutReanimation/animationsManager.ts
piaskowyk: Yes, this is intended. This parameter is related to Layout Aniamations. We don't want to clear Layout Animation configuration here.

src/reanimated2/layoutReanimation/sharedTransitions/ProgressTransitionManager.ts
```diff
@@ -158,14 +171,29 @@ function createProgressTransitionRegister() {
       }
     },
     onTransitionEnd: (removeViews = false) => {
+      if (currentTransitions.size === 0) {
+        toRemove.clear();
```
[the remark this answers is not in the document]
piaskowyk: yes, this is intentionally

tjzel (approved):
🪨

PR #4949, opened 2023-08-21
Better reanimated shadow tree commit detection

michalmaka:
## Summary

Proposed solution is more bullet proof solution for detecting whether currently committed shadow tree comes from reanimated.

Here are the main issues with current implementation:
- if there's another commit hook that is executed before `reanimated` one, then `propsRegistry_->isLastReanimatedRoot(newRootShadowNode)` may not work if the commit hook creates a copy of committed tree
- when `stateReconciliation` is enabled via commit options then new tree root is created and again above detection does not work.

It uses concept of `thread_local` c++ variable which has instance for each thread and RAII to ensure that the appropriate flag is set to `true` on entering commit section and reset to `false` after commit from reanimated on UI thread.

## Test plan

- enable state reconciliation for commit from `reanimated` and check whether the code goes into `isReanimatedCommit` `if` body in `ReanimatedCommitHook`

or

- create a custom commit hook and register it before `ReanimatedCommitHook` and perform same check as above

### DIFF
--- a/Common/cpp/Fabric/PropsRegistry.h
+++ b/Common/cpp/Fabric/PropsRegistry.h
@@ -25,16 +25,6 @@ class PropsRegistry {
 
   void remove(const Tag tag);
 
-  void setLastReanimatedRoot(
-      RootShadowNode::Shared const &lastReanimatedRoot) noexcept {
-    lastReanimatedRoot_.store(lastReanimatedRoot.get());
-  }
-
-  bool isLastReanimatedRoot(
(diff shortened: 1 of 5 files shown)


tomekzaw (approved):
Thanks for this PR! Very clever use of `thread_local`. Congratulations on the idea!

I've tested these changes and here's the proof that it indeed works like expected:

Zrzut ekranu 2023-08-21 o 17 58 02

Zrzut ekranu 2023-08-21 o 17 57 42

PR #4951, opened 2023-08-21
Use original function's name inside worklet

tjzel:
## Summary

Currently plugin uses `_f` as private function name inside of a worklet's function factory. However, @m-bert requested to use original function names to use for layout animations on web.

I added funny scripts as a bonus.

## Test plan

Unit tests, CI and Example app run


package.json
```diff
@@ -32,7 +32,9 @@
     "clean:deep": "cd android && rm -rf .cxx .gradle build && cd ../Example/android && rm -rf .gradle build app/build && cd ../.. && yarn clean",
     "reset:deep": "yarn clean:deep && yarn setup",
     "plugin": "cd plugin && yarn && cd ..",
-    "app": "cd app && yarn && cd .."
+    "app": "cd app && yarn && cd ..",
+    "run-example:ios": "cd Example && yarn && cd ios && pod install && open ReanimatedExample.xcworkspace && cd .. && yarn start --reset-cache",
+    "run-example:ios:cold-start": "cd Example && yarn && cd ios && git clean -xdf && pod install && open ReanimatedExample.xcworkspace && cd .. && yarn start --reset-cache"
```
tomekzaw, line 37: 1. Can we add an analogous script for FabricExample?
2. Can we move `git clean -xdf` to a separate command instead of having `run-example:ios:cold-start`?

tjzel: `git clean -xdf` works only in a given directory so it makes little sense to have it separately. Also, it will remove changes in tracked files in the given directory, so I'd rather use it in those specific scenarios. But yes, we can make it for fabric example.

tomekzaw (approved):
:like:

m-bert:
Indeed, I did requested that

m-bert:
Indeed, I did requested that. Thank you!

PR #4955, opened 2023-08-22
Remove `setNextLayoutAnimationGroup` call

tomekzaw:
## Summary

See https://github.com/Expensify/App/issues/20552#issuecomment-1687710922 for details.


tomekzaw:
@ntdiary Thanks for pointing out this problem as well as providing a repro. Looks like the fix which was supposed to resolve this issue doesn't work like expected so we'll need some more time to propose another solution.

piaskowyk (approved):

Issue #4943, opened 2023-08-21
SyntaxError: unknown: Support for the experimental syntax 'jsx' isn't currently enabled


iSplasher:
### Description

Trying to use this package in my project and have followed the installation steps to a T but still run into an error when I try to use jsx syntax in a file where I import the components from this package.
Same thing in a brand new project, see repro.

### Steps to reproduce

1. Install expo as per their docs
2. Install this package as per docs
3. Run `npm start -- --reset-cache`

App won't compile and you get: `TypeError: Cannot assign to read only property 'message' of object 'SyntaxError: unknown: Support for the experimental syntax 'jsx' isn't currently enabled (13:14):`

### Snack or a link to a repository

https://github.com/iSplasher/expo-react-react-native-reanimated-error

### Reanimated version

~3.3.0

### React Native version

0.72.3

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

None

### Build type

Debug mode

### Device

Real device

### Acknowledgements

Yes

tjzel:
Hi @iSplasher. I took a look at your repro and found the reason for this error. You are trying to define a component inside `useAnimatedStyle` and this is not supported - this function is meant only to create animated styles for components - its code is ran on UI thread where creating `Animated.View` or `View` doesn't make much sense. You should move that logic elsewhere.

iSplasher:
Thanks for letting me know. That piece of info sounds incredibly valuable. Would it make sense to add a notice/warning in the docs?

iSplasher:
That sounds reasonable. Perhaps you could just point it out briefly under the UI/JavaScript thread explanation. Thanks for the reply!

PR #4957, opened 2023-08-22
Run plugin inside a try-catch block

tjzel:
## Summary

Currently when Reanimated Babel plugin fails to transpile a file, regardless of the reason, we don't have any indication that it failed - we just get some generic error from e.g. metro bundler transpilation process. With this PR now we will get additional `[Reanimated] Babel plugin exception` substring in the error, allowing us to narrow down the issue.

Inspired by #4943 

## Test plan

Added two simple tests for it.


__tests__/plugin.test.ts
```diff
@@ -1439,6 +1439,31 @@ describe('babel plugin', () => {
...
+
+    it('uses `callGuardDev` for worklets in dev builds', () => {
+      const input = html`<script>
+        const foo = useAnimatedStyle(() => {
+          return <Image />;
+        });
+      </script>`;
+
+      expect(() => runPlugin(input)).toThrow('[Reanimated]');
+      console.log('Ended first test');
```
tomekzaw, line 1451:
```suggestion
```

plugin/src/plugin.ts
```diff
@@ -6,32 +6,51 @@ import type { ExplicitWorklet, ReanimatedPluginPass } from './types';
...
-          globals.add(name);
-        });
-      }
+      maybeUseCallGuardDEV(() => {
+        // allows adding custom globals such as host-functions
+        if (this.opts != null && Array.isArray(this.opts.globals)) {
+          this.opts.globals.forEach((name: string) => {
+            globals.add(name);
+          });
+        }
```
tomekzaw, line 34: I know this is out of scope for this PR but could we move it to a separate function for the purpose of consistency?

plugin/src/plugin.ts
```diff
@@ -6,32 +6,51 @@ import type { ExplicitWorklet, ReanimatedPluginPass } from './types';
 import { processIfWorkletNode } from './processIfWorkletNode';
 import { processInlineStylesWarning } from './processInlineStylesWarning';
 import { processIfCallback } from './processIfCallback';
+import { isRelease } from './utils';
 
 module.exports = function (): PluginItem {
+  function callGuardDEV(fun: () => void) {
```
tomekzaw, line 12: I'm not fully convinced if we should use the name `callGuardDEV` here since it may be easily confused with `callGuardDEV` for running worklets.

plugin/src/plugin.ts
```diff
@@ -6,32 +6,51 @@ import type { ExplicitWorklet, ReanimatedPluginPass } from './types';
...
+    try {
+      fun();
+    } catch (e) {
+      throw new Error('[Reanimated] Babel plugin exception: ' + e);
+    }
+  }
+
+  function releaseCaller(fun: () => void) {
```
tomekzaw, line 20: The name of this function should be analogous to the debug counterpart.

__tests__/plugin.test.ts
```diff
@@ -1439,6 +1439,31 @@ describe('babel plugin', () => {
...
+      const input = html`<script>
+        const foo = useAnimatedStyle(() => {
+          return <View />;
+        });
+      </script>`;
+
+      const current = process.env.BABEL_ENV;
+      process.env.BABEL_ENV = 'production';
+      expect(() => runPlugin(input)).toThrow();
+      expect(() => runPlugin(input)).not.toThrow('[Reanimated]');
```
piaskowyk, line 1464: Just add comment here with explanation

piaskowyk (approved):

plugin/src/plugin.ts
```diff
@@ -6,32 +6,51 @@ import type { ExplicitWorklet, ReanimatedPluginPass } from './types';
...
+    fun();
+  }
+
+  const maybeUseCallGuardDEV = isRelease() ? releaseCaller : callGuardDEV;
```
tomekzaw, line 24: What about `captureError`?

tomekzaw (changes requested):
Let's use try/catch for the release bundle as well

plugin/src/plugin.ts
```diff
@@ -1,37 +1,46 @@
...
   return {
     pre() {
-      // allows adding custom globals such as host-functions
-      if (this.opts != null && Array.isArray(this.opts.globals)) {
-        this.opts.globals.forEach((name: string) => {
-          globals.add(name);
-        });
-      }
+      runWithTaggedExceptions(() => {
+        addCustomGlobals.call(this);
```
tomekzaw, line 22: Will it work?
```suggestion
        addCustomGlobals(this);
```

plugin/src/addCustomGlobals.ts
```diff
@@ -0,0 +1,12 @@
+import { globals } from './commonObjects';
+
+// This function allows adding custom globals such as host-functions.
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const addCustomGlobals = function (this: any) {
```
tomekzaw, line 6: plz no `any`

__tests__/plugin.test.ts
```diff
@@ -1439,6 +1439,16 @@ describe('babel plugin', () => {
       expect(code).not.toContain('version: ');
       expect(code).toMatchSnapshot();
     });
+
+    it('uses `callGuardDev` for worklets in dev builds', () => {
```
tomekzaw:
```suggestion
    it('intercepts exceptions', () => {
```

__tests__/plugin.test.ts
```diff
@@ -1439,6 +1439,16 @@ describe('babel plugin', () => {
       expect(code).not.toContain('version: ');
       expect(code).toMatchSnapshot();
     });
+
+    it('uses `runWithTaggedErrors` for worklets', () => {
```
tomekzaw, line 1443:
```suggestion
    it('uses `runWithTaggedExceptions` for worklets', () => {
```

plugin/src/plugin.ts
```diff
@@ -1,37 +1,48 @@
...
     pre() {
-      // allows adding custom globals such as host-functions
-      if (this.opts != null && Array.isArray(this.opts.globals)) {
-        this.opts.globals.forEach((name: string) => {
-          globals.add(name);
-        });
-      }
+      runWithTaggedExceptions(() => {
+        initializeGlobals();
+        addCustomGlobals.call(this);
```
tomekzaw, line 24: Will it work?
```suggestion
        addCustomGlobals(this);
```

plugin/src/addCustomGlobals.ts
```diff
@@ -0,0 +1,12 @@
+import { globals } from './globals';
+import type { ReanimatedPluginPass } from './types';
+
+// This function allows adding custom globals such as host-functions.
+
+export const addCustomGlobals = function (this: ReanimatedPluginPass) {
```
tomekzaw, line 6: What's the advantage of using an arrow function here instead of a function declaration (which seems to be the primary way of creating functions in JavaScript)?
```suggestion
export function addCustomGlobals(this: ReanimatedPluginPass) {
```

plugin/src/addCustomGlobals.ts
```diff
@@ -0,0 +1,12 @@
+import { globals } from './globals';
+import type { ReanimatedPluginPass } from './types';
+
+// This function allows adding custom globals such as host-functions.
```
tomekzaw, line 4:
```suggestion
// This function allows for adding custom globals such as host-functions.
```

__tests__/plugin.test.ts
```diff
@@ -1439,6 +1457,16 @@ describe('babel plugin', () => {
       expect(code).not.toContain('version: ');
       expect(code).toMatchSnapshot();
     });
+
+    it('uses `runWithTaggedExceptions` for worklets', () => {
```
tomekzaw, line 1461: Not sure if test name should contain implementation details like the function name
```suggestion
    it('throws a tagged exception when processing worklet fails', () => {
```

tomekzaw (changes requested):

plugin/src/addCustomGlobals.ts
tjzel: This is not an arrow function but we could use simple function declaration regardless.

plugin/src/plugin.ts
tjzel: No, declaring `this` in a function as an argument in TS is only to assign a type to it, it's not a real argument. The function has to be called with current `this` bound to it and this (imo) is a clean way to do this. Alternate would be

`addCustomGlobals.bind(this)()`

Which just looks bad.

PR #4958, opened 2023-08-22
Added optimization to shouldSkipCommit mechanism

michalmaka:
## Summary

React Native 0.73 introduces concept of mount hooks. Such hook is useful to notify Reanimated that shadowTree commit coming from React Native has finished so that Reanimated can commit its own changes.

The solution has shown about 15% performance gain in number of successful commits per second.

## Test plan

None

### DIFF
--- a/Common/cpp/Fabric/PropsRegistry.h
+++ b/Common/cpp/Fabric/PropsRegistry.h
@@ -31,8 +31,8 @@ class PropsRegistry {
 
   bool shouldReanimatedSkipCommit() {
 #if REACT_NATIVE_MINOR_VERSION >= 73
-    // In RN 0.73 we have a mount hook will properly set this flag after
-    // non-Reanimated commit
+    // In RN 0.73+ we have a mount hook that will properly set this flag
+    // after a non-Reanimated commit.
     return shouldReanimatedSkipCommit_;
 #else
     return shouldReanimatedSkipCommit_.exchange(false);
--- a/Common/cpp/Fabric/PropsRegistry.h
+++ b/Common/cpp/Fabric/PropsRegistry.h
@@ -33,6 +33,10 @@ class PropsRegistry {
     return letMeIn_.exchange(false);
   }
 
+  void resetSkipCommitFlag() {
+    letMeIn_ = false;
+  }
+
  private:
   std::unordered_map<Tag, std::pair<ShadowNode::Shared, folly::dynamic>> map_;
 
--- a/Common/cpp/Fabric/PropsRegistry.h
+++ b/Common/cpp/Fabric/PropsRegistry.h
@@ -26,15 +26,15 @@ class PropsRegistry {
   void remove(const Tag tag);
 
   void pleaseSkipCommit() {
-    letMeIn_ = true;
+    shouldReanimatedSkipCommit_ = true;
   }
 
   bool shouldSkipCommit() {
-    return letMeIn_.exchange(false);
+    return shouldReanimatedSkipCommit_.exchange(false);
   }
 
   void resetSkipCommitFlag() {
-    letMeIn_ = false;
+    shouldReanimatedSkipCommit_ = false;
   }
 
  private:
@@ -42,7 +42,7 @@ class PropsRegistry {
 
   mutable std::mutex mutex_; // Protects `map_`.
 
-  std::atomic<bool> letMeIn_;
+  std::atomic<bool> shouldReanimatedSkipCommit_;
 };
 
 } // namespace reanimated
--- a/Common/cpp/Fabric/PropsRegistry.h
+++ b/Common/cpp/Fabric/PropsRegistry.h
@@ -31,7 +31,7 @@ class PropsRegistry {
 
   bool shouldReanimatedSkipCommit() {
 #if REACT_NATIVE_MINOR_VERSION >= 73
-    // In RN 0.73+ we have a mount hook that will properly set this flag
+    // In RN 0.73+ we have a mount hook that will properly unset this flag
     // after a non-Reanimated commit.
     return shouldReanimatedSkipCommit_;
 #else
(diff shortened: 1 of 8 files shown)


android/src/main/cpp/NativeProxy.cpp
```diff
@@ -53,6 +48,10 @@ NativeProxy::NativeProxy(
   uiManager_ = binding->getScheduler()->getUIManager();
   commitHook_ =
       std::make_shared<ReanimatedCommitHook>(propsRegistry_, uiManager_);
+#if REACT_NATIVE_MINOR_VERSION >= 73
+  mountHook_ =
+      std::make_shared < ReanimatedMountHook(propsRegistry_, uiManager_);
```
tomekzaw, line 53:
```suggestion
      std::make_shared<ReanimatedMountHook>(propsRegistry_, uiManager_);
```

Common/cpp/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -57,8 +57,8 @@ RootShadowNode::Unshared ReanimatedCommitHook::shadowTreeWillCommit(
     });
   }
 
-  // request Reanimated to skip one commit so that React Native can mount the
-  // changes instead of failing 1024 times and crashing the app
+  // If the commit comes from React Native then skip one commit from Reanimated
+  // since the tree is already outdated and we have necessary updates.
```
tomekzaw, line 61: * "the tree is already outdated" &rarr; the tree to be committed by Reanimated doesn't include the new changes from React Native
* "we have necessary updates" &rarr; all changes of animated props will be applied from props registry

Common/cpp/Fabric/PropsRegistry.h
```diff
@@ -33,6 +33,10 @@ class PropsRegistry {
     return letMeIn_.exchange(false);
   }
 
+  void resetSkipCommitFlag() {
+    letMeIn_ = false;
```
tomekzaw, line 37: I think it's high time to rename this flag, what do you think about `shouldReanimatedSkipCommit_`?

tomekzaw (changes requested):

android/src/main/cpp/NativeProxy.cpp
michalmaka: oh, nice catch

Common/cpp/Fabric/PropsRegistry.h
```diff
@@ -26,19 +26,23 @@ class PropsRegistry {
   void remove(const Tag tag);
 
   void pleaseSkipCommit() {
-    letMeIn_ = true;
+    shouldReanimatedSkipCommit_ = true;
   }
 
   bool shouldSkipCommit() {
-    return letMeIn_.exchange(false);
+    return shouldReanimatedSkipCommit_.exchange(false);
```
tomekzaw, line 33: Since we already have `resetSkipCommitFlag` (which is called only in RN 0.73+), does it still make sense to exchange it with `false` on RN 0.73+ or should we just return the current value and not modify it?

Common/cpp/Fabric/PropsRegistry.h
```diff
@@ -26,19 +26,23 @@ class PropsRegistry {
   void remove(const Tag tag);
 
   void pleaseSkipCommit() {
```
tomekzaw, line 28: Should we specify who exactly should skip commit?
```suggestion
  void pleaseSkipReanimatedCommit() {
```

Common/cpp/Fabric/PropsRegistry.h
```diff
@@ -26,19 +26,23 @@ class PropsRegistry {
...
-    letMeIn_ = true;
+    shouldReanimatedSkipCommit_ = true;
   }
 
   bool shouldSkipCommit() {
```
tomekzaw, line 32:
```suggestion
  bool shouldReanimatedSkipCommit() {
```

Common/cpp/Fabric/ReanimatedMountHook.h
```diff
@@ -0,0 +1,34 @@
...
+
+ private:
+  std::shared_ptr<PropsRegistry> propsRegistry_;
+
+  std::shared_ptr<UIManager> uiManager_;
+};
+
+} // namespace reanimated
+
+#endif // defined(RCT_NEW_ARCH_ENABLED) &&REACT_NATIVE_MINOR_VERSION >= 73
```
tomekzaw, line 34:
```suggestion
#endif // defined(RCT_NEW_ARCH_ENABLED) && REACT_NATIVE_MINOR_VERSION >= 73
```

Common/cpp/Fabric/PropsRegistry.h
```diff
@@ -25,20 +25,30 @@ class PropsRegistry {
...
+  void pleaseSkipReanimatedCommit() {
+    shouldReanimatedSkipCommit_ = true;
   }
 
-  bool shouldSkipCommit() {
-    return letMeIn_.exchange(false);
+  bool shouldReanimatedSkipCommit() {
+#if REACT_NATIVE_MINOR_VERSION >= 73
+    // In RN 0.73 we have a mount hook will properly set this flag after
+    // non-Reanimated commit
```
tomekzaw, line 35:
```suggestion
    // In RN 0.73+ we have a mount hook that will properly set this flag
    // after a non-Reanimated commit.
```

Common/cpp/Fabric/ReanimatedMountHook.h
```diff
@@ -0,0 +1,34 @@
...
+  ~ReanimatedMountHook() noexcept override;
+
+  void shadowTreeDidMount(
+      RootShadowNode::Shared const &rootShadowNode,
+      double mountTime) noexcept override;
+
+ private:
+  std::shared_ptr<PropsRegistry> propsRegistry_;
+
+  std::shared_ptr<UIManager> uiManager_;
```
tomekzaw, line 29:
```suggestion
  const std::shared_ptr<PropsRegistry> propsRegistry_;
  const std::shared_ptr<UIManager> uiManager_;
```

tomekzaw (approved):

PR #4933, opened 2023-08-17
Fix `useAnimatedSensor` vector velocity on Android

kowczarz:
## Summary

Velocity of gyroscope and magnetic field sensors were inverted when compared to iOS.

## Test plan

Added repro in examples.

PR #4960, opened 2023-08-22
Add clamp worklet

kowczarz:
## Summary

It's a followup PR for a discussion from https://github.com/software-mansion/react-native-reanimated/pull/4933#discussion_r1297531333

## Test plan

Examples of gyroscope and accelerometer should work the same as previous.


kowczarz:
Cc. @tomekzaw @tjzel

tomekzaw (approved):
Thanks for this PR! ❤️

src/reanimated2/utils.ts
```diff
@@ -32,3 +32,14 @@ export function isSharedValue<T>(value: any): value is SharedValue<T> {
   'worklet';
   return value?._isReanimatedSharedValue === true;
 }
+
+/**
+ * @param val preferred value
```
tjzel, line 37:
```suggestion
 * @param value clamped value
```

src/reanimated2/utils.ts
tjzel, line 1: Let's move it to `interpolate.ts` since it's an edge-case of interpolation.

tjzel (changes requested):

src/reanimated2/utils.ts
```diff
@@ -32,3 +32,14 @@ export function isSharedValue<T>(value: any): value is SharedValue<T> {
...
+ * @param min minimum allowed value
+ * @param max maximum allowed value
+ * Clamps given value within a range of values between a defined minimum bound and a maximum bound.
+ */
+export function clamp(val: number, min: number, max: number) {
```
tjzel, line 42: No need for extra abbreviation - although I do love the idea of 3-letters variables.

```suggestion
export function clamp(value: number, min: number, max: number) {
```

src/reanimated2/utils.ts
tomekzaw: `clamped value` would be the result value, no? Did you mean `value to be clamped`?

tjzel: It's the value you are clamping - clamped value, same as shareable value or animated value

tjzel: But I guess value to he clamped sounds more precise :like:

piaskowyk (approved):
I posted just two small comments, but overall it is 👌

tjzel (approved):

Issue #2975, opened 2022-02-09
transform styles not affecting Animated.View when using LayoutAnimation on the same element


masarbazi:
## Description
`transform` style is not taking effect when using `style` prop along with `entering` animation on the same `Animated.View`.
 * However this problem could be resolved by wrapping `Animated.View` by another parent `Animated.View` and defining entering prop on that, this report is just FYI.

## what's happening vs. expected behavior
Entering animation works but not taking effect by transform style prop.
![github-reanimted-issue-report](https://user-images.githubusercontent.com/29546473/153130925-2bad9716-28bd-451c-9063-24fd6246d93b.png)

## Minimal code example
```javascript
   ...
    <Animated.View
      style={{
        transform: [
          {perspective: 1000},
          {rotateZ: '10deg'},
          {scale: 0.85},
          {rotateX: '10deg'},
        ]
      }}
      entering={RollInLeft} // none of above styles are working because of this guy here.
     >
         ... 
```

## Snack
Expo sdk 43 installs reanimated 2.2.0 which does not support layout animations, so couldn't produce issue there but here is the code: 
[Snack (Just code)](https://snack.expo.dev/@masarbazi/reanimated_2975)

## Package versions
react-native 65.1
react-native-reanimated 2.4.1
nodejs 16.13.2

## Affected platforms
Just using Android here. don't know if other platforms are affected.

Latropos:
Both Android and IOS are affected

Latropos:
The suggested solution is to add a wrapper with entry/exit animation, we have also added an extra warning to inform users about this behaviour

PR #4969, opened 2023-08-23
Add warning - component style is overwritten with animation style.

Latropos:
## Summary

In this PR I want to add a warning to prevent cases described here: https://github.com/software-mansion/react-native-reanimated/issues/2975

 
The problem is that some styles are sometimes not affecting Animated.View that has layout animation. It happens when the style property we pass to Animated.View is also animated inside given layout animation. Then it gets overwritten.

Added warning:
> **Warning**
> [Reanimated] Property "transform" of AnimatedComponent(View) may be overwritten with layout animation. Please create a wrapper with the layout animation you want to apply.

<table>
<tr><th>BEFORE</th><th>AFTER</th></tr>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/8a02aa3b-fc97-4401-9993-d15f184a0c3f

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/84b627a3-43dd-4584-8f69-bb1b724dddf0

</td></tr>

</table>

## Test plan

Try this code:
<details>

```js
import Animated, { FadeIn, RollInRight } from 'react-native-reanimated';

  return (
      <Animated.View
        entering={RollInRight.duration(2000)}
        style={{
          width: 200,
          height: 200,
          backgroundColor: 'steelblue',
          transform: [
            { perspective: 1000 },
            { rotateZ: '10deg' },
            { scale: 0.9 },
            { rotateX: '10deg' },
          ],
        }}
      />
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    paddingTop: 100,
    backgroundColor: '#fff',
    padding: 8,
  },
});
```

</details>

### DIFF
--- a/src/animationBuilder.tsx
+++ b/src/animationBuilder.tsx
@@ -1,0 +1,59 @@
+import type {
+  ILayoutAnimationBuilder,
+  LayoutAnimationFunction,
+} from './reanimated2/layoutReanimation';
+import type { StyleProps } from './reanimated2/commonTypes';
+
+const mockTargetValues = {
+  targetOriginX: 0,
+  targetOriginY: 0,
+  targetWidth: 0,
+  targetHeight: 0,
+  targetGlobalOriginX: 0,
+  targetGlobalOriginY: 0,
+  windowWidth: 0,
+  windowHeight: 0,
+  currentOriginX: 0,
+  currentOriginY: 0,
+  currentWidth: 0,
+  currentHeight: 0,
+  currentGlobalOriginX: 0,
+  currentGlobalOriginY: 0,
+};
+
+export function maybeBuild(
+  layoutAnimationOrBuilder:
+    | ILayoutAnimationBuilder
+    | LayoutAnimationFunction
+    | Keyframe,
+  style: StyleProps | undefined
+): LayoutAnimationFunction | Keyframe {
+  const isAnimationBuilder = (
+    value: ILayoutAnimationBuilder | LayoutAnimationFunction | Keyframe
+  ): value is ILayoutAnimationBuilder =>
+    'build' in layoutAnimationOrBuilder &&
+    typeof layoutAnimationOrBuilder.build === 'function';
+
+  if (isAnimationBuilder(layoutAnimationOrBuilder)) {
+    const animatedStyle =
+      layoutAnimationOrBuilder.build()(mockTargetValues).animations;
+
+    const getCommonProperties = (obj1: object, obj2: object) =>
+      Object.keys(obj1).filter((key) =>
+        Object.prototype.hasOwnProperty.call(obj2, key)
+      );
+
+    const commonProperties = getCommonProperties(animatedStyle, style || {});
+    if (commonProperties) {
+      console.warn(
+        `[Reanimated] ${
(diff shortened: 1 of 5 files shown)


src/animationBuilder.tsx
```diff
@@ -0,0 +1,59 @@
...
+    'build' in layoutAnimationOrBuilder &&
+    typeof layoutAnimationOrBuilder.build === 'function';
+
+  if (isAnimationBuilder(layoutAnimationOrBuilder)) {
+    const animatedStyle =
+      layoutAnimationOrBuilder.build()(mockTargetValues).animations;
+
+    const getCommonProperties = (obj1: object, obj2: object) =>
+      Object.keys(obj1).filter((key) =>
+        Object.prototype.hasOwnProperty.call(obj2, key)
```
piaskowyk, line 44:
```suggestion
        obj2.hasOwnProperty(key)
```

src/animationBuilder.tsx
```diff
@@ -0,0 +1,59 @@
...
+): LayoutAnimationFunction | Keyframe {
+  const isAnimationBuilder = (
+    value: ILayoutAnimationBuilder | LayoutAnimationFunction | Keyframe
+  ): value is ILayoutAnimationBuilder =>
+    'build' in layoutAnimationOrBuilder &&
+    typeof layoutAnimationOrBuilder.build === 'function';
+
+  if (isAnimationBuilder(layoutAnimationOrBuilder)) {
+    const animatedStyle =
+      layoutAnimationOrBuilder.build()(mockTargetValues).animations;
```
piaskowyk, line 40: Lets spit it to more than one line for readability purposes
```js
const animationFactory = layoutAnimationOrBuilder.build()
const layoutAnimation = animationFactory(mockTargetValues)
const animatedStyle = layoutAnimation.animations
```

src/animationBuilder.tsx
```diff
@@ -0,0 +1,60 @@
+import type {
+  ILayoutAnimationBuilder,
+  LayoutAnimationFunction,
+} from './reanimated2/layoutReanimation';
+import type { StyleProps } from './reanimated2/commonTypes';
+
+const mockTargetValues = {
```
piaskowyk, line 7: Let's add to this proper type `LayoutAnimationsValues` to keep this mock up to date with type.

piaskowyk (changes requested):

src/animationBuilder.tsx
Latropos: This is banned in eslint recommended: https://eslint.org/docs/latest/rules/no-prototype-builtins
Do we want to turn this rule off?

piaskowyk: Ok, it makes sense. But did you check if it works when one of objects is null or undefined?

Latropos: It doesn't work, therefore I always pass an object to `getCommonProperties`:
```js
    const commonProperties = getCommonProperties(animatedStyle, style || {});
```

piaskowyk (approved):
👏

PR #4975, opened 2023-08-24
Add navigation stack to macOS example app

jfedak:

### DIFF
--- a/MacOSExample/App.tsx
+++ b/MacOSExample/App.tsx
@@ -1,3 +1,3 @@
-import BokehExample from '../app/src/examples/BokehExample';
+import App from '../app/';
 
-export default BokehExample;
+export default App;
--- a/MacOSExample/ios/Podfile.lock
+++ b/MacOSExample/ios/Podfile.lock
@@ -329,6 +329,12 @@ PODS:
   - React-jsinspector (0.71.12)
   - React-logger (0.71.12):
     - glog
+  - react-native-pager-view (5.4.24):
+    - React-Core
+  - react-native-safe-area-context (4.7.1):
+    - React-Core
+  - react-native-slider (4.4.2):
+    - React-Core
   - React-perflogger (0.71.12)
   - React-RCTActionSheet (0.71.12):
     - React-Core/RCTActionSheetHeaders (= 0.71.12)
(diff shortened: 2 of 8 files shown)


app/src/App.tsx
```diff
@@ -39,6 +44,10 @@ type RootStackParamList = { [P in keyof typeof EXAMPLES]: undefined } & {
 };
 
 interface HomeScreenProps {
+  navigation: StackNavigationProp<RootStackParamList, 'Home'>;
+}
+
+interface HomeScreenNativeProps {
   navigation: NativeStackNavigationProp<RootStackParamList, 'Home'>;
 }
```
tomekzaw, line 52: Instead of having `function HomeScreen({ navigation }: HomeScreenProps | HomeScreenNativeProps)`, why don't we just use the union type here?

```suggestion
interface HomeScreenProps {
  navigation: StackNavigationProp<RootStackParamList, 'Home'> | NativeStackNavigationProp<RootStackParamList, 'Home'>;
}
```

app/src/App.tsx
```diff
@@ -147,7 +168,11 @@ export default function App() {
       try {
         const initialUrl = await Linking.getInitialURL();
 
-        if (Platform.OS !== 'web' && initialUrl == null) {
+        if (
+          Platform.OS !== 'web' &&
+          Platform.OS !== 'macos' &&
+          initialUrl == null
+        ) {
```
tomekzaw, line 175: I don't mind using `Platform.OS` so in example app code, but what do you think about the following?

```ts
const IS_WEB = Platform.OS === 'web';
const IS_MACOS = Platform.OS === 'macos';
```

```suggestion
        if (!IS_WEB && !IS_MACOS && initialUrl == null) {
```

app/src/App.tsx
```diff
@@ -100,6 +109,15 @@ interface ItemProps {
...
+    return (
+      <Pressable style={styles.button} onPress={onPress}>
+        {icon && <Text style={styles.title}>{icon + '  '}</Text>}
+        <Text style={styles.title}>{title}</Text>
+      </Pressable>
+    );
+  }
+
   return (
     <RectButton style={styles.button} onPress={onPress}>
```
tomekzaw, line 122: I don't remember what was the reason to use `RectButton` here instead of `Pressable`, can you please check it? Maybe we could use `Pressable` on all platforms.

tomekzaw (approved):

PR #4977, opened 2023-08-24
Fix `useAnimatedProps` on web

### LINKED ISSUE #4909: useAnimatedProps doesnt update props on web
### Description

The issue is occurs only on web. It works ok on android and iOS. I've first notice the issue by trying to set showsVerticalScrollIndicator that was not disapearing on web. Then i tried to work around the issue by hacking it using the role attribute (accessibilityRole on react native) and then style it using css on the web. The role didn't change either. I works the same on Reanimated 2 as on reanimated 3. 

_Update extra info: pointerEvents doesn't work either because at first it maps to a style with !important rule but when changed it changes the style of the element without the important and it doesn't take effect. On web on the other hand you can set pointerEvents by setting it as animated style and **not setting it at all as prop or animatedprop**._

### Steps to reproduce

1. Create an Animted.View 
2. add the props we are going to use as whitelisted. addWhitelistedNativeProps({accessibilityRole: true});
3. Declare a shared value const SharedRole=useSharedValue("initial");
4. Declare animatedProps that contain { accessibilityRole: SharedRole.value }
5. change SharedRole using a button or any other event. 
6. inspect using dev tools and check the element attributes 

https://snack.expo.dev/@albi91/reanimated-useanimatedprops-web

### Reanimated version

3.3.0

### React Native version

0.72.3

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Summary

This PR changes behavior of `useAnimatedProps` so that props get updated properly on web.

Fixes #4909

## Test plan

Tested on [example from issue](https://github.com/apatozi/issue-reanimated-useanimatedprops-web).

### DIFF
--- a/src/reanimated2/UpdateProps.ts
+++ b/src/reanimated2/UpdateProps.ts
@@ -11,15 +11,16 @@ import { runOnUIImmediately } from './threads';
 let updateProps: (
   viewDescriptor: SharedValue<Descriptor[]>,
   updates: StyleProps | AnimatedStyle<any>,
-  maybeViewRef: ViewRefSet<any> | undefined
+  maybeViewRef: ViewRefSet<any> | undefined,
+  isAnimatedProps?: boolean
 ) => void;
 
 if (shouldBeUseWeb()) {
-  updateProps = (_, updates, maybeViewRef) => {
+  updateProps = (_, updates, maybeViewRef, isAnimatedProps = false) => {
     'worklet';
     if (maybeViewRef) {
       maybeViewRef.items.forEach((item, _) => {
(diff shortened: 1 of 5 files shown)


tomekzaw:
Should we rename this PR to "Fix `useAnimatedStyle` on web"? 😅

m-bert:
Sure! Seems like I had something else on my mind and overlooked it 😅

piaskowyk (approved):
👏

PR #4986, opened 2023-08-25
Replace old shadow nodes in PropsRegistry with new ones

michalmaka:
## Summary

I found out that many shadow nodes are stored in `PropsRegistry` even if newer versions were passed to `update` method. This PR replaces old `ShadowNode` with new one if needed.

## Test plan

None

### DIFF
--- a/Common/cpp/Fabric/PropsRegistry.cpp
+++ b/Common/cpp/Fabric/PropsRegistry.cpp
@@ -18,7 +18,6 @@ void PropsRegistry::update(
     // returns `ShadowNodeFamily const &` which is non-owning
     map_[tag] = std::make_pair(shadowNode, props);
   } else {
-    // no need to update `.first` because ShadowNodeFamily doesn't change
     // merge new props with old props
     it->second.second.update(props);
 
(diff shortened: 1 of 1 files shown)


Common/cpp/Fabric/PropsRegistry.cpp
```diff
@@ -21,6 +21,12 @@ void PropsRegistry::update(
     // no need to update `.first` because ShadowNodeFamily doesn't change
     // merge new props with old props
     it->second.second.update(props);
+
+    // Update ShadowNode stored in the map in case it was replaced
```
tomekzaw, line 25: I'm okay with this change but maybe it would be faster to always assign a new pair? This way we can remove most of the code here 😄 
```cpp
void PropsRegistry::update(
    const ShadowNode::Shared &shadowNode,
    folly::dynamic &&props) {
  map_[tag] = std::make_pair(shadowNode, props);
}
```

tomekzaw: Turns out we can't do this because we still need to call `.update()` on `folly::dynamic`.

tomekzaw (approved):

Common/cpp/Fabric/PropsRegistry.cpp
```diff
@@ -21,6 +21,12 @@ void PropsRegistry::update(
     // no need to update `.first` because ShadowNodeFamily doesn't change
```
tomekzaw, line 21:
```suggestion
```

PR #4997, opened 2023-08-28
Fix cleaning layout animation config on iOS

### LINKED ISSUE #4516: Layout Animation breaks after navigate
### Description

We are using `@react-navigation/stack`, and we are animating our UI search bar with the `<Animation.View entering={FadeIn} exiting={FadeOut} .../>` and everything insides `<Animation.View layout={Layout} .../>`

The UI animation works right in Android and in iOS. But, only in iOS, after do a `navigate`, and then going back, seems like something is break and the Layout animation doesn't work.

We tried with `detachPreviousScreen` to be sure that the screen is not unmounted but the bug still appears.

https://github.com/software-mansion/react-native-reanimated/assets/22961489/eb873bc4-6e5c-46e1-9315-2d0c7b00e8cd

### Steps to reproduce

1. Create a Layout Animation in Screen 1
2. Navigate to Screen 2
3. Go back to Screen 1.

https://github.com/nicolascavallin/reanimated-bug

Thanks to @dayze for the repro

### Reanimated version

3.1.0

### React Native version

0.71.8

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

### LINKED ISSUE #4816: Animated.View layout animations break inside react-navigation
### Description

Layout animations stop working in some situations using `@react-navigation/bottom-tabs`. In the linked repository, the layout animations on tab one works until you navigate to tab two. The same layout animation works on tab two regardless of navigation state.

https://github.com/software-mansion/react-native-reanimated/assets/1399969/ca37c41d-084f-4037-8f5a-743e3d42806d

### Steps to reproduce

1. Run the app from the linked repository on an iOS simulator
2. Press the toggle button on the initial screen (tab one) - observe expected layout animation
3. Switch to tab two, then back to tab one
4. Press the toggle button again - observe that the layout animation doesn't run

https://github.com/joeporpeglia/repro-tab-navigator-layout-animation

### Reanimated version

3.3.0

### React Native version

0.72.3

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 14

### Acknowledgements

Yes

Latropos:
## Summary
Closes #4816
Closes #4516

Fixing issue with react navigation. If given view is inside some view controller and it has layout animation don't remove its animations from subtree
<table>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/e1581853-f2ba-4b55-a07a-977ca4a84808

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/c366c6ff-1dc7-4b70-b9ed-33582f49da4d

</td></tr>
</table>

## Test plan

Code:
<details>

```js
import {NavigationContainer, useNavigation} from '@react-navigation/native';
import {createNativeStackNavigator} from '@react-navigation/native-stack';
import React from 'react';
import {Pressable, Text, View} from 'react-native';
import Animated, {Layout} from 'react-native-reanimated';

const Screen1 = () => {
  const [isOpen, setIsOpen] = React.useState(false);
  const navigation = useNavigation();
  return (
    <>
      <Animated.View
        style={{backgroundColor: 'red'}}
        layout={Layout.duration(1000)}>
        <Pressable onPress={() => setIsOpen(!isOpen)}>
          <Text>Open</Text>
        </Pressable>
        {isOpen && <View style={{width: 300, height: 400}} />}
      </Animated.View>
      <Text onPress={() => navigation.navigate('Screen2')}>nav</Text>
    </>
  );
};

const Screen2 = () => {
  const [isOpen, setIsOpen] = React.useState(false);
  const navigation = useNavigation();
  return (
    <>
      <Animated.View
        style={{backgroundColor: 'red'}}
        layout={Layout.duration(1000)}>
        <Pressable onPress={() => setIsOpen(!isOpen)}>
          <Text>Open</Text>
        </Pressable>
        {isOpen && <View style={{width: 300, height: 400}} />}
      </Animated.View>
    </>
  );
};

const App = (props: Props) => {
  return (
    <NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen name="Screen1" component={Screen1} />
        <Stack.Screen name="Screen2" component={Screen2} />
      </Stack.Navigator>
    </NavigationContainer>
  );
};

export default App;
```
</details>


tjzel (commented):
TO BE REVIEWED

apple/LayoutReanimation/REASwizzledUIManager.mm
```diff
@@ -90,7 +90,9 @@ - (void)reanimated_manageChildren:(NSNumber *)containerTag
       for (REAUIView *view in permanentlyRemovedChildren) {
         [originalSelf.animationsManager endAnimationsRecursive:view];
       }
-      [originalSelf.animationsManager removeAnimationsFromSubtree:(REAUIView *)container];
+      if (![originalSelf.animationsManager hasLayoutAnimation:(REAUIView *)container]) {
+        [originalSelf.animationsManager removeAnimationsFromSubtree:(REAUIView *)container];
+      }
```
piaskowyk, line 95: This solution can cause a memory leaks unfortunately. I think we should call removeAnimationsFromSubtree for every view from permanentlyRemovedChildren array.

piaskowyk (changes requested):

apple/LayoutReanimation/REAAnimationsManager.m
```diff
@@ -336,6 +336,14 @@ - (BOOL)wantsHandleRemovalOfView:(REAUIView *)view
   });
 }
 
+- (BOOL)hasLayoutAnimation:(REAUIView *)view
```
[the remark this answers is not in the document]
Latropos: * I've found that `wantsHandleRemovalOfView` so I've removed it
* I've made the function more universal, however we already have function `hasLayoutAnimation` with similar arguments doing the same on different level on abstraction. So maybe I should keep the name?

apple/LayoutReanimation/REASwizzledUIManager.mm
piaskowyk: @Latropos ⬆️ bump

piaskowyk (approved):
The original source of this issue was this line:
`removeAnimationsFromSubtree:(REAUIView *)container`
Since container represents an entire screen stack, every layout animation config for that stack is being removed after a screen change. Instead, we should only remove the configuration for the specific screen from the `permanentlyRemovedChildren`.

PR #4998, opened 2023-08-28
Improve entry exit animation types

Latropos:
## Summary

1. Use typescript regex types to generate object with `current` and `target` prefixes based on a single source of truth
   - This may be a good change, since we consider supporting more animation properties.
3. Extract common properties from similar types to reduce boiler-plate code, use inheritance instead of copy-pasting them
4. Get rid of `[key: string]: number | number[];` 🥳 
5. Remove lots of type casts in `SharedTransition.ts` 🥳 

## Test plan
`yarn type:check` & `yarn type:check-api`

### DIFF
--- a/src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
+++ b/src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
@@ -2,6 +2,28 @@ import type { ExtractArrayItemType } from '../../helperTypes';
 import type { EasingFunction } from '../../Easing';
 import type { StyleProps } from '../../commonTypes';
 
+export type LayoutAnimationsOptions =
+  | 'originX'
+  | 'originY'
+  | 'width'
+  | 'height'
+  | 'borderRadius'
+  | 'globalOriginX'
+  | 'globalOriginY';
+
+type CurrentLayoutAnimationsValues = {
+  [K in LayoutAnimationsOptions as `current${Capitalize<string & K>}`]: number;
+};
+
+type TargetLayoutAnimationsValues = {
+  [K in LayoutAnimationsOptions as `target${Capitalize<string & K>}`]: number;
+};
+
+interface WindowDimensions {
+  windowWidth: number;
+  windowHeight: number;
+}
+
 export interface KeyframeProps extends StyleProps {
   easing?: EasingFunction;
   [key: string]: any;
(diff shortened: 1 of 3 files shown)


src/reanimated2/layoutReanimation/sharedTransitions/SharedTransition.ts
```diff
@@ -14,14 +15,14 @@ import { ReduceMotion } from '../../commonTypes';
 import { configureLayoutAnimations } from '../../core';
 import { ProgressTransitionManager } from './ProgressTransitionManager';
 
-const supportedProps = [
+const SUPPORTED_PROPS = [
```
tjzel, line 18: Why do we go for all capital with underscores here? Also, a better approach here would be:
`const myArray = [...] as const`.

src/reanimated2/layoutReanimation/sharedTransitions/SharedTransition.ts
```diff
@@ -201,10 +207,14 @@ export class SharedTransition {
...
-            propertyName.charAt(0).toUpperCase() + propertyName.slice(1);
-          const currentValue = values['current' + PropertyName] as number;
-          const targetValue = values['target' + PropertyName] as number;
+          const PropertyName = (propertyName.charAt(0).toUpperCase() +
+            propertyName.slice(1)) as Capitalize<LayoutAnimationsOptions>;
+          const currentPropertyName = `current${PropertyName}` as const;
+          const targetPropertyName = `target${PropertyName}` as const;
+
+          const currentValue: number = values[currentPropertyName];
+          const targetValue: number = values[targetPropertyName];
```
tjzel, line 216:
```suggestion
          const currentValue = values[currentPropertyName];
          const targetValue = values[targetPropertyName];
```

tjzel (changes requested):

src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
```diff
@@ -14,28 +36,13 @@ export type LayoutAnimation = {
 };
 
 export type AnimationFunction = (a?: any, b?: any, c?: any) => any; // this is just a temporary mock
+export interface EntryAnimationsValues
```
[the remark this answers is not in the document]
Latropos: You are right. However interfaces have better performance inside VSCode than types. On the other hand this is only noticeable if you have complex generic types, so I think it's a good idea to change this

src/reanimated2/layoutReanimation/sharedTransitions/SharedTransition.ts
Latropos: * We usually use UPPERSCORE for global const. I think that using capital letter shows the purpose of this array better (we emphasise it is some kind of hardcoded SoT)
* Yes `as const` looks better, but it would throw stupid errors, since it doesn't work as we want (it keeps the order inside a type). So the inferred type is not
```
Array<'width' | 'height | ...>
```
It is
```
['width', 'height', ...]
```

tjzel: It is an immutable array - I think it's the right thing. The only pitfall is the need to do `SUPPORTED_PROPS as readonly string[]` instead of `SUPPORTED_PROPS as string[]` but since we cast it anyway I think this is irrelevant.

tjzel (approved):

PR #5000, opened 2023-08-28
Add wrapper component for disabling entering animation on the first render

bartlomiejbloniarz:

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -56,6 +56,7 @@ import { removeFromPropsRegistry } from './reanimated2/PropsRegistry';
 import { JSPropUpdater } from './JSPropUpdater';
 import { getReduceMotionFromConfig } from './reanimated2/animation/util';
 import { maybeBuild } from './animationBuilder';
+import { LayoutConfigContext } from './reanimated2/component/LayoutConfig';
 
 const IS_WEB = isWeb();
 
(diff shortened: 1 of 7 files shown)


src/reanimated2/component/SkipInitialEnteringAnimations.tsx
```diff
@@ -0,0 +1,21 @@
+import { createContext, useEffect, useRef } from 'react';
+
+export const SkipEnteringContext =
+  createContext<React.MutableRefObject<boolean> | null>(null);
+
+export function SkipInitialEnteringAnimations(props: {
+  value?: boolean;
+  children?: React.ReactNode;
+}) {
+  const skipValueRef = useRef(props.value !== false);
```
piaskowyk, line 10: it doesn't work when `value` is `null` or `undefined`
image

I think we should revers check to `props.value === true`

src/createAnimatedComponent.tsx
```diff
@@ -601,12 +604,11 @@ export default function createAnimatedComponent(
         >,
       setLocalRef: (ref) => {
         // TODO update config
-
         const tag = IS_WEB
           ? (ref as HTMLElement)
           : findNodeHandle(ref as Component);
-
         const { layout, entering, exiting, sharedTransitionTag } = this.props;
+        const skipEntering = !!this.context?.current;
```
piaskowyk, line 611: small think, but we can move the declaration of `skipEntering` right before `if (entering && !skipEntering) {`

src/reanimated2/component/SkipInitialEnteringAnimations.tsx
bartlomiejbloniarz: I want the ref to be true by default (i.e. when props.value === undefined), that's why I went with that. This way you can just write `<Skip...>` with no props and it will remove the animation.

piaskowyk: Ok, I see. Let's add comment that this is a intentional behaviour

PR #5003, opened 2023-08-29
Reduce file `createAnimatedComponent` - part 2

Latropos:
## Summary
File `createAnimatedComponent.tsx` contains our most important logic. It is already very big and since in the future we want to support ${\textcolor{teal}{\textsf{inline Styles}}}$ this file may grow a lot. Thats why we should make it clear and easy to rid - for example by splitting it across many small files.

This refactor will be made in smaller parts.

### Part 1 ([LINK](https://github.com/software-mansion/react-native-reanimated/pull/5053))
* Create folder 📁 `createAnimatedComponent` 
* Move all files used only in `createAnimatedComponent.tsx` into that folder. Move `createAnimatedComponent.tsx` as well.
* Extract all util function into separate file.
### Part 2 ${\textcolor{green}{\textsf{(This PR)}}}$ 
* Create class `inlinePropUpdater.tsx` which handles logic related with inline styles.
### Parts >2
This is not fixed yet, but I suggest we should:
* Move function `_filterNonAnimatedProps` away (70 lines of code)
  * Maybe create a separate class that would remember `initialStyle`? 
* Move away `setLocalRef` (inside `_setComponentRef`).


src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -8,53 +8,45 @@ import type {
...
-  isWeb,
-} from './reanimated2/PlatformChecker';
-import { initialUpdaterRun } from './reanimated2/animation';
-import type {
-  BaseAnimationBuilder,
-  EntryExitAnimationFunction,
-  ILayoutAnimationBuilder,
-} from './reanimated2/layoutReanimation';
+} from '../reanimated2/core';
+
```
tjzel, line 20:
```suggestion
```

src/createAnimatedComponent/inlinePropUpdater.ts
```diff
@@ -0,0 +1,190 @@
...
+    for (const [key, styleValue] of Object.entries(style)) {
+      if (
+        !isSharedValue(styleValue) &&
+        !(key === 'transform' && isInlineStyleTransform(styleValue))
+      ) {
+        newStyle[key] = styleValue;
+      }
+    }
+    return newStyle;
+  }
```
tjzel, line 189: Why is `hasInlineStyles` a static method and this is non-static?

tjzel (commented):
We need to clear some things out

src/createAnimatedComponent/inlinePropUpdater.ts
Latropos: Good point, I'll fix it

tjzel: I think that actually only `attach` and `detach` should be the members of this class - the rest of the functions present in this file should either all be static methods or outside of this class and I think the former is better.

src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -281,21 +127,21 @@ export default function createAnimatedComponent(
       }
     }
 
+    componentDidMount() {
+      this._attachNativeEvents();
+      this._JSPropUpdater.addOnJSPropsChangeListener(this);
+      this._attachAnimatedStyles();
+      this._inlinePropUpdater.attachInlineProps(this, this._getViewInfo());
```
tjzel, line 134: I'm not a fan of the fact that this PR is supposed to split `createAnimatedComponent` into several files, but actually changes the logic slightly in some places - like here. Could you maybe split it into two separate PRs - one just does the split and the other changes things like this?

tjzel (changes requested):

src/createAnimatedComponent/createAnimatedComponent.tsx
Latropos: Sure, see here for part one: [LINK](https://github.com/software-mansion/react-native-reanimated/pull/5053)

tjzel:
Let's apply suggestions that @piaskowyk made in the review of the [previous PR](https://github.com/software-mansion/react-native-reanimated/pull/5053) here.

src/createAnimatedComponent/createAnimatedComponent.tsx
tjzel, line 1: How about we get rid of `export default` from this file as well in favor of just `export`?

src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -0,0 +1,190 @@
...
+        this._inlinePropsViewDescriptors.add({
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          tag: viewTag as number,
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          name: viewName!,
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          shadowNodeWrapper: shadowNodeWrapper!,
+        });
+      }
+      const sharableViewDescriptors =
```
tjzel, line 159:
```suggestion
      const shareableViewDescriptors =
```

src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -0,0 +1,190 @@
...
+import { isSharedValue } from '../reanimated2/utils';
+import NativeReanimatedModule from '../reanimated2/NativeReanimated';
+
+function isInlineStyleTransform(transform: unknown): boolean {
+  if (!transform) {
+    return false;
+  }
+  if (!Array.isArray(transform)) {
+    return false;
+  }
```
tjzel, line 21: This should suffice
```suggestion
  if (!Array.isArray(transform)) {
    return false;
  }
```

src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  return inlineProps;
+}
+
+export class InlinePropManager {
+  _inlinePropsViewDescriptors: ViewDescriptorsSet | null = null;
+  _inlinePropsMapperId: number | null = null;
+  _inlineProps: StyleProps = {};
+
+  public static hasInlineStyles(style: StyleProps): boolean {
```
tjzel, line 99: I think I'd prefer these static functions to be just exported functions, what do you think about it @piaskowyk?

src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -0,0 +1,190 @@
...
+    } else {
+      update[key] = styleValue;
+    }
+  }
+  return update;
+}
+
+function extractSharedValuesMapFromProps(
+  props: AnimatedComponentProps<
+    Record<string, unknown> /* Initial component props */
```
tjzel, line 62: Can we move this comment outside of generic parameter?

tjzel (changes requested):
After addressing some code quality I think it's good to merge

src/createAnimatedComponent/InlinePropManager.ts
Latropos: It would make it harder to understand that `Record<string, unknown>` are the initial component props, not `AnimatedComponentProps<Record<string, unknown>>`

src/createAnimatedComponent/InlinePropManager.ts
Latropos: It turned out to be quite a common typo in our codebase

src/createAnimatedComponent/InlinePropManager.ts
piaskowyk: `isInlineStyleTransform`, `inlinePropsHasChanged`, `getInlinePropsUpdate` methods have already defined outside the class, so we can move the rest of static function outside the class too.

tjzel (approved):
Let's go!

PR #5005, opened 2023-08-29
Expose `libreanimated` to be consumed via prefab

tomekzaw:
## Summary

This PR exposes `libreanimated.so` along with necessary headers as a prefab on Android so it can be easily consumed by other third-party libraries (e.g. VisionCamera) without a need to manually specify paths to node_modules.

```gradle
android {
  buildFeatures {
    prefab true
  }
}
```

```cmake
find_package(react-native-reanimated REQUIRED CONFIG)

target_link_libraries(${PACKAGE_NAME} react-native-reanimated::reanimated)
```

> [!NOTE]  
> For now, we expose all header files, but this is subject to change as ultimately we want to expose only headers from public C++ API.

> [!NOTE]  
> When using `com.android.tools.build:gradle:7.3.1` the following error may occur. Please upgrade to 7.4.2 or newer.

```
Usage: prefab [OPTIONS] PACKAGE_PATH...

Error: Invalid value for "PACKAGE_PATH": Directory "/Users/tomekzaw/(...)/node_modules/react-native-reanimated/android/build/intermediates/prefab_package_configuration/debug/prefab" is not readable.
```

## Changes

1. We need to set `prefabPublishing true` in order to enable exposing libraries as prefabs.
2. We need `prefabHeadersDir` because `headers` accepts only one path. So we copy all headers from `src/main/cpp` and `Common/cpp/**` directly to `prefabHeadersDir` in `prepareHeadersForPrefab` task.
3. We need to list all directories in `Common/cpp` because in our codebase we use `#include "SomeHeader.h"` without any directory prefix so `SomeHeader.h` needs to be directly in `prefabHeadersDir`.
4. We need to call `prefabHeadersDir.mkdirs()` because otherwise Gradle Sync fails due to non-existent path.
5. We need `preBuild.dependsOn(prepareHeadersForPrefab)` so that the headers are already copied before CMake starts building other parts that depend on it.

## Test plan

🧨

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -482,6 +482,7 @@ task printVersions {
 task createNativeDepsDirectories() {
     downloadsDir.mkdirs()
     thirdPartyNdkDir.mkdirs()
+    prefabHeadersDir.mkdirs()
 }
 
 def resolveTaskFactory(String taskName, String artifactLocalName, File reactNativeAndroidDownloadDir, File reanimatedDownloadDir) {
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):
👏

PR #5008, opened 2023-08-30
Don't warn against overwriting styles in case of layout animation

Latropos:
## Summary

Don't warn against overwriting styles in case of layout animation, since most of use cases are valid

## Test plan

Tested with entering, layout and exiting animations

### DIFF
--- a/src/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent.tsx
@@ -621,7 +621,7 @@ export default function createAnimatedComponent(
               LayoutAnimationType.LAYOUT,
               maybeBuild(
                 layout,
-                this.props?.style,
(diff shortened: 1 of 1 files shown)


src/createAnimatedComponent.tsx
```diff
@@ -621,7 +621,7 @@ export default function createAnimatedComponent(
               LayoutAnimationType.LAYOUT,
               maybeBuild(
                 layout,
-                this.props?.style,
+                undefined /* We don't have to warn user if style has common properties with animation for LAYOUT */,
```
tomekzaw, line 624: What is "animation for LAYOUT"?

Latropos: Usually its `layout={Layout}`

PR #5009, opened 2023-08-30
Add JSProps on Fabric

piaskowyk:
## Summary

This pull request adds support for JS props in Fabric. The implementation is available in the common codebase, so there is no platform-specific code required. This approach doesn't rely on events since, based on my research, using events would require messy workarounds with React Native.
Same thing on Paper: https://github.com/software-mansion/react-native-reanimated/pull/4821

## Test plan

<details>
<summary>code</summary>

```js
import { TextInput } from 'react-native';
import Svg, {
  Path,
  Circle,
  G,
} from 'react-native-svg';
import Animated, {
  runOnJS,
  useSharedValue,
  useDerivedValue,
  useAnimatedProps,
  useAnimatedGestureHandler,
  interpolate,
import {PanGestureHandler} from 'react-native-gesture-handler';

const BORDER_WIDTH = 25;
const DIAL_RADIUS = 22.5;

export const {PI} = Math;
export const TAU = 2 * PI;

const AnimatedG = Animated.createAnimatedComponent(G);
const AnimatedInput = Animated.createAnimatedComponent(TextInput);

const polarToCartesian = (
  angle: number,
  radius: number,
  {x, y}: {x: number; y: number},
) => {
  'worklet';
  const a = ((angle - 90) * Math.PI) / 180.0;
  return {x: x + radius * Math.cos(a), y: y + radius * Math.sin(a)};
};

const cartesianToPolar = (
  x: number,
  y: number,
  {x: cx, y: cy}: {x: number; y: number},
  step = 1,
) => {
  'worklet';

  const value =
    Math.atan((y - cy) / (x - cx)) / (Math.PI / 180) + (x > cx ? 90 : 270);

  return Math.round(value * (1 / step)) / (1 / step);
};

const unMix = (value: number, x: number, y: number) => {
  'worklet';
  return (value - x) / (y - x);
};

type Props = {
  width: number;
  height: number;
  fillColor: string[];
  value: number;
  meterColor: string;
  min?: number;
  max?: number;
  onValueChange: (value: any) => void;
  children: (
    props: Partial<{defaultValue: string; text: string}>,
  ) => React.ReactNode;
  step?: number;
  decimals?: number;
};

const CircleSlider = (props: Props) => {
  const {
    width,
    height,
    value,
    meterColor,
    children,
    min,
    max,
    step = 1,
    decimals,
  } = props;
  const smallestSide = Math.min(width, height);

  const cx = width / 2;
  const cy = height / 2;
  const r = (smallestSide / 2) * 0.85;

  const start = useSharedValue(0);
  const end = useSharedValue(unMix(value, min! / 360, max! / 360));

  useEffect(() => {
    end.value = unMix(value, min! / 360, max! / 360);
  }, [value, end, max, min]);

  const startPos = useDerivedValue(() =>
    polarToCartesian(start.value, r, {x: cx, y: cy}),
  );

  const endPos = useDerivedValue(() =>
    polarToCartesian(end.value, r, {x: cx, y: cy}),
  );

  const animatedPath = useAnimatedProps(() => {
    const p1 = startPos.value;
    const p2 = endPos.value;
    return {
      d: `M${p1.x} ${p1.y} A ${r} ${r} 0 ${end.value > 180 ? 1 : 0} 1 ${p2.x} ${
        p2.y
      }`,
    };
  });

  const animatedCircle = useAnimatedProps(() => {
    const p2 = endPos.value;
    return {
      x: p2.x - 7.5,
      y: p2.y - 7.5,
    };
  });

  const animatedChildrenProps = useAnimatedProps(() => {
    const decimalCount = (num: number) => {
      if (decimals) {
        return decimals;
      }
      const numStr = String(num);
      if (numStr.includes('.')) {
        return numStr.split('.')[1].length;
      }
      return 0;
    };

    const value = interpolate(end.value, [min! / 360, max! / 360], [min! / 360, max! / 360]);

    return {
      defaultValue: `${value.toFixed(decimalCount(step))}`,
      text: `${value.toFixed(decimalCount(step))}`,
    };
  });

    onActive: ({x, y}: {x: number; y: number}, ctx: any) => {
      const value = cartesianToPolar(x, y, {x: cx, y: cy}, step);

      ctx.value = interpolate(value, [min! / 360, max! / 360], [min! / 360, max! / 360]);
      end.value = value;
    },
    onFinish: (_, ctx) => {
      runOnJS(props.onValueChange)(ctx.value);
    },
  });
  
  return (
      <Animated.View>
        <Svg width={width} height={height}>
          <Circle cx={cx} cy={cy} r={r + BORDER_WIDTH / 2 - 1} fill="blue" />
          <Circle
            cx={cx}
            cy={cy}
            r={r}
            strokeWidth={BORDER_WIDTH}
            fill="url(#fill)"
            stroke="rgba(255, 255, 525, 0.2)"
          />
          <AnimatedPath
            stroke={meterColor}
            strokeWidth={BORDER_WIDTH}
            fill="none"
            animatedProps={animatedPath}
          />
          <AnimatedG animatedProps={animatedCircle} onPress={() => {}}>
            <Circle cx={7.5} cy={7.5} r={DIAL_RADIUS} fill={meterColor} />
          </AnimatedG>
        </Svg>
        {children && children(animatedChildrenProps)}
      </Animated.View>
    </PanGestureHandler>
  );
};

CircleSlider.defaultProps = {
  width: 300,
  height: 300,
  fillColor: ['#fff'],
  meterColor: '#fff',
  min: 0,
  max: 359,
  step: 1,
  onValueChange: (x: any) => x,
};

export default function EmptyExample(): JSX.Element {
  const handleChange = value => console.log(value);
  console.log('EmptyExample');
  return (
    <CircleSlider
      width={325}
      height={325}
      value={0}
      meterColor={'#ffffff'}
      onValueChange={handleChange}>
      {animatedProps => (
        <AnimatedInput
          keyboardType="numeric"
          maxLength={3}
          selectTextOnFocus={false}
          animatedProps={animatedProps}
        />
      )}
    </CircleSlider>
  );
}

```

</details>


src/JSPropUpdater.ts
```diff
@@ -55,11 +66,49 @@ export class JSPropUpdater {
...
+
+  public removeOnJSPropsChangeListener(
+    animatedComponent: React.Component<unknown, unknown>
+  ) {
+    const viewTag = findNodeHandle(animatedComponent);
+    JSPropUpdaterFabric._tagToComponentMapping.delete(viewTag);
+  }
+}
+
+export default global._IS_FABRIC ? JSPropUpdaterFabric : JSPropUpdaterPaper;
```
tjzel, line 114: Do we want to make a fallback for e.g. web here or not necessarily?

tjzel (approved):
Looks good, except for minor typos!

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -392,6 +398,30 @@ bool NativeReanimatedModule::isThereAnyLayoutProp(
   }
   return false;
 }
+
+jsi::Value NativeReanimatedModule::getNonAnimatableProp(
+    jsi::Runtime &rt,
+    const jsi::Value &props) {
+  jsi::Object nonAnimatableProps(rt);
+  bool isAnyProp = false;
```
tomekzaw, line 406: I'm not sure if I understand this name.

src/reanimated2/globals.d.ts
```diff
@@ -100,4 +100,5 @@ declare global {
   var LayoutAnimationsManager: LayoutAnimationsManager;
   var UpdatePropsManager: UpdatePropsManager;
   var ProgressTransitionRegister: ProgressTransitionRegister;
+  var _jsPropsUpdater: (viewTag: number, props: unknown) => void;
```
tomekzaw, line 103: According to our convention, we use single leading underscore to represent JSI bindings, but this function is actually written in pure JavaScript. Can we rename it? Also, can we rename it to `updateJSProps` instead of `updater`? Finally, can we type `props` accordingly instead of using `unknown`?

src/JSPropUpdater.ts
```diff
@@ -55,11 +66,49 @@ export class JSPropUpdater {
...
+    animatedComponent: React.Component<unknown, unknown>
+  ) {
+    const viewTag = findNodeHandle(animatedComponent);
+    JSPropUpdaterFabric._tagToComponentMapping.set(viewTag, animatedComponent);
+  }
+
+  public removeOnJSPropsChangeListener(
+    animatedComponent: React.Component<unknown, unknown>
+  ) {
+    const viewTag = findNodeHandle(animatedComponent);
```
tomekzaw, line 109:
```suggestion
    const shadowNodeWrapper = findNodeHandle(animatedComponent);
```

tomekzaw (changes requested):

src/JSPropUpdater.ts
```diff
@@ -55,11 +66,49 @@ export class JSPropUpdater {
...
+  constructor() {
+    if (!JSPropUpdaterFabric.isInitialized) {
+      const updater = (viewTag: number, props: unknown) => {
+        const component =
+          JSPropUpdaterFabric._tagToComponentMapping.get(viewTag);
+        component?._updateFromNative(props);
+      };
+      runOnUIImmediately(() => {
+        'worklet';
+        global._jsPropsUpdater = (viewTag: number, props: unknown) => {
```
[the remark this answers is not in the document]
tomekzaw: Also, why can't we just install JS props updater in the main RN runtime and use it from C++ via `JSScheduler`?

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -289,11 +290,16 @@ jsi::Value NativeReanimatedModule::configureProps(
     const jsi::Value &uiProps,
     const jsi::Value &nativeProps) {
 #ifdef RCT_NEW_ARCH_ENABLED
-  (void)uiProps; // unused variable on Fabric
-  jsi::Array array = nativeProps.asObject(rt).asArray(rt);
+  jsi::Array array = uiProps.asObject(rt).asArray(rt);
+  for (size_t i = 0; i < array.size(rt); ++i) {
+    std::string name = array.getValueAtIndex(rt, i).asString(rt).utf8(rt);
+    animatableProps_.insert(name);
```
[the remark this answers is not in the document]
piaskowyk: This is sum of UI and native props sets

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -489,6 +519,23 @@ void NativeReanimatedModule::performOperations() {
...
+  for (const auto &[shadowNode, props] : copiedOperationsQueue) {
+    const jsi::Value &nonAnimatableProps = getNonAnimatableProp(rt, *props);
+    if (nonAnimatableProps.isUndefined()) {
+      continue;
+    }
+    Tag viewTag = shadowNode->getTag();
+    uiScheduler_->scheduleOnUI([&]() {
+      jsi::Value jsPropsUpdaterValue =
+          rt.global().getProperty(rt, "_jsPropsUpdater");
+      if (jsPropsUpdaterValue.isObject()) {
```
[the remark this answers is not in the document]
piaskowyk: I'll solve this problem by usage of JSLogger 🫡

src/reanimated2/globals.d.ts
piaskowyk: > "can we type props accordingly instead of using unknown?" 

The best what I can do is `Record` because there is no rules (or I don't know about them) how JS props can looks like 🤔

src/JSPropUpdater.ts
```diff
@@ -55,11 +66,49 @@ export class JSPropUpdater {
...
+    if (JSPropUpdaterPaper._tagToComponentMapping.size === 0) {
       this._reanimatedEventEmitter.removeAllListeners(
         'onReanimatedPropsChange'
       );
     }
   }
 }
+
+class JSPropUpdaterFabric implements JSPropUpdater {
+  private static _tagToComponentMapping = new Map();
```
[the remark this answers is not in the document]
piaskowyk: In this case this is really mapping from ViewTag to Component not ShadowNodeWrapper to Component

src/JSPropUpdater.ts
piaskowyk: Are you sure that findNodeHandle returns shadowNodeWrapper? According to the source code it should return just a number - https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/ReactNative/RendererImplementation.js#L47-L49. Also in every other place in our code we use `findNodeHandle` to obtain just a view tag. I think that `findHostInstance_DEPRECATED` returns a shadow node wrapper, but correct me if I'm wrong.

Common/cpp/NativeModules/NativeReanimatedModule.cpp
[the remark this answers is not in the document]
tjzel: 🫡

src/JSPropUpdater.ts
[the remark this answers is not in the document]
piaskowyk: We could use this approach, but I believe it would introduce more problems. The issue lies in copying the `nonAnimatableProps` to the second runtime. Since it is a `jsi::Value`, we cannot simply copy it.

However, we rely on the `nonAnimatableProps` in the UI to check for any JS prop and determine if we need to schedule an operation to the React runtime. Additionally, we need the `nonAnimatableProps` on React runtime to apply changes to the view.

One proposed solution is to calculate the `nonAnimatableProps` twice. However, I consider this approach to be both costly and unnecessary.

src/JSPropUpdater.ts
piaskowyk: It seems that it never worked on web 🫣

src/JSPropUpdater.ts
tomekzaw: You're right, `findHostInstance_DEPRECATED` returns a ShadowNodeWrapper. I'm not sure what's the result of `findNodeHandle`, I would need to check it. Anyway, the question is, why do we need this number on Fabric? I thought the only way to identify a RN component in C++ when updating props is to pass a ShadowNodeWrapper 🤔

src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -104,7 +104,7 @@ export function createAnimatedComponent(
     animatedStyle: { value: StyleProps } = { value: {} };
     _component: ComponentRef | HTMLElement | null = null;
     _sharedElementTransition: SharedTransition | null = null;
-    _JSPropUpdater = new JSPropUpdater();
+    _JSPropsUpdater = new JSPropsUpdater();
```
tomekzaw, line 107:
```suggestion
    _jsPropsUpdater = new JSPropsUpdater();
```

src/createAnimatedComponent/JSPropsUpdater.ts
```diff
@@ -0,0 +1,153 @@
...
+    _animatedComponent: React.Component<unknown, unknown>
+  ) {
+    // noop
+  }
+}
+
+function getJSPropsUpdater():
+  | typeof JSPropsUpdaterWeb
+  | typeof JSPropsUpdaterFabric
+  | typeof JSPropsUpdaterPaper {
```
tomekzaw, line 144: Can we assign this type above?
```ts
type JSPropsUpdater =
  | typeof JSPropsUpdaterWeb
  | typeof JSPropsUpdaterFabric
  | typeof JSPropsUpdaterPaper;

function getJSPropsUpdater(): JSPropsUpdater {
  // ...
}
```
Also, do we really need to specify the return type here?

src/createAnimatedComponent/JSPropsUpdater.ts
```diff
@@ -0,0 +1,153 @@
...
+  if (shouldBeUseWeb()) {
+    return JSPropsUpdaterWeb;
+  } else if (global._IS_FABRIC) {
+    return JSPropsUpdaterFabric;
+  } else {
+    return JSPropsUpdaterPaper;
+  }
+}
+export default getJSPropsUpdater();
```
tomekzaw, line 153: Do we even need a function here? Why not just export a value like we do in `NativeMethods.ts`?

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -487,6 +517,28 @@ void NativeReanimatedModule::performOperations() {
...
+      jsi::Function jsPropsUpdater =
+          maybeJSPropUpdater.asObject(rt).asFunction(rt);
+      jsPropsUpdater.call(rt, viewTag, nonAnimatableProps);
+#if DEBUG
+    } else {
+      JSLogger jsLogger(jsScheduler_);
+      jsLogger.warnOnJS("[Reanimated] Unable to update JS props.");
+    }
+#else
+    }
```
tomekzaw, line 538: The compiler will optimize away the empty `else` branch here on its own, let's keep things simple.
```suggestion
    } else {
#if DEBUG
      JSLogger jsLogger(jsScheduler_);
      jsLogger.warnOnJS("[Reanimated] Unable to update JS props.");
#endif
    }
```

src/createAnimatedComponent/JSPropsUpdater.ts
piaskowyk: I think we should because Typescript can't recognise correct type of JSPropsUpdater. Here is example:
Without type specification:
Screenshot 2023-10-13 at 10 45 02
With type specification:
Screenshot 2023-10-13 at 10 44 30

src/JSPropUpdater.ts
piaskowyk: In Fabric, the view tag is not required. However, I utilized the result of `findNodeHandle()` (which is a number) to store React JavaScript components in the JavaScript dictionary called `_tagToComponentMapping`.

```js
class JSPropsUpdaterFabric implements JSPropsUpdater {
  private static _tagToComponentMapping = new Map();
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -393,6 +399,30 @@ bool NativeReanimatedModule::isThereAnyLayoutProp(
   }
   return false;
 }
+
+jsi::Value NativeReanimatedModule::getNonAnimatableProps(
```
tomekzaw, line 403:
```suggestion
jsi::Value NativeReanimatedModule::filterNonAnimatableProps(
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -393,6 +399,30 @@ bool NativeReanimatedModule::isThereAnyLayoutProp(
...
+      const auto &propNameStr = propName.c_str();
+      const jsi::Value &propValue = propsObject.getProperty(rt, propNameStr);
+      nonAnimatableProps.setProperty(rt, propNameStr, propValue);
+    }
+  }
+  if (isAnyNonAnimatableProp) {
+    return nonAnimatableProps;
+  } else {
+    return jsi::Value::undefined();
+  }
```
tomekzaw, line 424:
```suggestion
  if (!hasAnyNonAnimatableProp) {
    return jsi::Value::undefined();
  }
  return nonAnimatableProps;
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -393,6 +399,30 @@ bool NativeReanimatedModule::isThereAnyLayoutProp(
   }
   return false;
 }
+
+jsi::Value NativeReanimatedModule::getNonAnimatableProps(
+    jsi::Runtime &rt,
+    const jsi::Value &props) {
+  jsi::Object nonAnimatableProps(rt);
+  bool isAnyNonAnimatableProp = false;
```
tomekzaw, line 407:
```suggestion
  bool hasAnyNonAnimatableProp = false;
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -494,6 +524,26 @@ void NativeReanimatedModule::performOperations() {
...
+    if (maybeJSPropUpdater.isObject()) {
+      jsi::Function jsPropsUpdater =
+          maybeJSPropUpdater.asObject(rt).asFunction(rt);
+      jsPropsUpdater.call(rt, viewTag, nonAnimatableProps);
+    } else {
+#ifndef NDEBUG
+      JSLogger jsLogger(jsScheduler_);
+      jsLogger.warnOnJS("[Reanimated] Unable to update JS props.");
+#endif
+    }
```
tomekzaw, line 544:
```suggestion
    assert(maybeJSPropUpdater.isObject() && "[Reanimated `updateJSProps` not found");
    jsi::Function jsPropsUpdater =
        maybeJSPropUpdater.asObject(rt).asFunction(rt);
    jsPropsUpdater.call(rt, viewTag, nonAnimatableProps);
```

src/createAnimatedComponent/JSPropsUpdater.ts
```diff
@@ -0,0 +1,170 @@
...
+    this._reanimatedEventEmitter = new NativeEventEmitter(reanimatedModule);
+  }
+
+  public addOnJSPropsChangeListener(
+    animatedComponent: React.Component<
+      AnimatedComponentProps<InitialComponentProps>
+    > &
+      IAnimatedComponentInternal
+  ) {
+    const viewTag = findNodeHandle(animatedComponent);
```
tomekzaw, line 53: Let's just make sure that we never use `findNodeHandle` on web.

src/createAnimatedComponent/JSPropsUpdater.ts
```diff
@@ -0,0 +1,170 @@
+'use strict';
+import {
+  NativeEventEmitter,
+  NativeModules,
+  findNodeHandle,
+} from 'react-native';
+import {
+  nativeShouldBeMock,
```
tomekzaw, line 8: There's already a PR which eliminates `nativeShouldBeMock` (#5310), do you think we should use `!shouldBeUseWeb()` here or do it in a separate PR?

src/createAnimatedComponent/JSPropsUpdater.ts
```diff
@@ -0,0 +1,170 @@
...
+    },
+  };
+
+  constructor() {
+    let reanimatedModule: typeof JSPropsUpdaterPaper._reanimatedModuleMock;
+    if (nativeShouldBeMock()) {
+      reanimatedModule = JSPropsUpdaterPaper._reanimatedModuleMock;
+    } else {
+      reanimatedModule = NativeModules.ReanimatedModule;
+    }
```
tomekzaw, line 43:
```suggestion
    const reanimatedModule = nativeShouldBeMock()
        ? JSPropsUpdaterPaper._reanimatedModuleMock;
        : NativeModules.ReanimatedModule;
```

tomekzaw (changes requested):

src/createAnimatedComponent/JSPropsUpdater.ts
piaskowyk: I did it in this PR 🫡

src/createAnimatedComponent/JSPropsUpdater.ts
piaskowyk: I think this is enough to keep it safe:
```js
let JSPropsUpdater: JSPropsUpdaterOptions;
if (SHOULD_BE_USE_WEB) {
  JSPropsUpdater = JSPropsUpdaterWeb;
} else if (global._IS_FABRIC) {
  JSPropsUpdater = JSPropsUpdaterFabric;
} else {
  JSPropsUpdater = JSPropsUpdaterPaper;
}
export default JSPropsUpdater;
```

Issue #5002, opened 2023-08-29
[nightly-20230828] 'JsiUtils.h' file not found


efstathiosntonas:
### Description

After installing `3.5.0-nightly-20230828-e860ae5c4` iOS cannot compile, it throws `'JsiUtils.h' file not found` on `WorkletRuntimeDecorator`. This started on `20230826` build. 

Android builds fine. Please note that I use `expo-modules` in a bare react-native project.

I've uninstalled and reinstalled node_modules, cleared derived data, deleted pods folder and podfile.lock.

![Screenshot 2023-08-29 at 09 09 58](https://github.com/software-mansion/react-native-reanimated/assets/717975/071e8cf9-cf7f-4442-b474-93517fb51ee2)

### Steps to reproduce

1. Update to `3.5.0-nightly-20230828-e860ae5c4`
2. Run app with Xcode

### Snack or a link to a repository

/

### Reanimated version

3.5.0-nightly-20230828-e860ae5c4

### React Native version

0.72.4

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

efstathiosntonas:
probably offending [PR](https://github.com/software-mansion/react-native-reanimated/commit/cb04295c9b574342bcc3e68db42a83742acbdadd#diff-02b248e4321a5e3cf1a4866f35dc89a7c8e310c3eeaf3a4f15d1dc3e3f690a9b)

piaskowyk:
Hey there! 👋 Thanks for your report! We really appreciate having such active and committed people like you in our community 😍 We've also noticed an issue with expo-dev-client on our CI. You can find more information about it here: [CI Issue](https://github.com/software-mansion/react-native-reanimated/actions/runs/6004184473). I'll try to fix it.

efstathiosntonas:
Thanks @piaskowyk, let me know if you want me to test any fixes!

efstathiosntonas:
Thanks @piaskowyk! Keep it up!

PR #5010, opened 2023-08-30
Resolve headers name conflict with expo-dev-client

piaskowyk:
## Summary

The `Check Expo dev-client nightly build` CI (https://github.com/software-mansion/react-native-reanimated/actions/runs/6016772546) fails with error:
```bash
❌  /Users/runner/work/react-native-reanimated/react-native-reanimated/app/
node_modules/react-native-reanimated/Common/cpp/ReanimatedRuntime/
WorkletRuntimeDecorator.cpp:59:3: use of undeclared identifier 'jsi_utils'

  jsi_utils::installJsiFunction(
```

This PR fixes the issue of header name conflict with expo-dev-client. The Expo package includes its own JsiUtils.h file, which leads to a collision in header mapping within Xcode. To resolve this problem, we have three possible options:
- Use an absolute import path (doesn't work on Android).
- Rename JsiUtils.h (as done in this PR).
- Utilize namespace imports like #include <Reanimated/...>.

In the past we encountered similar issue with Xcode flat header mapping - https://github.com/software-mansion/react-native-reanimated/pull/1775

Fixes https://github.com/software-mansion/react-native-reanimated/issues/5002

## Test plan

Check CI status: https://github.com/software-mansion/react-native-reanimated/actions/runs/6034169391?pr=5010


tomekzaw (approved):
👏

Issue #4992, opened 2023-08-27
[iOS building] [React native 0.72.4] No visible @interface for 'RNSScreenView' declares the selector 'isModal'


konri:
### Description

Hi guys, 
I have a problem after updating react native to the latest version. When I want to build my application via xcode I got the information: 
/my_path/node_modules/react-native-reanimated/ios/LayoutReanimation/REAScreensHelper.m:41:28 No visible @interface for 'RNSScreenView' declares the selector 'isModal'

![Screenshot 2023-08-27 at 23 35 41](https://github.com/software-mansion/react-native-reanimated/assets/3826473/694016ee-fcbf-4432-a6c2-0a59ca5b2ae2)

I need to remove this piece of code to run my application. 
```
bool isModal = [screen isModal];
    if (!isModal) {
      // case for modal with header
      RNSScreenView *parentScreen = (RNSScreenView *)[REAScreensHelper getScreenForView:screen.reactSuperview];
      if (parentScreen != nil) {
        isModal = [parentScreen isModal];
      }
    }
```

### Steps to reproduce

1. Update with https://react-native-community.github.io/upgrade-helper/?from=0.70.3&to=0.72.4 to latest react native version 
2. build application for ios with xcode 
3. problem occured at the end of building. 

### Snack or a link to a repository

https://react-native-community.github.io/upgrade-helper/?from=0.70.3&to=0.72.4

### Reanimated version

3.4.0

### React Native version

0.72.4

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

piaskowyk:
Ok, I know how to fix it. But overall you just need to upgrade you screens version to solve this issue

manurodriguezDL:
@piaskowyk Could you give any more information on how to fix it? I'm facing the same issue.

Korin92:
I have same problema, someone know how to fix this?

Korin92:
I fix this, I had to update react-native-screens to the last version

noviogroup:
@Korin92  I fix this, I had to update react-native-reanimated, react-native-screens and react-native-safe-area-context to the last version

Yes this worked, some reason i had to use yarn add because npm gave errors.

eisodev:
[...]

PR #5014, opened 2023-08-31
Fix compatibility with older rn-screens version

piaskowyk:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/4992
This issue may happen when someone uses an older version of RN screens without the `isModal` method. This PR adds a check to detect if this selector exists. If not, the whole SET feature is disabled.

## Test plan

Try to compile the App with the newest Reaniamted and the old version of rn-screens.

### DIFF
--- a/apple/LayoutReanimation/REASharedTransitionManager.m
+++ b/apple/LayoutReanimation/REASharedTransitionManager.m
@@ -341,7 +341,8 @@ static REASharedTransitionManager *_sharedTransitionManager;
     Class screenClass = [RNSScreen class];
     Class screenViewClass = [RNSScreenView class];
     BOOL allSelectorsAreAvailable = [RNSScreen instancesRespondToSelector:viewDidLayoutSubviewsSelector] &&
-        [RNSScreenView instancesRespondToSelector:notifyWillDisappearSelector];
+        [RNSScreenView instancesRespondToSelector:notifyWillDisappearSelector] &&
+        [RNSScreenView instancesRespondToSelector:@selector(isModal)]; // used by REAScreenHelper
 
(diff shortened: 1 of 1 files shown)


apple/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -341,7 +341,8 @@ - (void)swizzleScreensMethods
     Class screenClass = [RNSScreen class];
     Class screenViewClass = [RNSScreenView class];
     BOOL allSelectorsAreAvailable = [RNSScreen instancesRespondToSelector:viewDidLayoutSubviewsSelector] &&
-        [RNSScreenView instancesRespondToSelector:notifyWillDisappearSelector];
+        [RNSScreenView instancesRespondToSelector:notifyWillDisappearSelector] &&
+        [RNSScreenView instancesRespondToSelector:@selector(isModal)]; // used by REAScreenHelper
```
tomekzaw, line 345: Why do we need `@selector` here but we don't in the line above?

piaskowyk: We also use selectors but define a couple of lines above: https://github.com/software-mansion/react-native-reanimated/pull/5014/files#diff-b62ac3002c1da6d8f9df0d7d5cce57bd26cf59c1ceaeb77dcab4ac4475e2b093R340

`@selector(viewDidLayoutSubviews)` and `@selector(notifyWillDisappear)` are assigned to `SEL` object because I used them later, but `@selector(isModal)` is used only in one place.

tomekzaw (approved):

PR #5015, opened 2023-08-31
Fix crash on reload in `ShareableRemoteFunction`

tomekzaw:
## Summary

This PR fixes a crash that occurs during a reload of the app. `ShareableRemoteFunction` holds a reference to a remote function in `function_` field. When `ShareableRemoteFunction` is deallocated, `function_` also gets deallocated. If the runtime has already been terminated, the assert in `~jsi::Pointer` will fail.

Zrzut ekranu 2023-08-31 o 10 48 42

## Test plan

Tested on VisionCameraExample.


piaskowyk (approved):
🚀

PR #5022, opened 2023-09-01
Headers sufficiency check

michalmaka:
## Summary

Script for checking whether all headers in Common directory are self-sufficient, i.e. really includes whatever is needed by them.

Currently it needs to have android prefab extracted somewhere and path to it set as `ANDROID_PREFAB_CMAKE_DIR` in `scripts/headers-sufficiency-check.sh`. For testing purposes I've built android `FabricExample` app and taken prefab from one of downloaded dirs.

It also need to have Android NDK path set in the script.

What's needed is integration with CI pipelines.

## Test plan

None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ react-native-reanimated-tests.js
 # plugin
 !plugin/build
 plugin/types
+
+# headers-sufficiency-checker
+headers-sufficiency-check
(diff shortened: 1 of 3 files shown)


tomekzaw:
Closing this PR for now because it's a low-prio and we don't have enough capacity to resume working on this

PR #5032, opened 2023-09-04
Add reduce motion on macOS

jfedak:
## Summary

This PR add support for Reduce Motion setting on macOS platform  by checking `accessibilityDisplayShouldReduceMotion` property from `NSWorkspace`.

## Test plan

Tested with `Reduced Motion` example.

### DIFF
--- a/apple/REAModule.mm
+++ b/apple/REAModule.mm
@@ -270,7 +270,7 @@ RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(installTurboModule)
 #if __has_include(<UIKit/UIAccessibility.h>)
     auto isReducedMotion = UIAccessibilityIsReduceMotionEnabled();
 #else
-    auto isReducedMotion = false;
+    auto isReducedMotion = NSWorkspace.sharedWorkspace.accessibilityDisplayShouldReduceMotion;
 #endif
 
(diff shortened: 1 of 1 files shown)


apple/REAModule.mm
```diff
@@ -270,7 +270,7 @@ - (void)sendEventWithName:(NSString *)eventName body:(id)body
 #if __has_include(<UIKit/UIAccessibility.h>)
     auto isReducedMotion = UIAccessibilityIsReduceMotionEnabled();
 #else
-    auto isReducedMotion = false;
+    auto isReducedMotion = NSWorkspace.sharedWorkspace.accessibilityDisplayShouldReduceMotion;
```
tomekzaw, line 273: Which syntax should we use?
```suggestion
    auto isReducedMotion = [[NSWorkspace sharedWorkspace] accessibilityDisplayShouldReduceMotion];
```

jfedak: According to the [Apple documentation](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html), both syntaxes are equivalent. 

> Getting a value using somePerson.firstName is the same as using [somePerson firstName]

tomekzaw (approved):

PR #5037, opened 2023-09-05
Update useSharedValue.mdx - Example with an object as a shared value

cdehaan:
Example of how to update a share value which is an object.

## Summary

Just an addition to the documentation, add if you think it might be helpful. I got it wrong initially, and actually ChatGPT gave incorrect advice when I asked it.

## Test plan

No test needed, I think, since it's documentation.

### DIFF
--- a/docs/docs/core/useSharedValue.mdx
+++ b/docs/docs/core/useSharedValue.mdx
@@ -88,6 +88,22 @@ function App() {
 
 </Indent>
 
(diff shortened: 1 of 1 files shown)


kacperkapusciak:
Hi @cdehaan 

Thanks for pointing this out! 

I fixed the CI and simplify the idea/code a bit in https://github.com/software-mansion/react-native-reanimated/pull/5071 so I'm going to close this PR

Cheers ✌️

PR #5041, opened 2023-09-05
Fix accelerometer on android

bartlomiejbloniarz:
## Summary

This PR fixes the accelerometer type used in the listener on android.

## Test plan

Check `useAnimatedSensor - accelerometer` example


tomekzaw (approved):
Thanks for fixing this! 🤩

PR #5043, opened 2023-09-05
Notify users to run `setNativeProps` on the UI runtime

kacperkapusciak:
## Summary

This PR adds a check whether the user calls `setNativeProps` on the UI runtime. Currently, running `setNativeProps` from JS thread results in a non-relevant runtime crash.

## Test plan

```jsx
import React from 'react';
import Animated, {
  runOnUI,
  setNativeProps,
  useAnimatedRef,

  const animatedRef = useAnimatedRef<TextInput>();

  const handlePress = () => {
    // runOnUI(() => {
    setNativeProps(animatedRef, { backgroundColor: 'blue' });
    // })();
  };

  return (
      <Animated.View ref={animatedRef} style={styles.box} />
      <Button title="set native prop" onPress={handlePress} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: { width: 100, height: 100, backgroundColor: 'red' },
});

```

### DIFF
--- a/src/reanimated2/SetNativeProps.ts
+++ b/src/reanimated2/SetNativeProps.ts
@@ -28,7 +28,7 @@ if (isWeb()) {
     'worklet';
     if (!_WORKLET) {
       console.warn(
-        '[Reanimated] `setNativeProps` can only be used on the UI runtime.'
+        '[Reanimated] setNativeProps() can only be used on the UI runtime.'
       );
       return;
     }
@@ -42,7 +42,7 @@ if (isWeb()) {
     'worklet';
     if (!_WORKLET) {
       console.warn(
-        '[Reanimated] `setNativeProps` can only be used on the UI runtime.'
+        '[Reanimated] setNativeProps() can only be used on the UI runtime.'
       );
       return;
     }
@@ -55,17 +55,17 @@ if (isWeb()) {
 } else if (isChromeDebugger()) {
   setNativeProps = () => {
     console.warn(
-      '[Reanimated] `setNativeProps` is not supported with Chrome Debugger.'
+      '[Reanimated] setNativeProps() is not supported with Chrome Debugger.'
     );
   };
 } else if (isJest()) {
   setNativeProps = () => {
-    console.warn('[Reanimated] `setNativeProps` is not supported with Jest.');
+    console.warn('[Reanimated] setNativeProps() is not supported with Jest.');
   };
 } else {
   setNativeProps = () => {
     console.warn(
-      '[Reanimated] `setNativeProps` is not supported on this configuration.'
+      '[Reanimated] setNativeProps() is not supported on this configuration.'
     );
   };
 }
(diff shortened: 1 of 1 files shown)


src/reanimated2/SetNativeProps.ts
```diff
@@ -26,6 +26,11 @@ if (isWeb()) {
 } else if (IS_NATIVE && global._IS_FABRIC) {
   setNativeProps = (animatedRef, updates) => {
     'worklet';
+    if (!_WORKLET) {
+      throw new Error(
+        '[Reanimated] `setNativeProps` can only be used on the UI runtime. Please call the function with `runOnUI` or run it in a worklet instead.'
```
tjzel, line 31: While this comment is definitely good, I can't help but think that "run it in a worklet instead" could be misleading at times. Some worklets are called on JS thread, even deliberately. Also we have such cases like:

```TS
function MyInnerComponent(props: { someRef: any }) {
    setNativeProps(props.someRef, {});
    return {};
  });
  return <Animated.View style={animatedStyle} />;
}

function MyOuterComponent() {
  const someRef = React.useRef(null);
  return (
    <View ref={someRef}>
      <MyInnerComponent someRef={someRef} />
    </View>
  );
}
```

`useAnimatedStyle` executes on JS thread first, therefore we get crash here (I checked). User would have to check first if he's on UI thread first, same as it was with `measure` and we don't want that.

I would lean more toward the way we handle it with `measure` - silently failing - but I'm open for discussion.

tomekzaw: I agree with Tomek that "run it in a worklet instead" can be misleading in this case. `setNativeProps` is meant only to be used on the UI runtime. In this sense, it's kinda similar to `measure`. However, a silent fail is not a great solution here because `setNativeProps` has no return value and `measure` simply returns `null` (which is also sometimes does on the UI thread so the developer needs to handle it). I'd suggest using `console.warn` here with an error message "[Reanimated] `setNativeProps` can only be used on the UI runtime." which effectively is a no-op in release mode.

tjzel: But then in this case you'd get constant warnings in the console. And we all know the tragic story from gesture handler about this...

tomekzaw: Yep but only if you use it incorrectly (i.e. from the JS thread) and also it's a totally new feature so there should be no regressions.

kacperkapusciak: It won't be no-op in case of a warn because without this error it crashes later on with:

image

because `animatedRef.viewName.value` is undefined on JS runtime

kacperkapusciak: I've updated the error message to "[Reanimated] `setNativeProps` can only be used on the UI runtime." in [20be9ea](https://github.com/software-mansion/react-native-reanimated/pull/5043/commits/20be9eae511b783a1914c1a91044cd1b0de1d68d)

tomekzaw (approved):
Thanks for adding this! ❤️

src/reanimated2/SetNativeProps.ts
kacperkapusciak: I made it no-op in the end 😅 

image

tjzel (approved):
:prayge:

PR #5048, opened 2023-09-06
Add missing HostObject serialization in `makeShareableCloneOnUIRecursive`

tjzel:
## Summary

#4617 Changed the logic of `makeShareableCloneOnUIRecursive`, detecting if the object is a `HostObject` and skipping its serialization, on the assumption that it's already good to go. This caused crashes in `extractShareableOrThrow` due to how `isHostObject<T>` function works - it not only checks if the object is a `HostObject` but also if it's of type `T`.

image

Therefore, given a `HostObject` from `react-native`, in this case `ShadowNodeWrapper` on Fabric, it wasn't converted to our `Shareable` type and failed the second check in `extractShareableOrThrow`.

Big thanks to @piaskowyk who debugged the most of it 🥳 

## Test plan

See that without the change in `makeShareableCloneOnUIRecursive` the new error is thrown.

## Note

I have added some (seemingly) unnecessary `HostObject` creation elision, please check if it's relevant to the problem or maybe should be deleted or expanded.

### DIFF
--- a/Common/cpp/SharedItems/Shareables.cpp
+++ b/Common/cpp/SharedItems/Shareables.cpp
@@ -66,6 +66,9 @@ jsi::Value makeShareableClone(
         shareable = std::make_shared<ShareableArray>(rt, object.asArray(rt));
       }
     } else if (object.isHostObject(rt)) {
+      if (object.isHostObject<ShareableJSRef>(rt)) {
+        return object;
+      }
       shareable =
           std::make_shared<ShareableHostObject>(rt, object.getHostObject(rt));
     } else {
--- a/Common/cpp/SharedItems/Shareables.cpp
+++ b/Common/cpp/SharedItems/Shareables.cpp
@@ -109,6 +109,8 @@ std::shared_ptr<Shareable> extractShareableOrThrow(
     if (object.isHostObject<ShareableJSRef>(rt)) {
       return object.getHostObject<ShareableJSRef>(rt)->value();
     }
+    throw std::runtime_error(
+        "[Reanimated] Attempted to extract from a HostObject that wasn't converted to a Shareable.");
   } else if (maybeShareableValue.isUndefined()) {
     return Shareable::undefined();
   }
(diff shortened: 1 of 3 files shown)


tomekzaw (approved):
👏

piaskowyk (changes requested):
We need to test it well 🫡
I tested it with custom Host Object and it seems to works without problems.

src/reanimated2/shareables.ts
```diff
@@ -273,9 +273,16 @@ export function makeShareableCloneOnUIRecursive<T>(
       typeof value === 'function'
     ) {
       if (isHostObject(value)) {
-        return value as FlatShareableRef<T>;
+        // We call `_makeShareableClone` because provided HostObject
+        // might not be a Shareable.
+        return _makeShareableClone(
+          _makeShareableClone(value)
```
tomekzaw, line 279: Why do we call `_makeShareableClone` twice?

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -66,6 +66,9 @@ jsi::Value makeShareableClone(
         shareable = std::make_shared<ShareableArray>(rt, object.asArray(rt));
       }
     } else if (object.isHostObject(rt)) {
+      if (object.isHostObject<ShareableJSRef>(rt)) {
+        return object;
+      }
```
tomekzaw, line 71: We should never call `makeShareableClone` with an object that is already an `ShareableJSRef`. This would be an error in our logic. Let's simply fail in such case.
```suggestion
      assert(!object.isHostObject<ShareableJSRef>(rt) && "[Reanimated] Provided value is already an instance of ShareableJSRef.");
```

tomekzaw (changes requested):

src/reanimated2/shareables.ts
tjzel: Yes, I accidentally comitted it 😅

piaskowyk (approved):

src/reanimated2/shareables.ts
```diff
@@ -273,9 +273,14 @@ export function makeShareableCloneOnUIRecursive<T>(
       typeof value === 'function'
     ) {
       if (isHostObject(value)) {
-        return value as FlatShareableRef<T>;
+        // We call `_makeShareableClone` because provided HostObject
+        // might not be a Shareable.
```
tomekzaw, line 277:
```suggestion
        // We call `_makeShareableClone` to wrap the provided HostObject
        // inside ShareableJSRef.
```

tomekzaw (approved):

tomekzaw (approved):

PR #5051, opened 2023-09-06
Finish clamped spring animation

terrysahaidak:
## Summary

Disclaimer:
I'm unsure if this is a bug in the spring implementation or should be addressed elsewhere in the code, but it does happen only with `withSpring`, and kinda randomly, I could not reproduce it with `withTiming`.

While developing the screen transition animation, I noticed that the animation would stop near the end and then we unmount the screen, but it would never just finish completely. After some investigation, realized that the reason for that is the `restDisplacementThreshold` flag in the spring animation. 

When I tried to log the values in both my `useAnimatedStyle` and `springOnFrame` callback, I realized that useAnimatedStyle is never called with the end values.

```
 LOG  0
 LOG  0.054379944363520116
 LOG  0.17171456096392557
 LOG  0.3075683139784404
 LOG  0.4390611486117857
 LOG  0.5557570807744285
 LOG  0.6541357679058646
 LOG  0.7342995250647393
 LOG  0.798059955279234
 LOG  0.8478644812657916
 LOG  0.8862253751352963
 LOG  0.9154425173610802
 LOG  0.93749238762078
 LOG  0.9540066834990276
 LOG  0.966295548106543
 LOG  0.9753897143466889
 LOG  0.982087551556752
 LOG  0.9869998817539096
 LOG  0.990589401501771
 LOG  0.9932037270600896
 LOG  springIsNotInMove current: 0.9951022153102753, toValue: 1
 ```
 
 Notice there is no log after last `springIsNotInMove` - since `useAnimatedStyle` was not called with the ending `toValue = 1`.
 
 This happens since even if we assign `toValue` to the animation, the frame is never called again since we return `true` from the frame and the animation runner code thinks that it's finished and we don't really need to call it again, but that's not true - we do need to call all the mappers again with the resulting values.

### DIFF
--- a/src/reanimated2/animation/spring.ts
+++ b/src/reanimated2/animation/spring.ts
@@ -78,6 +78,10 @@ export const withSpring = ((
     ): boolean {
       const { toValue, startTimestamp, current } = animation;
 
+      if (current === toValue) {
+        return true;
+      }
+
       const timeFromStart = now - startTimestamp;
 
       if (config.useDuration && timeFromStart >= config.duration) {
@@ -85,7 +89,7 @@ export const withSpring = ((
 
         // clear lastTimestamp to avoid using stale value by the next spring animation that starts after this one
         animation.lastTimestamp = 0;
-        return true;
+        return false;
       }
 
       if (config.configIsInvalid) {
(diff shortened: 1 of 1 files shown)


Latropos (approved):

Latropos:
Thank you!

Latropos:
@terrysahaidak Could you provide repro?

terrysahaidak:
@Latropos yeah, i'll try to reproduce it in the example at least a bit later this week.

PR #5053, opened 2023-09-08
Reduce file createAnimatedComponent - part 1

Latropos:
## Summary
File `createAnimatedComponent.tsx` contains our most important logic. It is already very big and since in the future we want to support ${\textcolor{teal}{\textsf{inline Styles}}}$ this file may grow a lot. Thats why we should make it clear and easy to rid - for example by splitting it across many small files.

This refactor will be made in smaller parts.

### Part 1 ${\textcolor{green}{\textsf{(This PR)}}}$ 
* Create folder 📁 `createAnimatedComponent` 
* Move all files used only in `createAnimatedComponent.tsx` into that folder. Move `createAnimatedComponent.tsx` as well.
* Extract all util function into separate file.
### Part 2 ([LINK](https://github.com/software-mansion/react-native-reanimated/pull/5003))
* Create class `inlinePropUpdater.tsx` which handles logic related with inline styles.
### Parts >2
This is not fixed yet, but I suggest we should:
* Move function `_filterNonAnimatedProps` away (70 lines of code)
  * Maybe create a separate class that would remember `initialStyle`? 
* Move away `setLocalRef` (inside `_setComponentRef`).


tjzel (commented):
Looks good! I'll approve once it's not a draft.

tjzel (approved):
🚀

PR #5055, opened 2023-09-08
Use strict everywhere

tjzel:
## Summary

I'm not sure if we need this pull request, but it only took me 
```bash
#!/bin/bash
for file in $(find . -type f \( -name "*.ts" -o -name "*.d.ts" \))
do
  sed -i '' -e $'1i\\\n'"'use strict';"$'\n' "$file"
done
```
to make it.

It would help us in the past when @bartlomiejbloniarz ran into issues when `Object.freeze` froze his object and he couldn't modify it - and without `use strict` this is a silent fail.

## Test plan

Many clicks in Example app.

### DIFF
--- a/src/Animated.ts
+++ b/src/Animated.ts
@@ -1,3 +1,4 @@
+'use strict';
 import type { Extrapolate as _Extrapolate } from './reanimated2/interpolateColor';
 import type { SharedValue as _SharedValue } from './reanimated2/commonTypes';
 import type { DerivedValue as _DerivedValue } from './reanimated2/hook/useDerivedValue';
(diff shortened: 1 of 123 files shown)


tomekzaw (commented):
What about `*.tsx` files? Can we also add a CI workflow to ensure that each `*.(js|ts|tsx)` file has the directive?

reichhartd:
I'm not sure if this solves your problem, but TypeScript has the option [`alwaysStrict`](https://www.typescriptlang.org/tsconfig#alwaysStrict) for this, which ensures that your files are parsed in the ECMAScript strict mode, and emit “use strict” for each source file.

tjzel:
@tomekzaw Yeah I forgot we have `.tsx` files in source. 
@reichhartd Good to know that but it sadly won't cut it since using this option doesn't change behavior of compiled ESModule files.

piaskowyk (approved):

PR #5056, opened 2023-09-08
Properly type event hooks [9/9]

tjzel:
## Summary

After spending a lot of time in these parts of our code I came to a few conclusions in regard to our event hooks.

1. If event hook types are generic, they should be generically based on the type of event they are handling - instead of their payload.
E.g. React Native events are packed into an object called `NativeSyntheticEvent`. This object has `nativeEvent` field where the payload of the event is located. Beside that, some other information regarding the event is outside of that field as well.

Therefore in React Native the user would have to deal with that `NativeSyntheticEvent` and had his code expecting that event's payload is located in the `nativeEvent` field. Then it's only natural for the user to use these types of events as generics for our hooks. Consider the following:

```TSX
// Some component without Reanimated
<CustomComponent
     onSomeEvent={(event: CustomEvent) => { // This event is a NativeSyntheticEvent
       // code
     }}
   />
```

When user decided to makes his component an Animated Component, we can't know in TypeScript context which props are event handlers unless:
- we force the user to respecify his component's props (obviously a bad idea)
- detect if the object property's type is a function that receives a `NativeSyntheticEvent` - doable, but I doubt this method's reliability. 

Therefore, type-wise our event handler has to pretend to get `NativeSyntheticEvent` event type for the types to be correct and for the simplicity of the API event handler has to use this type as a generic. When defining the handler inside the hook (the lambda) there we should use the real event type - so the user can have a proper type. This is why e.g. `useEvent` has (in oversimplification) the following type:

```TSX
function useEvent<Event>(
  handler: (event: ReanimatedEvent<Event>)
 ): (event: Event) => void
```

We satisfy the constraint of the interface (component expects `(event: Event) => void` type handler) and the actual lambda uses the actual type that is emitted by Reanimated (obviously, ReanimatedEvent will be a part of public API).

While this looks good on the outside, it might look bad on the inside - we declare that the return type is not what it is. Since we cannot have both of those things at the same time (or can we?) I'm adding a fallback type that you can cast back to, which is the real type that `useEvent` would return. Therefore with the event hooks we should have three types that represent them:
1. Hook argument handler (worklet) type.
2. Real returned type (for internal purposes).
3. Interface-compliant type (for external purposes).

Using this schema we get:
```TSX
// Argument handler.
type EventHandler<Event> = (event: ReanimatedEvent<Event>) => void;

// Interface compliant type.
type EventHandlerProcessed<Event> = (event: Event) => void;

// Internal type (real return type of useEvent).
// I added `useRef` to `useEvent` to signify that
// the return type can actually be quite complex.
type EventHandlerInternal<Event> = MutableRefObject<Event>;

function useEvent<Event>(
  handler: EventHandler<Event>
): EventHandlerProcessed<Event> {
  const ref = useRef<EventHandler<Event>>(handler);
  return ref.current;
}
```

It works out of the box for the user, and for us only one type cast is required to have everything typed correctly (`useEvent<Event>(...) as EventHandlerInternal<Event>`).
Keep in mind that the names are just a proposal and the schema I presented is to be followed in principle, not literally. We don't always have to lie about the return type and sometimes the concerns raised by me are irrelevant.

Hooks that were changed according to this principle were:
- `useEvent`,
- `useAnimatedGestureHandler` (only in part since it's deprecated anyway - however it was a great playground to understand the situation here),
- `useAnimatedScrollHandler`,
- `useHandler` (partially).

On the other side, `WorkletEventHandler` is a class that is internal. Therefore, it would be much more convenient to use event's payload as it's generic argument - however I tried it and it turned to be not so simple, so instead I decided to get event's payload inside `WorkletEventHandler` - at least it makes it more consistent with the others.

### This overload is required by our API.

You might see this comment in the code alongside with `@ts-expect-error`. This is due to fact that we have to comply to interfaces and because we have Reanimated Babel plugin. When the user uses e.g. `runOnUI` he should provide a worklet as an argument. However, workletization happens during bundling, therefore his TypeScript cannot know that the function he passed will be workletized. Therefore, on the outside `runOnUI` should accept any function and only internally we type it as a worklet. Since this is an illegal overload (arguments for overloaded functions do not match) we have to ignore the error there.

## Test plan

Type tests will come (soonTM).


src/reanimated2/hook/useAnimatedGestureHandler.ts
```diff
@@ -31,78 +17,108 @@ const EventType = {
...
-}
+type StateType = (typeof EVENT_TYPE)[keyof typeof EVENT_TYPE];
 
-export interface GestureHandlerEvent<T> extends __NativeEvent<T> {
-  nativeEvent: T;
+interface PropsUsedInUseAnimatedGestureHandler {
+  handlerTag?: number;
+  numberOfPointers?: number;
+  state?: StateType;
+  oldState?: StateType;
```
Latropos, line 26: Should this be 
```suggestion
  previousState?: StateType;
```

src/reanimated2/hook/useAnimatedGestureHandler.ts
```diff
@@ -31,78 +17,108 @@ const EventType = {
...
-interface GestureHandlerNativeEvent {
-  handlerTag: number;
-  numberOfPointers: number;
-  state: (typeof EventType)[keyof typeof EventType];
-}
+type StateType = (typeof EVENT_TYPE)[keyof typeof EVENT_TYPE];
 
-export interface GestureHandlerEvent<T> extends __NativeEvent<T> {
-  nativeEvent: T;
+interface PropsUsedInUseAnimatedGestureHandler {
```
Latropos, line 22: Could we have something more concise here?

src/reanimated2/hook/useAnimatedGestureHandler.ts
```diff
@@ -31,78 +17,108 @@ const EventType = {
...
-    // This will be removed in the next PR in the series.
-    any,
-    TContext
-  >(handlers, dependencies);
+  Event extends NativeEventWrapper<PropsUsedInUseAnimatedGestureHandler> = DefaultEvent,
+  Context extends Record<string, unknown> = Record<string, unknown>
+>(handlers: GestureHandlers<Event, Context>, dependencies?: DependencyList) {
+  // type RealWebEvent = WrappedNativeEvent<GestureHandlersPayload<Event>>;
+  // type RealNativeEvent = ReanimatedEvent<GestureHandlersPayload<Event>>;
+  // type RealEvent = RealWebEvent | RealNativeEvent;
```
Latropos, line 67: Is this to be 🧹?

src/reanimated2/hook/useAnimatedGestureHandler.ts
Latropos: I see we use oldState a lot in our app. Is this previous state or is it even older?

tjzel: This is coming from `react-native-gesture-handler` https://github.com/software-mansion/react-native-gesture-handler/blob/main/src/handlers/gestureHandlerCommon.ts#L51, therefore we shouldn't rename it. `previousState` is also something, but it's not the case here (it's something different). Also keep in mind that `useAnimatedGestureHandler` is deprecated API and I we don't really want to bother that much about having it pretty.

src/reanimated2/hook/useAnimatedGestureHandler.ts
tjzel: Yes, I forgor

src/reanimated2/hook/useAnimatedGestureHandler.ts
tjzel: I could do `GestureHandlerEventPropsUsedInUseAnimatedGestureHandler` 🤔

tjzel: I don't want to add reliance on RNGH types here since it's deprecated API and we don't want to worry about this part of code breaking.

src/reanimated2/hook/useAnimatedGestureHandler.ts
```diff
@@ -1,28 +1,14 @@
...
-} from '../commonTypes';
-import type { DependencyList } from './commonTypes';
-import { useEvent } from './useEvent';
+  DependencyList,
+  NativeEventWrapper,
+  ReanimatedEvent,
+} from './commonTypes';
 import { useHandler } from './useHandler';
+import { useEvent } from './useEvent';
+import type { PanGestureHandlerGestureEvent as DefaultEvent } from 'react-native-gesture-handler';
```
piaskowyk, line 9: We should remove dependency to gesture-handler if it is possible

piaskowyk (changes requested):

src/reanimated2/hook/useAnimatedGestureHandler.ts
tjzel: ready

piaskowyk (approved):

PR #5060, opened 2023-09-11
Fix `measure` crash on web

m-bert:
## Summary

While going through our example app I stumbled upon crash in "Article progress" example. Turns out that sometimes we try to call `getBoundingClientRect` on element that has not been rendered yet. This PR fixes this problem.

## Test plan

Tested on "Article progress" example.


tomekzaw (approved):
Thanks for fixing this!

PR #5071, opened 2023-09-12
Document storing objects in shared values

kacperkapusciak:
Added note about handling objects in shared values

Thanks to @cdehaan for pointing this out! 🎖️ 

Supersedes https://github.com/software-mansion/react-native-reanimated/pull/5037


tjzel (approved):
⚡

Issue #5067, opened 2023-09-12
Crashing in Production Mode When Trying to Open the Drawer


aslihanturkdonmez:
### Description

I'm experiencing a crash when attempting to open the drawer in a React Native project using Reanimated v3. Debug mode is functioning correctly, but in production, the app crashes on both Android and iOS. I've created a new React Native project (0.72.4) and installed the necessary navigation and drawer packages.

![reanimated_crash](https://github.com/software-mansion/react-native-reanimated/assets/43846857/6e8810a1-7a34-457e-93e7-a0c204c9679a)

### Steps to reproduce

I've installed the plugin's latest version and am currently working on a project built with the most recent React Native version, but I'm encountering crashes.

### Snack or a link to a repository

https://github.com/aslihanturkdonmez/test

### Reanimated version

3.5.1

### React Native version

0.72.4

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

iOS simulator

### Device model

iPhone 14 pro (16.2)

### Acknowledgements

Yes

aurotones:
I'm having the same problem with crash message "maximum call stack size exceeded".
Problem goes away by rolling back to 3.4.2.

n-ii-ma:
Same issue here on Android:

```
com.swmansion.reanimated.AndroidUIScheduler.triggerUI SourceFile
com.swmansion.reanimated.AndroidUIScheduler$1.run
com.swmansion.reanimated.AndroidUIScheduler$2.runGuarded
com.facebook.react.bridge.GuardedRunnable.run
android.os.Handler.handleCallback Handler.java:942
android.os.Handler.dispatchMessage Handler.java:99
android.os.Looper.loopOnce Looper.java:226
android.os.Looper.loop Looper.java:313
android.app.ActivityThread.main ActivityThread.java:8762
java.lang.reflect.Method.invoke Method.java
com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run RuntimeInit.java:604
com.android.internal.os.ZygoteInit.main ZygoteInit.java:1067
```

Reverting back to v3.4.2 fixes the problem.

ranaavneet:
+1

efstathiosntonas:
same here, only on release scheme, debug works fine.

souna23:
the same

aslihanturkdonmez:
[...]

PR #5065, opened 2023-09-11
fix ReferenceError: Property '_IS_FABRIC' doesn't exist

efstathiosntonas:
fixes https://github.com/software-mansion/react-native-reanimated/issues/5061

PR #5076, opened 2023-09-13
Fix bug on release

Latropos:
## Summary
Fix to https://github.com/software-mansion/react-native-reanimated/issues/5067

This fix makes reanimated work in release build

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -104,11 +104,11 @@ body:
     id: workflow
     attributes:
       label: Workflow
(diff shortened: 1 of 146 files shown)


ranaavneet:
@Latropos this check is already present in 3.5.1 but it still crashes

https://github.com/software-mansion/react-native-reanimated/blob/3.5.1/src/reanimated2/NativeMethods.ts#L65

tjzel (approved):
👍

tjzel:
Was already done in #5065

ranaavneet:
@tjzel @Latropos is there a separate PR to fix #5067 ?

PR #5077, opened 2023-09-13
Properly type event hooks [0/9]

tjzel:
_This PR is made only to split #5056 to separate PRs for review comfort._

Hooks were cramped together for a long time. Let's allow them to be in separate files. This PR only does the split.


src/reanimated2/hook/useScrollViewOffset.ts
```diff
@@ -6,8 +6,8 @@ import type Animated from '../../index'; // TODO: fixme?
 import type { ScrollEvent } from './useAnimatedScrollHandler';
 import type { SharedValue } from '../commonTypes';
 import { findNodeHandle } from 'react-native';
-import { useEvent } from './utils';
 import { useSharedValue } from './useSharedValue';
+import { useEvent } from './useEvent';
```
Latropos, line 10: Please undo this to get smaller diff. It may be "nothing" but it simplifies usage of some git tools

Latropos (approved):

PR #5081, opened 2023-09-14
Properly type event hooks [4/9]

tjzel:
_This PR is made only to split #5056 to separate PRs for review comfort._

Making `WorkletEventHandler` use proper types.


src/reanimated2/WorkletEventHandler.ts
```diff
@@ -1,25 +1,39 @@
...
 import { registerEventHandler, unregisterEventHandler } from './core';
+import type { EventPayload, ReanimatedEvent } from './hook/commonTypes';
 
-function jsListener<T extends __NativeEvent<T>>(
+type AssertNativeEvent<Event extends object> = NativeSyntheticEvent<
+  EventPayload<Event>
+>;
+
+// In JS implementation (e.g. for web) we don't use Reanimated's
+// event emitter therefore we have to handle here
```
Latropos, line 12: Comma before therefore?

src/reanimated2/WorkletEventHandler.ts
```diff
@@ -1,25 +1,39 @@
 'use strict';
-import type { __NativeEvent } from './commonTypes';
+import type { NativeSyntheticEvent } from 'react-native';
 import NativeReanimatedModule from './NativeReanimated';
 import { registerEventHandler, unregisterEventHandler } from './core';
+import type { EventPayload, ReanimatedEvent } from './hook/commonTypes';
 
-function jsListener<T extends __NativeEvent<T>>(
+type AssertNativeEvent<Event extends object> = NativeSyntheticEvent<
```
Latropos, line 7: I would recommend some name without "assert", because it reminds me this usage of generics:
```
type AssertNumber<T> = T extends number ? T : never;

const a = 42;
const b = 'Hello';

const assertedA: AssertNumber<typeof a> = a; //OK
const assertedB: AssertNumber<typeof b> = b; //ERROR
```

tjzel: What do you suggest? That's a tough nut to crack.

Latropos (approved):

PR #5089, opened 2023-09-14
Bump react-native-macos version

jfedak:
## Summary

This PR bumps `react-native-macos` version from 0.71 to 0.72. It also features some minor changes that unifies `MacOSExample` and `Example` apps.

## Test plan

Tested with MacOSExample app.


package.json
```diff
@@ -155,7 +155,7 @@
     "react-native": "0.72.3",
     "react-native-builder-bob": "^0.18.3",
     "react-native-gesture-handler": "^2.12.0",
-    "react-native-macos": "^0.71.0-0",
+    "react-native-macos": "^0.72.0",
```
tjzel, line 158: Is this dependency actually needed in the root of our repo?

MacOSExample/metro.config.js
```diff
@@ -25,13 +33,6 @@ module.exports = {
       return acc;
     }, {}),
   },
-
-  transformer: {
```
tjzel, line 29: Why do we drop this setting? Is it included in the default config?

tjzel (changes requested):
Looks good to merge, but please answer my questions first 😅

MacOSExample/metro.config.js
jfedak: Exactly, it is present in [default config](https://github.com/facebook/react-native/blob/5323221d1442d1573bc65daff618478cb6f056f0/packages/metro-config/index.js)

package.json
jfedak: It turned out it is not. I'm about to push appropriate changes.

tjzel (approved):
Me likes

piaskowyk (approved):
👏

PR #5090, opened 2023-09-14
Split `decay` into multiple files

Latropos:
## Summary
This is only splitting a single file int multiple ones. This PR should make this [type improvements](https://github.com/software-mansion/react-native-reanimated/pull/5088) more readable.


src/reanimated2/animation/decay/index.ts
```diff
@@ -0,0 +1,3 @@
+export { withDecay } from './decay';
```
tjzel, line 1: 'use strict';
```suggestion
export { withDecay } from './decay';
```

src/reanimated2/animation/decay/rigidDecay.ts
```diff
@@ -0,0 +1,33 @@
+import type { DefaultDecayConfig, InnerDecayAnimation } from './utils';
```
tjzel, line 1:
```suggestion
'use strict';
import type { DefaultDecayConfig, InnerDecayAnimation } from './utils';
```

src/reanimated2/animation/decay/rubberBandDecay.ts
```diff
@@ -0,0 +1,44 @@
+import { SLOPE_FACTOR, VELOCITY_EPS } from './utils';
```
tjzel, line 1: 'use strict';
```suggestion
import { SLOPE_FACTOR, VELOCITY_EPS } from './utils';
```

src/reanimated2/animation/decay/utils.ts
```diff
@@ -0,0 +1,36 @@
+import type { AnimationObject, AnimatableValue } from '../../../index.web';
```
tjzel, line 1:
```suggestion
'use strict';
import type { AnimationObject, AnimatableValue } from '../../../index.web';
```

tjzel (changes requested):
after adding 'use strict's we are good to go

tjzel (approved):

PR #5074, opened 2023-09-12
Update type exports to fix build warnings

khevamann:
## Summary

This PR addresses an issue where building for web produces a ton of warnings about missing imports. closes #5073 

## Test plan

N/A

PR #5105, opened 2023-09-20
Fix exporting enums as types and type of global in PlatformChecker

tjzel:
## Summary

Fixes regression that occurred in #5074.

Later on we also have to rethink if we want to promote `Extrapolate` or `Extrapolation` since both are exactly the same and we only use the former.

## Test plan

I actually made `Extrapolation` used in our examples so such mistake in the future would be caught on CI (only regarding `Extrapolation` sadly).

> [!Note]
> I recommend using this
> ```json
> "editor.tokenColorCustomizations": {
>    "textMateRules": [
>      {
>        "scope": "entity.name.type.enum",
>        "settings": {
>          "foreground": "#a3e577"
>        }
>      },
>      {
>        "scope": "entity.name.type.class",
>        "settings": {
>          "foreground": "#e8b55e"
>        }
>      }
>    ]
>  }
>```
> In your VSCode settings so Classes, Enums and Types are not in the same color.

### DIFF
--- a/app/src/examples/DragAndSnapExample.tsx
+++ b/app/src/examples/DragAndSnapExample.tsx
@@ -6,7 +6,7 @@ import Animated, {
   useAnimatedStyle,
   useAnimatedGestureHandler,
   interpolate,
-  Extrapolate,
+  Extrapolation,
 } from 'react-native-reanimated';
 import {
   PanGestureHandler,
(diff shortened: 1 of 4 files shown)


bartlomiejbloniarz (approved):
🚀

piaskowyk (approved):

PR #5108, opened 2023-09-21
Add example for Next.js

bartlomiejbloniarz:
## Summary

This PR adds an example with Next.js and a CI to test build fails. It is intended to prevent using api that is only available in a browser (w.g. window) without an appropriate check. This ensures that reanimated can be used with SSG.

## Test plan

Build the example with `yarn build` and start with `yarn start`. You can also start it in dev mode using `yarn dev` and check if it responds changes in reanimated core.


tjzel (changes requested):
Requesting changes in blanco

nandorojo:
FYI – the optimal setup for Next.js / web to confirm tree shaking would be to get rid of `shouldBeUseWeb()` which is used in many places in favor of platform-based files (`index.web.ts`). This would ensure what we aren't importing any unused variables or using conditional declarations for things in the wrong places. It would result in more files but be far more resilient and less prone to tree shaking regressions.

bartlomiejbloniarz:
@nandorojo I am currently working on that in #5203

nandorojo:
Awesome, glad to see this getting attention.

Just left some comments.

PR #5113, opened 2023-09-21
Replaced DEBUG macro with NDEBUG

michalmaka:
## Summary

During working on integration of some other library @t0ms0n00 and I found out that our definition od `DEBUG` preprocessor macro was causing compilation error when we compile 3rd pary library. Such definition should not be used and better way is to use standard way of `NDEBUG` definition which is defined in release builds by default.

The problem was due to 3rd party having an enum with one entry named `DEBUG` which was replaced by preprocessor macro to `1` and caused compilation error.

We should use `NDEBUG` or name our preprocessor macro differently like `REANIMATED_DEBUG`.

`NDEBUG` is mentioned several times in C++ standard https://isocpp.org/files/papers/N4860.pdf

## Test plan

Check if compilation is successful


tomekzaw (approved):

tomekzaw:
Added CI check that disallows `#if DEBUG` and `#ifdef DEBUG` in `common/Cpp`, `android/src/main/cpp` and `apple/`. Should we also forbid `#elif DEBUG`? 🤔

piaskowyk (approved):

Issue #5112, opened 2023-09-21
error: node_modules/react-native-reanimated/src/reanimated2/threads.ts: /Users/zhuxiuwu/space/native/nativeapp/node_modules/react-native-reanimated/src/reanimated2/threads.ts: [Reanimated] Babel plugin exception: SyntaxError: unknown: Unexpected token


boonook:
### Description

error: node_modules/react-native-reanimated/src/reanimated2/threads.ts: /Users/zhuxiuwu/space/native/nativeapp/node_modules/react-native-reanimated/src/reanimated2/threads.ts: [Reanimated] Babel plugin exception: SyntaxError: unknown: Unexpected token
<img width="1115" alt="image" src="https://github.com/software-mansion/react-native-reanimated/assets/34494413/53825641-52eb-4ea2-a154-1ba799dc4ab3">

### Steps to reproduce

System:
  OS: macOS 13.1
  CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
  Memory: 46.64 MB / 8.00 GB
  Shell:
    version: 3.2.57
    path: /bin/bash
Binaries:
  Node:
    version: 16.14.0
    path: ~/.nvm/versions/node/v16.14.0/bin/node
  Yarn:
    version: 1.22.17
    path: /usr/local/bin/yarn
  npm:
    version: 8.3.1
    path: ~/.nvm/versions/node/v16.14.0/bin/npm
  Watchman:
    version: 2022.10.31.00
    path: /usr/local/bin/watchman
Managers:
  CocoaPods:
    version: 1.12.1
    path: /usr/local/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 22.4
      - iOS 16.4
      - macOS 13.3
      - tvOS 16.4
      - watchOS 9.4
  Android SDK:
    API Levels:
      - "21"
      - "22"
      - "23"
      - "24"
      - "25"
      - "26"
      - "27"
      - "28"
      - "29"
      - "30"
      - "31"
      - "32"
      - "33"
    Build Tools:
      - 23.0.1
      - 23.0.2
      - 23.0.3
      - 25.0.0
      - 25.0.1
      - 25.0.2
      - 25.0.3
      - 26.0.0
      - 26.0.1
      - 26.0.3
      - 27.0.3
      - 28.0.2
      - 28.0.3
      - 29.0.0
      - 29.0.2
      - 30.0.2
      - 30.0.3
[...]

PR #5119, opened 2023-09-22
fix syntax error

lafest:
## Summary

스크린샷 2023-09-22 14 46 10

I met this error in v3.5.4, so fixed syntax error to work well.

### DIFF
--- a/src/reanimated2/threads.ts
+++ b/src/reanimated2/threads.ts
@@ -212,7 +212,7 @@ export function runOnJS<Args extends unknown[], ReturnValue>(
     // and pass the worklet as a first argument followed by original arguments.
 
     return (...args) =>
-      runOnJS(runWorkletOnJS<Args, ReturnValue>)(
+      runOnJS(runWorkletOnJS as typeof runWorkletOnJS<Args, ReturnValue>)(
         fun as WorkletFunction<Args, ReturnValue>,
         ...args
       );


tjzel:
Fix necessary due to user's corrupted `yarn.lock` (#5112 for more context).

lafest:
@tjzel I tried following steps, but yarn did not resolved this issue.
```sh
rm -rf node_modules yarn.lock
yarn cache clean
yarn install
```

Is there any missing steps in my script?

lafest:
How [this line](https://github.com/software-mansion/react-native-reanimated/blob/main/src/reanimated2/threads.ts#L215) in main branch is resolved automatically by yarn?

I think this type error cannot be fixed by yarn.

tjzel:
This is not a type error - `yarn.lock`s tend to get corrupted and then fail to resolve some dependencies properly. Maybe you also need to reset your bundler's cache or something. Try `yarn react-native bundle --reset-cache --platform ios --dev true --entry-file index.js --bundle-output bundle.js`.

lafest:
```sh
rm -rf node_modules yarn.lock
yarn cache clean
yarn install
yarn react-native bundle --reset-cache --platform ios --dev true --entry-file index.js --bundle-output bundle.js
```

this worked well for me. @tjzel thanks for help!
but.. it seems risky that re-install all the yarn.lock file because other unrelative libraries also updated. so, if I find proper dependencies for reanimated, I'll post it.

PR #5124, opened 2023-09-22
Add e2e tests in NextExample

bartlomiejbloniarz:
## Summary

This PR adds e2e tests for web. Tests are implemented with cypress.

## Test plan
Check if the test passes, then change the target width in `test.js` and check that the test doesn't pass. Execute `yarn build` to build `NextExample` and then run `yarn e2e:headless` to start tests.


piaskowyk (approved):
🚀

PR #5133, opened 2023-09-25
Add default initial value to Participant List example

m-bert:
## Summary

While doing tests on this example I've found it really frustrating that input doesn't have default value, so I've decided to add it.

## Test plan

Tested on Participant List example.


piaskowyk (approved):
inblanco 🫡

PR #5134, opened 2023-09-25
Get rid of native=true from ReanimatedModule

Latropos:
## Test plan
Let's notce that `NativeReanimatedModule.native` is equal to `!shouldBeUseWeb()`. We can simplify our code a little substituting it.


src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -154,9 +154,9 @@ export class InlinePropManager {
       const shareableViewDescriptors =
         this._inlinePropsViewDescriptors.shareableViewDescriptors;
 
-      const maybeViewRef = NativeReanimatedModule.native
-        ? undefined
-        : ({ items: new Set([this]) } as ViewRefSet<any>); // see makeViewsRefSet
+      const maybeViewRef = shouldBeUseWeb()
```
tjzel, line 157: No need to evaluate it here each time. Instead evaluate `IS_NATIVE` in top level of this file.

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -453,7 +452,7 @@ For more, see the docs: \`https://docs.swmansion.com/react-native-reanimated/doc
 
   const { initial, remoteState, viewDescriptors } = initRef.current;
   const shareableViewDescriptors = viewDescriptors.shareableViewDescriptors;
-  const maybeViewRef = NativeReanimatedModule.native ? undefined : viewsRef;
+  const maybeViewRef = shouldBeUseWeb() ? viewsRef : undefined;
```
tjzel, line 455: Also evaluate it in the top level of the file.

src/reanimated2/WorkletEventHandler.ts
```diff
@@ -39,7 +39,7 @@ export default class WorkletEventHandler<Event extends object> {
     this.viewTag = undefined;
     this.registrations = [];
 
-    if (!NativeReanimatedModule.native) {
+    if (shouldBeUseWeb()) {
```
tjzel, line 42: top level

src/reanimated2/mutables.ts
```diff
@@ -64,7 +65,7 @@ export function makeMutable<T>(
 ): SharedValue<T> {
   let value: T = initial;
   let syncDataHolder: ShareableSyncDataHolderRef<T> | undefined;
-  if (!oneWayReadsOnly && NativeReanimatedModule.native) {
+  if (!oneWayReadsOnly && !shouldBeUseWeb()) {
```
tjzel, line 68: Definitely top level here.

tjzel (changes requested):

tjzel (approved):
Looks good, but make sure to check if it works on Web and simulators

piaskowyk (changes requested):
Everything look great, but please confirm that it works on Web and MacOS too

piaskowyk (approved):

PR #5135, opened 2023-09-25
Fix types in `createAnimatedComponent`

Latropos:
## Summary

I've been recently refactoring `createAnimatedComponent`, but I've forgot to update types.
* Replace any with unknown
* Type viewConfig as ViewConfig
* Add 'use strict'


src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -35,14 +45,20 @@ function inlinePropsHasChanged(
...
+) {
   'worklet';
-  const update: Record<string, any> = {};
+  const update: Record<string, unknown> = {};
   for (const [key, styleValue] of Object.entries(inlineProps)) {
     if (key === 'transform') {
-      update[key] = styleValue.map((transform: Record<string, any>) => {
-        return getInlinePropsUpdate(transform);
-      });
+      update[key] = inlineProps[key]?.map(
```
tjzel, line 57: This change potentially changes the behavior. Please try not to change how is JS executed on type fixing pull request. Also, why we stop using `styleValue` here?

src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -35,14 +45,20 @@ function inlinePropsHasChanged(
   return false;
 }
 
-function getInlinePropsUpdate(inlineProps: Record<string, any>) {
+function getInlinePropsUpdate(
+  inlineProps: Record<string, unknown> &
+    TransformsStyle &
+    Partial<TransformArrayItem>
```
tjzel, line 51: How can `inlineProps` be of type `Partial<TransformArrayItem>`?

tjzel (changes requested):

src/createAnimatedComponent/InlinePropManager.ts
Latropos: We have `const [key, styleValue] of Object.entries(inlineProps)` so `styleValue` is equal to `inlineProps[key]`. However due to typescript limitations inside conditional
`if (key === 'transform')` only type of `key` is narrowed.

Latropos: Previously we were casting `styleValue` to `any` to use it like an array. But when we substitute it with `inlineProps[key]` then it is automatically typed as `TransformsStyle`, so we TS knows it is indeed an array.

src/createAnimatedComponent/InlinePropManager.ts
Latropos: This function is recursive and it can be so only in second iteration, it happens when you use ` return getInlinePropsUpdate(transform);`

src/createAnimatedComponent/InlinePropManager.ts
tjzel: I meant the `?` operator. Before we just called `map`, so calling improperly would throw an error. Now we check if it's available, altering the behavior.

src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -35,12 +45,14 @@ function inlinePropsHasChanged(
...
+function getInlinePropsUpdate(
+  inlineProps: (Record<string, unknown> & TransformsStyle) | TransformArrayItem
+) {
   'worklet';
-  const update: Record<string, any> = {};
+  const update: Record<string, unknown> = {};
   for (const [key, styleValue] of Object.entries(inlineProps)) {
-    if (key === 'transform') {
-      update[key] = styleValue.map((transform: Record<string, any>) => {
+    if (key === 'transform' && 'transform' in inlineProps) {
```
tjzel, line 54: This is also altering the behavior now.

tjzel (approved):

PR #5136, opened 2023-09-25
Prevent top-level calls to `requestAnimationFrame`

bartlomiejbloniarz:
## Summary

Currently there are instances where `scheduleOnUiImmediately` is called in the top level to create global structures which are not needed on web. This results in a call to `requestAnimationFrame`, which is unavailable during static website generation. We already have a mock for `raf` which prevents the error, but in this PR I changed the logic so the unnecessary calls to `raf` don't happen on web.


src/reanimated2/layoutReanimation/sharedTransitions/ProgressTransitionManager.ts
```diff
@@ -203,10 +204,25 @@ function createProgressTransitionRegister() {
...
 }
 
-runOnUIImmediately(() => {
-  'worklet';
-  global.ProgressTransitionRegister = createProgressTransitionRegister();
-})();
+if (shouldBeUseWeb()) {
+  const throwError = () => {
+    throw new Error(
+      '[Reanimated] ProgressTransitionRegister should not be used on web.'
```
tjzel, line 210:
```suggestion
      '[Reanimated] `ProgressTransitionRegister` is not available on non-native platform.'
```

tjzel (changes requested):

src/reanimated2/UpdateProps.ts
```diff
@@ -117,14 +117,19 @@ const createUpdatePropsManager = global._IS_FABRIC
     };
 
 if (shouldBeUseWeb()) {
-  const throwError = () => {
-    throw new Error(
-      '[Reanimated] UpdatePropsManager should not be used on web.'
-    );
+  const maybeThrowError = () => {
+    if (!isJest()) {
```
tjzel, line 121: Please add a comment about why do we have to if `jest` out here.

src/reanimated2/layoutReanimation/sharedTransitions/ProgressTransitionManager.ts
```diff
@@ -205,16 +205,21 @@ function createProgressTransitionRegister() {
 }
 
 if (shouldBeUseWeb()) {
-  const throwError = () => {
-    throw new Error(
-      '[Reanimated] ProgressTransitionRegister should not be used on web.'
-    );
+  const maybeThrowError = () => {
+    if (!isJest()) {
```
tjzel, line 209: Please add a comment about why do we have to if `jest` out here.

tjzel (changes requested):

tjzel (approved):

tjzel:
Please double check if web behavior is still proper.

piaskowyk (approved):

PR #5137, opened 2023-09-25
Remove some occurrences of `any` type

Latropos:
https://github.com/software-mansion/react-native-reanimated/pull/4421/files

## Summary
Remove some unnecessary casts to `any` and provide valid types for `yogaValues`
We have found out that:
* most of keys in `yogaValues` are optional (only `windowWidth` and `windowHeight` are required
* `yogaValues` may have more keys (`currentTransformMatrix`, `targetTransformMatrix`)

So I've changed it type as follows
```diff
-   yogaValues: LayoutAnimationsValues,
+   yogaValues: Partial<SharedTransitionAnimationsValues>,
```

## Test plan
I've logged a lot of `yogaValues` to compare its keys so I'm pretty sure the new type is valid.


tjzel (changes requested):

piaskowyk (approved):

src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
```diff
@@ -79,7 +79,7 @@ export type LayoutAnimationFunction = (
 export type LayoutAnimationStartFunction = (
   tag: number,
   type: LayoutAnimationType,
-  yogaValues: LayoutAnimationsValues,
+  yogaValues: Partial<SharedTransitionAnimationsValues>,
```
tjzel, line 82: Why did these become `SharedTransitionAnimationsValues`?

src/reanimated2/layoutReanimation/animationsManager.ts
```diff
@@ -43,8 +43,10 @@ function createLayoutAnimationManager() {
     start(
       tag: number,
       type: LayoutAnimationType,
-      yogaValues: LayoutAnimationsValues,
-      config: LayoutAnimationFunction
+      yogaValues: Partial<SharedTransitionAnimationsValues>,
```
tjzel, line 46: Why did these become `SharedTransitionAnimationsValues`?

tjzel (changes requested):

src/reanimated2/layoutReanimation/animationBuilder/commonTypes.ts
Latropos: Well, the answer is:
1. We don't use type `LayoutAnimationStartFunction` at all, we just export it
2. This type is equal to input arguments of function ` createLayoutAnimationManager`, which works with both layout and shared transition tag animations. So once I've updated `createLayoutAnimationManager` I've updated type `LayoutAnimationStartFunction` to keep then consistent.

tjzel (approved):
Improve the comment and we gucci

src/reanimated2/layoutReanimation/animationsManager.ts
```diff
@@ -34,6 +34,7 @@ function stopObservingProgress(
   _notifyAboutEnd(tag, removeView);
 }
 
+// This function works with both layout and shared transition tag animations
```
tjzel, line 37:
```suggestion
/**
* Creates an animation manager for both Layout animations and Shared Transition Elements animations.
*/
```

tjzel (approved):

PR #5145, opened 2023-09-26
Reorder checks in `getInlinePropsUpdate` and remove `transform`-specific syntax

tjzel:
This PR makes the behavior of `useAnimatedStyle` and InlineStyles API consistent. Because of wrong order of checks in `getInlinePropsUpdate` potential Shared Values were treated as arrays and threw errors - this didn't happen for the same usage of Shared Value with `useAnimatedStyle`.

```TSX
  const sv = useSharedValue([{ translateX: 50 }, { translateY: 50 }]);
  const handlePress = () => {
    sv.value = [{ translateX: 100 }, { translateY: 100 }];
  };
    return {
      transform: sv.value,
    };
  });
  return (
      <Button title="Press me" onPress={handlePress} />
      {/* This works. */}
      {/* This doesn't. */}
      <Animated.View style={[styles.box, { transform: sv }]} />
    </View>
  );
}
```

## Test plan

With these changes this snippet works:

<details>
<summary>
code
</summary>

```TSX
import Animated, { useSharedValue } from 'react-native-reanimated';

import React from 'react';

function TransformSharedValue() {
  const sv = useSharedValue([{ translateX: 100 }, { rotate: '30deg' }]);
  function handlePress() {
    sv.value = [{ translateX: 200 }, { rotate: '60deg' }];
  }
  return (
    <>
      <Animated.View style={[styles.box, { transform: sv }]} />
      <Button title="Press me" onPress={handlePress} />
    </>
  );
}

function TextShadowOffsetSharedValue() {
  const sv = useSharedValue({ width: 5, height: 5 });
  function handlePress() {
    sv.value = { width: 10, height: 10 };
  }
  return (
    <>
      <Animated.Text
        ref={null}
        style={[
          styles.text,
          { textShadowColor: 'black', textShadowRadius: 10 },
          { textShadowOffset: sv },
        ]}>
        Serious text
      </Animated.Text>
      <Button title="Press me" onPress={handlePress} />
    </>
  );
}

  return (
      <TransformSharedValue />
      <TextShadowOffsetSharedValue />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: {
    width: 100,
    height: 100,
    backgroundColor: 'red',
  },
  text: {
    width: 100,
    height: 100,
  },
});
```
</details>

### DIFF
--- a/src/ConfigHelper.ts
+++ b/src/ConfigHelper.ts
@@ -83,6 +83,7 @@ let NATIVE_THREAD_PROPS_WHITELIST: Record<string, boolean> = {
   fontSize: true,
   lineHeight: true,
   textShadowRadius: true,
+  textShadowOffset: true,
   letterSpacing: true,
   aspectRatio: true,
   columnGap: true, // iOS only
--- a/src/createAnimatedComponent/InlinePropManager.ts
+++ b/src/createAnimatedComponent/InlinePropManager.ts
@@ -35,16 +35,18 @@ function inlinePropsHasChanged(
   return false;
 }
 
-function getInlinePropsUpdate(inlineProps: Record<string, any>) {
+function getInlinePropsUpdate(inlineProps: Record<string, unknown>) {
   'worklet';
-  const update: Record<string, any> = {};
+  const update: Record<string, unknown> = {};
   for (const [key, styleValue] of Object.entries(inlineProps)) {
-    if (key === 'transform') {
-      update[key] = styleValue.map((transform: Record<string, any>) => {
-        return getInlinePropsUpdate(transform);
-      });
-    } else if (isSharedValue(styleValue)) {
+    if (isSharedValue(styleValue)) {
       update[key] = styleValue.value;
+    } else if (Array.isArray(styleValue)) {
+      update[key] = styleValue.map((item) => {
+        return getInlinePropsUpdate(item);
+      });
+    } else if (typeof styleValue === 'object') {
+      update[key] = getInlinePropsUpdate(styleValue as Record<string, unknown>);
     } else {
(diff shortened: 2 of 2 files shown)


src/ConfigHelper.ts
```diff
@@ -83,6 +83,7 @@ let NATIVE_THREAD_PROPS_WHITELIST: Record<string, boolean> = {
   fontSize: true,
   lineHeight: true,
   textShadowRadius: true,
+  textShadowOffset: true,
```
Latropos, line 86: I know its not related to you PR directly, but
1️⃣ Do you think we should have some checks to verify that there are no common properties in UI_THREAD_PROPS_WHITELIST and NATIVE_THREAD_PROPS_WHITELIST?
2️⃣ Do you think we should update type `Record<string, boolean>` and replace `string` with `keyof ViewStyle | TextStyle | ...` and `boolean` with `true`?

Latropos:
I've tested, everything works. Also now you can pass an array to `fontVariant` in inline styles and it works although it haven't been working before 👏

Latropos (approved):

piaskowyk (approved):
LGTM, but update a bit PR description

src/ConfigHelper.ts
tjzel: 🥇 That's a good idea, we can add a check in `configureProps` function.
🥈 While using `boolean: true` seems a good idea, the former proposition doesn't sound good. If we do a dependency on `ViewStyle | TextStyle | ...` we risk that types for this part of code will break depending on changes in React Native - some props might be gone, or added.

PR #5146, opened 2023-09-26
Add delay to duration when clearing animations

m-bert:
## Summary

Code that cleans animations on web version doesn't take into account delay value. Because of that, animations with delay do not finish properly.

## Test plan

Check that animations with delay finish properly on example app.


piaskowyk (approved):
🫡

PR #5147, opened 2023-09-26
Refactor `animationsManager` on web

m-bert:
## Summary

`animationsManager.ts` file grew pretty big during development of layout animations. This PR aims to split it into multiple files to separate the logic of DOM manipulation from the animations logic.
## Test plan

Check that WebExample app works as it used to.


src/reanimated2/layoutReanimation/web/elementManager.ts
```diff
@@ -0,0 +1,297 @@
+import type { TransformsStyle } from 'react-native';
```
piaskowyk, line 1: What is the different between DOMManager and elementManager?

src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -39,12 +39,11 @@ import setAndForwardRef from './setAndForwardRef';
...
 import { InlinePropManager } from './InlinePropManager';
 import { PropsFilter } from './PropsFilter';
-
 import {
   startWebLayoutAnimation,
-  configureWebLayoutAnimations,
   tryActivateLayoutTransition,
-} from '../reanimated2/layoutReanimation/web/animationsManager';
+} from '../reanimated2/layoutReanimation/web/elementManager';
+import { configureWebLayoutAnimations } from '../reanimated2/layoutReanimation/web/DOMManager';
```
piaskowyk, line 46: I think that WebAnimaitonManager should contains whole public interface available to consume by other modules of Reaniamted.

src/reanimated2/layoutReanimation/web/elementManager.ts
```diff
@@ -0,0 +1,297 @@
...
+import { Animations, TransitionType } from '.';
```
piaskowyk, line 2: For me `Manager` suggest that this is a class with internal state and expose public API to deal with this state. Those files are more like a set of functions. - This is just my subject feel. I'm open to discussion.

piaskowyk (changes requested):
Although there has been some improvement, I still find it confusing to determine the responsibilities of specific modules/files. 🤔 In order to enhance the maintainability of this feature in the future, it is crucial that we organize it properly. I understand that I haven't provided clear instructions on how to improve it, but we can discuss it in person, possibly tomorrow. However, as a starting point, it would be good to define the public interface in the AnimationManager.

src/reanimated2/layoutReanimation/web/elementManager.ts
m-bert: I agree. I've named it this way since this is how we originally called it. But that's true, I've also thought that using name with `manager` will sound more like a class. 

While we are on this topic, I considered creating such class, but when I saw `animationsManager` file that is responsible for native platforms I thought that we probably don't want to use that much of OOP (unlike in Gesture Handler)

src/reanimated2/layoutReanimation/web/elementManager.ts
m-bert: `DOMManager` is responsible for inserting and removing animations in the document. `elementManager` performs operations on component that we want to animate. Initially there was only `DOMManager`, but to avoid import cycles I've decided to split it into two files.

piaskowyk (approved):
Much better than it was before 🤗

PR #5152, opened 2023-09-27
Extend createNPMPackage.sh

tjzel:
Sometimes we need to test Reanimated in an outside repo on a fresh package made by `createNPMPackage.sh`. It's annoying when we have to do it multiple times, since package managers cache packages and package installed from a new tarball might not be included in the project.

With these changes I extend the functionality of `createNPMPackage.sh` so now you can call it as follows:

- `./createNPMPackage.sh` works just like before,
- `./createNPMPackage.sh --nightly` or `./createNPMPackage.sh -n` instead of `./createNPMPackage.sh nightly`,
- `./createNPMPackage.sh --fresh` or `./createNPMPackage.sh -f` to generate a fresh tarball that will have the current timestamp appended to its version,
- `./createNPMPackage.sh [version]` to generate a fresh tarball that will have its version set to provided argument.

Respectively, changes were made to `scripts/set-nightly-version.js` - it was moved to `set-reanimated-version.js` and you can call it as follows:
- `node scripts/set-reanimated-version` does nothing,
- `node scripts/set-reanimated-version --help` or `node scripts/set-reanimated-version -h` shows usage info,
- `node scripts/set-reanimated-version --nightly` or `node scripts/set-reanimated-version -n` sets nightly version,
- `node scripts/set-reanimated-version --fresh` or `node scripts/set-reanimated-version -f` sets fresh version,
- `node scripts/set-reanimated-version` sets version to provided argument.

## Test plan

Create all the tarballs and see that they are working.


scripts/set-reanimated-version.js
```diff
@@ -0,0 +1,70 @@
...
+
+    dateIdentifier = dateIdentifier.slice(0, -6);
+    const currentCommit = exec('git rev-parse HEAD', {
+      silent: true,
+    }).stdout.trim();
+    const shortCommit = currentCommit.slice(0, 9);
+
+    const [major, minor, patch] = currentVersion.split('.');
+    const nextMinor = Number(minor) + 1;
+    version = `${major}.${nextMinor}.${patch}-nightly-${dateIdentifier}-${shortCommit}`;
```
bartlomiejbloniarz, line 47: Maybe `patch` here should be hardcoded to 0? I know that the outcome will be the same, since we never change it in `package.json`, but this way it might be a bit confusing. (I got confused 😭)

bartlomiejbloniarz (approved):

scripts/set-reanimated-version.js
tjzel: I think it's a good idea.

createNPMPackage.sh
```diff
@@ -1,11 +1,12 @@
...
+PREVIOUS_VERSION=$(node scripts/set-reanimated-version.js $@)
 yarn install --frozen-lockfile
 yarn bob build
-if [ "$1" = "nightly" ];
-then
-  node scripts/set-nightly-version.js
-fi
 npm pack
+# To preserve previous behavior we don't set back the version on nightly.
+if [ "$1" != "--nightly" ] && [ "$1" != "-n" ]; then
```
piaskowyk, line 8: I think we can safely remove this check because is not necessary

PR #5156, opened 2023-09-28
Add assert to props whitelists

tjzel:
## Summary
Just QoL assert.

## Test plan
```TSX
Animated.addWhitelistedNativeProps({ yes: true });
Animated.addWhitelistedUIProps({ yes: true });
```

### DIFF
--- a/src/ConfigHelper.ts
+++ b/src/ConfigHelper.ts
@@ -1,5 +1,5 @@
 'use strict';
-import { configureProps as jsiConfigureProps } from './reanimated2/core';
+import { jsiConfigureProps } from './reanimated2/core';
 
 /**
  * Styles allowed to be direcly updated in UI thread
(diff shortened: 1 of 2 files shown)


src/ConfigHelper.ts
```diff
@@ -121,7 +121,18 @@ let NATIVE_THREAD_PROPS_WHITELIST: Record<string, boolean> = {
...
+    if (key in UI_THREAD_PROPS_WHITELIST) {
+      throw new Error(
+        `[Reanimated] Property \`${key}\` was whitelisted both as UI and native prop. Please remove it from one of the lists.`
+      );
+    }
+  }
+}
+
 function configureProps(): void {
+  assertNoOverlapInLists();
```
piaskowyk, line 135: I think this check should be at the end of function, to check overlapping after modification of both set.

tjzel: yes

piaskowyk: Actually it is 😅
Sorry, my bad.

piaskowyk (approved):

src/ConfigHelper.ts
tjzel: :swiniok:

PR #5183, opened 2023-10-05
Fix JSISerializer for empty jsi::HostObject

tomekzaw:
## Summary

This PR fixes a crash when passing a `jsi::HostObject` with no props to `_log` or `_toString` functions caused by accessing the last element of an empty list.

## Test plan

```ts
runOnUI(() => {
  'worklet';
  const x = makeShareableCloneOnUIRecursive(() => {
    'worklet';
    console.log(42);
  });
  _log(x);
})();
```


piaskowyk (approved):
🫡

Issue #5181, opened 2023-10-05
[iOS Crash] 3.3.0 - 3.5.4: Slow scroll, fast crash. Memory Leak. useAnimatedScrollHandler Animated ExpoImage


tyrauber:
### Description

Animating an expo-image height within (or outside of) a ScrollView, using useAnimatedScrollHandler and useAnimatedStyle, will crash on an iOS device if you touch and scroll **and do not release your finger**.  The memory will balloon until the app crashes.

https://github.com/software-mansion/react-native-reanimated/assets/109972/5bc9983d-82c0-44e3-9d1a-cd4137bbeeae

Happens in Expo Go or Dev Client on iOS, debug and release.

https://snack.expo.dev/@tyrauber/reanimated-3-useanimatedstyle-crash

![download](https://github.com/software-mansion/react-native-reanimated/assets/109972/783119b3-473e-40ed-9cbe-25366a65da49)

```
  const HEADER_HEIGHT = 600;
  const scrollY = useSharedValue(0);

  const scrollHandler = useAnimatedScrollHandler((event) => {
    scrollY.value = event.contentOffset.y;
  });

  const headerHeight = useAnimatedStyle(() => {
    const height = interpolate(scrollY.value, [0, HEADER_HEIGHT], [HEADER_HEIGHT, 20 + Constants.statusBarHeight], Extrapolate.CLAMP);
    return { height: height };
  });

  return (
    <View style={styles.container}>
       <Animated.ScrollView
        style={styles.scroll}
        onScroll={scrollHandler}
        scrollEventThrottle={16}>
          <AnimatedImage
            style={[{position: 'relative', left: 0, right: 0, flex: 1, height: HEADER_HEIGHT }, headerHeight]}
            source='https://images.unsplash.com/photo-1682686581362-796145f0e123?auto=format&fit=crop&w=2940&q=80'
          />
          <View style={{backgroundColor: 'blue', flex: 1, height: 1200, padding: 10 }}/>
       </Animated.ScrollView>
     </View>
  )
      
```

### Steps to reproduce

1. Create an AnimatedComponent with expo-image
2. Wrap AnimatedImage with an AnimatedScrollView
3.  Add a scrollHandler with `useAnimatedScrollHandler` that sets a sharedValue with `event.contentOffset.y`
4. Use `useAnimatedStyle` to calculate a height based on the sharedValue
5. Use the style to animated the height of the Image
[...]

PR #5184, opened 2023-10-05
Fix unregistering from event inside `useScrollViewOffset` hook

Latropos:
## Summary

Add missing call of `event.current?.unregisterFromEvents();` when component is unmounted. We have similar code inside our `useFrameCallback` hook, but with correct unmounting.
### Now it works like below:

https://github.com/software-mansion/react-native-reanimated/assets/56199675/80802aee-81d9-4c92-b21b-175d9d40d8b9

### Previously some lists and registers were growing bigger and bigger:

https://github.com/software-mansion/react-native-reanimated/assets/56199675/40e91c7c-8f30-47a0-b7e6-6d4580bf477f


efstathiosntonas:
@Latropos does this fixes this: https://github.com/software-mansion/react-native-reanimated/issues/5181

Latropos:
@efstathiosntonas No, it doesn't. 😢 I see that inside our `useAnimatedScrollHandler` hook we also subscribe for events without unsubscribing so maybe we will be able to fox it in similar way.

piaskowyk (approved):

Issue #4932, opened 2023-08-17
Child component does not resize properly if it remounts when parent changes dimensions


tpcstld:
### Description

When a child component is using relative layouts such as `height: 100%`, it sometimes does not properly resize when the parent resizes instantaneously via `useAnimatedStyle` and the child component is remounting.

This affects iOS, and I am unsure if it affects other platforms. From my testing, this occurs in both debug and release builds, and on both physical devices and simulators.

### Steps to reproduce

The minimal repro is below. In the repro, the "parent" element is has a red background and the child component has a green background. Clicking "Press Me" toggles the height of the parent element while remouting the child. I expect that I would just see the green rectangle expand and shrink.

Instead, sometimes when I click the green rectangle does not shrink. Other times, the green rectangle does not expand and I see the red peek out.

https://github.com/software-mansion/react-native-reanimated/assets/5456182/e580c151-28fa-47fa-a3ae-649bdcd94aef

I notice that:

1. If I animate the height transition (e.g. using `withSpring`) then the issue does not occur.
2. If I remove the `key` prop on the child, thus no longer remounting, then the issue does not occur.

```typescript
/**
 * Sample React Native App
 * https://github.com/facebook/react-native
 *
 * @format
 */

import React from 'react';
import Reanimated, {
  useAnimatedStyle,
  useSharedValue,
} from 'react-native-reanimated';
import {Pressable, View, StyleSheet, Text} from 'react-native';

function App(): JSX.Element {
  const [value, setValue] = React.useState(false);
  const height = useSharedValue(100);

  const animatedStyle = useAnimatedStyle(() => {
    return {
      backgroundColor: 'red',
      height: height.value,
      width: '100%',
    };
  });

  const animate = React.useCallback(() => {
    setValue(v => !v);
    height.value = height.value === 100 ? 200 : 100;
  }, [height]);

  return (
    <View style={styles.container}>
      <Reanimated.View style={animatedStyle}>
        <Reanimated.View key={value ? 1 : 2} style={styles.inner} />
      </Reanimated.View>
      <Pressable onPress={animate} style={styles.button}>
        <Text>Press Me</Text>
      </Pressable>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    height: '100%',
  },
  button: {
    position: 'absolute',
    bottom: 100,
  },
  inner: {
    backgroundColor: 'green',
    height: '100%',
  },
});

export default App;
```

### Snack or a link to a repository

[...]

PR #5187, opened 2023-10-05
Fix race-condition during render

piaskowyk:
## Summary

This PR fixes a problem with race conditions during updating parent view and mounting children. The problem occurs in some specific circumstances:
- Parent has an animated style
- Parent update style without animation, just assigns new value to shared value (single frame update, without animation)
- Child doesn't have animated style
- Child's styles depend on parent size
- Child is mounting during update of parent style

In this situation (repro in attached example code) may happen race conditions. See at the graph:

image

### Style updating by Reanimated
image

### Mounting new view by React
image

### So when we merge both graph in single one, It should behave in this way:
image

### But because Reanimated is trying to perform update synchronously it looks like that:
image

In this specific situation, the React Shadow Tree and UI Native Tree are not synchronized. To resolve this issue, I disabled the synchronous style update through Reanimated until those trees can be synchronized once more.

| before | after |
| --- | --- |
| <video src="https://github.com/software-mansion/react-native-reanimated/assets/36106620/07c6d74f-f00a-4879-a523-0a5652c5b5f1" /> | <video src="https://github.com/software-mansion/react-native-reanimated/assets/36106620/484724af-f0e3-44af-b2e6-1efd452c80a4" /> |

Fixes https://github.com/software-mansion/react-native-reanimated/issues/4932

## Test plan

<details>

```js
import Reanimated, {
  useAnimatedStyle,
  useSharedValue,
import {Pressable, View, StyleSheet, Text} from 'react-native';

function App() {
  const height = useSharedValue(100);

    return { height: height.value };
  });

  return (
      <Reanimated.View style={[styles.parent, animatedStyle]}>
      {toggle && <Reanimated.View style={[styles.child]} />}
      </Reanimated.View>

      <Pressable onPress={() => {
        setToggle(!toggle);
        height.value = height.value === 100 ? 200 : 100;
      }} style={styles.button}>
        <Text>Press Me</Text>
      </Pressable>

    </View>
  );
}

  container: {
    height: '100%',
  },
  parent: {
    backgroundColor: 'red',
    width: '100%',
  },
  child: {
    backgroundColor: 'green',
    height: '100%',
    width: '50%'
  },
  button: {
    position: 'absolute',
    bottom: 100,
  },
});

export default App;
```

</details>


piaskowyk:
Closed due to: https://github.com/software-mansion/react-native-reanimated/pull/5224

PR #5188, opened 2023-10-05
Improve spring config verification

Latropos:
## Summary

Each time a spring animation is triggered we check if provided config is valid. When it isn't we skip animation and change the animated view to its final position after waiting the time passed as `duration` (which may be zero)

This change fixes the following bug:
* Warning when spring used with zero duration (which should be a valid usage)
* No warning for negative mass

Additional benefits are:
* Invalid config warning informs what exactly is wrong
* Types are both more accurate and less verobse

Behaviour when duration is zero:

https://github.com/software-mansion/react-native-reanimated/assets/56199675/c16caa2a-9df0-4ba0-9890-4fe81b5b090d

### DIFF
--- a/src/reanimated2/animation/spring.ts
+++ b/src/reanimated2/animation/spring.ts
@@ -19,7 +19,7 @@ import {
   underDampedSpringCalculations,
   criticallyDampedSpringCalculations,
   isAnimationTerminatingCalculation,
-  isConfigValid,
+  checkIfConfigIsValid,
 } from './springUtils';
 
 // TODO TYPESCRIPT This is a temporary type to get rid of .d.ts file.
@@ -58,7 +58,7 @@ export const withSpring = ((
       skipAnimation: false,
     };
 
-    config.skipAnimation = !isConfigValid(config);
+    config.skipAnimation = !checkIfConfigIsValid(config);
 
     if (config.duration === 0) {
       config.skipAnimation = true;
@@ -80,13 +80,9 @@ export const withSpring = ((
       }
 
       if (config.skipAnimation) {
-        // If we use duration we want to wait the provided time before stopping
-        if (config.useDuration) return false;
-        else {
-          animation.current = toValue;
-          animation.lastTimestamp = 0;
-          return true;
-        }
+        animation.current = toValue;
+        animation.lastTimestamp = 0;
+        return true;
       }
       const { lastTimestamp, velocity } = animation;
 
--- a/src/reanimated2/animation/spring.ts
+++ b/src/reanimated2/animation/spring.ts
@@ -19,7 +19,7 @@ import {
   underDampedSpringCalculations,
   criticallyDampedSpringCalculations,
   isAnimationTerminatingCalculation,
-  validateConfig,
+  isConfigValid,
 } from './springUtils';
 
 // TODO TYPESCRIPT This is a temporary type to get rid of .d.ts file.
@@ -58,7 +58,7 @@ export const withSpring = ((
       skipAnimation: false,
     };
 
-    config.skipAnimation = validateConfig(config);
+    config.skipAnimation = !isConfigValid(config);
 
     if (config.duration === 0) {
       config.skipAnimation = true;
--- a/src/reanimated2/animation/spring.ts
+++ b/src/reanimated2/animation/spring.ts
@@ -11,6 +11,7 @@ import type {
   SpringAnimation,
   InnerSpringAnimation,
   SpringConfigInner,
+  DefaultSpringConfig,
 } from './springUtils';
 import {
   initialCalculations,
@@ -18,6 +19,7 @@ import {
   underDampedSpringCalculations,
   criticallyDampedSpringCalculations,
   isAnimationTerminatingCalculation,
+  validateConfig,
 } from './springUtils';
 
 // TODO TYPESCRIPT This is a temporary type to get rid of .d.ts file.
@@ -36,7 +38,7 @@ export const withSpring = ((
 
   return defineAnimation<SpringAnimation>(toValue, () => {
     'worklet';
-    const defaultConfig: Record<keyof SpringConfig, any> = {
+    const defaultConfig: DefaultSpringConfig = {
       damping: 10,
       mass: 1,
       stiffness: 100,
@@ -49,28 +51,17 @@ export const withSpring = ((
       reduceMotion: undefined,
     } as const;
 
-    const config: Record<keyof SpringConfig, any> & SpringConfigInner = {
+    const config: DefaultSpringConfig & SpringConfigInner = {
       ...defaultConfig,
       ...userConfig,
       useDuration: !!(userConfig?.duration || userConfig?.dampingRatio),
-      configIsInvalid: false,
+      skipAnimation: false,
     };
 
-    if (
-      [
-        config.stiffness,
-        config.damping,
-        config.duration,
-        config.dampingRatio,
-        config.restDisplacementThreshold,
-        config.restSpeedThreshold,
-      ].some((x) => x <= 0) ||
-      config.mass === 0
-    ) {
-      config.configIsInvalid = true;
-      console.warn(
-        "You have provided invalid spring animation configuration! \n Value of stiffness, damping, duration and damping ratio must be greater than zero, and mass can't equal zero."
-      );
+    config.skipAnimation = validateConfig(config);
+
+    if (config.duration === 0) {
+      config.skipAnimation = true;
     }
 
     function springOnFrame(
@@ -88,7 +79,7 @@ export const withSpring = ((
         return true;
       }
 
-      if (config.configIsInvalid) {
+      if (config.skipAnimation) {
         // We don't animate wrong config
         if (config.useDuration) return false;
         else {
--- a/src/reanimated2/animation/spring.ts
+++ b/src/reanimated2/animation/spring.ts
@@ -80,7 +80,7 @@ export const withSpring = ((
       }
 
       if (config.skipAnimation) {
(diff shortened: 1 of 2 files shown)


Latropos:
@bartlomiejbloniarz Do we have two ways of skipping the spring animation now? Do they have exactly the same behaviour?

bartlomiejbloniarz:
@Latropos there is a slight difference between those behaviours. When reduced motion is enabled, the `toValue` is reached immediately. But when you pass an invalid config like `{duration:1000, dampingRatio:0}` the `toValue` will be reached after a second. Is this the intended behaviour?

Latropos:
@bartlomiejbloniarz It was deliberate decision to wait the `duration` time, however now I think I wasn't a good idea and immediate change would be better.

src/reanimated2/animation/spring.ts
```diff
@@ -49,28 +51,17 @@ export const withSpring = ((
...
-        config.restDisplacementThreshold,
-        config.restSpeedThreshold,
-      ].some((x) => x <= 0) ||
-      config.mass === 0
-    ) {
-      config.configIsInvalid = true;
-      console.warn(
-        "You have provided invalid spring animation configuration! \n Value of stiffness, damping, duration and damping ratio must be greater than zero, and mass can't equal zero."
-      );
+    config.skipAnimation = validateConfig(config);
```
tomekzaw, line 61: It's a little bit counterintuitive, I would expect that `validateConfig` returns whether the config is valid. 
```suggestion
    config.skipAnimation = !isConfigValid(config);
```

src/reanimated2/animation/spring.ts
```diff
@@ -49,28 +51,17 @@ export const withSpring = ((
...
+
+    if (config.duration === 0) {
+      config.skipAnimation = true;
```
tomekzaw, line 64: If the current value is 100, the target value is 200 and duration is 0, do we set the final value in such case?

src/reanimated2/animation/springUtils.ts
```diff
@@ -52,6 +58,38 @@ export interface InnerSpringAnimation
   toValue: number;
   current: number;
 }
+export function validateConfig(config: DefaultSpringConfig): boolean {
+  'worklet';
+  let errorLog = '[Reanimated] Invalid spring config, ';
+  let configIsInvalid = false;
```
tomekzaw, line 64:
```suggestion
  let isConfigInvalid = false;
```

src/reanimated2/animation/springUtils.ts
```diff
@@ -52,6 +58,38 @@ export interface InnerSpringAnimation
...
+  (
+    [
+      'stiffness',
+      'damping',
+      'dampingRatio',
+      'restDisplacementThreshold',
+      'restSpeedThreshold',
+      'mass',
+    ] as const
+  ).forEach((property) => {
```
tomekzaw, line 75: In React world, `prop` is usually used here.
```suggestion
  ).forEach((prop) => {
```

src/reanimated2/animation/springUtils.ts
```diff
@@ -52,6 +58,38 @@ export interface InnerSpringAnimation
...
+    if (config[property] <= 0) {
+      configIsInvalid = true;
+      errorLog += `${property} must be grater than zero but got ${config[property]}, `;
+    }
```
tomekzaw, line 79:
```suggestion
    const value = config[prop];
    if (value <= 0) {
      isConfigInvalid = true;
      errorLog += `, ${prop} must be grater than zero but got ${value}`;
    }
```

src/reanimated2/animation/springUtils.ts
```diff
@@ -52,6 +58,38 @@ export interface InnerSpringAnimation
   toValue: number;
   current: number;
 }
+export function validateConfig(config: DefaultSpringConfig): boolean {
+  'worklet';
+  let errorLog = '[Reanimated] Invalid spring config, ';
```
tomekzaw, line 63: We should remove comma from here and instead add it to concrete error messages below.
```suggestion
  let errorMessage = '[Reanimated] Invalid spring config';
```
Apart from this, there's no need to allocate this string each time we call `withSpring`, maybe we could optimize the memory usage of the happy path here?

src/reanimated2/animation/springUtils.ts
```diff
@@ -52,6 +58,38 @@ export interface InnerSpringAnimation
...
+      'damping',
+      'dampingRatio',
+      'restDisplacementThreshold',
+      'restSpeedThreshold',
+      'mass',
+    ] as const
+  ).forEach((property) => {
+    if (config[property] <= 0) {
+      configIsInvalid = true;
+      errorLog += `${property} must be grater than zero but got ${config[property]}, `;
```
tomekzaw, line 78:
```suggestion
      errorLog += `, ${property} must be grater than zero but got ${config[property]}`;
```

src/reanimated2/animation/springUtils.ts
```diff
@@ -52,6 +58,38 @@ export interface InnerSpringAnimation
...
+    }
+  });
+
+  if (config.duration < 0) {
+    configIsInvalid = true;
+    errorLog += `duration can't be negative, got ${config.duration}, `;
```
tomekzaw, line 84:
```suggestion
    errorLog += `, duration can't be negative, got ${config.duration}`;
```

src/reanimated2/animation/springUtils.ts
```diff
@@ -52,6 +58,38 @@ export interface InnerSpringAnimation
...
+  }
+
+  if (configIsInvalid) {
+    console.warn(errorLog);
```
tomekzaw, line 88:
```suggestion
  if (isConfigInvalid) {
    console.warn(errorMessage);
```

tomekzaw (changes requested):

src/reanimated2/animation/spring.ts
Latropos: Yes it is.
In this line we set `skipAnimation` flag, which is used in this clause:
```
      if (config.skipAnimation) {
        // If we use duration we want to wait the provided time before stopping
        if (config.useDuration) return false;
        else {
          animation.current = toValue;
          animation.lastTimestamp = 0;
          return true;
        }
      }
```
Therefore the animation without duration sets the value inside this if condition and animation with `duration = 0` sets the final value in next iteration, here:

```javascript

        animation.current = toValue;
        animation.lastTimestamp = 0;
        return true;
      }
```

Latropos: @tomekzaw  Now I regret the decision to wait the duration time for invalid config and I think we should always update the value immediately, WDYT?

tomekzaw: Sounds ok

Latropos: @tomekzaw Is this a change for this PR or for a separate one?

src/reanimated2/animation/spring.ts
```diff
@@ -88,8 +79,8 @@ export const withSpring = ((
         return true;
       }
 
-      if (config.configIsInvalid) {
-        // We don't animate wrong config
+      if (config.skipAnimation) {
+        // If we use duration we want to wait the provided time before stopping
         if (config.useDuration) return false;
```
tomekzaw, line 84: There's no curly braces after `if`, can we add them and also check why linter doesn't complain about it? We should have `curly` rule enabled. Also, what's the conclusion when it comes to the expected behavior when config is invalid? Why does it depend on whether we use duration or not?
```suggestion
        if (config.useDuration) {
          return false;
        }
```

src/reanimated2/animation/springUtils.ts
```diff
@@ -52,6 +58,35 @@ export interface InnerSpringAnimation
...
+      errorMessage += `, ${prop} must be grater than zero but got ${value}`;
+    }
+  });
+
+  if (config.duration < 0) {
+    errorMessage += `, duration can't be negative, got ${config.duration}`;
+  }
+
+  if (errorMessage !== '') {
+    console.warn('[Reanimated] Invalid spring config' + errorMessage);
```
tomekzaw, line 85: The name of this method `isConfigValid` suggests that it has no side-effects, however it can call `console.warn`. Can we think of a better name here?

tomekzaw (changes requested):

src/reanimated2/animation/spring.ts
Latropos: I've thought we have decided that changing this behaviour should be done in separate PR (to avoid big changes on a reviewed PR), but since this is not a big change I'll delete it. It is because of the previous behaviour (we used to wait duration time before updating the value).

Latropos: The rule 'curly' is only enabled in plugin

src/reanimated2/animation/springUtils.ts
Latropos: I think that `checkIfConfigIsValid` may be better here.

src/reanimated2/animation/spring.ts
Latropos: We can enable it (we have vialotion of this rule only in one file)

tomekzaw: Thanks for checking this, let's do it (enable it everywhere)

src/reanimated2/animation/springUtils.ts
tomekzaw: Sounds good

tomekzaw (approved):
Looks good to me. Let's wait for @piaskowyk's approval before merging.

piaskowyk (approved):

PR #5189, opened 2023-10-05
Fix article progress example on android

Latropos:
## Summary

An error is thrown when you measure en element that is not mounted yet. So I create a boolean flag which I set when component is ready to avoid this error.

Probably we also should improve error log.

Current error log:
> `[Reanimated] The view with tag ${viewTag} returned an invalid measurement response.`

New error log:
> `[Reanimated] The view with tag ${viewTag} returned an invalid measurement response. Please make sure the view is currently rendered.`

Both logs are not perfect, since in our example our measured element is a Text component, not a View one, however adding component type to log is not easy.

### DIFF
--- a/app/src/examples/ArticleProgressExample.tsx
+++ b/app/src/examples/ArticleProgressExample.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import { SafeAreaView, StyleSheet } from 'react-native';
 import Animated, {
   measure,
@@ -11,16 +11,12 @@ export default function ArticleProgressExample() {
   const scrollViewRef = useAnimatedRef<Animated.ScrollView>();
   const textRef = useAnimatedRef<Animated.Text>();
   const scrollHandler = useScrollViewOffset(scrollViewRef);
-  const [mounted, setMounted] = useState(false);
-
-  useEffect(() => {
-    setMounted(true);
-  }, []);
 
   const progressBarAnimatedStyle = useAnimatedStyle(() => {
-    if (!mounted) {
+    if (!textRef.current || !scrollViewRef.current) {
       return {};
     }
+
     const measuredText = measure(textRef);
     if (measuredText === null) {
       return { width: 0 };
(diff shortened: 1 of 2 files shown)


app/src/examples/ArticleProgressExample.tsx
```diff
@@ -11,8 +11,16 @@ export default function ArticleProgressExample() {
...
   const textRef = useAnimatedRef<Animated.Text>();
   const scrollHandler = useScrollViewOffset(scrollViewRef);
+  const [mounted, setMounted] = useState(false);
+
+  useEffect(() => {
+    setMounted(true);
+  }, []);
 
   const progressBarAnimatedStyle = useAnimatedStyle(() => {
+    if (!mounted) {
```
piaskowyk, line 21: W shouldn't use state value in worklet, lets change it to shared value.

src/reanimated2/NativeMethods.ts
```diff
@@ -75,7 +75,7 @@ if (isWeb()) {
       return null;
     } else if (measured.x === -1234567) {
       console.warn(
-        `[Reanimated] The view with tag ${viewTag} returned an invalid measurement response.`
+        `[Reanimated] The view with tag ${viewTag} returned an invalid measurement response. Please make sure the view is currently rendered.`
```
piaskowyk, line 78:
```suggestion
        `[Reanimated] The view with tag ${viewTag} returned an invalid measurement response. Please ensure that the view has been rendered.`
```

piaskowyk (changes requested):

app/src/examples/ArticleProgressExample.tsx
```diff
@@ -13,6 +13,10 @@ export default function ArticleProgressExample() {
   const scrollHandler = useScrollViewOffset(scrollViewRef);
 
   const progressBarAnimatedStyle = useAnimatedStyle(() => {
+    if (!textRef.current || !scrollViewRef.current) {
```
tomekzaw, line 16: Looks like we pass React refs to a worklet, do we really want to do it?

Latropos: Why not? We were already using ref inside worklet:
```javascript
const measuredText = measure(textRef);
```
And this seems to be the best way to check if component is already rendered...

tomekzaw: @Latropos Turns out it doesn't work as expected, animated refs don't have `.current` property when used on UI runtime. We'll probably need to revert this PR and find another fix. cc @jfedak whom I've complained to about ArticleProgressExample not working on macOS

PR #5190, opened 2023-10-05
Fix mapper getting stuck on an error in worklet

tjzel:
## Summary

Currently, in `mapperRun` function, when we set `processingMappers` to `true` and then a breaking worklet is called in
```TSX
      for (const mapper of sortedMappers) {
        if (mapper.dirty) {
          mapper.dirty = false;
          mapper.worklet();
        }
      }
```
      
We get stuck in `processingMappers` always set to `true` and mappers are never run again without reloading the app. This PR fixes it.

## Test plan

Animations here break after first button press without this PR.
<details>
<summary>
Snippet
</summary>

```TSX

import React from 'react';

import Animated, {
  useSharedValue,
  useAnimatedStyle,

    if (sv.value === 110) {
      throw new Error('Break animations');
    }
    return {
      width: sv.value,
    };
  });

  function handleAnimate() {
    sv.value += 10;
  }

  return (
      <Button title="Animate" onPress={handleAnimate} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: {
    height: 100,
    backgroundColor: 'red',
  },
});
```
</details>

### DIFF
--- a/src/reanimated2/mappers.ts
+++ b/src/reanimated2/mappers.ts
@@ -85,17 +85,20 @@ function createMapperRegistry() {
     if (processingMappers) {
       return;
     }
-    processingMappers = true;
-    if (mappers.size !== sortedMappers.length) {
-      updateMappersOrder();
-    }
-    for (const mapper of sortedMappers) {
-      if (mapper.dirty) {
-        mapper.dirty = false;
-        mapper.worklet();
+    try {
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):
tha.nks

PR #4332, opened 2023-04-04
Experimental transpilation functionality for nested worklets

tjzel:
## Summary

This PR adds optional functionality to plugin that allows nesting worklets. This is necessary to be able to call functions in-between threads and is a step forward to allow multi-threading in react-native-reanimated.

Current use cases are provided in the comment below.

## Test plan

Added some test suites.

PR #5192, opened 2023-10-06
Run worklets in background

tomekzaw:
## Summary

Work in progress.


Common/cpp/ReanimatedRuntime/WorkletRuntimeDecorator.cpp
```diff
@@ -125,6 +140,13 @@ void WorkletRuntimeDecorator::decorate(
             rt, synchronizedDataHolderRef, newData);
       });
 
+  jsi_utils::installJsiFunction(
+      rt,
+      "_getDataSynchronously",
+      [](jsi::Runtime &rt, const jsi::Value &synchronizedDataHolderRef) {
+        return reanimated::getDataSynchronously(rt, synchronizedDataHolderRef);
+      });
+
```
tomekzaw, line 149: Adding `_getDataSynchronously` to worklet runtime could be moved to a separate PR.

src/reanimated2/background.ts
```diff
@@ -0,0 +1,62 @@
...
+    return (...args) =>
+      NativeReanimatedModule.scheduleOnBackgroundQueue(
+        backgroundQueue,
+        workletRuntime,
+        makeShareableCloneRecursive(() => {
+          'worklet';
+          worklet(...args);
+        })
+      );
+  }
```
tomekzaw, line 61: We need to have two separate implementations here, one for RN runtime and the other for worklet runtimes (including the UI runtime) because of two reasons:
1. In the RN runtime, we inject the native code using a HostObject with native methods, but in the UI (worklet) runtime we inject native code as individual functions. We'll need to unify the injecting logic.
2. In the RN runtime, we need to use `makeShareableCloneRecursive`, as opposed to `makeShareableCloneOnUIRecursive` on the UI (worklet) runtime. We'll need to unify the cloning logic.

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -50,6 +50,22 @@ export default class JSReanimated {
...
+    throw new Error(
+      '[Reanimated] createBackgroundQueue is not available in JSReanimated.'
+    );
+  }
+
+  scheduleOnBackgroundQueue<T>(
+    _backgroundQueue: BackgroundQueue,
+    _workletRuntime: WorkletRuntime,
+    _worklet: ShareableRef<T>
+  ): BackgroundQueue {
```
Latropos, line 63: WDYT? Such a change could be made with all function throwing error in this file.
```suggestion
  ): never {
```

Example/babel.config.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
   presets: ['module:metro-react-native-babel-preset'],
-  plugins: ['../plugin'],
+  plugins: [['../plugin', {processNestedWorklets: true}]],
```
Latropos, line 3: We have never documented this plugin option (processNestedWorklets). Should we?

tomekzaw: Yes, it was added in #4332 and should be documented in a separate PR.

src/reanimated2/js-reanimated/index.ts
```diff
@@ -46,6 +46,12 @@ global._scheduleOnJS = () => {
   );
 };
 
+global._scheduleOnBackgroundQueue = () => {
+  throw new Error(
+    '[Reanimated] _scheduleOnBackgroundQueue should never be called in JSReanimated.'
```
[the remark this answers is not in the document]
tomekzaw: 1. Three times is not that much, but yeah, we could do it
2. This particular part of code needs a small refactor since there's way more functions that shouldn't be called in JSReanimated mode. We could define all such functions in an array and loop through it.

src/reanimated2/js-reanimated/JSReanimated.ts
tomekzaw: Thanks for this suggestion, indeed `never` works. I think we could go one step ahead and also remove the unused arguments along with the template and also remove the return type. I remember I tested it some time ago but it didn't work for me, but it's okay now. I'd be more than happy to change all methods in the next PR.
```ts
  scheduleOnBackgroundQueue() {
    throw new Error(
      '[Reanimated] scheduleOnBackgroundQueue is not available in JSReanimated.'
    );
  }
```

src/reanimated2/background.ts
tjzel: Should we change the name (or provide an alias) for `makeShareableCloneOnUIRecursive` to `makeShareableCloneOnWorkletRuntime`?

src/reanimated2/background.ts
```diff
@@ -0,0 +1,62 @@
...
+): (...args: Args) => void {
+  'worklet';
+  if (__DEV__ && !SHOULD_BE_USE_WEB && worklet.__workletHash === undefined) {
+    throw new Error(
+      '[Reanimated] The function passed to `runOnBackgroundQueue` is not a worklet.'
+    );
+  }
+  if (_WORKLET) {
+    return (...args) =>
+      _scheduleOnBackgroundQueue(
```
tjzel, line 43: Should we access it with `global`? We were doing it with e.g. `IS_FABRIC` until we didn't.

src/reanimated2/background.ts
tomekzaw: Yep, we could do it

src/reanimated2/background.ts
tomekzaw: Currently, we access all JSI bindings directly, but I think we could go for `global._foo`. This way we don't need to whitelist identifiers in Reanimated Babel plugin.

src/reanimated2/background.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { WorkletFunction } from './commonTypes';
+import type { WorkletRuntime } from './runtimes';
+import {
+  makeShareableCloneOnUIRecursive,
+  makeShareableCloneRecursive,
+} from './shareables';
+
+const SHOULD_BE_USE_WEB = shouldBeUseWeb();
+
+export type BackgroundQueue = {
```
tjzel, line 13: So, from my understanding this type is just to signify that it's a HostObject and nothing should be accessed on it. How about this pattern then?
BackgroundQueue

src/reanimated2/background.ts
```diff
@@ -0,0 +1,62 @@
...
+    return (...args) =>
+      _scheduleOnBackgroundQueue(
+        backgroundQueue,
+        workletRuntime,
+        makeShareableCloneOnUIRecursive(() => {
+          'worklet';
+          worklet(...args);
+        })
+      );
+  } else {
```
tjzel, line 51: redundant else

app/src/examples/WorkletRuntimeExample.tsx
```diff
@@ -122,6 +130,72 @@ function PerformanceNowDemo() {
...
+let runtime: WorkletRuntime | undefined;
+
+function BackgroundQueueLongRunningTaskDemo() {
+  const handlePress = () => {
+    if (queue === undefined) {
+      queue = createBackgroundQueue('bar');
+    }
+    if (runtime === undefined) {
+      runtime = createWorkletRuntime('foo');
+    }
```
tjzel, line 174: Is it possible that one would already be destroyed but not the other? Or they don't get destroyed if you get out of the scope?

app/src/examples/WorkletRuntimeExample.tsx
tjzel, line 1: How about an example that actually utilizes the queue out of the box (immediately queueing 2 tasks or so)?

Example/babel.config.js
tjzel: Let's be nice guys and write the docs for it first. I could do that if you want.

Common/cpp/Tools/BackgroundQueue.cpp
```diff
@@ -0,0 +1,97 @@
+#include "BackgroundQueue.h"
+
+#include <utility>
+
+namespace reanimated {
+
+BackgroundQueue::BackgroundQueue(const std::string &name) : name_(name) {
+  thread_ = std::thread([this] {
+#if __APPLE__
```
tjzel, line 9: Why #if here? Can it be set to a false value on Android?

Common/cpp/Tools/BackgroundQueue.h
```diff
@@ -0,0 +1,63 @@
...
+
+class BackgroundQueue : public jsi::HostObject {
+ public:
+  explicit BackgroundQueue(const std::string &name);
+
+  ~BackgroundQueue() override;
+
+  std::string toString() const {
+    return "[BackgroundQueue \"" + name_ + "\"]";
+  }
```
tjzel, line 29: More of a programming question than a review bit. Is it a convention/good practice, to make a `toString` method that does not "make sense" as a string? Or is it for Java purposes maybe?

Common/cpp/Tools/BackgroundQueue.h
```diff
@@ -0,0 +1,63 @@
...
+
+  const std::string name_;
+  std::thread thread_;
+  std::atomic_bool running_{true};
+  std::mutex mutex_;
+  std::condition_variable cv_;
+  std::queue<std::pair<
+      std::shared_ptr<WorkletRuntime>,
+      std::shared_ptr<ShareableWorklet>>>
+      queue_;
```
tjzel, line 50: I wonder, do we actually want to have a multi-thread queue or rather to tightly couple queues to their designated threads? What's the philosophy behind it?

Common/cpp/Tools/BackgroundQueue.cpp
```diff
@@ -0,0 +1,97 @@
...
+  });
+#ifdef ANDROID
+  pthread_setname_np(thread_.native_handle(), name_.c_str());
+#endif
+}
+
+BackgroundQueue::~BackgroundQueue() {
+  running_ = false;
+  cv_.notify_all();
+  thread_.join();
```
tjzel, line 22: Why `join` instead of `detach`?

Common/cpp/Tools/BackgroundQueue.cpp
```diff
@@ -0,0 +1,97 @@
...
+        0,
+        [this](jsi::Runtime &rt, const jsi::Value &, const jsi::Value *, size_t)
+            -> jsi::Value {
+          return jsi::String::createFromUtf8(rt, toString());
+        });
+  }
+  if (name == "name") {
+    return jsi::String::createFromUtf8(rt, name_);
+  }
+  return jsi::Value::undefined();
```
tjzel, line 42: Do we actually want to return undefined here instead of raising some errors? Or is it just a pattern for HostObjects?

Common/cpp/Tools/BackgroundQueue.cpp
```diff
@@ -0,0 +1,97 @@
...
+}
+
+std::vector<jsi::PropNameID> BackgroundQueue::getPropertyNames(
+    jsi::Runtime &rt) {
+  std::vector<jsi::PropNameID> result;
+  result.push_back(jsi::PropNameID::forUtf8(rt, "toString"));
+  result.push_back(jsi::PropNameID::forUtf8(rt, "name"));
```
tjzel, line 49: How about moving it to initialization list here?

Common/cpp/Tools/BackgroundQueue.cpp
```diff
@@ -0,0 +1,97 @@
...
+    cv_.wait(lock, [this] { return !queue_.empty() || !running_; });
+    if (!running_) {
+      return;
+    }
+    if (!queue_.empty()) {
+      auto [workletRuntime, shareableWorklet] = std::move(queue_.front());
+      queue_.pop();
+      lock.unlock();
+      workletRuntime->runGuarded(shareableWorklet);
+    }
```
tjzel, line 73: Is the check for `running_` present here because we can imperatively terminate a running queue with the destructor?

Common/cpp/Tools/BackgroundQueue.cpp
```diff
@@ -0,0 +1,97 @@
...
+    jsi::Runtime &rt,
+    const jsi::Value &backgroundQueueValue,
+    const jsi::Value &workletRuntimeValue,
+    const jsi::Value &shareableWorkletValue) {
+  auto backgroundQueue = extractBackgroundQueue(rt, backgroundQueueValue);
+  auto workletRuntime = extractWorkletRuntime(rt, workletRuntimeValue);
+  auto shareableWorklet = extractShareableOrThrow<ShareableWorklet>(
+      rt,
+      shareableWorkletValue,
+      "[Reanimated] Function passed to `_scheduleOnBackgroundQueue` is not a shareable worklet. Please make sure that `processNestedWorklets` option in Reanimated Babel plugin is enabled.");
```
tjzel, line 93: Let's add the docs first and link it here alongside the error.

Common/cpp/ReanimatedRuntime/WorkletRuntime.h
```diff
@@ -41,7 +41,7 @@ class WorkletRuntime : public jsi::HostObject {
   }
 
   std::string toString() const {
-    return "<WorkletRuntime \"" + name_ + "\">";
+    return "[WorkletRuntime \"" + name_ + "\"]";
```
tjzel, line 44: Have you thought about `[ReanimatedWorkletRuntime` ?

Common/cpp/NativeModules/NativeReanimatedModuleSpec.cpp
tjzel, line 1: Explain to me one day what is this file 😆

src/reanimated2/globals.d.ts
tjzel, line 1: what about _getDataSynchronously?

tjzel (changes requested):
I read it all and the PR looks extremely solid and clean! I might need to take another look into the mutex/cv shenanigans but only after you answer some (at least one) of my questions. Reading it was a pleasure (as with every pull request that doesn't base on an existing code I guess 💀).

Common/cpp/NativeModules/NativeReanimatedModuleSpec.cpp
piaskowyk: This file is typically generated by codegen, but in our case, we don't use codegen 😅

Common/cpp/ReanimatedRuntime/WorkletRuntimeDecorator.cpp
piaskowyk: Where `_getDataSynchronously` is used? 🤔

Common/cpp/Tools/BackgroundQueue.h
```diff
@@ -0,0 +1,63 @@
...
+      const std::shared_ptr<ShareableWorklet> &worklet);
+
+ private:
+  void runLoop();
+
+  const std::string name_;
+  std::thread thread_;
+  std::atomic_bool running_{true};
+  std::mutex mutex_;
+  std::condition_variable cv_;
```
piaskowyk, line 46: Could we rename `cv_` to something more informative? 🙏

app/src/examples/WorkletRuntimeExample.tsx
```diff
@@ -122,6 +130,72 @@ function PerformanceNowDemo() {
   return <Button title="performance.now" onPress={handlePress} />;
 }
 
+function BackgroundQueueFromJSDemo() {
+  const handlePress = () => {
+    const runtime = createWorkletRuntime('foo');
+    const queue = createBackgroundQueue('bar');
+    runOnBackgroundQueue(queue, runtime, () => {
```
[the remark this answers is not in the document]
tomekzaw: He will be punished with a EXC_BAD_ACCESS. Do you think we should prevent this (e.g. using locks) or at least warn about it (i.e. tell the user that the runtime is already being used on another thread)?

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -50,6 +50,18 @@ export default class JSReanimated {
...
+  createBackgroundQueue(_name: string): BackgroundQueue {
+    throw new Error(
+      '[Reanimated] createBackgroundQueue is not available in JSReanimated.'
+    );
+  }
+
+  scheduleOnBackgroundQueue() {
+    throw new Error(
+      '[Reanimated] scheduleOnBackgroundQueue is not available in JSReanimated.'
+    );
```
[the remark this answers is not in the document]
tomekzaw: Yes, at least that's the plan.

Example/babel.config.js
tomekzaw: Please do 🙏

Common/cpp/Tools/BackgroundQueue.cpp
tomekzaw: In pure JS, if you have `const foo = { bar: 1 };` then accessing `foo.xyz` returns `undefined`. So I'd say this is a pattern for JS objects and here I'm just trying to mimic it on a `jsi::HostObject`. And yes, this is also how other people implement `jsi::HostObject::get`.

Common/cpp/Tools/BackgroundQueue.cpp
tomekzaw: I've tried it once but it didn't work for me. 🤷 Also, other libs usually cache these props but in this case it sounds like an overkill to me.

Common/cpp/Tools/BackgroundQueue.cpp
tomekzaw: Nope, one should never call the destructor directly. But there could be `BackgroundQueue::shutdown`/`terminate` method which sets `running_` to false.

Common/cpp/Tools/BackgroundQueue.cpp
tomekzaw: Makes sense to me, will do

app/src/examples/WorkletRuntimeExample.tsx
```diff
@@ -122,6 +130,72 @@ function PerformanceNowDemo() {
...
+      'worklet';
+      console.log('Hello from background!', Math.random());
+    })();
```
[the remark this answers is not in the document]
tomekzaw: Great idea, will do! It should be easy since it's analogous to `runOnUI`. Do you mind if this will be a separate PR to keep this one minimal?

app/src/examples/WorkletRuntimeExample.tsx
```diff
@@ -122,6 +130,72 @@ function PerformanceNowDemo() {
...
+
+function BackgroundQueueFromUIDemo() {
+  const handlePress = () => {
+    const runtime = createWorkletRuntime('foo');
+    const queue = createBackgroundQueue('bar');
+    runOnUI(() => {
+      'worklet';
+      const x = Math.random();
+      console.log('Hello from UI thread!', x);
+      runOnBackgroundQueue(queue, runtime, () => {
```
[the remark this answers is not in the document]
tomekzaw: Yep, it's possible to call `runOnBackgroundQueue` from any worklet runtime (including the UI runtime) since it just calls `_scheduleOnBackgroundQueue` which is injected to all worklet runtimes.

Common/cpp/Tools/BackgroundQueue.cpp
```diff
@@ -0,0 +1,97 @@
...
+    const std::shared_ptr<WorkletRuntime> &runtime,
+    const std::shared_ptr<ShareableWorklet> &worklet) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  queue_.emplace(runtime, worklet);
+  cv_.notify_one();
+}
+
+void BackgroundQueue::runLoop() {
+  while (running_) {
+    std::unique_lock<std::mutex> lock(mutex_);
```
piaskowyk, line 63: If I understand correctly we can move `std::unique_lock<std::mutex> lock(mutex_);` befor loop declaration, and then we don't need to unblock the lock.

app/src/examples/WorkletRuntimeExample.tsx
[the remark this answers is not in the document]
piaskowyk: sure, let's do it in another PR

app/src/examples/WorkletRuntimeExample.tsx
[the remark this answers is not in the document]
piaskowyk: I believe that for now, it would suffice to mention this in the documentation. We can observe how people use it and determine in the future if there is a need to incorporate a locking mechanism or similar solution.

Common/cpp/Tools/BackgroundQueue.cpp
tjzel: 🤔

Common/cpp/Tools/BackgroundQueue.cpp
tjzel: So is it currently possible that queue won't be empty and running_ will be false?

Common/cpp/ReanimatedRuntime/WorkletRuntimeDecorator.cpp
tomekzaw: I'm using it in react-native-vision-camera. Currently, there seems to be some problem with copying shared values (or rather mutables) to worklet runtimes as they become regular JS objects (e.g. `{value: 42}`) and thus lose reactivity. The workaround for me was to use `SynchronousDataHolder` which is a `jsi::HostObject` so it can be passed to arbitrary worklet runtime. In there, I'm using `_getDataSynchronously` to read the current value and `_updateDataSynchronously` to update it. Nice thing about this is that it's indeed synchronous and uses mutexes instead of being asynchronous. Since we already inject `_updateDataSynchronously`, I thought there's no reason not to inject `_getDataSynchronously` as well.

src/reanimated2/background.ts
tomekzaw: Done in 74529a65b0c3e52bf5da71f653c83bf5a16b084f

src/reanimated2/background.ts
tomekzaw: It was there on purpose to emphasize the similarity between the implementation of both branches. Removed in a9938bb65f00d1ff3a479c593a7f8b13b99d12db

app/src/examples/WorkletRuntimeExample.tsx
tomekzaw: `queue` and `runtime` are global variables defined in lines 164-165 so technically they never get destroyed in this example.

app/src/examples/WorkletRuntimeExample.tsx
tomekzaw: You could achieve this effect by pressing the button 2 times. Done in 4b12419d8dbd6f58220b247d9c8dacb6789ff5f1

Common/cpp/Tools/BackgroundQueue.cpp
tomekzaw: Probably not, the function has different number of arguments depending on the platform.

Common/cpp/Tools/BackgroundQueue.h
tomekzaw: This is solely for developer experience purposes so that `String(queue)` returns a nice description of the object. Instead of implementing this directly in `BackgroundQueue::get` method (for `jsi::HostObject`), I moved it to a separate C++ method `BackgroundQueue::toString`.

Common/cpp/Tools/BackgroundQueue.h
tomekzaw: At this point there's no philosophy, we simply want to allow users to run worklets in background. That being said, we don't want to create a new thread per each job, so we want to re-use existing threads. Since threads (and runtimes) can do at most one thing at once, we need to queue these operations. Currently, a `BackgroundQueue` represents a single background thread which has its own queue. There's no concept of thread pool or executors, these things will come later on.

Common/cpp/Tools/BackgroundQueue.cpp
tomekzaw: That's an excellent question, I've copied this code from somewhere else. We'll need to decide if we want to change it or not. cc @kmagiera

src/reanimated2/globals.d.ts
tomekzaw: Good catch! Added in 00d4346874559331db2b3dcd059bbbc25bf8dc53

Common/cpp/Tools/BackgroundQueue.h
tjzel: So is it necessary at the moment to store pairs here? Why wouldn't `WorkletRuntime` just be a class private field?

tomekzaw: Currently there's no relationship between thread/queue and worklet runtime, you can submit any worklet runtime to any background queue. The API is like this: `runOnBackgroundQueue(queue, runtime, worklet)`. We could pass the worklet runtime to `createWorkletRuntime` but then we bind runtime to the queue which is less flexible.

tjzel: Thanks, it makes more sense now!

Common/cpp/ReanimatedRuntime/WorkletRuntime.h
tomekzaw: yes, let's leave it as it is now to avoid changes in the future 😉

kmagiera (commented):
Code looks mostly good. I left some small comments inline.

Regarding the public API, I'm not completely convinced to the current shape of it though. Here is what I don't like:
1) the term 'background queue' can be confused with background execution given the main target are mobile platforms. I'd prefer to avoid using 'background' altogether and instead use 'worklet tasks queue' or something else that's distinct.
2) I'd like us to consider making this API not as functional. Queue objects can have methods for scheduling. I'd vote for prefixing the method that creates the queue with `experimental` as we chatted offline, but then its methods wouldn't need that prefix.
3) Do we still support executing worklets in runtimes w/o any queue? Maybe it should be limited to always running on some pre-allocated "main queue"?

tomekzaw:
Thanks for the review!

> 1. the term 'background queue' can be confused with background execution given the main target are mobile platforms. I'd prefer to avoid using 'background' altogether and instead use 'worklet tasks queue' or something else that's distinct.

Fair point. "Worklet tasks queue" is a bit long and worklet is a task on its own, how about "worklet queue" and `createWorkletQueue`?

Also, do we want to bind worklet queue with a specific runtime?

```ts
// independent of runtime
const queue = experimental_createWorkletQueue();
queue.push(runtime1, worklet1);
queue.push(runtime2, worklet2);

// bound to a specific runtime
const queue = experimental_createWorkletQueue(runtime);
queue.push(worklet1);
queue.push(worklet2);
```

> 2. I'd like us to consider making this API not as functional. Queue objects can have methods for scheduling. I'd vote for prefixing the method that creates the queue with experimental as we chatted offline, but then its methods wouldn't need that prefix.

Here's some options:
a) `queue.post(worklet);`
b) `queue.push(worklet);` &ndash; can be confused with a regular queue (data structure)
c) `queue.submit(worklet);`  &ndash; "submit" is usually used for executors

> 3. Do we still support executing worklets in runtimes w/o any queue? Maybe it should be limited to always running on some pre-allocated "main queue"?

We never did support running worklets synchronously directly from JS, but we expose `WorkletRuntime::runGuarded` in C++ because we use it in Reanimated internals and also in `BackgroundQueue::runLoop`. Currently, `WorkletRuntime` is a wrapper of `jsi::Runtime` so it's unaware of threads and other stuff and I would like to keep it this way.

> if I understand correctly, tasks on the queue may retain JSI runtime and other stuff, so we can't let them live past the runtime lifecycle.

That's right, currently queue holds a vector of pairs (worklet runtime, worklet) so it retains `jsi::Runtime`.

app/src/examples/WorkletRuntimeExample.tsx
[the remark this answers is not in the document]
tomekzaw: I've removed `createBackgroundQueue` and `runOnBackgroundQueue` in favor of `runOnRuntime` which creates one worklet queue automatically so it's fully transparent to the user and there's no way of using runtime concurrently using JS API. It's still possible to invoke such runtime synchronously from C++ code but it's meant to be used by advanced users.

Common/cpp/Tools/BackgroundQueue.h
tomekzaw: I don't see anything wrong about using `cv` in this case, it's the usual name for a conditional variable in many places so maybe let's leave it that way.

Common/cpp/Tools/BackgroundQueue.cpp
tomekzaw: If we do so, then it will not be possible to push new jobs to the queue because `runLoop` will hold the lock forever.

Common/cpp/Tools/BackgroundQueue.cpp
tomekzaw: Changed to `thread_.detach()`. Background queue holds a queue of jobs (`std::function`) which retain worklet runtime and worklet (as shared pointers).

Common/cpp/Tools/BackgroundQueue.cpp
tomekzaw: After some investigation, turns out that this error is never shown as there's an additional check in `runOnRuntime` with a proper message (however, without a link to the docs).

kmagiera (approved):
Looks good after applying the recent updates. Thanks for this.

Would love if you could improve the description a bit. Specifically it should mention some caveats related to memory management (i.e. runtime and queue need to remain active until the current task is over). Would also be good if you could list tests you did related to queue destructor in the test plan section.

piaskowyk (approved):

tjzel (approved):

XantreGodlike:
Now we can use this stuff to move some computations into a separate thread?

tomekzaw:
@XantreGodlike yes

XantreGodlike:
Is there a way to polyfill Web Workers using this api? Maybe it can be profitable for ecosystem

tomekzaw:
> Is there a way to polyfill Web Workers using this api? Maybe it can be profitable for ecosystem

Yes, this is my next step 😈

XantreGodlike:
Hyped

PR #5194, opened 2023-10-06
Add `performance.now` in worklet runtimes

tomekzaw:
## Summary

This PR implements `performance.now` for worklet runtimes as well as aligns its behavior in the UI runtime with the behavior in the RN runtime.

Previously, `performance.now` was affected by slow animations so it would give different results depending on the setting. This PR decouples animation code from `performance.now` in favor of a separate JSI binding `_getAnimationTimestamp` which lets us implement `performance.now` without worrying about slow animations.

1. Inject `performance.now` into worklet runtimes. There were at least 3 options to achieve this:
    * obtain `performance.now` as a `jsi::HostFunction` and copy it to worklet runtime via worklet closure
      * pros: JS-only implementation
      * cons: too high-level and dependent on other functionalities (e.g. shareables)
    * inject JSI binding which calls some function from React Native C++ codebase
      * pros: consistent behavior with RN
      * cons: need to declare a dependency and link with symbols from RN, prone to changes in the framework
    * copy C++ source code from React Native codebase into Reanimated
      * pros: stable, separate, no linking required
      * cons: code duplication
1. Inject `_getAnimationTimestamp` (previously `_getCurrentTime`) into UI runtime
    * Rename all occurrences of `getCurrentTime` to `getAnimationTimestamp`
    * Rename all occurrences of `TimeProviderFunction` to `GetAnimationTimestampFunction`
1. Use `_getAnimationTimestamp()` instead of `performance.now()` in animation code

## Test plan

BokehExample


src/reanimated2/core.ts
```diff
@@ -45,10 +45,11 @@ export const isReanimated3 = () => true;
...
 
 // this is for web implementation
-global._WORKLET = false;
-global._log = function (s: string) {
-  console.log(s);
-};
+if (shouldBeUseWeb()) {
+  global._WORKLET = false;
+  global._log = console.log;
+  global._getAnimationTimestamp = () => performance.now();
```
tomekzaw, line 51: We can't assign `performance.now` directly here because we get `TypeError: Illegal invocation` when calling `_getAnimationTimestamp`. Another solution would be to assign `performance.now.bind(performance)` (it works in runtime) but this breaks Jest tests.

piaskowyk (approved):
👌

PR #5195, opened 2023-10-06
Add new spring feature - clamp

Latropos:
## Summary

Add possibility to define limits that won't be crossed by bouncing spring.
In simplification we can assume that range of bounces is related just with damping ratio,
therefore we recalculate new damping ratio, that satisfies our expectation - the spring moves within its range.

Simplified alghoritm:
1. Calculate $\mathtt{zeta}$ (dampingRatio), $\mathtt{omega1}$ and $\mathtt{omega0}$
2. Calculate such a two values for $\mathtt{zeta}$ that the extrema match both edges of the range exactly
3. Compare the calculate $\mathtt{zeta}$ s with the provided damping ratio. Substitute the damping ratio with the biggest of the values
5. Ignore the fact that now values for $\mathtt{omega0}$ and $\mathtt{omega1}$ are outdated
   * The spring still looks like a real physical one and seems very convincing
6. We changed damping ratio, but the duration stays the same 🥳 

## Examples

Two examples with different arbitrary chosen config. 

<table>
<tr><td>Example 1</td><td>Example 2</td></tr>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/81857b94-1942-4e52-b7b3-fceb3d7d00a3

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/43ebadf6-bfc5-4cc9-862f-a47e17fa84d8

</td></tr></table>

## Test plan

<details><summary>Resting example - code</summary>

```javascript

import Animated, {
  useSharedValue,
  useAnimatedStyle,
  withSpring,
import { View, Text, Button, StyleSheet, ViewStyle } from 'react-native';

const VIOLET = '#b58df1';
const BORDER_WIDTH = 4;
const FRAME_WIDTH = 300;

const LOWER_CLAMP = 120;
const UPPER_CLAMP = 220;
function renderFramedExample(testedStyle: ViewStyle, description: string) {
  return (
    <>
      <Text style={styles.text}>{description}</Text>
      <View
        style={{
          width: FRAME_WIDTH + 2 * BORDER_WIDTH,
          borderWidth: BORDER_WIDTH,
          borderColor: VIOLET,
        }}>
        <Animated.View
          style={[
            styles.clampMarker,
            { marginBottom: -40, width: UPPER_CLAMP },
          ]}
        />
        <Animated.View style={[styles.movingBox, testedStyle]} />
        <Animated.View
          style={[
            styles.clampMarker,
            {
              marginTop: -40,
              width: FRAME_WIDTH - LOWER_CLAMP,
              alignSelf: 'flex-end',
            },
          ]}
        />
      </View>
    </>
  );
}

  const randomWidth = useSharedValue(100);

  const config = {
    duration: 5000,
    dampingRatio: 0.4,
  };

  const clampedStyle = useAnimatedStyle(() => {
    return {
      width: withSpring(randomWidth.value, {
        ...config,
        clamp: [LOWER_CLAMP, UPPER_CLAMP],
      }),
    };
  });
    return {
      width: withSpring(randomWidth.value, config),
    };
  });

  return (
      {renderFramedExample(clampedStyle, 'Clamp example')}

      {renderFramedExample(style, 'Default')}
      <Button
        title="toggle"
        onPress={() => {
          randomWidth.value = toggle ? 130 : 210;
          setToggle(!toggle);
        }}
      />
    </View>
  );
}

  container: {
    flex: 1,
    flexDirection: 'column',
    padding: 40,
  },
  clampMarker: {
    margin: 0,
    opacity: 0.5,
    height: 40,
    backgroundColor: VIOLET,
  },
  movingBox: {
    height: 100,
    opacity: 0.5,
    backgroundColor: 'black',
  },
  text: {
    fontSize: 16,
    marginVertical: 4,
  },
});
```
</details>


src/reanimated2/animation/springUtils.ts
```diff
@@ -154,6 +162,57 @@ export function initialCalculations(
...
+   *     Math.exp(-zeta * Math.PI);      (over the target)
+   *     Math.exp(-zeta * 2 * Math.PI);  (before the target)
+   */
+
+  const newZeta1 = Math.abs(Math.log(relativeExtremum1) / Math.PI);
+  const newZeta2 = Math.abs(Math.log(relativeExtremum2) / (2 * Math.PI));
+
+  // The bigger is zeta the smaller are bounces, we return the biggest one
+  // because it should satisfy all conditions
+  return Math.max(Math.max(newZeta1, newZeta2), zeta);
```
tomekzaw, line 213: `Math.max` supports multiple arguments.
```suggestion
  return Math.max(newZeta1, newZeta2, zeta);
```

src/reanimated2/animation/springUtils.ts
```diff
@@ -84,6 +87,11 @@ export function validateConfig(config: DefaultSpringConfig): boolean {
     errorLog += `duration can't be negative, got ${config.duration}, `;
   }
 
+  if (config.clamp && config.clamp[0] > config.clamp[1]) {
+    configIsInvalid = true;
+    errorLog += `clamp should have format [number1, number2], where number1 is smaller than number2, got [${config.clamp[0]}, ${config.clamp[1]}]`;
```
tomekzaw, line 92: This error message sounds too detailed, let's just tell the user that the lower bound should be not greater than the upper bound.

src/reanimated2/animation/springUtils.ts
```diff
@@ -154,6 +162,57 @@ export function initialCalculations(
...
+/** We make an assumption that we can manipulate zeta without changing duration of movement.
+ *  According to theory this change is small and tests shows that we can indeed ignore it.
+ */
+export function scaleZetaToMatchClamps(
+  animation: SpringAnimation,
+  clamp: [number, number]
+) {
+  'worklet';
+  const { zeta, toValue, startValue } = animation;
+  if (Number(toValue) === startValue) {
```
tomekzaw, line 174: Why do we convert `toValue` into a number here? Does it mean that we no longer support colors?

src/reanimated2/animation/springUtils.ts
```diff
@@ -154,6 +162,57 @@ export function initialCalculations(
...
+    return zeta;
+  }
+
+  const signum = Number(toValue) - startValue < 0 ? -1 : +1;
```
tomekzaw, line 178: Why not just `Math.sign`?

src/reanimated2/animation/springUtils.ts
```diff
@@ -154,6 +162,57 @@ export function initialCalculations(
...
+  const firstClamp = signum === +1 ? clamp[1] : clamp[0];
+  const secondClamp = signum === +1 ? clamp[0] : clamp[1];
```
tomekzaw, line 180:
```ts
const [lowerBound, upperBound] = Math.sign(toValue) - startValue > 0 ? clamp : [clamp[1], clamp[0]];
```

src/reanimated2/animation/springUtils.ts
```diff
@@ -154,6 +162,57 @@ export function initialCalculations(
...
+
+  /** The extrema we get from equation below are relative (we obtain a ratio),
+   *  To get absolute extrema we convert it as follows:
+   *
+   *  AbsoluteExtremum = startValue ± RelativeExtremum * (toValue - startValue)
+   *  Where ± denotes:
+   *    + if extremum is over the target and signum = +1
+   *    - overwise
```
tomekzaw, line 188:
```suggestion
   *    - otherwise
```

tomekzaw (changes requested):
Left some syntax-related comments.

src/reanimated2/animation/springUtils.ts
Latropos: I've just tested and it looks that color handling is a bit buggy with our main code 😱 
Something like
```javascript
    return {
      backgroundColor: withSpring(randomColor.value, config),
    };
  });
```
where randomColor is a colorname like 'red' or 'blue' doesn't always run.

Anyway toValue is expected to be a number, since in `src/reanimated2/animation/util.ts` we have a lot of functions like `colorOnStart` and `numberOnStart` that are expected to convert input to numbers and then join it again.

Latropos: Also if `toValue` weren't `number` it would be difficult wo handle it, since value is always a number.

Latropos: Btw. logic inside `src/reanimated2/animation/util.ts` is very important and complicated, I think I'd suggest splitting it into multiple files, WDYT?

app/src/examples/WithClampExample.tsx
```diff
@@ -0,0 +1,106 @@
...
+const VIOLET = '#b58df1';
+const BORDER_WIDTH = 4;
+const FRAME_WIDTH = 300;
+const LOWER_CLAMP = 120;
+const UPPER_CLAMP = 220;
+const CLAMP_MARKER_HEIGHT = 40;
+
+function renderFramedExample(testedStyle: ViewStyle, description: string) {
+  return (
+    <React.Fragment>
```
piaskowyk, line 18:
```suggestion
    <>
```

app/src/examples/WithClampExample.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const style = useAnimatedStyle(() => {
+    return {
+      width: withSpring(randomWidth.value, config),
+    };
+  });
+
+  return (
+    <View style={styles.container}>
+      {renderFramedExample(clampedStyle, 'Clamp example')}
```
piaskowyk, line 73:
```suggestion
      {renderFrametExample(clampedStyle, 'Clamp example')}
```
or
```suggestion
      {renderExample(clampedStyle, 'Clamp example')}
```

piaskowyk:
What is the difference between example 1 and 2?

src/reanimated2/animation/spring.ts
```diff
@@ -49,6 +50,7 @@ export const withSpring = ((
       duration: 2000,
       dampingRatio: 0.5,
       reduceMotion: undefined,
+      clamp: undefined,
```
piaskowyk, line 53: Can we simply avoid defining it here?

src/reanimated2/animation/springUtils.ts
```diff
@@ -151,6 +158,56 @@ export function initialCalculations(
...
+  clamp: [number, number]
+) {
+  'worklet';
+  const { zeta, toValue, startValue } = animation;
+  if (Number(toValue) === startValue) {
+    return zeta;
+  }
+
+  const [lowerBound, upperBound] =
+    Number(toValue) - startValue > 0 ? clamp : [clamp[1], clamp[0]];
```
piaskowyk, line 175: You can cast `toValue` to a number once and assign it to a new variable, rather than casting it every time you use it.

piaskowyk:
You need to update documentation also

app/src/examples/WithClampExample.tsx
Latropos: Add an eslint rule?
``` diff
+     "react/jsx-fragments": ['error', 'syntax']
```

src/reanimated2/animation/spring.ts
Latropos: No, we can't do it without changing our types

app/src/examples/WithClampExample.tsx
tomekzaw: Yes, but make sure that `<React.Fragment key={key}>` works (you can't use `key` with `<></>`).

Latropos: It works with key 💪

Latropos:
@piaskowyk 
> You need to update documentation also

Yes, however I've found out that we should make more changes to the documentation, so I would prefer to do it in separate PR: https://github.com/software-mansion/react-native-reanimated/pull/5340

app/src/examples/WithClampExample.tsx
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
+import Animated, {
+  useSharedValue,
+  useAnimatedStyle,
+  withSpring,
+} from 'react-native-reanimated';
+import { View, Text, Button, StyleSheet, ViewStyle } from 'react-native';
+
+const VIOLET = '#b58df1';
```
Latropos, line 9: This example is very similar to https://github.com/software-mansion/react-native-reanimated/pull/5239/files#r1381390672
Once one of this two PRs gets merged (this one and #5195 creating Higher Order Animation WithClamp) I should unify these examples.

piaskowyk (commented):
Looks good to me 👍 We'll test it together in the office and we can merge it.

app/.eslintrc.js
```diff
@@ -9,5 +9,6 @@ module.exports = {
     'react-native/no-unused-styles': 'error',
     'react-native/no-raw-text': 'off',
     'react-native/no-single-element-style-arrays': 'error',
+    'react/jsx-fragments': ['error', 'syntax'],
```
tomekzaw, line 12: why do we need to have both `error` and `syntax` here and not only `error`?

app/src/examples/PendulumExample.tsx
```diff
@@ -166,20 +166,18 @@ export default function SpringExample() {
...
-        })}
-      </React.Fragment>
+      {fields.map((item) => {
+        return (
+          <InputField
+            fieldName={item.fieldName}
+            value={item.value}
+            setValue={(value) => {
+              item.setValue(value);
+            }}
```
tomekzaw, line 176: does it work this way?
```suggestion
            setValue={item.setValue}
```

app/.eslintrc.js
Latropos: We have to choose whether we prefer `<>` or `<React.Fragment>`, the rule has form `"react/jsx-fragments": [<enabled>, <mode>]` https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-fragments.md

tomekzaw: it's okay then, thanks for explanation!

piaskowyk (approved):

app/src/examples/WithClampExample.tsx
```diff
@@ -27,23 +29,47 @@ function renderExample(testedStyle: ViewStyle, description: string) {
           borderWidth: BORDER_WIDTH,
           borderColor: VIOLET,
         }}>
-        <Animated.View
-          style={[
-            styles.clampMarker,
-            { marginBottom: -CLAMP_MARKER_HEIGHT, width: UPPER_BOUND },
-          ]}
-        />
+        <View>
```
tomekzaw, line 32: Do we need this wrapping view? Can we remove it or use `<>` instead?

app/src/examples/WithClampExample.tsx
```diff
@@ -27,23 +29,47 @@ function renderExample(testedStyle: ViewStyle, description: string) {
...
-          style={[
-            styles.clampMarker,
-            {
-              marginTop: -CLAMP_MARKER_HEIGHT,
-              width: FRAME_WIDTH - LOWER_BOUND,
-              alignSelf: 'flex-end',
-            },
-          ]}
-        />
+        <View>
```
tomekzaw, line 51: same question here

app/src/examples/WithClampExample.tsx
```diff
@@ -59,20 +85,20 @@ export default function AnimatedStyleUpdateExample() {
...
 
-  const clampedStyle = useAnimatedStyle(() => {
+  const clampedStyleWithHoc = useAnimatedStyle(() => {
     return {
       width: withClamp(
         { min: LOWER_BOUND, max: UPPER_BOUND },
         withSpring(randomWidth.value, config)
       ),
     };
   });
```
tomekzaw, line 95: add a newline
```suggestion
  });

```

app/src/examples/WithClampExample.tsx
```diff
@@ -59,20 +85,20 @@ export default function AnimatedStyleUpdateExample() {
     dampingRatio: 0.075,
   };
 
-  const clampedStyle = useAnimatedStyle(() => {
+  const clampedStyleWithHoc = useAnimatedStyle(() => {
```
tomekzaw, line 88: What's "hoc" in this context? I know it stands for "higher-order component" but why this appears here for animated style?

src/reanimated2/animation/spring.ts
```diff
@@ -209,6 +211,10 @@ export const withSpring = ((
         animation.zeta = zeta;
         animation.omega0 = omega0;
         animation.omega1 = omega1;
+
+        if (config.clamp !== undefined) {
+          animation.zeta = scaleZetaToMatchClamps(animation, config.clamp);
```
tomekzaw, line 216: Can we add some unit tests for this function `scaleZetaToMatchClamps`?

tomekzaw (changes requested):

app/src/examples/WithClampExample.tsx
Latropos: We can't use `<>` instead, this view here is needed for absolute styling,

app/src/examples/WithClampExample.tsx
Latropos: I've though of higher-order-animation and made typo. However, I'll call it "animation modifier" to match our docs

src/reanimated2/animation/springUtils.ts
```diff
@@ -96,14 +99,24 @@ export function checkIfConfigIsValid(config: DefaultSpringConfig): boolean {
     errorMessage += `, duration can't be negative, got ${config.duration}`;
   }
 
+  if (
+    config?.clamp &&
```
tomekzaw, line 103: Do we need `?.` here?

piaskowyk:
I've tested it and it seems to work 🎉

Issue #5165, opened 2023-09-29
Broken symbolication for worklet functions


gaearon:
### Description

If a worklet function crashes, I get useless stack traces that point to the wrong place.

![F7HqFv8XwAATCb3](https://github.com/software-mansion/react-native-reanimated/assets/810438/b09067cb-3f12-4623-9b81-cb6e63cbb3b7)

### Steps to reproduce

1. Write `blablabla;` inside a worklet function
2. Observe broken stack and wrong line numbers in the redbox

### Snack or a link to a repository

sorry

### Reanimated version

3.4.2

### React Native version

0.72.4

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

PR #5197, opened 2023-10-06
Fix stack trace on worklet factory crash

tjzel:
## Summary

Fixes: #5165 

Since a worklet factory doesn't have it's counterpart in source code, when it crashes we can get a useless stack trace. It can crash when a variable that doesn't exists is referenced - since it's pulled into the worklet's factory closure. This PR fixes the stack trace on such crashes.

As a bonus I remove plugin's maps from the npm package.

## Test plan

<details>
<summary>
Snippet that forces the factory crash
</summary>

```TSX
//
//
//
//
//
//

  function fooOOO() {
    'worklet';
    stackOrderAscendinggg;
  }

  return null;
}
```
</details>

| Before | After |
|--------|--------|
|Simulator Screenshot - iPhone 14 Pro - 2023-10-06 at 19 23 48|Simulator Screenshot - iPhone 14 Pro - 2023-10-06 at 19 23 16|


plugin/src/makeWorklet.ts
```diff
@@ -319,6 +319,7 @@ function makeArrayFromCapturedBindings(
   fun: NodePath<WorkletizableFunction>
 ) {
   const closure = new Map<string, Identifier>();
+  const isLocationSet = new Map<string, boolean>();
```
tomekzaw, line 322: not really happy about this name, first I thought it's a `Set` (maybe it could be a set since we only assign booleans?)

package.json
```diff
@@ -60,7 +60,7 @@
     "scripts/reanimated_utils.rb",
     "mock.js",
     "plugin/index.js",
-    "plugin/build/",
+    "plugin/build/plugin.js",
```
tomekzaw, line 63: I know it's a simple change but maybe it would be better to have it as a separate PR?

plugin/src/makeWorklet.ts
```diff
@@ -368,6 +369,25 @@ function makeArrayFromCapturedBindings(
         currentScope = currentScope.parent;
       }
       closure.set(name, path.node);
+      isLocationSet.set(name, false);
+    },
+  });
+
+  /*
+  For reasons I don't exactly understand, the above traversal will
+  crash the whole bundle if we traversed original node instead of generated
```
tomekzaw, line 378: what do you mean by "the above traversal will crash the whole bundle"?

tjzel: The bundling of Reanimated Example (at least) crashes during bundling with a very vague error.

package.json
tjzel: We can.

plugin/src/makeWorklet.ts
tjzel: I think it was a set originally but then I changed it to the map and didn't update the name. I think the code is simpler with a Map, but if you prefer a Set I'm happy to listen to suggestions.

tomekzaw: Sounds good

plugin/src/makeWorklet.ts
tomekzaw: Suggestion: replace with "the above traversal will cause the whole bundle to crash"

piaskowyk:
Is it possible to add tests for this feature? 🤔

tjzel:
I could try, but this could be hacky 💻 🟢

tjzel:
Unfortunately, after many attempts, an Example is the best I can do in this situation. 😢 @piaskowyk

app/src/examples/WorkletFactoryCrashExample.tsx
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }
+
+  return (
+    <View style={styles.container}>
+      <Text style={styles.text}>
+        Clicking the button below should give a stack trace that denotes the
+        error in `badWorklet` function, pointing to `unexistingVariable` usage.
+      </Text>
+      <Button title={'Crash me'} onPress={handleOnPress} />
```
tomekzaw, line 20:
```suggestion
      <Button title="Crash me" onPress={handleOnPress} />
```

piaskowyk (approved):

app/src/examples/WorkletFactoryCrashExample.tsx
```diff
@@ -0,0 +1,34 @@
...
+      // @ts-expect-error
+      unexistingVariable;
+    }
+  }
+
+  return (
+    <View style={styles.container}>
+      <Text style={styles.text}>
+        Clicking the button below should give a stack trace that denotes the
+        error in `badWorklet` function, pointing to `unexistingVariable` usage.
```
tomekzaw, line 18: Instead of providing the instructions here, could we `try/catch` the exception and check the stack trace (as a string) automatically? If this works, let's just `console.log('OK')`.

tjzel: @tomekzaw Unfortunately this cannot be done like this. The stack trace is an unrefined error that points to the bundled code instead of the source code. We need the redbox to do its magic and point to the correct place.

tomekzaw (approved):

tomekzaw (approved):

PR #5206, opened 2023-10-10
Curve animation

Latropos:
## Summary

Right now we support curve animation as one of layout animation type. In this PR I am going to make it easy to use curve animation as a style animation too.

- ✅ Both curve animations (style animations & layout animation) have the same effect
- ❌ You don't have intuitive API to springify the curve animation (but this is true for layout animation as well).
 
## Suggested future changes

* Change file structure inside 📁animation
  * separate folders for first order and higher order animations
  * Split utils into multiple files
 

## Test plan

I've made sure that Layout Animation and animatedStyle Animation have the same look:

https://github.com/software-mansion/react-native-reanimated/assets/56199675/cb987c25-c394-41e3-a1ac-4362ae73c898


app/src/examples/CurveAnimationExample.tsx
```diff
@@ -0,0 +1,80 @@
...
+  useSharedValue,
+  withTiming,
+  useAnimatedStyle,
+  Easing,
+  CurvedTransition,
+} from 'react-native-reanimated';
+import { View, Button, StyleSheet } from 'react-native';
+import React, { useEffect, useState } from 'react';
+
+//TODO Make this function a parto of our API
```
piaskowyk, line 11: This is the goal of this PR 😅 so let's do it

piaskowyk (changes requested):

Issue #4505, opened 2023-05-30
[Android] The app crashes Unable to update properties for view tag XX. com.facebook.react.uimanager.IllegalViewOperationException: ViewManager for tag 75 could not be found.


SarojnetObjex:
`Unable to update properties for view tag 75
                                                                                                    com.facebook.react.uimanager.IllegalViewOperationException: ViewManager for tag 75 could not be found.
                                                                                                    
                                                                                                    	at com.facebook.react.uimanager.NativeViewHierarchyManager.resolveViewManager(NativeViewHierarchyManager.java:111)
                                                                                                    	at com.facebook.react.uimanager.NativeViewHierarchyManager.updateProperties(NativeViewHierarchyManager.java:138)
                                                                                                    	at com.facebook.react.uimanager.UIImplementation.synchronouslyUpdateViewOnUIThread(UIImplementation.java:306)
                                                                                                    	at com.swmansion.reanimated.NodesManager.updateProps(NodesManager.java:589)
                                                                                                    	at com.swmansion.reanimated.NativeProxy.updateProps(NativeProxy.java:185)
                                                                                                    	at com.swmansion.reanimated.NativeProxy$AnimationFrameCallback.onAnimationFrame(Native Method)
                                                                                                    	at com.swmansion.reanimated.NodesManager.onAnimationFrame(NodesManager.java:274)
                                                                                                    	at com.swmansion.reanimated.NodesManager.access$000(NodesManager.java:68)
                                                                                                    	at com.swmansion.reanimated.NodesManager$1.doFrameGuarded(NodesManager.java:173)
                                                                                                    	at com.facebook.react.uimanager.GuardedFrameCallback.doFrame(GuardedFrameCallback.java:29)
                                                                                                    	at com.facebook.react.modules.core.ReactChoreographer$ReactChoreographerDispatcher.doFrame(ReactChoreographer.java:175)
                                                                                                    	at com.facebook.react.modules.core.ChoreographerCompat$FrameCallback$1.doFrame(ChoreographerCompat.java:85)
                                                                                                    	at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1299)
                                                                                                    	at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1309)
                                                                                                    	at android.view.Choreographer.doCallbacks(Choreographer.java:923)
                                                                                                    	at android.view.Choreographer.doFrame(Choreographer.java:847)
[...]

Issue #4231, opened 2023-03-15
RELEASE ONLY: E/unknown:NativeViewHierarchyManager: Unable to update properties for view tag xxxx


frankoprifti:
### Description

On the Release build this error is happening a lot. The modals on my app are not opening a lot of times. The error can be seen in the logcat: 
`
E/unknown:NativeViewHierarchyManager: Unable to update properties for view tag 3395
    com.facebook.react.uimanager.IllegalViewOperationException: ViewManager for tag 3395 could not be found.
        at com.facebook.react.uimanager.NativeViewHierarchyManager.resolveViewManager(NativeViewHierarchyManager.java:111)
        at com.facebook.react.uimanager.NativeViewHierarchyManager.updateProperties(NativeViewHierarchyManager.java:138)
        at com.facebook.react.uimanager.UIImplementation.synchronouslyUpdateViewOnUIThread(UIImplementation.java:281)
        at com.swmansion.reanimated.NodesManager.updateProps(NodesManager.java:354)
        at com.swmansion.reanimated.nativeProxy.NativeProxyCommon.updateProps(NativeProxyCommon.java:97)
        at com.swmansion.reanimated.nativeProxy.AnimationFrameCallback.onAnimationFrame(Native Method)
        at com.swmansion.reanimated.NodesManager.onAnimationFrame(NodesManager.java:245)
        at com.swmansion.reanimated.NodesManager.access$000(NodesManager.java:42)
        at com.swmansion.reanimated.NodesManager$1.doFrameGuarded(NodesManager.java:142)
        at com.facebook.react.uimanager.GuardedFrameCallback.doFrame(GuardedFrameCallback.java:29)
        at com.facebook.react.modules.core.ReactChoreographer$ReactChoreographerDispatcher.doFrame(ReactChoreographer.java:175)
        at com.facebook.react.modules.core.ChoreographerCompat$FrameCallback$1.doFrame(ChoreographerCompat.java:85)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:970)
        at android.view.Choreographer.doCallbacks(Choreographer.java:796)
        at android.view.Choreographer.doFrame(Choreographer.java:727)
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:957)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:223)
[...]

Issue #3860, opened 2022-12-13
NativeViewHierarchyManager: Unable to update properties for view tag 12187


OrenMe:
### Description

App crashes and I see this appearing in our Sentry hitting many end users.
```
E/unknown:NativeViewHierarchyManager: Unable to update properties for view tag 12187
    com.facebook.react.uimanager.IllegalViewOperationException: ViewManager for tag 12187 could not be found.
    
        at com.facebook.react.uimanager.NativeViewHierarchyManager.resolveViewManager(NativeViewHierarchyManager.java:111)
        at com.facebook.react.uimanager.NativeViewHierarchyManager.updateProperties(NativeViewHierarchyManager.java:138)
        at com.facebook.react.uimanager.UIImplementation.synchronouslyUpdateViewOnUIThread(UIImplementation.java:305)
        at com.swmansion.reanimated.NodesManager.updateProps(NodesManager.java:588)
        at com.swmansion.reanimated.NativeProxy.updateProps(NativeProxy.java:167)
        at com.swmansion.reanimated.NativeProxy$AnimationFrameCallback.onAnimationFrame(Native Method)
        at com.swmansion.reanimated.NodesManager.onAnimationFrame(NodesManager.java:273)
        at com.swmansion.reanimated.NodesManager.access$000(NodesManager.java:65)
        at com.swmansion.reanimated.NodesManager$1.doFrameGuarded(NodesManager.java:170)
        at com.facebook.react.uimanager.GuardedFrameCallback.doFrame(GuardedFrameCallback.java:29)
        at com.facebook.react.modules.core.ReactChoreographer$ReactChoreographerDispatcher.doFrame(ReactChoreographer.java:175)
        at com.facebook.react.modules.core.ChoreographerCompat$FrameCallback$1.doFrame(ChoreographerCompat.java:85)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:964)
        at android.view.Choreographer.doCallbacks(Choreographer.java:790)
        at android.view.Choreographer.doFrame(Choreographer.java:721)
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:951)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #5216, opened 2023-10-11
fix(android): Unable to update properties for view tag

MasonLe2497:
### Platform: Android

I catch this issue when use [FlashList](https://shopify.github.io/flash-list/) with large data and have animated update on each item.
One element can be update/replace/recycle, so view tag can be update/remove too
So, before update props, we check current view tag exist in UIManager or not.

Fix [#4505](https://github.com/software-mansion/react-native-reanimated/issues/4505)
Fix [#4231](https://github.com/software-mansion/react-native-reanimated/issues/4231)
Fix [#3860](https://github.com/software-mansion/react-native-reanimated/issues/3860)

## Test plan
Before:

https://github.com/software-mansion/react-native-reanimated/assets/43195241/8f24027e-01d1-4af0-9bf6-8f611a14c6fa

After:

https://github.com/software-mansion/react-native-reanimated/assets/43195241/ccf439b9-8c76-4448-ad83-aaeae91e032b

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -344,6 +344,19 @@ public class NodesManager implements EventDispatcherListener {
   }
 
   public void updateProps(int viewTag, Map<String, Object> props) {
+    // We need to check current view exist in UIManager or not
+    // Sometime, view can be replace, recycle, ... with wrong way (like FlashList)
+    // So before update, We need to check view tag exist in UIManager
+    try {
+        View view = mUIManager.resolveView(viewTag);
+        if(view == null) {
+          return;
+        }
+    } catch (Exception ex) {
+        Log.d("Reanimated-updateProps", "Skip update props cause viewTag not exist or have been removed!");
+        return;
+    }
+
     // TODO: update PropsNode to use this method instead of its own way of updating props
     boolean hasUIProps = false;
     boolean hasNativeProps = false;


efstathiosntonas:
Hi @MasonLe2497 `Log` needs to be imported: 

`NodesManager.java:356: error: cannot find symbol
      Log.d("Reanimated-updateProps", "Skip update props cause viewTag not exist or have been removed!");`

tomekzaw:
@MasonLe2497 Thanks for submitting the PR! I agree we should do something about this error. Instead of checking if the view exists before updating props, maybe we could wrap `mUIImplementation.synchronouslyUpdateViewOnUIThread` and `mUIManager.updateView` calls into `try/catch` blocks? This way we can also avoid the error in case when the view exists when calling `updateProps` on the UI thread but no longer exists when calling `mUIManager.updateView` on the native modules thread.

MasonLe2497:
> @MasonLe2497 Thanks for submitting the PR! I agree we should do something about this error. Instead of checking if the view exists before updating props, maybe we could wrap `mUIImplementation.synchronouslyUpdateViewOnUIThread` and `mUIManager.updateView` calls into `try/catch` blocks? This way we can also avoid the error in case when the view exists when calling `updateProps` on the UI thread but no longer exists when calling `mUIManager.updateView` on the native modules thread.

Ok. I'll try tomorrow.(from UTC +7:00)

MasonLe2497:
@tomekzaw i tried to wrap try/catch to mUIImplementation.synchronouslyUpdateViewOnUIThread, but i still catch the issue.
I think synchronouslyUpdateViewOnUIThread run on different thread, so we cant catch nothing here

https://github.com/software-mansion/react-native-reanimated/assets/43195241/6bf0b55a-73c3-4ba5-9156-1d2ee667342a

MasonLe2497:
> Hi @MasonLe2497, `Log` needs to be imported:
> 
> `NodesManager.java:356: error: cannot find symbol Log.d("Reanimated-updateProps", "Skip update props cause viewTag not exist or have been removed!");`

Thanks, I just update my PR

ranaavneet:
Hey @tomekzaw, is there anything else we can do about this?

damianbeles:
@tomekzaw any update?

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -344,6 +344,19 @@ public void sendEvent(String name, WritableMap body) {
   }
 
   public void updateProps(int viewTag, Map<String, Object> props) {
+    // We need to check current view exist in UIManager or not
+    // Sometime, view can be replace, recycle, ... with wrong way (like FlashList)
+    // So before update, We need to check view tag exist in UIManager
```
tomekzaw, line 349: Let's rephrase this comment

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -344,6 +344,19 @@ public void sendEvent(String name, WritableMap body) {
...
+    try {
+        View view = mUIManager.resolveView(viewTag);
+        if(view == null) {
+          return;
+        }
```
tomekzaw, line 354:
```suggestion
        if (view == null) {
          return;
        }
```

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -344,6 +344,19 @@ public void sendEvent(String name, WritableMap body) {
...
+    } catch (Exception ex) {
```
tomekzaw, line 355:
```suggestion
    } catch (Exception e) {
```

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -344,6 +344,19 @@ public void sendEvent(String name, WritableMap body) {
...
+        Log.d("Reanimated-updateProps", "Skip update props cause viewTag not exist or have been removed!");
```
tomekzaw, line 356: Let's remove this warning so that we don't pollute the logs.
```suggestion
```

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -344,6 +344,19 @@ public void sendEvent(String name, WritableMap body) {
...
+        return;
```
tomekzaw, line 357: Can you please describe in which cases the exception will be thrown?

tomekzaw (changes requested):

damianbeles:
@ngocle2497 do you plan to review the change requests? We really need this in prod. If not, let me know and I'll do it.

joe-sam (changes requested):
Include valueUnpackerCode argument for latest API interface compatibility

```
public void initWithContext(ReactApplicationContext reactApplicationContext, String valueUnpackerCode) {
   mReactApplicationContext = reactApplicationContext;
   mNativeProxy = new NativeProxy(reactApplicationContext, valueUnpackerCode);
   mAnimationManager.setAndroidUIScheduler(getNativeProxy().getAndroidUIScheduler());
   compatibility = new ReaCompatibility(reactApplicationContext);
   compatibility.registerFabricEventListener(this);
 }
```

thomas-rx:
@damianbeles  @ngocle2497 I opened ticket #5767 to address that important error.

PR #5218, opened 2023-10-12
Bump react-native to 0.72.6

tomekzaw:
## Summary

This PR bumps react-native to 0.72.6 in order to fix issues with building with Xcode 15 as well as updates native third-party dependencies to latest versions.


Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
```diff
@@ -653,6 +654,13 @@
 					"-DFOLLY_MOBILE=1",
 					"-DFOLLY_USE_LIBCPP=1",
 				);
+				OTHER_LDFLAGS = (
+					"$(inherited)",
+					"-Wl",
+					"-ld_classic",
+					" ",
+					"-Wl -ld_classic ",
```
piaskowyk, line 662: Just to be sure: Are these changes auto-generated or you edited this file manually?

piaskowyk (approved):
👍

Example/ios/ReanimatedExample.xcodeproj/project.pbxproj
tomekzaw: Yes, these changes were auto-generated during `pod install` when I run it again. For some reason, it didn't work for me before and this was the reason why CIs were failing. I decided to run `rm -rf Pods Podfile.lock build` and `pod install` and then I noticed that this file has been changed, so I pushed the changes.

PR #5223, opened 2023-10-12
Add shareable ArrayBuffer, typed arrays and DataView

tomekzaw:
## Summary

This PR adds support for shareable `ArrayBuffer` as well as makes it possible to pass typed arrays or `DataView` into a worklet on top of it.

## Test plan

ShareablesExample.tsx

### DIFF
--- a/Common/cpp/SharedItems/Shareables.h
+++ b/Common/cpp/SharedItems/Shareables.h
@@ -84,6 +84,7 @@ class Shareable {
     SynchronizedDataHolder,
     HostObjectType,
     HostFunctionType,
+    ArrayBufferType,
   };
 
   explicit Shareable(ValueType valueType) : valueType_(valueType) {}
@@ -221,6 +222,20 @@ class ShareableHostFunction : public Shareable {
   jsi::HostFunctionType hostFunction_;
   std::string name_;
   unsigned int paramCount_;
+};
+
+class ShareableArrayBuffer : public Shareable {
+ public:
+  ShareableArrayBuffer(jsi::Runtime &rt, const jsi::ArrayBuffer &arrayBuffer)
+      : Shareable(ArrayBufferType),
+        data_(
+            arrayBuffer.data(rt),
+            arrayBuffer.data(rt) + arrayBuffer.size(rt)) {}
+
+  jsi::Value toJSValue(jsi::Runtime &rt) override;
+
+ protected:
+  const std::vector<uint8_t> data_;
 };
 
 class ShareableWorklet : public ShareableObject {
--- a/app/src/examples/ShareablesExample.tsx
+++ b/app/src/examples/ShareablesExample.tsx
@@ -1,0 +1,62 @@
+import { Button, StyleSheet, View } from 'react-native';
+
+import React from 'react';
+import { runOnUI } from 'react-native-reanimated';
+
+export default function ShareablesExample() {
+  return (
+    <View style={styles.container}>
+      <CyclicObjectDemo />
+      <InaccessibleObjectDemo />
+      <ArrayBufferDemo />
+    </View>
+  );
+}
+
+function CyclicObjectDemo() {
+  const handlePress = () => {
+    type RecursiveArray = (number | RecursiveArray)[];
+    const x: RecursiveArray = [];
+    x.push(1);
+    x.push(x);
+    runOnUI(() => {
+      console.log(x);
+    })();
+  };
+
+  return <Button title="Cyclic object" onPress={handlePress} />;
+}
+
+function InaccessibleObjectDemo() {
+  const handlePress = () => {
+    const x = new Set();
+    runOnUI(() => {
+      console.log(x);
+    })();
+  };
+
+  return <Button title="Inaccessible object" onPress={handlePress} />;
+}
+
+function ArrayBufferDemo() {
+  const handlePress = () => {
+    const ab = new ArrayBuffer(8);
+    const ta = new Uint8Array(ab);
+    ta[7] = 42;
+    runOnUI(() => {
+      console.log(ab instanceof ArrayBuffer);
+      const ta = new Uint8Array(ab);
+      console.log(ta[7]);
+    })();
+  };
+
+  return <Button title="ArrayBuffer" onPress={handlePress} />;
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+});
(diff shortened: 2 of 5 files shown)


Common/cpp/SharedItems/Shareables.h
```diff
@@ -223,6 +224,28 @@ class ShareableHostFunction : public Shareable {
...
+      : Shareable(ArrayBufferType),
+        data_(
+            arrayBuffer.data(rt),
+            arrayBuffer.data(rt) + arrayBuffer.size(rt)) {
+  }
+
+  jsi::Value toJSValue(jsi::Runtime &rt) override;
+
+ protected:
+  const std::vector<uint8_t> data_;
```
piaskowyk, line 246: Could you confirm that `uint8_t` works as well for example with `Float64Array`?

app/src/examples/ShareablesExample.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const dv = new DataView(buffer);
+    dv.setInt16(7, 12345);
+    runOnUI(() => {
+      console.log(dv instanceof DataView);
+      console.log(dv.getInt16(7) === 12345);
+    })();
+  };
+
+  return <Button title="DataView" onPress={handlePress} />;
+}
```
piaskowyk, line 88: Could you add another example that uses all data types listed in  VALID_ARRAY_VIEWS_NAMES? It could be simpletes. For example create a specific buffer and use it in the worklet.

Common/cpp/SharedItems/Shareables.h
tomekzaw: `uint8_t` is often used to represent a byte so effectively `std::vector<uint8_t>` is just an array of bytes. We never access or modify the data from C++ side, we just treat it as a binary buffer. Theoretically, it should work, but I will confirm it by adding more examples as you suggested in the next comment.

piaskowyk: Thank you 😊 I also think that is should work, but it is better to check to be sure.

kmagiera:
@wcandillon Another option would be to wait little bit until this is merged, and then use the nightly release (i.e. installing from npm with `nightly` tag)

app/src/examples/ShareablesExample.tsx
tomekzaw: Done in ebc26986c454e7a4b4637f9a08d54905d7193862

piaskowyk (approved):
💪

app/src/examples/ShareablesExample.tsx
wcandillon: Do you have something that opens the example pages to see if they crash or not as a test? (asking because I'd be interested to setup something like this maybe in RN Skia (we do have e2e tests separately)).

tomekzaw: Not yet, but recently we've been talking with @piaskowyk about adding it. In our case, we could also overwrite the implementation `_getAnimationTimestamp()` function and test the animations in a frame-by-frame manner. We couldn't find any simple way for testing whole features since they are implemented in multiple programming languages at once and also heavily depend on the RN itself, so we would need some kind of E2E framework to run the feature tests. I've experimeneted with E2E tests for Reanimated in the past (see https://github.com/tomekzaw/reanimated-e2e/tree/appium) but they were too flaky and time-consuming so we didn't add them to our CIs.

piaskowyk: Not yet, but we're working on something like that. But now we do it manually before every release 😅

wcandillon: On the Skia side, we found e2e frameworks to be also too "flaky" (more specifically we found that they were many too many assumptions and giving us enough control over the setup) and we have built our own e2e solution using the WebSocket API. This has worked really well for us.

1. People write a function in Jest: https://github.com/Shopify/react-native-skia/blob/main/package/src/renderer/__tests__/e2e/Point.spec.tsx#L8 and behind the scene we sent the source code of the function over the write to the RN app (the function needs to be pure).
2. On our example app, we have a screen that just waits and see if it receives tests to be executed on its side: https://github.com/Shopify/react-native-skia/blob/main/example/src/Tests/useClient.ts
3. Write now we use it to eval JS code on the RN side or doing a Skia drawing and returns an image. But we are thinking of extending it with more complex commands.

wcandillon:
I tried to use the PR but it doesn't seem like it's possible to return a typed array as a derived or shared value. Is it possible also that the cyclic object example as a regression on main? It seems to crash.

tomdoeslinux:
Seems to be issues with this

wcandillon:
To give a bit of context there, in RN Skia we can expose native objects quite easily however most of the types we expose could easily be replaced by a typed array. This actually makes us think about exposing on own native object that would be very similar to a float32array. It feels a bit hacky.

tomekzaw:
Thanks for your comments, I will try to reproduce the following scenario:
> it doesn't seem like it's possible to return a typed array as a derived or shared value

tomekzaw:
> Is it possible also that the cyclic object example as a regression on main? It seems to crash.

At first sight it looks like it's working correctly, the following error shows up (as expected):

tomekzaw:
The following example shows that the derived value correctly stores `Float32Array` in the UI runtime but once you access it in the RN runtime, it collapses into a regular JS object with keys `0`, `1`, `2` and `3`:

```ts
function DerivedValueDemo() {
  const sv = useSharedValue(0);

  const dv1 = useDerivedValue(() => {
    const ta = new Float32Array(4);
    ta[0] = sv.value;
    ta[1] = Math.random();
    ta[2] = Math.random();
    ta[3] = Math.random();
    return ta;
  }, [sv]);

  const dv2 = useDerivedValue(() => {
    if (!_WORKLET) {
      return;
    }
    console.log('UI', dv1.value instanceof Float32Array, dv1.value);
    return dv1.value;
  }, [dv1]);

  const handlePress = () => {
    sv.value += 1;
    console.log('JS', dv2.value instanceof Float32Array, dv2.value);
  };

  return <Button title="Derived value" onPress={handlePress} />;
}
```

cc @wcandillon In your use case, do you need to access `.value` property in the RN runtime or you just use it in the UI runtime?

wcandillon:
Is it possible to have the typed array as a shared value? or a trick to have it live only the the UI thread only would be good enough. the trick to return if on the JS thread is a nice one.

tomekzaw:
Yes, you can have a typed array in a shared value and use it on the UI thread only. The missing part is sending the typed array from UI to JS thread, e.g. if you `console.log` it or access `.value` property in the RN runtime.

PR #5232, opened 2023-10-13
Fix ./createNPMPackage

tjzel:
## Summary

In #5152 I made the change that uses node script before `yarn` was called and dependencies required for `set-reanimated-version.js` were not yet installed. This PR fixes this.


piaskowyk (approved):

piaskowyk:
I will merge it because we need to fix publishing of nightly builds.

PR #5233, opened 2023-10-13
Proposal for Adding Custom Easing Function

alabsi91:
# Introduction
React Native Reanimated already includes a wide range of easing functions that is enough to most user needs. However, introducing custom easing capabilities would be a valuable addition, particularly for advanced animations. I understand that you might question its necessity, I would appreciate your approval and input before initiating any changes

# Summary
The new `customEasing` function will accept a string representing an SVG path, The API might look something like this:
```ts
const path = "M 0 1 C 0.2 1 0.365 0 0.365 0 C 0.365 0 0.45 0.255 0.546 0.25 C 0.642 0.245 0.727 0 0.727 0 C 0.727 0 0.77 0.063 0.816 0.063 C 0.862 0.063 0.91 0 0.91 0 C 0.91 0 0.925 0.016 0.955 0.016 C 0.985 0.016 1 0 1 0";
withTiming(width, {
  easing: Easing.customEasing(path),
})
```

## Requirements

The SVG path string should adhere to the following conditions:

- The viewBox for the SVG should be defined as `"0 0 1 1"`, resulting in points ranging from 0 to 1 inside the box.
- The SVG path should always start with the `"M"` command, followed by `"C"` commands. No other commands are allowed, and relative values are not permitted.

# How it works:

The concept is simple. For example, for a single `"C"` command, similar to what we use in CSS with the `cubic-bezier` function, we map it to a `cubic-bezier` function. However, for more than one `"C"` command, we divide the animation into segments, each with its own `cubic-bezier` function. We then use the appropriate function based on the progress value.
Frame 4

# UI Easing Visualizer Tool

I believe implementing such a tool is essential because not everyone is comfortable reading SVG paths. Visualizing and editing the curves will likely encourage more users to make use of it. I've been working on a similar tool that can draw and test animations. You can also paste SVG paths from other sources, and they will be automatically converted to the correct format.

👉 Please check it out [Easing Visualizer](https://animare-ease-visualizer.netlify.app/).

# Example

Here's a very simple example of custom easing. Please note that in this example, pre-generated points are used, but in the implementation, the points will be calculated in real-time.
👉 [Expo Snack](https://snack.expo.dev/@alabsi91/creating-custom-easing-in-react-native).

### DIFF
--- a/src/reanimated2/CustomEasing.ts
+++ b/src/reanimated2/CustomEasing.ts
@@ -1,0 +1,1 @@
+// Waiting for approval


piaskowyk:
Hey @alabsi91 Thanks for you proposal! Interesting idea! I'll discuss it with the rest of team and let you know what we can do with it 🫡

alabsi91:
Any new updates?

PR #5237, opened 2023-10-16
Add shareable BigInt

tomekzaw:
## Summary

This PR adds support for passing instances of `BigInt` to worklet runtimes.

## Test plan

ShareablesExample &rarr; BigIntDemo

```ts
const bigint = BigInt('1234567890');
runOnUI(() => {
  console.log(typeof bigint === 'bigint');
  console.log(bigint === BigInt('1234567890'));
})();
```

### DIFF
--- a/Common/cpp/SharedItems/Shareables.cpp
+++ b/Common/cpp/SharedItems/Shareables.cpp
@@ -92,6 +92,10 @@ jsi::Value makeShareableClone(
     shareable = std::make_shared<ShareableScalar>(value.getBool());
   } else if (value.isNumber()) {
     shareable = std::make_shared<ShareableScalar>(value.getNumber());
+#if REACT_NATIVE_MINOR_VERSION >= 71
+  } else if (value.isBigInt()) {
+    shareable = std::make_shared<ShareableBigInt>(rt, value.getBigInt(rt));
+#endif
   } else if (value.isSymbol()) {
     // TODO: this is only a placeholder implementation, here we replace symbols
     // with strings in order to make certain objects to be captured. There isn't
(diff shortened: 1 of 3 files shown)


Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -92,6 +92,10 @@ jsi::Value makeShareableClone(
     shareable = std::make_shared<ShareableScalar>(value.getBool());
   } else if (value.isNumber()) {
     shareable = std::make_shared<ShareableScalar>(value.getNumber());
+#if REACT_NATIVE_MINOR_VERSION >= 71
+  } else if (value.isBigInt()) {
+    shareable = std::make_shared<ShareableBigInt>(rt, value.getBigInt(rt));
```
piaskowyk, line 97: Does JSC and V8 support BigInt as well?

tomekzaw: It's a part of JSI since RN 0.70 (see https://github.com/facebook/react-native/commit/11bae63bb1f833802ec6ce01342ebdd1d61e9252) but this implementation uses `jsi::BigInt::toString` which was added in RN 0.71 (see https://github.com/facebook/react-native/commit/a21a1f845bd854f01cac905a4737a360cca0cd12).

The implementation of these methods depends on the runtime, it works on Hermes, on V8 it's also implemented (see [here](https://github.com/Kudo/react-native-v8/blob/b5ce9d8344ee2abc9e7c1ea82281ba7d6a5ad511/src/v8runtime/V8Runtime.cpp#L692-L733)) but haven't run it yet, on JSC it throws an exception "Not implemented" (see [here](https://github.com/facebook/react-native/blob/a2f11895c2a80d677fdcd34d70d379cc381c438c/packages/react-native/ReactCommon/jsc/JSCRuntime.cpp#L682-L684)).

Speaking of `toString`, we could call it via `jsi::Function::call` so we could also support 0.70, but I think 0.71+ support with cleaner code is fine for us.

piaskowyk: Absolutely make sense 🫡

piaskowyk (approved):

PR #5239, opened 2023-10-16
Add new higher order animation - withClamp

Latropos:
## Summary

Add new higher order function withClamp that allows to limit range of movement. Usable especially with `spring` animation.

Allows to pass number or string for both of the clamps.

https://github.com/software-mansion/react-native-reanimated/assets/56199675/26641e51-7bb2-468d-adf9-a1f686e67824

## Test plan

<details><summary>Testing code </summary>

```javascript
'use strict';
import { defineAnimation, getReduceMotionForAnimation } from './util';
import type {
  Animation,
  AnimationCallback,
  AnimatableValue,
  Timestamp,
} from '../commonTypes';
import type {
  SpringConfig,
  SpringAnimation,
  InnerSpringAnimation,
  SpringConfigInner,
  DefaultSpringConfig,
} from './springUtils';
import {
  initialCalculations,
  calculateNewMassToMatchDuration,
  underDampedSpringCalculations,
  criticallyDampedSpringCalculations,
  isAnimationTerminatingCalculation,
  scaleZetaToMatchClamps,
  isConfigValid,
} from './springUtils';

// TODO TYPESCRIPT This is a temporary type to get rid of .d.ts file.
type withSpringType = <T extends AnimatableValue>(
  toValue: T,
  userConfig?: SpringConfig,
  callback?: AnimationCallback
) => T;

export const withSpring = ((
  toValue: AnimatableValue,
  userConfig?: SpringConfig,
  callback?: AnimationCallback
): Animation<SpringAnimation> => {
  'worklet';

  return defineAnimation<SpringAnimation>(toValue, () => {
    'worklet';
    const defaultConfig: DefaultSpringConfig = {
      damping: 10,
      mass: 1,
      stiffness: 100,
      overshootClamping: false,
      restDisplacementThreshold: 0.01,
      restSpeedThreshold: 2,
      velocity: 0,
      duration: 2000,
      dampingRatio: 0.5,
      reduceMotion: undefined,
      clamp: undefined,
    } as const;

    const config: DefaultSpringConfig & SpringConfigInner = {
      ...defaultConfig,
      ...userConfig,
      useDuration: !!(userConfig?.duration || userConfig?.dampingRatio),
      skipAnimation: false,
    };

    config.skipAnimation = !isConfigValid(config);

    if (config.duration === 0) {
      config.skipAnimation = true;
    }

    function springOnFrame(
      animation: InnerSpringAnimation,
      now: Timestamp
    ): boolean {
      const { toValue, startTimestamp, current } = animation;

      const timeFromStart = now - startTimestamp;

        animation.current = toValue;
        animation.lastTimestamp = 0;
        return true;
      }

      if (config.skipAnimation) {
        // If we use duration we want to wait the provided time before stopping
        if (config.useDuration) return false;
        else {
          animation.current = toValue;
          animation.lastTimestamp = 0;
          return true;
        }
      }
      const { lastTimestamp, velocity } = animation;

      const t = deltaTime / 1000;
      const v0 = -velocity;
      const x0 = toValue - current;

      const { zeta, omega0, omega1 } = animation;

      const { position: newPosition, velocity: newVelocity } =
        zeta < 1
          ? underDampedSpringCalculations(animation, {
              zeta,
              v0,
              x0,
              omega0,
              omega1,
              t,
            })
          : criticallyDampedSpringCalculations(animation, {
              v0,
              x0,
              omega0,
              t,
            });

      animation.current = newPosition;
      animation.velocity = newVelocity;

      const { isOvershooting, isVelocity, isDisplacement } =
        isAnimationTerminatingCalculation(animation, config);

      const springIsNotInMove =
        isOvershooting || (isVelocity && isDisplacement);

      if (!config.useDuration && springIsNotInMove) {
        animation.velocity = 0;
        animation.current = toValue;
        animation.lastTimestamp = 0;
        return true;
      }

      return false;
    }

    function isTriggeredTwice(
      previousAnimation: SpringAnimation | undefined,
      animation: SpringAnimation
    ) {
      return (
        previousAnimation?.lastTimestamp &&
        previousAnimation?.startTimestamp &&
        previousAnimation?.toValue === animation.toValue &&
        previousAnimation?.duration === animation.duration &&
        previousAnimation?.dampingRatio === animation.dampingRatio
      );
    }

    function onStart(
      animation: SpringAnimation,
      value: number,
      now: Timestamp,
      previousAnimation: SpringAnimation | undefined
    ): void {
      animation.current = value;
      animation.startValue = value;

      let mass = config.mass;
      const triggeredTwice = isTriggeredTwice(previousAnimation, animation);

      const duration = config.duration;

      const x0 = triggeredTwice
        ? // If animation is triggered twice we want to continue the previous animation
          // form the previous starting point
          previousAnimation?.startValue
        : Number(animation.toValue) - value;

      if (previousAnimation) {
        animation.velocity =
          (triggeredTwice
            ? previousAnimation?.velocity
            : previousAnimation?.velocity + config.velocity) || 0;
      } else {
        animation.velocity = config.velocity || 0;
      }

      if (triggeredTwice) {
        animation.zeta = previousAnimation?.zeta || 0;
        animation.omega0 = previousAnimation?.omega0 || 0;
        animation.omega1 = previousAnimation?.omega1 || 0;
      } else {
        if (config.useDuration) {
          const actualDuration = triggeredTwice
            ? // If animation is triggered twice we want to continue the previous animation
              // so we need to include the time that already elapsed
              duration -
              ((previousAnimation?.lastTimestamp || 0) -
                (previousAnimation?.startTimestamp || 0))
            : duration;

          config.duration = actualDuration;
          mass = calculateNewMassToMatchDuration(
            x0 as number,
            config,
            animation.velocity
          );
        }

        const { zeta, omega0, omega1 } = initialCalculations(mass, config);
        animation.zeta = zeta;
        animation.omega0 = omega0;
        animation.omega1 = omega1;

        if (config.useDuration && config.clamp !== undefined) {
          animation.zeta = scaleZetaToMatchClamps(animation, config.clamp);
        }
      }

      animation.lastTimestamp = previousAnimation?.lastTimestamp || now;

      animation.startTimestamp = triggeredTwice
        ? previousAnimation?.startTimestamp || now
        : now;
    }

    return {
      onFrame: springOnFrame,
      onStart,
      toValue,
      velocity: config.velocity || 0,
      current: toValue,
      startValue: 0,
      callback,
      lastTimestamp: 0,
      startTimestamp: 0,
      zeta: 0,
      omega0: 0,
      omega1: 0,
      reduceMotion: getReduceMotionForAnimation(config.reduceMotion),
    } as SpringAnimation;
  });
}) as withSpringType;

```
</summary>


src/reanimated2/animation/clamp.ts
```diff
@@ -0,0 +1,92 @@
...
+} from '../commonTypes';
+import type { ClampAnimation } from './commonTypes';
+
+type withClampType = <T extends AnimatableValue>(
+  clamp: { min: number; max: number },
+  delayedAnimation: T,
+  reduceMotion?: ReduceMotion
+) => T;
+
+// TODO This feature is not documented yet
```
Latropos, line 18: I think that documentation should be done in separate PR

tomekzaw: Could be, but maybe we can try in this one?

src/reanimated2/animation/clamp.ts
```diff
@@ -0,0 +1,92 @@
...
+  AnimationObject,
+  ReduceMotion,
+} from '../commonTypes';
+import type { ClampAnimation } from './commonTypes';
+
+type withClampType = <T extends AnimatableValue>(
+  clamp: { min: number; max: number },
+  delayedAnimation: T,
+  reduceMotion?: ReduceMotion
+) => T;
```
tomekzaw, line 16: 1. I don't have a strong opinion here but I think that the animation should be the first argument. Indeed, in `withDelay`, the animation is passed as second argument and the first argument is the delay, however in `withRepeat` the number of repetitions is declared after the animation. In this case, the clamp is applied "after" the animation yields the value, so this would correspond better with the order of params.
2. Both `min` and `max` should be optional parameters so that the animation can be clamped from one side only.
3. I don't like the fact that we pass `reduceMotion` as yet another argument. Maybe we can group `min`, `max` and `reduceMotion` in a single `config` object, like we do for spring and timing?
4. I'm wondering how `withClamp` would work for colors 😄 Can we come up with some DX (developer experience) tricks to ensure that it's used only for numbers (e.g. enforce it using TypeScript)?

Latropos: 1. You are right that it is justified to have the animation as the first parameter. On the other hand people need `withDelay` much more often than `withRepeat` and I guess that they are more familiar with syntax with animation being the second, so I think I would keep it like that.
2. Of course!
3. Sure! (BTW. Wouldn't it be better not to pass `reduceMotion` anywhere, but to have  higher order animation withReduceMotion?)
4. Colours have some weird bugs with spring. Generally animation between two colours works worse than animation single numerical progress and interpolating that color.
Yes, we can make sure that user provides number

tomekzaw: 1. Let's discuss it with the whole team
2. Thank you!
3. Thank you! I think we were considering it at some point but please ask @bartlomiejbloniarz for details.
4. Thank you Some time ago we greatly improved the implementation of `interpolateColor`. I thought `withSpring(color)` also works fine, are you able to prepare a repro of the problem you've mentioned?

.eslintrc.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
     },
   },
   rules: {
+    'no-case-declarations': 'off',
```
Latropos, line 27: This rule is not helpful, it encourages to obfuscate code, it enforces sth like this:

``` diff
+ const temp = foo + 5;
 switch (foo) {
     case 1:
-        const temp = foo + 5;
        return temp;
}
```

src/reanimated2/animation/clamp.ts
bartlomiejbloniarz: ad 3. we considered it at some point, but decided that adding a reduced motion prop / config would be a smaller change to the api convention

src/reanimated2/animation/clamp.ts
Latropos: Could we add examples then?
In our examples in docs we use reanimated from node_modules. We can't add example using a function that hasn't been merged yet. I suppose that splitting this in two PRs is a better solution.

src/reanimated2/animation/clamp.ts
piaskowyk, line 1: The example of usage in Example app is missing. Could you add one?

.eslintrc.js
piaskowyk: is this change related to this PR?

src/reanimated2/animation/clamp.ts
```diff
@@ -0,0 +1,93 @@
...
+  AnimatableValue,
+  AnimationObject,
+  ReduceMotion,
+} from '../commonTypes';
+import type { ClampAnimation } from './commonTypes';
+
+type withClampType = <T extends AnimatableValue>(
+  config: { min?: number; max?: number; reduceMotion?: ReduceMotion },
+  delayedAnimation: T,
+  reduceMotion?: ReduceMotion
```
piaskowyk, line 15: Why we can set `reduceMotion` twice here?

piaskowyk (changes requested):

src/reanimated2/animation/clamp.ts
Latropos: Do you think we should? Our example app is quite big and I am going to create an example for docs anyway.

piaskowyk: Yes, I think that we need. Every new feature in Reanimated has own example. We uses those examples as a tests before release.

piaskowyk:
#### Could you confirm that it works for different type of animation value?
- standard numbers
- colors - as HEX, as string rgb(), hsv()
- matrixes
- postfix values - deg, px

#### What will happen if someone use this syntax:
- withClamp(withSequence(...))
- withClamp(withDelay(...))

Latropos:
@piaskowyk I've just realised that our spring doesn't work with `withRepeat` very well.
1. In the first recording the first animation has EMPTY config, and the second has NO config. So they should have exactly the same behaviour. However in reality, once we trigger animation again the upper stops and immediately goes to the value it has just receives, and the bottom ones starts animation towards the new value.

https://github.com/software-mansion/react-native-reanimated/assets/56199675/7513d840-1509-4055-9153-9bdfee3857be

2. Spring withDuration doesn't work with `withRepeat`. It is assumed that new duration should match the total time of repeated animation, so when you want your animation to repeat forever you get bad results. 😢

Latropos:
@piaskowyk I've just tested and:
* withClamp(withSequence(...)) works ✅
* withClamp(withDelay(...)) works ✅
With clamp doesn't work, if inner animation returns something other than number. It is because we have to implement comparison for each data type:
* number ✅ - this is simple, we compare numbers by value 
* colors ❌ - how to say if colour has exceeded some predefined value? Is blue bigger than red? Do we want to implement it?
* matrixes ❌ - nope, matrix contains too much values
* postfix values - deg, px ❌ - not implemented as well. Because we would have to answer more questions, for example imagine sb provides lower clamp in radians and upper one in degrees. Should we handle it?

app/src/examples/WithClampExample.tsx
```diff
@@ -0,0 +1,100 @@
...
+  };
+
+  const clampedStyle = useAnimatedStyle(() => {
+    return {
+      width: withClamp(
+        { min: 120, max: 220 },
+        withSpring(randomWidth.value, config)
+      ),
+    };
+  });
```
piaskowyk, line 56: Why to pass `withSpring` with config `clamp: [120, 220]` to `withClamp`? 🤔

src/reanimated2/animation/clamp.ts
```diff
@@ -0,0 +1,102 @@
...
+      }
+
+      function onStart(
+        animation: Animation<any>,
+        value: AnimatableValue,
+        now: Timestamp,
+        previousAnimation: Animation<any> | null
+      ): void {
+        animation.startTime = now;
+        animation.started = false;
```
piaskowyk, line 63: It seems that those values are used only by `withDelay`, we can remove it.

src/reanimated2/animation/clamp.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      function clampOnFrame(
+        animation: ClampAnimation,
+        now: Timestamp
+      ): boolean {
+        const finished = clampedAnimation.onFrame(clampedAnimation, now);
+
+        if (clampedAnimation.current === undefined) {
+          // This should never happen
+          // TODO check if nextAnimation.current should be optional
```
piaskowyk, line 40: Instead of just adding `// TODO`, let's do it 😅

src/reanimated2/animation/clamp.ts
```diff
@@ -0,0 +1,102 @@
...
+            "[Reanimated] Error inside 'withClamp' animation, the inner animation has invalid current value"
+          );
+          return true;
+        } else {
+          const clampUpper = config.max ? config.max : clampedAnimation.current;
+          const clampLower = config.min ? config.min : clampedAnimation.current;
+          animation.current = Math.max(
+            clampLower,
+            Math.min(clampUpper, clampedAnimation.current)
+          );
```
piaskowyk, line 51: Unfortunately it doesn't work for some edge cases:
image
I think that the better option will be just it:
```suggestion
if (config.max < clampedAnimation.current) {
  animation.current = config.max;
}
if (config.min > clampedAnimation.current) {
  animation.current = config.min;
}
```

src/reanimated2/animation/clamp.ts
```diff
@@ -0,0 +1,102 @@
...
+): Animation<ClampAnimation> {
+  'worklet';
+  return defineAnimation<ClampAnimation, T>(
+    _clampedAnimation,
+    (): ClampAnimation => {
+      'worklet';
+      const clampedAnimation =
+        typeof _clampedAnimation === 'function'
+          ? _clampedAnimation()
+          : _clampedAnimation;
```
piaskowyk, line 30: When clampedAnimation can be a function?

piaskowyk (changes requested):

src/reanimated2/animation/clamp.ts
Latropos: I don't know. I've copied this part from our `withDelay` function. I don't know when the `delayedAnimation` can be a function as well. This check (inside withDelay) was added two years ago by you, and if nb knows why we have it we may consider removing it in the future.

src/reanimated2/animation/clamp.ts
Latropos: You're right I checked, and this should required. However I don't want to change anything within this PR cause I know that @tjzel is still working on typing our animations, so I'll just add a comment with information

tjzel: This is hard 😵

app/src/examples/WithClampExample.tsx
```diff
@@ -0,0 +1,114 @@
+import Animated, {
+  useSharedValue,
+  useAnimatedStyle,
+  withSpring,
+  withClamp,
+} from 'react-native-reanimated';
+import { View, Text, Button, StyleSheet, ViewStyle } from 'react-native';
+import React, { useState } from 'react';
+
+/**  TODO #5239 As soon as 'withClamp' is included in reanimated version used in docs
```
Latropos, line 10: Personally I like the idea of giving github :octocat:  PR number here, because
- If you don't know why sb added this TODO comment you can quicker find a valid PR
- Some of TODO comments are duplicated in multiple files, since the change should affect several files. You can easily search in code for all the TODOs created in given PR to check if you should address more of them.

src/reanimated2/index.ts
```diff
@@ -60,12 +60,13 @@ export type {
 export {
   cancelAnimation,
   defineAnimation,
-  withTiming,
-  withSpring,
+  withClamp,
```
Latropos, line 63: I've sorted it alphabetically.

src/reanimated2/animation/util.ts
```diff
@@ -49,23 +49,25 @@ interface RecognizedPrefixSuffix {
...
-    if (!match) {
-      throw new Error("[Reanimated] Couldn't parse animation value.");
-    }
-    const prefix = match[1];
-    const suffix = match[4];
-    // number with scientific notation
-    const number = match[2] + (match[3] ?? '');
-    return { prefix, suffix, strippedValue: parseFloat(number) };
-  } else {
+  if (typeof value === 'number') {
```
Latropos, line 56: I've moved this up, to make this less nested.
Do we need `prefix`?

Example/ios/Podfile.lock
```diff
@@ -807,4 +807,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: f37cea0ea1b6dc523fb88a9c2bdb9993f4b567ce
 
-COCOAPODS: 1.14.0
+COCOAPODS: 1.13.0
```
tomekzaw: Please bump CocoaPods to 1.14.0 or even 1.14.2. Here's the command: `brew upgrade cocoapods`.

src/reanimated2/animation/util.ts
piaskowyk: > I've moved this up, to make this less nested.
but why? 🤔 You significantly changed the logic of this method.

piaskowyk (changes requested):

src/reanimated2/animation/util.ts
Latropos: The logic isn't changed drastically, since value is typed to be either a 'number' or 'string'. 
> Why?

Many devs encourage to prefer the less nested solution, since it is considered simpler. https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

Latropos: The logic isn't changed drastically, since value is typed to be either a 'number' or 'string'. 
> Why?

Many devs encourage to prefer the less nested solution, since it is considered simpler. https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

piaskowyk: But what about for example this cast?
https://github.com/software-mansion/react-native-reanimated/blob/4d4f2118553e330829b2d49d4ca3b4d92d3a8a55/src/reanimated2/animation/util.ts#L160

piaskowyk (approved):

PR #5250, opened 2023-10-16
Refactor `IS_NATIVE` and `global._IS_FABRIC`

bartlomiejbloniarz:
## Summary

This PR removes `IS_NATIVE` in favor of `SHOULD_BE_USE_WEB` to be more consistent with `PlatformChecker`. Additionally `isFabric()` was added to `PlatformChecker`.


tomekzaw (approved):
Looks good to me 👏 

BTW Does it mean that we no longer need `global._IS_FABRIC` on the UI runtime? Technically, we could remove injecting it but is it safe to do so? Also, if we leave it and add `'worklet';` directive in the `isFabric` function, this will also work. Are there any use cases where we need to check if Fabric is enabled on the UI thread?

piaskowyk (approved):

bartlomiejbloniarz:
@tomekzaw as mentioned offline, IS_FABRIC is currently not used on the UI thread but I think we should keep it around until at least we are done with layout animations on Fabric

PR #5254, opened 2023-10-17
Add more eslint rules

Latropos:
## Summary
Add some more eslint rules. I've also removed rule 'curly' from `plugin/.eslintrc.js`, since this config extends config with this rule enabled anyway.

<table>
<tr><td>Rule</td><td>Change</td><td width="50%">Why?</td></tr>
<tr><td>
'no-case-declarations'
</td><td>
Explicitly enable ✅
</td><td>
This rule will be enabled explicitly (although it is already inside recommended set of rules), since we had a discussion about it and we probably want to remember it.

</td></tr>
<tr><td>
'no-use-before-define'
</td><td>
Explicitly turn off ❌
</td><td>
Could be handy, since worklets have to be defined in order, but requires too many changes
</td></tr>
<tr><td>
'@typescript-eslint/no-shadow'
</td><td>
enable ✅
</td><td>
Disable overwriting variables. Makes code a bit clearer. We often overwrite function arguments when we have two nested functions.

I've added underscore "_" prefix to argument of **inner** function, since this is a common custom and we use underscore for private variables and indeed these vars are more private than the external ones.

```diff
function outerFunction(variableName: number){
    // some code

-    function innerFunction(variableName: number){
+    function innerFunction(_variableName: number){
-         console.log(variableName)
+         console.log(_variableName)
    }
}
```

</td></tr>

</table>


src/reanimated2/UpdateProps.ts
```diff
@@ -20,7 +28,7 @@ if (shouldBeUseWeb()) {
   updateProps = (_, updates, maybeViewRef, isAnimatedProps) => {
     'worklet';
     if (maybeViewRef) {
-      maybeViewRef.items.forEach((item, _) => {
+      maybeViewRef.items.forEach((item, __) => {
```
m-bert, line 31: Is there any way to avoid such change without doing `// eslint-disable-next-line`?  Like adding exception to eslint rule?

src/reanimated2/animation/util.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-shadow */
```
m-bert, line 1: Why do we disable this rule?

m-bert (commented):
I'm not a huge fan of using `_` to avoid shadowing since we use the same prefix in RNGH to mark unused variables. Also, there are some places where you used different approach (like changing `key` to `initialKey`).

I don't think it's a problem though, let's hear what others have to say.

src/reanimated2/animation/util.ts
Latropos: Since it requires almost rewriting this file, which is very important one. I would prefer to do it in some other PR.

tjzel:
I agree @m-bert, we comply to the convention of marking unused variables with `_` prefix. Also we mark JSI bindings with `_` prefix - let's not overuse this. Ideally we just need to change the name for the internal functions so they would differ from the external. I'm a fan of suffixing `Internal` to the function or something similar.

src/reanimated2/UpdateProps.ts
Latropos: Unfortunately no 😔

Latropos:
@tjzel I've updated variable names, however I've used `inner` instead if `internal`. The latter sounds like a synonym of private and the first one suggest that there was some code nesting.

.eslintrc.js
```diff
@@ -25,12 +25,14 @@ module.exports = {
...
     curly: 'error',
+    'no-case-declarations': 'off', // This recommended rule is actually harmful instead of helpful
+    '@typescript-eslint/no-shadow': 'error',
     'import/no-unresolved': 'error',
     'import/consistent-type-specifier-style': ['error', 'prefer-top-level'],
     'react/jsx-uses-vars': 'error',
     'react/jsx-uses-react': 'error',
-    'no-use-before-define': 'off',
-    '@typescript-eslint/no-use-before-define': 'off', // TODO consider enabling this (currently it reports styles defined at the bottom of the file)
+    'no-use-before-define': 'error',
```
tjzel, line 34: I really don't think we should enable this rule. While it could be handy when working with worklets, it would overall lower the quality of our code.

I know we aren't really strict about this, but in my opinion code files in this order:
```TSX
// important thing

// less important thing

// helper function

// stylesheet
```

Improves code quality since you read code from top to bottom, instead of the opposite. Instead we could probably define a separate rule just for worklets (if possible).
cc @tomekzaw @piaskowyk

Latropos: I haven't turned this rule on in our example app because of stylesheet, so this particular element can be omitted.

tjzel: What do you mean by 'this particular element'?

Latropos: Fixed, didn't want to use word "stylesheet" twice in one sentence 😆

tjzel: Yes, but stylesheet is just a part of the problem.

.eslintrc.js
```diff
@@ -25,12 +25,14 @@ module.exports = {
   },
   rules: {
     curly: 'error',
+    'no-case-declarations': 'off', // This recommended rule is actually harmful instead of helpful
```
piaskowyk, line 28: Is any place in code where we are using `case-declarations`? 🤔

.eslintrc.js
```diff
@@ -25,12 +25,14 @@ module.exports = {
...
+    '@typescript-eslint/no-shadow': 'error',
```
piaskowyk, line 29: This rule is actually helpful 👍

.eslintrc.js
piaskowyk: I agree with Tomek. I don't see any benefits to using this rule either. If there are any, please let me know. However, I think we should disable this rule.

app/.eslintrc.js
```diff
@@ -3,11 +3,12 @@ module.exports = {
   extends: '@react-native',
   plugins: ['eslint-plugin-no-inline-styles'],
   rules: {
+    '@typescript-eslint/no-use-before-define': 'off',
+    '@typescript-eslint/no-shadow': 'off',
```
piaskowyk, line 7: why we want to disable it here?

plugin/.eslintrc.js
```diff
@@ -1,7 +1,8 @@
 module.exports = {
   root: true,
   extends: '../.eslintrc.js',
-  rules: {
-    'curly': 'error',
+  rules:{
+    '@typescript-eslint/no-use-before-define':'off'
```
piaskowyk, line 5: In the documentation is information that you need to disable more than this to make it wokrs - https://typescript-eslint.io/rules/no-use-before-define/#how-to-use

piaskowyk (changes requested):

Latropos (commented):
.

app/.eslintrc.js
tjzel: +1

src/reanimated2/Bezier.ts
```diff
@@ -52,16 +52,16 @@ export function Bezier(
     aX: number,
     aA: number,
     aB: number,
-    mX1: number,
-    mX2: number
+    innerMX1: number,
```
tjzel, line 55: I think this should be a suffix, not a prefix. While I understand that semantically it looks a lot better and is a better resemblance of a natural language, you are polluting the `inner` namespace, e.g. autocompletion will suggest a lot of things. @piaskowyk, @m-bert  what do you think?

app/.eslintrc.js
Latropos: To limit number of changes inside one PR

tjzel: I don't think we really care about number of changes in `/app`.

src/reanimated2/Bezier.ts
m-bert: I don't mind it being prefix. It looks more natural this way.

piaskowyk: I agree, `inner` isn't the best prefix in this case

src/reanimated2/animation/util.ts
tjzel: @Latropos Let's add a TODO here and a respective task to our Trello then

app/.eslintrc.js
```diff
@@ -3,11 +3,12 @@ module.exports = {
...
     'no-inline-styles/no-inline-styles': 'error',
     'react-native/no-inline-styles': 'off',
-    '@typescript-eslint/no-shadow': 'off',
     'react-native/no-unused-styles': 'error',
-    'react-native/no-raw-text': 'off',
+    'react-native/no-raw-text': 'off', //This rule is great, we don't enable it because of its performance. If we ever find similar rule we should enable it.
```
tomekzaw, line 11:
```suggestion
    'react-native/no-raw-text': 'off', // This rule is great, we don't enable it because of its performance. If we ever find similar rule we should enable it.
```

src/reanimated2/Bezier.ts
```diff
@@ -52,16 +52,16 @@ export function Bezier(
...
+    innerMX2: number
```
tomekzaw, line 56: Instead of renaming the arguments here, can we move the function to outer scope? Looks like we don't use values from closure inside this function.

src/createAnimatedComponent/PropsFilter.tsx
```diff
@@ -59,8 +59,8 @@ export class PropsFilter implements IPropsFilter {
           AnimatedComponentProps<AnimatedProps>
         >;
         if (animatedProp.initial !== undefined) {
-          Object.keys(animatedProp.initial.value).forEach((key) => {
-            props[key] = animatedProp.initial?.value[key];
+          Object.keys(animatedProp.initial.value).forEach((initialKey) => {
```
tomekzaw, line 62: `initialKey` suggests that key changes, the name is a bit confusing.

src/reanimated2/PropsRegistry.ts
```diff
@@ -21,7 +21,7 @@ function flush() {
   viewTags = [];
 }
 
-function removeFromPropsRegistryOnUI(viewTags: number[]) {
+function removeFromPropsRegistryOnUI(innerViewTags: number[]) {
```
tomekzaw, line 24: Let's restore the original name of the argument here (`viewTags`) and rename the variable in line 7:
```diff
-let viewTags: number[] = [];
+const VIEW_TAGS: number[] = [];
```
```diff
-viewTags = [];
+VIEW_TAGS.length = 0; // clear the array
```

src/reanimated2/UpdateProps.ts
```diff
@@ -20,7 +20,7 @@ if (shouldBeUseWeb()) {
...
```
tomekzaw, line 23: Instead of adding two underscores, let's just come up with some name here.

src/reanimated2/animation/spring.ts
```diff
@@ -68,19 +68,19 @@ export const withSpring = ((
       animation: InnerSpringAnimation,
       now: Timestamp
     ): boolean {
-      const { toValue, startTimestamp, current } = animation;
+      const { innerToValue, startTimestamp, current } = animation;
```
tomekzaw, line 71: Shouldn't it be `toValue: innerToValue` instead? 🤯 

As you can see, renaming variables in existing code can lead to potential bugs.

Finally, why this regression has not been caught by any of our unit tests?

tomekzaw (commented):
Instead of renaming the variables in existing code, let's add `/* eslint-disable @typescript-eslint/no-shadow */` in all existing files. This way we don't modify existing working code and have some time to refactor it later.

src/reanimated2/animation/spring.ts
tjzel: My guess would be because typeof animation is tragic 🤣

tjzel:
@tomekzaw That's a great idea to keep the warning for now.

piaskowyk (approved):

PR #5256, opened 2023-10-17
Fix "`valueUnpacker` is not a worklet" error when Reanimated Babel plugin is missing

### LINKED ISSUE #5209: Error: [Reanimated] `valueUnpacker` is not a worklet, js engine: hermes
### Description

When the application is opened it returns the error in question.

package:

     "@react-native-firebase/app": "^18.5.0",
     "@react-native-firebase/messaging": "^18.5.0",
     "@react-navigation/bottom-tabs": "^6.5.9",
     "@react-navigation/drawer": "^6.6.4",
     "@react-navigation/native": "^6.1.8",
     "@react-navigation/native-stack": "^6.9.14",
     "react": "18.2.0",
     "react-native": "0.72.5",
     "react-native-gesture-handler": "^2.13.1",
     "react-native-reanimated": "^3.5.4",
     "react-native-safe-area-context": "^4.7.2",
     "react-native-screens": "^3.25.0",

### Steps to reproduce

 yarn && cd ios && pod install && cd .. && yarn ios cache clean

.

### Reanimated version

3.5.4

### React Native version

0.72.5

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

None

### Build type

Debug mode

### Device

iOS simulator

### Device model

Any

### Acknowledgements

Yes

tomekzaw:
## Summary

This PR fixes a regression in troubleshooting when user forgot to setup Reanimated Babel plugin. Currently, a generic error is shown, instead of a developer-friendly user message with a link to the docs.

Closes #5209.

| Before | After |
|:-:|:-:|
| Zrzut ekranu 2023-10-17 o 18 20 45 | Zrzut ekranu 2023-10-17 o 18 25 23 |


src/reanimated2/valueUnpacker.ts
```diff
@@ -74,6 +74,14 @@ type ValueUnpacker = WorkletFunction<
...
+  const f = (() => {
+    'worklet';
+  }) as WorkletFunction<[], void>;
+  if (f.__workletHash === undefined) {
+    throw new Error(
+      `[Reanimated] Failed to create a worklet. See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooting#failed-to-create-a-worklet for more details.`
+    );
+  }
   if (!('__workletHash' in valueUnpacker)) {
     throw new Error('[Reanimated] `valueUnpacker` is not a worklet');
```
piaskowyk, line 86: Is it still possible for this situation to occur?

src/reanimated2/valueUnpacker.ts
```diff
@@ -74,6 +74,14 @@ type ValueUnpacker = WorkletFunction<
...
 
 if (__DEV__ && IS_NATIVE) {
+  const f = (() => {
+    'worklet';
+  }) as WorkletFunction<[], void>;
+  if (f.__workletHash === undefined) {
+    throw new Error(
+      `[Reanimated] Failed to create a worklet. See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooting#failed-to-create-a-worklet for more details.`
+    );
+  }
```
tjzel, line 84: I'm not sure if this is the file we want this check in. Is this file or `initializers.ts` called sooner?

tjzel (changes requested):

src/reanimated2/valueUnpacker.ts
tomekzaw: Looks like it is, hence the bug 😈

src/reanimated2/valueUnpacker.ts
tomekzaw: Technically yes, for instance if we remove `'worklet';` directive from `valueUnpacker`. But then it will fail on `valueUnpacker.__closure` below so I think we can remove this check.

src/reanimated2/valueUnpacker.ts
tjzel: :swiniok:

tjzel: But this check seems pretty unrelated to `valueUnpacker` - it's just a coincidence that lack of plugin failed here. I'd love to see this check be in a more meaningful place, but if you think it's okay here I will accept.

tomekzaw: Yes, it's unrelated to `valueUnpacker` but we need to check if worklet conversion was successful before we access `valueUnpacker.__closure` &ndash; otherwise, we get the error from the issue. I guess we could move the check to a separate file and call it from here but do we really need a separate file for this? Alternatively, we could move these two checks (1. workletHash, 2. valueUnpacker closure) to a new file but I think it's better to have valueUnpacker check in `valueUnpacker.ts`.

tjzel: 🆗

tjzel (approved):

ZachYanez:
Hello, I ran into this issue today and It's crashing my app. How can I get the fixed version of this library in my app? Running 3.5.4

src/reanimated2/valueUnpacker.ts
```diff
@@ -74,6 +74,14 @@ type ValueUnpacker = WorkletFunction<
 >;
 
 if (__DEV__ && IS_NATIVE) {
```
tomekzaw, line 76: I will need to use `!SHOULD_BE_USE_WEB` instead here

piaskowyk (approved):

PR #5257, opened 2023-10-17
Remove plugin.js.map

tomekzaw:
## Summary

This PR removes `plugin.js.map` from the repository since ackchyually it is not required when using Reanimated installed from branch (I hope so) and also causes conflicts when there are plugin changes in two or more PRs.

## Test plan

Install Reanimated from branch and check if it works


piaskowyk (approved):
🚀

tjzel (approved):
My heart cries

PR #5259, opened 2023-10-17
Memoize `_getViewInfo` result

tomekzaw:
## Summary

This PR adds memoization of `_getViewInfo` result in `createAnimatedComponent` in order to avoid calling it on each re-render. On Fabric, the expected behavior is that `getShadowNodeWrapperFromRef` will be called only once after the component is mounted.

Additionally, this PR reverts changes introduced in #4986 because now `ShadowNode` is read only once and never changes so the condition is always false.

## Test plan

1. Add `console.log` in `getShadowNodeWrapperFromRef`
2. Open `WidthExample.tsx`
3. Click "Increase height"/"Increase padding" buttons to force renders
4. There should be no logs in the output

### DIFF
--- a/Common/cpp/Fabric/PropsRegistry.cpp
+++ b/Common/cpp/Fabric/PropsRegistry.cpp
@@ -18,7 +18,7 @@ void PropsRegistry::update(
     // returns `ShadowNodeFamily const &` which is non-owning
     map_[tag] = std::make_pair(shadowNode, props);
   } else {
-    // no need to update `.first` because ShadowNodeFamily never changes
+    // no need to update `.first` because ShadowNode's family never changes
     // merge new props with old props
     it->second.second.update(props);
   }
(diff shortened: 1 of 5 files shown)


src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -251,6 +252,10 @@ export function createAnimatedComponent(
     }
 
     _getViewInfo(): ViewInfo {
+      if (this._viewInfo !== undefined) {
+        return this._viewInfo;
+      }
```
piaskowyk, line 257: Just to be sure - did you tested those changes with fast refresh and hot reload?

Common/cpp/Fabric/PropsRegistry.cpp
```diff
@@ -18,14 +18,9 @@ void PropsRegistry::update(
     // returns `ShadowNodeFamily const &` which is non-owning
     map_[tag] = std::make_pair(shadowNode, props);
   } else {
+    // no need to update `.first` because ShadowNodeFamily never changes
```
michalmaka, line 21:
```suggestion
    // no need to update `.first` because ShadowNode's family never changes
```

michalmaka (approved):

src/createAnimatedComponent/createAnimatedComponent.tsx
tomekzaw: Yep, it seems to work fine.

PR #5264, opened 2023-10-18
Fix `runOnJS` crash in release mode again

tomekzaw:
## Summary

This PR fixes a crash described in https://github.com/software-mansion/react-native-reanimated/issues/4613#issuecomment-1741018643.

Turns out that the condition in assert is always true in debug mode but there's an optimization for RemoteFunction in release mode which breaks it.

## Test plan

See repro in #4617.

### DIFF
--- a/Common/cpp/SharedItems/Shareables.cpp
+++ b/Common/cpp/SharedItems/Shareables.cpp
@@ -69,9 +69,9 @@ jsi::Value makeShareableClone(
       shareable =
           std::make_shared<ShareableArrayBuffer>(rt, object.getArrayBuffer(rt));
     } else if (object.isHostObject(rt)) {
-      assert(
-          !object.isHostObject<ShareableJSRef>(rt) &&
-          "[Reanimated] Provided value is already an instance of ShareableJSRef.");
+      if (object.isHostObject<ShareableJSRef>(rt)) {
+        return object;
+      }
       shareable =
           std::make_shared<ShareableHostObject>(rt, object.getHostObject(rt));
     } else {


piaskowyk (approved):

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -69,9 +69,9 @@ jsi::Value makeShareableClone(
       shareable =
           std::make_shared<ShareableArrayBuffer>(rt, object.getArrayBuffer(rt));
     } else if (object.isHostObject(rt)) {
-      assert(
-          !object.isHostObject<ShareableJSRef>(rt) &&
-          "[Reanimated] Provided value is already an instance of ShareableJSRef.");
+      if (object.isHostObject<ShareableJSRef>(rt)) {
+        return object;
+      }
```
tjzel, line 74: Why do we return `object` instead of `value`? Or maybe we should actually make a copy here?

tjzel (changes requested):

Common/cpp/SharedItems/Shareables.cpp
tomekzaw: Thanks for pointing out this. I hope I can explain it better. First, `ShareableJSRef` is a container for `Shareable` (as a `shared_ptr`) which is a container for any JS value (there are specialized classes like `ShareableString` or `ShareableObject`). It also extends `jsi::HostObject` so we can create `jsi::HostObject<ShareableJSRef>` and use it as a regular JS value in our JS code.

Technically, here we should obtain the underlying `std::shared_ptr<Shareable>`. In other branches for other types we simply create it, and in this case we just obtain it using `->value()` method:
```cpp
if (object.isHostObject<ShareableJSRef>(rt)) {
  shareable = object.getHostObject<ShareableJSRef>(rt)->value();
}
```
All branches (for all types) end with the following line which wraps the returned `Shareable` in a new instance of `jsi::HostObject<ShareableJSRef>`:
```cpp
return ShareableJSRef::newHostObject(rt, shareable);
```
This means that effectively we would first unwrap the `jsi::HostObject<ShareableJSRef>`, then unwrap `ShareableJSRef` to obtain `std::shared_ptr<Shareable>`, then wrap it again in a new `ShareableJSRef` object and finally wrap it with `jsi::HostObject<ShareableJSRef>`. And this is not eco-friendly, my friend.

Can we re-use `ShareableJSRef`? Yes, because it's just a regular C++ object which holds only `std::shared_ptr<Shareable> value_;`. We never overwrite `value_` and it doesn't hold any runtime-bound JSI data types so it's safe to hold.

Can we re-use `jsi::HostObject<ShareableJSRef>`? Yes, because we call `makeShareableClone` and pass the runtime which would be used to unwrap it just to wrap it again. Also, under the hood `jsi::HostObject<ShareableJSRef>` holds `std::shared_ptr<ShareableJSRef>` so it's perfectly fine to hold it. The only difference would be that we have a new instance of `jsi::HostObject<ShareableJSRef>` but nothing should break if we re-use the same instance of `jsi::Value` since it's managed by JS VM GC.

That's why I decided to go for `return object;`. Note that we can't have `return value;` because we obtain `const jsi::Value &value` as an argument but we need to return `jsi::Value` so we would need to create a copy, i.e. `return jsi::Value(rt, value);`.

tjzel: That's an excellent answer, thanks for explaining that!

PR #5266, opened 2023-10-18
Add option to omit `initData` in babel plugin

bartlomiejbloniarz:
## Summary

This PR changes the babel plugin to remove `initData` on web. It aims to reduce the bundle size.

❗ EDIT: I reworked this PR to instead have a more narrow configuration option (`omitNativeOnlyData`). This will work well in conjunction with our babel preset which will be able to configure the plugin automatically based on the target platform.

## Test plan
Modify the `NextExample/babel.config.js` file:
```js
['react-native-reanimated/plugin', { omitWorkletInitData: true }]
```
Run `yarn build` in NextExample and verify that the difference between reanimated and noreanimated is around 35KB.


plugin/src/utils.ts
```diff
@@ -5,3 +5,7 @@ export function isRelease() {
     process.env.NODE_ENV?.match(pattern)
   );
 }
+
+export function isWeb() {
+  return process.env.REANIMATED_BABEL_PLUGIN_IS_WEB === '1';
+}
```
tjzel, line 11: I don't like that being an environmental variable - I think it'd much better if it was passed as an option like `processNestedWorklets`.
cc @tomekzaw

__tests__/plugin.test.ts
```diff
@@ -1582,4 +1582,18 @@ describe('babel plugin', () => {
...
+      function foo() {
+        'worklet';
+        var foo = 'bar';
+      }
+    </script>`;
+
+    const { code } = runPlugin(input);
+    expect(code).toHaveWorkletData(0);
+    expect(code).toMatchSnapshot();
+  });
```
tjzel, line 1598: Make sure to clean-up after a test. This way it's left all future tests will not include initData. 
For reference see https://github.com/software-mansion/react-native-reanimated/blob/main/__tests__/plugin.test.ts#L1441

Also the name of the test is somehow misleading. In current implementation, our plugin doesn't know on what platform it will bundle - we have to specify this explicitly (and we can do this regardless of platform!).

Therefore the name of the test should be more like `doesn't include initData when REANIMATED_BABEL_PLUGIN_IS_WEB is set` (with current implementation)

tjzel (changes requested):

tjzel (approved):
Looks good but let's hear what the others think about the API here.

piaskowyk (approved):
👌

__tests__/plugin.test.ts
```diff
@@ -1582,4 +1582,67 @@ describe('babel plugin', () => {
       expect(code).toMatchSnapshot();
     });
   });
+
+  describe('for Web configuration', () => {
```
tomekzaw, line 1586:
```suggestion
  describe('for web configuration', () => {
```

plugin/src/types.ts
```diff
@@ -10,6 +10,7 @@ export interface ReanimatedPluginOptions {
   relativeSourceLocation?: boolean;
   disableInlineStylesWarning?: boolean;
   processNestedWorklets?: boolean;
+  isWeb?: boolean;
```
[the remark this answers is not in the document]
tomekzaw: I think the flag should represent the action (like `processNestedWorklets`), not the target platform. Imagine we introduce a second optimization for web; how should we name it so that the user can enable only one of them? On the other hand, init data (in particular, worklet code as string) is necessary only on native (for now) and probably will never be needed on web.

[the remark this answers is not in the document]
bartlomiejbloniarz: We could do something like `useWebOptimizations` (`shouldBeUseWebOptimizations` 😭). But I think it might be a bit over the top?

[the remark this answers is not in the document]
tjzel: How about `omitNativeOnlyData`?

plugin/src/types.ts
```diff
@@ -10,6 +10,7 @@ export interface ReanimatedPluginOptions {
   relativeSourceLocation?: boolean;
   disableInlineStylesWarning?: boolean;
   processNestedWorklets?: boolean;
+  omitWorkletInitData?: boolean;
```
tjzel, line 13: I understand it will be mostly used in the preset, but still I think we should allow for a more fine-grained control for the user. `omitWorkletInitData` seems mysterious, a user might ask himself: 'what is it and should I really omit it?'. I think we should really signify with the name of that option, that enabling it omits data that is required only on native platforms.

tomekzaw: Do you mean `omitWorkletInitData_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`? 🤔

tjzel: > Do you mean `omitWorkletInitData_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`? 🤔

what

tjzel: reminded me of
image

tjzel:
Mergele? 🙏

tomekzaw (approved):

PR #5273, opened 2023-10-19
Fix ArticleProgressExample again

tomekzaw:
## Summary

This PR fixes ArticleProgressExample again. In current code, `textRef.current` is passed as a copy so it never changes and thus the example doesn't work. I've also decided to simplify the code and name intermediate results of mathematical formulas used to calculate the width of the progress bar.

cc @jfedak whom I've complained to that the example doesn't work on macOS

cc @tjzel who never believed in `clamp`


tjzel (approved):
image

piaskowyk (approved):

PR #5278, opened 2023-10-19
[Web] Fix hiding component with `entering` on every render

tjzel:
## Summary

Some TS shenanigans as a necessary bonus.

<table>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/40713406/d13c9186-f1e5-4e12-bbb1-704b1eac0c08

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/40713406/804fbcfb-f0f0-4b3f-9af3-c46617ffed18

</td></tr>
</table>

## Test plan

<details>

```tsx

import Animated, { FadeIn } from 'react-native-reanimated';

import React from 'react';

  const [shouldBeUseEntering, setShouldBeUseEntering] = React.useState(false);

  function handlePress() {
    setShouldBeUseEntering(!shouldBeUseEntering);
  }

  return (
      <Button title="ShouldBeUseEntering" onPress={handlePress} />
      <Animated.View
        style={styles.box}
        entering={shouldBeUseEntering ? FadeIn : undefined}
      />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: {
    width: 100,
    height: 100,
    backgroundColor: 'blue',
  },
});
```

</details>

### DIFF
--- a/src/reanimated2/layoutReanimation/web/animationsManager.ts
+++ b/src/reanimated2/layoutReanimation/web/animationsManager.ts
@@ -35,18 +35,12 @@ function chooseConfig<ComponentProps extends Record<string, unknown>>(
 
 function checkUndefinedAnimationFail(
   initialAnimationName: string,
-  isLayoutTransition: boolean,
-  hasEnteringAnimation: boolean,
-  element: HTMLElement
+  isLayoutTransition: boolean
 ) {
   // This prevents crashes if we try to set animations that are not defined.
   // We don't care about layout transitions since they're created dynamically
   if (initialAnimationName in Animations || isLayoutTransition) {
     return false;
-  }
-
-  if (hasEnteringAnimation) {
-    makeElementVisible(element);
   }
 
   console.warn(
@@ -56,17 +50,9 @@ function checkUndefinedAnimationFail(
   return true;
 }
 
-function checkReduceMotionFail(
-  animationConfig: AnimationConfig,
-  hasEnteringAnimation: boolean,
-  element: HTMLElement
-) {
+function checkReduceMotionFail(animationConfig: AnimationConfig) {
   if (!animationConfig.reduceMotion) {
     return false;
-  }
-
-  if (hasEnteringAnimation) {
-    makeElementVisible(element);
   }
 
   return true;
@@ -105,6 +91,7 @@ export function startWebLayoutAnimation<
   props: Readonly<AnimatedComponentProps<ComponentProps>>,
   element: HTMLElement,
   animationType: LayoutAnimationType,
+  shouldMakeVisible = false,
   transitionData?: TransitionData
 ) {
   const config = chooseConfig(animationType, props);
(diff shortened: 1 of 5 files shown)


src/reanimated2/layoutReanimation/web/animationsManager.ts
```diff
@@ -141,10 +125,14 @@ export function startWebLayoutAnimation<
...
 
-  if (checkReduceMotionFail(animationConfig, hasEnteringAnimation, element)) {
+  if (checkReduceMotionFail(animationConfig)) {
     return;
   }
 
+  if (shouldMakeVisible) {
+    makeElementVisible(element);
+  }
+
```
m-bert, line 135: Correct me if I'm wrong, but now if user has enabled reduced motion (or we encounter undefined animation) element will stay `hidden`, because that part of code won't be reached.

m-bert (changes requested):
Thanks for fixing! There's only one thing that I'd like you to explain and we're good to go.

src/reanimated2/layoutReanimation/web/animationsManager.ts
```diff
@@ -94,6 +94,10 @@ export function startWebLayoutAnimation<
   shouldMakeVisible = false,
   transitionData?: TransitionData
 ) {
+  if (shouldMakeVisible) {
+    makeElementVisible(element);
+  }
+
```
m-bert, line 100: It looks like now we ignore delay (look [here](https://github.com/software-mansion/react-native-reanimated/blob/4aac80fbd90f76a778bd10e42576dbb0822e0699/src/reanimated2/layoutReanimation/web/componentUtils.ts#L162) for context)

m-bert (changes requested):

src/reanimated2/layoutReanimation/web/animationsManager.ts
tjzel: How do we ignore it? From what I see previous 'unhides' were made without the delay.

m-bert: Yes, they were, but only because we didn't trigger any animation after revealing this element. If I understand correctly, now if element has `entering` prop `makeElementVisible` will trigger, which means that we are ignoring delay that I mentioned in link above.

tjzel: I read it very closely and you're absolutely right - but I just hate it that we have different control flows and every one of them calls this function. I will give it a big think later on and come up with a universal solution.

m-bert (approved):
Now it looks fine 🟢

src/reanimated2/layoutReanimation/web/animationsManager.ts
```diff
@@ -141,17 +122,46 @@ export function startWebLayoutAnimation<
...
+  transitionData?: TransitionData
+) {
+  let visibilityDelay = 0;
+
+  const maybeAnimationConfigWithTransform = tryGetAnimationConfigWithTransform(
+    props,
+    animationType
   );
+
+  if (maybeAnimationConfigWithTransform) {
```
piaskowyk, line 148: It seems that `tryGetAnimationConfigWithTransform` should return `null` when it is not the first render and `animationType` is `entering`, in order to avoid unnecessary code execution.

src/reanimated2/layoutReanimation/web/animationsManager.ts
```diff
@@ -141,17 +122,46 @@ export function startWebLayoutAnimation<
...
+    const { animationConfig, transform } = maybeAnimationConfigWithTransform;
+
+    visibilityDelay = animationConfig.delay;
```
piaskowyk, line 151: If it is a second render, will `makeElementVisible` with `visibilityDelay` be called again?

piaskowyk (changes requested):

piaskowyk (approved):

PR #5283, opened 2023-10-20
Add option to substitute `isWeb` and `shouldBeUseWeb` in babel plugin.

bartlomiejbloniarz:
## Summary

This PR changes the Babel plugin to substitute `isWeb()` and `shouldBeUseWeb()` calls with `true` so that Webpack is able to recognize dead code.

## Test plan
Modify the NextExample/babel.config.js file:
```js
['react-native-reanimated/plugin', { omitNativeOnlyData: true, substituteWebPlatformChecks: true }]
```
Run yarn build in NextExample and verify that the difference between reanimated and noreanimated is around 34KB.


tomekzaw (commented):
I have a crazy idea, why don't we just create a new file `PlatformChecker.web.ts` with the following contents? 😄 

```ts
function isWeb() {
  return true;
}

function shouldBeUseWeb() {
  return true;
}

// ...
```

plugin/src/plugin.ts
```diff
@@ -27,7 +28,12 @@ module.exports = function (): PluginItem {
     visitor: {
       CallExpression: {
         enter(path: NodePath<CallExpression>, state: ReanimatedPluginPass) {
-          runWithTaggedExceptions(() => processForCalleesWorklets(path, state));
+          runWithTaggedExceptions(() => {
+            processForCalleesWorklets(path, state);
+            if (state.opts.substituteWebPlatformChecks) {
+              substituteWebCallExpression(path);
+            }
+          });
```
tjzel, line 36: We have to keep in mind this won't transform `shouldBeUseWeb` or `isWeb` inside worklets. While it's not a valid case to call them there, we should at least have it included in tests.

plugin/src/types.ts
tjzel, line 1: I don't quite remember, didn't we agree to use some better names here? Or was this regarding some other PR.

tjzel (changes requested):

plugin/src/types.ts
bartlomiejbloniarz: We agreed on `omitNativeOnlyData` in a separate pr. I just need to merge those changes here.

tjzel (approved):
Me likes 🌵

PR #5284, opened 2023-10-20
Improve mockedRequestAnimationFrame generated types.

td-tomasz-joniec:
## Summary

Applying this change makes build process avoid including `/// <reference types="node" />` annotation in generated output.
In effect developers using package won't automatically have node types imported into their codebase.

**Before**

image

**After**

image

### DIFF
--- a/src/reanimated2/mockedRequestAnimationFrame.ts
+++ b/src/reanimated2/mockedRequestAnimationFrame.ts
@@ -4,6 +4,6 @@
 // by React Native for test purposes.
 export function mockedRequestAnimationFrame(
   callback: (timestamp: number) => void
-) {
+): ReturnType<typeof setTimeout> {
   return setTimeout(() => callback(performance.now()), 0);
 }


tjzel (approved):

tomekzaw:
1. Do we want to add a CI job that ensures that there's no `/// <reference types="node" />` in the output files?
2. Why not `ReturnType<typeof requestAnimationFrame>`? I know it's still a `number`, but I think the declaration should match the original declaration of `requestAnimationFrame` rather than the actual implementation.

tjzel:
@tomekzaw I'm not sure if `typeof requestAnimationFrame` will be allowed when `requestAnimationFrame` doesn't exist in some environment.

td-tomasz-joniec:
> @tomekzaw
> 
> 1. That's a good idea.
> 2. I'm not sure if `typeof requestAnimationFrame` will be allowed when `requestAnimationFrame` doesn't exist in some environment.

There is also possibility of wrapping returned value in Number(...) ,  Number(NodeJS.Timeout) === number,
so that will ensure that return type is always some number in most cases integer, in worst (unsupported) case NaN

PR #5287, opened 2023-10-20
Update first sentence tense in getting-started.mdx

cwclancy:
Having "was" confused me, and I went back to the fundamentals header, thinking I missed something. This makes it more clear I am in the right spot, and I don't need to go search for a "fundaments section".

To me, it currently reads "now that you have a strong understanding from the fundamentals section, here is a new sections with ..."


kacperkapusciak (approved):
I agree!

PR #3765, opened 2022-11-11
Fix inverted cell style for inverted Animated.FlatList

levibuzolic:
## Description

`Animated.FlatList` currently presents each item flipped when using the `inverted` prop, this due to how the `inverted` props works internally (it literally flips the ScrollView). [In the React  Native codebase](https://github.com/facebook/react-native/blob/29caed22cc1c9a426e524e498cf44cf18df0577e/Libraries/Lists/VirtualizedList.js#L1845-L1850) they then apply a style to each cell item to flip it back to the correct orientation. However this fix isn't applied to Reanimated's `FlatList` implementation because it overrides the `CellRendererComponent` prop.

Fixes #2769

## Changes

I've used the same fix from React Native to correct the implementation in Reanimated and included an example in the Example app so it's easy to verify if this gets regressed in the future.

Before | After
---|---
Simulator Screen Recording - iPhone 8 - 2022-11-11 at 14 19 48 | Simulator Screen Recording - iPhone 8 - 2022-11-11 at 14 18 16

## Test code and steps to reproduce

Check the `Inverted FlatList Example` in the Example app.

## Checklist

- [x] Included code example that can be used to test this change
- [x] ~Updated TS types~
- [x] ~Added TS types tests~
- [ ] Added unit / integration tests
- [x] ~Updated documentation~
- [x] Ensured that CI passes

PR #5295, opened 2023-10-23
Fix inverted `FlatList`

bartlomiejbloniarz:
## Summary

This PR changes how reanimated treats inverted `FlatLists`. In #3765 our implementation was changed to mimic the react-native behavior by assigning a special style that reverts the cell inversion (inverted FlatList flips the entire view and then flips back every cell). In react-native@0.72.4 the inversion style was [changed](https://github.com/facebook/react-native/pull/38073) on android due to some performance issues. That change combined with our implementation was causing the cell to remain inverted on android (since both styles were being applied). This PR changes our implementation to instead allow the `CellRendererComponent` to receive a style prop, making react solely responsible for the cell inversion.

## Test plan
Check whether the Inverted FlatList Example behaves properly on android.
I tested the example on all rn versions from rn@0.69

### DIFF
--- a/src/reanimated2/component/FlatList.tsx
+++ b/src/reanimated2/component/FlatList.tsx
@@ -23,7 +23,6 @@ const createCellRendererComponent = (
   itemLayoutAnimation?: ILayoutAnimationBuilder
 ) => {
   const CellRendererComponent = (props: CellRendererComponentProps) => {
-    console.log(itemLayoutAnimation);
     return (
       <AnimatedView
         // TODO TYPESCRIPT This is temporary cast is to get rid of .d.ts file.
--- a/src/reanimated2/component/FlatList.tsx
+++ b/src/reanimated2/component/FlatList.tsx
@@ -2,7 +2,7 @@
 import type { ForwardedRef } from 'react';
 import React, { Component, forwardRef } from 'react';
 import type { FlatListProps, LayoutChangeEvent } from 'react-native';
-import { FlatList, StyleSheet } from 'react-native';
+import { FlatList } from 'react-native';
 import { AnimatedView } from './View';
 import { createAnimatedComponent } from '../../createAnimatedComponent';
 import type { ILayoutAnimationBuilder } from '../layoutReanimation/animationBuilder/commonTypes';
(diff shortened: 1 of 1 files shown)


src/reanimated2/component/FlatList.tsx
```diff
@@ -79,8 +69,8 @@ export const ReanimatedFlatList = forwardRef(
     }
 
     const cellRenderer = React.useMemo(
```
tomekzaw, line 71:
```suggestion
    const CellRendererComponent = React.useMemo(
```

src/reanimated2/component/FlatList.tsx
```diff
@@ -79,8 +69,8 @@ export const ReanimatedFlatList = forwardRef(
...
-      () => createCellRenderer(itemLayoutAnimation, cellStyle),
-      [cellStyle]
+      () => createCellRenderer(itemLayoutAnimation),
+      []
```
tomekzaw, line 73: why is it empty

src/reanimated2/component/FlatList.tsx
```diff
@@ -12,25 +12,21 @@ import { LayoutAnimationConfig } from './LayoutAnimationConfig';
...
+const createCellRenderer = (itemLayoutAnimation?: ILayoutAnimationBuilder) => {
+  const cellRenderer = (props: CellRendererProps) => {
     return (
       <AnimatedView
         // TODO TYPESCRIPT This is temporary cast is to get rid of .d.ts file.
         layout={itemLayoutAnimation as any}
         onLayout={props.onLayout}
-        style={cellStyle}>
+        style={props.style}>
         {props.children}
```
tomekzaw, line 30: why not `{...restProps}` here?

src/reanimated2/component/FlatList.tsx
bartlomiejbloniarz: Adding `itemLayoutAnimation` to the dependency array would cause issues e.g. if we had `itemLayoutAnimation={Layout.springify()}`, every time the FlatList rerenders (e.g. when an item was added) a new `Layout` object would be created causing a new `CellRendererComponent` to be created. Since every element of the list is wrapped with `CellRendererComponent` it would make all items exit and reenter on every render.

So normally for 
```js
<Animated.FlatList
          itemLayoutAnimation={Layout.springify()}
          ...
/>
```
we have:

https://github.com/software-mansion/react-native-reanimated/assets/56109050/8ceb1b5e-f77e-4c5f-beee-ce7d7a4b3a1d

but if we changed our implementation to
```js
      () => createCellRenderer(itemLayoutAnimation),
      [itemLayoutAnimation]

```
we would have:

https://github.com/software-mansion/react-native-reanimated/assets/56109050/fa1da14f-5426-4ae8-89c6-fab07a030ddf

tomekzaw: okay, makes sense to me 👍 tanks for clarification:

src/reanimated2/component/FlatList.tsx
```diff
@@ -12,31 +12,30 @@ import { LayoutAnimationConfig } from './LayoutAnimationConfig';
...
 
-const createCellRenderer = (
-  itemLayoutAnimation?: ILayoutAnimationBuilder,
-  cellStyle?: StyleProps
+const createCellRendererComponent = (
+  itemLayoutAnimation?: ILayoutAnimationBuilder
 ) => {
-  const cellRenderer = (props: AnimatedFlatListProps) => {
+  const CellRendererComponent = (props: CellRendererComponentProps) => {
+    console.log(itemLayoutAnimation);
```
piaskowyk, line 26: I think we want to remove this console log 😅

piaskowyk (commented):
Just one more thinks, leave a note in PR description that you check if it works fine for older RN versions too

src/reanimated2/component/FlatList.tsx
bartlomiejbloniarz: 😭

piaskowyk (approved):

tomekzaw (approved):

PR #5298, opened 2023-10-24
Fix `delay` in web LA

m-bert:
## Summary

Right now if you modify layout transition with `delay` on web, component will go to its destination, then it will appear back on its original place and the animation will trigger. To fix that I set `transform` of the element to keep it in original place, and then remove it in `onanimationstart`.

## Test plan

Checked that examples run correctly. Also checked on `BasicLayoutAnimation` example with added `delay` and on the code below:

<details> <summary> Test code </summary>

```jsx
import Animated, { Layout } from 'react-native-reanimated';

  const [r, setR] = useState(-100);
  const [y, setY] = useState(-100);

  return (
      <Animated.View
        style={{
          width: 100,
          height: 100,
          backgroundColor: 'red',
          marginLeft: r,
          transform: [{ translateX: 10 }],
        }}
        layout={Layout.delay(250)}
      />
      <Animated.View
        style={[
          styles.box2,
          {
            marginLeft: y,
            transform: [{ translateX: -10 }],
          },
        ]}
        layout={Layout.delay(250)}
      />

      <View style={{ display: 'flex', flexDirection: 'row' }}>
        <Button onPress={() => setR(-r)} title="Change R" />
        <Button onPress={() => setY(-y)} title="Change Y" />
      </View>
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },

  box2: {
    width: 100,
    height: 100,
    backgroundColor: 'yellow',
  },
});
```

</details>


src/reanimated2/layoutReanimation/web/animationParser.ts
```diff
@@ -65,3 +67,33 @@ export function convertAnimationObjectToKeyframes(
...
+  transform.forEach((transformObject: RNTransformProp) => {
+    for (const [key, value] of Object.entries(transformObject)) {
+      if (key === 'reversed') {
+        continue;
+      }
+
+      if (key.indexOf('translate') < 0) {
+        transformString += `${key}(${value}) `;
+      } else {
+        transformString += `${key}(${value}px) `;
```
piaskowyk, line 93: What about `deg` from `rotation` property?

src/reanimated2/layoutReanimation/web/componentUtils.ts
```diff
@@ -88,6 +90,27 @@ function getReversedFromConfig(config: CustomConfig) {
...
+
+  if (typeof style.transform === 'string') {
+    throw new Error('[Reanimated] String transform is currently unsupported.');
+  }
+
+  if (!Array.isArray(style)) {
+    return style.transform;
+  }
+
+  // Only last transform should be considered
```
piaskowyk, line 106: what in situation like that:
```js
style={[{ transform: [{ rotation: 90deg }] }, { transform: [{ scale: 1.5 }] }]}
```

We can try to merge those transforms, or at least we should show a warning to the user that we consider only the last transform from style.

src/reanimated2/layoutReanimation/web/componentUtils.ts
```diff
@@ -201,7 +226,16 @@ export function handleLayoutTransition(
     existingTransform
   );
 
-  setElementAnimation(element, animationConfig);
+  const tmpTransform = existingTransform
```
piaskowyk, line 229:
```suggestion
  const transformCopy = existingTransform
```
sounds better for me 😅

src/reanimated2/layoutReanimation/web/componentUtils.ts
m-bert: If you do something like this, only last transform will be applied to element. I've thought about merging them, but it would be inconsistent with default behavior without reanimated.

src/reanimated2/layoutReanimation/web/componentUtils.ts
m-bert: Done in [68f3259](https://github.com/software-mansion/react-native-reanimated/pull/5298/commits/68f325953fa2d9b6c7cffbb6b421b8387f6152a9)

src/reanimated2/layoutReanimation/web/animationParser.ts
m-bert: I don't think you can pass `rotate` property without specifying either `deg` or `rad`. I've tried that, adding any of them works, but using just a number results in component having no rotation.

piaskowyk (approved):

nandorojo:
Hey guys, are layout animations for web published?

tomekzaw:
@nandorojo They will be released in 3.6.0 and we plan to announce it officially after the release. But you can already use them via `react-native-reanimated@nightly`.

PR #5306, opened 2023-10-25
Run listeners when calling sv.modify

tomekzaw:
## Summary

This PR fixes `modify` method of shared value so that it runs all listeners after the change even if the reference to the value does not change.

Co-authored-by: @kmagiera 
Co-authored-by: @piaskowyk 

## Test plan

ModifyExample.tsx

### DIFF
--- a/app/src/examples/ModifyExample.tsx
+++ b/app/src/examples/ModifyExample.tsx
@@ -1,0 +1,46 @@
+import { Button, StyleSheet, View } from 'react-native';
+
+import React from 'react';
+import {
+  runOnUI,
+  useDerivedValue,
+  useSharedValue,
+} from 'react-native-reanimated';
+
+export default function ModifyExample() {
+  const sv = useSharedValue([1]);
+
+  const handleModify = () => {
+    sv.modify((value) => {
+      'worklet';
+      value.push(value.length + 1);
+      return value;
+    });
+  };
+
+  useDerivedValue(() => {
+    console.log('useDerivedValue', sv.value);
+  });
+
+  const handleRead = () => {
+    console.log('JS', sv.value);
+    runOnUI(() => {
+      console.log('UI', sv.value);
+    })();
+  };
+
+  return (
+    <View style={styles.container}>
+      <Button title="Modify" onPress={handleModify} />
+      <Button title="Read" onPress={handleRead} />
+    </View>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+});
--- a/app/src/examples/ModifyExample.tsx
+++ b/app/src/examples/ModifyExample.tsx
@@ -18,6 +18,10 @@ export default function ModifyExample() {
     });
   };
 
+  const handleModifyWithoutArgument = () => {
+    sv.modify();
+  };
+
   useDerivedValue(() => {
     console.log('useDerivedValue', sv.value);
   });
@@ -32,6 +36,10 @@ export default function ModifyExample() {
   return (
     <View style={styles.container}>
       <Button title="Modify" onPress={handleModify} />
+      <Button
+        title="Modify without argument"
+        onPress={handleModifyWithoutArgument}
+      />
       <Button title="Read" onPress={handleRead} />
     </View>
   );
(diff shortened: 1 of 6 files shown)


kmagiera:
One think we discussed with @wcandillon amd that'd be cool to add us to allow for modified function to accept no arguments in which case it should act as if you passed an identity function, that is it should trigger the effects only and not do any updates. This can be used in situations when it isn't practical to place the code that updates value inside modify but we still want to mutate the value and then let the framework know that changes have been applied.

wcandillon:
@tomekzaw The examples from the PR are compelling.

Currently we a trick to get it to work on Skia: https://github.com/Shopify/react-native-skia/pull/1928/files#diff-a7bfffe910a0f5b40dc0e93e13c3b4f262313ddfe907d0971fa91f7696fe9ab0R60
The no argument version of this would be nice but probably something we can live without too.

tomekzaw:
@kmagiera Correct me if I'm wrong, do you mean that `sv.modify()` should behave like `sv.modify((x) => x)`?

Personally, I find `sv.modify()` on its own a bit unclear; we modify the value but at the same time don't give a recipe what to do. I like the second variant better, it gives a correct impression that it's a trick on the library user side to trigger the effects.

How about `sv.notifyListeners()`, `sv.triggerListeners()` or `sv.callListeners()`? The name is analogous to already existing `addListener`/`removeListener` methods and shifts the focus from the value to the effects.

wcandillon:
@tomekzaw about the name, this might sound strange but to me anything with the word listeners in it sounds slow to me (weird bias 😅). Maybe if you replace the word listeners with Change. or `modify` sounds good too :)

tomekzaw:
@wcandillon That's an interesting point, the shorter the name of the method, the faster it is? 😄 

Another alternative would be `sv.runEffects()`, running is pretty fast, right? 😄

kmagiera:
I don't like the idea of adding more methods to cover this type of very infrequent usecases. In addition I'd consider names revolving around "effects" and "listeners"  to leak abstraction details that we don't expose in other ways. It is true that empty argument is kind of cryptic but seems to me like a good tradeoff between not wanting to increasing the API surface area while giving a decent escape hatch that covers this one specific usecase.

app/src/examples/ModifyExample.tsx
```diff
@@ -0,0 +1,46 @@
...
+      return value;
+    });
+  };
+
+  useDerivedValue(() => {
+    console.log('useDerivedValue', sv.value);
+  });
+
+  const handleRead = () => {
+    console.log('JS', sv.value);
```
kmagiera, line 26: I assume that we expect in this case for these to be different. We should add a comment about that so that it doesn't confuse people testing this in the future.

kmagiera (approved):
Accepting as the discussion about `modify(undefined)` should't impact the code this much. 

One more thing to check is that I remember @wcandillon reporting earlier that `modify` isn't exposed in public TS types. Would be good to verify that. Also, given that there are no docs to documentation it'd be good to add some details there or at least make a note of that such that we can work on that in the near future.

tomekzaw:
> I don't like the idea of adding more methods to cover this type of very infrequent usecases.

Since it's an infrequent usecase indeed, why do we need to have additional support for the no-argument variant when passing `(x) => x` works just fine?

> In addition I'd consider names revolving around "effects" and "listeners" to leak abstraction details that we don't expose in other ways.

Both `addListener` and `removeListener` are already a part of the public API (see [here](https://github.com/software-mansion/react-native-reanimated/blob/1210203c4bd044f875610b27bbd0a29b2403775a/src/reanimated2/commonTypes.ts#L11-L16)) so we already leak this abstraction.

> I remember @wcandillon reporting earlier that `modify` isn't exposed in public TS types. Would be good to verify that.

I believe it's already exposed. Accessing `.modify` in the example seems to work fine as opposed to `_value` which isn't meant to be public indeed.

Zrzut ekranu 2023-10-26 o 11 53 59

> Also, given that there are no docs to documentation it'd be good to add some details there or at least make a note of that such that we can work on that in the near future.

Sure, I'll mention `modify` method in the "Remarks" sections (see [here](https://docs.swmansion.com/react-native-reanimated/docs/core/useSharedValue/#remarks)).

piaskowyk (approved):

piaskowyk (approved):

src/reanimated2/commonTypes.ts
```diff
@@ -12,7 +12,7 @@ export interface SharedValue<Value> {
   value: Value;
   addListener: (listenerID: number, listener: (value: any) => void) => void;
   removeListener: (listenerID: number) => void;
-  modify: (modifier: (value: any) => any) => void;
+  modify: (modifier?: (value: any) => any) => void;
```
wcandillon, line 15: yep this is perfect thks ☺️☺️☺️

PR #5309, opened 2023-10-26
Fix version routing for animations

Latropos:
## Summary

1. We want to keep the same topic when changing docs version
2. This taks is automatically done when we have analogous routing paths for both versions
3. We can use [this docusarus plugin](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects) to automatically redirect client fro one page to some other
4. Unfortunately changing version in dropdown skips this step (redirect) and we can't use  `plugin-client-redirects` to link two doc versions
5. But we change default paths of files in versioned docs and link previous paths to the new ones. T**his would make all external links work.**

> **Note**
> Redirection of `plugin-client-redirects` works only in production build.

<table>
<tr><td>BEFORE</td><td>AFTER</td><tr>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/b0d99d6b-83bf-4258-928e-27d9346555ad

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/d26c0b31-0721-4232-af6e-b367355a9519

</td><tr>
</table>


kacperkapusciak (changes requested):
This is going to break the links someone has saved on their tab bar in the browser or links in the older blogposts etc.

We need to take a step back and think why does Google rank the older pages higher than the new ones rather than artifically redirecting it on the frontend

How can we make Google to rank Reanimated 3 pages higher? This would solve our problem the proper way

Latropos:
We won't fix it. We will change doc priorities instead

PR #5318, opened 2023-10-27
Improve counter example

tomekzaw:
## Summary

1. Rename `AnimatedTextInputExample` to `CounterExample`
2. Use `defaultValue` prop instead of `value`
3. Add `font-variant: 'tabular-nums';`

### DIFF
--- a/app/src/examples/CounterExample.tsx
+++ b/app/src/examples/CounterExample.tsx
@@ -4,13 +4,7 @@ import Animated, {
   useSharedValue,
   withTiming,
 } from 'react-native-reanimated';
-import {
-  Button,
-  TextInput,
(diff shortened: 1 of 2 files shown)


piaskowyk (approved):

Noitidart:
Just curious, why was `defaultValue` added? What impact does that have?

tomekzaw:
If you pass `defaultValue` in animated style, you can omit `value` prop in render function and the counter still works as expected.

Noitidart:
Thanks very much for your fast reply. Is this done for performance or something though? Because in the example we are setting both at the same time, value and defaultValue. Sorry for the follow up.

PR #5319, opened 2023-10-27
Update paths in iOS/tvOS/macOS workflows

tomekzaw:
## Summary

I've noticed that MacOSExample CIs are started even for JS-only PRs because `on.push.paths` was not specified.

Additionally, I fixed paths in iOS and tvOS workflows (`ios` &rarr; `apple`).


tjzel (approved):

piaskowyk (approved):
👏

PR #5320, opened 2023-10-27
Organise type tests better

tjzel:
## Summary

I tried to move tests into their separate 'categories', cleaned-up the file a bit and moved kind-of irrelevant tests to new `legacy.tsx` file.

## Test plan

Test the tests 🧠


piaskowyk (commented):
It can be challenging to assess the testing coverage and identify any missing aspects when dealing with a single file that has 1500 lines of code. Splitting the file into separate domain-specific files could be a helpful approach. For instance, you could consider categorizing the code into files such as:
- Event types
- Component creation types
- List types
- Style types
- Animation types
- ...

This kind of structure can enhance code organisation and make it easier to understand and maintain. It allows us to focus on specific domains while working with the tests. What do you think about this kind of structure?

tjzel:
@piaskowyk I think it's a very good idea - do you think it should be done in this PR or maybe do the split in the next one?

piaskowyk:
I don't have a strong opinion about splitting it into another PR. Let's proceed with whatever option works best for you.

piaskowyk (approved):

tjzel:
Then I'd rather do it in the future PR for a smoother transition. But if it's not done in a week or two, remember to nudge me about it 😄

piaskowyk:
sure thing 🫡

PR #5321, opened 2023-10-27
Add babel preset

bartlomiejbloniarz:
## Summary

This PR adds a babel preset for reanimated. This preset can be used to automatically configure our plugin based on the platform. It allows us to remove unnecessary native code on web, without bothering users. It relies on bundler setting the `platform` (or `target`) property on its `caller`. From what I've seen metro and weback both set the `platform` property. In next.js however it is not defined, but from what I've seen it sets the `target` property to `web` when it is generating the client side.

## Test plan

Verify the behavior of example apps.


WebExample/babel.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = function (api) {
     console.log('Starting Web example without Babel plugin.');
   }
   return {
-    presets: ['babel-preset-expo'],
+    presets: [['../preset', {disableBabelPlugin}], 'babel-preset-expo'],
```
tomekzaw, line 9: The idea here was that the user doesn't need to setup Babel plugin, so now we should disable Babel preset.

preset/index.js
```diff
@@ -0,0 +1,16 @@
...
+  const isWeb = platform === 'web';
+
+  return {
+    plugins: [['../plugin', { isWeb }]],
+  };
+};
+
+function getPlatform(caller) {
+  return caller && (caller.platform ?? caller.target);
+}
```
tomekzaw, line 16: I don't like the fact that `getPlatform` returns `platform` or `target`. How about renaming this function to `isWeb(caller)` and returning boolean from it?

preset/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
```
tomekzaw, line 1: No need for `tsconfig.json` if we don't have any TS files here?

bartlomiejbloniarz: Yes, I just added it for this draft, because otherwise commit hooks wouldn't let me push it and I decided to deal with those issues later.

PR #5329, opened 2023-10-30
Fix Expo Dev Client CI

piaskowyk:
## Summary

By default, Expo uses npm as a package manager. When you execute `npm install <library>`, npm installs the dependencies and runs the `prepare` script from the library. This setup generally works well. However, on our CI environment, we install Reanimated directly from the branch, without the autogenerated files and types. These files are usually generated by the `prepare` command. The problem arises with the plugin since there may be a Babel version mismatch between Expo and the plugin. This mismatch causes the plugin build to fail, resulting in an error.
```
npm ERR!   Type 'import("/home/runner/.npm/_cacache/tmp/git-clonesZNSUp/plugin/node_modules/@babel/types/lib/index").FunctionExpression' is not assignable to type 'babel.types.FunctionExpression'.
npm ERR!     Types of property 'id' are incompatible.
npm ERR!       Type 'import("/home/runner/.npm/_cacache/tmp/git-clonesZNSUp/plugin/node_modules/@babel/types/lib/index").Identifier | null | undefined' is not assignable to type 'babel.types.Identifier | null | undefined'.
```
More logs here: https://github.com/software-mansion/react-native-reanimated/actions/runs/6697889174/job/18198727220

My solution is to use the nightly version of Reanimated from NPM. If anyone has a better approach, I'm open to discussion and would love to hear alternative suggestions. 🤗

## Test plan

See CI reults


tomekzaw:
> My solution is to use the nightly version of Reanimated from NPM. If anyone has a better approach, I'm open to discussion and would love to hear alternative suggestions. 🤗

If we use nightly version from npm, this CI will be out-of-date by 1 day which may be confusing.

How about we modify this CI so it first builds the package (using `./createNPMPackage.sh`) and then adds a dependency on Reanimated from .tgz (using `yarn add file://...`)?

piaskowyk:
> If we use nightly version from npm, this CI will be out-of-date by 1 day which may be confusing.

Actually, what I did was reschedule this CI to run after the CI that publishes the nightly version on NPM. However, I understand that this is not a perfect solution.

> How about we modify this CI so it first builds the package (using ./createNPMPackage.sh) and then adds a dependency on Reanimated from .tgz (using yarn add file://...)?

Sure, I can do that 🫡

tomekzaw:
> Actually, what I did was reschedule this CI to run after the CI that publishes the nightly version on NPM. 

oh right, I must have overlooked it, sorry

tjzel (approved):

PR #5330, opened 2023-10-30
Add react-hooks eslint plugin

szydlovsky:
## Summary

I added _react-hooks_ eslint plugin to repo. It is a rule that checks integrity of used hooks, as well as their dependencies and general standards of their implementation and use. I have stumbled upon a custom hook that doesn't start with _"use"_ - which has been changed. There are some conditional hook calls in the repo - I have temporarily disabled eslint errors regarding them and await any ideas on what to do there.

## Test plan

Delete any of the _"eslint-disable..."_ comments added in [this commit](https://github.com/software-mansion/react-native-reanimated/commit/00c37e8c0d4fe2529cbca68b9d87ec2a7ce0c35a).  To see the dependencies' fixes and suggestions change dependencies [in this line](https://github.com/software-mansion/react-native-reanimated/blob/c11b671b5dd58b0b18df941727d9d8336d9227e6/app/src/examples/MatrixTransform.tsx#L46) to: `[matrix2, currentTransformIndex]`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'prettier',
     'plugin:import/typescript',
+    'plugin:react-hooks/recommended',
   ],
   plugins: ['react', 'react-native', 'import', 'jest', '@typescript-eslint'],
   env: {
--- a/package.json
+++ b/package.json
@@ -145,6 +145,7 @@
     "eslint-plugin-no-inline-styles": "^1.0.5",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^6.0.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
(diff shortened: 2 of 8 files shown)


tjzel:
Why do we enable those rules if we're ignoring the warnings/errors they produce? I'm confused (or maybe the ignores are just on the draft version and I'm commenting here too hasty 😵)

szydlovsky:
@tjzel exactly the ignores are on the draft version and I hope we can come up with something for each of these cases 👌

tomekzaw (approved):

tomekzaw:
@tjzel I've decided to merge this PR immediately because I think it adds some value even in its current form. First of all, it prevents us from breaking rules of hooks in the future, not only in the library but also in the example app. Secondly, all misuses of hooks will need to be addressed in separate PRs anyway because they are used in critical pieces of Reanimated.

piaskowyk (approved):

tjzel:
🤔

PR #5332, opened 2023-10-31
useWorkletCallback  dependencies fix

szydlovsky:
## Summary

Custom hook named _**useWorkletCallback**_ (a workletized version of _**useCallback**_) turns out to be a bit broken - when given undefined dependencies array, it doesn't try to infer them from worklet function. This PR aims to change it.

## Test plan

You can put my sample code into EmptyExample.tsx (the example only shows the dependencies' fix - it doesn't capitalize on any intended memoization):

```js
import { Text, StyleSheet, View, TouchableOpacity } from 'react-native';
import { useWorkletCallback } from 'react-native-reanimated';

  const [text, setText] = useState<string>('');

  const makeRandomString = () => {
    let result = '';
    const length = 6;
    const characters =
      'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
    const charactersLength = characters.length;
    let counter = 0;
    while (counter < length) {
      result += characters.charAt(Math.floor(Math.random() * charactersLength));
      counter += 1;
    }
    return result;
  };

  const changeString = () => {
    const newText = makeRandomString();
    setText(newText);
  };

  const unneededDependency = 0;
  /*
    If you change the hook's dependencies to unneededDepdendency - it will stop properly updating.
    However, if you completely remove the dependencies array, improved useWorkletCallback will properly
    add 'text' to dependencies (as it rightfully should from the very beginning).
  */
  const getNewestText = useWorkletCallback(() => {
    console.log(text);
    return text;
  }, [text]);

  return (
      <TouchableOpacity
        style={styles.button}
        onPress={() => {
          changeString();
        }}>
        <Text style={styles.text}>Change string</Text>
      </TouchableOpacity>
      <Text style={styles.text}>
        The random string is
        <Text style={styles.randomString}>{` ${getNewestText()}`}</Text>
      </Text>
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  button: {
    width: 200,
    height: 60,
    backgroundColor: 'cyan',
    justifyContent: 'center',
    marginBottom: 50,
  },
  text: {
    fontSize: 25,
    fontWeight: '300',
    color: 'black',
    textAlign: 'center',
  },
  randomString: {
    fontWeight: '500',
  },
});

```

### DIFF
--- a/src/reanimated2/hook/useWorkletCallback.ts
+++ b/src/reanimated2/hook/useWorkletCallback.ts
@@ -1,10 +1,14 @@
 'use strict';
 import { useCallback } from 'react';
 import type { DependencyList } from './commonTypes';
+import type { WorkletFunction } from '../commonTypes';
 
 export function useWorkletCallback<Args extends unknown[], ReturnValue>(
   worklet: (...args: Args) => ReturnValue,
   deps?: DependencyList
 ) {
-  return useCallback(worklet, deps ?? []);
+  return useCallback(
+    worklet,
+    deps ?? Object.values((worklet as WorkletFunction).__closure)
+  );
 }


src/reanimated2/hook/useWorkletCallback.ts
```diff
@@ -1,10 +1,14 @@
...
+import type { WorkletFunction } from '../commonTypes';
 
 export function useWorkletCallback<Args extends unknown[], ReturnValue>(
   worklet: (...args: Args) => ReturnValue,
   deps?: DependencyList
 ) {
-  return useCallback(worklet, deps ?? []);
+  return useCallback(
+    worklet,
+    deps ?? Object.values((worklet as WorkletFunction).__closure)
```
tomekzaw, line 12: I'm wondering if this also works on web in case we don't use Reanimated Babel plugin.

Here's a code snippet from `useDerivedValue` that does exactly the same thing, maybe we can copy it over to here?
https://github.com/software-mansion/react-native-reanimated/blob/a2ae77e0386da38989eaebcd3ca1699ca00b6559/src/reanimated2/hook/useDerivedValue.ts#L22-L28

szydlovsky:
Turns out that basing off of `useDerivedValue`'s code regarding dependencies is complicated and difficult to read, so the PR goes into draft mode and gets postponed until more important things are dealt with 👍

piaskowyk:
At this moment we decided to close PR, due to lower usage of this hook, but we need to consider how we deal with dependencies in our other hooks

PR #5353, opened 2023-11-09
Fix lint staged

Latropos:
## Summary
BEFORE:
* Lint-staged lints all files and prints all warnings and errors

NOW:
* Lint only staged files and shows both errors and warnings
* If the previous tasks success all files get lined, but only errors are printed

Drawbacks:
Lint runs twice. It should not cause burdensome delays since the first lint runs only for the staged files, so if you stage <10 files it runs very fast.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,4 +4,3 @@ docs/*
 plugin/*
 eslintrc.js
 jest.config.js
-__typetests__/*
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,7 @@
 .eslintrc.js
 scripts/*.js
+docs/*
+plugin/*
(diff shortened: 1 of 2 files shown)


piaskowyk (changes requested):
Let's lint all file as it was but display only errors without warnings

Latropos:
>Let's lint all file as it was but display only errors without warnings
It would be better to display all errors and only the new warnings. Best I can do is to display errors for the whole `src` folder and both errors and warnings for the staged files

piaskowyk:
> Best I can do is to display errors for the whole src folder and both errors and warnings for the staged files

If you want you can, but displaying just errors from `/src` is enough

PR #5363, opened 2023-11-10
Improve TS and ESLint in src

marmor157:
## Summary

This PR removes as many `any` as possible and removes unnecessary `// eslint-disable` comments alongside fixing some of the eslint errors


src/reanimated2/Colors.ts
```diff
@@ -496,14 +493,18 @@ export function RGBtoHSV(r: any, g?: any, b?: any): HSV {
       h = r - g + d * 4;
       h /= 6 * d;
       break;
+    case min:
+    /* fallthrough */
```
tomekzaw, line 497:
```suggestion
```

src/reanimated2/ViewDescriptorsSet.ts
```diff
@@ -21,10 +21,10 @@ export function makeViewDescriptorsSet(): ViewDescriptorsSet {
   const data: ViewDescriptorsSet = {
     shareableViewDescriptors,
     add: (item: Descriptor) => {
-      shareableViewDescriptors.modify((descriptors: Descriptor[]) => {
+      shareableViewDescriptors.modify((descriptors: any): Descriptor[] => {
```
tomekzaw, line 24: 😱

src/reanimated2/ViewDescriptorsSet.ts
```diff
@@ -36,10 +36,10 @@ export function makeViewDescriptorsSet(): ViewDescriptorsSet {
     },
 
     remove: (viewTag: number) => {
-      shareableViewDescriptors.modify((descriptors: Descriptor[]) => {
+      shareableViewDescriptors.modify((descriptors: any): Descriptor[] => {
```
tomekzaw, line 39: 🙀

src/reanimated2/commonTypes.ts
```diff
@@ -8,13 +9,22 @@ export interface StyleProps extends ViewStyle, TextStyle {
   [key: string]: any;
 }
 
+interface SharedValueInternals<T> {
```
tomekzaw, line 12: is it a common practice to prefix such types with `Internals`?

src/reanimated2/commonTypes.ts
```diff
@@ -8,13 +9,22 @@ export interface StyleProps extends ViewStyle, TextStyle {
...
+  _value?: T | Descriptor | AnimatableValue;
+}
+
 export interface SharedValue<Value> {
   value: Value;
-  addListener: (listenerID: number, listener: (value: any) => void) => void;
+  addListener: (listenerID: number, listener: (value: Value) => void) => void;
   removeListener: (listenerID: number) => void;
-  modify: (modifier?: (value: any) => any) => void;
+  modify: (modifier?: (value: Value) => unknown) => void;
```
tomekzaw, line 22:
```suggestion
  modify: (modifier?: (value: Value) => Value | AnimatableValue) => void;
```

src/reanimated2/component/ScrollView.tsx
```diff
@@ -25,9 +25,7 @@ interface AnimatedScrollViewInterface extends ScrollView {
   getNode(): ScrollView;
 }
 
-const AnimatedScrollViewComponent = createAnimatedComponent(
-  ScrollView as any
-) as any;
+const AnimatedScrollViewComponent = createAnimatedComponent(ScrollView);
```
tomekzaw, line 28: ❤️

src/reanimated2/hook/useAnimatedSensor.ts
```diff
@@ -138,5 +138,5 @@ export function useAnimatedSensor(
     };
   }, [sensorType, userConfig]);
 
-  return ref.current;
+  return ref.current as AnimatedSensor<ValueRotation> | AnimatedSensor<Value3D>;
```
tomekzaw, line 141: the return type here depends on `SensorType`, can we enforce it, perhaps with function overloading?

src/reanimated2/isSharedValue.ts
```diff
@@ -1,7 +1,7 @@
 'use strict';
 import type { SharedValue } from './commonTypes';
 
-export function isSharedValue<T>(value: any): value is SharedValue<T> {
+export function isSharedValue<T>(value: unknown): value is SharedValue<T> {
   'worklet';
-  return value?._isReanimatedSharedValue === true;
+  return (value as Record<string, unknown>)?._isReanimatedSharedValue === true;
```
tomekzaw, line 6: `isSharedValue` is a public function and user can pass `null` or `undefined` to it so we probably shouldn't cast it to `Record` here

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -160,16 +160,14 @@ class InnerKeyframe implements IEntryExitAnimationBuilder {
...
-                    index.toString() + '_transform:' + transformProp,
-                    transformStyle[transformProp]
-                  );
-                });
-              }
-            );
+            keyframe.transform.forEach((transformStyle, index) => {
+              Object.keys(transformStyle).forEach((transformProp: string) => {
+                addKeyPointWith(
+                  index.toString() + '_transform:' + transformProp,
```
tomekzaw, line 166: Let's use string interpolation here (in a separate PR)

src/reanimated2/layoutReanimation/web/componentUtils.ts
```diff
@@ -17,15 +17,12 @@ import { _updatePropsJS } from '../../js-reanimated';
...
+import { isReducedMotion } from '../../PlatformChecker';
 
 function getEasingFromConfig(config: CustomConfig): string {
   const easingName = (
-    config.easingV !== undefined &&
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    config.easingV!.name in WebEasings
-      ? // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        config.easingV!.name
+    config.easingV !== undefined && config.easingV?.name in WebEasings
```
tomekzaw, line 24: what about simply `config?.easingV?.name in WebEasings`

src/reanimated2/mappers.ts
```diff
@@ -6,18 +6,18 @@ import { isSharedValue } from './isSharedValue';
...
 const IS_JEST = isJest();
 
-type Mapper = {
+type Mapper<T> = {
   id: number;
   dirty: boolean;
   worklet: () => void;
-  inputs: SharedValue<any>[];
-  outputs?: SharedValue<any>[];
+  inputs: unknown[];
```
tomekzaw, line 13: somewhere above it was `Record<string, unknown>`, no?

src/reanimated2/mappers.ts
```diff
@@ -131,17 +131,21 @@ function createMapperRegistry() {
     }
   }
 
-  function extractInputs(
-    inputs: any,
-    resultArray: SharedValue<any>[]
-  ): SharedValue<any>[] {
+  function isBaseObject(obj: unknown): obj is object {
```
tomekzaw, line 134:
```suggestion
  function isPlainObject(obj: unknown): obj is object {
```

btw I thought we already have such function somewhere

tomekzaw (commented):
Looks good so far! 🚀

src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -88,7 +88,9 @@ export function createAnimatedComponent<P extends object>(
 export function createAnimatedComponent(
   Component: ComponentType<InitialComponentProps>,
   options?: Options<InitialComponentProps>
-): any {
+):
+  | FunctionComponent<AnimateProps<InitialComponentProps>>
+  | ComponentClass<AnimateProps<InitialComponentProps>> {
```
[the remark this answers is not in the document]
marmor157: I've taken it from types above, but as I've looked in  the code it turns out it's the other way around, it always returns `AnimatedComponent` wrapped in `React.forwardRef`. So actually I will remove the explicit return type and let it infer it.

src/reanimated2/commonTypes.ts
marmor157: I've tried looking it up in Google, but haven't found anything. I've just went here with a gut-feeling

src/reanimated2/commonTypes.ts
marmor157: This does not work. It breaks `InlineStylesTest`. If I understand this correctly, we are making here a little bit of circular dependency, in case of `Animated.View` `style` which is of type `StyleProp<ViewStyle>` needs to extend `Value`(as it is passed to `modifier`) and `Value` needs to extend `StyleProp<ViewStyle>`(as it's returned from `modifier`).

The problem is that when we do it like this
```
  function InlineStylesTest13() {
    const sv = useSharedValue({ transform: [{ translateX: 0 }] });
    return <Animated.View style={sv} />;
  }
```
type of `sv` is 

```
SharedValue<{
    transform: {
        translateX: number;
    }[];
}>
```
and `ViewStyle` is not compatible wit as `transform` in it is not required.

If we do it like this, it will work 
```
  function InlineStylesTest13() {
    const sv = useSharedValue<ViewStyle>({ transform: [{ translateX: 0 }] });
    return <Animated.View style={sv} />;
  }
```

but this requires users to specify the `ViewStyle` in each `SharedValue` they use for that, which is terrible.

That being said, I've found another solution xD

src/reanimated2/isSharedValue.ts
marmor157: Yeah, and `?.` operator will just return `undefined` in those cases

src/reanimated2/hook/useAnimatedSensor.ts
marmor157: It is enforced in function overloading starting at line 66. The problem is that Typescript doesn't care about function declaration types in function implementation typing, and it requires it to handle union of input types and union of return types in EVERY configuration possible.

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
marmor157: https://github.com/software-mansion/react-native-reanimated/pull/5376

src/reanimated2/layoutReanimation/web/componentUtils.ts
marmor157: Typescript for some reason doesn't allow to pass `undefined` to `in` operator but it's not something simple `as` can't fix xD

src/reanimated2/ViewDescriptorsSet.ts
marmor157: ([5ab6d88](https://github.com/software-mansion/react-native-reanimated/pull/5363/commits/5ab6d8838cb543cb10c953e5a71841e8fbf51dda))

src/reanimated2/ViewDescriptorsSet.ts
marmor157: ([5ab6d88](https://github.com/software-mansion/react-native-reanimated/pull/5363/commits/5ab6d8838cb543cb10c953e5a71841e8fbf51dda))

src/reanimated2/mappers.ts
marmor157: [203e122](https://github.com/software-mansion/react-native-reanimated/pull/5363/commits/203e12242a33d3199c7060fa9ec926763d785c58)

src/reanimated2/mappers.ts
marmor157: [98d3164](https://github.com/software-mansion/react-native-reanimated/pull/5363/commits/98d3164603d97e5f6800c266e7d7f79507de2a9e)

src/reanimated2/shareables.ts
```diff
@@ -44,7 +44,7 @@ export function registerShareableMapping(
   _shareableCache.set(shareable, shareableRef || _shareableFlag);
 }
 
-function isPlainJSObject(object: object) {
+export function isPlainJSObject(object: unknown): object is object {
```
tomekzaw, line 47: Calling `Object.getPrototypeOf(null)` or `Object.getPrototypeOf(undefined)` throws a `TypeError` so it cannot be `unknown` I guess.

marmor157: Yeah, you are right, do you prefer to keep this only for `object` type and do the checks for `undefined` and `null` in `extractInputs` or make this function also accept `unknown` and do the checks here?

marmor157: IMO given that `value` in `shareables` that is passed to this function is `any` and `inputs` in `mappers` are `unknown`, it's better to do the checks in `isPlainJSObject`

tomekzaw: Let's leave it as it is now (`object`) and also make sure that you cannot pass `null` to it

marmor157: [0f7a4e8](https://github.com/software-mansion/react-native-reanimated/pull/5363/commits/0f7a4e85de00f9a4b0cccbf848d0f89f2bc7bfca)

src/reanimated2/mappers.ts
```diff
@@ -142,11 +142,7 @@ function createMapperRegistry<T>() {
       }
     } else if (isSharedValue<T>(inputs)) {
       resultArray.push(inputs);
-    } else if (
-      inputs !== null &&
-      inputs !== undefined &&
-      isPlainJSObject(inputs)
-    ) {
+    } else if (inputs && isPlainJSObject(inputs)) {
```
marmor157, line 145: I've made it like that initially to do not throw away any other possible values, but actually we do not care here about any other [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) values as they would no pass `isPlainJSObject` anyway.

src/reanimated2/isSharedValue.ts
tomekzaw: yep but can we cast `value` which can be `null` to a `Record<string, unknown>`?

marmor157: Well, in typescript you define the types, so yes, as long as you know why you are doing it. In this case, we know that `?.` operator will handle those cases, so we can assume that if `value` isn't `Record<string, unknown>` it will evaluate to `undefined`.

src/reanimated2/Colors.ts
```diff
@@ -462,14 +462,16 @@ export const rgbaColor = (
...
+export function RGBtoHSV(r: RGB | number, g?: number, b?: number): HSV {
   'worklet';
-  /* eslint-disable */
-  if (arguments.length === 1) {
+  if (typeof r === 'object') {
     g = r.g;
     b = r.b;
     r = r.r;
   }
+  // Typescript doesn't know that g and b are now sure to not be undefined
```
tomekzaw, line 472: Can we somehow achieve this using function overloading?

src/reanimated2/Colors.ts
```diff
@@ -496,14 +493,17 @@ export function RGBtoHSV(r: any, g?: any, b?: any): HSV {
       h = r - g + d * 4;
       h /= 6 * d;
       break;
+    case min:
```
tomekzaw, line 496: Do we need `case min:` since we also have `default:` here?

src/reanimated2/commonTypes.ts
```diff
@@ -36,12 +46,12 @@ export type ShareableSyncDataHolderRef<T> = {
...
 };
 
-export type MapperRegistry = {
+export type MapperRegistry<T> = {
   start: (
     mapperID: number,
     worklet: () => void,
-    inputs: SharedValue<any>[],
-    outputs?: SharedValue<any>[]
+    inputs: Record<string, unknown>[],
```
tomekzaw, line 53: is it really correct?

src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts
tomekzaw: plz merge main

src/reanimated2/mappers.ts
```diff
@@ -3,21 +3,22 @@ import type { SharedValue } from './commonTypes';
...
 const IS_JEST = isJest();
 
-type Mapper = {
+type Mapper<T> = {
   id: number;
   dirty: boolean;
   worklet: () => void;
-  inputs: SharedValue<any>[];
-  outputs?: SharedValue<any>[];
+  inputs: Record<string, unknown>[];
```
tomekzaw, line 14: is it correct type?

src/reanimated2/valueSetter.ts
```diff
@@ -13,12 +21,12 @@ export function valueSetter(sv: any, value: any, forceUpdate = false): void {
     typeof value === 'function' ||
     (value !== null &&
       typeof value === 'object' &&
-      (value as AnimationObject).onFrame !== undefined)
+      (value as unknown as AnimationObject).onFrame !== undefined)
```
tomekzaw, line 24: me not likes

src/reanimated2/platformFunctions/measure.ts
```diff
@@ -55,17 +54,16 @@ function measurePaper<T extends Component>(animatedRef: AnimatedRef<T>) {
...
       `[Reanimated] The view with tag ${viewTag} is not a valid argument for measure(). This may be because the view is not currently rendered, which may not be a bug (e.g. an off-screen FlatList item).`
     );
     return null;
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const measured = _measurePaper!(viewTag as number);
-  if (measured === null) {
+  const measured = _measurePaper?.(viewTag as number);
+  if (measured == null) {
```
tomekzaw, line 66:
```suggestion
  if (!measured) {
```

src/reanimated2/Colors.ts
tjzel: Actually took the time to profile those and more solutions, we will make a decision later on how to handle problems with types in function overloads.

src/reanimated2/Colors.ts
tjzel, line 1: Actually removed function overloads here since currently in the code we only use non-object signatures for them.

src/reanimated2/commonTypes.ts
tjzel: Apparently, yes!

Screenshot 2023-12-08 at 10 41 17

src/reanimated2/commonTypes.ts
```diff
@@ -36,12 +46,12 @@ export type ShareableSyncDataHolderRef<T> = {
   __hostObjectShareableJSRefSyncDataHolder: T;
 };
 
-export type MapperRegistry = {
+export type MapperRegistry<T> = {
```
tjzel, line 49: This shouldn't be bound to one type. In extreme situations outputs could be two shared values of different type parameter.

src/reanimated2/isSharedValue.ts
tjzel: This is the best we can do here, since actually TS would like us to use `in` operator here, but we can't because of Host Objects.

src/reanimated2/component/ScrollView.tsx
```diff
@@ -25,7 +25,9 @@ interface AnimatedScrollViewInterface extends ScrollView {
   getNode(): ScrollView;
 }
 
-const AnimatedScrollViewComponent = createAnimatedComponent(ScrollView);
+const AnimatedScrollViewComponent = createAnimatedComponent(
+  ScrollView as any
+) as any;
```
tomekzaw, line 30: why :(

tjzel: Screenshot 2023-12-08 at 11 50 34

src/reanimated2/hook/useAnimatedSensor.ts
```diff
@@ -81,7 +81,7 @@ export function useAnimatedSensor(
 export function useAnimatedSensor(
   sensorType: SensorType,
   userConfig?: Partial<SensorConfig>
-): AnimatedSensor<ValueRotation | Value3D> {
+): AnimatedSensor<ValueRotation> | AnimatedSensor<Value3D> {
```
tjzel, line 84: I'm wondering if it's not a potentially breaking change.

src/reanimated2/valueSetter.ts
tjzel: we will get to Animation Objects one day...

tjzel:
Closing because of #5481 and its follow-ups.

PR #4902, opened 2023-08-09
Enable `NDEBUG` in production builds

tomekzaw:
## Summary

Related to #4901.

This PR properly sets `NDEBUG` flag when building in release mode.

References:
* https://github.com/facebook/react-native/pull/36172
* https://github.com/facebook/react-native/pull/38314
* https://github.com/facebook/react-native/pull/36194

PR #5366, opened 2023-11-10
Fix NDEBUG macro not present in Release builds

tjzel:
## Summary

This PR aims to fix the issue when Release builds would fail on Paper on iOS.
The reason this regression happened is #5113.

XCode ships by default to new projects `DEBUG` macro for Debug Configuration. Due to some issues (#4902) we decided to switch over to `NDEBUG` macro. However, the `NDEBUG` is not provided by XCode out of the box. Incidentally, it's provided by React Native which puts its install script into project's Podfile - however, [only in Fabric](https://github.com/facebook/react-native/commit/421df9ffd58092b1a2dec455a048edb6db1739de). 
What is more, some time ago React Native made it so `PRODUCTION=1 pod install` is [no longer a necessary step](https://github.com/facebook/react-native/commit/93fdcbaed0f69b268e1ae708a52df9463aae2d53) to make a release build, therefore we cannot rely on it. Until React Native defines `NDEBUG` for Paper too we have to detect, based on available options, whether we potentially have a Release build and this pull request does this.

Thanks to https://github.com/software-mansion/react-native-reanimated/pull/5383 we finally figures out how to do it the most agreeable way. We hardcode that for a build config named `Release` NDEBUG=1 will be set. If the user uses some custom release build config e.g. "MyRelease", NDEBUG=1 will not be set - in this situation we'll require the user to use `PRODUCTION=1`.

## Test plan

Run all example apps (except TVOS cause I couldn't make it cooperate with me, and Web) on both iOS and Android in both Release and Debug and see if they work properly (e.g. add a throw in #ifndef fragment).


apple/RNReanimated.xcodeproj/project.pbxproj
tjzel, line 1: This isn't necessary unless the library get's pre-compiled.

TVOSExample/ios/TVOSExample.xcodeproj/xcshareddata/xcschemes/TVOSExample-tvOS.xcscheme
```diff
@@ -41,7 +41,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Debug"
+      buildConfiguration = "Release"
```
tomekzaw, line 44: revert it

Common/cpp/Tools/ReanimatedMacros.h
```diff
@@ -0,0 +1,27 @@
...
+automatically (on Fabric only). Therefore, we do the following:
+- on Android we just look-up NDEBUG
+- on iOS we look-up NDEBUG (we still inject it in .podspec when `PRODUCTION=1`)
+  - if NDEBUG is not defined we check if Fabric is disabled
+    - if Fabric is disabled, we check for out-of-the-box DEBUG
+      - if DEBUG isn't defined it's almost safe to assume a release build
+*/
+
+#ifndef NDEBUG
+#ifdef __APPLE__
```
tomekzaw, line 20: Let's swap these lines to be consistent with the description above.
```suggestion
#ifdef __APPLE__
#ifndef NDEBUG
```

tomekzaw (changes requested):

apple/RNReanimated.xcodeproj/project.pbxproj
tjzel, line 1: This might be required for pre-compilation.

tomekzaw (approved):
i like it

PR #5371, opened 2023-11-13
Allow layout animations to be updated on rerender

bartlomiejbloniarz:
## Summary

This PR changes the current implementation to allow changes in layout animations on rerender.

This PR is the first part of the layout animation configuration overhaul. In a next PR I will extend those changes to SETs and remove `configureLayoutAnimations` altogether.

TODO:

- [x] make it work
- [x] add example
- [x] add batchinator
- [x] take care of web
- [x] docs
- [x] fix issues with changing animations when an animation is already running

# Issues

### Simultaneous layout animations

I encountered an interesting issue, that I was able to reproduce with the following custom layout animation function on the main branch (so without my changes).
```js
function CustomLayoutTransiton(): LayoutAnimationFunction {
  const isEven = makeMutable(1);
  return (values) => {
    'worklet';
    const isEvenLocal = isEven.value;
    isEven.value = 1 - isEven.value;

    if (isEvenLocal) {
      return {
        animations: {
          opacity: withTiming(1, { duration: 1000 }),
        },
        initialValues: {
          opacity: 0,
        },
      };
    }
    return {
      animations: {
        originX: values.targetOriginX,
      },
      initialValues: {
        originX: values.currentOriginX,
      },
    };
  };
}
```
The issue is that if a layout animation is started before the previous animation finished, then the previous animation is cancelled. If those animations operate on different sets of props, then we end up with broken results. In this example we have the opacity of the box stay lower than 1, if the second animation is requested too quickly.

https://github.com/software-mansion/react-native-reanimated/assets/56109050/ef18cea5-4a8b-44cd-b8c4-2fc5c6eb677e

In this PR I changed the implementation of `animationsManager` to merge the layout animations. This is the same solution that was previously implemented to avoid conflicts between entering and layout animations.

### Web

There are some issues with `visibility:hidden` on web, but they should be fixed after #5278 gets merged


jslok:
Testing your PR and it looks like it fixes the issue #5375 I was seeing so thank you! It does introduce a new problem I have found. On goBack, the animated shared element disappears from a flatlist.

https://github.com/software-mansion/react-native-reanimated/assets/22921326/afa3287e-0a3f-4b76-9c23-70f37c9f32f5

bartlomiejbloniarz:
@jslok This PR doesn't really involve SETs, so I don't think my changes really fixed the other problem, they only changed the way it manifests.

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -323,6 +323,31 @@ jsi::Value NativeReanimatedModule::configureLayoutAnimation(
...
 
+jsi::Value NativeReanimatedModule::configureLayoutAnimationBatch(
+    jsi::Runtime &rt,
+    const jsi::Value &layoutAnimationsBatch) {
+  auto array = layoutAnimationsBatch.asObject(rt).asArray(rt);
+  size_t length = array.size(rt);
+  std::vector<LayoutAnimationConfig> batch(length);
+  for (int i = 0; i < length; i++) {
+    auto item = array.getValueAtIndex(rt, i).asObject(rt);
+    batch[i].tag = item.getProperty(rt, "viewTag").asNumber();
```
piaskowyk, line 334:
```suggestion
    auto &batchItem = batch[i];
    batchItem.tag = item.getProperty(rt, "viewTag").asNumber();
```

src/reanimated2/UpdateLayoutAnimations.ts
```diff
@@ -0,0 +1,48 @@
...
+    config?: Keyframe | LayoutAnimationFunction;
+  }[] = [];
+
+  return {
+    update(config: {
+      viewTag: number;
+      type: LayoutAnimationType;
+      config?: Keyframe | LayoutAnimationFunction;
+    }) {
+      animations.push(config);
```
piaskowyk, line 22: Lets do what we talked last time, something like that but with better naming
```suggestion
      config.config = makeShareableCloneRecursive(config.config);
      animations.push(config);
```
and remove `map()` from `core.ts`

bartlomiejbloniarz: I changed it a bit differently, I call `makeShareableCloneRecursive` in the `updateLayoutAnimations` function. From there I can always simply pass the batch as it is.

piaskowyk (approved):
🫡

Issue #4850, opened 2023-07-30
missing mock of LayoutAnimations


YOEL311:
### Description

I am having an issue with using layout animation in my components. It is causing some problems.

````
        <Animated.View entering={SlideInDown.duration(500)}>
````
the test failed I got this error
````
    TypeError: Cannot read properties of undefined (reading 'duration')
````

This issue is because the mock file lacks a mock of LayoutAnimations.

### Steps to reproduce

use with layout animation  (like SlideInDown.duration(500) etc...)
``$ yarn test ``

### Snack or a link to a repository

https://snack.expo.dev/@yoel301/reanimated-2

### Reanimated version

2.17.0

### React Native version

0.70.9

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Real device

### Acknowledgements

Yes

Latropos:
Hi @YOEL311!
I am unable to reproduce your issue, you have written you can't run your tests and you haven't provided code of any tests. I was trying to put the example you've given inside our component tests and everything have been working fine. Please provide a valid reproduction, and tell us if you are using any testing library.

YOEL311:
I'm looking for a good way to individually recover my failing test code
but for now see here same issues with same fix

https://github.com/software-mansion/react-native-reanimated/issues/3284

YOEL311:
@Latropos 
I created a repo to recover this behaver

https://github.com/YOEL311/bugReanimatedRecovery

to recover this bug you need to make these steps

````
git clone https://github.com/YOEL311/bugReanimatedRecovery

cd bugReanimatedRecovery
yarn 
yarn test
````
you can see if you replace the line in the file ``jest-setup.js`` with code in command the test passing

https://github.com/YOEL311/bugReanimatedRecovery/blob/8e3e6cc89d6d2aabf0c02fc92f2cbf6070e3cf3c/jest-setup.js#L5
````
jest.doMock('react-native-reanimated', () =>
  require('react-native-reanimated/mock'),
);

// const NOOP = () => {};
// const mock = require('react-native-reanimated/mock');
// jest.doMock('react-native-reanimated', () => {
//   return {
//     ...mock,
//     SlideInDown: {
//       duration: NOOP,
//     },
//   };
// });
````
also i see that in main branch (with reanimated 3 version) you have mock on layoutAnimation
see here

[...]

PR #5372, opened 2023-11-13
Improve typing in mock.ts

tomekzaw:
## Summary

This PR is a small cleanup of `mock.ts` in terms of typing.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/4850

## Test plan


tjzel:
@tomekzaw I added all of our public API to the mocks, however I'm not sure how do we want to handle all of those functions so they are temporarily marked with a question mark with regard to their implementation.

src/reanimated2/mock.ts
```diff
@@ -100,197 +241,175 @@ class BaseAnimationMock {
...
-    EXTEND: 'extend',
-    CLAMP: 'clamp',
-    IDENTITY: 'identity',
-  },
+const layoutReanimation = {
+  BaseAnimationBuilder: new BaseAnimationMock(),
+  ComplexAnimationBuilder: new BaseAnimationMock(),
+  Keyframe: new BaseAnimationMock(),
+  // Flip
+  FlipInXUp: new BaseAnimationMock(),
```
Latropos, line 267: WDYT about sth like this:
```tsx
Object.fromEntries(['FlipInXUp', ...].map(animationName => 
    [animationName, new BaseAnimationMock()])
```

.eslintrc.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
   },
   rules: {
     curly: 'error',
+    '@typescript-eslint/no-explicit-any': 'warn',
```
Latropos, line 36: This is already on main, please merge

src/mock.ts
```diff
@@ -0,0 +1,442 @@
...
+  isReanimated3: () => true,
+  // isConfigured: ADD ME IF NEEDED
+  enableLayoutAnimations: NOOP,
+  // getViewProp: ADD ME IF NEEDED
+};
+
+const layoutReanimation = {
+  BaseAnimationBuilder: new BaseAnimationMock(),
+  ComplexAnimationBuilder: new BaseAnimationMock(),
+  Keyframe: new BaseAnimationMock(),
```
Latropos, line 275: Why is this preferred over our previous solution
```tsx
].forEach((k) =>
  Object.assign(ReanimatedV2, {
    [k]: new BaseAnimationMock(),
  })
);
```
?
Anyway, you can specify type, to make sure nobody makes a typo when assigning  baseAnimations:

```suggestion
const layoutReanimation: Record<string, BaseAnimationMock> = {
  BaseAnimationBuilder: new BaseAnimationMock(),
  ComplexAnimationBuilder: new BaseAnimationMock(),
  Keyframe: new BaseAnimationMock(),
```

mock.js
```diff
@@ -1,56 +1,10 @@
...
-
-  interpolate: NOOP,
-  interpolateColor: NOOP,
-  clamp: NOOP,
-  createAnimatedComponent: (Component) => Component,
-  addWhitelistedUIProps: NOOP,
-  addWhitelistedNativeProps: NOOP,
-};
+const Reanimated = require('./src/mock');
+const Animated = Reanimated.default;
```
tomekzaw, line 2: Not sure if we need to unpack this stuff, how about just re-exporting `./src/mock`?

tomekzaw (commented):
Let's merge this PR in the current shape and iterate on it.

tjzel (approved):

PR #5373, opened 2023-11-13
Updated peerDependencies for accepted ECMAScript standard plugins

jahead:
## Summary
A number of the babel plugins proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. They need to be moved to their transform varients.

* `@babel/plugin-proposal-class-properties"` -> `@babel/plugin-transform-class-properties`
* `@babel/plugin-proposal-nullish-coalescing-operator` -> `@babel/plugin-transform-nullish-coalescing-operator`
* `@babel/plugin-proposal-optional-chaining` -> `@babel/plugin-transform-optional-chaining`
* `@babel/plugin-proposal-private-methods` -> `@babel/plugin-transform-private-methods`

## Test plan

Ran the tests on the repo. The biggest impact on the devX is that downstream will need to move the to the new packages


jahead:
I understand the reasoning, when and if RN team make the changes, we will have to update these peers right?

tjzel:
Yes, but we will also have to take into account backward compatibility then.

PR #5383, opened 2023-11-16
[ios] Only add HERMES_ENABLE_DEBUGGER flag to debug builds

gabrieldonadel:
## Summary

Building on release mode using react-native 0.73.0-rc.4 results in a build error due to undefined symbols

> Undefined symbols for architecture x86_64:
  "facebook::hermes::inspector_modern::chrome::enableDebugging(std::__1::unique_ptr<facebook::hermes::inspector_modern::RuntimeAdapter, std::__1::default_delete<facebook::hermes::inspector_modern::RuntimeAdapter>>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&)", referenced from:
      reanimated::ReanimatedHermesRuntime::ReanimatedHermesRuntime(std::__1::unique_ptr<facebook::hermes::HermesRuntime, std::__1::default_delete<facebook::hermes::HermesRuntime>>, std::__1::shared_ptr<facebook::react::MessageQueueThread> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) in libRNReanimated.a(ReanimatedHermesRuntime.o)
  "facebook::hermes::inspector_modern::chrome::disableDebugging(int)", referenced from:
      reanimated::ReanimatedHermesRuntime::~ReanimatedHermesRuntime() in libRNReanimated.a(ReanimatedHermesRuntime.o)

This happens because react native only sets `HERMES_ENABLE_DEBUGGER` in debug mode https://github.com/facebook/react-native/blob/23cf10428eb58cfcbf614b8b0728f2535f2d252f/packages/react-native/scripts/cocoapods/utils.rb#L44-L47

So in order to fix this I believe that instead of always adding the `HERMES_ENABLE_DEBUGGER` flag we should only it on debug mode as well, using `GCC_PREPROCESSOR_DEFINITIONS[config=Debug]`

## Test plan

Build using react-native 0.73.0-rc.4 on release mode


Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.h
```diff
@@ -39,7 +39,9 @@ namespace reanimated {
 using namespace facebook;
 using namespace react;
 #if REACT_NATIVE_MINOR_VERSION >= 73
+#if HERMES_ENABLE_DEBUGGER
 using namespace facebook::hermes::inspector_modern;
+#endif // HERMES_ENABLE_DEBUGGER
 #else
 using namespace facebook::hermes::inspector;
 #endif
```
tomekzaw, line 47:
```suggestion
#if HERMES_ENABLE_DEBUGGER
#if REACT_NATIVE_MINOR_VERSION >= 73
using namespace facebook::hermes::inspector_modern;
#else
using namespace facebook::hermes::inspector;
#endif
#endif // HERMES_ENABLE_DEBUGGER
```

Common/cpp/ReanimatedRuntime/ReanimatedHermesRuntime.cpp
```diff
@@ -23,7 +23,9 @@ namespace reanimated {
 using namespace facebook;
 using namespace react;
 #if REACT_NATIVE_MINOR_VERSION >= 73
+#if HERMES_ENABLE_DEBUGGER
 using namespace facebook::hermes::inspector_modern;
+#endif // HERMES_ENABLE_DEBUGGER
 #else
 using namespace facebook::hermes::inspector;
 #endif
```
tomekzaw, line 31: I think we're safe to swap the order of these ifdefs here.
```suggestion
#if HERMES_ENABLE_DEBUGGER
#if REACT_NATIVE_MINOR_VERSION >= 73
using namespace facebook::hermes::inspector_modern;
#else
using namespace facebook::hermes::inspector;
#endif
#endif // HERMES_ENABLE_DEBUGGER
```

tomekzaw (commented):
Thanks for this PR! Left one minor suggestion, let me know what you think.

cc @piaskowyk we'll need to include it in 3.6.0 for RN 0.73 support

tjzel:
@gabrieldonadel Thanks for this PR, it helped us a great deal with some other problem we had.

I'm a bit worried about doing `config=Debug`. If the user has custom debug config (`MyDebug`) then this macro will not be defined. Could you elaborate if this is some standard way of specifying build flags? (hardcoding for Debug/Release)

gabrieldonadel:
> I'm a bit worried about doing `config=Debug`. If the user has custom debug config (`MyDebug`) then this macro will not be defined. Could you elaborate if this is some standard way of specifying build flags? (hardcoding for Debug/Release)

Hi @tjzel, if the user had a custom debug config that could indeed be a problem but `React-hermes` and `hermes-engine` would also not have these build flags as React native also uses this same logic to set `HERMES_ENABLE_DEBUGGER` only in debug mode

https://github.com/gabrieldonadel/react-native/blob/3c259240f0ea68a9105be817d71b06f3e046d6d3/packages/react-native/scripts/cocoapods/utils.rb#L44-L47

tjzel:
I see. In other ruby scripts RN tends to regex 'Debug' but here where you pointed they just go straight for the literal.

tjzel:
@gabrieldonadel I took a look into backwards compatibility and on 0.72 and lower we still have to rely on `PRODUCTION=1`. If the user on 0.72 chooses a Release build in XCode but doesn't do `PRODUCTION=1` we wouldn't define `HERMES_ENABLE_DEBUGGER` when it should be defined.

For consistency - could you add a conditional here that for 0.72 and below we set this flag only when `PRODUCTION=1` is not set?

gabrieldonadel:
> For consistency - could you add a conditional here that for 0.72 and below we set this flag only when `PRODUCTION=1` is not set?

Sure @tjzel, updated

tjzel (approved):

piaskowyk (approved):
👏

PR #5380, opened 2023-11-16
add missing easing method to BaseAnimationMock

ugurcanatas:
## Summary

This pr is a fix for #5379 which is related to missing easing method in BaseAnimationMock class.

PR #5384, opened 2023-11-16
add some of the missing mock methods to BaseAnimationMock

ugurcanatas:
## Summary

Since with [this PR](https://github.com/software-mansion/react-native-reanimated/pull/5380), a missing easing mock was added, I figured I can extend this a little bit and add some more mock functions to `BaseAnimationMock`. 

PS: I can also add the rest if that's fine with the maintainers(Including everything that is accessible from `ComplexAnimationBuilder` and `BaseAnimationBuilder` via dot property accessor).


piaskowyk:
Thanks again! 😍 

>PS: I can also add the rest if that's fine with the maintainers(Including everything that is accessible from ComplexAnimationBuilder and BaseAnimationBuilder via dot property accessor).

It would be great if you could prepare a pull request with these changes!

src/reanimated2/mock.ts
```diff
@@ -12,6 +12,10 @@ const ID = (t) => t;
 const IMMEDIATE_CB_INVOCATION = (cb: () => unknown) => cb();
 
 class BaseAnimationMock {
+  createInstance() {
+    return this;
+  }
```
ugurcanatas, line 17: since this is also exposed I added it. 
However `createInstance` might only makes sense internally inside the library(from my perspective, correct me if I am wrong). I can also delete it if it does not makes sense.

src/reanimated2/mock.ts
```diff
@@ -48,6 +52,50 @@ class BaseAnimationMock {
...
+    return 300;
+  }
+
+  getReduceMotion() {
+    return ReduceMotion.System;
+  }
+
+  getAnimationAndConfig() {
+    return [NOOP, {}];
+  }
```
ugurcanatas, line 97: here and in `getDelayFunction`  I used `NOOP`.  Because:
- `getDelayFunction` returns `AnimationFunction`
- `getAnimationAndConfig` has `AnimationFunction` as a first property inside of the array.
`AnimationFunction` type has a comment: `// this is just a temporary mock`

If this also does not work, I can change it based on the suggestion.

ugurcanatas:
hey @piaskowyk I added the methods to the mock class. I have couple questions which i've written ⬆️ 
Didn't add the properties because test didn't break in my case when I try to access properties of the classes 
(such as `FadeInDown.duration(600).delayV`). I feel like these are sufficient enough.

piaskowyk:
Thanks! 😍
As you said, `createInstance` and `getAnimationAndConfig` are internal methods, but those values are available in the object so we can mock them.

piaskowyk (approved):

ugurcanatas:
🎉🎉

PR #5399, opened 2023-11-23
Update `sideEffects` in package.json to `true`

raon0211:
## Summary

I've proposed a change in the package.json of [react-native-reanimated](https://github.com/software-mansion/react-native-reanimated/blob/main/package.json#L188) regarding the sideEffects property. Currently, it specifies two directories:

```
"sideEffects: [
    "./lib/reanimated2/core",
    "./lib/index"
  ]
```

However, this configuration seems to be inaccurate for the following reasons:

- **React Native Usage**: In React Native environments, the TypeScript files from the src folder are primarily used for bundling.
- **Presence of Side Effects in src**: Numerous files in the src folder, such as src/reanimated2/layoutReanimation/animationsManager.ts, contain side effects.

The current setup leads to issues when using bundlers like esbuild, where files with side effects may be inadvertently omitted.

To address this, I've set the sideEffects field to true in package.json. This change aims to ensure that bundling processes correctly include necessary files. 

However, if there's a more precise way to specify the side effects, I'm open to suggestions and feedback in the pull request comments!

## Test plan

- Make a metro build with the updated `package.json` and check if all tests are passing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -185,8 +185,5 @@
       "typescript"
     ]
   },
-  "sideEffects": [
-    "./lib/reanimated2/core",
-    "./lib/index"
-  ]
+  "sideEffects": true
 }


bartlomiejbloniarz:
Hi @raon0211 👋 Thanks for this PR. I would like to make some changes, is it fine by you?

bartlomiejbloniarz:
Hi @raon0211. I did some experimenting and unfortunately I don't see how this should work. With the sideEffects setting set to `true` our `next.js` example stops working, because the `Animated.View` component somehow is `undefined`. I don't really understand why this is happening, since marking side effects as `true` should lead to more code being included. However, when I mark side effects as false everything works fine (which again is surprising to me, because I would expect more code to be removed here).

Our web implementation doesn't really depend on side effects and for native there is currently no tree shaking in metro, so for now I think we should leave it as it is. I would be grateful if you have any more insight into this topic.

piaskowyk:
I truly appreciate your dedication in preparing this pull request, and I want to thank you for your time! 🙌 Unfortunately, we can't see a clear reason why this could fix the mentioned issue. I need to close this PR for now, but if you wish to continue working on it in the future, feel free to reach out to us.

PR #5400, opened 2023-11-23
reJest - mikro-framework for Reanimated public API runtime tests

Latropos:
## Summary

This PR comes with mikro-framework for Reanimated public API runtime tests
Based on: https://github.com/software-mansion/react-native-reanimated/pull/5406

### 1. Create testing container in your application

```js
import React from 'react';
import RuntimeTestsRunner from './ReanimatedRuntimeTestsRunner/RuntimeTestsRunner';

import './Animations.test'; // import tests

export default function RuntimeTestsExample() {
  return <RuntimeTestsRunner />;
}
```

### 2. Implement tests `Animations.test.ts`

```js
import Animated, { useSharedValue, useAnimatedStyle, withTiming } from 'react-native-reanimated';
import { describe, test, expect, render, useTestRef, getTestComponent, wait } from './ReanimatedRuntimeTestsRunner/RuntimeTestsApi';

const AnimatedComponent = () => {
  const widthSV = useSharedValue(0);
  const ref = useTestRef('AnimatedComponent');

  const style = useAnimatedStyle(() => ( {
    width: withTiming(widthSV.value, { duration: 500 }),
  }));

  useEffect(() => {
    widthSV.value = 100;
  }, [widthSV]);

  return (
    <View
      style={{ flex: 1, flexDirection: 'column' }}>
      <Animated.View
        ref={ref}
        style={[{ height: 80,  backgroundColor: 'blue' }, style]}
      />
    </View>
  );
};

describe('Tests of animations', () => {
  test('withTiming - expect pass', async () => {
    await render(<AnimatedComponent />);
    const component = getTestComponent('AnimatedComponent');
    await wait(600);
    expect(await component.getAnimatedStyle('width')).toBe('100');
  });
});
```

## API

- `describe`
- `test`
- `expect`
- `render`
- `useTestRef`
- `getTestComponent`
- `component.getAnimatedStyle`
- `mockAnimationTimer`
- `recordAnimationUpdates`
- `updatesContainer.getUpdates`
- `updatesContainer.getNativeSnapshots`
- `callTracker`
- `callTrackerFn`
- `getTrackerCallCount`
- `registerValue`
- `getRegisteredValue`
- `Presets`
- `notify`
- `waitForNotify`
- `wait`


app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/logMessageUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+    yellow: '\x1b[33m',
+    cyan: '\x1b[36m',
+  };
+
+  return `${COLOR_CODES[color]}${text}\x1b[0m`;
+}
+
+export function defaultTestErrorLog(
+  expected: any,
+  received: any,
```
tjzel, line 29:
```suggestion
  expected: unknown,
  received: unknown,
```

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/logMessageUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+    "Seems that you've forgot to call `mockAnimationTimer()`",
+};
+
+export function logInFrame(text: string) {
+  console.log(`\t╔${'═'.repeat(text.length + 2)}╗`);
+  console.log(`\t║ ${text} ║`);
+  console.log(`\t╚${'═'.repeat(text.length + 2)}╝`);
+}
+
+export function color(text: any, color: 'yellow' | 'cyan' | 'green' | 'red') {
```
tjzel, line 16: ?
```suggestion
export function color(text: string, color: 'yellow' | 'cyan' | 'green' | 'red') {
```

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsRunner.tsx
```diff
@@ -0,0 +1,35 @@
+import { View, Button, StyleSheet, Text } from 'react-native';
+import React, { useEffect, useState } from 'react';
+import { runTests, configure } from './RuntimeTestsApi';
+import { LockObject } from './types';
+
+let renderLock: LockObject = { lock: false };
+
+export default function RuntimeTestsRunner() {
+  const [component, renderComponent] = useState(false);
```
tjzel, line 9:
```suggestion
  const [component, setComponent] = useState(null);
```

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsRunner.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <View style={styles.container}>
+      <Button
+        title="Run tests"
+        onPress={async () => {
+          renderLock = configure({ render: renderComponent });
+          await runTests();
+        }}
+      />
+      {component}
+      {!component && <Text> Press "Run tests" button to start tests </Text>}
```
tjzel, line 25:
```suggestion
      {component || <Text> Press "Run tests" button to start tests </Text>}
```

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestComponent.ts
```diff
@@ -0,0 +1,29 @@
...
+  | 'zIndex'
+  | 'opacity'
+  | 'width'
+  | 'height'
+  | 'top'
+  | 'left'
+  | 'backgroundColor';
+
+export class TestComponent {
+  constructor(private ref: React.MutableRefObject<any>) {}
```
tjzel, line 15: :c

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsApi.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export function afterAll(job: () => void) {
+  testRunner.afterAll(job);
+}
+
+export function test(name: string, testCase: () => void) {
+  testRunner.test(name, testCase);
+}
+
+export async function render(component: any) {
```
tjzel, line 31: :c

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsApi.ts
```diff
@@ -0,0 +1,121 @@
...
+    'worklet';
+    testRunnerCallTrackerFn(name);
+  };
+}
+
+export function getTrackerCallCount(name: string) {
+  return testRunner.getTrackerCallCount(name);
+}
+
+export function registerValue(name: string, value: any) {
```
tjzel, line 57: :c

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Presets.ts
```diff
@@ -0,0 +1,221 @@
...
+  '✿༺ 𝐻𝑒𝓁𝓁𝑜 𝓌𝑜𝓇𝒹 ༻✿',
+  '꧁༺ 𝓗𝓮𝓵𝓵𝓸 𝔀𝓸𝓻𝓭 ༻꧂',
+  '🅷🅴🅻🅻🅾 🆆🅾🆁🅳',
+  '🅗🅔🅛🅛🅞 🅦🅞🅡🅓',
+];
+const NOT_NUMBERS = [Infinity, -Infinity, NaN];
+
+const EMPTIES = [[], null, undefined, {}, [[]], [{}], [null]];
+
+const FANCY_OBJECTS = [globalThis];
```
tjzel, line 96: scary o.o

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+  | Array<unknown>
+  | number
+  | bigint
+  | Record<string, unknown>;
+
+export class Matchers {
+  constructor(private currentValue: TestValue, private testCase: TestCase) {}
+
+  private assertValueIsCallTracker(
+    value: any
```
tjzel, line 17:
```suggestion
    value: unknown
```

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (!('name' in value && 'JS' in value && 'UI' in value)) {
+      throw Error('Invalid value');
+    }
+  }
+
+  public toBe(
+    expectedValue: TestValue,
+    comparisonMode = ComparisonMode.DISTANCE
+  ) {
+    const CONDITIONS: { [Key: string]: (e: any, v: any) => Boolean } = {
```
tjzel, line 28:
```suggestion
    const CONDITIONS: { [Key: string]: (e: unknown, v: unknown) => Boolean } = {
```

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+      [ComparisonMode.STRING]: (expected, value) => {
+        return value !== expected;
+      },
+      [ComparisonMode.NUMBER]: (e, v) => {
+        return isNaN(Number(v)) ? !isNaN(e) : Number(v) !== Number(e);
```
tjzel, line 33: Don't you want to check here for +0 and -0?

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+            `Invalid color format "${e}", please use lowercase hex color (like #123abc) `
+          );
+        }
+        return typeof v !== 'string' || v !== e;
+      },
+      [ComparisonMode.DISTANCE]: () => {
+        const valueAsNumber = Number(this.currentValue);
+        if (Platform.OS === 'ios') {
+          return (
+            isNaN(valueAsNumber) || valueAsNumber !== Number(expectedValue)
```
tjzel, line 48: Can we really not include a margin of error here?

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+      const name = color(this.currentValue.name, 'green');
+      const expected = color(times, 'green');
+      const received = color(callsCount, 'red');
+      this.testCase.errors.push(
+        `Expected ${name} to be called ${expected} times, but was called ${received} times`
+      );
+    }
+  }
+
+  public toBeCalledUI(times: number) {
```
tjzel, line 83: Why no default here?

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+      const threadName = color('UI thread', 'cyan');
+      const expected = color(times, 'green');
+      const received = color(this.currentValue.UI, 'red');
+      this.testCase.errors.push(
+        `Expected ${name} to be called ${expected} times on ${threadName}, but was called ${received} times`
+      );
+    }
+  }
+
+  public toBeCalledJS(times: number) {
```
tjzel, line 96: Why no default here?

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+      this.testCase.errors.push(
+        `Expected ${name} to be called ${expected} times on ${threadName}, but was called ${received} times`
+      );
+    }
+  }
+
+  public toMatchSnapshot(expectedValue: TestValue) {
+    if (Array.isArray(expectedValue) && Array.isArray(expectedValue)) {
+      let errorString = '';
+      (this.currentValue as any).forEach((val: any, idx: number) => {
```
tjzel, line 112: :c

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+      const expected = color(times, 'green');
+      const received = color(this.currentValue.UI, 'red');
+      this.testCase.errors.push(
+        `Expected ${name} to be called ${expected} times on ${threadName}, but was called ${received} times`
+      );
+    }
+  }
+
+  public toMatchSnapshot(expectedValue: TestValue) {
+    if (Array.isArray(expectedValue) && Array.isArray(expectedValue)) {
```
tjzel, line 110:
```suggestion
    if (Array.isArray(expectedValue)) {
```

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  public toMatchNativeSnapshots(
+    nativeSnapshots: Array<Record<string, unknown>>
+  ) {
+    let errorString = '';
+    const jsUpdates = this.currentValue as Array<Record<string, unknown>>;
+    for (let i = 0; i < jsUpdates.length; i++) {
+      const jsUpdate = jsUpdates[i];
+      const nativeUpdate = nativeSnapshots[i + 1];
```
tjzel, line 141: Why +1 here?

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,170 @@
...
+      [ComparisonMode.DISTANCE]: () => {
+        const valueAsNumber = Number(this.currentValue);
+        if (Platform.OS === 'ios') {
+          return (
+            isNaN(valueAsNumber) || valueAsNumber !== Number(expectedValue)
+          );
+        } else {
+          return (
+            isNaN(valueAsNumber) ||
+            Math.abs(valueAsNumber - Number(expectedValue)) > 1
```
tjzel, line 53: How about moving this `1` to some constant?

tjzel (changes requested):
I managed to get through everything except for `TestRunner` and `UpdaterContainer`.

Changes in the native code seem to be fine, only some little details there that could be improved.

As for the TS code of ReanimatedTestsRunner I find there's lack of one thing that's also missing in this PR's description - some general explanation of the concept.

I think we really need an explanation what, why and how is tested with this functionality added. Reviewing the code is pretty difficult if I don't exactly know what's the general idea behind it, especially when seeing magic `-1` and `+1` when accessing snapshot arrays etc.

Common/cpp/NativeModules/NativeReanimatedModule.h
```diff
@@ -72,6 +72,7 @@ class NativeReanimatedModule : public NativeReanimatedModuleSpec {
       jsi::Runtime &rt,
       const jsi::Value &viewTag,
       const jsi::Value &propName,
+      const jsi::Value &shadowNodeWrapper,
```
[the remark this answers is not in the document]
piaskowyk: Ola did it this way, and I haven't found a reason to change the order and move it to the last argument. So I just left it as it is. Both of these last arguments are optional, so it should be fine.

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -390,17 +393,57 @@ public void synchronouslyUpdateUIProps(int viewTag, ReadableMap uiProps) {
...
+      case "left":
+        {
+          float left = PixelUtil.toDIPFromPixel(view.getLeft());
+          return Float.toString(left);
+        }
+      case "backgroundColor":
+        {
+          Drawable background = view.getBackground();
+          if (!(background instanceof ReactViewBackgroundDrawable)) {
+            return "unknown";
```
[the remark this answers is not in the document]
Latropos: We want to always return a string, so we sometimes return "unknown". But we could extract it into a variable
```java
String defaultValue = "unknown"
```

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -390,17 +393,57 @@ public void synchronouslyUpdateUIProps(int viewTag, ReadableMap uiProps) {
...
-      Float zIndex = view.getElevation();
-      result = Float.toString(zIndex);
+    View view;
+    try {
+      view = mUIManager.resolveView(viewTag);
+    } catch (Exception e) {
+      return "unknown";
+    }
+    switch (propName) {
+      case "opacity":
```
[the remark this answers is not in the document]
piaskowyk: I don't think it's necessary 🤔
- We can't share it across our Android and iOS codebase
- Using string enums would require additional code like constructors, It produces boilerplate
- We only use it in one specific part of our code

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
Latropos: We need a "margin of error" since on Android we get all sizes in pixels. So the actual amount of pixels displaying certain element is always an integer, and then we convert it into dpi.

$$ px=dp \cdot(dpi/160)$$

$$ dp=px \cdot (160/dpi)$$

Since value of $px$ gets rounded, we have an error of $160/dpi$, and the baseline density is exactly 160, so I've decided to set error of $\pm 1$.

@piaskowyk Actually we should use https://reactnative.dev/docs/pixelratio#get here to adjust this setting for devices with non-default density

Latropos (commented):
.

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
Latropos: We will move it to some function, like described above

app/src/examples/RuntimeTests/tests/withTiming.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+  getTestComponent,
+  wait,
+  callTracker,
+  callTrackerFn,
+  getTrackerCallCount,
+} from '../ReanimatedRuntimeTestsRunner/RuntimeTestsApi';
+
+const EXAMPLE_COLORS = {
+  coral: [0xff7f50, 'rgb(255,127,80)', '#ff7f50', 'coral'],
+  cornflowerblue: [0x6495ed, 'rgb(100,149,237)', '#6495ed', 'cornflowerblue'],
```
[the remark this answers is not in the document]
Latropos: I wanted to test some default colors that are less common than "red" or "pink"

Common/cpp/Tools/UIRuntimeDecorator.cpp
```diff
@@ -59,6 +60,22 @@ void UIRuntimeDecorator::decorate(
...
       uiRuntime, "_maybeFlushUIUpdatesQueue", maybeFlushUIUpdatesQueue);
+  jsi_utils::installJsiFunction(
+      uiRuntime,
+      "_obtainProp",
+#ifdef RCT_NEW_ARCH_ENABLED
+      [obtainPropFunction](
+          jsi::Runtime &rt,
+          const jsi::Value &shadowNodeWrapper,
+          const jsi::Value &propName) {
+        return obtainPropFunction(rt, shadowNodeWrapper, propName.asString(rt));
```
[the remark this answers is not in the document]
piaskowyk: Here we're simply converting from `jsi::Value` to `jsi::String`, while in the `obtainProp` function, there's a conversion from `jsi::String` to `std::string` 🫣

src/reanimated2/NativeReanimated/NativeReanimated.ts
```diff
@@ -172,9 +181,21 @@ See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooti
   getViewProp<T>(
     viewTag: number,
     propName: string,
+    component: React.Component | undefined, // required on Fabric
     callback?: (result: T) => void
   ) {
-    return this.InnerNativeModule.getViewProp(viewTag, propName, callback);
+    let shadowNodeWrapper;
+    if (isFabric()) {
+      shadowNodeWrapper = getShadowNodeWrapperFromRef(
```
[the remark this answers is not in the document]
piaskowyk: In theory, it is possible, but in C++, we don't know which component we want to obtain the reference for. On the Fabric view, tags do not work.

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -230,20 +246,78 @@ void NativeReanimatedModule::unregisterEventHandler(
...
+    const std::string &propName,
+    const jsi::Value &shadowNodeWrapper) {
+  ShadowNode::Shared shadowNode = shadowNodeFromValue(rt, shadowNodeWrapper);
+  auto newestCloneOfShadowNode =
+      uiManager_->getNewestCloneOfShadowNode(*shadowNode);
+  Props::Shared props = newestCloneOfShadowNode->getProps();
+  auto staticProps = std::static_pointer_cast<const ViewProps>(props);
+  auto layoutableShadowNode =
+      traitCast<LayoutableShadowNode const *>(newestCloneOfShadowNode.get());
+  const auto &frame = layoutableShadowNode->layoutMetrics_.frame;
```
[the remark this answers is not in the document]
piaskowyk: Which part are you referring to? In the new architecture, all styles and properties of an object are stored in shadow nodes, so we simply read those values from the component's shadow node. Unfortunately, there are no external sources of information about it other than the source code itself. 😕

app/src/examples/RuntimeTests/tests/withTiming.test.tsx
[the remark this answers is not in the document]
Latropos: I often have to change colours when working since some of the most frequently used colours (red, yellow, green etc.) are too bright and it is uncomfortable to look at them on the screen for too long. I don't mean the aesthetics, I just have sensible eyes and bright colours hurt me a little.

app/src/examples/RuntimeTests/RuntimeTestsExample.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import RuntimeTestsRunner from './ReanimatedRuntimeTestsRunner/RuntimeTestsRunner';
+
+// load tests
+import './tests/Animations.test';
+// import './tests/sharedValue.test';
+// import './tests/withTiming.test';
```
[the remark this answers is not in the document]
piaskowyk: I removed the comments and disable some tests temporarily since they require rewriting. However, including navigation here might not be the best approach. We should find a better solutions. Let's address this in another pull request.

app/src/examples/RuntimeTests/tests/withTiming.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+  color2 = EXAMPLE_COLORS.cornflowerblue[3],
+  duration = 200,
+}: {
+  color1?: string | number;
+  color2?: string | number;
+  duration?: number;
+}) => {
+  const widthSV = useSharedValue(0);
+  const colorSV = useSharedValue(color1);
+  const ref = useTestRef('AnimatedComponent');
```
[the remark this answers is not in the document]
piaskowyk: This whole file needs to be refactored

android/src/main/java/com/swmansion/reanimated/NodesManager.java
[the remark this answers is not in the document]
tjzel: 😭

android/src/main/java/com/swmansion/reanimated/NodesManager.java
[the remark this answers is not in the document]
tjzel: Yeah, but shouldn't we return a string that would be more descriptive of what exactly has gone wrong?

Common/cpp/Tools/UIRuntimeDecorator.cpp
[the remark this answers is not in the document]
tjzel: 🤯

android/src/main/java/com/swmansion/reanimated/NodesManager.java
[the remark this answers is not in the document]
piaskowyk: I'll change this value

Common/cpp/Tools/UIRuntimeDecorator.cpp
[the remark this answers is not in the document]
tjzel: Then you should fix function signature in `PlatformDepMethodsHolder.h` and `NativeReanimatedModule.cpp` since it has `jsi::Value` there instead of `jsi::String`.

Common/cpp/NativeModules/NativeReanimatedModule.cpp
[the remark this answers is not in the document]
tjzel: I was wondering why do we have to get the `newestCloneOfShadowNode`, then take `layoutableShadowNode` etc.

app/src/examples/RuntimeTests/tests/sharedValue.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+const SharedValueComponent = ({ initialValue }: { initialValue: any }) => {
+  const sharedValue = useSharedValue(initialValue);
+  registerValue('sv', sharedValue);
+  return <Text>{sharedValue.value}</Text>;
+};
+
+describe('Tests of SharedValue', () => {
+  test('SharedValue - test number preset', async () => {
+    for (const preset of Presets.numbers) {
+      await render(null);
```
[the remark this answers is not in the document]
tjzel: I was asking for `render(null)` here to be more specific.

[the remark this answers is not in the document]
piaskowyk: This test checks the value of sharedValue after the component mounts. Therefore, we need to clear the render output to ensure that a new component will be fully mounted, not just rerendered.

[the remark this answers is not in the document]
tjzel: Put this in the comment as well please 🙏

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
piaskowyk: It seems like we don't need 😅 
image

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
tjzel: Yes, but here you don't seem to take this margin of error into account, no?

src/reanimated2/NativeReanimated/NativeReanimated.ts
```diff
@@ -172,9 +181,21 @@ See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooti
   getViewProp<T>(
     viewTag: number,
     propName: string,
+    component: React.Component | undefined, // required on Fabric
     callback?: (result: T) => void
   ) {
-    return this.InnerNativeModule.getViewProp(viewTag, propName, callback);
+    let shadowNodeWrapper;
+    if (isFabric()) {
```
tomekzaw, line 188: Not a huge fan of calling `isFabric` here, let's have two separate functions, one for Paper, one for Fabric.

tomekzaw (changes requested):
I apologize that you had to wait for so long for my review. I've left a couple of comments and suggestions in files that affect the final package. Even though these APIs are mostly meant to be private, let's improve their shape so they are analogous to existing public APIs for better maintainability.

jest.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   modulePathIgnorePatterns: [
     'Example',
     'FabricExample',
+    'app',
```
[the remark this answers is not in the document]
Latropos: This is necessary for CI to pass, without this line `yarn test` tries to execute runtime tests as well.

image

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -230,23 +247,79 @@ void NativeReanimatedModule::unregisterEventHandler(
...
-#ifdef RCT_NEW_ARCH_ENABLED
-  throw std::runtime_error(
-      "[Reanimated] getViewProp is not implemented on Fabric yet");
-#else
-  const int viewTagInt = viewTag.asNumber();
   const auto propNameStr = propName.asString(rnRuntime).utf8(rnRuntime);
   const auto funPtr = std::make_shared<jsi::Function>(
       callback.getObject(rnRuntime).asFunction(rnRuntime));
+#ifdef RCT_NEW_ARCH_ENABLED
+  auto shadowNode = shadowNodeFromValue(rnRuntime, shadowNodeWrapper);
```
piaskowyk, line 303:
```suggestion
  const auto shadowNode = shadowNodeFromValue(rnRuntime, shadowNodeWrapper);
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -112,18 +115,32 @@ NativeReanimatedModule::NativeReanimatedModule(
...
     this->dispatchCommand(rt, shadowNodeValue, commandNameValue, argsValue);
   };
+
+  auto obtainProp = [this](
+                        jsi::Runtime &rt,
+                        const jsi::Value &shadowNodeWrapper,
+                        const jsi::Value &propName) {
+    jsi::Runtime &uiRuntime = uiWorkletRuntime_->getJSIRuntime();
+    const auto propNameStr = propName.asString(rt).utf8(rt);
+    auto shadowNode = shadowNodeFromValue(rt, shadowNodeWrapper);
```
piaskowyk, line 125:
```suggestion
    const auto shadowNode = shadowNodeFromValue(rt, shadowNodeWrapper);
```

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -390,17 +393,57 @@ public void synchronouslyUpdateUIProps(int viewTag, ReadableMap uiProps) {
   }
 
   public String obtainProp(int viewTag, String propName) {
```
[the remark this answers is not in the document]
Latropos: After making some research I am afraid that the answer is no, we can't have multiple return types in java function: https://stackoverflow.com/questions/28946291/how-to-return-string-and-int-type-in-a-method. Perhaps we could bypass it, but it would require a lot of extra code 😔

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -230,23 +248,81 @@ void NativeReanimatedModule::unregisterEventHandler(
...
+
+  return jsi::Value::undefined();
+
+#else
+  const int viewTagInt = viewTag.asNumber();
 
   uiScheduler_->scheduleOnUI([=]() {
     jsi::Runtime &uiRuntime = uiWorkletRuntime_->getJSIRuntime();
-    const auto propNameValue =
+    const jsi::Value propNameValue =
```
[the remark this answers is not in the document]
Latropos: It looks that this is `jsi::String` by default, not `jsi::Value`

src/reanimated2/NativeReanimated/NativeReanimated.ts
Latropos: That would be a big change, since we have a lot of usages of `getViewProp` in our files, renaming it into `getViewPropPaper` and `getViewPropFabric` would require changing all these files. However I've prepared `getViewProp` for slitting in the future:
* Separated cpp code of `getViewProp` into separate functions
* changed declaration of `getViewProp` to accept either viewTag or shadowNodeWrapper

src/reanimated2/core.ts
```diff
@@ -55,10 +55,14 @@ if (SHOULD_BE_USE_WEB) {
   global._getAnimationTimestamp = () => performance.now();
 }
 
-export function getViewProp<T>(viewTag: number, propName: string): Promise<T> {
-  if (IS_FABRIC) {
+export function getViewProp<T>(
+  viewTag: number,
+  propName: string,
+  component?: React.Component // required on Fabric
```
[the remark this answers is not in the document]
Latropos: The code in `src/reanimated2/NativeReanimated/NativeReanimated.ts` is as follows:

```jsx
    if (isFabric()) {
      shadowNodeWrapper = getShadowNodeWrapperFromRef(
        component as React.Component
      );
      return this.InnerNativeModule.getViewProp(
        shadowNodeWrapper,
        propName,
        callback
      );
    }
```
We use component to get `shadowNodeWrapper`

apple/REANodesManager.mm
```diff
@@ -534,6 +536,27 @@ - (NSString *)obtainProp:(nonnull NSNumber *)viewTag propName:(nonnull NSString
...
+#if !TARGET_OS_OSX
+    UIColor *color = view.backgroundColor;
+    const size_t totalComponents = CGColorGetNumberOfComponents(color.CGColor);
+    const CGFloat *components = CGColorGetComponents(color.CGColor);
+    int r = 255 * components[MIN(0, totalComponents - 2)];
+    int g = 255 * components[MIN(1, totalComponents - 2)];
+    int b = 255 * components[MIN(2, totalComponents - 2)];
+    result = [NSString stringWithFormat:@"#%02x%02x%02x", r, g, b];
+#else
+    result = @"Cant get background color on macos";
```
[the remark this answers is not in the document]
Latropos: > Maybe we should come up with macOS implementation anyway?

I think it's a good idea, but I'd rather do it in separate PR. This one needs to be tested on both Paper and Fabric and both IOS and Android, which means that it is difficult to test. Its also very big.

> Maybe we should throw an exception instead?

We could, but it would crash the app, instead of reporting one of the tests as a failing, so I think I would like to keep returning a string here. I've tested it and, although it's not a pattern, it is very easy to debug. And it may be just a temporary solution, until we implement it for macOS

> The error message is poorly written.

I'll try to write it better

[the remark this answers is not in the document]
Latropos: > Maybe we should come up with macOS implementation anyway?

Sure!

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -230,23 +248,81 @@ void NativeReanimatedModule::unregisterEventHandler(
...
+    jsi::Runtime &uiRuntime = uiWorkletRuntime_->getJSIRuntime();
+    std::string resultStr =
+        obtainPropFromShadowNode(uiRuntime, propNameStr, shadowNode);
+
+    jsScheduler_->scheduleOnJS([=](jsi::Runtime &rnRuntime) {
+      const auto resultValue =
+          jsi::String::createFromUtf8(rnRuntime, resultStr);
+      funPtr->call(rnRuntime, resultValue);
+    });
+  });
```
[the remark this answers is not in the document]
Latropos: I got the following error when calling `obtainPropFunction_` on js:
>NSException: This function must be called on the main queue;

Common/cpp/NativeModules/NativeReanimatedModule.cpp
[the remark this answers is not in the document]
Latropos: This error appears otherwise:
image

Common/cpp/NativeModules/NativeReanimatedModule.h
```diff
@@ -130,6 +131,11 @@ class NativeReanimatedModule : public NativeReanimatedModuleSpec {
   jsi::Value measure(jsi::Runtime &rt, const jsi::Value &shadowNodeValue);
 
   void initializeFabric(const std::shared_ptr<UIManager> &uiManager);
+
+  std::string obtainPropFromShadowNode(
```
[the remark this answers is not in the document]
Latropos: Changing `std::string` into `jsi::Value` everywhere is a big change, since jsi::Value is a temporary object and cannot be always accessed from other threads and it would require a lot of testing. Can I do this in separate PR?

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -230,23 +242,92 @@ void NativeReanimatedModule::unregisterEventHandler(
...
+  } else if (propName == "zIndex") {
+    std::optional<int> zIndex = viewProps->zIndex;
+    if (zIndex) {
+      return std::to_string(*zIndex);
+    }
+  } else if (propName == "backgroundColor") {
+    return int_to_hex(*viewProps->backgroundColor);
+  }
+
+  return "Getting property '" + propName +
```
tomekzaw, line 284: Not a huge fan of returning strings with error message here

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -230,23 +242,92 @@ void NativeReanimatedModule::unregisterEventHandler(
...
+      "' with function 'getViewProp' is not supported";
+}
+
+#endif
+
+#ifdef RCT_NEW_ARCH_ENABLED
```
tomekzaw, line 290: This `#endif` ends `#ifdef RCT_NEW_ARCH_ENABLED` so I think we can just remove these lines?
```suggestion
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -230,23 +242,92 @@ void NativeReanimatedModule::unregisterEventHandler(
...
-#ifdef RCT_NEW_ARCH_ENABLED
-  throw std::runtime_error(
-      "[Reanimated] getViewProp is not implemented on Fabric yet");
+  const auto propNameStr = propName.asString(rnRuntime).utf8(rnRuntime);
+  const auto funPtr = std::make_shared<jsi::Function>(
+      callback.getObject(rnRuntime).asFunction(rnRuntime));
+  const auto shadowNode = shadowNodeFromValue(rnRuntime, shadowNodeWrapper);
+  uiScheduler_->scheduleOnUI([=]() {
+    jsi::Runtime &uiRuntime = uiWorkletRuntime_->getJSIRuntime();
+    std::string resultStr =
```
tomekzaw, line 303:
```suggestion
    const auto resultStr =
```

Common/cpp/Tools/UIRuntimeDecorator.cpp
```diff
@@ -59,6 +60,11 @@ void UIRuntimeDecorator::decorate(
   jsi_utils::installJsiFunction(uiRuntime, "_setGestureState", setGestureState);
   jsi_utils::installJsiFunction(
       uiRuntime, "_maybeFlushUIUpdatesQueue", maybeFlushUIUpdatesQueue);
+
+  jsi_utils::installJsiFunction(
+      uiRuntime, "_obtainPropFabric", obtainPropFunction);
+  jsi_utils::installJsiFunction(
+      uiRuntime, "_obtainPropPaper", obtainPropFunction);
```
tomekzaw, line 67: Can we move these lines into `#ifdef RCT_NEW_ARCH_ENABLED` above, next to `_updatePropsPaper` and `_measurePaper`?

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -390,17 +393,43 @@ public void synchronouslyUpdateUIProps(int viewTag, ReadableMap uiProps) {
...
+    }
+
+    switch (propName) {
+      case "opacity":
+        return Float.toString(view.getAlpha());
+      case "zIndex":
+        return Float.toString(view.getElevation());
+      case "width":
+        float width = PixelUtil.toDIPFromPixel(view.getWidth());
+        return Float.toString(width);
```
tomekzaw, line 410: Honestly I don't think we need to introduce intermediate variables here, what do you think?
```suggestion
        return Float.toString(PixelUtil.toDIPFromPixel(view.getWidth()));
```

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -390,17 +393,43 @@ public void synchronouslyUpdateUIProps(int viewTag, ReadableMap uiProps) {
...
+      case "backgroundColor":
+        Drawable background = view.getBackground();
+        if (!(background instanceof ReactViewBackgroundDrawable)) {
+          return "unable to resolve background color";
+        }
+        int actualColor = ((ReactViewBackgroundDrawable) background).getColor();
+        return String.format("#%06x", (0xFFFFFF & actualColor));
+      default:
+        throw new IllegalArgumentException(
+            "[Reanimated] Attempted to get  unsupported property"
```
tomekzaw, line 429:
```suggestion
            "[Reanimated] Attempted to get unsupported property"
```

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -390,17 +393,43 @@ public void synchronouslyUpdateUIProps(int viewTag, ReadableMap uiProps) {
...
+                + propName
+                + " with function 'getViewProp'");
```
tomekzaw, line 431:
```suggestion
                + " with function `getViewProp`");
```

src/reanimated2/globals.d.ts
```diff
@@ -102,4 +102,9 @@ declare global {
   var UpdatePropsManager: UpdatePropsManager;
   var ProgressTransitionRegister: ProgressTransitionRegister;
   var updateJSProps: (viewTag: number, props: Record<string, unknown>) => void;
+  var _obtainPropPaper: (viewId: number, propName: string) => string;
```
tomekzaw, line 105:
```suggestion
  var _obtainPropPaper: (viewTag: number, propName: string) => string;
```

apple/REANodesManager.mm
```diff
@@ -517,26 +517,50 @@ - (void)updateProps:(nonnull NSDictionary *)props
...
+#endif
+    if (color != nil) {
+      const size_t totalComponents = CGColorGetNumberOfComponents(color.CGColor);
+      const CGFloat *components = CGColorGetComponents(color.CGColor);
+      int r = 255 * components[MIN(0, totalComponents - 2)];
+      int g = 255 * components[MIN(1, totalComponents - 2)];
+      int b = 255 * components[MIN(2, totalComponents - 2)];
+      return [NSString stringWithFormat:@"#%02x%02x%02x", r, g, b];
+    } else {
+      return @"Invalid background color is 'nil'";
```
tomekzaw, line 556:
```suggestion
      return @"nil";
```

apple/REANodesManager.mm
```diff
@@ -517,26 +517,50 @@ - (void)updateProps:(nonnull NSDictionary *)props
...
+    return [@(view.frame.origin.y) stringValue];
+  } else if ([propName isEqualToString:@"left"]) {
+    return [@(view.frame.origin.x) stringValue];
+  } else if ([propName isEqualToString:@"backgroundColor"]) {
+#if !TARGET_OS_OSX
+    UIColor *color = view.backgroundColor;
+#else
+    NSColor *color = view.backgroundColor;
+#endif
+    if (color != nil) {
```
tomekzaw, line 548:
```suggestion
    if (color == nil) {
      return @"nil";
    }
    // rest of the code
```

Common/cpp/NativeModules/NativeReanimatedModule.h
```diff
@@ -127,9 +131,19 @@ class NativeReanimatedModule : public NativeReanimatedModuleSpec {
       const jsi::Value &commandNameValue,
       const jsi::Value &argsValue);
 
+  jsi::String obtainProp(
```
tomekzaw, line 134: Not really sure why we need all of those functions, why do we have so much of them?
* `getViewProp`
* `obtainProp`
* `obtainPropFromShadowNode`

What's the difference between `getViewProp` and `obtainProp`? I assume that one is used from the RN runtime and the other is used from the UI runtime. Can we somehow unify it so that it's consistent across runtimes? (We can compare `jsi::Runtime &rt` with `uiRuntime_->getJSIRuntime()` to see if we're on the UI runtime.)

What's the difference between `obtainProp` and `obtainPropFromShadowNode`? Can we merge one into another?

tomekzaw (changes requested):

Common/cpp/NativeModules/NativeReanimatedModule.h
Latropos: * `obtainPropFromShadowNode` is a helper function used both in `getViewProp` and `obtainProp`
* The difference between `obtainProp` and `getViewProp` is exactly in runtimes. The current form of these functions was introduced here: https://github.com/software-mansion/react-native-reanimated/pull/4929 We could try unify them, but that would be a vary big change (and a very dangerous), so I wouldn't recommend doing it in this PR

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -230,23 +242,92 @@ void NativeReanimatedModule::unregisterEventHandler(
...
+    jsi::Runtime &rt,
+    const std::string &propName,
+    const ShadowNode::Shared &shadowNode) {
+  auto newestCloneOfShadowNode =
+      uiManager_->getNewestCloneOfShadowNode(*shadowNode);
+  auto props = newestCloneOfShadowNode->getProps();
+  auto viewProps = std::static_pointer_cast<const ViewProps>(props);
+  auto layoutableShadowNode =
+      traitCast<LayoutableShadowNode const *>(newestCloneOfShadowNode.get());
+  const auto &frame = layoutableShadowNode->layoutMetrics_.frame;
```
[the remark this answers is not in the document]
Latropos: We don't have to, I've fixed that

Common/cpp/NativeModules/NativeReanimatedModule.cpp
Latropos: So I'll throw an error

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -242,53 +242,56 @@ void NativeReanimatedModule::unregisterEventHandler(
...
-  } else if (propName == "opacity") {
-    return std::to_string(viewProps->opacity);
-  } else if (propName == "zIndex") {
-    std::optional<int> zIndex = viewProps->zIndex;
-    if (zIndex) {
-      return std::to_string(*zIndex);
+
+  if (propName == "width" || propName == "height" || propName == "top" ||
+      propName == "left") {
+    // These pops are calculated from frame
```
Latropos, line 261:
```suggestion
    // These props are calculated from frame
```

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -242,53 +242,56 @@ void NativeReanimatedModule::unregisterEventHandler(
...
+      return std::to_string(frame.size.width);
+    } else if (propName == "height") {
+      return std::to_string(frame.size.height);
+    } else if (propName == "top") {
+      return std::to_string(frame.origin.y);
+    } else if (propName == "left") {
+      return std::to_string(frame.origin.x);
+    }
+  } else {
+    // These pops are calculated from viewProps
```
Latropos, line 276:
```suggestion
    // These props are calculated from viewProps
```

Common/cpp/NativeModules/NativeReanimatedModule.h
tjzel: I think we can make those functions tied to runtimes for now, and can make them runtime-agnostic in the future, if needed.

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Asserts.ts
tjzel, line 1: Circle back to this

Common/cpp/NativeModules/NativeReanimatedModule.h
tomekzaw: I would love to see some unification, right now there's too many functions and it's confusing.

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -0,0 +1,178 @@
...
+      typeof value !== 'object' ||
+      !('name' in value && 'onJS' in value && 'onUI' in value)
+    ) {
+      throw Error('Invalid value');
+    }
+  }
+
+  public toBe(
+    expectedValue: TestValue,
+    comparisonMode = ComparisonMode.DISTANCE
```
tjzel, line 27: Doesn't quite make sense that the default comparisonMode is not AUTO 🤔

tjzel (approved):
Let's get this bad boy merged!

Common/cpp/NativeModules/NativeReanimatedModule.h
tjzel: I agree about the unification, but I think this PR is too important too keep it waiting at function splitting/merging.

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Comparators.ts
```diff
@@ -0,0 +1,98 @@
...
+        !COMPARATORS[ComparisonMode.AUTO](
+          expected[key as keyof typeof expected],
+          value[key as keyof typeof value]
+        )
+      ) {
+        return false;
+      }
+    }
+    return true;
+  },
```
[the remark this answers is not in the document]
Latropos: I don't think so, since expected value should be hardcoded and the actual one calculated. Therefore references shouldn't match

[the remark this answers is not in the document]
tjzel: Oh, I didn't realize you had computed ones here, my bad 😅

Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -230,23 +242,95 @@ void NativeReanimatedModule::unregisterEventHandler(
...
+        return std::to_string(*viewProps->zIndex);
+      }
+    } else if (propName == "backgroundColor") {
+      return intColorToHex(*viewProps->backgroundColor);
+    }
+  }
+
+  throw std::runtime_error(std::string(
+      "Getting property '" + propName +
+      "' with function 'getViewProp' is not supported"));
```
Latropos, line 292: ` instead of '

android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -390,17 +393,39 @@ public void synchronouslyUpdateUIProps(int viewTag, ReadableMap uiProps) {
...
+      case "height":
+        return Float.toString(PixelUtil.toDIPFromPixel(view.getHeight()));
+      case "top":
+        return Float.toString(PixelUtil.toDIPFromPixel(view.getTop()));
+      case "left":
+        return Float.toString(PixelUtil.toDIPFromPixel(view.getLeft()));
+      case "backgroundColor":
+        Drawable background = view.getBackground();
+        if (!(background instanceof ReactViewBackgroundDrawable)) {
+          return "unable to resolve background color";
```
Latropos, line 419: Throw an error

tomekzaw (approved):

PR #5405, opened 2023-11-24
Fix entering flickers in web LA

m-bert:
## Summary

Sometimes entering animations flicker. This happens because we've recently merged [this PR](https://github.com/software-mansion/react-native-reanimated/pull/5278), which always sets component visibility of the component on its first render, even though now we are doing it inside `onanimationstart`.

## Test plan

Tested on example app on [LA] examples


piaskowyk (approved):
🫡

PR #5433, opened 2023-11-29
Deprecate old signature

Latropos:
## Summary
Deprecate
`const Flatlist = Animated.createAnimatedComponent(FlatList)`

### DIFF
--- a/src/createAnimatedComponent/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent/createAnimatedComponent.tsx
@@ -54,6 +54,7 @@ import {
   getReducedMotionFromConfig,
 } from '../reanimated2/layoutReanimation/web';
 import type { CustomConfig } from '../reanimated2/layoutReanimation/web/config';
+import type { FlatList, FlatListProps } from 'react-native';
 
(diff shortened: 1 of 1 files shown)


tomekzaw (commented):
This is not exactly what I imagined; I though we'll deprecate `Animated.createAnimatedComponent(FlatList)` call in TypeScript using function overloading &ndash; i.e. deprecate this specific call.

src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -77,13 +78,28 @@ type Options<P> = {
   setNativeProps: (ref: AnimatedComponentRef, props: P) => void;
 };
 
+function assertComponentIsComponent(
```
tjzel, line 81: What is this assert for?

Latropos: If we want to add this overload we have to include `typeof FlatList` in definition of `createAnimatedComponent`. Then it starts to throw error since in fact `ComponentType<InitialComponentProps>` has some props that FlatList don't.

PR #5435, opened 2023-11-30
fix(android): delete native instance of Scheduler on deactivate

BasixKOR:
## Summary

We found a leak that `Scheduler` was holding mContext after the activity has been closed, and found out `mHybridData` was holding a native JNI global reference to the Scheduler instance, preventing GC.


BasixKOR:
We have found some more some resources not deinitialized properly, so I'll make a new PR with all the patches we've made.

PR #5436, opened 2023-11-30
Fix type of `_log` function

tomekzaw:
## Summary

This PR fixes the type of argument of `_log` function.

## Test plan

```ts
_log(42);
```


tjzel (approved):
Solid pull request. I wish all of them were so thorough.

PR #5444, opened 2023-11-30
Fix `exiting` on web

m-bert:
## Summary

This PR aims to fix wrong positions of elements during `exiting` layout animation.

## Test plan

Tested on example app (and other LA examples provided by @kacperkapusciak)


src/reanimated2/layoutReanimation/web/componentUtils.ts
```diff
@@ -135,6 +135,13 @@ export function getProcessedConfig(
   };
 }
 
+export function saveSnapshot(element: HTMLElement) {
+  element.setAttribute(
+    'data-reanimated-snapshot',
+    JSON.stringify(element.getBoundingClientRect())
+  );
```
tomekzaw, line 142: We could also store these snapshots in a `WeakMap`, this way we don't need to serialize and parse it.

m-bert: Done in [bedb8f7](https://github.com/software-mansion/react-native-reanimated/pull/5444/commits/bedb8f73bed613dc4842082d807dd388640e2a6a)

piaskowyk (approved):
🫡

PR #5452, opened 2023-12-01
Remove dummy element on animation cancel

m-bert:
## Summary

In exiting animation, we create a dummy - clone of the original element that will be removed. When animation ends, we remove dummy from parent. The problem is, if something cancels animation, our dummy component won't be removed. 

I've stumbled upon this problem while working on custom keyframes.

## Test plan

Tested on example app. Mostly on Keyframe animation example (on branch with custom keyframes implementation).

### DIFF
--- a/src/reanimated2/layoutReanimation/web/componentUtils.ts
+++ b/src/reanimated2/layoutReanimation/web/componentUtils.ts
@@ -290,4 +290,10 @@ export function handleExitingAnimation(
     // Given that this function overrides onAnimationEnd, it won't be null
     originalOnAnimationEnd?.call(this, event);
   };
+
+  dummy.addEventListener('animationcancel', () => {
+    if (parent?.contains(dummy)) {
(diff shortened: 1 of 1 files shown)


src/reanimated2/layoutReanimation/web/componentUtils.ts
```diff
@@ -290,4 +290,10 @@ export function handleExitingAnimation(
     // Given that this function overrides onAnimationEnd, it won't be null
     originalOnAnimationEnd?.call(this, event);
   };
+
+  dummy.addEventListener('animationcancel', () => {
```
piaskowyk, line 294: Why the listener from https://github.com/software-mansion/react-native-reanimated/pull/5452/files#diff-31934cfe147067532824ca4f698abb5b5582f9f8a656986f77efe30a459178f6R177-R180 isn't enough?

piaskowyk (approved):

PR #5376, opened 2023-11-14
Extract keyframe key construction to a function

marmor157:
## Summary
This PR uses string interpolation to construct keyframe keys in `Keyframe.ts` as requested it [PR comment](https://github.com/software-mansion/react-native-reanimated/pull/5363#discussion_r1390286269). Additionally it extracts it to a helper function

PR #5454, opened 2023-12-01
Fix keyframes crash

m-bert:
## Summary

[Recent changes](https://github.com/software-mansion/react-native-reanimated/pull/5376) in Keyframe.ts introduced crashes on iOS and Android, because of calling non-worklet function from UI thread.

## Test plan

Tested on Olympic example.


tomekzaw (approved):

OwenMelbz:
Can I just ask which npm version this fix will be in? Thanks

tomekzaw:
@OwenMelbz This PR is already released in 3.7.0 and 3.7.1.

PR #5458, opened 2023-12-02
Replace useRef with useState for useSharedValue

amadeus:
## Summary
The general idea here is that we can avoid the `makeMutable` call on every single render.  This change is also compatible with hot reloading.

We (at Discord) have been running this patch on top of our own reanimated now for a few weeks and haven't noticed any side effects of the change.

There has been a bit of additional discussion I've seen surrounding this issue here: https://github.com/software-mansion/react-native-reanimated/pull/3199

## Test plan
N/A

### DIFF
--- a/src/reanimated2/hook/useSharedValue.ts
+++ b/src/reanimated2/hook/useSharedValue.ts
@@ -1,5 +1,5 @@
 'use strict';
-import { useEffect, useRef } from 'react';
+import { useEffect, useState } from 'react';
 import { cancelAnimation } from '../animation';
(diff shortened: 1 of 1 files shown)


tomekzaw:
@amadeus Thanks a lot for submitting this PR!

piaskowyk (approved):
I've tested it and everything works as expected 🎉

tomekzaw (approved):
Thanks again for the PR!

amadeus:
Yay Thanks!

PR #5463, opened 2023-12-04
Fix overloads in createAnimatedComponent

Latropos:
## Summary

1. Move the overload with type deprecation to be the last one. It means that if user provides invalid type which isn't covered by any overload he will not get the warning about FlatList deprecation. This warning was very confusing if user wasn't using FlatList at all
2. Add overload of `ComponentType`. Since `ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>` we have assume that it is enough to have two overloads (ComponentClass & FunctionComponent), but it isn't

|BEFORE|AFTER|
|--|--|
|Error for FastImage and deprecation for FastImage ❌ | Deprecation for FlatList only 😎|
|Screenshot 2023-12-12 at 13 16 37|image|
|image||

### DIFF
--- a/src/createAnimatedComponent/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent/createAnimatedComponent.tsx
@@ -87,15 +87,7 @@ type Options<P> = {
  * @see https://docs.swmansion.com/react-native-reanimated/docs/core/createAnimatedComponent
  */
 
-/**
- * @deprecated Please use `Animated.FlatList` component instead of calling `Animated.createAnimatedComponent(FlatList)` manually.
- */
-// @ts-ignore This is required to create this overload, since type of createAnimatedComponent is incorrect and doesn't include typeof FlatList
-export function createAnimatedComponent(
-  component: typeof FlatList<unknown>,
-  options?: Options<any>
-): ComponentClass<AnimateProps<FlatListProps<unknown>>>;
-
+// Don't change the order of overloads, since such a change breaks current behavior
 export function createAnimatedComponent<P extends object>(
   component: FunctionComponent<P>,
   options?: Options<P>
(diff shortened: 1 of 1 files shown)


m-bert (approved):

tjzel (changes requested):
Can you provide examples in the PR description that show why those changes are necessary?

src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -87,15 +87,7 @@ type Options<P> = {
...
-/**
- * @deprecated Please use `Animated.FlatList` component instead of calling `Animated.createAnimatedComponent(FlatList)` manually.
- */
-// @ts-ignore This is required to create this overload, since type of createAnimatedComponent is incorrect and doesn't include typeof FlatList
-export function createAnimatedComponent(
-  component: typeof FlatList<unknown>,
-  options?: Options<any>
-): ComponentClass<AnimateProps<FlatListProps<unknown>>>;
-
+// Don't change the order of overloads, since such a change breaks current behavior
```
piaskowyk, line 90: It appears to be a flawed approach when our only means of ensuring that we don't break anything is through a comment. Is it possible to implement some TypeScript tests instead? I believe that @tjzel has already worked on something similar.

Latropos: Unfortunately there are no tests for deprecation

src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -106,6 +98,22 @@ export function createAnimatedComponent<P extends object>(
...
+  options?: Options<P>
+): FunctionComponent<AnimateProps<P>> | ComponentClass<AnimateProps<P>>;
+
+/**
+ * @deprecated Please use `Animated.FlatList` component instead of calling `Animated.createAnimatedComponent(FlatList)` manually.
+ */
+// @ts-ignore This is required to create this overload, since type of createAnimatedComponent is incorrect and doesn't include typeof FlatList
+export function createAnimatedComponent(
+  component: typeof FlatList<unknown>,
+  options?: Options<any>
```
tjzel, line 114: Can we get rid of `any` here?

tjzel: I see, let's keep it for now then.

tjzel (approved):

PR #5464, opened 2023-12-04
Use relative file location in source maps with `relativeSourceLocation` enabled

tjzel:
## Summary

This option was added back when source maps were not included in https://github.com/software-mansion/react-native-reanimated/pull/3141.

Source map addition for better LogBox experience was added in https://github.com/software-mansion/react-native-reanimated/pull/3846 - but it didn't take into account `relativeSourceLocation`. This PR fixes this and there is no more absolute location in the source map while `relativeSourceLocation` is on.

I haven't found a way to generate the source map with relative location out-of-the-box.

## Test plan

In `WorkletExample` see that LogBox is working as intended. Also plugin tests maybe.


bartlomiejbloniarz (approved):
🚀

Issue #5475, opened 2023-12-07
Shared Element Transition does not take current scale into account


mrousavy:
### Description

Hey all!

I've noticed a small bug in Shared Element Transitions, where the Animation does not take the current scale into account. See this video recording:

https://github.com/software-mansion/react-native-reanimated/assets/15199031/da40c895-ed5b-4d65-87ad-b7d030b362af

You might want to watch this in QuickTime and go through this frame by frame - but as you can see I drag down the view to this scale:

<img src="https://github.com/software-mansion/react-native-reanimated/assets/15199031/29ad3e50-d693-465e-8a23-223c211d96d2" width="35%" />

...where it should start animating from. If we go to the next frame, we can see that the animation does not start at that scaled down state, but rather at the original scale:

<img src="https://github.com/software-mansion/react-native-reanimated/assets/15199031/92f57954-d57e-4364-9c6d-fcd0618d3955" width="35%" />

...and then just scales down from that to it's target position.

It does work for the translation (it starts at the correct originY and originY), but not for scale. Both translation and scale are passed in the same parent view's style.

### Steps to reproduce

1. Create Shared Element Transition
2. Scale down the View that contains the shared element
3. Run shared element transition at scaled down state
4. Notice that it does not start at scaled down state, but rather at scale 1.

### Snack or a link to a repository

https://github.com/mrousavy/react-native-reanimated/tree/repro/shared-element-transition-bugs

### Reanimated version

3.6.1

### React Native version

0.72.7

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

Real device

### Device model

iPhone 11 Pro

### Acknowledgements

Yes

Laityned:
As an addition, I noticed that transformations applied to any parent element are also not taking into account. Therefore, transistion is happening from its original location and not from his current location

bartlomiejbloniarz:
This issue was auto-closed as it pertains to the unsupported Old Architecture. If you think this is a mistake, please re-open.

Issue #5473, opened 2023-12-07
Shared Element Transition is offset more and more each time it runs


mrousavy:
### Description

Hey!
I've been playing around with Shared Element Transitions (they are amazing btw!), but I found a funny little bug.

For every time the Shared Element Transition runs, it gets offset by more and more pixels to the top left.

https://github.com/software-mansion/react-native-reanimated/assets/15199031/f8528f5d-1329-4127-8ec6-fad24f008e3f

This happens also if I don't pass a custom `sharedTransitionStyle`.

### Steps to reproduce

1. Create View (expo-image's `<Image>` in my case) with Shared Element Transition Tag:
    ```tsx
    <ReanimatedImage
      sharedTransitionTag={uri}
      contentFit="cover"
      cachePolicy="memory"
      allowDownscaling={true}
      style={styles.image}
      source={source}
    />
    ```
2. Run animation between components multiple times and see shift in positions

### Snack or a link to a repository

https://github.com/mrousavy/react-native-reanimated/tree/repro/shared-element-transition-bugs

### Reanimated version

3.6.1

### React Native version

0.72.7

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

Real device

### Device model

iPhone 11 Pro

### Acknowledgements

Yes

mrousavy:
Okay - just created a minimal reproduction! I just updated the "[SET] Custom Animations" example in the Reanimated playground in my branch here: https://github.com/mrousavy/react-native-reanimated/tree/repro/shared-element-transition-bugs


https://github.com/software-mansion/react-native-reanimated/assets/15199031/82a50b84-4174-43c1-9590-8266193b57c4

bartlomiejbloniarz:
This issue was auto-closed as it pertains to the unsupported Old Architecture. If you think this is a mistake, please re-open.

PR #5476, opened 2023-12-07
Repro/shared element transition bugs

mrousavy:
## Summary

This is just a reproduceable example for the SET bug reports I mentioned in https://github.com/software-mansion/react-native-reanimated/issues/5473 and https://github.com/software-mansion/react-native-reanimated/issues/5475.


arasrezaei:
i am actually experiencing same...
i home https://github.com/software-mansion/react-native-reanimated/pull/5793 fixes this bugs too.

PR #5480, opened 2023-12-08
Improve TS in src [6/6] - better return type in useAnimatedSensor

tjzel:
## Summary
Turns out there is a difference between
```TS
AnimatedSensor<ValueRotation | Value3D>
```
and
```TS
AnimatedSensor<ValueRotation> | AnimatedSensor<Value3D>
```
The latter is the correct type in this case.

## Test plan

CI

Co-authored by @marmor157.


src/reanimated2/hook/useAnimatedSensor.ts
```diff
@@ -146,5 +146,5 @@ export function useAnimatedSensor(
     };
   }, [sensorType, userConfig]);
 
-  return ref.current;
+  return ref.current as AnimatedSensor<ValueRotation> | AnimatedSensor<Value3D>;
```
tomekzaw, line 149: Can we set the appropriate return type based on `sensorType` value?

tjzel: Would need to make this function generic in regard to the `sensorType` - I don't think this would make it actually simpler.

tomekzaw: `ValueRotation` and `Value3D` have different sets of fields and our users deserve better type-checking and autocomplete in their code.

tjzel: Yes, but this overload is not seen by the user - it's only the implementation. He will get the type that is provided by two previous overloads.

tomekzaw: You're right, I missed overloads. All good!

tomekzaw (approved):

PR #5483, opened 2023-12-08
Improve TS in src [2/6] - Add default type parameter to SharedValue

tjzel:
## Summary

Just a QoL change so we can type `SharedValue` and `ShareableRef` instead of `SharedValue<unknown>` and `ShareableRef<unknown>`.

## Test plan

CI

Co-authored by: @marmor157.


src/reanimated2/isSharedValue.ts
```diff
@@ -1,7 +1,10 @@
...
 import type { SharedValue } from './commonTypes';
 
-export function isSharedValue<T>(value: any): value is SharedValue<T> {
+export function isSharedValue<T = unknown>(
+  value: unknown
+): value is SharedValue<T> {
   'worklet';
-  return value?._isReanimatedSharedValue === true;
+  // We cannot use `in` operator here because `value` could be a HostObject and therefore we cast.
+  return (value as Record<string, unknown>)?._isReanimatedSharedValue === true;
```
tomekzaw, line 9: Since we export `isSharedValue` to the public API, should we extend the current implementation so that it supports arbitrary JS values? Currently `isSharedValue(42)` will fail.

btw why not `object` instead of `Record<string, unknown>`?

tjzel: With this implementation it works.
Screenshot 2023-12-14 at 17 42 43

Regarding the use of `Record<string,unknown>` - `object` is a type that doesn't have any properties so trying to access anything results in a TS error. Proper type to check for props is unfortunately `Record`.

tomekzaw: :like:

.eslintrc.js
```diff
@@ -60,5 +60,6 @@ module.exports = {
       { fixMixedExportsWithInlineTypeSpecifier: false },
     ],
     'tsdoc/syntax': 'error',
+    '@typescript-eslint/no-non-null-assertion': 'off',
```
Latropos, line 63: I think it would be much more safe to have this rule on. We can use asserts return type instead of eslint disable.
Such an asserts function would throw error
```typescript
function assertNotNull<T>(x: T): asserts x is NonNullable<T> {
  if (x === null || x === undefined) {
    throw new Error('error');
  }
}
```

src/reanimated2/hook/utils.ts
```diff
@@ -42,8 +42,10 @@ export function areDependenciesEqual(
   prevDeps: DependencyList
 ) {
   function is(x: number, y: number) {
-    // eslint-disable-next-line no-self-compare
-    return (x === y && (x !== 0 || 1 / x === 1 / y)) || (x !== x && y !== y);
+    return (
+      (x === y && (x !== 0 || 1 / x === 1 / y)) ||
```
Latropos, line 46: What is the purpose of this code? Handling `Infinity`? Could we compare directly x with infinity to make it easier to read instead?

src/reanimated2/isSharedValue.ts
Latropos: Why don't we implement this function like this:
```jsx

export function isSharedValue<T>(value: unknown): value is SharedValue<T> {
  'worklet';
  return typeof value === 'object' &&
    value !== null &&
    '_isReanimatedSharedValue' in value
    ? value?._isReanimatedSharedValue === true
    : false;
}
```

Latropos: Longer code, but I'd prefer to avoid unnecessary type casts

Latropos (changes requested):
I don't think we should disable `'@typescript-eslint/no-non-null-assertion'`

.eslintrc.js
tjzel: @Latropos more motivations on enabling non-null-assertions you can find here https://github.com/software-mansion/react-native-reanimated/pull/5481#issue-2032988675.

While we could use those assertions, there are two problems with them.
1. For short functions function call on UI is relatively expensive at the moment.
2. Why call an assert when we are certain that a given value is not a null? Like all the Fabric specific functions on Fabric.

src/reanimated2/isSharedValue.ts
tjzel: We can't use `in` operator because of HostObjects.

src/reanimated2/hook/utils.ts
tjzel: I'm actually not sure either what do we achieve with
`(x !== 0 || 1 / x === 1 / y)`.

tjzel:
@Latropos It was already enabled in #5481.

src/reanimated2/hook/utils.ts
tjzel: Okay, I know the answer. In JavaScript the number type has values for `-0` and `0`. However, `-0 === 0` is `true`. So if we want to check if we have this case we need to compare their reciprocals since `1/0` gives `Number.POSITIVE_INFINITY` and `1/(-0)` gives `NUMBER.NEGATIVE_INFINITY` and `Number.POSITIVE_INFINITY === Number.NEGATIVE_INFINITY` is `false`.

piaskowyk (approved):

src/reanimated2/hook/utils.ts
Latropos: Why did you undo all this changes? Could you add a comment explaining this code (since it was unintuitive for me)?

Latropos (approved):

.eslintrc.js
Latropos: > Why call an assert when we are certain that a given value is not a null? Like all the Fabric specific functions on Fabric.

Btw. why don't we group our architecture-specific props into a single object inside global? We would get a better type validation, if we had something like:

```jsx

architecureSpecificProps = {
    _IS_FABRIC: false,
    _updatePropsFabric: ((
        operations: {
          shadowNodeWrapper: ShadowNodeWrapper;
          updates: StyleProps | AnimatedStyle<any>;
        }[]
      ) => void)
    _updatePropsPaper: undefined
    // and _measurePaper, _measureFabric etc.
} |  {
    _IS_FABRIC: true,
    _updatePropsFabric:undefined
    _updatePropsPaper: ((
        operations: {
          shadowNodeWrapper: ShadowNodeWrapper;
          updates: StyleProps | AnimatedStyle<any>;
        }[]
      ) => void)
}
```

so instead of

```jsx
if (!!(global as localGlobal)._IS_FABRIC;){ //code of isFabric()
          _updatePropsFabric!(operations);
}
```
we would have

```jsx
if (!!(global as localGlobal).architecureSpecificProps._IS_FABRIC;){
          architecureSpecificProps._updatePropsFabric(operations);
}
```
without any non-null-assertion

src/reanimated2/hook/utils.ts
tjzel: I didn't undo them, I forgot to merge this branch into current main and it had stale changes.

src/reanimated2/commonTypes.ts
```diff
@@ -13,11 +13,14 @@ export interface StyleProps extends ViewStyle, TextStyle {
  *
  * Shared values are defined using [useSharedValue](https://docs.swmansion.com/react-native-reanimated/docs/core/useSharedValue) hook. You access and modify shared values by their `.value` property.
  */
-export interface SharedValue<Value> {
+export interface SharedValue<Value = unknown> {
```
tomekzaw, line 16: Shouldn't we also update our docs (in particular "Type definitions" section [here](https://docs.swmansion.com/react-native-reanimated/docs/core/useSharedValue/))?

tomekzaw (approved):

src/reanimated2/commonTypes.ts
tjzel: Yes.

.eslintrc.js
tjzel: The problem is not with `as localGlobal`, but with the file `globals.d.ts`. TypeScript is just bad with dealing with contexts. If you want you can make a PoC of this solution and I'd gladly see it but at the moment I'm in doubt whether it will actually work without overcomplicating the code.

bartlomiejbloniarz (approved):

src/reanimated2/commonTypes.ts
tjzel: Added in 233b65d

PR #5484, opened 2023-12-08
Improve TS in src [3/6] - Add Mutable type

tjzel:
## Summary

Adding proper type for a `Mutable` instead of using `any`, also making `SharedValue` related to `Mutable` type (making it a public subset).

## Test plan

CI

Co-authored by: @marmor157.


src/reanimated2/commonTypes.ts
```diff
@@ -23,6 +24,12 @@ export interface SharedValue<Value = unknown> {
   ) => void;
 }
 
+export interface Mutable<Value = unknown> extends SharedValue<Value> {
+  _isReanimatedSharedValue: true;
+  _animation?: AnimationObject | null; // only in Native
+  _value: Value | Descriptor | AnimatableValue;
```
Latropos, line 30: Why have you chosen this type?
Why, if `_value` is of type `Descriptor` we don't call `Mutable<Descriptor>`?

tjzel: Good idea, I changed it in 22bbdf1

bartlomiejbloniarz (approved):
🚀

PR #5486, opened 2023-12-08
Improve TS in src [5/6] - Acknowledge UI type of AnimatedRef

tjzel:
## Summary

Currently `useAnimatedRef` is an object that is of certain type on JS thread but a significantly other on the UI thread. This UI thread type is added and used in every relevant place.

## Test plan

CI

Co-authored by: @marmor157.


src/reanimated2/platformFunctions/scrollTo.ts
```diff
@@ -18,25 +25,25 @@ import type { Component } from 'react';
...
+  animatedRef: AnimatedRefOnUI,
   x: number,
   y: number,
   animated: boolean
 ) {
   'worklet';
-  dispatchCommand(animatedRef as any, 'scrollTo', [x, y, animated]);
+  dispatchCommand(
+    // This assertion is needed to comply to `dispatchCommand` interface
+    animatedRef as unknown as AnimatedRef<Component>,
```
bartlomiejbloniarz, line 39: Wouldn't it be easier to just type `animatedRef` as `AnimatedRef<Component>` in this function? Or do you think it's better to stick to the convention that worklet functions should be typed with "UI types"?

tjzel: Now that I'm thinking about it maybe the parameter should be defined as a union. This function can be called on JS therefore JS representation of `animatedRef` will be present there. I think I'll add that union with a proper comment.

And to answer your question - I believe types are supposed to help us. I want the next developer that has to change something in `scrollTo` implementation to have all the information he needs in the type, instead of running the Example app and scraping `Object.keys` to understand what's going on.

PR #5495, opened 2023-12-13
Allow SETs configuration to be updated on rerender

bartlomiejbloniarz:
## Summary
This PR changes the current SETs implementation to allow for:
- changes in configuration on rerender
- tracking the changes in components on the hidden screen (the target snapshot is now created right before the hidden screen is shown)

## Additional changes
In this PR I also changed the logic behind applying transforms to views in Snapshots on iOS. Now during the transition (so when the shared element is removed from it's parent) we apply a transform to the shared element, that is a combination of all the transforms that were applied by superviews before removing this element from its parent. I decided to do this in this PR, since those changes were closely tied to other changes I was already making in the iOS Snapshot logic. I, however, decided to not apply analogous changes on Android in this PR, because due to the API differences, the logic there would be too complicated, making this already humongous PR too big/dense.

TODO:
- [x] Android
- [x] `isInteractive` issues

## Test plan

Check the behavior of `[SET] Change theme` example (and possibly all the other SET examples). Also on iOS check the `[SET] Nested transforms` example.


Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -8,48 +8,49 @@
...
+      sharedTransitionConfigs.push_back(std::move(layoutAnimationConfig));
+    } else {
+      if (config == nullptr) {
+        getConfigsForType(type).erase(tag);
+      } else {
+        getConfigsForType(type)[tag] = config;
+      }
+    }
+  }
+  for (auto [tag, type, config, sharedTransitionTag] :
```
tomekzaw, line 29: Does it work?
```suggestion
  for (const auto &[tag, type, config, sharedTransitionTag] :
```

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -8,48 +8,49 @@
...
-    const std::shared_ptr<Shareable> &config) {
-  auto lock = std::unique_lock<std::mutex>(animationsMutex_);
-  if (type == SHARED_ELEMENT_TRANSITION ||
-      type == SHARED_ELEMENT_TRANSITION_PROGRESS) {
+void LayoutAnimationsManager::configureAnimationBatch(
+    const std::vector<LayoutAnimationConfig> &layoutAnimationsBatch) {
+  auto lock = std::unique_lock<std::recursive_mutex>(animationsMutex_);
+  std::vector<LayoutAnimationConfig> sharedTransitionConfigs;
+  for (auto layoutAnimationConfig : layoutAnimationsBatch) {
+    auto [tag, type, config, sharedTransitionTag] = layoutAnimationConfig;
```
tomekzaw, line 16: Does it work?
```suggestion
    const auto &[tag, type, config, sharedTransitionTag] = layoutAnimationConfig;
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -155,11 +189,43 @@ private boolean tryStartSharedTransitionForViews(
...
     orderByAnimationTypes(sharedElements);
+    return true;
+  }
+
+  protected void onScreenWillDisappear() {
+    if (!mIsTransitionPrepared) {
+      return;
+    }
+    mIsTransitionPrepared = false;
+    for (var sharedElement : mSharedElementsWithAnimation) {
```
tomekzaw, line 200: In our codebase we mostly use concrete types instead of `var`, is this a deliberate decision to use `var` here?

apple/LayoutReanimation/REASnapshot.m
```diff
@@ -110,26 +146,31 @@ - (void)makeSnapshotForView:(REAUIView *)view useAbsolutePositionOnly:(BOOL)useA
...
+  }
+  return transform;
+}
 
-  if (isTransformed && transformedView != nil) {
-    return transformedView;
+- (REAUIView *)maybeFindTransitionView:(REAUIView *)view
+{
+  while (view != nil && ![REAScreensHelper isRNSScreenType:view]) {
+    if ([view.superview isKindOfClass:NSClassFromString(@"UITransitionView")]) {
```
tomekzaw, line 170: I'm curious to know what's the difference between this and 
```suggestion
    if ([view.superview isKindOfClass:[UITransitionView class]]) {
```

Common/cpp/LayoutAnimations/LayoutAnimationsManager.h
```diff
@@ -22,17 +22,13 @@ struct LayoutAnimationConfig {
   int tag;
   LayoutAnimationType type;
   std::shared_ptr<Shareable> config;
+  std::string sharedTransitionTag;
```
tomekzaw, line 25: Just wondering if we can make these fields `const`?

apple/LayoutReanimation/REASnapshot.m
piaskowyk: Unfortunately we can't because this is an internal Apple API and not published in the SDK :/

Common/cpp/LayoutAnimations/LayoutAnimationsManager.h
piaskowyk: image

android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -56,13 +82,22 @@ protected View getTransitioningView(int tag) {
...
+    if (eventDispatcher != null) {
+      eventDispatcher.addListener(new TopWillAppearListener(eventDispatcher));
+    }
   }
 
   protected void viewDidLayout(View view) {
-    maybeRestartAnimationWithNewLayout(view);
+    // this was causing problems when I moved the start of transition to the willAppear event
+    // handler
+    //    maybeRestartAnimationWithNewLayout(view);
```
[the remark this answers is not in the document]
piaskowyk: Yes, we want to keep it. Although it interferes with Bartek's changes, the functionality to restart the animation is less crucial compared to ensuring the proper configuration on rerender. We'll fix this in the upcoming PRs.

tomekzaw (approved):

PR #5499, opened 2023-12-14
Always inject JS version

tjzel:
## Summary

Previously we injected JS version only in the Debug bundle. This caused combination of Debug app & Release bundle to throw warnings about failed version resolution. No more.

## Test plan

:rocket:

- [x] Debug App & Debug Bundle runs without warnings
- [x] Debug App & Release Bundle runs without warnings
- [x] Release App & Debug Bundle runs without warnings
- [x] Release App & Release Bundle runs without warnings


tjzel:
@bartlomiejbloniarz removed in 3ec9f92

Common/cpp/Tools/ReanimatedVersion.cpp
```diff
@@ -68,11 +76,16 @@ void checkJSVersion(
         "See `https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooting#mismatch-between-c-code-version-and-javascript-code-version` for more details.");
     return;
   }
+}
+#else // NDEBUG
```
tomekzaw, line 80: The comment is quite misleading, maybe let's get rid of it?
```suggestion
#else
```

Common/cpp/Tools/ReanimatedVersion.cpp
```diff
@@ -68,11 +76,16 @@ void checkJSVersion(
...
+void checkJSVersion(
+    jsi::Runtime &rnRuntime,
+    const std::shared_ptr<JSLogger> &jsLogger) {
+  // NOOP
```
tomekzaw, line 84: It would be good to explain why this function is NOOP and why the next function always returns true (i.e. in which configuration exactly we skip the check).

Common/cpp/ReanimatedRuntime/RNRuntimeDecorator.cpp
```diff
@@ -32,6 +32,7 @@ void RNRuntimeDecorator::decorate(
 #ifndef NDEBUG
   checkJSVersion(rnRuntime, nativeReanimatedModule->getJSLogger());
 #endif // NDEBUG
+  injectReanimatedVersion(rnRuntime);
```
tomekzaw, line 35:
```suggestion
  injectReanimatedCppVersion(rnRuntime);
```

PR #5502, opened 2023-12-17
Layout-transition docs - Empty brackets removed

azashi:
## Summary

Layout transition docs had empty brackets, which are removed in this PR


tomekzaw (approved):
Thanks for this PR!

piaskowyk (approved):
Thank you! 😃

PR #5512, opened 2023-12-20
Mark `create-expo-app` as latest

kacperkapusciak:
It seems like npx sometimes uses an old cached version of the `create-expo-app` cli 

This PR adds the `latest` modifier to make sure that the invoked command have the `-e` command available


lukmccall (approved):
🔥🔥🔥

bartlomiejbloniarz (approved):

PR #5513, opened 2023-12-20
[Fix TS] Introduce executeOnUIRuntimeSync and use it to replace Sync Data Holder

piaskowyk:
## Summary

This PR fixes some small issues from https://github.com/software-mansion/react-native-reanimated/pull/4300

### DIFF
--- a/src/reanimated2/js-reanimated/JSReanimated.ts
+++ b/src/reanimated2/js-reanimated/JSReanimated.ts
@@ -290,6 +290,12 @@ export default class JSReanimated {
       '[Reanimated] configureProps is not available in JSReanimated.'
     );
   }
+
+  executeOnUIRuntimeSync<T, R>(_shareable: ShareableRef<T>): R {
+    throw new Error(
+      '[Reanimated] executeOnUIRuntimeSync is not available in JSReanimated.'
+    );
+  }
 }
 
 enum Platform {
--- a/src/reanimated2/js-reanimated/JSReanimated.ts
+++ b/src/reanimated2/js-reanimated/JSReanimated.ts
@@ -293,7 +293,7 @@ export default class JSReanimated {
 
   executeOnUIRuntimeSync<T, R>(_shareable: ShareableRef<T>): R {
     throw new Error(
-      '[Reanimated] executeOnUIRuntimeSync is not available in JSReanimated.'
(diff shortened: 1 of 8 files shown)


src/reanimated2/threads.ts
```diff
@@ -130,6 +130,11 @@ export function runOnUI<Args extends unknown[], ReturnValue>(
   };
 }
 
+// @ts-expect-error Check `executeOnUIRuntimeSync` overload above.
+export function executeOnUIRuntimeSync<Args extends unknown[], ReturnValue>(
+  worklet: (...args: Args) => ReturnValue
+): WorkletFunction<Args, ReturnValue>;
```
tjzel, line 136: It doesn't return a worklet actually.

```suggestion
export function executeOnUIRuntimeSync<Args extends unknown[], ReturnValue>(
  worklet: (...args: Args) => ReturnValue
): (...args: Args) => ReturnValue;
```

src/reanimated2/js-reanimated/JSReanimated.ts
```diff
@@ -290,6 +290,12 @@ export default class JSReanimated {
       '[Reanimated] configureProps is not available in JSReanimated.'
     );
   }
+
+  executeOnUIRuntimeSync<T, R>(_shareable: ShareableRef<T>): R {
+    throw new Error(
+      '[Reanimated] executeOnUIRuntimeSync is not available in JSReanimated.'
```
tjzel, line 296:
```suggestion
      '[Reanimated] `executeOnUIRuntimeSync` is not available in JSReanimated.'
```

src/reanimated2/mutables.ts
```diff
@@ -82,9 +82,9 @@ export function makeMutable<T>(initial: T): SharedValue<T> {
         })();
       }
     },
-    get value() {
+    get value(): T {
       if (!SHOULD_BE_USE_WEB) {
-        return executeOnUIRuntimeSync(uiValueGetter)(mutable);
+        return uiValueGetter(mutable) as T;
```
tjzel, line 87: Screenshot 2023-12-21 at 18 48 17

with the above change you don't need a cast here
```suggestion
        return uiValueGetter(mutable);
```

tjzel (approved):
suggested minor improvements

PR #5516, opened 2023-12-21
Fix TS and initialization order for `executeUIRuntimeSync`.

tjzel:
## Summary

Minor changes to #4300.

## Test plan

🚀

### DIFF
--- a/Common/cpp/ReanimatedRuntime/WorkletRuntime.h
+++ b/Common/cpp/ReanimatedRuntime/WorkletRuntime.h
@@ -65,10 +65,10 @@ class WorkletRuntime : public jsi::HostObject,
   std::vector<jsi::PropNameID> getPropertyNames(jsi::Runtime &rt) override;
 
  private:
+  std::recursive_mutex runtimeMutex_;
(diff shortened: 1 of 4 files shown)


Common/cpp/ReanimatedRuntime/WorkletRuntime.h
```diff
@@ -65,10 +65,10 @@ class WorkletRuntime : public jsi::HostObject,
   std::vector<jsi::PropNameID> getPropertyNames(jsi::Runtime &rt) override;
 
  private:
+  std::recursive_mutex runtimeMutex_;
```
tjzel, line 68: I think this should be moved because it's used in `WorkletRuntime::WorkletRuntime` constructor as an argument of `makeRuntime` before its own `std::recursive_mutex` default constructor was called.
While it will probably work because we are only passing the address there, I think it's safer to have it initialized already.

piaskowyk: That's a good point, and you're right! Yes, it works because we are passing a pointer instead of the value of the field

piaskowyk (approved):

PR #5518, opened 2023-12-22
Remove `makeRemote`

tjzel:
## Summary

`makeRemote` function seems to be no longer necessary.
Prettifying `useAnimatedStyle` as a bonus.

### Why can `makeRemote` be removed

`makeRemote` utilizes the `ShareableHandle` idea. Let's look at its code:

```ts
export function makeRemote(initial) {
  const handle = makeShareableCloneRecursive({
    __init: () => {
      'worklet';
      return initial;
    },
  });
  registerShareableMapping(initial, handle);
  return initial;
}
```

The object `handle` defined inside this function has `__init` worklet property. This worklet has the object `initial` in its closure. When `makeShareableCloneRecursive` is called with this worklet, `initial's` value will be copied to the `handle`'s closure. Therefore, calling `__init` on UI runtime will return the copied value of initial.

So at this point we have a `ShareableObject` `handle` with a worklet property `__init`. Both this object and its worklet were saved in `_shareableCache` when `makeShareableCloneRecursive` was called. Current mappings are:

`_shareableCache[handle]` - `handle`
`_shareableCache[handle.__init]` - `handle.__init`

Now `registerShareableMapping` is called. This function sets a new mapping for `_shareableCache`. It maps `initial` to `handle`. Therefore any worklet that has `initial` in its closure will take the mapping for `initial` instead, which is the `handle`.

Then on the UI thread, when we evaluate worklet that uses `initial`, we run it through `valueUnpacker`. `valueUnpacker` checks if the object has `__init` property, if yes, then it return the result of `__init` worklet. So it returns the copied value of `initial` that was copied during evaluation of `__init`.

To sum it up:

`makeRemote` is a function that maps a given JS reference to a function that on UI thread will return the same value. With current implementation of Shareables this hardly serves any purpose, since we could just copy the value like a plain JS value, using its copy.

For example, code:

```ts
const a = { b: 5 };
makeRemote(a);

function foo() {
  'worklet';
  console.log(a);
}
```

Provides the same results as

```ts
const a = { b: 5 };

function foo(){
  `worklet`;
  console.log(a);
}
```

Since all that `makeRemote` does is add unnecessary identity unpacking function.

## Test plan

🚀 

To check `areAnimationsActive`:

<details>

```tsx

import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withTiming,

import React from 'react';

  const sv = useSharedValue(0);

    return {
      transform: [{ rotate: `${sv.value}deg` }],
    };
  });

  return (
      <Button
        title="Start animation"
        onPress={() => (sv.value = withTiming(180, { duration: 30000 }))}
      />
      <Button title="Rerender" onPress={() => setState(state + 1)} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: {
    height: 100,
    width: 100,
    backgroundColor: 'blue',
  },
});
```

</details>


src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -244,9 +244,7 @@ export function createAnimatedComponent(
     _detachStyles() {
       if (IS_WEB && this._styles !== null) {
         for (const style of this._styles) {
-          if (style?.viewsRef) {
-            style.viewsRef.remove(this);
-          }
+          style.viewsRef.delete(this);
```
tomekzaw, line 247: Shouldn't it be `style.viewsRef?.delete(this);` here instead of in line 40?

src/reanimated2/hook/useHandler.ts
```diff
@@ -66,7 +65,7 @@ export function useHandler<
   const initRef = useRef<ContextWithDependencies<Context> | null>(null);
   if (initRef.current === null) {
     initRef.current = {
-      context: makeRemote<Context>({} as Context),
+      context: {} as Context,
```
tomekzaw, line 68: Could you please explain why is it possible to remove `makeRemote` here? How exactly can we make sure that this change will not break anything?

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -510,20 +512,27 @@ For more, see the docs: \`https://docs.swmansion.com/react-native-reanimated/doc
...
   }, []);
 
   checkSharedValueUsage(initial.value);
 
   if (isJest()) {
-    return { viewDescriptors, initial, viewsRef, animatedStyle };
+    return {
+      viewDescriptors,
+      initial,
+      maybeViewsRef,
```
tomekzaw, line 532: Do we really want to rename the key here?
```suggestion
      viewsRef: maybeViewsRef,
```

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -510,20 +512,27 @@ For more, see the docs: \`https://docs.swmansion.com/react-native-reanimated/doc
...
   }, dependencies);
 
   useEffect(() => {
-    animationsActive.value = true;
+    areAnimationsActive.value = true;
     return () => {
-      animationsActive.value = false;
+      areAnimationsActive.value = false;
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tomekzaw, line 523: I think we should add `areAnimationsActive` to the dependencies array and also ensure that `useSharedValue` returns a stable reference.

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -442,27 +442,29 @@ For more, see the docs: \`https://docs.swmansion.com/react-native-reanimated/doc
...
+      validateAnimatedStyles(initialStyle);
+    }
+    propUpdaterData.current = {
       initial: {
         value: initialStyle,
-        updater: updater,
+        updater,
       },
-      remoteState: makeRemote<AnimatedState>({
+      remoteState: {
```
tomekzaw, line 455: Why is it safe to remove `makeRemote` here?

src/createAnimatedComponent/PropsFilter.tsx
```diff
@@ -37,7 +37,7 @@ export class PropsFilter implements IPropsFilter {
         const processedStyle: StyleProps = styles.map((style) => {
           if (style && style.viewDescriptors) {
             // this is how we recognize styles returned by useAnimatedStyle
-            style.viewsRef.add(component);
+            style.viewsRef?.add(component);
```
tomekzaw, line 40: Do we really need to make this change here?

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -37,7 +37,7 @@ interface AnimatedState {
   isAnimationCancelled: boolean;
 }
 
-interface AnimationRef {
+interface PropUpdaterData {
```
tomekzaw, line 40: We already use the term "prop updater" for `JSPropsUpdater`, but this thing looks way different, we'll need to come up with a different name.

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -411,16 +411,16 @@ export function useAnimatedStyle<Style extends DefaultStyle>(
   isAnimatedProps = false
 ) {
   const viewsRef: ViewRefSet<unknown> = useViewRefSet();
```
tomekzaw, line 413: 1. Do we really want to call `useViewRefSet()` on each render?
2. How about calling `useViewRefSet`  conditionally only on web?
```suggestion
  const maybeViewsRef = SHOULD_BE_USE_WEB ? useViewRefSet() : undefined;
```

tomekzaw (changes requested):

src/reanimated2/hook/useAnimatedStyle.ts
tjzel: I was considering doing that initially but then scrapped the idea do make this PR touch only `makeRemote`. I can add it here however.

src/reanimated2/hook/useAnimatedStyle.ts
tjzel: Seems like a mistake.

src/reanimated2/hook/useHandler.ts
tjzel: I will add a detailed description in the PR why `makeRemote` is no longer necessary and why are those changes safe.

tjzel: @tomekzaw added!

src/reanimated2/hook/useAnimatedStyle.ts
tjzel: Regarding 1. `useViewRefSet` uses `useRef` inside so there's no problem calling it on every render.

src/reanimated2/hook/useAnimatedStyle.ts
tjzel: Doesn't `useSharedValue` return a stable reference now with `useState` implementation?

src/reanimated2/hook/useAnimatedStyle.ts
tjzel: It's now in PR's description.

src/createAnimatedComponent/PropsFilter.tsx
tjzel: Because of inconsistency in `useAnimatedStyle`. We would always make `viewsRef` object in it, but also make `maybeViewsRef` object. The latter was sent to some functions while the former was always returned.

Therefore in native implementations we would unnecessarily return it and add to it in this place. Now that we return undefined instead we need to check if it's defined.

src/createAnimatedComponent/createAnimatedComponent.tsx
tjzel: No need to use `?` here since we have `IS_WEB` check above.

src/reanimated2/hook/useAnimatedStyle.ts
tjzel: Name for this thing is very hard 😭

src/reanimated2/hook/useAnimatedStyle.ts
```diff
@@ -510,19 +511,25 @@ For more, see the docs: \`https://docs.swmansion.com/react-native-reanimated/doc
...
 
   useEffect(() => {
-    animationsActive.value = true;
+    areAnimationsActive.value = true;
     return () => {
-      animationsActive.value = false;
+      areAnimationsActive.value = false;
     };
-  }, []);
+  }, [areAnimationsActive]);
```
piaskowyk, line 522: Did you tested this change? It could potentially break animation during render of component? Could you check what will happen if you trigger rerender on component with running animation?

piaskowyk (commented):
Overall, everything is going great. However, I do have one doubt regarding potential regression.

src/reanimated2/hook/useAnimatedStyle.ts
tjzel: Tested it and it seems to be fine.

src/reanimated2/hook/useAnimatedStyle.ts
tjzel: Relevant test was added to the PR description.

tomekzaw (approved):
🚢 🇮🇹

src/reanimated2/hook/useHandler.ts
tomekzaw: Fixed in #5586.

src/reanimated2/hook/useAnimatedStyle.ts
tomekzaw: @bartlomiejbloniarz Here's another call to `makeRemote` that was removed in this PR. Can you please confirm if it doesn't break anything? Maybe we should introduce a fix similar to #5586 here? We can also rename `remoteState` to `uiState`.

PR #5528, opened 2023-12-27
Fix FlatList item generic types

jkadamczyk:
## Summary

I think I ran into a typescript types regression, where the `Animated.FlatList` type was no longer generic. It always used the any type and I couldn't provide a custom item type myself anymore. I found it while upgrading from Reanimated v2 to v3. I think I managed to patch the generated .d.ts file properly locally and this is what did the trick for me.

If this isn't matching your code style or requirements, I'm happy to work with you on making this work.

## Test plan

Probably makes sense to see if it breaks anything in the example app but other than that 🤷

### DIFF
--- a/src/reanimated2/component/FlatList.tsx
+++ b/src/reanimated2/component/FlatList.tsx
@@ -1,5 +1,4 @@
 'use strict';
-import type { ForwardedRef } from 'react';
 import React, { forwardRef } from 'react';
 import type {
   FlatListProps,
@@ -64,7 +63,7 @@ interface AnimatedFlatListComplement<T> extends FlatList<T> {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const FlatListForwardRefRender = function <Item>(
   props: ReanimatedFlatListPropsWithLayout<Item>,
-  ref: ForwardedRef<FlatList>
+  ref: React.ForwardedRef<FlatList>
 ) {
   const { itemLayoutAnimation, skipEnteringExitingAnimations, ...restProps } =
     props;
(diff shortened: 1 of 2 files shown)


src/reanimated2/component/FlatList.tsx
```diff
@@ -60,48 +60,58 @@ interface AnimatedFlatListComplement<T> extends FlatList<T> {
...
+export const ReanimatedFlatList = forwardRef(FlatListForwardRefRender) as <
+  // We need explicit any here, because this is the exact same type that is used in React Native types.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ItemT = any
+>(
+  props: ReanimatedFlatListPropsWithLayout<ItemT> & {
+    ref?: ForwardedRef<FlatList>;
   }
-);
+) => React.ReactElement;
```
tjzel, line 114: This makes me curious, are we certain that this type of general type here won't cause any issues?

tjzel (changes requested):
Thank you for finding that @jkadamczyk! I took the liberty and committed some minor changes directly into your PR. Everything looks well, apart from that one question that I asked. Once you assure me everything is fine I'm happy to merge it 😃

src/reanimated2/component/FlatList.tsx
jkadamczyk: Hey! Let me look into that!

jkadamczyk: @tjzel I wasn't able to find a more elegant solution. If you have some ideas, please share them with me.
To back using it like it is I can share the function call type declaration in the ExoticComponent interface. It also returns a `ReactElement`. But for sure there are parts of that missing.
CleanShot 2023-12-28 at 19 06 51@2x

In my solution which I have in a patch already I was using below sources:
- https://fettblog.eu/typescript-react-generic-forward-refs/
- https://stackoverflow.com/questions/58469229/react-with-typescript-generics-while-using-react-forwardref

jkadamczyk: Also https://github.com/Microsoft/TypeScript/issues/17574

tjzel: Hey, that's some amazing research there, very interesting reads! (it's tempting to request more changes just to read some more 😄)

Let's merge this bad boy

tjzel (approved):

PR #5530, opened 2023-12-28
Move `shareableCache` related code to a dedicated file and streamline its logic

tjzel:
## Summary

Extension of #5434.

## Test plan

🚀

### DIFF
--- a/src/reanimated2/shareableCache.ts
+++ b/src/reanimated2/shareableCache.ts
@@ -1,4 +1,5 @@
 'use strict';
+import { shouldBeUseWeb } from './PlatformChecker';
 import type { ShareableRef } from './commonTypes';
 
 /*
@@ -12,4 +13,32 @@ There we use Object.entries to iterate over the keys which throws an error on ac
 For convenience we move this cache to a separate file so it doesn't scare us with red squiggles.
 */
 
-export const shareableCache = new WeakMap<object, ShareableRef | symbol>();
+const SHOULD_BE_USE_WEB = shouldBeUseWeb();
+
+/**
+ * This symbol is used to represent a mapping from the value to itself.
+ *
+ * It's used to prevent converting a shareable that's already converted -
+ * for example a Shared Value that's in worklet's closure.
+ **/
+export const shareableFlag = Symbol('shareable flag');
+
+const cache = SHOULD_BE_USE_WEB
+  ? null
+  : new WeakMap<object, ShareableRef | symbol>();
+
+export const shareableCache = cache
+  ? {
+      set(shareable: object, shareableRef?: ShareableRef): void {
+        cache.set(shareable, shareableRef || shareableFlag);
+      },
+      get: cache.get,
+    }
+  : {
+      set() {
+        // NOOP
+      },
+      get() {
+        return null;
+      },
+    };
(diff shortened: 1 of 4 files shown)


bartlomiejbloniarz (approved):
💯

src/reanimated2/shareableCache.ts
```diff
@@ -12,4 +23,22 @@ There we use Object.entries to iterate over the keys which throws an error on ac
 For convenience we move this cache to a separate file so it doesn't scare us with red squiggles.
 */
 
-export const shareableCache = new WeakMap<object, ShareableRef | symbol>();
+const cache = SHOULD_BE_USE_WEB
+  ? null
+  : new WeakMap<object, ShareableRef | symbol>();
+
+export const shareableCache = cache
```
tomekzaw, line 30: This is more of a philosophical decision but I think we should use base conditions instead of previous effects. This leads to duplicated code (in some sense) but probably easier to understand.
```suggestion
export const shareableCache = SHOULD_BE_USE_WEB
```

src/reanimated2/shareableCache.ts
```diff
@@ -12,4 +23,22 @@ There we use Object.entries to iterate over the keys which throws an error on ac
...
+        cache.set(shareable, shareableRef || shareableFlag);
+      },
+      get: cache.get,
+    }
+  : {
+      set() {
+        // NOOP
+      },
+      get() {
+        return null;
```
tomekzaw, line 42: Maybe let's throw an error here that says "`shareableCache.get` should not be called on web"? Or, even better, we could add some helper error function that would be called here.

tomekzaw (commented):
Looks good to me, left some minor suggestions

src/reanimated2/shareableCache.ts
tjzel: I wanted to do it, but you know... TypeScript...
but this is the right thing to do, since checking for `cache` here might not be a scalable solution.

src/reanimated2/shareableCache.ts
tjzel: What do you mean by 'helper error function'?

tomekzaw: Some function that we call from here and other places that shouldn't be called on web that throws an error that this function should not be called on web. Something like `__builtin_trap()`.

tjzel: If you want to do it in other places that seems like a separate PR. In this one I didn't want to change the behavior, since currently on web `shareableCache` is always empty.

src/reanimated2/shareableCache.ts
```diff
@@ -1,15 +1,44 @@
...
+During cloning we use `Object.entries` to iterate over the keys which throws an error on accessing `_value`.
+For convenience we moved this cache to a separate file so it doesn't scare us with red squiggles.
 */
 
-export const shareableCache = new WeakMap<object, ShareableRef | symbol>();
+const cache = SHOULD_BE_USE_WEB
+  ? null
+  : new WeakMap<object, ShareableRef | symbol>();
+
+export const shareableCache = SHOULD_BE_USE_WEB
```
piaskowyk, line 30: In my opinion, a more precise name for it could be `shareableMapping` or `shareableMappingCache`.

tjzel: `shareableMappingCache` is a great idea!

PR #5531, opened 2023-12-28
Add suggestions to errors for accessing mutable's `value`

### LINKED ISSUE #5117: Confusing error when passing animated style to a non-animated component
### Description

Suppose I have this.

```js
    return {
      // ...
    }
  })
 return <Image style={[animatedStyle]} />
```

I get an error that's very difficult to understand. It doesn't give me any information that would hint at what actually went wrong. It says `Reading from _value directly is only possible on the UI runtime` but I don't know what `_value` is, and I definitely don't read it directly myself. And the stack frame it shows doesn't even have `_value` access.

Screenshot 2023-09-21 at 19 06 57

The problem with the code above is that I forgot to do `Animated.createAnimatedComponent(Image)`.

I'd like the error message to be clearer if possible — at least in development.

### Steps to reproduce

see above

sorry

### Reanimated version

3.4.2

### React Native version

0.72.4

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

tjzel:
## Summary

Adding it as an answer to feedback from #5117. Fixes #5117.

## Test plan

🪨

### DIFF
--- a/src/reanimated2/mutables.ts
+++ b/src/reanimated2/mutables.ts
@@ -108,7 +108,7 @@ export function makeMutable<T>(
     set _value(newValue: T) {
       if (!SHOULD_BE_USE_WEB) {
         throw new Error(
-          '[Reanimated] Setting `_value` directly is only possible on the UI runtime.'
+          '[Reanimated] Setting `_value` directly is only possible on the UI runtime. Perhaps you want to assign to `value` instead?'
         );
       }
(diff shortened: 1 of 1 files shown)


bartlomiejbloniarz (approved):
👀

PR #5536, opened 2023-12-29
Split TS type tests into separate files

tjzel:
## Summary

Previously file `common.tsx` in typetests had over 2000 lines. Now it's split into separate files. Type testing scripts were changed accordingly.

## Test plan

CI


piaskowyk (approved):
🫡

PR #5537, opened 2023-12-29
useAnimatedSensor | TypeError: Cannot assign to read-only property 'sensor' fixed.

talut:
TypeError: Cannot assign to read-only property 'sensor'

## Summary

This pull request aims to resolve the issue related to the **TypeError: Cannot assign to read-only property 'sensor'.** When attempting to modify a read-only property. In this case, the issue arises from reassigning ref.current.sensor within the useRef hook.

While the object returned by useRef is mutable, modifying a read-only property within this object, as done with ref.current.sensor, leads to the error in question.

In the current implementation, ref.current.sensor is initially set with the function initializeSensor(sensorType, config) and is later reassigned in the useEffect hook. It is this reassignment that triggers the error.

To fix this issue, I suggest a code restructuring to avoid the direct reassignment of the sensor property. Instead, we can create a new object for ref.current each time an update is needed. This approach sidesteps the limitations of the read-only property while retaining the intended functionality.

## Test plan

```sh
yarn add react-native-reanimated
```

```ts
import { SensorType, useAnimatedSensor } from 'react-native-reanimated';

const SensorTest = () => {
  /*
   * Issue:  TypeError: Cannot assign to read-only property 'sensor'
   *
   * Why: This issuse happening while we setting custom userConfig
   * Example: { interval: 10 }
   */
  const animatedSensor = useAnimatedSensor(SensorType.GYROSCOPE, {
    interval: 10,
  });

 useDerivedValue(() => {
    const x = animatedSensor.sensor.value.x;
    const y = animatedSensor.sensor.value.y;

    console.log('x', x);
    console.log('y', y);
  }, [animatedSensor.sensor.value.x, animatedSensor.sensor.value.y]);

  return (
      <Button title="Re-render" onPress={() => setState(state + 1)} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

### DIFF
--- a/src/reanimated2/hook/useAnimatedSensor.ts
+++ b/src/reanimated2/hook/useAnimatedSensor.ts
@@ -130,7 +130,7 @@ export function useAnimatedSensor(
 
     ref.current = {
       sensor: newSensor,
-		  unregister: id !== -1 ? () => unregisterSensor(id) :  () => {
+      unregister: id !== -1 ? () => unregisterSensor(id) :  () => {
         // NOOP
       },
       isAvailable: id !== -1,
--- a/src/reanimated2/hook/useAnimatedSensor.ts
+++ b/src/reanimated2/hook/useAnimatedSensor.ts
@@ -110,13 +110,12 @@ export function useAnimatedSensor(
       ...config,
       ...userConfig,
     };
-    ref.current.sensor = initializeSensor(sensorType, newConfig);
 
-    const sensorData = ref.current.sensor;
-    const adjustToInterfaceOrientation =
-      ref.current.config.adjustToInterfaceOrientation;
+    const newSensor = initializeSensor(sensorType, newConfig);
+    const sensorData = newSensor;
+    const adjustToInterfaceOrientation = newConfig.adjustToInterfaceOrientation;
 
-    const id = registerSensor(sensorType, config, (data) => {
+    const id = registerSensor(sensorType, newConfig, (data) => {
       'worklet';
       if (adjustToInterfaceOrientation) {
         if (sensorType === SensorType.ROTATION) {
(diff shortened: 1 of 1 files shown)


tjzel:
Oh I see, I found https://github.com/software-mansion/react-native-reanimated/issues/5538 and it happens on re-render. I updated your snippet accordingly.

tjzel (commented):
What concerns me here is the fact that with these changes we are no longer returning a stable reference. This is potentially a breaking change. 

Maybe we should instead re-work the mechanism due to which this object is getting frozen in the first place, I'll make a PoC of this.

cc @piaskowyk

tjzel:
I opened an alternate pull request to this one. We have to think which one do we want to apply here.

Keep in mind that the problem arises only because return value of `useAnimatedSensor` is not accessed granularly enough by you (it's not your fault however, since we haven't put any information about it in general). If you instead did:

```ts
  const { sensor } = useAnimatedSensor(SensorType.GYROSCOPE, {
    interval: 10,
  });

 useDerivedValue(() => {
    const x = sensor.value.x;
    const y = sensor.value.y;

    console.log('x', x);
    console.log('y', y);
  }); // No need to specify dependencies when Reanimated Babel plugin isn't disabled.
```

Then everything would be fine.

talut:
Hey @tjzel thanks for the updates. 

You're right my solution returns an unstable reference.

> I opened an alternate pull request to this one. We have to think which one do we want to apply here.
> 
> Keep in mind that the problem arises only because return value of `useAnimatedSensor` is not accessed granularly enough by you (it's not your fault however, since we haven't put any information about it in general). If you instead did:
> 
> ```ts
>   const { sensor } = useAnimatedSensor(SensorType.GYROSCOPE, {
>     interval: 10,
>   });
> 
>  useDerivedValue(() => {
>     const x = sensor.value.x;
>     const y = sensor.value.y;
> 
>     console.log('x', x);
>     console.log('y', y);
>   }); // No need to specify dependencies when Reanimated Babel plugin isn't disabled.
> ```
> 
> Then everything would be fine.


I'll try this one.

Thanks.

tjzel:
Hi @talut, after a debate we decided to actually use a solution very similar to yours, but with added memoization in #5646. I hope you don't mind it, since you were kind-of first here.

PR #5548, opened 2024-01-02
Give more meaningful warning when modifying a Shareable

tjzel:
## Summary

Wanted to open this PR for a long time.

Currently we use `Object.freeze` in development on objects after they were converted to a Shareable. This is problematic for users since there is no information about why is the object immutable all of a sudden.

While this PR won't give a meaningful error when user tries to add a new property to the object when doing something like:

```ts
const a = {};

function foo(){
'worklet';
console.log(a.b);
};

runOnUI(worklet)();
a.b = 1;
runOnUI(worklet)();
```

but it seems to be a rare case and mostly the problems are with modifying existing properties.

## Test plan

```ts
const a = { b: 1 };
makeShareableCloneRecursive(a);
a.b = 2; // Warning here.
```


docs/docs/guides/troubleshooting.mdx
```diff
@@ -155,6 +155,72 @@ LogBox.ignoreLogs([
...
+obj.prop = 2; // Warning will be displayed here.
+runOnUI(worklet)();
+```
+
+and expect the results to be `1` and `2`. However, the results will be `1` and `1` because `obj` is not a Shared Value and is only copied to UI runtime once. Therefore, in development builds, we make the object immutable and add this warning after copying it to signal that it's not a valid use of Reanimated. To fix this, you should use a Shared Value instead:
+
+**Solution:**
+
+```js
+const obj = useSharedValue({ prop: 1 });
```
tomekzaw, line 182: Can we rename obj to sv? Also, solution should be a diff, what do you think?

src/reanimated2/shareables.ts
```diff
@@ -241,14 +241,28 @@ Offending code was: \`${getWorkletCode(value)}\``);
...
+      if (__DEV__ && !isHostObject(value) && isPlainJSObject(value)) {
+        // We freeze objects that are transformed to Shareable with a meaningful warning.
+        // This should help detect issues when someone modifies data after it's been converted to
+        // Shareable. Meaning that they may be doing a faulty assumption in their
         // code expecting that the updates are going to automatically populate to
         // the object sent to the UI thread. If the user really wants some objects
-        // to be mutable they should use shared values instead.
-        Object.freeze(value);
+        // to be mutable they should use Shared Values instead.
+        Object.entries(value).forEach(([key, element]) => {
```
tomekzaw, line 251: Can we come up with some metric here and measure the performance impact of this change?

app/src/examples/CubesExample.tsx
```diff
@@ -293,10 +293,7 @@ export default function CubesExample() {
 
   return (
     <View style={componentStyle.container}>
-      <Button
-        title={'log data'}
-        onPress={() => console.log(animatedSensor.sensor.value)}
-      />
+      <Button title={'log data'} onPress={() => console.log(sensor.value)} />
```
tomekzaw, line 296:
```suggestion
      <Button title="log data" onPress={() => console.log(sensor.value)} />
```

app/src/examples/OldAnimatedSensorExample.tsx
```diff
@@ -7,20 +7,17 @@ import Animated, {
...
     };
   });
 
   return (
     <View style={componentStyle.container}>
-      <Button
-        title={'log data'}
-        onPress={() => console.log(animatedSensor.sensor.value)}
-      />
+      <Button title={'log data'} onPress={() => console.log(sensor.value)} />
```
tomekzaw, line 20:
```suggestion
      <Button title="log data" onPress={() => console.log(sensor.value)} />
```

docs/docs/guides/troubleshooting.mdx
```diff
@@ -155,6 +155,72 @@ LogBox.ignoreLogs([
...
+### Tried to modify key of an object which has been converted to a Shareable.
+
+**Problem:** This warning is displayed to inform the user that a Shared Value should be used or an object used in a worklet should be accessed more granularly.
+
+#### 1. Not using Shared Values.
+
+You might get this warning when you do something along the lines of:
+
+```js
+const obj = { prop: 1 };
```
tomekzaw, line 167: Let's add newlines around the function for better readability.

docs/docs/guides/troubleshooting.mdx
```diff
@@ -155,6 +155,72 @@ LogBox.ignoreLogs([
...
+function worklet() {
+  'worklet';
+  console.log(obj.prop);
+}
+runOnUI(worklet)();
+obj.prop = 2; // Warning will be displayed here.
```
tomekzaw, line 173: Let's paste the beginning of an actual warning here?

src/reanimated2/shareables.ts
```diff
@@ -241,14 +241,28 @@ Offending code was: \`${getWorkletCode(value)}\``);
         shareableMappingCache.set(value, inaccessibleObject);
         return inaccessibleObject;
       }
-      if (__DEV__) {
-        // we freeze objects that are transformed to shareable. This should help
-        // detect issues when someone modifies data after it's been converted to
-        // shareable. Meaning that they may be doing a faulty assumption in their
+      if (__DEV__ && !isHostObject(value) && isPlainJSObject(value)) {
+        // We freeze objects that are transformed to Shareable with a meaningful warning.
```
tomekzaw, line 245:
```suggestion
        // We freeze objects that are transformed to a shareable with a meaningful warning.
```

src/reanimated2/shareables.ts
```diff
@@ -241,14 +241,28 @@ Offending code was: \`${getWorkletCode(value)}\``);
...
-        Object.freeze(value);
+        // to be mutable they should use Shared Values instead.
+        Object.entries(value).forEach(([key, element]) => {
+          Object.defineProperty(value, key, {
+            get() {
+              return element;
+            },
+            set() {
+              console.warn(
+                `[Reanimated] Tried to modify key \`${key}\` of an object which has been converted to a Shareable. This will not affect worklets which have already received the object. If you want this change to be reflected in the worklets, consider using a Shared Value instead. See
```
tomekzaw, line 258: Shareable is an implementation details, let's not leak internals into error messages.
```suggestion
                `[Reanimated] Tried to modify key \`${key}\` of an object which has been already passed to a worklet. The new value will not be propagated to worklets which have already received the object which may lead to inconsistencies. If you want this change to be reflected in the worklets, consider using a shared value instead. See 
```
Also, let's just provide a link to troubleshooting instead of shooting the full solution here.

tomekzaw (changes requested):
Certainly a PR that we all need and deserve, let's just polish some rough edges.

src/reanimated2/shareables.ts
tjzel: Since it was development only change I neglected that but sure we can.

docs/docs/guides/troubleshooting.mdx
tjzel: Yeah, diff is a nice idea here.

src/reanimated2/shareables.ts
tjzel: I'd like to keep capital Shareable to give a distinction that it's a specific concept

tjzel:
@tomekzaw I moved docs and examples part to a separate PR #5560 after applying review suggestions

src/reanimated2/shareables.ts
tjzel: That's debatable 🤓

tomekzaw: From the results you have provided, the difference is negligible 👍

src/reanimated2/shareables.ts
tomekzaw: I won't fight about this one but I still think we need to stick with lower-case terms (i.e. shared value, mutable, shareable).

src/reanimated2/shareables.ts
```diff
@@ -241,14 +241,28 @@ Offending code was: \`${getWorkletCode(value)}\``);
         shareableMappingCache.set(value, inaccessibleObject);
         return inaccessibleObject;
       }
-      if (__DEV__) {
-        // we freeze objects that are transformed to shareable. This should help
-        // detect issues when someone modifies data after it's been converted to
-        // shareable. Meaning that they may be doing a faulty assumption in their
-        // code expecting that the updates are going to automatically populate to
+      if (__DEV__ && !isHostObject(value) && isPlainJSObject(value)) {
```
tomekzaw, line 244: Would it make sense to move `&& !isHostObject(value)` inside `isPlainJSObject`?

docs/docs/guides/troubleshooting.mdx
```diff
@@ -155,6 +155,86 @@ LogBox.ignoreLogs([
 
 See [the accessibility overview](accessibility) to learn more about Reduced Motion.
 
+### Tried to modify key of an object which has been converted to a Shareable.
+
+**Problem:** This warning is displayed to inform the user that a Shared Value should be used or an object used in a worklet should be accessed more granularly.
+
+#### 1. Not using Shared Values.
```
tomekzaw, line 162:
```suggestion
### Tried to modify key of an object which has been converted to a shareable.

**Problem:** This warning is displayed to inform the user that a shared value should be used or an object used in a worklet should be accessed more granularly.

#### 1. Not using shared values.
```

docs/docs/guides/troubleshooting.mdx
```diff
@@ -155,6 +155,86 @@ LogBox.ignoreLogs([
...
+ function worklet() {
+   'worklet';
+-  console.log(obj.prop);
++  console.log(sv.value.prop);
+ }
+
+ runOnUI(worklet)();
+-obj.prop = 2; // Warning: Tried to modify key `prop` of an object which has been already passed to a worklet.
++sv.value = { prop: 2 }; // Everything is fine here.
++// Keep in mind that you cannot modify the property directly with `sv.value.prop = 2`
```
tomekzaw, line 196: You can also use `.modify` here but usually it's not recommended.

docs/docs/guides/troubleshooting.mdx
```diff
@@ -155,6 +155,86 @@ LogBox.ignoreLogs([
...
++
+ function worklet() {
+   'worklet';
+-  console.log(obj.propAccessedInWorklet);
++  console.log(propAccessedInWorklet);
+ }
+
+ runOnUI(worklet)();
+-obj.propNotAccessedInWorklet = 3; // Warning: Tried to modify key `prop` of an object which has been already passed to a worklet.
++obj.propNotAccessedInWorklet = 3; // Everything is fine here.
```
tomekzaw, line 235:
```suggestion
obj.propNotAccessedInWorklet = 3; // Everything is fine here.
```

src/reanimated2/shareables.ts
```diff
@@ -241,14 +241,28 @@ Offending code was: \`${getWorkletCode(value)}\``);
...
+        // We freeze objects that are transformed to a Shareable with a meaningful warning.
+        // This should help detect issues when someone modifies data after it's been converted.
+        // Meaning that they may be doing a faulty assumption in their
+        // code expecting that the updates are going to automatically propagate to
         // the object sent to the UI thread. If the user really wants some objects
-        // to be mutable they should use shared values instead.
-        Object.freeze(value);
+        // to be mutable they should use Shared Values instead.
+        Object.entries(value).forEach(([key, element]) => {
```
piaskowyk, line 251: Can we use here just JSProxy instead of it?

tjzel: We don't have access to user's reference sadly :c

docs/docs/guides/troubleshooting.mdx
tjzel: This was diff produced by `diff -u`, I don't think I should change that here.

piaskowyk (approved):
For me is ok 👌

src/reanimated2/shareables.ts
tjzel: I moved this part to a separate function in 9d40d1f.

docs/docs/guides/troubleshooting.mdx
tjzel: Actually let's add it.

src/reanimated2/shareables.ts
```diff
@@ -138,9 +138,12 @@ export function makeShareableCloneRecursive<T>(
         toAdapt = value.map((element) =>
           makeShareableCloneRecursive(element, shouldPersistRemote, depth + 1)
         );
+        freezeObjectIfDev(value);
       } else if (isTypeFunction && value.__workletHash === undefined) {
         // this is a remote function
         toAdapt = value;
+        console.log('Remote function');
```
tomekzaw, line 145:
```suggestion
```

src/reanimated2/shareables.ts
```diff
@@ -200,7 +204,6 @@ Offending code was: \`${getWorkletCode(value)}\``);
             return new RegExp(pattern, flags);
           },
         });
-        shareableMappingCache.set(value, handle);
```
tomekzaw, line 203: why removed?

tjzel: accident

app/src/examples/ShareablesExample.tsx
```diff
@@ -159,10 +193,166 @@ function DataViewDemo() {
...
 }
 
+function FreezingShareables() {
+  const navigation =
+    useNavigation<NativeStackNavigationProp<StackParams, 'Freezing'>>();
+
+  return (
+    <View style={styles.container}>
+      <Button
+        title={'Try modify converted array'}
```
tomekzaw, line 203:
```suggestion
        title="Try modify converted array"
```

app/src/examples/ShareablesExample.tsx
```diff
@@ -159,10 +193,166 @@ function DataViewDemo() {
...
+      <Button
+        title={'Try modify converted host object'}
+        onPress={tryModifyConvertedHostObject}
+      />
+      <Button
+        title={'Try modify converted plain object'}
+        onPress={tryModifyConvertedPlainObject}
+      />
+      <Button
+        title={'Try modify converted regex'}
```
tomekzaw, line 222:
```suggestion
        title={'Try modify converted regex literal'}
```

app/src/examples/ShareablesExample.tsx
```diff
@@ -159,10 +193,166 @@ function DataViewDemo() {
...
+      <Button
+        title={'Try modify converted regex instance'}
+        onPress={tryModifyConvertedRegexInstance}
+      />
+      <Button
+        title={'Try modify converted array buffer'}
+        onPress={tryModifyConvertedArrayBuffer}
+      />
+      <Button
+        title={'Try modify converted int32 array'}
```
tomekzaw, line 234:
```suggestion
        title={'Try modify converted Int32Array'}
```

app/src/examples/ShareablesExample.tsx
```diff
@@ -159,10 +193,166 @@ function DataViewDemo() {
...
+      <Button
+        title={'Try modify converted regex'}
+        onPress={tryModifyConvertedRegexLiteral}
+      />
+      <Button
+        title={'Try modify converted regex instance'}
+        onPress={tryModifyConvertedRegexInstance}
+      />
+      <Button
+        title={'Try modify converted array buffer'}
```
tomekzaw, line 230:
```suggestion
        title={'Try modify converted ArrayBuffer'}
```

app/src/examples/ShareablesExample.tsx
```diff
@@ -159,10 +193,166 @@ function DataViewDemo() {
...
+      <Button
+        title={'Try modify converted plain object'}
+        onPress={tryModifyConvertedPlainObject}
+      />
+      <Button
+        title={'Try modify converted regex'}
+        onPress={tryModifyConvertedRegexLiteral}
+      />
+      <Button
+        title={'Try modify converted regex instance'}
```
tomekzaw, line 226:
```suggestion
        title={'Try modify converted RegExp instance'}
```

app/src/examples/ShareablesExample.tsx
```diff
@@ -1,9 +1,38 @@
...
+  NativeStackNavigationProp,
+  createNativeStackNavigator,
+} from '@react-navigation/native-stack';
+
+type StackParams = {
+  Creating: undefined;
+  Freezing: undefined;
+};
+
+const Stack = createNativeStackNavigator<StackParams>();
```
tomekzaw, line 20: Nested stack makes the example overcomplicated.

app/src/examples/ShareablesExample.tsx
```diff
@@ -159,10 +193,118 @@ function DataViewDemo() {
...
+        title="Try modify converted host object"
+        onPress={tryModifyConvertedHostObject}
+      />
+      <Button
+        title="Try modify converted plain object"
+        onPress={tryModifyConvertedPlainObject}
+      />
+      <Button
+        title="Try modify converted regex literal"
+        onPress={tryModifyConvertedRegexLiteral}
```
tomekzaw, line 220:
```suggestion
        title="Try modify converted RegExp literal"
        onPress={tryModifyConvertedRegExpLiteral}
```

app/src/examples/ShareablesExample.tsx
```diff
@@ -159,10 +193,118 @@ function DataViewDemo() {
...
+      />
+      <Button
+        title="Try modify converted RegExp instance"
+        onPress={tryModifyConvertedRegexInstance}
```
tomekzaw, line 224:
```suggestion
        onPress={tryModifyConvertedRegExpInstance}
```

app/src/examples/ShareablesExample.tsx
```diff
@@ -159,10 +193,118 @@ function DataViewDemo() {
...
+  };
+  makeShareableCloneRecursive(obj);
+  obj.a = 2; // should warn because it's frozen
+}
+
+function tryModifyConvertedRegexLiteral() {
+  const regexLiteral = /a/;
+  makeShareableCloneRecursive(regexLiteral);
+  // @ts-expect-error
+  regexLiteral.regexProp = 2; // shouldn't warn because it's not frozen
```
tomekzaw, line 276: Let's just use `obj` and `prop` to make the code more repeatable and easier to spot differences.
```suggestion
  const obj = /a/;
  makeShareableCloneRecursive(obj);
  // @ts-expect-error
  obj.prop = 2; // shouldn't warn because it's not frozen
```

tomekzaw (approved):

giantslogik:
@tomekzaw @tjzel this PR crashes our App (after upgrade to 3.12.0) . It crashes on launch with 'property is not configurable'. Any insights into debugging will be appreciated , it crashes at shareables.ts: 321 .
https://github.com/software-mansion/react-native-reanimated/issues/6082

tjzel:
@giantslogik It's fixed.

MicroDroid:
How do I ignore this warning? I have a situation like:

```tsx
<LibraryComponent ref={ref => forwardedRef.current = ref} />
```

And it's triggering this warning, and (1) I don't know why `forwardedRef` is passed to UI thread to begin with, (2) I don't have control over the library component (unless I fork it), and (3) I'm really just setting the ref here and it's never used in the UI thread so I just don't care overall

MicroDroid:
In fact it happens with just `ref={forwardedRef}`

PR #5553, opened 2024-01-04
Setup custom eslint plugin

Latropos:
## Summary
Setup custom eslint plugi with one rule - no animated style in non animated component. This rule is turned on in example app.


app/yarn.lock
```diff
@@ -331,6 +331,121 @@
   dependencies:
     "@types/hammerjs" "^2.0.36"
 
+"@esbuild/aix-ppc64@0.20.0":
```
Latropos, line 334: It looks that esbuild has to be linked here too, it is a dependency of eslintPlugin

eslintPlugin/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "license": "MIT",
+  "types": "lib/typescript/index",
+  "scripts": {
+    "test": "yarn jest",
+    "bundle": "yarn esbuild ./public/index.js --external:espree --external:@typescript-eslint --platform=node --bundle --outfile=./index.js",
+    "build": "yarn tsc; yarn bundle; rm -rf ./public/*"
+  },
+  "dependencies": {
+    "esbuild": "^0.20.0"
+  },
```
tjzel, line 13: This should be a `devDependency` too.

package.json
```diff
@@ -92,6 +92,7 @@
   "dependencies": {
     "@babel/plugin-transform-object-assign": "^7.16.7",
     "@babel/preset-typescript": "^7.16.7",
+    "@typescript-eslint/rule-tester": "^6.20.0",
```
tjzel, line 95: This has to be a devDependency.

__tests__/eslintPlugin/noAnimatedStyleToNonAnimatedComponent.test.tsx
```diff
@@ -0,0 +1,121 @@
+import { RuleTester } from '@typescript-eslint/rule-tester';
+import noAnimatedStyleToNonAnimatedComponent from '../../eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent';
```
tjzel, line 2: I don't quite like the fact that we are not testing the bundled output, but source code here instead. We should do a require here for `../eslintPlugin/`.

__tests__/eslintPlugin/noAnimatedStyleToNonAnimatedComponent.test.tsx
tjzel, line 1: Where did you get this test structure? It would be good to mimic test structure of popular eslint plugins, if you haven't done this already.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,155 @@
+import { AST_NODE_TYPES, TSESLint, TSESTree } from '@typescript-eslint/utils';
+
+const rule: TSESLint.RuleModule<'animatedStyle' | 'sharedValue'> = {
+  create: function (context) {
+    return {
+      JSXOpeningElement(node: TSESTree.JSXOpeningElement) {
+        if (node.name.type === AST_NODE_TYPES.JSXMemberExpression) return; // Property-like namespace syntax <Animated.View>
+        if (node.name.type === AST_NODE_TYPES.JSXNamespacedName) return; // XML-based namespace syntax: <Animated:View>
```
tjzel, line 8: Is this actually the case in React Native? 👀

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,155 @@
...
+              checkArrayNode(node);
+            } else if (node?.type === 'ObjectExpression') {
+              checkObjectNode(node);
+            }
+          });
+        }
+
+        if (
+          isVariableDefinedAs(componentName, 'Animated') ||
+          isVariableDefinedAs(componentName, 'createAnimatedComponent')
```
tjzel, line 93: does it work if an animated component is exported from another file, etc.?

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
tjzel, line 1: Could you rearrange this file to have top-down flow (that is, most important things on top of the file, helper functions etc. at the bottom). That way it's more readable in my opinion.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,155 @@
...
+            } else if (node?.type === 'ArrayExpression') {
+              checkArrayNode(node);
+            } else if (node?.type === 'ObjectExpression') {
+              checkObjectNode(node);
+            }
+          });
+        }
+
+        if (
+          isVariableDefinedAs(componentName, 'Animated') ||
```
tjzel, line 92: People tend to do:

`import Animated as Reanimated` or so, especially in React Native .Animated & Reanimated brownfields. Will it work in this case?

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,155 @@
...
+
+          variableNameTokenIds.forEach((idx) => {
+            if (tokensBefore[idx + 2].value === expectedToken) {
+              isAnimated = true;
+            }
+          });
+          return isAnimated;
+        }
+
+        function checkIdentifierNode(styleExpression: TSESTree.Identifier) {
```
tjzel, line 35: check for what? we need better names here

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,155 @@
...
+          const variableNameTokenIds: Array<number> = [];
+
+          tokensBefore.forEach((token, idx) => {
+            if (token.value === variableName) {
+              variableNameTokenIds.push(idx);
+            }
+          });
+
+          variableNameTokenIds.forEach((idx) => {
+            if (tokensBefore[idx + 2].value === expectedToken) {
```
tjzel, line 28: this seems to be a magical +2

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,155 @@
...
+          if (isAnimatedStyle) {
+            context.report({
+              node,
+              messageId: 'animatedStyle',
+              data: { componentName, variableName },
+            });
+          }
+        }
+
+        function checkObjectNode(styleExpression: TSESTree.ObjectExpression) {
```
tjzel, line 52: check for what?

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,155 @@
...
+                    propertyValue: variableName,
+                    componentName,
+                  },
+                });
+              }
+            }
+          });
+        }
+
+        function checkArrayNode(styleExpression: TSESTree.ArrayExpression) {
```
tjzel, line 78: check for what?

tjzel (changes requested):
I will need you to explain bits of this code to me IRL to make sure I understand them correctly.

__tests__/eslintPlugin/noAnimatedStyleToNonAnimatedComponent.test.tsx
Latropos: Sure!
I've found that using string literals to create new code snippets is common: [prefer-nullish-coalescing](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts)
They even use `.map` to create more tests.
However formats of eslint rule tests are very inconsistent, for example here [adjacent-overload-signatures](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/adjacent-overload-signatures.test.ts) each example has its code and error is hardcoded in place and here [member-ordering-alphabetically-order](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/member-ordering/member-ordering-alphabetically-order.test.ts) all test cases are grouped in an array and run in the end.
Also In the first linked example the error location is not tested and in the remaining two examples it is. I don't test my rule for error location too, it doesn't seem to be a thing that could break in my rule.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
Latropos: It's not, React components cannot include JSX namespace names.
Yet, I want to keep this condition to avoid unncecessary risk.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
Latropos: No, it doesn't work. Unfortunately in such a case you just have to disable this rule (or give your component name starting with "(Rea/A)nimated)"

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
Latropos: I've added a comment.

eslintPlugin/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "allowSyntheticDefaultImports": true,
+    "moduleResolution": "NodeNext",
+    "module":"NodeNext",
+    "strict": true,
+    "forceConsistentCasingInFileNames": true,
+    "resolveJsonModule": true,
+    "outDir": "./public",
+    "rootDir": "./src"
+  },
+}
```
tjzel, line 16: newline still missing here

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
tjzel: Add a comment about it here then please.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        const sourceCode = context.getSourceCode();
+        const tokensBefore = sourceCode.getTokensBefore(node);
+        const componentName = node?.name?.name;
+
+        main();
+        function main() {
+          if (
+            isVariableDefinedAs(componentName, 'Animated') || //TODO allow to customize rule and provide custom import name
+            isVariableDefinedAs(componentName, 'Reanimated') ||
```
tjzel, line 18:
```suggestion
// People tend to import `Animated` as `Reanimated`.
            isVariableDefinedAs(componentName, 'Reanimated') ||
```

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,175 @@
...
+        if (node.name.type === AST_NODE_TYPES.JSXNamespacedName) return; // XML-based namespace syntax: <Animated:View>
+
+        const sourceCode = context.getSourceCode();
+        const tokensBefore = sourceCode.getTokensBefore(node);
+        const componentName = node?.name?.name;
+
+        main();
+        function main() {
+          if (
+            isVariableDefinedAs(componentName, 'Animated') || //TODO allow to customize rule and provide custom import name
```
tjzel, line 17: This seems to also be problematic. One of many common cases here is accidentaly using `Animated` from React Native instead of Reanimated. Therefore, the plugin would miss a lot of its functionality here.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,175 @@
...
+            return;
+          }
+          const styleAttribute = node.attributes
+            .map((attribute) => {
+              return attribute.type === AST_NODE_TYPES.JSXAttribute &&
+                attribute.name.name === 'style'
+                ? [attribute]
+                : [];
+            })
+            .flat();
```
tjzel, line 30: How about using `.find` method here instead of `.map`?

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,175 @@
...
+            case AST_NODE_TYPES.ArrayExpression: // style={[style1, style2]}
+              checkArrayNodeForBeingAnimated(styleExpression);
+              break;
+            case AST_NODE_TYPES.ObjectExpression: //style={{backgroundColor:'pink'}}
+              checkObjectNodeForBeingAnimated(styleExpression);
+              break;
+            case AST_NODE_TYPES.MemberExpression: //style={{backgroundColor:styles.myStyle}}
+              //We assume that all member expressions are correct
+              break;
+          }
```
tjzel, line 62: How about moving it to a separate function?

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
tjzel: I like the comment a lot, but it seems to be misleading, since `idx+2` is `=` there. I assume that `a` is actually `idx` there. You should make that table to refer to the actual situation here - meaning that `const` should be `idx - 1`.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,175 @@
...
+          });
+        }
+      },
+    };
+  },
+  meta: {
+    docs: {
+      recommended: 'recommended',
+      description: 'Avoid looping over enums.',
+    },
```
tjzel, line 162: What's this?

tjzel (changes requested):

tjzel:
If you're adding this and it's supposed to be a part of our public API you should also acknowledge it in the docs.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
Latropos: Oh, this is actually a trick here ;-)
I've tried with `.find` first and for some reason it require you to make a type case:
image
But with map you get the correct type
image

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
Latropos: Yes, you're right. We could parse imports to detect actual import name. However this improvement could be done in the future, this PR is already big.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
Latropos: I am not really convinced if its worth moving. We could split the whole `main` function into smaller functions instead.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
Latropos: Indeed I was wondering whether to start with -1 or 0

Latropos:
@tjzel Yes, but we probably should publish it before adding documentation and we also shouldn't make this PR too big.

tjzel:
Why though? If it's tested sufficiently then I see no reason to add the docs immediately. Also change in the docs is independent of the code you already made, that's not really bloating this PR.

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
tjzel: This is what type assertions are for. 

You shouldn't use `map` and `flat` just to get a type _you want_. You should use `find` and type assertion if you know better than TypeScript.

What you are doing at the moment is an actual type assertion but with unnecessary runtime decisions and I believe this is a bad pattern.

Latropos:
@tjzel 
This plugin isn't going to be published with reanimated yet 🤫

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
Latropos: @tjzel Yes, you're right, in this case I really know the actual type.

tjzel:
Ah, then nvm my rant 😅 🤯

__tests__/eslintPlugin/noAnimatedStyleToNonAnimatedComponent.test.tsx
tjzel: We should also include all the current shortcomings of this plugin in the tests as well, so we have their existence well documented. Let's add a test that
- gives a false positive when an animated component is passed as a parameter
- gives a false positive when `Animated` is imported as e.g. `ReaAnimated`
- gives a false negative when `Animated` from react native is used
- gives a false negative when animated style is passed as a parameter
- gives a false negative when shared value is passed as a parameter

and some more if you think about some most basic examples (we can substitue importing from file with parameters imo)

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,176 @@
...
+        const sourceCode = context.getSourceCode();
+        const tokensBefore = sourceCode.getTokensBefore(node);
+        const componentName = node?.name?.name;
+
+        main();
+        function main() {
+          if (
+            isVariableDefinedAs(componentName, 'Animated') ||
+            // People tend to import `Animated` as `Reanimated`.
+            //TODO parse imports to detect actual import name
```
tjzel, line 19:
```suggestion
            // TODO parse imports to detect actual import name
```

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,176 @@
...
+
+          const styleExpression = styleValue.expression;
+          switch (styleExpression.type) {
+            case AST_NODE_TYPES.Identifier: // style={myVariable}
+              checkIdentifierNodeForBeingAnimated(styleExpression);
+              break;
+            case AST_NODE_TYPES.ArrayExpression: // style={[style1, style2]}
+              checkArrayNodeForBeingAnimated(styleExpression);
+              break;
+            case AST_NODE_TYPES.ObjectExpression: //style={{backgroundColor:'pink'}}
```
tjzel, line 56:
```suggestion
            case AST_NODE_TYPES.ObjectExpression: // style={{backgroundColor:'pink'}}
```

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,176 @@
...
+              checkObjectNodeForBeingAnimated(styleExpression);
+              break;
+            case AST_NODE_TYPES.MemberExpression: //style={{backgroundColor:styles.myStyle}}
+              //We assume that all member expressions are correct
```
tjzel, line 60:
```suggestion
              // We assume that all member expressions are correct
```

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,176 @@
...
+              ├───────────┼───────┼───────┼───────┼─────────────┤
+              │ Token     │ idx-1 │ idx   │ idx+1 │ idx+2       │
+              ╰───────────┴───────┴───────┴───────┴─────────────╯
+             */
+            if (tokensBefore[idx + 2].value === expectedToken) {
+              isAnimated = true;
+            }
+          });
+          return isAnimated;
+        }
```
tjzel, line 92:
```suggestion
        function isVariableDefinedAs(
          variableName: string,
          expectedToken: string
        ) {
          const variableNameTokenIds: Array<number> = [];

          tokensBefore.forEach((token, idx) => {
            if (token.value === variableName) {
              variableNameTokenIds.push(idx);
            }
          });

          return variableNameTokenIds.some(
            (idx) =>
              /* 
              Lets count tokens from variable name to its definition:
              ╭───────────┬───────┬───────┬───────┬────────────────╮
              │ Code      │ const │  sv   │   =   │ useSharedValue │
              ├───────────┼───────┼───────┼───────┼────────────────┤
              │ Token     │ idx-1 │ idx   │ idx+1 │      idx+2     │
              ╰───────────┴───────┴───────┴───────┴────────────────╯
             */
              tokensBefore[idx + 2].value === expectedToken
          );
        }
```

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,176 @@
...
+  },
+  meta: {
+    docs: {
+      recommended: 'recommended',
+      description:
+        "Don't pass a reanimated animated style into a non-animated component.",
+    },
+    messages: {
+      sharedValue:
+        "Property  '{{propertyName}}: {{propertyValue}}' is using a sharedValue '{{propertyValue}}', but was used in a default component. Replace your {{componentName}} with an Animated.{{componentName}}",
```
tjzel, line 166: Given component doesn't have to be in `Animated` namespace and we shouldn't be that specific here.
```suggestion
        "Property  '{{propertyName}}: {{propertyValue}}' is using a shared value '{{propertyValue}}', but was used in a default component. Replace {{componentName}} with an animated component from Reanimated.",
```

eslintPlugin/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -0,0 +1,176 @@
...
+      animatedStyle:
+        "Style '{{variableName}}' is an animated style, but was used in a default component. Replace your '{{componentName}}' with an Animated.{{componentName}}",
+    },
```
tjzel, line 169:
```suggestion
      animatedStyle:
        "Style '{{variableName}}' is an animated style, but was used in a default component. Replace your '{{componentName}}' with an animated component from Reanimated.",
    },
```

tjzel (changes requested):
Add the tests and correct some minor things and we should be good to go!

__tests__/eslintPlugin/noAnimatedStyleToNonAnimatedComponent.test.tsx
Latropos: >gives a false positive when Animated is imported as e.g. ReaAnimated

Its a false negative, which IMO is better (if sb have custom import name he has to mind the errors on their own, but we won't spam them with errors)

__tests__/eslintPlugin/noAnimatedStyleToNonAnimatedComponent.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+    componentName: 'View',
+    propertyName: 'width',
+    propertyValue: 'width',
+  },
+} as const;
+
+/**
+ * Correct code correctly classified as satisfying the rule
+ */
+const trueNegative = [
```
tjzel, line 52: I don't quite like the names of the tests now. Names of the test cases should mainly say what they are testing, not what is their outcome. It should something among the lines of

- non-animated component with non-animated style
- non-animated component with animated style
- non-animated component with shared value prop
- animated component with non-animated style
- etc.

Latropos: Unfortunately we don't have names of test cases when testing the eslint rules, the way we have in normal jest tests. Instead, we have the code snippet injected in its place.
I suppose you don't mean "names of the tests" but names of the variables storing the tests and additional description in comments. Then I understand that your suggestions have many benefits, but I would keep it the way it is for the following reasons:

* Variable names don't contain spaces, so `nonAnimatedComponentWithNonAnimatedStyle` maybe more descriptive, but `nonAnimatedComponentWithNonAnimatedStyle` and `nonAnimatedComponentWithAnimatedStyle` look similar and shorter names like `truePositive` and `trueNegative` are easier to work with
* It's intuitive that `trueNegative` are cases, that should be detected as breaking the rule, but they ere not
* The whole code snippet is acting like a test name, so it is easy to read that we are dealing, for example, with "non-animated component with animated style"

__tests__/eslintPlugin/noAnimatedStyleToNonAnimatedComponent.test.tsx
Latropos: No, the first line in my rule is
```jsx
    if (node.name.type === AST_NODE_TYPES.JSXMemberExpression) {
      return; // Property-like namespace syntax <Animated.View>
    }
```
So we assume that we never report error for any member expression (like <ReaAnimaterd.View style={animatedStyle/> )

Latropos: As I've typed above
> Its a false negative, which IMO is better (if sb have custom import name he has to mind the errors on their own, but we won't spam them with errors)

And I really believe that its a good solution

__tests__/eslintPlugin/noAnimatedStyleToNonAnimatedComponent.test.tsx
tjzel: I'm considering DX here.

Information that something broke in `falsePositive` doesn't really tell you much, you need to dive into it to get what exactly is failing.

On the other hand, if I want to look up something, for e.g. "what should I expect when I pass `animatedStyle` to a `ReaAnimated` component" I must look through all the test suites since I don't know if this case is even handled in the tests.

If you want to keep the `falseNegative` logic etc. at least put those tests in some domains. E.g.:

```ts
const withAnimatedComponent = {
  withAnimatedStyle: {
    truePositive: [...]
    falsePositive: [...]
    ...
  },
  withSharedValue: {...},
}

const withNonAnimatedComponent = {
  ...
};

__tests__/eslintPlugin/noAnimatedStyleToNonAnimatedComponent.test.tsx
tjzel: > So we assume that we never report error for any member expression (like <ReaAnimaterd.View style={animatedStyle/> )

Oh I see, I forgot about it. That's not too great however, since using `Animated` from `react-native` is a very common mistake. We can leave it as is for now, but definitely need to improve it in the future.

Latropos:
@tjzel 
I've changed test order of the tests. Here is also the screenshot of how the testing look in terminal:

image

PR #5560, opened 2024-01-05
Fix `useAnimatedSensor` usage in docs and examples

tjzel:
## Summary

Currently `useAnimatedSensor` returns a stable ref which is also modified inside of `useAnimatedSensor`. This leads to issues when a user passes the whole return object of `useAnimatedSensor` to a worklet - the whole object gets frozen in `makeShareableCloneRecursive` and the hook stops working properly on re-render.

Currently in our documentation and examples we use it in this exact way - passing the whole object to a worklet. This was unnoticed until now since we never did any additional re-renders there.

There are two ways out of this problem.

1. Amend documentation and examples to show proper usage. Obviously it won't reach users fast. (Unless they get warnings from #5548).

2. Make it so `useAnimatedSensor` doesn't return a stable ref. I don't think we ever claimed it to be stable, but nevertheless this is potentially a breaking change since someone might rely on it being stable.

I prefer the first option since the latter is a breaking change but I'd like to hear you out.

## Test plan

🚀

### DIFF
--- a/app/src/examples/AnimatedSensorAccelerometerExample.tsx
+++ b/app/src/examples/AnimatedSensorAccelerometerExample.tsx
@@ -13,14 +13,14 @@ const OFFSET_X = 100; // in px
 const OFFSET_Y = 100; // in px
 
 export default function AnimatedSensorAccelerometerExample() {
-  const accelerometer = useAnimatedSensor(SensorType.ACCELEROMETER);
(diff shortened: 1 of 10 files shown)


tomekzaw:
> Currently `useAnimatedSensor` returns a stable ref which is also modified inside of `useAnimatedSensor`. 

Where exactly do we modify the ref? Would it be possible to simply avoid such modification?

tjzel:
Closing in favor of #5646.

Issue #5561, opened 2024-01-06
Web layout animation fails to load on production mode


linhvovan29546:
## Expected behavior: 
- The layout animation should work correct on production mode.
## Actual behavior:
- The layout animation is not working on production mode. 
-  The initAnimationName is set to `c` instead of the correct value.

<img width="808" alt="Screen Shot 2024-01-06 at 4 38 07 PM" src="https://github.com/software-mansion/react-native-reanimated/assets/84692183/85ed9251-8bd6-4dd2-bbff-9ff6d085d966">
<img width="883" alt="image" src="https://github.com/software-mansion/react-native-reanimated/assets/84692183/6283bbc1-1463-4ec4-847a-6c81f69dedd7">

### Steps to reproduce

1. Build the application with the production mode `npx expo export -p web`
2. Deploy app (netlify,...)
3. Open the application on a web browser and navigate to the page with the layout animation.
5. See result

### Snack or a link to a repository

https://snack.expo.dev/@vovanlinh29546/react-native-web-reanimated-layout-animation

### Reanimated version

3.6.1

### React Native version

0.72.3

### Platforms

Web

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Paper

### Build type

Release app & production bundle

### Device

None

### Device model

None

### Acknowledgements

Yes

linhvovan29546:
My solution is to create a className for the animation class https://github.com/software-mansion/react-native-reanimated/blob/4ad3a517d80f1680f9d9c8897f2211358bd142a6/src/reanimated2/layoutReanimation/defaultAnimations/Slide.ts#L335C23-L335C23
```
export class SlideOutDown extends ComplexAnimationBuilder implements IExitAnimationBuilder {
  className: string = 'SlideOutDown'; // Adding className property

  // ... existing code ...

}
```
<img width="809" alt="Screen Shot 2024-01-06 at 5 04 30 PM" src="https://github.com/software-mansion/react-native-reanimated/assets/84692183/112ad4ce-29ca-4c62-849c-f6d02579657c">

tomekzaw:
Thanks for reporting this issue! cc @m-bert

m-bert:
Hi @linhvovan29546! Could you please check if [this PR](https://github.com/software-mansion/react-native-reanimated/pull/5562) fixes this issue?

m-bert:
[...]

PR #5562, opened 2024-01-08
[LA Web] Fix animations in production bundle

m-bert:
## Summary

As pointed in #5561, layout animations don't work with production bundle. It is caused by fragile method of obtaining animation name, i.e. using `config.name` if config is function. During bundling, name of this function may change (in linked issue to `c`, in my case to `r`), so instead of animation user will see warning that reanimated couldn't load animation. 

This PR adds static field `className` to animations builders in order to provide access to animation name, even if config name gets changed during bundling.  

## Test plan

1.  Create new app (`npx create-expo-app`)
2. Install Reanimated from package (`npm pack` -> `yarn add <path_to_package>`) 
3. `npx expo export -p web`
4. `npx serve dist --single`

I used reanimated from local package to make sure that `expo export` works correctly.

<details>
<summary> Tested on the following code </summary>

```jsx
import Animated, { FadeOut, FadeOutRight } from "react-native-reanimated";
import { View, Button } from "react-native";
import { useState } from "react";

  return (
    <View
      style={{
        flex: 1,
        alignItems: "center",
        justifyContent: "center",
        flexDirection: "column",
      }}
    >
      {show && (
        <Animated.View
          style={[{ width: 100, height: 80, backgroundColor: "black" }]}
          exiting={FadeOut}
        />
      )}

      <Button
        title="toggle"
        onPress={() => {
          setShow((prev) => !prev);
        }}
      />
    </View>
  );
}

```

</details>

### DIFF
--- a/src/reanimated2/layoutReanimation/web/animationsManager.ts
+++ b/src/reanimated2/layoutReanimation/web/animationsManager.ts
@@ -89,13 +89,15 @@ function tryGetAnimationConfigWithTransform<
     return null;
   }
 
+  type ConstructorWithStaticContext = {
+    className: string;
+  } & typeof config.constructor;
+
   const isLayoutTransition = animationType === LayoutAnimationType.LAYOUT;
   const initialAnimationName =
     typeof config === 'function'
-      ? // @ts-expect-error ignore for now
-        config.className
-      : // @ts-expect-error ignore for now
-        config.constructor.className;
+      ? config.className
+      : (config.constructor as ConstructorWithStaticContext).className;
 
   const shouldFail = checkUndefinedAnimationFail(
     initialAnimationName,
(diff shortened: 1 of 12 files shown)


src/reanimated2/layoutReanimation/web/animationsManager.ts
```diff
@@ -89,9 +89,15 @@ function tryGetAnimationConfigWithTransform<
     return null;
   }
 
+  type ConstructorWithStaticContext = {
```
tjzel, line 92: This should use imported `LayoutAnimationStaticContext`.

tjzel (changes requested):
As discussed, let's change `className` to something more relevant like `presetName`.

src/reanimated2/layoutReanimation/web/animationsManager.ts
m-bert: Both done in [fd57b63](https://github.com/software-mansion/react-native-reanimated/pull/5562/commits/fd57b63bbddc9d9a55fe555a828f6040b3814072).

tjzel (approved):
🚢 🇮🇹

PR #5565, opened 2024-01-09
Pass down generator boolean in babel plugin

wcandillon:
fixes #5564

PR #5577, opened 2024-01-12
Properly transpile async functions

tjzel:
## Summary

Inspired by #5565. Currently we don't respect if the workletized function is an async one. This PR fixes this.

Keep in mind that even with these change **async worklets will not work properly on the UI thread**. This is because Hermes doesn't support them out of the box and they need to be transformed into generators first (you can see that in the snapshots).

Properly transforming them so they are valid on UI thread should be considered a future, very low priority task, since it wasn't requested yet by our users.

## Test plan

Relevant tests were added to `plugin.test.ts`.

<details>
<summary>
To see the difference between JS and UI behavior
</summary>

```tsx

import React from 'react';

  async function asyncWorklet() {
    'worklet';
    await Promise.resolve();
    console.log('Hello world!');
  }

  const runAsyncWorklet = runOnUI(asyncWorklet);

  return (
      <Button title="Run async worklet on UI" onPress={runAsyncWorklet} />
      <Button title="Run async worklet on JS" onPress={asyncWorklet} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

</details>

### DIFF
--- a/__tests__/plugin.test.ts
+++ b/__tests__/plugin.test.ts
@@ -1722,4 +1722,34 @@ describe('babel plugin', () => {
       expect(code).toMatchSnapshot();
     });
   });
+
+  describe('for async functions', () => {
+    it('makes an async worklet factory', () => {
+      const input = html`<script>
+        async function foo() {
+          'worklet';
+          await Promise.resolve();
+        }
+      </script>`;
+
+      const { code } = runPlugin(input);
+      expect(code).toContain('asyncToGenerator');
+      expect(code).toMatchSnapshot();
+    });
+
+    it('makes an async worklet string', () => {
+      const input = html`<script>
+        async function foo() {
+          'worklet';
+          await Promise.resolve();
+        }
+      </script>`;
+
+      const { code } = runPlugin(input);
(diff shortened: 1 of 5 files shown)


piaskowyk (approved):

__tests__/plugin.test.ts
```diff
@@ -1722,4 +1722,34 @@ describe('babel plugin', () => {
...
+      expect(code).toMatchSnapshot();
+    });
+
+    it('makes an async worklet string', () => {
+      const input = html`<script>
+        async function foo() {
+          'worklet';
+          await Promise.resolve();
+        }
+      </script>`;
```
tomekzaw, line 1746: This is the same input code as in the previous test. We can merge it into a single unit test (so we don't have two identical snapshots) or leave it as it is now for better granularity.

tomekzaw (approved):
Thanks for this PR!

__tests__/plugin.test.ts
tjzel: I like granular

PR #5586, opened 2024-01-15
Fix `useHandler` context issue

bartlomiejbloniarz:
## Summary

This PR fixes a regression introduced in #5518. `useHandler` works by maintaining a `context` object that can be accessed and modified in event handlers (e.g. within `useAnimatedGestureHandler`). Removing `makeRemote` call there made it so that any changes to the `context` object were not visible in event handlers, rendering the `context` object useless. In this PR I wrapped the context object with `makeShareable`, which works basically in the same way as `makeRemote` did.

## Test plan

Check whether the `[SET] Modals` example behaves properly (i.e. if the modal closing gestures are captured correctly).

### DIFF
--- a/src/reanimated2/hook/useHandler.ts
+++ b/src/reanimated2/hook/useHandler.ts
@@ -4,6 +4,7 @@ import type { WorkletFunction } from '../commonTypes';
 import { isWeb, isJest } from '../PlatformChecker';
 import type { DependencyList, ReanimatedEvent } from './commonTypes';
 import { areDependenciesEqual, buildDependencies } from './utils';
+import { makeShareable } from '../shareables';
 
 interface GeneralHandler<
   Event extends object,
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):
LGTM, let's wait for @tjzel's approval

piaskowyk (approved):

tjzel:
Thanks for this PR! Indeed that was a big oversight by me when removing `makeRemote`. I was wondering why is it possible to modify this value if it was made in `makeRemote`. It's because the following line in `valueUnpacker`:

Screenshot 2024-01-16 at 14 27 33

ShareableHandles are memoized on the UI thread so that's why the changes persist. Because of that I think that all changes in my PR #5518 should be revised to see if it didn't introduce any more regressions.

tjzel (approved):

piaskowyk:
I'll merge it because I need this changes 🙏

PR #5592, opened 2024-01-17
Pass shared_ptr by reference

piaskowyk:
## Summary

The aim of this PR is to replace passing shared pointer by value with reference passing. In addition I marked some fields as `const`.

## Test plan

See the CI


tomekzaw (approved):
👏

PR #5594, opened 2024-01-17
Update typescript-eslint

Latropos:
## Summary
* We were using @typescript-eslint/X in version 5.x, but we should use 6.x
* Now we can use type-checked rules:
  * I've turned some of them off:
    * '@typescript-eslint/no-unsafe-call': 'off',
    * '@typescript-eslint/no-unsafe-member-access': 'off',
    * '@typescript-eslint/no-unsafe-return': 'off',
    * '@typescript-eslint/no-unsafe-assignment': 'off',
    * '@typescript-eslint/no-unsafe-argument': 'off',
  * fixed the others:
    * Don't use objects in string literals
    * Fix shorthand in https://github.com/software-mansion/react-native-reanimated/pull/5595
 * I didn't turn type-checked rules for plugin. It would require fixing the configuration, but our eslint configuration should be migrated to new format anyway. 
* In the future I want to migrate `eslintrc` config files and to turn object shorthand on


src/reanimated2/js-reanimated/index.ts
```diff
@@ -112,8 +113,10 @@ export const _updatePropsJS = (
         component._touchableNode.setAttribute(dashedKey, rawStyles[key]);
       });
     } else {
+      const componentName =
+        'className' in component ? component?.className : '';
       console.warn(
-        `[Reanimated] It's not possible to manipulate the component ${component}`
+        `[Reanimated] It's not possible to manipulate the component ${componentName}`
```
tjzel, line 119: Is it actually what we want here?

src/reanimated2/platformFunctions/measure.ts
tjzel, line 1: I'm not sure if we want these changes in this file. `viewTag` should be a `number` on Paper and I don't know if `ShadowNodeWrapper` has `toString()` method. Could you check that? Maybe we should also split `AnimatedRef` type internally so it would specify it's platform (e.g. AnimatedRefOnJSPaper).

src/reanimated2/UpdateLayoutAnimations.ts
```diff
@@ -17,6 +17,7 @@ function createUpdateManager() {
     update(batchItem: LayoutAnimationBatchItem) {
       animations.push(batchItem);
       if (animations.length === 1) {
+        // eslint-disable-next-line @typescript-eslint/unbound-method
```
tjzel, line 20: Please change the signature of flush to
`flush(this: void)` instead.

src/reanimated2/UpdateProps.ts
```diff
@@ -72,6 +73,7 @@ const createUpdatePropsManager = isFabric()
               updates,
             });
             if (operations.length === 1) {
+              // eslint-disable-next-line @typescript-eslint/unbound-method
```
tjzel, line 76: Please change the signature of flush to
`flush(this: void)` instead.

src/reanimated2/UpdateProps.ts
```diff
@@ -102,6 +104,7 @@ const createUpdatePropsManager = isFabric()
...
```
tjzel, line 107: Please change the signature of flush to
`flush(this: void)` instead.

src/reanimated2/initializers.ts
```diff
@@ -52,12 +52,14 @@ export function setupConsole() {
   if (!IS_CHROME_DEBUGGER) {
     // @ts-ignore TypeScript doesn't like that there are missing methods in console object, but we don't provide all the methods for the UI runtime console version
     global.console = {
+      /* eslint-disable @typescript-eslint/unbound-method */
```
tjzel, line 55: This is actually fine here 👍

tjzel (changes requested):
I have a general question regarding those changes - if I remove some of those `eslint-disable` comments I'm not getting any errors in my VSCode, only when I manually run the lint from CLI. Is this the same case for you? Or maybe there's some method for VSCode to acknowledge those rules?

Also this PR could use some polishing, but seems good overall after we tackle the things I pointed out.

src/reanimated2/core.ts
```diff
@@ -62,6 +62,7 @@ export function getViewProp<T>(viewTag: number, propName: string): Promise<T> {
     );
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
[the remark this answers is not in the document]
Latropos: Here is the explanation of the rule: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-misused-promises.md

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -256,7 +256,9 @@ function assertVectorsHaveEqualLengths(a: number[], b: number[]) {
   'worklet';
   if (__DEV__ && a.length !== b.length) {
     throw new Error(
-      `[Reanimated] Cannot calculate inner product of two vectors of different lengths. Length of ${a} is ${a.length} and length of ${b} is ${b.length}.`
+      `[Reanimated] Cannot calculate inner product of two vectors of different lengths. Length of ${a.toString()} is ${
+        a.length
+      } and length of ${b.toString()} is ${b.length}.`
```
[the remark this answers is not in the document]
Latropos: Since here it looks better to keep these numbers in "[ ]", but in animationBuilder.tsx it doesn't

src/reanimated2/js-reanimated/index.ts
Latropos: Well, I've tested the following code:
```tsx
          const object = { a: 6, b: 7 };
          console.log(`1. I've got an object: ${object}`);
          console.log(`2. I've got an object: ${object.toString()}`);
```

It prints:  
1. I've got an object: [object Object]
2. I've got an object: [object Object]

I can't reproduce this particular error (it should be unaccessible), but we definitely don't want to print "[object Object]" and className is the best we can extract from the component.

src/reanimated2/animation/transformationMatrix/matrixUtils.tsx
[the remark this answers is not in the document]
Latropos: I've just tested that `.toString` an `",".join` work the same way. However I'll add extra space when concatenating strings. Numbers don't look as bad when concatenated without string.

No spaces: "This","looks","very","bad"
With spaces: "This", "looks", "much", "better"

No spaces: 1,2,6,34,1
With spaces: 1, 2, 6, 34, 1

src/reanimated2/platformFunctions/measure.ts
Latropos: Shadow node wrapper doesn't have toString() method. Throwing any of this warnings on Fabric were causing additional error (Cannot determine default value of object). I'll refactor this to display view tag only on paper (since there are no view tags on fabric anyway)

src/reanimated2/core.ts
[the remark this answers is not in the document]
tjzel: I read this but still didn't understand 😆

src/reanimated2/js-reanimated/index.ts
tjzel: Yeah but in web it might be different. Just try logging this string in the beginning of this function and see if it makes sense.

src/reanimated2/core.ts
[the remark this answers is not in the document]
Latropos: In this particular case you return promise from inside of promise (getViewProp is a promise) which is considered a bad pattern

src/reanimated2/js-reanimated/index.ts
Latropos: Tested on web too, still the same log

src/reanimated2/core.ts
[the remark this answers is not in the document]
tjzel: Took me some time but I finally understood this rule!

tjzel (approved):

PR #5599, opened 2024-01-19
Fix plugin sometimes not workletizing `ObjectMethod`

### LINKED ISSUE #5555: Worklets nested within objects throw error on update to RN 73, Reanimated 3.6.1
### Description

This may be an RNGH issue and not Reanimated, but posting here since the error is a reanimated error. I have a pan gesture that must register an `onTouchesMove` callback, but cannot do it inline and take advantage of the auto-workletizing that is documented here: https://docs.swmansion.com/react-native-gesture-handler/docs/#automatic-workletization-of-gesture-callbacks 

So, I am explicitly marking this method as `"worklet";`. In expo 49/react-native 0.72.3 and Reanimated 3.5.4 and RNGH 2.13.4 , the following code worked:

```ts
function Demo() {

  const panGesture = Gesture.Pan()
  
  panGesture.onTouchesMove((evt, mgr) => {
    'worklet'
    console.log('move!!')
  })

  return (
    <GestureHandlerRootView style={{ flex: 1, backgroundColor: 'seashell' }}>
      <GestureDetector gesture={panGesture}>
        <Animated.View style={{ flex: 1 }} />
      </GestureDetector>
    </GestureHandlerRootView>
  )
}
```

After upgrading to expo 50 and Reanimated 3.6.1 and RNGH 2.14.0 I see this error on pan:

```
Error: ENOENT: no such file or directory, open '/Users/REDACTED/installValueUnpacker'
    at Object.openSync (node:fs:601:3)
    at Object.readFileSync (node:fs:469:35)
    at getCodeFrame (/Users/REDACTED/node_modules/metro/src/Server.js:927:18)
    at Server._symbolicate (/Users/REDACTED/node_modules/metro/src/Server.js:1010:22)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Server._processRequest (/Users/REDACTEDnode_modules/metro/src/Server.js:422:7) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/Users/REDACTED/installValueUnpacker'
}
 ERROR  ReanimatedError: [Reanimated] Tried to synchronously call a non-worklet function on the UI thread.
See `https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooting#tried-to-synchronously-call-a-non-worklet-function-on-the-ui-thread` for more details., js engine: reanimated

```

Note that only the `onTouchesX` callbacks trigger the error -- the workletized `onBegin` `onEnd` do not.
(something to do with handlers that set `config.needsPointerData = true`?)

I have tried resetting metro cache and `watchman watch-del-all`

Possibly related to https://github.com/software-mansion/react-native-reanimated/issues/5541

### Steps to reproduce

Run the above demo on: 
- expo 50.0.0-preview.7
- react-native 0.73.1
- RNGH 2.14.0

n/a, reanimated version not yet supported 

### Reanimated version

3.6.1

### React Native version

0.73.1

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

tjzel:
## Summary

Fixes #5555.

Currently, when transforming functions with a `'worklet'` directive we are only looking for:

1. FunctionDeclaration
```ts
function foo(){'worklet'}
```

2. FunctionExpression
```ts
const foo = function (){'worklet'}}
```

3. ArrowFunctionExpression
```ts
const foo = () => {'worklet'};
```

And not taking into consideration ObjectMethod:

4.
```ts
const foo = {
  bar() {'worklet'}
}
```

This PR fixes that and does a slight refactor of the upper control flow and renames some functions.

This wasn't an issue before because plugins used in Expo and React Native used to transform ObjectMethods to ObjectProperties:

```ts
const foo {
  bar: function() {'worklet'}
}
```

But it's no longer always the case.

## Test plan

Added a step to the CI if our App is actually bundling properly and added a test.

## Note

I'm considering changing the names of relevant files afterwards, but in this PR it would just be too messy I think.

### DIFF
--- a/.github/workflows/validate-plugin.yml
+++ b/.github/workflows/validate-plugin.yml
@@ -52,4 +52,4 @@ jobs:
 
       - name: Check Example App bundling
         working-directory: Example
-        run: yarn && yarn react-native bundle --reset-cache --entry-file='App.tsx' --dev=true --platform='ios' > /dev/null
+        run: yarn && yarn react-native bundle --reset-cache --entry-file='App.tsx' --bundle-output='/dev/null' --dev=true --platform='ios'
--- a/.github/workflows/validate-plugin.yml
+++ b/.github/workflows/validate-plugin.yml
@@ -49,3 +49,7 @@ jobs:
         run: git diff
       - name: Test
         run: yarn jest plugin
+
+      - name: Check Example App bundling
+        working-directory: Example
+        run: yarn && yarn react-native bundle --reset-cache --entry-file='App.tsx' --dev=true --platform='ios' > /dev/null
--- a/__tests__/__snapshots__/plugin.test.ts.snap
+++ b/__tests__/__snapshots__/plugin.test.ts.snap
@@ -1826,20 +1826,20 @@ var foo = function () {
 `;
 
 exports[`babel plugin for worklet nesting transpiles nested worklets embedded in runOnJS in runOnUI 1`] = `
-"var _worklet_16739363980177_init_data = {
-  code: "function anonymous(){const{runOnJS}=this.__closure;console.log('Hello from UI thread');runOnJS(function(){'worklet';console.log('Hello from JS thread');})();}",
-  location: "/dev/null",
-  sourceMap: "\\"mock source map\\"",
-  version: "x.y.z"
-};
-var _worklet_13427774179957_init_data = {
+"var _worklet_13427774179957_init_data = {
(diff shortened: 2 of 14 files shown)


bartlomiejbloniarz (approved):
🚀

PR #5606, opened 2024-01-23
Fix horse with reduced motion

Latropos:
## Summary

Our hook gets produces buggy behaviour on each page refresh, so I've replaced it with css media-query.

The reduced motion warning works correcrtly.


kacperkapusciak (approved):
It works! 🎉 

What about the clouds and the blinking stars?

Latropos:
@kacperkapusciak These elements already have slow and calm animations so its safe to keep them

PR #5609, opened 2024-01-25
Update CI deps

piaskowyk:
## Summary

This PR updated CI dependencies to get rid off following warnings:
Screenshot 2024-01-25 at 14 20 52

Contains update for:
- actions/checkout
- actions/setup-node
- actions/upload-artifact

I also removed node version specification where it isn't necessary.

## Test plan

See at CI output

### DIFF
--- a/.github/workflows/build-npm-package-action.yml
+++ b/.github/workflows/build-npm-package-action.yml
@@ -25,25 +25,27 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref }}
 
       - name: Setup Node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: 16.x
+          node-version: 20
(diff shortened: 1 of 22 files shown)


.github/workflows/build-npm-package-action.yml
```diff
@@ -25,25 +25,27 @@ jobs:
...
+        uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref }}
 
       - name: Setup Node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: 16.x
+          node-version: 20
```
tomekzaw, line 35: Do we need to pass `node-version` here? In other places we don't.

tomekzaw (approved):

.github/workflows/build-npm-package-action.yml
piaskowyk: I think that having more control over the crucial CI that builds the package is important to ensure a reproducible environment. However, in other CIs where node is simply used as a test script runner, its version becomes less important.

tomekzaw: Okay, this makes perfect sense to me, thanks!

PR #5615, opened 2024-01-30
Fix cpp build failing due to unused variable

DrRefactor:
## Summary

Building app with nightly (29-01-24) build fails due to unused 'supportsLocking_' variable in WorkletRuntime.h

Fixes https://github.com/software-mansion/react-native-reanimated/issues/5616

## Test plan

Create bare react native project or use mine https://github.com/DrRefactor/reanimated-issue-5614-repro
```
yarn add react-native-reanimated@nightly
yarn android --mode=release
```


piaskowyk (approved):
Thanks for the PR! 🙌

tomekzaw:
Not a huge fan of the fact that `supportsLocking_` field is present only in debug mode, the ABI should be the same for debug and release mode. Instead of this, can we just add `(void)supportsLocking_;` or [`[[maybe_unused]]` attribute](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused)?

DrRefactor:
@tomekzaw [[maybe_unused]] is how I approached this: https://github.com/software-mansion/react-native-reanimated/pull/5615/commits/402206ae7665aca92b8baea1fd07040a7ba33872

@piaskowyk could you take a look at @tomekzaw 's question?

PR #5617, opened 2024-01-31
Capture exiting element snapshot only right before hiding it to preve…

DrRefactor:
…nt enforcing (browser) layout on each render

## Summary

Current implementation of saving snapshot on each component update causes browser to re-layout elements with defined 'exiting' animations. This is caused by .getBoundingClientRect() called in each componentDidUpdate.

Moreover, component might not necessarily update, but its bounding client rect might have changed due to ancestor (of some level) layout change, that did not cause element to render (e.g. due to memo).

Getting component "snapshot" (boundingClientRect) right before hiding element seems to be enough, and has no significant performance overhead.

Motivation: performance optimization, removing unnecessary overhead.

## Test plan

I will provide repo for tests later.

### DIFF
--- a/src/reanimated2/layoutReanimation/web/componentUtils.ts
+++ b/src/reanimated2/layoutReanimation/web/componentUtils.ts
@@ -127,6 +127,10 @@ export function getProcessedConfig(
   };
 }
 
+export function saveSnapshot(element: HTMLElement) {
+  snapshots.set(element, element.getBoundingClientRect());
+}
+
 export function setElementAnimation(
   element: HTMLElement,
   animationConfig: AnimationConfig,
@@ -218,7 +222,7 @@ export function handleExitingAnimation(
   const dummy = element.cloneNode() as ReanimatedHTMLElement;
   dummy.reanimatedDummy = true;
 
-  const snapshot = element.getBoundingClientRect();
+  const snapshot = snapshots.get(element) ?? element.getBoundingClientRect();
 
   element.style.animationName = '';
   // We hide current element so only its copy with proper animation will be displayed
--- a/src/reanimated2/layoutReanimation/web/componentUtils.ts
+++ b/src/reanimated2/layoutReanimation/web/componentUtils.ts
@@ -127,10 +127,6 @@ export function getProcessedConfig(
   };
 }
 
-export function saveSnapshot(element: HTMLElement) {
-  snapshots.set(element, element.getBoundingClientRect());
-}
-
 export function setElementAnimation(
   element: HTMLElement,
   animationConfig: AnimationConfig,
@@ -222,6 +218,8 @@ export function handleExitingAnimation(
   const dummy = element.cloneNode() as ReanimatedHTMLElement;
   dummy.reanimatedDummy = true;
 
+  const snapshot = element.getBoundingClientRect();
+
   element.style.animationName = '';
   // We hide current element so only its copy with proper animation will be displayed
   element.style.visibility = 'hidden';
@@ -238,7 +236,6 @@ export function handleExitingAnimation(
   setElementAnimation(dummy, animationConfig);
   parent?.appendChild(dummy);
 
-  const snapshot = snapshots.get(element)!;
   snapshots.set(dummy, snapshot);
 
   setDummyPosition(dummy, snapshot);
--- a/src/createAnimatedComponent/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent/createAnimatedComponent.tsx
@@ -34,7 +34,7 @@ import type {
   IAnimatedComponentInternal,
   ViewInfo,
 } from './commonTypes';
-import { has, flattenArray } from './utils';
+import { has, flattenArray, disposableCallback } from './utils';
 import setAndForwardRef from './setAndForwardRef';
 import {
   isFabric,
@@ -49,6 +49,7 @@ import {
   tryActivateLayoutTransition,
   configureWebLayoutAnimations,
   getReducedMotionFromConfig,
+  saveSnapshot,
 } from '../reanimated2/layoutReanimation/web';
 import { updateLayoutAnimations } from '../reanimated2/UpdateLayoutAnimations';
 import type { CustomConfig } from '../reanimated2/layoutReanimation/web/config';
@@ -168,10 +169,15 @@ export function createAnimatedComponent(
       }
 
       if (IS_WEB) {
+        if (this.props.exiting) {
+          saveSnapshot(this._component as HTMLElement);
+        }
+
         if (
           !this.props.entering ||
           getReducedMotionFromConfig(this.props.entering as CustomConfig)
         ) {
+          this._observeLayout(this._component);
           this._isFirstRender = false;
           return;
         }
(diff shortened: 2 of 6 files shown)


src/reanimated2/layoutReanimation/web/componentUtils.ts
```diff
@@ -222,6 +218,8 @@ export function handleExitingAnimation(
   const dummy = element.cloneNode() as ReanimatedHTMLElement;
   dummy.reanimatedDummy = true;
 
+  const snapshot = element.getBoundingClientRect();
+
   element.style.animationName = '';
   // We hide current element so only its copy with proper animation will be displayed
   element.style.visibility = 'hidden';
```
DrRefactor, line 225: This is the important thing - we have to capture boundingClientRect before hiding element. After hiding element with visibility:hidden, browser won't calculate proper boundingClientRect. For some reason element hidden this way acts as a zero-size node.

DrRefactor:
@m-bert Please take a look.

tjzel:
@m-bert Notice me senpai

DrRefactor:
For some reason, I only see React's layout effects in Chrome debugger, but no browser native layouts, but I'm sure you'll get the idea from below.
My example renders Animated.View with `exiting` prop, with dummy 50ms interval that enforces Animated.View's parent to render. On below screenshots, you see a capture from Chrome performance tab of one interval tick render:

## main (nightly 30-01)
image


## this PR
image


Useful reference for things that cause relayouts:
https://gist.github.com/paulirish/5d52fb081b3570c81e3a

DrRefactor:
This is a repo with reproduction:
https://github.com/DrRefactor/reanimated-repro-PR-5617

1. clone above repo
2. `npm install`
3. `npm run web`
4. Chrome -> Performance -> start
5. save results
6. reinstall reanimated version from this PR and do 1-5 again
7. compare results

m-bert:
I've just looked at this PR and it seems that "Habits" example doesn't work. "Continue" buttton should stay at its place when clicked, but in this PR it jumps to the middle of the screen.

m-bert:
After further investigation I've found out what is the problem. `Dummy` gets reattached based on `snapshot`.  If you capture snapshot like in this PR, you don't take into account that `element` may have changed its layout based on other elements.

In case of "Habits" it would mean that `button` slides up because we remove habits list and then we capture snapshot, thus it ends up at wrong spot. Correct me if I'm wrong.

DrRefactor:
I'm looking into it, but it seems like it used to work "correctly" just by an accident. It just happened to be lucky race outcome for continue button to last render before unmounting other elements, therefore the snapshot was in old position.
I'd argue that the example code should actually be corrected 😛 What do you think @m-bert ?

Still, I'll dive deeper into it, just need a second opinion about the approach.

m-bert:
What do you mean by "race condition"? JS is single threaded and it shouldn't fall into race condition. Also, as far as I know, order of unmounting is deterministic, so it shouldn't be a problem.

DrRefactor:
> as far as I know, order of unmounting is deterministic, so it shouldn't be a problem.

Indeed, it is

I was writing in a hurry earlier, let me elaborate a bit more about the race.

##  race
1. Component Button (from Habits, with exit animation) updates, last "snapshot" is saved
2. <We hope that no layout update to Button (or its parent) happens here, as snapshot won't be updated anymore>
3. Component ListItem (from Habits) unmounts
4. <we hope that no Button update/render occurs here>
5. <point 2. again>
6. Component Button unmounts

## another argument
Moreover, architecture changes in React such as concurrent mode, encourage us to treat implementation details with a grain of salt. Assuming and basing library core on order of updates/unmounts might be dangerous.

DrRefactor:
I'm trying to dive deeper in Habits example case to figure out way on how to treat it correctly. Maybe I'll get to the point where exit animation code state is in the "best of two worlds" state:
1. Performant (from this PR)
2. Does not jump if siblings unmount first (like on main)
3. Handles non-render layouts well (on main, if component changes its position without render, exit animation won't be correct). This is quite frequent use case, e.g. when parent moves children in scrollview-like manner.
4. Race-proof / not depending on 2,4,5 from https://github.com/software-mansion/react-native-reanimated/pull/5617#issuecomment-1919715313

DrRefactor:
@m-bert I've added version based on ResizeObserver as a second commit. Please re-review 🙏

m-bert:
> 2. <We hope that no layout update to Button (or its parent) happens here, as snapshot won't be updated anymore>

I assume that you mean that we've already clicked the buttton, because otherwise snapshot will be saved. Also what do you mean by "we hope no layout update happens here"? If it does, current approach doesn't consider it anyway, because it would break exiting animation. That's why we use snapshot that was taken before the unmounting phase was started.

> 4. <we hope that no Button update/render occurs here>

There will be no render. This component will be unmounted, so it won't be rerendered. We don't have to hope that it won't, it just won't be.

> (...) e.g. when parent moves children in scrollview-like manner.

We are aware of this problem and I've already started working on it.

> I've added version based on ResizeObserver as a second commit. Please re-review 🙏

I don't think this approach will be enough. Moreover, there are problems with BasicLayoutAnimation example (i.e. exiting doesn't work)

DrRefactor:
> I assume that you mean that we've already clicked the buttton, because otherwise snapshot will be saved

No, I mean before we click a button. Just change layout of "Continue" button, without rendering it. It can be done e.g. by changing parent style, if we memo "Continue" button to prevent extra renders.

> There will be no render. This component will be unmounted, so it won't be rerendered. We don't have to hope that it won't, it just won't be.

Considering this example's code (the same condition for ListItems and Continue button) that's true, we don't have to worry.

> We are aware of this problem and I've already started working on it.

I just checked your PR, good job. Please be aware though, that "true" scroll is just one of the ways that can move DOM element without changing its own style. E.g. changing parent styles: transform, scale, alignItems, justifyContent (...) all can move the item without rendering it.

> I don't think this approach will be enough. Moreover, there are problems with BasicLayoutAnimation example (i.e. exiting doesn't work)

Thanks, nice catch! I've not taken entering animation into account. I pushed new changes that start observing element after entering animation is done. This is consistent with behavior of current implementation - first snapshot is taken before entering animation.

m-bert:
Hi @DrRefactor! I've looked at your changes. As I said earlier, I'm not sure if `ResizeObserver` is a good choice here. There are still problems with "BasicLayoutAnimation" example that you can see on the video below:


https://github.com/software-mansion/react-native-reanimated/assets/63123542/91adffde-09d9-41b7-ba36-1028ea3caeb4

DrRefactor:
Right, it turns out I've only fixed the problem with BasicLayoutAnimation without pressing "update" button first...

I'm having second thoughts now, maybe it's getting too complicated vs current simple approach. Current (main) solution is easily understandable and stable.

I just hate that React calls `componentWillUnmount` and synchronously unmounts elements, instead of calling `componentWillUnmount` for the whole subtree first and unmount it from DOM later. It would be much easier to implement unmount animations that do not require additional layouts on each render. If it'd be otherwise, capturing snapshots would not be required at all.

I've run out of ideas for now, feel free to close this PR, if you have no additional suggestions. I'd prioritize readability & stability over these performance gains.

m-bert:
Unfortunately I don't have any more suggestions. I will close this PR since we cannot find a good way to reconcile our approaches. 

Even though it won't get merged, thank you for this PR and for the discussion! You've pointed out some things that we could have easily missed (like layout changes without re-renders). Also using `ResizeObserver` was interesting idea and perhaps we will be able to utilize it in the future.

PR #5626, opened 2024-02-01
[LA Web] Adjust scroll in exiting

m-bert:
## Summary

Right now exiting animation doesn't take into consideration that element can be inside a ScrollView (i.e. its position can be changed without rerender by scrolling through its parent). This results in weird behavior of animations. This PR aims to fix that problem.

### Before

https://github.com/software-mansion/react-native-reanimated/assets/63123542/b704195b-a72f-45a6-aa61-8d76d7c7b93f

### After

https://github.com/software-mansion/react-native-reanimated/assets/63123542/bb49ee50-c5ac-47d2-b19e-23829baaf1e9

## Limitations

This implementation doesn't take into account nested ScrollViews (only the scroll value from the closest ancestor will be used).

Sometimes `FlatList` elements that were not visible may become visible when their exiting animations triggers.

## Test plan

Tested on example app (mostly on Default Layout Animations example).

<details>
<summary>Horizontal scroll tested on this example</summary>

```jsx
import { Text, StyleSheet, View, FlatList, Pressable } from 'react-native';
import Animated, { FadeOut } from 'react-native-reanimated';

function Rect() {
  return <Animated.View exiting={FadeOut} style={styles.rect} />;
}

const DATA = new Array(100).fill(0).map((_, index) => `${index}`);

  return (
      <Pressable
        onPress={() => setShow((prev) => !prev)}
        style={styles.button}
      />

      {show && (
        <>
          <FlatList
            style={{ marginTop: 200, width: 750 }}
            horizontal={true}
            data={DATA}
            renderItem={({ item }) => <Rect />}
            keyExtractor={(item) => item}
          />
        </>
      )}
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    overflow: 'scroll',
  },

  rect: {
    width: 250,
    height: 50,
    backgroundColor: 'magenta',
    margin: 10,
  },

  button: {
    width: 100,
    height: 50,
    backgroundColor: 'blue',
    position: 'absolute',
    top: 25,
    left: 25,
  },
});

```

</details>


piaskowyk (approved):
👍

src/reanimated2/layoutReanimation/web/componentUtils.ts
```diff
@@ -239,6 +281,23 @@ export function handleExitingAnimation(
...
   const snapshot = snapshots.get(element)!;
+
+  const scrollOffsets = getElementScrollValue(element);
+
+  if (
+    scrollOffsets.scrollTopOffset !== snapshot.scrollOffsets.scrollTopOffset
+  ) {
+    snapshot.top +=
+      snapshot.scrollOffsets.scrollTopOffset - scrollOffsets.scrollTopOffset;
+  }
```
piaskowyk, line 292: The line `scrollOffsets.scrollTopOffset !== snapshot.scrollOffsets.scrollTopOffset` is a bit complicated. Let's use new variables:
```
const currentScrollTopOffset = scrollOffsets.scrollTopOffset;
const lastScrollTopOffset = snapshot.scrollOffsets.scrollTopOffset;
```
just in case you can add comment why `currentScrollTopOffset` can be different with `lastScrollTopOffset`, because this part potentially can be unintuitive for someone in the future.

piaskowyk (approved):

src/reanimated2/layoutReanimation/web/componentUtils.ts
m-bert: Done in [9278f1c](https://github.com/software-mansion/react-native-reanimated/pull/5626/commits/9278f1c6ea28f6cf5925e00728c55a74a8c1468d)

PR #5628, opened 2024-02-01
Add planets example

tomekzaw:
## Summary

This PR adds an example with animated planets.

https://github.com/software-mansion/react-native-reanimated/assets/20516055/24dd3bfc-c78b-435a-958e-d2408361965f


piaskowyk (approved):
🚀🪐

PR #5632, opened 2024-02-01
Add amount example

tomekzaw:
## Summary

This PR adds amount example with animated text input content.

https://github.com/software-mansion/react-native-reanimated/assets/20516055/260f3c11-c3ad-41f6-a5a0-13b0cd00706c

### DIFF
--- a/app/src/examples/AmountExample.tsx
+++ b/app/src/examples/AmountExample.tsx
@@ -1,0 +1,77 @@
+import Animated, {
+  useAnimatedProps,
+  useDerivedValue,
+  useSharedValue,
+  withTiming,
+} from 'react-native-reanimated';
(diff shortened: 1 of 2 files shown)


piaskowyk (approved):

aminroosta:
Hey @tomekzaw
Thank you for the great example. 

However, when I look up the [TextInput](https://reactnative.dev/docs/textinput) documentation, there is no `text` property.   
I tried your example substituting it with `value`, (`Animated.addWhitelistedNativeProps({ value: true });` and it doesn't work. Where does `text` come from? Is this documented somewhere 🙏

tomekzaw:
Hi @aminroosta, `text` is a native prop, it doesn't exist on the JS side.

PR #5636, opened 2024-02-02
Change file structure in plugin

tjzel:
Requires #5599.

## Summary

Currently files in plugin were named after their most important function. This was a temporary solution applied in the past. Now files are properly aggregated by their usage domain.

## Test plan

All tests and CI pass.


bartlomiejbloniarz (approved):
🚀

PR #5639, opened 2024-02-05
Better error when passing easing from 'react-native' instead of 'reanimated'

Latropos:
## Summary
Function `withTiming` should throw an error when easing is not a worklet (or a bound function  run from UI thread).
This is a common bug, because it happens when user is mixing imports from reanimated and animated.

## Test plan
|Before|After|
|--|--|
||The actual names of all easing functions from react-native are all "fun"|
|image|image|

<details><summary>timing animation (example from the table) - code </summary>

```jsx
import Animated, {
  useSharedValue,
  withTiming,
  useAnimatedStyle,
  // Easing, <- this should be the correct import
import { View, Button, StyleSheet, Easing } from 'react-native';
import React from 'react';

    return {
      width: withTiming(randomWidth.value, {
        easing: Easing.linear,
      }),
    };
  });

  return (
      <Animated.View style={[styles.box, style]} />
      <Button
        title="toggle"
        onPress={() => {
        }}
      />
    </View>
  );
}

  container: {
    flex: 1,
    flexDirection: 'column',
  },
  box: {
    width: 100,
    height: 80,
    backgroundColor: 'black',
    margin: 30,
  },
});


```

</details>

<details><summary> Keyframe animation - code </summary>

``` jsx

import Animated, { Keyframe } from 'react-native-reanimated';
import { Button, Easing, View, StyleSheet } from 'react-native';

export default function KeyframeAnimation() {
  const [show, setShow] = useState(false);
  const enteringAnimation = new Keyframe({
    from: {
      originX: 50,
      transform: [{ rotate: '45deg' }, { scale: 0.5 }],
    },
    30: {
      transform: [{ rotate: '-90deg' }, { scale: 2 }],
    },
    50: {
      originX: 70,
    },
    100: {
      originX: 0,
      transform: [{ rotate: '0deg' }, { scale: 1 }],
      easing: Easing.quad,
    },
  })
    .duration(2000)
    .withCallback((finished: boolean) => {
      'worklet';
      if (finished) {
        console.log('callback');
      }
    });
  const exitingAnimation = new Keyframe({
    0: {
      opacity: 1,
      originX: 0,
    },
    30: {
      originX: -50,
      easing: Easing.exp,
    },
    to: {
      opacity: 0,
      originX: 500,
    },
  }).duration(2000);
  return (
    <View style={styles.columnReverse}>
      <Button
        title="animate"
        onPress={() => {
          setShow((last) => !last);
        }}
      />
      <View style={styles.blueBoxContainer}>
        {show && (
          <Animated.View
            entering={enteringAnimation}
            exiting={exitingAnimation}
            style={styles.blueBox}
          />
        )}
      </View>
    </View>
  );
}

  columnReverse: { flexDirection: 'column-reverse' },
  blueBoxContainer: {
    height: 400,
    alignItems: 'center',
    justifyContent: 'center',
  },
  blueBox: {
    height: 100,
    width: 200,
    backgroundColor: 'blue',
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```
</details>

<details> <summary> Curved transition animation - code  </summary>

```jsx
import Animated, {
  BounceOut,
  CurvedTransition,
  LightSpeedInRight,
import {
  Image,
  LayoutChangeEvent,
  Text,
  View,
  StyleSheet,
  Easing,
} from 'react-native';
import React, { useCallback, useEffect, useMemo, useState } from 'react';
import { ScrollView, TapGestureHandler } from 'react-native-gesture-handler';

const AnimatedImage = Animated.createAnimatedComponent(Image);
type Props = {
  columns: number;
  pokemons: number;
};
function getRandomColor() {
  const randomColor = Math.floor(Math.random() * 16777215).toString(16);
  return randomColor;
}
type PokemonData = {
  ratio: number;
  address: string;
  key: number;
  color: string;
};

export function WaterfallGrid({ columns = 3, pokemons = 100 }: Props) {
  const [poks, setPoks] = useState<Array<PokemonData>>([]);
  const [dims, setDims] = useState({ width: 0, height: 0 });
  const handleOnLayout = useCallback(
    (e: LayoutChangeEvent) => {
      const newLayout = e.nativeEvent.layout;
      if (
        dims.width !== +newLayout.width ||
        dims.height !== +newLayout.height
      ) {
        setDims({ width: newLayout.width, height: newLayout.height });
      }
    },
    [dims, setDims]
  );
  const margin = 10;
  const width = (dims.width - (columns + 1) * margin) / columns;
  useEffect(() => {
    if (dims.width === 0 || dims.height === 0) {
      return;
    }
    const poks: {
      ratio: number;
      address: string;
      key: number;
      color: string;
    }[] = [];

    for (let i = 0; i < pokemons; i++) {
      const ratio = 1 + Math.random();
      poks.push({
        ratio,
        address: `https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/${i}.png`,
        key: i,
        color: `#${getRandomColor()}`,
      });
    }
    setPoks(poks);
  }, [dims, setPoks, pokemons]);
  const layoutTransition = CurvedTransition.delay(1000).easingX(Easing.linear);

  const [cardsMemo, height] = useMemo<[Array<JSX.Element>, number]>(() => {
    if (poks.length === 0) {
      return [[], 0];
    }
    const cardsResult: Array<JSX.Element> = [];
    const heights = new Array(columns).fill(0);
    for (const pok of poks) {
      const cur = Math.floor(Math.random() * (columns - 0.01));
      const pokHeight = width * pok.ratio;
      heights[cur] += pokHeight + margin / 2;
      cardsResult.push(
        <Animated.View
          entering={LightSpeedInRight.delay(cur * 200 * 2).springify()}
          exiting={BounceOut}
          layout={layoutTransition}
          key={pok.address}
          style={[
            {
              width: width,
              height: pokHeight,
              backgroundColor: pok.color,
              left: cur * width + (cur + 1) * margin,
              top: heights[cur] - pokHeight,
            },
            styles.pok,
          ]}>
          <TapGestureHandler
            onHandlerStateChange={() => {
              setPoks(poks.filter((it) => it.key !== pok.key));
            }}>
            <AnimatedImage
              layout={layoutTransition}
              source={{ uri: pok.address }}
              style={{ width: width, height: width }}
            />
          </TapGestureHandler>
        </Animated.View>
      );
    }
    return [cardsResult, Math.max(...heights) + margin / 2];
  }, [poks, columns, layoutTransition, width]);
  return (
    <View onLayout={handleOnLayout} style={styles.flexOne}>
      {cardsMemo.length === 0 && <Text> Loading </Text>}
      {cardsMemo.length !== 0 && (
        <ScrollView>
          <View style={{ height: height }}>{cardsMemo}</View>
        </ScrollView>
      )}
    </View>
  );
}
export default function WaterfallGridExample() {
  return (
    <View style={styles.flexOne}>
      <WaterfallGrid columns={3} pokemons={10} />
    </View>
  );
}

  flexOne: {
    flex: 1,
  },
  pok: {
    alignItems: 'center',
    justifyContent: 'center',
    position: 'absolute',
  },
});

```
</details>


src/reanimated2/animation/timing.ts
```diff
@@ -65,14 +65,16 @@ export const withTiming = function (
...
-        'name' in userConfig?.easing ? ` '${userConfig?.easing.name}'` : ''
-      } provided to 'withTiming' is not a worklet. Are you sure you didn't import it from react-native? `
-    );
+  if (userConfig?.easing) {
+    const { easing } = userConfig;
+    if (!('__workletHash' in easing)) {
+      throw new Error(
+        `[Reanimated] The easing function${
+          'name' in easing ? ` '${easing.name}'` : ''
+        } provided to 'withTiming' is not a worklet. Are you sure you didn't import it from react-native? `
```
tomekzaw, line 74:
```suggestion
        } provided to `withTiming` is not a worklet. Are you sure you didn't import it from react-native?`
```

src/reanimated2/animation/timing.ts
```diff
@@ -65,6 +65,16 @@ export const withTiming = function (
 ): Animation<TimingAnimation> {
   'worklet';
 
+  if (userConfig?.easing) {
+    const { easing } = userConfig;
+    if (!('__workletHash' in easing)) {
```
tjzel, line 70: While this is probably safe here, I really don't think we should be using `in` operator in worklets.

piaskowyk, line 70: It seems that the solution doesn't work specifically for that one bezier object in 😅
https://github.com/software-mansion/react-native-reanimated/blob/main/src/reanimated2/Easing.ts#L226

We can evaluate the performance and determine whether this optimization is still necessary. If it doesn't significantly affect the performance, we can consider removing it.

piaskowyk (changes requested):

src/reanimated2/animation/timing.ts
```diff
@@ -67,14 +67,20 @@ export const withTiming = function (
 
   if (userConfig?.easing) {
     const { easing } = userConfig;
-    if (!('__workletHash' in easing)) {
+    const isFunction: boolean = typeof easing === 'function';
+    const isWorklet: boolean = !!(
+      easing as unknown as { __workletHash: number }
+    )?.__workletHash;
+    const functionName = typeof easing === 'function' ? easing?.name : '';
+    const isBound = functionName.startsWith('bound');
```
tomekzaw, line 75: does it work correctly across different JS engines?

Latropos: According to documentation it does: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind

src/reanimated2/animation/timing.ts
```diff
@@ -65,14 +65,16 @@ export const withTiming = function (
...
+  if (userConfig?.easing) {
+    const { easing } = userConfig;
+    if (!('__workletHash' in easing)) {
+      throw new Error(
+        `[Reanimated] The easing function${
+          'name' in easing ? ` '${easing.name}'` : ''
+        } provided to 'withTiming' is not a worklet. Are you sure you didn't import it from react-native? `
+      );
+    }
   }
```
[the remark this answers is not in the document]
tomekzaw: now there are 2 newlines 😄

tjzel (changes requested):
Please also add some test plan in the PR description (with before/after screenshots maybe?).

src/reanimated2/animation/timing.ts
```diff
@@ -65,6 +65,26 @@ export const withTiming = function (
...
+    // validate config
+    if (userConfig?.easing) {
+      const { easing } = userConfig;
+      const isFunction: boolean = typeof easing === 'function';
+      const isWorklet: boolean = !!(
+        easing as unknown as { __workletHash: number }
+      )?.__workletHash;
+      const functionName =
+        typeof easing === 'function' ? easing?.name + ' ' : '';
+      const isBound = functionName.startsWith('bound');
```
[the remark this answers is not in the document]
Latropos: I've added. But Github failed to automatically mark this as "outdated"

src/reanimated2/animation/timing.ts
```diff
@@ -65,6 +65,29 @@ export const withTiming = function (
...
+    // validate config
+    if (userConfig?.easing) {
+      const { easing } = userConfig;
+      const isFunction: boolean = typeof easing === 'function';
+      const isWorklet: boolean = !!(
+        easing as unknown as { __workletHash: number }
+      )?.__workletHash;
+      const functionName = typeof easing === 'function' ? easing?.name : '';
+      /** When calling `withTiming` on UI we don't pass worklets directly, but we
+      bound them first. This happens, for example, when animating with keyframes */
```
tjzel, line 78:
```suggestion
      /** Worklets ran on UI thread are bound first. Therefore if a function wasn't bound it cannot be a worklet. See `valueUnpacker` code for reference. */
```

tjzel (approved):
Just make the comment more accurate and we are good to go!

src/reanimated2/animation/timing.ts
```diff
@@ -65,6 +65,29 @@ export const withTiming = function (
 ): Animation<TimingAnimation> {
   'worklet';
 
+  if (__DEV__) {
+    // validate config
+    if (userConfig?.easing) {
```
tomekzaw, line 70: Can we merge it into one condition (`__DEV__ && userConfig?.easing`) and save one level of indentation?

src/reanimated2/animation/timing.ts
```diff
@@ -65,6 +65,29 @@ export const withTiming = function (
...
+      const { easing } = userConfig;
+      const isFunction: boolean = typeof easing === 'function';
+      const isWorklet: boolean = !!(
+        easing as unknown as { __workletHash: number }
+      )?.__workletHash;
```
tomekzaw, line 75: Whaat?

src/reanimated2/animation/timing.ts
```diff
@@ -65,6 +65,29 @@ export const withTiming = function (
...
+      const functionName = typeof easing === 'function' ? easing?.name : '';
+      /** When calling `withTiming` on UI we don't pass worklets directly, but we
+      bound them first. This happens, for example, when animating with keyframes */
+      const isBound = functionName.startsWith('bound');
```
tomekzaw, line 79: Why exactly do we need to check that function is not bound?

tomekzaw (commented):
Looks good, left some questions in the comments.

Also, shouldn't we also perform the same check here:
* ComplexAnimationBuilder https://github.com/software-mansion/react-native-reanimated/blob/deb741ba42e6cb39e200e7c84806e9371d99587d/src/reanimated2/layoutReanimation/animationBuilder/ComplexAnimationBuilder.ts#L42
* CurvedTransition https://github.com/software-mansion/react-native-reanimated/blob/deb741ba42e6cb39e200e7c84806e9371d99587d/src/reanimated2/layoutReanimation/defaultTransitions/CurvedTransition.ts#L48
* JumpingTransition https://github.com/software-mansion/react-native-reanimated/blob/deb741ba42e6cb39e200e7c84806e9371d99587d/src/reanimated2/layoutReanimation/defaultTransitions/JumpingTransition.ts#L7
* Keyframe https://github.com/software-mansion/react-native-reanimated/blob/deb741ba42e6cb39e200e7c84806e9371d99587d/src/reanimated2/layoutReanimation/animationBuilder/Keyframe.ts#L125

If yes, it probably makes sense to move it to a utility function `assertEasingIsWorklet` and call it from all these places.

src/reanimated2/animation/timing.ts
tjzel: @tomekzaw I described it in my comment.

tomekzaw: can't find the comment :(

tjzel: Here you go! 
Screenshot 2024-02-12 at 17 21 00

tomekzaw: Aaaa, this comment, thanks. What about bound functions in the RN runtime?

tjzel: Now when I think about it, I think it's enough to just check for `isWorklet` here, isn't it?

tomekzaw: Probably yes

tjzel:
@Latropos Can you check if it works on Web?

Latropos:
@tjzel I've tested, it works on web

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,29 @@ if (__DEV__ && IS_REDUCED_MOTION) {
   );
 }
 
+export function assertEasingIsWorklet(
+  easing: EasingFunction | EasingFunctionFactory
+) {
+  'worklet';
+
+  if (isWeb()) {
```
tjzel, line 52: This is the pattern we currently use throughout the code.
```suggestion
const SHOULD_BE_USE_WEB = SHOULD_BE_USE_WEB();

export function assertEasingIsWorklet(
  easing: EasingFunction | EasingFunctionFactory
) {
  'worklet';

  if (SHOULD_BE_USE_WEB) {
```

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,29 @@ if (__DEV__ && IS_REDUCED_MOTION) {
...
+    /* It is possible to run reanimated on web without plugin, so let's skip this check on web */
+    return true;
+  }
+  const isFunction: boolean = typeof easing === 'function';
+  const functionName = typeof easing === 'function' ? easing?.name : '';
+  /* Worklets ran on UI thread are bound first. Therefore if a function wasn't bound it cannot be a worklet. See `valueUnpacker` code for reference. */
```
tjzel, line 58: I feel bad for reworking this comment over and over, but I want to make it as informative as possible.

```suggestion
  /* Just checking with `isWorkletFunction` is not sufficient in case of Keyframe animations. Therefore, we also check for binding. */
```

tjzel (approved):

src/reanimated2/animation/timing.ts
tjzel: Actually it's not sufficient in case of Keyframe animations - however I think understanding why is that should be a separate task (if we actually want to investigate it).

src/reanimated2/animation/util.ts
```diff
@@ -44,18 +44,20 @@ if (__DEV__ && IS_REDUCED_MOTION) {
   );
 }
 
+const SHOULD_BE_USE_WEB = SHOULD_BE_USE_WEB();
```
tomekzaw, line 47:
```suggestion
const SHOULD_BE_USE_WEB = shouldBeUseWeb();
```

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,31 @@ if (__DEV__ && IS_REDUCED_MOTION) {
...
+  'worklet';
+
+  if (SHOULD_BE_USE_WEB) {
+    /* It is possible to run reanimated on web without plugin, so let's skip this check on web */
+    return true;
+  }
+  const isFunction: boolean = typeof easing === 'function';
+  const functionName = typeof easing === 'function' ? easing?.name : '';
+  /* Just checking with `isWorkletFunction` is not sufficient in case of keyframe animations. Therefore, we also check for binding. */
+  const isBound = functionName.startsWith('bound');
```
piaskowyk, line 61: Lets try to avoid this check by fixing the Keyframe easing.

piaskowyk (changes requested):

src/reanimated2/animation/util.ts
Latropos: It was fixed with the following line
```tsx
    functionInstance.__workletHash = workletHash;
```
Now function `isWorkletFunction` returns true for bound worklets as well

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,31 @@ if (__DEV__ && IS_REDUCED_MOTION) {
...
+
+  if (!isBound && isFunction && !isWorkletFunction(easing)) {
+    throw new Error(
+      `[Reanimated] The easing function ${
+        functionName ? `\`${functionName}\` ` : ''
+      }provided to \`withTiming\` is not a worklet. Are you sure you didn't import it from react-native? `
```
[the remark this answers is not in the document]
Latropos: Since handling bound functions was fixed, we don't need to check `functionName` anymore.

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,31 @@ if (__DEV__ && IS_REDUCED_MOTION) {
...
+  easing: EasingFunction | EasingFunctionFactory
+) {
+  'worklet';
+
+  if (SHOULD_BE_USE_WEB) {
+    /* It is possible to run reanimated on web without plugin, so let's skip this check on web */
+    return true;
+  }
+  const isFunction: boolean = typeof easing === 'function';
+  const functionName = typeof easing === 'function' ? easing?.name : '';
```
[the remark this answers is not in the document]
Latropos: Since handling bound functions was fixed, we don't need to check `functionName` anymore.

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,26 @@ if (__DEV__ && IS_REDUCED_MOTION) {
...
+) {
+  'worklet';
+
+  if (SHOULD_BE_USE_WEB) {
+    /* It is possible to run reanimated on web without plugin, so let's skip this check on web */
+    return true;
+  }
+  const isFunction: boolean = typeof easing === 'function';
+
+  if (isFunction && !isWorkletFunction(easing)) {
```
tomekzaw, line 60: Maybe we should move `typeof easing === 'function'` function to `isWorkletFunction`?

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,26 @@ if (__DEV__ && IS_REDUCED_MOTION) {
...
+    throw new Error(
+      `[Reanimated] The easing function provided to \`withTiming\` is not a worklet. Are you sure you didn't import it from react-native? `
```
tomekzaw, line 62: It looks like `assertEasingIsWorklet` is called from various functions, not only from `withTiming`, so the error message is incorrect in other cases.

src/reanimated2/valueUnpacker.ts
```diff
@@ -46,6 +46,8 @@ function valueUnpacker(objectToUnpack: any, category?: string): any {
       workletsCache.set(workletHash, workletFun);
     }
     const functionInstance = workletFun.bind(objectToUnpack);
+    functionInstance.__workletHash = workletHash;
```
tomekzaw, line 49: Why do we need this change?

tomekzaw (changes requested):

src/reanimated2/valueUnpacker.ts
Latropos: Because `functionInstance` if a binding of workletFun, so when you run `functionInstance` you run a worklet. Therefore `isWorkletFunction(functionInstance)` should be true, and since we decide if given function is a worklet reading its hash I set it here.
I was considering passing here some other values as well
```suggestion
    functionInstance.__workletHash = 0; //Provide some hash value, so isWorkletFunction can correctly detect it as a worklet
```

tomekzaw (changes requested):

src/reanimated2/commonTypes.ts
```diff
@@ -131,7 +131,11 @@ export function isWorkletFunction<
   'worklet';
   // Since host objects always return true for `in` operator, we have to use dot notation to check if the property exists.
   // See https://github.com/facebook/hermes/blob/340726ef8cf666a7cce75bc60b02fa56b3e54560/lib/VM/JSObject.cpp#L1276.
-  return !!(value as Record<string, unknown>).__workletHash;
+  return (
+    typeof value === 'function' &&
+    '__workletHash' in value &&
+    !!value.__workletHash
+  );
```
tomekzaw, line 138: The reason why we check `!!value.__workletHash` instead of `'__workletHash' in value` is that `in` operator always returns `true` for `jsi::HostObject`. We actually take advantage of that in `isHostObject` function:

https://github.com/software-mansion/react-native-reanimated/blob/4a15cf467b709a9365ae57a8eb7ba28397984a29/src/reanimated2/shareables.ts#L25-L32

Since we first check `typeof value === 'function'`, I guess it's safe to assume that once it passes, `value` is not a `jsi::HostObject`, however it could be a `jsi::HostFunction` and I'm not sure how `in` operator behaves in this case.

If possible, I would avoid changing that and I recommended reverting to the previous version (just with correct order of checks):

```suggestion
  return (
    typeof value === 'function' &&
    !!(value as Record<string, unknown>).__workletHash;
  );
```

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,27 @@ if (__DEV__ && IS_REDUCED_MOTION) {
...
+const SHOULD_BE_USE_WEB = shouldBeUseWeb();
+
+export function assertEasingIsWorklet(
+  easing: EasingFunction | EasingFunctionFactory
+) {
+  'worklet';
+  if (_WORKLET) {
+    // If 'assertEasingIsWorklet' is called on UI runtime then easing is a copy and it is not a worklet
+    return true;
+  }
```
tomekzaw, line 56: We don't plan to call `assertEasingIsWorklet` on the UI runtime. This function is meant to work only on the RN runtime.

1. Let's remove `'worklet';` directive to forbid calling it from the UI runtime
2. Let's remove `if (_WORKLET) { ... }` since we don't need to support that
```suggestion
  'worklet';
  if (_WORKLET) {
    // If 'assertEasingIsWorklet' is called on UI runtime then easing is a copy and it is not a worklet
    return true;
  }
```

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,27 @@ if (__DEV__ && IS_REDUCED_MOTION) {
...
+  if (SHOULD_BE_USE_WEB) {
+    // It is possible to run reanimated on web without plugin, so let's skip this check on web
+    return true;
```
tomekzaw, line 59: This function just asserts so there's no need for it to have a return value.
```suggestion
    return;
```

src/reanimated2/animation/util.ts
```diff
@@ -43,6 +44,27 @@ if (__DEV__ && IS_REDUCED_MOTION) {
   );
 }
 
+const SHOULD_BE_USE_WEB = shouldBeUseWeb();
+
+export function assertEasingIsWorklet(
+  easing: EasingFunction | EasingFunctionFactory
+) {
```
tomekzaw, line 51:
```suggestion
): void {
```

tomekzaw (changes requested):

src/reanimated2/animation/util.ts
Latropos: I've added assertions each time `assertEasingIsWorklet` is called. Maybe we should move both `__DEV__` and `!_WORKLET` inside the fnciton?
```
 if (__DEV__ && userConfig?.easing && !_WORKLET) {
    assertEasingIsWorklet(userConfig?.easing);
  }
```

tomekzaw: No, `assertEasingIsWorklet` should only be called on the RN runtime. If it is called on the UI runtime, this means something went wrong and we should fix the root cause. There's no need to check `!_WORKLET`, please revert that change.

src/reanimated2/animation/util.ts
```diff
@@ -33,16 +33,38 @@ import {
...
+    // but its bound copy, which is not a worklet. We don't want to throw any error then.
+    return;
+  }
+  if (SHOULD_BE_USE_WEB) {
+    // It is possible to run reanimated on web without plugin, so let's skip this check on web
+    return;
+  }
+  if (!isWorkletFunction(easing)) {
+    throw new Error(
+      `[Reanimated] The easing function is not a worklet. Are you sure you didn't import it from react-native? `
```
tomekzaw, line 63:
```suggestion
      '[Reanimated] The easing function is not a worklet. Please make sure you import `Easing` from react-native-reanimated.'
```

tomekzaw (commented):
Looks good to me, let's just slightly improve the error message.

src/reanimated2/animation/timing.ts
```diff
@@ -65,6 +69,10 @@ export const withTiming = function (
 ): Animation<TimingAnimation> {
   'worklet';
 
+  if (__DEV__ && userConfig?.easing) {
+    assertEasingIsWorklet(userConfig?.easing);
```
piaskowyk, line 73: You can consider if this additional nullcheck is necessary 
```suggestion
    assertEasingIsWorklet(userConfig.easing);
```

piaskowyk (approved):

piaskowyk (approved):

PR #5644, opened 2024-02-06
Fix warning about overwriting style with layout animation

### LINKED ISSUE #5395: LayoutAnimations `entering` and `exiting` not working
### Description

Hey yall,

I tried playing around with LayoutAnimations and I love the idea - they're so simple to implement, amazing work!

https://github.com/software-mansion/react-native-reanimated/assets/15199031/3fac8af4-25c7-4482-9a4d-877fe70c30c0

I did notice that there are a few bugs though.

I created a repro here that showcases those bugs: https://github.com/software-mansion/react-native-reanimated/pull/5394

## 1. Exiting not running

The `exiting` animation is not running in the given example.

## 2. Custom `useAnimatedStyle` not running on first mount

On the first mount, the `useAnimatedStyle` for the second card is not properly applied. I think this is a race condition where the LayoutAnimation takes precedence over the `useAnimatedStyle`.

## 3. Some items not showing

Sometimes the items just don't show up.

## 4. `layout` instead of `useAnimatedStyle` does not work

When using `layout` instead of manually using `useAnimatedStyle` it will not consider properties in the `transform` style.

This is how I animate the slide & rotate to the right now:

```tsx
const rotation = useDerivedValue(() => withSpring(index * 15), [index]);

  return {
    zIndex: -index,
    backgroundColor: item.color,
    transform: [
      {
        translateX: interpolate(rotation.value, [0, 100], [0, 100]),
      },
      {
        translateY: interpolate(rotation.value, [0, 100], [0, -30]),
      },
      {
        scale: interpolate(rotation.value, [0, 100], [1, 0.7]),
      },
      {
        rotate: `${rotation.value}deg`,
      },
    ],
  };
}, [rotation, index, item]);

return (
  <Animated.View
    entering={Entering}
    exiting={Exiting}
    style={[styles.item, animatedStyle]}
  />
);
```

This is how I want to do it:

```tsx
const style = useMemo(() => {
  return {
    zIndex: -index,
    backgroundColor: item.color,
    transform: [
      {
        translateX: x,
      },
      {
        translateY: y,
      },
      {
        scale: scale,
      },
      {
        rotate: `${rotation}deg`,
      },
    ],
  };
}, [rotation, index, item]);

return (
  <Animated.View
    entering={Entering}
    exiting={Exiting}
    layout={Move} // <-- just a default Layout.springify() animation that should animate the props in transform: [...]
    style={[styles.item, style]}
  />
);
```

Let me know if you need more details!

### Steps to reproduce

https://github.com/software-mansion/react-native-reanimated/pull/5394 <-- open the new page

https://github.com/software-mansion/react-native-reanimated/pull/5394

### Reanimated version

3.5.0 (main)

### React Native version

0.72.6

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 15 Pro Simulator

### Acknowledgements

Yes

Latropos:
## Summary
In this PR https://github.com/software-mansion/react-native-reanimated/pull/4969 I didn't include neither animatedStyles nor styles nested in arrays.
That's why the problem wasn't reported in this example: https://github.com/software-mansion/react-native-reanimated/issues/5395

closes #5395 

|Before|After|
|---|---|
|image|image|

## Test plan

```jsx
import Animated, {
  useAnimatedStyle,
  SlideInDown,
  ZoomOut,
  useDerivedValue,
  withSpring,
  interpolate,

interface Item {
  color: string;
}

function getRandomColor(): string {
  return '#' + Math.floor(Math.random() * 16777215).toString(16);
}

const Entering = SlideInDown.springify().mass(1).stiffness(1000).damping(500);
const Exiting = ZoomOut.springify().mass(1).stiffness(1000).damping(500);

function ItemComponent({ item, index }: { item: Item; index: number }) {
  const rotation = useDerivedValue(() => withSpring(index * 15), [index]);

    return {
      zIndex: -index,
      backgroundColor: item.color,
      transform: [
        {
          translateX: interpolate(rotation.value, [0, 100], [0, 100]),
        },
        {
          translateY: interpolate(rotation.value, [0, 100], [0, -30]),
        },
        {
          scale: interpolate(rotation.value, [0, 100], [1, 0.7]),
        },
        {
          rotate: `${rotation.value}deg`,
        },
      ],
    };
  }, [rotation, index, item]);

  return (
    <Animated.View
      entering={Entering}
      exiting={Exiting}
      style={[styles.item, animatedStyle]}
    />
  );
}

export default function StackedLayoutAnimationExample() {
  const [items, setItems] = useState<Item[]>([
    { color: 'red' },
    { color: 'green' },
  ]);

  useEffect(() => {
    const interval = setInterval(() => {
      setItems((i) => [{ color: getRandomColor() }, ...i]);
    }, 1500);
    return () => clearInterval(interval);
  }, []);

  return (
      {items.map((item, index) => {
        if (index > 2) {
          return null;
        }
        return <ItemComponent key={item.color} item={item} index={index} />;
      })}
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  item: {
    width: 100,
    height: 150,
    borderRadius: 15,
    position: 'absolute',
  },
});
```
</details>


src/animationBuilder.tsx
```diff
@@ -25,12 +26,52 @@ const mockTargetValues: LayoutAnimationsValues = {
   currentBorderRadius: 0,
 };
 
+function maybeReportOverwrittenProperties(
+  layoutAnimationStyle: StyleProps,
+  style: NestedArray<StyleProps>,
+  displayName: string
+) {
+  function getCommonProperties(
```
tomekzaw, line 34: Does function `getCommonProperties` need to be nested inside `maybeReportOverwrittenProperties`?

src/animationBuilder.tsx
```diff
@@ -25,12 +26,52 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+
+  const commonProperties = getCommonProperties(layoutAnimationStyle, style);
+
+  if (commonProperties.length > 0) {
+    console.warn(
+      `[Reanimated] ${
+        commonProperties.length === 1 ? 'Property' : 'Properties: '
+      } "${commonProperties.join(
+        ', '
+      )}" of ${displayName} may be overwritten with layout animation. Please create a wrapper with the layout animation you want to apply.`
```
tomekzaw, line 64:
```suggestion
      )}" of ${displayName} may be overwritten by a layout animation. Please wrap your component with an animated view and apply the layout animation on the wrapper.`
```

src/animationBuilder.tsx
```diff
@@ -25,12 +26,52 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+      : [secondStyle];
+    const commonKeys: Array<string> = [];
+
+    secondStyleFlat.forEach((s) => {
+      if ('initial' in s) {
+        s = s.initial.value;
+      }
+
+      const commonStyleKeys = Object.keys(s).filter((key) =>
+        Object.prototype.hasOwnProperty.call(layoutStyle, key)
```
tomekzaw, line 49: Why not just `key in layoutStyle`?

src/animationBuilder.tsx
```diff
@@ -25,12 +26,52 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+      );
+      commonKeys.push(...commonStyleKeys);
```
tomekzaw, line 51: Why not just use `commonKeys.concat`?

src/animationBuilder.tsx
```diff
@@ -42,21 +83,13 @@ export function maybeBuild(
...
-        `[Reanimated] ${
-          commonProperties.length === 1 ? 'Property' : 'Properties: '
-        } "${commonProperties.join(
-          ', '
-        )}" of ${displayName} may be overwritten with layout animation. Please create a wrapper with the layout animation you want to apply.`
+    if (__DEV__) {
+      maybeReportOverwrittenProperties(
+        layoutAnimationStyle,
+        style || {},
+        displayName
```
tomekzaw, line 92: Does it make sense to call `maybeReportOverwrittenProperties` for an empty object if `style` is undefined?
```suggestion
    if (__DEV__ && style) {
      maybeReportOverwrittenProperties(
        layoutAnimationStyle,
        style,
        displayName
```

src/animationBuilder.tsx
```diff
@@ -42,21 +83,13 @@ export function maybeBuild(
...
-    if (commonProperties.length > 0) {
-      console.warn(
-        `[Reanimated] ${
-          commonProperties.length === 1 ? 'Property' : 'Properties: '
-        } "${commonProperties.join(
-          ', '
-        )}" of ${displayName} may be overwritten with layout animation. Please create a wrapper with the layout animation you want to apply.`
+    if (__DEV__) {
+      maybeReportOverwrittenProperties(
+        layoutAnimationStyle,
```
tomekzaw, line 90: Let's use `layoutAnimation.animations` directly here instead of assigning it to `layoutAnimationStyle`.
```suggestion
        layoutAnimation.animations,
```

src/animationBuilder.tsx
```diff
@@ -25,12 +26,52 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+    });
+    return commonKeys;
+  }
+
+  const commonProperties = getCommonProperties(layoutAnimationStyle, style);
+
+  if (commonProperties.length > 0) {
+    console.warn(
+      `[Reanimated] ${
+        commonProperties.length === 1 ? 'Property' : 'Properties: '
```
tomekzaw, line 61:
```suggestion
        commonProperties.length === 1 ? 'Property' : 'Properties '
```

src/animationBuilder.tsx
```diff
@@ -25,12 +26,52 @@ const mockTargetValues: LayoutAnimationsValues = {
   currentBorderRadius: 0,
 };
 
+function maybeReportOverwrittenProperties(
+  layoutAnimationStyle: StyleProps,
+  style: NestedArray<StyleProps>,
+  displayName: string
```
tomekzaw, line 32: Maybe we should rename it to `componentName`?

src/animationBuilder.tsx
```diff
@@ -25,12 +26,52 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+) {
+  function getCommonProperties(
+    layoutStyle: object,
+    secondStyle: object | Array<object>
```
tomekzaw, line 36: Can we use something more specific than `object` here?

src/animationBuilder.tsx
```diff
@@ -25,12 +26,52 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+  ) {
+    const secondStyleFlat = Array.isArray(secondStyle)
+      ? secondStyle.flat()
+      : [secondStyle];
+    const commonKeys: Array<string> = [];
+
+    secondStyleFlat.forEach((s) => {
+      if ('initial' in s) {
+        s = s.initial.value;
+      }
```
tomekzaw, line 46: Why do we need this piece of code?

tomekzaw (changes requested):

src/animationBuilder.tsx
Latropos: No, it doesn't

src/animationBuilder.tsx
Latropos: Good catch, It seems, that concat should be a little faster: https://www.measurethat.net/Benchmarks/Show/7699/0/array-concat-vs-spread-vs-push-spread-loop-apply#latest_results_block

src/animationBuilder.tsx
Latropos: We use this string from `createAnimatedComponent`:
```tsx
  AnimatedComponent.displayName = `AnimatedComponent(${
    Component.displayName || Component.name || 'Component'
  })`;
```
So I would keep the current name

src/animationBuilder.tsx
Latropos: To include animated style, for example here we should throw an error: https://github.com/software-mansion/react-native-reanimated/issues/5395

tomekzaw: Okay, this makes sense, so `'initial' in s` is to check if `s` is an animated style?

If yes, maybe we should add `isAnimatedStyle` function in a separate PR for cases like this one.

src/animationBuilder.tsx
```diff
@@ -25,12 +26,50 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+    : [secondStyle];
+
+  let commonKeys: Array<string> = [];
+  secondStyleFlat.forEach((s) => {
+    if ('initial' in s) {
+      s = s.initial.value; // Include properties of animated style
+    }
+
+    const commonStyleKeys = Object.keys(s).filter((key) => key in layoutStyle);
+    commonKeys = commonKeys.concat(...commonStyleKeys);
```
tomekzaw, line 44: It should work without `...`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat
```suggestion
    commonKeys = commonKeys.concat(commonStyleKeys);
```

src/animationBuilder.tsx
tomekzaw: Okay, thanks for the response

tomekzaw (approved):
Thanks for adding this warning!

src/animationBuilder.tsx
```diff
@@ -25,12 +26,50 @@ const mockTargetValues: LayoutAnimationsValues = {
   currentBorderRadius: 0,
 };
 
+function getCommonProperties(
+  layoutStyle: StyleProps,
+  secondStyle: StyleProps | Array<StyleProps>
```
tjzel, line 31:
```suggestion
  componentStyle: StyleProps | Array<StyleProps>
```

src/animationBuilder.tsx
```diff
@@ -25,12 +26,50 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+) {
+  const secondStyleFlat = Array.isArray(secondStyle)
```
tjzel, line 33:
```suggestion
  const stylesFlatArray = Array.isArray(secondStyle)
```

src/animationBuilder.tsx
```diff
@@ -25,12 +26,50 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+    ? secondStyle.flat()
+    : [secondStyle];
+
+  let commonKeys: Array<string> = [];
+  secondStyleFlat.forEach((s) => {
```
tjzel, line 38:
```suggestion
  secondStyleFlat.forEach((style) => {
```

src/animationBuilder.tsx
```diff
@@ -25,12 +26,50 @@ const mockTargetValues: LayoutAnimationsValues = {
...
+    if ('initial' in s) {
+      s = s.initial.value; // Include properties of animated style
+    }
+
+    const commonStyleKeys = Object.keys(s).filter((key) => key in layoutStyle);
+    commonKeys = commonKeys.concat(commonStyleKeys);
+  });
+  return commonKeys;
```
tjzel, line 46: You might get duplicate keys that way. How about:

```ts
function getCommonProperties(
  layoutStyle: StyleProps,
  secondStyle: StyleProps | Array<StyleProps>
) {
  let stylesFlatArray = Array.isArray(secondStyle)
    ? secondStyle.flat()
    : [secondStyle];

  stylesFlatArray = stylesFlatArray.map((style) => {
    if ('initial' in style) {
      return style.initial.value;
    }
    return style;
  });

  const componentStylesKeys = stylesFlatArray.flatMap((style) =>
    Object.keys(style)
  );

  const commonKeys = Object.keys(layoutStyle).filter((key) =>
    componentStylesKeys.includes(key)
  );

  return commonKeys;
}
```

tjzel (approved):
All looks good, names of some variables could be better but it's very minor. Left a suggestion if you are interested in it, but your solution is good as well.

TS types are not the best, but since this file hasn't undergone proper typing it's okay.

PR #5645, opened 2024-02-06
Add global prefix

jakub-gonet:
## Summary

Added `global` prefix to properties accessed on global object.

## Test plan

- Test various screens in example app.

### DIFF
--- a/plugin/src/globals.ts
+++ b/plugin/src/globals.ts
@@ -131,7 +131,6 @@ export const defaultGlobals = new Set([
   '_setGestureState',
   '_notifyAboutProgress',
   '_notifyAboutEnd',
-  '_runOnUIQueue',
   '_getAnimationTimestamp',
 ]);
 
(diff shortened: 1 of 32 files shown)


plugin/src/globals.ts
```diff
@@ -111,28 +111,6 @@ export const defaultGlobals = new Set([
...
-  '_measurePaper',
-  '_measureFabric',
-  '_scrollToPaper',
-  '_dispatchCommandPaper',
-  '_dispatchCommandFabric',
-  '_setGestureState',
-  '_notifyAboutProgress',
-  '_notifyAboutEnd',
-  '_runOnUIQueue',
-  '_getAnimationTimestamp',
```
tjzel, line 135: This is potentially a breaking change if someone uses some of those identifiers in worklets. Let's move on to using `global` in Reanimated but keep this list for now.

tjzel (changes requested):
As @tomekzaw mentioned, since `_WORKLET` is a part of public API we should keep it accessed without `global`.

tjzel (approved):
:shipit:

PR #5647, opened 2024-02-07
Fix higher-order animations re-render behaviour

szydlovsky:
## Summary

Currently, 3/4 of higher order animations - `withRepeat`, `withDelay` and `withSequence` - show undefined behaviour when the component they are in gets re-rendered. They either completely stop in the middle or continue the included lower-order animation beginning with the older `.currentValue`, but of a full animation duration (which makes them feel much slower). This fix makes them reset to the very beginning every time re-render of their component happens.

## Test plan

All three animations in the example should properly reset when re-render button is pressed while they are ongoing.

``` TYPESCRIPT
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withRepeat,
  withDelay,
  withSequence,
  withTiming,
import { StyleSheet, View, Text, Button } from 'react-native';

function RepeatChild({ color }) {
  const sv = useSharedValue(0);
  sv.value = withRepeat(withTiming(200, { duration: 1500 }), 2, true);

    return { width: 1 + sv.value };
  });

  return (
    <View style={{ ...styles.child, backgroundColor: color }}>
      <Text>withRepeat</Text>
    </View>
  );
}

function DelayChild({ color }) {
  const sv = useSharedValue(0);
  sv.value = withDelay(
    1000,
    withRepeat(withTiming(200, { duration: 1000 }), 2, true)
  );

    return { width: 1 + sv.value };
  });

  return (
    <View style={{ ...styles.child, backgroundColor: color }}>
      <Text>withDelay</Text>
    </View>
  );
}

function SequenceChild({ color }) {
  const sv = useSharedValue(0);
  sv.value = withSequence(
    withTiming(200, { duration: 750 }),
    withTiming(0, { duration: 750 }),
    withTiming(200, { duration: 750 }),
    withTiming(0, { duration: 750 })
  );

    return { width: 1 + sv.value };
  });

  return (
    <View style={{ ...styles.child, backgroundColor: color }}>
      <Text>withSequence</Text>
    </View>
  );
}

export default function RerendersExample() {
  const [dummy, setDummy] = useState<boolean>(false);
  const [renderCount, setRenderCount] = useState<number>(0);
  const onPress = () => {
    setDummy(!dummy);
    setRenderCount(renderCount + 1);
  };

  return (
      <Text>Children re-render count: {renderCount}</Text>
      <RepeatChild color={dummy ? '#57607A' : '#57697a'} />
      <DelayChild color={dummy ? '#57607A' : '#57697a'} />
      <SequenceChild color={dummy ? '#57607A' : '#57697a'} />
      <Button title="Re-render children" color="#000000" onPress={onPress} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    marginVertical: 50,
  },
  child: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    alignSelf: 'stretch',
  },
  box: {
    height: 100,
    backgroundColor: 'purple',
  },
});

```
</details>


Example/ios/Podfile.lock
```diff
@@ -1449,4 +1449,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: fca071226b19ce84770ff3ba64fd881fcb93dead
 
-COCOAPODS: 1.14.3
+COCOAPODS: 1.15.2
```
Latropos, line 1452: 👀

src/reanimated2/animation/delay.ts
```diff
@@ -31,6 +32,21 @@ export const withDelay = function <T extends AnimationObject>(
   reduceMotion?: ReduceMotion
 ): Animation<DelayAnimation> {
   'worklet';
+
+  const initialAnimationBounds: AnimationBounds =
+    typeof _nextAnimation === 'function'
+      ? {
+          start: (_nextAnimation() as AnimationObject)
+            .startValue as AnimatableValue,
```
Latropos, line 40: Please get rid of all the typecast here:
```
start: _nextAnimation().startValue,
end: _nextAnimation().toValue,
```
Either provide default value for `undefined` or allow `undefined` in AnimationBounds.

Latropos (commented):
Please fix types and remove all unnecessary casts

src/reanimated2/animation/repeat.ts
```diff
@@ -100,7 +114,14 @@ export const withRepeat = function <T extends AnimationObject>(
         now: Timestamp,
         previousAnimation: Animation<any> | null
       ): void {
-        animation.startValue = value;
+        // Detect re-render
+        const hasBeenInterrupted =
+          value !== initialAnimationBounds.start &&
+          value !== initialAnimationBounds.end;
+        const startingValue = hasBeenInterrupted
+          ? initialAnimationBounds.start
```
Latropos, line 122: Could this be undefined?

Latropos (commented):
Btw, you can use the following markdown commands to make your PR description better 

```
<details><summary>Code</Summary>

YOUR CODE HERE
</details>
```
<details><summary>Code</Summary>

YOUR CODE HERE
</details>

Example/ios/Podfile.lock
szydlovsky: @Latropos Apparently, .2 fixed the 1.15.1 issues

src/reanimated2/animation/repeat.ts
```diff
@@ -39,6 +40,13 @@ export const withRepeat = function <T extends AnimationObject>(
 ): Animation<RepeatAnimation> {
   'worklet';
 
+  const a =
```
Latropos, line 43: We already define this inside defineAnimation.
```suggestion
  const nextAnimation =
```
Please name this variable `nextAnimation` instead of `a` and remove the second declaration in line 55.

src/reanimated2/animation/delay.ts
```diff
@@ -31,6 +32,15 @@ export const withDelay = function <T extends AnimationObject>(
   reduceMotion?: ReduceMotion
 ): Animation<DelayAnimation> {
   'worklet';
+
+  const a =
```
Latropos, line 36: Same comment as in `repeat.ts`, we don't need to have this condition twice

src/reanimated2/animation/sequence.ts
```diff
@@ -58,6 +59,14 @@ export function withSequence(
     });
   }
 
+  const initialAnimationBounds: AnimationBounds[] = _animations.map((anim) => {
+    const a = typeof anim === 'function' ? anim() : anim;
+    return {
+      start: a.startValue,
+      end: a.toValue,
+    };
+  });
```
Latropos, line 68: Please don't use one-letter-names:
```suggestion
  const initialAnimationBounds: AnimationBounds[] = _animations.map((animationObject) => {
    const animation = typeof animationObject === 'function' ? animationObject() : animationObject;
    return {
      start: animation.startValue,
      end: animation.toValue,
    };
  });
```

Latropos (changes requested):
Please refactor your code to get rid of code duplication

src/reanimated2/animation/sequence.ts
```diff
@@ -140,10 +149,18 @@ export function withSequence(
           ] as SequenceAnimation;
         }
 
+        // Detect re-render
+        const hasBeenInterrupted =
+          value !== initialAnimationBounds[animation.animationIndex].start &&
```
Latropos, line 154: Extract `initialAnimationBounds[animation.animationIndex]` into a variable please

src/reanimated2/animation/repeat.ts
```diff
@@ -100,7 +108,14 @@ export const withRepeat = function <T extends AnimationObject>(
...
         previousAnimation: Animation<any> | null
       ): void {
-        animation.startValue = value;
+        // Detect re-render
+        const hasBeenInterrupted =
+          value !== initialAnimationBounds.start &&
+          value !== initialAnimationBounds.end;
+        const startingValue = hasBeenInterrupted
+          ? initialAnimationBounds.start ?? value
+          : value;
```
Latropos, line 117: I think that this is easier to read:

```suggestion
        const startingValue = (hasBeenInterrupted && initialAnimationBounds.start)
          ? initialAnimationBounds.start
          : value;
```

Latropos (changes requested):
I am not sure if the change really solves the problem. Intuitively, `withRepeat` repeating only once should behave exactly like if its inner animation wasn't wrapped in it.
Now not only these behaviours are different, but the behaviour of interrupting `withRepeat` is inconsistent (sometimes it restarts the animation, but once it finishes it doesn't)

https://github.com/software-mansion/react-native-reanimated/assets/56199675/b4af7a5e-bafa-4f71-949d-6620bb00f71a

szydlovsky:
@Latropos 
> Intuitively, `withRepeat` repeating only once should behave exactly like if its inner animation wasn't wrapped in it.

The lower-order animations can properly continue after re-render. Unfortunately, with higher-order ones, we would have to compute the remaining time for the inner animation (and a few more factors e.g. when the inner animation is withSpring), so resetting it is the most we can do without adding a ton of edge cases for each higher-order animation paired with each lower-order one.

> the behaviour of interrupting withRepeat is inconsistent (sometimes it restarts the animation, but once it finishes it doesn't)

This one is strictly paired with `sharedValues`, namely if the animation finishes at the exact same value as it is supposed to begin, then it will correctly reset after a re-render. However, having an animation with a shared value going from `0` to `1` and re-rendering it after it finishes will give us an animation from `1` to `1` (so basically no animation in most cases). I think we shouldn't mess with changing the animation's shared value since it might be used in other components throughout the view and disrupt their behaviour.

src/reanimated2/animation/repeat.ts
```diff
@@ -100,7 +108,14 @@ export const withRepeat = function <T extends AnimationObject>(
         now: Timestamp,
         previousAnimation: Animation<any> | null
       ): void {
-        animation.startValue = value;
+        // Detect re-render
+        const hasBeenInterrupted =
+          value !== initialAnimationBounds.start &&
+          value !== initialAnimationBounds.end;
```
tomekzaw, line 114: What about animations with easing which can go through end point while still in progress?

src/reanimated2/animation/repeat.ts
szydlovsky: This cannot be done, I am getting a 'Tried to synchronously call a non-worklet function on the UI thread'

src/reanimated2/animation/delay.ts
szydlovsky: Same as above

src/reanimated2/animation/repeat.ts
szydlovsky: It is close to impossible to force a re-render in the exact moment that an animation with easing goes through `toValue` point, but I am currently investigating whether we need comparing to `.end` whatsoever.

src/reanimated2/animation/delay.ts
```diff
@@ -31,6 +31,13 @@ export const withDelay = function <T extends AnimationObject>(
   reduceMotion?: ReduceMotion
 ): Animation<DelayAnimation> {
   'worklet';
+
+  const initialAnimationStart =
+    typeof _nextAnimation === 'function'
+      ? _nextAnimation().startValue
+      : _nextAnimation.startValue;
+  let didResetAfterInterruption = false;
```
Latropos, line 39:
```suggestion
// This is code from lines 45-48, it can be removed from lines 45-48 and placed here
  const nextAnimation =
    typeof _nextAnimation === 'function'
      ? _nextAnimation()
      : _nextAnimation;
          
  const initialAnimationStart = nextAnimation.startValue;
```

src/reanimated2/animation/repeat.ts
```diff
@@ -39,6 +39,11 @@ export const withRepeat = function <T extends AnimationObject>(
 ): Animation<RepeatAnimation> {
   'worklet';
 
+  const initialAnimationStart =
+    typeof _nextAnimation === 'function'
+      ? _nextAnimation().startValue
+      : _nextAnimation.startValue;
+
```
Latropos, line 46: // This is code from lines 52-55, it can be removed from lines 52-55 and placed here
  const nextAnimation =
    typeof _nextAnimation === 'function'
      ? _nextAnimation()
      : _nextAnimation;
          
  const initialAnimationStart = nextAnimation.startValue;

src/reanimated2/animation/delay.ts
```diff
@@ -31,6 +31,13 @@ export const withDelay = function <T extends AnimationObject>(
   reduceMotion?: ReduceMotion
 ): Animation<DelayAnimation> {
   'worklet';
+
+  const initialAnimationStart =
```
tomekzaw, line 35: Not sure if I understand the meaning of this variable correctly but maybe let's rename it so it's more consistent with `startValue`?
```suggestion
  const initialAnimationStartValue =
```

src/reanimated2/animation/sequence.ts
```diff
@@ -58,6 +58,12 @@ export function withSequence(
     });
   }
 
+  const initialAnimationStarts = _animations.map((animationObject) => {
```
tomekzaw, line 61:
```suggestion
  const initialAnimationStartValues = _animations.map((animationObject) => {
```

src/reanimated2/animation/delay.ts
tomekzaw, line 39: Is it correct to initialize `didResetAfterInterruption` with `false`? This means we will reset the animation at the beginning so before it could have been interrupted

src/reanimated2/animation/delay.ts
```diff
@@ -62,7 +69,12 @@ export const withDelay = function <T extends AnimationObject>(
           const finished =
             previousAnimation.finished ||
             previousAnimation.onFrame(previousAnimation, now);
-          animation.current = previousAnimation.current;
+          // We need to prevent setting previous animation value when there was an interruption - so that it can reset properly
+          if (didResetAfterInterruption) {
+            didResetAfterInterruption = false;
+          } else {
+            animation.current = previousAnimation.current;
+          }
```
tomekzaw, line 77: What if we negate the logic here and replace `!didResetAfterInterruption` with `shouldResetAfterInterruption`? This way we can simplify the logic and make it easier to understand.
```suggestion
          if (shouldResetAfterInterruption) {
            animation.current = previousAnimation.current;
            shouldResetAfterInterruption = false;
          }
```

src/reanimated2/animation/delay.ts
```diff
@@ -76,9 +88,14 @@ export const withDelay = function <T extends AnimationObject>(
         now: Timestamp,
         previousAnimation: Animation<any> | null
       ): void {
+        // Detect re-render
+        const hasBeenInterrupted = value !== initialAnimationStart;
+        didResetAfterInterruption = hasBeenInterrupted;
+
+        animation.current = hasBeenInterrupted ? initialAnimationStart : value;
```
tomekzaw, line 95:
```suggestion
        animation.current = hasBeenInterrupted ? initialAnimationStartValue : value;
        shouldResetAfterInterruption = ???;
```

src/reanimated2/animation/delay.ts
szydlovsky: I think I should rename it to something like `shouldSkipNextValueSet`. This flag helps us prevent setting current value to the previous one (which makes resetting impossible).

src/reanimated2/animation/delay.ts
```diff
@@ -31,6 +31,13 @@ export const withDelay = function <T extends AnimationObject>(
   reduceMotion?: ReduceMotion
 ): Animation<DelayAnimation> {
   'worklet';
+
+  const initialAnimationStartValue =
+    typeof _nextAnimation === 'function'
+      ? _nextAnimation().startValue
+      : _nextAnimation.startValue;
+  let shouldSkipNextValueSet = false;
```
tjzel, line 39:
```suggestion
  let wasInterrupted = false;
```

src/reanimated2/animation/delay.ts
```diff
@@ -62,7 +69,14 @@ export const withDelay = function <T extends AnimationObject>(
...
             previousAnimation.finished ||
             previousAnimation.onFrame(previousAnimation, now);
-          animation.current = previousAnimation.current;
+
+          // We need to prevent setting previous animation value when there was an interruption - so that it can reset properly
+          if (shouldSkipNextValueSet) {
+            shouldSkipNextValueSet = false;
+          } else {
+            animation.current = previousAnimation.current;
+          }
```
tjzel, line 78:
```suggestion
          if (wasInterrupted) {
            wasInterrupted = false;
          } else {
            animation.current = previousAnimation.current;
          }
```

src/reanimated2/animation/delay.ts
```diff
@@ -76,9 +90,16 @@ export const withDelay = function <T extends AnimationObject>(
         now: Timestamp,
         previousAnimation: Animation<any> | null
       ): void {
+        // Detect re-render
+        const hasBeenInterrupted = value !== initialAnimationStartValue;
+        shouldSkipNextValueSet = hasBeenInterrupted;
+
+        animation.current = hasBeenInterrupted
+          ? initialAnimationStartValue
+          : value;
```
tjzel, line 99:
```suggestion
        // Detect re-render
        const wasInterrupted = value !== initialAnimationStartValue;

        animation.current = wasInterrupted
          ? initialAnimationStartValue
          : value;
```

Also, I don't really like comparing floats like this, but I'll let this slide ⛷️

src/reanimated2/animation/repeat.ts
```diff
@@ -100,7 +105,13 @@ export const withRepeat = function <T extends AnimationObject>(
...
         previousAnimation: Animation<any> | null
       ): void {
-        animation.startValue = value;
+        // Detect re-render
+        const hasBeenInterrupted = value !== initialAnimationStartValue;
+        const startingValue =
+          hasBeenInterrupted && initialAnimationStartValue !== undefined
+            ? initialAnimationStartValue
+            : value;
+        animation.startValue = startingValue;
```
tjzel, line 114:
```suggestion
        const wasInterrupted = value !== initialAnimationStartValue;
        const startingValue =
          wasInterrupted && initialAnimationStartValue !== undefined
            ? initialAnimationStartValue
            : value;
        animation.startValue = startingValue;
```

Also, why do we check for `undefined` here? We don't do it in case of `withSequence` and `withDelay`.

src/reanimated2/animation/sequence.ts
```diff
@@ -140,10 +146,16 @@ export function withSequence(
           ] as SequenceAnimation;
         }
 
+        // Detect re-render
+        const animationStart =
+          initialAnimationStartValues[animation.animationIndex];
+        const hasBeenInterrupted = value !== animationStart;
+        const startingValue = hasBeenInterrupted ? animationStart : value;
+
```
tjzel, line 154:
```suggestion
        const animationStart =
          initialAnimationStartValues[animation.animationIndex];
        const wasInterrupted = value !== animationStart;
        const startingValue = hasBeenInterrupted ? animationStart : value;

```

tjzel (commented):
Looks legit, just need some little clarification on `undefined` here.

src/reanimated2/animation/delay.ts
tomekzaw: https://github.com/software-mansion/react-native-reanimated/pull/5647#discussion_r1488033114 :grief:

Latropos (approved):

PR #5648, opened 2024-02-07
Fix `remoteState` in `useAnimatedStyle`

tjzel:
## Summary

When we merged #5518 we weren't aware that the mechanism of `ShareableHandles` (objects with `__init` worklet property) are memoized on UI thread with the use of `std::unique_ptr` and a map. Because of that `remoteState` currently resets to its initial state after `dependencies` change. This happens, because:

- worklet is stateful during one call of `runOnUI`,
- `startMapper` calls `runOnUI` exactly once,
- worklet has access to its state and can change it (and access changed state) as long as the mapper runs,
- when `dependencies` change, mapper is stopped and started again, so another `runOnUI` is called and previous state is lost
- worklet starts with initial instead 

## Test plan

<details>
<summary>
Test code
</summary>

```tsx
import React from 'react';
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withTiming,

  const [render, setRender] = React.useState(0);
  const [deps, setDeps] = React.useState(0);

  const animatedStyle = useAnimatedStyle(() => ({ width: sv.value }), [deps]);

  function handleRerender() {
    console.log('Rerendering');
    setRender((prev) => prev + 1);
  }

  function handleChangeDeps() {
    console.log('Changing deps & Re-rendering');
    setDeps((prev) => prev + 1);
  }

  function handleAnimate() {
    sv.value = withTiming(Math.random() * 100 + 50, { duration: 1000 });
  }

  return (
      <Button title="Rerender" onPress={handleRerender} />
      <Button title="Animate" onPress={handleAnimate} />
      <Button title="Change deps" onPress={handleChangeDeps} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: {
    height: 100,
    backgroundColor: 'red',
  },
});
```

</details>

Add this to `useAnimatedStyle`:

```diff
diff --git a/src/reanimated2/hook/useAnimatedStyle.ts b/src/reanimated2/hook/useAnimatedStyle.ts
index d6d015572..18dd46d17 100644
--- a/src/reanimated2/hook/useAnimatedStyle.ts
+++ b/src/reanimated2/hook/useAnimatedStyle.ts
@@ -505,6 +505,7 @@ For more, see the docs: \`https://docs.swmansion.com/react-native-reanimated/doc
     } else {
       fun = () => {
         'worklet';
+        console.log(remoteState.last);
         styleUpdater(
           shareableViewDescriptors,
           updaterFn,
```

|Before|After|
|--|--|
|Screenshot 2024-02-07 at 18 20 03|Screenshot 2024-02-07 at 18 21 15|

### DIFF
--- a/src/reanimated2/shareables.ts
+++ b/src/reanimated2/shareables.ts
@@ -329,8 +329,12 @@ export function makeShareableCloneOnUIRecursive<T>(
   return cloneRecursive(value);
 }
 
-export function makeShareable<T extends object>(value: T): T {
-  if (SHOULD_BE_USE_WEB) {
+function makeShareableJS<T extends object>(value: T): T {
+  return value;
+}
+
+function makeShareableNative<T extends object>(value: T): T {
+  if (shareableMappingCache.get(value)) {
     return value;
   }
   const handle = makeShareableCloneRecursive({
@@ -342,3 +346,12 @@ export function makeShareable<T extends object>(value: T): T {
   shareableMappingCache.set(value, handle);
   return value;
 }
+
+/**
+ * This function creates a value on UI with persistent state - changes to it on the UI
+ * thread will be seen by all worklets. Use it when you want to create a value
+ * that is read and written only on the UI thread.
+ */
+export const makeShareable = SHOULD_BE_USE_WEB
+  ? makeShareableJS
+  : makeShareableNative;
--- a/src/reanimated2/hook/useAnimatedStyle.ts
+++ b/src/reanimated2/hook/useAnimatedStyle.ts
@@ -2,7 +2,7 @@
 import type { MutableRefObject } from 'react';
 import { useEffect, useRef } from 'react';
 
-import { startMapper, stopMapper } from '../core';
+import { makeShareable, startMapper, stopMapper } from '../core';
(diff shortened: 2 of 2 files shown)


src/reanimated2/shareables.ts
```diff
@@ -329,8 +329,12 @@ export function makeShareableCloneOnUIRecursive<T>(
...
 }
 
-export function makeShareable<T extends object>(value: T): T {
-  if (SHOULD_BE_USE_WEB) {
+function makeShareableJS<T extends object>(value: T): T {
+  return value;
+}
+
+function makeShareableNative<T extends object>(value: T): T {
+  if (shareableMappingCache.get(value)) {
```
tjzel, line 337: Adding simple optimization here so we won't call `makeShareableCloneRecursive` and `shareableMappingCache` multiple times for no effect. If for safety reason you don't want it to be in this PR, I can move it to separate PR.

Although if `value` was mapped to some other shareable already, then the previous worklet that was using it could get broken with current implementation.

tomekzaw (approved):

piaskowyk (approved):

PR #5663, opened 2024-02-13
Remove unused _REANIMATED_VERSION_BABEL_PLUGIN

tjzel:
As in the title.

🚀

Requires #5667.


app/src/examples/WithoutBabelPluginExample.tsx
```diff
@@ -14,9 +14,10 @@ import { StyleSheet, Text, View } from 'react-native';
 enableExperimentalWebImplementation(true);
 
 function isBabelPluginEnabled() {
-  'inject Reanimated Babel plugin version';
-  // @ts-ignore
-  return global._REANIMATED_VERSION_BABEL_PLUGIN !== undefined;
+  function worklet() {
+    'worklet';
+  }
+  return '__workletHash' in worklet;
```
tomekzaw, line 20:
```suggestion
  function foo() {
    'worklet';
  }
  return isWorklet(foo);
```

tomekzaw (approved):

app/src/examples/WithoutBabelPluginExample.tsx
tjzel: @tomekzaw isWorklet is not a part of public API (should it be?).

PR #5666, opened 2024-02-14
[useAnimatedKeyboard][Android] Fix some keyboard issues

piaskowyk:
## Summary

This pull request addresses a few Android keyboard-related issues:
- It resolves the problem of non-monotonic values during animation. Previously, we were intercepting all keyboard events, but not all of them were relevant to keyboard animation.
- It fixes the issue where the state was incorrect if a transition was canceled.
- It ensures proper respect for auto-hiding navigation bars on Android devices.

## Test plan

Run useAnimatedKeyboard Example on physical Android device.


android/src/main/java/com/swmansion/reanimated/keyboard/KeyboardAnimationCallback.java
```diff
@@ -31,14 +34,31 @@ public WindowInsetsAnimationCompat.BoundsCompat onStart(
   public WindowInsetsCompat onProgress(
       @NonNull WindowInsetsCompat insets,
       @NonNull List<WindowInsetsAnimationCompat> runningAnimations) {
+    boolean isAnyKeyboardAnimation = false;
```
tomekzaw, line 37:
```suggestion
    boolean isAnyKeyboardAnimationRunning = false;
```

android/src/main/java/com/swmansion/reanimated/keyboard/KeyboardAnimationCallback.java
```diff
@@ -31,14 +34,31 @@ public WindowInsetsAnimationCompat.BoundsCompat onStart(
...
+        isAnyKeyboardAnimation = true;
+        break;
+      }
+    }
+    if (!isAnyKeyboardAnimation) {
+      return insets;
+    }
     keyboard.updateHeight(insets);
     notifyAboutKeyboardChange.call();
     return insets;
```
tomekzaw, line 49:
```suggestion

    if (isAnyKeyboardAnimationRunning) {
      keyboard.updateHeight(insets);
      notifyAboutKeyboardChange.call();
    }
    return insets;
```

android/src/main/java/com/swmansion/reanimated/keyboard/KeyboardAnimationCallback.java
```diff
@@ -31,14 +34,31 @@ public WindowInsetsAnimationCompat.BoundsCompat onStart(
...
   }
 
   @Override
   public void onEnd(@NonNull WindowInsetsAnimationCompat animation) {
+    if (!isKeyboardAnimation(animation)) {
+      return;
+    }
     keyboard.onAnimationEnd();
     notifyAboutKeyboardChange.call();
```
tomekzaw, line 58: This is a matter of taste, I know the code with early returns is more maintainable but this is probably easier to understand. Just a suggestion
```suggestion
    if (isKeyboardAnimation(animation)) {
      keyboard.onAnimationEnd();
      notifyAboutKeyboardChange.call();
    }
```

android/src/main/java/com/swmansion/reanimated/keyboard/KeyboardAnimationCallback.java
```diff
@@ -31,14 +34,31 @@ public WindowInsetsAnimationCompat.BoundsCompat onStart(
...
   }
+
+  private boolean isKeyboardAnimation(@NonNull WindowInsetsAnimationCompat animation) {
```
tomekzaw, line 61: This method could be static
```suggestion
  private static boolean isKeyboardAnimation(@NonNull WindowInsetsAnimationCompat animation) {
```

tomekzaw (changes requested):

android/src/main/java/com/swmansion/reanimated/keyboard/Keyboard.java
```diff
@@ -20,15 +23,30 @@ public int getHeight() {
...
+    int keyboardHeight = (int) PixelUtil.toDIPFromPixel(Math.max(0, keyboardHeightDip));
+    if (keyboardHeight == 0 && state == KeyboardState.OPEN) {
+      return;
+    }
+    height = keyboardHeight;
   }
 
   public void onAnimationStart() {
-    state = height == 0 ? KeyboardState.OPENING : KeyboardState.CLOSING;
+    if (activeTransitionCounter > 0) {
```
tjzel, line 37: Please check if when switching between two text inputs rapidly you won't get this even re-triggered even though the keyboard shouldn't close.

tjzel (approved):

android/src/main/java/com/swmansion/reanimated/keyboard/Keyboard.java
piaskowyk: I've checked this and it works 🎉

android/src/main/java/com/swmansion/reanimated/keyboard/Keyboard.java
```diff
@@ -1,34 +1,51 @@
...
     int systemBarBottomInset = insets.getInsets(SYSTEM_BAR_TYPE_MASK).bottom;
-    int keyboardHeightDip = contentBottomInset - systemBarBottomInset;
-    height = (int) PixelUtil.toDIPFromPixel(Math.max(0, keyboardHeightDip));
+    boolean hasNavigationBar = KeyCharacterMap.deviceHasKey(KeyEvent.KEYCODE_HOME);
+    int keyboardHeightDip =
+        hasNavigationBar ? contentBottomInset - systemBarBottomInset : contentBottomInset;
+    int keyboardHeight = (int) PixelUtil.toDIPFromPixel(Math.max(0, keyboardHeightDip));
+    if (keyboardHeight == 0 && mState == KeyboardState.OPEN) {
+      return;
+    }
```
piaskowyk, line 32: TODO: Add comment here

PR #5667, opened 2024-02-14
Expose `isWorklet` in public API

tjzel:
I don't think adding it to the docs is necessary, TSDoc is sufficient IMO.

### DIFF
--- a/src/reanimated2/commonTypes.ts
+++ b/src/reanimated2/commonTypes.ts
@@ -103,8 +103,25 @@ export type WorkletFunction<
 > = ((...args: Args) => ReturnValue) & (WorkletBaseRelease | WorkletBaseDev);
 
 /**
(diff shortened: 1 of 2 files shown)


src/reanimated2/commonTypes.ts
```diff
@@ -103,8 +103,25 @@ export type WorkletFunction<
...
+ *
+ * function myWorklet() {
+ *   'worklet';
+ * };
+ * ```
+ *
+ * ### Maintainer note
  * This function works well on the JS thread performance-wise, since the JIT can inline it.
  * However, on other threads it will not get optimized and we will get a function call overhead.
+ * We want to change in in the future, but it's not feasible at the moment.
```
tomekzaw, line 124:
```suggestion
 * We want to change it in the future, but it's not feasible at the moment.
```

tomekzaw (approved):

Issue #5660, opened 2024-02-12
[3.7.0] Data type not recognized by value unpacker. at valueUnpacker (WorkletRuntime::WorkletRuntime:1:1477)


efstathiosntonas:
### Description

Cannot reproduce it locally, for now I've encountered in production builds on Androids 11,12,13,14. 

Error: 
`Data type not recognized by value unpacker. at valueUnpacker (WorkletRuntime::WorkletRuntime:1:1477)`

Full stacktrace: 

```
﻿﻿com.facebook.jni.CppException: [Reanimated] Data type not recognized by value unpacker.

Error: [Reanimated] Data type not recognized by value unpacker.
    at valueUnpacker (WorkletRuntime::WorkletRuntime:1:1477)
        at com.swmansion.reanimated.AndroidUIScheduler.triggerUI(AndroidUIScheduler.java:-2)
        at com.swmansion.reanimated.AndroidUIScheduler$1.run(AndroidUIScheduler.java:24)
        at com.swmansion.reanimated.AndroidUIScheduler$2.runGuarded(AndroidUIScheduler.java:43)
        at com.facebook.react.bridge.GuardedRunnable.run(GuardedRunnable.java:29)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:250)
        at android.app.ActivityThread.main(ActivityThread.java:7755)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:958)
```

### Steps to reproduce

1. Updated to 3.7.0
2. crashes

### Snack or a link to a repository

no-repo-cannot-reproduce

### Reanimated version

3.7.0

### React Native version

0.73.4

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Release app & production bundle

### Device

Real device

### Device model

LG G8 ThinQ™ (LM-G820), Mi Note 10, Galaxy S22 Ultra (SM-S908U), moto g(8) power

### Acknowledgements

Yes

joe-sam:
[...]

PR #5688, opened 2024-02-19
Extract worklet logic from value unpacker

tomekzaw:
## Summary

This PR moves the logic behind creation of worklets from JS `valueUnpacker` directly into C++ `ShareableWorklet::toJSValue`.

It's a step towards removing `valueUnpacker` which has been causing some problems recently (see #5660).

### DIFF
--- a/Common/cpp/SharedItems/Shareables.cpp
+++ b/Common/cpp/SharedItems/Shareables.cpp
@@ -1,5 +1,7 @@
 #include "Shareables.h"
 
+#include <sstream>
+
 using namespace facebook;
 
 namespace reanimated {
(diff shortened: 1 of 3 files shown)


Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -218,15 +220,80 @@ jsi::Value ShareableHostFunction::toJSValue(jsi::Runtime &rt) {
...
+    }
+
+    workletsCache.getPropertyAsFunction(rt, "set").callWithThis(
+        rt, workletsCache, workletHash, workletFun);
+  }
+
+  const auto workletFunObj = workletFun.asObject(rt);
+  auto functionInstance = workletFunObj.getPropertyAsFunction(rt, "bind")
+                              .callWithThis(rt, workletFunObj, obj);
+  obj.setProperty(rt, "_recur", functionInstance);
```
tomekzaw, line 294: Just wondering if we can run this code once before we add worklet to the cache?

tomekzaw:
Some performance measurements:

main debug 136360 us
main release 86487 us

this PR debug 397071 us
this PR release 242305 us

😢

tjzel:
Haven't had the time to check it yet, but did you verify that the caches are actually utilized? Maybe that's the reason for performance issues.

tomekzaw:
@tjzel You are right, it looks like after the changes the cache does not work as expected.

Before:
```
worklet hash: creation time
17031627990684: 109 us
17031627990684: 5 us
17031627990684: 3 us
17031627990684: 3 us
17031627990684: 3 us
17031627990684: 3 us
17031627990684: 3 us
17031627990684: 3 us
```

After:
```
worklet hash: creation time
17031627990684: 123 us
17031627990684: 110 us
17031627990684: 110 us
17031627990684: 108 us
17031627990684: 107 us
17031627990684: 107 us
17031627990684: 113 us
17031627990684: 106 us
```

tomekzaw:
After:
this PR debug: 122862 us
this PR release: 82824 us

So it's faster 🥳

tomekzaw:
## CWCT (cumulative worklet creation time)

```cpp
#include <iostream>
#include <sstream>
#include <chrono>

using namespace std::chrono;

auto start = high_resolution_clock::now();
// ShareableWorklet::toJSValue
auto end = high_resolution_clock::now();
auto duration = duration_cast<microseconds>(end - start);
static unsigned long acc = 0;
acc += duration.count();
std::cout << static_cast<unsigned long long>(workletHash.asNumber()) << ": " << acc << " us" << std::endl;
```

| | main | this PR |
|-|:-:|:-:|
| **debug** | 122455 us | 122862 us |
| **release** | 82094 us | 82824 us |

So it's just a 1% slower 🥳

tomekzaw:
## WIT (worklet invocation time)

```ts
function worklet() {
  'worklet';
  // do nothing
}
runOnUI(() => {
  const start = performance.now();
  for (let i = 0; i < 10000000; i++) {
    worklet();
  }
  const end = performance.now();
  console.log(end - start, 'ms');
})();
```

| | main | this PR |
|-|:-:|:-:|
| **debug** | 373 us | 371 us |
| **release** | 347 us | 347 us |

PR #5692, opened 2024-02-20
[Example] Fix counter being editable

tjzel:
## Summary

Currently the counter is editable - meaning, the user can put in some value from keyboard - which doesn't make a lot of sense.

## Test plan

🔢


tomekzaw (approved):
`ReText` from `react-native-redash` also sets `underlineColorAndroid="transparent"`: https://github.com/wcandillon/react-native-redash/blob/master/src/ReText.tsx

tjzel:
@tomekzaw I checked and tt doesn't seem to change anything in our case.

PR #5694, opened 2024-02-20
[Example] Fix interrupted animation of Letters in development

tjzel:
## Summary

Currently, in development the animation in Letters gets interrupted near its end and stays like this, which doesn't look right. This doesn't happen in production builds however - it seems to be related to `setInterval` repetition time parameter - making it longer makes the behavior uniform in both development and production.

## Test plan

💌

### DIFF
--- a/app/src/examples/LettersExample.tsx
+++ b/app/src/examples/LettersExample.tsx
@@ -18,16 +18,19 @@ interface LettersProps {
 function Letters({ text }: LettersProps) {
   return (
     <Animated.View style={styles.line}>
-      {splitLetters(text).map(({ char, key }, index) => (
-        <Animated.Text
-          key={key}
-          style={styles.text}
-          entering={FadeIn.duration(500).delay(index)}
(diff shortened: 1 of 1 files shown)


app/src/examples/LettersExample.tsx
```diff
@@ -18,16 +18,19 @@ interface LettersProps {
...
-          key={key}
-          style={styles.text}
-          entering={FadeIn.duration(500).delay(index)}
-          exiting={FadeOut.duration(500).delay(index)}
-          layout={Layout.duration(Math.random() * 700 + 200).delay(index)}>
-          {char}
-        </Animated.Text>
-      ))}
+      {splitLetters(text).map(({ char, key }, index) => {
+        console.log(index);
```
tomekzaw, line 22:
```suggestion
```

tjzel: @tomekzaw fixed, approve pls 
 🥺 
 🙏
🦵🦵

PR #5695, opened 2024-02-20
Rename `isWorklet` to `isWorkletFunction`

tjzel:
## Summary

To avoid confusion with `_WORKLET`.

## Test plan

🤯

### DIFF
--- a/app/src/examples/WithoutBabelPluginExample.tsx
+++ b/app/src/examples/WithoutBabelPluginExample.tsx
@@ -1,5 +1,5 @@
 import Animated, {
-  isWorkletized,
+  isWorkletFunction,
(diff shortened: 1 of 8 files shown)


tomekzaw:
The motivation behind this PR was that the function `isWorklet()` added in #4894 and made public in #5667 could be easily confused with a not-yet-implemented function that checks if the current runtime is a worklet runtime, which was supposed to be given the same name.

After some consideration, that the latter function could be named `isWorkletRuntime()` which is more specific and makes more sense so `isWorklet` is actually a pretty good name (like `Array.isArray` or `Number.isInteger`).

tomekzaw (approved):

PR #5705, opened 2024-02-21
[useAnimatedKeyboard][iOS] Keyboard interpolation

piaskowyk:
## Summary

The `useAnimatedKeyboard()` hook doesn't sync with the keyboard animation because keyboard animations use CAAnimations that are computed on the window server. This means they have their own timer and presentation layer. The info we get about the actual keyboard height is just an approximation of the real size shown by the window server.

Our hook, on the other hand, uses a DisplayLink timer, which is different from the CAAnimation timer. With this PR, we are adding a function that attempts to better estimate the keyboard position using easing and Bezier curves. While this solution might not be perfect down to the pixel because window server animations aren't entirely straightforward, it's still a big improvement compared to before.

The Bezier curves are based on the measured keyboard height from screen recordings, and then I adjusted the curves to fit them better.

Basic estimation function:
image

Where:
- x - time progress of animation
- a1, a2, b1, b2, d1, d3 - curve parameters

Example of curve:

image

Keyboard appearing: https://www.desmos.com/calculator/aw4ocksord
Keyboard hiding: https://www.desmos.com/calculator/6zay2c6jou

Related info: https://forums.developer.apple.com/forums/thread/712762

## The curve parameters

"The curve parameters were calculated using an optimization script that aimed to minimize approximation errors. The script consists of three stages:
- Analyzing the video to compute the keyboard frame during animations
- Calculating the average keyboard height during the animation
- Initial stage of curve approximation with a step of 0.1
- Additional stage of curve approximation with a step of 0.01"

Results:
best_factors_opening: 1, 4.88, 2.59, 9.89, 0.29, 2.45, error: 0.0149869186740412
best_factors_closing: 1, 5.65, 2.74, 8.38, 0.93, 3.29, error: 0.02679889950216266

<details>
<summary>code</summary>

```python
import cv2
import numpy as np

cap = cv2.VideoCapture('video.mp4')
if not cap.isOpened():
  print("Error opening video file")
  exit()

opening_measurements = []
closing_measurements = []
current_measurements = []

last_keyboard_height = 0
state = ""

while cap.isOpened():
  ret, frame = cap.read()
  if not ret:
    break
  frame_height = int(cap.get(cv2.CAP_PROP_FRAME_HEIGHT))
  frame_width = int(cap.get(cv2.CAP_PROP_FRAME_WIDTH))
  gray_frame = cv2.cvtColor(frame, cv2.COLOR_BGR2GRAY)
[...]

Issue #5584, opened 2024-01-15
`useAnimatedKeyboard` does not support floating keyboard (iPad)


renchap:
### Description

On iPads, you can switch the keyboard mode and have a floating keyboard, where the keyboard appears on top on the app (no longer docker at the bottom of the screen) and can be dragged around.

When in floating mode, `useAnimatedKeyboard` incorrectly computes the height, which seems to change depending on the position on the screen.

Please note than on a real device, there are two other Keyboard modes (Undock and split). They do not appear on a Similator iPad pro, I am not sure why, but they probably need to be handled as well.

<img width="1080" alt="image" src="https://github.com/software-mansion/react-native-reanimated/assets/42070/e0377ceb-a961-4706-986c-c4421045d7c7">

### Steps to reproduce

1. Open and compile the Example app from the repro branch below using an iPad in simulator (or an iPad device) as the target. Note that I configured the Example app to support iPad devices in this fork
2. Open the `useAnimatedKeyboard` example
3. Tap the text field at the bottom
4. Long-press the Keyboard + bottom arrow icon on the bottom right of the keyboard
5. Choose "floating"
6. Move the keyboard. The text field disappear, and you can see in Metro console that the height is now completely invalid

### Snack or a link to a repository

https://github.com/renchap/react-native-reanimated/tree/use-animated-keyboard-floating-bug

### Reanimated version

3.6.0

### React Native version

0.72.6

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

Real device

### Device model

Any iPad I tested

### Acknowledgements

Yes

renchap:
I am trying to write a PR for this, but I never wrote ObjectiveC before and I am struggling to find something that works.

From what I have been able to gather:
- `UIKeyboardWillShowNotification` and `UIKeyboardWillHideNotification ` should not be dispatched when the keyboard is in floating mode, so those events could be used to subscribe to keyboard events only when needed
- Other people are advising to check for the keyboard width, and consider that if the width is the width of the screen, then the keyboard is floating. But I suspect this will cause issues with Split View or other modes (like split keyboard)

szydlovsky:
[...]

PR #5710, opened 2024-02-22
Fix useAnimatedKeyboard hook for iPad detached keyboard (floating/undocked/split)

szydlovsky:
## Summary

According to https://github.com/software-mansion/react-native-reanimated/issues/5584, `useAnimatedKeyboard` returns incorrect height when iPad keyboard is in detached mode (so either floating, undocked or split). Moreover, there is no way to tell if the keyboard is detached in order to e.g. ignore emitted values. This PR tries to address the issues by:

1.  Adding new KeyboardState case `DETACHED`, which is the hook's state as long as the keyboard is detached (this state's raw integer value is 5)
2. Returning `0` height value as long as KeyboardState returned by hook is `DETACHED`

This way the user can listen for state changes and the hook doesn't emit values disrupting behaviour dependent on it (e.g. at some point it would return entire window's height as keyboard height).

iPad simulator supports only `FLOATING` out of these keyboard modes. `SPLIT` and `UNDOCKED` need physical device.
| before | after |
| --- | --- |
| <video src="https://github.com/software-mansion/react-native-reanimated/assets/77503811/29bad57e-c6d3-469a-8b37-10c80c63d02b" /> | <video src="https://github.com/software-mansion/react-native-reanimated/assets/77503811/17fdb2ac-83a9-4208-8c73-7662a55f1507" /> |

This PR will be a draft until https://github.com/software-mansion/react-native-reanimated/pull/5705 is merged, because it substantially changes the logic in REAKeyboardEventObserver and this PR will need some changes.

## Test plan

Reanimated's example app needs some additional setup to properly run fullscreen-mode on iPad (both simulator and physical device):

1. Add a few properties to `/Example/ios/ReanimatedExample.xcodeproj/project.pbxproj` file:

- lines 557 - 561 should look like this:
```
				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
				SUPPORTS_MACCATALYST = NO;
				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
				SWIFT_VERSION = 5.0;
				TARGETED_DEVICE_FAMILY = "1,2";
```

- lines 586 -  589 should look like this:
```
				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
				SUPPORTS_MACCATALYST = NO;
				SWIFT_VERSION = 5.0;
				TARGETED_DEVICE_FAMILY = "1,2";
```
2. Add a console log to `/app/src/examples/AnimatedKeyboardExample.tsx` to see output of the hook, `translateStyle` code might look this way:
```typescript
  console.log(
    `STATE: ${keyboard.state.value}, HEIGHT: ${keyboard.height.value}`
  );
  return {
  };
});
```
3. Run Example on some iPad simulator (only floating keyboard) or physical iPad device (all of the mentioned options) and observe logged values in console.


szydlovsky:
With the new REAKeyboardEventObserver file logic, it is either possible to correctly handle detached keyboards or correctly return keyboard height with interactive dismiss mode. Never two at the same time. Hence, closing the PR for now.

PR #5716, opened 2024-02-22
Move Reanimated Babel Plugin dependencies to dependencies

tjzel:
## Summary

After thinking about this, there's no reason for us to keep up the weird situation, with actually used Babel plugins in peer dependencies.

Reasons for that are:
- those plugins are supposed to perform a given transformation, regardless of their version and whether they are proposals or not. 
- React Native is currently in progress of changing those proposal plugins to accepted ones, it would give it unnecessary maintenance to check which are available.

## Test plan

I checked that our Example app works properly, there are no differences in Jest tests and also Reanimated works with Expensify's app.


package.json
```diff
@@ -91,31 +91,23 @@
   },
   "homepage": "https://github.com/software-mansion/react-native-reanimated#readme",
   "dependencies": {
-    "@babel/plugin-transform-object-assign": "^7.16.7",
+    "@babel/plugin-transform-arrow-functions": "^7.23.3",
+    "@babel/plugin-transform-nullish-coalescing-operator": "^7.23.4",
+    "@babel/plugin-transform-optional-chaining": "^7.23.4",
+    "@babel/plugin-transform-shorthand-properties": "^7.23.3",
+    "@babel/plugin-transform-template-literals": "^7.23.3",
```
tomekzaw, line 98: Can we use `^7.0.0-0` for all of them?

tjzel: Smart, will do.

piaskowyk (approved):
Good for me 🫡

tomekzaw (approved):
Let it be

PR #5720, opened 2024-02-23
Fix eslint config

tjzel:
## Summary

Because we use ESlint rules that require a parses, proper amendments to the `.eslintrc.js` had to be made. Because of that we can also finally extend main `.eslintrc.js` config in sub-directiories.

You can read more [here](https://stackoverflow.com/questions/58510287/parseroptions-project-has-been-set-for-typescript-eslint-parser).

As a note, I recommend `eslint --print-config <file>` to debug if the rules are actually extended etc.

## Test plan

🚀


plugin/package.json
```diff
@@ -2,8 +2,11 @@
   "devDependencies": {
     "@react-native/eslint-config": "^0.72.1",
     "@types/node": "^18.15.11",
+    "@typescript-eslint/eslint-plugin": "^7.0.2",
+    "@typescript-eslint/parser": "^7.0.2",
     "esbuild": "^0.17.11",
-    "eslint": "^8.35.0",
+    "eslint": "^8.56.0",
```
Latropos, line 8: Could you update all eslint dependencies in each of our `package.json`s? So we have the same version in all the examples:
image

Also `@typescript-eslint/eslint-plugin` in main folder has version `"^6.19.0"`, could you make sure all the versions match?

Probably this change should be done in separate PR preceding this one

tjzel: We should probably use ^8.0.0-0 everywhere for simplicity.

tjzel:
@Latropos done!

Latropos (approved):
LGTM, thank you for your change!
Are you sure, that all the rules (especially our custom one) keep working? The newest eslint v9 introduced some breaking changes.
Also probably in the future we will have to migrate to newer config format.

tjzel:
@Latropos you were right, I actually had to downgrade it to ESlint 8 due to breaking changes 🤣

PR #5723, opened 2024-02-23
Delete removed methods from ReanimatedUIManager

tboba:
## Summary

Since RN 0.74.0-rc.0, methods `replaceExistingNonRootView` and `removeSubviewsFromContainerWithID` have been removed from UIManagerModule. This means that they cannot be also used in ReanimatedUIManager anymore. Even more, apps that are using Reanimated cannot be built on Android, because of the error about undeclared symbols. Since I couldn't find any places where these methods were executed, I've decided to remove them.

## Test plan

I believe that checking if examples from Example app work properly should be sufficient.


piaskowyk (changes requested):
Good changes! Could you check if it works for older react native version too? (0.72, 0.73)

tboba:
@piaskowyk I just tested layout animations on Example app (which has react-native version 3.7.1 installed) and created a fresh RN app with 0.72 with HabitsExample and I can't find any issues there!

tomekzaw:
This will be fixed in #5734 which also introduces a new sourceset for RN 0.74+.

piaskowyk:
> This will be fixed in #5734 which also introduces a new sourceset for RN 0.74+.

But thanks for the PR 😊

PR #5724, opened 2024-02-23
Make `makeShareableCloneRecursive` more readable

tjzel:
## Summary

Refactoring `makeShareableCloneRecursive` so it would be actually readable now.

The behavior of the function (and it's flow) shouldn't change.

## Test plan

- [x] Runtime tests


packages/react-native-reanimated/src/shareables.ts
```diff
@@ -100,14 +101,72 @@ const DETECT_CYCLIC_OBJECT_DEPTH_THRESHOLD = 30;
...
+  }
+  if (isFunction && !isWorkletFunction(value)) {
+    // this is a remote function
+    return cloneRemoteFunction(value, shouldPersistRemote);
   }
+  if (isHostObject(value)) {
+    return cloneHostObject(value, shouldPersistRemote);
+  }
+  if ((isPlainJSObject(value) || isFunction) && isWorkletFunction(value)) {
+    return cloneWorklet(value, shouldPersistRemote, depth);
```
bartlomiejbloniarz, line 141: wrong order

packages/react-native-reanimated/src/shareables.ts
```diff
@@ -124,176 +183,225 @@ export function makeShareableCloneRecursive<T>(
...
+
+function cloneHostObject<T extends object>(
+  value: T,
+  shouldPersistRemote: boolean
+): ShareableRef<T> {
+  // for host objects we pass the reference to the object as shareable and
+  // then recreate new host object wrapping the same instance on the UI thread.
+  // there is no point of iterating over keys as we do for regular objects.
+  const clone = ReanimatedModule.makeShareableClone(value, shouldPersistRemote);
+  shareableMappingCache.set(value, clone);
```
bartlomiejbloniarz, line 234: missing `shareableMappingCache.set(clone);`

packages/react-native-reanimated/src/shareables.ts
```diff
@@ -124,176 +183,225 @@ export function makeShareableCloneRecursive<T>(
...
+): ShareableRef<T> {
+  return ReanimatedModule.makeShareableClone(value, shouldPersistRemote);
+}
+
+function cloneArray<T extends unknown[]>(
+  value: T,
+  shouldPersistRemote: boolean,
+  depth: number
+): ShareableRef<T> {
+  const NAME_ME = value.map((element) =>
```
bartlomiejbloniarz, line 200: name them :((

packages/react-native-reanimated/src/shareables.ts
```diff
@@ -124,176 +183,225 @@ export function makeShareableCloneRecursive<T>(
...
-        const flags = value.flags;
-        const handle = makeShareableCloneRecursive({
-          __init: () => {
-            'worklet';
-            return new RegExp(pattern, flags);
-          },
-        });
-        shareableMappingCache.set(value, handle);
-        return handle as ShareableRef<T>;
-      } else if (value instanceof Error) {
```
bartlomiejbloniarz, line 222: where error

bartlomiejbloniarz (changes requested):

packages/react-native-reanimated/src/shareables.ts
tjzel: [94f9b37](https://github.com/software-mansion/react-native-reanimated/pull/5724/commits/94f9b377f51c545f93a72a8e6bf09497ebb09ed0)

packages/react-native-reanimated/src/shareables.ts
tjzel: [94f9b37](https://github.com/software-mansion/react-native-reanimated/pull/5724/commits/94f9b377f51c545f93a72a8e6bf09497ebb09ed0)

packages/react-native-reanimated/src/shareables.ts
tjzel: [94f9b37](https://github.com/software-mansion/react-native-reanimated/pull/5724/commits/94f9b377f51c545f93a72a8e6bf09497ebb09ed0)

bartlomiejbloniarz (approved):
🚀

packages/react-native-reanimated/src/shareables.ts
```diff
@@ -124,176 +192,243 @@ export function makeShareableCloneRecursive<T>(
...
+  return handle as unknown as ShareableRef<T>;
+}
+
+function cloneArrayBuffer<T extends ArrayBuffer>(
+  value: T,
+  shouldPersistRemote: boolean
+): ShareableRef<T> {
+  const clone = WorkletsModule.makeShareableClone(value, shouldPersistRemote);
+  shareableMappingCache.set(value, clone);
+  shareableMappingCache.set(clone);
```
piaskowyk, line 383: This part of the code repeats a few times, so it might be a good idea to move it to a separate function. However, I'm concerned it might make the code less readable. What do you think?

piaskowyk (commented):
Did you run the RuntimeTests against those changes?

packages/react-native-reanimated/src/shareables.ts
tjzel: In the OG code sometimes there's

```ts
shareableMappingCache.set(value, clone);
shareableMappingCache.set(clone);
```

And sometimes only

```ts
shareableMappingCache.set(value, clone);
```

The reason for this is unclear (I guess somebody just forgot), but let's address it in a separate pull request. Leaving it as is now hints us that there's a problem and that's useful in this case IMO.

tjzel: @piaskowyk Runtime tests behave exactly as on the `main` branch 🥳

piaskowyk (approved):

Issue #5731, opened 2024-02-26
Use `useMemo` for reference stability instead of `useRef`


tjzel:
### Description

Currently, Reanimated uses a pattern of `useRef` with follow-up initialization for memoization:

```js
function useCustomHook(){
  const ref = useRef(null);
  if (ref.current === null){
    ref.current = initialize();
  };
};
```

---

List of files that need the aforementioned change:

- [ ] ViewDescriptorSet.ts #5732 
- [ ] useAnimatedKeyboard.ts 
- [ ] useAnimatedRef.ts
- [ ] useAnimatedStyle.ts
- [ ] useDerivedValue.ts
- [ ] useEvent.ts (and all other hooks that use it):
    1. useAnimatedScrollHandler.ts
    2. useScrollViewOffset.ts
    3. useAnimatedGestureHandler.ts (deprecated)
- [ ] useHandler.ts
    1. useAnimatedScrollHandler.ts
    2. useAnimatedGestureHandler.ts (deprecated)
- [ ] useScrollViewOffset.ts

---
We should also asses if memoization if those files is acceptable:

- [ ] FlatList.tsx
- [ ] useAnimatedSensor.ts
- [ ] LayoutAnimationConfig.ts
- [ ] useFrameCallback.ts

---

Then we will have to assure that all our hooks return new reference had some of their dependencies changed as it can happen for example in https://github.com/software-mansion/react-native-reanimated/issues/5364.

### Steps to reproduce

n/a

### Snack or a link to a repository

n/a

### Reanimated version

n/a

### React Native version

n/a

### Platforms

Android, iOS, Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

tjzel:
Closing since it seems to be irrelevant.

Rexogamer:
looks like this issue wasn't closed?

tjzel:
@Rexogamer I must've been very tired already 😄

PR #5732, opened 2024-02-26
Use `useMemo` in `useViewRefSet`

tjzel:
## Summary

This PR is a part of #5731.

## Test plan

Use the following and see that no exceptions are thrown in WebExample (this change doesn't affect Native).

<details>
<summary>
testing diff
</summary>

```patch
diff --git a/src/reanimated2/hook/useAnimatedStyle.ts b/src/reanimated2/hook/useAnimatedStyle.ts
index 6b74e82cb..49d631944 100644
--- a/src/reanimated2/hook/useAnimatedStyle.ts
+++ b/src/reanimated2/hook/useAnimatedStyle.ts
@@ -424,6 +424,12 @@ export function useAnimatedStyle<Style extends DefaultStyle>(
   isAnimatedProps = false
 ): AnimatedStyleHandle<Style> | JestAnimatedStyleHandle<Style> {
   const viewsRef: ViewRefSet<unknown> | undefined = useViewRefSet();
+  const viewsRefRef = useRef(viewsRef);
+  if (viewsRefRef.current !== viewsRef) {
+    throw new Error('TEST FAILED');
+  } else {
+    console.log('TEST PASSED');
+  }
   const animatedUpdaterData = useRef<AnimatedUpdaterData>();
   let inputs = Object.values(updater.__closure ?? {});
   if (SHOULD_BE_USE_WEB) {
```

</details>

### DIFF
--- a/src/reanimated2/ViewDescriptorsSet.ts
+++ b/src/reanimated2/ViewDescriptorsSet.ts
@@ -1,5 +1,5 @@
 'use strict';
-import { useRef } from 'react';
+import { useMemo } from 'react';
 import { makeMutable } from './core';
 import type { SharedValue } from './commonTypes';
 import type { Descriptor } from './hook/commonTypes';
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):

tjzel:
On second thought, I think this PR violated rules of React once again. We modify the object returned by `useMemo` which doesn't seem to be problematic, unless the cache will actually get dropped. Then we will lose all the registered views and lose reactivity. Should we do this instead?

```ts
function useViewRefSetJS<T>(): ViewRefSet<T> {
  const viewRefSet = useRef<ViewRefSet<T>>({
    viewSet: new Set(),
    add(item: T) {
      this.viewSet.add(item);
    },

    remove(item: T) {
      this.viewSet.delete(item);
    },
  });

  return viewRefSet.current;
}
```

kmagiera (commented):
I believe that this particular instance should actually rely on `useState` as it is important that the list of views the style is attached to doesn't get thrown away. The semantics of `useMemo` is that it should be used as a performance optimization, meaning we should use it in places where it is safe for whatever it returns to re-run. In this case we want this hook to return a container that we will put some values in

kmagiera:
After thinking a bit more about this one, it feels like it should really be just ref. In this particular instance it isn't used for rendering (the result isn't used for rendering) and we really never update the value itself. There is a similar pattern presented in the official docs so I'd assume that for this usecase we can actually keep it this way: https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents

tjzel:
Closing since it seems to be irrelevant.

PR #5733, opened 2024-02-27
Correctly handle passing null and undefined as animatedProps

Latropos:
## Summary
Correctly handle passing null and undefined as animatedProps.
The chosen solution was to update invalid value of animatedProps inside the constructor of createAnimatedComponent. However we are unable to modify props (even in constructor) so I'm creating a copy of props (named `safeProps`) which can be modified and should be used instead of props.
## Test plan

Try passing null as animatedProps, I used our planet example for testing 🪐


src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -135,8 +135,23 @@ export function createAnimatedComponent(
...
       super(props);
+
+      const animatedProps = this.props.animatedProps;
+
+      this.safeProps =
+        animatedProps === null || animatedProps === undefined
+          ? { animatedProps: {} }
+          : {
+              animatedProps,
+            };
```
piaskowyk, line 153: can we write it like that?
```suggestion
          : { animatedProps };
```

piaskowyk (dismissed):

PR #5740, opened 2024-02-28
[useAnimatedKeyboard][Android] Filter keyboard events

piaskowyk:
## Summary

It resolves the problem of non-monotonic values during animation. Previously, we were intercepting all keyboard events, but not all of them were relevant to keyboard animation.


tomekzaw (approved):
Looks good to me!

PR #5741, opened 2024-02-28
[useAnimatedKeyboard][Android] Fix state after animation dismiss

piaskowyk:
## Summary

It fixes the issue where the state was incorrect if a keyboard transition was canceled.


android/src/main/java/com/swmansion/reanimated/keyboard/Keyboard.java
```diff
@@ -20,15 +21,31 @@ public int getHeight() {
...
+    int keyboardHeightDip = Math.max(0, contentBottomInset - systemBarBottomInset);
+    if (keyboardHeightDip == 0 && mState == KeyboardState.OPEN) {
+      /*
+      When the keyboard is being canceling, for one frame the insets show a keyboard height of 0,
+      causing a jump of the keyboard. We can avoid it by ignoring that frame and calling
+      the listeners on the following frame.
+      */
+      return;
+    }
+    mHeight = (int) PixelUtil.toDIPFromPixel(keyboardHeightDip);
```
tomekzaw, line 33: This change looks similar to the one from #5742, is it supposed to be in this PR?

piaskowyk: This change may seem minor, but it's crucial to implement in this PR to ensure correct state management.

It sets the keyboard height to 0, so when the next animation begins, in the `onAnimationStart` method, we will switch the state to `opening` instead of `closing`.

PR #5742, opened 2024-02-28
[useAnimatedKeyboard][Android] Fix navigation bar with auto hide

piaskowyk:
## Summary

It ensures proper respect for auto-hiding navigation bars on Android devices.


android/src/main/java/com/swmansion/reanimated/keyboard/Keyboard.java
```diff
@@ -21,8 +23,11 @@ public int getHeight() {
   public void updateHeight(WindowInsetsCompat insets) {
     int contentBottomInset = insets.getInsets(CONTENT_TYPE_MASK).bottom;
     int systemBarBottomInset = insets.getInsets(SYSTEM_BAR_TYPE_MASK).bottom;
-    int keyboardHeightDip = Math.max(0, contentBottomInset - systemBarBottomInset);
-    if (keyboardHeightDip == 0 && mState == KeyboardState.OPEN) {
+    boolean hasNavigationBar = KeyCharacterMap.deviceHasKey(KeyEvent.KEYCODE_HOME);
+    int keyboardHeightDip =
+        hasNavigationBar ? contentBottomInset - systemBarBottomInset : contentBottomInset;
+    int keyboardHeight = (int) PixelUtil.toDIPFromPixel(Math.max(0, keyboardHeightDip));
```
tomekzaw, line 29: Could you please specify when exactly `keyboardHeightDip` is less than zero? Can we somehow catch this phenomenon by setting a breakpoint inside `if (keyboardHeightDip < 0)`?

tomekzaw (approved):

android/src/main/java/com/swmansion/reanimated/keyboard/Keyboard.java
piaskowyk: Sure! This situation is described in comment here:
https://github.com/software-mansion/react-native-reanimated/pull/5741/files#diff-f6b59cc981edc898c89a7edf397f87eae2874bdd13b388b976f9adb71d7f7629R27-R29
>When the keyboard is being canceling, for one frame the insets show a keyboard height of 0,
      causing a jump of the keyboard. We can avoid it by ignoring that frame and calling
      the listeners on the following frame.

PR #5743, opened 2024-02-28
Make `useEvent` return a different reference if event handler should be reattached

### LINKED ISSUE #5364: Stale closure with useAnimatedScrollHandler when passed to memo()
### Description

In idiomatic React, it is presumed that it is safe to skip re-rendering if all props are referentially the same. Built-in optimizations like `memo` rely on that, and the upcoming React compiler will rely on that even stronger.

It seems like `useAnimatedScrollHandler` breaks this constraint. I haven't looked at how it works, but it seems like it retains referential identity but relies on re-rendering to swap out its callback. This leads to non-deterministic behavior — add `memo()` anywhere in the middle of the path of `onScroll` being passed down, and the callback will have a stale closure.

This probably points to breaking the rules of React (e.g. mutation in render).

### Steps to reproduce

Code:

```js
import * as React from 'react'
import {View} from 'react-native'
import Animated, {useAnimatedScrollHandler} from 'react-native-reanimated'

  const [x, setX] = React.useState(0)
  React.useEffect(() => {
    let id = setInterval(() => setX(x => x + 1), 1000)
    return () => clearInterval(id)
  }, [])

  const handler = useAnimatedScrollHandler({
    onScroll() {
      console.log(x)
    },
  })

  return <Child onScroll={handler} />
}

const Child = React.memo(function ({onScroll}) {
  return (
    <View style={{flex: 1, flexDirection: 'row'}}>
      <Animated.FlatList
        onScroll={onScroll}
        style={{flex: 1, backgroundColor: 'red'}}
        contentContainerStyle={{
          paddingTop: 1000,
        }}
      />
    </View>
  )
})

```

Note the `x` never updates.

Then remove `React.memo` around `Child`.

Now `x` updates.

## Runnable demo

```
git clone https://github.com/bluesky-social/social-app.git
cd social-app

git checkout reanimated-stuck-onscroll-repro-memo
yarn
yarn start --clear
```

The repro is in `App.native.tsx`.

This definitely reproes on iOS and Android. Haven't checked web.

https://github.com/bluesky-social/social-app/tree/reanimated-stuck-onscroll-repro-memo

### Reanimated version

3.4.3

### React Native version

0.72.5

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

tjzel:
## Summary

Currently Reanimated relies on reattaching event handlers on render. However, we give any notice to React that it should occur, because `useEvent` always returns the same reference.

With these changes, the reference for a given handler will change if it had to be rebuild and components using `memo` will get re-rendered.

Fixes #5364 

## Test plan

See added `MemoExample.tsx`.

### DIFF
--- a/src/reanimated2/hook/useEvent.ts
+++ b/src/reanimated2/hook/useEvent.ts
@@ -1,5 +1,4 @@
 'use strict';
-import type { MutableRefObject } from 'react';
 import { useRef } from 'react';
 import WorkletEventHandler from '../WorkletEventHandler';
 import type { ReanimatedEvent } from './commonTypes';
@@ -17,9 +16,9 @@ export type EventHandlerProcessed<
   Context extends Record<string, unknown> = never
 > = (event: Event, context?: Context) => void;
 
-export type EventHandlerInternal<Event extends object> = MutableRefObject<
-  WorkletEventHandler<Event>
->;
+export type EventHandlerInternal<Event extends object> = {
+  workletEventHandler: WorkletEventHandler<Event>;
+};
 
 /**
  * Lets you run a function whenever a specified native event occurs.
@@ -49,13 +48,16 @@ export function useEvent<Event extends object, Context = never>(
   eventNames: string[] = [],
   rebuild = false
 ): EventHandlerInternal<Event> {
-  const initRef = useRef<WorkletEventHandler<Event> | null>(null);
+  const initRef = useRef<EventHandlerInternal<Event>>(null!);
   if (initRef.current === null) {
-    initRef.current = new WorkletEventHandler<Event>(handler, eventNames);
+    initRef.current = {
+      workletEventHandler: new WorkletEventHandler<Event>(handler, eventNames),
+    };
   } else if (rebuild) {
-    initRef.current.updateWorklet(handler);
+    const workletEventHandler = initRef.current.workletEventHandler;
+    workletEventHandler.updateWorklet(handler);
+    initRef.current = { workletEventHandler };
   }
 
-  // We cast it since we don't want to expose initial null value outside.
-  return initRef as EventHandlerInternal<Event>;
+  return initRef.current;
 }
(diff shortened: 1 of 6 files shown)


piaskowyk (approved):

src/reanimated2/hook/useEvent.ts
```diff
@@ -49,13 +48,18 @@ export function useEvent<Event extends object, Context = never>(
...
+    const workletEventHandler = new WorkletEventHandler<Event>(
+      handler,
+      eventNames
+    );
+    initRef.current = { workletEventHandler };
   } else if (rebuild) {
-    initRef.current.updateWorklet(handler);
+    const workletEventHandler = initRef.current.workletEventHandler;
+    workletEventHandler.updateWorklet(handler);
+    initRef.current = { workletEventHandler };
```
gaearon, line 61: Is there any way that reassignment during render can be avoided? This still violates React's rules — it means a render is "observable" even if it has not been committed. In React, this can happen with newer features like Transitions. If you update state in a Transition, but this Transition has not completed yet, it should *not* reflect in the user interface. However, assigning a mutable value during rendering and then reading from that value means that the change from an uncommitted Transition render would reflect for any further re-renders (even before this Transition commits — or even if it never ends up being committed).

tjzel: @gaearon I think we could reassign with `useEffect` here, would that be good enough?

As a side note, I'm a bit confused with all these React features. From what you're saying it turns out that now you shouldn't assign to Refs during render - meaning that everything has to be pushed to `useEffect`. Is this intentional or am I missing some information here?

gaearon: It's not really new — in React, rendering is supposed to be pure. This means that the fact that something was rendered should be unobservable and shouldn't influence rendering output of other renders. 

Do these look helpful? https://react.dev/reference/rules

gaearon: Also here: https://react.dev/reference/react/useRef

Screenshot 2024-05-29 at 00 08 39

tjzel: Damn, now I feel stupid 😅

PR #5744, opened 2024-02-28
Prevent multiple registration of worklet event handlers

tjzel:
## Summary

Currently `WorkletEventHandler` can only be attached to a single component - however, we don't check for that and it can lead to issues when using Reanimated. This PR adds QoL warning if the handler is attached to multiple components.

## Test plan

TBD


piaskowyk:
After changes in:
- https://github.com/software-mansion/react-native-reanimated/pull/5845
- https://github.com/software-mansion/react-native-reanimated/pull/6030
That changes are no longer relevant, so we we can close this one 🫡

PR #5751, opened 2024-03-01
feat: migrate to turbomodule

WoLewicki:
## Summary

PR converting the module to TurboModule.


android/src/paper/java/com/swmansion/reanimated/NativeReanimatedModuleSpec.java
```diff
@@ -0,0 +1,36 @@
...
+ * This code was generated by [react-native-codegen](https://www.npmjs.com/package/react-native-codegen).
+ *
+ * Do not edit this file as changes may cause incorrect behavior and will be lost
+ * once the code is regenerated.
+ *
+ * @generated by codegen project: GenerateModuleJavaSpec.js
+ *
+ * @nolint
+ */
+
```
tomekzaw, line 11: Is this file really generated by react-native-codegen or just copied from somewhere?

apple/REAModule.h
```diff
@@ -1,14 +1,24 @@
+#ifdef RCT_NEW_ARCH_ENABLED
+#import <rnreanimated/rnreanimated.h>
+#else
 #import <React/RCTBridgeModule.h>
+#endif
```
tomekzaw, line 5: Can we add a newline here for better readability?
```suggestion
#endif // RCT_NEW_ARCH_ENABLED

```

apple/REAModule.h
```diff
@@ -1,14 +1,24 @@
...
-
-@interface REAModule : RCTEventEmitter <RCTBridgeModule, RCTEventDispatcherObserver, RCTUIManagerObserver>
+@interface REAModule : RCTEventEmitter
+#ifdef RCT_NEW_ARCH_ENABLED
+                       <NativeReanimatedModuleSpec,
+#else
+                       <RCTBridgeModule,
+#endif
+                        RCTEventDispatcherObserver,
+                        RCTUIManagerObserver>
```
tomekzaw, line 21: I'm wondering if this clever code is parsed correctly by Xcode intellisense; it certainly does not work well in GitHub syntax highlighting. How about the following?

```suggestion
#ifdef RCT_NEW_ARCH_ENABLED
typedef NativeReanimatedModuleSpec BaseModule;
#else
typedef RCTBridgeModule BaseModule;
#endif // RCT_NEW_ARCH_ENABLED

@interface REAModule : RCTEventEmitter <BaseModule, RCTEventDispatcherObserver, RCTUIManagerObserver>
```

src/reanimated2/NativeReanimated/NativeReanimated.ts
```diff
@@ -10,6 +9,7 @@ import { jsVersion } from '../platform-specific/jsVersion';
 import type { WorkletRuntime } from '../runtimes';
 import { getValueUnpackerCode } from '../valueUnpacker';
 import type { LayoutAnimationBatchItem } from '../layoutReanimation/animationBuilder/commonTypes';
+import NativeReanimatedModule from '../../specs/NativeReanimatedModule';
```
tomekzaw, line 12: Actually, `Native` prefix in `NativeReanimatedModule.ts` is just for react-native-codegen to detect this file. The actual name of the module is `ReanimatedModule` so maybe let's import it under that name?

Also, `NativeReanimatedModule` already represents something else in our codebase so this ambiguity might be confusing.

```suggestion
import ReanimatedModule from '../../specs/NativeReanimatedModule';
```

src/reanimated2/NativeReanimated/NativeReanimated.ts
```diff
@@ -90,9 +90,8 @@ export class NativeReanimated {
     }
     global._REANIMATED_VERSION_JS = jsVersion;
     if (global.__reanimatedModuleProxy === undefined) {
-      const { ReanimatedModule } = NativeModules;
       const valueUnpackerCode = getValueUnpackerCode();
-      ReanimatedModule?.installTurboModule(valueUnpackerCode);
+      NativeReanimatedModule?.installTurboModule(valueUnpackerCode);
```
tomekzaw, line 94: This is probably out of the scope of this PR but it's quite confusing that we call `installTurboModule` method on an existing turbo module. We should probably rename `installTurboModule` to `installJSIBindings`. Alternatively, we could even use `ReanimatedModule` directly instead of `this.InnerNativeModule` (but let's do it in a separate PR).

tomekzaw (changes requested):
I'm getting the following warnings in Example app running on Android:

Screenshot 2024-03-04 at 10 40 57

Screenshot 2024-03-04 at 10 39 12

It works correctly on main:

Screenshot 2024-03-04 at 10 43 15

android/build.gradle
```diff
@@ -130,6 +130,10 @@ def toPlatformFileString(String path) {
   return path
 }
 
+if (isNewArchitectureEnabled()) {
+    apply plugin: "com.facebook.react"
+}
+
```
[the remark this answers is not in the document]
WoLewicki: I believe it does: https://github.com/software-mansion/react-native-gesture-handler/blob/65578672d09993123fb02b27bbdbdb24d928c396/android/build.gradle#L58. It is needed for compilation, not sure for what exactly, but it is listed in the transition docs: https://github.com/reactwg/react-native-new-architecture/blob/main/docs/turbo-modules.md#the-buildgradle-file

android/src/paper/java/com/swmansion/reanimated/NativeReanimatedModuleSpec.java
WoLewicki: It is copied from the `build` folder generated by codegen when using new arch. We don't change those files for better diffs is we changed anything in the spec.

apple/REAModule.h
WoLewicki: Ok I'll add it everywhere

apple/REAModule.h
WoLewicki: I works well in Xcode but I can change it.

WoLewicki: Hmm it looks like it has some problems with such syntax: 
image

android/build.gradle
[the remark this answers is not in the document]
tomekzaw: You're right, I was searching for `apply plugin: "com.facebook.react"` and forgot that apostrophes exist.

src/reanimated2/NativeReanimated/NativeReanimated.ts
WoLewicki: >  We should probably rename installTurboModule to installJSIBindings.

I can do it for this PR and the rest of the work in the next one. Let's merge it ASAP.

tomekzaw (approved):
Tested on the following configurations:
* ✅ Example on Android
* ✅ Example on iOS
* ✅ FabricExample on Android
* ✅ FabricExample on iOS
* ✅ WebExample

piaskowyk (approved):

PR #5752, opened 2024-03-01
fix: don't use deprecated onCatalystInstanceDestroy API

WoLewicki:
## Summary

PR moving from deprecated method `onCatalystInstanceDestroy` to `invalidate` as suggested by the core.

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -109,13 +109,13 @@ public class NodesManager implements EventDispatcherListener {
     return mAnimationManager;
   }
 
-  public void onCatalystInstanceDestroy() {
+  public void onInvalidate() {
     if (mAnimationManager != null) {
-      mAnimationManager.onCatalystInstanceDestroy();
+      mAnimationManager.onInvalidate();
     }
 
     if (mNativeProxy != null) {
-      mNativeProxy.onCatalystInstanceDestroy();
+      mNativeProxy.onInvalidate();
       mNativeProxy = null;
     }
   }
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -109,13 +109,13 @@ public AnimationsManager getAnimationsManager() {
     return mAnimationManager;
   }
 
-  public void onCatalystInstanceDestroy() {
+  public void onInvalidate() {
```
tomekzaw, line 112: Shouldn't it be just `invalidate` instead of `onInvalidate` as in `ReanimatedModule`?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -67,7 +67,7 @@ public AnimationsManager(ReactContext context, UIManagerModule uiManagerModule)
     mSharedTransitionManager = new SharedTransitionManager(this);
   }
 
-  public void onCatalystInstanceDestroy() {
+  public void onInvalidate() {
```
tomekzaw, line 70: same as above

android/src/main/java/com/swmansion/reanimated/nativeProxy/NativeProxyCommon.java
```diff
@@ -226,7 +226,7 @@ public void unsubscribeFromKeyboardEvents(int listenerId) {
 
   protected abstract HybridData getHybridData();
 
-  public void onCatalystInstanceDestroy() {
+  public void onInvalidate() {
```
tomekzaw, line 229: same as above

tomekzaw (approved):

piaskowyk (approved):

Issue #5728, opened 2024-02-26
BUG - Animated.FlatList / itemLayoutAnimation - ANDROID


JeanSilvany:
### Description

Currently, there is a bug that occurs only on Android when using the itemLayoutAnimation in Animated.FlatList.

The animations work perfectly when an item enters, but when removing an item, there is always a UI bug where the last item becomes invisible during rendering. Sometimes, more than one item disappears as well.

Throughout various tests, I found that using the property contentContainerStyle={{ height: '100%' }} prevents the bug from occurring. I also noticed that sometimes the exit animation works perfectly if the item is being controlled by a useState / visible, but even then the last item remains invisible. But this is not the expected behavior for a list.

The current problem occurs only on Android, in both Debug and Release modes with a physical device.

The expectation is that the exit animations function perfectly in the List and that there are no invisible item bugs.

https://github.com/software-mansion/react-native-reanimated/assets/73763375/3595047f-1a54-49d8-8aaf-484e845d5243

### Steps to reproduce

- Add more than 2 items to the list
- Remove any item that is not the last one.

### Snack or a link to a repository

https://snack.expo.dev/@jeansilvany/reanimated-layout-issue

### Reanimated version

3.7.1

### React Native version

0.73.4

### Platforms

Android

### JavaScript runtime

None

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & production bundle

### Device

None

### Acknowledgements

Yes

BrunoADPaixao:
Same here!

gabrielsoaresm94:
Same issue here!!

Latropos:
The problem is fully reproducible, even in simulator

IgorDantasDev:
Any updates?

chiratnachakraborty-betterhalf:
> Right now the best workaround is to hardcode height of your flatList. If for some reason you can't do it, you may consider sth like this: [#5758 (review)](https://github.com/software-mansion/react-native-reanimated/pull/5758#pullrequestreview-1943414942)

Hi @Latropos 
I checked you PR for the same... if you closely look at the video that you shared for the Test Plan... the item which is partially visible , and not fully visible doesn't animate correctly. I really appreciate your effort in putting out a solve for this issue, but any idea as why the above mentioned thing is occuring? and any potential fix for the same?
Thanks

Latropos:
@chiratnachakraborty-betterhalf 
[...]

PR #5758, opened 2024-03-04
Fix android flatlist bug - DO NOT MERGE

Latropos:
## Summary
This is a quick fix of this bug: https://github.com/software-mansion/react-native-reanimated/issues/5728
The problem is that flatList shows odd behaviour on Android, if `itemLayoutAnimation` is provided and height is not hardcoded. The implemented, imperfect solution is to read the actual height and hardcode it (but only if all the 3 conditions are satisfied).

Implementing a workaround instead of fixing the problem is generally not a good strategy, but in this particular case it may be justified:

> [!IMPORTANT]
The bug went unnoticed for so long probably, because `itemLayoutAnimation` is not documented

> [!CAUTION]
This PR is not expected to be merged. We will keep it open until #5728 is fixed to serve as a demonstration of bug workaround.

#### Cons
❌ This is just a symptomatic treatment of the problem

#### Pros
✅ Fighting the real cause of the problem would be very difficult, since it seems located in native code. Developing a native solution would probably take a month to ship (or longer), and community shows an interest in fixing this
✅ The workaround is applied if and only if the bug will appear, so we can be sure not to break any working app

<table>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/480b1bd5-fd17-4654-b667-b9a4713258a5

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/de588671-73aa-49d2-807d-b553dc1c294f

</td></tr>
</table>

## Test plan

When testing the app I used code from the example below. I also made sure to include changing size of flatList, by showing or hiding the yellow view in the recording.

https://github.com/software-mansion/react-native-reanimated/assets/56199675/f49d1b96-84e5-43a1-a05f-2a72f8ea08f1

<details>
<summary> Testing code </summary>
/* eslint-disable no-inline-styles/no-inline-styles */
import {
  Button,
  ColorValue,
  ListRenderItemInfo,
  StyleSheet,
  Text,
  TouchableOpacity,
  View,
} from 'react-native';

import Animated, {
  Layout,
  SlideInRight,
  SlideOutLeft,

  const [data, setData] = useState<ColorValue[]>([]);
  const [toggle, setToggle] = useState(true);

  const handleAddItem = (addItem: ColorValue) => {
    setData((oldState) => [...oldState, addItem]);
  };

  const handleRemoveItem = (removeItem: ColorValue) => {
    setData((oldState) => oldState.filter((item) => item !== removeItem));
  };

  const renderItem = ({ item }: ListRenderItemInfo<ColorValue>) => {
    return (
      <Animated.View entering={SlideInRight} exiting={SlideOutLeft}>
        <TouchableOpacity
          activeOpacity={0.8}
          style={[styles.item, { backgroundColor: item }]}
          onPress={() => handleRemoveItem(item)}>
          <Text>{item.toString()}</Text>
        </TouchableOpacity>
      </Animated.View>
    );
  };

  return (
      <Animated.FlatList
        itemLayoutAnimation={Layout}
        data={data}
        renderItem={renderItem}
        keyExtractor={(item) => item.toString()}
        contentContainerStyle={{ backgroundColor: 'pink' }}
        onLayout={() => {
          console.log('ON LAYOUT');
        }}
        ItemSeparatorComponent={() => <View style={styles.separator} />}
      />

      {toggle && <View style={{ height: 300, backgroundColor: 'yellow' }} />}
      <Button title="Show additional view" onPress={() => setToggle(!toggle)} />

      <Button
        title="Add Item"
        onPress={() =>
          handleAddItem(
            '#' +
              (0x1000000 + Math.random() * 0xffffff).toString(16).substr(1, 6)
          )
        }
      />
    </View>
  );
}

  container: {
    flex: 1,
    padding: 16,
  },
  item: {
    width: '100%',
    height: 80,
    justifyContent: 'center',
    alignItems: 'center',
    borderRadius: 15,
  },
  separator: {
    height: 10,
  },
});

```jsx
```
</details>

### DIFF
--- a/src/reanimated2/component/FlatList.tsx
+++ b/src/reanimated2/component/FlatList.tsx
@@ -1,5 +1,5 @@
 'use strict';
-import React, { forwardRef, useRef } from 'react';
+import React, { forwardRef, useRef, useState } from 'react';
 import type {
   FlatListProps,
   LayoutChangeEvent,
@@ -12,8 +12,10 @@ import { createAnimatedComponent } from '../../createAnimatedComponent';
 import type { ILayoutAnimationBuilder } from '../layoutReanimation/animationBuilder/commonTypes';
 import { LayoutAnimationConfig } from './LayoutAnimationConfig';
 import type { AnimatedProps, AnimatedStyle } from '../helperTypes';
+import { isAndroid } from '../PlatformChecker';
 
 const AnimatedFlatList = createAnimatedComponent(FlatList);
+const IS_ANDROID = isAndroid();
 
 interface CellRendererComponentProps {
   onLayout?: ((event: LayoutChangeEvent) => void) | undefined;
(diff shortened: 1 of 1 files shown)


szydlovsky (approved):
Looks good to me 👍

Tetraminoes:
So i tried this solution and my FlatList stopped becoming scrollable. I assume its because `onLayout` gives the height of the section of the flatlist thats being rendered. I dont want to use a ScrollView because i heard FlatLists are faster for large lists. Any ideas?

Latropos:
@Tetraminoes Try hardcoding flatList height with styles, `{flex:1 }` should work

metkm:
> @Tetraminoes Try hardcoding flatList height with styles, `{flex:1 }` should work

Trying this workaround with horizontal list. It doesn't work. The list can't be scrolled because onLayout nativeEvent returns the size of screen.

metkm:
any ideas @szydlovsky ?

PR #5759, opened 2024-03-04
Fix race condition when ShareableHandle is accessed on multiple threads

### LINKED ISSUE #5660: [3.7.0] Data type not recognized by value unpacker. at valueUnpacker (WorkletRuntime::WorkletRuntime:1:1477)

tjzel:
## Summary

Fixes #5660, a regression introduced in #4300.

The aforementioned race condition happens this way:

1. An object using a `ShareableHandle` underneath (e.g. a shared value) is created.
2. This object is accessed on UI thread.
3. The initializer of this `ShareableHandle` get called on the UI thread.
4. At the same time, JS thread schedules an operation on this object (e.g. setting a value of shared value).
5. Access on UI thread forces the initialization. The condition of the if clause resolves to true and UI thread tries to access the runtime.
Screenshot 2024-03-04 at 13 00 07
6. However, access from JS thread has locked the runtime and causes the UI runtime to wait.
7. Then, JS thread enters the same if clause body and initializes the whole shareable.
8. After initializing and releasing the runtime, UI thread gets unblocked.
9. However, now `initializer_` has been nulled and causes memory access issues.

It's difficult to change the whole flow of locking to prevent such scenarios. Therefore we won't null the `initializer_` object anymore. However, this won't fix the problem of potential double initialization. Luckily, the code of `valueUnpacker` already prevents that with its shareable handle cache and the fact that runtime operations must be sequential. 

## Test plan

Run the following race condition reproduction made - you should re-run the app several times (probably up to 20ish) since it's most likely to happen when the app starts. It's also more likely to happen on Android simulators (at least for me).

<details>
<summary>
Test code
</summary>

```tsx
import {useSharedValue} from 'react-native-reanimated';
import {useEffect, useState} from 'react';

const value = 666666;

const Screen = () => {
  const sv = useSharedValue(value);
  useSharedValue(1);
  useSharedValue(2);
  useSharedValue(3);
  sv.value = value;

  useEffect(() => {
    // eslint-disable-next-line @typescript-eslint/no-unused-vars
    const currentValue = sv.value;
  }, [sv]);

  return null;
};
const ReanimatedCrashReproduction = () => {
  const [render, setRender] = useState(false);

  useEffect(() => {
    const interval = setInterval(() => setRender(r => !r), 500);
    return () => clearInterval(interval);
  }, []);

  return render ? <Screen /> : null;
};

export default ReanimatedCrashReproduction;
```

</summary>

You can do the following to see that double initialization happens still, but all is well 🚰.

Screenshot 2024-03-06 at 10 06 50

Big thanks to @piaskowyk for the debugging help 🚀

### DIFF
--- a/Common/cpp/SharedItems/Shareables.cpp
+++ b/Common/cpp/SharedItems/Shareables.cpp
@@ -251,8 +251,8 @@ jsi::Value ShareableHandle::toJSValue(jsi::Runtime &rt) {
     auto value = std::make_unique<jsi::Value>(getValueUnpacker(rt).call(
         rt, initObj, jsi::String::createFromAscii(rt, "Handle")));
 
-    std::unique_lock<std::mutex> lock(initializationMutex_, std::try_to_lock);
-    if (lock.owns_lock() && remoteValue_ == nullptr) {
+    std::unique_lock<std::mutex> lock(initializationMutex_);
+    if (remoteValue_ == nullptr) {
       remoteValue_ = std::move(value);
       remoteRuntime_ = &rt;
     }
--- a/Common/cpp/SharedItems/Shareables.cpp
+++ b/Common/cpp/SharedItems/Shareables.cpp
@@ -252,7 +252,7 @@ jsi::Value ShareableHandle::toJSValue(jsi::Runtime &rt) {
         rt, initObj, jsi::String::createFromAscii(rt, "Handle")));
 
     std::unique_lock<std::mutex> lock(initializationMutex_, std::try_to_lock);
-    if (lock.owns_lock()) {
+    if (lock.owns_lock() && remoteValue_ == nullptr) {
       remoteValue_ = std::move(value);
       remoteRuntime_ = &rt;
     }
--- a/Common/cpp/SharedItems/Shareables.cpp
+++ b/Common/cpp/SharedItems/Shareables.cpp
@@ -248,9 +248,6 @@ jsi::Value ShareableRemoteFunction::toJSValue(jsi::Runtime &rt) {
 jsi::Value ShareableHandle::toJSValue(jsi::Runtime &rt) {
   if (remoteValue_ == nullptr) {
     auto initObj = initializer_->getJSValue(rt);
-    if (remoteValue_ != nullptr) {
-      jsi::String::createFromAscii(rt, "Handle");
-    }
     remoteValue_ = std::make_unique<jsi::Value>(getValueUnpacker(rt).call(
         rt, initObj, jsi::String::createFromAscii(rt, "Handle")));
     remoteRuntime_ = &rt;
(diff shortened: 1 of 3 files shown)


Latropos:
@tjzel I was looking into linked issue and it seemed that the problem was occurring only in newest release. Could you please track down and link here the exact PR causing the bug?

callaars:
@Latropos, it's been occurring since 3.7.1.

tomekzaw:
@Latropos It seems like the offending PR was #4300.

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -245,13 +246,16 @@ jsi::Value ShareableRemoteFunction::toJSValue(jsi::Runtime &rt) {
...
     auto initObj = initializer_->getJSValue(rt);
-    remoteValue_ =
-        std::make_unique<jsi::Value>(getValueUnpacker(rt).call(rt, initObj));
-    remoteRuntime_ = &rt;
-    initializer_ = nullptr; // we can release ref to initializer as this
-    // method should be called at most once
+    auto value = std::make_unique<jsi::Value>(getValueUnpacker(rt).call(
+        rt, initObj, jsi::String::createFromAscii(rt, "Handle")));
+
+    std::unique_lock<std::mutex> lock(initializationMutex_, std::try_to_lock);
```
piaskowyk, line 254: Wy need to remove `std::try_to_lock` because this will lead to calling `*remoteValue_` on `null` :/

tjzel: smart

piaskowyk (approved):

tomekzaw (approved):

PR #5769, opened 2024-03-08
Fix android layout animation z-index bug

### LINKED ISSUE #5715: [Android] Layout animations break z-index behavior when unmouting/remounting Views
### Description

Z-index positioning behaves as you would expect consistently until the first Reanimated layout animation (entering/exiting) is triggered at any point on the app.

From that point, if Views that rely on z-index to be positioned are unmounted and remounted, the behavior will not be realiable anymore.

This is only happening on Android.

https://github.com/software-mansion/react-native-reanimated/assets/11333829/f2eb10b6-f526-47ce-86ba-8fb98070ec71

## Known workarounds
- If you make sure the view that should be on top is rendered after the View lower on the stack, you get consistent behavior.
   ```
   <AnotherView />
   <ViewOnTop />
   ```
  Instead of
   ```
   <ViewOnTop />
   <AnotherView />
   ```
- You can also use strategies like conditional `display: "none"` instead of unmounting/remounting.

## Similar issues
It looks like z-index behavior on Android is not a new discussion: https://github.com/facebook/react-native/issues/35565 . What was surprising to me was the fact that Reanimated can be the culprit. As in my app, z-index works just fine on Android with the latest react-native until you have an Animated.View with entering/exiting.

Some other people also realized that on their apps this was the cause: https://github.com/facebook/react-native/issues/35565#issuecomment-1453458685

If it helps, I also found these similar old issues:
https://github.com/software-mansion/react-native-reanimated/issues/3368 (mentioned on the comment I linked)
https://github.com/software-mansion/react-native-reanimated/issues/3124 (iOS specific)
Looks like those are solved but since they are of a similar nature, it can be helpful in the process of finding the cause.

### Steps to reproduce

1. On Android, position Views with `position:absolute` on top of each other strictly based on z-index - So the View on top is rendered first but z-index ensures it's on top anyways.
2. Trigger a layout animation on an Animated.View with the entering/exiting props
3. Try to unmount and remount the view with higher z-index multiple times

https://snack.expo.dev/@leonardorib/layout-animations-and-zindex-issue-on-android (Reanimated 3.6.2)
or
https://github.com/leonardorib/layout-animation-and-z-index-issue (Reanimated 3.7.1)

### Reanimated version

3.6.2, 3.7.1

### React Native version

0.73.4

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

Samsung Galaxy M14 (Android 14)

### Acknowledgements

Yes

bartlomiejbloniarz:
## Summary
This PR fixes the bug that was causing z-index to behave incorrectly on Android after a layout animation was triggered. The issue was that sometimes our layout animation implementation was causing `removeView` to be called more than once for the same view. This is not a problem on plain Android, since there is a conditional check there for removal. However, react-native's implementation of `ReactViewGroup` does not check if the view is a part of this `ViewGroup` when handling removal. 

Screenshot 2024-03-08 at 11 17 45

Screenshot 2024-03-08 at 11 12 28
The handler code decreases the counter, that maintains the number of views that have 'z-index' defined by the user. Calling `removeView` too many times causes the counter to go into negative numbers, causing react to disable custom z-index order, even when there are views with user defined z-index.

We decided to add an additional check when calling the `removeView` function in `AnimationsManager.java`. There is a possibility we could remove it altogether, but as of now this is not certain and requires further investigation.

closes #5715

## Test plan

Check if the repro from #5715 works properly

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
@@ -649,7 +649,9 @@ public class AnimationsManager implements ViewHierarchyObserver {
       mReanimatedNativeHierarchyManager.publicDropView(view);
     }
 
-    if (parent != null) {
+    // this removal might be redundant, however we decided to keep it for now to avoid introducing
+    // breaking changes
+    if (parent != null && parent.indexOfChild(view) != -1) {
       parent.removeView(view);
     }
   }


tomekzaw (approved):
LGTM

piaskowyk (approved):

mmmoussa:
This PR needs to be reverted. It broke a core part of our app. We have an animated view which we show and hide when tapping somewhere else on the screen which is no longer being removed correctly.

mmmoussa:
It seems the offending PR was actually merged in 3.9.0 and is not this one. Apologies for the confusion.

PR #5770, opened 2024-03-08
FPSCounter

piaskowyk:
## Summary

This PR adds a new Reanimated component which displays current fps for both JS and UI threads.

## Test plan

Open Reanimated Example on your android/ios emulator and go into the `Performance Monitor Example`


LatekVo:
PreviousTimestamp was an integer, while timestamp was a float.
This cause around 1ms of disparity, giving an error of roughly 6% per frame.
After running with this fix for around 5 minutes, FPS remained at a stable 60 per second.
Previously, after 5 or so minutes fps was either 0. something, or above 5000.

LatekVo:
Hot-reloads break this feature. Looks like old `loopAnimationFrame()` keeps on running after hot-reload.

LatekVo:
Do we push this component just as an example for now, or should we make a standalone component out of it?

src/reanimated2/component/PerformanceMonitor.tsx
```diff
@@ -0,0 +1,203 @@
...
+        animatedProps={animatedProps}
+        editable={false}
+      />
+    </View>
+  );
+}
+
+export function PerformanceMonitor() {
+  return (
+    <View style={[styles.monitor]}>
```
szydlovsky, line 175: You don't need to add square brackets when using only one stylesheet style 😄 
```suggestion
    <View style={styles.monitor}>
```

app/src/examples/PerfomanceMonitorExample.tsx
```diff
@@ -0,0 +1,11 @@
...
+import { View } from 'react-native';
+import { PerformanceMonitor } from 'react-native-reanimated';
+
+export default function PerformanceMonitorExample() {
+  return (
+    <View>
+      <PerformanceMonitor />
+    </View>
+  );
+}
```
szydlovsky, line 11: We usually create examples so, that they showcase our features in action. It would be great if you added some things that the performance can be tracked on. For example, you can add some of our animations - feel free to base on any other example in the app.

szydlovsky (commented):
@LatekVo  The PR looks great! There are some smaller quality-of-life things to add. It would also be great if you featured a `Summary` and a `Test plan` for the PR. If it comes to the summary, just a few sentences describing the feature (and possibly its limitations if there are any), some video showcasing it would also be great. Test plan usually comes down to pointing to a specific example in our app - but if you have any extra code that the feature can be tested on, make sure to include it!

LatekVo:
Thank you for the review @szydlovsky :)
I've applied most of these changes locally, but while at that, i found some new issues which i will have to resolve before requesting another review.
Thank you for your time

LatekVo:
Current performance, now that i think of it i will make it much thinner and more transparent.
https://github.com/software-mansion/react-native-reanimated/assets/74246391/ab65ff21-e6a0-48ce-a8bf-cf53e3cac9b5

app/src/examples/PerfomanceMonitorExample.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const exampleElements = useRef(
+    new Map<Examples, JSX.Element>([
+      [Examples.Empty, <EmptyExample />],
+      [Examples.Bokeh, <BokehExample />],
+      [Examples.Planets, <PlanetsExample />],
+      [Examples.Emojis, <EmojiWaterfallExample />],
+    ])
+  );
+
+  const [currentExample, setCurrentExample] = useState(<EmptyExample />);
```
szydlovsky, line 27: It is generally not suggested to store as heavy objects as entire components in `useState` hook. I would suggest to store `Examples` enum values in there: 
```suggestion
  const [currentExample, setCurrentExample] = useState<Examples>(Examples.Empty);
```

app/src/examples/PerfomanceMonitorExample.tsx
```diff
@@ -0,0 +1,72 @@
...
+        {[
+          Examples.Empty,
+          Examples.Bokeh,
+          Examples.Planets,
+          Examples.Emojis,
+        ].map((element) => (
+          <Pressable
+            key={element}
+            style={styles.button}
+            onPress={() => setElementByExample(element)}>
```
szydlovsky, line 47: According to previous comments, here I would suggest rendering example based solely on the state (not 100% sure if this compiles but you should get the idea):

```suggestion
      {exampleElements.current.get(currentExample)!}
      <View style={styles.buttonContainer}>
        {[
          Examples.Empty,
          Examples.Bokeh,
          Examples.Planets,
          Examples.Emojis,
        ].map((element) => (
          <Pressable
            key={element}
            style={styles.button}
            onPress={() => setCurrentExample(element)}>
```

app/src/examples/PerfomanceMonitorExample.tsx
```diff
@@ -0,0 +1,72 @@
...
+      [Examples.Planets, <PlanetsExample />],
+      [Examples.Emojis, <EmojiWaterfallExample />],
+    ])
+  );
+
+  const [currentExample, setCurrentExample] = useState(<EmptyExample />);
+
+  const setElementByExample = (example: Examples) => {
+    setCurrentExample(exampleElements.current.get(example)!);
+  };
```
szydlovsky, line 31: Then, this function is not needed anymore.
```suggestion
```

szydlovsky (changes requested):

PR #5779, opened 2024-03-11
fix: remove top level is_fabric check

WoLewicki:
## Summary

PR removing top level check for `isFabric` since it should be evaluated after it is added to `global` which may not be done at that point.


tomekzaw (approved):
LGTM

piaskowyk (commented):
Can we be certain that on the very first call to one of those functions, `isFabric()` will return the correct value?

WoLewicki:
@piaskowyk I don't think so, I am not fully aware when those checks are used and if they are used before this line: https://github.com/software-mansion/react-native-reanimated/blob/1ab8c224197b55247034060b24bf71b71afd88d3/Common/cpp/ReanimatedRuntime/RNRuntimeDecorator.cpp#L30 is evaluated then it will not work. I think it is still better than having it in top-level scope.

PR #5780, opened 2024-03-11
fix: prevent error due to undefined right operand of  "in" check in animationBuilder

computerjazz:
## Summary

There's currently an error when a style in an array is `undefined`: 
 `ERROR  TypeError: right operand of 'in' is not an object`

```tsx

function App() {
  return <Animated.View entering={FadeIn} style={[{ flex: 1, backgroundColor: 'red' }, undefined]} />
}

```

not sure why this isn't a TS error?

Screenshot 2024-03-11 at 10 21 28 AM

### DIFF
--- a/src/animationBuilder.tsx
+++ b/src/animationBuilder.tsx
@@ -35,7 +35,7 @@ function getCommonProperties(
     : [componentStyle];
 
   componentStyleFlat = componentStyleFlat.map((style) =>
-    'initial' in style
+    style && 'initial' in style
       ? style.initial.value // Include properties of animated style
       : style
   );


koreymadden:
After updating, I started getting this error too.

src/animationBuilder.tsx
```diff
@@ -35,7 +35,7 @@ function getCommonProperties(
     : [componentStyle];
 
   componentStyleFlat = componentStyleFlat.map((style) =>
-    'initial' in style
+    style && 'initial' in style
       ? style.initial.value // Include properties of animated style
       : style
   );
```
thespacemanatee, line 41: I think we should filter undefined styles as well:

```ts
componentStyleFlat = componentStyleFlat.map((style) =>
  style && 'initial' in style
    ? style.initial.value // Include properties of animated style
    : style
).filter(Boolean);
```

m-bert:
Hi @computerjazz! Thanks for this PR! We decided to merge #5784. I'm closing this one since they do the same thing.
Thank you one more time!

PR #5782, opened 2024-03-12
DEL deprecated override in react-native 0.72-rc

Cura-inthelab:


piaskowyk:
Hello 👋 Thank you for highlighting that. We applied your changes while adding support for React Native 0.74 - [PR link](https://github.com/software-mansion/react-native-reanimated/pull/5815). Once again, thank you for your commitment and time - we truly appreciate it!

PR #5734, opened 2024-02-27
feat: Bridgeless with runtimeExecutor

WoLewicki:
PR bringing 0.74/bridgeless mode support to the library.

PR #5785, opened 2024-03-12
feat: Support `jsi::NativeState` in Shareables

mrousavy:
## Summary

Allows Shared Objects to also retain their `NativeState` when copied between runtimes.

Assuming this object:

```cpp
jsi::Object obj;
obj.setProperty(runtime, "name", jsi::String::forUtf8(runtime, "Marc"));

std::shared_ptr<User> user = std::make_shared<User>("Marc");
obj.setNativeState(user);
```

..gets sent to JS, if it will be used in a Worklet runtime it will only contain the `"name"` property, and it's native state field will be null.

With this PR, `jsi::NativeState` also gets copied over to the JS object if it is shared between runtimes, which is safe to do because it is a `shared_ptr` already.

## Test plan

Try to use an object with Native State (e.g. VisionCamera v4 tests) in a Worklet runtime.

### DIFF
--- a/Common/cpp/SharedItems/Shareables.cpp
+++ b/Common/cpp/SharedItems/Shareables.cpp
@@ -198,6 +198,9 @@ ShareableObject::ShareableObject(jsi::Runtime &rt, const jsi::Object &object)
     auto value = extractShareableOrThrow(rt, object.getProperty(rt, key));
     data_.emplace_back(key.utf8(rt), value);
   }
+  if (object.hasNativeState(rt)) {
+    nativeState_ = object.getNativeState(rt);
+  }
 }
 
 jsi::Value ShareableObject::toJSValue(jsi::Runtime &rt) {
@@ -205,6 +208,9 @@ jsi::Value ShareableObject::toJSValue(jsi::Runtime &rt) {
   for (size_t i = 0, size = data_.size(); i < size; i++) {
     obj.setProperty(
         rt, data_[i].first.c_str(), data_[i].second->getJSValue(rt));
+  }
+  if (nativeState_ != nullptr) {
+    obj.setNativeState(rt, nativeState_);
   }
   return obj;
 }
--- a/Common/cpp/SharedItems/Shareables.h
+++ b/Common/cpp/SharedItems/Shareables.h
@@ -186,6 +186,9 @@ class ShareableObject : public Shareable {
 
  protected:
   std::vector<std::pair<std::string, std::shared_ptr<Shareable>>> data_;
+
+ private:
+  std::shared_ptr<jni::NativeState> nativeState_ = nullptr;
 };
 
 class ShareableHostObject : public Shareable {


mrousavy:
I think `jsi::NativeState` was introduced in RN 0.71, right? (https://github.com/facebook/react-native/commit/61792339657b12f43318e842aba48c7043e40168)

If so, I could also add guards to make sure this is not part of builds on RN 0.70 or lower. lmk

mrousavy:
done

WoLewicki:
I see you did it exactly the same as in my PR, nice. The other thing we would probably need is something similar to https://github.com/software-mansion/react-native-reanimated/pull/5734/files#diff-de926588e6942029aeecf8c65e8d8292808b725b95b6bcf22b421425fd10664eR150. Did you check it on 0.74 maybe?

mrousavy:
Ah lol - that was coincidentally exactly the same approach, didn't even see your PR 😂 

Okay well I mean my PR unblocks me right now, if you intend to merge your PR soon then I guess we can close my PR? Otherwise we can use my PR as an intermediate solution where all `jsi::Object`s can have native state.

I only tried this in RN 0.73, not 0.74 - I don't think we need that check as it is always a `jsi::Object` that has NativeState, never a `jsi::HostObject`, right? I could be mistaken here, I'm mainly working on my use-cases for NativeState..

tomekzaw:
Added some more code to properly handle passing objects with `jsi::NativeState` between JS and UI runtime as well as in the opposite direction (18a83f8).

Here's how this PR can be tested in Reanimated Example app:

<details>
<summary>RNRuntimeDecorator.h</summary>

```cpp
class MyNativeState : public jsi::NativeState {
 public:
  explicit MyNativeState(int number) : number_(number) {}
  int number_;
};
```

</details>

<details>
<summary>RNRuntimeDecorator.cpp</summary>

```cpp
  auto setNativeState = jsi::Function::createFromHostFunction(
      rnRuntime,
      jsi::PropNameID::forAscii(rnRuntime, "setNativeState"),
      2,
      [](jsi::Runtime &rt,
         const jsi::Value &thisValue,
         const jsi::Value *arguments,
         size_t count) -> jsi::Value {
        auto state = std::make_shared<MyNativeState>(arguments[1].asNumber());
        arguments[0].asObject(rt).setNativeState(rt, state);
      });
  rnRuntime.global().setProperty(
      rnRuntime, "setNativeState", std::move(setNativeState));

  auto getNativeState = jsi::Function::createFromHostFunction(
      rnRuntime,
      jsi::PropNameID::forAscii(rnRuntime, "getNativeState"),
      1,
      [](jsi::Runtime &rt,
         const jsi::Value &thisValue,
         const jsi::Value *arguments,
         size_t count) -> jsi::Value {
        const auto &obj = arguments[0].asObject(rt);
        if (!obj.hasNativeState(rt)) {
        }
        return jsi::Value(obj.getNativeState<MyNativeState>(rt)->number_);
      });
  rnRuntime.global().setProperty(
      rnRuntime, "getNativeState", std::move(getNativeState));
```

</details>

<details>
<summary>ShareablesExample.tsx</summary>

```tsx
function NativeStateJSToUIDemo() {
  const handlePress = () => {
    const obj = {};
    // @ts-expect-error
    const { setNativeState, getNativeState } = global;
    setNativeState(obj, Math.random() * 1000);
    console.log(_WORKLET ? 'UI' : 'JS', getNativeState(obj));
    runOnUI(() => {
      console.log(_WORKLET ? 'UI' : 'JS', getNativeState(obj));
    })();
  };

  return <Button title="NativeState JS->UI" onPress={handlePress} />;
}

function NativeStateUIToJSDemo() {
  const handlePress = () => {
    // @ts-expect-error
    const { setNativeState, getNativeState } = global;
    function jsCallback(obj: object) {
      console.log(_WORKLET ? 'UI' : 'JS', getNativeState(obj));
    }
    runOnUI(() => {
      const obj = {};
      setNativeState(obj, Math.random() * 1000);
      console.log(_WORKLET ? 'UI' : 'JS', getNativeState(obj));
      runOnJS(jsCallback)(obj);
    })();
  };

  return <Button title="NativeState UI->JS" onPress={handlePress} />;
}
```

</details>

piaskowyk (approved):
👏

PR #5793, opened 2024-03-15
SET everywhere

piaskowyk:

### DIFF
--- a/apple/LayoutReanimation/REASharedTransitionManager.m
+++ b/apple/LayoutReanimation/REASharedTransitionManager.m
@@ -31,6 +31,9 @@
   BOOL _clearScreen;
   BOOL _isInteractive;
   REAUIView *_disappearingScreen;
+  BOOL _screenWasAddedInUIManager;
+  NSMutableSet<NSNumber *> *_affectedSharedViewTags;
+  NSMutableArray<REASharedElement *> *_sharedTransitionsOnLayoutQueue;
 }
 
 /*
@@ -61,6 +64,9 @@ static REASharedTransitionManager *_sharedTransitionManager;
     _reattachedViews = [NSMutableSet new];
     _isAsyncSharedTransitionConfigured = NO;
     _isConfigured = NO;
+    _screenWasAddedInUIManager = NO;
+    _affectedSharedViewTags = [NSMutableSet new];
+    _sharedTransitionsOnLayoutQueue = [NSMutableArray new];
     [self swizzleScreensMethods];
   }
   return self;
@@ -107,16 +113,168 @@ static REASharedTransitionManager *_sharedTransitionManager;
   _layoutedSharedViewsFrame[view.reactTag] = [[REAFrame alloc] initWithX:x y:y width:width height:height];
 }
 
+- (void)notifyAboutNewScreenAddedInUIManager
+{
+  _screenWasAddedInUIManager = YES;
+}
+
+- (void)notifyAboutAffectedViewTags:(NSArray<NSNumber *> *)affectedViewTags
+{
+  [_affectedSharedViewTags removeAllObjects];
+  for (NSNumber *viewTag in affectedViewTags) {
+    if ([_animationManager hasAnimationForTag:viewTag type:SHARED_ELEMENT_TRANSITION]) {
+      [_affectedSharedViewTags addObject:viewTag];
+    }
+  }
+}
+
 - (void)viewsDidLayout
 {
   if (!_isConfigured) {
     return;
   }
-  [self configureAsyncSharedTransitionForViews:_addedSharedViews];
+  if (_screenWasAddedInUIManager) {
+    [self configureAsyncSharedTransitionForViews:_addedSharedViews];
+    [_addedSharedViews removeAllObjects];
+    [self maybeRestartAnimationWithNewLayout];
+    [_layoutedSharedViewsTags removeAllObjects];
+    [_layoutedSharedViewsFrame removeAllObjects];
+    _screenWasAddedInUIManager = NO;
+    return;
+  }
+
+  if ([_addedSharedViews count] == 0) {
+    return;
+  }
+  [_sharedElements removeAllObjects];
+  for (REAUIView *sharedView in _addedSharedViews) {
+    NSNumber *siblingViewTag = _findPrecedingViewTagForTransition(sharedView.reactTag);
+    REAUIView *siblingView = [_animationManager viewForTag:siblingViewTag];
+    if (siblingView == nil) {
+      continue;
+    }
+    REAUIView *viewSource = siblingView;
+    REAUIView *viewTarget = sharedView;
+    REASnapshot *sourceViewSnapshot = [[REASnapshot alloc] initWithAbsolutePosition:viewSource];
+    REASnapshot *targetViewSnapshot = [[REASnapshot alloc] initWithAbsolutePosition:viewTarget];
+    REASharedElement *sharedElement = [[REASharedElement alloc] initWithSourceView:viewSource
+                                                                sourceViewSnapshot:sourceViewSnapshot
+                                                                        targetView:viewTarget
+                                                                targetViewSnapshot:targetViewSnapshot];
+    sharedElement.animationType = SHARED_ELEMENT_TRANSITION;
+    [_sharedElements addObject:sharedElement];
+
+    _snapshotRegistry[viewSource.reactTag] = sourceViewSnapshot;
+    _sharedElementsLookup[viewSource.reactTag] = sharedElement;
+    [_viewsToHide addObject:viewSource.reactTag];
+  }
+  for (REASharedElement *sharedElement in _sharedTransitionsOnLayoutQueue) {
+    REAUIView *viewTarget = sharedElement.targetView;
+    sharedElement.targetViewSnapshot = [[REASnapshot alloc] initWithAbsolutePosition:viewTarget];
+    [_sharedElements addObject:sharedElement];
+  }
+  [_sharedTransitionsOnLayoutQueue removeAllObjects];
   [_addedSharedViews removeAllObjects];
-  [self maybeRestartAnimationWithNewLayout];
-  [_layoutedSharedViewsTags removeAllObjects];
-  [_layoutedSharedViewsFrame removeAllObjects];
+  if ([_sharedElements count] == 0) {
+    return;
+  }
+  [self configureTransitionContainer];
+  [self reparentSharedViewsForCurrentTransition:_sharedElements];
+  [self startSharedTransition:_sharedElements];
+}
+
+- (void)viewsWillRemove:(NSArray<REAUIView *> *)viewsToRemove
+{
+  if ([viewsToRemove count] == 0) {
+    return;
+  }
+  for (REAUIView *view in viewsToRemove) {
+    if ([_animationManager hasAnimationForTag:view.reactTag type:SHARED_ELEMENT_TRANSITION]) {
+      _snapshotRegistry[view.reactTag] = [[REASnapshot alloc] initWithAbsolutePosition:view];
+    }
+  }
+}
+
+- (void)viewsDidRemoved:(NSArray<REAUIView *> *)removedViews
+{
+  if ([removedViews count] == 0) {
+    return;
+  }
+
+  NSMutableArray<REAUIView *> *filteredViews;
+  if ([_affectedSharedViewTags count] > 0) {
+    for (REAUIView *view in removedViews) {
+      NSNumber *siblingViewTag = _findPrecedingViewTagForTransition(removedViews[0].reactTag);
+      if ([_affectedSharedViewTags containsObject:siblingViewTag]) {
+        REAUIView *viewSource = view;
+        REAUIView *viewTarget = [_animationManager viewForTag:siblingViewTag];
+        REASnapshot *sourceViewSnapshot = _snapshotRegistry[viewSource.reactTag];
+        REASnapshot *targetViewSnapshot = nil; // wait untile the target view will be apply proper layout
+        REASharedElement *sharedElement = [[REASharedElement alloc] initWithSourceView:viewSource
+                                                                    sourceViewSnapshot:sourceViewSnapshot
+                                                                            targetView:viewTarget
+                                                                    targetViewSnapshot:targetViewSnapshot];
+        sharedElement.animationType = SHARED_ELEMENT_TRANSITION;
+        [_sharedTransitionsOnLayoutQueue addObject:sharedElement];
+        _sharedElementsLookup[viewSource.reactTag] = sharedElement;
+        _currentSharedTransitionViews[viewSource.reactTag] = viewSource;
+        [_affectedSharedViewTags removeAllObjects];
+      } else {
+        [filteredViews addObject:view];
+      }
+    }
+    return;
+  }
+
+  if ([filteredViews count] == 0) {
+    return;
+  }
+  NSMutableArray<REAUIView *> *sharedViews = [NSMutableArray new];
+  for (REAUIView *view in filteredViews) {
+    if ([_animationManager hasAnimationForTag:view.reactTag type:SHARED_ELEMENT_TRANSITION]) {
+      [sharedViews addObject:view];
+    }
+  }
+  if ([sharedViews count] == 0) {
+    return;
+  }
+  [_sharedElements removeAllObjects];
+  [_currentSharedTransitionViews removeAllObjects];
+  for (REAUIView *sharedView in sharedViews) {
+    NSNumber *siblingViewTag = _findPrecedingViewTagForTransition(sharedView.reactTag);
+    REAUIView *siblingView = [_animationManager viewForTag:siblingViewTag];
+    if (siblingView == nil) {
+      continue;
+    }
+    REAUIView *viewSource = sharedView;
+    REAUIView *viewTarget = siblingView;
+    REASnapshot *sourceViewSnapshot = _snapshotRegistry[viewSource.reactTag];
+    REASnapshot *targetViewSnapshot = [[REASnapshot alloc] initWithAbsolutePosition:viewTarget];
+    REASharedElement *sharedElement = [[REASharedElement alloc] initWithSourceView:viewSource
+                                                                sourceViewSnapshot:sourceViewSnapshot
+                                                                        targetView:viewTarget
+                                                                targetViewSnapshot:targetViewSnapshot];
+    sharedElement.animationType = SHARED_ELEMENT_TRANSITION;
+    [_sharedElements addObject:sharedElement];
+    _sharedElementsLookup[viewSource.reactTag] = sharedElement;
+    _currentSharedTransitionViews[viewSource.reactTag] = viewSource;
+  }
+  if ([_sharedElements count] == 0) {
+    return;
+  }
+  [self configureTransitionContainer];
+  [self reparentSharedViewsForCurrentTransition:_sharedElements];
+  [self startSharedTransition:_sharedElements];
+}
+
+- (bool)canClearAnimationConfig:(NSNumber *)viewTag
+{
+  for (REASharedElement *sharedElement in _sharedTransitionsOnLayoutQueue) {
+    if (sharedElement.sourceView.reactTag == viewTag) {
+      return false;
+    }
+  }
+  return true;
 }
 
 - (void)configureAsyncSharedTransitionForViews:(NSArray<REAUIView *> *)views
@@ -358,19 +516,19 @@ static REASharedTransitionManager *_sharedTransitionManager;
 - (void)swizzleScreensMethods
 {
 #if LOAD_SCREENS_HEADERS
-  static dispatch_once_t onceToken;
-  dispatch_once(&onceToken, ^{
-    SEL viewDidLayoutSubviewsSelector = @selector(viewDidLayoutSubviews);
-    SEL notifyWillDisappearSelector = @selector(notifyWillDisappear);
-    SEL viewIsAppearingSelector = @selector(viewIsAppearing:);
-    Class screenClass = [RNSScreen class];
-    Class screenViewClass = [RNSScreenView class];
-    BOOL allSelectorsAreAvailable = [RNSScreen instancesRespondToSelector:viewDidLayoutSubviewsSelector] &&
-        [RNSScreenView instancesRespondToSelector:notifyWillDisappearSelector] &&
-        [RNSScreen instancesRespondToSelector:viewIsAppearingSelector] &&
-        [RNSScreenView instancesRespondToSelector:@selector(isModal)]; // used by REAScreenHelper
+  SEL viewDidLayoutSubviewsSelector = @selector(viewDidLayoutSubviews);
+  SEL notifyWillDisappearSelector = @selector(notifyWillDisappear);
+  SEL viewIsAppearingSelector = @selector(viewIsAppearing:);
+  Class screenClass = [RNSScreen class];
+  Class screenViewClass = [RNSScreenView class];
+  BOOL allSelectorsAreAvailable = [RNSScreen instancesRespondToSelector:viewDidLayoutSubviewsSelector] &&
+      [RNSScreenView instancesRespondToSelector:notifyWillDisappearSelector] &&
+      [RNSScreen instancesRespondToSelector:viewIsAppearingSelector] &&
+      [RNSScreenView instancesRespondToSelector:@selector(isModal)]; // used by REAScreenHelper
 
-    if (allSelectorsAreAvailable) {
+  if (allSelectorsAreAvailable) {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
       [REAUtils swizzleMethod:viewDidLayoutSubviewsSelector
                      forClass:screenClass
                          with:@selector(reanimated_viewDidLayoutSubviews)
@@ -383,9 +541,9 @@ static REASharedTransitionManager *_sharedTransitionManager;
                      forClass:screenClass
                          with:@selector(reanimated_viewIsAppearing:)
                     fromClass:[self class]];
-      _isConfigured = YES;
-    }
-  });
+    });
+    _isConfigured = YES;
+  }
 #endif
 }
 
--- a/app/src/examples/EmptyExample.tsx
+++ b/app/src/examples/EmptyExample.tsx
@@ -1,11 +1,37 @@
-import { Text, StyleSheet, View } from 'react-native';
+import React, { useState } from 'react';
+import { StyleSheet, View, Button, Text } from 'react-native';
+import Animated from 'react-native-reanimated';
 
-import React from 'react';
+// export default function EmptyExample() {
+//   const [toggle1, setToggle1] = useState(false);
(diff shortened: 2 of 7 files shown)


Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -159,6 +159,11 @@ int LayoutAnimationsManager::findPrecedingViewTagForTransition(const int tag) {
   return -1;
 }
 
+const std::vector<int> LayoutAnimationsManager::getSharedGroup(const int tag) {
+  auto const &groupName = viewTagToSharedTag_[tag];
```
tomekzaw, line 163: This is a nitpick but we mostly use `const auto &` in our codebase.
```suggestion
  const auto &groupName = viewTagToSharedTag_[tag];
```

Common/cpp/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -159,6 +159,11 @@ int LayoutAnimationsManager::findPrecedingViewTagForTransition(const int tag) {
   return -1;
 }
 
+const std::vector<int> LayoutAnimationsManager::getSharedGroup(const int tag) {
```
tomekzaw, line 162: Since the `std::vector` is stored somewhere, can we return a reference instead of a copy and is it safe to do so?
```suggestion
const std::vector<int> &LayoutAnimationsManager::getSharedGroup(const int tag) {
```

android/src/main/cpp/LayoutAnimations.cpp
```diff
@@ -1,4 +1,5 @@
 #include "LayoutAnimations.h"
+#include <vector>
```
tomekzaw, line 2: This is a nitpick but can we move this import above or below? Currently it's between two double-quote import and this is an angle-bracket import.

android/src/main/cpp/LayoutAnimations.cpp
```diff
@@ -100,10 +101,22 @@ void LayoutAnimations::setFindPrecedingViewTagForTransition(
...
+void LayoutAnimations::setGetSharedGroupBlock(
+    GetSharedGroupBlock getSharedGroupBlock) {
+  getSharedGroupBlock_ = getSharedGroupBlock;
+}
+
 int LayoutAnimations::findPrecedingViewTagForTransition(int tag) {
   return findPrecedingViewTagForTransitionBlock_(tag);
 }
 
+jni::local_ref<JArrayInt> LayoutAnimations::getSharedGroup(int tag) {
```
tomekzaw, line 113: This is a nitpick but we can add `const` here.
```suggestion
jni::local_ref<JArrayInt> LayoutAnimations::getSharedGroup(const int tag) {
```

android/src/main/cpp/LayoutAnimations.cpp
```diff
@@ -100,10 +101,22 @@ void LayoutAnimations::setFindPrecedingViewTagForTransition(
...
+  std::vector<int> group = getSharedGroupBlock_(tag);
```
tomekzaw, line 114:
```suggestion
  const auto &group = getSharedGroupBlock_(tag);
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/ReanimatedNativeHierarchyManager.java
```diff
@@ -214,12 +214,14 @@ public class ReanimatedNativeHierarchyManager extends NativeViewHierarchyManager
...
   private final HashMap<Integer, Set<Integer>> mPendingDeletionsForTag = new HashMap<>();
   private boolean initOk = true;
+  private final TabNavigatorObserver tabNavigatorObserver;
 
   public ReanimatedNativeHierarchyManager(
       ViewManagerRegistry viewManagers, ReactApplicationContext reactContext) {
     super(viewManagers);
 
     mReaLayoutAnimator = new ReaLayoutAnimator(reactContext, this);
+    tabNavigatorObserver = new TabNavigatorObserver(mReaLayoutAnimator);
```
tomekzaw, line 224:
```suggestion
    mTabNavigatorObserver = new TabNavigatorObserver(mReaLayoutAnimator);
```

android/src/main/java/com/swmansion/reanimated/layoutReanimation/ScreensHelper.java
```diff
@@ -0,0 +1,66 @@
...
+import android.view.View;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class ScreensHelper {
+
+  public static View getTabNavigator(View view) {
+    View currentView = view;
+    while (currentView != null) {
+      if (currentView.getClass().getSimpleName().equals("ScreenContainer")) {
```
tomekzaw, line 12: I already saw this check in `ReanimatedNativeHierarchyManager.java`, maybe we can extract it into a separate static method like `isViewScreenContainer`? Same for `Screen` and `ScreenCoordinatorLayout`.

android/src/main/java/com/swmansion/reanimated/layoutReanimation/ScreensHelper.java
```diff
@@ -0,0 +1,66 @@
...
+        return currentView;
+      }
+      if (currentView.getClass().getSimpleName().equals("Screen")
+          && currentView.getParent() != null
+          && currentView
+              .getParent()
```
tomekzaw, line 18: We call `.getParent()` twice, maybe we could store the result somewhere and break this condition into two separate conditionals?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/ScreensHelper.java
```diff
@@ -0,0 +1,66 @@
...
+              .equals("ScreensCoordinatorLayout")) {
+        View screen = currentView;
+        Class<?> screenClass = screen.getClass();
+        try {
+          Method getContainer = screenClass.getMethod("getContainer");
+          currentView = (View) getContainer.invoke(screen);
+        } catch (NoSuchMethodException
+            | InvocationTargetException
+            | IllegalAccessException ignored) {
+        }
```
tomekzaw, line 30: What happens if this exception is thrown? I think it could lead to infinite loop. Maybe let's store the previous `currentView` and break the loop if it doesn't change during the iteration?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/ScreensHelper.java
```diff
@@ -0,0 +1,66 @@
...
+      } else if (currentView.getParent() instanceof View) {
+        currentView = (View) currentView.getParent();
+      } else {
+        break;
+      }
+    }
+    return null;
+  }
+
+  public static boolean isChildOfScreen(View view, View screen) {
```
tomekzaw, line 40:
```suggestion
  public static boolean isViewChildOfScreen(View view, View screen) {
```

apple/LayoutReanimation/REAScreensHelper.h
```diff
@@ -17,5 +18,10 @@
 + (REAUIView *)getScreenWrapper:(REAUIView *)view;
 + (int)getScreenType:(REAUIView *)screen;
 + (bool)isRNSScreenType:(REAUIView *)screen;
++ (bool)isStackChanged:(REAUIView *)view;
++ (REAUIView *)findTopScreenInChildren:(REAUIView *)screen;
++ (REAUIView *)getActiveTabForTabNavigator:(REAUIView *)tabNavigator;
++ (bool)isChild:(REAUIView *)child OfScreen:(REAUIView *)screen;
```
tomekzaw, line 24: This is a nitpick but I think it makes more sense this way
```suggestion
+ (bool)isView:(REAUIView *)view ChildOfScreen:(REAUIView *)screen;
```

apple/LayoutReanimation/REAScreensHelper.h
```diff
@@ -17,5 +18,10 @@
...
++ (bool)isOnTop:(REAUIView *)view;
```
tomekzaw, line 25: What does this method do? Can we make its name more descriptive?
```suggestion
+ (bool)isViewOnTopOfScreenStack:(REAUIView *)view;
```

apple/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -636,7 +829,8 @@ - (void)reparentSharedViewsForCurrentTransition:(NSArray *)sharedElements
 
 - (void)startSharedTransition:(NSArray *)sharedElements
 {
-  for (REASharedElement *sharedElement in sharedElements) {
+  NSArray *sharedElementCopy = [sharedElements copy];
+  for (REASharedElement *sharedElement in sharedElementCopy) {
```
tomekzaw, line 833: Why do we need to make a copy here?

android/src/main/java/com/swmansion/reanimated/layoutReanimation/Snapshot.java
```diff
@@ -117,9 +119,39 @@ public class Snapshot {
     borderRadii = new ReactNativeUtils.BorderRadii(0, 0, 0, 0, 0);
   }
 
+  private int[] tryToFixLocation(View view) {
```
tomekzaw, line 122: Let's rename this method since it returns a value.

app/src/examples/SharedElementTransitions/TabNavigatorExample.tsx
```diff
@@ -0,0 +1,105 @@
...
+    </View>
+  );
+};
+
+const createStack = () => {
+  const Stack = createNativeStackNavigator();
+  return () => (
+    <Stack.Navigator>
+      <Stack.Screen
+        name={'Details'}
```
tomekzaw, line 53:
```suggestion
        name="Details"
```

app/src/examples/SharedElementTransitions/TabNavigatorExample.tsx
```diff
@@ -0,0 +1,105 @@
...
+  [key: string]: {
+    id?: number;
+    showButtons?: boolean;
+  };
+};
+const Screen = ({ navigation, route }: NativeStackScreenProps<ScreenProps>) => {
+  const id = route.params?.id ?? 0;
+  const showButtons = !!route.params?.showButtons;
+  return (
+    <View style={{ ...styles.container }}>
```
tomekzaw, line 29:
```suggestion
    <View style={styles.container}>
```

app/src/examples/SharedElementTransitions/TabNavigatorExample.tsx
```diff
@@ -0,0 +1,105 @@
...
+} from '@react-navigation/native-stack';
+import Animated from 'react-native-reanimated';
+
+function getStyle(index: number) {
+  if (index === 0) {
+    return styles.box1;
+  } else if (index === 1) {
+    return styles.box2;
+  }
+  return styles.box3;
```
tomekzaw, line 16: This is a perfect place to use `switch/case`.

apple/LayoutReanimation/REAScreensHelper.h
```diff
@@ -17,5 +18,10 @@
 + (REAUIView *)getScreenWrapper:(REAUIView *)view;
 + (int)getScreenType:(REAUIView *)screen;
 + (bool)isRNSScreenType:(REAUIView *)screen;
++ (bool)isStackChanged:(REAUIView *)view;
```
tomekzaw, line 21: What does this method do? What does it mean that stack ~~was~~ has been changed?

android/src/main/cpp/LayoutAnimations.cpp
tjzel: Afaik this should be handled by clang-format.

apple/LayoutReanimation/REAScreensHelper.h
tjzel: I prefer the current way.

tjzel: I prefer the current way.

apple/LayoutReanimation/REAScreensHelper.h
tjzel: I'm not sure if you want to add 'View' here, since it's a strongly typed language.

tjzel: On the other hand it's a recurring pattern in the rest of the code, so maybe it's a good idea.

apple/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: We no longer need this copy

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/ScreensHelper.java
```diff
@@ -0,0 +1,83 @@
...
+      if (isScreen(currentView) && isScreensCoordinatorLayout(currentView.getParent())) {
+        View screen = currentView;
+        Class<?> screenClass = screen.getClass();
+        try {
+          Method getContainer = screenClass.getMethod("getContainer");
+          currentView = (View) getContainer.invoke(screen);
+        } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
+          String message =
+              e.getMessage() != null ? e.getMessage() : "Unable to invoke the getContainer method";
+          Log.e("Reanimated", message);
```
tjzel, line 25:
```suggestion
          Log.e("[Reanimated]", message);
```

packages/react-native-reanimated/apple/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -351,26 +361,69 @@ - (NSArray *)sortViewsByTags:(NSArray *)views
...
+  SEL notifyWillDisappearSelector = @selector(notifyWillDisappear);
+  SEL viewIsAppearingSelector = @selector(viewIsAppearing:);
+  Class screenClass = [RNSScreen class];
+  Class screenViewClass = [RNSScreenView class];
+  BOOL allSelectorsAreAvailable = [RNSScreen instancesRespondToSelector:viewDidLayoutSubviewsSelector] &&
+      [RNSScreenView instancesRespondToSelector:notifyWillDisappearSelector] &&
+      [RNSScreen instancesRespondToSelector:viewIsAppearingSelector] &&
+      [RNSScreenView instancesRespondToSelector:@selector(isModal)]; // used by REAScreenHelper
+
+  if (allSelectorsAreAvailable) {
```
tjzel, line 424: Why did you move it like this?

packages/react-native-reanimated/apple/LayoutReanimation/REASharedTransitionManager.m
```diff
@@ -383,21 +436,46 @@ - (void)swizzleScreensMethods
...
+    return _disappearingScreens[count - 1];
+  }
+}
+
+- (NSMutableArray<REAUIView *> *)getDisappearingScreens
+{
+  return _disappearingScreens;
+}
+
+- (void)dismissAssyncTransition
```
tjzel, line 470:
```suggestion
- (void)dismissAsyncTransition
```

tjzel: let's not be vulgar here

tjzel (changes requested):

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/TabNavigatorObserver.java
```diff
@@ -0,0 +1,122 @@
...
+
+  class FragmentLifecycleCallbacks extends FragmentManager.FragmentLifecycleCallbacks {
+    private View firstScreen;
+    private final Method getScreen;
+    private final Method getActivityState;
+    private final Set<Integer> screenTagsWithListener = new HashSet<>();
+    private final List<View> nextTransition = new ArrayList<>();
+
+    public FragmentLifecycleCallbacks(Fragment fragment)
+        throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
```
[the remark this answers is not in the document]
tjzel: Probably we could pass the values the constructor needs so it wouldn't fail, but that's more of an aesthetic decision in this case.

apps/common-app/package.json
```diff
@@ -57,5 +57,8 @@
     "react-native-screens": "3.31.0-rc.1",
     "react-native-svg": "^15.2.0-rc.0",
     "typescript": "~5.3.0"
+  },
```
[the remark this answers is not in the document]
piaskowyk: ok, but i'm curious what is the reason why? 🤔

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -680,7 +680,7 @@ private void cancelAnimationsInSubviews(ViewGroup view) {
     }
   }
 
-  private View resolveView(int tag) {
+  protected View resolveView(int tag) {
```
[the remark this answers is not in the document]
piaskowyk: The `protected` modificator in Java means private inside a package

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/ScreensHelper.java
```diff
@@ -0,0 +1,83 @@
...
+
+import android.util.Log;
+import android.view.View;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class ScreensHelper {
+
+  public static View getTabNavigator(View view) {
+    View currentView = view;
```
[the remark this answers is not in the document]
piaskowyk: Yes, I know, but I wanted to rename a variable, because just `view` is too generic

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/ScreensHelper.java
```diff
@@ -0,0 +1,83 @@
...
+
+  public static boolean isScreenContainer(Object maybeView) {
+    return isInstanceOf(maybeView, "ScreenContainer");
+  }
+
+  public static boolean isScreensCoordinatorLayout(Object maybeView) {
+    return isInstanceOf(maybeView, "ScreensCoordinatorLayout");
+  }
+
+  private static boolean isInstanceOf(Object maybeView, String className) {
```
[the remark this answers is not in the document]
piaskowyk: There is possibility to call `Class.forName(StringName)` but it can throw `ClassNotFoundException`

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -379,6 +396,33 @@ private List<SharedElement> getSharedElementsForCurrentTransition(
...
+    int siblingTag = siblingView.getId();
+    int[] sharedGroup = mNativeMethodsHolder.getSharedGroup(sharedView.getId());
+    int siblingIndex = -1;
+    for (int i = 0; i < sharedGroup.length; i++) {
+      if (sharedGroup[i] == siblingTag) {
+        siblingIndex = i;
+      }
+    }
+
+    for (int i = siblingIndex; i >= 0; i--) {
```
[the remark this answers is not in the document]
piaskowyk: This section of code will only be executed when there is a screen transition within a screen stack under the tab navigators. In the case of a stack, the assumption is that the animation is initiated by the topmost screen. Hence, the matching view to the current view needs to be registered in the array before the shared view of the currently managed top screen.
The order of a views inside the shared group is by the mounting.

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -696,4 +741,58 @@ void orderByAnimationTypes(List<SharedElement> sharedElements) {
...
+    sortViewsByTags(sharedViews);
+    for (View sharedView : sharedViews) {
+      int[] sharedGroup = mNativeMethodsHolder.getSharedGroup(sharedView.getId());
+      for (int i = sharedGroup.length - 1; i >= 0; i--) {
+        View targetView = mAnimationsManager.resolveView(sharedGroup[i]);
+        if (!ScreensHelper.isViewChildOfScreen(targetView, newTab)) {
+          continue;
+        }
+        Snapshot sourceViewSnapshot = mSnapshotRegistry.get(sharedView.getId());
+        if (sourceViewSnapshot == null) {
```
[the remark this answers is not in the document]
piaskowyk: I don't know about any case, but I added it preventively to avoid theoretically possible null ptr exceptions, if you against it I can remove it.

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
```diff
@@ -293,33 +300,43 @@ private List<SharedElement> getSharedElementsForCurrentTransition(
...
-          continue;
+        int stackId = sourceStack.getId();
+        ViewGroupManager stackViewManager =
+            (ViewGroupManager) reanimatedNativeHierarchyManager.resolveViewManager(stackId);
+        boolean isInSameStack = false;
+        for (int i = 0; i < stackViewManager.getChildCount(sourceStack); i++) {
+          if (stackViewManager.getChildAt(sourceStack, i) == viewTargetScreen) {
+            isInSameStack = true;
+          }
         }
```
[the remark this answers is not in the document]
piaskowyk: make sense, any idea how to apply it here?

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/TabNavigatorObserver.java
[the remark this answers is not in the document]
piaskowyk: I understand your concerns, but the constructor must ensure that a constructed object is valid. In this case, it will not, so I would need to add additional checks in every method within this class to verify if the object was properly initialized.

packages/react-native-reanimated/apple/LayoutReanimation/REASharedTransitionManager.m
piaskowyk: I resolved some issues related to missing animations after reload by ensuring that the instance of `_isConfigured` is set to true every time after the `swizzleScreensMethods` call, upon completing the swizzling process (it happens only once for the class).

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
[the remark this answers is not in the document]
tjzel: In such case we should probably add a comment here explaining that.

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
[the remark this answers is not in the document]
tjzel: oh, my bad 😅

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
```diff
@@ -494,7 +494,7 @@ public boolean shouldAnimateExiting(int tag, boolean shouldAnimate) {
   }
 
   public boolean hasAnimationForTag(int tag, int type) {
-    return mNativeMethodsHolder.hasAnimation(tag, type);
+    return mNativeMethodsHolder != null && mNativeMethodsHolder.hasAnimation(tag, type);
```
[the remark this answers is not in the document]
tjzel: So how comes we didn't need it before? What was happening under those conditions?

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/ScreensHelper.java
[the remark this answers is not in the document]
tjzel: I guess it's fine then 🥳

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/SharedTransitionManager.java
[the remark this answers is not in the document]
tjzel: https://react.dev/reference/rules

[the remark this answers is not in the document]
tjzel: oops, wrong link 🙈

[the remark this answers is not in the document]
tjzel: https://stackoverflow.com/questions/21512250/how-can-i-make-my-class-iterable-so-i-can-use-foreach-syntax

it's old but I guess the pattern hasn't changed

apps/common-app/package.json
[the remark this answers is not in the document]
tjzel: It's supposed only to be a container for the common javascript code. I want to avoid the case where its dependencies would be pulled into the example apps and potentially duplicate some libraries. We only have those libraries are dev dependencies for TypeScript here.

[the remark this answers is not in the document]
tjzel: I remember the pain in the previous repo where app and ExampleApp had different versions of some package 💀

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/layoutReanimation/AnimationsManager.java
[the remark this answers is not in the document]
piaskowyk: We always should check it before native method call, but this is outside the scope of this PR

apps/common-app/src/examples/SharedElementTransitions/TabNavigatorExample.tsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
+import { StyleSheet, View, Button, Text } from 'react-native';
+import { createBottomTabNavigator } from '@react-navigation/bottom-tabs';
+import type {NativeStackScreenProps} from '@react-navigation/native-stack';
```
tomekzaw, line 4: Is this file linted properly?

tomekzaw (approved):
LGTM, let's just make sure that `getSharedGroupBlock` is properly mocked on iOS.

tjzel (approved):

PR #5794, opened 2024-03-15
Create first "real" runtime test suite

Latropos:
## Summary

Previous tests were created to test and debug testing environment.

<table>
<tr><td width="50%">Changes</td><td width="50%">Why</td></tr>
<tr><td>
Increase max line width for all files in app/src/examples/RuntimeTests
</td><td>
It's convenient to have one set of test parameters in one line
</td></tr>
<tr><td>
Disable prettier for all snapshot files

``` diff
+ app/**/*.snapshot.ts
```

</td><td>It allows to have one snapshot per line. The file is better organised and you can immediately see how many individual snapshots you have</td>
</tr>

<tr><td>Improve snapshot mismatch formatting, create the following function to have equal length for all the printed numbers

```jsx
export function appendWhiteSpaceToMatchLength(
  message: string | number,
  length: number
)
```

> [!WARNING]
> Temporarily the function throws an error when message has less characters than length

</td>
<td>

It allows to quickly see the difference between the recorded and actual snapshot. 

image

</td>
</tr>

<tr><td>Change the way we render component in test - don't display anything if no component is provided</td><td>We can render invisible components (white views on white backgrounds) to reduce blinking and flickering </td></tr>

<tr><td>
Add test.each syntax like in jest <a href="https://jestjs.io/docs/api#testeachtablename-fn-timeout">LINK</a>
</td><td>
Possibility to create simpler tests
</td></tr>

</table>

## Recording

https://github.com/software-mansion/react-native-reanimated/assets/56199675/eb923b01-226a-42c5-b6a2-d202e90414d1

## Full test logs
image


piaskowyk (approved):
👏👏👏

PR #5803, opened 2024-03-18
Fix ocassional text interactivity issues.

LatekVo:
### Issue 1
After deployment, website behaved like it does after a hot-reload in dev, which isn't it's intended look.
### Issue 2
When announcement banner was visible, text was not given the `interactiveHeaderText` css class, thus breaking it.

### DIFF
--- a/docs/src/components/Hero/SelectedLabel/index.tsx
+++ b/docs/src/components/Hero/SelectedLabel/index.tsx
@@ -15,8 +15,8 @@ const SelectedLabel: React.FC<{
   isInteractive: boolean;
 }> = ({ children, isInteractive = false }) => {
   // DOM refs
-  const selectionRef = useRef(null);
-  const selectionContainerRef = useRef(null);
+  const selectionRef = useRef<HTMLSpanElement | null>(null);
+  const selectionContainerRef = useRef<HTMLDivElement | null>(null);
   const textLabelRef = useRef<HTMLSpanElement | null>(null);
 
   // Render-persistent label positioning styles
--- a/docs/src/components/Hero/SelectedLabel/index.tsx
+++ b/docs/src/components/Hero/SelectedLabel/index.tsx
@@ -88,8 +88,6 @@ const SelectedLabel: React.FC<{
     );
   };
 
-  const classList = styles.headerText;
-
   return (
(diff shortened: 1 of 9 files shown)


docs/src/components/Hero/SelectedLabel/index.tsx
```diff
@@ -88,12 +88,7 @@ const SelectedLabel: React.FC<{
     );
   };
 
-  const classList = clsx(
-    isInteractive ? styles.preEnabledTextInteractivity : null,
-    staticStyles.current.enabledTextInteractivity
-      ? styles.interactiveHeaderText
-      : styles.headerText
-  );
+  const classList = styles.headerText;
```
piaskowyk, line 91: You don't need variable here `const classList`, you can just write className={styles.headerText}

docs/src/components/Hero/SelectedLabel/utils.ts
```diff
@@ -78,8 +77,8 @@ export const computeTextStyles = (
   staticStyles: StaticStyles
 ): TextScaleStyles => {
   // these magic numbers are a result of disparity between font's apparent and actual size
-  const sizeOffsetX = 0.99;
-  const sizeOffsetY = 1.21;
+  const sizeOffsetX = 1.0;
+  const sizeOffsetY = 1.22;
```
piaskowyk, line 81: Why you changed those values? 🤔

LatekVo: Changes in CSS affect the alignment when scaling.
I changed some margins, so i had to correct these.
I know this sounds like a really bad system, but they correct the disparities between font height, and it's actual strokes position which cannot be easily detected within JS without going into the TTF files themselves.
Perhaps i should move these values outside of the function, like it's done here?
Example: 
```js
. . .

const DEFAULT_MAX_DURATION_MS = 800;
const DEFAULT_MIN_VELOCITY = 700;

export default class FlingGestureHandler extends GestureHandler {
  private numberOfPointersRequired = DEFAULT_NUMBER_OF_TOUCHES_REQUIRED;
  private direction: Directions = DEFAULT_DIRECTION;

. . .
```

piaskowyk (approved):

PR #5804, opened 2024-03-18
Improve code quality of helper classes of createAnimatedComponent

tjzel:
## Summary

It was getting me mad for quite some time. Here I remove passing `this` to the methods of the helper classes - instead they are passed in the constructor.

## Test plan

🚀

### DIFF
--- a/src/createAnimatedComponent/InlinePropManager.ts
+++ b/src/createAnimatedComponent/InlinePropManager.ts
@@ -1,7 +1,7 @@
 'use strict';
 import type { StyleProps } from '../reanimated2';
 import type {
-  IAnimatedComponentInternal,
+  AnimatedComponent,
   AnimatedComponentProps,
   IInlinePropManager,
   ViewInfo,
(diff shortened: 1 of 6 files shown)


src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -132,14 +132,15 @@ export class InlinePropManager implements IInlinePropManager {
   _inlinePropsViewDescriptors: ViewDescriptorsSet | null = null;
   _inlinePropsMapperId: number | null = null;
   _inlineProps: StyleProps = {};
```
Latropos, line 134:
```suggestion
 private _inlinePropsViewDescriptors: ViewDescriptorsSet | null = null;
 private _inlinePropsMapperId: number | null = null;
 private _inlineProps: StyleProps = {};
```

src/createAnimatedComponent/InlinePropManager.ts
```diff
@@ -132,14 +132,15 @@ export class InlinePropManager implements IInlinePropManager {
...
+  private _component: AnimatedComponent;
```
Latropos, line 135: Why this is called `_component` and in JSPropUpdater we call it `_animatedComponent`? Both are animated, aren't they?

Latropos (commented):
Please consider small changes from this review, and fill up "Test plan" section with the following information:
* What examples from our example app you checked to work with your changes
* What platforms did you check, please check web too

PR #5805, opened 2024-03-18
Add web implementation for useScrollViewOffset

szydlovsky:
## Summary

Our `useScrollViewOffset` had no support for web, so I added it using some basic web scroll listeners.
There are some funny typing shenanigans, if anyone comes up with better solutions I will happily use it.

| before | after |
| --- | --- |
| <video src="https://github.com/software-mansion/react-native-reanimated/assets/77503811/fd61dd7d-3070-4d6e-a8e5-d00b25fe32db"/> | <video src="https://github.com/software-mansion/react-native-reanimated/assets/77503811/5b49a3dd-1a34-4d01-8c09-ae4bd723c9c3" /> |

## Test plan

Check `useScrollViewOffset` example from Reanimated WebExample and watch console logs or copy [example code](https://docs.swmansion.com/react-native-reanimated/docs/scroll/useScrollViewOffset/#example) from our docs.

More in-depth testing can be done using the following code - it tests switching and mouting/dismounting components that the hook refers to:

``` TYPESCRIPT
import Animated, {
  useAnimatedRef,
  useDerivedValue,
  useSharedValue,
  useScrollViewOffset,
import { Button, StyleSheet, Text, View } from 'react-native';

export default function ScrollViewOffsetExample() {
  const bref = useAnimatedRef<Animated.ScrollView>();
  const scrollHandler = useSharedValue(0);
  const [scrollAMounted, setScrollAMounted] = useState(true);
  const [scrollBMounted, setScrollBMounted] = useState(true);
  const [scrollAPassed, setScrollAPassed] = useState(true);

  useDerivedValue(() => {
    console.log(scrollHandler.value);
  });

  const onAMountPress = () => {
    setScrollAMounted(!scrollAMounted);
  };
  const onBMountPress = () => {
    setScrollBMounted(!scrollBMounted);
  };
  const onPassTogglePress = () => {
    setScrollAPassed(!scrollAPassed);
  };

  useScrollViewOffset(scrollAPassed ? aref : bref, scrollHandler);

  return (
    <>
      <View style={styles.positionView}>
        <Text>Test</Text>
      </View>
      <View style={styles.divider} />
      <Button
        title={`${scrollAMounted ? 'Dismount' : 'Mount'} scroll A`}
        onPress={onAMountPress}
      />
      <Button
        title={`${scrollBMounted ? 'Dismount' : 'Mount'} scroll B`}
        onPress={onBMountPress}
      />
      <Button
        title={`Toggle the ref, currently passed to ${
          scrollAPassed ? 'scroll A' : 'scroll B'
        }`}
        onPress={onPassTogglePress}
      />
      {scrollAMounted ? (
        <Animated.ScrollView
          ref={aref}
          style={[styles.scrollView, { backgroundColor: 'purple' }]}>
          {[...Array(100)].map((_, i) => (
            <Text key={i} style={styles.text}>
              A: {i}
            </Text>
          ))}
        </Animated.ScrollView>
      ) : null}
      {scrollBMounted ? (
        <Animated.ScrollView
          ref={bref}
          style={[styles.scrollView, { backgroundColor: 'lime' }]}>
          {[...Array(100)].map((_, i) => (
            <Text key={i} style={styles.text}>
              B: {i}
            </Text>
          ))}
        </Animated.ScrollView>
      ) : null}
    </>
  );
}

  positionView: {
    margin: 20,
    alignItems: 'center',
  },
  scrollView: {
    flex: 1,
    width: '100%',
  },
  text: {
    fontSize: 50,
    textAlign: 'center',
  },
  divider: {
    backgroundColor: 'black',
    height: 1,
  },
});

```

### DIFF
--- a/src/reanimated2/hook/useScrollViewOffset.ts
+++ b/src/reanimated2/hook/useScrollViewOffset.ts
@@ -34,6 +34,7 @@ function useScrollViewOffsetJS(
     // eslint-disable-next-line react-hooks/rules-of-hooks
     initialRef !== undefined ? initialRef : useSharedValue(0)
   );
+  const scrollRef = useRef<AnimatedScrollView | null>(null);
 
   const eventHandler = useCallback(() => {
     'worklet';
@@ -45,6 +46,12 @@ function useScrollViewOffsetJS(
   }, [animatedRef, animatedRef.current]);
 
   useEffect(() => {
+    // We need to make sure that listener for old animatedRef value is removed
+    if(scrollRef.current !== null) {
+      (scrollRef.current as unknown as HTMLElement).removeEventListener('scroll', eventHandler);
+    }
+    scrollRef.current = animatedRef.current;
+
     const element = animatedRef.current as unknown as HTMLElement;
     element.addEventListener('scroll', eventHandler);
     return () => {
--- a/src/reanimated2/hook/useScrollViewOffset.ts
+++ b/src/reanimated2/hook/useScrollViewOffset.ts
@@ -15,14 +15,6 @@ import { isWeb } from '../PlatformChecker';
 
 const IS_WEB = isWeb();
 
-const scrollEventNames = [
-  'onScroll',
-  'onScrollBeginDrag',
-  'onScrollEndDrag',
-  'onMomentumScrollBegin',
-  'onMomentumScrollEnd',
-];
-
 /**
  * Lets you synchronously get the current offset of a `ScrollView`.
  *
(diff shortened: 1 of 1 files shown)


tjzel (changes requested):
The level of branching in this hook is high - let's split it into two separate implementations and export the relevant one, see `ViewDescriptorSet.ts` for reference.

tjzel (changes requested):

tjzel (approved):
:shipit:

src/reanimated2/hook/useScrollViewOffset.ts
```diff
@@ -15,22 +15,69 @@ import { isWeb } from '../PlatformChecker';
...
  * Lets you synchronously get the current offset of a `ScrollView`.
  *
  * @param animatedRef - An [animated ref](https://docs.swmansion.com/react-native-reanimated/docs/core/useAnimatedRef) attached to an Animated.ScrollView component.
  * @returns A shared value which holds the current offset of the `ScrollView`.
  * @see https://docs.swmansion.com/react-native-reanimated/docs/scroll/useScrollViewOffset
  */
-export function useScrollViewOffset(
+export const useScrollViewOffset = IS_WEB
+  ? useScrollViewOffsetJS
+  : useScrollViewOffsetNative;
```
tomekzaw, line 27: Should be `useScrollViewOffsetWeb` instead of `useScrollViewOffsetJS`

src/reanimated2/hook/useScrollViewOffset.ts
```diff
@@ -15,22 +15,69 @@ import { isWeb } from '../PlatformChecker';
...
+
+function useScrollViewOffsetJS(
+  animatedRef: AnimatedRef<AnimatedScrollView>,
+  initialRef?: SharedValue<number>
+): SharedValue<number> {
+  const offsetRef = useRef(
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    initialRef !== undefined ? initialRef : useSharedValue(0)
+  );
```
tomekzaw, line 36: Do we really need to call `useSharedValue` inside and ignore rules of hooks?

src/reanimated2/hook/useScrollViewOffset.ts
```diff
@@ -15,22 +15,69 @@ import { isWeb } from '../PlatformChecker';
...
+  }, [animatedRef, animatedRef.current]);
+
+  useEffect(() => {
+    // We need to make sure that listener for old animatedRef value is removed
+    if (scrollRef.current !== null) {
+      (scrollRef.current as unknown as HTMLElement).removeEventListener(
+        'scroll',
+        eventHandler
+      );
+    }
```
tomekzaw, line 55: Why do we need to call `removeEventListener` here as well instead of only in `useEffect` cleanup function?

src/reanimated2/hook/useScrollViewOffset.ts
szydlovsky: This is exactly the situation we mentioned today, I made it this way so it mirrors the native implementation

src/reanimated2/hook/useScrollViewOffset.ts
szydlovsky: @tomekzaw because the animatedRef reference might change to another scroll and we then want to clean the previous listener (see example in code section of PR description)

tomekzaw: Can we just assign `const element = animatedRef.current` first (when calling `addEventListener`) and then call `removeEventListener` on `element`?

szydlovsky: That's exactly what `scrollRef` does here

szydlovsky: To be precise: we store a ref to the affected component in `scrollRef` each time `useEffect` runs. Since its dependencies are `animatedRef`, `animatedRef.current` and `eventHandler` (which also regenerates only when `animatedRef` and `animatedRef.current` change), it re-runs ONLY when we switch the hook to another component. In such situation, cleanup on `scrollRef` runs first, ensuring there is no listener to a scroll that we don't care about, and then we can safely add new listener to a new component.

src/reanimated2/hook/useScrollViewOffset.ts
szydlovsky: We want to either return values to a given sharedValue - `initialRef`, or return a new sharedValue that get the updates. So far, no one came up with a better way to do it other than conditionally calling the hook. As soon as we find it or change this logic, it will get updated.

src/reanimated2/hook/useScrollViewOffset.ts
tomekzaw: Oh, I see now. Anyway, is it possible to modify the code so that `removeEventListener` appear only once?

szydlovsky: I'm afraid I don't see any other way. The `removeEventListener` calls match two different scenarios - the `scrollRef` one makes sure we disconnect from an older scroll component during a switch, while the `return` cleanup runs when the component (or its parent) gets dismounted for good.

PR #5806, opened 2024-03-18
Add information about type of `toValue` to our documentation

Latropos:
## Summary
Our documentation is missing information about type of `toValue`

> [!Note]
> The information added in this PR should be included in our documentation 3 times, on pages `withTiming`, `withSpring` and `withDecay`. I will copy-paste it once the text is accepted

image

### DIFF
--- a/docs/docs/animations/withTiming.mdx
+++ b/docs/docs/animations/withTiming.mdx
@@ -54,6 +54,26 @@ enum ReduceMotion {
 #### `toValue`
 
 The value on which the animation will come at rest.
+There are many possible types for `toValue`, currently the following kinds of values are supported:
+
+* **colors** - we support wide range of color formats:
+  * hex numbers - `0xff1234`
+  * rgb - `rgb(100,50,0)`
+  * rgba - `rgba(255,105,180,0)`
+  * hsl - `hsl(0, 50%, 50%)`
+  * hex strings - `#53575E`
+  * css color names - `dodgerblue`
+* **numbers**
+* **suffixed numbers** - strings being a number with a unit, like `"100%"` or `"90deg"`
+* **objects** - object with properties that will be animated separately
+* **array** - array of values, each value will be animated separately.
+* **transformation matrix** - each array of exactly 16 numerical values is by default 
+animated as a transformation matrix. It means that we don't apply the animation to each value separately.
+Instead we decompose the transformation matrix into matrices of 3 basic transformations - rotation, scale and translation.
+Then we animate each of these 3 transforms separately.
+
+Please mind, that `toValue` and the animated shared value should share the same category
+(for example you can't animate width from 100px to 50%)
 
 #### `config` <Optional />
 
--- a/docs/docs/animations/withTiming.mdx
+++ b/docs/docs/animations/withTiming.mdx
@@ -56,21 +56,21 @@ enum ReduceMotion {
 The value on which the animation will come at rest.
 There are many possible types for `toValue`, currently the following kinds of values are supported:
 
-* **colors** - we support wide range of color formats:
-  * hex numbers - `0xff1234`
-  * rgb - `rgb(100,50,0)`
-  * rgba - `rgba(255,105,180,0)`
-  * hsl - `hsl(0, 50%, 50%)`
-  * hex strings - `#53575E`
-  * css color names - `dodgerblue`
-* **numbers**
-* **suffixed numbers** - strings being a number with a unit, like `"100%"` or `"90deg"`
-* **objects** - object with properties that will be animated separately
-* **array** - array of values, each value will be animated separately.
-* **transformation matrix** - each array of exactly 16 numerical values is by default 
-animated as a transformation matrix. It means that we don't apply the animation to each value separately.
-Instead we decompose the transformation matrix into matrices of 3 basic transformations - rotation, scale and translation.
-Then we animate each of these 3 transforms separately.
+- **colors** - we support wide range of color formats:
+  - hex numbers - `0xff1234`
+  - rgb - `rgb(100,50,0)`
+  - rgba - `rgba(255,105,180,0)`
+  - hsl - `hsl(0, 50%, 50%)`
+  - hex strings - `#53575E`
+  - css color names - `dodgerblue`
+- **numbers**
+- **suffixed numbers** - strings being a number with a unit, like `"100%"` or `"90deg"`
+- **objects** - object with properties that will be animated separately
+- **array** - array of values, each value will be animated separately.
+- **transformation matrix** - each array of exactly 16 numerical values is by default
+  animated as a transformation matrix. It means that we don't apply the animation to each value separately.
+  Instead we decompose the transformation matrix into matrices of 3 basic transformations - rotation, scale and translation.
+  Then we animate each of these 3 transforms separately.
 
 Please mind, that `toValue` and the animated shared value should share the same category
 (for example you can't animate width from 100px to 50%)
--- a/docs/docs/animations/withTiming.mdx
+++ b/docs/docs/animations/withTiming.mdx
@@ -53,27 +53,22 @@ enum ReduceMotion {
 
 #### `toValue`
 
-The value on which the animation will come at rest.
-There are many possible types for `toValue`, currently the following kinds of values are supported:
+The value on which the animation will come at rest. Supported categories:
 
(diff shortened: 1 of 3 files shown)


szydlovsky (approved):

docs/docs/animations/withTiming.mdx
```diff
@@ -54,6 +54,26 @@ enum ReduceMotion {
...
 The value on which the animation will come at rest.
+There are many possible types for `toValue`, currently the following kinds of values are supported:
+
+- **colors** - we support wide range of color formats:
+  - hex numbers - `0xff1234`
+  - rgb - `rgb(100,50,0)`
+  - rgba - `rgba(255,105,180,0)`
+  - hsl - `hsl(0, 50%, 50%)`
+  - hex strings - `#53575E`
+  - css color names - `dodgerblue`
```
kacperkapusciak, line 65:
```suggestion
- **colors**
  - RGB (Red, Green, Blue) - e.g. `rgb(100, 50, 0)`,
  - RGBA (Red, Green, Blue, Alpha) - e.g. `rgba(255, 105, 180, 0)`,
  - RGB Hexadecimal - e.g. `#53575E`,
  - Hexadecimal integer literal - e.g. `0xff1234`,
  - HSL (Hue, Saturation, Lightness) - e.g.`hsl(0, 50%, 50%)`,
  - Named colors - e.g. `dodgerblue`.
```

let's use the official MDN naming

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value
https://developer.mozilla.org/en-US/docs/Web/CSS/named-color
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#numeric_literals

docs/docs/animations/withTiming.mdx
```diff
@@ -54,6 +54,26 @@ enum ReduceMotion {
...
+- **numbers**
+- **suffixed numbers** - strings being a number with a unit, like `"100%"` or `"90deg"`
```
kacperkapusciak, line 67: The "suffixed numbers" category isn't very precise. MDN specifies directly what "suffixed numbers" are supported and it's separated as https://developer.mozilla.org/en-US/docs/Web/CSS/angle or https://developer.mozilla.org/en-US/docs/Web/CSS/percentage

Are `150px` "suffixed number" supported? From this sentence I'm not sure.

Please check the source code and write precisely what types (angle, percentage, or more?) are supported.

docs/docs/animations/withTiming.mdx
```diff
@@ -54,6 +54,26 @@ enum ReduceMotion {
...
+- **objects** - object with properties that will be animated separately
+- **array** - array of values, each value will be animated separately.
+- **transformation matrix** - each array of exactly 16 numerical values is by default
+  animated as a transformation matrix. It means that we don't apply the animation to each value separately.
+  Instead we decompose the transformation matrix into matrices of 3 basic transformations - rotation, scale and translation.
+  Then we animate each of these 3 transforms separately.
+
+Please mind, that `toValue` and the animated shared value should share the same category
+(for example you can't animate width from 100px to 50%)
```
kacperkapusciak, line 76:
```suggestion
Please mind, that `toValue` and the animated shared value have to share the same category (e.g. you can't animate `width` from `100px` to `50%`).
```
nitpick

docs/docs/animations/withTiming.mdx
```diff
@@ -54,6 +54,26 @@ enum ReduceMotion {
...
+  - hex numbers - `0xff1234`
+  - rgb - `rgb(100,50,0)`
+  - rgba - `rgba(255,105,180,0)`
+  - hsl - `hsl(0, 50%, 50%)`
+  - hex strings - `#53575E`
+  - css color names - `dodgerblue`
+- **numbers**
+- **suffixed numbers** - strings being a number with a unit, like `"100%"` or `"90deg"`
+- **objects** - object with properties that will be animated separately
+- **array** - array of values, each value will be animated separately.
```
kacperkapusciak, line 69:
```suggestion
- **array** - array of numbers, each value will be animated separately,
```
any values? i think it only works with numbers

docs/docs/animations/withTiming.mdx
```diff
@@ -54,6 +54,26 @@ enum ReduceMotion {
...
+- **colors** - we support wide range of color formats:
+  - hex numbers - `0xff1234`
+  - rgb - `rgb(100,50,0)`
+  - rgba - `rgba(255,105,180,0)`
+  - hsl - `hsl(0, 50%, 50%)`
+  - hex strings - `#53575E`
+  - css color names - `dodgerblue`
+- **numbers**
+- **suffixed numbers** - strings being a number with a unit, like `"100%"` or `"90deg"`
+- **objects** - object with properties that will be animated separately
```
kacperkapusciak, line 68:
```suggestion
- **objects** - object with properties that will be animated separately,
```
comma

docs/docs/animations/withTiming.mdx
```diff
@@ -54,6 +54,26 @@ enum ReduceMotion {
...
+- **array** - array of values, each value will be animated separately.
+- **transformation matrix** - each array of exactly 16 numerical values is by default
+  animated as a transformation matrix. It means that we don't apply the animation to each value separately.
+  Instead we decompose the transformation matrix into matrices of 3 basic transformations - rotation, scale and translation.
+  Then we animate each of these 3 transforms separately.
```
kacperkapusciak, line 73:
```suggestion
- **transformation matrix** - an array consisting of exactly 16 numerical values is by default animated as a transformation matrix. The numbers in the matrix aren't animated separately. Instead, the array gets decomposed into 3 basic transformations - rotation, scale, and translation – which are then animated separately.
```
a bit simpler I believe

kacperkapusciak (changes requested):

kacperkapusciak:
it is possible to reuse the text by using components like we do in RNGH https://github.com/software-mansion/react-native-gesture-handler/blob/main/docs/docs/gestures/fling-gesture.md?plain=1#L82 but copying and pasting it 3 times is more than fine. 

If i were you I'd copy'n'paste it 3 times

docs/docs/animations/withTiming.mdx
Latropos: `0xff1234` is not a hexadecimal integer literal, but an actual number. Since even you misunderstood it I must have been not precise enough 😆 
Anyway, maybe we should skip it and stop supporting it (at least officialy) anyway? Jut because this is not supported by React-Native.

docs/docs/animations/withTiming.mdx
Latropos: Yes, it does, I've just verified

docs/docs/animations/withTiming.mdx
Latropos: Sure, so technically we support all 'normal' literal suffixes applied with no space to the number.
"Normal" means here that neither "zł" or "$" works, the suffix should consist of letters only. I will change the message.

<details><summary> Example code </summary>

```jsx
import Animated, {
  useAnimatedProps,
  useSharedValue,
  withTiming,
  Easing,
import { Button, TextInput, StyleSheet, View } from 'react-native';

import React from 'react';

Animated.addWhitelistedNativeProps({ text: true });

export default function CounterExample() {
  const suffix = 'PLN';
  const sv = useSharedValue(`0${suffix}`);
  const toggle = useSharedValue(false);

    return { text: sv.value, defaultValue: sv.value };
  });

  const handleToggle = () => {
    const newValue = toggle.value ? `0${suffix}` : `10${suffix}`;
    toggle.value = !toggle.value;

    sv.value = withTiming(newValue, {
      duration: 1000,
      easing: Easing.steps(10, true),
    });
  };

  return (
    <>
      <View style={styles.buttons}>
        <Button onPress={handleToggle} title="Toggle" />
      </View>
      <AnimatedTextInput
        animatedProps={animatedProps}
        style={styles.text}
        editable={false}
      />
    </>
  );
}

  buttons: {
    marginVertical: 50,
  },
  text: {
    fontSize: 100,
    fontVariant: ['tabular-nums'],
    textAlign: 'center',
  },
});
```
</details>

https://github.com/software-mansion/react-native-reanimated/assets/56199675/12787d19-d693-4262-9d93-b663383cd0e8

kacperkapusciak (approved):

PR #5807, opened 2024-03-18
Fixes for Reanimated with Bridgeless

tjzel:
## Summary

Fix some stuff in @WoLewicki's PR #5734.

### DIFF
--- a/src/createAnimatedComponent/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent/createAnimatedComponent.tsx
@@ -348,18 +348,23 @@ export function createAnimatedComponent(
             '[Reanimated] Cannot find host instance for this component. Maybe it renders nothing?'
           );
         }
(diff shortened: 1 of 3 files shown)


src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -348,18 +348,25 @@ export function createAnimatedComponent(
...
-         * RN uses viewConfig for components for storing different properties of the component(example: https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Components/ScrollView/ScrollViewNativeComponent.js#L24).
-         * The name we're looking for is in the field named uiViewClassName.
-         */
-        viewName = hostInstance?.viewConfig?.uiViewClassName;
-
-        viewConfig = hostInstance?.viewConfig;
 
+        // TODO Move each implentation to separate function so it's not that bloated here.
+        // Theoretically I could just add another underscoe to each property dynamically
+        // but it's ugly af.
```
WoLewicki, line 354:
```suggestion
        // TODO Move each implementation to separate function so it's not that bloated here.
        // Theoretically I could just add another underscore to each property dynamically
        // but it does not sound good either.
```

Issue #5345, opened 2023-11-07
Passing the same useAnimatedScrollHandler to different scroll views breaks


gaearon:
### Description

We're seeing an issue where passing the same `useAnimatedScrollHandler({ onScroll })` to multiple flat list nested inside a pager fails to register scroll events for the last scroll view.

Replacing it with a vanilla `onScroll` handler fixes it which makes me think it's a problem with Reanimated.

Moving the `useAnimatedScrollHandler` Hook into the leaf components (so creating a handler per component instead of passing down the same handler to many components) also works around the issue.

### Steps to reproduce

I'm sorry, but I don't have time to extract an isolated repro right now.

I was thinking to file the issue anyway because maybe you can guess the source of the problem. 

If you do want to try my non-isolated repro, you'd need a Bluesky account — happy to send an invite (pls DM me on twitter or email me). The repro step there is to open the profile tab, click on "Posts & Replies" and observe the `onScroll` handler from `<PagerWithHeader>` only firing for the first `<Feed>` to which it's passed to from `screens/Profile.tsx`.

If you want to try to recreate the issue from scratch, maybe you can try:

1. Declaring one scroll handler
2. Passing it to multiple flat lilsts
3. If that's not enough, maybe wrapping those flat lists into an RN pager

### Snack or a link to a repository

https://github.com/bluesky-social/social-app/commits/profile-tab-reanimated-bug-repro

### Reanimated version

3.4.2

### React Native version

0.72.5

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Acknowledgements

Yes

gaearon:
There is a repro link, but it is not under my username. That said, it is indeed not a minimal repro.

gaearon:
Maybe related, has a repro: https://github.com/software-mansion/react-native-reanimated/issues/5360

gaearon:
Yea I think https://github.com/software-mansion/react-native-reanimated/issues/5360 is a separate issue but it *also* contains a reduced repro for this one.

gaearon:
Here's a reduced repro for this specific issue.

```js
import * as React from 'react'
import {View, Button} from 'react-native'
import Animated, {useAnimatedScrollHandler} from 'react-native-reanimated'

export default function App() {
  const [attachBoth, setAttachBoth] = React.useState(false)

  const [x, setX] = React.useState(0)
[...]

PR #5809, opened 2024-03-19
Add documentation that event handlers cannot be used with more than one component.

Latropos:
## Summary

In this comment https://github.com/software-mansion/react-native-reanimated/issues/5345#issuecomment-1859198802:

>Unfortunately, with current implementation handlers provided by **useAnimatedScrollHandler** or any other handlers that are an instance of WorkletEventHandler, that being: **useScrollViewOffset**, **useEvent** and **useAnimatedGestureHandler** (last one deprecated) are supposed to be passed to only one component. This is because their class instances keep the data of the latest component only. Passing them to multiple components leads to undefined behavior.

we mention that event handlers cannot be used with more than one component. This should be documented.

|Handler|Screenshot|Where in docs|
|--|--|--|
|useAnimatedScrollHandler|image|Section "remarks"|
|useScrollViewOffset|image|Section "remarks"|
|useEvent|image|Section "returns"|
|useAnimatedGestureHandler| Deprecated ❌| No documentation ❌|

### DIFF
--- a/docs/docs/advanced/useEvent.mdx
+++ b/docs/docs/advanced/useEvent.mdx
@@ -31,7 +31,7 @@ Value indicating whether handler should be rebuilt.
 
 ### Returns
 
-The hook returns event handler that will be invoked when native event is dispatched.
+The hook returns event handler that will be invoked when native event is dispatched. That handler should be connected to only one component.
 
 ## Example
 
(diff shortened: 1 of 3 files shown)


tjzel (changes requested):
Even though `useEvent` doesn't have a high-quality documentation, I think we should add it there so we don't forget it later on.

Regarding `useScrollViewOffset`, I think we should add to the remarks section that the animatedRef parameter there mustn't be changed. E.g.

```ts
useScrollViewOffset(someState ? animatedRef1 : animatedRef2)
```

tjzel (approved):

PR #5812, opened 2024-03-20
Fix default Magnetic Sensor interval causing a security exception

LatekVo:
## Summary

Currently default magnetic sampling interval equals 2ms, which triggers a `HIGH_SAMPLE_RATE` security exception.
This fix defaults the probing interval to 8ms.

Note: `'auto'` setting defaults to `-1`, which is then replaced by `2`.

## Test plan

Download a new emulator on Android Studio - I'm using `Pixel 8 UpsideDownCake`
Launch the magnetic sensor example before, and after applying this patch.

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensor.java
+++ b/android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensor.java
@@ -17,6 +17,7 @@ public class ReanimatedSensor {
   Sensor sensor;
   ReanimatedSensorType sensorType;
   int interval;
+  public static final int MINIMAL_UNPRIVILEGED_INTERVAL = 8;
 
   ReanimatedSensor(
       WeakReference<ReactApplicationContext> reactContext,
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensor.java
```diff
@@ -17,6 +17,7 @@ public class ReanimatedSensor {
   Sensor sensor;
   ReanimatedSensorType sensorType;
   int interval;
+  private static final int MINIMAL_UNPRIVILEGED_SAMPLING_RATE = 8;
```
tomekzaw, line 20:
```suggestion
  private static final int DEFAULT_INTERVAL = 8;
```

android/src/main/java/com/swmansion/reanimated/sensor/ReanimatedSensor.java
```diff
@@ -30,7 +31,7 @@ public class ReanimatedSensor {
         (SensorManager) reactContext.get().getSystemService(reactContext.get().SENSOR_SERVICE);
     this.sensorType = sensorType;
     if (interval == -1) {
-      this.interval = SensorManager.SENSOR_DELAY_UI;
+      this.interval = MINIMAL_UNPRIVILEGED_SAMPLING_RATE;
```
tomekzaw, line 34:
```suggestion
      this.interval = DEFAULT_INTERVAL;
```

piaskowyk (approved):

LatekVo:
Note: this PR fixed both the magnetic and the accelerometer sensor.
Other sensors are not affected.

PR #5813, opened 2024-03-20
Implement  `not` for runtime tests

Latropos:
## Summary

Add possibility to negate some of the matchers. I've decided that there is no need to negate snapshot matchers, so I've handled just the other functions. Both API and implementation are inspired with jest tests.

image

## Test plan
All runtime tests work


app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -22,64 +32,121 @@ export class Matchers {
...
+  public toBeCalled = this.makeThrowingMatcher(Matchers._toBeCalledMatcher);
+  public toBeCalledUI = this.makeThrowingMatcher(Matchers._toBeCalledUIMatcher);
+  public toBeCalledJS = this.makeThrowingMatcher(Matchers._toBeCalledJSMatcher);
+
+  public not = {
+    toBe: this.makeThrowingMatcher(Matchers._toBeMatcher, true),
+    toBeCalled: this.makeThrowingMatcher(Matchers._toBeCalledMatcher, true),
+    toBeCalledUI: this.makeThrowingMatcher(Matchers._toBeCalledUIMatcher, true),
+    toBeCalledJS: this.makeThrowingMatcher(Matchers._toBeCalledJSMatcher, true),
+  };
```
piaskowyk, line 138: I think we can do it simpler by create getter for property `not`
```js
get not() {
  this._negation = true;
  return this;
}
```
In this case, we don't need to reassign methods to class field.

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -22,64 +32,120 @@ export class Matchers {
...
+      } to be called ${expected} times on ${threadName}, but was called ${received} times`,
+    };
+  };
+
+  private _toBeCalledJSMatcher: MatcherFunction = (
+    currentValue: TestValue,
+    times = 1
+  ) => {
+    Matchers._assertValueIsCallTracker(currentValue);
+    Matchers._assertValueIsCallTracker(currentValue);
```
piaskowyk, line 101: We can call it only once
```suggestion
    Matchers._assertValueIsCallTracker(currentValue);
```

piaskowyk (approved):
👏

piaskowyk (approved):

PR #5818, opened 2024-03-21
Point metro resolver for react-native to lib

### LINKED ISSUE #5789: Compiling JS failed: Invalid expression Buffer size
### Description

Starting with Version 3.4.0 I get the error `Compiling JS failed: 232124:78:invalid expression Buffer size 17924054 starts with: 766172205f5f42554444c455f535441` (see screenshot).
I stayed with version 3.3.0 for as long as I could (up to React Native 0.71) but now that I'm finally upgrading my app to React Native 0.73 I need to upgrade to at least 3.6.0 and desperately need a fix for this.

Simulator Screenshot - iPhone 15 Pro - 2024-03-13 at 13 56 55

### Steps to reproduce

1. Install "react-native-reanimated" >= 3.4.0
2. Run app on iOS

n.a.

### Reanimated version

3.8.0

### React Native version

0.73.6

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Device model

iPhone 15 Pro (iOS 17)

### Acknowledgements

Yes

DanielKuhn:
react-native-builder-bob uses standard directories for "main", "module" and "react-native" in package.json. These expect precompilation before release.
"react-native" usually points to the source directory. 
Since in the case of react-native-reanimated sometimes typescript is not transpiled correctly (see #5789 and https://github.com/facebook/react-native/issues/39296#issuecomment-1707525728) and react-native uses javascript anyways, it should be safe to point "react-native" to the compiled javascript sources instead of the typescript sources (which are referenced in "source" anyways).

## Summary

Fixes #5789

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   },
   "main": "lib/module/index",
   "module": "lib/module/index",
-  "react-native": "src/index",
+  "react-native": "lib/module/index",
   "source": "src/index",
   "types": "lib/typescript/index",
   "files": [


tjzel:
Hi Daniel, unfortunately we will not merge this pull request. The problem with transpilation comes somewhere from your App configuration, perhaps invasive dependencies. As I suggested, you can safely specify other targets that `react-native` in your `metro.config.js` if you don't have the time to investigate it.

The issue you mentioned does not appear for most of the users - and for the little few that encounter it, bumping some outdated dependencies fixes it.

React Native has TypeScript plugins for its Metro Bundler for a reason and we don't want its users experience to downgrade with this change.

PR #5843, opened 2024-03-27
Include original remote function name in "Tried to synchronously call" error message

tomekzaw:
## Summary

This PR improves developer experience by including the remote function name in the error message that is shown when trying to call remote function synchronously.

| Named function | Anonymous function |
|:-:|:-:|
| Screenshot 2024-03-27 at 16 35 31 | Screenshot 2024-03-27 at 16 34 23 |

## Test plan

ShareablesExample.tsx


Common/cpp/SharedItems/Shareables.h
```diff
@@ -265,13 +265,20 @@ class ShareableRemoteFunction
       public std::enable_shared_from_this<ShareableRemoteFunction> {
  private:
   jsi::Runtime *runtime_;
+#ifndef NDEBUG
```
mrousavy, line 268: why not just `#ifdef DEBUG`?

piaskowyk (approved):

Common/cpp/SharedItems/Shareables.h
tomekzaw: More details in PR description in #5113.

mrousavy: Ah, interesting! 

image

tjzel (approved):

PR #5844, opened 2024-03-27
Add support for cloning `Error`

tomekzaw:
## Summary

This PR adds support for cloning JS `Error` instances from RN runtime to UI runtime (or any worklet runtime).

The copied error inherits the original name, message as well as stack trace.

Initially, I also wanted to implement cloning in the opposite direction (UI &rarr; RN) but currently `makeShareableCloneOnUIRecursive` does not support initializers and most likely will be subject to change as we plan to unify the implementation of cloning across the runtimes.

## Test plan

ShareableExample.tsx

### DIFF
--- a/app/src/examples/ShareablesExample.tsx
+++ b/app/src/examples/ShareablesExample.tsx
@@ -162,7 +162,6 @@ function DataViewDemo() {
 
 function ErrorDemo() {
   const handlePress = () => {
-    // const e = new Error('error message');
     const e = new Error('error message');
     console.log(_WORKLET ? 'UI' : 'RN', e instanceof Error);
     console.log(_WORKLET ? 'UI' : 'RN', String(e));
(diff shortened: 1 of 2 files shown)


app/src/examples/ShareablesExample.tsx
```diff
@@ -159,6 +160,23 @@ function DataViewDemo() {
   return <Button title="DataView" onPress={handlePress} />;
 }
 
+function ErrorDemo() {
+  const handlePress = () => {
+    // const e = new Error('error message');
```
piaskowyk, line 165:
```suggestion
```

piaskowyk (commented):
Looks good to me 👍

piaskowyk (approved):

piaskowyk (approved):

PR #5854, opened 2024-03-30
feat: suggest useComposedScrollHandler hook

ACHP:
The goal of this hook is to be able to compose multiple scrollHandler, and therefore "hide" the fact that a custom scrollView already use a scrollHandler.

## why

Let's say I want to create a Fancy ScrollView that use `useAnimatedScrollHandler` internally to do some animation, on the background color for example.
I want to expose this new component to the team and let them use their own animations (in addition to mines) using `reanimated` primitives.

Maybe I missed something in the documentation, but as of today it seems that there is no **easy** way to do it.
We have to provide a custom API for our `FancyScrollView` to handle `onScroll`, `onMomentumScrollEnd` etc ...
This is not ideal because now the implementation details of our FancyScrollView have somehow leaked outside our component.

## Existing discussion

https://github.com/software-mansion/react-native-reanimated/discussions/3611

## Status
This is still a work in progress, I would really appreciate some feedback on this:
- Would it actually make sense to add this API in reanimated ? (Or should it be an external lib ?)
- Is this the right way to do it ?
- How to type the handlerList params properly ?

### DIFF
--- a/app/src/examples/ComposedScrollHandler/index.tsx
+++ b/app/src/examples/ComposedScrollHandler/index.tsx
@@ -1,0 +1,54 @@
+import Animated, {
+  AnimatedScrollViewProps,
+  useAnimatedScrollHandler,
+} from 'react-native-reanimated';
+import { StyleSheet, Text } from 'react-native';
+
+import React from 'react';
+import { useComposedScrollHandler } from '../../../../src/reanimated2/hook/composeAnimatedScrollHandler';
+
+export default function ComposedAnimatedScrollHandlerExample() {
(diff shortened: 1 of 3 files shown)


tjzel:
Hi @ACHP. We were discussing how to implement event handler merging in Reanimated for a while now and soon we will come to some conclusions. We have a couple of approaches here to choose from and we will include your, with adding a new hook, in the process.

szydlovsky:
Hey @ACHP ! Great thanks for suggesting the hook - we have undertaken this approach and are currently working on it in https://github.com/software-mansion/react-native-reanimated/pull/5890 - feel free to join the discussion as well!

PR #5859, opened 2024-04-02
Include factory in easing warning

Latropos:
## Summary
Include easing factories in easing warning

### DIFF
--- a/src/reanimated2/animation/util.ts
+++ b/src/reanimated2/animation/util.ts
@@ -58,15 +58,15 @@ export function assertEasingIsWorklet(
     // It is possible to run reanimated on web without plugin, so let's skip this check on web
     return;
   }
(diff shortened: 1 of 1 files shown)


src/reanimated2/animation/util.ts
```diff
@@ -58,10 +58,15 @@ export function assertEasingIsWorklet(
     // It is possible to run reanimated on web without plugin, so let's skip this check on web
     return;
   }
-  if (!isWorkletFunction(easing)) {
-    throw new Error(
-      '[Reanimated] The easing function is not a worklet. Please make sure you import `Easing` from react-native-reanimated.'
-    );
+
+  if ('factory' in easing) {
+    assertEasingIsWorklet(easing.factory());
```
tjzel, line 63: Can you make sure that it's safe to call factory here? i.e. there's no side effects related to calling it 'early'.

tjzel (approved):

PR #5863, opened 2024-04-03
Migrate to yarn4

tjzel:


.github/workflows/check-TS-react-native.yml
```diff
@@ -27,8 +29,12 @@ jobs:
           cache: 'yarn'
       - name: Clear annotations
         run: scripts/clear-annotations.sh
+      - name: Install dependencies
+        run: yarn install
+      - name: Build repo
+        run: yarn build
       - name: Install react-native ${{ matrix.react-native-version }}
-        run: yarn add react-native@${{ matrix.react-native-version }}
+        run: yarn add -D react-native@${{ matrix.react-native-version }}
```
tomekzaw, line 37: Let's use full flags for better readability, i.e. `--save-dev` here.

.github/workflows/tvos-build.yml
```diff
@@ -83,4 +85,4 @@ jobs:
 
       - name: Build app
         working-directory: ${{ env.WORKING_DIRECTORY }}
-        run: yarn react-native run-ios --no-packager --scheme TVOSExample-tvOS --simulator "Apple TV"
+        run: yarn react-native-tvos run-ios --no-packager --scheme TVOSExample-tvOS --simulator "Apple TV"
```
tomekzaw, line 88: Q: Why was this change necessary?

app/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "private": true,
   "scripts": {
-    "lint": "eslint --max-warnings=0 .",
-    "postinstall": "patch-package"
+    "build": "yarn patch-package",
```
tomekzaw, line 4: I think yarn supports patching dependencies out-of-the-box; let's migrate from `patch-package` to `yarn patch` but in a separate PR.

package.json
```diff
@@ -30,7 +30,9 @@
     "type:check:tests:0.72+": "./scripts/test-ts.sh __typetests__/0.72+",
     "type:check:tests:legacy": "./scripts/test-ts.sh __typetests__/legacy",
     "type:check:all": "yarn type:check:src && yarn type:check:plugin && ./scripts/test-ts.sh app/src/App.tsx __typetests__/common __typetests__/0.72+ __typetests__/legacy",
-    "prepare": "yarn plugin && bob build && husky install && yarn app",
+    "build": "yarn build:plugin && bob build && husky install && yarn build:app ",
```
tomekzaw, line 33:
```suggestion
    "build": "yarn build:plugin && bob build && husky install && yarn build:app",
```

package.json
```diff
@@ -30,7 +30,9 @@
...
+    "build:app": "cd app && yarn install && yarn build && cd ..",
+    "build:plugin": "cd plugin && yarn install && yarn build && cd ..",
```
tomekzaw, line 35: Not sure if `&& cd ..` at the end are required; as far as I remember yarn restores the original current working directory.

plugin/package.json
```diff
@@ -11,7 +11,7 @@
   "license": "MIT",
   "readmeFilename": "README.md",
   "scripts": {
-    "prepare": "yarn generate",
+    "build": "yarn generate",
     "generate": "yarn tsc && yarn bundle && yarn format && yarn clean",
```
tomekzaw, line 15: Can we simply rename `generate` to `build`?
```suggestion
    "build": "yarn tsc && yarn bundle && yarn format && yarn clean",
```

plugin/src/workletFactory.ts
```diff
@@ -88,7 +87,7 @@ export function makeWorkletFactory(
   // bracket would become part of the comment thus resulting in an error, since
   // there is a missing closing bracket.
   codeObject.code =
-    '(' + (isObjectMethod(fun) ? 'function ' : '') + codeObject.code + '\n)';
+    '(' + (fun.isObjectMethod() ? 'function ' : '') + codeObject.code + '\n)';
```
tomekzaw, line 90: Why do we need these changes in the Babel plugin?

.github/workflows/tvos-build.yml
tjzel: There's no package named `react-native` in this catalogue. I haven't checked why is that so, but `react-natvie-tvos` forks `react-native`'s CLI so it's what we want to use there.

.github/workflows/check-expo-dev-client-nightly.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         run: npx create-expo-app app
       - name: Install expo-dev-client
         working-directory: app
-        run: npm install expo-dev-client@next
+        run: if npm view expo dist-tags | grep -q 'next:' ; then npm install expo@next else npm install expo@latest fi
```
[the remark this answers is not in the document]
tjzel: `@next` is only available when expo releases an RC version. This CI fails all the time if it's not available - I wanted to do it some time ago, but I forgot.

package.json
tjzel: You are right, it's not necessary.

plugin/src/workletFactory.ts
```diff
@@ -26,7 +26,6 @@ import {
   isObjectExpression,
   isObjectMethod,
```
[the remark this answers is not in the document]
tjzel: Yes.

plugin/src/workletFactoryCall.ts
```diff
@@ -42,6 +42,8 @@ function addStackTraceDataToWorkletFactory(
   const originalWorkletLocation = path.node.loc;
   if (originalWorkletLocation) {
     workletFactoryCall.callee.loc = {
+      filename: originalWorkletLocation.filename,
+      identifierName: originalWorkletLocation.identifierName,
```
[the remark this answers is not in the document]
tjzel: Babel has improved its types a bit. In this case it requires for the object to be fully specified - but I already checked, it doesn't change the behavior.

plugin/src/workletFactory.ts
tjzel: Babel has improved its types a bit. Now `isSomething` method can only be applied to nodes, not traversal objects - instead you should use its' methods.

piaskowyk (approved):
I have no more comments than Tomek 👍

.github/workflows/tvos-build.yml
tomekzaw: good, thanks!

.github/workflows/check-expo-dev-client-nightly.yml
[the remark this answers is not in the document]
tomekzaw: Thanks for your explanation. This could be done in a separate PR but I don't mind it being here.

package.json
```diff
@@ -18,7 +18,7 @@
     "format:js": "prettier --write --list-different src __tests__ __typetests__",
     "format:plugin": "cd plugin && yarn format && cd ..",
     "format:java": "node ./scripts/format-java.js",
-    "format:ios": "find apple/ -iname *.h -o -iname *.m -o -iname *.mm -o -iname *.cpp | xargs clang-format -i --Werror",
+    "format:ios": "find apple/ -iname \"*.h\" -o -iname \"*.m\" -o -iname \"*.mm\" -o -iname \"*.cpp\" | xargs clang-format -i --Werror",
```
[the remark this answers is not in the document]
tomekzaw: hmm we already use `'` here: https://github.com/software-mansion/react-native-reanimated/blob/6a283cae3c20c80c1a012f748f7eafa2f2534247/package.json#L11

.github/workflows/check-TS-react-native.yml
```diff
@@ -27,8 +29,12 @@ jobs:
           cache: 'yarn'
       - name: Clear annotations
         run: scripts/clear-annotations.sh
+      - name: Install dependencies
+        run: yarn install
+      - name: Build repo
+        run: yarn build
```
tomekzaw, line 35: Please ensure that this pattern is present in all workflows prior to merging.

plugin/package.json
```diff
@@ -11,8 +11,7 @@
   "license": "MIT",
   "readmeFilename": "README.md",
   "scripts": {
-    "prepare": "yarn generate",
-    "generate": "yarn tsc && yarn bundle && yarn format && yarn clean",
+    "build": "yarn tsc && yarn bundle && yarn format && yarn clean",
```
tomekzaw, line 14: Please make sure that we don't have `yarn generate` anywhere left prior to merging.

tomekzaw (approved):
Thanks!

PR #5876, opened 2024-04-08
Test transparent colors in runtime tests

Latropos:
## Summary
Functions `getViewProp` and `getAnimatedStyle` should return the correct colour, even if given colour contains transparency like `#12345678`.

Also I update docs as below to indicate that both short hex colour and hex colours with transparency are supported.
image

## Test plan
I've added runtime tests


app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Comparators.ts
```diff
@@ -15,13 +15,19 @@ const COMPARATORS: {
     return bothAreNaN || value === expected;
   },
 
-  [ComparisonMode.COLOR]: (expected, value) => {
-    if (typeof value !== 'string' || typeof expected !== 'string') {
+  [ComparisonMode.COLOR]: (expectedUnknown, value) => {
```
piaskowyk, line 18: Can we restore the `expected` name or change name to something more accurate?

piaskowyk (approved):

PR #5851, opened 2024-03-29
Fix calculating keyboard height

maciekstosio:
## Summary

Should fix: #5811

Keyboard height seems to be improperly calculated. Based on my investigation it seems that ` WindowInsetsCompat.Type.ime()` includes height of the keyboard + bottom navigation. What Reanimated does currently is checking if navigation bar is visible (`KeyCharacterMap.deviceHasKey(KeyEvent.KEYCODE_HOME);`) and then subtracts the bottom inset `int keyboardHeight = DiphasNavigationBar ? contentBottomInset - systemBarBottomInset : contentBottomInset;`. But to my understanding `hasNavigationBar` is always true, thus when keyboard is hidden keyboard height becomes `-contentBottomInset`. So I changed the navigation height to be `Math.max(0, keyboardHeightDip)`

Additionally in WindowInsetsManager.java I applied default paddings and added padding bottom which should help correctly position element from the bottom. 

### Remarks
- on older androids < 11, hidden status bar rolls back to adjustPan keyboard behavior
- flag `isStatusBarTranslucentAndroid` doesn't seems to do anything when screen has native header (`headerShown: true`)

### Examples before vs after 

*Adding useAnimatedKeyboard breaks the layout*
|Before|After|
|-|-|
|<video src="https://github.com/software-mansion/react-native-reanimated/assets/11800297/bb3bec01-7e57-451f-8d43-2271739402eb" />|<video src="https://github.com/software-mansion/react-native-reanimated/assets/11800297/46eff754-5eea-41d0-9ffe-20657cdc4e0c" />|

*After closing the keyboard there is unexpected gap (negative height)*
|Before|After|
|-|-|
|<video src="https://github.com/software-mansion/react-native-reanimated/assets/11800297/284d0be5-5cb2-439a-b0f3-5881525f0ec6" />|<video src="https://github.com/software-mansion/react-native-reanimated/assets/11800297/5e6ebf04-6d53-4284-8ad0-d079fbe242bd" />|

*When StatusBar is hidden keyboard nor working properly on Android 10 - not fixed but not a regression*
|Before|After|
|-|-|
|<video src="https://github.com/software-mansion/react-native-reanimated/assets/11800297/75041b95-7db9-4a12-89f5-a68ab5006ea9" />|<video src="https://github.com/software-mansion/react-native-reanimated/assets/11800297/608ac6e6-9dcc-40bc-8f41-d6132e21b006" />|

## Test plan

<details>
<summary>Example showing the problem</summary>

```jsx
[...]

PR #5889, opened 2024-04-11
[WIP] Handle status bar without is status bar translucent android option

maciekstosio:
## Summary

During the discussion #5851 it was suggested that we could handle status bar without passing special flag. I attached observer that listens to StatusBar changes and keep track if the color is translucent and what is its height. Thanks to that I can change top insets automatically. *The PR is WIP - if we decide to go with it I need to remove unused flag or deprecate it*

|Before without useAnimatedKeyboard|Before with useAnimatedKeyboard|After|
|-|-|-|
|<video src="https://github.com/software-mansion/react-native-reanimated/assets/11800297/8dbd29b0-d11f-4932-9964-6d1af0b5261c" />|<video src="https://github.com/software-mansion/react-native-reanimated/assets/11800297/6ae853eb-ba17-4bea-bb80-b490c81dc7bb" />|<video src="https://github.com/software-mansion/react-native-reanimated/assets/11800297/69d5c15a-93e2-4701-a05f-d165c6730074" />|

## Test plan
Turn of navigation headers if visible, and then you use this Example:
<details>
<summary>Example</summary>

```jsx
import Animated, {
  useAnimatedKeyboard,
  useAnimatedReaction,
  useAnimatedStyle,
import {
  Platform,
  StatusBar,
  StyleSheet,
  TextInput,
  View,
  Text,
  Button,
} from 'react-native';

  const [statusBarHidden, setStatusBarHidden] = useState(false);
  const [statusBarTranslucent, setStatusBarTranslucent] = useState(false);

  const animatedStyles = useAnimatedStyle(() => ({
  }));

  useAnimatedReaction(
    () => {
      return keyboard.height.value;
    },
    (currentValue) => console.log(currentValue)
  );

  return (
    <Animated.View
      style={[
        styles.container,
        animatedStyles,
        { justifyContent: 'flex-end', borderWidth: 5, borderColor: '#ff0', backgroundColor: "#222" },
      ]}>
      <StatusBar
        barStyle="default"
        hidden={statusBarHidden}
        backgroundColor={statusBarTranslucent ? 'transparent' : undefined}
        translucent={statusBarTranslucent}
      />
      <Button
        title="Toggle statusBackgroundColor"
        onPress={() => setStatusBarTranslucent((hidden) => !hidden)}
      />
      <Button
        title="Toggle StatusBar"
        onPress={() => setStatusBarHidden((hidden) => !hidden)}
      />
      <Button
        title="Show Warning"
        onPress={() => console.warn('WARNING!!!!')}
      />
      <View
        style={[
          styles.center,
          {
            height: 200,
            backgroundColor: '#f0f',
            borderWidth: 5,
            borderColor: '#0ff',
          },
        ]}>
        <Text>{`Android ${Platform.constants['Release']}`}</Text>
        <TextInput placeholder="Test" />
      </View>
    </Animated.View>
  );
}

  container: { flex: 1 },
  center: { justifyContent: 'center', alignItems: 'center' },
});

```

</details>

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/keyboard/WindowsInsetsManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/keyboard/WindowsInsetsManager.java
@@ -9,7 +9,6 @@ import androidx.core.view.ViewCompat;
 import androidx.core.view.WindowCompat;
 import androidx.core.view.WindowInsetsCompat;
 import com.facebook.react.bridge.ReactApplicationContext;
-import com.swmansion.reanimated.BuildConfig;
 import java.lang.ref.WeakReference;
 
 public class WindowsInsetsManager {
@@ -58,24 +57,19 @@ public class WindowsInsetsManager {
   }
 
   private WindowInsetsCompat onApplyWindowInsetsListener(View view, WindowInsetsCompat insets) {
+    WindowInsetsCompat defaultInsets = ViewCompat.onApplyWindowInsets(view, insets);
     if (mKeyboard.getState() == KeyboardState.OPEN) {
       mKeyboard.updateHeight(insets);
       mNotifyAboutKeyboardChange.call();
     }
-    setWindowInsets(insets);
-    return insets;
+    setWindowInsets(defaultInsets);
+    return defaultInsets;
   }
 
   private void setWindowInsets(WindowInsetsCompat insets) {
-    int paddingBottom = 0;
-    boolean isOldPaperImplementation =
-        !BuildConfig.IS_NEW_ARCHITECTURE_ENABLED && BuildConfig.REACT_NATIVE_MINOR_VERSION < 70;
-    if (isOldPaperImplementation) {
-      int navigationBarTypeMask = WindowInsetsCompat.Type.navigationBars();
-      paddingBottom = insets.getInsets(navigationBarTypeMask).bottom;
-    }
     int systemBarsTypeMask = WindowInsetsCompat.Type.systemBars();
     int paddingTop = insets.getInsets(systemBarsTypeMask).top;
+    int paddingBottom = insets.getInsets(systemBarsTypeMask).bottom;
     updateInsets(paddingTop, paddingBottom);
   }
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/reanimated/keyboard/WindowsInsetsManager.java
```diff
@@ -14,10 +16,32 @@
...
 
+  private final ViewTreeObserver.OnDrawListener statusBarChangeListener =
+      new ViewTreeObserver.OnDrawListener() {
+        @Override
+        public void onDraw() {
+          View statusBar = getStatusBarView();
+          int currentStatusBarHeight = statusBar.getHeight();
+          boolean isStatusBarTranslucent = getWindow().getStatusBarColor() == Color.TRANSPARENT;
+
+          if (isStatusBarTranslucent != mIsStatusBarTranslucent
```
maciekstosio, line 36: This is fired few times during the changes so I don't want to do update multiple times (it blocks the UI).

maciekstosio:
This solution works, but we're investigating this in terms of performance, as in current state it may cause problems.

efstathiosntonas:
@maciekstosio recently I had to use `react-native-keyboard-controller` and `react-native-avoid-softinput` and they both had issues with `useAnimatedKeyboard` and edge to edge setup with `react-native-unistyles`. The status bar is flickering when `useAnimatedKeyboard` does it's magic causing weird layout effects. 

Both input handlers set `decorFitsSystemWindow` dynamically but on `react-native-unistyles@2.8.1` it's static -> false. 

Just wanted to bring this to your attention. cc: @jpudysz

maciekstosio:
> Using `ViewTreeObserver.OnDrawListener` is not performant. It triggers the callback multiple times. After months of debugging insets IMO there are two reliable ways to measure it:
> 
> * `WindowInsetsCompat` - but requires `decorFitsSystemWindow` set to false
> * measuring root view and decor view and compare them - but we need to handle multiple edge cases
> 
> @maciekstosio if you want I can share some knowledge here

Yeah, that's why it's on hold for so long :D Although I'm not sure if those two solution are feasible - we have decorFitsSystemWindow(false) to observe keyboard height, but as far as I remember the problem with transparent status bar is that insets do not change, just the background and I couldn't find a way to observe that (other then on every draw).
Anyway, I'm more than happy to discuss it with you, I'll write you.

piaskowyk:
Unfortunately, we've dropped support for useAnimatedKeyboard 😢

PR #5893, opened 2024-04-12
Test two ways to use sharedValue in `useAnimatedStyle`

Latropos:
## Summary
In many animations we have two ways of implementing them - we can use withTiming either inside `useAnimatedStyle` or outside. For simplicity I've decided to name test cases with `useAnimatedStyle` actively updating shared value as "ACTIVE", and if from `useAnimatedStyle` we observe it I call test case "PASSIVE". So these test cases are either ACTIVE or PASSIVE from the point of view of the hook `useAnimatedStyle`


app/src/examples/RuntimeTests/tests/snapshotTests/easing.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+    expect(activeUpdates).toMatchNativeSnapshots(activeNativeUpdates, true);
+  });
+
+  test.each([(Easing.in, Easing.out, Easing.inOut)])('Easing.%p(Easing.elastic(10))', async easing => {
+    const [activeUpdates, activeNativeUpdates, passiveUpdates] = await getSnaphotUpdates(easing(Easing.elastic(10)));
+    expect(activeUpdates).toMatchSnapshots(EasingSnapshots[easing.name as keyof typeof EasingSnapshots]);
+    expect(passiveUpdates).toMatchSnapshots(EasingSnapshots[easing.name as keyof typeof EasingSnapshots]);
+
+    expect(activeUpdates).toMatchNativeSnapshots(activeNativeUpdates, true);
+  });
```
szydlovsky, line 148: The empty lines in both `test.each` bodies seem to be inconsistent 😄

szydlovsky (approved):
Looks good to me - left just a super small cosmetic comment

PR #5901, opened 2024-04-15
Fix bridgeless for rn@0.74.0-rc.8

piaskowyk:
## Summary

This pull request adjusts the Reanimated initialisation process to changes in React Native versions between 0.74.0-rc.4 and 0.74.0-rc.8. Now, the method `setBridge` is called in both bridgeless and bridgefull modes.


apple/REAModule.mm
```diff
@@ -158,47 +159,35 @@ - (void)handleJavaScriptDidLoadNotification:(NSNotification *)notification
...
   _surfacePresenter = surfacePresenter;
+  _isBridgeless = true;
 }
 
 - (void)setBridge:(RCTBridge *)bridge
 {
-  // This method isn't called on Bridgeless mode.
   [super setBridge:bridge];
-
-  [bridge.uiManager.observerCoordinator addObserver:self];
```
piaskowyk, line 174: We don't use it on the new architecture

WoLewicki (approved):
Hope it works now 🎉

apple/REAModule.mm
tomekzaw: 1. I was wondering if we use it in RNGH to pass events to Reanimated but turns out we don't (in RNGH there's a function `sendEventForReanimated` that on new arch calls `reanimatedEventDispatcher.sendEvent()`  which in turn calls `reanimatedModule?.nodesManager?.onEventDispatch(event)` so I guess that's okay).
2. If it's not needed on the new architecture, let's wrap `<RCTEventDispatcherObserver>` in header file and `eventDispatcherWillDispatchEvent:` method inside `#ifdef RCT_NEW_ARCH_ENABLED`, what do you think?

apple/REAModule.mm
```diff
@@ -158,47 +159,35 @@ - (void)handleJavaScriptDidLoadNotification:(NSNotification *)notification
...
- * BridgelessTurboModuleSetup.
+ * This selector is invoked via BridgelessTurboModuleSetup.
  */
 - (void)setSurfacePresenter:(id<RCTSurfacePresenterStub>)surfacePresenter
 {
   _surfacePresenter = surfacePresenter;
+  _isBridgeless = true;
 }
 
 - (void)setBridge:(RCTBridge *)bridge
```
tomekzaw, line 170: It's a bit confusing that we have two separate places where we define `setBridge:` method (line 170 and line 207). I know it was like that prior to this PR. It's definitely not a blocker but we should do something about it in a separate PR.

apple/REAModule.mm
```diff
@@ -158,47 +159,35 @@ - (void)handleJavaScriptDidLoadNotification:(NSNotification *)notification
...
                                            selector:@selector(handleJavaScriptDidLoadNotification:)
                                                name:RCTJavaScriptDidLoadNotification
                                              object:nil];
 
   [[self.moduleRegistry moduleForName:"EventDispatcher"] addDispatchObserver:self];
-
-//  [bridge.uiManager.observerCoordinator addObserver:self]; // TODO: Check if it's needed on new arch.
-#ifndef NDEBUG
-  [self setReaSurfacePresenter];
-#endif // NDEBUG
```
tomekzaw, line 199: I wonder why is this code DEBUG-only? It was like that before, just curious to know if that's right.

apple/REAModule.mm
```diff
@@ -295,7 +279,7 @@ - (void)sendEventWithName:(NSString *)eventName body:(id)body
 
 RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(installTurboModule : (nonnull NSString *)valueUnpackerCode)
 {
-  if (!self.bridge) {
+  if (_isBridgeless) {
```
tsapeta, line 282: Why do you need a separate implementation for bridgeless here?

The implementation for non-bridgeless looks good to me, just don't use `[self.bridge respondsToSelector:@selector(runtime)]` check – it no longer returns `YES` even if the runtime is there. The reason is that the bridge is now an instance of `RCTBridgeProxy` that inherits from `NSProxy` which probably has a custom implementation for `respondsToSelector`, different than `NSObject`.
We know that `bridge.runtime` has been available (as long as you have a category interface) for many RN versions back, so it seems to be safe to call it.

Two more nitpicks regarding the implementation for bridgeless:
- Using `[RCTBridge currentBridge]` is very risky and I guess pretty easy to break in an app with multiple bridges (or now React instances).
- `bridge.jsCallInvoker` was initially not available in the bridge proxy, but it is now. I think there is no need to use the runtime executor in its current shape (call invoker, runtime executor and runtime scheduler might be unified in the next RN versions).

apple/REAModule.mm
piaskowyk: As we discussed offline, we refactor the whole initialization process in the following PR.

apple/REAModule.mm
piaskowyk: It shouldn't be 🫣 I fixed it in https://github.com/software-mansion/react-native-reanimated/pull/5901/commits/3e54ad02b263b1d087582efe9e4f609efaa8b30f. Also. We still need `REAInitializerRCTFabricSurface` class for the Fabric with bridge :/

FabricExample/ios/FabricExample.xcodeproj/project.pbxproj
```diff
@@ -46,24 +47,25 @@
 		13B07FB51A68108700A75B9A /* Images.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; name = Images.xcassets; path = FabricExample/Images.xcassets; sourceTree = "<group>"; };
 		13B07FB61A68108700A75B9A /* Info.plist */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.plist.xml; name = Info.plist; path = FabricExample/Info.plist; sourceTree = "<group>"; };
 		13B07FB71A68108700A75B9A /* main.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; name = main.m; path = FabricExample/main.m; sourceTree = "<group>"; };
-		192DDE7394C0421AABAD1E39 /* Poppins-ExtraBold.ttf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = undefined; includeInIndex = 0; lastKnownFileType = unknown; name = "Poppins-ExtraBold.ttf"; path = "../assets/fonts/Poppins/Poppins-ExtraBold.ttf"; sourceTree = "<group>"; };
```
tjzel, line 49: Is this coming from the upgradeHelper?

tjzel (approved):
🙏

apple/REAModule.mm
tjzel: > bridge.jsCallInvoker was initially not available in the bridge proxy, but it is now. I think there is no need to use the runtime executor in its current shape (call invoker, runtime executor and runtime scheduler might be unified in the next RN versions).

Are you telling me that all our work on refactoring Reanimated initialization pipeline to be able to dynamically utilize either `jsCallInvoker` or `runtimeExecutor` was mostly pointless? 💀

tsapeta: Yes xD

Maybe not entirely pointless, because the runtime executor might become the recommended way when the RN team figures out the final API for libraries. Right now we have two ways and both are lacking of features (e.g. synchronous executions), so I would recommend to use the call invoker and don't complicate the code too much until then.

FabricExample/ios/FabricExample.xcodeproj/project.pbxproj
piaskowyk: This specific change wasn't intentional; it was autogenerated during pod installation. Kacper added those fonts some time ago.

PR #5906, opened 2024-04-16
Add skip and only for tests

Latropos:
## Summary

Implement skip and only for tests.
Especially `only` is important when creating new tests, since you generally want to add a decorator to your test instead of commenting out most of the file to run only one test.

- `only` - if `only` decorator is used in any testfile, then only the decorater tests and suites will be executed
  - `describe.only`
  - `test.only`
  - `test.only.each`
- skip - skip decorated tests and suites
  - `describe.skip`
  - `test.skip`
  - `test.skip.each`

Additionally the number of the skipped tests will be displayed.
## Test plan

image

<details><summary>Testing code</summary>

```tsx
describe('Big wardrobe', () => {
  test('Test of wardrobe', async () => {
    await render(<View />);
    expect(1).toBe(1);
  });

  describe('Shelf 1', () => {
    test('Test of Shelf 1', async () => {
      await render(<View />);
      expect(1).toBe(1);
    });
    describe('Box 1 on Shelf 1', () => {
      describe('Book 1 in the Box 1 on Shelf 1', () => {
        test('Test of Book 1 in the Box 1  on Shelf 1', async () => {
          await render(<View />);
          expect(1).toBe(1);
        });
      });
      describe('Book 2 in the Box 1 on Shelf 1', () => {
        test('Test of Book 2 in the Box 1  on Shelf 1', async () => {
          await render(<View />);
          expect(1).toBe(1);
        });
      });
    });
  });
  describe('Shelf 2', () => {
    test('Test of Shelf 2', async () => {
      await render(<View />);
      expect(1).toBe(1);
    });
  });
  describe('Shelf 3', () => {
    test('Test of Shelf 3', async () => {
      await render(<View />);
      expect(1).toBe(1);
    });
    describe('Box 1 on Shelf 3', () => {
      describe('Book 1 in the Box 1 on Shelf 3', () => {
        test('Test of Book 1 in the Box 1  on Shelf 3', async () => {
          await render(<View />);
          expect(1).toBe(1);
        });

        test.each(['Page1', 'Page2', 'Page3'])('Test of %p in Box 1 on Shelf 3', async () => {
          await render(<View />);
          expect(1).toBe(1);
        });
      });
      describe('Book 2 in the Box 1 on Shelf 3', () => {
        test('Test of Book 2 in the Box 1  on Shelf 3', async () => {
          await render(<View />);
          expect(1).toBe(1);
        });
      });
    });
  });
});
```
</summary>


app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
```diff
@@ -172,40 +208,55 @@ export class TestRunner {
...
+    console.log(`${indentNestingLevel(testSuite.nestingLevel)} ${testSuite.name}`);
 
-    testSuite.buildSuite();
     if (testSuite.beforeAll) {
       await testSuite.beforeAll();
     }
 
     for (const testCase of testSuite.testCases) {
-      await this.runTestCase(testSuite, testCase, summary);
+      if ((!this._includesOnly || testSuite.only || testCase.only) && !testCase.skip) {
```
piaskowyk, line 255: It seems a bit too complicated. We currently have four different flags to determine if a test case should be executed or not: `this._includesOnly`, `testSuite.only`, `testCase.only`, and `testCase.skip`. What we can do is simplify this by using just one flag, for instance `testCase.skip`. Instead of removing test cases using `this._testSuites.splice`, each test case can be explicitly assigned the `skip` property. This way, we only need to check if a test should be skipped in one specific location. Currently, the `runTestSuite` function has two additional early returns to handle skipping.

In summary, let's simplify things. We will utilize a single flag to decide whether a test case should be executed or not, and we will check this flag in just one place.

piaskowyk (changes requested):

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
Latropos: The `splice` function is not needed to remove tests from test suits, but to insert them in order. Unfortunately, since we build testSuites before running them, it is needed to keep them in order. 
When we append each test suite at the end the order of the nested suites is wrong.

Latropos: I've added an additional loop to iterate over all the decorators and correctly assign skip depending on other flags.

Latropos: I can't get rid of the splice function inside `describe`, since the order of tests would be wrong. Yet I can propose the following changes to simplify the code, which I would like to implement:
* Move logic to find index in describe into separate util file
* Create folder testRunner, and rename testRunner into testRunner/index.ts, so we could store all testRunnerHelpers in one place (asserts, Matchers, stringFormatUtils and so on)
* Have all the decorators grouped in additional object
* Move function printSummary to stringUtils etc.
* Move notificationRegistry and the related functions to a separate file.

I would like to do it in separate PR to reduce the number of conflicts with my implementation of decorators `failing` i `warning`

piaskowyk (approved):

Issue #5912, opened 2024-04-18
react-native-reanimated 3.9.0-rc.1 fails to build on iOS with new architecture enabled in Release build


timbocole:
### Description

When building an application for iOS with new architecture enabled, with a non-release build, compilation fails with:
```
❌  app/node_modules/react-native-reanimated/apple/REAModule.mm:179:9: no visible @interface for 'REAModule' declares the selector 'setReaSurfacePresenter'

  [self setReaSurfacePresenter];
```
This is due to #5901 removing the conditional `#ifndef NDEBUG` around the invocation of `setReaSurfacePresenter`, however the definition of this api is still wrapped in a  conditional`#ifndef NDEBUG`  block.

Reapplying the conditional to the invocation resolves the problem however I am unsure of expected behaviour here.

### Steps to reproduce

Attempt to build a NON-RELEASE app for iOS using new architecture.

### Snack or a link to a repository

NONE

### Reanimated version

3.9.0-rc.1

### React Native version

0.74.0-rc.9

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

None

### Acknowledgements

Yes

piaskowyk:
@timbocole thanks a lot for the report 🙌

timbocole:
> @timbocole thanks a lot for the report 🙌

Thanks for the fix - I can see now that I totally inverted the report though as was actually using a Release config without realising and go thrown by the double negative in the code 😂

PR #5925, opened 2024-04-22
Fix production build

piaskowyk:
## Summary

This PR fixes production build for iOS. In https://github.com/software-mansion/react-native-reanimated/pull/5901 I called `setReaSurfacePresenter` method but this one was defined only on debug build.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/5912

## Test plan

Run production build for iOS


bartlomiejbloniarz (approved):
🪨

PR #5934, opened 2024-04-24
fix: copy state in reanimated commit hook

WoLewicki:
## Summary

Fix for the issue: https://github.com/Expensify/App/issues/40048 which root cause of seems to be the not copied state of `ScrollView` when this code is not added. I'll try to provide a simple repro when I get some time to do it.


tomekzaw (approved):
LGTM, thanks for the PR!

piaskowyk (approved):

tomekzaw:
Just wondering if we should also add explicit state here?

https://github.com/software-mansion/react-native-reanimated/blob/e1eedbb85a284ca4e3f9888378b7251784dbafbb/Common/cpp/Fabric/ShadowTreeCloner.cpp#L29

tomekzaw (changes requested):
I think we should pass state explicitly there as well

WoLewicki:
Done

PR #5942, opened 2024-04-26
Make Reanimated more compliant with rules of React

tjzel:

### DIFF
--- a/app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsRunner.tsx
+++ b/app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsRunner.tsx
@@ -1,15 +1,13 @@
 import { View, Button, StyleSheet } from 'react-native';
-import React, { ReactNode, useEffect, useState } from 'react';
+import React, { ReactNode, useEffect, useRef, useState } from 'react';
 import { runTests, configure } from './RuntimeTestsApi';
-import { LockObject } from './types';
-
-let renderLock: LockObject = { lock: false };
 
 export default function RuntimeTestsRunner() {
+  const isRenderLocked = useRef<boolean>(false);
   const [component, setComponent] = useState<ReactNode | null>(null);
   useEffect(() => {
-    if (renderLock) {
-      renderLock.lock = false;
+    if (isRenderLocked.current) {
+      isRenderLocked.current = false;
     }
   }, [component]);
   return (
@@ -17,7 +15,7 @@ export default function RuntimeTestsRunner() {
       <Button
         title="Run tests"
         onPress={async () => {
-          renderLock = configure({ render: setComponent });
+          isRenderLocked.current = configure({ render: setComponent }).lock;
           await runTests();
         }}
       />
--- a/src/reanimated2/component/FlatList.tsx
+++ b/src/reanimated2/component/FlatList.tsx
@@ -75,8 +75,9 @@ const FlatListForwardRefRender = function <Item = any>(
   // react-native defaults it to 50 for FlatLists.
   // We set it to 1, so we have peace until
   // there are 960 fps screens.
-  if (!('scrollEventThrottle' in restProps)) {
-    restProps.scrollEventThrottle = 1;
+  const copiedProps = { ...restProps };
+  if (!('scrollEventThrottle' in copiedProps)) {
+    copiedProps.scrollEventThrottle = 1;
   }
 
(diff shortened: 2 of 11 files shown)


app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsRunner.tsx
```diff
@@ -1,23 +1,21 @@
 import { View, Button, StyleSheet } from 'react-native';
-import React, { ReactNode, useEffect, useState } from 'react';
+import React, { ReactNode, useEffect, useRef, useState } from 'react';
 import { runTests, configure } from './RuntimeTestsApi';
-import { LockObject } from './types';
-
-let renderLock: LockObject = { lock: false };
 
 export default function RuntimeTestsRunner() {
+  const isRenderLocked = useRef<boolean>(false);
```
tjzel, line 6: cc @Latropos

How to properly define this lock inside this component?

src/reanimated2/component/ScrollView.tsx
```diff
@@ -25,27 +24,22 @@ const AnimatedScrollViewComponent = createAnimatedComponent(ScrollView);
...
-
-    if (scrollViewOffset) {
-      // eslint-disable-next-line react-hooks/rules-of-hooks
-      useScrollViewOffset(animatedRef, scrollViewOffset);
-    }
+    const animatedRef = useAnimatedRef<AnimatedScrollView>();
+    const outRef = ref ?? animatedRef;
+
+    // @ts-expect-error TODO: Allow null as `useScrollViewOffset` ref argument.
+    useScrollViewOffset(null, scrollViewOffset);
```
tjzel, line 31: cc @szydlovsky

Pls make it work in `useScrollViewOffset` (we are prisoners of our own API 💀)

src/reanimated2/hook/useFrameCallback.ts
```diff
@@ -41,13 +41,16 @@ export function useFrameCallback(
   });
 
   useEffect(() => {
+    // Unregister previous callback.
+    frameCallbackRegistry.unregisterFrameCallback(ref.current.callbackId);
```
tjzel, line 45: Was unregistering here overlooked or is it not necessary? cc @tomekzaw @piaskowyk

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsRunner.tsx
Latropos: Isn't it outdated already?

src/reanimated2/component/ScrollView.tsx
szydlovsky: @tjzel do we really need nullable ref here that badly? Changing it will really complicate the hook implementation. The current native implementation also throws a warning when there is no ref - which I remember may have been your idea. After all, this hook with a null ref does essentially nothing so what do we need it for?

tjzel: Definitely we don't need to use nullable ref here. However:

1. We shouldn't call this hook conditionally like before.
2. The user can pass to `Animated.ScrollView` either:
    2.1 An AnimatedRef - then we have no problems whatsoever.
    2.2 A plain ref from React - then it will cause a runtime error.

In scenario 2.2 we would need to check if the passed ref is not an Animated one. If it's not, my idea would be to pass `null` to `useScrollViewOffset` hook as a special case that it shouldn't do anything. Keep in mind that we cannot just pass `animatedRef` to it as for now, since in this scenario it's never getting attached to anything and causes another error.

szydlovsky: @tjzel get yourself a new main 👍

tjzel: 🥳

src/reanimated2/component/FlatList.tsx
```diff
@@ -91,7 +92,7 @@ const FlatListForwardRefRender = function <Item = any>(
     // @ts-expect-error In its current type state, createAnimatedComponent cannot create generic components.
     <AnimatedFlatList
       ref={ref}
-      {...restProps}
+      {...copiedProps}
```
mrousavy, line 95: Instead of copying, we could just also pass it here; 
```suggestion
      {...restProps}
      scrollEventThrottle={restProps.scrollEventThrottle ?? 1}
```
no?

PR #5953, opened 2024-04-29
Prevent iOS bridgeless initialization from getting the wrong bridge

### LINKED ISSUE #5497: Crash with expo custom build (49.x or 50.x, or 51.x) when newArchEnabled set to true on iOS and with expo-dev-client installed
### Description

When reloading (using `r`) and sometimes when starting the app, the app crash (on iOS, not Android) as soon as react-native-reanimated is installed, even if not used inside app, with the newArchEnabled.

As soon as your remove `react-native-reanimted` and `expo prebuild --clean` and `yarn ios` the app will no longer crash on boot or reload.

Just an fyi, you no longer need to add 'react-native-reanimated/plugin' in your babel.config.js with expo 50 as per their latest blog post.

### Steps to reproduce

1. yarn create expo-app --template tabs@beta
2. yarn expo install --fix    
3. yarn expo install expo-build-properties
4. add to app.json
```
      [
        "expo-build-properties",
        {
          "ios": {
            "newArchEnabled": true
          },
          "android": {
            "newArchEnabled": true
          }
        }
      ]
```
5. yarn expo install expo-dev-client
6. Change  "react-native-safe-area-context": "4.8.1",
7.  Change     "react-native-screens": "3.29.0",
8. Add this to package.json
```
  "expo": {
    "install": {
      "exclude": [
        "react-native-safe-area-context",
        "react-native-screens"
      ]
    }
  }
```
9. Change this inside package.json
```
    "android": "expo run:android",
    "ios": "expo run:ios",
```
10. yarn expo install react-native-reanimated
11. yarn ios

https://github.com/sync/new-arch-crash

### Reanimated version

3.6.1

### React Native version

0.73.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Device model

iPhone 15 Pro

### Acknowledgements

Yes

gabrieldonadel:
## Summary

As I pointed out [here](https://github.com/software-mansion/react-native-reanimated/pull/5901#discussion_r1567743828), using `[RCTBridge currentBridge]` may lead to the wrong bridge in apps that have multiple React instances, as is the case for expo-dev-client.

https://github.com/software-mansion/react-native-reanimated/blob/7e1a19ed8190763456e96939eec71c2528c1eed4/apple/REAModule.mm#L284

Fixes #5497 

Instead of using `[RCTBridge currentBridge]` we can just call `self.bridge` and ensure the correct bridge is always used 

## Test plan

Run FabricExample app on iOS

### DIFF
--- a/apple/REAModule.mm
+++ b/apple/REAModule.mm
@@ -281,7 +281,7 @@ RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(installTurboModule : (nonnull NSString *)
 {
   if (_isBridgeless) {
 #if REACT_NATIVE_MINOR_VERSION >= 74 && defined(RCT_NEW_ARCH_ENABLED)
-    RCTCxxBridge *cxxBridge = (RCTCxxBridge *)[RCTBridge currentBridge];
+    RCTCxxBridge *cxxBridge = (RCTCxxBridge *)self.bridge;
     auto &rnRuntime = *(jsi::Runtime *)cxxBridge.runtime;
     auto executorFunction = ([executor = _runtimeExecutor](std::function<void(jsi::Runtime & runtime)> &&callback) {
       // Convert to Objective-C block so it can be captured properly.


tomekzaw (approved):
Thanks for the PR!

tomekzaw:
I wonder if there's anything we can do to prevent us from using `currentBridge` in the future, maybe some CI checks?

tsapeta:
> I wonder if there's anything we can do to prevent us from using `currentBridge` in the future, maybe some CI checks?

Why would you need that? This API is going to be deprecated and removed anyway, like everything that related to the bridge. Also, it's always risky to depend on global things like this. See the comment for `currentBridge`: https://github.com/facebook/react-native/blob/6262158c11709ea0cbc7603732608046296f39e0/packages/react-native/React/Base/RCTBridge.mm#L252-L261

tomekzaw:
> Why would you need that?

To prevent future us from using this API.

> This API is going to be deprecated and removed anyway

Okay, that sounds good

piaskowyk (approved):

mym0404:
I'm looking for a way to pass a bridge from `ViewManager` to `View` in Fabric.

It seems this code is not run in Fabric `ViewManager` class. 

```objc
- (UIView*)view {
  RNCNaverMapViewImpl* ret = [[RNCNaverMapViewImpl alloc] init];
  ret.bridge = self.bridge;

  return ret;
}
```

Is there any example inject `Bridge` into `View` from Fabric `ViewManager`? Any suggestion or example github repo is welcome.

tsapeta:
@mym0404 As far as I know, there is no way to achieve this. Fabric creates the view instance itself. I think this is intentional as the views shouldn't have any dependencies other than props in order to work properly with the view recycling mechanism.

mym0404:
@tsapeta understood, thank you for response.

Issue #4730, opened 2023-07-13
On android useAnimatedStyle fails when Animated.FlatList has the onRefresh property.


LhZzZz:
### Description

```
const opacityStyle = useAnimatedStyle(() => {
    console.warn('inited', inited.value);
    return {
      opacity: inited.value ? 1 : 0,
    };
  });

<Animated.FlatList
    onRefresh={props.onRefresh}
    refreshing={props.refreshing}
    data={props.data}
    renderItem={props.renderItem}
    style={[{flex: 1}, opacityStyle]}
/>

```

the opacityStyle not working but if i remove onRefresh it works again. it only happens on android.

### Steps to reproduce

1. set onRefresh  and animated style for a Animated.FlatList
2. animated style not working
3. remove onRefresh
4.  animated style works

### Snack or a link to a repository

https://gitlab.com/dogofe/enhance-tab-view.git

### Reanimated version

2.17.0

### React Native version

0.71.2

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

Android emulator

### Acknowledgements

Yes

LhZzZz:
> Hey! 👋
> 
> The issue doesn't seem to contain a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example).
> 
> Could you provide a [snack](https://snack.expo.dev/) or a link to a GitHub repository under your username that reproduces the problem?

https://snack.expo.dev/@lhz19960823/tactless-pretzel

Latropos:
Hey! I've tested your code and everything works perfectly with newest reanimated
Please update your react-native-reanimated to fix the problem.

https://github.com/software-mansion/react-native-reanimated/assets/56199675/d449ec4b-7d57-48e0-aa5b-3df90b3541a2

Latropos:
@LhZzZz Thank you for your submitting the bug. (Your silver kitten is very beautiful) The bug appears on android devices only. We are investigating the problem, and I hope we will solve it soon.

Latropos:
The reason is probably that when you have a FlatList with refreash on android an extra view for refresh is added under the hood. This particular view is a native-only one and probably it causes the problem. We copy it's style to inner component that can be animated. 

As you can see in this quickly made repro:
1. the bottom square and the flatlist have the very same animated style 
```
  const opacityStyle = useAnimatedStyle(() => {
    return {
      opacity: 0.5,
[...]

PR #5960, opened 2024-04-30
Make animated components use different tags for events

szydlovsky:
## Summary

Turns out the usual component tag (used for Layout Animations, Shared Element Transitions and Animated Styles) won't always work for events if the component with nested event emitters is made animated. Thus I am splitting these two into different tag variables.

## Test plan

You can check out the following code (logs in console): 

``` TYPESCRIPT
import React from 'react';
import Animated, { useAnimatedScrollHandler } from 'react-native-reanimated';
import { StyleSheet, View, Text } from 'react-native';
import { FlashList } from '@shopify/flash-list';

const AnimatedFlashList = Animated.createAnimatedComponent(FlashList);

const data = new Array(45);

export default function EmptyApp() {
  const scrollHandler = useAnimatedScrollHandler(
    {
      onBeginDrag: () => {
        'worklet';
        console.log('ON BEGIN DRAG');
      },
      onScroll: () => {
        'worklet';
        console.log('ON SCROLL');
      },
      onMomentumEnd: () => {
        'worklet';
        console.log('ON END');
      },
    },
    []
  );

  const renderFlashListItem = () => {
    return <View style={styles.itemFlash} />;
  };

  const renderFlatListItem = () => {
    return <View style={styles.itemFlat} />;
  };

  return (
      <View style={styles.listContainer}>
        <Text>Animated FlashList</Text>
        <AnimatedFlashList
          onScroll={scrollHandler}
          renderItem={renderFlashListItem}
          data={data}
          estimatedItemSize={40}
        />
      </View>
      <View style={styles.listContainer}>
        <Text>Animated FlatList</Text>
        <Animated.FlatList
          onScroll={scrollHandler}
          renderItem={renderFlatListItem}
          data={data}
        />
      </View>
    </View>
  );
}

  container: {
    flex: 1,
    flexDirection: 'row',
  },
  listContainer: {
    flex: 1,
    flexDirection: 'column',
  },
  itemFlash: {
    width: 20,
    height: 20,
    backgroundColor: 'red',
    margin: 10,
  },
  itemFlat: {
    width: 20,
    height: 20,
    backgroundColor: 'green',
    margin: 10,
  },
});

```
</details>

Current tests work on:

- [x] Paper
- [x] Fabric
- [x] Web

### DIFF
--- a/src/createAnimatedComponent/commonTypes.ts
+++ b/src/createAnimatedComponent/commonTypes.ts
@@ -96,7 +96,8 @@ export interface AnimatedComponentRef extends Component {
 export interface IAnimatedComponentInternal {
   _styles: StyleProps[] | null;
   _animatedProps?: Partial<AnimatedComponentProps<AnimatedProps>>;
-  _viewTag: number;
+  _componentViewTag: number;
+  _eventViewTag: number;
   _isFirstRender: boolean;
   jestAnimatedStyle: { value: StyleProps };
   _component: AnimatedComponentRef | HTMLElement | null;
--- a/src/createAnimatedComponent/createAnimatedComponent.tsx
+++ b/src/createAnimatedComponent/createAnimatedComponent.tsx
@@ -122,7 +122,8 @@ export function createAnimatedComponent(
   {
     _styles: StyleProps[] | null = null;
     _animatedProps?: Partial<AnimatedComponentProps<AnimatedProps>>;
-    _viewTag = -1;
+    _componentViewTag = -1;
+    _eventViewTag = -1;
     _isFirstRender = true;
     jestAnimatedStyle: { value: StyleProps } = { value: {} };
     _component: AnimatedComponentRef | HTMLElement | null = null;
@@ -143,7 +144,8 @@ export function createAnimatedComponent(
     }
 
     componentDidMount() {
-      this._viewTag = this._getViewInfo().viewTag as number;
+      this._setComponentViewTag();
+      this._setEventViewTag();
       this._attachNativeEvents();
       this._jsPropsUpdater.addOnJSPropsChangeListener(this);
       this._attachAnimatedStyles();
@@ -185,7 +187,10 @@ export function createAnimatedComponent(
       if (this.props.sharedTransitionTag) {
         this._configureSharedTransition(true);
       }
-      this._sharedElementTransition?.unregisterTransition(this._viewTag, true);
+      this._sharedElementTransition?.unregisterTransition(
+        this._componentViewTag,
+        true
+      );
 
       const exiting = this.props.exiting;
       if (
@@ -209,7 +214,7 @@ export function createAnimatedComponent(
             : getReduceMotionFromConfig();
         if (!reduceMotionInExiting) {
           updateLayoutAnimations(
-            this._viewTag,
+            this._componentViewTag,
             LayoutAnimationType.EXITING,
             maybeBuild(
               exiting,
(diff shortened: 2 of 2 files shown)


gronxb:
I haven't checked detail, but `gorhom/react-native-bottom-sheet` was behaving strangely after applying that patch.

szydlovsky:
@gronxb if you were able to provide a small example where the react-native-bottom-sheet behaves strangely I’d be able to check it out and make sure it works as well

gronxb:
I've been busy and haven't prepared a repro, I'll do it as soon as I can.

When calling Pressable onPress from within the bottom sheet, the bottom sheet is forced down.

tjzel (changes requested):
While it's fine for me that we might have two tags, first for events and the second for other operations, I think that we should focus on the "why" here.

1. How comes that we actually need two different tags? Do all operations that require `_componentViewTag` work properly when `_eventViewTag` is correct

2. Why didn't we need it before? Which change exactly forces us to make the split here? Or maybe it was always like this and only now we have detected it?

3. What about `useScrollViewOffset` here? Is it always able to get the tag it actually requires?

We definitely need to try to answer those questions.

gronxb:
@szydlovsky 

Sorry for the delay, here's the reproduction, I think it's the backdrop, if you press anywhere on the bottom sheet, it forces it to go down.

If I don't apply that patch, it only goes down normally when I press backdrop. 

It seems like something conflicts and thinks backdrop is pressed even if you press anywhere.

https://github.com/gronxb/reanimated-repro/blob/main/App.tsx

* Apply Patch 

https://github.com/software-mansion/react-native-reanimated/assets/41789633/9bdc88fa-de0e-4bad-a79b-e08a86145749

* No Patch

https://github.com/software-mansion/react-native-reanimated/assets/41789633/ead7ab65-8a66-4f18-a576-be11797f9a29

szydlovsky:
@gronxb alright I found the issue, could you please try out this patch - 
[react-native-reanimated+3.11.0.patch](https://github.com/software-mansion/react-native-reanimated/files/15236996/react-native-reanimated%2B3.11.0.patch)

gronxb:
@szydlovsky  it works fine 👍

szydlovsky:
@tjzel `@gorhom/bottom-sheets` now works fine with the fix AND turns out `useScrollViewOffset` gets its tags very well for all scenarios (Web/Paper/Fabric) - it didn't need any changes. Therefore it is now reviewable again

tjzel:
@szydlovsky I assume you tested it natively on iOS only? Android can have a different view hierarchy, please check it too.

Is this PR something that could be verified by runtime tests?

szydlovsky:
@tjzel I tested it on all: iOS and Android (both Paper and Fabric) and Web. Works fine everywhere. If it comes to runtime tests, we would have to use `flash-list` and `botom-sheets` as dependencies so I am not so sure - taking under consideration that this fix is quite highly demanded.

tjzel:
Oh, that's great then! We can list those as dev dependencies, that's not a problem, but let's not incorporate 3rd party libraries into our runtime tests yet. I will add it as a potential task for the future.

tjzel (approved):
Looks good to me, but I'd like @piaskowyk or @tomekzaw to take a look at it as well.

tomekzaw (approved):

ChadyAyoub4:
@szydlovsky Works perfectly.

sraut1:
@szydlovsky I am facing this issue https://github.com/software-mansion/react-native-reanimated/issues/4730. 
should this patch fix this one too?

szydlovsky:
@sraut1 This patch will be in the next release ***after*** 3.11. And no, this one most likely doesn't fix the issue since it has nothing to do with animated styles. Nevertheless, feel free to [use the patch](https://github.com/software-mansion/react-native-reanimated/pull/5960#issuecomment-2098548412), maybe it sorts something out.

sraut1:
@szydlovsky the patch fixes issue https://github.com/software-mansion/react-native-reanimated/issues/4730. thanks

szydlovsky:
@sraut1 woah, I'll let the ppl know then 🤯

PR #5980, opened 2024-05-06
fix: build issues on visionOS

okwasniewski:
## Summary

This PR fixes build issues for visionOS. The `CACurrentMediaTime` API comes from QuartzCore which was not imported in this file. Unfortunately, visionOS is more strict and requires explicit imports, this shouldn't cause any issues on other platforms.

## Test plan

Build the app for visionOS

### DIFF
--- a/apple/REASlowAnimations.mm
+++ b/apple/REASlowAnimations.mm
@@ -1,6 +1,7 @@
 #import <RNReanimated/REASlowAnimations.h>
 #if TARGET_IPHONE_SIMULATOR
 #import <dlfcn.h>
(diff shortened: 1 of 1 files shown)


tomekzaw:
@okwasniewski Thanks for the PR!

I've noticed that you've added an import in `#if TARGET_IPHONE_SIMULATOR` which is set only on simulators, is this intended?

okwasniewski:
Thanks for checking the PR @tomekzaw ! You are right this import should be added for every configuration not only simulators. I've moved the import 👍🏻

bartlomiejbloniarz (approved):

PR #5987, opened 2024-05-07
Change navbar to a reusable component

patrycjakalinska:
In order to use navbar and footer theme component from `@swmansion/t-rex-ui`, we need to rewrite current navbar to be more simplified. To achieve that we moved landing background from `Navbar` to `Hero` component and footer svg to `FooterBackground`.

Screenshot 2024-05-14 at 10 47 45
Screenshot 2024-05-14 at 10 47 57


piaskowyk (approved):
🎉

PR #5988, opened 2024-05-07
Add border radii to snapshots

bartlomiejbloniarz:
## Summary
This PR adds the possibility to animate the border radius of all 4 corners separately with shared elements transitions.

## Test plan

Check the behavior of `BorderRadiiExample` and also check for regressions in other SET/LA examples.


piaskowyk (approved):
🫡

android/src/main/java/com/swmansion/reanimated/ReactNativeUtils.java
```diff
@@ -5,33 +5,68 @@
...
   private static Field mBorderRadiusField;
+  private static Method getCornerRadiiMethod;
 
-  public static float getBorderRadius(View view) {
+  public static class BorderRadii {
+    public float full, topLeft, topRight, bottomLeft, bottomRight;
+
+    public BorderRadii(
+        float full, float topLeft, float topRight, float bottomLeft, float bottomRight) {
+      this.full = full;
```
tomekzaw, line 21: Do we need to call `Float.isNaN` here as well?

tomekzaw (approved):

PR #5990, opened 2024-05-08
Change the docs to mention, that `.springify()` works with `.duration()`

Latropos:
## Summary
Right now in our docs [[LINK](https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/entering-exiting-animations#modifiers)] we say, thet `springify` doesn't work with duration, however it does. So I've moved it from T**ime-based** modifiers  section into **common** ones.

## Test plan

See in the recording that springified layout can have different duration.
https://github.com/software-mansion/react-native-reanimated/assets/56199675/519eacb6-d011-4700-89fc-a62ac6b93684


kacperkapusciak (approved):
LGTM, good catch!

PR #5991, opened 2024-05-08
Move spellcheck action to precommit

tjzel:
## Summary

Spellcheck configuration is far from acceptable and reports numerous false positives. Since it looks like task that's a bit time consuming, I decided to move this action to precommit hook and only trigger it on changed files. We might restore it in the future if we consider it necessary.

## Test plan

🚀


piaskowyk (approved):

tomekzaw (approved):
Let it be

PR #5992, opened 2024-05-08
Monorepo 

tjzel:
## About

This pull request finally moves Reanimated repo to desired monorepo structure.

The migration process will be divided into steps, so then code review would be feasible.

## Summary

This particular pull request does only the following:

- Creates a monorepo structure
- Moves current contents of the repository, excluding `.git`, to `packages/react-native-reanimated`.
- Creates a copy of the `LICENSE, .gitignore` in the root of the repository.

When reviewing, please only take those changes into account - clone this repository and browse the files.

After this PR monorepo's functionalities are not utilized. `packages/react-native-reanimated` still has its own `node_modules` and `yarn.lock`, as well as embedded packages. This will be changed in the following PRs.

Also, a number of scripts/CI won't work either due to path differences. This will also be amended in the upcoming PRs.

## Test plan

Run `yarn` in the root of the repository, `yarn && yarn build && yarn build:app` in `packages/react-native-reanimated` and build the Example app as usual. It's working as expected.

## Steps achieved towards monorepo

- [ ] All CI pass
- [x] Precommit scripts work properly
- [x] react-native-reanimated is a separate workspace which uses hoisted node_modules
- [x] app is a separate workspace which uses hoisted node_modules
- [x] paper example is a separate workspace which uses hoisted node_modules
- [x] fabric example is a separate workspace which uses hoisted node_modules
- [x] web example is a separate workspace which uses hoisted node_modules
- [x] macos example is a separate workspace which uses hoisted node_modules
- [x] next example is a separate workspace which uses hoisted node_modules
- [x] reanimated eslint plugin is a separate workspace which uses hoisted node_modules
- [x] the npm package builds and works as expected
- [x] reanimated2 directory is removed
- [x] TypeScript, ESLint and Prettier configs are unified and only extended when necessary in workspaces


piaskowyk (approved):
🔥

PR #5993, opened 2024-05-08
Monorepo [0/?] - move move github actions to root

tjzel:
## About

This pull request is a part of the effort of migrating to monorepo structure:
- #5992

## Summary

Moving `.github` directory to the root of the project, alongside with fixes that regard new repository structures.

This pull request won't fix CIs which are being tackled in
- #5996 

## Test plan

No CI should fail on a step related to the change of the monorepo structure.

## Steps achieved towards monorepo

- [ ] All CI pass
- [ ] Precommit scripts work properly
- [ ] react-native-reanimated is a separate workspace which uses hoisted node_modules
- [ ] app is a separate workspace which uses hoisted node_modules
- [ ] paper example is a separate workspace which uses hoisted node_modules
- [ ] fabric example is a separate workspace which uses hoisted node_modules
- [ ] web example is a separate workspace which uses hoisted node_modules
- [ ] macos example is a separate workspace which uses hoisted node_modules
- [ ] example is a separate workspace which uses hoisted node_modules
- [ ] reanimated eslint plugin is a separate workspace which uses hoisted node_modules
- [ ] the npm package builds and works as expected
- [ ] reanimated2 directory is removed
- [ ] TypeScript, ESLint and Prettier configs are unified and only extended when necessary in workspaces


.github/workflows/android-build.yml
```diff
@@ -41,20 +45,24 @@ jobs:
...
-          distribution: 'zulu'
+          distribution: "zulu"
           java-version: 17
 
       - name: Restore Reanimated node_modules from cache
         uses: actions/cache@v3
         with:
-          path: node_modules
-          key: ${{ runner.os }}-android-node-modules-reanimated-${{ hashFiles('yarn.lock') }}
+          path: packages/react-native-reanimated/node_modules
```
tjzel, line 60: cc @tomekzaw

If you could explain to me (preferably offline) how does the memoization work here.

PR #5995, opened 2024-05-09
Test predefined entering animation

Latropos:
## Summary

The tests require about 1minute 40s to run.

The new tests:
* verify that all the exported predefined entering animations have stable behaviour
* verify that they work with our most important modifiers: springify & duration

Also I improve the way I deep copy updates

## Test plan

image


app/src/examples/RuntimeTests/tests/layoutAnimations/entering/predefinedEntering.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+  ZoomInDown,
+  ZoomInEasyDown,
+  ZoomInEasyUp,
+  ZoomInLeft,
+  ZoomInRight,
+  ZoomInRotate,
+  ZoomInUp,
+];
+
+const ENTERING_SETS: Array<[string, unknown[], number]> = [
```
szydlovsky, line 76: Is there any advantage of `unknown[]` over `any[]`? Just curious since later on we treat each Entering class as `any`.

szydlovsky (commented):
Looks great, I've ran the tests from the branch myself and 125 tests got skipped - is that supposed to be that way?
image

app/src/examples/RuntimeTests/tests/layoutAnimations/entering/predefinedEntering.test.tsx
Latropos: Not really, however its considered to be a good custom to always use `unknown` instead of `any` if possible.

Latropos:
@szydlovsky 
> Looks great, I've ran the tests from the branch myself and 125 tests got skipped - is that supposed to be that way?

I've forgot to delete only decorator, now all the tests run

szydlovsky (commented):
All the tests now run, but some of them fail and we even get a runtime crash with one of them: 
image
Do we want to address it now or in a different PR?

Latropos:
@szydlovsky This is an expected error, and we want to check that it is thrown. Unfortunately you can't hide render error dialogs in `__DEV__` mode, you can only dismis all the yellow warnings.

szydlovsky:
@Latropos Alright, if all this is expected I don't see any more issues :shipit:

szydlovsky (approved):

app/src/examples/RuntimeTests/tests/layoutAnimations/entering/predefinedEntering.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+      <Animated.View entering={enteringAnimation} style={styles.animatedBox} />
+    </View>
+  );
+};
+
+async function getSnapshotUpdates(entering: any, waitTime: number, duration: number | undefined, springify = false) {
+  await mockAnimationTimer();
+
+  const updatesContainer = await recordAnimationUpdates();
+  const componentEntering = springify ? entering : entering.springify();
```
piaskowyk, line 101: Why do we test only `springify` here, what about others modifiers?

Latropos: I want to do it in separate PR in separate file. 👍

We have about 40 entering animations, and 14 modifiers that can be composed in many ways. I was thinking about the best way to test them in a way that doesn't take too long to execute and I've chosen the following:
* Create test "test **predefined** entering animation" testing ALL the predefined animations with **basic** modifiers (maybe delay should also be considered basic)
* Create separated tests for various modifier compositions (Which I would name "test entering animation **modifiers**")

The reason I thought to split it into two separate test files is that the first suite should be animation-oriented, and the second should be modifier-oriented.

I was also thinking about creating all the possible test  cases and running a random subset of them, but I've rejected this idea.

piaskowyk (approved):

PR #5996, opened 2024-05-09
Fix CI issues with yarn

tjzel:

### DIFF
--- a/.github/workflows/build-monorepo-action.yml
+++ b/.github/workflows/build-monorepo-action.yml
@@ -1,6 +1,6 @@
 name: Test build in monorepo
 env:
-  YARN_ENABLE_HARDENED_MODE: 0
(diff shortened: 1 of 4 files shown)


tjzel:
I fixes issues with regard to RN template using Yarn 3.6.4 by default.

@tomekzaw @piaskowyk Could you take a look at the failing V8 and monorepo CIs? I have the following questions.

1. Does V8 CI fail because it doesn't yet support RN 0.74? Maybe you can see it on the spot, I haven't investigated it.

2. What's exactly the purpose of the monorepo CI? It creates the following file structure:

```tree
.
└── monorepo
    ├── RootApp
    └── packages
        └── PackageApp
```

Which doesn't make a lot of sense - creating the `RootApp` step. Could you explain a bit your intentions behind this?

piaskowyk (approved):

PR #5997, opened 2024-05-09
fix: use proper classes for bridgeless

WoLewicki:
## Summary

Thanks to @lukmccall we discovered that on bridgeless mode invoking `getReactInstanceManager` creates another instance of application. This PR makes it behave correctly there. 

## Test plan

Run application and see that without this code, on bridgeless, there is no additional button in dev menu since it is created and the button is added in another instance.


tomekzaw (commented):
Can we do something to avoid duplicating whole `NativeProxyCommon.java`?

WoLewicki:
I can move this method only to a new file and make sourceSets from it. Do you want it?

tomekzaw:
Yes, please

WoLewicki:
Done

tomekzaw (approved):
better

PR #6014, opened 2024-05-14
Fix slow animation on Android

piaskowyk:
## Summary

This PR fixes the issue of slow animations on Android. We utilize two different sources of timestamps on Android, so we need to slow down both of them instead of just one to fix the problem. Previously, almost all animations were canceled immediately after starting because the current frame timestamp was 10 times higher than the slowed-down start timestamp.

| before | after |
| --- | --- |
| <video src="https://github.com/software-mansion/react-native-reanimated/assets/36106620/0221e94e-b1b7-48d2-9092-85265dbb36b0
" /> | <video src="https://github.com/software-mansion/react-native-reanimated/assets/36106620/5dcd926e-8c94-44e5-a5a5-7b8e60ebf333" /> |

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/NodesManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/NodesManager.java
@@ -3,6 +3,7 @@ package com.swmansion.reanimated;
 import static java.lang.Float.NaN;
 
 import android.graphics.drawable.Drawable;
+import android.os.SystemClock;
 import android.view.View;
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.GuardedRunnable;
@@ -49,6 +50,9 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import javax.annotation.Nullable;
 
 public class NodesManager implements EventDispatcherListener {
+
+  private Long mFirstUptime = SystemClock.uptimeMillis();
+  private boolean mSlowAnimationsEnabled = false;
 
   public void scrollTo(int viewTag, double x, double y, boolean animated) {
     View view;
@@ -268,6 +272,11 @@ public class NodesManager implements EventDispatcherListener {
     // Systrace.beginSection(Systrace.TRACE_TAG_REACT_JAVA_BRIDGE, "onAnimationFrame");
 
     double currentFrameTimeMs = frameTimeNanos / 1000000.;
+    if (mSlowAnimationsEnabled) {
+      final long ANIMATIONS_DRAG_FACTOR = 10;
+      currentFrameTimeMs =
+          mFirstUptime + (currentFrameTimeMs - mFirstUptime) / ANIMATIONS_DRAG_FACTOR;
+    }
 
     if (currentFrameTimeMs > lastFrameTimeMs) {
       // It is possible for ChoreographerCallback to be executed twice within the same frame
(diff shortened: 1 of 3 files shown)


efstathiosntonas:
@piaskowyk I believe PR title is confusing, I read it like "improve the performance of animations" 😅

piaskowyk:
Hey, this PR is about our feature that slow down all animation for debugging purposes 😅
https://github.com/software-mansion/react-native-reanimated/assets/36106620/08bcfae4-cb24-4fe6-b2dc-22fb4944b50c

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -50,6 +51,10 @@
 
 public class NodesManager implements EventDispatcherListener {
 
+  private Long mFirstUptime = SystemClock.uptimeMillis();
+  private boolean mSlowAnimationsEnabled = false;
+  private long mAnimationsDragFactor;
```
bartlomiejbloniarz, line 56: Why do we have this as a variable if we only ever pass the hard-coded value from `NativeProxyCommon.java` here

bartlomiejbloniarz (approved):

packages/react-native-reanimated/cspell.json
tjzel, line 1: This file is starting to make me angry...

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -268,6 +273,10 @@ private void onAnimationFrame(long frameTimeNanos) {
     // Systrace.beginSection(Systrace.TRACE_TAG_REACT_JAVA_BRIDGE, "onAnimationFrame");
 
     double currentFrameTimeMs = frameTimeNanos / 1000000.;
+    if (mSlowAnimationsEnabled) {
+      currentFrameTimeMs =
+          mFirstUptime + (currentFrameTimeMs - mFirstUptime) / mAnimationsDragFactor;
+    }
```
tjzel, line 279: Smart 🧠

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -523,4 +532,12 @@ private static void addProp(WritableMap propMap, String key, Object value) {
...
     }
   }
+
+  public void setSlowAnimations(boolean slowAnimationsEnabled, long animationsDragFactor) {
+    mSlowAnimationsEnabled = slowAnimationsEnabled;
+    mAnimationsDragFactor = animationsDragFactor;
+    if (slowAnimationsEnabled) {
+      mFirstUptime = SystemClock.uptimeMillis();
+    }
+  }
```
tjzel, line 542:
```suggestion
  public void enableSlowAnimations(boolean slowAnimationsEnabled, long animationsDragFactor) {
    mSlowAnimationsEnabled = slowAnimationsEnabled;
    mAnimationsDragFactor = animationsDragFactor;
    if (slowAnimationsEnabled) {
      mFirstUptime = SystemClock.uptimeMillis();
    }
  }
```

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NodesManager.java
tjzel: +1 on this.

tjzel (approved):
The animations are so slow now they made the 'after' video 3x longer!

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NodesManager.java
piaskowyk: I made this change because `ANIMATIONS_DRAG_FACTOR` is also utilized in the `NativeProxyCommon::getAnimationTimestamp()` method. I wanted to define it in only one place and pass it as an argument here.

PR #6016, opened 2024-05-14
Tests: cancelAnimation, useFrameCallback, measure, withDecay

Latropos:
## Summary
**New tests take approx 20 seconds to run ⏰**

Add tests for the following functionalities:
* cancelAnimation
* useFrameCallback
* measure
* withDecay

## Test plan

https://github.com/software-mansion/react-native-reanimated/assets/56199675/2e5285a2-c203-48fa-adbd-2d528932a68a


app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
```diff
@@ -45,6 +45,26 @@ export class Matchers {
...
+  ) => {
+    const currentValueAsNumber = Number(Number(currentValue));
+    const validInputTypes = typeof minimumValue === 'number' && typeof maximumValue === 'number';
+    const isWithinRange = Number(minimumValue) <= currentValueAsNumber && currentValueAsNumber <= Number(maximumValue);
+
+    const coloredExpected = color(`[${minimumValue}, ${maximumValue}]`, 'green');
+    const coloredReceived = color(currentValue, 'red');
+
+    return {
+      pass: isWithinRange && validInputTypes,
```
szydlovsky, line 61: So if someone creates a test with wrong input types, they will see it as not passing test. Hmmm.. wouldn't it be better to just force all `currentValue`, `minimumValue` and `maximumValue` arguments to be `number` types? Then it forces test suite users to use correct types because the API says so.

szydlovsky (changes requested):

app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/Matchers.ts
Latropos: Sure! I've fixed all the types

app/src/examples/RuntimeTests/RuntimeTestsExample.tsx
```diff
@@ -2,22 +2,22 @@ import React from 'react';
 import RuntimeTestsRunner from './ReanimatedRuntimeTestsRunner/RuntimeTestsRunner';
 
 // load tests
-import './tests/Animations.test';
+// import './tests/Animations.test';
```
szydlovsky, line 5: Do we want it commented out? 😄

app/src/examples/RuntimeTests/tests/animations/withDecay/basic.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    { velocity: 900, deceleration: 0.997 },
+    { velocity: 900, clamp: [0, 150] },
+    { velocity: 900, clamp: [0, 150], rubberBandEffect: true },
+    { velocity: 2000, clamp: [0, 150], rubberBandEffect: true },
+    { velocity: 2000, clamp: [0, 150], rubberBandEffect: true, rubberBandFactor: 2 },
+  ] as Array<WithDecayConfig>)('Config %p', async config => {
+    await render(<DecayComponent config={config} />);
+    await wait(3000);
+  });
+});
```
szydlovsky, line 33: There appears to be no `expect`-like statement, these tests will just run without any info whether they passed or not. Also, `3000ms` wait is quite a lot for each of the tests (maybe let's make it dependant on the config?)

szydlovsky (changes requested):

app/src/examples/RuntimeTests/RuntimeTestsExample.tsx
Latropos: Probably yes, this was just an "example suite", which was expected to show the API. This test file will be removed. Probably we are going to provide some test of testing frameworks instead

app/src/examples/RuntimeTests/tests/animations/withDecay/basic.test.tsx
Latropos: Oh, thank you so much! 😱 I'm always adding snapshots at the very end (because you have to replace them if you modify test case) and forgot.
Yes, `3000ms` is a lot, I'll make it quicker

app/src/examples/RuntimeTests/tests/advancedAPI/useFrameCallback.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+      );
+    };
+    test.each([500, 1000, 2000, 3000])('Run frameCallback for **%p**ms', async timeToStop => {
+      await render(<CancelAfterDelayComponent timeToStop={timeToStop} />);
+      const animatedComponent = getTestComponent(STOP_AFTER_DELAY_REF);
+
+      await wait(timeToStop + 200);
+
+      const expectedWidth = timeToStop / 8;
+      expect(await animatedComponent.getAnimatedStyle('width')).toBeWithinRange(expectedWidth - 3, expectedWidth + 10);
```
szydlovsky, line 51: What leads to such uncertainty in the result?

szydlovsky (approved):
Looks good! There is one thing that looks suspicious to me but other than that a solid test suite!

app/src/examples/RuntimeTests/tests/advancedAPI/useFrameCallback.test.tsx
Latropos: setTimeout is not perfectly precise and useFrameCallback is not synchronised with setTimeout, however the range may still be reduced.

szydlovsky: @Latropos alright, I trust you with this one 🫡

Issue #5892, opened 2024-04-12
method 'android.view.Window android.app.Activity.getWindow()' on a null object reference


TomCorvus:
### Description

Hi 👋

I have an error on some Android devices since I use `useAnimatedKeyboard` hook from `react-native-reanimated`.
This fatal crash seems to not appear each time.

Here the error message:
`Fatal Exception: java.lang.NullPointerException
Attempt to invoke virtual method 'android.view.Window android.app.Activity.getWindow()' on a null object reference`
`com.swmansion.reanimated.keyboard.WindowsInsetsManager.getWindow (WindowsInsetsManager.java:12)`
causes by 
```
  private Window getWindow() {
    return mReactContext.get().getCurrentActivity().getWindow();
  }
```
in
`react-native-reanimated/android/src/main/java/com/swmansion/reanimated/keyboard/WindowsInsetsManager.java` L31

It appears that react context is not init and can't get window width yet.

I have saw in `react-native` core that there is a condition about that to avoid crash (I think)
```
    Context context = mReactApplicationContext.getCurrentActivity();
    if (context != null) {
      return context;
    }
```
`react-native/ReactAndroid/src/main/java/com/facebook/react/animated/ColorAnimatedNode.java` L106

or in react-native-firebase
```
    // protect against NPEs
    if (currentActivity == null) {
      promise.resolve(null);
      return;
    }

```
`@react-native-firebase/perf/android/src/reactnative/java/io/invertase/firebase/perf/ReactNativeFirebasePerfModule.java` L93

If I understand, it needs maybe to get this condition too in `WindowsInsetsManager` class to avoid this crash.

The reproduction links to `react-native-reanimated` example because it is difficult to reproduce in production. This crash seems to be "random" but I tried to give as much detail as possible to fix it.

Thanks

### Steps to reproduce

1. Init fresh project
2. Install react-native-reanimated
3. Use `useAnimatedKeyboard` hook

### Snack or a link to a repository

https://github.com/software-mansion/react-native-reanimated/tree/main/Example

### Reanimated version

3.8.1

### React Native version

0.73.6

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Release app & production bundle

### Device

Real device

### Device model

Galaxy A32 (Android13), Pixel 7 (Android13), Pixel 5 64bit (Android 14)

### Acknowledgements

Yes

[...]

PR #6022, opened 2024-05-15
Fix crash when current activity is null

maciekstosio:
## Summary

Should fix: #5892. We do not check if the current activity is `null` and it seems that it randomly can be null. It seems like ongoing issue on react native side: https://github.com/facebook/react-native/issues/37518, https://github.com/facebook/react-native/issues/18345. The `useAnimatedKeyboard` hook won't work when the app when current activity is `null`, but it should not crash the app anymore. Our assumption is that most of the libs handle it this way (#5892), so it must be some corner case when most functionality won't work - that's why we don't try to recover from this state.

### DIFF
--- a/android/src/main/java/com/swmansion/reanimated/keyboard/WindowsInsetsManager.java
+++ b/android/src/main/java/com/swmansion/reanimated/keyboard/WindowsInsetsManager.java
@@ -1,9 +1,10 @@
 package com.swmansion.reanimated.keyboard;
 
+import android.app.Activity;
 import android.os.Handler;
 import android.os.Looper;
+import android.util.Log;
 import android.view.View;
-import android.view.Window;
 import android.widget.FrameLayout;
 import androidx.core.view.ViewCompat;
 import androidx.core.view.WindowCompat;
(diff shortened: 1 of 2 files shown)


cspell.json
```diff
@@ -36,7 +36,9 @@
     "workletize",
     "workletized",
     "workletizes",
-    "worklets"
+    "worklets",
+    "androidx",
+    "appcompat"
```
maciekstosio, line 41: Local CI was failing on those two.

szydlovsky: @maciekstosio please try keeping the `cspell.json` file in an alphabetical order 😄

maciekstosio: Sure, fixed!

szydlovsky (approved):
Looks good :shipit:

android/src/main/java/com/swmansion/reanimated/keyboard/WindowsInsetsManager.java
```diff
@@ -27,33 +30,55 @@ public WindowsInsetsManager(
...
+    Activity currentActivity = getCurrentActivity();
+    if (currentActivity == null) {
+      Log.e("Reanimated", MissingContextErrorMsg);
+      return;
+    }
+
+    ViewCompat.setOnApplyWindowInsetsListener(
+        currentActivity.getWindow().getDecorView(), this::onApplyWindowInsetsListener);
+    ViewCompat.setWindowInsetsAnimationCallback(
+        currentActivity.getWindow().getDecorView(), keyboardAnimationCallback);
```
piaskowyk, line 51:
```suggestion
    View rootView = currentActivity.getWindow().getDecorView();
    ViewCompat.setOnApplyWindowInsetsListener(rootView, this::onApplyWindowInsetsListener);
    ViewCompat.setWindowInsetsAnimationCallback(rootView, keyboardAnimationCallback);
```

piaskowyk (approved):
Just one small change and we can merge 🫡

PR #6028, opened 2024-05-17
Fix "Not implemented" exception on JSC

### LINKED ISSUE #5974: Exception in Hostfunction: not implemented
### Description

Screenshot 2024-05-05 at 12 00 10 AM

this error produce after installation of the app

### Steps to reproduce

Add library to package.json file
run yarn
pod install
install app

I have problem in my project

### Reanimated version

3.8.1

### React Native version

0.73.0

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

tjzel:
## Summary

In some implementations of JSC, at least `react-native-macos` [here](https://www.npmjs.com/package/react-native-macos/v/0.73.26?activeTab=code) native state methods are not implemented and they are throwing errors. This PR adds logic that will go around this issue, by probing the `hasNativeState` method to determine whether it should use that.

Fixes #5974 

## Test plan

- [x] MacOS example builds now.
- [x] Check if all "not implemented" methods in the link are safely accessed.


tomekzaw (commented):
Why can't we just call `hasNativeState` in `try/catch` block?

tjzel:
@tomekzaw It's for performance. Why would we surround it with the try-catch on every call, since we will know everything after the first call.

tomekzaw:
@tjzel I wouldn't worry much about performance in this case.

SInce we know that some versions react-native-macos don't support `jsi::NativeState` when using JSC, why can't we just use `#ifdef`s instead?

Also, the names `nativeStateAccess`, `AccessProbe`, `Safe` and `Unsafe` are a bit overwhelming, how about simply `supportsNativeState` with `Yes` and `No` instead?

Finally, I don't like the fact that some unrelated files have been formatted, can we move them to a separate PR?

mrousavy:
Sorry about that, I added `jsi::NativeState` here.. I didn't test on MacOS / JSC.

tjzel:
@tomekzaw I read an interesting [article about try/catch blocks performance](https://pspdfkit.com/blog/2020/performance-overhead-of-exceptions-in-cpp/) from which it seems that try blocks are fast but the catch ones are expensive. Since these calls would constantly throw when `nativeState` is not implemented, I'd rather keep the current approach.

Regarding `#ifdef` I don't know how would we do that. I don't think there's any flag that would tell us if `nativeState` is implemented on a given engine.

I'll revert formatting and make the names more approachable.

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -221,11 +229,45 @@ jsi::Value ShareableObject::toJSValue(jsi::Runtime &rt) {
...
+      nativeStateSource.asObject(rt).hasNativeState(rt)) {
+    nativeState_ = nativeStateSource.asObject(rt).getNativeState(rt);
+    isNativeStateImplemented_ = IsNativeStateImplemented::Yes;
+  }
+}
+
+void ShareableObject::runWithNativeStateProbe(
+    std::function<void()> &&block) {
+  try {
+    block();
```
piaskowyk, line 264: Why we need to pass an lambda here instead of just execute `obj.setNativeState(rt, nativeState_);`?

Common/cpp/SharedItems/Shareables.cpp
```diff
@@ -221,11 +229,45 @@ jsi::Value ShareableObject::toJSValue(jsi::Runtime &rt) {
...
+    isNativeStateImplemented_ = IsNativeStateImplemented::Yes;
+  } catch (...) {
```
piaskowyk, line 266: Does this 'Not implemented' exception have a specific type that we can implicitly catch here?

piaskowyk (commented):
Based on the information from here - [PR to React Native core](https://github.com/facebook/react-native/commit/7b7f128968310c466ce0a929e9964b08fe098503), the native state has been available on JSC since version 0.74. Hence, it appears that we can simply create a straightforward compilation flag instead of runtime checks. What do you think?

tjzel:
Well I guess we can do it that way for now, if the problem persists we will resort to runtime checks.

tjzel:
Superseded by #6137

PR #5845, opened 2024-03-27
WorkletEventHandler revamp

szydlovsky:
## Summary
Our `WorkletEventHandler` class (used in `useScrollViewOffset` and `useEvent` hooks as well in internal logic of `AnimatedComponent`) was made for handling only one component at a time. I made it so multiple of them can work at the same time. This way, anything based on `useEvent` (so for example `useAnimatedScrollHandler`) can be used for multiple components without needlessly copying the logic. This fixes https://github.com/software-mansion/react-native-reanimated/issues/5345 and https://github.com/software-mansion/react-native-reanimated/issues/5488 as well.

## Test plan
Effectively, this needs testing on `useAnimatedScrollHandler` and `useScrollViewOffset` as well as generally checking for any regression in AnimatedComponents.

Testing code for `useAnimatedScrollHandler` on multiple components:
<details><summary>Code</summary>

``` TYPESCRIPT
import * as React from 'react';
import { Text, View, Button, StyleSheet } from 'react-native';
import Animated, { useAnimatedScrollHandler } from 'react-native-reanimated';

type ItemValue = { title: string };
const items: ItemValue[] = [...new Array(101)].map((_each, index) => {
  return { title: `${index}` };
});

type ItemProps = { title: string };
const Item = ({ title }: ItemProps) => (
  <View style={styles.item}>
    <Text style={styles.title}>{title}</Text>
  </View>
);

export default function EmptyExample() {
  const [attachBoth, setAttachBoth] = React.useState(false);
  const [onlyOnScroll, setOnlyOnScroll] = React.useState(true);

  // props to pass into children render functions
  const onScroll = useAnimatedScrollHandler(
    onlyOnScroll
      ? {
          onScroll(e) {
            'worklet';
            console.log(e.eventName);
          },
        }
      : {
          onScroll(e) {
            'worklet';
            console.log(e.eventName);
          },
          onBeginDrag(e) {
            'worklet';
            console.log(e.eventName);
          },
          onEndDrag(e) {
            'worklet';
            console.log(e.eventName);
          },
          onMomentumBegin(e) {
            'worklet';
            console.log(e.eventName);
          },
          onMomentumEnd(e) {
            'worklet';
            console.log(e.eventName);
          },
        }
  );

  return (
    <View style={styles.container}>
      <Button
        title={`Swtich to attach ${attachBoth ? 'first' : 'both'}`}
        onPress={() => {
          setAttachBoth(!attachBoth);
        }}
      />
      <Button
        title={`Swtich events to ${
[...]

PR #6030, opened 2024-05-17
Make animated component's event tag properly update

szydlovsky:
## Summary
Turns out (thanks to @j-piasecki for noticing) that animated component's event-emitting viewTag can change through re-rendering. The following fix takes care of that.

## Test plan

You can go through all `ComposedEventHandler` examples form Example app, as well as code snippets in https://github.com/software-mansion/react-native-reanimated/pull/5845#issue-2211354383.

Also, this should work perfectly regardless of re-renders:

``` TYPESCRIPT
import { GestureDetector, Gesture } from 'react-native-gesture-handler';
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  withSpring,

function Ball(props) {
  const isPressed = useSharedValue(false);
  const offset = useSharedValue({ x: 0, y: 0 });

    return {
      transform: [
        { translateX: offset.value.x },
        { translateY: offset.value.y },
        { scale: withSpring(isPressed.value ? 1.2 : 1) },
      ],
      backgroundColor: isPressed.value ? 'yellow' : 'blue',
    };
  });

  const gesture = Gesture.Pan()
    .onBegin(() => {
      'worklet';
      isPressed.value = true;
    })
    .onChange((e) => {
      'worklet';
      offset.value = {
        x: e.changeX + offset.value.x,
        y: e.changeY + offset.value.y,
      };
    })
    .onFinalize(() => {
      'worklet';
      isPressed.value = false;
    });

  return (
    <GestureDetector gesture={gesture}>
      <Animated.View style={[styles.ball, animatedStyles]} key={props.counter} />
    </GestureDetector>
  );
}

  const [counter, setCounter] = useState(0);

  return (
      <Button title="Remount" onPress={() => setCounter((prev) => prev + 1)} />
      <Ball counter={counter} />
    </View>
  );
}

  container: {
    flex: 1,
  },
  ball: {
    width: 100,
    height: 100,
    borderRadius: 100,
    backgroundColor: 'blue',
    alignSelf: 'center',
  },
});
```
</details>


src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -291,6 +293,29 @@ export function createAnimatedComponent(
...
+          const prevProp = prevProps[key];
+          if (
+            has('workletEventHandler', prevProp) &&
+            prevProp.workletEventHandler instanceof WorkletEventHandler
+          ) {
+            prevProp.workletEventHandler.unregisterFromEvents(
+              this._eventViewTag
+            );
+          }
+        }
```
piaskowyk, line 310: It seems like a code duplication with `_detachNativeEvents`

src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -291,6 +293,29 @@ export function createAnimatedComponent(
     _updateNativeEvents(
       prevProps: AnimatedComponentProps<InitialComponentProps>
     ) {
+      // If the event view tag changes, we need to completely re-mount all events
+      const computedEventTag = this._getEventViewTag();
```
piaskowyk, line 297: `computedEventTag` or `componentEventTag`? 🤔

src/createAnimatedComponent/createAnimatedComponent.tsx
szydlovsky: @piaskowyk `_detachNativeEvents` works on current props. This code needs to remove previous ones

tjzel: +1

src/createAnimatedComponent/createAnimatedComponent.tsx
szydlovsky: @piaskowyk computed - it is being calculated on the go and used for some calculations, but not yet applied

src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+    // If the event view tag changes, we need to completely re-mount all events
+    if (this._managedComponent._eventViewTag !== computedEventTag) {
+      // Remove all bindings from previous props that ran on the old viewTag
+      executeForEachEventHandler(prevProps, (_key, handler) => {
+        handler.unregisterFromEvents(this._managedComponent._eventViewTag);
+      });
+      // We don't need to unregister from current (new) props, because their events weren't registered yet
+      // Replace the view tag
+      this._managedComponent._eventViewTag = computedEventTag;
```
tjzel, line 88: I don't like the fact that this is where the `eventViewTag` is set but it's calculated in `createAnimatedComponent`. It's looks like it could be a confusing side effect in the future. Let's move all the responsibility for handling event tags to this manager. We should have a method `updateEventTag` that would be called inside `updateNativeEvents`.

tjzel (changes requested):
Nice work, you got the idea, but now we need some polishing to ensure proper maintainability in the future.

src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -0,0 +1,118 @@
...
+  AnimatedComponentProps,
+  InitialComponentProps,
+} from './commonTypes';
+import { has } from './utils';
+import { WorkletEventHandler } from '../reanimated2/WorkletEventHandler';
+import { isWeb } from '../reanimated2/PlatformChecker';
+
+const IS_WEB = isWeb();
+
+type ManagedAnimatedComponent = React.Component<
```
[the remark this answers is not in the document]
szydlovsky: Yes it is, for example in `JSPropsUpdater` and it is done the same exact way:
```typescript
public addOnJSPropsChangeListener(
    animatedComponent: React.Component<
      AnimatedComponentProps<InitialComponentProps>
    > &
      IAnimatedComponentInternal
  ) {
   ...
  ```

File: src/createAnimatedComponent/createAnimatedComponent.tsx [hunk 5]
```diff
@@ -141,12 +143,13 @@ export function createAnimatedComponent(
...
       }
+      this._NativeEventsManager = new NativeEventsManager(this);
     }
 
     componentDidMount() {
-      this._setComponentViewTag();
-      this._setEventViewTag();
-      this._attachNativeEvents();
+      this._componentViewTag = this._getComponentViewTag();
+      this._eventViewTag = this._getEventViewTag();
```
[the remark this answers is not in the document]
szydlovsky: We could do that with `eventViewTag`, but `componentViewTag` is responsible for LA/SET and animated styles, it isn't even the scope of native events

src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -0,0 +1,144 @@
...
+    handler: InstanceType<typeof WorkletEventHandler>
+  ) => void
+) {
+  for (const key in props) {
+    const prop = props[key];
+    if (isWorkletEventHandler(prop)) {
+      callback(key, prop.workletEventHandler);
+    }
+  }
+}
```
tjzel, line 125: Why won't we make it a private method of the class?

src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -0,0 +1,144 @@
...
+    const scrollableNode = componentAnimatedRef.getScrollableNode();
+    newTag = findNodeHandle(scrollableNode) ?? -1;
+  } else {
+    newTag =
+      findNodeHandle(
+        options?.setNativeProps ? component : componentAnimatedRef
+      ) ?? -1;
+  }
+  return newTag;
+}
```
tjzel, line 144: Same here, instead of passing the `component` we could make it a private method.

tjzel (changes requested):
We're almost there!

src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -0,0 +1,144 @@
...
+  constructor(component: ManagedAnimatedComponent, options?: ComponentOptions) {
+    this._managedComponent = component;
+    this._componentOptions = options;
+    this._eventViewTag = getEventViewTag(
+      this._managedComponent,
+      this._componentOptions
+    );
+  }
+
+  public attachNativeEvents() {
```
piaskowyk, line 27:
```suggestion
  public attachEvents() {
```
The class name already includes the `NativeEvents` prefix, so there's no need to repeat this information in the method name. You can apply this change also to `detachNativeEvents` and `updateNativeEvents` methods.

tjzel: smart

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -0,0 +1,144 @@
...
+  AnimatedComponentRef,
+} from './commonTypes';
+import { has } from './utils';
+import { WorkletEventHandler } from '../WorkletEventHandler';
+import { findNodeHandle } from 'react-native';
+
+export class NativeEventsManager implements INativeEventsManager {
+  private _managedComponent: ManagedAnimatedComponent;
+  private _componentOptions?: ComponentOptions;
+  private _eventViewTag = -1;
```
tjzel, line 16: Me thinks

```suggestion
  private #managedComponent: ManagedAnimatedComponent;
  private #componentOptions?: ComponentOptions;
  private #eventViewTag = -1;
```

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -0,0 +1,144 @@
...
+      (key, handler) => {
+        if (!prevProps[key]) {
+          // Prop got added
+          handler.registerForEvents(this._eventViewTag);
+        }
+      }
+    );
+  }
+
+  private isWorkletEventHandler(
```
tjzel, line 86: This should be a helper function instead of a method IMO.

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -0,0 +1,144 @@
...
+      handler: InstanceType<typeof WorkletEventHandler>
+    ) => void
+  ) {
+    for (const key in props) {
+      const prop = props[key];
+      if (this.isWorkletEventHandler(prop)) {
+        callback(key, prop.workletEventHandler);
+      }
+    }
+  }
```
tjzel, line 108: This seems nice to avoid code duplication, but I find the manager itself hard to read with it. I'm considering if we should move to inline `Object.entries()` instead.

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -0,0 +1,144 @@
...
+    } else {
+      newTag =
+        findNodeHandle(
+          this._componentOptions?.setNativeProps
+            ? this._managedComponent
+            : componentAnimatedRef
+        ) ?? -1;
+    }
+    return newTag;
+  }
```
tjzel, line 127: Why does it have to be so ugly 😭

tjzel (changes requested):
😎

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
szydlovsky: It was moved as is from animated component file and unfortunately precisely this works in all use cases we have so far...

tjzel: Yes I know :c

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
szydlovsky: There can't be both at once, `#` symbol is sufficient

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -11,30 +11,27 @@ import { WorkletEventHandler } from '../WorkletEventHandler';
 import { findNodeHandle } from 'react-native';
 
 export class NativeEventsManager implements INativeEventsManager {
-  private _managedComponent: ManagedAnimatedComponent;
-  private _componentOptions?: ComponentOptions;
-  private _eventViewTag = -1;
+  #managedComponent: ManagedAnimatedComponent;
+  #componentOptions?: ComponentOptions;
```
tjzel, line 15: Will it work?
```suggestion
  readonly #managedComponent: ManagedAnimatedComponent;
  readonly #componentOptions?: ComponentOptions;
```

tjzel (approved):
Just check one more thing for me and we gucci. Just make sure to test it thoroughly!

piaskowyk (approved):
Well done 🫡

PR #6034, opened 2024-05-20
Add missing unit tests (bezier, color converter, interpolate)

Latropos:
## Summary
Add unit test we have been missing in our repo.
Tests of normalizeColor were moved together with other test files and expanded

## Test plan

<details><summary> Bezier test </summary>

image
</details>

<details><summary> Colors test</summary>

<table>
<tr>
<td>

image
</td>
<td>

image
</td>
</tr>
</table>
</details>

<details><summary> Interpolate test </summary>

image
</details>


__tests__/bezier.test.ts
```diff
@@ -0,0 +1,237 @@
...
+      ).toBeTruthy();
+    });
+
+    const MONKEY_TRIES = 20000;
+    const PRECISION_0 = 0.01;
+    let allTestPass0 = true;
+    repeat(MONKEY_TRIES)(() => {
+      const a = Math.random();
+      const b = Math.random();
+      const c = Math.random();
```
maciekstosio, line 194: Is it something that we agreed that we do in tests? Random may cause tests to be flaky, imo it should be better to have hardcoded few corner cases than random.

Latropos: > Is it something that we agreed that we do in tests?

Yes, @piaskowyk and I were talking about it and decided that it is OK as long as we have hardcoded corner cases too. It is a controversial testing technique, but its not uncommon. These particular tests are verifying bezier function, which has implementation copy-pasted from a linked external library. The tests are based on tests common in that library.

>  Random may cause tests to be flaky, imo it would be better to have hardcoded few corner cases than random.

You may notice, that there is quite a lot of repeat in random tests, each is run 1000 times, this was to reduce the risk of flakiness - when you have 1000 tries it is much easier to catch some error if it occur.

Few corner cases can't be as reliable as 1000 cases. We could hardcode 1000 random values, but this would be much more complicated.

maciekstosio: >  We could hardcode 1000 random values, but this would be more complicated.
That's definitely something we wouldn't like to do :D The only thing that could be added, or maybe is added (?), is constant seed then random values will be "hardcoded". But if this is something we agreed and 
>as long as we have hardcoded corner cases too 
then it looks ok to me 🎉

maciekstosio (approved):

PR #6049, opened 2024-05-27
Change passing of example components

patrycjakalinska:
Because of new InteractiveExample structure for `Examples` section to work properly we needed to change passed example components from JSX.Elements to functional components. Also refreshing Examples that were not based on width of the parent container was fixed.

### DIFF
--- a/docs/src/components/InteractiveExample/index.tsx
+++ b/docs/src/components/InteractiveExample/index.tsx
@@ -93,7 +93,6 @@ export default function InteractiveExample({
   const [tab, setTab] = React.useState<Tab>(Tab.PREVIEW);
   const [jsxCode, setJsxCode] = React.useState(() => compileTSXtoJSX(tsxCode));
   const [width, setWidth] = React.useState<number | null>(null);
-  const [showPreview, setShowPreview] = React.useState(!showCode);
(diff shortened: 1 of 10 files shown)


docs/src/components/InteractiveExample/index.tsx
```diff
@@ -93,7 +93,6 @@ export default function InteractiveExample({
   const [tab, setTab] = React.useState<Tab>(Tab.PREVIEW);
   const [jsxCode, setJsxCode] = React.useState(() => compileTSXtoJSX(tsxCode));
   const [width, setWidth] = React.useState<number | null>(null);
-  const [showPreview, setShowPreview] = React.useState(!showCode);
```
m-bert, line 96: `showCode` is now marked as unused, are you sure it should be removed in this PR?

m-bert (commented):
Looks good, please merge current `main` into this branch

docs/src/components/InteractiveExample/index.tsx
patrycjakalinska: yes, because apparently showPreview was unused as well

m-bert (approved):

PR #6057, opened 2024-05-28
Add safer refresh on InteractiveExample

patrycjakalinska:
Before when page was loading, there was a chance that `width` value wouldn't be passed to component in `InteractiveComponent` - in other words the `width` will be passed to component before it had become something other than `null`, and there won't be a rerender after width gets a number value. After this change, with `useEffect` hook, we monitor if width has changed and if so, we trigger a rerender that guarantees our component will always get correct width.


kacperkapusciak (approved):
looks good!

sidenote: Shouldn't the InteractiveExample be a part of T-rex as well? We've started to use it in RNGH as well. Just asking

PR #6058, opened 2024-05-28
Fixes of dark-theme and values in docs

patrycjakalinska:
### Add offset 'margins' so withDecay example would look cleaner

Before:

https://github.com/software-mansion/react-native-reanimated/assets/59940332/752191ee-8202-42f0-be3e-1255533417ba

After:

https://github.com/software-mansion/react-native-reanimated/assets/59940332/e91d5701-efeb-4255-a350-c516381b27b9

### Add dark theme to __Section List__ and __Bottom Sheet__

Before:

image
image

After:

image
image

### Fix `--swm-dropdown-versions-item` color on light-theme

Before:

image

After:

image

### Remove different background on Examples' sidebar

Before: 

image

After: 

image

### DIFF
--- a/packages/docs-reanimated/src/css/colors.css
+++ b/packages/docs-reanimated/src/css/colors.css
@@ -258,7 +258,7 @@
 
   /* Versions dropdown */
   --swm-dropdown-versions-background: var(--swm-off-white);
-  --swm-dropdown-versions-item: var(--swm-off-white);
+  --swm-dropdown-versions-item: var(--swm-navy-light-100);
   --swm-dropdown-versions-item-border: var(--swm-purple-light-40);
   --swm-dropdown-versions-item-background: var(--swm-purple-light-20);
 
--- a/packages/docs-reanimated/src/css/overrides.css
+++ b/packages/docs-reanimated/src/css/overrides.css
@@ -115,6 +115,8 @@ button[class*='DocSearch-Button'] {
   margin: 0 !important;
 }
 
+/* versions dropdown on landing */
+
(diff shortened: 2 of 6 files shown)


docs/static/examples/BottomSheet.js
```diff
@@ -71,12 +73,18 @@ const sheetStyles = StyleSheet.create({
...
   const isOpen = useSharedValue(false);
 
   const toggleSheet = () => {
     isOpen.value = !isOpen.value;
   };
 
+  const contentStyle = useAnimatedStyle(() => ({
+    color: colorScheme === 'light' ? '#001a72' : '#f8f9ff',
+    textDecorationColor: colorScheme === 'light' ? '#001a72' : '#f8f9ff',
+  }));
```
piaskowyk, line 86: I don't see any shared value inside `useAnimatedStyle`. Perhaps we can simply use a plain JavaScript object 🤔 What are your thoughts on that?

patrycjakalinska: It's actually good idea (like proper way to do this), i was thinking about it, but was not sure - thanks!

piaskowyk (approved):

kacperkapusciak (approved):
I found one more visual glitch where material ui inputs are always in dark mode but it can be fixed in another PR

image
image

kacperkapusciak:
don't merge it yet please, let guys release 3.12.0 from main

kacperkapusciak:
Okay, let's resolve the conflicts and I think it's ready to merge

PR #6064, opened 2024-05-29
[Web LA] Add warning about `transform`

m-bert:
## Summary

In #6060 I've removed adding already existing transform into layout animations on web (you can read more about reasoning in the mentioned PR). This PR adds `warning` for users that suggests creating animated wrapper if `Animated.View` with layout animation contains `transform` in `style`.

## Test plan

Tested on example app (`MountingUnmounting` example) and on the following code:

<details>
<summary> Test code </summary>

```jsx
import React from 'react';
import Animated, { LightSpeedInLeft } from 'react-native-reanimated';

  return (
      <Animated.View
        entering={LightSpeedInLeft.duration(2500)}
        style={{
          width: 100,
          height: 100,
          backgroundColor: 'crimson',
          transform: [{ skewX: '25deg' }],
        }}
      />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

</details>


src/reanimated2/layoutReanimation/web/animationsManager.ts
```diff
@@ -115,6 +115,12 @@ export function startWebLayoutAnimation<
   animationType: LayoutAnimationType,
   transitionData?: TransitionData
 ) {
+  if (element.style.transform !== '') {
+    console.warn(
+      `[Reanimated] Property "transform" will not be included in layout animation. Please wrap your component with an animated view and apply the layout animation on the wrapper.`
```
szydlovsky, line 120: I am not sure whether every user will understand that `apply the layout animation on the wrapper` means `move the layout animation to the wrapper`. I'd make it a little bit more straightforward stating that it has to be moved there.

szydlovsky (approved):
Looks good to me but let's see what @piaskowyk has to say 👀

packages/react-native-reanimated/src/layoutReanimation/web/animationsManager.ts
```diff
@@ -50,6 +50,35 @@ function checkUndefinedAnimationFail(
...
+  keyframe: string,
+  styles: CSSStyleDeclaration
+) {
+  const propertyRegex = /([a-zA-Z-]+)(?=:)/g;
+  const animationProperties = new Set();
+
+  let match;
+  while ((match = propertyRegex.exec(keyframe)) !== null) {
+    animationProperties.add(match[1]);
+  }
```
piaskowyk, line 63: You can simply use matchAll instead of looping over all match:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll

```
const matches = str.matchAll(regexp);
```

piaskowyk (approved):

packages/react-native-reanimated/src/layoutReanimation/web/animationsManager.ts
m-bert: Done in [c144817](https://github.com/software-mansion/react-native-reanimated/pull/6064/commits/c144817fd72be2064d2df9d5c66cd5663776cc32).

PR #6079, opened 2024-06-05
Add Floating Action Button example to Reanimated Cookbook

patrycjakalinska:
We added FAB example to `Reanimated Cookbook`.

Result:

https://github.com/software-mansion/react-native-reanimated/assets/59940332/83dfa958-7b27-4980-9ecf-6f3823112557


docs/static/examples/FABButton.js
```diff
@@ -0,0 +1,157 @@
...
+      //highlight-next-line
+      const delay = index * 100;
+      return {
+        transform: [
+          { translateY: translateValue },
+          {
+            scale: withDelay(
+              delay,
+              //highlight-next-line
+              withTiming(interpolate(moveValue, [OFFSET * index, 0], [1, 0]))
```
piaskowyk, line 53: I'm not sure if 'interpolate' is necessary here because the value of interpolation function will evaluate only one during the animation start 🤔 Correct me if I'm wrong.

packages/docs-reanimated/static/examples/FABButton.js
```diff
@@ -0,0 +1,156 @@
...
+      return {
+        transform: [
+          { translateY: translateValue },
+          {
+            scale: withDelay(delay, withTiming(scaleValue)),
+          },
+        ],
+      };
+    });
+
```
kacperkapusciak, line 58: I'm not a fan of generating animated styles as it breaks my "rules of hooks" mental model.

I believe it would be cleaner if we had this animated style defined once inside the FloatingActionButton component and pass only the required values (for example `isExpanded` shared value) as props.

packages/docs-reanimated/static/examples/FABButton.js
```diff
@@ -0,0 +1,156 @@
...
+const SPRING_CONFIG = {
+  duration: 1200,
+  overshootClamping: true,
+  dampingRatio: 0.8,
+};
+
+const OFFSET = 60;
+
+const FABButton = ({ animatedStyles, buttonLetter }) => (
```
kacperkapusciak, line 22:
```suggestion
const FloatingActionButton = ({ animatedStyles, buttonLetter }) => (
```
Let's use the full name as `FABButton` expands to `Floating Action Button Button` 😄

kacperkapusciak (changes requested):

packages/docs-reanimated/static/examples/FABButton.js
patrycjakalinska: 😭 😆

kacperkapusciak (approved):
💯

packages/docs-reanimated/static/examples/FloatingActionButton.js
```diff
@@ -0,0 +1,160 @@
...
+const OFFSET = 60;
+
+const FloatingActionButton = ({ isExpanded, index, buttonLetter }) => {
+  const animatedStyles = useAnimatedStyle(() => {
+    const moveValue = interpolate(
+      Number(isExpanded.value),
+      [0, 1],
+      //highlight-next-line
+      [0, OFFSET * index]
+    );
```
piaskowyk, line 29: You can simply replace it with this:
```suggestion
const moveValue = isExpanded.value ? OFFSET * index : 0;
```

https://github.com/software-mansion/react-native-reanimated/assets/36106620/fcd7e4d1-741e-4ae8-b517-927f3afb66f2

piaskowyk (approved):
I've tested (only) the `FloatingActionButton` code in the Expo snack and it works for me 👍

PR #6089, opened 2024-06-07
docs(example): update readme with build step

frankcalise:
## Summary

I had trouble running the Example app after following the README steps. After building from the root of the monorepo, the iOS app began working so I updated the documentation steps.

## Test plan

1. Clone the `react-native-reanimated` repository
2. Follow step exactly in `Example/README.md` and observe the iOS app will not start
3. Go back to the root of the monorepo and run `yarn build`
4. `cd Example && yarn ios` should now work successfully


SolankiYogesh (approved):

tomekzaw (approved):
Thanks for the PR!

PR #6103, opened 2024-06-11
Fix eslint rules and types of tests

Latropos:
## Summary
Since we have introduced **monorepo** code of example application was provided with more restrictive eslint rules, so I have to either fix runtime tests or disable some of the rules. I do both depending on the case.

* Override eslint rule `no-template-curly-in-string`, because of test-each syntax uses string slooking like string template
* Fix all the type errors, refactor `RuntimeTestsApi.ts`, to rewrite types of test & describe
* Fix imports and ignore most of other rules
 
## Test plan
Tested on IOS


apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsApi.ts
```diff
@@ -209,5 +185,5 @@ export async function recordAnimationUpdates() {
 }
 
 export async function stopRecordingAnimationUpdates() {
-  testRunner.stopRecordingAnimationUpdates();
+  await testRunner.stopRecordingAnimationUpdates();
```
tjzel, line 188: Was it caught by eslint? Is it how it was intended to work?

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
```diff
@@ -127,11 +128,12 @@ export class TestRunner {
       buildSuite,
       testCases: [],
       nestingLevel: (this._currentTestSuite?.nestingLevel || 0) + 1,
+      // eslint-disable-next-line no-unneeded-ternary
```
tjzel, line 131: Why do we disable that? I assume eslint wants us to use nullish coalescent operator here?

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
```diff
@@ -323,6 +328,7 @@ export class TestRunner {
       console.error = newConsoleFuncJS;
       console.warn = newConsoleFuncJS;
 
+      // eslint-disable-next-line @typescript-eslint/unbound-method
```
tjzel, line 331: Why is this method unbound?

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/types.ts
```diff
@@ -25,22 +25,22 @@ export type SharedValueSnapshot = {
 export type ComponentRef = MutableRefObject<(Component & { props: { style: Record<string, unknown> } }) | null>;
 
 export enum DescribeDecorator {
-  ONLY = 'ONLY',
-  SKIP = 'SKIP',
-  NONE = 'NONE',
+  ONLY = 'only',
+  SKIP = 'skip',
```
tjzel, line 29: Why did you remove `NONE`? How is it related to monorepo & eslint?

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/types.ts
```diff
@@ -104,6 +104,7 @@ export type TestConfiguration = {
   render: Dispatch<SetStateAction<ReactNode | null>>;
 };
 
+/* eslint-disable no-var */
```
tjzel, line 107: This is ok, but I wonder if with newer TypeScript there's actually a nicer way to do it.

apps/common-app/src/examples/RuntimeTests/tests/animations/withTiming/arrays.test.tsx
```diff
@@ -102,6 +101,7 @@ describe('withTiming animation of ARRAY', () => {
       );
       const components = Object.values(COMPONENT_REF).map(refName => getTestComponent(refName));
 
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
tjzel, line 104: Why disable that? 👀

apps/common-app/src/examples/RuntimeTests/tests/animations/withTiming/objects.test.tsx
```diff
@@ -85,6 +79,7 @@ describe('withTiming animation of WIDTH', () => {
       await render(<AnimatedComponent startStyle={startStyle} finalStyle={finalStyle} />);
       const component = getTestComponent(COMPONENT_REF);
       await wait(1000);
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
tjzel, line 82: Why disable that? 👁️ 👃 👁️

apps/common-app/src/examples/RuntimeTests/tests/core/useSharedValue.test.tsx
```diff
@@ -136,7 +136,7 @@ describe('Tests of *****sharedValue*****', () => {
         await render(<ProgressBar progress={index / Presets.numbers.length} />);
       }
     });
-    test.each([BigInt(2), BigInt(-2), BigInt(123456789), BigInt(123456789123456789123456789)])(
+    test.each([BigInt(2), BigInt(-2), BigInt(123456789), BigInt(1234567891234567)])(
```
tjzel, line 139: ??

piaskowyk:
The name of this PR is a bit misleading because it does more than just fixing monorepo compatibility

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
Latropos: I disable it because this was already handled in https://github.com/software-mansion/react-native-reanimated/pull/6047

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsApi.ts
Latropos: Yes, there should be `await` here, and preciously eslint didn't catch it.

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsApi.ts
```diff
@@ -158,8 +133,9 @@ export async function wait(delay: number) {
   return testRunner.wait(delay);
 }
 
+// eslint-disable-next-line @typescript-eslint/unbound-method
```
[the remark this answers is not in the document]
Latropos: Because we copy class method, so the state is not preserved. We can 

- disable eslint
- change notify function into arrow function
- change notify into a static function and provide option to ignore static unboud functions https://typescript-eslint.io/rules/unbound-method/#ignorestatic in eslint config

Personally I suggest using ignorestatic, because static functions don't rely on class scope and should be safe to use.

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/types.ts
Latropos: Monorepo required me to use correct types. I wanted to use type `Record<DescribeDecorator, ...>`, so I removed `NONE` which is never used (and now I think I shouldn't have added it to this enum).
Also thats why I changed the values of enum keys - they should match object properties.
Of course I could define more new types without modifying existing ones, but this solution seems the nicest.

apps/common-app/src/examples/RuntimeTests/tests/animations/withTiming/arrays.test.tsx
Latropos: Because syntax
```tsx
array.foreach(async () => {})
```
is not allowed. I'll replace it with `for (const X of array)` loop.

apps/common-app/src/examples/RuntimeTests/tests/core/useSharedValue.test.tsx
Latropos: image

apps/common-app/src/examples/RuntimeTests/tests/animations/withTiming/arrays.test.tsx
```diff
@@ -102,17 +101,19 @@ describe('withTiming animation of ARRAY', () => {
...
       });
       await wait(speed + 200);
-
-      components.forEach(async (component, index) => {
+      let index = 0;
+      for (const component of components) {
         expect(await component.getAnimatedStyle('width')).toBe(finalWidths[index], ComparisonMode.DISTANCE);
-      });
+        index += 1;
+      }
```
piaskowyk, line 116: Why we need this change?

piaskowyk (approved):
ok for me 👍

apps/common-app/src/examples/RuntimeTests/tests/animations/withTiming/arrays.test.tsx
Latropos: The syntax `array.foreach(async () => {})` is not allowed, one of eslint rules ban it.
After reading [this page](https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971) I got convincded that `for` loop is a valid way to write this code.

tjzel (approved):

PR #6105, opened 2024-06-12
Make all links from Reanimated Cookbook relative

patrycjakalinska:
Before some links were absolute, and other were relative. This PR makes all links from examples relative.


kacperkapusciak (approved):
Thanks! Now docs properly build with `yarn build` 💯 

Because the docs CI is broken it failed to prevent merging incorrect links to repo

PR #6106, opened 2024-06-12
Fix nightly build CI

szydlovsky:
## Summary

Nightly build CI got hit with a recoil from monorepo integration.
Also, running `npm publish` from Reanimated package root (that contained packed `.tgz` file) would use files of Reanimated instead of the package, thus the move operation.

## Test plan

image


tjzel (approved):
Me likes

PR #6115, opened 2024-06-13
fix: don't overwrite `propKeysManagedByAnimated` with `nil`

kirillzyusko:
## Summary

Do not overwrite `propKeysManagedByAnimated` with `nil`

## Explanation

When you copy `propKeysManagedByAnimated` - you are copying `nil`. Later on in 

```objc
[_mountingManager synchronouslyUpdateViewOnUIThread:tag changedProps:props componentDescriptor:*componentDescriptor];
```

props gets set to `nil` and then `NSSet` with `transform` key, but in 

```objc
[componentView setPropKeysManagedByAnimated_DO_NOT_USE_THIS_IS_BROKEN:propKeysManagedByAnimated];
```

you set it back to `nil`.

As a result if re-render during animation happens, it may lead to unexpected jumps during the animation.

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp
+++ b/packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp
@@ -519,7 +519,7 @@ jsi::Object NativeReanimatedModule::getUIProps(
         propNames.getValueAtIndex(rt, i).asString(rt).utf8(rt);
     bool isLayoutProp =
         nativePropNames_.find(propName) != nativePropNames_.end();
-    if (isLayoutProp) {
+    if (!isLayoutProp) {
       const jsi::Value &propValue = props.getProperty(rt, propName.c_str());
       res.setProperty(rt, propName.c_str(), propValue);
     }
@@ -675,14 +675,15 @@ void NativeReanimatedModule::performOperations() {
     }
   }
 
+  // If there's no layout props to be updated, we can apply the updates
+  // directly onto the components and skip the commit.
+  for (const auto &[shadowNode, props] : copiedOperationsQueue) {
+    Tag tag = shadowNode->getTag();
+    jsi::Object uiProps = getUIProps(rt, props->asObject(rt));
+    synchronouslyUpdateUIPropsFunction_(rt, tag, uiProps);
+  }
+
   if (!hasLayoutUpdates) {
-    // If there's no layout props to be updated, we can apply the updates
-    // directly onto the components and skip the commit.
-    for (const auto &[shadowNode, props] : copiedOperationsQueue) {
-      Tag tag = shadowNode->getTag();
-      jsi::Object uiProps = getUIProps(rt, props->asObject(rt));
-      synchronouslyUpdateUIPropsFunction_(rt, tag, uiProps);
-    }
     return;
   }
 
--- a/packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp
+++ b/packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp
@@ -509,6 +509,24 @@ bool NativeReanimatedModule::isThereAnyLayoutProp(
   return false;
 }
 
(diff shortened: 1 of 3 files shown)


tomekzaw:
Hi @kirillzyusko, first of all thanks for this PR!

> As a result if re-render during animation happens, it may lead to unexpected jumps during the animation.

Could you please share a repro of this incorrect behavior so we can get a better understanding of the issue?

---

> Do not overwrite propKeysManagedByAnimated with `nil`

I guess that there's a reason why we call
```objc
NSSet<NSString *> *propKeysManagedByAnimated = [componentView propKeysManagedByAnimated_DO_NOT_USE_THIS_IS_BROKEN];
```
and
```objc
[componentView setPropKeysManagedByAnimated_DO_NOT_USE_THIS_IS_BROKEN:propKeysManagedByAnimated];
```
around calling
```objc
[surfacePresenter synchronouslyUpdateViewOnUIThread:viewTag props:uiProps];
```

`git blame` shows that these lines were added in this PR: https://github.com/software-mansion/react-native-reanimated/pull/5050

Here's the PR description that explains why this change was introduced:
>> The `synchronouslyUpdateViewOnUIThread` method adds all received props to the `propKeysManagedByAnimated_DO_NOT_USE_THIS_IS_BROKEN` registry. If some style prop is added to this registry, then it is unable to animate it from the commit hook. This PR prevents to addition of props managed by Reanimated to the `propKeysManagedByAnimated_DO_NOT_USE_THIS_IS_BROKEN` registry. A similar solution exists in the RN codebase: https://github.com/facebook/react-native/blob/850349b1d274f0cc2595eee2f6bb361a958bb2e2/packages/react-native/React/Fabric/Mounting/RCTMountingManager.mm#L308-L310

Here's an analogous code in react-native repository:
https://github.com/facebook/react-native/blob/4076dbfc8651fcd193b886bf63b23728fad466d7/packages/react-native/React/Fabric/Mounting/RCTMountingManager.mm#L298-L302 and it still exists on main. It looks a bit different, perhaps we should align our code with the current code snippet from react-native?

---
 
Finally, we are considering removing all `[surfacePresenter synchronouslyUpdateViewOnUIThread:props:]` calls because they only apply changes to native views (which is very performant) but at the same time they don't reflect the changes to ShadowTree – this can lead to inconsistencies, e.g. when reading props like `transform` directly from `ShadowTree` which don't include changes made by Reanimated. So maybe a proper fix would be to just get rid of handling non-layout (ex. "UI props") in a special way and commit all props changes to `ShadowTree`, no matter if they affect layout or not. However, we're still quite worried that this would significantly affect performance especially on lower-end Android devices.

kirillzyusko:
Thank you @tomekzaw for your response!

> Could you please share a repro of this incorrect behavior so we can get a better understanding of the issue?

It was an issue with `react-native-reanimated@3.8.0` in Expensify project. This patch helped to get rid off the issue.

However with latest `react-native-reanimated` version (3.15.1) this bug doesn't happen anymore and we removed the patch from our feature branch.

> Finally, we are considering removing all [surfacePresenter synchronouslyUpdateViewOnUIThread:props:] calls because they only apply changes to native views (which is very performant) but at the same time they don't reflect the changes to ShadowTree – this can lead to inconsistencies, e.g. when reading props like transform directly from ShadowTree which don't include changes made by Reanimated. So maybe a proper fix would be to just get rid of handling non-layout (ex. "UI props") in a special way and commit all props changes to ShadowTree, no matter if they affect layout or not. However, we're still quite worried that this would significantly affect performance especially on lower-end Android devices.

Since I made this patch in a collaboration with @Szymon20000 I think it would better to ask him what is the best way to handle this.

From my perspective - if the issue is not reproducible anymore and we deleted the patch, then this PR can be simply closed 🙂

piaskowyk:
Hey Kirill 👋 Thanks for the PR but since we've removed `synchronouslyUpdateViewOnUIThread` and `setPropKeysManagedByAnimated_DO_NOT_USE_THIS_IS_BROKEN`, it seems that these changes are no longer relevant, so I'll go ahead and close this PR.

PR #6116, opened 2024-06-13
Fix SET in Expo Go

piaskowyk:
## Summary

This pull request aims to address the lack of Shared Element Transition in Expo Go. The issue stemmed from the initialization process of SET. In Expo Go, Reanimated gets instantiated twice (once for the Expo client and once for the actual application). During initialization, `REASharedTransitionManager` attempts to swizzle certain methods from `react-native-screens` but aims to do so only once:

```objc
static dispatch_once_t onceToken;
dispatch_once(&onceToken, ^{
  ...
  _isConfigured = YES;
});
```

The variable `_isConfigured` was only set to true if the swizzling process completed successfully. The problem appeared when the second instance attempted to set up the `screens` integration, as it had already been done, resulting in the new instance failing to set `_isConfigured` to true, even if the method swizzling was successful.

The solution involves changing `_isConfigured` to a class field instead of an instance field.

## Test plan

Run Shared Element Transition in Expo Go

### DIFF
--- a/packages/react-native-reanimated/apple/LayoutReanimation/REASharedTransitionManager.m
+++ b/packages/react-native-reanimated/apple/LayoutReanimation/REASharedTransitionManager.m
@@ -27,7 +27,6 @@
   NSMutableSet<REAUIView *> *_reattachedViews;
   BOOL _isStackDropped;
   BOOL _isAsyncSharedTransitionConfigured;
-  BOOL _isConfigured;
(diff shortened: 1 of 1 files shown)


piaskowyk:
@EvanBacon - fixed, thanks for the help yesterday 🤝

tomekzaw (approved):

PR #6121, opened 2024-06-14
Fix hot-reload in web-example

piaskowyk:
## Summary

This PR resolves the issue of hot reload not working after modifying the example code in a common app or the source code of Reanimated.

## Test plan

- open web-example
- modify an example in common-app and see if hot-reload happened
- modify a source code of reanimated and see if hot-reload happened


m-bert (approved):
Thanks a lot ❤️

PR #6123, opened 2024-06-14
Merge ViewDescriptors implementation

piaskowyk:
## Summary

The motivation behind this PR was the lack of animation in the web implementation when React.StrictMode was enabled. While debugging, I discovered we had two similar mechanisms for storing connection between components and animations: one for mobile animations and one for web animations. I realized that the native implementation worked flawlessly without any issues on the web as well, as it had been recently refactored to align with React's guidelines. I decided to remove the flawed web implementation and replace it with mobile implementation.

## Test plan

1. Open web-example and check if animations works.
2. Wrap any web example from example app with `<React.StrictMode></React.StrictMode>` and see if animations works.


packages/react-native-reanimated/src/hook/commonTypes.ts
```diff
@@ -13,13 +13,14 @@ import type {
...
-import type { ViewDescriptorsSet, ViewRefSet } from '../ViewDescriptorsSet';
+import type { ViewDescriptorsSet } from '../ViewDescriptorsSet';
 import type { AnimatedStyle } from '../helperTypes';
+import { ReanimatedHTMLElement } from '../js-reanimated';
 
 export type DependencyList = Array<unknown> | undefined;
 
 export interface Descriptor {
-  tag: number;
+  tag: number | ReanimatedHTMLElement;
```
piaskowyk, line 23: I'm considering whether we should change the name of this prop from `tag` to something more accurate, but I'm unsure if it's a good idea. If anyone has a strong opinion, I'm open to suggestions.

tomekzaw: We can do that in a separate PR.

m-bert: Imho `tag` is fine, the only thing that has changed is that now it may represent `HTMLElement`. On the other hand, those elements are also called `tags` so we could leave that.

But if you have en example of more descriptive name, we can rename it.

m-bert (approved):
LGTM

tjzel (changes requested):
Just blocking it for now because I know this code more in-depth and want to review it properly before we merge it. 🙏

tjzel (approved):
We need some TS fixes for these changes in the pipeline, but it's of the least concern for now. Beside this, all is solid!

PR #6136, opened 2024-06-19
Add tests of predefined exiting

Latropos:
## Summary
Add snapashot tests of predefined exiting

## Test plan

Run tests


apps/common-app/src/examples/RuntimeTests/tests/layoutAnimations/exiting/predefinedExiting.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  await wait(waitTime);
+  const updates = updatesContainer.getUpdates();
+  await unmockAnimationTimer();
+  await clearRenderOutput();
+
+  return updates;
+}
+
+describe.only('Test predefOuted exiting', () => {
```
szydlovsky, line 117: What does `predefOuted` mean? 😄

szydlovsky (approved):
Everything fine, just a small nit

latekvo (approved):
Couldn't find any issues 👍

apps/common-app/src/examples/RuntimeTests/tests/layoutAnimations/exiting/predefinedExiting.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  await wait(waitTime);
+  const updates = updatesContainer.getUpdates();
+  await unmockAnimationTimer();
+  await clearRenderOutput();
+
+  return updates;
+}
+
+describe.only('Test predefined exiting', () => {
```
piaskowyk, line 117: Do we need `describe.only` here? It seem like just `describe` should be enough.

apps/common-app/src/examples/RuntimeTests/tests/layoutAnimations/exiting/predefinedExiting.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+  describe('exiting on mount, no modifiers', () => {
+    test.each(EXITING_SETS)('Test suite of ${0}Out', async ([_setName, exitingSet, waitTime]) => {
+      for (const exiting of exitingSet) {
+        const snapshotName = (exiting as any).name;
+        const updates = await getSnapshotUpdates(exiting, waitTime, undefined);
+
+        expect(updates).toMatchSnapshots(
+          NoModifierExitingSnapshots[snapshotName as keyof typeof NoModifierExitingSnapshots],
+        );
```
piaskowyk, line 126: Why do we only check for JS updates instead of both JS and native updates? 🤔 The primary goal is to verify that updates applies correctly on the native side as well.

piaskowyk (changes requested):

piaskowyk (approved):
When it comes to native snapshots, it requires some additional work, so let's leave it for now.

PR #6142, opened 2024-06-20
Run formatter

piaskowyk:
## Summary

I have just ran `yarn format` and this is an effect 🤷‍♂️

PR #6147, opened 2024-06-20
Fix C++ linter on CI

piaskowyk:
## Summary

This PR aims to fix the C++ linter on CI after transition to the monorepo.


piaskowyk:
image
This means that it is working again. When we merge [this PR](https://github.com/software-mansion/react-native-reanimated/pull/6142) into this one, it will show as green 🎉

tomekzaw (approved):
Thanks for this PR!

PR #6151, opened 2024-06-21
[LA] Replace Layout Transitions example

patrycjakalinska:
We replaced `Default Layout Transition` with refreshed example.

Before:
image

After:

https://github.com/software-mansion/react-native-reanimated/assets/59940332/bafb040f-8d4b-4656-b6ea-3a12cafd2ea5


apps/common-app/src/examples/LayoutAnimations/LayoutTransitionExample.tsx
piaskowyk, line 1: On iOS appears strange warning 🤔 
image

piaskowyk, line 1: Fading transition acting a bit strange 🤔 should we report it as a bug?

https://github.com/software-mansion/react-native-reanimated/assets/36106620/3f589872-6e62-4871-943e-a2b471823ff3

apps/common-app/src/examples/LayoutAnimations/LayoutTransitionExample.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+type TransitionProps = StackScreenProps<
+  RootStackParamList,
+  'linear' | 'fading' | 'sequenced' | 'jumping' | 'curved' | 'entryexit'
+>;
+
+function Transition({ route }: TransitionProps) {
+  const { title, transition } = route.params;
+  const [items, setItems] = useState(INITIAL_LIST);
+  const [selected, _] = useState({ label: title, value: transition });
```
piaskowyk, line 186: Then we don't need setter, so it seems like we can just write:
```suggestion
  const selected = { label: title, value: transition };
```

apps/common-app/src/examples/LayoutAnimations/LayoutTransitionExample.tsx
```diff
@@ -0,0 +1,295 @@
...
+        }
+      />
+      <Button
+        title="Entry/Exit Transition"
+        onPress={() =>
+          navigation.navigate('entryexit', {
+            title: 'EntryExit',
+            transition:
+              EntryExitTransition.entering(FlipInEasyX).exiting(FlipOutYRight),
+          })
```
piaskowyk, line 110: "Why do we need to define those parameters twice - here and as `initialParams`? Is there a way to simplify this?"

apps/common-app/src/examples/LayoutAnimations/LayoutTransitionExample.tsx
```diff
@@ -0,0 +1,295 @@
...
+];
+
+type RootStackParamList = {
+  Home: undefined;
+  linear: { title: string; transition: any };
+  fading: { title: string; transition: any };
+  sequenced: { title: string; transition: any };
+  jumping: { title: string; transition: any };
+  curved: { title: string; transition: any };
+  entryexit: { title: string; transition: any };
```
piaskowyk, line 50: I'm aware that we don't have any public type for transitions, but I'm considering whether replacing `any` with `unknown` would be a better choice. What are your thoughts on this?

apps/common-app/src/examples/LayoutAnimations/LayoutTransitionExample.tsx
```diff
@@ -0,0 +1,295 @@
...
+interface ItemsProps {
+  selected: { label: string; value: any };
+  items: Item[];
+  onRemove: (id: number) => void;
+}
+
+function Items({ selected, items, onRemove }: ItemsProps) {
+  return (
+    <View style={styles.gridContainer}>
+      {items.map((item: { id: number; content: string; color: string }) => (
```
piaskowyk, line 212: Can we just do in this way? Because you have already defined `Item` type.
```suggestion
      {items.map((item: Item) => (
```

apps/common-app/src/examples/LayoutAnimations/LayoutTransitionExample.tsx
patrycjakalinska: 🫢

patrycjakalinska: afaik the transition is working on all the elements that are changing position, maybe that's because of my implementation, not the problem in the core?

apps/common-app/src/examples/LayoutAnimations/LayoutTransitionExample.tsx
patrycjakalinska: As we discussed on dms, it's not possible to use `unknown` - `any` stays 🆗

piaskowyk (approved):
Just two small things, but apart from that, it is ready to merge 👍

PR #6153, opened 2024-06-21
Properly stringify tiny color values

swansontec:
## Summary

If the interpolated alpha value is really tiny, the code might emit a string like "rgba(0, 0, 0, 1e-11)". This is not valid syntax, and can cause worklet crashes. Replace these small values with 0.

## Test plan

This PR includes the necessary unit-test update:

```ts
  it('handles tiny values', () => {
    const colors = ['#00000000', '#ff802001'];

    // We don't want output like "rgba(4, 2, 0, 3.921568627450981e-7)":
    const interpolatedColor = interpolateColor(0.0001, [0, 1], colors);
    expect(interpolatedColor).toBe(`rgba(4, 2, 0, 0)`);
  });
```


tomekzaw (approved):
LGTM, thanks for the PR!

piaskowyk (approved):
Thanks for your PR! 🎉

PR #6160, opened 2024-06-24
Add remark about reading SharedValue from the js thread

j-piasecki:
## Summary

Reading from a shared value on the js thread may block the JS thread until the UI runtime finishes whatever it's doing. This isn't really communicated clearly and adding it to the docs seems like the best option as we don't want to show warnings.

## Test plan

I asked ChatGPT to point out problems (it did).

### DIFF
--- a/packages/docs-reanimated/docs/core/useSharedValue.mdx
+++ b/packages/docs-reanimated/docs/core/useSharedValue.mdx
@@ -67,6 +67,8 @@ import SharedValueSrc from '!!raw-loader!@site/src/examples/SharedValue';
 
 - When you change the `sv.value` Reanimated will update the styles and keep the shared value in sync between the threads. However, this won't trigger a typical React re-render because a shared value is a plain JavaScript object.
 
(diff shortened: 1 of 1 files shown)


packages/docs-reanimated/docs/core/useSharedValue.mdx
```diff
@@ -67,6 +67,8 @@ import SharedValueSrc from '!!raw-loader!@site/src/examples/SharedValue';
 
 - When you change the `sv.value` Reanimated will update the styles and keep the shared value in sync between the threads. However, this won't trigger a typical React re-render because a shared value is a plain JavaScript object.
 
+- When you read the `sv.value` on the [JavaScript thread](/docs/fundamentals/glossary#javascript-thread), the thread will get blocked until the value is fetched from the [UI thread](/docs/fundamentals/glossary#ui-thread). In most cases it will be negligible, but if you are running a lot of animations or reading a value multiple times, the wait time needed to synchronize both threads may significantly increase.
```
piaskowyk, line 70: "but if you are running a lot of animations" - This suggests that the animations themselves can impact JS thread performance. I think a more accurate statement would be something more like this - "but if the UI thread is busy" 🤔 what do you think?

j-piasecki: Sounds good to me

piaskowyk (approved):
🫡

kacperkapusciak (approved):

joe-sam:
This is confusing 🫡 and possibly requires some investigation. 
Shouldn't all **Read** operations on a shared objects be non-blocking unless there are one or more multiple **_write_** operation (object locks) on the object in question on the UI thread.

PR #6144, opened 2024-06-20
docs: Rewrite Layout Transition documentation page

patrycjakalinska:
To be up to date we rewritten [Layout Transition](https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/layout-transitions), implemented InteractiveExample with selectable transition and added new, refreshed videos.

- [x] Jumping, Curved and Entry/Exit Transition are available on web

- [x] This needs latest react-native-reanimated@nightly

- [x] It includes new updated Fading default `duration` from this [PR](https://github.com/software-mansion/react-native-reanimated/pull/6174).

- [x] check default duration in transitions

Before:

https://github.com/software-mansion/react-native-reanimated/assets/59940332/9d845bcf-776c-4514-aa99-c825fb4ceef6

After:

https://github.com/software-mansion/react-native-reanimated/assets/59940332/1e721176-273f-4740-b0a5-27f69f4a6140

How interactivExample works:

https://github.com/software-mansion/react-native-reanimated/assets/59940332/cbe09f76-9579-4b13-9c4d-cbbd23911934

PR #6161, opened 2024-06-24
Deprecate `combineTransition`

piaskowyk:
## Summary

The `combineTransition` method doesn't seem to be necessary. It doesn't add any additional logic, duplicates the public API, and misleadingly suggests that `combineTransition` and `EntryExitTransition` are distinct entities. In fact, it has even been removed from the documentation page [here](https://github.com/software-mansion/react-native-reanimated/pull/6144).


szydlovsky (approved):
Looks about right

kacperkapusciak (approved):

PR #6162, opened 2024-06-24
Fix all autoFixable problems inside app/common

Latropos:
## Summary
Run eslint with `--fix` flag.
These test are broken since when introducing monorepo we unified eslint rules and made them more demanding.

Most of the changes is to `prefer import type`.

The others are:
* Remove type cast when its not needed
image

* Don't have unnecessary conditions
image

* Use property shorthand
image

* When possible use `const` instead of `let`
image

## Test plan

All changes were made automatically.


piaskowyk (approved):
Seems ok 👍

PR #6163, opened 2024-06-24
Name every worklet function

tjzel:
## Summary

Currently, in case of unnamed worklets, such as

```ts
() => {
  'worklet';
  // some logic
}

function () {
  'worklet'
  // some logic
}
```

Those functions are named `anonymous` and parsed as such on the Worklet Runtime. However, this is a pitfall for Reanimated. When the `anonymous` function crashes and the stack trace is vague - it usually is, especially in production - the first suspicion is the error in Reanimated's core logic, since we don't know which function in particular has failed.

In case of a faulty logic in a third-party library this is double as harmful. It doesn't lead the creators of these libraries into the pit of success, since they might not be aware it's their code causing issues. Conversely, Reanimated maintainers receive a very obscure issue and engage in rather unpleasant debugging process - only to discover later the library's fault.

My idea is to:

1. Add source library name to every worklet function.
2. Add source file name to every worklet function.
3. Add ID to every worklet function.

## Test plan

- [x] Update the plugin test snapshots.
- [x] Add relevant test cases if not present already.
- [x] Try to recreate a crash for a third party library, compare the clarity of raised exceptions.

<table>
<th>Before</th><th>After</th>
<tr><td colspan=2>Production developer environment</td></tr>
<tr><td></td><td>Screenshot 2024-06-26 at 11 26 01</td></tr>
<tr><td colspan=2>Sentry stack trace</td></tr>
<tr><td></td><td></td>
</tr>
</table>

Sentry stack traces are coming from Android. Unfortunately, regardless of these changes iOS Sentry stack traces look like this:

Screenshot 2024-06-26 at 12 42 24


bartlomiejbloniarz (approved):

packages/react-native-reanimated/__tests__/plugin.test.ts
```diff
@@ -147,7 +148,8 @@ describe('babel plugin', () => {
       </script>`;
 
       const { code } = runPlugin(input);
-      expect(code).toContain('const foo=this._recur');
+      // expect(code).toContain('const foo=this._recur');
```
tomekzaw, line 151:
```suggestion
```

packages/react-native-reanimated/__tests__/plugin.test.ts
```diff
@@ -1698,7 +1700,10 @@ describe('babel plugin', () => {
       </script>`;
 
       const { code } = runPlugin(input);
-      expect(code).toContain('var foo = function* foo() {');
+      expect(code).toMatch(
+        /var foo_null[0-9]* = function\* foo_null[0-9]*\(\) {/gm
+      );
+      // expect(code).toContain('var foo = function* foo() {');
```
tomekzaw, line 1706:
```suggestion
```

packages/react-native-reanimated/plugin/src/workletFactory.ts
```diff
@@ -325,17 +325,44 @@ function hash(str: string): number {
...
+    return fun.node.id.name + suffix;
   }
   if (isFunctionExpression(fun.node) && isIdentifier(fun.node.id)) {
-    return fun.node.id.name;
+    return fun.node.id.name + suffix;
   }
-  return 'anonymous'; // fallback for ArrowFunctionExpression and unnamed FunctionExpression
+
+  // Fallback for ArrowFunctionExpression and unnamed FunctionExpression.
+  return suffix;
```
tomekzaw, line 365: `suffix` starts with an underscore, can we get rid of it here?
```ts
const suffix = `${source}${functionId++}`;

return `${fun.node.key.name}_${suffix}`;
```

tomekzaw (changes requested):

packages/react-native-reanimated/plugin/src/workletFactory.ts
tjzel: @tomekzaw Done!

packages/react-native-reanimated/plugin/src/workletFactory.ts
```diff
@@ -325,17 +325,44 @@ function hash(str: string): number {
...
+    source = basename(filepath).split('.')[0];
+    const splitFilepath = filepath.split('/');
+    // Get the library name from the path.
+    const nodeModulesIndex = splitFilepath.indexOf('node_modules');
+    if (nodeModulesIndex !== -1) {
+      // Remove all non-alphanumeric characters.
+      const libraryName = splitFilepath[nodeModulesIndex + 1].replace(
+        /\W/g,
+        ''
+      );
```
piaskowyk, line 348: I think it would be good to implement the same data sanitization for the file name (`source`) as well

packages/react-native-reanimated/plugin/src/workletFactory.ts
```diff
@@ -325,17 +325,44 @@ function hash(str: string): number {
...
+      source = libraryName + '_' + source;
+    }
+  }
+
+  const suffix = source + functionId++;
```
piaskowyk, line 353: Maybe just to be safe and prevent potentially strange errors when `source` is `undefined` or `null` (theoreticly `source = basename(filepath).split('.')[0];` can produce `undefined` for a file without extension), let's replace it with the following syntax:
```js
const suffix = `${source}${functionId++}`;
```

But if you think that is overkill you don't need to 😅

tjzel: @piaskowyk `source` is initialized with `unknown_file` in line 334 😅

packages/react-native-reanimated/plugin/src/workletFactory.ts
tjzel: That's actually a very good remark!

packages/react-native-reanimated/plugin/src/workletFactory.ts
tjzel: Also nothing bad will happen if there's no extension.

Screenshot 2024-07-03 at 09 26 17

piaskowyk: ok, you convinced me 😅

packages/react-native-reanimated/__tests__/plugin.test.ts
```diff
@@ -147,7 +149,109 @@ describe('babel plugin', () => {
...
+            'worklet';
+          },
+        };
+      </script>`;
+
+      // TODO: this is an edge case that wasn't ever handled.
+      expect(() => runPlugin(input)).toThrow();
+      // const { code } = runPlugin(input);
+      // expect(code).toMatch(/function foo_null[0-9]+\(\)/gm);
+      // expect(code).toMatchSnapshot();
```
tomekzaw, line 199:
```suggestion
```

packages/react-native-reanimated/plugin/src/workletFactory.ts
```diff
@@ -325,17 +327,38 @@ function hash(str: string): number {
...
+    return toIdentifier(fun.node.key.name + '_' + suffix);
   }
   if (isFunctionDeclaration(fun.node) && isIdentifier(fun.node.id)) {
-    return fun.node.id.name;
+    return toIdentifier(fun.node.id.name + '_' + suffix);
   }
   if (isFunctionExpression(fun.node) && isIdentifier(fun.node.id)) {
-    return fun.node.id.name;
+    return toIdentifier(fun.node.id.name + '_' + suffix);
   }
```
tomekzaw, line 358: I don't like using `+` for concatenation, why not go for old trusted `$`?

packages/react-native-reanimated/plugin/src/workletFactory.ts
```diff
@@ -325,17 +327,38 @@ function hash(str: string): number {
...
+  if (state.file.opts.filename) {
+    const filepath = state.file.opts.filename;
+    source = basename(filepath);
+
+    // Get the library name from the path.
+    const splitFilepath = filepath.split('/');
+    const nodeModulesIndex = splitFilepath.indexOf('node_modules');
+    if (nodeModulesIndex !== -1) {
+      const libraryName = splitFilepath[nodeModulesIndex + 1];
+      source = libraryName + '_' + source;
```
tomekzaw, line 345: and here

packages/react-native-reanimated/plugin/src/plugin.ts
```diff
@@ -50,6 +52,14 @@ module.exports = function (): PluginItem {
           });
         },
       },
+      Program: {
+        enter(_path, state) {
+          runWithTaggedExceptions(() => {
+            // Reset worklet number.
+            state.workletNumber = 1;
```
tomekzaw, line 59: Let's use worklet hash instead as it is more stable than autoincrement.

piaskowyk (approved):

packages/react-native-reanimated/__tests__/plugin.test.ts
tjzel: I keep it here on purpose since this the correct assertion.

packages/react-native-reanimated/plugin/src/plugin.ts
tjzel: Worklet hashes depend on the function name 🤓

tomekzaw: Please do anyway

packages/react-native-reanimated/plugin/src/workletStringCode.ts
```diff
@@ -29,16 +29,19 @@ import { strict as assert } from 'assert';
...
+import type { ReanimatedPluginPass, WorkletizableFunction } from './types';
 
 const MOCK_SOURCE_MAP = 'mock source map';
 
 export function buildWorkletString(
   fun: BabelFile,
+  state: ReanimatedPluginPass,
   closureVariables: Array<Identifier>,
-  name: string,
+  newName: string,
```
tomekzaw, line 40:
```suggestion
  nameWithFilename: string,
```

packages/react-native-reanimated/plugin/src/workletStringCode.ts
```diff
@@ -72,7 +75,7 @@ export function buildWorkletString(
 
   assert(inputMap, '[Reanimated] `inputMap` is undefined.');
 
-  const includeSourceMap = !isRelease();
+  const includeSourceMap = !(isRelease() || state.opts.disableSourceMaps);
```
tomekzaw, line 78: How about `excludeSourceMap`?

packages/react-native-reanimated/plugin/src/plugin.ts
tjzel: TBD

packages/react-native-reanimated/plugin/src/workletFactory.ts
```diff
@@ -325,17 +327,38 @@ function hash(str: string): number {
...
+    const filepath = state.file.opts.filename;
+    source = basename(filepath);
+
+    // Get the library name from the path.
+    const splitFilepath = filepath.split('/');
+    const nodeModulesIndex = splitFilepath.indexOf('node_modules');
+    if (nodeModulesIndex !== -1) {
+      const libraryName = splitFilepath[nodeModulesIndex + 1];
+      source = `${libraryName}_${source}`;
+    }
```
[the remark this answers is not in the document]
tjzel: I wanted to point away from Reanimated, not necessarily to the exact culprit - I don't want to bloat the bundle too much with such long names.

PR #6164, opened 2024-06-25
Add `ReducedMotionConfig` component

piaskowyk:
## Summary

This PR adds `ReduceMotionConfig` component that allows to determine the default animation behavior in response to the device's reduced motion accessibility setting. It affects application globally. The default behavior disables all animation if reduced motion is enabled on a device. You can utilize this component to override that behavior.

### Usage

```js
function App() {
  
  return (
    // ...
    <ReduceMotionConfig mode={ReduceMotion.Never} />
    // ...
  );
}
```

### Demo

https://github.com/software-mansion/react-native-reanimated/assets/36106620/e15a484e-bdc1-454c-b9ff-3ae8863db62e

## Test plan

Open `Reduce Motion` example from example app.


packages/react-native-reanimated/src/component/ReduceMotionConfig.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * @see https://docs.swmansion.com/react-native-reanimated/docs/components/ReduceMotionConfig
+ */
+export function ReduceMotionConfig({ mode }: { mode: ReduceMotion }) {
+  const firstRender = useRef(true);
+  if (__DEV__ && firstRender.current) {
+    firstRender.current = false;
+    console.warn(
+      `[Reanimated] Reduced motion setting is overwritten with mode '${mode}'.`
+    );
+  }
```
szydlovsky, line 19: According to React purity dogma, we should move it to its own `useEffect` block with empty dependency array (runs just once on mount). Then, we don't need to read and write `firstRender` ref during render (which is against function purity).

szydlovsky (changes requested):
I like the idea of overriding it this simply!

packages/react-native-reanimated/src/component/ReduceMotionConfig.tsx
szydlovsky: I think we even don't need to use refs if we put it into a `useEffect` with empty deps, since it then behaves like `componentDidMount` callback (and I think we want the warning every mount, don't we?)

szydlovsky: Update: this approach doesn't interfere with purity of functions BUT omitting using a flag is always a good idea

packages/react-native-reanimated/src/component/ReduceMotionConfig.tsx
```diff
@@ -0,0 +1,33 @@
...
+    if (!__DEV__) {
+      return;
+    }
+    console.warn(
+      `[Reanimated] Reduced motion setting is overwritten with mode '${mode}'.`
+    );
+  }, []);
+
+  useEffect(() => {
+    const wasEnabled = IS_REDUCED_MOTION.value;
```
tomekzaw, line 23: Looks like here we're reading a ~~shared value~~ mutable on the JS thread. and there are some performance concerns around that. Is it really safe to do so here?

piaskowyk: I believe it shouldn't have a significant impact, but I've thought of a new idea. Let me know what you think.

packages/react-native-reanimated/src/component/ReduceMotionConfig.tsx
```diff
@@ -0,0 +1,33 @@
...
+    console.warn(
+      `[Reanimated] Reduced motion setting is overwritten with mode '${mode}'.`
+    );
+  }, []);
+
+  useEffect(() => {
+    const wasEnabled = IsReduceMotion.jsValue;
+    if (mode === ReduceMotion.Never) {
+      IsReduceMotion.setEnabled(false);
+    }
```
szydlovsky, line 26: What about a situation where someone sets component's mode to `.Always` and system's reduced motion is turned off? The reduced motion will stay OFF. This case is not handled. I'd add an `else if` case for `mode === ReduceMotion.Always`

szydlovsky (changes requested):
I believe it lacks some logic

packages/react-native-reanimated/src/component/ReduceMotionConfig.tsx
piaskowyk: In my assumptions, this component was meant to disable reduced motion, but you are right. It could be confusing that we can pass more types of actions here. I have implemented it in this [commit](https://github.com/software-mansion/react-native-reanimated/pull/6164/commits/3bf2eea1bf9f11b54f590ce1abd61ab6864f71d3).

packages/react-native-reanimated/src/component/ReduceMotionConfig.tsx
```diff
@@ -0,0 +1,43 @@
+'use strict';
+import { useEffect } from 'react';
+import { ReduceMotion } from '../commonTypes';
+import { IsReduceMotion } from '../hook/useReducedMotion';
+import { isReducedMotion } from '../PlatformChecker';
```
szydlovsky, line 5: The naming convention 😵‍💫😵‍💫

piaskowyk: I need to find a better name 😅

szydlovsky (approved):
Love the refactor - just please tak a look at linter

kacperkapusciak (approved):
lgtm

diegolmello:
Awesome. Thanks for working on this important topic ❤️

vyobukhov:
Hi guys, 

do we really need this annoying warning on every app init in dev mode?
```
console.warn(
      `[Reanimated] Reduced motion setting is overwritten with mode '${mode}'.`
    );
```

If it was overwritten, it was done on purpose, so what is the point to remind about it on every app reload in dev?

piaskowyk:
To avoid any confusion, if any library that you depend on call it in their code. If you want to silence this warning, just use:
```js
import { LogBox } from 'react-native';
LogBox.ignoreLogs(['[Reanimated] Reduced motion']);
```

PR #6168, opened 2024-06-25
Layout Transition test

Latropos:
## Summary
Create tests of **LAYOUT TRANSITIONS**

Tests include:
* components moving UP, DOWN, LEFT, RIGHT, LEFT & UP, RIGHT & UP
* Various modifiers
* Some verification that changing the order of the modifiers doesn't modify the snapshot


packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/FadingTransition.ts
```diff
@@ -29,7 +29,7 @@ export class FadingTransition
     const delayFunction = this.getDelayFunction();
     const callback = this.callbackV;
     const delay = this.getDelay();
-    const duration = this.durationV ?? 500;
+    const duration = (this.durationV ?? 500) / 2;
```
piaskowyk, line 32:
```suggestion
    const duration = this.durationV / 2 ?? 250;
```

piaskowyk: This is great change! But this change is beyond the scope of this PR. Could you please move it to a separate PR, and then rebase this PR onto the new one with the updated defaults?

apps/common-app/src/examples/RuntimeTests/tests/layoutAnimations/layout/testComponents.test.tsx
```diff
@@ -0,0 +1,119 @@
+import React, { useState, useEffect } from 'react';
+import { View, StyleSheet } from 'react-native';
+import Animated from 'react-native-reanimated';
+import { useTestRef } from '../../../ReanimatedRuntimeTestsRunner/RuntimeTestsApi';
+
+export const TRANSITION_REF = 'TRANSITION_REF';
```
piaskowyk, line 6: Technically this is not a ref but just a name of ref 😅

apps/common-app/src/examples/RuntimeTests/tests/layoutAnimations/layout/testComponents.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+  return (
+    <View style={styles.containerVertical}>
+      {direction === Direction.UP && (
+        <>
+          {show && <Animated.View layout={layout} style={styles.animatedBox} />}
+          <Animated.View layout={layout} style={styles.animatedBox} />
+          <Animated.View ref={ref} layout={layout} style={[styles.animatedBox, styles.mainBox]} />
+        </>
+      )}
+      {direction === Direction.DOWN && (
```
piaskowyk, line 33: What does the `direction` property describe?

packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/FadingTransition.ts
Latropos: Yes, I just want to merge the other PR first

apps/common-app/src/examples/RuntimeTests/tests/layoutAnimations/layout/testComponents.test.tsx
Latropos: It is the direction of movement of the component that we record a snapshot of.

Latropos: Please see that the orange view is moving in different directions:

https://github.com/software-mansion/react-native-reanimated/assets/56199675/58c19501-f6dc-47c3-800c-4c5cd4ddb4c6

piaskowyk (approved):

apps/common-app/src/examples/RuntimeTests/tests/layoutAnimations/layout/testComponents.test.tsx
Latropos: You are completely right, but I was using this style a lot. I think that it is not bad to keep this misnomer, since:
* it is shorter
* it is still easy to guess that it is the ref name, not the ref, because we use UPPER_CASE syntax for hardcoded constant data.

PR #6171, opened 2024-06-26
fix empty export on web

EvanBacon:
## Summary

React Native Babel preset uses a loose plugin to convert imports to commonjs, we're investigating switching to a more standard approach in Expo CLI here https://github.com/expo/expo/pull/30005 which aligns with how Metro is used at Meta.
 
When running on the kitchen sink, I found that there was an invalid expression in reanimated where it exports null and then destructures null. This error didn't occur previously because the commonjs plugin deferred access to the null object until later.

**Before**

```js
var _Math = _$$_REQUIRE(_dependencyMap[0]);

// Later ...

_Math.add
```

This coincidentally didn't throw because the deferred code is never accessed.

**After**

```js
var add = _$$_REQUIRE(_dependencyMap[0]).add;
```

This throws because the `null` is accessed when the module is loaded.

## Test plan

Bundle a Metro web project with experimentalImportSupport enabled:

```js
config.transformer.getTransformOptions = async () => ({
  transform: {
    experimentalImportSupport: true,
    inlineRequires: false,
  },
});
```

Then run `npx serve dist` -> Error was there but not anymore.

### DIFF
--- a/packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
+++ b/packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
@@ -1,3 +1,3 @@
 // RNRender is not used for web. An export is still defined to eliminate warnings from bundlers such as esbuild.
-
+'use strict';
 export {};
--- a/packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
+++ b/packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
@@ -1,3 +1,3 @@
-'use strict';
 // RNRender is not used for web. An export is still defined to eliminate warnings from bundlers such as esbuild.
-module.exports = null;
+
+export {};


packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
```diff
@@ -1,3 +1,3 @@
-'use strict';
```
piaskowyk, line 1: May I restore `'use strict';` here if it does not break anything for you?

piaskowyk:
Thank you for the fix and detailed description!

packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
piaskowyk, line 1: May I restore `‘use strict’;` here if it does not break anything for you?

EvanBacon: should be fine to add back!

packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
```diff
@@ -1,3 +1,3 @@
...
 // RNRender is not used for web. An export is still defined to eliminate warnings from bundlers such as esbuild.
-module.exports = null;
+
```
EvanBacon, line 2:
```suggestion
'use strict';
```

packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
piaskowyk: Thanks!

codeion:
Hi,
It's breaking build with vite bundler, maybe to consider this patch...
[react-native-reanimated+3.13.0.patch](https://github.com/user-attachments/files/16107917/react-native-reanimated%2B3.13.0.patch)

djaffer:
WARNING in ./node_modules/react-native-reanimated/lib/module/createAnimatedComponent/createAnimatedComponent.js:265
export 'RNRenderer' (imported as 'RNRenderer') was not found in '../platform-specific/RNRenderer' (module has no exports)
  263 |           viewConfig = null;
  264 |         } else {
> 265 |           var hostInstance = RNRenderer.findHostInstance_DEPRECATED(component);
  266 |           if (!hostInstance) {
  267 |             throw new Error('[Reanimated] Cannot find host instance for this component. Maybe it renders nothing?');
  268 |           }

PR #6172, opened 2024-06-26
[Web LA] Fix chrome flickers

m-bert:
## Summary

Currently some of our animations flicker on google chrome browser. This PR moves setting animation into `requestAnimationFrame` callback - this way we will be sure that animation will be properly set before next repaint.

### Before

https://github.com/software-mansion/react-native-reanimated/assets/63123542/fe674d23-d1bb-44c3-a870-335c330c7073

### After

https://github.com/software-mansion/react-native-reanimated/assets/63123542/2a4833a7-c3d9-4a59-b108-a7f4cc166f1f

## Test plan

Tested on default animations example

### DIFF
--- a/packages/react-native-reanimated/src/layoutReanimation/web/componentUtils.ts
+++ b/packages/react-native-reanimated/src/layoutReanimation/web/componentUtils.ts
@@ -156,6 +156,8 @@ export function setElementAnimation(
   };
 
   if (animationConfig.animationType === LayoutAnimationType.ENTERING) {
+    // On chrome sometimes entering animations flicker. This is most likely caused by animation being interrupted
+    // by already started tasks. To avoid flickering, we use `requestAnimationFrame`, which will run callback right before repaint.
     requestAnimationFrame(configureAnimation);
   } else {
     configureAnimation();
--- a/packages/react-native-reanimated/src/layoutReanimation/web/componentUtils.ts
+++ b/packages/react-native-reanimated/src/layoutReanimation/web/componentUtils.ts
@@ -148,12 +148,18 @@ export function setElementAnimation(
 ) {
   const { animationName, duration, delay, easing } = animationConfig;
 
-  requestAnimationFrame(() => {
+  const configureAnimation = () => {
     element.style.animationName = animationName;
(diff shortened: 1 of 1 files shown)


packages/react-native-reanimated/src/layoutReanimation/web/componentUtils.ts
```diff
@@ -148,10 +148,12 @@ export function setElementAnimation(
 ) {
   const { animationName, duration, delay, easing } = animationConfig;
 
-  element.style.animationName = animationName;
-  element.style.animationDuration = `${duration}s`;
-  element.style.animationDelay = `${delay}s`;
-  element.style.animationTimingFunction = easing;
+  requestAnimationFrame(() => {
```
piaskowyk, line 151: We definitely should add a comment here `requestAnimationFrame` is required here

piaskowyk, line 151: Have you checked if it works with `setTimeout(() => {}, 0)` instead of `requestAnimationFrame`?

piaskowyk, line 151: If I understand correctly, we are delaying every animation start by one frame. Would it be better to only delay entering animations without layout changes and exiting animations?

m-bert: I've just tested that - it seems to work (though I think I've seen one flicker, but I couldn't reproduce it afterwards)

m-bert: Good catch! Done in [b61d796](https://github.com/software-mansion/react-native-reanimated/pull/6172/commits/b61d79600bd2c8169a779e1807d127d0db0ea774)

m-bert: i agree! Done in [9a3ec76](https://github.com/software-mansion/react-native-reanimated/pull/6172/commits/9a3ec76202a25337992b5daa76ec95a096646464)

piaskowyk (approved):
🎉

PR #6174, opened 2024-06-27
Fix duration of fading transition

Latropos:
## Summary
Now the fading transition executes for `2*duration` miliseconds.
We should divide the duration by 2 to fix it, the same way it done for the following transitions:

https://github.com/software-mansion/react-native-reanimated/blob/82d1d9582d96b463ab95d6ba9b8c1c20e7aa867b/packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/JumpingTransition.ts#L33

https://github.com/software-mansion/react-native-reanimated/blob/82d1d9582d96b463ab95d6ba9b8c1c20e7aa867b/packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/SequencedTransition.ts#L44

Also I rename the variables that hold half of the duration and are named "duration" into "halfDuration", as it was suggested during the review
## Test plan

Tests are in this PR: https://github.com/software-mansion/react-native-reanimated/pull/6168


packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/FadingTransition.ts
```diff
@@ -29,7 +29,7 @@ export class FadingTransition
     const delayFunction = this.getDelayFunction();
     const callback = this.callbackV;
     const delay = this.getDelay();
-    const duration = this.durationV ?? 500;
+    const duration = (this.durationV ?? 500) / 2;
```
piaskowyk, line 32: What do you think about:
```suggestion
    const halfAnimationDuration = this.durationV / 2 ?? 250;
```

Latropos: The current change is consistent with implementations of both JumpingTransition and SequencedTransition:


https://github.com/software-mansion/react-native-reanimated/blob/82d1d9582d96b463ab95d6ba9b8c1c20e7aa867b/packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/JumpingTransition.ts#L33

https://github.com/software-mansion/react-native-reanimated/blob/82d1d9582d96b463ab95d6ba9b8c1c20e7aa867b/packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/SequencedTransition.ts#L44

piaskowyk: so, to keep it consistent we can change syntax in `JumpingTransition.ts` and `SequencedTransition.ts` too

Latropos: Sure, I'll update them!
However I would prefer:
`const halfDuration = (this.durationV ?? 500) /2;` instead of
`const halfDuration = this.durationV / 2 ?? 250;`
It's just easier to see that the default is 500, not 250

packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/CurvedTransition.ts
```diff
@@ -111,28 +111,28 @@ export class CurvedTransition
           originX: delayFunction(
             delay,
             withTiming(values.targetOriginX, {
-              duration,
+              duration: halfDuration,
```
piaskowyk, line 114: this is not a half of duration but whole duration of an animation

packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/FadingTransition.ts
```diff
@@ -29,7 +29,7 @@ export class FadingTransition
     const delayFunction = this.getDelayFunction();
     const callback = this.callbackV;
     const delay = this.getDelay();
-    const duration = (this.durationV ?? 500) / 2;
+    const halfDuration = (this.durationV ?? 500) / 2; // The minimal duration is 100ms, otherwise the animation will be halfDuration + 50ms long
```
piaskowyk, line 32: Let't wait with changes in this particular file until the https://github.com/software-mansion/react-native-reanimated/pull/6183 will be merged.

packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/JumpingTransition.ts
```diff
@@ -31,7 +31,8 @@ export class JumpingTransition
     const callback = this.callbackV;
     const delay = this.getDelay();
     const duration = (this.durationV ?? 300) / 2;
-    const config = { duration: duration * 2 };
+    const halfDuration = duration / 2;
+    const config = { duration };
```
piaskowyk, line 35: I think, here is a small mistake, probably it should be more like this
```
const duration = this.durationV ?? 300;
const halfDuration = duration / 2;
const config = { duration };
```

piaskowyk (changes requested):

packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/FadingTransition.ts
```diff
@@ -45,8 +45,8 @@ export class FadingTransition
           opacity: delayFunction(
             delay,
             withSequence(
-              withTiming(0, { duration }),
-              withTiming(1, { duration })
+              withTiming(0, { duration: duration / 2 }),
```
piaskowyk, line 48:
```suggestion
              withTiming(0, { duration: halfDuration }),
```

piaskowyk (changes requested):

packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/FadingTransition.ts
```diff
@@ -29,7 +29,8 @@ export class FadingTransition
     const delayFunction = this.getDelayFunction();
     const callback = this.callbackV;
     const delay = this.getDelay();
-    const duration = (this.durationV ?? 500) / 2;
+    const duration = this.durationV ?? 500;
+    const halfDuration = duration / 2;
```
piaskowyk, line 33:
```suggestion
    const halfDuration = (this.durationV ?? 500) / 2;
```

We don't want to use `duration` anywhere because `originX`, `originY`, `width`, and `height` should be changed in the middle of the animation when the component becomes invisible. Therefore, the animation should look like this:

```js
originX: withDelay(
  delay + halfDuration,
  withTiming(values.targetOriginX, { duration: 0 })
),
```

piaskowyk (changes requested):

piaskowyk (approved):

PR #6180, opened 2024-06-28
[LA] Replace Keyframe example

patrycjakalinska:
We replaced `Keyframe Animation` with refreshed example.

Before:

https://github.com/software-mansion/react-native-reanimated/assets/59940332/0c2b752d-5108-4005-adf7-f5e9fab513b6

After:

https://github.com/software-mansion/react-native-reanimated/assets/59940332/805fc753-e994-4a53-8f9b-229e0690562a


piaskowyk (commented):
Have you verified if it works correctly on the Web platform as well?

PR #6184, opened 2024-06-28
Remove fontawesome dependency from example app

szarbartosz:
## Summary
I noticed that `@fontawesome` dependency is present in the repository but is used in 2 demo screens only - the size of unpacked dependencies are following:
- `@fortawesome/fontawesome-svg-core`: 397 kB
- `@fortawesome/free-solid-svg-icons`: 5.09 MB
- `@fortawesome/react-native-fontawesome`: 33.4 kB

I believe that plain text emojis (which are present on eg. examples list) could be used instead of fontawesome icons.

## Test plan
Open `Tab bar` and `Habits` example from the example app

### DIFF
--- a/apps/common-app/package.json
+++ b/apps/common-app/package.json
@@ -8,9 +8,6 @@
     "format": "prettier --write --list-different ."
   },
   "peerDependencies": {
-    "@fortawesome/fontawesome-svg-core": "*",
-    "@fortawesome/free-solid-svg-icons": "*",
-    "@fortawesome/react-native-fontawesome": "*",
     "@react-native-async-storage/async-storage": "*",
     "@react-native-community/slider": "*",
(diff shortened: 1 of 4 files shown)


kacperkapusciak (commented):
if these icons aren't bundled to the production reanimated package (I believe they aren't) then this change doesn't impact anything

comparing: development hermes engine weights over 1 GB and _it doesn't matter_

why we use fontawesome in the first place? i have no idea but for years we never have any problems with it

kacperkapusciak:
Closing after contacting on DMs

kacperkapusciak (approved):
After consulting on DMs

Removing fontawesome dependency makes it easier to port Reanimated to some out-of-tree platforms on which SVGs aren't fully supported yet which is a valid usecase for removing it from the example app

PR #6188, opened 2024-06-29
fix side effects

EvanBacon:
## Summary

Been playing with an implementation of side effects in Expo CLI and noticed that reanimated has an unmarked side effect which causes tree shaking to remove required globals.


piaskowyk (approved):
Thank you for the fix! 🎉

PR #6191, opened 2024-07-01
Add eslint verification of app-common back to CI

Latropos:
## Summary

CI tests of common-app were removed when introducing monorepo.
In this PR I fix all the broken rules and enable CI again.

## Test plan
I've made a deliberate mistake in app-common to make sure the CI tests work
image

### DIFF
--- a/.github/workflows/static-root-checks.yml
+++ b/.github/workflows/static-root-checks.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Lint react-native-reanimated
         working-directory: packages/react-native-reanimated
         run: yarn lint:js
-      - name: Lint app-common
+      - name: Lint common-app
         working-directory: apps/common-app
         run: yarn lint
       - name: Find unused code
--- a/.github/workflows/static-root-checks.yml
+++ b/.github/workflows/static-root-checks.yml
@@ -39,9 +39,12 @@ jobs:
       - name: Check types
         working-directory: packages/react-native-reanimated
         run: yarn type:check:all
-      - name: Lint
+      - name: Lint reanimated
         working-directory: packages/react-native-reanimated
         run: yarn lint:js
+      - name: Lint app-common
+        working-directory: packages/react-native-reanimated
+        run: yarn lint:app
       - name: Find unused code
         working-directory: packages/react-native-reanimated
         run: yarn find-unused-code:js
--- a/.github/workflows/static-root-checks.yml
+++ b/.github/workflows/static-root-checks.yml
@@ -43,8 +43,8 @@ jobs:
         working-directory: packages/react-native-reanimated
         run: yarn lint:js
       - name: Lint app-common
-        working-directory: packages/react-native-reanimated
-        run: yarn lint:app
+        working-directory: apps/common-app
+        run: yarn lint
       - name: Find unused code
         working-directory: packages/react-native-reanimated
         run: yarn find-unused-code:js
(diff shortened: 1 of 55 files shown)


piaskowyk (dismissed):

.github/workflows/static-root-checks.yml
```diff
@@ -39,9 +39,12 @@ jobs:
       - name: Check types
         working-directory: packages/react-native-reanimated
         run: yarn type:check:all
-      - name: Lint
+      - name: Lint reanimated
```
tomekzaw, line 42:
```suggestion
      - name: Lint react-native-reanimated
```

apps/common-app/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "build": "yarn patch-package",
-    "lint": "eslint --max-warnings=0 .",
+    "lint": "eslint .",
```
tomekzaw, line 7: Why was `--max-warnings=0` removed?

tomekzaw (changes requested):

apps/common-app/package.json
piaskowyk: +1
I had assumed that the default value was `0` previously, but based on the [documentation](https://eslint.org/docs/latest/use/command-line-interface#--max-warnings), it appears that we require this flag.

Latropos: Using `--max-warnings=0` makes the tests fail if there is any warning. It was removed as we probably want CI to pass despite the warnings.

>Normally, if ESLint runs and finds no errors (only warnings), it exits with a success exit status. However, if --max-warnings is specified and the total warning count is greater than the specified threshold, ESLint exits with an error status.

After reading the attached documentation I still think we should either omit this flag or provide `--max-warnings=-1`...
Or we can have `--max-warnings="current number of warnings"` to make sure no new warnings are generated.

piaskowyk: so, why did it work before with this flag?

.github/workflows/static-root-checks.yml
tomekzaw, line 42: I'm submitting the same suggestion again. It looks like it was resolved but for some reason it has been overwritten. Maybe you've force pushed the commits from your local branch?
```suggestion
      - name: Lint react-native-reanimated
```

apps/common-app/package.json
Latropos: >so, why did it work before with this flag?
Probably some of our warnings are generated by rules that weren't enabled before. The broken warnings are `@typescript-eslint/no-explicit-any` and `@typescript-eslint/no-var-requires`. We use var requires a lot for importing assets and most of explicit any are in runtime tests (for example because we don't export type for EntryExitAnimation).

In the future we can config out rule to allow import of tests and assets: https://typescript-eslint.io/rules/no-var-requires/ and fix and ban `@typescript-eslint/no-explicit-any` outside Runtime Tests. However this would be beyond the scope of this PR.

Also if certain rule should stop a PR from being merged then I would recommend marking it as error.

.github/workflows/static-root-checks.yml
```diff
@@ -39,9 +39,12 @@ jobs:
       - name: Check types
         working-directory: packages/react-native-reanimated
         run: yarn type:check:all
-      - name: Lint
+      - name: Lint react-native-reanimated
         working-directory: packages/react-native-reanimated
         run: yarn lint:js
+      - name: Lint app-common
```
tomekzaw, line 45:
```suggestion
      - name: Lint common-app
```

tomekzaw (approved):

PR #6192, opened 2024-07-01
Fix eslint rules inside apps/common-app

Latropos:
## Summary
When introducing monorepo we unified all the eslint rules across the project.
That's why we the eslint rules got more restrictive and we had to fix all the existing problems.

## Test plan
Run `yarn lint` inside apps/common-app

### DIFF
--- a/apps/common-app/src/examples/ShareablesExample.tsx
+++ b/apps/common-app/src/examples/ShareablesExample.tsx
@@ -62,6 +62,7 @@ function RemoteNamedFunctionSyncCallDemo() {
 
 function RemoteAnonymousFunctionSyncCallDemo() {
   const handlePress = () => {
+    // eslint-disable-next-line no-constant-condition
     const foo = true ? () => {} : () => {};
     runOnUI(() => {
       foo();
--- a/apps/common-app/src/examples/SharedElementTransitions/ImageStack.tsx
+++ b/apps/common-app/src/examples/SharedElementTransitions/ImageStack.tsx
@@ -44,9 +44,7 @@ export function ImageStack({
         }
         const marginTop = index === 0 || !collapsed ? 20 : -size + 20;
         return (
-          <View
-            style={[{ zIndex: -index }, styles.center]}
-            key={`${image}@${index}`}>
+          <View style={[{ zIndex: -index }, styles.center]} key={index}>
             <Animated.Image
               source={image}
               style={[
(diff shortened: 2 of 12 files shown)


apps/common-app/src/examples/ShareablesExample.tsx
```diff
@@ -62,6 +62,7 @@ function RemoteNamedFunctionSyncCallDemo() {
 
 function RemoteAnonymousFunctionSyncCallDemo() {
   const handlePress = () => {
+    // eslint-disable-next-line no-constant-condition
     const foo = true ? () => {} : () => {};
```
Latropos, line 66: @tomekzaw I have no idea why do we need this constant condition...

apps/common-app/src/examples/SharedElementTransitions/ImageStack.tsx
```diff
@@ -57,7 +55,7 @@ export function ImageStack({
                 },
                 styles.roundedBorder,
               ]}
-              sharedTransitionTag={`${image}@${index}`}
+              sharedTransitionTag={`SET${index}`}
```
Latropos, line 58: `${image}` evaluates to string "[Object Object]", I've tested that the gallery example still works

piaskowyk (approved):
For me is ok, but let's wait for @tomekzaw review

apps/common-app/src/examples/WorkletRuntimeExample.tsx
```diff
@@ -74,7 +74,7 @@ function CreateWorkletRuntimeDemo() {
     const runtime = createWorkletRuntime('foo');
     console.log(runtime);
     console.log(runtime.name);
-    console.log(`${runtime}`);
+    console.log(`${runtime.name}`);
```
tomekzaw, line 77: This is the same as `console.log` above. The purpose of this `console.log` was to check the output of `${runtime}`. Can we restore the original `console.log`?

apps/common-app/src/examples/WorkletFactoryCrashExample.tsx
```diff
@@ -6,8 +6,9 @@ export default function App() {
   function handleOnPress() {
     function badWorklet() {
       'worklet';
-      // @ts-expect-error
-      unexistingVariable;
+      // @ts-expect-error This is an expected error
+      // eslint-disable-next-line no-unused-expressions
+      unExistingVariable;
```
tomekzaw, line 11:
```suggestion
      nonExistentVariable;
```

apps/common-app/src/examples/ShareablesExample.tsx
tomekzaw: Me neither but let's leave it as it is now

tomekzaw (approved):
LGTM

PR #6194, opened 2024-07-01
docs: Add OG images for Examples

patrycjakalinska:
Before OG images script only covered `/docs` folder, but after this change it also generates OG images for `/blog` folder which containes Examples.

image

## Test plan

In `/docs-reanimated` run:
```
yarn build && yarn serve

### DIFF
--- a/packages/docs-reanimated/scripts/build-og-images.jsx
+++ b/packages/docs-reanimated/scripts/build-og-images.jsx
@@ -6,7 +6,7 @@ import path from 'path';
 import fs from 'fs';
 import OGImageStream from './og-image-stream';
 
-const getMarkdownHeader = (path) => {
+const getDocsMarkdownHeader = (path) => {
   const content = fs.readFileSync(path, 'utf-8');
   const headers = content
     .split('\n')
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/docs-reanimated/scripts/build-og-images.jsx [hunk 1]
```diff
@@ -57,9 +69,25 @@ async function buildOGImages() {
...
-  dirs.map(async ({ dir, files }) => {
-    files.map(async (file) => {
-      const header = getMarkdownHeader(path.resolve(baseDirPath, dir, file));
+  const blogFiles = await fs.promises.readdir(blogDirPath);
+  const blogFilesFiltered = blogFiles
+    .filter((file) => file.endsWith('.md') || file.endsWith('.mdx'))
+    .map((file) => ({
+      dir: '',
+      files: [file],
+    }));
```
piaskowyk, line 78: Why we can obtains docs files in the same way as blog files? Is it possible to unify implementation?

piaskowyk (approved):
Well done!

PR #6198, opened 2024-07-02
Fix error persistancy in Performance Monitor

latekvo:
## UPDATE:

Replaced with this much simpler PR: [link](/software-mansion/react-native-reanimated/pull/6241)

#### Reasons:
- original mechanism could've been modified to achieve a very similar error-fixing effect
- dynamic array fill size is an overcomplicated for what it does.

### issue:
Issues which occur on CircularBuffer's total runtime accumulator, persist over the entire runtime
these issues are frequent, and occur due to dropped frames, which are almost guaranteed to happen to some degree.
After a couple minutes of runtime, effects were noticeable.
Previous implementation is also very prone to higher error rates at lower frame-rates.

### solution:
Implement Performance Monitor's frame storage in a self-regenerating way.
While I still use the circular buffer, the accumulator is saved and reset every revolution, it's state persists for one more revolution, but is completely wiped after that, thus errors don't persist for more than 2 revolutions of the frame buffer (by default: `66ms`).
  
---

Additionally, I added a dynamic update speed, which normalizes the time to complete the circular buffer revolution. By default, this time is equal to `0.33s`, but can be changed by adjusting newly added `expectedFps` and `smoothingCoefficient` props.

### DIFF
--- a/packages/react-native-reanimated/src/component/PerformanceMonitor.tsx
+++ b/packages/react-native-reanimated/src/component/PerformanceMonitor.tsx
@@ -9,25 +9,21 @@ import { createAnimatedComponent } from '../createAnimatedComponent';
 import { addWhitelistedNativeProps } from '../ConfigHelper';
 
 class CircularAccumulator {
-  private mainAccumulator: number = 0;
-  private memoryAccumulator: number = 0;
+  mainAccumulator: number = 0;
+  memoryAccumulator: number = 0;
 
-  private iterator: number = 0;
-  private circularArray: Float32Array;
+  iterator: number = 0;
+  circularArray: Float32Array | null = null;
+  length: number = 0;
 
   // fixme: may be negatively influenced by the first empty run
   // definietely a flawed system, remove or find a different error source
-  private errorRateAccumulator: number = 0;
+  errorRateAccumulator: number = 0;
 
-  length: number;
   previousTimestamp: number = 0;
 
-  constructor(length: number) {
-    this.length = length;
-    this.circularArray = new Float32Array(length);
-  }
-
-  private arrayEndHandler() {
+  arrayEndHandler() {
+    'worklet';
     this.errorRateAccumulator += this.memoryAccumulator;
     this.memoryAccumulator = this.mainAccumulator;
     this.mainAccumulator = 0;
@@ -35,8 +31,13 @@ class CircularAccumulator {
   }
 
   pushTimeDelta(timeDelta: number) {
+    'worklet';
     if (this.iterator === this.length) {
       this.arrayEndHandler();
+    }
+
+    if (this.circularArray === null) {
+      return;
     }
 
     this.mainAccumulator += timeDelta;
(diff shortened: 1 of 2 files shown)


packages/react-native-reanimated/src/component/PerformanceMonitor.tsx
```diff
@@ -4,56 +4,151 @@ import React, { useEffect, useRef } from 'react';
...
 import type { FrameInfo } from '../frameCallback';
-import type { SharedValue } from '../commonTypes';
 import { useSharedValue, useAnimatedProps, useFrameCallback } from '../hook';
 import { createAnimatedComponent } from '../createAnimatedComponent';
 import { addWhitelistedNativeProps } from '../ConfigHelper';
 
-type CircularBuffer = ReturnType<typeof createCircularDoublesBuffer>;
-function createCircularDoublesBuffer(size: number) {
+type CircularAccumulator = ReturnType<typeof contructCircularAccumulator>;
+const contructCircularAccumulator = (length: number, expectedFps: number) => {
```
szydlovsky, line 12: small typo 😄 
```suggestion
type CircularAccumulator = ReturnType<typeof constructCircularAccumulator>;
const constructCircularAccumulator = (length: number, expectedFps: number) => {
```

latekvo: thanks for the catch

packages/react-native-reanimated/src/component/PerformanceMonitor.tsx
```diff
@@ -180,11 +241,44 @@ function UiPerformance() {
...
+  expectedFps,
+  smoothingCoefficient,
+}: PerformanceMonitorProps) {
   return (
     <View style={styles.monitor}>
-      <JsPerformance />
-      <UiPerformance />
+      <JsPerformance
+        expectedFps={(expectedFps ?? DEFAULT_EXPECTED_FPS) / 2} // /2 due to 2x lower sampling rate on JS due to performance issues
+        smoothingCoefficient={smoothingCoefficient ?? DEFAULT_BUFFER_SIZE}
```
szydlovsky, line 276: I am not sure if we just don't want a default values for the props. Then we don't need the `PerformanceMonitorProps` type and we can describe props with a TSDoc comment like here: https://github.com/software-mansion/react-native-reanimated/blob/ac8b705d1bd1ed770679e8009144254fe34bd517/packages/react-native-reanimated/src/hook/useAnimatedScrollHandler.ts#L28-L37

szydlovsky (approved):
Everything looks fine, I have a more organisation-wise suggestions 😄

packages/react-native-reanimated/src/component/PerformanceMonitor.tsx
latekvo: Sure, I can do that.

Although `expectedFps` has a very specific use case - people testing fps on phones where they expect fps higher than 60, and in some cases higher than 120 might want to change this value to their expected fps amount, having this value either be too low or too high will only really affect the smoothing quality, which is not all that essential.

I'll keep this value at 60 by default, but will change smoothing rate to 30, and make both of these values constant by removing the props associated with them.

Changing smoothing rate to 30 will make it reasonably smooth on both 60fps `(0.5s smoothing look-back)` and 120fps `(0.25s smoothing look-back)` devices.

szydlovsky: Seems reasonable to me 😄

piaskowyk:
please wait a bit with merge, i want to test it on my machine

szydlovsky:
Ok, one more thing - take a look at linter. It suggests you have some unused types in the component file 😄

latekvo:
Closing, replaced with [link](/software-mansion/react-native-reanimated/pull/6241) PR.

PR #6199, opened 2024-07-02
Fix test of animations on Android

Latropos:
## Summary
This PR contains changes necessary to make tests of animations pass on Android. I also make function `getAnimatedStyle` return correct type of a prop (previously it was always returning a string)
## Test plan
Run tests of animations on Paper on both IOS and Android.

The only error is the following one on IOS:
image

image

### DIFF
--- a/apps/common-app/src/examples/RuntimeTests/tests/animations/withDelay/addDelays.test.tsx
+++ b/apps/common-app/src/examples/RuntimeTests/tests/animations/withDelay/addDelays.test.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from 'react';
-import { View, StyleSheet } from 'react-native';
+import { View, StyleSheet, Platform } from 'react-native';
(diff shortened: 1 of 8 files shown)


apps/common-app/src/examples/RuntimeTests/tests/animations/withDelay/addDelays.test.tsx
```diff
@@ -158,8 +161,10 @@ describe('Add three delays', () => {
       );
 
       expect([{ width: 0 }, { width: 0 }, ...updatesOneDelay]).toMatchSnapshots(updates);
-      expect(updatesOneDelay).toMatchNativeSnapshots(nativeUpdatesOneDelay);
-      expect(updates).toMatchNativeSnapshots(nativeUpdates);
+      if (componentType !== 'INLINE' || Platform.OS !== 'android') {
```
piaskowyk, line 164: There are quite a few workarounds in this PR. I understand that they fix some issues, but let me take some time to consider a potential solution to fix the root cause of the problem.

piaskowyk (changes requested):

PR #6201, opened 2024-07-02
Fix hermesCommand path for android release bundle

szydlovsky:
## Summary

Yeah so it turns out `hermesCommand` path was commented out (and wrong) in PaperExample `build.gradle`.
I've seen that this change was introduced while ensuring RN0.71 compatibility so I am a bit confused..

## Test plan

Launch android paper example with release bundle

### DIFF
--- a/apps/paper-example/android/app/build.gradle
+++ b/apps/paper-example/android/app/build.gradle
@@ -45,7 +45,7 @@ react {
 
     /* Hermes Commands */
     //   The hermes compiler command to run. By default it is 'hermesc'
-    hermesCommand = "../../node_modules/react-native/sdks/hermesc/osx-bin/hermesc"
+    hermesCommand = "../../node_modules/react-native/sdks/hermesc/%OS-BIN%/hermesc"
     //
     //   The list of flags to pass to the Hermes compiler. By default is "-O", "-output-source-map"
     // hermesFlags = ["-O", "-output-source-map"]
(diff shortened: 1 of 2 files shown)


tomekzaw (approved):
LGTM

tomekzaw (commented):
What about fabric-example?

szydlovsky:
@tomekzaw Fabric already has a correct path

apps/paper-example/android/app/build.gradle
```diff
@@ -45,7 +45,7 @@ react {
 
     /* Hermes Commands */
     //   The hermes compiler command to run. By default it is 'hermesc'
-    // hermesCommand = "$rootDir/my-custom-hermesc/bin/hermesc"
+    hermesCommand = "../../node_modules/react-native/sdks/hermesc/osx-bin/hermesc"
```
tjzel, line 48: Why is it needed? From my experiences it's only necessary when you are doing some "custom" operations with Hermes.

Also, hardcoding `osx-bin` isn't a good idea if some of our contributors wanted to build it on Linux/Windows. AFAIK there was some way to put here a platform-agnostic path.

tjzel (changes requested):

apps/paper-example/android/app/build.gradle
piaskowyk: +1

tomekzaw (approved):

piaskowyk (approved):

Issue #6119, opened 2024-06-13
Using `interpolateColor` inside `useDerivedValue` creates colors glitches on iOS


swansontec:
### Description

Returning an `interpolateColor` value from a `useDerivedValue` will cause incorrect colors to appear on iOS. Here is the setup:

```js
const blue = '#2080c0';
const derivedColor = useDerivedValue(() =>
  interpolateColor(0, [0, 1], [blue, blue]),
);
```

This *should* always render as blue if we put it in an animated style:

```js
const animatedStyle = useAnimatedStyle(() => ({
  color: derivedColor.value,
}));
```

However, on iOS specifically, the text will glitch to pink after the first button press:

```js
  // Cause a re-render on press:
  const [count, setCount] = React.useState(0);
  const handleCount = () => setCount(count + 1);

  return (
    <>
      <Animated.Text style={animatedStyle}>
        {count}
      </Animated.Text>
      <Button onPress={handleCount} title="Count" />
    </>
  );
```

### Steps to reproduce

Here is what seems to be happening:

1. First render
  a. JS thread: `useDerivedValue` returns the string `"rgba(32, 128, 192, 1)"`
  b. JS thread: `useAnimatedStyle` returns the style `{ color: "rgba(32, 128, 192, 1)" }`
  c. The text renders as blue
  d. UI thread: `useDerivedValue` runs again, and returns the *integer* 0xff2080c0. This seems to be some sort of optimization, since this differs from what the JS thread returns.
2. User presses the button
3. Second render
  a. JS thread: `useDerivedValue` does *not* run, but returns the saved 0xff2080c0 from the UI thread.
  b. JS thread: `useAnimatedStyle` returns the style `{ color: 0xff2080c0 }`
  c. The text renders as pink

Directly rendering `<Text style={{ color: 0xff2080c0 }}>Hello</Text>` produces the same pink color (although TypeScript complains that numbers aren't valid colors). Messing around with the values, it seems that `useAnimatedStyle` on the JS thread interprets integers as 0xrrggbbaa, whereas the Reanimated UI thread interprets integers as 0xaarrggbb.

### Snack or a link to a repository

https://github.com/swansontec/PinkEyeTest

### Reanimated version

3.12.0

### React Native version

0.74.2

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Any

### Device

Any

### Acknowledgements

Yes

[...]

PR #6213, opened 2024-07-04
Remove unnecessary color processing in `interpolateColor`

szydlovsky:
## Summary

Our `interpolateColor` function seems to have some unnecessary color processing in it. It was added some 3-4 years ago and should not be needed by now.

## Why

There is a function called `rgbaColor` which is used in `hsvToColor` (which in turn is used only in `interpolateColor`) as well as directly in `interpolateColor`. So it's safe to say it is used ***only*** there. All it does is take r,g,b,a values, runs a small check on them and then returns a rgba string.
It has a check whether we are on UI thread or not. If we were on UI, it returned a number-formatted color (formatted for native platforms) instead of the usual rgba string. The problem is, these values are then directly returned from the function. I believe it was needed back then, but now we have a function called `processColorInProps` which does exactly that + it doesn't interfere with returned values of `interpolateColor`. Taking all these things into consideration, I believe this check in `rgbaColor` function is not needed anymore.

Moreover, it creates problems. When `interpolateColor` gets called on UI thread (for example by using it inside `useDerivedValue` hook as shown by this issue: https://github.com/software-mansion/react-native-reanimated/issues/6119), the user gets these number-formatted colors which don't work when passed anywhere.

## Test plan

Copy `App.tsx` from snack linked to the mentioned issue and make sure the text stays blue. Also, check out Color Interpolation example from the example app.

Tested on iOS and Android (both Paper and Fabric) as well as Web. I've noticed neither issues nor regressions.

### DIFF
--- a/packages/react-native-reanimated/src/Colors.ts
+++ b/packages/react-native-reanimated/src/Colors.ts
@@ -9,7 +9,7 @@
 /* eslint no-bitwise: 0 */
 import type { StyleProps } from './commonTypes';
 import { makeShareable } from './core';
-import { isAndroid, isWeb } from './PlatformChecker';
+import { isAndroid } from './PlatformChecker';
 
 interface RGB {
   r: number;
--- a/packages/react-native-reanimated/src/Colors.ts
+++ b/packages/react-native-reanimated/src/Colors.ts
@@ -511,9 +511,6 @@ export const blue = (c: number): number => {
   return c & 255;
 };
 
-const IS_WEB = isWeb();
-const IS_ANDROID = isAndroid();
-
 export const rgbaColor = (
   r: number,
   g: number,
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):
LGTM but let's wait for @piaskowyk's approval

piaskowyk (approved):
According to my test and our runtimetests it works 🎉

jesse-one-app:
this change has caused an issue in our project when using `interpolateColor` inside of a `useAnimatedProps` hook with an `Animated.createAnimatedComponent` using a `react-native-svg` path. The color changes briefly then resets back to black. 

When i roll back to 3.13.0 that doesn't include this change the color works as expected. Can we revert this change or fix the issues it has with animated props?

tomekzaw:
It seems like this should be fixed with https://github.com/software-mansion/react-native-svg/pull/2397

bohdanprog:
Hello @jesse-one-app,
I tried to reproduce your issue but couldn't. Could you please check that example? It should work.

```
import React from 'react';
import {Svg, Circle} from 'react-native-svg';
import Animated, {
  createAnimatedPropAdapter,
  processColor,
  useAnimatedProps,
  useSharedValue,
  withRepeat,
  withTiming,
  interpolateColor,

  props => {
    if (Object.keys(props).includes('fill')) {
      props.fill = {type: 0, payload: processColor(props.fill)};
    }
  },
  ['fill'],
);

  const opacity = useSharedValue(0);

  React.useEffect(() => {
    opacity.value = withRepeat(withTiming(1), -1, true);
  }, []);

  const circleAnimatedProps = useAnimatedProps(
    () => {
      const coordinates = {cx: 50, cy: 50, r: 50};

      return {
        cx: coordinates.cx,
        cy: coordinates.cy,
        r: coordinates.r,
        fill: interpolateColor(opacity.value, [0, 1], ['red', 'green']),
        opacity: opacity.value,
        strokeWidth: 2,
      };
    },
    [],
    adapter,
  );

  return (
    <Svg width="100%" height="100%">
      <AnimatedCircle animatedProps={circleAnimatedProps} />
    </Svg>
  );
}

```

jesse-one-app:
@bohdanprog we were not calling `processColor` in the `createAnimatedPropAdapter` function

thanks for the example and the help!

PR #6214, opened 2024-07-04
Optimize Shadow Tree cloning

bartlomiejbloniarz:
## Summary

This PR optimizes the algorithm used to clone the Shadow Tree used in `ReanimatedCommitHook` and `performOperations`. 

## Current approach
The current algorithm works as follows. After receiving a batch of updates, we iterate over the list and apply the changes one by one. To apply changes we have to:
1. calculate the path from the affected ShadowNode to the root using `ShadowNodeFamily::getAncestors`
2. traverse the path upwards cloning all the nodes up to the root

This way we unfortunately clone some ShadowNodes multiple times. For example for a batch of size `n` we will clone the root node `n` times. 

Cloning ShadowNodes is expensive, so we had implemented an optimization - whenever a node is unsealed we would change it in place instead of cloning it. Unfortunately this didn't work, since the `getSealed` method always returns `true` in Production mode. This is not a bug, but the intended behavior, as sealing is only intended to help finding bugs in the Debug Mode. This still could be salvaged by memoizing which nodes were already cloned by us, but this approach still wouldn't be perfect, as modyfing nodes in place is still a heavy operation.

## New approach
To mitigate those issues we split the process into two phases:
1. calculate the subtree of the ShadowTree that contains all the nodes that we want to update
2. traverse the ShadowTree and clone nodes (that belong to the subtree) in the (reversed) topological order

By calculating the subtree first we ensure that in the second phase:
1. we traverse only nodes that absolutely have to be traversed
2. we clone only nodes that absolutely have to be cloned
3. we clone every node at most once

With this approach the second phase is performed in the optimal number of operations.

## Limitations
The current implementation of phase one (building the subtree) is not optimal. It is implemented by simply calling `getAncestors` on every node from the batch. This is fortunately not a huge problem, because cloning had a much heavier impact on the performance. To optimize this there will have to be some changes done in RN (because the `parent` field in `ShadowNodeFamily` is private, so traversing the tree upwards is only possible through `getAncestors`). I hope to soon open a suitable PR.

## Some examples
I checked the performance of our heavier examples on some devices in the Release Mode. For the `BokehExample.tsx` the results are:

| Phone  | Example size | Before [FPS] | After [FPS] | 
| -------- | ------- | -------- | ------- |
| iPhone 12 mini  | 200   | 30-40 | 60 |
| One+ A6 | 100   | 10-20 | 30-40 |
| iPhone 15 Pro  | 250 | 30-40 | 120 |
| Samsung Galaxy S23 | 100 | 55-70 | 120 |

I also tested through Xcode Instruments how much time does the `performOperations` function take on the same example. Tests were conducted on the iPhone simulator, but they should give an idea on the order of the number of operations this function makes (and how fast that number grows in relation to the example size).

| Example size | Before [ms] | After [ms] | Before/After |
| ------- | -------- | ------- | ------- |
| 1   | 1.95 | 2.1 | 0.92 |
| 20   | 2.4 | 2.1 | 1.14 |
| 100   | 5.3 | 2.3 | 2.3 |
| 250   | 22 | 4 | 5.5 |
| 500 | 77 | 7.7 | 10 |

## Test plan
Check the behavior of examples in the `FabricExample` app. Verify that heavy examples have improved, while simpler examples have not regressed.


packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -38,23 +40,21 @@ RootShadowNode::Unshared ReanimatedCommitHook::shadowTreeWillCommit(
   // ShadowTree not commited by Reanimated, apply updates from PropsRegistry
 
   auto rootNode = newRootShadowNode->ShadowNode::clone(ShadowNodeFragment{});
+  std::unordered_map<
+      const ShadowNodeFamily *,
+      std::vector<std::shared_ptr<RawProps>>>
```
tomekzaw, line 45: Can we use `std::unique_ptr` instead of `std::shared_ptr` here? Also, it's not recommended to allocate `RawProps` on heap and hold it for too long, can we avoid using pointers here at all?

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -38,23 +40,21 @@ RootShadowNode::Unshared ReanimatedCommitHook::shadowTreeWillCommit(
   // ShadowTree not commited by Reanimated, apply updates from PropsRegistry
 
   auto rootNode = newRootShadowNode->ShadowNode::clone(ShadowNodeFragment{});
+  std::unordered_map<
+      const ShadowNodeFamily *,
```
tomekzaw, line 44: It doesn't seem safe to store pointers to `ShadowNodeFamily`, is there any alternative? If that's not possible, can we at least hold references instead of pointers (using `std::reference_wrapper`)?

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -38,23 +40,21 @@ RootShadowNode::Unshared ReanimatedCommitHook::shadowTreeWillCommit(
...
-              cloneShadowTreeWithNewProps(rootNode, family, RawProps(props));
-
-          if (newRootNode == nullptr) {
-            // this happens when React removed the component but Reanimated
-            // still tries to animate it, let's skip update for this specific
-            // component
-            return;
-          }
-          rootNode = newRootNode;
+          propsMap[&family].push_back(std::make_shared<RawProps>(props));
```
tomekzaw, line 53: Let's use `emplace_back` instead if possible.

packages/react-native-reanimated/Common/cpp/Fabric/ShadowTreeCloner.cpp
```diff
@@ -6,60 +6,67 @@
...
+    std::unordered_map<const ShadowNodeFamily *, std::vector<int>> &childrenMap,
+    const ShadowNode::Shared &shadowNode,
+    std::unordered_map<
+        const ShadowNodeFamily *,
+        std::vector<std::shared_ptr<RawProps>>> &propsMap) {
+  auto family = &shadowNode->getFamily();
+  auto children = shadowNode->getChildren();
+  auto &affectedChildren = childrenMap[family];
+
+  for (auto index : affectedChildren) {
```
tomekzaw, line 19:
```suggestion
  for (const auto index : affectedChildren) {
```

packages/react-native-reanimated/Common/cpp/Fabric/ShadowTreeCloner.cpp
```diff
@@ -6,60 +6,67 @@
...
-  // adapted from ShadowNode::cloneTree
+ShadowNode::Unshared cloneShadowTreeWithNewPropsRecursive(
+    std::unordered_map<const ShadowNodeFamily *, std::vector<int>> &childrenMap,
+    const ShadowNode::Shared &shadowNode,
+    std::unordered_map<
+        const ShadowNodeFamily *,
+        std::vector<std::shared_ptr<RawProps>>> &propsMap) {
+  auto family = &shadowNode->getFamily();
+  auto children = shadowNode->getChildren();
+  auto &affectedChildren = childrenMap[family];
```
tomekzaw, line 17: Let's use `const` wherever possible.
```suggestion
  const auto family = &shadowNode->getFamily();
  const auto children = shadowNode->getChildren();
  const auto &affectedChildren = childrenMap[family];
```

packages/react-native-reanimated/Common/cpp/Fabric/ShadowTreeCloner.h
```diff
@@ -14,8 +15,9 @@ namespace reanimated {
 
 ShadowNode::Unshared cloneShadowTreeWithNewProps(
     const ShadowNode::Shared &oldRootNode,
-    const ShadowNodeFamily &family,
-    RawProps &&rawProps);
+    std::unordered_map<
```
tomekzaw, line 18: Let's define `std::unordered_map<const ShadowNodeFamily *, std::vector<std::shared_ptr<RawProps>>>` as some type using `using` keyword.

packages/react-native-reanimated/Common/cpp/Fabric/ShadowTreeCloner.cpp
```diff
@@ -6,60 +6,67 @@
...
+  std::unordered_map<const ShadowNodeFamily *, std::vector<int>> childrenMap;
+
+  for (auto &[family, _] : propsMap) {
+    auto ancestors = family->getAncestors(*oldRootNode);
+
+    for (auto it = ancestors.rbegin(); it != ancestors.rend(); ++it) {
+      auto &parentNode = it->first.get();
+      auto index = it->second;
+      auto parentFamily = &parentNode.getFamily();
+      auto &affectedChildren = childrenMap[parentFamily];
```
tomekzaw, line 58:
```suggestion
      const auto &parentNode = it->first.get();
      const auto index = it->second;
      const auto parentFamily = &parentNode.getFamily();
      const auto &affectedChildren = childrenMap[parentFamily];
```

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp
szydlovsky: out of curiosity, what's the difference?

tomekzaw: `emplace_back` uses move semantics

szydlovsky: ooh, didn't know 😄

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp
bartlomiejbloniarz: I only use `ShadowNodeFamily` whilst still holding onto corresponding `ShadowNodes`. Since `ShadowNode` owns its `family` through a `shared_ptr` it should be safe.

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp
bartlomiejbloniarz: In the `ReanimatedCommitHook` we have props stored as `folly::dynamic` and in `performOperations` we have them as `jsi::Value`. We have to turn them into `RawProps` at some point before cloning `ShadowNodes`. We could use some abstraction over `jsi::Value` and `folly::dynamic` but this is exactly what `RawProps` do. They need to be allocated on the heap, because we want to process the entire batch efficiently. Also, we only hold on to them for the duration of the `clone` function, we never persist them across multiple invocations.

ReactorCSA:
Thanks for this great library!!!
My question may be stupid but anyways, will this fix also improve performance on Paper architecture?
Thanks in advance :)

bartlomiejbloniarz:
Hi @ReactorCSA! 
This optimization is only for the New Architecture, since Shadow Tree is only a New Architecture concept.

janicduplessis:
@bartlomiejbloniarz I've tested this in my app and it causes a crash on screens that use `@gorhom/bottom-sheet`. Sorry I don't have an exact repro, but here's the crash report I get.

image

bartlomiejbloniarz:
@janicduplessis Thank you for testing this PR 🙏

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp
bartlomiejbloniarz: Nevermind

bartlomiejbloniarz:
@janicduplessis could you test your app again? I changed the implementation so that we don't allocate `RawProps` on the heap (sadly @tomekzaw was right). From my tests `@gorhom/bottom-sheet` doesn't crash anymore, but it would be cool to see if your app is fine now 🙏

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp
bartlomiejbloniarz: Never-nevermind

bartlomiejbloniarz:
The issue wasn't caused by the usage of `RawProps` on its own. It was actually due to using the same `RawProps` multiple times. This happened because the `ChildMap` was populated wrongly (i.e. my skill issue). I decided to remove the abstraction and go back to using `RawProps`.

packages/react-native-reanimated/Common/cpp/Fabric/ShadowTreeCloner.h
```diff
@@ -15,12 +16,18 @@ using namespace react;
 
 namespace reanimated {
 
-using PropsMap = std::unordered_map<const ShadowNodeFamily *, std::vector<std::unique_ptr<PropsWrapper>>>;
+template<class T>
+concept Derived = std::is_base_of_v<PropsWrapper, T>;
+
+template<Derived T>
+using PropsMap = std::unordered_map<const ShadowNodeFamily *, std::vector<std::unique_ptr<T>>>;
```
tomekzaw: Do we need `PropsMap` to be generic?

bartlomiejbloniarz: This is outdated 😅

mrousavy:
A bit unrelated, but looking at this code I can see that you can access the props of a component as `jsi::Value`s - how did you get to that point?

I'm trying to build a simple component (iOS/Android) where the props layer is implemented in C++/JSI, and I manually pass down the props to ObjC/Java myself - but I couldn't figure out how to get access to raw props as `jsi::Value`s...

janicduplessis:
I’ll test it again in the next few days

janicduplessis:
@mrousavy I think here the props as jsi::Value is new props to update the components with, not the current props of the component.

bartlomiejbloniarz:
@mrousavy As @janicduplessis mentioned, the props here are the new props. They come from the [reanimated updateProps function](https://github.com/software-mansion/react-native-reanimated/blob/243be4aede2ac26afb355dd100beb2cd3701f203/packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp#L619)

tjzel (approved):
Looks good! 🚢 🇮🇹

packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -736,22 +736,9 @@ void NativeReanimatedModule::performOperations() {
...
-              // specific component
-              continue;
-            }
-            rootNode = newRootNode;
           }
-
-          auto newRoot = std::static_pointer_cast<RootShadowNode>(rootNode);
-
-          return newRoot;
+          return std::static_pointer_cast<RootShadowNode>(
```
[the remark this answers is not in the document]
bartlomiejbloniarz: I refactored the logic, so that now the `cloneShadowTreeWithNewProps` returns a pointer to `RootShadowNode`

tjzel (approved):

piaskowyk (approved):

zibs:
Hey @bartlomiejbloniarz!

I just ran a git bisect between 3.14.0 and 3.15.0 where this commit is the one that causes an issue in my app: `a4c5c5d39ac27c5d04f825af30f172ab8087844d is the first bad commit`.

My issue occurs on bridgeless enabled + fabric

I'm attaching two videos of the issue, that seems to be related to custom scroll bars and Safe Areas, although it doesn't matter which version the `react-native-safe-area-context` is.


You can see that in the first video, `3.14.0`, the custom scroll bar/safe areas work fine.

https://github.com/user-attachments/assets/4fc8d8b6-f6c9-4da2-97fd-a3be6f28550b

You can see that in the first video, `3.15.0` (`a4c5c5d39ac27c5d04f825af30f172ab8087844d is the first bad commit`), the custom scroll bar/safe areas doesn't work.

https://github.com/user-attachments/assets/15954cb4-71ff-42ae-b05c-fc09993dc42f

Here is the most relevant code:

```tsx
    const onScrollHandler = useAnimatedScrollHandler(event => {
      if (onScroll) {
        runOnJS(onScroll)({nativeEvent: event});
      }
      let newOffset =
        ((event.contentOffset.y + largeTitleOffsetForScrollIndicator) /
          completeScrollbarHeight) *
        (difference + scrollIndicatorSize);

      if (newOffset < 0) {
        newOffset = 0;
      }

      if (newOffset > difference) {
        newOffset = difference;
      }

      scrollIndicatorOffset.value = newOffset;
    });

    const scrollStyle = useAnimatedStyle(() => {
      return {
        height: scrollIndicatorSize - headerOffsetForScrollIndicator,
        transform: [
          {
            translateY: scrollIndicatorOffset.value,
          },
        ],
      };
    });
```

And I can say that if I remove the `translateY: scrollIndicatorOffset.value,`, it "works" in the sense that the safe area doesn't break, but it does break my custom scrolling implementation.

Do you have any thoughts on what might be happening here? It would be very much appreciated, as this is one the last remaining things for me to enable bridgeless/fabric.

I will also say that this example is running `"react-native": "0.74.5",`, but issue is the same on `75` too.

bartlomiejbloniarz:
Hi @zibs. I think that this issue is related to #6659. I need some more time to address it properly.

PR #6216, opened 2024-07-05
Introduce Import buttons to run tests more easily

Latropos:
## Summary

image

## Android Header Fix
This PR should fix problem with additional value added to "top" property on Android. It fixes test cases similar to:  withSequence -> Cascade of callbacks -> Test that all callbacks have been called a correct number of times, this test case includes snapshot including top property.

## Test plan
I've run all tests on paper, on both IOS and Anroid and didn't notice any unexpected failures.
Several failures of layout transition tests were expected, as their timing was updated after test creation.


apps/common-app/src/examples/RuntimeTests/RuntimeTestsExample.tsx
```diff
@@ -1,30 +1,55 @@
...
+        {
+          testSuiteName: 'Tests of testing framework',
+          importTest: () => {
+            require('./tests/TestsOfTestingFramework.test');
+          },
+        },
+        {
+          testSuiteName: 'animations',
+          importTest: () => {
+            require('./tests/animations');
```
piaskowyk, line 17: I'm considering if it would be a better idea to move `requires` from `./tests/animations/index.ts` here to keep everything in one place.

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsRunner.tsx
```diff
@@ -21,30 +26,61 @@ export class ErrorBoundary extends React.Component<
...
+    <View>
+      {importButtons.map(({ testSuiteName, importTest }) => {
+        return (
+          <TouchableOpacity
+            key={testSuiteName}
+            onPress={() => {
+              importTest();
+              if (!importedTests.includes(testSuiteName)) {
+                setImportedTests([...importedTests, testSuiteName]);
+              }
```
piaskowyk, line 49: let's move it to `handleClick` function to avoid unnecessary nesting in code

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/RuntimeTestsRunner.tsx
```diff
@@ -21,30 +26,61 @@ export class ErrorBoundary extends React.Component<
...
-      return <Text>Something went wrong.</Text>;
+      return <Text>Unable to render component</Text>;
     }
 
     return this.props.children;
   }
 }
 
-export default function RuntimeTestsRunner() {
+function ImportButtons({ importButtons }: { importButtons: Array<ImportButton> }) {
```
piaskowyk, line 36: Is there a way to run all tests at once instead of selecting them one by one?

Latropos: I've added an additional button to run all the tests (except for tests of testing framework)

piaskowyk (approved):

Issue #6203, opened 2024-07-03
Animations in recycled components seem to be broken from v3.9.0 and up


LcTwisk:
### Description

The changes in https://github.com/software-mansion/react-native-reanimated/pull/5268 result in animations not properly resetting when a component is being recycled in e.g. a FlashList (even if the `SharedValue`'s backing the animation are reset.

Was able to work around it by explicitly setting a property that would change when recycling. That results in the `_requiresNewInitials`[here](https://github.com/software-mansion/react-native-reanimated/blob/4115e83875bcd70becd5e02f770e4bd21c24294a/packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx#L106-L109) to become true (because of a change in props). However, I'm not sure if this is the intended solution for recycling. Any thoughts?
```diff
      <Animated.View
+       key={index}
        style={animatedStyle}
        height={height}
        onLayout={onLayout}
      />
```

#### A video showing the issue in our app:
The following can be observed: an animation is triggered on component 1, but the state is not being reset when the component is recycled on index 11 (there's a recycle pool of 10 components).
The backing `SharedValue`'s are being reset when the component is recycled:
```ts
if (id !== previousId.current) {
  stepperOpacity.value = 0;
  stepperTranslateX.value = 0;
  previousId.current = id;
}
```

https://github.com/software-mansion/react-native-reanimated/assets/3274157/f74fb7a2-fe73-47b5-9eab-65248320a8ed

### Steps to reproduce

1. Create a large list with uniform components that will be recycled using e.g. a FlashList
2. Animate a state change in the first item of the list
3. Scroll down and see the animated style change appear on the recycled component (while the actual state change is reset)

### Snack or a link to a repository

Will create a reproduction project later

### Reanimated version

3.9.0 (and up)

### React Native version

0.72.14

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Acknowledgements

Yes

erquhart:
I'm also seeing this - specifically, `onScroll` from a handler created with `useAnimatedScrollHandler` is no longer called on scroll in a `FlashList` animated via reanimated `createAnimatedComponent`. Happy to try early releases to test any fix candidates if it's helpful. Building with Expo for iOS.

LeslieOA:
[...]

Issue #6102, opened 2024-06-11
glitches and crashes after migrating to reanimated >= 3.9


AleCatSS:
### Description

Hello Reanimated team!

In our React Native project, we use Reanimated for different animations on multiple screens. Recently, we attempted to update to version 3.11.0 to support React Native 0.74 (as per the [compatibility table](https://docs.swmansion.com/react-native-reanimated/docs/guides/compatibility/)). After this update, we started experiencing crashes and glitches on animations across the project. 

During our investigation, we localized the issue starting from Reanimated version 3.9 by gradually upgrading Reanimated alone, version by version, and keeping all other packages unaltered (including React Native).

From our investigation, it seems crashes and glitches are more prominent in debug mode and on low-end devices (we tested with a real iPhone SE second generation). We were not able to reproduce crashes in release mode, but we were able to see the same glitches in animations (although less frequently).

NOTE: in our project we also use react-native-gesture-handler v2.16.2

### Steps to reproduce

Unfortunately we currently don't have specific steps to reproduce, however I suggest:

- create a project with React Native version >= 0.72.12 (our first tested version) and re-animated version >= 3.9;
- add few relatively heavy animations;
- build on iOS/Android in debug mode (also release mode but in debug issues are more prominent);
- test with low-end devices (e.g. iPhone SE second generation both real device or simulator)

### Snack or a link to a repository

/

### Reanimated version

3.11.0 (issue starts with 3.9.0)

### React Native version

0.72.12

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

Real device

### Device model

iPhone SE 2nd generation

### Acknowledgements

Yes

crrobinson14:
We can confirm here that the same change was very helpful in our app. (We downgraded to 3.7.2.) A few screen transitions are still slower for us on Android than iOS (using react-native-navigation, which uses this library for its transitions) but before we were getting complete hangs/crashes and <=3.7.2 resolves that for us. We haven't had a chance yet to try making an isolated repro but wanted to at least add the data point.

AleCatSS:
After more investigation we isolated the issue to in this PR merged in 3.9:
-  https://github.com/software-mansion/react-native-reanimated/pull/5268 
and in particular on this file: 
[...]

Issue #6037, opened 2024-05-20
Flickering in animation after updating react native


ArsenSheripov:
### Description

Hello everyone, I had a problem after updating react native to version 0.74.1; after the update, the header animation began to flicker heavily when activated. 

### Steps to reproduce

```
const Component = ({ }) => {
 const [readerMode, setReaderMode]=useState({mode:'off'})
  const marginTop = useSharedValue(0);
  const animatedStyles = useAnimatedStyle(() => ({
    marginTop: withTiming(marginTop.value, { duration: 500 }),
  }));

  return (
      <View style={{flex: 1,}>
        <Animated.View style={[styles.safeBlock, safeBlockAnimatedStyles]} /> */}
        <Animated.View
          style={[{ height: 100, backgroundColor: 'red' }, animatedStyles]}
          onLayout={onLayout}
          {...props}
        />

        <ScrollView>
          <View style={{ height: 400, backgroundColor: 'blue' }} />
          <View style={{ height: 400, backgroundColor: 'green' }} />
          <View style={{ height: 400, backgroundColor: 'yellow' }} />
          <View style={{ height: 400, backgroundColor: 'teal' }} />
        </ScrollView>
        <Button
          text="On"
          onPress={() => setReaderMode({ mode: 'on' })}
        />
        <Button
          text="Off"
          onPress={() => setReaderMode({mode: 'off' })}
        />
      </View>
  );
};

```
### Snack or a link to a repository
https://snack.expo.dev/@arsen.sheripov/upbeat-blue-waffle

### Reanimated version

3.11.0

### React Native version

0.74.1

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Release app & production bundle

### Device

iOS simulator

### Acknowledgements

Yes

pavelbabenko:
Same issue with animating `height` in production

exploIF:
@ArsenSheripov Can I ask you to provide working repro? Unfortunately your example is not working, also there is nothing in the code that triggers animation. Hard to help without that

ArsenSheripov:
@exploIF https://snack.expo.dev/kLsBpRavc-_8FcsALDcbE?platform=android
ofcource, here I corrected the code

exploIF:
@ArsenSheripov I've checked your repro on both Android and iOS and everything works fine on my end

ArsenSheripov:
@exploIF 
[...]

PR #5268, opened 2023-10-18
Fix dynamically added/removed animated styles on iOS/Android

tjzel:
## Summary

Fixes #5296.

1. Since this is potentially a JS-heavy fix, I'll gladly listen to suggestions and receive help with how to profile it.

2. I personally dislike how those changes look like. I would much more prefer to simplify some logic present in `createAnimatedComponent` and its devil spawn, reducing some code, instead of building on top of it - although I'm not sure if we need it at the moment.

Just watch the video to understand what is happening:

<table>
<tr><td>BEFORE (buggy)</td><td>AFTER (desired)</td></tr>
<tr><td>

https://github.com/software-mansion/react-native-reanimated/assets/40713406/07495711-a909-4499-89b8-a0be17f3d8ef

</td><td>

https://github.com/software-mansion/react-native-reanimated/assets/40713406/47cbace1-bf98-4011-8b4c-79bc6f6a80d4

</td></tr>
</table>

## Test plan

💣 

## Note

JS-Reanimated (Web implementation) still has those issues and I will produce a separate solution for it in a finite amount of time.

PR #6218, opened 2024-07-05
Revert dynamic styles PR 

szydlovsky:
## Summary

Unfortunately, turns out the PR with dynamic styles #5268 introduced more problems than it solved. Thus, proposing a revert.
This revert most likely fixed:
- https://github.com/software-mansion/react-native-reanimated/issues/6203
- https://github.com/software-mansion/react-native-reanimated/issues/6102
- https://github.com/software-mansion/react-native-reanimated/issues/6037

as well as one more not-published issue and potentially a few more.

## Notes

Even if it gets approved, to be merged ONLY after @tjzel agrees with everything

## Test plan

😢 😭


apps/common-app/src/examples/index.ts
```diff
@@ -491,11 +490,6 @@ export const EXAMPLES: Record<string, Example> = {
     title: 'Log test',
     screen: LogExample,
   },
-  DynamicStylesExample: {
```
tjzel, line 494: 🧨

packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx
```diff
@@ -24,30 +23,23 @@ function dummyListener() {
...
         const styleProp = inputProps.style;
         const styles = flattenArray<StyleProps>(styleProp ?? []);
-        if (this._requiresNewInitials) {
-          this._initialStyle = {};
-        }
         const processedStyle: StyleProps = styles.map((style) => {
           if (style && style.viewDescriptors) {
             // this is how we recognize styles returned by useAnimatedStyle
-            if (this._requiresNewInitials) {
+            style.viewsRef?.add(component);
```
tjzel, line 41: viewRefs no longer exist after #6123

packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx
```diff
@@ -70,6 +62,7 @@ export class PropsFilter implements IPropsFilter {
           Object.keys(animatedProp.initial.value).forEach((initialValueKey) => {
             props[initialValueKey] =
               animatedProp.initial?.value[initialValueKey];
+            animatedProp.viewsRef?.add(component);
```
tjzel, line 65: viewRefs no longer exist after #6123

tjzel (changes requested):
This is a sad day in the history of Reanimated - with this change we are going one step further from mimicking React with our API 😭.

We should add relevant information to our documentation about the fact the user can't change animated styles dynamically.

tjzel (approved):

PR #6226, opened 2024-07-08
Speed up precommit

tjzel:
I measured which bits of precommit take so long:

Screenshot 2024-07-08 at 11 04 17

I decided to simply call it conditionally.


Latropos (approved):
LGTM 👍

PR #6227, opened 2024-07-08
Extend plugin's documentation

tjzel:
Since the number of features in Reanimated Babel Plugin grows it would be decent to have them nicely documented.

As a bonus plugin becomes a separate workspace - it was overlooked during the monorepo migration.


packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
```diff
@@ -0,0 +1,217 @@
...
+    }
+  : () => {
+      'worklet'; // Won't work without it.
+      return { width: 200 };
+    };
+
+const style = useAnimatedStyle(foo);
+```
+
+In such cases we recommend you to handle the conditional logic in the worklet itself or to refactor your code to avoid the need for conditional worklets.
```
patrycjakalinska, line 192: just loose suggestion, something felt off here with 'to refactor your code to ...'

```suggestion
In such cases we recommend either handling the conditional logic in the worklet itself or refactoring your code to eliminate the need for conditional worklets.
```

patrycjakalinska (approved):
inspiring ✨

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
tjzel: I wanted to use active voice as our guru @kacperkapusciak has taught us.

patrycjakalinska: i'm still learning this mystic knowledge, sorry :c

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
```diff
@@ -0,0 +1,217 @@
...
+  'worklet';
+  console.log('Hello from worklet');
+}
+```
+
+- A function that is _autoworkletizable_, i.e.:
+
+```ts
+useAnimatedStyle(() => {
+  // This function will be ran on the Worklet Runtime,
```
kacperkapusciak, line 32:
```suggestion
  // This function will be ran on the UI thread,
```
i know from the internals point of view a worklet runtime would be ackchyually more correct

but we never use this nomenclature in the documentation. the devs have a mental model of 2 threads - JS and UI.

I feel like saying JS runtime, Worklet runtime is kinda exposing exposing the internals of Reanimated

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
```diff
@@ -0,0 +1,217 @@
...
+};
+```
+
+## Autoworkletization
+
+To reduce boilerplate code and relieve the user from knowing the internals of Reanimated, Reanimated Babel Plugin tries to detect automatically whether a function should be workletized. For example, all callbacks in Reanimated API that are ran on the Worklet runtime are autoworkletized, such as:
+
+- `useAnimatedStyle`
+- `useAnimatedProps`
+- `runOnUI`
```
kacperkapusciak, line 89: Can we have a full list here? Or a link to full list in the plugin's source code?

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
```diff
@@ -0,0 +1,217 @@
...
+  foo() {
+    'worklet';
+    console.log('Hello from ObjectMethod');
+  },
+};
+```
+
+## Autoworkletization
+
+To reduce boilerplate code and relieve the user from knowing the internals of Reanimated, Reanimated Babel Plugin tries to detect automatically whether a function should be workletized. For example, all callbacks in Reanimated API that are ran on the Worklet runtime are autoworkletized, such as:
```
kacperkapusciak, line 85: i don't understand this paragraph

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
```diff
@@ -0,0 +1,217 @@
...
+function foo() {
+  // This function will be autoworkletized.
+  return { width: 100 };
+}
+
+function bar() {
+  // This function will be autoworkletized.
+  function foobar() {
+    // This function won't since it's not defined in top-level scope.
+    console.log('Hello from worklet');
```
kacperkapusciak, line 142: the comments in this snippet and the console log are confusing

normally when we comment functions we put a comment before the name
```js
// this function does something
function superFun() {
``` 

and the console log says `"Hello from worklet"` where the comment says the function won't get workletized

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
```diff
@@ -0,0 +1,217 @@
...
+    };
+
+const style = useAnimatedStyle(foo);
+```
+
+In such cases we recommend you to handle the conditional logic in the worklet itself or to refactor your code to avoid the need for conditional worklets.
+
+## Pitfalls
+
+There are some patterns that won't work with the plugin, regardless of whether they are autoworkletizable or not.
```
kacperkapusciak, line 196:
```suggestion
There are some patterns that won't work with the plugin.
```
the second part of the sentence is a bit of a mouthful

kacperkapusciak (changes requested):
Nice! Finally a place where its all collected and described 🎖️ 

Left a couple of comments but overall I'm very happy about it!

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
tjzel: On second thought it makes sense to stay with _UI thread_ here. 👍

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
tjzel: Some of the comments I put where the user would expect the `'worklet'` directive to be to get the "wow effect" 😄 However it might make more sense to put all such comments before the functions. What do you think?

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
tjzel: @kacperkapusciak I rewrote it. Do you think I should drop the "motives" altogether?

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
tjzel: I changed the ambiguous fragments so now they explicitly say what they refer to.

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
```diff
@@ -0,0 +1,239 @@
...
+const style = useAnimatedStyle(() => {
+  // You don't need to add the 'worklet' directive here,
+  // since plugin detects this callback as autoworkletizable.
+  return {
+    width: 100,
+  };
+});
+```
+
+This isn't limited to `useAnimatedStyle` hook - Reanimated Babel Plugin autoworkletizes all callbacks for the API of Reanimated. The whole list can be found in the [plugin source code](https://github.com/software-mansion/react-native-reanimated/blob/main/plugin/src/autoworkletization.ts).
```
szydlovsky, line 93: This link doesn't work   🦅 ❗❗ 
```suggestion
This isn't limited to `useAnimatedStyle` hook - Reanimated Babel Plugin autoworkletizes all callbacks for the API of Reanimated. The whole list can be found in the [plugin source code](https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/plugin/src/autoworkletization.ts).
```

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
```diff
@@ -0,0 +1,239 @@
...
+  'worklet'; // Won't work without it.
+  return {
+    width: 100,
+  };
+}
+
+// bar.ts
+import { foo } from './foo';
+const style = useAnimatedStyle(foo);
+```
```
szydlovsky, line 173: I'd swap places of these two file snippets. I think it is more clear if we firstly define the context (importing function to use it in useAnimatedStyle) and then show a snippet (similar to many others before), but with a forced 'worklet' directive

szydlovsky (changes requested):

packages/docs-reanimated/docs/reanimated-babel-plugin/about.md
```diff
@@ -0,0 +1,217 @@
...
+    console.log('Scrolling...');
+  },
+};
+
+const handler = useAnimatedScrollHandler(handlerObject);
+```
+
+### Workletizing whole files (experimental)
+
+You can mark a file as a workletizable file by adding the `'worklet'` directive to the top of the file. This will make all _top-level_ entities workletized. This can come in handy for files that contain multiple worklets.
```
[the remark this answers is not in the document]
jakub-gonet: Maybe just stick with functions for now and when we add support for objects and classes we can just list them there (or just go with "objects, classes, functions")?
In Erlang world the common name for that is "term" but dunno if it'd help in JS world.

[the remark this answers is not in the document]
tjzel: The idea with listing seems good to me.

szydlovsky (approved):

PR #6230, opened 2024-07-08
Workletizable Classes

tjzel:
## Summary

This pull request features running classes on the Worklet runtime. Classes aren't supported by Hermes and require polyfilling.

- Class instances can be created in worklets.
- Class instances cannot be sent between runtimes - afaik it's impossible to create host objects with a prototype.
- Static class methods aren't supported - a feature for the future perhaps.

Current implementation is a bit naive.
1. We create worklets from all required polyfills.
1. We create __the whole class__ (not the class instance!) in the worklet.
1. We create the instance of this ad-hoc made class.

We could could put the class in the global scope, however it doesn't seem exactly necessary at the moment.

You can obtain a worklet class explicitly by adding a ClassProperty `__workletClass` to it

```ts
// Explicit WorkletClass
class Clazz {
  __workletClass = true;
}
```

or implicitly by placing it in a file with a worklet directive.

```ts
'worklet';
// Implicit WorkletClass
class Clazz {}
```

## Test plan

- [x] Add unit tests
- [x] Add runtime tests suite


packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+
+  let factory: VariableDeclaration;
+
+  let hasHandledClass = false;
+
+  traverse(ast, {
+    [WorkletizableFunction]: {
+      // @ts-expect-error TS has some trouble inferring here.
+      enter: (functionPath: NodePath<WorkletizableFunction>) => {
+        if (functionPath.parentPath.isObjectProperty()) {
```
bartlomiejbloniarz, line 102: Why do we skip `functionPaths`?

packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+          return;
+        }
+
+        const workletDirective = directive(directiveLiteral('worklet'));
+
+        if (functionPath.parentPath.isCallExpression() && !hasHandledClass) {
```
bartlomiejbloniarz, line 108: please revise those if statements

packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+            ),
+          ]);
+          hasHandledClass = true;
+
+          return;
+        }
+
+        const bodyPath = functionPath.get('body');
+        if (!bodyPath.isBlockStatement()) {
+          bodyPath.replaceWith(
```
bartlomiejbloniarz, line 128: this can be extracted to a utility function

packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+
+        (functionPath.node.body as BlockStatement).directives.push(
+          workletDirective
+        );
+      },
+    },
+  });
+
+  const body = ast.program.body;
+  body.push(factory!);
```
bartlomiejbloniarz, line 141: add assert here

packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+          identifier(className + 'ClassFactory')
+        ),
+        identifier(className + 'ClassFactory')
+      )
+    )
+  );
+
+  sortPolyfills(ast);
+
+  const transformedNewCode = transformSync(generate(ast).code, {
```
bartlomiejbloniarz, line 158: this name could be more descriptive

packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+    ast: true,
```
bartlomiejbloniarz, line 159: Is the `ast` here necessary if we already have it?

packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+    filename: state.file.opts.filename,
+  });
+
+  assert(transformedNewCode);
+  assert(transformedNewCode.ast);
+
+  const parent = classPath.parent as Program;
```
bartlomiejbloniarz, line 166: Add better assertions here

packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+function getPolyfillsToSort(ast: BabelFile): SortElement[] {
+  const polyfills: SortElement[] = [];
+
+  traverse(ast, {
+    Program: {
+      enter: (functionPath: NodePath<Program>) => {
+        const statements = functionPath.get('body');
+        statements.forEach((statement, index) => {
+          const bindingIdentifiers = statement.getBindingIdentifiers();
+          if (!statement.isFunctionDeclaration() || !statement.node.id?.name) {
```
bartlomiejbloniarz, line 262: Add a comment here about polyfills being FunctionDeclarations

packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+            index,
+            dependencies: new Set(),
+          };
+          polyfills.push(element);
+          statement.traverse({
+            Identifier(path: NodePath<Identifier>) {
+              if (
+                !path.isReferencedIdentifier() ||
+                path.node.name in bindingIdentifiers ||
+                statement.scope.hasOwnBinding(path.node.name) ||
```
bartlomiejbloniarz, line 277: Explain these conditions better

packages/react-native-reanimated/plugin/src/file.ts
```diff
@@ -73,15 +71,16 @@ function getCandidate(statementPath: NodePath<Statement>) {
...
+    return statementPath.get('declaration') as NodePath<unknown>;
   } else {
     return statementPath;
   }
 }
 
-function processWorkletizableEntity(nodePath: NodePath<BabelNode>) {
+function processWorkletizableEntity(
+  nodePath: NodePath<unknown>,
+  state: ReanimatedPluginPass
```
bartlomiejbloniarz, line 82: Is the state necessary anymore?

packages/react-native-reanimated/plugin/src/jestMatchers.ts
```diff
@@ -70,11 +70,12 @@ expect.extend({
 
   toIncludeInWorkletString(received: string, expected: string) {
     // Regular expression pattern to match the code field
-    const pattern = /code: "((?:[^"\\]|\\.)*)"/s;
-    const match = received.match(pattern);
+    // @ts-ignore This regex works well in Jest.
+    const pattern = /code: "((?:[^"\\]|\\.)*)"/gs;
+    const matches = received.match(pattern);
 
     // If a match was found and the match group 1 (content within quotes) includes the expected string
```
bartlomiejbloniarz, line 77: Update comment here

bartlomiejbloniarz (changes requested):

packages/react-native-reanimated/plugin/src/class.ts
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  if (!hasWorkletClassMarker(classPath.node.body)) {
+    return false;
+  }
+
+  const parentPath = classPath.parentPath;
+  const className = classPath.node.id.name;
+
+  classPath.skip();
```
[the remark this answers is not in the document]
tjzel: It's no longer necessary.

packages/react-native-reanimated/plugin/src/class.ts
tjzel: I refactored the whole logic there.

packages/react-native-reanimated/plugin/src/class.ts
tjzel: I refactored the whole logic there.

packages/react-native-reanimated/plugin/src/class.ts
tjzel: I refactored the whole logic there.

packages/react-native-reanimated/plugin/src/class.ts
tjzel: Renamed.

packages/react-native-reanimated/plugin/src/class.ts
tjzel: It wasn't needed anymore.

packages/react-native-reanimated/plugin/src/class.ts
tjzel: I refactored the whole logic there.

packages/react-native-reanimated/plugin/src/class.ts
tjzel: Done.

packages/react-native-reanimated/plugin/src/class.ts
tjzel: I put the logic in the comments.

packages/react-native-reanimated/plugin/src/file.ts
tjzel: Removed.

packages/react-native-reanimated/plugin/src/class.ts
tjzel: I don't need it anymore.

packages/react-native-reanimated/plugin/src/jestMatchers.ts
tjzel: Done.

PR #6232, opened 2024-07-08
[Web LA] Add `EntryExitTransition`

m-bert:
## Summary

This PR adds `EntryExitTransition` to set of transitions available on `web`. 

This transition is different from others, therefore logic behind its creation is also different.

### Two animations

First approach was to run 2 animations - `exiting` and `entering`. While it seemed like an easier approach, there were few problems:

- It lead to many code duplications (especially when starting animation, handling `end`/`cancel`)
- I couldn't find a way to keep elements in good positions (`transform` was overriden by animations and we need to use it to put element back in its starting position) 

While I haven't observed any flickers caused by having 2 animations, arguments above made me try seconds approach)

### Dynamic keyframe

Second approach to this problem involves creating dynamic keyframe animation - just like in case of other transitions. First we obtain data of both `exiting` and `entering` animations. Then we perform two operations:

1. We add calculated transform into `exiting` animation (`unshift` is used to make sure that it will be placed at the beginning of keyframe timestamps
2. We merge two animations object into one keyframe data object. 

In 2. we have to make sure that both animations will fit in one keyframe, therefore `exiting` animation is squeezed into interval `[0,49]` and `entering` animation runs from `50` to `100`.

### Examples

https://github.com/software-mansion/react-native-reanimated/assets/63123542/6ee7fcfe-4ebd-41cf-bc33-9ab7d262b20b

https://github.com/software-mansion/react-native-reanimated/assets/63123542/a9d15f48-c158-461b-8343-33c6d20ae5bd

## Test plan

Verified on `DefaultLayoutTransitions` example.


packages/react-native-reanimated/src/layoutReanimation/web/transition/EntryExit.web.ts
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  const keyframeData: AnimationData = {
+    name,
+    style:
+      '0' in exitingAnimationData.style
+        ? {}
+        : { 0: { transform: [additionalData] } },
+    duration: 300,
+  };
```
piaskowyk, line 29: Let's think about whether we still need this

packages/react-native-reanimated/src/layoutReanimation/web/transition/EntryExit.web.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  for (const [timestamp, styles] of Object.entries(
+    exitingAnimationData.style
+  )) {
+    if (styles.transform !== undefined) {
+      styles.transform.unshift(additionalData);
+    } else {
+      styles.transform = [additionalData];
+    }
```
piaskowyk, line 38: Let's add comment or move it to another function

packages/react-native-reanimated/src/layoutReanimation/web/transition/EntryExit.web.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const newTimestamp = parseInt(timestamp) / 2;
+
+    keyframeData.style[`${newTimestamp === 50 ? 49 : newTimestamp}`] = styles;
```
piaskowyk, line 42: `Math.min()`

m-bert: Done in [d571951](https://github.com/software-mansion/react-native-reanimated/pull/6232/commits/d571951851d72110a4c51a7b3f44dfc393240fe0)

packages/react-native-reanimated/src/layoutReanimation/web/transition/EntryExit.web.ts
m-bert: Removed in [25fbb28](https://github.com/software-mansion/react-native-reanimated/pull/6232/commits/25fbb283ac85afae124c08a801f92e8c9482823d) since it is no longer required

packages/react-native-reanimated/src/layoutReanimation/web/transition/EntryExit.web.ts
m-bert: Done in [026c08f](https://github.com/software-mansion/react-native-reanimated/pull/6232/commits/026c08f5e422fd8a8443d20bed2d4bb8f3b2391c)

piaskowyk (approved):

PR #6233, opened 2024-07-08
[reJest] Fixes problem with opacity comparation

piaskowyk:
## Summary

Due to some platform limitations, certain properties may not be exact values as originally set, for example, opacity. This PR introduces new comparators, `CLOSE_DISTANCE` and `FLOAT`, which can be used in cases where we can't achieve the exact value as originally set.


piaskowyk:
I named it `CLOSE_DISTANCE` to maintain naming conventions, as the `DISTANCE` mode already exists.

PR #6236, opened 2024-07-08
[reJest] waitForAnimationUpdates

piaskowyk:
## Summary

This pull request adds a new method to the runtime test API - `waitForAnimationUpdates`. The method accepts the expected snapshot of the animation and waits until it receives the exact amount of JS updates defined in the snapshot.


apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
```diff
@@ -564,6 +569,25 @@ export class TestRunner {
...
+        await this.runOnUIBlocking(() => {
+          'worklet';
+          assertMockedAnimationTimestamp(global.framesCount);
+          flag.value = global.framesCount >= updatesCount - 1;
+        });
+        if (flag.value) {
+          clearInterval(interval);
+          resolve(true);
+        }
+      }, 20);
```
Latropos, line 587: What do you think about extracting it into a separate variable?

Latropos (commented):
Looks good!
However I suppose that `waitForAnimationUpdates(snapshot.length)` allows more flexibility.

For example we sometimes want to check that concatenation of snapshots of `Animation_A` and `Animation_B` matches the snapshot of `withSequence(Animation_A, Animation_B)`.

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
```diff
@@ -564,6 +569,25 @@ export class TestRunner {
...
+  public waitForAnimationUpdates(snapshot: Array<any>): Promise<boolean> {
+    const updatesCount = snapshot.length;
+    const flag = makeMutable(false);
+    return new Promise<boolean>(resolve => {
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
+      const interval = setInterval(async () => {
+        await this.runOnUIBlocking(() => {
+          'worklet';
+          assertMockedAnimationTimestamp(global.framesCount);
+          flag.value = global.framesCount >= updatesCount - 1;
```
Latropos, line 581: Do we need updatesCount at all? Perhaps it would be enough to wait until the number of recorded frames doesn't increse during 1000/FPS ms? 🤔

piaskowyk: As we discussed offline, we have decided to abandon this approach

Latropos (approved):

PR #6243, opened 2024-07-09
Assess and remove unnecessary usage of `Float32Array`

latekvo:
## Summary

This PR removes a 4-year-old TODO in the `Bezier` library, keeping `Arrays` and removing mentions of `Float32Arrays`.

This is due to `Float32Arrays` having a very specific `WebGL` use case, and generally being slower than regular `Arrays`, especially when they're dynamically re-allocated over and over again as it is the case in the `Bezier` library.

Live speed demo: [link](https://jsperf.app/float32array-slowness-test/2)
Live demo results:
image

## Test plan

not-applicable

## Notes

Using `Float32Array` here has no effect on precision, and no measurable speed improvement/degradation was noted on the production-like-conditions tests that I did.

### DIFF
--- a/packages/react-native-reanimated/src/Bezier.ts
+++ b/packages/react-native-reanimated/src/Bezier.ts
@@ -90,7 +90,7 @@ export function Bezier(
   mX2: number,
   mY2: number
 ): (x: number) => number {
-  'worklet';
+  ('worklet');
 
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):
🫡

PR #6258, opened 2024-07-11
Exract async code to separate classes

Latropos:
## Summary
Previously to use function `runOnUIBlocking` we need a reference to TestRunner. 
I wanted to use `runOnUIBlocking` while implementing some other feature 
(`expect(()=>{}).toThrow()`) and found it a bit troublesome. Therefore I've decided to extract this code into separate class first.

### DIFF
--- a/apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
+++ b/apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
@@ -80,7 +80,7 @@ export class TestRunner {
     return this._renderLock;
   }
 
-  public render(component: ReactElement<Component> | null) {
+  public async render(component: ReactElement<Component> | null) {
     if (!component && this._wasRenderedNull) {
       return;
     }
--- a/apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
+++ b/apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
@@ -2,7 +2,6 @@ import type { Component, MutableRefObject, ReactElement } from 'react';
 import { useRef } from 'react';
 import type {
   BuildFunction,
-  LockObject,
   NullableTestValue,
   Operation,
   SharedValueSnapshot,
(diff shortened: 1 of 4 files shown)


apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
```diff
@@ -36,15 +37,42 @@ function notifyJS(name: string) {
   notificationRegistry[name] = true;
 }
 
+export class RenderLock {
+  private _renderLock: LockObject = { lock: false };
+
+  public lockRender() {
```
piaskowyk, line 43: we don't need to duplicate a name
```suggestion
  public lock() {
```

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
```diff
@@ -36,15 +37,42 @@ function notifyJS(name: string) {
...
+    this._renderLock = { lock: true };
+  }
+
+  public unlockRender() {
```
piaskowyk, line 47:
```suggestion
  public unlock() {
```

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
```diff
@@ -36,15 +37,42 @@ function notifyJS(name: string) {
...
+    this._renderLock = { lock: false };
+  }
+
+  public waitForUnlock(maxWaitTime?: number) {
```
piaskowyk, line 51: is it the same implementation as `waitForThreadUnlock` from `SyncUIRunner`? Can we avoid code duplication? For example you can use inheritance

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/TestRunner.ts
```diff
@@ -83,18 +107,19 @@ export class TestRunner {
     return this._renderLock;
   }
 
-  public async render(component: ReactElement<Component> | null) {
+  public render(component: ReactElement<Component> | null) {
```
piaskowyk, line 110: Why now we can remove `async` form here?

apps/common-app/src/examples/RuntimeTests/ReanimatedRuntimeTestsRunner/UpdatesContainer.ts
```diff
@@ -18,7 +18,7 @@ type NativeUpdate = {
   jsUpdateIndex: number;
 };
 
-export function createUpdatesContainer(testRunner: TestRunner) {
+export function createUpdatesContainer() {
```
piaskowyk, line 21: that is a good change! 👍

piaskowyk (approved):

PR #6259, opened 2024-07-11
[Web LA] Fill in missing keyframes in default animations

m-bert:
## Summary

This PR adds missing `keyframe` steps definitions in default layout animations. While they are not necessary for animations to work, I've found some problems with it while implementing `withInitialValues` modifier.

It turns out if you do not specify a property in initial `keyframe` step and then you decide to change it in next `keyframe` steps, it may not work.

> [!NOTE]
> Example below was written in pure `HTML` and `CSS` on [W3Schools Tryit Editor](https://www.w3schools.com/cssref/tryit.php?filename=trycss3_keyframes). However, since this PR is about web layout animations, it applies to them in the same manner as to animations in `HTML` and `CSS`.

### Initial state
Consider this example:

```CSS
#box {
  width: 100px;
  height: 100px;
  background: red;
  animation: myAnimation 2s;
}

@keyframes myAnimation {
  100% {
    opacity: 0;
    transform: rotate(5rad) scale(0);
   }
}
```
In this case, `div` with id `box` will perform `PinwheelOut` animation:

https://github.com/software-mansion/react-native-reanimated/assets/63123542/0e7f1b71-658b-45cf-b3e9-0737c9066728

### Adding initial values

Now, suppose that we want to use `withInitialValues` modifier and add `scale: 2`. This will result in the following `keyframe`:

```CSS
@keyframes myAnimation {
  0% {
    transform: scale(2);
  }
  100% {
    opacity: 0;
    transform: rotate(5rad) scale(0);
   }
}
```
Now the animation looks like this:

https://github.com/software-mansion/react-native-reanimated/assets/63123542/772ef0e8-f829-40c6-87a9-5a1558da855b

This animation is broken for two reasons:

1. There is no rotation applied to the component
2. There is a moment when component simply disappears, instead of smoothly changing `opacity` from `1` to `0`

### Filling remaining properties

Now, if we fill in remaining `keyframe` properties, it will look like this:

```CSS
@keyframes myAnimation {
  0% {
    opacity: 1;
    transform: rotate(0rad) scale(2);
  }
  100% {
    opacity: 0;
    transform: rotate(5rad) scale(0);
   }
}
```

And the resulting animation looks like this:

https://github.com/software-mansion/react-native-reanimated/assets/63123542/97778a16-9117-4723-8e90-42cca5b0d06d

### Order of properties

> [!CAUTION]
> Order of properties between `keyframe` steps does matter!

It is obvious that order of operations inside `transform` matters. However, it is also important to keep order of those operations between `keyframes`. In the example above, if we switch `rotation` with `scale` like this:

```CSS
@keyframes myAnimation {
  0% {
    opacity: 1;
    transform: scale(2) rotate(0rad);
  }
  100% {
    opacity: 0;
    transform: rotate(5rad) scale(0);
   }
}
```
The animation will look like this:

https://github.com/software-mansion/react-native-reanimated/assets/63123542/216d679b-3e3d-4fab-af07-869ad65984f4

So even though all properties were provided, we still do not get proper animation.

## Potential issues

Problem described in the _**Order of properties**_ paragraph may lead to some problems, since layout animations on web are created dynamically from objects that contain `keyframe` description. While mixing order of `keyframe` steps doesn't change anything (you can for example specify `to` before `from` and it will work correctly), order inside transform does matter. This may be problematic because we are iterating over objects using `Object.entries` method. However, up to this point, no issues were reported.

## Test plan

Tested on **_Default Layout Animations_** example.

### DIFF
--- a/packages/react-native-reanimated/src/layoutReanimation/web/animation/Zoom.web.ts
+++ b/packages/react-native-reanimated/src/layoutReanimation/web/animation/Zoom.web.ts
@@ -81,6 +81,7 @@ export const ZoomOutData = {
   ZoomOut: {
     name: 'ZoomOut',
     style: {
+      0: { transform: [{ scale: 1 }] },
       100: { transform: [{ scale: 0 }] },
     },
     duration: DEFAULT_ZOOM_TIME,
@@ -89,6 +90,7 @@ export const ZoomOutData = {
   ZoomOutRotate: {
     name: 'ZoomOutRotate',
     style: {
+      0: { transform: [{ scale: 1, rotate: '0rad' }] },
       100: { transform: [{ scale: 0, rotate: '0.3rad' }] },
     },
     duration: DEFAULT_ZOOM_TIME,
@@ -97,6 +99,7 @@ export const ZoomOutData = {
   ZoomOutRight: {
     name: 'ZoomOutRight',
     style: {
+      0: { transform: [{ translateX: '0vw', scale: 1 }] },
       100: { transform: [{ translateX: '100vw', scale: 0 }] },
     },
     duration: DEFAULT_ZOOM_TIME,
@@ -105,7 +108,8 @@ export const ZoomOutData = {
   ZoomOutLeft: {
     name: 'ZoomOutLeft',
     style: {
-      100: { transform: [{ translateX: '-100vw', scale: 1 }] },
+      0: { transform: [{ translateX: '0vw', scale: 1 }] },
+      100: { transform: [{ translateX: '-100vw', scale: 0 }] },
     },
     duration: DEFAULT_ZOOM_TIME,
   },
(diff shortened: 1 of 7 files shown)


bartlomiejbloniarz (approved):

packages/react-native-reanimated/src/layoutReanimation/web/animation/Zoom.web.ts
```diff
@@ -105,14 +108,16 @@ export const ZoomOutData = {
   ZoomOutLeft: {
     name: 'ZoomOutLeft',
     style: {
-      100: { transform: [{ translateX: '-100vw', scale: 1 }] },
+      0: { transform: [{ translateX: '0vw', scale: 1 }] },
+      100: { transform: [{ translateX: '-100vw', scale: 0 }] },
```
m-bert, line 112: Here I changed `scale` to `0` - probably we have overlooked that during first review of LA 😅

piaskowyk (commented):
luks 👍

PR #6266, opened 2024-07-15
[Web] Fix missing RNRenderer export (#6265)

### LINKED ISSUE #6265: [Vite] [Web] No matching export in RNRenderer.web.js for import "RNRenderer"
### Description

As of 3.13 I believe this error starts to occur in Vite:

```
 ERROR  Build failed with 1 error:                                                            9:11:57 AM
../../node_modules/react-native-reanimated/lib/module/createAnimatedComponent/createAnimatedComponent.js:12:9: ERROR: No matching export in "../../node_modules/react-native-reanimated/lib/module/platform-specific/RNRenderer.web.js" for import "RNRenderer"
```

### Steps to reproduce

Sorry to not have a great repro, but if you create a Vite app and import the latest Reanimated you will see it! Only posting here without full repro because the actual fix is pretty simple!

The file `react-native-reanimated/lib/module/platform-specific/RNRenderer.web.js` doesn't export anything:

```
'use strict';

// RNRender is not used for web. An export is still defined to eliminate warnings from bundlers such as esbuild.
export {};
//# sourceMappingURL=RNRenderer.web.js.map
```

Which cases Vite to break. Instead, maybe we can export `null`?

https://github.com/universal-future/vxrn

### Reanimated version

3.14.0

### React Native version

0.73.1

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

lkinasiewicz:
## Summary

Fixes #6265 

## Test plan

Applications using reanimated should build properly using bundlers that use ESM - e.g. Vite.

### DIFF
--- a/packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
+++ b/packages/react-native-reanimated/src/platform-specific/RNRenderer.web.ts
@@ -1,3 +1,4 @@
 'use strict';
 // RNRender is not used for web. An export is still defined to eliminate warnings from bundlers such as esbuild.
-export {};
+const RNRenderer = {};
+export { RNRenderer };


piaskowyk (approved):
Thanks for the fix! 🎉

PR #6284, opened 2024-07-17
Workletizable Context Objects

tjzel:
This pull request introduces the concept of _Context Objects_ on the Worklet Runtime.

## What?

Context Object is a shareable object that doesn't lose its `this` binding when it goes to the Worklet Runtime. To define an object as a Context Object, the user has to define property `__workletContextObject` on it. The value of this property doesn't matter.

```ts
// Explicit Context Object
const obj = {
  foo() { return 1; },
  bar() { return this.foo(); },
  __workletContextObject: true,
}
```

Context Objects are also auto-detected when Babel processes files marked with a worklet directive. If a top level object has a method that references `this` inside we treat is as a Context Object.

```ts
'worklet';
// Implicit Context Object
const obj = {
  foo() { return 1; },
  bar() { return this.foo(); },
};
```

## Why?

Currently it's not possible to refer to an object's methods from its other methods.

```ts
const obj = {
  foo() {
    console.log("foo");
  }
  bar() {
    this.foo();
  }
}

runOnUI(() => {
  obj.bar(); // Crash, the binding is lost on serialization.
})();
```

## How?

Context Objects are handled as another type of shareable entity on the React Runtime.
1. The plugin adds a special property `__workletObjectFactory` to Context Objects. It's a function that creates an identical object.
2. We serialize the factory instead of the object itself.
3. We recreate the object on the Worklet Runtime via ShareableHandle mechanism.

## Test plan

- [x] Add unit tests
- [x] Add a runtime test suite

### DIFF
--- a/apps/common-app/src/examples/RuntimeTests/tests/plugin/contextObjects.test.tsx
+++ b/apps/common-app/src/examples/RuntimeTests/tests/plugin/contextObjects.test.tsx
@@ -22,7 +22,7 @@ describe('Test context objects', () => {
         foo() {
           return 1;
         },
-        __workletObject: true,
+        __workletContextObject: true,
       };
 
       useEffect(() => {
@@ -45,7 +45,7 @@ describe('Test context objects', () => {
       registerValue(SHARED_VALUE_REF, output);
       const contextObject = {
         foo: () => 1,
-        __workletObject: true,
+        __workletContextObject: true,
       };
 
       useEffect(() => {
@@ -73,7 +73,7 @@ describe('Test context objects', () => {
         bar() {
           return this.foo() + 1;
         },
-        __workletObject: true,
+        __workletContextObject: true,
       };
 
       useEffect(() => {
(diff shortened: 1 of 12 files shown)


apps/common-app/src/examples/RuntimeTests/tests/plugin/contextObjects.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+const SHARED_VALUE_REF = 'SHARED_VALUE_REF';
+
+describe('Test context objects', () => {
+  test('methods are workletized', async () => {
+    const ExampleComponent = () => {
+      const output = useSharedValue<number | null>(null);
+      registerValue(SHARED_VALUE_REF, output);
+      const contextObject = {
+        foo() {
```
bartlomiejbloniarz, line 22: Please use `this` in this example

bartlomiejbloniarz (approved):

PR #6287, opened 2024-07-17
Use seed in randomised tests

Latropos:
## Summary
Unfortunately it is impossible to seed Math.random function, so this code includes custom implementation of pseudorandom number generator.

https://stackoverflow.com/questions/521295/seeding-the-random-number-generator-in-javascript/47593316#47593316


packages/react-native-reanimated/__tests__/bezier.test.ts
```diff
@@ -8,6 +8,23 @@ import { Bezier } from '../src/Bezier';
  */
 // spell-checker:enable
 
+function splitMix32(seed: number) {
```
piaskowyk, line 11: If it is copied from somewhere, let's add a link to the source of the implementation.

piaskowyk (approved):

PR #6288, opened 2024-07-17
Fix Reduced Motion example

piaskowyk:
## Summary

Previously, animations were defined in the global scope, so if they were run once, they couldn't be run again. I simply replaced static animations with a lambda that creates a new animation every time.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/21de2620-24f2-4d27-8e82-3217ddeab07e" /> | <video src="https://github.com/user-attachments/assets/ac178ce6-c6d5-4cc0-9029-43341b20b278" /> |

## Test plan

Run `ReducedMotion` example


szydlovsky (approved):
Looks and works about right 👍

PR #6289, opened 2024-07-18
Remove header to fix broken tests on android

Latropos:
## Summary
The height of information "Reload the app to run the tests again" was causing broken tests on android. This height was included in results of `getViewProp("top")` including snapshots**BEFORE:**
image
**AFTER:**
image


tjzel (changes requested):
Please do one of the following:

1. Restore the text with `position:absolute` so it doesn't affect layout.

2. Make the text appear after the tests are finished - probably with some callback that will notify the component.

tjzel (approved):

PR #6291, opened 2024-07-18
Prefer `waitForAnimationUpdates` over `wait`

Latropos:
## Summary
Prefer waitForAnimationUpdates over wait to make the animation faster and more reliable


piaskowyk (approved):
👍

PR #6292, opened 2024-07-18
Rename ReJest main folder

Latropos:
## Summary

Rename `ReanimatedRuntimeTestsRunner` into `ReJest`, because RuntimeTestsRunner is not the only class exported from that folder, it contains all the functionalities of reJest micro-framework.


apps/paper-example/ios/Podfile.lock
```diff
@@ -1578,4 +1578,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 746b518f69244f6a3f6b04dabf476339073cb5fe
 
-COCOAPODS: 1.14.3
+COCOAPODS: 1.14.2
```
piaskowyk, line 1581: Just restore this change and we can merge 👍

piaskowyk (changes requested):

piaskowyk (approved):

PR #6293, opened 2024-07-18
Introduce file structure to tests

Latropos:
## Summary
The new file structure is:

image


piaskowyk (approved):
That's generally okay, but using the name `utils` is not the best practice. Unfortunately, I don't have a better suggestion so we can use it 👍

PR #6295, opened 2024-07-19
Unify snapshots comparison

Latropos:
## Summary
Unify code for snapshot comparison on Android and IOS.
Also a lot of outdated code was cleaned out, as we have decided not to return snapshots of multiple views, but always for a certain predefined one.


piaskowyk (approved):
👏

PR #6304, opened 2024-07-22
[Web LA] Fix `easing`

m-bert:
## Summary

After #6163 was merged, each `worklet` function follows new naming convention. Since `easings` are `worklets`, their `name` property has also been changed. In web layout animations we were applying `easing` based on its name, therefore changing it resulted in only default `linear` easing being available on web.

This PR adds new `Symbol` into easing functions so that we can once again get their name and apply correct easing to animations.

## Test plan

Tested on _**BasicLayoutAnimation**_ example with `Easing.exp`

### DIFF
--- a/packages/react-native-reanimated/src/layoutReanimation/web/componentUtils.ts
+++ b/packages/react-native-reanimated/src/layoutReanimation/web/componentUtils.ts
@@ -25,7 +25,7 @@ import { EasingNameSymbol } from '../../Easing';
 
 function getEasingFromConfig(config: CustomConfig): string {
   if (!config.easingV) {
-    return `cubic-bezier(${WebEasings['linear'].toString()})`;
+    return `cubic-bezier(${WebEasings.linear.toString()})`;
   }
 
   const easingName = config.easingV[EasingNameSymbol];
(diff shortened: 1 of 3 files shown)


tjzel (approved):
This pattern is still far from perfect but 🚀

packages/react-native-reanimated/src/layoutReanimation/web/componentUtils.ts
```diff
@@ -21,14 +21,26 @@ import type { ReanimatedSnapshot, ScrollOffsets } from './componentStyle';
...
+  }
+
+  const easingName = config.easingV[EasingNameSymbol];
+
+  if (!(easingName in WebEasings)) {
+    console.warn(
+      `[Reanimated] Selected easing is not currently supported on web.`
+    );
+
+    return `cubic-bezier(${WebEasings['linear'].toString()})`;
```
tjzel: extract this to a separate string since its repeated

m-bert: It will be extracted into separate function in #6239.

m-bert: I really wanted to merge `CurvedTransition` last week, but it would be useless without this PR so I've decided to postpone it.

PR #6305, opened 2024-07-22
Add runtime tests for animations in StrictMode

piaskowyk:
## Summary

This PR adds test for animation and layout animation for StrictMode.

I also needed to fix the type of  the `beforeEach` method,


apps/common-app/src/examples/RuntimeTests/ReJest/types.ts
```diff
@@ -53,15 +53,17 @@ export type TestCase = {
   | { decorator: Exclude<TestDecorator, TestDecorator.WARN | TestDecorator.FAILING> | null }
 );
 
+export type MaybeAsyncVoidFunction = () => void | Promise<void>;
```
Latropos, line 56: This type is already declared in this file in line 39: 
```
export type BuildFunction = () => void | Promise<void>;
```
Feel free to rename it if you think it doesn't match the new usages. If it is possible try finding something shorter, maybe `MaybeAsyncVoid` since it is easy to guess it must be a function.

apps/common-app/src/examples/RuntimeTests/RuntimeTestsExample.tsx
```diff
@@ -75,6 +75,12 @@ export default function RuntimeTestsExample() {
             require('./tests/plugin/contextObjects.test');
           },
         },
+        {
+          testSuiteName: 'StrictMode',
+          importTest: () => {
+            require('./tests/StrictMode/StrictMode.test');
+          },
+        },
```
Latropos, line 83: One of the main goals of splitting the tests into sections was to avoid tests running for too long, like more than 15 minutes as the simulator or computer could start sleep mode, which have negative impact on tests. Could we have this sections more uniform when considering duration? For example could you merge StrictMode with some other section and rename it into sth like "other"? I am open for discussion on how we want to manage the scope and number of sections.

Latropos (changes requested):

Latropos (approved):

PR #6307, opened 2024-07-22
refactor: split makeMutable to web and native implementations

tjzel:
## Summary

Currently Web and Native implementations of `makeMutable` are bundled together and are hard to read. I split them for more clarity in the code, and also to prepare `makeMutable` for its further changes.

## Test plan

- [x] Runtime tests pass.


tomekzaw (approved):
LGTM. We could rename `makeMutableUI` to `makeMutableNativeUI` but the current name is fine as well.

tjzel:
> LGTM. We could rename `makeMutableUI` to `makeMutableNativeUI` but the current name is fine as well.

This would suggest that there would be `makeMutableWebUI` which isn't going to be the case soon.

PR #6310, opened 2024-07-22
Add mutable value read/write during render warning

MatiPl01:
## Summary

This PR adds warnings shown when the SharedValue `.value` property is accessed (read or written) while the component is being rendered.
The initial render is ignored as all reanimated hooks are executed for the first time during the initial render (e.g. `useAnimatedStyle` builds the initial component style).

## Test plan

- open the **Invalid read/write during render** example in the example app,
- press on the **Re-render** button to trigger re-render, which will access `.value` property of the shared value from the component function body,
- see warnings about invalid read and invalid write

### DIFF
--- a/packages/react-native-reanimated/src/mutables.ts
+++ b/packages/react-native-reanimated/src/mutables.ts
@@ -15,7 +15,7 @@ function shouldWarnAboutAccessDuringRender() {
 }
 
 function checkInvalidReadDuringRender() {
-  if (shouldWarnAboutAccessDuringRender()) {
+  if (__DEV__ && shouldWarnAboutAccessDuringRender()) {
     console.warn(
       '[Reanimated] Reading from `value` during component render. Please ensure that you do not access the `value` property while React is rendering a component.'
     );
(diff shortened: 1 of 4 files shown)


packages/react-native-reanimated/src/reactUtils.ts
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+function getCurrentReactOwner() {
+  // @ts-expect-error React secret internals aren't typed
+  return React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
```
MatiPl01, line 5: I hope I won't get fired because of using this secret variable

tomekzaw: We'll find out 😈

tomekzaw (approved):
LGTM

packages/react-native-reanimated/src/mutables.ts
```diff
@@ -69,12 +90,14 @@ function makeMutableNative<Value>(initial: Value): Mutable<Value> {
 
   const mutable: Mutable<Value> = {
     get value(): Value {
+      checkInvalidReadDuringRender();
```
piaskowyk, line 93: Do we want to check it always or only in the Development mode?

MatiPl01: Yea, makes sense to limit only to development. I will change that.

piaskowyk (approved):

PR #6311, opened 2024-07-22
refactor: Less error prone `_value` in Mutables.

tjzel:
## Summary

At the moment the `_value` property is an ordinary property, which makes it visible via `Object.keys` or `const prop in sv`. This way sometimes when it's accidentally passed to React it throws errors and we had to add troubleshooting for that.

This PR makes it so it would be hidden via `enumerable: false`.

## Test plan

- [x] Runtime tests pass.


packages/react-native-reanimated/src/mutables.ts
```diff
@@ -10,6 +10,25 @@ const SHOULD_BE_USE_WEB = shouldBeUseWeb();
...
+ *
+ * This way when the user accidentally sends the SharedValue to React, he won't get an obscure
+ * error message.
+ *
+ * We hide for both _React runtime_ and _Worklet runtime_ mutables for uniformity of behavior.
+ */
+function hideInternalValueProp(mutable: Mutable) {
+  'worklet';
+  Object.defineProperty(mutable, '_value', {
+    configurable: false,
```
Latropos, line 27:
```suggestion
    writable: true,
    configurable: false, // the user is not allowed to change properties of sharedValue._value, the user must reassign it.
```

Latropos: WDYT about adding some description and explicitly marking this as writable?

Latropos (approved):
Great improvement!

packages/react-native-reanimated/src/mutables.ts
tjzel: Relevant comment is already present in the function's TSDoc and in the other comments in the file.

Writable will be true by default since we are modifying an already created 'standard' object, I don't think re-setting it here has some real value.

PR #6316, opened 2024-07-23
feat: Move commonjs exports down for worklet files

tjzel:
## Summary

Since Reanimated Babel Plugin breaks hoisting and [newer Typescript hoists CommonJS exports by default](https://github.com/microsoft/TypeScript/pull/57669) we sometimes need to dehoist those exports.

## Test plan

- [x] Add unit test suite.
- [x] Runtime tests pass.


tomekzaw (approved):

tjzel (commented):
👍

PR #6320, opened 2024-07-23
Remove worklets

tjzel:


mrousavy:
👀

packages/react-native-worklets/src/publicGlobals.ts
```diff
@@ -0,0 +1,24 @@
...
+   * shouldn't depend on it, we keep it here
+   * mainly for backward compatibility for our users.
+   */
+  var _WORKLET: boolean | undefined;
+
+  /**
+   * This ArrayBuffer contains the memory address of `jsi::Runtime`
+   * which is the Reanimated UI runtime.
+   */
+  var _WORKLET_RUNTIME: ArrayBuffer;
```
mrousavy, line 23: Why not `bigint`, which is uintptr_t in C++?

tjzel: AFAIK there was no support for bigints back in the day so this was the way 😅

tjzel:
It was replaced by many other PRs

Issue #6245, opened 2024-07-09
Race condition in NativeReanimatedModule::performOperations()


yard:
#### Background
We have been debugging a weird issue where _sometimes_ the app would incorrectly account for keyboard state. To spare you from all the steps on that journey, we were able to confirm that the styles computed by `useAnimatedStyle` hook will not apply (although debug logs confirmed they did re-compute and did that correctly). After carefully studying the issue I think we have found the root cause for that behavior.

#### Issue description
[performOperations()](https://github.com/software-mansion/react-native-reanimated/blob/77644ca17d54b20a5547e7a1be8ec279cd2b5ded/packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp#L637C30-L637C47) method of `NativeReanimatedModule` can submit updated props to the registry while React is committing a new shadow tree. Specifically, it can happen so that [ReanimatedCommitHook](https://github.com/software-mansion/react-native-reanimated/blob/77644ca17d54b20a5547e7a1be8ec279cd2b5ded/packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp#L24) has already gathered the props from the registry and returned the updated tree to React, enabled `shouldReanimatedSkipCommit_` flag and React is now reconciling the tree; at this very moment updates applied to props registry will effectively be lost: `performOperations()` would [bail out](https://github.com/software-mansion/react-native-reanimated/blob/77644ca17d54b20a5547e7a1be8ec279cd2b5ded/packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp#L703) (as the flag is set). Until next re-render happens, the UI would be presented with the old props.

Surely, this is a hard thing to reproduce – sorry for not providing a snack/repo, I am trying to think of a way of how to demonstrate that without too many external dependencies, but I hope that explanation is clear enough to at least confirm the issue is present and valid.

#### Proposed fix
We have applied a patch that adds another flag to props registry, which gets set whenever any prop update happens during mount phase. This way, [ReanimatedMountHook](https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedMountHook.cpp) can check if anything was changed during mount phase and trigger a re-render. I am not exactly sure if that's the best way to fix the problem, but it does seem to work and does so reliably and without causing excessive renders.

[...]

PR #6330, opened 2024-07-25
Modify reanimated props updates

bartlomiejbloniarz:
## Summary

This PR changes the way `performOperations` and `ReanimatedCommitHook` are synchronized. The current implementations faces 2 problems:

1. Updates that come through `performOperations` after `pleaseSkipReanimatedCommit` is called in the commit hook will be deferred until the next commit. Usually it's fine since the next commit will be triggered by the next animation frame. But if there was a singular update scheduled through Reanimated, we might not see the change for a long time. This issue is more thoroughly explained in [this issue](https://github.com/software-mansion/react-native-reanimated/issues/6245).

2. In `ReanimatedCommitMarker` there is an assumption that there can be at most one commit happening on a given thread (i.e. there can't be nested commits). This isn't true, since there can be an event listener that commits some changes in a reaction to a native mount/unmount (which is a part of the commit function). [This exact scenario was observed in the New Expensify App with `react-native-keyboard-controller`](https://github.com/Expensify/App/issues/44437#issuecomment-2250696773). At first I thought this is a mistake, but [this PR in RN](https://github.com/facebook/react-native/commit/5f0435fb852cbb8abfe6cab77784ebddf1ad38ac) seems to allow for scenarios like that.

Applied fixes:

ad 1. Now instead of resetting the skip flag in reanimated after a transaction is mounted (via MountHook), we reset the flag whenever a non-empty batch is read in `performOperations`. This ensures that we don't make an unnecessary commit, but never skip any updates.

ad 2. Now we mark reanimated commits through `ReanimatedCommitShadowNode`. This is a class derived from ShadowNode that allows us to modify the root nodes traits_ (and add our custom trait). This ensures that even if the root node gets cloned it will retain the information. We couldn't derive from `RootShadowNode` since it is a final class.


packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedMountHook.cpp
```diff
@@ -18,13 +17,7 @@ ReanimatedMountHook::~ReanimatedMountHook() noexcept {
...
 void ReanimatedMountHook::shadowTreeDidMount(
     RootShadowNode::Shared const &,
-    double) noexcept {
-  // When commit from React Native has finished, we reset the skip commit flag
-  // in order to allow Reanimated to commit its tree
-  if (!ReanimatedCommitMarker::isReanimatedCommit()) {
-    propsRegistry_->resetReanimatedSkipCommitFlag();
-  }
-}
+    double) noexcept {}
```
tomekzaw, line 20: Do we still need this mount hook?

packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -736,7 +736,19 @@ void NativeReanimatedModule::performOperations() {
...
           }
-          return cloneShadowTreeWithNewProps(oldRootShadowNode, propsMap);
+
+          auto rootNode =
+              cloneShadowTreeWithNewProps(oldRootShadowNode, propsMap);
+
+          // Mark the commit as Reanimated commit so that we can distinguish it
+          // in ReanimatedCommitHook.
+
+          auto reaShadowNode =
```
tomekzaw, line 746: Let's avoid using "rea" if needed.
```suggestion
          auto markedShadowNode =
```
or
```suggestion
          auto shadowNodeWithTrait =
```

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -31,9 +31,15 @@ RootShadowNode::Unshared ReanimatedCommitHook::shadowTreeWillCommit(
 #else
     RootShadowNode::Unshared const &newRootShadowNode) const noexcept {
 #endif
-  if (ReanimatedCommitMarker::isReanimatedCommit()) {
+
+  auto reaShadowNode =
+      std::reinterpret_pointer_cast<ReanimatedCommitShadowNode>(
+          newRootShadowNode);
+
+  if (reaShadowNode->isReanimatedCommit()) {
```
tomekzaw, line 39:
```suggestion
  if (reaShadowNode->hasReanimatedCommitTrait()) {
```

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitShadowNode.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+#ifdef RCT_NEW_ARCH_ENABLED
+
+namespace reanimated {
+
+const ShadowNodeTraits::Trait reanimatedCommitTrait{
```
tomekzaw, line 6:
```suggestion
constexpr ShadowNodeTraits::Trait ReanimatedCommitTrait{
```

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitShadowNode.h
```diff
@@ -0,0 +1,24 @@
...
+    ShadowNodeTraits::Trait(1 << 31)};
```
tomekzaw, line 7:
```suggestion
    1 << 31};
```

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitShadowNode.h
```diff
@@ -0,0 +1,24 @@
...
+
+class ReanimatedCommitShadowNode : public ShadowNode {
```
tomekzaw, line 9:
```suggestion
class ReanimatedTraitsEnabledShadowNode : public ShadowNode {
```

tomekzaw (changes requested):

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedMountHook.cpp
bartlomiejbloniarz: no

packages/react-native-reanimated/Common/cpp/NativeModules/NativeReanimatedModule.cpp
tomekzaw: Right, how about `ReanimatedTraitAccessEnabledShadowNode`? That's probably too long.

packages/react-native-reanimated/Common/cpp/Fabric/ReanimatedCommitShadowNode.h
```diff
@@ -0,0 +1,30 @@
...
+
+namespace reanimated {
+
+// We use this trait to mark that a commit was created by reanimated.
+// Traits are copied when nodes are cloned, so this information
+// won't be lost unless someone explicitly overrides it.
+// We need this information to skip unnecessary updates in
+// the commit hook.
+// Currently RN traits go up to ten, so hopefully
+// the arbitarly chosen number 27 will be safe :)
```
tomekzaw, line 12:
```suggestion
// We use this trait to mark that a commit was created by Reanimated.
// Traits are copied when nodes are cloned, so this information
// won't be lost unless someone explicitly overrides it.
// We need this information to skip unnecessary updates in
// the commit hook.
// Currently RN traits go up to 10, so hopefully
// the arbitrarily chosen number 27 will be safe :)
```

tomekzaw (approved):

PR #6350, opened 2024-07-30
refactor: Rename NativeReanimatedModule to CommonReanimatedModule

tjzel:
The name didn't scale and was misleading as the common c++ module was called 'native' and objC module is called 'REAModule'.


tjzel:
Turns out codegen requires the name `NativeReanimatedModule` 😭

PR #6355, opened 2024-07-31
docs: fix broken link in reanimated babel plugin docs

kacperkapusciak:
After the change to monorepo link leads to 404


tjzel (approved):
🦃

PR #6356, opened 2024-07-31
[reJest] Migrate to use the new runtimetests API

Latropos:
## Summary
Replace all uses of `FAILING` and `WARN` test decorators with `expect().toThrow()`
The full removal of the logic behind these decorators is located here: https://github.com/software-mansion/react-native-reanimated/pull/6358


piaskowyk:
I changed the name of the PR to avoid confusion with Reanimated API.

piaskowyk (approved):

PR #6357, opened 2024-07-31
refactor: Remove 'worklets' dependency on `NativeReanimatedModule`

tjzel:
## Summary

Just applying some dependency inversion principle. Basically now `RNRuntimeDecorator` defines an interface that its callers have to conform to.

Keep in mind this isn't the final shape of how the code is split, I'm submitting this PR for granularity of the changes.

## Test plan

- Android builds & runs
    - on Paper
        - [x] in Development
        - [x] in Release
    - on Fabric
        - [x] in Development
        - [x] in Release
- iOS builds & runs
    - on Paper
        - [x] in Development
        - [x] in Release
    - on Fabric
        - [x] in Development
        - [x] in Release


bartlomiejbloniarz (approved):
🪨

tomekzaw (changes requested):
Requesting changes so that I can peacefully review this PR

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -24,6 +24,7 @@
 #include "CollectionUtils.h"
 #include "EventHandlerRegistry.h"
 #include "FeaturesConfig.h"
+#include "REAWorkletRuntimeDecorator.h"
```
tomekzaw, line 27: We're using `REA` prefix only in Obj-C codebase; otherwise it should always be `Reanimated`, i.e. `ReanimatedCommitHook`, `ReanimatedHermesRuntime` etc.
```suggestion
#include "ReanimatedWorkletRuntimeDecorator.h"
```

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/REAWorkletRuntimeDecorator.cpp
```diff
@@ -0,0 +1,14 @@
+#include "REAWorkletRuntimeDecorator.h"
+#include "JSISerializer.h"
+#include "PlatformLogger.h"
+#include "ReanimatedJSIUtils.h"
+#include "WorkletRuntime.h"
+
+namespace reanimated {
```
tomekzaw, line 7: Add newline
```suggestion
namespace reanimated {

```

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/REAWorkletRuntimeDecorator.cpp
```diff
@@ -0,0 +1,14 @@
...
+void REAWorkletRuntimeDecorator::decorate(jsi::Runtime &rt) {
+  jsi_utils::installJsiFunction(
+      rt, "_log", [](jsi::Runtime &rt, const jsi::Value &value) {
+        PlatformLogger::log(stringifyJSIValue(rt, value));
+      });
+}
+} // namespace reanimated
```
tomekzaw, line 14: Add newline
```suggestion

} // namespace reanimated
```

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/REAWorkletRuntimeDecorator.cpp
```diff
@@ -0,0 +1,14 @@
...
+#include "PlatformLogger.h"
+#include "ReanimatedJSIUtils.h"
+#include "WorkletRuntime.h"
+
+namespace reanimated {
+void REAWorkletRuntimeDecorator::decorate(jsi::Runtime &rt) {
+  jsi_utils::installJsiFunction(
+      rt, "_log", [](jsi::Runtime &rt, const jsi::Value &value) {
+        PlatformLogger::log(stringifyJSIValue(rt, value));
+      });
```
tomekzaw, line 12: Personally I think that `_log` function should be moved to `worklets` since it can be used without Reanimated.

On Android, the tag should be taken from `workletRuntime->name_`.

packages/react-native-reanimated/apple/native/NativeProxy.mm
```diff
@@ -45,13 +45,22 @@ - (void *)runtime;
...
 using namespace react;
 
+static inline bool getIsReducedMotion()
+{
+#if __has_include(<UIKit/UIAccessibility.h>)
+  auto isReducedMotion = UIAccessibilityIsReduceMotionEnabled();
+#else
+  auto isReducedMotion = NSWorkspace.sharedWorkspace.accessibilityDisplayShouldReduceMotion;
+#endif
+  return isReducedMotion;
```
tomekzaw, line 55:
```suggestion
#if __has_include(<UIKit/UIAccessibility.h>)
  return UIAccessibilityIsReduceMotionEnabled();
#else
  return NSWorkspace.sharedWorkspace.accessibilityDisplayShouldReduceMotion;
#endif
```

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/RNRuntimeDecorator.cpp
```diff
@@ -38,7 +37,9 @@ void RNRuntimeDecorator::decorate(
   injectReanimatedCppVersion(rnRuntime);
 
   rnRuntime.global().setProperty(
-      rnRuntime, "_REANIMATED_IS_REDUCED_MOTION", isReducedMotion);
+      rnRuntime,
+      "_REANIMATED_IS_REDUCED_MOTION",
+      nativeReanimatedModule->isReducedMotion());
```
tomekzaw, line 42:
```suggestion
      nativeReanimatedModule->getIsReducedMotion());
```

packages/react-native-reanimated/android/src/main/reanimated/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ add_library(
     SHARED
     ${REANIMATED_COMMON_SOURCES}
     ${SOURCES_ANDROID}
+    "${ANDROID_SRC_DIR}/main/cpp/PlatformLogger.cpp"
```
tomekzaw, line 11: This should be moved to worklets.

tomekzaw (changes requested):

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/REAWorkletRuntimeDecorator.cpp
tjzel: All of this code was through our formatters. Seems like we are missing some settings?

tjzel: I've been through clang-format config options and there's no option for this specific case. I can add a newline here manually but I don't think it's something we should enforce in the codebase.

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/REAWorkletRuntimeDecorator.cpp
tjzel: I will do it in the follow-up PR. I don't want to have any non-cpp code separated in this PR.

packages/react-native-reanimated/android/src/main/reanimated/CMakeLists.txt
tjzel: I will do it in the follow-up PR. I don't want to have any non-cpp code separated in this PR.

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/RNRuntimeDecorator.cpp
tjzel: Why `getIsReducedMotion`?

tomekzaw (approved):

PR #6361, opened 2024-07-31
fix: proper types for userAgent and vendor

WoLewicki:
## Summary

PR restoring proper types for `Navigator`.

When this code would be removed, VSCode takes me to definitions from TS, which show it the same way: 
image


szydlovsky:
@WoLewicki I'm not sure if we can just change it this way, see:
https://github.com/software-mansion/react-native-reanimated/blob/da2e2ef4f944d81d7b17528ad5bfbde67a2a847f/packages/react-native-reanimated/src/js-reanimated/JSReanimated.ts#L268-L279

szydlovsky (approved):
:shipit:

PR #6363, opened 2024-07-31
Remove `ReanimatedHiddenHeaders`, `LoggerInterface` and `SpeedChecker`

tomekzaw:
## Summary

This PR removes some rusty and unused parts of the code like `SpeedChecker` as well as removes `Logger` and `LoggerInterface` in favor of static `PlatformLogger` which also enables us to finally get rid of `ReanimatedHiddenHeaders` and `hidden_headers` along with its configuration in `CMakeLists.txt` and `RNReanimated.podspec`.


packages/react-native-reanimated/android/src/main/cpp/PlatformLogger.cpp
```diff
@@ -0,0 +1,29 @@
+#include <android/log.h>
+
+#include "PlatformLogger.h"
+
+constexpr const char *tag = "Reanimated";
```
tjzel, line 5:
```suggestion
constexpr const auto tag = "Reanimated";
```

tjzel (approved):

PR #6375, opened 2024-08-05
fix: Light mode colors in the make mutable example in docs

MatiPl01:
## Summary

I used `useColorScheme` hook from `react-native` instead of the one from the `@mui/material` library, which returned `'dark'` color scheme name irrespective of the actual selected theme.

This PR changes the hook to the correct one which fixes the text color in the light mode.

### Light theme 

| Before | After |
|-|-|
| Screenshot 2024-08-05 at 18 18 19 | Screenshot 2024-08-05 at 18 16 25 |

### Dark theme 
Works fine, as before.

| Before | After |
|-|-|
| Screenshot 2024-08-05 at 18 17 42 | Screenshot 2024-08-05 at 18 17 03 |

## Test plan

- `cd packages/reanimated-docs`,
- `yarn start`,
- navigate to the `makeMutable` page in docs and see the example

### DIFF
--- a/packages/docs-reanimated/src/examples/MakeMutable.tsx
+++ b/packages/docs-reanimated/src/examples/MakeMutable.tsx
@@ -1,10 +1,5 @@
-import {
-  Text,
-  StyleSheet,
-  View,
-  TouchableOpacity,
-  useColorScheme,
-} from 'react-native';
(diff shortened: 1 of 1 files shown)


bartlomiejbloniarz (approved):
🚀

PR #6383, opened 2024-08-06
Fix colour comparison on Android

Latropos:
## Summary

Due to some recent changes some views are instances of CSSBackgroundDrawable instead of ReactViewBackgroundDrawable. I didn't track which PR introduced this change, since `ReactViewBackgroundDrawable` is marked as deprecated anyway. 

This PR contains changes to make snapshot colour comparison work again.

image


packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -453,13 +454,27 @@ public String obtainProp(int viewTag, String propName) {
...
+          // By default transparency is first, color second
+          return "#" + invertedColor.substring(2, 8) + invertedColor.substring(0, 2);
         }
-        int actualColor = ((ReactViewBackgroundDrawable) background).getColor();
-        String invertedColor = String.format("%08x", (0xFFFFFFFF & actualColor));
-        // By default transparency is first, color second
-        return "#" + invertedColor.substring(2, 8) + invertedColor.substring(0, 2);
+        ;
+
+        return "Unable to resolve background color";
```
piaskowyk, line 477: Something like that would be more readable
```java
int actualColor = -1;
if (background instanceof ReactViewBackgroundDrawable) {
  actualColor = ((ReactViewBackgroundDrawable) background).getColor();
}

if (background instanceof CSSBackgroundDrawable) {
  actualColor = ((CSSBackgroundDrawable) background).getColor();
}

if (actualColor == -1) {
  return "Unable to resolve background color";
}

String invertedColor = String.format("%08x", (0xFFFFFFFF & actualColor));
// By default transparency is first, color second
return "#" + invertedColor.substring(2, 8) + invertedColor.substring(0, 2);
```

piaskowyk (changes requested):

piaskowyk (approved):

PR #6364, opened 2024-08-01
feat: Implement reanimated logger with stack trace customization

MatiPl01:
## Summary

This PR implements better logger to ensure that call stacks in logs are clear and properly show the culprit of the issue.

he default default stack trace, included after calling `console.warn`/`console.error`, contains all calls up with function calls in the library source code and even `console` call itself. It is a bit confusing, because the real source of the problem often is not visible in the call stack until the complete call stack is revealed and the proper call is found.

This new logger is intended to make calls stacks more readable by excluding unnecessary calls from reanimated source code in some warnings where it's better to show the place in the user code which caused the issue. It also adds the `[Reanimated]` prefix to all logs automatically, so it is no longer necessary to remember about adding this prefix manually in every log message.

## Remarks

Some error/warning call stacks cannot be beautified, especially if these are native stacks or execution of the specific function is not tied to the component (e.g. the component schedules the animation but the error is thrown after the animation starts).

PR #6385, opened 2024-08-07
feat: Implement better reanimated logger with clean stack traces

MatiPl01:
## Summary

This PR is a much cleaner approach than proposed in #6364. It includes metro-config modification which is essential to collapse logs from reanimated source code, which aren't helpful to the user while tracking down the issue. The previous approach was trimming logs from reanimated source code completely - this approach just collapses them, so that they are still available to the user and can be revealed above the presented stack trace part.

## General idea

To get better logs, I had to implement the following 2 changes:
1. **metro config** - the `symbolicator` must have been added to properly collapse stack frames that aren't meaningful to the user,
2. **logger object** - the new logger object uses `LogBox.addLog` method, thanks to which we can get pretty stack traces when we log a warning from the UI thread (before such warnings didn't include meaningful stack trace as error stack was created inside `LogBox` after `runOnJS` was called, so we were getting a bit limited JS stack - see [example 11](https://github.com/software-mansion/react-native-reanimated/pull/6387#ui-warning-examle) in the follow up PR).

## Example improvement
(tested on a real project to see if it works there as well)

- current logs either point to the reanimated source code or aren't readable at all (if warning is logged from the UI thread as in the example below)
- new logger shows correct destination of the issue culprit in the code frame, collapses stack frames in the call stack that aren't interesting to the user (reanimated source code) and focuses on the file where the user included problematic code

| Before | After |
|-|-|
| <video src="https://github.com/user-attachments/assets/a5302586-f4d0-4636-8bd8-6c406c9d8c73" /> | <video src="https://github.com/user-attachments/assets/3121636f-69a2-4b6f-8f38-b1889d4c62e1" /> |

## Test plan

See the example in the next PR (#6387).


packages/react-native-reanimated/metro-config/index.js
```diff
@@ -0,0 +1,35 @@
...
+    "/packages/react-native-reanimated/.+\\.(t|j)sx?$",
+    // When reanimated is installed as a dependency (node_modules)
+    "/node_modules/react-native-reanimated/.+\\.(t|j)sx?$",
+  ]
+    // Make patterns work with both Windows and POSIX paths.
+    .map((pathPattern) => pathPattern.replaceAll("/", "[/\\\\]"))
+    .join("|")
+);
+
+function withReanimated(config) {
```
MatiPl01, line 13: I export `withReanimated` function that takes the merged user config as a parameter instead of exporting a separate config in order not to override the `symbolicator` completely. I want to be able to access user's `symbolicator` (usually the default one from `@react-native/metro-config` or `@expo/metro-config`) to keep the default behavior for logs from `react-native` source code, etc. and just a custom behavior for `react-native-reanimated`.

apps/fabric-example/metro.config.js
```diff
@@ -1,4 +1,7 @@
 const { getDefaultConfig, mergeConfig } = require('@react-native/metro-config');
+const {
+  withReanimated,
+} = require('../../packages/react-native-reanimated/metro-config');
```
MatiPl01, line 4: Users will be able to import it using this path:
`react-native-reanimated/metro-config`

packages/react-native-reanimated/metro-config/index.js
tjzel: Great work, however I'd rename it to `wrapWithReanimatedMetroConfig` or something like this for clarity.

packages/react-native-reanimated/metro-config/index.js
tjzel, line 1: On one hand I'd want this file in .ts for source, but it's so small that it seems like an overkill.

packages/react-native-reanimated/src/logger/LogBox.ts
```diff
@@ -0,0 +1,43 @@
...
+  message: Message;
+  category: Category;
+  componentStack: ComponentStack;
+  componentStackType: ComponentStackType | null;
+  stack?: string;
+};
+
+interface LogBoxExtended extends LogBoxStatic {
+  addLog(data: LogData): void;
+}
```
tjzel, line 41: Where are those types coming from? If from React-Native, can we just import them instead of re-declaring them?

packages/react-native-reanimated/metro-config/index.js
```diff
@@ -0,0 +1,35 @@
+const COLLAPSED_STACK_REGEX = new RegExp(
+  [
+    // For internal usage in the example app
+    "/packages/react-native-reanimated/.+\\.(t|j)sx?$",
+    // When reanimated is installed as a dependency (node_modules)
+    "/node_modules/react-native-reanimated/.+\\.(t|j)sx?$",
```
tjzel, line 6: Does it work with hoisted node_modules? Also, what happens if it's actually the code in Reanimated which crashes?

packages/react-native-reanimated/src/logger/LogBox.ts
tjzel, line 1: Since this file is all about types, let's name it something along the lines of `LogBoxInterface`.

packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -0,0 +1,75 @@
+'use strict';
+import { runOnJS } from '../threads';
+import { LogBox } from './LogBox';
+import type { LogLevel, LogData } from './LogBox';
+
+const addLog = LogBox.addLog.bind(LogBox);
```
tjzel, line 6: Since you don't use the rest of the `LogBox` here, how about exporting only `addLog` from the previous file?

tjzel (changes requested):

packages/react-native-reanimated/metro-config/index.js
MatiPl01: I can add `d.ts` file for type declarations, similarly as metro does with its exported default config. I think this will be sufficient in such a case.

packages/react-native-reanimated/src/logger/LogBox.ts
MatiPl01: Unfortunately not, `LogBox.addLog` method is hidden form users and TS has no exported type declarations related to this method. I needed to copy these types from RN LogBox flow types declarations and include them manually to create the proper `LogData` type.

Screenshot 2024-08-08 at 12 32 15

The only exported type is `LogBoxStatic` which looks like this:

```tsx
export interface LogBoxStatic {
  /**
   * Silence any logs that match the given strings or regexes.
   */
  ignoreLogs(patterns: (string | RegExp)[]): void;

  /**
   * Toggle error and warning notifications
   * Note: this only disables notifications, uncaught errors will still open a full screen LogBox.
   * @param ignore whether to ignore logs or not
   */
  ignoreAllLogs(ignore?: boolean): void;

  install(): void;
  uninstall(): void;
}

export const LogBox: LogBoxStatic;
export type LogBox = LogBoxStatic;
```

MatiPl01: I can add a comment at the top of the file that these types are copied from RN sources with the file location I copied them from.

packages/react-native-reanimated/metro-config/index.js
MatiPl01: It should work the same as with hoisted `react-native` module where ignore paths are specified in the default metro config. The path will always end width `node_modules/react-native-reanimated`, even if the module is hoisted, so the regex will match a path in such a case as well.

Recordings in the description are made in a project which uses hoisting and the applied metro config worked fine.

apps/fabric-example/metro.config.js
MatiPl01: Because of this:

```
> yarn start --reset-cache
error Cannot find module 'react-native-reanimated/metro-config'
Require stack:
- /Users/mateu/Work/projects/react-native-reanimated/apps/fabric-example/metro.config.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/cosmiconfig/node_modules/import-fresh/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/cosmiconfig/dist/loaders.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/cosmiconfig/dist/createExplorer.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/cosmiconfig/dist/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/metro-config/src/loadConfig.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/metro-config/src/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native/community-cli-plugin/dist/utils/loadMetroConfig.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native/community-cli-plugin/dist/commands/bundle/buildBundle.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native/community-cli-plugin/dist/commands/bundle/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native/community-cli-plugin/dist/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/react-native.config.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/node_modules/cosmiconfig/dist/loaders.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/node_modules/cosmiconfig/dist/defaults.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/node_modules/cosmiconfig/dist/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/build/readConfigFromDisk.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/build/loadConfig.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/build/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native-community/cli/build/commands/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native-community/cli/build/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/cli.js.
Error: Cannot find module 'react-native-reanimated/metro-config'
Require stack:
- /Users/mateu/Work/projects/react-native-reanimated/apps/fabric-example/metro.config.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/cosmiconfig/node_modules/import-fresh/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/cosmiconfig/dist/loaders.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/cosmiconfig/dist/createExplorer.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/cosmiconfig/dist/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/metro-config/src/loadConfig.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/metro-config/src/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native/community-cli-plugin/dist/utils/loadMetroConfig.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native/community-cli-plugin/dist/commands/bundle/buildBundle.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native/community-cli-plugin/dist/commands/bundle/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native/community-cli-plugin/dist/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/react-native.config.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/node_modules/cosmiconfig/dist/loaders.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/node_modules/cosmiconfig/dist/defaults.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/node_modules/cosmiconfig/dist/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/build/readConfigFromDisk.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/build/loadConfig.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli-config/build/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native-community/cli/build/commands/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/node_modules/@react-native-community/cli/build/index.js
- /Users/mateu/Work/projects/react-native-reanimated/node_modules/react-native/cli.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Module._load (node:internal/modules/cjs/loader:985:27)
    at Module.require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at Object.<anonymous> (/Users/mateu/Work/projects/react-native-reanimated/apps/fabric-example/metro.config.js:2:43)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
```

MatiPl01: It works in another project's monorepo setup, though, so I think the problem is only with path resolution in our reanimated monorepo.

MatiPl01: It seems that I needed to run `yarn build` and it started to work then

packages/react-native-reanimated/src/logger/LogBox.ts
MatiPl01: I moved `addLog` (renamed to `addLogBoxLog`) implementation to this file, so I think it's better to keep the current name.

packages/react-native-reanimated/metro-config/index.js
tjzel: I wanted types for internal usage, but let's just ignore this for now.

packages/react-native-reanimated/metro-config/index.d.ts
tjzel, line 1: Does it work?

packages/react-native-reanimated/metro-config/index.js
```diff
@@ -0,0 +1,35 @@
+const COLLAPSED_STACK_REGEX = new RegExp(
+  [
+    // For internal usage in the example app
+    '/packages/react-native-reanimated/.+\\.(t|j)sx?$',
```
tjzel, line 4: Internal packages also acquire it via linked `node_modules`, is it necessary?

packages/react-native-reanimated/src/errors.ts
```diff
@@ -1,6 +1,20 @@
...
+    // Set the name of the error to the class name
+    this.name = this.constructor.name;
+
+    if (Error.captureStackTrace) {
+      // Exclude constructor from stack trace
+      Error.captureStackTrace(this, this.constructor);
+    }
+  }
+}
+
```
tjzel, line 17: As discussed, let it be a usual function that you can call with `new`.

packages/react-native-reanimated/src/logger/LogBox.ts
tjzel: Let's add such comment!

packages/react-native-reanimated/src/threads.ts
```diff
@@ -255,3 +258,11 @@ export function runOnJS<Args extends unknown[], ReturnValue>(
...
   };
 }
+
+// Override the logFunction implementation with the one that adds logs
+// with better stack traces to the LogBox
+loggerImpl.logFunction = (data: LogData) => {
+  'worklet';
+  runOnJS(logToLogBoxAndConsole)(data);
+};
+shareableMappingCache.set(logger, makeShareableCloneRecursive(logger));
```
tjzel, line 268: Let's export a function from `logger` called something like `replaceLoggerImplementation` since this might be a bit confusing.

packages/react-native-reanimated/src/publicGlobals.ts
```diff
@@ -21,4 +23,9 @@ declare global {
    * which is the Reanimated UI runtime.
    */
   var _WORKLET_RUNTIME: ArrayBuffer;
+
+  /**
+   * This is a custom error class that is thrown by Reanimated.
+   */
+  var ReanimatedError: typeof ReanimatedErrorClass;
```
tjzel, line 30: I don't think we should make it a public global. Let's move it to `globals.d.ts` to make it private instead.

packages/react-native-reanimated/metro-config/index.d.ts
MatiPl01: Yes, it shows the type properly when I hover over the function call.

Screenshot 2024-08-12 at 15 46 21

And when I press `CMD + LMB` to see the source, it shows the proper type declaration.

Screenshot 2024-08-12 at 15 46 28

Since `metro.config.js` file is a JS file, it won't show errors if I pass incorrect config but it works the same for imports from `@react-native/metro-config`

packages/react-native-reanimated/metro-config/index.js
MatiPl01: Not quite. The path I am getting from metro is pointing to `packages/react-native-reanimated` instead of `node_modules/react-native-reanimated`.

Screenshot 2024-08-12 at 15 56 03

What is more, there is no `react-native-reanimated` symbolic link in `node_modules` in example apps so library sources are imported just from the `packages/react-native-reanimated`.

Screenshot 2024-08-12 at 15 58 05

packages/react-native-reanimated/src/errors.ts
```diff
@@ -1,6 +1,23 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-misused-new
+  new (message: string): ReanimatedError;
+}
+
+const ReanimatedErrorFactory: ReanimatedError = ((message: string) => {
+  'worklet';
+  const errorInstance = new Error(`[Reanimated] ${message}`);
+  errorInstance.name = 'ReanimatedError';
+  return errorInstance;
+}) as unknown as ReanimatedError;
```
tjzel, line 14: We should try something like this, `ErrorConstructor` seems to be the name pattern for such terms.

```tsx
type ReanimatedError = Error & 'ReanimatedError'; // Signed type.

interface ReanimatedErrorConstructor extends Error {
  new (message?: string): ReanimatedError;
  (message?: string): ReanimatedError;
  readonly prototype: ReanimatedError;
}

const ReanimatedError: ReanimatedErrorConstructor = function ReanimatedError(
  message?: string
) {
  // Implementation.
} as ReanimatedErrorConstructor;

const error: ReanimatedError = new Error('ReanimatedError'); // Error due to signed type.

const reaError: Error = new ReanimatedError('ReanimatedError'); // All ok.

const rea2Error: ReanimatedError = new ReanimatedError('ReanimatedError'); // All ok.
```

packages/react-native-reanimated/tsconfig.json
```diff
@@ -5,5 +5,5 @@
       "react-native-reanimated": ["./src"]
     }
   },
-  "include": ["src"]
+  "include": ["src", "metro-config"]
```
tjzel, line 8: Why is it needed?

tjzel (commented):
Looks all good, let's figure out the types and also let's make sure we have defined behavior when copying errors from React Runtime to Workler Runtime, i.e.:

```tsx
// React Runtime
const error = new ReanimatedError();
runOnUI(() => {
  console.log(error); // Should fail, let's make it verbose.
  };
)();
```

packages/react-native-reanimated/metro-config/index.js
tjzel: The link is not in the `node_modules` of the package, but in `node_modules` of the whole monorepo.

Screenshot 2024-08-13 at 09 50 46


I guess metro simply resolves this symbolic link to the actual path. How about instead of hardcoding the path in here we add an option to add ignored paths to the wrapper?

tjzel: Or maybe we should just forget whether it's `packages` or `node_modules` and just look for `react-native-reanimated`

MatiPl01: > Or maybe we should just forget whether it's packages or node_modules and just look for react-native-reanimated

No, it won't work as well. We won't be able to correct stack traces in our reanimated monorepo where the root directory is named `react-native-reanimated`. See the path `.../react-native-reanimated/packages/react-native-reanimated/...` in the screenshot above.

> I guess metro simply resolves this symbolic link to the actual path. How about instead of hardcoding the path in here we add an option to add ignored paths to the wrapper?

It might be useful but the benefit of exporting our metro-config wrapper will fade away. The user still can add custom logic to filter out stack frames via the `customizeFrame` method of the `symbolicator`.

I would prefer to keep the current metro config. If the user has a custom path to `react-native-reanimated` (rather a rare case, usually all packages leave in the `node_modules` directory), they can create a custom metro-config entry to filter out unwanted stack frames instead of using the provided wrappeer. I would rather add information to docs how it works and how to properly create such a config.

MatiPl01: Alternatively, we can add an option to provide a path to `node_modules` (but it seems unnecessary to me as metro in its default config assumes that `react-native` sources live under the `node_modules` directory, so I think we can do something similar).

packages/react-native-reanimated/src/errors.ts
MatiPl01: Can't create `const ReanimatedError` variable as it is added to the ignore list on UI and results in being not defined when I use this name.

MatiPl01: It makes more sense to me too keep the proper name for the type, which shows up when I hover over the `ReanimatedError` call instead of the name of the variable itself, which is assigned to `global.ReanimatedError`, so the name of the variable I initially use can be anything. The `Error` type that is globally available in TS is called `Error`, not `ErrorConstructor`, so I would prefer to follow this convention.

packages/react-native-reanimated/tsconfig.json
MatiPl01: Because of `index.d.ts` in the `metro-config` directory

Screenshot 2024-08-13 at 11 36 31

MatiPl01: I can also add it to eslintignore but it seems better to keep eslint and ts checks

packages/react-native-reanimated/src/errors.ts
MatiPl01: And, why do I need to make these strange typing hacks whilst it already works fine?

Screenshot 2024-08-13 at 11 42 06

packages/react-native-reanimated/plugin/src/globals.ts
```diff
@@ -116,7 +116,6 @@ const notCapturedIdentifiers = [
 
   // Reanimated
   '_WORKLET',
-  'ReanimatedError',
```
tjzel: It should stay here. We don't want to copy it to the Worklet runtime.

packages/react-native-reanimated/src/errors.ts
```diff
@@ -18,11 +18,6 @@ const ReanimatedErrorConstructor: ReanimatedErrorConstructor =
     return errorInstance;
   } as ReanimatedErrorConstructor;
 
-export function registerReanimatedError() {
-  'worklet';
-  global.ReanimatedError = ReanimatedErrorConstructor;
-}
-
```
tjzel: Why did you remove it? We still want to register it in the global scope on Worklet runtimes, just not on the React runtime.

packages/react-native-reanimated/src/globals.d.ts
```diff
@@ -113,5 +112,4 @@ declare global {
     shadowNodeWrapper: ShadowNodeWrapper,
     propName: string
   ) => string;
-  var ReanimatedError: ReanimatedErrorConstructor;
```
tjzel: Keep it here with annotation that it's only present on Worklet Runtimes.

tjzel (changes requested):

packages/react-native-reanimated/src/errors.ts
MatiPl01: Yeah, I can see that. I wanted to revert these changes but decided to keep them for now.
I don't like the idea of making `ReanimatedError` available globally in the worklet runtime but make it different for the react runtime. It totally makes no sense to me and is much more confusing than having `ReanimatedError` available globally.

MatiPl01: I would prefer to keep things consistent and either import the error in each file or use the global definition. If the first is not possible, I would just stick to the globally available `ReanimatedError`.

packages/react-native-reanimated/src/globals.d.ts
```diff
@@ -112,4 +113,5 @@ declare global {
     shadowNodeWrapper: ShadowNodeWrapper,
     propName: string
   ) => string;
+  var ReanimatedError: ReanimatedErrorConstructor;
```
tjzel, line 116:
```suggestion
  /**
  * ReanimatedError is present in the global scope only on Worklet runtimes. To use it on the React runtime you have to import it.
  **/
  var ReanimatedError: ReanimatedErrorConstructor;
```

packages/react-native-reanimated/src/initializers.ts
```diff
@@ -9,6 +9,8 @@ import {
 } from './threads';
 import { mockedRequestAnimationFrame } from './mockedRequestAnimationFrame';
 
+registerReanimatedError();
```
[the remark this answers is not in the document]
MatiPl01: Needed to add `registerReanimatedError` call to callback of `makeShareableCloneRecursive` in `createWorkletRuntime`. Now works fine.
Thanks for suggestion.

packages/react-native-reanimated/src/threads.ts
```diff
@@ -255,3 +263,19 @@ export function runOnJS<Args extends unknown[], ReturnValue>(
     );
   };
 }
+
+// Override the logFunction implementation with the one that adds logs
+// with better stack traces to the LogBox (need to override it after runOnJS
```
tjzel, line 268:
```suggestion
// with better stack traces to the LogBox (need to override it after `runOnJS`
```

packages/react-native-reanimated/src/threads.ts
```diff
@@ -162,13 +170,13 @@ export function runOnUIImmediately<Args extends unknown[], ReturnValue>(
...
-      '[Reanimated] `runOnUIImmediately` cannot be called on the UI runtime. Please call the function synchronously or use `queueMicrotask` or `requestAnimationFrame` instead.'
+    throw new ReanimatedError(
+      '`runOnUIImmediately` cannot be called on the UI runtime. Please call the function synchronously or use `queueMicrotask` or `requestAnimationFrame` instead.'
     );
   }
   if (__DEV__ && !SHOULD_BE_USE_WEB && !isWorkletFunction(worklet)) {
-    throw new Error(
-      '[Reanimated] `runOnUIImmediately` can only be used on worklets.'
+    throw new ReanimatedError(
+      '`runOnUIImmediately` can only be used on worklets.'
```
tjzel, line 179:
```suggestion
      '`runOnUIImmediately` can only be used with worklets.'
```

packages/react-native-reanimated/src/threads.ts
```diff
@@ -76,12 +84,12 @@ export function runOnUI<Args extends unknown[], ReturnValue>(
...
   if (__DEV__ && !SHOULD_BE_USE_WEB && _WORKLET) {
-    throw new Error(
-      '[Reanimated] `runOnUI` cannot be called on the UI runtime. Please call the function synchronously or use `queueMicrotask` or `requestAnimationFrame` instead.'
+    throw new ReanimatedError(
+      '`runOnUI` cannot be called on the UI runtime. Please call the function synchronously or use `queueMicrotask` or `requestAnimationFrame` instead.'
     );
   }
   if (__DEV__ && !SHOULD_BE_USE_WEB && !isWorkletFunction(worklet)) {
-    throw new Error('[Reanimated] `runOnUI` can only be used on worklets.');
+    throw new ReanimatedError('`runOnUI` can only be used on worklets.');
```
tjzel, line 92:
```suggestion
    throw new ReanimatedError('`runOnUI` can only be used with worklets.');
```

packages/react-native-reanimated/src/errors.ts
```diff
@@ -1,6 +1,35 @@
...
+    errorInstance.name = 'ReanimatedError';
+    return errorInstance;
+  } as ReanimatedErrorConstructor;
+
+export { ReanimatedErrorConstructor as ReanimatedError };
+
+/**
+ * Registers ReanimatedError constructor in global scope.
+ * Used to add ReanimatedError in the worklet runtime.
+ */
```
tjzel, line 26:
```suggestion
/**
 * Registers `ReanimatedError` in global scope.
 * Use it only for Worklet runtimes.
 */
```

packages/react-native-reanimated/src/errors.ts
```diff
@@ -1,6 +1,35 @@
...
+export function registerReanimatedError() {
+  'worklet';
+  (global as Record<string, unknown>).ReanimatedError =
+    ReanimatedErrorConstructor;
+}
```
tjzel, line 31: How about such addition here:

```ts
if(!_WORKLET) {
throw new Error(...);
}
```

tjzel (approved):
Finally! I added only some minor comments but you can ignore them if you feel so. Let's merge this!

Just a reminder, please make sure to make a follow-up PR which adds information about this feature in our Docs.

packages/react-native-reanimated/src/errors.ts
MatiPl01: I think this will be useful. I will add this change.

PR #6387, opened 2024-08-07
feat: Use new logger and ReanimatedError for lib warnings and errors

MatiPl01:
## Summary

This PR replaces all `console.warn`, `console.error` and `throw new Error` statements with custom logger calls (`logger.warn`, `logger.error` and `logger.newError` respectively).

It declares the logger object as a global variable for easier usage and to prevent circular imports (as it imports `runOnJS`, which was the cause of circular dependencies).

## Test plan

- open the `EmptyExample` in the example app,
- change the `SHOW_EXAMPLE` constant to see the respective error/warning,
- see better error logs with code frame that was the culprit (if possible - sometimes we cannot get pretty error stack trace, especially when functions are called asynchronously)


MatiPl01:
## Logs comparison
(Based on 16 examples that I tested out)

### Improved logs

#### 1. Error: [Reanimated] Property `text` was whitelisted both as UI and native prop. Please remove it from one of the lists.

| Before | After |
|-|-|
| Simulator Screenshot - iPhone 15 Pro - 2024-08-07 at 17 53 25 |  |

#### 2. [Reanimated] Property "opacity" of AnimatedComponent(View) may be overwritten by a layout animation. Please wrap your component with an animated view and apply the layout animation on the wrapper.

| Before | After |
|-|-|
| Simulator Screenshot - iPhone 15 Pro - 2024-08-07 at 17 57 03 |  |

#### 3. ReanimatedError: [Reanimated] Bezier x values must be in [0, 1] range., js engine: reanimated

| Before | After |
|-|-|
| simulator_screenshot_6D519664-8F46-4494-9192-E284CB9E6297 |  |

#### 4. Warning: Error: [Reanimated] Function `getViewProp` requires a component to be passed as an argument on Fabric.

| Before | After |
|-|-|
| simulator_screenshot_9554B659-06FF-4587-A6ED-9F92621B8689 |  |

#### 5. Warning: Error: [Reanimated] Invalid color space provided: INVALID. Supported values are: ['RGB', 'HSV'].

| Before | After |
|-|-|
| simulator_screenshot_FBBC89DE-0145-4AAC-BE45-419E92380E42 |  |

#### 6. Warning: Error: [Reanimated] Unsupported value for "interpolate"

| Before | After |
|-|-|
| simulator_screenshot_7F209BED-7096-485E-8DDF-53B35BA23108 |  |

### Unchanged logs

apps/common-app/src/examples/EmptyExample.tsx
```diff
@@ -1,19 +1,248 @@
-import { Text, StyleSheet, View } from 'react-native';
-
 import React from 'react';
```
MatiPl01, line 1: I will revert changes from this file before merging. I don't think we need to include this in examples.

packages/react-native-reanimated/src/Bezier.ts
```diff
@@ -7,6 +8,8 @@
 
 // These values are established by empiricism with tests (tradeoff: performance VS precision)
 
+const logger = global.__reanimatedLogger;
```
MatiPl01, line 11: I had to access `__reanimatedLogger` via `global` as next example was not seeing `__reanimatedLogger` when I accessed it directly, thus it showed this error:

```
ReferenceError: __reanimatedLogger is not defined
```

apps/next-example/babel.config.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
         alias: {
           'react-native': '../../node_modules/react-native-web',
           // Uncomment this if you want fast-refresh to work with reanimated:
-          // 'react-native-reanimated': '../src/index',
```
tjzel, line 11: What's this?

packages/react-native-reanimated/jest-setup.js
```diff
@@ -1,2 +1,21 @@
...
+  error(msg) {
+    console.log(formatMessage(msg));
+  },
+  fatal(msg) {
+    console.log(formatMessage(msg));
+  },
+  newError(msg) {
+    return new Error(formatMessage(msg));
+  }
+}
```
tjzel, line 21: Shouldn't this go into `mocks`?

packages/react-native-reanimated/jest.config.js
```diff
@@ -5,4 +5,7 @@ module.exports = {
   setupFilesAfterEnv: ['@testing-library/jest-native/extend-expect'],
   testEnvironment: 'node',
   transformIgnorePatterns: [],
+  "globals": {
```
tjzel, line 8: Why do we need this?

packages/react-native-reanimated/src/Bezier.ts
tjzel: Have you tried `globalThis` instead?

packages/react-native-reanimated/src/Bezier.ts
```diff
@@ -98,7 +101,7 @@ export function Bezier(
   }
 
   if (!(mX1 >= 0 && mX1 <= 1 && mX2 >= 0 && mX2 <= 1)) {
-    throw new Error('[Reanimated] Bezier x values must be in [0, 1] range.');
+    throw logger.newError('Bezier x values must be in [0, 1] range.');
```
tjzel, line 104: I'd much more prefer a syntax among the lines of `throw new ReanimatedError`

packages/react-native-reanimated/src/animation/util.ts
```diff
@@ -88,12 +90,13 @@ export function recognizePrefixSuffix(
   value: string | number
 ): RecognizedPrefixSuffix {
   'worklet';
+  console.log('recognizePrefixSuffix');
```
tjzel, line 93:
```suggestion
```

packages/react-native-reanimated/src/hook/utils.ts
```diff
@@ -1,7 +1,10 @@
 'use strict';
 import type { WorkletFunction } from '../commonTypes';
+
```
tjzel, line 3:
```suggestion
```

packages/react-native-reanimated/src/layoutReanimation/web/domUtils.ts
```diff
@@ -1,11 +1,14 @@
 'use strict';
 
 import type { ReanimatedHTMLElement } from '../../js-reanimated';
+
```
tjzel, line 4:
```suggestion
```

packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -57,4 +72,4 @@ function createLogger(prefix: string) {
   };
 }
 
-export const logger = createLogger('[Reanimated]');
+global.__reanimatedLogger = createLogger('[Reanimated]');
```
tjzel, line 75: I wonder if we actually want to add this to the global scope. It's definitely doable to just export it with some dependency inversion.

Also, are we certain that the logger will be defined in the global scope already for every case where we use it?

packages/react-native-reanimated/src/mutables.ts
```diff
@@ -1,11 +1,14 @@
 'use strict';
 import { shouldBeUseWeb } from './PlatformChecker';
 import type { Mutable } from './commonTypes';
+
 import { shareableMappingCache } from './shareableMappingCache';
 import { makeShareableCloneRecursive } from './shareables';
 import { executeOnUIRuntimeSync, runOnUI } from './threads';
 import { valueSetter } from './valueSetter';
 
+const logger = global.__reanimatedLogger;
```
tjzel, line 10: All usages of `global.__reanimatedLogger` that happened before this file was evaluated will be broken.

packages/react-native-reanimated/src/publicGlobals.ts
```diff
@@ -21,4 +21,16 @@ declare global {
    * which is the Reanimated UI runtime.
    */
   var _WORKLET_RUNTIME: ArrayBuffer;
+
+  /**
+   * Logger is set as a global variable for easier usage without the
+   * necessity of importing it in every file. It also caused dependency
+   * cycle when imported directly from the logger file as it uses runOnJS.
+   */
```
tjzel, line 29: Let's try to solve this issue.

packages/react-native-reanimated/src/runtimes.ts
```diff
@@ -2,13 +2,16 @@
 import { isWorkletFunction } from './commonTypes';
 import type { WorkletFunction } from './commonTypes';
 import { setupCallGuard, setupConsole } from './initializers';
+
```
tjzel, line 5:
```suggestion
```

tjzel (changes requested):

apps/next-example/babel.config.js
MatiPl01: This is an incorrect path 😄

MatiPl01: I think it wasn't updated after migrating to monorepo. There is no src directory in any of ancestor directories relative to the `next-example`.

packages/react-native-reanimated/jest-setup.js
MatiPl01: You mean the `packages/react-native-reanimated/src/mock.ts` file?

packages/react-native-reanimated/jest.config.js
```diff
@@ -5,4 +5,7 @@ module.exports = {
   setupFilesAfterEnv: ['@testing-library/jest-native/extend-expect'],
   testEnvironment: 'node',
   transformIgnorePatterns: [],
+  globals: {
+    ReanimatedError: true
+  }
```
tjzel, line 10: Why do we set it to `true` here?

packages/react-native-reanimated/plugin/src/globals.ts
```diff
@@ -116,7 +116,6 @@ const notCapturedIdentifiers = [
 
   // Reanimated
   '_WORKLET',
-  'ReanimatedError',
```
tjzel, line 119: Merge to upstream 😜

packages/react-native-reanimated/src/index.ts
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+import './errors';
```
tjzel, line 3: AFAIK this isn't necessary anymore, right?

packages/react-native-reanimated/src/screenTransition/RNScreensTurboModule.ts
```diff
@@ -1,11 +1,14 @@
 'use strict';
+
```
tjzel, line 2:
```suggestion
```

tjzel (approved):
All looks good!

Since now we want to enforce the use of `ReanimatedError` and `logger` in our source code, please add a CI step that ensures no plain errors (except places where we must) or no `console.something` is used.

I guess we'd need to make an ESLint plugin for that, so don't consider it a priority task.

packages/react-native-reanimated/jest.config.js
MatiPl01: It was needed when `ReanimatedError` was declared as a global variable. Now is no longer needed.

PR #6391, opened 2024-08-08
chore: Make GitHub actions names match their titles, cleanup

tjzel:
## Summary

Currently, the name of the (failing) GitHub action mismatches its filename, which is annoying  in development. This PR adds such matching along with some necessary cleanups.

## Test plan

🚀


.github/workflows/expo-devclient-build-check-nightly.yml
tjzel, line 1: This CI will fail, fixing it is out of scope of this PR.

bartlomiejbloniarz (approved):

.github/workflows/example-ios-build-check.yml
```diff
@@ -45,29 +45,22 @@ jobs:
       - name: Check out Git repository
         uses: actions/checkout@v4
 
-      - name: Install monorepo node dependencies
-        run: yarn install
-      - name: Restore Reanimated node_modules from cache
+      # Does this optimization even work with new yarn?
+      # Perhaps we should cache `yarn's` cache instead?
```
m-bert, line 49: I think we should stay consistent with these comments and stick with `TODO` convention:

`# TODO: Check if this optimization works with new yarn`

.github/workflows/example-macos-build-check.yml
```diff
@@ -0,0 +1,89 @@
...
+    concurrency:
+      group: macos-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v4
+
+      # Does this optimization even work with new yarn?
+      # Perhaps we should cache `yarn's` cache instead?
+      # - name: Restore node_modules from cache
```
m-bert, line 46: Do we need to keep this previous version in comment?

.github/workflows/example-typescript-check-and-lint.yml
```diff
@@ -16,12 +16,15 @@ jobs:
...
       matrix:
-        working-directory:
-          [
-            apps/paper-example,
+        working-directory: [
+            apps/common-app,
             apps/fabric-example,
-            apps/web-example,
+            apps/macos-example,
+            # We don't check for Next.
```
m-bert, line 23: Maybe we should explain why we don't check for `Next` instead of just leaving it commented out.

.github/workflows/npm-reanimated-package-build.yml
```diff
@@ -29,6 +29,8 @@ jobs:
   build:
     if: github.repository == 'software-mansion/react-native-reanimated'
     runs-on: ubuntu-latest
+    env:
+      PACKAGE_NAME:
```
m-bert, line 33: Shouldn't we define package name here? Correct me if I'm wrong, but at this moment it is empty

tjzel: This is just a placeholder so the .yml linter doesn't yell at me for using probably non-existing env variable 🤣

PR #6396, opened 2024-08-09
Detox setup for IOS

Latropos:
## Summary
Setup detox environment for IOS. The detox should only run the runtime Tests, the decision whether tests passed is made on re-jest micro-framework.
Therefore depending on the output from our micro-framework we display component marking if the test passed or not.

Right now only one test suite was created, we run only tests of `babel plugin` 

## Test plan

Use
```
yarn detox-debug-paper-build
yarn detox-debug-paper-test
```
To build and test debug version of the IOS app.

https://github.com/user-attachments/assets/c6f9cf95-9b39-427d-85e8-1a8e24df9e48


piaskowyk:
We no longer want to continue this approach

PR #6397, opened 2024-08-09
Update RN pager view patch file

szydlovsky:
## Summary

Our patch to `react-native-pager-view` for `RN 0.75-rc.6` works, but seems like it isn't the best way to do that, as stated [here](https://github.com/callstack/react-native-pager-view/issues/850#issuecomment-2275240214). In short, removing the `override` keyword will cause tab bar navigation crashes. Instead, I adjusted the functions' signature, so that they actually override something.

## Test plan

Try building android examples


piaskowyk (approved):

tomekzaw (approved):

Romick2005:
It is better to go with this [fix](https://github.com/callstack/react-native-pager-view/issues/850#issuecomment-2275240214).

szydlovsky:
@Romick2005 yeah, I use exactly that fix (brought back `override`, removed nullability of `View` arguments)

Issue #6341, opened 2024-07-29
WithDelay of 0ms should not cause any delay


Latropos:
### Description

Definitely `withDelay(0,s omeAnimation);` shouldn't cause any delay, and even if it cause it should be to small to be noticeable.
Yet the delay is clearly visible;
The lower square in the recording has the delay of 0ms attached.

https://github.com/user-attachments/assets/261d92d2-f302-4240-96c8-3d14352431c2

### Steps to reproduce

Replace current EmptyExample with

```tsx
import { StyleSheet, View, Button } from 'react-native';
import React from 'react';
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withDelay,
  withTiming,
} from 'react-native-reanimated';

export default function EmptyExample() {
  const leftSV = useSharedValue(0);

  const animatedStyle = useAnimatedStyle(() => {
    return { left: withTiming(leftSV.value) };
  });

  const animatedStyleDelayed = useAnimatedStyle(() => {
    return { left: withDelay(0, withTiming(leftSV.value)) };
  });

  return (
    <View style={styles.container}>
      <Button
        title="toggle"
        onPress={() => {
          leftSV.value = leftSV.value === 200 ? 0 : 200;
        }}
      />
      <Animated.View style={[styles.box, animatedStyle]} />
      <Animated.View style={[styles.box, animatedStyleDelayed]} />
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    padding: 50,
  },
  box: {
    height: 50,
    width: 50,
    backgroundColor: 'navy',
    borderRadius: 5,
    margin: 5,
  },
});
```

### Snack or a link to a repository

https://snack.expo.dev/tOBz0-UpuWHJjKJqu6p9-

### Reanimated version

nightly

### React Native version

0.74.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Acknowledgements

Yes

TomCorvus:
For me, it is not a bug, it is just the delay to execute the code like a native `setTimeout`.
If you execute a function directly and execute the same function in a the setTimeout 0, the function called by the setTimeout will be always called after even if the `setTimeout` is called before.

```js
function consoleLog(mode = "directly") {
    console.log("called " + mode)
}

setTimeout(() =>{
    consoleLog("with timeout");
[...]

PR #6398, opened 2024-08-09
Skip zero delay

Latropos:
## Summary

Closes https://github.com/software-mansion/react-native-reanimated/issues/6341

Probably we want to fix the comparison inside `withDelay` to start the delayed animation as soon as the conditions are fulfilled, instead of waiting an additional frame.

```diff
-if (now - startTime > delayMs || animation.reduceMotion) {
+if (now - startTime >= delayMs || animation.reduceMotion) {
```

With this fix the delay of 0ms is only visible if you are really picky:

<table>
<tr>
<td> BEFORE </td>
<td> AFTER </td>
</tr>
<tr>
<td> 

https://github.com/user-attachments/assets/261d92d2-f302-4240-96c8-3d14352431c2

</td>
<td> 

https://github.com/user-attachments/assets/cf833359-7440-41d1-ab0a-daa3050bae15

</td>
</tr>

### Test plan

Includes simplification of tests of withDelay. Tested on both IOS and Android


apps/common-app/src/examples/RuntimeTests/tests/animations/withDelay/addDelays.test.tsx
```diff
@@ -105,89 +105,55 @@ describe('Add three delays', () => {
 
     const updates = updatesContainer.getUpdates(getTestComponent(componentType));
     const updatesNative = await updatesContainer.getNativeSnapshots(getTestComponent(componentType));
+    updatesNative.forEach(update => {
+      if ('width' in update && update.width === '[Reanimated] Unable to resolve view') {
+        update.width = 0;
+      }
+    });
```
piaskowyk, line 112: Why we do need that?

piaskowyk (approved):
Please take to consider my previous comments, but overall seems to be good 👍

PR #6400, opened 2024-08-09
Fix `inline style` tests in `Jest`

m-bert:
## Summary

It was brought to our attention that `inline styles` do not work with `Jest`. This PR aims to fix this problem.

## Test plan

Run tests


yarn.lock
```diff
@@ -7174,16 +7174,22 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@testing-library/react-native@npm:^7.1.0":
-  version: 7.2.0
-  resolution: "@testing-library/react-native@npm:7.2.0"
+"@testing-library/react-native@npm:^12.5.2":
```
tjzel, line 7177: Nice bump 👀

packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -584,6 +584,7 @@ export function createAnimatedComponent(
         <Component
           nativeID={nativeID}
           {...filteredProps}
+          jestInlineStyle={this.props.style}
```
tjzel, line 587: Do we want to attach it when we're not in Jest environment?

packages/react-native-reanimated/src/jestUtils.ts
```diff
@@ -27,22 +27,81 @@ const defaultFramerateConfig = {
...
+      };
+    }
+
     currentStyle = {
       ...styleObject,
-      ...component.props.jestAnimatedStyle?.value,
+      ...currentStyle,
     };
+
+    if (!isEmpty(jestAnimatedStyleValue as object)) {
```
tjzel, line 89: Afaik spread operator on empty objects is fine, is it needed?

packages/react-native-reanimated/src/jestUtils.ts
```diff
@@ -27,22 +27,81 @@ const defaultFramerateConfig = {
...
 };
 
+const isEmpty = (obj: object) => Object.keys(obj).length === 0;
+const getStylesFromObject = (obj: object) => {
+  return obj === undefined
+    ? {}
+    : Object.fromEntries(
+        Object.entries(obj).map(([property, value]) => [
+          property,
+          value.value ?? value,
```
tjzel, line 37: maybe `value._isReanimatedSharedValue ? value.value : value` instead?

packages/react-native-reanimated/src/jestUtils.ts
```diff
@@ -27,22 +27,81 @@ const defaultFramerateConfig = {
...
+        ])
+      );
+};
+
+type StyleValue = { value: unknown };
+type JestInlineStylesType =
+  | {
+      [s: string]: StyleValue;
+    }
+  | ArrayLike<StyleValue>;
```
tjzel, line 47:
```suggestion
type JestInlineStyle =
  | {
      [s: string]: StyleValue;
    }
  | ArrayLike<StyleValue>;
```

packages/react-native-reanimated/src/jestUtils.ts
```diff
@@ -27,22 +27,81 @@ const defaultFramerateConfig = {
...
+
   let currentStyle = {};
+
   if (Array.isArray(styleObject)) {
     styleObject.forEach((style) => {
       currentStyle = {
         ...currentStyle,
         ...style,
       };
     });
```
tjzel, line 60: What in case of nested array? i.e.

`style={[styleA, styleB, [animatedStyle], styleWithInlineStyle]}`

packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
tjzel: Also what about non-inline jest style? It would be more maintainable if it was used here as well.

packages/react-native-reanimated/src/jestUtils.ts
```diff
@@ -27,22 +27,81 @@ const defaultFramerateConfig = {
...
-  } else {
+
+    return currentStyle;
+  }
+
+  const jestInlineStyles = component.props
```
tjzel, line 65: No update to the type of AnimatedComponent?

packages/react-native-reanimated/__tests__/inline.test.tsx
tjzel, line 1: rename it to `inlineStyles` please

tjzel (changes requested):

packages/react-native-reanimated/src/jestUtils.ts
m-bert: Using `flattenArray` doesn't solve this problem and I'm not sure if we want to dive into recursion to solve it (at least for now)

packages/react-native-reanimated/src/jestUtils.ts
m-bert: Changed in [19bae71](https://github.com/software-mansion/react-native-reanimated/pull/6400/commits/19bae71cfcf30605d8f720542886999ae8108803)

packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
m-bert: Probably we won't, changed in [e147d2b](https://github.com/software-mansion/react-native-reanimated/pull/6400/commits/e147d2b0eb362ac0a51663dbe9dce8b791a027f5), along with adding prop for non-inline styles

yarn.lock
m-bert: I had to bump this version, otherwise `screen` was not defined. Lmk if you prefer lower version 😝

packages/react-native-reanimated/src/jestUtils.ts
tjzel: Let's annotate it with a relevant comment then, so we see in the future that such cases aren't handled on purpose.

tjzel (approved):
👏

piaskowyk (approved):
looks good 👍

PR #6403, opened 2024-08-11
[Web] Add `findNodeHandle` web version

m-bert:
## Summary

In many places we use `findNodeHandle` function to get native view that is represented by React component. The problem is that usage of this function on `web` leads to warnings (which are shown via `console.error`), that this function in deprecated and should not be used.

I've already done 3 PRs that remove usage of `findNodeHandle` on `web`, but it keeps reappearing. That's why I've decided to create our version of this function. On native platforms it is just export from `react-native`. On `web` it returns component that is passed into the function.

>[!CAUTION]
>Note that `findNodeHandle`(or to be exact, `findDOMNode`) will be [removed in next major version of React](https://react.dev/reference/react-dom/findDOMNode) (i.e. React 19)

## Test plan

Verifed on our web-example that it uses web version of `findNodeHandle`. 

>[!NOTE]
>In our example app you can still see depracation warning. This is because of `GestureDetector` from `react-native-gesture-handler` and it has to be fixed in mentioned library.

### DIFF
--- a/packages/react-native-reanimated/src/component/LayoutAnimationConfig.tsx
+++ b/packages/react-native-reanimated/src/component/LayoutAnimationConfig.tsx
@@ -8,7 +8,7 @@ import React, {
 } from 'react';
 import type { ReactNode } from 'react';
 import { setShouldAnimateExitingForTag } from '../core';
-import { findNodeHandle } from 'react-native';
+import { findNodeHandle } from '../platformFunctions/findNodeHandle';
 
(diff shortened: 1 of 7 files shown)


tjzel (approved):
Solid work!

piaskowyk (approved):

gaearon:
`findNodeHandle` is just as bad on React Native. See https://github.com/software-mansion/react-native-reanimated/issues/6719.

PR #6409, opened 2024-08-12
feat: User-defined logger config

MatiPl01:
## Summary

This PR adds a possibility to customize log level via the `level` property (limit to just see errors or show errors and warnings) and enable/disable strict mode (via `strict` property), which will be useful for changes implemented in [this](https://github.com/software-mansion/react-native-reanimated/pull/6310) PR to show invalid `.value` access warnings only in the strict mode.

## Example recording

https://github.com/user-attachments/assets/232d3b69-d99c-4114-989f-967d9d8c9e22

## Test plan

- Open `EmptyExample` (which will be reverted to the original implementation before merging this PR),
- Change logger settings (e.g. the `level` property) via `Animated.configureLogger`,


packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -55,27 +66,59 @@ export function logToLogBoxAndConsole(data: LogData) {
...
+export function updateLoggerConfig(config?: Partial<LoggerConfigInternal>) {
+  'worklet';
+  global.__loggerConfig = {
+    logFunction:
+      // Re-use previously assigned log function if it exists
+      config?.logFunction ?? global.__loggerConfig?.logFunction ?? logToConsole,
+    // Don't reuse previous level and strict values from the global config
+    level: config?.level ?? 'warn',
+    strict: config?.strict ?? false,
+  };
```
tjzel, line 86: Why global scope? 👀

packages/react-native-reanimated/src/Animated.ts
```diff
@@ -27,6 +27,7 @@ export { ReanimatedFlatList as FlatList } from './component/FlatList';
 export {
   addWhitelistedNativeProps,
   addWhitelistedUIProps,
+  configureLogger,
```
tjzel, line 30: Please don't add stuff to `Animated` 😭

packages/react-native-reanimated/src/ConfigHelper.ts
```diff
@@ -52,6 +54,11 @@ export function addWhitelistedUIProps(props: Record<string, boolean>): void {
   }
 }
 
+export function configureLogger(config: LoggerConfig) {
+  updateLoggerConfig(config);
+  executeOnUIRuntimeSync(updateLoggerConfig)(config);
+}
```
tjzel, line 60: What about other runtimes? Maybe we should instead do `configureLoggerOnJS`, `configureLoggerOnUI` for now?

packages/react-native-reanimated/jest-setup.js
```diff
@@ -1,2 +1,19 @@
...
+      case 'error':
+      case 'fatal':
+      case 'syntax':
+        console.error(data.message.content);
+        break;
+    }
+  },
+  level: 'warn',
+  strict: false,
+}
```
tjzel, line 19: is it really necessary?

packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -55,27 +66,59 @@ export function logToLogBoxAndConsole(data: LogData) {
...
+}
+
+const logLevelImportance = {
+  warn: 1,
+  error: 2,
+  fatal: 3,
+} as const;
```
tjzel, line 93: Let's turn this into an enum.

packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -55,27 +66,59 @@ export function logToLogBoxAndConsole(data: LogData) {
...
+  }
+  config.logFunction(createLog(level, message));
 }
 
 export const logger = {
-  warn(message: string) {
+  warn(message: string, options: LogOptions = {}) {
     'worklet';
-    loggerImpl.logFunction(createLog('warn', message));
+    handleLog('warn', message, options);
```
tjzel, line 114: I assume the options parameter got added to allow for dynamically silencing/escalating warnings etc., is it correct?

packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -55,27 +66,59 @@ export function logToLogBoxAndConsole(data: LogData) {
...
+  options: LogOptions
 ) {
-  loggerImpl.logFunction = logFunction;
+  'worklet';
+  const config = global.__loggerConfig;
+  if (
+    (options.strict && !config.strict) ||
+    logLevelImportance[level] < logLevelImportance[config.level]
+  ) {
+    return;
```
tjzel, line 106: What does this do?

packages/react-native-reanimated/src/logger/logger.ts
MatiPl01: This was global for the same reason as the `ReanimatedError` in the first PR. Because it needs to be global in the Worklet runtime, I made it global in the React runtime as well.

I will adjust changes based on your recent suggestions I already applied to previous PRs.

packages/react-native-reanimated/src/Animated.ts
MatiPl01: Mhm. I thought it's similar to calling `Animated.addWhitelistedNativeProps` in terms of behavior. In both cases, you change the configuration of something so I wanted to keep a similar approach.

I can export it as a standalone function renamed to something more meaningful like `configureReanimatedLogger` or something similar if you prefer.

packages/react-native-reanimated/src/ConfigHelper.ts
MatiPl01: You mean we would want to have a separate configuration for both runtimes? I think it will be a bit confusing and, by default, the user would expect to change the configuration for all runtimes (e.g. the user doesn't have to know if the error/warning comes from the UI or JS runtime and it will be hard to predict if the change made to the config will work as intended).

I can agree that current configuration options are a bit limited and we might think of adding more options. Having just a `level` and `strict` fields don't provide enough configuration options.

packages/react-native-reanimated/jest-setup.js
MatiPl01: If the `__loggerConfig` is global, yes it is necessary. `__loggerConfig` is not initialized in jest as the file in which we register `__loggerConfig` in globals is not executed.

When I change `__loggerConfig` to use import/export way as I made with `ReanimatedError` it should no longer be necessary.

packages/react-native-reanimated/src/logger/logger.ts
MatiPl01: currently it is only used to mark specific logs as **strict-only**. I don't know which logs will be suitable to convert to strict mode only but probably these that are most annoying (like overriding layout animation with animated style which is often done on purpose or my [PR](https://github.com/software-mansion/react-native-reanimated/pull/6310) that should be merged with invalid shared value read/write).

See `LogOptions` type, currently it can take just one option.
```tsx
type LogOptions = {
  strict?: boolean;
};
```

packages/react-native-reanimated/src/ConfigHelper.ts
MatiPl01: Ok, it makes sense now. Yeah, this might be problematic. I didn't have other runtimes in my mind.

packages/react-native-reanimated/src/logger/logger.ts
tjzel: I think you already explained this to me once but I forgot 😅

packages/react-native-reanimated/src/Animated.ts
tjzel: The Animated namespace was added to make the migration from React Native's Animated API to Reanimated API. At the moment it's (in my opinion) unnecessary and I'd like to see it removed in Reanimated 4 😅

packages/react-native-reanimated/src/logger/logger.ts
MatiPl01: I will add some comments in there because it might be a bit confusing at the first glance.

packages/react-native-reanimated/src/runtimes.ts
```diff
@@ -35,11 +36,16 @@ export function createWorkletRuntime(
...
+  // Assign to a different variable as __reanimatedLoggerConfig is not captured
+  // identifier on the Worklet runtime.
+  const config = __reanimatedLoggerConfig;
+
   return NativeReanimatedModule.createWorkletRuntime(
     name,
     makeShareableCloneRecursive(() => {
       'worklet';
       registerReanimatedError();
+      registerLoggerConfig(config);
```
MatiPl01, line 48: I think this will be enough. The `configureReanimatedLogger` function should be called called in the global react scope before calling `createWorkletRuntime` so changes made to the config will be also visible in other runtimes when they are created.

We can add to docs information how to use `configureReanimatedLogger` properly (call it before using anything else from reanimated).

packages/react-native-reanimated/src/runtimes.ts
```diff
@@ -35,11 +36,16 @@ export function createWorkletRuntime(
   name: string,
   initializer?: WorkletFunction<[], void>
 ): WorkletRuntime {
+  // Assign to a different variable as __reanimatedLoggerConfig is not captured
```
tjzel, line 39:
```suggestion
  // Assign to a different variable as __reanimatedLoggerConfig is not a captured
```

packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -55,27 +78,84 @@ export function logToLogBoxAndConsole(data: LogData) {
...
+export function updateLoggerConfig(options?: Partial<LoggerConfigInternal>) {
+  'worklet';
+  // Clone the config in the worklet runtime thus the object sent to
+  // the worklet runtime is immutable
+  const config = _WORKLET
+    ? {
+        ...(((global as Record<string, unknown>)
+          .__reanimatedLoggerConfig as LoggerConfigInternal) ?? {}),
+      }
+    : __reanimatedLoggerConfig;
```
tjzel, line 114: I don't get it, could you elaborate?

packages/react-native-reanimated/src/initializers.ts
```diff
@@ -1,18 +1,50 @@
...
+if (SHOULD_BE_USE_WEB) {
+  global._WORKLET = false;
+  global._log = console.log;
+  global._getAnimationTimestamp = () => performance.now();
+}
+
+// Register ReanimatedError and update logger config in the UI global scope.
+// (we are using `executeOnUIRuntimeSync` here to make sure that the changes
+// are applied before any async operations are executed on the UI runtime)
+if (!shouldBeUseWeb()) {
```
tjzel, line 30:
```suggestion
if (SHOULD_BE_USE_WEB) {
  global._WORKLET = false;
  global._log = console.log;
  global._getAnimationTimestamp = () => performance.now();
} else {
// Register ReanimatedError and update logger config in the UI global scope.
// (we are using `executeOnUIRuntimeSync` here to make sure that the changes
// are applied before any async operations are executed on the UI runtime)
```

packages/react-native-reanimated/jest-setup.js
tjzel: Is it still necessary? 😅

packages/react-native-reanimated/src/initializers.ts
```diff
@@ -1,18 +1,50 @@
...
+// Override the logFunction implementation with the one that adds logs
+// with better stack traces to the LogBox (need to override it after `runOnJS`
+// is defined).
+updateLoggerConfig({
+  logFunction(data) {
+    'worklet';
+    runOnJS(logToLogBoxAndConsole)(data);
+  },
+});
+shareableMappingCache.set(logger, makeShareableCloneRecursive(logger));
```
tjzel, line 46: I don't understand. You call `updateLoggerConfig` via `executeOnUIRuntimeSync` and then you call it again without it. Could you explain?

packages/react-native-reanimated/src/logger/logger.ts
MatiPl01: Well, I cannot just assign `__reanimatedLoggerConfig` to the `context` variable as `__reanimatedLoggerConfig` is not assigned to `global` variable on the Worklet runtime yet, thus, trying to access `__reanimatedLoggerConfig` in the same way as I do in the `handleLog` (`const config = __reanimatedLoggerConfig;`) will crashes the app.

The result was to access it via the `global` variable, which is defined. We will get `undefined` for the `global.__reanimatedLoggerConfig` if `__reanimatedLoggerConfig` hadn't been registered in the global scope, so in that case, I am returning the empty object as a fallback, which will be filled in below in this function.

Destructuring is no longer necessary and I forgot to remove that, so this implementation can be simplified to:

```tsx
  const config = _WORKLET
    ? ((global as Record<string, unknown>)
        .__reanimatedLoggerConfig as LoggerConfigInternal) ?? {}
    : __reanimatedLoggerConfig;
```

packages/react-native-reanimated/src/initializers.ts
MatiPl01: ### The first call

```tsx
executeOnUIRuntimeSync(updateLoggerConfig)();
```

Is needed to create the logger config object on the UI runtime. What `updateLoggerConfig` does without any parameter when called for the first time, before the `__reanimatedLoggerConfig` exists in the UI runtime, is creating the config object and registering it in the global scope:

```tsx
export function updateLoggerConfig(options?: Partial<LoggerConfigInternal>) {
  'worklet';
  const config = _WORKLET
    ? ((global as Record<string, unknown>)
        .__reanimatedLoggerConfig as LoggerConfigInternal) ?? {} // <-- this empty object will be returned
    : __reanimatedLoggerConfig;

  // <-- will be filled with default values from below
  config.logFunction =
    config?.logFunction ?? options?.logFunction ?? logToConsole; // <-- `logToConsole`
  config.level = options?.level ?? LogLevel.warn; // <-- `LogLevel.warn`
  config.strict = options?.strict ?? false; // <-- `false`

  if (_WORKLET) { // <-- Because it is called with `executeOnUIRuntimeSync`, it is a worklet
    registerLoggerConfig(config); // <-- and will store the config object in the UI runtime global scope
  }
}
```

---

### The second call

It is executed on the React runtime and it replaces the `logFunction` in the config.

```tsx
updateLoggerConfig({
  logFunction(data) {
    'worklet';
    runOnJS(logToLogBoxAndConsole)(data);
  },
});
```

MatiPl01: But, I can see now that this is invalid as `logFunction` is always the default implementation then instead of the `logToLogBoxAndConsole`

tjzel (approved):
All good now!

PR #6412, opened 2024-08-13
Fix github workflows

Latropos:
## Summary

1. **Fix paths**
   - Workflow files have been renamed, but the paths inside the workflows weren't updated
   - Due to this mismatch the workflow is not triggered once it is edited
4. **Fix caching**
   - Remove caching when building apps (it wasn't working well)

## Test plan
There are failing tests, but their fail in not depending on this PR
image
image

### DIFF
--- a/.github/workflows/example-android-build-check.yml
+++ b/.github/workflows/example-android-build-check.yml
@@ -18,7 +18,7 @@ on:
     branches:
       - main
     paths:
-      - .github/workflows//example-android-build-check.yml
+      - .github/workflows/example-android-build-check.yml
       - packages/react-native-reanimated/android/**
       - packages/react-native-reanimated/Common/**
       - apps/paper-example/package.json
--- a/.github/workflows/example-android-build-check.yml
+++ b/.github/workflows/example-android-build-check.yml
@@ -47,6 +47,7 @@ jobs:
           distribution: 'zulu'
           java-version: 17
 
+      # TODO: Add caching for node_modules and artifacts that will work with monorepo setup.
       - name: Install monorepo node dependencies
         run: yarn install --immutable
 
--- a/.github/workflows/example-android-build-check.yml
+++ b/.github/workflows/example-android-build-check.yml
@@ -47,6 +47,9 @@ jobs:
           distribution: 'zulu'
           java-version: 17
 
+      - name: Install monorepo node dependencies
+        run: yarn install --immutable
+
       - name: Build app
         working-directory: ${{ matrix.working-directory }}/android
         run: ./gradlew assembleDebug --build-cache -PreactNativeArchitectures=arm64-v8a
--- a/.github/workflows/example-android-build-check.yml
+++ b/.github/workflows/example-android-build-check.yml
@@ -4,7 +4,7 @@ env:
 on:
   pull_request:
     paths:
-      - .github/workflows/android-build.yml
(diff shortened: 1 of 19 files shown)


.github/workflows/example-android-build-check.yml
```diff
@@ -18,7 +18,7 @@ on:
     branches:
       - main
     paths:
-      - .github/workflows/android-build.yml
+      - .github/workflows//example-android-build-check.yml
```
tjzel, line 21:
```suggestion
      - .github/workflows/example-android-build-check.yml
```

.github/workflows/example-ios-build-check.yml
```diff
@@ -1,10 +1,10 @@
-name: Example iOS build check
+name: Example app, iOS build check (paper and fabric)
```
tjzel, line 1: Please revert this change, I made the workflow names match their filename for convenience.

.github/workflows/example-ios-build-check.yml
```diff
@@ -45,43 +45,15 @@ jobs:
       - name: Check out Git repository
         uses: actions/checkout@v4
 
-      # TODO: Rework caching in this CI for monorepo setup.
```
tjzel, line 48: Let's keep such comment here instead:

`# TODO: Add caching for node_modules and artifacts that will work with monorepo setup.`

.github/workflows/example-tvos-build-check.yml
```diff
@@ -42,51 +43,20 @@ jobs:
       - name: Install monorepo node dependencies
         run: yarn install --immutable
 
-      - name: Restore Reanimated node_modules from cache
```
tjzel, line 45: Let's keep such comment here instead:

`# TODO: Add caching for node_modules and artifacts that will work with monorepo setup.`

.github/workflows/example-android-build-check.yml
```diff
@@ -47,19 +47,6 @@ jobs:
           distribution: 'zulu'
           java-version: 17
 
-      # TODO: Rework caching in this CI for monorepo setup.
```
tjzel, line 50: Let's keep such comment here instead:

`# TODO: Add caching for node_modules and artifacts that will work with monorepo setup.`

tjzel (changes requested):

.github/workflows/example-android-build-check.yml
```diff
@@ -47,19 +47,6 @@ jobs:
...
-          path: .
-          key: ${{ runner.os }}-node-modules-${{ hashFiles('yarn.lock') }}
-          restore-keys: ${{ runner.os }}-node-modules-
-      - name: Restore app node_modules from cache
-        uses: actions/cache@v3
-        with:
-          path: ${{ matrix.working-directory }}/node_modules
-          key: ${{ runner.os }}-node-modules-${{ matrix.working-directory }}-${{ hashFiles(format('{0}/yarn.lock', matrix.working-directory)) }}
-          restore-keys: ${{ runner.os }}-node-modules-${{ matrix.working-directory }}-
       - name: Install monorepo node dependencies
```
Latropos, line 50:
```suggestion
      # TODO: Add caching for node_modules and artifacts that will work with monorepo setup.
      - name: Install monorepo node dependencies
```

PR #6423, opened 2024-08-17
chore: Pretty pre-commit

tjzel:
## Summary

Making `pre-commit` script more readable, improving the overall quality of precommit checks and some random repo improvements.

## Test plan

🚀

PR #6424, opened 2024-08-17
chore: Improve build pipeline of Reanimated Babel plugin

tjzel:
## Summary

Simplifying the workspace structure, minifying the output file and more.

Requires:
- #6423 

## Test plan

- [ ] CI passes.


packages/react-native-reanimated/plugin/.gitignore
```diff
@@ -0,0 +1,2 @@
+lib
```
Latropos, line 1: I am not sure if we shouuld add it to top level .gitignore instead.

Latropos (approved):

packages/react-native-reanimated/plugin/.gitignore
tjzel: This is plugin's .gitignore file 😅

PR #6430, opened 2024-08-19
feat: Enable Clangd in repo

tjzel:
## Summary

Bringing `clangd` to our monorepo.

### What's clangd?

It's a language server that plugs into your text editor (yes, I'm looking at you VSCode) and provides you with syntax highlighting, autocompletion, inspections and much more. You can read about it on [clangd's webpage](https://clangd.llvm.org/).

### Why do we need this?

Often times when we develop C++ code, we have to jump between Xcode and Android Studio just to get autosuggestions or be able to inspect some functions. This slows down development and these IDEs have their downsides.

With this change, you'll be able to edit the code in VSCode and have all the necessary information there, including static checks. In my experience it also works much faster than native IDEs.

Just to name a few useful features:
- CMD+click leads you to source
- Macros are resolved and you can see which branch of the code is active and which isn't
- It parses objective C 🤯

### How do I use it?

I automated the whole pipeline for clangd, so __all you need to do is install VSCode's clangd extension__.

You also need to __compile the project once__ to generate all the artifacts utilized by clangd. For example, if you are developing for Fabric, iOS, just build FabricExample.

As a bonus, built-in VSCode's C++ plugin also benefits from this, so you don't need to use clangd at all. However, I still recommend clangd as a replacement. Try it out yourself!

Before:
Screenshot 2024-08-21 at 18 26 23

After:
Screenshot 2024-08-21 at 18 15 42

## Test plan

🚀


packages/react-native-reanimated/android/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,2 @@
 
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.swmansion.reanimated">
-
-</manifest>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"/>
```
tjzel, line 2: Gradle yells at me that it's deprecated and unused to define a namespace here.

piaskowyk (approved):
Looks good to me 🫡 

I want to use IT!

packages/react-native-reanimated/scripts/clangd-add-xcode-step.rb
```diff
@@ -0,0 +1,5 @@
+def add_clangd_generation_step()
+  generator_script_path = '../../../packages/react-native-reanimated/scripts/clangd-generate-xcode-metadata.sh'
+
+  script_phase :name => 'Generate metadata for clangd', :script => generator_script_path, :shell_path => '/bin/bash', :always_out_of_date => 'true', :execution_position => :after_compile
+end
```
WoLewicki, line 5: no newline at the end of file, not very nice

tjzel: Contributions are welcome 🤗

Issue #6432, opened 2024-08-20
debuggerWorker.js:69 TypeError: Cannot convert undefined or null to object


TonyFTannous-Byd:
### Description

![352525954-22331002-5e09-405b-a1a4-d439aec6f2e6](https://github.com/user-attachments/assets/a6e26434-e90d-4faf-bae3-da14b0b81f54)

### Steps to reproduce

i use react native 0.75.1, when launching react native debugger, this error displayed and the app crashed

### Snack or a link to a repository

https://github.com/software-mansion/react-native-reanimated

### Reanimated version

^3.15.0

### React Native version

0.75.1

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

MatiPl01:
Hey!

Could you pleas provide more information? Did you use the new (experimental) react native debugger, chrome debugger or some other debugging tool?

Can you provide a repro or share the repository?

TonyFTannous-Byd:
i user react native debugger [https://github.com/jhen0409/react-native-debugger](url) because it support debug request network.
i use react native 0.75.1, and i need to debug my request network. so i used react native debugger with enabled network inspect. because i use @react-navigation/drawer which it use react-native-reanimated 3.15.0. an error occured: Cannot convert undefined or null to object located in /node_modules/react-native-reanimated/src/js-reanimated/index.ts --> _updatePropsJS --> component.props

MatiPl01:
Thank you for response. I can reproduce the issue while I am debugging JS remotely with `react-native-debugger`. I am not sure it we will able to fix that, though, as this debugger doesn't support JSI, which reanimated uses.

From the `react-native-debugger` README:

> ⚠️ This app is currently only supported old [Remote Debugger](https://reactnative.dev/docs/debugging#chrome-developer-tools), if you're looking for new debugger support (e.g. Hermes / JSI / New Architecture) of React Native Debugger, please follow https://github.com/jhen0409/react-native-debugger/discussions/774.

I will look into this issue and check it we can do something.

MatiPl01:
@TonyFTannous-Byd Could you check if changes I made in the #6437 PR fix your issue? You can make the same changes in `node_modules/react-native-reanimated` and see if you still see this error.

If you still have some problems, don't hesitate to re-open the issue and ask.

abdulmoizshaikh:
[...]

PR #6437, opened 2024-08-21
fix: Chrome debugger crash caused by incorrect viewTag

MatiPl01:
## Summary

The problem was reported in [this](https://github.com/software-mansion/react-native-reanimated/issues/6432) issue and [this](https://github.com/software-mansion/react-native-reanimated/discussions/6359) discussion seems to report the same problem.

In short, the crash was caused by incorrect check in `_getViewInfo` in `createAnimatedComponent` which assigned the `_component` to the `viewTag` only for web but not for chrome debugger. I replaced `IS_WEB` check with `shouldBeUseWeb`, which also checks whether the current environment is a chrome debugger.

## Test plan

- I created a bare app with `npx @react-native-community/cli init ReactNativeDebugger --version 0.75.1`
- Installed latest reanimated and downloaded [react-native-debugger](https://github.com/jhen0409/react-native-debugger?tab=readme-ov-file) mentioned in [this](https://github.com/software-mansion/react-native-reanimated/issues/6432) issue
- Launched `react-native-debugger` and enabled remote debugging with js,
- When I tried to animate component with animated style, I saw the reported crash

### DIFF
--- a/packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
+++ b/packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
@@ -289,7 +289,7 @@ export function createAnimatedComponent(
         ? (this._component as AnimatedComponentRef).getAnimatableRef?.()
         : this;
 
-      if (IS_WEB) {
+      if (shouldBeUseWeb()) {
         // At this point I assume that `_setComponentRef` was already called and `_component` is set.
         // `this._component` on web represents HTMLElement of our component, that's why we use casting
         viewTag = this._component as HTMLElement;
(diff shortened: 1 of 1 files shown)


packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -289,7 +289,7 @@ export function createAnimatedComponent(
         ? (this._component as AnimatedComponentRef).getAnimatableRef?.()
         : this;
 
-      if (IS_WEB) {
+      if (shouldBeUseWeb()) {
```
MatiPl01, line 292: Not sure if some other places where we use `!IS_WEB`/`IS_WEB` checks should be updated as well. This change fixes the problem reported in the linked issue but I am not sure about other places where we use web check instead of `shouldBeUseWeb`.

piaskowyk (approved):
SHOULD_BE_OK_🫡

PR #6439, opened 2024-08-22
fix: add missing include

okwasniewski:
## Summary

This PR adds missing import for the header file. 

This was breaking visionOS builds. 

The `CFTimeInterval` comes from Core Foundation so we need to have foundation imported to make this work. 

## Test plan

CI Green


piaskowyk (approved):
Thanks for the PR! 😊

PR #6441, opened 2024-08-22
Fix getViewProp on Android, RN.74

Latropos:
## Summary

We can't import CSSBackgroundDrawable on Reanimated lower than 75, so I had to change the affected code not to require it.


packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NodesManager.java
```diff
@@ -456,25 +455,17 @@ public String obtainProp(int viewTag, String propName) {
...
 
-        // ReactViewBackgroundDrawable got deprecated in react-native 0.75.
-        //noinspection deprecation
-        if (background instanceof ReactViewBackgroundDrawable) {
-          //noinspection deprecation
-          actualColor = ((ReactViewBackgroundDrawable) background).getColor();
-        }
+        try {
+          Method getColor = background.getClass().getMethod("getColor");
+          actualColor = (int) getColor.invoke(background);
```
piaskowyk, line 460: It can be scope variable
```suggestion
          int actualColor = (int) getColor.invoke(background);
```

piaskowyk (commented):
The `getViewProps` changes appear good, but why do we need to update the ESLint plugin in the same PR? 🤔 Should we revert the plugin changes from this PR?

piaskowyk (approved):

PR #6442, opened 2024-08-22
Change `requestAnimationFrame` flush condition

### LINKED ISSUE #6371: Reanimated Stops Working Completely on iOS
### Description

Reanimated completely stops working after a while on production on iOS. I've attached a video and a minimum reproducible repo.

Some thoughts:
- This bug occurs 100 out of 100 times with Video but if you remove it, it only happens around 1 out of 100 times.
- Occurs with both react-native-video and expo-av.
- Once the bug occurs everything that touches reanimated stops working but everything else including the native Animated API works fine. 
- Only occurs on iOS on release mode on actual device.

https://github.com/user-attachments/assets/beb2353c-1e71-48c9-a118-d8e4d4366b2d

### Steps to reproduce

Clone https://github.com/phantom-xrn/reanimated-bug/tree/main

0. Run it on release mode on an 120hz iOS device(Replicated on 14 pro/15 pro/15 pro max but not on 12 mini/13)
1. Wait for the video to start playing.
2. Press and Long Press on the bottom two cards a few times.
3. Reanimated will stop working after a while. The buttons will not scale anymore when pressed and any component that uses reanimated like bottom sheets stop working.
4. Only error on Xcode console is `Unbalanced calls start/end for tag 19` 

https://github.com/phantom-xrn/reanimated-bug/tree/main

### Reanimated version

3.14.0

### React Native version

0.73.6

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Paper (Old Architecture)

### Build type

### Device

Real device

### Device model

Iphone 15 Pro

### Acknowledgements

Yes

bartlomiejbloniarz:
## Summary
Currently we schedule`rAF` flush when the first callback is added to the list. However the flush method can also be called by an event, meaning that sometimes we have a flush scheduled to run on a given frame, but the callbacks array is empty. If then, another callback is requested, the array will contain 1 element, triggering another flush request (even though one is already scheduled). To prevent this, from causing countless requests on a singe frame we check the frame timestamp and abort if it's repeated.

This approach unfortunately causes some problems when video is playing in the app. On iPhone devices sometimes the displayLink can fire its callback twice in a single frame (with the same timestamp). This leads to us cancelling the `rAF` flush, which means that until an event triggers a flush, none updates from reanimated will come through.

This PR changes the way we request a flush. Instead of checking the callbacks array size, we instead remember whether a flush was requested and request a new one only when there was no request. This prevents us from requesting unnecessary flushes when an event has caused the callbacks array to be emptied, while also allowing for repeated frame timestamps.

closes #6371 

## Test plan

Check for regressions in the example app and in this example:
https://gist.github.com/kmagiera/b2df85f9512951f5e6ceee7bc569f5f1


kmagiera (approved):
This looks good. The only thing I'm wondering now is that when duplicated rAFs come at the same timestamp, we will run all the mappers twice with the same input, which may not be as efficient as needed, especially given it happens for example when video is being played (relatively heavy task as well). Normally, I wouldn't be concerned about running mappers twice in a frame from time to time, but it it happens continuously for the whole duration of the video, I'd love if we try to measure the consequences here and at least add a comment about that in the PR description.

bartlomiejbloniarz:
In my testing it wasn't a continuous issue, it only happened from time to time. Sometimes I had to try for a few minutes to get it to break.

tjzel (approved):
:shipit:

piaskowyk (approved):

PR #6445, opened 2024-08-23
fix: Recursion in worklets

tjzel:
## Summary

Turn out that in #6163 I overlooked the fact that we shouldn't rename React part of worklet functions. This PR changes the naming behavior, only worklet counterparts are renamed for better error verbosity. See differences in snapshot for more details.

## Test plan

- [x] Upgraded jest unit tests pass
- [x] New runtime tests pass


packages/react-native-reanimated/plugin/src/jestMatchers.ts
```diff
@@ -68,27 +69,49 @@ expect.extend({
     };
   },
 
-  toIncludeInWorkletString(received: string, expected: string) {
-    // Regular expression pattern to find the `code` field in `initData`.
-    // @ts-ignore This regex works well in Jest.
-    const pattern = /code: "((?:[^"\\]|\\.)*)"/gs;
-    const matches = received.match(pattern);
+  toContainInWorkletString(received: string, expected: string) {
```
tjzel, line 72: Renamed `toInclude` -> `toContain` and added `toMatch` to align with Jest names.

packages/react-native-reanimated/plugin/src/utils.ts
```diff
@@ -60,7 +60,7 @@ export function addCustomGlobals(this: ReanimatedPluginPass) {
  * becomes
  *
  * ```ts
- *   const foo = factoryCall();
+ *   export const foo = factoryCall();
```
tjzel, line 63: Overlooked previously.

packages/react-native-reanimated/plugin/src/workletFactory.ts
```diff
@@ -219,12 +218,12 @@ export function makeWorkletFactory(
     VariableDeclaration | ExpressionStatement | ReturnStatement
   > = [
     variableDeclaration('const', [
-      variableDeclarator(functionIdentifier, funExpression),
```
tjzel, line 222: It's safer to create a new identifier node every time.

apps/common-app/src/examples/RuntimeTests/tests/plugin/recursion.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+  test('from Worklet to React runtime', async () => {
+    const ExampleComponent = () => {
+      const output = useSharedValue<number | null>(null);
+      registerValue(SHARED_VALUE_REF, output);
+      function recursiveWorklet(a: number) {
+        if (a === 2) {
+          output.value = a;
+        } else if (a === 1) {
+          try {
+            // TODO: Such case isn't supported at the moment -
```
tjzel, line 92: WDYT?

bartlomiejbloniarz (approved):

PR #6454, opened 2024-08-27
chore: Enforce the use of ReanimatedError instead of Error

MatiPl01:
## Summary

This PR adds eslint rule that will enforce the use of the new `ReanimatedError` in favor of `Error` in the library source files.

## Example image

Screenshot 2024-08-28 at 17 25 11

## Test plan

Just try to use `throw new Error` inside the `packages/react-native-reanimated` instead of `throw new ReanimatedError` and see that eslint complains about incorrect usage.


tjzel (commented):
Looks good, let's just work on some of these TODOs first.

I recall having some issues with Jest hanging on ESLint unit tests. From what I remember I wasn't able to solve it back in the day though. You can run Jest from the JavaScript debug terminal and, if you have source maps generated, the breakpoints will work - hope this piece of info is going to be helpful 🤞

Latropos (approved):
LGTM, thank you!

MatiPl01:
> Looks good, let's just work on some of these TODOs first.
> 
> I recall having some issues with Jest hanging on ESLint unit tests. From what I remember I wasn't able to solve it back in the day though. You can run Jest from the JavaScript debug terminal and, if you have source maps generated, the breakpoints will work - hope this piece of info is going to be helpful 🤞

Thanks for suggestion about fixing tests. I spent some time looking into a potential cause of the issue, tried upgrading `@typescript-eslint/*` dependencies and nothing helps. When I was digging in the source code of `@typescript-eslint/rule-tester` I noticed that it hangs no the import of the `@typescript-eslint/parser` dependency and `@typescript-eslint/typescript-estree`.

I will merge this PR for now and look into the issue later on, since it seems to be better to solve it separately from this PR that just adds a new rule and doesn't break test which already don't work.

PR #6460, opened 2024-08-28
Fix handling of layout animations coinciding with view flattening

bartlomiejbloniarz:
## Summary

This PR fixes the way a reparenting of views is handled in Layout Animations. It can so happen that a view is being removed and in the same transaction its parent is being moved (due to view flattening). The current implementation in this scenario would move the parent correctly and remove the `MutationNode` from our registry. However, the pointer to this `MutationNode` was still stored as a parent by the child node. This would lead to us trying to remove the previously moved node when the child animation ended. This would lead to a crash, since we tried to remove the view from its old parent.

## Test plan

Check the `[LA] View flattening` example, and other examples for regressions.


tjzel:
Please catch me offline so you could explain this fix to me in person 🙏

piaskowyk (approved):

PR #6467, opened 2024-08-30
docs: Add metro config wrapper and reanimated logger docs

MatiPl01:


apps/common-app/src/examples/InvalidValueAccessExample.tsx
```diff
@@ -9,8 +9,7 @@ import Animated, {
 } from 'react-native-reanimated';
 
 configureReanimatedLogger({
-  // change to `false` or remove the `configureReanimatedLogger` call to
-  // disable the warning
+  // change to `false` to disable the warning
   strict: true,
```
MatiPl01, line 13: I decided to make logger `strict` by default. I think that nobody will enable strict mode as nobody reads docs until something doesn't work, thus no one will see my new invalid value read/write warning from the #6310 PR.

packages/react-native-reanimated/src/ConfigHelper.ts
```diff
@@ -58,7 +61,9 @@ export function configureReanimatedLogger(config: LoggerConfig) {
   // Update the configuration object in the React runtime
   updateLoggerConfig(config);
   // Register the updated configuration in the UI runtime
-  executeOnUIRuntimeSync(updateLoggerConfig)(config);
+  if (!SHOULD_BE_USE_WEB) {
+    executeOnUIRuntimeSync(updateLoggerConfig)(config);
+  }
```
MatiPl01, line 66: I tried to add an interactive playground and was getting crash in there. After that I decided that interactive example makes no much sense in this case as logs on web are logged to the console and many people may assume that this example because they won't see any warnings/errors until they open dev tools.

At least, thanks to this I was able to find the issue and fix it.

packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -36,7 +40,7 @@ function logToConsole(data: LogData) {
 export const DEFAULT_LOGGER_CONFIG: LoggerConfigInternal = {
   logFunction: logToConsole,
   level: LogLevel.warn,
-  strict: false,
+  strict: true,
```
MatiPl01, line 43: I think that we'd better enable strict mode by default and tell users how to disable it if warnings are too annoying.

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
```diff
@@ -0,0 +1,39 @@
...
+id: accurate-call-stacks
+sidebar_label: Accurate Call Stacks
+sidebar_position: 1
+---
+
+# Accurate Call Stacks
+
+When debugging Reanimated code, you may encounter error or warning call stacks that don’t clearly indicate the root cause of a problem. This occurs because, by default, React Native displays the entire call stack up to the point where the error was thrown or the warning was logged. This can be misleading, as it often highlights code from Reanimated internals instead of the specific problematic code within your project.
+
+To address this, Metro provides a way to enhance stack traces through the Metro `symbolicator`. Reanimated utilizes this feature by exporting a Metro config wrapper called `wrapWithReanimatedMetroConfig`. This wrapper configures the symbolicator to improve the accuracy of stack traces, making it easier to identify issues within your code.
```
tjzel, line 11: While this is well written, I'm afraid it might not address its target audience properly.

First of all, these paragraphs could suggest like we're moving the responsibility of knowing whether an error comes from Reanimated or the user's code to the user himself with little guidance.

Secondly, there's no point at which Reanimated users would be incentivized to look-up this page in the docs. In my opinion they'd still be stuck on vague errors, googling them or creating new issues.

Ideally, we'd have the wrapper enabled by default, but I guess it isn't really feasible, analogous to the case of Reanimated Babel plugin.

To solve this issue, I have a few ideas:
- Let's put the information about the Metro Wrapper somewhere in the installation guide as an __recommended, optional step__.
- If it's possible in runtime to obtain the information whether the Metro Wrapper is present, we could embed some runtime solution. For example, we could add a one-time warning during Reanimated initialization that Metro Wrapper isn't present or we could append information about the Wrapper to every error thrown by Reanimated.

I think we should consider who exactly the target of our Metro Wrapper is, probably specify it in these docs and edit them accordingly.

As a side note, I think technical details here might be unnecessarily intertwined with DX-relevant information. Let's make sure the most important information don't include technicalities and provide them later on.

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+| Before                                                                                              | After                                                                                             |
+| --------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
+| ![Call Stack before applying Reanimated Metro config wrapper](/img/debugging/MetroConfigBefore.png) | ![Call Stack after applying Reanimated Metro config wrapper](/img/debugging/MetroConfigAfter.png) |
+
+## Remarks
+
+- Some errors, especially those caused by asynchronous code, may still result in stack traces that point to internal Reanimated rather than the specific line in your code that triggered the error. This happens because the stack trace loses track of the component or function that initiated the asynchronous operation.
```
tjzel, line 39: I think it would be nice to add here whether the user can or can't do something about it.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
```diff
@@ -0,0 +1,70 @@
+---
+id: logger-configuration
+sidebar_label: Logger configuration
+sidebar_position: 2
+---
+
+# Logger configuration
+
+Reanimated shows warnings that indicate potential issues in the user code. These logs can be configured to be more or less verbose. The default configuration is to show all warnings. You can change this behavior by using the `configureReanimatedLogger` function.
+
```
tjzel, line 10: Let's give an example of such an warning here. Let's also make it more prominent that we warn of misuses of Reanimated API, than general "potential issues in the user code".

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
```diff
@@ -0,0 +1,70 @@
...
+#### `strict`
+
+A boolean value that enables or disables strict mode. When strict mode is enabled, Reanimated will show more warnings that can help you to catch potential issues in your code.
+
+## Remarks
+
+- The logger configuration is global and affects all Reanimated animations in your app.
+
+- The `configureReanimatedLogger` function should be called before any Reanimated animations are created, e.g. in the root file of your app.
```
tjzel, line 62: Is such information present as TSDoc in our codebase?

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+- Each call to `configureReanimatedLogger` overrides the previous configuration.
```
tjzel, line 64: What's the point of calling it again if we want it to be called in the App's root?

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+- There are currently no `fatal` logs in Reanimated, but the `fatal` level is reserved for future use.
```
tjzel, line 66: I don't think such implementation detail should be present in the documentation.

tjzel (changes requested):

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
MatiPl01: I wonder if I should remove `fatal` from the enum, since we don't use it anywhere. I added this kind of error level as LogBox had `fatal` in it's log level union type but it might be unnecessary in our case.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
MatiPl01: Mhm, I just wanted to let users know that if `configureReanimatedLogger` is called again, it will override the previous configuration. I can agree that this bullet point might be confusing and unnecessary.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
MatiPl01: Nope, I didn't add a TSDoc for `configureReanimatedLogger`.
I will add it as it may be useful.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
MatiPl01: You think, that including the example within a sentence (like: "Reanimated shows warnings that indicate misuses of the library API, such as modifying the Shared Value during component re-render") will be enough? I don't want to paste a screenshot with a warning in here as the description will become a bit messy.

Alternatively, I can add the section with example warning/warnings showing how they look like.

tjzel: Yes, I think such example would be sufficient 😄

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
tjzel: Well, if it bothers you and we don't have any short-term plans to implement anything `fatal`, I guess we can just remove it.

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
MatiPl01: > If it's possible in runtime to obtain the information whether the Metro Wrapper is present, we could embed some runtime solution. For example, we could add a one-time warning during Reanimated initialization that Metro Wrapper isn't present or we could append information about the Wrapper to every error thrown by Reanimated.

I'm afraid this won't be possible. We don't have a direct access to metro from react native so we likely won't be able to detect if this config is used.

Even though we pass an error stack trace in out custom logger to the `addLog` method from `LogBox`, this stack trace is not symbolicated and we don't know how it will look after metro symbolication is applied (and whether our custom metro symbolicator is used).

> Let's put the information about the Metro Wrapper somewhere in the installation guide as an recommended, optional step.

Will this look fine or should I change something?

Screenshot 2024-09-02 at 13 37 36

MatiPl01: > I think we should consider who exactly the target of our Metro Wrapper is, probably specify it in these docs and edit them accordingly.

I think that this config wrapper will be helpful for everyone, also for devs. Do you think that we wouldn't want to recommend it for everyone? I think that most developers, who use Reanimated as a dependency of their library, would prefer to see detailed errors with proper code snippet.

If you meant developers of Reanimated, we can still unwrap wrapped stack frames. They aren't removed, so adding this wrapper won't remove anything form stack. These errors would be helpful even for us, e.g. while developing an example app.

packages/react-native-reanimated/src/logger/logger.ts
kacperkapusciak, line 1: a documentation change also changing the reanimated source code? for me it's a no no

or at least don't mark it as a docs change

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
```diff
@@ -91,7 +91,32 @@ To learn more about the plugin head onto to [Reanimated babel plugin](/docs/fund
 
 </details>
 
-### Step 3: Clear Metro bundler cache (recommended)
+### Step 3: Wrap metro config with reanimated wrapper (recommended)
```
kacperkapusciak, line 94: I strongly advocate for **not** adding installation steps to Reanimated in the fundamentals 

As this metro config is not required I would omit it from the fundamentals/getting-started

We could ask Expo to include it in their metro config tho

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
```diff
@@ -91,7 +91,32 @@ To learn more about the plugin head onto to [Reanimated babel plugin](/docs/fund
...
+
+Wrap your existing Metro configuration in the `metro.config.js` file with the `wrapWithReanimatedMetroConfig` function.
+
+```js
+const {
+  wrapWithReanimatedMetroConfig,
```
kacperkapusciak, line 100: is it still possible to change the name of this function?

i'd prefer `adaptReanimatedMetroConfig(config)` or something

as this arguably doesn't wrap the config tho, it changes the fields in the config object

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
```diff
@@ -91,7 +91,32 @@ To learn more about the plugin head onto to [Reanimated babel plugin](/docs/fund
...
+  // Your existing Metro configuration options
+};
+
+module.exports = wrapWithReanimatedMetroConfig(config);
+```
+
+<details>
+<summary>Why should I do this?</summary>
+
+By wrapping your Metro configuration with the Reanimated Metro config wrapper, you can enhance reanimated errors and warnings with more accurate call stacks. This makes it easier to identify issues within your code.
```
kacperkapusciak, line 113:
```suggestion
By wrapping your Metro configuration with the Reanimated Metro config wrapper, you can improve reanimated errors and warnings with more accurate call stacks. This makes it easier to identify issues within your code.
```

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
```diff
@@ -0,0 +1,67 @@
+---
+id: logger-configuration
+sidebar_label: Logger configuration
+sidebar_position: 2
+---
+
+# Logger configuration
+
+Reanimated shows warnings that indicate misuses of the library API, such as modifying the [shared value](/docs/fundamentals/glossary#shared-value) during component re-render. These logs can be configured to be more or less verbose. The **default** configuration shows **all warnings and errors**. You can change this behavior by using the `configureReanimatedLogger` function.
```
kacperkapusciak, line 9: the docs doesn't say whether the logger is enabled by default or the user needs to enable it themself

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
```diff
@@ -0,0 +1,67 @@
...
+}
+```
+
+</details>
+
+### Configuration options
+
+#### `level`
+
+A value od the `ReanimatedLogLevel` enum that defines the minimum level of the logs that will be shown.
```
kacperkapusciak, line 51:
```suggestion
A value of the `ReanimatedLogLevel` enum that defines the minimum level of the logs that will be shown.
```

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+#### `strict`
+
+A boolean value that enables or disables strict mode. When strict mode is enabled, Reanimated will show more warnings that can help you to catch potential issues in your code.
+
+## Remarks
+
+- The logger configuration is global and affects all Reanimated animations in your app.
```
kacperkapusciak, line 59: the docs should say whether the logger is enabled in production build

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
```diff
@@ -0,0 +1,60 @@
...
+id: accurate-call-stacks
+sidebar_label: Accurate Call Stacks
+sidebar_position: 1
+---
+
+# Accurate Call Stacks
+
+When debugging Reanimated code, you may encounter error or warning call stacks that do not clearly indicate the root cause of the problem. These stacks can be misleading, as they often highlight code from Reanimated's internals rather than the misuse of the Reanimated API that is the source of the problem.
+
+To address this, Reanimated provides a Metro configuration wrapper called `wrapWithReanimatedMetroConfig`. This wrapper automatically adjusts your Metro config to enhance the accuracy of call stacks in warnings and errors generated by the Reanimated library.
```
kacperkapusciak, line 11:
```suggestion
To address this, Reanimated provides a Metro configuration wrapper called `wrapWithReanimatedMetroConfig`. This wrapper automatically adjusts your Metro config to improves the accuracy of call stacks in warnings and errors generated by the Reanimated library.
```
i detest the word `enhance` since the day the chatgpt was born

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
tjzel, line 113: This sentence sound like the user helps Reanimated, because he _enhances reanimated error and warnings_. Let's reword it so we'd signify the benefit to the user more, i.e.:

```suggestion
Errors and warnings you encounter when you develop an app with Reanimated are sometimes indirect and difficult to understand. Reanimated Metro config wrapper handles these indirections internally and instead guides you directly to the offending code, providing clean and comprehensible information.

Unfortunately, we can't enable the wrapper in your app by default. 
```

tjzel (changes requested):
Awesome work, much better than the original version! Still some little bits to improve though.

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
tjzel: Could you elaborate? We find this feature to be of substantial benefit to the user, since eventually they would be guided to their offending code in errors from Reanimated.

I also don't like the fact that it bloats the installation procedure and makes it less approachable. How about creating a new subpage (albeit close to `getting-started`) about recommended installation steps/good practices?

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
tjzel: The name `wrapWithReanimatedMetroConfig` was coined to be on par with the visuals of using it:

```js
module.exports = wrapWithReanimatedMetroConfig(config); // config is "wrapped" here
```

I agree that `wrapWith` seems a bit odd. We could try `apply`, `add` or `mergeWith` instead.

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
tjzel: GOAT

packages/react-native-reanimated/src/logger/logger.ts
tjzel: +1, let's turn it into something like `enable strict logger by default`

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
MatiPl01: I can agree with @tjzel that users will benefit from adding this metro config wrapper and we should somehow make it clear for users that such a wrapper exists. Can you suggest any better solution than the current one? I like the idea that @tjzel suggested to add separate page with recommended configuration steps if you don't want to include this step in the getting started page.

packages/react-native-reanimated/src/logger/logger.ts
MatiPl01: Yep, it makes sense. What about splitting code into 2 PRs to make changes clear?

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
MatiPl01: I wonder if `apply` or `merge` prefixes won't be even more confusing. They might suggest that some fields of the config would be overridden but the implementation of `wrapWithReanimatedMetroConfig` works in such a way that it takes into account users settings (including `symbolicator` field) and collapses these stack frames, which point to reanimated source code.

I previously used name `withReanimatedMetroConfig` (or `withReanimated`) but @tjzel suggested changing it to `wrapWithReanimatedMetroConfig`.

I have no strong opinion on how we should name this function so I can rename it if we decide on something.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
MatiPl01: Mhm, I thought that this part of the sentence `The **default** configuration shows **all warnings and errors**` is enough. This fragment explicitly tells that the user **can change** this behavior by using the `configureReanimatedLogger` function, but, when the default behavior is sufficient, no configuration must be made.

Also this part `Reanimated shows warnings that indicate misuses of the library API` tells that reanimated displays logs by default (so, at least to me, it means that the logger is enabled).

How do you want to make it more explicit? I don't think we have to describe internals in detail.

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
MatiPl01: Well, I like to use the `enhance` word, but can change it to `improve` if you prefer.

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+The following example shows the difference in call stacks before and after applying the Reanimated Metro config wrapper. The **Before** image displays Reanimated source code as the error source, while the **After** image shows the actual incorrect code that caused the error.
+
+| Before                                                                                              | After                                                                                             |
+| --------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
+| ![Call Stack before applying Reanimated Metro config wrapper](/img/debugging/MetroConfigBefore.png) | ![Call Stack after applying Reanimated Metro config wrapper](/img/debugging/MetroConfigAfter.png) |
+
+## Remarks
+
+- The `wrapWithReanimatedMetroConfig` doesn't remove any stack frames from the call stack; it only collapses the frames that are not relevant to the user. Collapsed stack frames can be expanded by pressing on the **See N more frames** text at the bottom of the **Call Stack**.
```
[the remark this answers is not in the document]
MatiPl01: I don't like overusing `you` and writing something to the user directly. I prefer to write things impersonally but, if you find that it's too formal, I can change as you wish.

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+| Collapsed                                                      | Expanded                                                     |
+| -------------------------------------------------------------- | ------------------------------------------------------------ |
+| ![Collapsed Call Stack](/img/debugging/CallStackCollapsed.png) | ![Expanded Call Stack](/img/debugging/CallStackExpanded.png) |
+
+</Indent>
+
+- Some errors, especially those caused by asynchronous code, may still result in stack traces that point to Reanimated internals rather than the specific line in your code that triggered the error. This happens because the stack trace loses track of the component or function that initiated the asynchronous operation.
+
+- Unfortunately, there's nothing you can do to prevent this behavior, as it's a limitation of how stack traces in asynchronous code are handled. You'll need to manually debug the issue based on the error message to identify the potential cause of the problem.
```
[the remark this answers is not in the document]
MatiPl01: This looked a bit too sparse in a single bullet point. Maybe I can somehow combine it and make it a bit less wordy than it's now.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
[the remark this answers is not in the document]
MatiPl01: Yup, I have been using some before but it must have got disabled for some reason. Thanks for reminding me to enable it.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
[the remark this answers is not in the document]
MatiPl01: > of your code indiscriminately

Doesn't it suggest that this may affect user's code that is not related to Reanimated?
You suggested me before to be more explicit that changes affect how Reanimated work. not how the app works in general.

I can agree that `animations` word might be confusing and should be changed to something else.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+- The `configureReanimatedLogger` function should be called before any Reanimated animations are created, e.g. in the root file of your app.
+
+- The `configureReanimatedLogger` function is intended for application developers. If you are creating a library that relies on Reanimated, do not include this function call in your library source code.
+
```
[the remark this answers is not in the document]
MatiPl01: > and it might override their own configurations

this is not technically possible. All node modules are executed first and the user's configuration cannot be overridden in this way.

It will override the default configuration. though, which might be confusing to the user.

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
MatiPl01: Too wordy for me. I can agree that the fragment _you can enhance_ of the sentence could be improved but I don't like your suggestion. I will think of something else.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
MatiPl01: It depends. The logger didn't change anything compared to the previous behavior that we had while using `console.warn`/`console.error`. Some warnings are shown only in `__DEV__` and some not.

I don't want to write that the logger will be included in production builds as it will include most of the logs but not all of them.

I think we should add a possibility to configure whether we want to see warnings in prod as well via the logger config in the future.

What do you suggest writing for now?

packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -2,6 +2,10 @@
 import { addLogBoxLog } from './LogBox';
 import type { LogData, LogBoxLogLevel } from './LogBox';
 
+const DOCS_URL =
+  'https://docs.swmansion.com/react-native-reanimated/docs/debugging/logger-configuration';
+const DOCS_REFERENCE = `If you don't want to see this message, you can disable the \`strict\` mode. Refer to:\n${DOCS_URL} for more details.`;
+
```
MatiPl01, line 8: This is related to docs and includes a link to the documentation page so I would prefer to merge this change together with docs.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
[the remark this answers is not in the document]
tjzel: > Doesn't it suggest that this may affect user's code that is not related to Reanimated?
Your right, it might suggest that.

> The logger configuration is global and affects all warnings and errors shown by Reanimated. There's no option to configure the logger per file/component.
Me likes 👍

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
tjzel: Good for me 👍

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+To address this, Reanimated provides a Metro configuration wrapper called `wrapWithReanimatedMetroConfig`. This wrapper automatically adjusts your Metro config to improve the accuracy of call stacks in warnings and errors generated by the Reanimated library.
+
+<details>
+<summary>How does it work?</summary>
+
+By default, React Native displays the entire call stack up to the point where an error is thrown or a warning is logged, including all stack frames except those from the React Native source code.
+
+To modify this behavior, we can use the `symbolicator` field in the Metro config, which allows customization of the displayed stack frames. Reanimated leverages this feature to adjust which stack frames are **hidden** (**collapsed**) in the stack trace. By doing so, stack frames from Reanimated internals are hidden, ensuring that the stack trace only highlights the relevant parts of the call stack.
```
tjzel, line 18: Is there any external source we could link with `symbolicator`?

tjzel (approved):
Good for me now, let's wait for @kacperkapusciak's review.

packages/docs-reanimated/docs/debugging/logger-configuration.mdx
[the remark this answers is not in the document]
MatiPl01: Ok, so you prefer to change this description to say that the user's configuration will be overridden instead of leaving this part of the sentence *which will override the default configuration*.

packages/docs-reanimated/docs/debugging/accurate-call-stacks.mdx
MatiPl01: I was searching for docs but, unfortunately, haven't found anything. I just came across this prop why looking at the source code in metro.

MatiPl01: There is no `symbolicator` field in the official [docs](https://metrobundler.dev/docs/configuration/#configuration-structure). I think, for some reason, metro prefers to hide this configuration option form the users.

This field is included in the config type, though:

```tsx
export interface ConfigT extends Readonly<MetalConfigT> {
  readonly resolver: Readonly<ResolverConfigT>;
  readonly server: Readonly<ServerConfigT>;
  readonly serializer: Readonly<SerializerConfigT>;
  readonly symbolicator: Readonly<SymbolicatorConfigT>; // <-- here
  readonly transformer: Readonly<TransformerConfigT>;
  readonly watcher: Readonly<WatcherConfigT>;
}
```

kacperkapusciak (approved):

PR #6477, opened 2024-09-03
chore: Enable strict mode by default, remove fatal log, add missing web check

MatiPl01:
## Summary

This PR is just a set of logger implementation improvements.

- fixes crash on web caused by missing check before the `executeOnUIRuntimeSync` call (which doesn't exist on web),
- enables `strict` mode by default,
- removes `fatal` log level as it is not used anywhere.


packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -36,7 +36,7 @@ function logToConsole(data: LogData) {
 export const DEFAULT_LOGGER_CONFIG: LoggerConfigInternal = {
   logFunction: logToConsole,
   level: LogLevel.warn,
-  strict: false,
+  strict: true,
```
MatiPl01, line 39: Better to set to `true` by default and add information how to disable it with link to docs - will be added with respective docs pages in the #6467 PR.

packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -117,7 +117,7 @@ type LogOptions = {
 };
 
 function handleLog(
-  level: Exclude<LogBoxLogLevel, 'syntax'>,
+  level: Exclude<LogBoxLogLevel, 'syntax' | 'fatal'>,
```
MatiPl01, line 121: We don't use `fatal` for now, so it's better to remove it to prevent confusion. I added it just because logbox has a `fatal` type separate from  `error`/`warn`/`syntax` types.

packages/react-native-reanimated/src/mutables.ts
```diff
@@ -18,7 +18,7 @@ function shouldWarnAboutAccessDuringRender() {
 function checkInvalidReadDuringRender() {
   if (shouldWarnAboutAccessDuringRender()) {
     logger.warn(
-      'Reading from `value` during component render. Please ensure that you do not access the `value` property while React is rendering a component.',
+      'Reading from `value` during component render. Please ensure that you do not access the `value` property of a shared value while React is rendering a component.',
```
tjzel, line 21:
```suggestion
      'Reading from `value` during component render. Please ensure that you do not access the `value` property or use `get` method of a shared value while React is rendering a component.',
```

packages/react-native-reanimated/src/mutables.ts
```diff
@@ -27,7 +27,7 @@ function checkInvalidReadDuringRender() {
 function checkInvalidWriteDuringRender() {
   if (shouldWarnAboutAccessDuringRender()) {
     logger.warn(
-      'Writing to `value` during component render. Please ensure that you do not access the `value` property while React is rendering a component.',
+      'Writing to `value` during component render. Please ensure that you do not access the `value` property of a shared value while React is rendering a component.',
```
tjzel, line 30:
```suggestion
      'Writing to `value` during component render. Please ensure that you do not access the `value` property or use `set` method of a shared value while React is rendering a component.',
```

tjzel (approved):
:shipit:

PR #6479, opened 2024-09-03
Implement styleTransition

Latropos:
### Style transition vs layout transition
The **layout transition** is a feature of reanimated that allows to transform each layout change into smooth animation. Analogously, a **style transition** will be generalisation of the layout transition - it animates any style change.

## Recordings

<table>
<tr><td width="50%">IOS</td><td width="50%">ANDROID</td></tr>
<tr><td>

https://github.com/user-attachments/assets/d9ddcebc-56a3-4f74-9fcf-2c0d3bd28d5b

</td><td>

</td></tr>
</table>


packages/react-native-reanimated/src/animationBuilder.tsx
```diff
@@ -7,25 +7,38 @@ import type {
...
+  ...(['TopLeft', 'TopRight', 'BottomLeft', 'BottomRight'] as const).map(
+    (direction) => `Border${direction}Radius` as const
+  ),
+];
+
+const mockTargetValues = Object.fromEntries(
+  propBaseList.map((propName) => [`target${propName}`, 0])
+);
+const mockCurrentValues = Object.fromEntries(
+  propBaseList.map((propName) => [`target${propName}`, 0])
```
piaskowyk, line 33:
```suggestion
  propBaseList.map((propName) => [`current${propName}`, 0])
```

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
```diff
@@ -53,6 +84,111 @@ struct Snapshot {
...
+
+    "BorderTopLeftRadius", // int
+    "BorderTopRightRadius",
+    "BorderBottomLeftRadius",
+    "BorderBottomRightRadius",
+
+    "BorderLeftWidth", // float
+    "BorderRightWidth",
+    "BorderTopWidth",
+    "BorderBottomWidth"};
```
piaskowyk, line 119: what about the rest of props, for example transform?

apps/common-app/src/examples/LayoutAnimations/BasicLayoutAnimation.tsx
```diff
@@ -17,14 +13,23 @@ export default function BasicLayoutAnimation() {
       <Button onPress={() => setState(!state)} title="Update" />
       {visible && (
         <Animated.View
-          entering={BounceInLeft}
-          layout={Layout.springify()}
-          exiting={FadeOutRight.duration(500)}
+          layout={LinearStyleTransition.duration(1000)}
```
piaskowyk, line 16: i think it would be better to keep it as new props called for example `transition`

packages/react-native-reanimated/src/layoutReanimation/defaultTransitions/LinearStyleTransition.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  build = (): LayoutAnimationFunction => {
+    const delayFunction = this.getDelayFunction();
+    const [animation, config] = this.getAnimationAndConfig();
+    const callback = this.callbackV;
+    const delay = this.getDelay();
+
+    return (values) => {
+      'worklet';
+      return {
```
piaskowyk, line 33: It seems better to generate an animation object based on properties provided in `values` instead of hardcoding them as one big object

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
```diff
@@ -53,6 +84,111 @@ struct Snapshot {
...
+
+struct StyleSnapshot {
+  std::array<double, 9> numericPropertiesValues;
+  std::array<std::string, 2> stringPropertiesValues;
+
+  StyleSnapshot(
```
piaskowyk, line 125: Is there a way to create a snapshot lazily only with values that have been changed in some way during component lifecycle?

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
Latropos: I've just added support for transform. There are much more props, that may require some additional consideration - do we want to implement shadow unification for Android and IOS? Also there are props with quantified values (borderStyle is either 'solid' or 'dotted').
I would keep the current number of props for now.

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
Latropos: I am afraid theres no

apps/common-app/src/examples/LayoutAnimations/BasicLayoutAnimation.tsx
Latropos: I think it's a good idea, but I am not sure how would you like layoutTransition and styleTransition to overwrite each others props. Perhaps we keep it for now the way it is?

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
Latropos: However, we want the transition to work for set of properties provided by the user, so we still have some space for optimisation here.

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsManager.h
```diff
@@ -74,6 +74,12 @@ class LayoutAnimationsManager {
   std::unordered_map<int, std::shared_ptr<Shareable>> enteringAnimations_;
   std::unordered_map<int, std::shared_ptr<Shareable>> exitingAnimations_;
   std::unordered_map<int, std::shared_ptr<Shareable>> layoutAnimations_;
+  // LAYOUT_AND_STYLE animations are a subset of LAYOUT animations and use the
+  // same array to store config
+  // We use this additional array to figure out if given LAYOUT animation is
+  // LAYOUT_AND_STYLE too
+  std::unordered_map<int, bool> doLayoutAnimationContainStyle_;
```
piaskowyk, line 81: I don't understand that name, could you change it to something more accurate? 🙏

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
```diff
@@ -6,11 +6,17 @@
 #include <react/renderer/mounting/MountingOverrideDelegate.h>
 #include <react/renderer/mounting/ShadowView.h>
 #include <memory>
+#include <string>
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
 #include <vector>
 
+#ifdef __ANDROID__ // parse color on android
```
piaskowyk, line 15: It appears that you no longer need those changes in this file

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/SnapshotUtils.h
```diff
@@ -0,0 +1,159 @@
...
+    "BorderTopRightRadius",
+    "BorderBottomLeftRadius",
+    "BorderBottomRightRadius",
+
+    "BorderLeftWidth", // float
+    "BorderRightWidth",
+    "BorderTopWidth",
+    "BorderBottomWidth"};
+
+struct StyleSnapshot {
```
piaskowyk, line 104: The names Snapshot and StyleSnapshot might be confusing. Could you come up with different naming?

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/SnapshotUtils.h
```diff
@@ -0,0 +1,159 @@
...
+#include <vector>
+
+#ifdef __ANDROID__ // parse color on android
+#include <iomanip>
+#include <sstream>
+#endif
+
+namespace reanimated {
+
+#ifdef __ANDROID__
```
piaskowyk, line 22: Could we use a const expression here instead of `#ifdef`? This approach is more type-safe and doesn't have any performance overhead also

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -684,40 +744,77 @@ void LayoutAnimationsProxy::startExitingAnimation(
 
 void LayoutAnimationsProxy::startLayoutAnimation(
```
piaskowyk, line 745: How about introducing a new method called `startLayoutTransition`? 🤔

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsManager.cpp
```diff
@@ -24,6 +24,11 @@ void LayoutAnimationsManager::configureAnimationBatch(
         enteringAnimationsForNativeID_[tag] = config;
         continue;
       }
+      if (type == LAYOUT_AND_STYLE) {
```
piaskowyk, line 27: Why you can't reuse the logic from `getConfigsForType`?

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -331,9 +332,14 @@ void LayoutAnimationsProxy::handleUpdatesAndEnterings(
...
-        if (!layoutAnimationsManager_->hasLayoutAnimation(tag, LAYOUT) ||
-            (!shouldAnimate && !layoutAnimations_.contains(tag))) {
+        auto hasLayoutAnimation =
+            layoutAnimationsManager_->hasLayoutAnimation(tag, LAYOUT);
+        auto hasLayoutAndStyleAnimation =
+            layoutAnimationsManager_->hasLayoutAnimation(tag, LAYOUT_AND_STYLE);
+        auto layoutChanged = hasLayoutChanged(mutation);
+        if (!hasLayoutAnimation ||
+            (!hasLayoutAndStyleAnimation && !layoutChanged &&
+             !layoutAnimations_.contains(tag))) {
```
piaskowyk, line 342: Since this check is getting larger, could you move it to a separate method?

apps/common-app/src/examples/SharedElementTransitions/ChangeTheme.tsx
```diff
@@ -41,7 +41,9 @@ function getTheme(theme: boolean, disabled: boolean) {
       height: withSpring(values.targetHeight, config),
       originX: withSpring(values.targetOriginX, config),
       originY: withSpring(values.targetOriginY, config),
-      borderRadius: withSpring(values.targetBorderRadius, config),
+      borderRadius: values.targetBorderRadius
```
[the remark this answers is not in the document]
Latropos: As we have `interface SharedTransitionAnimationsValues
  extends LayoutAnimationsValues` and I've marked `borderRadius?: number` as optional inside LayoutAnimationsValues.
  So I check if it is not undefined.

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/SnapshotUtils.h
Latropos: Let's have `LayoutSnapshot` and `StyleSnapshot`

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsManager.cpp
Latropos: `getConfigsForType` of LAYOUT_AND_STYLE (renamed STYLE_TRANSITION) returns config for LAYOUT animation:
```cpp
    case LAYOUT:
    case STYLE_TRANSITION: // a subset of LAYOUT
      return layoutAnimations_;
    case SHARED_ELEMENT_TRANSITION:
    case SHARED_ELEMENT_TRANSITION_PROGRESS:
      return sharedTransitionAnimations_;
```
I want to leverage code of LAYOUT as much as possible, and STYLE_TRANSITIONS  are a subset of LAYOUT animations. Therefore I use separate array to distinguish the LAYOUT animations being STYLE_TRANSITIONS. The similar approach was used for SHARED_ELEMENT_TRANSITION_PROGRESS.

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsProxy.cpp
Latropos: I was trying several approaches and I've noticed, that the variables defined to get this checked (like `hasLayoutAndStyleAnimation`) are used later in the code. So extraction would bring some redundancy. Also I was wondering whether it doesn't make more sense to extract all long cases from this large switch loop (Update and Insert), and such a refactor could be safely done after merging this feature.

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/SnapshotUtils.h
Latropos: Unfortunately `constexpr` can't be used with strings. I was trying several workarounds and each of them required many typecasts and was generating lot's of code.

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsProxy.cpp
Latropos: Do you mean `startStyleTransition` or do you think of renaming all the stuff related to layout Transitions to use naming `LayoutTransition` instead of `LayoutAnimation`?

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsManager.h
```diff
@@ -74,6 +74,12 @@ class LayoutAnimationsManager {
   std::unordered_map<int, std::shared_ptr<Shareable>> enteringAnimations_;
   std::unordered_map<int, std::shared_ptr<Shareable>> exitingAnimations_;
   std::unordered_map<int, std::shared_ptr<Shareable>> layoutAnimations_;
+  // LAYOUT_AND_STYLE animations are a subset of LAYOUT animations and use the
```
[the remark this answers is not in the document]
Latropos: Yes, but I really think I should mention this information above both here and inside `hasLayoutAnimation` function

piaskowyk:
We'll use some of these approaches in the SET implementation - https://github.com/software-mansion/react-native-reanimated/pull/7466

PR #6486, opened 2024-09-05
chore: Bump prettier

tjzel:
## Summary

- Applied new formatting.
- Added global .prettierignore for `build` so prettier won't try to format .jsons generated by Android

## Test plan

🚀


MatiPl01 (approved):
I don't like that prettier adds these unnecessary commas, but if it's not configurable then it doesn't matter to me.

tomekzaw (approved):

PR #6487, opened 2024-09-05
chore: Add separate tsconfig for metro-config

tjzel:
## Summary

Including both `src` and `metro-config` in `tsconfig.json` of Reanimated package sometimes falsely suggests TypeScript that there should be `lib/module/src` and `lib/module/metro-config` directories.

## Test plan

🚀

### DIFF
--- a/packages/react-native-reanimated/metro-config/tsconfig.json
+++ b/packages/react-native-reanimated/metro-config/tsconfig.json
@@ -1,0 +1,3 @@
+{
+  "extends": "../../tsconfig.json",
+}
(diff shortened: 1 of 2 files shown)


MatiPl01 (approved):
I wouldn't say that it suggests that there should be a separate directory in the built library as the `react-native-builder-bob` field in the `package.json` specifies which directories should be included in the build process. 

To me, adding `metro-config` directory to the `include` array in the tsconfig just says that the same typescript configuration should be used for this directory as for the `src`.

It makes no difference to me which approach we decide to use, so if you prefer this implementation, let's merge this PR.

tomekzaw (approved):

PR #6520, opened 2024-09-17
fix: Add JavaScript wrappers to captured console to prevent crashes in React Native 0.75

tjzel:
## Summary

Currently there seems to be a bug in the JSI layer which causes a crash when we try to copy some of the console methods, i.e. `clear` or `dirxml`.

The crash happens only in React Native 0.75. It's not reproducible in neither 0.76 nor 0.74. It also happens only in the configuration of a debug app and production bundle.

I haven't yet discovered what exactly causes the crash. It's tied to the console methods sometimes being `HostFunction`s. Therefore, as a workaround we don't copy the methods as they are in the original console object, we copy JavaScript wrappers instead.

The fix should't be treated as a complete solution. Same crash could also happen on copying of some other HostFunctions.

## Details

1. In Hermes runtime there's a global `console` object. Reanimated sends methods of this object via JSI to store "references" to them.
2. Sometimes, some of the console methods turn out to be HostFunctions. In this case, we obtain the underlying `std::function` from the `jsi::Function` with `getHostFunction`.
3. We copy the obtained `std::function` and store that copy. This is where the crash happens, on the copy assignment (memory - bad access).

Certain conditions have to be met for this flow to occur and for the crash to happen:

- It's present only in React Native 0.75.x. I've tried both 0.74 and RC of 0.76 and everything is okay there.
- I've been troubleshooting for iOS but allegedly it's platform agnostic. Also, the crash appears on both Paper and Fabric.
- When I mentioned that "some of the console methods turn out to be HostFunctions" - this only happens in RN 0.75. In both 0.74 and RC of 0.76 the methods we copy are never HostFunctions.
- It's present in configuration of a debug build of the app with a production bundle. In cases of debug&debug or prod&prod there's no crash.
- Methods which causes the crash are the more exotic ones - `clear`, `dirxml` etc. They all are HostFunctions when crash conditions are present. Common methods like `log`, `warn` etc. are standard JS functions and can be copied with no issues.

I rummaged through Hermes repo to find where i.e. `clear` method originates. I found this file which from I understand creates the "first" console object:
https://github.com/facebook/hermes/blob/649ceaa4c8b2ed331251152ccb7fcee07f19fc4c/tools/node-hermes/nodelib/wrappers/internal/console/constructor.js#L470

However, nothing in here suggests that any method should be a HostObject. Therefore I assume it's either replaced somewhere else by Hermes or React Native, i.e.

- When Chrome Debug Protocol is connected into the runtime
  https://github.com/facebook/hermes/blob/649ceaa4c8b2ed331251152ccb7fcee07f19fc4c/API/hermes/inspector/chrome/CDPHandler.cpp#L2372
- Initialization of Hermes, although I assume, `console` object is injected into the global scope for the first time here:
  https://github.com/facebook/hermes/blob/649ceaa4c8b2ed331251152ccb7fcee07f19fc4c/tools/node-hermes/node-hermes.cpp#L306

On the React Native side, I found the place where they replace the console with some polyfills in debug bundles. Since the bundle is in production mode, this doesn't happen:
https://github.com/facebook/react-native/blob/949296571b36080d4ebe0bc8fac9ef3907367475/packages/polyfills/console.js#L596

## Test plan

Apply the following patch to Metro to be able to juggle between bundle types:

```diff
diff --git a/src/lib/splitBundleOptions.js b/src/lib/splitBundleOptions.js
index 3f35705877cd2d91fc032ea6a75a12b449155ac7..f78656a8c5bdeff4d3915cf2a7262613b6f447ba 100644
--- a/src/lib/splitBundleOptions.js
+++ b/src/lib/splitBundleOptions.js
@@ -1,15 +1,18 @@
 "use strict";
 
+const overrideBundleType = true;
+const isDevBundle = false;
+
 function splitBundleOptions(options) {
   return {
     entryFile: options.entryFile,
     resolverOptions: {
       customResolverOptions: options.customResolverOptions,
-      dev: options.dev,
+      dev: overrideBundleType ? isDevBundle : options.dev,
     },
     transformOptions: {
       customTransformOptions: options.customTransformOptions,
-      dev: options.dev,
+      dev: overrideBundleType ? isDevBundle : options.dev,
       hot: options.hot,
       minify: options.minify,
       platform: options.platform,
```

## Notes

Sorry for the wall of text but this crash is way too annoying to document and screenshot everything thoroughly 🤷

Fixes
- #6415


packages/react-native-reanimated/src/initializers.ts
```diff
@@ -80,9 +80,48 @@ export function setupCallGuard() {
...
+    })
+  );
+
+  return consoleCopy as unknown as typeof console;
+}
+
 // We really have to create a copy of console here. Function runOnJS we use on elements inside
 // this object makes it not configurable
-const capturableConsole = { ...console };
+const capturableConsole = createMemorySafeCapturableConsole();
```
tomekzaw, line 124: Have you tried passing console methods to the worklet below directly instead of capturing `capturableConsole`? Something like

```typescript
const { assert, debug, log, warn, error, info } = console;

export function setupConsole() {
  'worklet';
  // ...
  global.console = { assert, debug, log, warn, error, info };
  // ...
}
```

tjzel: I thought about that we could just skip copying the methods we don't use on the UI Runtime. However, there's still no guarantee that some of the ones we do copy won't suddenly become uncopyable HostFunctions.

tomekzaw: Cool, so the general idea is to wrap `jsi::HostFunction` inside a regular JS function so that we copy it as `ShareableRemoteFunction` rather than a `ShareableHostFunction`, is that correct?

tjzel: Yes, that's correct 😸

tomekzaw (approved):

PR #6523, opened 2024-09-18
docs: Add AvailableFrom label

patrycjakalinska:
This PR introduces the `AvailableFrom` label (a standard version and a full-sized version). The purpose of this label is to inform users that feature is available starting from a specified version and in all subsequent versions.

https://github.com/user-attachments/assets/172c4930-75fd-41f1-8616-845e0832cdd1

### DIFF
--- a/packages/docs-reanimated/src/components/AvailableFrom/index.tsx
+++ b/packages/docs-reanimated/src/components/AvailableFrom/index.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import styles from './styles.module.css';
+
+interface Props {
+  version?: string;
+  big?: any;
+}
+
+export default function AvailableFrom({ version, big }: Props) {
+  return (
+    <>
+      {big ? (
+        <div className={styles.big}>
+          <div className={styles.info}>
+            <svg
+              xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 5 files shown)


packages/docs-reanimated/src/components/AvailableFrom/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const getBadgeContent = () => {
+    if (isFull) {
+      return (
+        <>
+          <div className={styles.infoIcon}>
+            {isLightMode ? <Danger /> : <DangerDark />}
+          </div>
+          <span>
+            Available from Reanimated{' '}
+            <span className={styles.version}>v{version}</span>
```
tomekzaw, line 28: Let's drop `v` prefix here for better readability
```suggestion
            <span className={styles.version}>{version}</span>
```

patrycjakalinska: Should we drop `v` prefix for both full version and normal?

tomekzaw: I think yes.

piaskowyk (approved):
Looks great for me 👍

PR #6526, opened 2024-09-20
Fix wrong exiting animation indices [New Architecture]

bartlomiejbloniarz:
## Summary

When an exiting animation is finished, we mark some views for removal. Then we trigger `pullTransaction` to apply those removals. However, it might happen that RN also tries to perform some transaction at the same time. In the algorithm responsible for triggering exiting animations, if we encounter a view marked for removal, we handle it right there by calling `endAnimationsRecursively`. However, we also have to add this view to the `toBeRemoved` list. Otherwise its neighbors will have incorrect indices, and when one of them is removed the app will crash. 

## Test plan

Test `[LA] View Flattening` example with:
```js
  useEffect(() => {
    setInterval(() => {
      setVisible((prev) => !prev);
    }, 0);
  }, []);
```


tjzel (approved):
Crazy Eyes

PR #6534, opened 2024-09-23
draft: fix infinite loop in REASharedTransitionManager

preynal:
## Summary

This fixes an infinite loop in `REASharedTransitionManager`, in `getSharedElementForCurrentTransition`, when `siblingViewTag` is nil and `_findPrecedingViewTagForTransition` returns nil, it will get stuck in an infinite loop and freeze the app.

## Test plan

You can test this by setting sharedTransitionTags

### DIFF
--- a/packages/react-native-reanimated/apple/reanimated/LayoutReanimation/REASharedTransitionManager.m
+++ b/packages/react-native-reanimated/apple/reanimated/LayoutReanimation/REASharedTransitionManager.m
@@ -251,7 +251,6 @@ static BOOL _isConfigured = NO;
     REAUIView *sharedViewScreen = [REAScreensHelper getScreenForView:sharedView];
     REAUIView *stack = [REAScreensHelper getStackForView:sharedViewScreen];
 
-
     NSMutableSet<NSNumber *> *visitedTags = [NSMutableSet new];
     const int maxIterations = 100; // Safety limit to prevent infinite loops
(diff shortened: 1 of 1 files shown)


arekkubaczkowski:
I've applied you changes as a patch and it seems like it fixes the crashing indeed but the shared element transition doesn't work when you try to repeat it within the same shareTag

preynal:
> I've applied you changes as a patch and it seems like it fixes the crashing indeed but the shared element transition doesn't work when you try to repeat it within the same shareTag

thanks, I've seen that too but I didn't know if it was related to my use case, or the crash, I will look into it

arekkubaczkowski:
@preynal sorry for bothering, do you plan to take a look on it in the nearest future?

preynal:
> > I've applied you changes as a patch and it seems like it fixes the crashing indeed but the shared element transition doesn't work when you try to repeat it within the same shareTag

I wasn't able to get it to work correctly, I might get back to it in the future when I have more time

TomAtterton:
This solved the crash for me too although not 100% sure if maxIterations is needed as it worked fine for me without.

piaskowyk:
Thanks for your PR and your time! We're currently rewriting the Shared Element Transition implementation for the new architecture and I hope we'll resolve all the issues that occurred in the Paper implementation. Since these changes are no longer relevant, I'll go ahead and close this PR.

PR #6542, opened 2024-09-24
refactor: Obtain JsCallInvoker directly from context

tjzel:
## Summary

Screenshot 2024-09-24 at 17 27 16

## Test plan

- [ ] Compatibility CI passes ([run](https://github.com/software-mansion/react-native-reanimated/actions/runs/11016769839))
- [x] 🚀


tomekzaw:
What about backward compatibility? Is it possible to do that in RN 0.73?

tjzel:
@tomekzaw https://github.com/software-mansion/react-native-reanimated/actions/runs/11016769839

PR #6543, opened 2024-09-24
Convert all iOS includes to angle-bracket imports

tomekzaw:
## Summary

This PR changes the way headers are imported in `apple/` directory.

Previously, we had the following types of imports:
* `#import <RNReanimated/REAModule.h>` – okay
* `#import <RNReanimated/REASnapshot.h>` – not good, `REASnapshot.h` is located in `apple/LayoutReanimation/REASnapshot.h`
* `#import <ReanimatedSensorContainer.h>` – not good, `RNReanimated/` is missing
* `#include <memory>` – `#import` is better than `#include` since it guarantees to actually paste the contents of the header file at most once and is reported to improve build times slightly

First, I've decided to add `apple/` in the include paths. The reason is to avoid potential conflicts with headers from `Common/cpp/` (perhaps that's why there's `LayoutAnimations` in common codebase and `LayoutReanimation` in Apple codebase?) and also indicate that the header is Apple-only. Another reason is to prevent importing headers directly from `RNReanimated/*.h`. Finally, I think it looks cool.
```diff
-#import <RNReanimated/REAModule.h>
+#import <RNReanimated/apple/REAModule.h>
```

Next, I fixed the header include paths so that they represent the actual directory tree in `apple/`, including intermediate directories like `sensor` or `keyboardObserver`. This also makes it easier to group imports since they are automatically reordered during `yarn format:apple`.
```diff
-#import <RNReanimated/REASnapshot.h>
+#import <RNReanimated/apple/LayoutReanimation/REASnapshot.h>
```

Then, I fixed imports of headers from `worklets` subspec:
```diff
-#import <RNReanimated/WorkletRuntime.h>
+#import <worklets/WorkletRuntime/WorkletRuntime.h>
```

I also fixed some direct imports:
```diff
-#import <REASlowAnimations.h>
+#import <RNReanimated/apple/REASlowAnimations.h>
```

Finally, I changed all `#include` to `#import` (actually, the use of `#include` in `apple/` directory is now disallowed and guarded by CI):
```diff
-#include <jsi/jsi.h>
+#import <jsi/jsi.h>
```

## Test plan

See if CI is green.


packages/react-native-reanimated/RNReanimated.podspec
```diff
@@ -83,62 +83,21 @@ Pod::Spec.new do |s|
   s.source       = { :git => "https://github.com/software-mansion/react-native-reanimated.git", :tag => "#{s.version}" }
 
   s.subspec "reanimated" do |ss|
+    ss.source_files = "Common/cpp/reanimated/**/*.{cpp,h}"
+    ss.header_dir = "reanimated"
+    ss.header_mappings_dir = "Common/cpp/reanimated"
```
tomekzaw, line 88: `header_mappings_dir` prevents flattening of header paths, i.e. if the file is located in `x/y/z/foo.h`, it would be available in `reanimated/foo.h` but with `header_mappings_dir` set it will be in `reanimated/x/y/z/foo.h`.

packages/react-native-reanimated/RNReanimated.podspec
```diff
@@ -83,62 +83,21 @@ Pod::Spec.new do |s|
...
-      sss.header_dir = "worklets/Tools"
-    end
-
-    ss.subspec "WorkletRuntime" do |sss|
-      sss.source_files = "Common/cpp/worklets/WorkletRuntime/*.{cpp,h}"
-      sss.header_dir = "worklets/WorkletRuntime"
-    end
+    ss.source_files = "Common/cpp/worklets/**/*.{cpp,h}"
+    ss.header_dir = "worklets"
+    ss.header_mappings_dir = "Common/cpp/worklets"
```
tomekzaw, line 100: Thanks to `header_mappings_dir` we don't need to ship all subdirectories as separate subsubspecs.

packages/react-native-reanimated/apple/reanimated/Fabric/REAInitializerRCTFabricSurface.h
```diff
@@ -1,7 +1,7 @@
 #ifdef RCT_NEW_ARCH_ENABLED
 
-#import <RNReanimated/REAModule.h>
 #import <React/RCTFabricSurface.h>
+#import <reanimated/apple/REAModule.h>
```
tomekzaw, line 4: I've decided to add `apple/` in the path to avoid conflicts with `Common/cpp/reanimated`. Also, it looks cool.

packages/react-native-reanimated/apple/reanimated/native/REAIOSUIScheduler.h
```diff
@@ -1,8 +1,9 @@
-#import <RNReanimated/UIScheduler.h>
+#import <worklets/Tools/UIScheduler.h>
+
 #import <React/RCTUIManager.h>
 #import <ReactCommon/CallInvoker.h>
 
-#include <memory>
+#import <memory>
```
tjzel, line 6: Should this be `import` or `include`?

tjzel (approved):

packages/react-native-reanimated/apple/reanimated/native/REAIOSUIScheduler.h
tjzel: Nvm I already know hehe

Issue #6512, opened 2024-09-17
Android build failing with `react-native-reanimated` 3.15.2 (RN 0.72.15, NDK 23)


jeanregisser:
### Description

Hey, we got the following compile errors on Android trying to upgrade `react-native-reanimated` from v3.15.1 to v3.15.2.

We use the following SDK/NDK versions, with React Native 0.72.15.

```gradle
ext {
    buildToolsVersion = "34.0.0"
    minSdkVersion = 24
    compileSdkVersion = 34
    targetSdkVersion = 34
    supportLibVersion = "31.0.0"
    kotlinVersion = "1.9.20"
    // We use NDK 23 which has both M1 support and is the side-by-side NDK version from AGP.
    ndkVersion = "23.1.7779620"
}
```

Note: I saw #6466, but compilation was working for us with v3.15.1.

```
  /home/runner/work/wallet/wallet/node_modules/react-native-reanimated/Common/cpp/reanimated/NativeModules/NativeReanimatedModule.cpp:198:34: error: explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Werror,-Wc++20-extensions]
    uiScheduler_->scheduleOnUI([=, this] {
                                   ^
  /home/runner/work/wallet/wallet/node_modules/react-native-reanimated/Common/cpp/reanimated/NativeModules/NativeReanimatedModule.cpp:264:34: error: explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Werror,-Wc++20-extensions]
    uiScheduler_->scheduleOnUI([=, this] {
                                   ^
  /home/runner/work/wallet/wallet/node_modules/react-native-reanimated/Common/cpp/reanimated/NativeModules/NativeReanimatedModule.cpp:278:11: error: explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Werror,-Wc++20-extensions]
        [=, this] { eventHandlerRegistry_->unregisterEventHandler(id); });
            ^
  /home/runner/work/wallet/wallet/node_modules/react-native-reanimated/Common/cpp/reanimated/NativeModules/NativeReanimatedModule.cpp:374:34: error: explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Werror,-Wc++20-extensions]
    uiScheduler_->scheduleOnUI([=, this]() {
                                   ^
[...]

PR #6553, opened 2024-09-26
fix: Restore backwards C++ compatibility for NDK23

tjzel:
## Summary

Turns out preliminary support for NDK27 in
- #6495

broke support for NDK23. This PR brings back support for NDK23 which is used in React Native 0.72.

## Test plan

- [ ] All GitHub actions pass.

## Notes

Fixes
- #6512


tomekzaw (approved):
Clever macro!

lkinasiewicz:
The c++ version recognition is wrong - it's not `20202L` but `202002L`. The fix does not work.
https://en.cppreference.com/w/cpp/preprocessor/replace#Predefined_macros

tjzel:
god damnit

jeanregisser:
I just proposed a fix for the value: #6590

PR #6557, opened 2024-09-27
refactor(Worklets): JavaScript NativeReanimatedModule

tjzel:
## Summary

Requires:
- #6556

QoL refactor that will allow easier NativeWorkletsModule embedment.

Summary:

`NativeReanimated.ts` -> `ReanimatedModule/NativeReanimated.ts`
`js-reanimated/` -> `ReanimatedModule/js-reanimated`

`NativeReanimatedModule` and `JSReanimatedModule` aren't exported anymore. The only way to obtain them is via `createReanimatedModule`, that should yield less error prone code. `NativeWorkletsModule` will follow the same pattern.

## Test plan

- [x] CI green :shrek:


packages/react-native-reanimated/__tests__/hooks.useSharedValue.test.tsx
```diff
@@ -2,8 +2,6 @@ import React from 'react';
 import renderer from 'react-test-renderer';
 import Animated, { useSharedValue } from '../src';
 
-jest.mock('../src/NativeReanimated/NativeReanimated');
```
tjzel, line 5: Turns out there's no need to mock it at all.

packages/react-native-reanimated/scripts/set-reanimated-version.js
```diff
@@ -19,8 +19,7 @@ if (IS_HELP) {
   console.warn(
     'Use --nightly or -n to set nightly version.\nUse --fresh or -f to set fresh version.\nElse pass the version as an argument.'
   );
-  process.exitCode = 1;
-  return;
```
tjzel, line 23: Linter is angry at this.

piaskowyk (approved):
Looks good to me 👍

tomekzaw:
I'd like to review this PR prior to merging

packages/react-native-reanimated/__tests__/hooks.useSharedValue.test.tsx
```diff
@@ -16,8 +14,12 @@ describe('useSharedValue', () => {
...
 
     function TestComponent(props: Props) {
-      const opacity = useSharedValue(props.value);
-      return <Animated.View style={{ opacity: opacity.value }} />;
+      const [opacity, setOpacity] = React.useState(props.value);
+      const opacitySv = useSharedValue(props.value);
+
+      React.useEffect(() => setOpacity(opacitySv.value), [opacitySv]);
+
+      return <Animated.View style={{ opacity }} />;
```
tomekzaw, line 22: Why was this change necessary?

packages/react-native-reanimated/src/ReanimatedModule/index.web.ts
```diff
@@ -0,0 +1,4 @@
+'use strict';
+// this file was created to prevent NativeReanimated from being included in the web bundle
+import { createJSReanimatedModule } from './js-reanimated';
+export const ReanimatedModule = createJSReanimatedModule();
```
tomekzaw, line 4: Previously, we used `export ... from ...` which works correctly with tree-shaking. Can you confirm that the new code also supports tree-shaking?

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
```diff
@@ -13,11 +14,15 @@ import { getValueUnpackerCode } from '../valueUnpacker';
 import { isFabric } from '../PlatformChecker';
 import type React from 'react';
 import { getShadowNodeWrapperFromRef } from '../fabricUtils';
-import ReanimatedModule from '../specs/NativeReanimatedModule';
+import { ReanimatedTurboModule } from '../specs';
 import { ReanimatedError } from '../errors';
 
+export function createNativeReanimatedModule() {
+  return new NativeReanimatedModule();
+}
```
tomekzaw, line 22: What's the deal with wrapping constructor call with a factory function? Why can't we just instantiate `NativeReanimatedModule` directly instead of calling `createNativeReanimatedModule()`?

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
```diff
@@ -13,11 +14,15 @@ import { getValueUnpackerCode } from '../valueUnpacker';
 import { isFabric } from '../PlatformChecker';
 import type React from 'react';
 import { getShadowNodeWrapperFromRef } from '../fabricUtils';
-import ReanimatedModule from '../specs/NativeReanimatedModule';
+import { ReanimatedTurboModule } from '../specs';
```
tomekzaw, line 17: At some point, I'd like to make a clear distinction between `ReanimatedModule` and `ReanimatedTurboModule`. The sole purpose of the latter is to install JSI bindings so let's rename it to `ReanimatedInstallerTurboModule`.

packages/react-native-reanimated/src/specs/index.ts
```diff
@@ -0,0 +1,4 @@
+'use strict';
+import ReanimatedTurboModule from './NativeReanimatedModule';
+
+export { ReanimatedTurboModule };
```
tomekzaw, line 4: Will `export { ReanimatedTurboModule } from './NativeReanimatedModule';` also work here?

packages/react-native-reanimated/src/commonTypes.ts
```diff
@@ -5,6 +5,70 @@ import type {
...
+    iosReferenceFrame: number,
+    handler: ShareableRef<(data: Value3D | ValueRotation) => void>
+  ): number;
+  unregisterSensor(sensorId: number): void;
+  registerEventHandler<TValue>(
+    eventHandler: ShareableRef<TValue>,
+    eventName: string,
+    emitterReactTag: number
+  ): number;
+  unregisterEventHandler(id: number): void;
```
tomekzaw, line 23: Let's add newlines in between for consistency.
```suggestion
  registerSensor(
    sensorType: number,
    interval: number,
    iosReferenceFrame: number,
    handler: ShareableRef<(data: Value3D | ValueRotation) => void>
  ): number;

  unregisterSensor(sensorId: number): void;

  registerEventHandler<TValue>(
    eventHandler: ShareableRef<TValue>,
    eventName: string,
    emitterReactTag: number
  ): number;

  unregisterEventHandler(id: number): void;
```

packages/react-native-reanimated/src/ReanimatedModule/js-reanimated/index.ts
```diff
@@ -1,38 +1,32 @@
...
-import { logger } from '../logger';
-import { ReanimatedError } from '../errors';
+import { PropsAllowlists } from '../../propsAllowlists';
+import { logger } from '../../logger';
+import { ReanimatedError } from '../../errors';
 
-const reanimatedJS = new JSReanimated();
+export { createJSReanimatedModule } from './JSReanimated';
 
+// TODO: Install these global functions in a more suitable location.
```
tomekzaw, line 14: We'll need to move it to a separate file and also mark as `sideEffects` in package.json.

packages/react-native-reanimated/src/ReanimatedModule/js-reanimated/JSReanimated.ts
```diff
@@ -290,7 +299,10 @@ export default class JSReanimated {
   }
 }
 
-enum Platform {
+// Lack of this export breaks TypeScript generation since
+// an enum transpiles into JavaScript code.
```
tomekzaw, line 303: Not sure why this change is necessary and why it didn't break TS generation prior to the changes.

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
```diff
@@ -99,38 +103,38 @@ See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooti
     if (__DEV__) {
       checkCppVersion();
     }
-    this.InnerNativeModule = global.__reanimatedModuleProxy;
+    this.#reanimatedModuleProxy = global.__reanimatedModuleProxy;
```
tomekzaw, line 106: Since we're assigning `global.__reanimatedModuleProxy` to `this.#reanimatedModuleProxy`, does it make sense to reset `global.__reanimatedModuleProxy` to undefined here?

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
```diff
@@ -77,8 +82,8 @@ See \`https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshoo
   }
 }
 
-export class NativeReanimated {
-  private InnerNativeModule: NativeReanimatedModule;
+class NativeReanimatedModule implements IReanimatedModule {
+  #reanimatedModuleProxy: ReanimatedModuleProxy;
```
tomekzaw, line 86: I guess we should also rename the C++ counterpart (`NativeReanimatedModule` to `ReanimatedModuleProxy`) but we can do it in a separate PR.

tomekzaw (changes requested):

tomekzaw:
Also, could you please give a general overview of the changes (e.g. which classes/functions/files have been renamed, which files have been moved to another directory) in the PR description? 🙏

packages/react-native-reanimated/src/specs/index.ts
tjzel: No, `ReanimatedTurboModule` has to be a default export by the rules of the Codegen. This bit of code turns it into a named export.

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
tjzel: AFAIK we can't because of the Codegen. I wanted to have `WorkletsCommonModule` in the first place but I was forced to change the name to `WorkletsNativeModule` - `NativeModule` suffix is required.

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
tjzel: I'd rather stay with `ReanimatedTurboModule`. `ReanimatedTurboModule` is indeed a JavaScript TurboModule instance - it's just that we only use it to call its' `install` function and handle JSI calls with `NativeReanimatedModule`.

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
tjzel: I think it's the standard to use a dedicated factory function when dealing with the singleton pattern. I don't want to expose the `NativeReanimatedModule` class outside at all - you can see that `NativeReanimatedModule` class has no `export` keyword preceding it.

packages/react-native-reanimated/src/ReanimatedModule/index.web.ts
tjzel: I'll check it.

packages/react-native-reanimated/src/ReanimatedModule/js-reanimated/index.ts
tjzel: I'll make a note for myself to add some ESLint plugin for sideEffect detection in the future.

packages/react-native-reanimated/src/ReanimatedModule/js-reanimated/JSReanimated.ts
tjzel: We talked about it offline.

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
tjzel: Leaving it as is as we discussed offline.

packages/react-native-reanimated/src/ReanimatedModule/index.web.ts
tjzel: It works! 👍

tomekzaw (approved):

PR #6564, opened 2024-10-02
Do not make a synchronous call when canceling animation. 

Szymon20000:
## Summary
Screenshot 2024-10-02 at 14 00 50
I noticed in client's code that cancelAnimation called for few values takes almost a second.
That seems to be caused by the fact that it's used on every useShareValue hook on unmount and across multiple libraries build on the top of reanimated. Looks like getter is synchronous and blocks the js until we schedule a block on UI. 
I suspect the idea was to make sure we use the same type as we can animate different types. I thought it can be possible to just take initial value if it's set. Please let me know if for some reason we have to make a synchronous call either way. 

## Test plan
- Start animation and then cancel it with cancelAnimation function


arasrezaei:
I noticed that on navigation backs, does that related to this?
BTW, What is the profiler name?

Szymon20000:
@tomekzaw Can I ask you for a bit of help here? :)

joe-sam:
It was interesting to note that an animation is cancelled by updating its corresponding  SharedValue.

What happens to all the DerivedValue(s) dependent on the SharedValue(s) on CancelAnimation. 

Does a batched call for cancellation with reverse tree dependency chain of responsibility  of the derived animations automatically fire ?

tomekzaw:
Hey @Amurmurmur, right now we're in the process of releasing Reanimated 3.16.0 with support for React Native 0.76.

Once that's complete, we'll review any PRs that have been submitted in the meantime as our capacity is pretty limited.

I'd love to have this problem resolved but this PR is a significant change and we'll need to test it thoroughly prior to merging.

packages/react-native-reanimated/src/animation/util.ts
```diff
@@ -541,5 +541,9 @@ export function defineAnimation<
 export function cancelAnimation<T>(sharedValue: SharedValue<T>): void {
   'worklet';
   // setting the current value cancels the animation if one is currently running
+  if (sharedValue._initial != null) {
+    sharedValue.value = sharedValue._initial;
+    return;
+  }
```
tjzel, line 547: This is unnacceptable, animations on cancel would restore to their initial state:



https://github.com/user-attachments/assets/0b43e6b1-7378-4e96-91b1-2c333059aa15

tjzel (commented):
This is a difficult problem, let me explain why:

There might be a situation, where:
1. `cancelAnimation` is called on React runtime
2. Slightly later another animation is started on UI runtime

This is fine with current implementation. `cancelAnimation` will synchronously stop the current one without interfering with the new animation. However, with your change we essentialy turn `cancelAnimation` to an asynchronous call.

Therefore, there might be a situation where another animation has started before previous got cancelled, resulting in the cancellation affecting the new animation. That's what I'm afraid of. Here's a video where I show this problem:


https://github.com/user-attachments/assets/5282fec5-16cc-48e3-b785-4f15a4b2cc23

<details><summary>Reproduction snippet</summary>
<p>

import { StyleSheet, View, Button, Text } from 'react-native';

import React from 'react';

import Animated, {
  useSharedValue,
  useAnimatedStyle,
  withTiming,
  cancelAnimation,
  runOnUI,
import { Gesture, GestureDetector } from 'react-native-gesture-handler';

  const rotation = useSharedValue(0);
  const opacity = useSharedValue(1);

    return {
      transform: [{ rotate: rotation.value + 'rad' }],
    };
  });

  const touchOpacity = useAnimatedStyle(() => {
    return {
      opacity: opacity.value,
    };
  });

  const tap = Gesture.Tap()
    .onBegin(() => {
      opacity.value = 0.5;
      rotation.value = withTiming(rotation.value + Math.PI, {
        duration: 2000,
      });
    })
    .onEnd(() => {
      opacity.value = 1;
    });

  return (
      <Button
        title="Start from React"
        onPress={() =>
          runOnUI(() => {
            rotation.value = withTiming(rotation.value + Math.PI, {
              duration: 2000,
            });
          })()
        }
      />
      <Button
        title="Stop from React"
        onPress={() => {
          // Let's pretend React/UI is busy:
          setTimeout(() => cancelAnimation(rotation), 500);
        }}
      />
      <GestureDetector gesture={tap}>
        <Animated.View style={touchOpacity}>
          <Text>Start from UI</Text>
        </Animated.View>
      </GestureDetector>
      <Animated.View style={[animatedStyle, styles.box]} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: {
    marginTop: 20,
    width: 100,
    height: 100,
    backgroundColor: 'powderblue',
  },
});


</p>
</details>

This behavior isn't necessarily bad, but we can't just make such change overnight - it should be well documented and we should provide an escape hatch for the edge case I presented.

I think a reasonable solution here would be to:
- Deprecate `cancelAnimation`.
- Add new `stopAnimation` function. `stopAnimation` would be async on React runtime and sync on UI runtime. It would be the default to replace `stopAnimation`.
- Add new `stopAnimationSync` function. It would be sync on both React and UI runtimes, therefore blocking. It would be the aforementioned escape hatch.

cc @tomekzaw @piaskowyk

tjzel:
@Szymon20000 you are right, I forgot we actually still have the write async in this scenario, since after the read the other runtime is unlocked.

joe-sam:
I don't know whether this is feasible to implement but perhaps the cancelAnimation actually needs to work through a PriorityQueue of Animations in a batched update. This may solve the asynchronous ordering problem, derived animations cancellations and mitigate the round robin blocking time. 
In addition it should be also be quite possible to override the terminal value of the animation from the API on interrupt exception to either default to reset or finalize.

tjzel:
@Szymon20000
Yes, synchronous call on a Worklet Runtime (`_WORKLET === TRUE`) and `runOnUI` on React Runtime should suffice.

There's another edge case where you can call `cancelAnimation` from another Worklet Runtime, not only UI one, but I don't even want to think about it... 🙈

Szymon20000:
@tjzel I've just updated the pr to do exactly what we discussed.

packages/react-native-reanimated/src/animation/util.ts
```diff
@@ -541,5 +542,12 @@ export function defineAnimation<
 export function cancelAnimation<T>(sharedValue: SharedValue<T>): void {
```
tjzel, line 542: Please add to this function's TSDoc information that it works asynchronously.

tjzel (approved):
@Szymon20000 I left only some minor corrections. Address them and we are good to merge. Please use `yarn format` in repo to fix CI.

I think this change is so important we should also port it back to 3.16, 3.15 and 3.14. Thoughts? @bartlomiejbloniarz @tomekzaw

tjzel:
Sorry @Amurmurmur I was OOO for a week

PR #6567, opened 2024-10-03
Add SVG color properties to list of color attributes to parse

jakex7:
## Summary

This PR will significantly enhance the developer experience when animating SVGs, as with these changes, the `SVGAdapter` is no longer required (at least for animating **colors**). It's related to the following PR: https://github.com/software-mansion/react-native-svg/pull/2471

## Test plan

At this point, the changes will not impact the current behavior and need to be tested within the `react-native-svg` repository until the release. Once the RNSVG release is out, I will add an example to the Reanimated repository.

### DIFF
--- a/packages/react-native-reanimated/src/Colors.ts
+++ b/packages/react-native-reanimated/src/Colors.ts
@@ -340,6 +340,12 @@ export const ColorProperties = makeShareable([
   'tintColor',
   'textShadowColor',
   'overlayColor',
+  // SVG color properties
+  'fill',
+  'floodColor',
+  'lightingColor',
+  'stopColor',
+  'stroke',
 ]);
 
 // // ts-prune-ignore-next Exported for the purpose of tests only


tomekzaw (approved):
Thanks for the PR! 🫡

tjzel:
@jakex7 is this change backwards compatible? Reanimated 3.16 will support at least React Native 0.74. Will it work with all `react-native-svg` versions compatible with React Native 0.74?

jakex7:
> @jakex7 is this change backwards compatible? Reanimated 3.16 will support at least React Native 0.74. Will it work with all react-native-svg versions compatible with React Native 0.74? 

Yes, it is. It will support all compatible `react-native` and `react-native-svg` versions.
Currently, it's required to write custom `createAnimatedPropAdapter` like this:

```ts
  props => {
    if (Object.keys(props).includes('fill')) {
      props.fill = {type: 0, payload: processColor(props.fill)};
    }
  },
  ['fill'],
);
```

These changes won't affect the users with custom adapters, as their colors will be passed as `ColorStruct` anyway.

> You said the current changes don't affect current behavior - how comes this will change once new RNSVG is out?

New version of RNSVG will support `ColorValue` type in addition to custom `ColorStruct` on `'color', 'fill', 'floodColor', 'lightingColor', 'stopColor', 'stroke'`

PR #6571, opened 2024-10-04
Add `HostFunction` handling to `_scheduleOnJS`

bartlomiejbloniarz:
## Summary

Currently, when a Host Function is called through `runOnJS` the app crashes, since we are only handling Remote Function in the `_scheduleOnJS` method. For example the following code leads to a crash:

```js
useEffect(() => {
    const id = setInterval(() => {
      console.log('interval');
    }, 1000);
    const intervalAlias = clearInterval;
    runOnUI(() => {
      runOnJS(intervalAlias)(id);
    })();
  }, []);
```

This example works fine on the old architecture, because Timer functions are polyfilled there. On the New Architecture they are Host Functions, so we need to handle them separately.

## Test plan

Check

```js
useEffect(() => {
    const id = setInterval(() => {
      console.log('interval');
    }, 1000);
    const intervalAlias = clearInterval;
    runOnUI(() => {
      runOnJS(intervalAlias)(id);
    })();
  }, []);
```


tomekzaw (approved):
LGTM

packages/react-native-reanimated/Common/cpp/worklets/WorkletRuntime/WorkletRuntimeDecorator.cpp
```diff
@@ -78,39 +94,61 @@ void WorkletRuntimeDecorator::decorate(
...
           } else {
-            auto argsArray =
-                shareableArgs->toJSValue(rt).asObject(rt).asArray(rt);
-            auto argsSize = argsArray.size(rt);
-            std::vector<jsi::Value> args(argsSize);
-            for (size_t i = 0; i < argsSize; i++) {
-              args[i] = argsArray.getValueAtIndex(rt, i);
-            }
-            remoteFun.asObject(rt).asFunction(rt).call(
+            std::vector<jsi::Value> args = parseArgs(rt, shareableArgs);
```
tomekzaw, line 120:
```suggestion
            const auto args = parseArgs(rt, shareableArgs);
```

PR #6577, opened 2024-10-05
refactor(Worklets): Move `messageQueue` to Worklets

tjzel:
## Summary

Moving `MessageQueueThread` to Worklets. It still has the name `REAMessageThread` or similar in some places. This will be addressed in the following PR.

## Test plan

- [x] iOS 0.76 paper/fabric
- [x] android 0.76 paper/fabric

CI does the rest.


packages/react-native-reanimated/Common/cpp/worklets/NativeModules/NativeWorkletsModule.cpp
```diff
@@ -17,9 +18,13 @@ using namespace facebook;
...
 namespace worklets {
 
-NativeWorkletsModule::NativeWorkletsModule(const std::string &valueUnpackerCode)
+NativeWorkletsModule::NativeWorkletsModule(
+    const std::string &valueUnpackerCode,
+    const std::shared_ptr<MessageQueueThread> &jsQueue)
     : NativeWorkletsModuleSpec(nullptr),
-      valueUnpackerCode_(valueUnpackerCode) {}
+      valueUnpackerCode_(valueUnpackerCode),
+      jsQueue_(jsQueue) {}
```
piaskowyk, line 26: It would be safer and more explicite in intentions when we do it like that:
```c++
NativeWorkletsModule::NativeWorkletsModule(
    const std::string &valueUnpackerCode,
    const std::shared_ptr<MessageQueueThread> jsQueue)
    : NativeWorkletsModuleSpec(nullptr),
      valueUnpackerCode_(valueUnpackerCode),
      jsQueue_(std::move(jsQueue)) {}
```

piaskowyk (changes requested):

packages/react-native-reanimated/Common/cpp/worklets/NativeModules/NativeWorkletsModule.cpp
tjzel: This is copy-pasted from how it's done currently in `NativeReanimatedModule.cpp`. Let's address such changes in follow-up pull requests.

piaskowyk: I have your word 🫡

piaskowyk (approved):

PR #6586, opened 2024-10-10
Lazy initalize layout aniamtions in the Commit Hook

bartlomiejbloniarz:
## Summary

Since RN 0.76 the first surface is created after our `initializeTurboModule` method is called on iOS. This change is probably a result of [this PR](https://github.com/facebook/react-native/commit/a778979ed604d1dc7707faf4a3836e3deba136c6#diff-ebba7a8b74cc36d5a03437c3999863076246c6c4eb5559fd1a58719b720fd755). This breaks our initialization of `layoutAnimationsProxy_`, so now we lazy initialize it for every surface in the Commit hook.

## Test plan
Check if layout animations work in the `fabric-example` app.

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
@@ -15,8 +15,11 @@ namespace reanimated {
 
 ReanimatedCommitHook::ReanimatedCommitHook(
     const std::shared_ptr<PropsRegistry> &propsRegistry,
-    const std::shared_ptr<UIManager> &uiManager)
-    : propsRegistry_(propsRegistry), uiManager_(uiManager) {
+    const std::shared_ptr<UIManager> &uiManager,
+    const std::function<void(SurfaceId)> initializeLayoutAnimations)
+    : propsRegistry_(propsRegistry),
+      uiManager_(uiManager),
+      initializeLayoutAnimations_(initializeLayoutAnimations) {
   uiManager_->registerCommitHook(*this);
 }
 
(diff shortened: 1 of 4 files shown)


packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -28,6 +31,16 @@ RootShadowNode::Unshared ReanimatedCommitHook::shadowTreeWillCommit(
...
     RootShadowNode::Shared const &,
     RootShadowNode::Unshared const &newRootShadowNode) noexcept {
+  auto surfaceId = newRootShadowNode->getSurfaceId();
+
+  {
+    auto lock = std::unique_lock<std::mutex>(mutex_);
+    if (!initializedSurfaces_.contains(surfaceId)) {
+      initializeLayoutAnimations_(surfaceId);
+      initializedSurfaces_.insert(surfaceId);
+    }
```
tomekzaw, line 41: Let's store max `SurfaceId` so far and enumerate all surfaces when max changes.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.h
```diff
@@ -15,7 +15,8 @@ class ReanimatedCommitHook : public UIManagerCommitHook {
  public:
   ReanimatedCommitHook(
       const std::shared_ptr<PropsRegistry> &propsRegistry,
-      const std::shared_ptr<UIManager> &uiManager);
+      const std::shared_ptr<UIManager> &uiManager,
+      const std::function<void(SurfaceId)> initializeLayoutAnimations);
```
tomekzaw, line 19:
```suggestion
      const std::function<void(const SurfaceId)> initializeLayoutAnimations);
```

tomekzaw (changes requested):

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
bartlomiejbloniarz: 🫡

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.h
```diff
@@ -44,6 +46,12 @@ class ReanimatedCommitHook : public UIManagerCommitHook {
   std::shared_ptr<PropsRegistry> propsRegistry_;
 
   std::shared_ptr<UIManager> uiManager_;
+
+  std::shared_ptr<LayoutAnimationsProxy> layoutAnimationsProxy_;
+
+  SurfaceId currentMaxSurfaceId_ = -1;
+
+  std::mutex mutex_;
```
tomekzaw, line 54:
```suggestion
  std::mutex mutex_; // Protects `currentMaxSurfaceId_`.
```

tomekzaw (approved):

PR #6598, opened 2024-10-15
Update ReducedMotionConfig.mdx

mzaien:
## Summary

Adding warning that ReducedMotionConfig is applied globally, as I was checking and docs and did not know until I checked the [pr](https://github.com/software-mansion/react-native-reanimated/blob/0bd65b35bbe33c69590e4bfcb836f618ff066534/packages/react-native-reanimated/src/component/ReducedMotionConfig.tsx) this component was introduced in 

## Test plan

check the website after it gets deployed


piaskowyk (approved):
Thanks for the PR! 🙌

PR #6599, opened 2024-10-15
Bump reanimated version to 3.16.0

bartlomiejbloniarz:
## Summary

This PR bumps the Reanimated version to 3.16.0 and updates the compatibility table (there is also a minor Keyframe fix). This PR doesn't follow the usual release PR scheme, as we want to change the way we do releases. Now the `next` reanimated version will be used on main and release package will be made from a separate branch.

### DIFF
--- a/packages/react-native-reanimated/src/layoutReanimation/animationBuilder/Keyframe.ts
+++ b/packages/react-native-reanimated/src/layoutReanimation/animationBuilder/Keyframe.ts
@@ -105,7 +105,7 @@ class InnerKeyframe implements IEntryExitAnimationBuilder {
     const duration: number = this.durationV ? this.durationV : 500;
     const animationKeyPoints: Array<number> = Array.from(
       Object.keys(this.definitions)
-    ).map((value) => parseInt(value, 10));
+    ).map(parseInt);
 
     const getAnimationDuration = (
       key: string,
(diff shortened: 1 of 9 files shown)


packages/react-native-reanimated/src/layoutReanimation/animationBuilder/Keyframe.ts
```diff
@@ -105,7 +105,7 @@ class InnerKeyframe implements IEntryExitAnimationBuilder {
     const duration: number = this.durationV ? this.durationV : 500;
     const animationKeyPoints: Array<number> = Array.from(
       Object.keys(this.definitions)
-    ).map(parseInt);
+    ).map((value) => parseInt(value, 10));
```
tomekzaw, line 108: Shouldn't we move it to a separate PR? 😄

tomekzaw, line 108: I think this change deserves it to be made in a separate PR.

tomekzaw:
What about the code snippet [here](https://github.com/software-mansion/react-native-reanimated/blob/608a9093a7967565b83fa11dd9cb00816ffdf16e/packages/react-native-reanimated/scripts/set-reanimated-version.js#L63-L64)? If we merge this PR but don't publish a release today, we'll accidentally release `3.17.0-nightly`. I think we should remove bumping minor version in this very PR if we want to go this way.

packages/react-native-reanimated/src/layoutReanimation/animationBuilder/Keyframe.ts
bartlomiejbloniarz: 👍

tomekzaw (approved):

PR #6600, opened 2024-10-16
Fix crash on macos when using Layout Animations

Latropos:
## Summary
Previously we were passing a window to function convertPoint, that accepts only view. It was causing a crash, therefore we should pass the contentView of the window.

## Test plan
Tested on several examples:
* Olympic example
* [LA] Basic Layout Animation
* [LA] Entering and Exiting with Layout 
* [LA] Carousel


piaskowyk (approved):
👏

PR #6612, opened 2024-10-18
Move headers to `apple` directory

tomekzaw:
## Summary

This PR moves iOS/macOS/tvOS headers from `apple/reanimated/` to yet another `apple` subdirectory so that the file tree reflects the paths in `#import <...>`.

After this change, `apple/reanimated` can be added to `HEADER_SEARCH_PATHS` and imports like `reanimated/sensor/ReanimatedSensor.h` will work properly.


efstathiosntonas:
Thanks @tomekzaw!

ps. fking headers are so annoying to work with

bartlomiejbloniarz (approved):

tomekzaw:
@efstathiosntonas Just FYI, today later on I will also submit another PR that should fix the headers when linking statically.

Issue #6607, opened 2024-10-17
[3.16.0] 'worklets/WorkletRuntime/ReanimatedHermesRuntime.h' file not found


efstathiosntonas:
### Description

After upgrading from 3.15.5 I get this error on `ReanimatedHermesRuntime`:

`'worklets/WorkletRuntime/ReanimatedHermesRuntime.h' file not found`

Tried with no luck:

1. pod deintegrate
2. removed derived data
3. rm -rf node_modules
4. install everything back

### Steps to reproduce

1. upgrade from 3.15.5
5. cannot build
6. xcode 16

### Snack or a link to a repository

none

### Reanimated version

7.16.0

### React Native version

0.75.4

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Acknowledgements

Yes

efstathiosntonas:
I believe this is the offending PR: 

* https://github.com/software-mansion/react-native-reanimated/pull/6543

I use static framework on Podfile: 

`use_frameworks! :linkage => :static`

piaskowyk:
You are so fast man! 🤯

tomekzaw:
Working on it ⌛

tomekzaw:
@werlensantos Please wait for 3.16.1

tomekzaw:
Here's a workaround: https://github.com/software-mansion/react-native-reanimated/pull/6614

It also requires changes from https://github.com/software-mansion/react-native-reanimated/pull/6612

Waiting for an internal review

efstathiosntonas:
> @werlensantos Please wait for 3.16.1 early next week

it will be available on nightly, right? 

thanks @tomekzaw for the quick fixes!

ethanneff:
should you remove the `3.16.0` version since it is broken?

efstathiosntonas:
> should you remove the `3.16.0` version since it is broken?

I believe that the only projects that are affected are the ones using react-native-firebase so it’s not entirely broken.

efstathiosntonas:
@tomekzaw sure thing, will report back tomorrow morning.

efstathiosntonas:
@tomekzaw tested `3.17.0-nightly-20241018-16297eb3c` and life got back to normal.

tomekzaw:
[...]

PR #6614, opened 2024-10-18
Fix static linking on iOS

tomekzaw:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/6607.

This is more of a workaround to fix issue with missing headers when building Reanimated with static linkage for iOS. I spent several hours on trying to make it right but it would work correctly only partially. I will post my findings here or internally later on and I hope to find a better solution sometime in the future.

This workaround adds paths to `Common/cpp` and `apple/` directories to `HEADER_SEARCH_PATHS` – the preprocessor looks for headers in these locations. I can't use an absolute path (calculated in Ruby) because it would make the output setup-dependent and thus affect checksums in `Podfile.lock` which is unwanted as some apps check against checksum changes on CI.
 
## Test plan

Tested on fabric-example app with `USE_FRAMEWORKS=static` and without.


bartlomiejbloniarz (approved):

efstathiosntonas:
@tomekzaw hi, quick question: 

when  `pod install` after the changes in the podspec, is it necessary to clean derived data/build folder/Pods folder before rebuilding?

tomekzaw:
Interesting comment: https://github.com/facebook/react-native/issues/38067#issuecomment-2474228095

PR #6618, opened 2024-10-21
[LA] Update window update check

bartlomiejbloniarz:
## Summary

This PR changes the way we keep track of the window dimensions in Layout Animations on the New Arch. The previous version didn't work on Android in release mode. This new approach is possible because we moved the initialization of layout animations to the commit hook in #6586.

## Test plan

Check if Layout Animations dependent on `windowWidth` (e.g. SlideInLeft) work properly.


tomekzaw (approved):

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -784,13 +784,12 @@ void LayoutAnimationsProxy::maybeRestoreOpacity(
 void LayoutAnimationsProxy::maybeUpdateWindowDimensions(
     facebook::react::ShadowViewMutation &mutation,
     SurfaceId surfaceId) const {
-  // This is a hacky way to obtain the window dimensions.
-  // We can identify the root, by checking if its tag is equal to the surfaceId
-  if (mutation.parentShadowView.tag == surfaceId) {
+  if (mutation.type == ShadowViewMutation::Update &&
+      !std::strcmp(mutation.oldChildShadowView.componentName, "RootView")) {
```
tomekzaw, line 788: Instead of hard-coding `"RootView"` here, can we use `RootComponentName` from `RootShadowNode.h`?

PR #6631, opened 2024-10-25
fix: Invalid shared value access displaying when code is correct

MatiPl01:
## Summary

The problem was caused by our `useDerivedValue` hook which set ref to `null` in the `useEffect` cleanup method.

```tsx
useEffect(() => {
  return () => {
    initRef.current = null;
  };
}, []);
```

The warning was shown during hot reload, because react calls cleanup functions in hooks, thus the `initRef` was reset in the `useDerivedValue` hook, which resulted in execution of this code that should run only when the component mounts for the first time:

```tsx
  initRef.current = makeMutable(initialUpdaterRun(updater));
}
```

Because of that, `initialUpdaterRun` executed the `updater` callback on the JS thread during hot reload, while the component was rendered, which resulted in reading shared value's `.value` property during the render phase.


packages/react-native-reanimated/src/logger/logger.ts
```diff
@@ -39,7 +39,7 @@ function logToConsole(data: LogData) {
 export const DEFAULT_LOGGER_CONFIG: LoggerConfigInternal = {
   logFunction: logToConsole,
   level: LogLevel.warn,
-  strict: true,
+  strict: false,
```
MatiPl01, line 42: I decided to change this to `false` as the invalid raed/write warning was reported to be too annoying and it might be better to keep it disabled by default.

packages/react-native-reanimated/src/hook/useDerivedValue.ts
```diff
@@ -73,11 +73,5 @@ export function useDerivedValue<Value>(
     };
   }, dependencies);
 
-  useEffect(() => {
-    return () => {
-      initRef.current = null;
-    };
-  }, []);
-
```
tjzel, line 81: It seems to be ok to remove it since `useSharedValue` doesn't have any cleanup function.

packages/react-native-reanimated/src/logger/logger.ts
MatiPl01: I saw a [tweet](https://x.com/hirbod_dev/status/1849491666610077954?s=46&t=rZ58uY0MGSMkjVkZUDCj3g) in which people complain about these warnings so I though that it may be too annoying. Maybe we shouldn't bother and we should encourage people to fix invalid shared value access issues on their side, though. I don't know which approach we prefer.

MatiPl01: We agreed to leave this warning enabled by default.

tjzel (approved):
Let's go!!!

PR #6632, opened 2024-10-25
docs: Change strict mode to false in default logger config in docs

MatiPl01:
## Summary

This PR adjusts docs to match changes introduced in the #6631 PR.


MatiPl01:
We [decided](https://github.com/software-mansion/react-native-reanimated/pull/6631#discussion_r1816723699) to keep the warning turned on by default. People should fix invalid usage of the shared value on their side.

PR #6634, opened 2024-10-25
fix: Side effects paths in package.json

MatiPl01:
## Summary

This PR adds missing initializers path to side effects (because some global variables are written in this file). It also fixes paths to remaining files (as there is no longer `reanimated2` directory).


tjzel (approved):
LGTM

PR #6647, opened 2024-10-29
feat: Support for multiple surfaces on Fabric

patrycjakalinska:
## Summary

This PR introduces handling multiple surfaces on Fabric.

## Test plan

🚀


packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -704,72 +701,77 @@ void NativeReanimatedModule::performOperations() {
...
-              return nullptr;
+  
+    
+    
+        bool hasLayoutUpdates = false;
+        
+        for (const auto &[shadowNode, props] : copiedOperationsQueue) {
+            if (isThereAnyLayoutProp(rt, props->asObject(rt))) {
+                hasLayoutUpdates = true;
+                break;
```
tomekzaw, line 712: Please format code here by running `yarn format:common` in `packages/react-native-reanimated` directory.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -704,72 +701,77 @@ void NativeReanimatedModule::performOperations() {
...
+            return;
+        }
+        
+        
+        react_native_assert(uiManager_ != nullptr);
+        const auto &shadowTreeRegistry = uiManager_->getShadowTreeRegistry();
+        
+    for (auto &[shadowNode, props] : copiedOperationsQueue) {
+        SurfaceId surfaceId_ = shadowNode->getSurfaceId();
+        shadowTreeRegistry.visit(surfaceId_, [&](ShadowTree const &shadowTree) {
```
tomekzaw, line 742: I'm not sure if this is an optimal approach here in terms of execution time – in particular, we don't know how expensive is a call to `shadowTreeRegistry.visit`. In the vast majority of apps, there will be only one surface so we should heavily optimize for that use case. If there are more surfaces, maybe we should group the updates by surface ID so we visit each surface exactly once?

patrycjakalinska: Grouping by surfaceId was my initial thought! I will change it accordingly

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/NativeReanimatedModule.cpp
```diff
@@ -735,41 +733,46 @@ void NativeReanimatedModule::performOperations() {
...
-            auto family = &shadowNode->getFamily();
-            react_native_assert(family->getSurfaceId() == surfaceId_);
-            propsMap[family].emplace_back(rt, std::move(*props));
+  std::unordered_map<SurfaceId, PropsMap> propsMapBySurface;
+
+  for (auto const &[shadowNode, props] : copiedOperationsQueue) {
+    SurfaceId surfaceId = shadowNode->getSurfaceId();
+    auto family = &shadowNode->getFamily();
+    react_native_assert(family->getSurfaceId() == surfaceId);
+    propsMapBySurface[surfaceId][family].emplace_back(rt, std::move(*props));
```
tomekzaw, line 742: Maybe it makes sense to populate `propsMapBySurface` directly in `updateProps` function? On the other hand, I don't think having it as a separate step makes much difference in terms of execution time and performance.

patrycjakalinska: I was thinking that it is more readable (and kinda related to neighbour code) in current place (at least for me), but if you would like to put it in `updateProps`, I'm open for any changes

tomekzaw:
Please merge latest main to fix macOS CI, looks like tvOS is still broken on other PRs

piaskowyk:
I have tested it, and it crashes sometimes for me after a reload. What I did was I opened the Bokeh Example and pressed R to reload the app - on Android, it stops the animation, but on iOS, it crashes. Could you verify if this happened before as well?

patrycjakalinska:
> I have tested it, and it crashes sometimes for me after a reload. What I did was I opened the Bokeh Example and pressed R to reload the app - on Android, it stops the animation, but on iOS, it crashes. Could you verify if this happened before as well?

We established that this issue is not related to changes from this PR.

piaskowyk (approved):
I've tested it and it seems to be ok 👍

PR #6658, opened 2024-10-31
chore: Remove dead code for RN 73 and below

tjzel:


packages/react-native-reanimated/__typetests__/72plus/index.tsx
tjzel, line 1: Tests cases from this file were taken to update those in `common`.

packages/react-native-reanimated/android/build.gradle
```diff
@@ -346,63 +346,41 @@ android {
                 srcDirs += "src/paper/java"
             }
 
-            // messageQueueThread
-            if (REANIMATED_MAJOR_VERSION > 2) {
-                if (REACT_NATIVE_MINOR_VERSION <= 72) {
-                    srcDirs += "src/reactNativeVersionPatch/messageQueueThread/72"
-                } else {
-                    srcDirs += "src/reactNativeVersionPatch/messageQueueThread/latest"
```
tjzel, line 354: This file was moved outside source sets.

packages/react-native-reanimated/android/build.gradle
```diff
@@ -346,63 +346,41 @@ android {
...
+                srcDirs += "src/reactNativeVersionPatch/ReanimatedUIManager/"
             } else {
                 srcDirs += "src/reactNativeVersionPatch/ReanimatedUIManager/latest"
             }
 
-            // ReactHost
-            if (REACT_NATIVE_MINOR_VERSION <= 72) {
-                srcDirs += "src/reactNativeVersionPatch/ReactHost/72"
-            } else {
-                srcDirs += "src/reactNativeVersionPatch/ReactHost/latest"
```
tjzel, line 371: This file was moved outside source sets.

packages/react-native-reanimated/src/createAnimatedComponent/getViewInfo.ts
tjzel, line 1: Changing this file would be too time consuming for now.

packages/react-native-reanimated/src/fabricUtils.ts
tjzel, line 1: Changing this file would be too time consuming for now.

packages/react-native-reanimated/__typetests__/common/inlineStylesTest.tsx
```diff
@@ -44,7 +44,7 @@ function InlineStylesTest() {
   }
 
   function InlineStylesTest4() {
-    const sv = useSharedValue(true);
+    const sv = useSharedValue('0');
```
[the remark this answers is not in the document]
tjzel: They are related. The original test was supposed to check if passing a string here would be caught, since it's not an allowed type. However, in previous versions it wasn't caught in types of React Native and since `AnimatedStyle` types are derived from React Native, we didn't catch it either. Now this test could finally get updated.

PR #6663, opened 2024-11-04
[Fabric LA] Fix `maybeDropAncestors` condition.

### LINKED ISSUE #6644: React Native 0.76: Previous Item Remains Visible on State-Driven Navigation
I'm experiencing an issue where I'm changing the view on the same page according to the state value, effectively toggling between two different views. However, when I switch from the first view (page) to the second one, an item that was selected in the first page unexpectedly reappears somewhere within the FlatList on the second page as I scroll. It’s as if the selected item from the previous page is "stuck" on the screen, appearing out of context on the new view.

This issue is specific to the new React Native 0.76 version and the new architecture; everything works as expected in React Native 0.75.

Here are some additional observations that might help clarify the problem:

 I expected entering and exiting animations to work when bound to a key value in the new architecture, as most examples demonstrate this usage. However, this doesn’t seem to be functioning correctly in my setup.

Since this issue only occurs in the new architecture, I suspect the problem could be library-related.

Any advice on resolving this would be greatly appreciated!

Ekran Resmi 2024-10-28 21 17 10
Ekran Resmi 2024-10-28 21 17 19

### Steps to reproduce

As you can see from the Snack website, that's the entire code

https://snack.expo.dev/QIF-21OUXgM12PC_8cqZR

### Reanimated version

3.16.1

### React Native version

0.76

### Platforms

iOS

### JavaScript runtime

None

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & production bundle

### Device

iOS simulator

### Device model

iPhone 16 Pro and Real Device iPhone 12

### Acknowledgements

Yes

bartlomiejbloniarz:
## Summary

In the `maybeDropAncestors` function we can remove the view if it has no remaining animating views. Let's say we have nested exiting animations:

 ```mermaid
 flowchart TD
     A((A: EXITING))
     B((B: EXITING))
     C((C: WAITING))
     D((D: WAITING))
     E((E: WAITING))
 
     A --> B
     A --> C
     A --> D
     A --> E
 
 ```
 In the current implementation in this case if the animation in `B` ended before `A`, we would visit `A` in `maybeDropAncestors` and decided to remove `A`, even though it still has some waiting children. Then `A` would be added to the view recycling pool while still having children. This would cause us to see some zombie views when the view is reused. 
 
I changed the `maybeDropAncestors` condition to check the size of the `children` list. I also removed `node->animatedChildren` as I think it is no longer necessary.

Fixes #6644 

## Test plan
Chceck `[LA] View recycling` example, if there are no zombie views in the `WheelPicker` component.


piaskowyk (approved):
Smart 👍  
Just to be sure if I understand correctly - We can simply rely on checking the count of children since they are detached before the parent, correct?

PR #6667, opened 2024-11-05
fix: Allow TypeScript in worklet classes

### LINKED ISSUE #6642: Type annotations are not supported in __workletClass
### Description

A class that has the __workletClass annotation doesn't support any kind of type annotations.
The following works:
```tsx
class Test {
  __workletClass = true;
  value;

  constructor(value) {
    this.value = value;
  }
}
```

But the following wouldn't:
```
class Test {
  __workletClass = true;
  value: number;

  constructor(value: number) {
    this.value = value;
  }
}
```
or
```
class Test<T> {
  __workletClass = true;
  value: T;

  constructor(value: T) {
    this.value = value;
  }
}
```

### Steps to reproduce

Add a type annotation to a class that has the __workletClass property. Adding a type annotations to https://github.com/software-mansion/react-native-reanimated/blob/main/apps/common-app/src/examples/RuntimeTests/tests/plugin/fileWorkletization.ts should be enough to reproduce the issue.

https://github.com/shopify/react-native-skia

### Reanimated version

3.5.1

### React Native version

0.75.2

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

tjzel:
## Summary

Turns out `@babel/preset-typescript` always has to be included when calling `transformSync` from `babel` (it cannot be included in a later call). Due to that I made all `transformSync` calls into a wrapper call that uses a required set of plugins and presets.

Fixes 
- #6642

## Test plan

- [x] Plugin unit tests pass
- [x] Runtime tests pass


packages/react-native-reanimated/plugin/src/transform.ts
```diff
@@ -0,0 +1,26 @@
...
+  return transformSync(code, {
+    ...rest,
+    plugins: [...plugins, ...extraPlugins],
+    presets: [...presets, ...extraPresets],
+  });
+}
+
+const presets: PluginItem[] = [require.resolve('@babel/preset-typescript')];
+
+const plugins: PluginItem[] = [];
```
tomekzaw, line 19:
```suggestion
const defaultPresets: PluginItem[] = [require.resolve('@babel/preset-typescript')];

const defaultPlugins: PluginItem[] = [];
```

packages/react-native-reanimated/plugin/src/workletFactory.ts
```diff
@@ -469,3 +454,14 @@ function makeArrayFromCapturedBindings(
 
   return Array.from(closure.values());
 }
+
+const plugins = [
```
tomekzaw, line 458:
```suggestion
const extraPlugins = [
```

packages/react-native-reanimated/plugin/src/workletFactory.ts
```diff
@@ -91,10 +77,9 @@ export function makeWorkletFactory(
   codeObject.code =
     '(' + (fun.isObjectMethod() ? 'function ' : '') + codeObject.code + '\n)';
 
-  const transformed = transformSync(codeObject.code, {
+  const transformed = workletTransformSync(codeObject.code, {
+    extraPlugins: plugins,
```
tomekzaw, line 81:
```suggestion
    extraPlugins,
```

apps/common-app/src/examples/RuntimeTests/tests/plugin/recursion.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useEffect } from 'react';
 import { View } from 'react-native';
-import { useSharedValue, runOnUI, runOnJS } from 'react-native-reanimated';
+import { useSharedValue, runOnUI } from 'react-native-reanimated';
 import { render, wait, describe, getRegisteredValue, registerValue, test, expect } from '../../ReJest/RuntimeTestsApi';
```
wcandillon, line 4: This is very cool. Could type annotations be added to the test that test file level "worklet" annotations? To be on the safe side?

bartlomiejbloniarz (approved):

PR #6670, opened 2024-11-06
fix: Break circular reference in `AndroidUIScheduler`

lukmccall:
## Summary

Breaks circular reference in the `AndroidUIScheduler`.

`AndroidUIScheduler` is a `HybridClass` that has a C++ counterpart holding a global reference to the Java object. This structure creates a circular reference between C++ and Java, which the garbage collector cannot clean up. To resolve this issue, I manually remove the reference between Java and C++ by calling `resetNative` when the scheduler is deactivated.

## Test plan

- tested in Expo Go ✅


tomekzaw (approved):

lukmccall:
Closes in favour of https://github.com/software-mansion/react-native-reanimated/pull/6697.

PR #6671, opened 2024-11-06
fix: Break circular reference in `LayoutAnimation`

lukmccall:
## Summary

Breaks circular reference in the `LayoutAnimations`.

`LayoutAnimations ` is a `HybridClass` with a C++ counterpart holding a global reference to the Java object. This structure creates a circular reference between C++ and Java, which the garbage collector cannot clean up. To resolve this issue, I manually removed the reference between C++ and Java by resetting the global ref during the destructor of the `NativeProxy`.

## Test plan

- tested in Expo Go ✅

### DIFF
--- a/packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
+++ b/packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
@@ -82,6 +82,8 @@ NativeProxy::~NativeProxy() {
   // has already been destroyed when AnimatedSensorModule's
   // destructor is ran
   reanimatedModuleProxy_->cleanupSensors();
+
+  layoutAnimations_->cthis()->invalidate();
 }
 
 jni::local_ref<NativeProxy::jhybriddata> NativeProxy::initHybrid(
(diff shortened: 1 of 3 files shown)


tomekzaw (approved):

packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
```diff
@@ -110,6 +110,8 @@ NativeProxy::~NativeProxy() {
   // has already been destroyed when AnimatedSensorModule's
   // destructor is ran
   nativeReanimatedModule_->cleanupSensors();
+
+  layoutAnimations_->cthis()->invalidate();
```
tjzel, line 114: I've been dealing a lot with missing invalidations in https://github.com/software-mansion/react-native-reanimated/pull/6850 and I have some thoughts.

Shouldn't we call it from Java when Native Modules are invalidated, for a more transparent flow?

`(java)ReanimatedModule.invalidate() -> ... -> (java)NativeProxy.invalidate() -> (hybrid)LayoutAnimations->invalidate()`

I don't think that relying on Native Proxy's destructor is the best choice here. On the other hand, this is safer since NativeProxy can be "safely" used until it's destroyed.

What do you think?

piaskowyk (approved):

packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
lukmccall: That would be better, but it really depends on what you want to accomplish. I'm okay with either approach: calling it in the destructor or during the invalidation flow. I don’t think relying on the Native Proxy's destructor is a bad idea, since we are clearing members of the Native Proxy, which means it can be called from the destructor. If I remember correctly, removing it from the Java side was somewhat challenging, but I don’t recall the exact reason.

Issue #6677, opened 2024-11-08
useAnimatedStyle gets out of sync with SharedValue modified inside useEffect on Android


gaearon:
### Description

Here's the code.

```js
import React, {useState, useEffect} from 'react';
import {Pressable, Text} from 'react-native';
import Animated, {
  useAnimatedStyle,
  useSharedValue,
} from 'react-native-reanimated';

export default function App() {
  const [show, setShow] = useState(false);
  return show ? (
    <Lightbox onClose={() => setShow(false)} />
  ) : (
    <Pressable
      style={{padding: 100, fontSize: 50, backgroundColor: 'purple'}}
      onPress={() => setShow(true)}>
      <Text>open</Text>
    </Pressable>
  );
}

function Lightbox({onClose}) {
  const openProgress = useSharedValue(0);

  useEffect(() => {
     console.log('set to 1');
     openProgress.value = 1;
  }, [openProgress]);

  const containerStyle = useAnimatedStyle(() => {
    console.log(openProgress.value);
    if (openProgress.value < 1) {
      return {backgroundColor: 'red'};
    }
    return {backgroundColor: 'blue'};
  });

  return (
    <Animated.View
      style={[
        {
          position: 'absolute',
          top: 0,
          left: 0,
          right: 0,
          bottom: 0,
        },
        containerStyle,
      ]}>
      <Pressable style={{flex: 1}} onPress={onClose} />
    </Animated.View>
  );
}
```

### Steps to reproduce

1. Use Paper (this doesn't repro on Fabric)
2. Put the code above in your project
3. Run on Android

Repro case:

1. Press "open"
2. Tap on the opened blue screen (which will close it)
3. Press "open" again

Expected: the value will change `0 -> 1` and the screen will again be blue

Actual: the value will change `0 -> 1` **but the screen will remain red**

This only reproduces on Android Paper.

Wrapping `.value` assignment in `setTimeout` with `1` seems to "fix" it.

https://github.com/user-attachments/assets/9d3a42f2-a4ae-459d-a5e5-507034a74837

### Snack or a link to a repository

https://github.com/gaearon/reanimated-repro-weird

### Reanimated version

3.16.1

### React Native version

0.76.1, but repros on 0.74.1 so is not new

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

Android emulator

### Acknowledgements

Yes

[...]

PR #5767, opened 2024-03-07
 Prevent crash on non-existent view updates in Android

thomas-rx:
## Summary

I'm opening this PR to address a critical issue previously tackled by PR #5216, which has stalled due to the original author's inactivity. 

This fix is crucial for resolving significant bugs affecting Android view property updates (referenced in issues #4505, #4231, #3860) that impact application stability and performance.

Typical error : 
```unknown:NativeViewHierarchyManager: Unable to update properties for view tag 943
03-07 19:54:33.543 19722 19722 E unknown:NativeViewHierarchyManager: com.facebook.react.uimanager.IllegalViewOperationException: ViewManager for tag 943 could not be found.
```

### Before patching
https://github.com/software-mansion/react-native-reanimated/assets/40337934/a7cd7f8c-c0a3-4a84-98f7-4853886799f0

### After patching
https://github.com/software-mansion/react-native-reanimated/assets/40337934/c4dfe495-fed9-48cf-a227-e7768cec26cc

PR #6685, opened 2024-11-10
fix(Android, Paper): Lost update on render

tjzel:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/6677

Due to the React Native Android's batching of Native Views creation, sometimes the update from Reanimated for a component might come before the corresponding view was actually created. This used to crash the app which was fixed in:
- https://github.com/software-mansion/react-native-reanimated/pull/5767)

However, the above fix didn't address the fact the update was lost. This pull request re-applies first Animated Style update so then the update would be applied on the second frame, instead of not at all.

## Test plan

Test it on the reproduction in https://github.com/software-mansion/react-native-reanimated/issues/6677

@bartlomiejbloniarz Do you think it can affect Layout Animations?

## Note

This should probably be handled in Android specific native code but I failed there to force the UIManager to create the View when we have an update pending.

### DIFF
--- a/packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
+++ b/packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
@@ -35,6 +35,7 @@ import type {
 } from '../commonTypes';
 import { isWorkletFunction } from '../commonTypes';
 import { ReanimatedError } from '../errors';
+import { Platform } from 'react-native';
 
 const SHOULD_BE_USE_WEB = shouldBeUseWeb();
 
@@ -43,6 +44,7 @@ interface AnimatedState {
   animations: AnimatedStyle<any>;
   isAnimationRunning: boolean;
   isAnimationCancelled: boolean;
+  isFirstRun: boolean;
 }
 
 interface AnimatedUpdaterData {
@@ -465,6 +467,7 @@ For more, see the docs: \`https://docs.swmansion.com/react-native-reanimated/doc
         animations: {},
         isAnimationCancelled: false,
         isAnimationRunning: false,
+        isFirstRun: true,
       }),
       viewDescriptors: makeViewDescriptorsSet(),
     };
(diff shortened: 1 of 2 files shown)


joe-sam:
Possibly the updater function argument  for _useAnimatedStyle_/_useAnimatedProps_ should just include previous style/prop argument and leave it to the end user of the API to decide whether to set the animation for style/prop or not.
If the component is not created yet the callback argument _prevStyleOrProp_ should be undefined instead of throwing some unknown view tag error which cannot be handled elsewhere in the stack.

_useAnimatedStyleOrProps ( (prevStyleOrProp : Style | Props | undefined) => { return nextStyleOrProp } , ...)_

This  updates applied to the later frame may not be idempotent, associative or distributive.
 The flags will also need to be cleaned up on unmounting.

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
```diff
@@ -511,11 +514,44 @@ For more, see the docs: \`https://docs.swmansion.com/react-native-reanimated/doc
...
+          */
+          requestAnimationFrame(() =>
+            styleUpdater(
+              shareableViewDescriptors,
+              updater,
+              remoteState,
+              areAnimationsActive,
+              isAnimatedProps
+            )
+          );
```
piaskowyk, line 544: We should consider resolving this in native code - similar to the fixes we've already implemented for iOS:
- https://github.com/software-mansion/react-native-reanimated/pull/2580
- https://github.com/software-mansion/react-native-reanimated/pull/5224

The current solution seems a bit undeterministic and could potentially cause some unexpected blinking 🤔

joe-sam:
Could this be the underlying issue that views are intentionally delayed while mounting, and that there appears to be a retry mechanism already in place at least on Android(Paper) [synchronise dispatching of view commands through RuntimeScheduler #47604 ](https://github.com/facebook/react-native/pull/47604).

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
tjzel: Honestly I don't believe it would lead to blinks, since the update is never stale - but probably during some heavy traffic and slow JS/UI the update could still get lost (although a lot less likely).

Implementing it in the native code is the best solution here, but then it's a code that would need to be maintained and it steers us away from improving Reanimated for Fabric 😿

tjzel:
After giving it some thought we decided that this PR won't be merged to the main branch. This implementation is too problem-specific and we don't want to put more effort into supporting Paper with more general approach.

Issue #6202, opened 2024-07-02
useAnimatedProps not updating on web


oxbits:
### Description

when i go to here :

https://docs.swmansion.com/react-native-reanimated/examples/slider

the TextInput updates as you slide the slider

when i run the snack web client the TextInput does not update

[same issue when i run the code locally in my workspace on the web]

### Steps to reproduce

go to this snack and run on web, the TextInput does not update : 

~~https://snack.expo.io/?platform=android&name=Slider%20%7C%20React%20Native%20Reanimated&dependencies=react-native-reanimated%40*%2Creact-native-gesture-handler%40*%2C%40shopify%2Fflash-list%40*&hideQueryParams=true&sourceUrl=https%3A%2F%2Fdocs.swmansion.com%2Freact-native-reanimated%2Fexamples%2FSlider.js~~

(better snack version:)

https://snack.expo.dev/@oxbits/slider-%7C-react-native-reanimated

### Snack or a link to a repository

~~https://snack.expo.io/?platform=android&name=Slider%20%7C%20React%20Native%20Reanimated&dependencies=react-native-reanimated%40*%2Creact-native-gesture-handler%40*%2C%40shopify%2Fflash-list%40*&hideQueryParams=true&sourceUrl=https%3A%2F%2Fdocs.swmansion.com%2Freact-native-reanimated%2Fexamples%2FSlider.js~~

(better snack version:)

https://snack.expo.dev/@oxbits/slider-%7C-react-native-reanimated

### Reanimated version

3.10.1 or 3.12.1 [i tried both and others too]

### React Native version

0.74

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

all of this makes me wonder why it works on the documentation web page but not in the snack or on my local box ?

oxbits:
@gihub-actions

i have provided the link to the snack above

here it is again :

https://snack.expo.io/?platform=android&name=Slider%20%7C%20React%20Native%20Reanimated&dependencies=react-native-reanimated%40*%2Creact-native-gesture-handler%40*%2C%40shopify%2Fflash-list%40*&hideQueryParams=true&sourceUrl=https%3A%2F%2Fdocs.swmansion.com%2Freact-native-reanimated%2Fexamples%2FSlider.js

andreshsingh:
+1, we're facing the same issue

oxbits:
[...]

PR #6691, opened 2024-11-12
fix: useAnimatedProps text not working on web

5ZYSZ3K:
fix: web-example not starting because of metro-resolver version mismatch
feature: Add AnimatedPathExample to the repo

## Summary

- Fixed an issue that on web you were not able to use useAnimatedProps to animate text on TextInput (#6202)
- Fixed the version mismatch for web-example app

## Test plan

You can check example app. Examples:
- AmountExample
- JSPropsExample

### DIFF
--- a/apps/common-app/src/apps/reanimated/examples/AmountExample.tsx
+++ b/apps/common-app/src/apps/reanimated/examples/AmountExample.tsx
@@ -3,11 +3,20 @@ import Animated, {
   useDerivedValue,
   useSharedValue,
   withTiming,
+  createAnimatedPropAdapter,
 } from 'react-native-reanimated';
-import { Button, StyleSheet, TextInput, View } from 'react-native';
+import { Button, Platform, StyleSheet, TextInput, View } from 'react-native';
 import React, { useCallback } from 'react';
 
 Animated.addWhitelistedNativeProps({ text: true });
+
+const textAdapter = createAnimatedPropAdapter(
+  (props: Record<string, unknown>) => {
+    props._setAttributeDirectly = true;
+    props.value = props.text;
+  },
+  ['value']
+);
 
 const AnimatedTextInput = Animated.createAnimatedComponent(TextInput);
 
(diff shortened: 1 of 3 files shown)


patrycjakalinska:
Hi @5ZYSZ3K, thanks for creating this pull request c: Can you split it so one PR would have only one responsibility? Currently we have 2 fixes and one feature, ideally each one of these would be separate. 

About `fix: web-example not starting because of metro-resolver version mismatch` - i think currently the `web-example` is starting (kudos to @tomekzaw and [this PR](https://github.com/software-mansion/react-native-reanimated/pull/6886)), so no need for that.

oxbits:
Hello @5ZYSZ3K and @patrycjakalinska 

We see our issue has been referenced in this pull request.

We see another pull request has been referenced but we do not think it fixes the issue we are experiencing.

> ## Summary
> * Fixed an issue that on web you were not able to use useAnimatedProps to animate text on TextInput ([useAnimatedProps not updating on web #6202](https://github.com/software-mansion/react-native-reanimated/issues/6202))

5ZYSZ3K:
> Hi @5ZYSZ3K, thanks for creating this pull request c: Can you split it so one PR would have only one responsibility? Currently we have 2 fixes and one feature, ideally each one of these would be separate.
> 
> About `fix: web-example not starting because of metro-resolver version mismatch` - i think currently the `web-example` is starting (kudos to @tomekzaw and [this PR](https://github.com/software-mansion/react-native-reanimated/pull/6886)), so no need for that.

Hi, I've rebased it and got rid of an additional example - now it's just the fix + a showcase on already existing examples (which were affected by the bug)

5ZYSZ3K:
A quick explanation on why wasn't it working:
Updating `text` prop on `TextInput` won't work because `React` implements `_valueTracker` there, so I added the possibility to set a value directly
https://github.com/facebook/react/issues/11488
https://github.com/facebook/react/blob/main/packages/react-dom-bindings/src/client/inputValueTracking.js

patrycjakalinska:
hi @5ZYSZ3K, thanks for finding the underlying issue in RN `_valueTracker`, big big kudos. Your quick fix was okay (as we talked!) but I've opened a PR with different approach, that would be easier for us to maintain. Again big thanks!

PR #6692, opened 2024-11-12
chore: Describe `isWorkletRuntime` behavior on different runtimes

tjzel:
## Summary

@latekvo told me it's poorly documented so I updated it


packages/react-native-reanimated/src/commonTypes.ts
```diff
@@ -408,10 +408,8 @@ export type WorkletFunction<
  *
  * ### Maintainer note
  *
- * This function works well on the JS thread performance-wise, since the JIT can
- * inline it. However, on other threads it will not get optimized and we will
- * get a function call overhead. We want to change it in the future, but it's
- * not feasible at the moment.
+ * This function is supposed to be used only in the React Runtime. It always
+ * returns `true` in Worklet Runtimes.
```
bartlomiejbloniarz, line 412:
```suggestion
 * returns `false` in Worklet Runtimes.
```
I think it's the other way around

bartlomiejbloniarz (approved):

PR #6694, opened 2024-11-12
chore(Worklets): SingleInstanceChecker

tjzel:
## Summary

Moving `SingleInstanceChecker` to Worklets and applying it there.

## Test plan

🚀

### DIFF
--- a/packages/react-native-reanimated/apple/worklets/apple/WorkletsModule.mm
+++ b/packages/react-native-reanimated/apple/worklets/apple/WorkletsModule.mm
@@ -1,6 +1,7 @@
 #import <React/RCTBridge+Private.h>
 #import <worklets/WorkletRuntime/RNRuntimeWorkletDecorator.h>
 #import <worklets/apple/WorkletsModule.h>
+#import <worklets/tools/SingleInstanceChecker.h>
 
 using worklets::NativeWorkletsModule;
 using worklets::RNRuntimeWorkletDecorator;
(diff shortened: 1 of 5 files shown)


packages/react-native-reanimated/apple/worklets/apple/WorkletsModule.mm
```diff
@@ -16,6 +17,9 @@ - (void)_tryAndHandleError:(dispatch_block_t)block;
 
 @implementation WorkletsModule {
   std::shared_ptr<NativeWorkletsModule> nativeWorkletsModule_;
+#ifndef NDEBUG
+  SingleInstanceChecker<WorkletsModule> singleInstanceChecker_;
```
piaskowyk, line 21: I think we should either change it to `worklets::SingleInstanceChecker` or add `using worklets` to avoid potential compilation errors

piaskowyk (changes requested):
Just one change, the rest of changes is ok 👍

packages/react-native-reanimated/apple/worklets/apple/WorkletsModule.mm
tjzel: Yeah I noticed this on CI, fixed [f47d857](https://github.com/software-mansion/react-native-reanimated/pull/6694/commits/f47d8576f9a01c09a3b304e21a40ae95da56e313)

piaskowyk (approved):

PR #6697, opened 2024-11-12
fix: Break circular reference in `NativeProxy` and `AndroidUIScheduler`

lukmccall:
## Summary

Breaks circular reference in `NativeProxy` and `AndroidUIScheduler`.
Needs https://github.com/software-mansion/react-native-reanimated/pull/6671 to work. 

`NativeProxy` and `AndroidUIScheduler` is a `HybridClass` with a C++ counterpart holding a global reference to the Java object. This structure creates a circular reference between C++ and Java, which the garbage collector cannot clean up. To resolve this issue, I manually removed the reference between C++ and Java by resetting the global ref during the invalidation of the `NativeProxy` and `AndroidUIScheduler`.

## Test plan

I tested it with a new project using Expo (SDK 52) and with Fabric enabled. I reloaded the app several times and performed a heap dump to verify that everything was removed correctly.


packages/react-native-reanimated/android/src/reactNativeVersionPatch/NativeProxyFabric/latest/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -87,11 +87,19 @@ private native HybridData initHybridBridgeless(
...
+
   @Override
   protected HybridData getHybridData() {
     return mHybridData;
   }
 
+  @Override
+  public void invalidate() {
+    super.invalidate();
+    invalidateCpp();
```
tjzel, line 100: Shouldn't we reverse the order here?
```suggestion
    invalidateCpp();
    super.invalidate();
```

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/AndroidUIScheduler.java
```diff
@@ -44,5 +46,6 @@ public void runGuarded() {
 
   public void deactivate() {
     mActive.set(false);
+    invalidateCpp();
   }
```
tjzel, line 50: Can we rename this method to `invalidate` instead?

packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
```diff
@@ -270,6 +275,11 @@ void NativeProxy::registerEventHandler() {
 }
 
 void NativeProxy::maybeFlushUIUpdatesQueue() {
+  // Module might be already destroyed.
+  if (!javaPart_) {
+    return;
+  }
```
tjzel, line 281: I'm not a fan of checking this every time, but it sure is the best way if we don't want any refactors here. However, I'd much prefer something among the lines for transparency:

```suggestion
  if (invalidated) {
    return;
  }
```

What do you think?

piaskowyk (approved):
For me, the approach is okay 👍

packages/react-native-worklets/android/src/main/cpp/worklets/android/AndroidUIScheduler.cpp
```diff
@@ -41,6 +39,9 @@ jni::local_ref<AndroidUIScheduler::jhybriddata> AndroidUIScheduler::initHybrid(
 }
 
 void AndroidUIScheduler::triggerUI() {
+  if(!uiScheduler_){
```
tomekzaw, line 42:
```suggestion
  if (!uiScheduler_) {
```

tomekzaw (changes requested):

tomekzaw (approved):

PR #6705, opened 2024-11-15
feat: Add name to the Reanimated Plugin

tjzel:
## Summary

Babel plugin can have names. Plugin will now be named `reanimated` instead of `node_modules/react-native-reanimated/plugin/index.js`.

Cleaning some files as a bonus.

## Test plan

🚀


packages/react-native-reanimated/plugin/src/plugin.ts
```diff
@@ -78,8 +76,6 @@ module.exports = function (): PluginItem {
       Program: {
         enter(path: NodePath<Program>, state: ReanimatedPluginPass) {
           runWithTaggedExceptions(() => {
-            // Reset worklet number.
-            state.workletNumber = 1;
```
tjzel, line 82: It's extraneous.

packages/react-native-reanimated/plugin/src/globals.ts
```diff
@@ -161,3 +169,23 @@ export let globals: Set<string>;
...
+ * ];
+ * ```
+ */
+export function addCustomGlobals(state: ReanimatedPluginPass) {
+  if (state.opts && Array.isArray(state.opts.globals)) {
+    state.opts.globals.forEach((name: string) => {
+      globals.add(name);
+    });
+  }
+}
```
tjzel, line 191: Brought here from `utils.ts`

bartlomiejbloniarz (approved):

Issue #6719, opened 2024-11-18
Renimated calls findNodeHandle with class instances, which is a slow path in React


gaearon:
### Description

`findNodeHandle` is a notoriously slow path in React — in fact, the function it's calling internally is literally called [`findCurrentFiberUsingSlowPath`](https://github.com/facebook/react/blob/92c0f5f85fed42024b17bf6595291f9f5d6e8734/packages/react-reconciler/src/ReactFiberTreeReflection.js#L123-L266). This is quite inefficient and should be avoided whenever possible.

As @kmagiera [suggested](https://bsky.app/profile/kzzzf.bsky.social/post/3lb5ihib43s2l), filing an issue about it.

While doing it with native refs is fine (it'll exit early before going into the slow path), calling it on a _class instance_ triggers the bad codepath. I see Reanimated doing this in a few places:

<img width="1183" alt="Screenshot 2024-11-18 at 01 33 31" src="https://github.com/user-attachments/assets/4cd019dc-9055-448e-a791-134260f333c0">

<img width="1074" alt="Screenshot 2024-11-18 at 01 28 51" src="https://github.com/user-attachments/assets/c9a57738-2909-4cf6-8256-fce772064efd">

<img width="1234" alt="Screenshot 2024-11-18 at 01 29 27" src="https://github.com/user-attachments/assets/9247730d-bd1f-4fab-874b-aba5096ba0f7">

It would be nice to audit all `findNodeHandle` callsites and remove any that pass class component instances.

### Steps to reproduce

See above

### Snack or a link to a repository

https://github.com/software-mansion/react-native-reanimated

### Reanimated version

3.16.1

### React Native version

0.74.1

### Platforms

Android, iOS, macOS, Web

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Acknowledgements

Yes

gaearon:
```js
<Animated.View />
```

gaearon:
Partial fix: https://github.com/software-mansion/react-native-reanimated/pull/6720

PR #6720, opened 2024-11-18
Replace findNodeHandle with reading _componentViewTag

gaearon:
## Summary

Partially addresses https://github.com/software-mansion/react-native-reanimated/issues/6719.

This is better because it doesn't enter the traversal codepath inside React. The `findNodeHandle` calculation inside the `Animated.View` itself is not so bad because it's short circuited inside for native refs (at least with `View`). Whereas passing a class instance like here always triggers the slower path in React. Regardless, it's already computed so why compute again?

## Test plan

I've added console logs in our app to verify whether `animatedComponent._componentViewTag === findNodeHandle(animatedComponent)` and it always turned out true in the cases I hit.

Not sure if there any cases where those could be different.


tjzel:
I recall there was an issue that if a component doesn't have its own native representation (it wraps a View for example) then we need to re-obtain the tag in some cases (when the underlying View changes). Not sure if it applies here however.

piaskowyk (approved):
I really appreciate that you reported the issue and submitted a PR! Thanks for your time 🙏. I wanted to address other areas where we use the findNode handler, so I've created another PR that builds upon yours - https://github.com/software-mansion/react-native-reanimated/pull/6736

PR #6730, opened 2024-11-19
Update ReanimatedPackage to supported stable API

jakex7:
## Summary

Replace the deprecated `TurboReactPackage` with the newer `BaseReactPackage` to improve compatibility with future React Native releases. Note that this change requires at least react-native@0.74+, but this should not be an issue as, according to the compatibility table, Reanimated@3.17.0 only supports versions 0.74 and above.

Similar change has been introduced in the SVG package https://github.com/software-mansion/react-native-svg/pull/2541

## Test plan

CI should pass


jakex7:
> @jakex7 Please merge main 🙏

@tjzel Done

tjzel (approved):

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedPackage.java
```diff
@@ -36,7 +36,7 @@
       ReanimatedModule.class,
       ReanimatedUIManager.class,
     })
-public class ReanimatedPackage extends TurboReactPackage implements ReactPackage {
+public class ReanimatedPackage extends BaseReactPackage implements ReactPackage {
```
javache, line 39: `BaseReactPackage` already implements `ReactPackage` so the implements is unnecessary.

PR #6732, opened 2024-11-19
feature: Detect edge-to-edge and set isStatusBarTranslucentAndroid / isNavigationBarTranslucentAndroid

zoontek:
## Summary

Similar to [the PR](https://github.com/software-mansion/react-native-screens/pull/2464) I opened on the `react-native-screens` repository (I highly recommend to read the discussion there to understand the motivation behind this), this PR detects if the user enabled edge-to-edge and act accordingly: `useAnimatedKeyboard` are ignored, set to `true` automatically. If those are set, a warning is logged:

> `isStatusBarTranslucentAndroid` and `isNavigationBarTranslucentAndroid` values are ignored when `using react-native-edge-to-edge`

It at some point [this proposal](https://github.com/facebook/react-native/pull/47554) lands in core, `react-native-is-edge-to-edge` will be updated to support both the library and the core edge-to-edge flag, making the transition seamless for the users.

## Test plan

- Install [react-native-edge-to-edge](https://github.com/zoontek/react-native-edge-to-edge) in the example app.
- Don't set `isStatusBarTranslucentAndroid` / `isNavigationBarTranslucentAndroid`, or set them to something else than `true`


tomekzaw:
Hi @zoontek, thanks for this PR. We briefly discussed this internally and @bartlomiejbloniarz will take a look on it.

zoontek:
@tomekzaw @bartlomiejbloniarz Any news? It continues to create [hard to detect issues](https://github.com/zoontek/react-native-edge-to-edge/issues/57).

bartlomiejbloniarz:
Hi @zoontek. I changed the version of `react-native-is-edge-to-edge` to a concrete version (without `^`) just to be extra safe, as we try to keep our dependencies to minimum.

zoontek:
@bartlomiejbloniarz Sure, no worry 🙂

zoontek:
@bartlomiejbloniarz I rebased against main to fix the conflict.

tomekzaw (approved):

piaskowyk (approved):

packages/react-native-reanimated/package.json
```diff
@@ -89,7 +89,8 @@
     "@babel/plugin-transform-unicode-regex": "^7.0.0-0",
     "@babel/preset-typescript": "^7.16.7",
     "convert-source-map": "^2.0.0",
-    "invariant": "^2.2.4"
+    "invariant": "^2.2.4",
+    "react-native-is-edge-to-edge": "1.1.6"
```
tjzel, line 93: I'm not sure about it. If `react-native-is-edge-to-edge` is supposed to be embraced by the community, we should have it as a peer dependency instead to avoid conflicts for the user, no?

zoontek: @tjzel Yes, that means an extra step for everyone, not a good idea. BTW `react-native-is-edge-to-edge` (the detection library, not `react-native-edge-to-edge`) is less then 100 bytes in production builds.

tjzel:
Don't worry @zoontek. I looked into the packages code and I understood the context more and I think there won't be problems even if there are multiple versions loaded in.

tjzel:
Currently we are freezing the main branch for maintenance reasons so it will land there a bit later.

zoontek:
I rebased against `main` to resolve conflicts.

PR #6736, opened 2024-11-20
Refactor findNodeHandler

piaskowyk:
## Summary

This PR addresses the issue reported at [https://github.com/software-mansion/react-native-reanimated/issues/6719](https://github.com/software-mansion/react-native-reanimated/issues/6719) and aims to:

- **Unify the method used to obtain the view tag**, as there are currently several approaches.
- **Avoid passing a class component to the findNodeHandler**. Instead, we'll pass a ref to the component, similar to what Expo implemented here: [https://github.com/expo/expo/pull/33016](https://github.com/expo/expo/pull/33016).
- **Limit unnecessary invocations of findNodeHandler** to no more than one call per render.
- **Remove the invocation of findHostInstance from Paper renderer** on the New Architecture.

**Additional Remarks:**
- When a class component is passed to `createAnimatedComponent`, it will still fall back to the slow path where we can get a native ref.
- In `NativeEventManager`, we need to call findNodeHandler again after every render to ensure that the children of `<GestureDetector />` haven't changed their native tags.
- `LayoutAnimationConfig` still uses findNodeHandler. It requires a complete refactor of their functionality to eliminate its use, and I plan to handle this in another PR.
- `findHostInstance_DEPRECATED` always follows the slow path even for native refs, which is why I've implemented our own version of `findHostInstance` to optimize the happy path.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/6719

Related PRs:
- https://github.com/software-mansion/react-native-reanimated/pull/6030
- https://github.com/software-mansion/react-native-reanimated/pull/5960
- https://github.com/software-mansion/react-native-reanimated/pull/4445
- https://github.com/expo/expo/pull/33016

## Test plan

- [x] check Paper
- [x] check Fabric
- [x] check Web


packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -249,31 +246,33 @@ export function createAnimatedComponent(
...
     }
 
-    _getComponentViewTag() {
+    getComponentViewTag() {
       return this._getViewInfo().viewTag as number;
     }
 
     _detachStyles() {
-      if (this._componentViewTag !== -1 && this._styles !== null) {
+      const { viewTag } = this._getViewInfo() as { viewTag: number };
```
tjzel, line 254:
```suggestion
      const viewTag = this.getComponentViewTag();
```

packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -314,7 +315,7 @@ export function createAnimatedComponent(
         viewName = viewInfo.viewName;
         viewConfig = viewInfo.viewConfig;
         shadowNodeWrapper = isFabric()
-          ? getShadowNodeWrapperFromRef(this)
+          ? getShadowNodeWrapperFromRef(this, hostInstance)
```
tjzel, line 318: This is very smart!

packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -491,21 +490,19 @@ export function createAnimatedComponent(
...
-          this._componentViewTag = tag;
+        if (ref !== this._componentRef) {
+          this._componentRef = ref;
+          // if ref is changed, reset viewInfo
+          this._viewInfo = undefined;
         }
+        if (!ref) {
+          // component is unmounted
+          return;
+        }
```
tjzel, line 501:
```suggestion
        if (ref == null) {
          // component is unmounted
          return;
        }
        if (ref !== this._componentRef) {
          this._componentRef = ref;
          // if ref is changed, reset viewInfo
          this._viewInfo = undefined;
        }
```

tjzel (approved):
Looks good to me overall. Make sure to test it with the newest Gesture Handler and SVG on their more complex examples, since they used to have cases where these refs mechanisms broke.

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -77,23 +77,38 @@ export class NativeEventsManager implements INativeEventsManager {
     });
   }
 
-  private getEventViewTag() {
+  private getEventViewTag(isComponentUpdated: boolean = false) {
```
tjzel, line 80: You've done an excellent job refactoring this function. The only thing that worries me is the order of execution, which seems to be black-boxy. Why do we check for `getScrollableNode` before checking for `setNativeProps`? Why are fast paths executed last?

Let's try do to our best to well document this flow as this is such a crucial part of our code. If you need assistance, I'm more than happy to help you investigating this.

packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -484,21 +480,31 @@ export function createAnimatedComponent(
...
-        if (ref !== this._componentRef) {
-          this._componentRef = ref;
-          // if ref is changed, reset viewInfo
-          this._viewInfo = undefined;
-        }
         if (!ref) {
           // component is unmounted
           return;
         }
+        if (ref !== this._componentRef) {
```
tjzel, line 503: Are you sure we can reorder this?

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
```diff
@@ -1,4 +1,44 @@
+/* eslint-disable camelcase */
```
tjzel, line 1: 🚬🐫

packages/react-native-reanimated/src/createAnimatedComponent/commonTypes.ts
```diff
@@ -117,6 +112,7 @@ export interface IAnimatedComponentInternal {
   _NativeEventsManager?: INativeEventsManager;
   _viewInfo?: ViewInfo;
   context: React.ContextType<typeof SkipEnteringContext>;
+  getComponentViewTag: () => number;
```
tjzel, line 115: Let's TSDoc this method with the previous comment:

```
  /**
   * Used for Shared Element Transitions, Layout Animations and Animated Styles.
   * It is not related to event handling.
   */
```

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
```diff
@@ -1,4 +1,44 @@
...
 // @ts-nocheck
 'use strict';
-export { default as RNRenderer } from 'react-native/Libraries/Renderer/shims/ReactNative';
+
+import { isFabric } from '../PlatformChecker';
+
+function findHostInstanceFastPath(ref: React.Component) {
+  if (ref.__internalInstanceHandle && ref.__nativeTag && ref._viewConfig) {
+    return ref;
+  }
```
tjzel, line 11: Let's add a comment (or better, link - if available) which explains this condition.

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
```diff
@@ -1,4 +1,44 @@
...
+
+let findHostInstance_DEPRECATED: (ref: unknown) => void;
+export function findHostInstance(component: React.Component): unknown {
+  // Fast path for native refs
+  const hostInstance = findHostInstanceFastPath(component._componentRef);
+  if (hostInstance !== undefined) {
+    return hostInstance;
+  }
+
+  resolveFindHostInstance_DEPRECATED();
```
tjzel, line 40: Can we call it in top-level instead or do some dynamic shenanigans instead? Or is it necessary to try resolve it on each `findHostInstance` call?

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -77,23 +77,38 @@ export class NativeEventsManager implements INativeEventsManager {
...
+      // have their own setNativeProps method passed as an option.
+      return findNodeHandle(this.#managedComponent) ?? -1;
+    }
+    if (!isComponentUpdated) {
+      // On the first render of a component, we may already receive a resolved view tag.
+      return this.#managedComponent.getComponentViewTag();
+    }
+    if (componentAnimatedRef.__nativeTag) {
+      // Fast path for native refs
+      return componentAnimatedRef?.__nativeTag;
```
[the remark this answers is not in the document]
piaskowyk: For handling events, things get a bit more complicated:
- If the component is a type of list or ScrollView, we want to avoid the potential issue of getting the native tag of a wrapper instead of the ScrollView itself.
- When someone specifies custom `componentOptions`, we should use them because it indicates that they intentionally want to do something different from the norm.
- If it's the first render, whichever method we use, the outcome will be the same.

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
piaskowyk: I explained it in details here - https://github.com/software-mansion/react-native-reanimated/pull/6736#discussion_r1853824290

packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
piaskowyk: I think it is safe 🫡

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
piaskowyk: The `resolveFindHostInstance_DEPRECATED` method uses the `isFabric` method, which has caused some race conditions in the past when used at the top level. Sometimes, the JS code was executed before the native part had a chance to inject information about the architecture type.

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -77,23 +77,52 @@ export class NativeEventsManager implements INativeEventsManager {
...
       const scrollableNode = componentAnimatedRef.getScrollableNode();
-      newTag = findNodeHandle(scrollableNode) ?? -1;
-    } else {
-      newTag =
-        findNodeHandle(
-          this.#componentOptions?.setNativeProps
-            ? this.#managedComponent
-            : componentAnimatedRef
-        ) ?? -1;
+      if (typeof scrollableNode === 'number') {
```
tomekzaw, line 95: Does this also work as intended on Fabric where there are shadow node wrappers instead of view tags (numbers)?

packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -478,34 +473,42 @@ export function createAnimatedComponent(
...
-        // TODO update config
-
-        const tag = findNodeHandle(ref as Component);
-
-        // callback refs are executed twice - when the component mounts with ref,
-        // and with null when it unmounts
-        if (tag !== null) {
-          this._componentViewTag = tag;
+        if (!ref) {
+          // component is unmounted
```
tomekzaw, line 500:
```suggestion
          // component has been unmounted
```

packages/react-native-reanimated/src/fabricUtils.ts
```diff
@@ -2,25 +2,16 @@
...
 
-let findHostInstance_DEPRECATED: (ref: unknown) => void;
 let getInternalInstanceHandleFromPublicInstance: (ref: unknown) => {
   stateNode: { node: unknown };
 };
 
 export function getShadowNodeWrapperFromRef(
-  ref: React.Component
+  ref: React.Component,
+  hostInstance?: unknown
```
tomekzaw, line 13: Is there a way to provide a "better" type for `hostInstance`?

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
tomekzaw, line 1: Not sure if we should name this file `RNRenderer.ts`, maybe there's a better name?

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
```diff
@@ -1,4 +1,49 @@
...
+function findHostInstanceFastPath(ref: React.Component) {
+  if (ref.__internalInstanceHandle && ref.__nativeTag && ref._viewConfig) {
+    // This is a native ref to a Fabric component
+    return ref;
+  }
+  if (ref._nativeTag && ref.viewConfig) {
+    // This is a native ref to a Paper component
+    return ref;
+  }
+  return undefined;
```
tomekzaw, line 17: Shouldn't we throw an error in such case for easier debugging?

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
```diff
@@ -1,4 +1,49 @@
...
+  if (findHostInstance_DEPRECATED !== undefined) {
+    return;
+  }
+  if (isFabric()) {
+    try {
+      findHostInstance_DEPRECATED =
+        require('react-native/Libraries/Renderer/shims/ReactFabric').findHostInstance_DEPRECATED;
+    } catch (e) {
+      findHostInstance_DEPRECATED = (_ref: unknown) => null;
+    }
```
tomekzaw, line 30: When will that happen? Shouldn't we throw an error indeed?

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
```diff
@@ -1,4 +1,49 @@
...
+export function findHostInstance(component: React.Component): unknown {
+  // Fast path for native refs
+  const hostInstance = findHostInstanceFastPath(component._componentRef);
+  if (hostInstance !== undefined) {
+    return hostInstance;
+  }
+
+  resolveFindHostInstance_DEPRECATED();
+  return findHostInstance_DEPRECATED(
+    isFabric() ? component : component._componentRef
```
tomekzaw, line 47: Looks like we can skip this `isFabric()` call here and just call a function that uses `findHostInstance_DEPRECATED` under the hood.

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
```diff
@@ -77,23 +77,52 @@ export class NativeEventsManager implements INativeEventsManager {
...
 
-  private getEventViewTag() {
+  private getEventViewTag(componentUpdate: boolean = false) {
     // Get the tag for registering events - since the event emitting view can be nested inside the main component
     const componentAnimatedRef = this.#managedComponent
-      ._component as AnimatedComponentRef;
-    let newTag: number;
+      ._componentRef as AnimatedComponentRef & {
+      __nativeTag?: number;
+      _nativeTag?: number;
```
tjzel, line 85:
```suggestion
			/** Fabric **/
      __nativeTag?: number;
      /** Paper **/
      _nativeTag?: number;
```

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
tjzel: When `findHostInstanceFastPath` returns `undefined` we call `findHostInstance_DEPRECATED()`

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
tjzel: This is the previous behavior. I wonder if we should change it 🤔

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
tjzel: What do you mean?

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
tjzel: `findHostInstance.ts`?

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
piaskowyk: I've just followed the previous implementation, but it might be more reasonable to throw an error instead

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
piaskowyk: This is not an error. That just means it’s a ref to a non-native component, and it’s necessary to call `findHostInstance_DEPRECATED` on them - I'll add a proper comment to code to explain it.

packages/react-native-reanimated/src/createAnimatedComponent/NativeEventsManager.ts
piaskowyk: On both platforms, `getScrollableNode` returns a number - I've checked it for both ScrollView and FlatList.

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
piaskowyk: make sense 👍

packages/react-native-reanimated/src/platform-specific/findHostInstance.ts
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable camelcase */
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
tjzel, line 2:
```suggestion
```

packages/react-native-reanimated/src/platform-specific/findHostInstance.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export type HostInstance = {
+  // Fabric fields
+  __internalInstanceHandle?: Record<string, unknown>;
+  __nativeTag?: number;
+  _viewConfig?: Record<string, unknown>;
+  // Paper fields
+  _nativeTag?: number;
+  viewConfig?: Record<string, unknown>;
+};
```
tjzel, line 17: Can we make `hostInstancePaper` and `hostInstanceFabric` types instead?

packages/react-native-reanimated/src/platform-specific/RNRenderer.ts
tjzel: Let's keep the `resolve` call on each invocation then.

tjzel (approved):
Let's go!

PR #6737, opened 2024-11-20
docs: worklets on the Web are just JavaScript functions

kacperkapusciak:
This PR explains the worklet behaviour on the Web and separates the [Worklets guide](https://docs.swmansion.com/react-native-reanimated/docs/guides/worklets) into easier to reference and scan sections.

### DIFF
--- a/packages/docs-reanimated/docs/guides/worklets.mdx
+++ b/packages/docs-reanimated/docs/guides/worklets.mdx
@@ -158,3 +158,5 @@ function App() {
 Worklets can run in other runtimes than the one provided by Reanimated. For example [VisionCamera](https://github.com/mrousavy/react-native-vision-camera) and [LiveMarkdown](https://github.com/Expensify/react-native-live-markdown) create their own worklet runtimes.
 
 You can create your own worklet runtimes with [`createWorkletRuntime`](/docs/threading/createWorkletRuntime) function.
+
+There's no UI thread available on the Web platform. Because of that, when Reanimated runs in the browser, worklets are resolved to plain JavaScript functions and the `'worklet';` directive has no effect.
--- a/packages/docs-reanimated/docs/guides/worklets.mdx
+++ b/packages/docs-reanimated/docs/guides/worklets.mdx
@@ -8,6 +8,8 @@ sidebar_position: 1
 
 Worklets are short-running JavaScript functions that can run on the [UI thread](/docs/fundamentals/glossary#ui-thread). Reanimated uses worklets to calculate view styles and react to events on the UI thread.
 
+## Defining worklets
+
 You can create your own worklets using the `'worklet';` directive at the top of a function.
 
 ```javascript
@@ -16,6 +18,8 @@ function myWorklet() {
   console.log('Hello from a worklet');
 }
 ```
+
+## Workletization
 
 The [Reanimated Babel Plugin](https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/plugin/README-dev.md#basics) looks for functions marked with the `'worklet'` directive and converts them into serializable objects. We call this process [workletization](/docs/fundamentals/glossary#to-workletize). These objects can then be copied and run over on the UI thread.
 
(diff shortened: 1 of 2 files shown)


piaskowyk (approved):

packages/docs-reanimated/docs/threading/runOnUI.mdx
```diff
@@ -71,6 +71,8 @@ import RunOnUISrc from '!!raw-loader!@site/src/examples/RunOnUI';
 
 - Make sure not to execute `runOnUI` on the UI thread as this will result in an error.
 
+- On the Web, `runOnUI` immediately calls the function given as the argument. In browsers there's no UI thread available.
```
tomekzaw, line 74: Actually, `runOnUI` on web behaves just like `requestAnimationFrame` – it creates a version of the function that, when called, will be scheduled to be run with given arguments on next animation frame.

packages/docs-reanimated/docs/guides/worklets.mdx
```diff
@@ -158,3 +158,5 @@ function App() {
 Worklets can run in other runtimes than the one provided by Reanimated. For example [VisionCamera](https://github.com/mrousavy/react-native-vision-camera) and [LiveMarkdown](https://github.com/Expensify/react-native-live-markdown) create their own worklet runtimes.
 
 You can create your own worklet runtimes with [`createWorkletRuntime`](/docs/threading/createWorkletRuntime) function.
+
+There's no UI thread available on the Web platform. Because of that, when Reanimated runs in the browser, worklets are resolved to plain JavaScript functions and the `'worklet';` directive has no effect.
```
tomekzaw, line 162:
```suggestion
There's no separate UI thread available on the Web platform. Because of that, when Reanimated runs in the browser, worklets are resolved to plain JavaScript functions and the `'worklet';` directive has no effect.
```
That's partially true but `'worklet';` directive has an effect (i.e. the code is transformed); it's just that the effect is not used – instead, the function is always called as a regular JS code.

tomekzaw (approved):
obraz

tomekzaw (changes requested):
obraz

packages/docs-reanimated/docs/guides/worklets.mdx
tomekzaw, line 162:
```suggestion
There's no separate UI thread available on the Web platform. Because of that, when Reanimated runs in the browser, worklets are resolved to plain JavaScript functions and the `'worklet';` directive has no effect.
```
That's partially true but `'worklet';` directive has an effect (i.e. the code is transformed); it's just that the effect is not used – instead, the function is always called as a regular JS code.

packages/docs-reanimated/docs/threading/runOnUI.mdx
tomekzaw, line 74: Actually, `runOnUI` on web behaves just like `requestAnimationFrame` – it creates a version of the function that, when called, will be scheduled to be run with given arguments on next animation frame.

tomekzaw (changes requested):
obraz

packages/docs-reanimated/docs/guides/worklets.mdx
```diff
@@ -158,3 +158,5 @@ function App() {
 Worklets can run in other runtimes than the one provided by Reanimated. For example [VisionCamera](https://github.com/mrousavy/react-native-vision-camera) and [LiveMarkdown](https://github.com/Expensify/react-native-live-markdown) create their own worklet runtimes.
 
 You can create your own worklet runtimes with [`createWorkletRuntime`](/docs/threading/createWorkletRuntime) function.
+
+There's no separate UI thread available on the Web platform. Because of that, when Reanimated runs in the browser, worklets are resolved to plain JavaScript functions and the `'worklet';` directive has no effect.
```
tjzel, line 162: This isn't exactly true. We still depend on the babel plugin for closure capturing to deduce dependencies for styles etc. When doing some advanced stuff on the web not using `worklet` could actually hurt your code. I know it shouldn't be like that, but what can we do 😿

kacperkapusciak: thanks! fixed in [cf8d59b](https://github.com/software-mansion/react-native-reanimated/pull/6737/commits/cf8d59b1919c9c826dfa1849d31b1e56d2a274d4)

packages/docs-reanimated/docs/threading/runOnUI.mdx
kacperkapusciak: thanks! fixed in [3e8f977](https://github.com/software-mansion/react-native-reanimated/pull/6737/commits/3e8f97740c25d66afc1e615af832f1ddab59df2d)

tjzel (approved):
Me likes

tomekzaw (approved):

PR #6739, opened 2024-11-20
fix: only intercept events with waiting handlers

### LINKED ISSUE #6684: onLayout stops firing after rotation on iOS when Animated.View is used
### Description

Since upgrading to React Native 0.76 (via Expo 52) I've been trying to track down an [issue](https://github.com/facebook/react-native/issues/47251) with `KeyboardAvoidingView` on iPadOS. Despite the resolution of a related [bug](https://github.com/facebook/react-native/issues/47262) in React Native, the issue persisted in a project using `react-native-reanimated` and the **New Architecture**.

After rotation, the `KeyboardAvoidingView` eventually stops updating properly. The root cause is that the `onLayout` [handler](https://github.com/facebook/react-native/blob/v0.76.1/packages/react-native/Libraries/Components/Keyboard/KeyboardAvoidingView.js#L118-L134) stops being called. This causes `_relativeKeyboardHeight` to calculate the wrong height due to a stale [frame height](https://github.com/facebook/react-native/blob/v0.76.1/packages/react-native/Libraries/Components/Keyboard/KeyboardAvoidingView.js#L101-L103). As such, the height is [set](https://github.com/facebook/react-native/blob/v0.76.1/packages/react-native/Libraries/Components/Keyboard/KeyboardAvoidingView.js#L171) to 0 and no padding is applied.

The `onLayout` bug is not limited to `KeyboardAvoidingView`. It seems that any React Native project using react-native-reanimated will eventually stop firing `onLayout` after rotation on ~iPadOS~ iOS. This issue is limited to the New Architecture. If I switch to the old architecture, the issue does not reproduce.

The provided repro shows this behavior. When running in a simulator or device, rotating the screen will show the keyboard either pushing up the main content with too much padding (simulator) or hiding the input text box completely (real device). The `console.log` also shows an extra `onLayout` being called when an `Animated.View` is present in the DOM, right before `onLayout` stops firing. Removing the `Animated.View` prevents this issue from occurring, and there is only a single `onLayout` call during rotation.

I thought this issue may have been [Expo](https://github.com/expo/expo/issues/32546) specific. However, after further testing and isolation of the reproducer, I found that it can be reproduced using [React Native](https://github.com/mhoran/keyboard-avoiding-view-repro/compare/react-native) without Expo as well. I also isolated the issue to react-native-reanimated, initially believing it was due to an interaction with [react-native-gesture-handler](https://github.com/software-mansion/react-native-gesture-handler/issues/3192).

### Steps to reproduce

1. Launch the reproducer *on an iPad simulator* with `npm run ios`
2. Focus the text input at the bottom
3. Open the on screen keyboard
4. Rotate screen
5. Note that `onLayout` has been printed twice in the log
6. Rotate the screen again
7. Note that the `KeyboardAvoidingView` has applied incorrect padding and that `onLayout` was not printed in the log

https://github.com/mhoran/keyboard-avoiding-view-repro/

### Reanimated version

3.16.1

### React Native version

0.76.1

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Device model

iPad Air 11-inch (M2)

### Acknowledgements

Yes

### Screenshots

|Prior to rotation|After rotation|
|---|---|
|Simulator Screenshot - iPad Air 11-inch (M2) - 2024-11-10 at 11 50 56|Simulator Screenshot - iPad Air 11-inch (M2) - 2024-11-10 at 11 51 03|

mhoran:
## Summary

Previously, NativeReanimatedModule::handleRawEvent would intercept all events received by the event listener. This resulted in an issue where onLayout would not fire in JS on the New Architecture.

Instead, only intercept events with waiting handlers. This prevents asJSIValue from being called on the Reanimated event loop and allows onLayout to bubble up in JS.

See
https://github.com/facebook/react-native/blob/v0.76.2/packages/react-native/ReactCommon/react/renderer/components/view/BaseViewEventEmitter.cpp#L82-L112, which prevents onLayout from being dispatched more than once.

asJSIValue evaluates the lambda above in
https://github.com/facebook/react-native/blob/v0.76.2/packages/react-native/ReactCommon/react/renderer/core/ValueFactoryEventPayload.cpp#L16.

Fixes #6684

## Test plan

See reproducer in #6684. Without this change, `onLayout` will not fire on device rotation. With this change, `onLayout` will fire as expected.


tomekzaw:
@mhoran Thanks for the PR. Just making sure if I understand the problem and the proposed solution correctly.

Currently, all `onLayout` events are intercepted by Reanimated (while still on the UI thread), and calling `asJSIValue` results in `layoutEventState->wasDispatched` being set to true. Then, once the event is processed on the JS thread, `asJSIValue` returns `null`.

After this change, the flow described above will be true only for events that pass `isAnyHandlerWaitingForEvent` test. For all other `onLayout` events (those which are not awaited by Reanimated), `asJSIValue` will not be called on the UI thread so `asJSIValue` will return the actual event data when called on the JS thread.

Is that correct?

mhoran:
> Currently, all onLayout events are intercepted by Reanimated (while still on the UI thread), and calling asJSIValue results in layoutEventState->wasDispatched being set to true. Then, once the event is processed on the JS thread, asJSIValue returns null.

Correct. While the Reanimated event listener will bail on events dispatched from the JS thread: https://github.com/software-mansion/react-native-reanimated/blob/254af50e07ad151ea940b858392f3396cc07f63f/packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm#L148-L153
rotation fires from the UI thread. This results in the handler being called with the layout event, which Reanimated intercepts.

> After this change, the flow described above will be true only for events that pass isAnyHandlerWaitingForEvent test. For all other onLayout events (those which are not awaited by Reanimated), asJSIValue will not be called on the UI thread so asJSIValue will return the actual event data when called on the JS thread.

Correct. This would ultimately happen when `NativeReanimatedModule::handleEvent` calls in to `EventHandlerRegistry::processEvent`, but unfortunately due to the way `asJSIValue` is implemented, and the debounce logic in the `layout` lambda, just getting the value will intercept the event, which is not desired.

I am not sure what would happen if one were to set up an `onLayout` handler that was intended to be handled by Reanimated. Is this even a valid use case? I think it would then get dispatched due to `isAnyHandlerWaitingForEvent` returning true, but then would not be dispatched on the React Native side. That seems fine, but I'm not sure how to even test this.

I added a test for this as well. Triggering a layout event by modifying a view's styles from the UI thread is equivalent to a device rotation (which fires on the UI thread.) The test will fail prior to this code change but passes after.

tomekzaw:
Thanks for clarification as well as adding the tests.

> I am not sure what would happen if one were to set up an `onLayout` handler that was intended to be handled by Reanimated. Is this even a valid use case?

Sounds like a valid use case, here's a draft PR where I register an event listener for `topLayout` event: https://github.com/software-mansion/react-native-reanimated/pull/5738

mhoran:
> Sounds like a valid use case, here's a draft PR where I register an event listener for topLayout event: https://github.com/software-mansion/react-native-reanimated/pull/5738

I've added a test to cover this use case as well.

seanpcoyle:
Is there a chance that this will be landed in an upcoming version? This fixes an outstanding issue in https://github.com/gorhom/react-native-bottom-sheet/issues/2051.

Much appreciated!

piaskowyk (approved):
Good catch! Thanks for your PR 🙌 I really appreciate the detailed description 🫡

rkstar:
any idea on when this might be shipped in a release? we're facing issues related to this `onLayout` that is blocking us from updating to expo sdk 52.

PR #6742, opened 2024-11-21
Fix bug in useSharedValueMock

sumo-slonik:
## Summary

The API for sharedValue has changed from using .value to get() and set() methods. As a result, the mock that previously returned an object with a single value property no longer works in the tests.

## Test plan

This is a PR for the mocks, so it's not necessary for it.

### DIFF
--- a/packages/react-native-reanimated/src/mock.ts
+++ b/packages/react-native-reanimated/src/mock.ts
@@ -51,7 +51,7 @@ const hook = {
   // useHandler: ADD ME IF NEEDED
   useWorkletCallback: ID,
   useSharedValue: <Value>(init: Value) => {
-    let value = { value: init }
+    let value = { value: init };
     const handler = {
       get(target: any, prop: string) {
         if (prop === 'value') {
@@ -61,8 +61,12 @@ const hook = {
           return () => target.value;
         }
         if (prop === 'set') {
-          return (newValue: Value) => {
-            target.value = newValue;
+          return (newValue: Value | ((currentValue: Value) => Value)) => {
+            if (typeof newValue === 'function') {
+              target.value = (newValue as (currentValue: Value) => Value)(target.value);
+            } else {
+              target.value = newValue;
+            }
           };
         }
         return undefined;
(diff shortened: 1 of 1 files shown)


packages/react-native-reanimated/src/mock.ts
```diff
@@ -50,7 +50,16 @@ const hook = {
...
+  useSharedValue: <Value>(init: Value) => {
+    let value = init;
+
+    return {
+      set: (newValue: Value) => {
+        value = newValue;
+      },
+      get: () => value
+    };
+  },
```
tjzel, line 62: We need to preserve the access by `sv.value`. Try something like this:

```suggestion
  useSharedValue: <Value>(init: Value) => {
    let value = {value: init};
    value.get = () => value.value;
    value.set = (newValue) => { value.value = newValue};
    return value;
  },
```

Or better yet, use a Proxy here.

tjzel (changes requested):

packages/react-native-reanimated/src/mock.ts
sumo-slonik: By proxy do you mean something like this?
`
  const useSharedValue = <Value>(init: Value) => {
    let internalValue = init;

    const handler = {
      get(target: any, prop: string) {
        if (prop === 'value') {
          return internalValue;
        }
        if (prop === 'get') {
          return () => internalValue;
        }
        if (prop === 'set') {
          return (newValue: Value) => {
            internalValue = newValue;
          };
        }
        return undefined;
      },
      set(target: any, prop: string, value: any) {
        if (prop === 'value') {
          internalValue = value;
          return true;
        }
        return false;
      }
    };
    return new Proxy({}, handler);
  },
  `

tjzel: I meant the literal [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) object.

tjzel (approved):

PR #6745, opened 2024-11-21
Update base-class for ReanimatedPackage

javache:
## Summary

TurboReactPackage is deprecated and will be removed in a future release. BaseReactPackage is a swap-in replacement and exists in older React Native versions too.

## Test plan

CI


javache:
Duplicate of https://github.com/software-mansion/react-native-reanimated/pull/6730

PR #6746, opened 2024-11-21
Fix `getAnimatedStyle` error when called with component that doesn't have animated styles

AcostaB:
`isEmpty` will return `false` if it is passed in `undefined`, instead of throwing an error.

## Summary

If `getAnimatedStyle` is passed in a component that doesn't have animated styles, it throws an error saying "Cannot convert undefined or null to object". This happens because `jestUtils.ts` has an `isEmpty` function that calls `Object.keys` with an argument that can potentially be `undefined`. The arg can potentially be `undefined` because it comes from `const jestAnimatedStyleValue = component.props.jestAnimatedStyle?.value;`

This issue was introduced in version 3.16.0.

## Test plan

Can be simply tested on a browser console. Just run:

`(obj => Object.keys(obj).length === 0)(undefined) // throws error` 

`(obj => !obj || Object.keys(obj).length === 0)(undefined) // returns true as expected` 

`(obj => !obj || Object.keys(obj).length === 0)({test: 'something'}) // returns false as expected`

### DIFF
--- a/packages/react-native-reanimated/src/jestUtils.ts
+++ b/packages/react-native-reanimated/src/jestUtils.ts
@@ -28,7 +28,7 @@ const defaultFramerateConfig = {
   fps: 60,
 };
 
-const isEmpty = (obj: object) => Object.keys(obj).length === 0;
+const isEmpty = (obj: object | undefined) => !obj || Object.keys(obj).length === 0;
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):
Thanks for your changes! 🙌 I'll merge them into the main branch and backport them to version 3.17.

PR #6749, opened 2024-11-22
feat: Add support for boxShadow

### LINKED ISSUE #6687: Can't animate boxShadow in RN 0.76+
### Description

useAnimatedStyle does not seem to support boxShadow animations - a new feature in React Native 0.76+. When trying to animate box shadows, the shadows disappear as soon as any animation code runs in the hook.

Please note that my repro is very simple but there are many ways to make this bug happen. For example, you can literally run any reanimated interpolator that affects the shadow and the shadow will immediately disappear, you don't even need to return and use a value for the shadow.

### Steps to reproduce

1. Set up RN 0.76.1+ with react native reanimated
2. Create an animated Pressable component with createAnimatedComponent
3. create animated box shadow styles and do anything, maybe animate on click for example
4. shadow disappears when button is pressed instead of being animated - and never returns

https://github.com/C0rren/ReactNativeReanimatedIssue

### Reanimated version

3.16.1

### React Native version

0.76.1

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

Emulator (iPhone 16, iOS 18)

### Device model

Emulator (iPhone 16, iOS 18)

### Acknowledgements

Yes

patrycjakalinska:
## Summary

This PR introduces support for boxShadow - a new feature from React Native 0.76+ NewArch. At the same time it address #6687 .

* Adds parser for `boxShadow` prop, that transforms string into a `boxShadow` object.
* Adds Jest tests
* Fix colors flickering when run with `withSpring`

Runtime test are added for future fabric support in RuntimeTests.

Fixes #6687 

https://github.com/user-attachments/assets/9051d98f-cfb0-41b0-a2b4-6a8c11b477da

## Test plan

To test, paste this code snippet to `EmptyExample` and run.

<details>
<summary>EmptyExample code</summary>

```js
import React from 'react';
import { Pressable, SafeAreaView, ScrollView, Text, View } from 'react-native';
import Animated, {
  Easing,
  interpolate,
  interpolateColor,
  useAnimatedStyle,
  useSharedValue,
  withTiming,
import processBoxShadow from 'react-native-reanimated/src/processBoxShadow';

const AnimatedPressable = Animated.createAnimatedComponent(Pressable);
const BUTTON_TRANSITION_TIME = 500;

const EmptyExample = () => {
  const pressed = useSharedValue<number>(0);

  const animatedBoxShadow = useAnimatedStyle(() => {
    const blurRadius = interpolate(pressed.value, [0, 1], [10, 0]);
    const color = interpolateColor(
      pressed.value,
      [0, 1],
      ['#57b495', '#31775d'] 
    );

    const boxShadow = `0px 4px ${blurRadius} 0px ${color}`;

    return {
      boxShadow,
    };
  });

  const backgroundColorStyle = useAnimatedStyle(() => {
    const backgroundColor = interpolateColor(
      pressed.value,
      [0, 1],
      ['#b1dfd0', '#57b495']
    );

    return {
      backgroundColor,
    };
  });

  const handlePressIn = () => {
    pressed.value = withTiming(1, {
      duration: BUTTON_TRANSITION_TIME,
      easing: Easing.out(Easing.ease),
    });
  };

  const handlePressOut = () => {
    pressed.value = withTiming(0, {
      duration: BUTTON_TRANSITION_TIME,
      easing: Easing.out(Easing.ease),
    });
  };

  return (
    <SafeAreaView style={backgroundStyle}>
      <ScrollView style={backgroundStyle}>
        <View
          style={{
            padding: 24,
          }}>
          <Text
            style={{
              fontSize: 18,
              color: '#001a72',
              fontWeight: 'bold',
              textAlign: 'center',
              marginBottom: 16,
            }}>
            The background color animation will work and so will the shadow ✨
          </Text>
          <Text
            style={{
              fontSize: 14,
              color: '#001a72',
              marginBottom: 8,
            }}>
            BoxShadow is officially supported by Reanimated 💅
          </Text>
          <AnimatedPressable
            style={[
              backgroundColorStyle,
              animatedBoxShadow,
              {
                padding: 16,
                borderRadius: 100,
                marginVertical: 16,
              },
            ]}
            onPressIn={handlePressIn}
            onPressOut={handlePressOut}>
            <Text style={{ color: '#001a72' }}>
              Hello I am button with shadow
            </Text>
          </AnimatedPressable>
        </View>
      </ScrollView>
    </SafeAreaView>
  );
};

const backgroundStyle = {
  backgroundColor: '#f8f9ff',
  flex: 1,
};

export default EmptyExample;

```

</details>


tjzel (commented):
Let's go a bit further. Let's try how `boxShadow` fares in both Jest and our runtime tests.

packages/react-native-reanimated/src/Colors.ts
```diff
@@ -164,6 +164,13 @@ function parsePercentage(str: string): number {
   return int / 100;
 }
 
+function clampRGBA(RGBA: ParsedColorArray): void {
+  'worklet';
+  for (let i = 0; i < 4; i++) {
+    RGBA[i] = Math.max(0, Math.min(RGBA[i], 1));
+  }
+}
+
```
tjzel, line 173: Why do we need this?

packages/react-native-reanimated/src/Colors.ts
```diff
@@ -675,6 +686,24 @@ export function processColorsInProps(props: StyleProps) {
   for (const key in props) {
     if (ColorProperties.includes(key)) {
       props[key] = processColor(props[key]);
+    } else if (
+      NestedColorProperties[key as keyof typeof NestedColorProperties]
+    ) {
+      const nestedPropGroup = props[key] as StyleProps;
+      // most of the time there is only one nested prop in boxShadow array
```
tjzel, line 693: What's the purpose of this comment?

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
```diff
@@ -150,7 +151,11 @@ function runAnimations(
         animation.callback && animation.callback(true /* finished */);
       }
     }
-    result[key] = animation.current;
+    if (typeof animation.current === 'object') {
+      result[key] = { ...animation.current };
+    } else {
+      result[key] = animation.current;
+    }
```
tjzel, line 158: Why do we need this?

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
```diff
@@ -226,7 +234,15 @@ function styleUpdater(
...
-          last[propName] = updates[propName];
+          if (Array.isArray(updates[propName])) {
+            updates[propName].forEach((obj: StyleProps) => {
+              for (const prop in obj) {
+                last[propName][prop] = obj[prop];
+              }
+            });
+          } else {
+            last[propName] = updates[propName];
+          }
```
tjzel, line 245: Why do we need this?

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
patrycjakalinska: because if we don't do that then not a copy is passed to result, rather a reference. It cause the color to not behave properly as it is being change in more than one place. Spreading the object fixes it

tjzel: Can you give a sample case? Why doesn't it affect other objects, only `boxShadow`?

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
patrycjakalinska: It fixes boxShadow color not properly animating. It was caused by the boxShadow being an object. The `toValue` and `current` were becoming the same value causing the shadow to just switch color

packages/react-native-reanimated/src/Colors.ts
patrycjakalinska: when we used color based prop (i.ex. backgroundColor or boxShadow) with `withSpring` animation the colors would flicker, as the RGBA values were going below `0` and returning `NaN`. This function is a guard function that makes sure RGBA values stays within the given limit

packages/react-native-reanimated/src/Colors.ts
patrycjakalinska: as if someone were scared of nested loop 😳 remove it?

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
patrycjakalinska: I'm not sure if we have other example than boxShadow itself. For example `backgroundColor` does simple value copying -> `result[key] = 'rgba(255,0,0,1)` but if we look at `boxShadow` to does it like that:
```
result[key] = { offsetX: 10, offsetY: 5, blurRadius: 3, color: 'rgba(233,0,0,1)
```
and by that it saves reference to object in `result[key]`

packages/react-native-reanimated/src/Colors.ts
tjzel: Yeah, I think it's a bit misleading in this form.

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
tjzel: I know we talked about this offline, but I don't remember the conclusion. `transform` cannot be animated like object as well? Does this affect how `transform` property animates?

patrycjakalinska: `transform` is fine, 'cause it only affects how `color` property is handled.

Normally we have `newValues` which in this case should hold not parsed `color` value. But if it was passed to `result[key]` (aka `updates[key]`) as an object, the `updates` held it as a reference resulting in both updates and newValues `color` prop being parsed. If both of them are parsed the animation cannot run, the `newValues` must stay in `rgba` format, to be assigned to `state.last`, to run the animation.

tl;dr it's all about `color` prop, every other prop is ok

packages/react-native-reanimated/src/Colors.ts
tjzel: Let's make this bit a separate PR then to limit the scope of this one.

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
tjzel: So the underlying issue is that `updates[propName]` doesn't have the updated array, yes? If that's true, then why don't we fix it so `updates` would have it right?

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
tjzel: Then let's annotate that with a comment here.

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
patrycjakalinska: sorry, i didn't explained it well 😞 

the `toValue` and `current`  end up being the same value because when we previously did `last[propName] = updates[propName]` we again referenced the object, not copied the prop to the `state.last`. So if our `last`['boxShadow']` is holding reference to `updates['boxShadow']`, the `toValue` relies on `last` state, then as soon as current changes, the `toValue` follows resulting in a jump rather than a animation. Not spreading but assigning directly prop to prop gave the expected result

(fields to look at is `current` and `toValue`)
Before:
image
https://github.com/user-attachments/assets/9474379d-8a31-4d97-b37d-b59733bf24ee
After:
image
https://github.com/user-attachments/assets/3e9e5986-864f-404f-9e95-058a89122dd8

tjzel: I understand now 👍

packages/react-native-reanimated/src/Colors.ts
```diff
@@ -349,6 +349,10 @@ export const ColorProperties = makeShareable([
   'stroke',
 ]);
 
+const NestedColorProperties = makeShareable({
+  boxShadow: ['color'],
```
piaskowyk, line 353:
```suggestion
  boxShadow: 'color',
```
I think we can just use a simple string instead of an array since we don't know of any more complicated structures. I understand that using an array would be a more flexible solution, but let's keep it simple for now.

packages/react-native-reanimated/src/Colors.ts
```diff
@@ -675,6 +679,23 @@ export function processColorsInProps(props: StyleProps) {
...
+
+        for (const propName in nestedProp) {
+          if (
+            NestedColorProperties[
+              key as keyof typeof NestedColorProperties
+            ].includes(propName)
+          ) {
+            nestedProp[propName] = processColor(nestedProp[propName]);
+          }
+        }
```
piaskowyk, line 697: Instead of iterating over the entire prop object, let's iterate over `NestedColorProperties` and check if those properties exist in the prop object because the prop object can potentially have more fields than those listed in `NestedColorProperties`

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
```diff
@@ -150,7 +151,16 @@ function runAnimations(
...
       }
     }
-    result[key] = animation.current;
+    /*
+     * If `animation.current` is an object, spread its properties into a new object
+     * to avoid modifying the original reference. This ensures when `newValues` has a nested color prop, it stays unparsed
+     * in rgba format, allowing the animation to run correctly.
+     */
+    if (typeof animation.current === 'object') {
+      result[key] = { ...animation.current };
```
piaskowyk, line 160: Do we also need to copy other objects from the style, such as transforms? 🤔

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
```diff
@@ -226,7 +239,15 @@ function styleUpdater(
...
         );
         if (finished) {
-          last[propName] = updates[propName];
+          if (Array.isArray(updates[propName])) {
+            updates[propName].forEach((obj: StyleProps) => {
+              for (const prop in obj) {
+                last[propName][prop] = obj[prop];
+              }
+            });
+          } else {
```
piaskowyk, line 248: 1. When `updates[propName]` is an array?
2. What about spreed operator? 
3. We also need a proper commentary, as this change is not obvious :/
4. Is any change to avoid that copying here? - just wondering 🤔

packages/react-native-reanimated/src/Colors.ts
patrycjakalinska: prop object being the general object with all the props passed (backgroundColor, width, boxShadow etc.)? Because in current approach I use the loop that was there before (loop through keys in props), and if there is a boxShadow in props (boxShadow being an array by default), I iterate through objects in array (most cases one object), and check for if there is a color prop in boxShadow object

Can you elaborate more? Do you want to move checking the NestedProp outside the parent loop?

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
patrycjakalinska: 1. transform and boxShadow
2. spread operator removed problem of reference, but only copying prop to prop assured the independence of `toValue` and `current` (in other words it made sure the animation wouldn't jump)
3. on it!
4. in a meaning to avoid copying prop to prop? I tried some other approaches but only this guaranteed smooth animation

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
patrycjakalinska: In this approach we also copy transform yes, but it doesn't affect it - it only makes sure any nested the color property will work

patrycjakalinska: In this line the original `key` value is `0` rather that `boxShadow`, so we cannot use it. It is because we are doing:
```js
    animation.forEach((entry, index) => {
      if (
        !runAnimations(
          entry,
          timestamp,
          index, // key
          result[key],
          animationsActive
        )
      ) {
```

I would suggest saving initial `key` to `currentKey` prop, and passing it. It supports your idea of avoiding performance degradation as well as it's not invasive.

```js
function runAnimations(
  animation: AnimatedStyle<any>,
  timestamp: Timestamp,
  key: number | string,
  result: AnimatedStyle<any>,
  animationsActive: SharedValue<boolean>,
  originalKey?: string
): boolean {
 // ...
  const currentKey = originalKey || (key as string);
  if (Array.isArray(animation)) {
 // ...
    animation.forEach((entry, index) => {
      if (
        !runAnimations(
          entry,
          timestamp,
          index,
          result[key],
          animationsActive,
          currentKey
        )
      ) {
      // ...
      }
  } else if (typeof animation === 'object' && animation.onFrame) {
   // ....
    /*
     * If `animation.current` is an object, spread its properties into a new object
     * to avoid modifying the original reference. This ensures when `newValues` has a nested color prop, it stays unparsed
     * in rgba format, allowing the animation to run correctly.
     */
    if (currentKey === 'boxShadow') {
      result[key] = { ...animation.current };
    } else {
      result[key] = animation.current;
    }
    return finished;
```

by default, on the first run, `runAnimations` function doesn't have `originalKey` so `currentKey` becomes `key`. But when we go deeper i.ex. run animation for each item of the array (like we do in `boxShadow` case), the `currentKey` stays as `boxShadow` (or any other propName) and lets us determine if we want to spread the `animation.current`

patrycjakalinska: I can push this changes so you can see first hand

packages/react-native-reanimated/src/Colors.ts
```diff
@@ -675,6 +679,22 @@ export function processColorsInProps(props: StyleProps) {
...
+
+        for (const propName in nestedProp) {
+          if (
+            NestedColorProperties[key as keyof typeof NestedColorProperties] ===
+            propName
+          ) {
+            nestedProp[propName] = processColor(nestedProp[propName]);
+          }
+        }
+      }
```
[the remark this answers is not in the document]
patrycjakalinska: yeah, I agree this is a better approach 😅

piaskowyk (approved):

bonnmh:
Hi team,

The previous boxShadow crash on Android is fixed, but it has been replaced by a severe performance issue. Scrolling through a <ScrollView /> containing multiple views with animated boxShadow causes significant JS and UI thread lag on Android. The issue occurs on both platforms, but the lag is significantly more severe on Android. On iOS, the lag is present but less pronounced.

The performance returns to normal immediately after removing the boxShadow property from the style object in useAnimatedStyle.

To Reproduce:
Render a list of components within a <ScrollView />. Each component must contain a view using useAnimatedStyle to apply a boxShadow. Scrolling the list on Android will cause noticeable stuttering and frame drops.

```
  const shadow = pressed.value ? innerShadow : defaultShadow;

  return {
    backgroundColor: '...', 
    boxShadow: shadow,
  };
});
```

Environment:

**react-native**: 0.77.2
**react-native-reanimated**: 3.17.3 | 3.17.4 | 3.17.5 | 3.18.0
**OS**: Android/iOS
**Device Test**: iPhone 14 Pro Max, Samsung Galaxy A16 5G

PR #6755, opened 2024-11-24
useAnimatedKeyboard fixes (iOS)

### LINKED ISSUE #5584: `useAnimatedKeyboard` does not support floating keyboard (iPad)

### LINKED ISSUE #6440: useAnimatedKeyboard gives wrong values when prefersCrossFadeTransitions is turned on.
### Description

`useAnimatedKeyboard` reports wrong values when used with `prefersCrossFadeTransitions` turned on with iOS.
_Prefer Cross-Fade Transitions_ is a sub option, after turning on _Reduce Motion_ on iOS. It seems when this is turned on, the `useAnimatedKeyboard` reports wrong info.

A similar issue was fixed in react natives keyboard avoiding view here: https://github.com/facebook/react-native/pull/34503

### Steps to reproduce

1. Can be reproduced in example app, I tested with `apps/paper-example` in `react-native-reanimated` repository. 
2. Install and get running on an iOS simulator as normally.
3. Go into settings on iOS device and turn on _Reduce Motion_ + _Prefer Cross-Fade Transitions_  
4. Go into reanimated paper example app, and open useAnimatedKeyboard example.
5. Tap input field, to get keyboard up. Then tap outside for it to close.
6. Keyboard focus works, but when hiding keyboard, the animated keyboard reports wrong height and status afterwards, when keyboard is closed.
  a. Have experienced some scenarios of it also giving to big a height on opening keyboard. 

https://github.com/user-attachments/assets/aca6932f-60cb-4fbe-ba5b-e3a0f4538673

#### Logged values
With added logs by usage of animated keyboard:

```
+    console.info("keyboard state", keyboard.height.value, keyboard.state.value)
    return {
    };
  });
```

Then on opening the values seems correct, going to correct height and open state.
But on closing it goes to 0 height, then straight up to a lot higher, with still reporting keyboard as open.
```
 INFO  keyboard state 0 0
 INFO  keyboard state 0 2
 INFO  keyboard state 291 2
 INFO  keyboard state 0 2
 INFO  keyboard state 852 2

```

https://github.com/software-mansion/react-native-reanimated/tree/main/apps/paper-example

### Reanimated version

3.15.0

### React Native version

0.75.1

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

### LINKED ISSUE #6727: useAnimatedKeyboard returns wrong height
### Description

When using `useAnimatedKeyboard` there are times when the `height` returned by the hook is not right, causing the same "jump" effect as `KeyboardAvoidingView`. This is caused by the height changing from 0 to `max` almost instantly (or in few 3-5 steps). Usually it happens when pressing rapidly on components, causing keyboard to open and close frequently. This happens both during opening and during closing. It is happening only on iOS, since on Android I'm not able to reproduce this.

It is happening also in a barebone react-native project, just with Reanimated.

https://github.com/user-attachments/assets/fbbbb6d4-6097-4406-ba1e-c3ddf70d4b3e

### Steps to reproduce

1. Have some inputs and a "sticky" component
2. Try to press multiple times in order to open-close the keyboard rapidly
3. --> The height will be wrong at some point

I configured a barebone app with just Reanimated and I was able to reproduce it. I'll leave the code of that App.tsx here.

<details>
<summary>App.tsx code</summary>

```tsx
import React from 'react';

import {
  Button,
  SafeAreaView,
  ScrollView,
  StyleSheet,
  Text,
  TextInput,
} from 'react-native';
import Animated, { useAnimatedKeyboard, useAnimatedProps, useAnimatedStyle, useSharedValue } from 'react-native-reanimated';


Animated.addWhitelistedNativeProps({text: true});

const AnimatedText = Animated.createAnimatedComponent(TextInput);

function App(): React.JSX.Element {


  const { height } = useAnimatedKeyboard();
  const oldHeight = useSharedValue(0);

    if (oldHeight.value - height.value > 100) {
      console.log('Keyboard moved too quickly!');
    }
    oldHeight.value = height.value;

    return {
      text: `${height.value}`,
      defaultValue: `${height.value}`,
    };
  }, [height]);

    return {
      marginBottom: height.value,
      backgroundColor: 'yellow',
      padding: 16,
    };
  });

  return (
    <SafeAreaView style={style.container}>
      <ScrollView contentInsetAdjustmentBehavior="automatic">
          <AnimatedText style={{ fontSize: 24 }} animatedProps={animatedProps} />

          <Text style={{ marginTop: 16 }}>Name</Text>
          <TextInput />
          <Text style={{ marginTop: 16 }}>Surname</Text>
          <TextInput />
          <Text style={{ marginTop: 16 }}>Secret</Text>
        </ScrollView>
        <Animated.View style={animatedStyle}>
          <Button title="Sticky Me"/>
        </Animated.View>
    </SafeAreaView>
  );
}

const style = StyleSheet.create({
  container: {
    backgroundColor: 'white',
    flex: 1,
  },
}
);

export default App;
```
</details>

https://snack.expo.dev/mtxg_VNC73hJaIz9hoMG0?platform=ios
Code of App.tsx available also in this issue

### Reanimated version

3.16.1

### React Native version

0.76.1

### Platforms

iOS

### JavaScript runtime

None

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & production bundle

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

mhoran:
## Summary

fix: return correct height for detached keyboard on iOS (fixes #5584, #6440)
fix: don't clobber keyboard height when keyboard did not animate on iOS
fix: convert keyboard position to window coordinate space on iOS (Stage Manager compatibility)
fix: don't prematurely update keyboard height during dismissal on iOS
fix: don't jump to height when iOS keyboard is rapidly opened/closed (fixes #6727)
fix: correct height when keyboard opened from shortcuts bar on iOS

## Test plan

Can be tested with the useAnimatedKeyboard example.

1. Enable iPad for the example project.
2. Launch the useAnimatedKeyboard example.
3. Open the keyboard, then undock or float the keyboard. The view should not transform when undocked or floating.
4. Dock the keyboard
5. Hide the keyboard, note that the view should now be transformed by the keyboard height
6. Enable Reduce Motion in Accessibility, and Prefers Cross-Fade Transitions
7. Open the keyboard, note that he transform should work as expected

### DIFF
--- a/packages/react-native-reanimated/apple/reanimated/apple/keyboardObserver/REAKeyboardEventObserver.mm
+++ b/packages/react-native-reanimated/apple/reanimated/apple/keyboardObserver/REAKeyboardEventObserver.mm
@@ -173,22 +173,11 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
 
 - (void)updateKeyboardFrame
 {
-  CGFloat keyboardHeight = 0;
   bool isKeyboardAnimationRunning = [self hasAnyAnimation:_measuringView];
   if (isKeyboardAnimationRunning) {
-    keyboardHeight = [self getAnimatingKeyboardHeight];
-  } else {
-    // measuring view is no longer running an animation, we should settle in OPEN/CLOSE state
-    if (_state == OPENING || _state == CLOSING) {
-      _state = _state == OPENING ? OPEN : CLOSED;
-    }
-    if (_state == OPEN || _state == CLOSED) {
-      keyboardHeight = _targetKeyboardHeight;
-    }
-    // stop display link updates if no animation is running
-    [[self getDisplayLink] setPaused:YES];
+    CGFloat keyboardHeight = [self getAnimatingKeyboardHeight];
+    [self runListeners:keyboardHeight];
   }
-  [self runListeners:keyboardHeight];
 }
 
 - (void)keyboardWillChangeFrame:(NSNotification *)notification
@@ -242,8 +231,6 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
                        self->_measuringView.frame = CGRectMake(0, -1, 0, self->_targetKeyboardHeight);
                      }];
     [self runUpdater];
-  } else {
-    [self runListeners:_targetKeyboardHeight];
   }
 }
 
@@ -271,6 +258,10 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
                              selector:@selector(keyboardWillHide:)
                                  name:UIKeyboardWillHideNotification
                                object:nil];
+      [notificationCenter addObserver:self
+                             selector:@selector(keyboardDidHide:)
+                                 name:UIKeyboardDidHideNotification
+                               object:nil];
     }
 
     [self->_listeners setObject:listener forKey:listenerId];
@@ -288,6 +279,7 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
       [notificationCenter removeObserver:self name:UIKeyboardWillShowNotification object:nil];
       [notificationCenter removeObserver:self name:UIKeyboardDidShowNotification object:nil];
       [notificationCenter removeObserver:self name:UIKeyboardWillHideNotification object:nil];
+      [notificationCenter removeObserver:self name:UIKeyboardDidHideNotification object:nil];
     }
   });
 }
@@ -310,10 +302,22 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
 
 - (void)keyboardDidShow:(NSNotification *)notification
 {
+  [[self getDisplayLink] setPaused:YES];
+
+  auto window = [[[UIApplication sharedApplication] delegate] window];
+  auto keyboardView = [self getKeyboardView];
+  if (keyboardView) {
+    CGRect frameIntersection = CGRectIntersection(window.bounds, keyboardView.frame);
+    _targetKeyboardHeight =
+        CGRectIsEmpty(frameIntersection) ? 0 : CGRectGetMaxY(window.bounds) - CGRectGetMinY(frameIntersection);
+  }
+  _state = OPEN;
+  [self runListeners:_targetKeyboardHeight];
+
   if (_isKeyboardObserverAttached) {
     return;
   }
-  if (auto keyboardView = [self getKeyboardView]) {
+  if (keyboardView) {
     [_keyboardView addObserver:self
                     forKeyPath:@"center"
                        options:NSKeyValueObservingOptionOld | NSKeyValueObservingOptionNew
@@ -331,6 +335,15 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
 
   _state = CLOSING;
   [self keyboardWillChangeFrame:notification];
+}
+
+- (void)keyboardDidHide:(NSNotification *)notification
+{
+  [[self getDisplayLink] setPaused:YES];
+
+  _targetKeyboardHeight = 0;
+  _state = CLOSED;
+  [self runListeners:_targetKeyboardHeight];
 }
 
 - (void)updateKeyboardHeightDuringInteractiveDismiss:(CGPoint)oldKeyboardFrame
--- a/packages/react-native-reanimated/apple/reanimated/apple/keyboardObserver/REAKeyboardEventObserver.mm
+++ b/packages/react-native-reanimated/apple/reanimated/apple/keyboardObserver/REAKeyboardEventObserver.mm
@@ -202,28 +202,38 @@ typedef NS_ENUM(NSUInteger, KeyboardState) {
   CGRect beginFrame = [[userInfo objectForKey:UIKeyboardFrameBeginUserInfoKey] CGRectValue];
   CGRect endFrame = [[userInfo objectForKey:UIKeyboardFrameEndUserInfoKey] CGRectValue];
   NSTimeInterval animationDuration = [[userInfo objectForKey:UIKeyboardAnimationDurationUserInfoKey] doubleValue];
-  CGSize windowSize = [[[UIApplication sharedApplication] delegate] window].frame.size;
+  auto window = [[[UIApplication sharedApplication] delegate] window];
 
-  _initialKeyboardHeight = windowSize.height - beginFrame.origin.y;
-  _targetKeyboardHeight = windowSize.height - endFrame.origin.y;
+  /*
+   In some cases, such as when the user has enabled "prefer cross-fade transitions" or
+   when the keyboard is floating, the begin or end frame is an empty rectangle. In
+   such cases we should treat the keyboard as closed, rather than at (0, 0).
+   */
+  CGRect beginFrameIntersection = CGRectIntersection(window.bounds, beginFrame);
+  _initialKeyboardHeight =
+      CGRectIsEmpty(beginFrameIntersection) ? 0 : CGRectGetMaxY(window.bounds) - CGRectGetMinY(beginFrameIntersection);
+
+  CGRect endFrameIntersection = CGRectIntersection(window.bounds, endFrame);
+  _targetKeyboardHeight =
+      CGRectIsEmpty(endFrameIntersection) ? 0 : CGRectGetMaxY(window.bounds) - CGRectGetMinY(endFrameIntersection);
 
   /*
     This may seem a bit confusing, but usually, the state should be either OPENED or CLOSED.
     However, if it shows as OPENING, it means that the interactive dismissal was canceled.
   */
-  bool isInteractiveMode = _state == OPENING;
-  if (_targetKeyboardHeight > 0 && _state != OPEN) {
-    _state = OPENING;
-  } else if (_targetKeyboardHeight == 0 && _state != CLOSED) {
-    _state = CLOSING;
-  }
+  bool isInteractiveMode = _state == OPENING && _isKeyboardObserverAttached;
   auto keyboardView = [self getKeyboardView];
-  bool hasKeyboardAnimation = [self hasAnyAnimation:keyboardView];
-  if (isInteractiveMode) {
-    // This condition can be met after canceling interactive dismissal.
-    _initialKeyboardHeight = windowSize.height - keyboardView.frame.origin.y;
+  if (_state == CLOSING) {
+    _targetKeyboardHeight = 0;
+  } else if (isInteractiveMode) {
+    // Prefer the keyboard view frame over notification frame which may not be current.
+    beginFrameIntersection = CGRectIntersection(window.bounds, keyboardView.frame);
+    _initialKeyboardHeight = CGRectIsEmpty(beginFrameIntersection)
+        ? 0
+        : CGRectGetMaxY(window.bounds) - CGRectGetMinY(beginFrameIntersection);
   }
 
+  bool hasKeyboardAnimation = [self hasAnyAnimation:keyboardView];
   if (hasKeyboardAnimation || isInteractiveMode) {
     _measuringView.frame = CGRectMake(0, -1, 0, _initialKeyboardHeight);
     [UIView animateWithDuration:animationDuration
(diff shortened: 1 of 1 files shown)


mhoran:
Sure thing! I tried to leave detailed notes in my commit messages for posterity, but happy to provide further context as well.

Regarding `UIKeyboardWillChangeFrameNotification` vs show/hide notifications, initially I tried to take an approach similar to the one in #5710. While I did find a working solution (checking the `endFrame` only, which worked with interactive dismissal), I ran into issues with this approach in split view as well as Stage Manager. Dealing with all the different coordinate systems was a real pain, and I figured the code would be so complex that it would end up regressing in the future.

I found that the most reliable way to determine if the keyboard has been detached is to listen for `UIKeyboardWillHideNotification`. This required reworking things a bit to use `UIKeyboardWillShowNotification` and `UIKeyboardWillHideNotification` instead of `UIKeyboardWillChangeFrameNotification`, due to ordering (`UIKeyboardWillChangeFrameNotification` fires before the show / hide notifications.)

However, this approach also has the benefit of simplifying the logic in the change handler. I was able to remove the logic that tried to guess whether the keyboard was opening or closing -- which I found to be inaccurate at times, especially when a hardware keyboard is attached (due to the autosuggest dock.) The `OPENING` and `CLOSING` states are now set only in the appropriate notification handlers (or overridden in `updateKeyboardHeightDuringInteractiveDismiss` -- more on that later.)

I fixed a few more issues along the way.
1. There was an issue with the hardware keyboard autosuggest bar not being respected, due to the notification handler expecting there to be a measurement frame with accurate height at all times. However, that's not true if the event handler doesn't happen to run at the exact moment a keyboard animation is running. I found this to be unnecessary complexity, since `_targetKeyboardHeight` can be used instead.
2. I fixed an issue with the calculated height when the user has enabled Reduced Motion (Accessibility) and Prefers Cross-Fade Transitions. The implemented fix is the same that is in [React Native](https://github.com/facebook/react-native/blob/v0.76.3/packages/react-native/Libraries/Components/Keyboard/KeyboardAvoidingView.js#L89-L95).
3. The coordinate space of keyboard frame notifications are in the screen's coordinate space. These need to be converted to the window's coordinate space to work correctly with Stage Manager. Without 3c2d160c0200deaeacc6ad15a9a8b454e0978459, the height returned by `useAnimatedKeyboard` will only be correct if an app is full screen when Stage Manager is enabled. Otherwise, the height will not take into consideration the position of the window within the stage.
4. I found that sometimes the wrong height would be returned -- often a quick jump to 0, then back to the correct height -- during animations, due to an issue with `updateKeyboardHeightDuringInteractiveDismiss`. I added a guard to make sure that the keyboard did not immediately go to 0 (this will be handled by `keyboardWillHide`) and corresponds to the user dismissing the keyboard with the close button or blurring the input. I also added a check to determine if the keyboard is floating (keyboard width != window width). These are not perfect solutions, but do make the animation less janky. I think someone with more iOS experience could come up with a more bullet proof solution (e.g. check if there is a touch somewhere in the main view), but I'm not sure how to do that myself.

There is a new API, [Keyboard Layout Guide](https://developer.apple.com/documentation/uikit/keyboards_and_input/adjusting_your_layout_with_keyboard_layout_guide), which will make this all a lot simpler in the future. However, it was introduced in iOS 15. So we will have to live with this complexity until RN 0.76 is the oldest supported version for Reanimated.

Note that there remain some issues with undocked and split keyboard modes, due to incorrect start frame positions sent in notifications. Apple has removed these modes in new devices, so I didn't try to implement workarounds.

mhoran:
Here is a screen recording showing the improvements: https://drive.google.com/file/d/1EFDhYZWNMXhJIwNlj1_ugeVwe3FqZqVF/view?usp=drivesdk.

mhoran:
It seems I may also have fixed #6727 as I was cleaning things up in b6c8364. When a hardware keyboard is connected and the autocorrect bar is on screen, the keyboard is considered "open". But when you show the on screen keyboard, it rapidly hides and then shows the keyboard -- inadvertently triggering the interactive dismissal logic. Cleaning up the interactive dismissal flags fixed this for the hardware keyboard case, and I believe it should fix the scenario described in #6727 as well.

janicduplessis:
I tested this in my app and can confirm it does fix the Prefer Cross Fade Transition issue (https://github.com/software-mansion/react-native-reanimated/issues/6440)

mhoran:
There is an issue with iOS 18.2 when opening the on-screen keyboard from the autosuggest bar when a hardware keyboard is attached. Opening the on-screen keyboard in this way will not properly update the height returned by `useAnimatedKeyboard`. However, the issue is not related to these changes (I can reproduce without them.) The keyboard handlers receive bogus end frame information in the notification, so we do not know where the keyboard actually is.

mhoran:
I managed to fix the iOS 18.2 hardware keyboard issue by updating the keyboard height in response to `UIKeyboardDidShowNotification` and `UIKeyboardDidHideNotification` notifications. The keyboard does jump to the target height instead of animating when opening the software keyboard with a hardware keyboard attached, so hopefully we'll get a proper fix from Apple.

This has cleaned up keyboard state handling as well. Now state changes are handled only by the corresponding notification handlers, and `updateKeyboardFrame` is only responsible for animation tracking. That should be a bit less error prone in the future.

On a side note, I managed to get the `keyboardLayoutGuide` working with React Native, sort of. Unfortunately it relies heavily on constraints, and I think this may be fundamentally incompatible with React Native's layout model. Happy to share my learnings with anyone on the Reanimated team who is interested.

tomekzaw:
@mhoran Thanks again for your work on this PR. We're quite focused in other areas but we'll try to find some availability to review and thoroughly test your changes.

packages/react-native-reanimated/apple/reanimated/apple/keyboardObserver/REAKeyboardEventObserver.mm
```diff
@@ -202,37 +186,46 @@ - (void)keyboardWillChangeFrame:(NSNotification *)notification
   CGRect beginFrame = [[userInfo objectForKey:UIKeyboardFrameBeginUserInfoKey] CGRectValue];
   CGRect endFrame = [[userInfo objectForKey:UIKeyboardFrameEndUserInfoKey] CGRectValue];
   NSTimeInterval animationDuration = [[userInfo objectForKey:UIKeyboardAnimationDurationUserInfoKey] doubleValue];
-  CGSize windowSize = [[[UIApplication sharedApplication] delegate] window].frame.size;
+  auto window = [[[UIApplication sharedApplication] delegate] window];
+  CGSize windowSize = window.frame.size;
+
+  beginFrame = [window convertRect:beginFrame fromCoordinateSpace:window.screen.coordinateSpace];
+  endFrame = [window convertRect:endFrame fromCoordinateSpace:window.screen.coordinateSpace];
```
szydlovsky, line 193: I'd add a small comment explaining that the coordinate space conversion is needed

packages/react-native-reanimated/apple/reanimated/apple/keyboardObserver/REAKeyboardEventObserver.mm
```diff
@@ -202,37 +186,46 @@ - (void)keyboardWillChangeFrame:(NSNotification *)notification
...
+  _initialKeyboardHeight = CGRectGetMaxY(beginFrame) > 0 ? windowSize.height - beginFrame.origin.y : 0;
   _targetKeyboardHeight = windowSize.height - endFrame.origin.y;
 
+  bool prefersCrossFade = false;
+#if __has_include(<UIKit/UIAccessibility.h>)
+  prefersCrossFade = UIAccessibilityPrefersCrossFadeTransitions();
+#endif
+  if (prefersCrossFade && beginFrame.origin.y == 0) {
+    _initialKeyboardHeight = 0;
+  }
```
szydlovsky, line 204: I'd definitely add a comment with explanation here as well. From what I've seen and tinkered with, only the begin frame seems to be off, the `endFrame` (and thus `targetKeyboardHeight`) is fine.

packages/react-native-reanimated/apple/reanimated/apple/keyboardObserver/REAKeyboardEventObserver.mm
```diff
@@ -323,11 +350,16 @@ - (void)updateKeyboardHeightDuringInteractiveDismiss:(CGPoint)oldKeyboardFrame
   }
   float windowHeight = keyboardView.window.bounds.size.height;
   float keyboardHeight = keyboardView.frame.size.height;
+  if (keyboardHeight == 0 || keyboardView.frame.size.width != keyboardView.window.bounds.size.width) {
+    return;
+  }
```
szydlovsky, line 355: I'd love some comment why this is needed too. Some person reading it in a 2 years time might not get that the width check is for floating keyboard.

szydlovsky (commented):
Overall, an incredible PR and I wish more folks from community made such impressive contributions!
I've left some rather stylistic comments to make sure the code is anyhow readable in the future, since most of the stuff we do here is already pretty hacky.
I am yet to test everything I need on a physical iPad, I'll get that done tomorrow most likely.

Nonetheless, great work!

packages/react-native-reanimated/apple/reanimated/apple/keyboardObserver/REAKeyboardEventObserver.mm
mhoran: That is what I have seen as well. And in fact, this code has no effect anymore after I added the guard above to check `CGRectGetMaxY` for other reasons. It may still make sense to have this here to explicitly set the height to `0`, but if you think it should just be removed that is fine as well.

szydlovsky:
Also, I found that the height values emitted when undocking the keyboard are some quite rough jumps.
See the vid below, the values in my case were:
- 340 (open)
- sudden jump to 54 (closing)
- a second later, jump to 0 (closed)

https://github.com/user-attachments/assets/9f27e737-bb79-488d-a5ef-1d7422f2d0a4

I wonder if we could maybe force a animation, just like you did with the canceled interactive dismiss.
It would probably need a way to determine if incoming `keyboardWillHide` notification hides it for a floating keyboard. I think I kinda had it in my own attempt that you mentioned: https://github.com/software-mansion/react-native-reanimated/pull/5710

packages/react-native-reanimated/apple/reanimated/apple/keyboardObserver/REAKeyboardEventObserver.mm
szydlovsky: Oh yeah I forgot - the CGRectGetMaxY would also need some explanation haha
I couldn't quite wrap my head around what it does here

szydlovsky:
For the record, the changes regarding jumping midpoints when transitioning from attached to detached keyboard turned out to be inconsistent. On some devices they fixed stuff, on others didn't do nothing and we believe it just added unnecessary complexity. Thus, the logic from that commit is gonna be rolled back, but the comments stay here

szydlovsky:
Unfortunately, I found a more severe mismatch with interactive dismiss emitted values - they seem to be quite a bit off 😢 
EDIT: I checked and it was already present before the PR. If we manage to do something about it - would be great but it's not essential

https://github.com/user-attachments/assets/437f9abe-4565-41ef-b4c6-fd8a8450bd2d

szydlovsky (approved):
After quite a bit of testing, looks and works great, fixes exactly what it is supposed to and introduces no regressions. Beautiful 🎉

tomekzaw (approved):
@mhoran Thanks again for the PR!

PR #6762, opened 2024-11-26
refactor(Worklets): move JSScheduler

tjzel:
## Summary

This pull requests moves `JSScheduler`, a class responsible for scheduling operations back to the JavaScript thread, to Worklets.

- `JSScheduler` is now available from `WorkletsModuleProxy` in cpp and that's how `ReanimatedModuleProxy` obtains it when it needs it. More sophisticated APIs will be made in the future.
- Both `ReanimatedModule` and `WorkletsModule` are now responsible to forward `jsCallInvoker` to their own Proxy modules. `jsCallInvoker` is no longer available in `JSScheduler`.

## Test plan

CI


packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
```diff
@@ -293,11 +291,12 @@ - (void)sendEventWithName:(NSString *)eventName body:(id)body
   auto jsCallInvoker = self.bridge.jsCallInvoker;
 #endif // defined(RCT_NEW_ARCH_ENABLED) && REACT_NATIVE_MINOR_VERSION >= 75
   auto jsiRuntime = reinterpret_cast<facebook::jsi::Runtime *>(self.bridge.runtime);
+  auto isBridgeless = ![self.bridge isKindOfClass:[RCTCxxBridge class]];
```
tjzel, line 294: I tested it on `0.74, 0.75, 0.76` and it worked well there. Moreover, it's only used to inject `isBridgeless` into the React Runtime. If we want, we can drop it altogether in some follow-up PRs.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -80,7 +79,8 @@ ReanimatedModuleProxy::ReanimatedModuleProxy(
         onRender(timestampMs);
       }),
       animatedSensorModule_(platformDepMethodsHolder),
-      jsLogger_(std::make_shared<JSLogger>(jsScheduler_)),
+      jsLogger_(
+          std::make_shared<JSLogger>(workletsModuleProxy->getJSScheduler())),
```
tomekzaw, line 83: Why can't we just use `jsScheduler_` like before?
```suggestion
      jsLogger_(jsScheduler_),
```

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -351,16 +351,17 @@ jsi::Value ReanimatedModuleProxy::getViewProp(
...
     jsi::Runtime &uiRuntime = uiWorkletRuntime_->getJSIRuntime();
     const auto resultStr =
         obtainPropFromShadowNode(uiRuntime, propNameStr, shadowNode);
 
-    jsScheduler_->scheduleOnJS([=](jsi::Runtime &rnRuntime) {
-      const auto resultValue =
-          jsi::String::createFromUtf8(rnRuntime, resultStr);
-      funPtr->call(rnRuntime, resultValue);
-    });
+    workletsModuleProxy_->getJSScheduler()->scheduleOnJS(
```
tomekzaw, line 359: Why can't we just store `jsScheduler_` and use it here? We use `uiScheduler_` anyways so let's be consistent.
```suggestion
    jsScheduler_->scheduleOnJS(
```

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -388,8 +389,8 @@ jsi::Value ReanimatedModuleProxy::getViewProp(
         const auto resultValue =
             obtainPropFunction_(uiRuntime, viewTagInt, propNameValue);
         const auto resultStr = resultValue.asString(uiRuntime).utf8(uiRuntime);
-
-        jsScheduler_->scheduleOnJS([=](jsi::Runtime &rnRuntime) {
+        const auto jsScheduler = workletsModuleProxy_->getJSScheduler();
+        jsScheduler->scheduleOnJS([=](jsi::Runtime &rnRuntime) {
```
tomekzaw, line 393: Why?

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.h
```diff
@@ -194,7 +194,6 @@ class ReanimatedModuleProxy : public ReanimatedModuleProxySpec {
   const bool isBridgeless_;
   const bool isReducedMotion_;
   const std::shared_ptr<WorkletsModuleProxy> workletsModuleProxy_;
-  const std::shared_ptr<JSScheduler> jsScheduler_;
```
tomekzaw, line 197: Let's restore `jsScheduler_`.

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
tomekzaw: I think we can safely assume that bridgeless mode is always enabled on Fabric. Let's just have `auto isBridgeless = true` for now, what do you think?

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
tjzel: I'm moving `UIScheduler` in https://github.com/software-mansion/react-native-reanimated/pull/6789 as well.

I removed the field in anticipation of an API coming. I don't think that with our idea of how Worklets&Reanimated should cooperate the ownership of schedulers should be shared.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
tjzel: Should be `workletsModuleProxy->getJSScheduler()->...`

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
tjzel: We still support 0.74 though, where Bridgeful could be a case.

piaskowyk (approved):
With the additional explanation that every turbo module has its own reference to the JS call invoker, and that this change is to leverage their own reference instead of cross-passing, this change makes sense to me. However, before we proceed with the merge, let's consider Tomek's comments.

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
tomekzaw: Right

tomekzaw (approved):

PR #6769, opened 2024-11-27
fix:`withDecay` animation not starting if inital value is set to 0

### LINKED ISSUE #6680: withDecay() never starts if initial value is 0
### Description

The `withDecay` animation never starts if the initial value is `0`.

```js
import React from 'react';
import {Pressable, Text} from 'react-native';
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withDecay,

  const translateY = useSharedValue(0); // <------- if you change it to 1, it'll work
    transform: [{translateY: translateY.value}],
  }));
  return (
    <>
      <Pressable
        style={{padding: 100, fontSize: 50, backgroundColor: 'purple'}}
        onPress={() => {
          translateY.value = withDecay({
            velocity: 200,
          });
        }}>
        <Text>open</Text>
      </Pressable>
      <Animated.View
        style={[
          {
            backgroundColor: 'blue',
            width: 100,
            height: 100,
          },
          style,
        ]}
      />
    </>
  );
}
```

https://github.com/user-attachments/assets/c51b50ce-7077-4737-bba2-2ca7d379d55c

### Steps to reproduce

See above, pretty easy repro.

Seems to affect both iOS and Android.

Testing on Paper, not sure about Fabric.

https://github.com/gaearon/reanimated-repro-weird/commits/withdecay-bug (note the branch!)

### Reanimated version

3.16.1

### React Native version

0.76.1

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

patrycjakalinska:
## Summary

This PR introduces a new property `forceRunAnimation` to our `DecayAnimation` object. It allows us to skip conditional check that prevented `withDecay` from running if the initial value was set to `0`. 

Fixes #6680. 

## Test plan

Run code snippet in React Native Paper example 🚀 

<details>
<summary>Code snippet used in issue</summary>

```ts
import React from 'react';
import { Pressable, Text } from 'react-native';
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withDecay,

  const translateY = useSharedValue(0); 

    transform: [{ translateY: translateY.value }],
  }));

  return (
    <>
      <Pressable
        style={{ padding: 100, backgroundColor: 'purple' }}
        onPress={() => {
          translateY.value = withDecay({
            velocity: 200,
          });
        }}>
        <Text style={{ fontSize: 50 }}>open</Text>
      </Pressable>
      <Animated.View
        style={[
          {
            backgroundColor: 'blue',
            width: 100,
            height: 100,
          },
          style,
        ]}
      />
    </>
  );
}

```
</details>

### DIFF
--- a/packages/react-native-reanimated/src/animation/decay/decay.ts
+++ b/packages/react-native-reanimated/src/animation/decay/decay.ts
@@ -113,6 +113,7 @@ export const withDecay = function (
       onFrame: decay,
       onStart,
       callback,
+      forceRunAnimation: true,
       velocity: config.velocity ?? 0,
       initialVelocity: 0,
       current: 0,
--- a/packages/react-native-reanimated/src/valueSetter.ts
+++ b/packages/react-native-reanimated/src/valueSetter.ts
@@ -31,6 +31,7 @@ export function valueSetter<Value>(
     // built in animations that are not higher order(withTiming, withSpring) hold target value in .current
     if (
       mutable._value === animation.current &&
+      !animation.forceRunAnimation &&
       !animation.isHigherOrder &&
       !forceUpdate
     ) {
--- a/packages/react-native-reanimated/src/animation/decay/utils.ts
+++ b/packages/react-native-reanimated/src/animation/decay/utils.ts
@@ -13,16 +13,23 @@ const IS_WEB = isWeb();
 export const VELOCITY_EPS = IS_WEB ? 1 / 20 : 1;
 export const SLOPE_FACTOR = 0.1;
 
+/*
+ * The `forceRunAnimation` prop is necessary to run withDecay animation when
+ * related sharedValue stays as `0`. It allows to skip animation check and start
+ * animation immediately
+ */
+
 export interface DecayAnimation extends Animation<DecayAnimation> {
   lastTimestamp: Timestamp;
   startTimestamp: Timestamp;
   initialVelocity: number;
   velocity: number;
   current: AnimatableValue;
+  forceRunAnimation: boolean;
 }
 
 export interface InnerDecayAnimation
-  extends Omit<DecayAnimation, 'current'>,
+  extends Omit<DecayAnimation, 'current' | 'forceRunAnimation'>,
     AnimationObject {
   current: number;
   springActive?: boolean;


packages/react-native-reanimated/src/valueSetter.ts
```diff
@@ -31,6 +31,7 @@ export function valueSetter<Value>(
     // built in animations that are not higher order(withTiming, withSpring) hold target value in .current
     if (
       mutable._value === animation.current &&
+      !animation.forceRunAnimation &&
```
gaearon, line 34: it seems like this check has a purpose (explained above), does introducing this flag not defeat that purpose? i think this needs to be rethought more holistically with the above comment in mind

gaearon: to be clear i'm not a maintainer, just thinking based on the code

packages/react-native-reanimated/src/animation/decay/decay.ts
```diff
@@ -113,6 +113,7 @@ export const withDecay = function (
       onFrame: decay,
       onStart,
       callback,
+      forceRunAnimation: true,
       velocity: config.velocity ?? 0,
       initialVelocity: 0,
       current: 0,
```
piaskowyk, line 119: Instead of adding that artificial prop, we should calculate the finish position like we do for `withTiming`, for example. You can see how it's done here: https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/src/animation/timing.ts#L154. This solution should be clearer and simpler

piaskowyk (changes requested):

packages/react-native-reanimated/src/animation/decay/decay.ts
patrycjakalinska: As I can see `timing` doesn't rely on previous `current` to calculate recent `current` value, as the `decay` does it. Wouldn't setting `current` in onStart to `toValue` break for example  this `decay` function fragment:

```js
  animation.current = current + (v * config.velocityFactor * deltaTime) / 1000;
```
Just wondering, because even this fragment can be rewritten to support your idea (so not a problem really)

In conclusion, to fully understand, you want to replace the `current` with `toValue` instead of - in this case - `0`, so that it will not go into this conditional?

```js
    if (
      mutable._value === animation.current &&
      !animation.isHigherOrder &&
      !forceUpdate
    ) {
      animation.callback && animation.callback(true);
      return;
    }
```
Should I also mark it as a `isHigherOrder` as it will hold its target value in `.current`?

packages/react-native-reanimated/src/animation/decay/decay.ts
```diff
@@ -115,7 +115,7 @@ export const withDecay = function (
       callback,
       velocity: config.velocity ?? 0,
       initialVelocity: 0,
-      current: 0,
+      current: undefined,
```
piaskowyk, line 118: Just before merging, let's add a comment explaining why that `undefined` is important.

piaskowyk (approved):

packages/react-native-reanimated/src/animation/decay/decay.ts
patrycjakalinska: added in [6a35393](https://github.com/software-mansion/react-native-reanimated/pull/6769/commits/6a353932cb5881dd3d999fe9142b837c94af3394)

PR #6774, opened 2024-11-29
fix: brownfield integration

WoLewicki:
## Summary

In brownfield, `    // we need only one instance because SurfacePresenter is the same during the application lifetime` is false since you can destroy RN and create it again.

## Test plan

Run Reanimated in brownfield and try to create and destroy the whole RN a couple of times.

### DIFF
--- a/packages/react-native-reanimated/apple/reanimated/apple/Fabric/REAInitializerRCTFabricSurface.h
+++ b/packages/react-native-reanimated/apple/reanimated/apple/Fabric/REAInitializerRCTFabricSurface.h
@@ -5,7 +5,7 @@
 
 @interface REAInitializerRCTFabricSurface : RCTFabricSurface
 
-@property REAModule *reaModule;
+@property __weak REAModule *reaModule;
 
 @end
 
--- a/packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
+++ b/packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
@@ -193,9 +193,10 @@ RCT_EXPORT_MODULE(ReanimatedModule);
 
 - (void)setReaSurfacePresenter
 {
-  if (reaSurface == nil) {
-    // we need only one instance because SurfacePresenter is the same during the application lifetime
-    reaSurface = [[REAInitializerRCTFabricSurface alloc] init];
+  if(_surfacePresenter && ![_surfacePresenter surfaceForRootTag:reaSurface.rootTag]) {
+    if (!reaSurface) {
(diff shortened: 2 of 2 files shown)


packages/react-native-reanimated/apple/reanimated/apple/Fabric/REAInitializerRCTFabricSurface.h
```diff
@@ -5,7 +5,7 @@
 
 @interface REAInitializerRCTFabricSurface : RCTFabricSurface
 
-@property REAModule *reaModule;
+@property __weak REAModule *reaModule;
```
WoLewicki, line 8: Since surface is `static`, when destroying the modules etc, `REAModule` would not be deallocated since `REAInitializerRCTFabricSurface` still held reference to it.

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
```diff
@@ -193,9 +193,10 @@ - (void)setBridge:(RCTBridge *)bridge
 
 - (void)setReaSurfacePresenter
 {
-  if (reaSurface == nil) {
-    // we need only one instance because SurfacePresenter is the same during the application lifetime
-    reaSurface = [[REAInitializerRCTFabricSurface alloc] init];
+  if(_surfacePresenter && ![_surfacePresenter surfaceForRootTag:reaSurface.rootTag]) {
```
WoLewicki, line 196: In brownfield scenario, when we create, destroy and then create the module again, `reaSurface is still the same, but `_surfacePresenter` is a new one so we need to add it again.

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
```diff
@@ -193,9 +193,10 @@ - (void)setBridge:(RCTBridge *)bridge
...
+    if (!reaSurface) {
+      reaSurface = [[REAInitializerRCTFabricSurface alloc] init];
```
piaskowyk, line 198: Is it possible to change `reaSurface` from a static to a regular instance? Maybe that approach would be easier and safer.

WoLewicki: I'm not sure why was it made static in the first place, probably because `SurfacePresenter is the same during the application lifetime` which can be not true. I can change it then.

piaskowyk (approved):
Let's merge this as a fix, especially since we've removed support for Bridgeful mode on Fabric 👍

Issue #5685, opened 2024-02-19
Height, margin and padding animations feel laggy and flickery on production (3.7.0)


domenicoprestia:
### Description

When animating Animated components on height/paddings and margins all animations feel laggy and flickery in production builds.

Simulator: 

https://github.com/software-mansion/react-native-reanimated/assets/46924943/5a805a5a-5d93-4038-9b16-c3612e2ba306

Production: 

https://github.com/software-mansion/react-native-reanimated/assets/46924943/8f8fcd34-197a-4192-93df-abc2a3ec54a1

### Steps to reproduce

1. Create an animation on a padding value or a height value with 2 states 
2. Try to animate it on a production build 

### Snack or a link to a repository

i cant link a repo

### Reanimated version

3.7.0

### React Native version

0.72.6

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

Release app & production bundle

### Device

Real device

### Device model

All devices

### Acknowledgements

Yes

pavelbabenko:
Same issue

lcarrettin:
I'm experiencing lags after updating from 3.6.1 to 3.7.0

szydlovsky:
Hey @domenicoprestia @pavelbabenko @lcarrettin 👋 We can't investigate anything without reproducible code. Do you have any example with mentioned issue that I can easily run and observe problems? Also, I suggest checking on newest versions (3.12.0, 3.12.1) as well

codegabk:
@szydlovsky animate any image with width and height in a useAnimatedStyle, you will notice flickering and lag

danilvalov:
@lcarrettin
> I'm experiencing lags after updating from 3.6.1 to 3.7.0

I see the same lags in v3.6.1.
On iOS these lags are much stronger.
I also checked v3.16.1, the same problem is there, too.

I think the problem is that all animation calculations for height changes and paddings use the CPU, not the GPU. But I could be wrong

mhoran:
@szydlovsky I've added a reproducer to the reanimated-padding-animation-issue branch here: https://github.com/mhoran/react-native-scratch/tree/reanimated-padding-animation-issue. This is using Reanimated 3.16.1 (also see on 3.17.1) and react-native 0.76.3.

[...]

PR #6776, opened 2024-12-02
Fix `ReanimatedCommitHook` cloning

### LINKED ISSUE #6659: Modal React Native is break when using useAnimatedStyle in component
### Description

before Im upgrade to react native 0.76.1(new arch), Im using react native 0.75.4 and its work fine with old arch. I have trace error why my modal children is not rendered, because i have useAnimatedStyle in component where my modal placed.
When i disabled useAnimatedStyle, its not break modal

### Steps to reproduce

1. I have card component, inside card is modal and bar component. Inside bar component im using useAnimatedStyle.
2. Open modal inside card component
3. Modal break (children not rendered)

https://snack.expo.dev/@tamacroft_expo/2fd4f5

### Reanimated version

3.16.1

### React Native version

0.76.1

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & production bundle

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

bartlomiejbloniarz:
## Summary

This PR changes the way the `ShadowTree` is updated in the commit hook. Previously we would always clone the updated nodes, leading to RN also cloning the remaining nodes (some of them have never been mounted, as we are in the commit hook). 

When RN clones uncommited nodes, it doesn't clone the state from the source node, but instead uses last mounted state, meaning that some of the updates will be lost.

To fix that, we don't clone unmounted `ShadowNodes` but instead change them in place. To recognize if a node was mounted we use the `getHasBeenPromoted` method. 

This change needs to be well tested as modifying nodes in place has to be done carefully, since it's not really the intended way.

fixes #6659

## Test plan
Check behaviour of the repro from #6659, ChessboardExample, BokehExample

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
@@ -70,7 +70,7 @@ RootShadowNode::Unshared ReanimatedCommitHook::shadowTreeWillCommit(
           propsMap[&family].emplace_back(props);
         });
 
-    rootNode = cloneShadowTreeWithNewProps(*rootNode, propsMap);
+    rootNode = cloneShadowTreeWithNewPropsUnmounted(rootNode, propsMap);
 
     // If the commit comes from React Native then pause commits from
     // Reanimated since the ShadowTree to be committed by Reanimated may not
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
@@ -77,6 +77,10 @@ ShadowNode::Unshared cloneShadowTreeWithNewPropsUnmountedRecursive(
   }
 
   auto shadowNode = std::const_pointer_cast<ShadowNode>(oldShadowNode);
+  auto layoutableShadowNode = std::dynamic_pointer_cast<LayoutableShadowNode>(shadowNode);
+  if (layoutableShadowNode){
+    layoutableShadowNode->dirtyLayout();
+  }
 
   const auto family = &shadowNode->getFamily();
   const auto affectedChildrenIt = childrenMap.find(family);
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
@@ -1,14 +1,16 @@
 #ifdef RCT_NEW_ARCH_ENABLED
 
-#include <reanimated/Fabric/ShadowTreeCloner.h>
 #include <react/renderer/core/DynamicPropsUtilities.h>
+#include <reanimated/Fabric/ShadowTreeCloner.h>
 
 #include <ranges>
 #include <utility>
 
 namespace reanimated {
 
-ChildrenMap calculateChildrenMap(const RootShadowNode &oldRootNode, const PropsMap &propsMap){
+ChildrenMap calculateChildrenMap(
+    const RootShadowNode &oldRootNode,
+    const PropsMap &propsMap) {
   ChildrenMap childrenMap;
 
   for (auto &[family, _] : propsMap) {
@@ -69,12 +71,13 @@ ShadowNode::Unshared cloneShadowTreeWithNewPropsUnmountedRecursive(
     ShadowNode::Shared const &oldShadowNode,
     const ChildrenMap &childrenMap,
     const PropsMap &propsMap) {
-  if (oldShadowNode->getHasBeenPromoted()){
-    return cloneShadowTreeWithNewPropsRecursive(*oldShadowNode, childrenMap, propsMap);
+  if (oldShadowNode->getHasBeenPromoted()) {
+    return cloneShadowTreeWithNewPropsRecursive(
+        *oldShadowNode, childrenMap, propsMap);
   }
-  
+
   auto shadowNode = std::const_pointer_cast<ShadowNode>(oldShadowNode);
-  
+
   const auto family = &shadowNode->getFamily();
   const auto affectedChildrenIt = childrenMap.find(family);
   const auto propsIt = propsMap.find(family);
@@ -82,8 +85,9 @@ ShadowNode::Unshared cloneShadowTreeWithNewPropsUnmountedRecursive(
 
   if (affectedChildrenIt != childrenMap.end()) {
     for (const auto index : affectedChildrenIt->second) {
-      auto clone = cloneShadowTreeWithNewPropsUnmountedRecursive(children[index], childrenMap, propsMap);
-      if (clone != children[index]){
+      auto clone = cloneShadowTreeWithNewPropsUnmountedRecursive(
+          children[index], childrenMap, propsMap);
+      if (clone != children[index]) {
         shadowNode->replaceChild(*children[index], clone, index);
       }
     }
@@ -100,23 +104,24 @@ ShadowNode::Unshared cloneShadowTreeWithNewPropsUnmountedRecursive(
           propsParserContext, newProps, RawProps(props));
     }
   }
-  
+
   if (newProps) {
-    auto& props = shadowNode->getProps();
-    auto& mutableProps = const_cast<Props::Shared&>(props);
+    auto &props = shadowNode->getProps();
+    auto &mutableProps = const_cast<Props::Shared &>(props);
 
 #ifdef ANDROID
-    auto& newPropsRef = const_cast<Props&>(*newProps);
+    auto &newPropsRef = const_cast<Props &>(*newProps);
     newPropsRef.rawProps = mergeDynamicProps(
-            mutableProps->rawProps,
-            newProps->rawProps,
-            NullValueStrategy::Override);
+        mutableProps->rawProps,
+        newProps->rawProps,
+        NullValueStrategy::Override);
 #endif
     mutableProps = newProps;
-    auto layoutableShadowNode = static_pointer_cast<YogaLayoutableShadowNode>(shadowNode);
+    auto layoutableShadowNode =
+        static_pointer_cast<YogaLayoutableShadowNode>(shadowNode);
     layoutableShadowNode->updateYogaProps();
   }
-  
+
   return shadowNode;
 }
 
@@ -124,20 +129,23 @@ RootShadowNode::Unshared cloneShadowTreeWithNewProps(
     const RootShadowNode &oldRootNode,
     const PropsMap &propsMap) {
   auto childrenMap = calculateChildrenMap(oldRootNode, propsMap);
-  
+
   // This cast is safe, because this function returns a clone
   // of the oldRootNode, which is an instance of RootShadowNode
-  return std::static_pointer_cast<RootShadowNode>(cloneShadowTreeWithNewPropsRecursive(oldRootNode, childrenMap, propsMap));
+  return std::static_pointer_cast<RootShadowNode>(
+      cloneShadowTreeWithNewPropsRecursive(oldRootNode, childrenMap, propsMap));
 }
 
 RootShadowNode::Unshared cloneShadowTreeWithNewPropsUnmounted(
     RootShadowNode::Unshared const &oldRootNode,
-    const PropsMap &propsMap){
+    const PropsMap &propsMap) {
   auto childrenMap = calculateChildrenMap(*oldRootNode, propsMap);
-  
+
   // This cast is safe, because this function returns a clone
   // of the oldRootNode, which is an instance of RootShadowNode
-  return std::static_pointer_cast<RootShadowNode>(cloneShadowTreeWithNewPropsUnmountedRecursive(oldRootNode, childrenMap, propsMap));
+  return std::static_pointer_cast<RootShadowNode>(
+      cloneShadowTreeWithNewPropsUnmountedRecursive(
+          oldRootNode, childrenMap, propsMap));
 }
 
 } // namespace reanimated
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
@@ -1,11 +1,33 @@
 #ifdef RCT_NEW_ARCH_ENABLED
 
 #include <reanimated/Fabric/ShadowTreeCloner.h>
+#include <react/renderer/core/DynamicPropsUtilities.h>
 
 #include <ranges>
 #include <utility>
 
 namespace reanimated {
+
+ChildrenMap calculateChildrenMap(const RootShadowNode &oldRootNode, const PropsMap &propsMap){
+  ChildrenMap childrenMap;
+
+  for (auto &[family, _] : propsMap) {
+    const auto ancestors = family->getAncestors(oldRootNode);
+
+    for (const auto &[parentNode, index] :
+         std::ranges::reverse_view(ancestors)) {
+      const auto parentFamily = &parentNode.get().getFamily();
+      auto &affectedChildren = childrenMap[parentFamily];
+
+      if (affectedChildren.contains(index)) {
+        continue;
+      }
+
+      affectedChildren.insert(index);
+    }
+  }
+  return childrenMap;
+}
 
 ShadowNode::Unshared cloneShadowTreeWithNewPropsRecursive(
     const ShadowNode &shadowNode,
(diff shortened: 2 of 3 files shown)


bartlomiejbloniarz:
@chrfalch I think it's ready to be tested now. Thanks!

j-piasecki (approved):
LGTM, but I'm still unsure whether `getHasBeenPromoted` is a strong enough condition.

bartlomiejbloniarz:
@j-piasecki Yeah, I'm also worried about that. I think that if there will be no issues during testing and this gets merged, we will release a RC version.

mhoran:
While this patch resolves the modal issue for me on iOS, it seems to have completely broken `useAnimatedStyle` for my [use case](https://github.com/mhoran/weechatRN/blob/f077cb7fe8b575cb93572cd2e129b2bfc89d3dc7/src/usecase/shared/KeyboardAvoidingView.tsx#L29-L39). Removing `useAnimatedStyle` and just applying padding (no animated style) works -- so it seems this patch isn't quite enough.

bartlomiejbloniarz:
@mhoran It breaks on both iOS and Android?

Vali-98:
Tested this on Android - this fixes the issue where modals are rendered off-screen or have 0 height/width causing them to be invisible.

However, this does not fix cases where Entering animations of an AnimatedView inside a modal seems to have 0 height/width, despite onLayout reporting the heights and widths properly.

mhoran:
> @mhoran It breaks on both iOS and Android?

Yes, in slightly different ways.

On iOS, I see two issues. First, issues with `TextInput` `onFocus` events triggering an animation. This gets stuck in a loop, opening and then immediately closing the keyboard. 

The second issue is an interaction with `FlatList`. The `FlatList` cannot be scrolled; it gets stuck.

[Screen Recording](https://github.com/user-attachments/assets/10c13c44-b681-4e42-9bd5-4e73b2ce13ec)

On Android, I only see the `FlatList` issue, and less often; but it is still reproducible.

I found that this seems to be some interaction with `react-native-drawer-layout`. The issue only seems to happen when the drawer is open when the app starts.

I have created a reproducer here: https://github.com/mhoran/react-native-scratch/tree/reanimated-commit-hook-issue. Note this is using a local checkout of `react-native-reanimated` and expects it to be located next to the repo. Change `metro.config.js` and `package.json` as necessary. It uses `npm` as the package manager and leverages symlinks (hence the `metro.config.js` changes.)

gaearon:
Is it possible to fix the root issue instead? It sounds like this is going against immutable tree updates — which is worrying to me because it is a fundamental constraint of the new RN architecture. It is worrying when Reanimated as a library keeps diverging from how RN works instead of getting closer to it.

Looking at https://github.com/software-mansion/react-native-reanimated/issues/6659#issuecomment-2468380063:

>But whenever we clone a ShadowNode that has not been mounted, YogaLayoutableShadowNode re-clones all of its children. When cloning, RN uses the last mounted state, meaning that it actually overrides the height assigned by RNScreens, with wrong value.

This sounds like the proper fix should either be in RNScreens (does _it_ mutate stuff it shouldn't?) or in how it interacts with Yoga. Can the fix be investigated there instead?

gaearon:
cc @sammy-SC maybe you have thoughts on the right way to handle this

gaearon:
Basically, the point I'm trying to make is that immutable tree updates is the _entire premise_ of the Fabric architecture. It's a fundamental constraint. Every technical decision about how RN works will be relying on this constraint — both existing decisions and future decisions. So, in my understanding, it is not appropriate for a library to break this constraint and mutate these nodes — or worse, fight over who gets to mutate these nodes. If one library starts undermining this constraint, it isn't a good reason for another library to give in and undermine it even further. Instead, the concern needs to be brought upstream and a resolution _within the architecture_ needs to be found. Reanimated is used very widely so if it starts violating the fundamental constraints of Fabric, this makes it hard for RN itself to change its implementation and take advantage of those constraints. Since changes that were previously unobservable would now become observable. I think this is very important and I want to make sure that this point of view is heard (you probably hear a lot of pragmatic "just fix it now somehow" so I want to balance that out a bit). Thanks.

chrfalch:
I agree with @gaearon as well. From Expo's side this is also something we feel very strongly about getting right - these libraries (REA/RNScreens) is as you say a fundamental part of almost all RN apps out there and will affect so many other libraries as well (React Navigation and a lot of other libs).

I don't think the problem is mutating stuff that shouldn't be mutated - If I remember correctly the issue started as an optimization to make sure Reanimated didn't clone shadow nodes too often. The problem was that when running animations the code ended up cloning single shadow nodes multiple times in each pass - causing a performance degradation (correct me if I'm wrong, @bartlomiejbloniarz).

A PR (#6214) was merged that optimized this to avoid "over-cloning" - but it seemed that the PR also introduced some side effects causing some layout issues that was mentioned in one of the first issues (#6659).

A suggested way forward would be to remove the optimizations to create a baseline and verify that all the reported issues are working correct. Then we can measure and investigate which optimizations that should be made. 

As it is now, I also feel a bit insecure about what Dan also pointed out - the root cause of this problem (in the new architecture).

bartlomiejbloniarz:
@chrfalch Reverting #6214 wouldn't actually solve the problem. I mentioned in #6659 that this issue was also present in the old algorithm. Funnily enough it didn't show up in the Debug mode there, because we were updating `ShadowNode`s in-place, as an optimization. The new algorithm actually stopped in-place updates, and that's why those issues manifested. 

I agree that the root cause of this issue is:
> But whenever we clone a `ShadowNode` that has not been mounted, `YogaLayoutableShadowNode` re-clones all of its children. When cloning, RN uses the last mounted state, meaning that it actually overrides the height assigned by RNScreens, with wrong value.

I am not sure when this behavior was introduced to RN, so that's something I need to examine.

@gaearon let me explain in more detail what we are doing here, as we are not *exactly* breaking the `ShadowTree` immutability, but as I was debugging the problems that this PR introduces, I am getting more convinced that our approach needs radical rethinking. Basically:
- our animations are applied by simply commiting a new `ShadowTree` every frame
- those updates are not seen by the RN renderer, as the new `ShadowNodes` are not pushed to JS 
- so, if RN performs a render when an animation is running, it will commit a `ShadowTree`, that doesn't contain our animation updates, resulting in a flicker
- to mitigate that, we use the `CommitHook` api. Basically, the `shadowTreeWillCommit` method is called whenever a new tree is about to be commited (before the layout is calculated). This method allows us to inspect the new tree, and return a new one, that will be commited instead. 
- so we have a big registry of all the updates, that we reapply on every non-reanimated commit

This PR is only about the algorithm that we use to modify the tree that we get in the `shadowTreeWillCommit` method. Before this PR (#6776), we would clone every `ShadowNode` that we wanted to animate (and also the entire path from there to the root). With my changes we instead check if a node *has been commited* and if not, then we modify it in place. So we are not *mutating* the mounted `ShadowTree`. We are just *intercepting* the tree that's about to be commited (using the dedicated CommitHook api) and *mutating* it before it's actually promoted to be the new `ShadowTree`. The motivation behind this change was that, when we cloned those un-commited `ShadowNodes` the whole tree would be re-cloned and we would lose `State` updates (and things like the `Modal` height in #6659 would break).

But again, this approach comes with its own problems, that came up in testing, and it's very likely that this needs to be rethinked.

gaearon:
Makes sense, thanks for explaining! Seems important that the team finds some approach that integrates nicely with RN itself. Does RN.Animated avoid these problems somehow?

chrfalch:
> @chrfalch Reverting #6214 wouldn't actually solve the problem. I mentioned in #6659 that this issue was also present in the old algorithm. Funnily enough it didn't show up in the Debug mode there, because we were updating `ShadowNode`s in-place, as an optimization. The new algorithm actually stopped in-place updates, and that's why those issues manifested.

You explained it very well - thanks @bartlomiejbloniarz (also for looking into this is such a throughout way)

bartlomiejbloniarz:
@gaearon 

Animated doesn't really go through the same pipeline, since it either applies transforms directly to views, when using `nativeDriver` or it re-renders the tree on every frame from JS.

bartlomiejbloniarz:
@mhoran Could you test this version of this PR on your app?

chrfalch:
@bartlomiejbloniarz - sorry for the long wait - but I've now spent some time testing this issue and can verify that it now seems to be working correct in Debug/Release builds using the reproduction project - both with/without the last commit you've added. Great work!

mhoran:
> @mhoran Could you test this version of this PR on your app?

On iOS, the issues I reported with the previous version of this patch look to be resolved with the latest patch. I still see a regression on new arch if padding is changed in the same `useAnimatedStyle` block as a `transform` (the screen flickers), but perhaps that is unrelated to the commit hook issue.

Things are looking much better on Android as well. This seems to resolve issues I was having with `TouchableOpacity` not knowing whether it was pressed. I'm now able to navigate to screens that were previously inaccessible due to that bug. I'm getting a hang with animated padding, but that may also be unrelated to the commit hook issue, as it is not affected by this patch.

Great work!

mhoran:
One small issue remains with modals on both Android and iOS. On Android, I've seen the modal animate almost entirely off-screen to the top left corner before jumping to the correct position. On iOS, modals sometimes animate in from the bottom right instead of bottom center. Unfortunately I've found this incredibly difficult to reproduce, but there still seems to be some sort of issue with modals in particular.

samducker:
This issue seems to affect useAnimatedKeyboard also I am having performance issues with react-native-gifted-chat (which uses useAnimatedKeyboard and reanimated) since I moved to the new architecture. When I open the keyboard there is noticeable heavy lag similar to the issue with modals like gorham/bottom-sheet.

mhoran:
> This issue seems to affect useAnimatedKeyboard also I am having performance issues with react-native-gifted-chat (which uses useAnimatedKeyboard and reanimated) since I moved to the new architecture. When I open the keyboard there is noticeable heavy lag similar to the issue with modals like gorham/bottom-sheet.

How are you animating the view? If you have padding (or margin or height) in an animated style, it will cause flickering. This is a regression in new arch, but I'm not sure this issue is the root cause.

Animated padding (margin and height) had poor performance on old arch (see #5685), but in new arch, even having a static padding, margin or heading value in an aimated style block will cause flickering.

I worked around this by spotting my animated styles in two (one for the transform, one for padding), and ensuring that the padding style updates as infrequently as possible.

samducker:
Hey @mhoran yeah my bad I think you're right as I just applied the new patch and it is not resolving my issues with gorhom/bottom-sheet lagging and lack of space at the bottom + the keyboard issue I mentioned

I made this assumption based on the other lagging issues related to this PR and reading this issue in gorhom bottom sheet [https://github.com/gorhom/react-native-bottom-sheet/issues/2046](https://github.com/gorhom/react-native-bottom-sheet/issues/2046)

It is also not resolving my keyboard issues in gifted chat either, the code is from gifted chat and is here [https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/GiftedChat.tsx](https://github.com/FaridSafi/react-native-gifted-chat/blob/master/src/GiftedChat.tsx)

bartlomiejbloniarz:
@mhoran Do you see the issue with modal animation only with this PR or is it something that also happens on main?

mhoran:
> @mhoran Do you see the issue with modal animation only with this PR or is it something that also happens on main?

On Android, modals always animate to the top left corner and remain off screen. With this PR, they only animate to the top left corner sometimes, and always wind up in the correct place. The animation is just wrong, sometimes.

On iOS, this PR seems to introduce a regression, but it's very hard to tell. I only see modal issues on iOS when the window has been resized to less than full screen using Stage Manager. The modals display in odd places, and do not show entirely on screen. As such, it's very difficult to tell where it is animating from, since the full modal is not visible. It could be that this issue exists on main or it could be a regression, and reproducing is very tricky.

latekvo:
When testing with the modal bug repro reported here: #6659, the invalid modal rendering is still visible during the modal entrance animation.
This issue seems to only occur on a physical device. 

<details>
<summary>
Test code
</summary>

```js
import {useState} from 'react';
import {Button, Modal, SafeAreaView, StatusBar, View} from 'react-native';
import Animated, {useAnimatedStyle} from 'react-native-reanimated';

  const [visible, setVisible] = useState(false);

  const animatedStyle = useAnimatedStyle(() => ({
    transform: [{translateX: 0}],
  }));

  return (
    <SafeAreaView
      style={{
        flex: 1,
        backgroundColor: '#fff',
        padding: 16,
      }}>
      <StatusBar />
      <View
        style={{
          paddingVertical: 16,
        }}
      />
      <Button
        title="Open Modal"
        onPress={() => {
          setVisible(true);
        }}
      />
      <Animated.View style={animatedStyle} />
      <Modal visible={visible} animationType="slide">
        <View
          style={{
            padding: 32,
            backgroundColor: 'pink',
          }}/>
      </Modal>
    </SafeAreaView>
  );
}

```

</details>

On the following video, the entrance animation is 8 frames long.
During the first 3 frames of the animation - there's a pink square moving on the bottom-left corner of the screen, before being covered up by the modal.

https://github.com/user-attachments/assets/876e069b-7105-410e-8b46-346fdd62f067

mhoran:
As mentioned in https://github.com/react-navigation/react-navigation/commit/9831292375ddf6daca1cbf82cbe904ed4d73f7db#commitcomment-150759476, this patch no longer worked for me on iOS once I upgraded react-native-drawer-layout to 4.1.1. I see issues with ScrollViews and cannot open the keyboard, as if `dirtyLayout()` was not added.

kslbdev:
Hello, getting a crash with this change, not always reproducible.
[reanimated_crash.txt](https://github.com/user-attachments/files/18593591/reanimated_crash.txt)

jiroscripts:
Hello @bartlomiejbloniarz 

Do you have any news on this PR, please?

As mentionned by @kslbdev, we also experienced many crashes on iOS which looks like what he shared.

karimserhan:
Hi @bartlomiejbloniarz , thank you for working on this. Would you mind sharing status update?

bartlomiejbloniarz:
I'm finally back to working on this issue! Sorry for leaving it in this state for so long.

From what I've seen this solution causes some new issues, so I have to carefully test it out, and maybe come up with a new solution. 

cc @chrfalch

mhoran:
@bartlomiejbloniarz I've been trying to triage issues that seem to be related (see #6865, #6915, #6937, and #7052). They seem to exhibit the behavior of Reanimated and React Native state conflicts, resulting in flickering animations. I've found that applying this patch does *not* fix those issues -- so it could still be something else, but it all seems related to the underlying issue you said this PR should fix.

I noticed that `main` has diverged quite a bit now and some refactorings have been made around the commit and mount hooks. However, I still see all the issues described here as well as the linked issues, so something is still amiss.

> On iOS, this PR seems to introduce a regression, but it's very hard to tell. I only see modal issues on iOS when the window has been resized to less than full screen using Stage Manager. The modals display in odd places, and do not show entirely on screen. As such, it's very difficult to tell where it is animating from, since the full modal is not visible. It could be that this issue exists on main or it could be a regression, and reproducing is very tricky.

This PR doesn't introduce the regression described here, but adding Reanimated to the project does. Without Reanimated, modals do animate from the bottom of the screen to the center as configured. When Reanimated is added to the project, they sometimes animate in from the bottom right corner. This only happens when using Stage Manager and resizing the application window to less than full screen.

As mentioned in my [comment](https://github.com/software-mansion/react-native-reanimated/pull/6776#issuecomment-2563348415) above, this PR seems to make things worse when certain properties are animated at certain times, e.g. `zIndex`.

I was hopeful that #7014 could help with these issues, but applying that commit did not seem to address any of the issues I've been encountering.

Happy to help test out any half-baked ideas you may have. I've been poking at this for a few weeks now but don't really know the internals of React Native or Reanimated well enough to get very far. Feel free to ping me on Discord, username is the same as my GitHub handle.

bartlomiejbloniarz:
@mhoran Thank you so much for looking into this 🙏

mhoran:
I tried taking a different approach by attempting to fix the underlying issue of React Native applying the last mounted rather than current `ShadowNode` state. I came across https://github.com/facebook/react-native/pull/44796 and applied the proposed change. Unfortunately it did not resolve issue with modals on Android or iOS. I'm not entirely sure if I made all the necessary changes in Reanimated, but I tried a few things and did not see any difference.

I also confirmed that the commit hook does seem to be responsible for the freezing animations I've been seeing (similar to #7052). This happens when a React Native commit occurs during an animation. I've narrowed the issue down to [`shadowNode.clone`](https://github.com/software-mansion/react-native-reanimated/blob/fa3aa31e6b314313d32d7ca8d148c4a76f3356e9/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp#L63-L66) in `cloneShadowTreeWithNewPropsRecursive`. If I skip this, then animations do not hang (though they do of course flicker due to the inconsistent state.)

bartlomiejbloniarz:
@mhoran Removing the `clone` call should also break all animations. Unless you are animating non-layout props that go through `synchronouslyUpdatePropsOnUi`.

The [PR](https://github.com/facebook/react-native/pull/44796) from j-piasecki won't solve this issue, as there are other places where the state can be lost.

chrfalch:
@bartlomiejbloniarz - would it be an idea to escalate this to meta to see if we get some help fixing the problem?

mhoran:
> @mhoran Removing the clone call should also break all animations. Unless you are animating non-layout props that go through synchronouslyUpdatePropsOnUi.

I created two paths through the `ShadowTreeCloner`, one for the commit hook that is effectively a no-op to the tree and one for the path through `ReanimatedModuleProxy` that still clones the nodes. This causes animations jump during commit but otherwise they run smooth, as compared to the freezing that happens otherwise. I just did this for troubleshooting to try and narrow down the issue; commenting out the call to `cloneShadowTreeWithNewProps` in the commit hook has the same effect, but I wanted to make sure the performance issues weren't due to the tree traversal in Reanimated, which was proved by this test.

bartlomiejbloniarz:
@mhoran ok, that makes sense

bartlomiejbloniarz:
@chrfalch This solution was based on some previous communication that we had with them. But now after seeing how many problems it introduces, I think escalating might be the best path forward.

bartlomiejbloniarz:
I think I finally have a good reproduction that separates the problem from reanimated. Will open an issue tomorrow.

mhoran:
I tried applying the change to `YogaLayoutableShadowNode` from https://github.com/facebook/react-native/issues/49694 to see if it fixed the modal position issue. On Android, both the animation and position issue are unaffected by the change (modal still animates partially offscreen to an position in the top left corner). On iOS, the position is now correct when using Stage Manager but the modal still animates in from the bottom right corner periodically. Perhaps the animation issue is something else -- but I was surprised to see the Android position was not fixed.

denysoleksiienko:
@danieljvdm In my project, I have a screen in the presentation 'modal' where I use a bottom sheet with reanimated. And I am experiencing the same problem on Android with flickering behaviour always when opening the modal screen, not a bottom sheet. Looks like when inizialization

lookis:
any update? @bartlomiejbloniarz 
this patch fixed my project!

bartlomiejbloniarz:
This PR is not a good solution. The problem that it was aiming to solve should be fixed in one of upcoming RN releases, when the `useShadowNodeStateOnClone` feature flag mentioned [here](https://github.com/facebook/react-native/issues/49694) becomes true by default.

j-piasecki:
> The problem that it was aiming to solve should be fixed in one of upcoming RN releases, when the useShadowNodeStateOnClone feature flag mentioned https://github.com/facebook/react-native/issues/49694 becomes true by default.

This flag should be enabled in RN 0.80 - https://github.com/facebook/react-native/pull/51041.

chrfalch:
`useShadowNodeStateOnClone` was back ported to [0.79.2](https://github.com/facebook/react-native/releases/tag/v0.79.2) as well :)

PR #6779, opened 2024-12-02
Clear `operationsInBatch_` before terminating UI runtime

tomekzaw:
## Summary

This PR fixes the following crash that happens during a reload due to `jsi::Value` outliving the UI runtime.

The offending `jsi::Value` comes from `operationsInBatch_` inside `ReanimatedModuleProxy` (formerly `NativeReanimatedModule`.

In `~ReanimatedModuleProxy` we manually clear data structures containing `jsi::Value` prior to resetting `uiWorkletRuntime_` which effectively terminates the runtimes. It looks like we forgot about `operationsInBatch_`.

Screenshot 2024-12-02 at 13 00 08

## Test plan

1. Launch FabricExample app
2. Reload the app several times
3. Repeat steps 1 and 2 several times

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
@@ -198,7 +198,9 @@ ReanimatedModuleProxy::~ReanimatedModuleProxy() {
   // runtime, so they have to go away before we tear down the runtime
   eventHandlerRegistry_.reset();
   frameCallbacks_.clear();
+#ifdef RCT_NEW_ARCH_ENABLED
   operationsInBatch_.clear();
+#endif // RCT_NEW_ARCH_ENABLED
   uiWorkletRuntime_.reset();
 }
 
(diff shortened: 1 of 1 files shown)


packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -198,6 +198,9 @@ ReanimatedModuleProxy::~ReanimatedModuleProxy() {
   // runtime, so they have to go away before we tear down the runtime
   eventHandlerRegistry_.reset();
   frameCallbacks_.clear();
+#ifdef RCT_NEW_ARCH_ENABLED
+  operationsInBatch_.clear();
+#endif // RCT_NEW_ARCH_ENABLED
   uiWorkletRuntime_.reset();
```
tjzel, line 204: IMO this kind of pattern in the code - explicit cleanups - is going to resurface in the future. Can't we somehow slightly refactor this bit so we'd use RAII instead?

tomekzaw: I think we could move `uiWorkletRuntime_` in header file so that it is deallocated at the end, what do you think?

PR #6768, opened 2024-11-27
Bump example apps to React Native 0.77

tomekzaw:
## Summary

This PR bumps `react-native` in paper-example and fabric-example from `0.76.5` to `0.77.0-rc.6` (golden RC) accordingly to https://react-native-community.github.io/upgrade-helper/?from=0.76.5&to=0.77.0-rc.6.

Apart from this, changes have been made in macos-example, tvos-example and web-example to align versions of common native dependencies so they are hoisted by yarn.

I had to patch `@shopify/flash-list`:
* https://github.com/Shopify/flash-list/issues/1466 – fixes Android build
* https://github.com/Shopify/flash-list/pull/1041 – fixes `scrollTo` example

Changes required by Reanimated to support RN 0.77 have been moved to separate PRs for the sake of atomicity and transparency:
* https://github.com/software-mansion/react-native-reanimated/pull/6875
* https://github.com/software-mansion/react-native-reanimated/pull/6876
* https://github.com/software-mansion/react-native-reanimated/pull/6883
* https://github.com/software-mansion/react-native-reanimated/pull/6884

PR #6783, opened 2024-12-03
refactor: restore JSCallInvoker

tjzel:
## Summary

Turns out that majority of work with regards to using `RuntimeExecutor` done in 
- https://github.com/software-mansion/react-native-reanimated/pull/5734
can be reverted since `jsCallInvoker` is staying in the API.

Thanks to that we can significantly cleanup the code which is helpful for Worklets.

## Test plan

I tested the changes on 0.74, 0.75 and 0.76 for Android/iOS Paper/Fabric and all worked well.


packages/react-native-reanimated/android/src/main/java/com/swmansion/worklets/JSCallInvokerResolver.java
```diff
@@ -0,0 +1,27 @@
...
+    try {
+      var catalystInstance = context.getClass().getMethod("getCatalystInstance").invoke(context);
+      assert catalystInstance != null;
+      var method = catalystInstance.getClass().getMethod("getJSCallInvokerHolder");
+      return (CallInvokerHolderImpl) method.invoke(catalystInstance);
+    } catch (Exception e) {
+      throw new RuntimeException("Failed to get JSCallInvokerHolder", e);
+    }
+  }
+}
```
tjzel, line 27: In my opinion it's more maintainable to use a reflection in this case compared to source sets. I will add relevant comments here.

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
```diff
@@ -279,38 +276,18 @@ - (void)sendEventWithName:(NSString *)eventName body:(id)body
...
+  auto jsCallInvoker = self.bridge.jsCallInvoker;
+  auto jsiRuntime = reinterpret_cast<facebook::jsi::Runtime *>(self.bridge.runtime);
+
+  assert(jsiRuntime != nullptr);
+
+  auto reanimatedModuleProxy =
+      reanimated::createReanimatedModule(self, self.bridge, jsCallInvoker, workletsModule, _isBridgeless);
+
+  jsi::Runtime &rnRuntime = *jsiRuntime;
+  [self commonInit:reanimatedModuleProxy withRnRuntime:rnRuntime];
```
tomekzaw, line 289: We shouldn't use `self.bridge.jsCallInvoker` in bridgeless mode. You have to do the following:
```objc
#ifdef RCT_NEW_ARCH_ENABLED
@synthesize callInvoker = _callInvoker;
#endif // RCT_NEW_ARCH_ENABLED

auto jsCallInvoker = _callInvoker.callInvoker;
```

Let's restore the original implementation without refactoring the whole function. See https://github.com/software-mansion/react-native-reanimated/pull/6768/files#diff-06f7028a5f8a14ac525b70098bbf48937abe91707aea813bc4528ec9c166f5bc

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
```diff
@@ -327,6 +304,7 @@ - (void)commonInit:(std::shared_ptr<ReanimatedModuleProxy>)reanimatedModuleProxy
   WorkletRuntimeCollector::install(rnRuntime);
   RNRuntimeDecorator::decorate(rnRuntime, reanimatedModuleProxy);
 #ifdef RCT_NEW_ARCH_ENABLED
+  [self attachReactEventListener];
```
tomekzaw, line 307: Why do we need this change here?

tjzel: We'd call it on Fabirc before `commonInit` here https://github.com/software-mansion/react-native-reanimated/pull/6783/files#diff-06f7028a5f8a14ac525b70098bbf48937abe91707aea813bc4528ec9c166f5bcL296 so I moved it into `commonInit` fabric section.

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
tjzel: [ff3eeb2](https://github.com/software-mansion/react-native-reanimated/pull/6783/commits/ff3eeb2ec53004767ab8aa97c08813fc12fec41d)

piaskowyk (approved):
Now looks good to me 👍

tomekzaw (approved):

PR #6785, opened 2024-12-03
chore: rephrase render shared value read warning

tjzel:
## Summary

Turns out it can be understood that you should be using `get` in render instead of `.value`.


MatiPl01 (approved):
LGTM

PR #6789, opened 2024-12-04
refactor(Worklets): move UIScheduler 

tjzel:
## Summary

This pull requests moves `UIScheduler`, a class responsible for scheduling operations to the Main Thread, to Worklets. While it seems that `Reanimated` should have some freedom scheduling to the Main Thread if it doesn't use the UI Runtime, I think it's better to move it now as is and add some secondary APIs later on. We'll have a better overview of what we need once more code is moved to Worklets.

## Test plan

- CI
- [x] Working 0.76 Paper/Fabric iOS/Android


piaskowyk (approved):
👍

PR #6790, opened 2024-12-04
feat(Worklets): react-native-worklets node package stub

tjzel:
## Summary

Creating a workspace for `react-native-worklets` which we'll gradually expand.

## Test plan

🚀


packages/react-native-worklets/RNWorklets.podspec
tjzel, line 1: It's partially generated via `npx create-react-native-library` and partially copied from `RNReanimated.podspec`. It has some code which currently doesn't do anything - it will be needed later on.

packages/react-native-worklets/android/CMakeLists.txt
tjzel, line 1: It's copied from `react-native-reanimated/android/CMakeLists.txt`. It has some code which currently doesn't do anything - it will be needed later on.

packages/react-native-worklets/android/build.gradle
tjzel, line 1: It's partially generated via `npx create-react-native-library` and partially copied from `react-native-reanimated/android/build.gradle`. It has some code which currently doesn't do anything - it will be needed later on.

packages/react-native-worklets/android/src/main/cpp/worklets/CMakeLists.txt
tjzel, line 1: It's copied from `react-native-reanimated/android/src/main/cpp/worklets/CMakeLists.txt`. It has some code which currently doesn't do anything - it will be needed later on.

packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/DummyWorkletsModule.java
tjzel, line 1: Dummy module seems to be necessary for the package to be registered and compiled into the app.

PR #6792, opened 2024-12-05
refactor(Worklets): move UIWorkletRuntime

tjzel:
## Summary

Reanimated no longer creates UI Worklet Runtime. Now it's created by Worklets and Reanimated only borrows it.

## Test plan

See that apps are working and zooming 👍


tjzel:
@piaskowyk Later on - definitely. A time will come when we'll think how the memory should be shared by the modules. For now though, let everything be borrowed when it's needed to decouple the code as much as we can.

PR #6796, opened 2024-12-08
Fix crash caused by using `ShareableHandle` in multiple remote runtimes

tomekzaw:
## Summary
Fixes https://github.com/Expensify/react-native-live-markdown/issues/574. See root cause analysis in https://github.com/Expensify/react-native-live-markdown/issues/574#issuecomment-2526377216.

This PR fixes a crash caused by calling `ShareableHandle::toJSValue` with second remote runtime after initializing `remoteValue_` with a `jsi::Value` belonging to the first remote runtime.

I assume that this is a rare scenario so we only memoize the value for the first remote runtime and we recreate the value for all subsequent runtimes.

## Test plan

Reproduction:

<details>
<summary>EmptyExample.tsx</summary>

```tsx
import { Text, StyleSheet, View } from 'react-native';

import React from 'react';
import {
  createWorkletRuntime,
  runOnRuntime,
  useAnimatedStyle,

const regex = /\d/;

const workletRuntime = createWorkletRuntime('another');

  useAnimatedStyle(() => {
    console.log('useAnimatedStyle', String(regex));
    return {};
  });

  runOnRuntime(workletRuntime, () => {
    'worklet';
    console.log('runOnRuntime', String(regex));
    return {};
  })();

  return (
      <Text>Hello world!</Text>
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

</details>

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/worklets/SharedItems/Shareables.cpp
+++ b/packages/react-native-reanimated/Common/cpp/worklets/SharedItems/Shareables.cpp
@@ -293,7 +293,8 @@ jsi::Value ShareableHandle::toJSValue(jsi::Runtime &rt) {
     return jsi::Value(rt, *remoteValue_);
   }
   auto initObj = initializer_->toJSValue(rt);
-  return getValueUnpacker(rt).call(rt, initObj, jsi::String::createFromAscii(rt, "Handle"));
+  return getValueUnpacker(rt).call(
+      rt, initObj, jsi::String::createFromAscii(rt, "Handle"));
 }
 
 jsi::Value ShareableString::toJSValue(jsi::Runtime &rt) {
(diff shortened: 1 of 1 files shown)


tomekzaw:
Please confirm that we don't need any additional synchronization involving `initializationMutex_` prior to merging.

cc @tjzel who touched this code some time ago

tjzel (approved):
LGTM

piaskowyk:
Changes look ok, and probably we also need to add that logic to RetainingShareable. What do you think? 🤔

tomekzaw:
Looks like `RetainingShareable` already supports multiple runtimes. There's a comparison of `&rt` with `primaryRuntime_` and `secondaryRuntime_` and there's a call to `BaseClass::toJSValue(rt)` that generates a new instance for yet another runtime:
https://github.com/software-mansion/react-native-reanimated/blob/2a947a894941c16ef25a715f689bdc8fa9540b84/packages/react-native-reanimated/Common/cpp/worklets/SharedItems/Shareables.cpp#L142-L163

PR #6807, opened 2024-12-11
feat: preliminary use of react-native-worklets in reanimated

tjzel:
## Summary

Adding necessary code to detect if `react-native-worklets` is available and change the build behavior. At the moment we forcefully make it `false` since `react-native-worklets` is a stub.

## Test plan

CI


packages/react-native-reanimated/android/src/main/cpp/reanimated/CMakeLists.txt
tjzel, line 1: Additional `find_package` and linking here is due to fact that `react-native-reanimated/android/src/main/cpp/reanimated/CMakeLists.txt` and `react-native-reanimated/android/src/main/cpp/worklets/CMakeLists.txt` aren't working in separation. Not including the worklets' CMakeLists.txt causes Reanimated to fail at compilation because it depends on it secretly.

packages/react-native-reanimated/android/src/main/cpp/reanimated/CMakeLists.txt
```diff
@@ -29,10 +37,44 @@ endif()
...
+  # TODO: Refactor this when adding support for newest V8
+  target_include_directories(reanimated PRIVATE "${JS_RUNTIME_DIR}/src")
+  file(GLOB V8_SO_DIR "${JS_RUNTIME_DIR}/android/build/intermediates/\
+    library_jni/**/jni/${ANDROID_ABI}")
+  find_library(
+    V8EXECUTOR_LIB v8executor
+    PATHS ${V8_SO_DIR}
+    NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
+  target_link_libraries(reanimated ${V8EXECUTOR_LIB})
+endif()
```
piaskowyk, line 80: Should that code be guarded by the condition `if(NOT ${HAS_EXTERNAL_WORKLETS})` too? 🤔

tjzel: It works like this now but you are right, it would be more explicit that way.

packages/react-native-reanimated/android/build.gradle
```diff
@@ -232,8 +236,10 @@ android {
     }
 
     prefab {
-        worklets {
-            headers workletsPrefabHeadersDir.absolutePath
+        if(!hasExternalWorklets){
```
tomekzaw, line 239:
```suggestion
        if (!hasExternalWorklets) {
```

packages/react-native-reanimated/android/build.gradle
```diff
@@ -406,11 +413,13 @@ def assertMinimalReactNativeVersion = task assertMinimalReactNativeVersionTask {
     }
 }
 
-task prepareWorkletsHeadersForPrefabs(type: Copy) {
-    from("$projectDir/src/main/cpp")
-    from("$projectDir/../Common/cpp")
-    include("worklets/**/*.h")
-    into(workletsPrefabHeadersDir)
+if(!hasExternalWorklets) {
```
tomekzaw, line 416:
```suggestion
if (!hasExternalWorklets) {
```

packages/react-native-reanimated/RNReanimated.podspec
```diff
@@ -78,18 +78,24 @@ Pod::Spec.new do |s|
...
+      ss.subspec "apple" do |sss|
+        # Please be careful with the snakes.
+        # 🐍🐍🐍
+        # Thank you for your understanding.
+        sss.source_files = "apple/worklets/**/*.{mm,h,m}"
+        sss.header_dir = "worklets"
+        sss.header_mappings_dir = "apple/worklets"
+      # end
     end
   end
```
tomekzaw, line 100:
```suggestion
      end
    end
  # end
```

piaskowyk (approved):
Looks good to me 👍 I've tested it on Fabric/Paper iOS/Android and it seems to work ok.

PR #6809, opened 2024-12-11
fix: missing `.get()` in `useDerivedValue` mock

kirillzyusko:
## Summary

In https://github.com/software-mansion/react-native-reanimated/pull/6312 new API with `get`/`set` methods was added.

However mock implementation wasn't updated and `useDerivedValue` still returns only `value`. This PR fixes this problem and returns getter as well.

## Test plan

Run test that uses `.get()` on a value returned from `useDerivedValue`,


tjzel (approved):
Thanks for your contribution!

PR #6811, opened 2024-12-12
chore: add cmake formatting and linting

tjzel:
## Summary

🚀 

## Test plan

Android linting CI


packages/react-native-reanimated/scripts/lint-cmake.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if ! which cmake-lint >/dev/null; then
+  echo "error: cmake-lint is not installed"
+  exit 1
+fi
+
+cmake-lint "$@"
```
tjzel, line 8: This also checks formatting.

packages/react-native-reanimated/scripts/format-cmake.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if ! which cmake-format >/dev/null; then
+  echo "error: cmake-format is not installed"
+  exit 1
+fi
+
+cmake-format -i "$@"
```
tjzel, line 8: At the moment we use the defaults for formatting. We could make some config if we want to, but I don't think we are opinionated about it.

bartlomiejbloniarz (approved):

PR #6812, opened 2024-12-12
chore: add external-worklets-example

tjzel:
## Summary

The new example is meant to test `react-native-worklets` and `react-native-reanimated` integration while both packages are installed in an app. The cases when `react-native-worklets` isn't installed are handled by all the other examples.

This is basically copying FabricExample and changing all the necessary bits to ExternalWorkletsExample.

For the time being we'll only test this on Fabric and newest React Native. Once we reach satisfying enough code we'll make necessary backports.

## Test plan

I added a brand new build CI for this example.


piaskowyk (commented):
What would be a different between ExternalWorkletsExample and Fabric Example?

tjzel:
@piaskowyk 
> What would be a different between ExternalWorkletsExample and Fabric Example?

ExternalWorkletsExample actually uses `react-native-worklets`, has it installed via its `package.json`. FabricExample doesn't have such dependency.

tjzel:
We decided against adding another example app for the sake of the simplicity of the repo.

PR #6813, opened 2024-12-12
refactor: restructure worklets source sets

tjzel:
## Summary

To use `react-native-worklets` dynamically we must have a convenient way of disabling Worklets code that's inside Reanimated. Therefore we need to adjust Worklets' Java file locations.

All Worklets Java files were moved to

`android/src/worklets/main`

with exception of 

`WorkletsModule.java` which is needed by Reanimated to link with `WorkletsModule`. It's Reanimated's version of this file - meaning, that `WorkletsModule.java` in `react-native-worklets` will be slightly different. This is necessary to work around codegen.

Similar story goes for `NativeWorkletsModuleSpec.java`.

## Test plan

CI


packages/react-native-reanimated/android/build.gradle
```diff
@@ -347,9 +347,19 @@ android {
     }
     sourceSets.main {
         java {
+            // Always present
+            srcDirs += "src/worklets/WorkletsModule/java"
```
piaskowyk, line 351: Why do we always do that? Is this just a temporary workaround or a permanent solution?

tjzel: I'm doing it to have a clear separation of Worklets and Reanimated in the file structure within `react-native-reanimated` so the "code duplication" trick would be simpler and more transparent. These exact files are necessary because Reanimated declares `WorkletsModule` in TypeScript and it can't be missing.

packages/react-native-reanimated/android/src/main/cpp/reanimated/CMakeLists.txt
```diff
@@ -18,10 +18,6 @@ if(${JS_RUNTIME} STREQUAL "hermes")
   find_package(hermes-engine REQUIRED CONFIG)
 endif()
 
-if(${JS_RUNTIME} STREQUAL "hermes")
-  find_package(hermes-engine REQUIRED CONFIG)
-endif()
-
```
tjzel, line 24: I accidentally duplicated (harmlessly) some code here in a previous PR after resolving merge conflict.

packages/react-native-reanimated/android/build.gradle
tjzel: I refurbished it slightly so it would be more clear now.

bartlomiejbloniarz (approved):

PR #6814, opened 2024-12-12
chore: remove extraneous packageManager field

tjzel:
## Summary

Added by accident during monorepo migration.


patrycjakalinska (approved):
🚋 🚆

PR #6827, opened 2024-12-17
feat: duplicate worklets code the smart way

tjzel:
## Summary

In the transitional period of `react-native-worklets` coming as a separate package, Reanimated should work with or without `react-native-worklets` installed. This requires the code of Worklets to be present both in `react-native-reanimated` and `react-native-worklets`.

This PR moves Worklets' files that were in Reanimated to `react-native-worklets` package and adds a script that makes hard links to these files. The linking happens during the `yarn build` step of `react-native-reanimated` - this guarantees that `react-native-reanimated` would still work without `react-native-worklets` and that its' npm package would still contain all the necessary files.

For Android some files had to be duplicated. This is because of codegen conflicts and namespace generation. However, that code is relatively small and put into respective source sets `externalWorklets/with` and `externalWorklets/without`.

Also:
- I moved `ReanimatedVersion.cpp/h` files back to Reanimated. We'll implement version checking for Worklets when we actually need it.
- I adjusted the CI to always do `yarn build` in Reanimated when necessary.
- Enabled detection of `react-native-worklets` in `react-native-reanimated`.
- I improved the detection mechanism of `react-native-worklets` for iOS.
- Renamed `NativeWorkletsModuleSpec.ts` to `NativeReaWorkletsModuleSpec.ts` to avoid codegen conflicts.
- Removed `DummyWorkletsModule` from `react-native-worklets` since now it has `WorkletsModule`.

## Test plan

Run an Example with these changes.

Then do `yarn add react-native-worklets` in `common-app` (it will use workspace version by default) in the example directory and run it again.


packages/react-native-reanimated/android/src/externalWorklets/with/com/swmansion/reanimated/ReanimatedPackage.java
tjzel, line 1: This is the version of `ReanimatedPackage.java` when `react-native-worklets` are installed. Note that it doesn't register `WorkletsModule`.

packages/react-native-reanimated/android/src/externalWorklets/without/com/swmansion/reanimated/ReanimatedPackage.java
```diff
@@ -25,16 +25,14 @@
...
-// we will host this module in Reanimated's package.
+// `WorkletsPackage` must be registered within `ReanimatedPackage` when
+// `react-native-worklets` isn't installed . Extra `...Package` files
+// are ignored by React Native tools.
 @ReactModuleList(
     nativeModules = {
-      WorkletsModule.class,
       ReanimatedModule.class,
       ReanimatedUIManager.class,
+      WorkletsModule.class,
```
bartlomiejbloniarz, line 35: Restore order!

bartlomiejbloniarz (approved):

PR #6833, opened 2024-12-19
fix: inject IS_FABRIC from worklets

tjzel:
## Summary

We need to inject it from the Decorator that executes sooner.

## Test plan

Example App doesn't throw errors anymore on bare BokehExample


tomekzaw (approved):
I can approve that but sounds like `_IS_FABRIC` is a Reanimated-specific concept and Worklets don't need to know the architecture.

tjzel:
@tomekzaw https://github.com/software-mansion/react-native-reanimated/pull/6836 refactors this bit of code, this PR is only a bandaid for now

tomekzaw:
Okay, thanks for explanation

piaskowyk (approved):

PR #6837, opened 2024-12-19
Remove `REAInitializerRCTFabricSurface`

piaskowyk:
## Summary

This PR cleans up the Reanimated initialization process for iOS on Fabric. The primary motivation for these changes is to remove `REAInitializerRCTFabricSurface`, which was used as a workaround to access the surface presenter on the new architecture in bridge-full mode.

⚠️ Please note that this PR drops support for bridge-full mode on the new architecture.

## Test plan

I need to wait until the reload issue is fixed on the main branch 😞


tomekzaw:
Finally! 😮

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
```diff
@@ -102,44 +97,12 @@ - (dispatch_queue_t)methodQueue
...
-  });
-}
-
-- (void)handleJavaScriptDidLoadNotification:(NSNotification *)notification
-{
-  [self attachReactEventListener];
-}
-
-- (void)attachReactEventListener
+- (void)attachReactEventListener:(std::shared_ptr<ReanimatedModuleProxy>)reanimatedModuleProxy
```
tomekzaw, line 105:
```suggestion
- (void)attachReactEventListener:(const std::shared_ptr<ReanimatedModuleProxy> &)reanimatedModuleProxy
```

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
```diff
@@ -148,19 +111,17 @@ - (void)attachReactEventListener
...
-              // as we cannot safely access worklet runtime here
-              // and also we don't care about topLayout events
-              return false;
-            }
-            return reanimatedModuleProxy->handleRawEvent(rawEvent, CACurrentMediaTime() * 1000);
-          });
-      [scheduler addEventListener:eventListener];
-    }
+    auto eventListener =
+        std::make_shared<facebook::react::EventListener>([reanimatedModuleProxy](const RawEvent &rawEvent) {
```
tomekzaw, line 115: Maybe we should pass `reanimatedModuleProxy` as `std::weak_ptr`?

tomekzaw (approved):

WoLewicki (commented):
Why are all those other changes done here, are they strictly connected to `REAInitializerRCTFabricSurface` removal? Or do they remove the `bridge` path completely? Or is there some other reason?

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
piaskowyk: Here, I can't pass a reference to a pointer because `attachReactEventListener` is called with a local pointer. If I pass a reference to smart pointer here and use that reference asynchronously in a block, the pointer will already be destroyed :/

packages/react-native-reanimated/apple/reanimated/apple/REAModule.mm
piaskowyk: I think using a shared pointer should be fine, but just to be safe, we could use a weak_ptr instead 🫡

piaskowyk:
> Why are all those other changes done here, are they strictly connected to `REAInitializerRCTFabricSurface` removal? Or do they remove the bridge path completely? Or is there some other reason?

@WoLewicki That PR also drops support for bridge-full mode

piaskowyk:
I've tested it with reloads on both Paper and Fabric, and everything works fine 👍

tomekzaw (approved):

PR #6838, opened 2024-12-19
Fix `JS_RUNTIME_HERMES` flag definition

piaskowyk:
## Summary

This PR streamlines the definition of the `JS_RUNTIME_HERMES` flag. Previously, the flag wasn't consistently defined across all usage points (which prevented the compilation of certain branches of code - for example in WorkletsModuleProxy.cpp), and it also had duplicated definitions. Additionally, I've removed outdated and unnecessary checks from the flag definition.


packages/react-native-reanimated/Common/cpp/worklets/Tools/defs.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+// JS_RUNTIME_HERMES is only set on Android so we have to check __has_include
+// on iOS.
```
tomekzaw, line 4: Let's mention in the comment that on Android, `JS_RUNTIME_HERMES` is set in CMakeLists.txt. On iOS, there's no simple way to detect if Hermes is enabled or not, so we check the existence of the header file.

packages/react-native-reanimated/Common/cpp/worklets/WorkletRuntime/ReanimatedHermesRuntime.cpp
```diff
@@ -5,19 +5,14 @@
...
+#include <hermes/hermes.h>
 #include <jsi/decorator.h>
 #include <jsi/jsi.h>
 
 #include <memory>
 #include <string>
 #include <utility>
 
-#if __has_include(<reacthermes/HermesExecutorFactory.h>)
-#include <reacthermes/HermesExecutorFactory.h>
```
tomekzaw, line 16: Can you please confirm in which version of RN `HermesExecutorFactory.h` file was removed?

tomekzaw (approved):

packages/react-native-reanimated/Common/cpp/worklets/WorkletRuntime/ReanimatedHermesRuntime.cpp
piaskowyk: Actually, both files exist, but `#include <hermes/hermes.h>` is all we need to use. I've run compatibility tests on CI, and this change appears to be safe (expect one flaky check)

piaskowyk:
I've also tested it with JSC, and it works properly 👍

packages/react-native-reanimated/Common/cpp/worklets/Tools/defs.h
tjzel, line 1: Change the name to capital `Defs.h` since for some reason all our cpp files are capital case.

tjzel (approved):

PR #6850, opened 2024-12-23
refactor: explicit invalidations for native and cpp

tjzel:
## Summary

This pull requests refactors memory management of Worklets and Reanimated.

Basically, since Reanimated can obtain WorkletsModuleProxy and the Worklet Runtimes as shared pointers, it has to release them explicitly during the invalidation stage of Native Modules. Releasing them later on (e.g. via deconstructors) might lead into issues and crashes.

Ideally we'd instead use some different solution here than shared pointers, but it can wait as it's not mandatory at the moment and could be a significant refactor.

Fixes:
- iOS crash on reload
- Android crash on SingleInstanceChecker during third reload

## Test plan

- [x] 0.76 iOS/Android Fabric works
- [x] 0.76 iOS/Android Paper works
- [x] 0.75 iOS/Android Fabric works
- [x] 0.75 iOS/Android Paper works
- [x] 0.74 iOS/Android Fabric works
- [x] 0.74 iOS/Android Paper works


packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.h
```diff
@@ -46,19 +46,19 @@ class WorkletsModule : public jni::HybridClass<WorkletsModule> {
   }
 
  private:
-  friend HybridBase;
-  jni::global_ref<WorkletsModule::javaobject> javaPart_;
-  jsi::Runtime *rnRuntime_;
-  std::shared_ptr<WorkletsModuleProxy> workletsModuleProxy_;
-
   explicit WorkletsModule(
-      jni::alias_ref<WorkletsModule::jhybridobject> jThis,
```
tjzel, line 55: Turns out we don't need to keep the reference to Java part of `WorkletsModule` in cpp.

packages/react-native-reanimated/Common/cpp/worklets/Tools/SingleInstanceChecker.h
```diff
@@ -52,10 +52,8 @@ SingleInstanceChecker<T>::SingleInstanceChecker() {
   std::string className =
       __cxxabiv1::__cxa_demangle(typeid(T).name(), nullptr, nullptr, &status);
 
-  // Only one instance should exist, but it is possible for two instances
-  // to co-exist during a reload.
   assertWithMessage(
-      instanceCount_ <= 1,
+      instanceCount_ < 1,
```
tjzel, line 56: When cleaning during invalidation stage, it shouldn't be possible for two instances to exist at the same time.

tomekzaw: Why not just `instanceCount_ == 0`?

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.cpp
```diff
@@ -52,7 +50,6 @@ jni::local_ref<WorkletsModule::jhybriddata> WorkletsModule::initHybrid(
       std::make_shared<worklets::JSScheduler>(*rnRuntime, jsCallInvoker);
   auto uiScheduler = androidUIScheduler->cthis()->getUIScheduler();
   return makeCxxInstance(
-      jThis,
       rnRuntime,
```
tomekzaw, line 53: This change looks weird, do we really need `WorkletsModule` to be a Hybrid class if we don't use `jThis` here in `makeCxxInstance`?

packages/react-native-reanimated/apple/worklets/apple/WorkletsModule.mm
```diff
@@ -52,4 +52,13 @@ @implementation WorkletsModule {
   return @YES;
 }
 
+- (void)invalidate
+{
+  // We have to destroy extra runtimes when invalidate is called. If we clean
+  // it up later instead there's a chance the runtime will retain references
+  // to invalidated memory and will crash on destruction.
+  workletsModuleProxy_.reset();
+  [super invalidate];
```
tomekzaw, line 61: Let's add a newline here since `[super invalidate];` is unrelated to the comment above.
```suggestion
  // We have to destroy extra runtimes when invalidate is called. If we clean
  // it up later instead there's a chance the runtime will retain references
  // to invalidated memory and will crash on destruction.
  workletsModuleProxy_.reset();

  [super invalidate];
```

packages/react-native-reanimated/android/src/main/java/com/swmansion/worklets/WorkletsModule.java
```diff
@@ -69,6 +69,12 @@ public boolean installTurboModule(String valueUnpackerCode) {
   }
 
   public void invalidate() {
+    // We have to destroy extra runtimes when invalidate is called. If we clean
+    // it up later instead there's a chance the runtime will retain references
+    // to invalidated memory and will crash on its destruction.
+    invalidateCpp();
     mAndroidUIScheduler.deactivate();
```
tomekzaw, line 76: Let's add a newline here as well
```suggestion
    // We have to destroy extra runtimes when invalidate is called. If we clean
    // it up later instead there's a chance the runtime will retain references
    // to invalidated memory and will crash on its destruction.
    invalidateCpp();

    mAndroidUIScheduler.deactivate();
```

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.h
```diff
@@ -46,19 +46,19 @@ class WorkletsModule : public jni::HybridClass<WorkletsModule> {
...
       jsi::Runtime *rnRuntime,
```
tomekzaw, line 50: If possible, let's pass the runtime as a reference instead of a pointer, i.e. `jsi::Runtime &`.

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.h
```diff
@@ -46,19 +46,19 @@ class WorkletsModule : public jni::HybridClass<WorkletsModule> {
...
       const std::string &valueUnpackerCode,
       jni::alias_ref<JavaMessageQueueThread::javaobject> messageQueueThread,
       const std::shared_ptr<facebook::react::CallInvoker> &jsCallInvoker,
       const std::shared_ptr<worklets::JSScheduler> &jsScheduler,
       const std::shared_ptr<UIScheduler> &uiScheduler);
+
+  void invalidateCpp();
+
+  friend HybridBase;
+  jsi::Runtime *rnRuntime_;
```
tomekzaw, line 60: If possible, let's store runtime as a reference.
```suggestion
  jsi::Runtime &rnRuntime_;
```

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.h
tomekzaw: Does it still have to be a `HybridClass` or can we convert it back to `JavaClass` or something?

tomekzaw (approved):
The changes look good to me, I left some remarks in the comments. Two more general questions:
1. Why do we need a separate `invalidate` method? Why can't we just clear the pointers in C++ destructors like we used to do?
2. Once `invalidate` method is called, shall we add some assertions to prevent us from calling any methods of an invalidated object?

piaskowyk (commented):
Could you elaborate more on why performing cleanup in the C++ destructor is considered too late? This isn't entirely clear to me, but it seems to be important.

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.h
tjzel: Probably it doesn't have to be a HybridClass at the moment. However, let's change it in another PR if bothers us.

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.h
tjzel: This is a mirror of how `NativeProxy` stores it. Let's change it in a different PR if we want to.

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.cpp
tjzel: Probably it will need to be a Hybrid class in the future.

tjzel:
@piaskowyk Native Modules are invalidated and destroyed in stages.

This is more or less how React Native gets cleaned up now (e.g. on reload):
1. `invalidate()` is invoked on all its Native Modules (probably in reverse order of their initialization, but I'm not exactly sure there).
2. React Native instance specific objects, like Shadow Nodes are deleted.
3. Native Modules are released (and deleted).

Our Worklet Runtime can hold instances to various Host Objects, like Shadow Nodes. Therefore, if we destroy our runtime on step 3. it might crash on garbage collection because the C++ side of these host objects is already destroyed. This probably could be mitigated with a different kind of bindings, but it's definitely outside of the scope of this PR.

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.h
tjzel: https://github.com/software-mansion/react-native-reanimated/pull/6850#discussion_r1908596428

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.h
tomekzaw: Nobody said this was the right way. But okay, let's do that in another PR.

packages/react-native-reanimated/android/src/main/cpp/worklets/android/WorkletsModule.cpp
tomekzaw: So why do we remove `jThis` now? We'll have to add it in the future anyway.

piaskowyk (approved):

PR #6875, opened 2025-01-08
Remove `CoreFeatures.h` include

tomekzaw:
## Summary

This PR removes `#include <react/utils/CoreFeatures.h>` in ReanimatedModuleProxy.cpp which breaks build on RN 0.77.

See also https://github.com/facebook/react-native/pull/45626


piaskowyk (approved):
Looks good, but let's wait until the CI checks for backward compatibility. Once that's confirmed, we can merge

PR #6878, opened 2025-01-08
Use `TurboModuleRegistry.getEnforcing` instead of `get`

tomekzaw:
## Summary

This PR migrates all usages of `TurboModuleRegistry.get` to `TurboModuleRegistry.getEnforcing` which throws an error if the module is not found.

### DIFF
--- a/apps/common-app/src/examples/ShareableFreezingExample.tsx
+++ b/apps/common-app/src/examples/ShareableFreezingExample.tsx
@@ -93,11 +93,7 @@ function tryModifyConvertedRemoteFunction() {
 }
 
 function tryModifyConvertedHostObject() {
-  const obj = TurboModuleRegistry.get('Clipboard');
-  if (!obj) {
-    console.warn('No host object found.');
-    return;
-  }
+  const obj = TurboModuleRegistry.getEnforcing('Clipboard');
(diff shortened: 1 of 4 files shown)


tomekzaw:
cc @tjzel Perhaps we need to add a mock for `WorkletsModule`?

tjzel (approved):

tjzel:
Oh, I think you actually meant the breaking CI. This is not worklets fault - native modules aren't loaded in the Jest environment (same error goes for Reanimated's Turbo Module) while on the other hand Jest naively resolves all imported files, even if they aren't actually used. I looked briefly for a nice solution and got that you need to add the following to `jest.config.js`:

```js
  moduleNameMapper: {
    reanimatedModuleInstance:
      '<rootDir>/src/ReanimatedModule/reanimatedModuleInstance.web.ts',
    workletsModuleInstance:
      '<rootDir>/src/worklets/WorkletsModule/workletsModuleInstance.web.ts',
    JSPropsUpdater:
      '<rootDir>/src/createAnimatedComponent/JSPropsUpdater.web.ts',
  },
```

However, I'm afraid that our users would also need to apply a derivative of this fix, which is kinda bad.

PR #6880, opened 2025-01-09
fix: __reanimatedLoggerConfig is not defined on edge runtime

lebleb:
## Summary

If you run NextJs application in edge runtime, library throws an error

`__reanimatedLoggerConfig is not defined`

we needs to declare global or globalThis on every usage in workers on edge runtime.

## Collaboration
**It was found and made in collaboration with @vault-developer** 

## Test plan

Tested with patch-package and it worked


MatiPl01 (approved):
LGTM. If this change fixes the issue, we can merge that PR.

PR #6892, opened 2025-01-13
Update versions in compatibility check workflow

tomekzaw:
## Summary

This PR adds React Native 0.77 for Paper and Fabric to the test matrix in compatibility check workflow for GitHub Actions.

### DIFF
--- a/.github/workflows/reanimated-compatibility-check-nightly.yml
+++ b/.github/workflows/reanimated-compatibility-check-nightly.yml
@@ -25,6 +25,8 @@ jobs:
             { version: '0.75', architecture: 'Fabric' },
             { version: '0.76', architecture: 'Paper' },
             { version: '0.76', architecture: 'Fabric' },
+            { version: '0.77', architecture: 'Paper' },
+            { version: '0.77', architecture: 'Fabric' },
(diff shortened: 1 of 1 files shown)


.github/workflows/reanimated-compatibility-check-nightly.yml
```diff
@@ -25,6 +25,8 @@ jobs:
             { version: '0.75', architecture: 'Fabric' },
             { version: '0.76', architecture: 'Paper' },
             { version: '0.76', architecture: 'Fabric' },
+            { version: '0.77', architecture: 'Paper' },
+            { version: '0.77', architecture: 'Fabric' },
```
piaskowyk, line 29: Do we want to remove checks for version 0.74 in order to align with our policy of supporting the last three versions?

tomekzaw: Sounds right

tomekzaw: Let's keep for support RN 0.74 for now since we're already dropping support RN 0.73 in Reanimated 3.17.

PR #6893, opened 2025-01-13
fix: Crash on UI Scheduling with invalidated WorkletsModuleProxy

tjzel:
## Summary

When React Native is destroyed on reload, it runs the destructors for Native Modules on the JavaScript thread. Therefore, even though React Native instance is being destroyed, UI thread can still perform operations on the UI runtime. This leads to a crash where `scheduleOnUI` is called from the UI thread when `WorkletsModuleProxy` destructor has already released `uiWorkletRuntime`.

To fix this, I'm adding a simple shared mutex. `WorkletsModuleProxy` can be cleaned up only when no scheduling takes place and scheduling cannot be performed after `WorkletsModuleProxy` was invalidated.

## Test plan

Try reloading Android now and see that it doesn't crash on `scheduleOnUI` anymore.


tjzel:
Closed in favor of 
- https://github.com/software-mansion/react-native-reanimated/pull/6900

Issue #4472, opened 2023-05-18
[Android 11] Production crash - Attempt to invoke virtual method 'java.lang.Object java.lang.ref.WeakReference.get()' on a null object reference


a-vikor:
### Description

Dear Software Mansion, Reader,

In Google's Play Console I got a crash report listing `Exception java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object java.lang.ref.WeakReference.get()' on a null object reference`.

![Screenshot 2023-05-18 at 13 06 02](https://github.com/software-mansion/react-native-reanimated/assets/115455866/e963e98e-e011-44e3-9968-8be0385f43ae)

Previously there was a [same issue opened](https://github.com/software-mansion/react-native-reanimated/issues/2369), but back in the days developers used to forget some installation steps. Since then, Android installation steps were automatised, so I don't think it is due my setup actions. 
My `babel.config.js` can be found here:

![Screenshot 2023-05-18 at 13 16 29](https://github.com/software-mansion/react-native-reanimated/assets/115455866/fe3c4587-85d8-4ddb-85ce-9190de7ccd99)

Thanks in advance!

### Steps to reproduce

I don't have any reproduction steps as an end user encountered the issue inside a chat application (where inverted FlatList has itemLayoutAnimation={Layout.mass(100)} property set).

### Snack or a link to a repository

Can't provide repository. Setup is similar like in following repository: https://github.com/passengerV/chat-issues

### Reanimated version

3.0.2

### React Native version

0.71.4

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

Real device

### Device model

Redmi begonia (Redmi Note 8 Pro)

### Acknowledgements

Yes

SiSa68:
@a-vikor Same here
Were you able to find a solution?

a-vikor:
@SiSa68 I haven't went for investigation as it was an isolated case. If it repeats, I will have definitely.

Latropos:
I am closing this ticket, since the problem isn't reproducible. If you ever see this happening again feel free to reopen it

matejkriz:
Could you please reopen the issue? It's number one in crashes reported from our latest version. So far we don't have isolated example, our repo is https://github.com/trezor/trezor-suite/tree/develop/suite-native/app

"react-native-reanimated": "3.16.1"

PR #6897, opened 2025-01-14
fix: layout animation NPE crash

Nodonisko:
## Summary

After update to Reanimated 3.16.x we started to have hundreds of crashes due to this error:
```
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object java.lang.ref.WeakReference.get()' on a null object reference.
```
More info could be found in this issue #4472 

## Test plan

I am not able to reproduce issue locally since it's happening for some small percentage of users. According to stack trace I am pretty sure it will help but I am not sure if it's proper fix because I don't have deeper understanding of surrounding code.


piaskowyk (approved):
Thanks for your help! I made a few adjustments to your approach, but your concept was spot on 👍

PR #6900, opened 2025-01-14
fix: Memory mismanagement with UI scheduled callbacks

tjzel:
### What
The pattern we used with scheduling callbacks on the UI thread - capturing `this` as a reference - is prone to memory issues. Given such code:

```c++
void WorkletsModuleProxy::scheduleOnUI(
    jsi::Runtime &rt,
    const jsi::Value &worklet) {
  auto shareableWorklet = extractShareableOrThrow<ShareableWorklet>(
      rt, worklet, "[Worklets] Only worklets can be scheduled to run on UI.");
  uiScheduler_->scheduleOnUI(
      [this, shareableWorklet] {
        this->uiWorkletRuntime_->runGuarded(shareableWorklet);
      });
}
```

We are likely to run into accessing invalidated memory during a reload. This is due to fact that `WorkletsModuleProxy` is managed by some object held by the instance of React Native. Let's look at the following scenario.

1. `WorkletsModuleProxy` is created on the JS thread and held by the `WorkletsModule` Native Module.
2. `WorkletsModuleProxy::scheduleOnUI` is invoked on the JS thread. The callback is scheduled to be executed on the UI thread.
3. Application's reload gets triggered. A tear down of React Native is starting on the JS thread.
4. `WorkletsModule` gets destroyed. Therefore, `WorkletsModuleProxy` is released and also destroyed.
5. The callback is finally executed on the UI thread by the scheduler. However, `this` has been invalidated. The App crashes.

Keep in mind that this isn't exclusive to thread jumps exclusively. Calling `scheduleOnUI` on the UI thread could still result in the callback executing after the memory has been invalidated.

`WorkletsModuleProxy` is only an example here, the problem could manifest in all the places where we pass lambdas that capture `this` by reference.

### Fix
To fix this I refactored the code so everytime we pass `this` to a scheduled callback, it would be done via a _weak pointer_ which would lock the object and prevent it from being destroyed while the callback is being executed on the UI thread.

Perhaps some bits of code don't need this safety measure due to a heuristic existing that guarantees that respective memory won't be invalidated before the callback gets executed. However, I found it extremely challenging and unreliable to come up with these heuristics, as they could possibly break at any future change of the code.

### Affected code:

- ReanimatedCommitHook
- LayoutAnimationProxy
- ReanimatedModuleProxy
- WorkletsModuleProxy
- WorkletRuntime
- NativeProxy

## Test plan

Reloading the app no longer causes a crash on a scheduled UI callback.


packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
```diff
@@ -50,6 +50,7 @@ NativeProxy::NativeProxy(
           isBridgeless,
           getIsReducedMotion())),
       layoutAnimations_(std::move(layoutAnimations)) {
+  reanimatedModuleProxy_->init(getPlatformDependentMethods());
```
tjzel, line 53: `weak_from_this` can only be used after the object has been instantiated. Therefore, all callbacks that use `weak_from_this` must be created in a subsequent init method.

packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
```diff
@@ -622,12 +629,7 @@ void NativeProxy::setupLayoutAnimations() {
 
 void NativeProxy::invalidateCpp() {
   layoutAnimations_->cthis()->invalidate();
-
   workletsModuleProxy_.reset();
-
-  if (reanimatedModuleProxy_ != nullptr) {
-    reanimatedModuleProxy_->invalidate();
-  }
```
tjzel, line 630: We can't invalidate it here because relevant callbacks might still need it. Once they get executed and cleaned up the destructor will take care of things.

packages/react-native-reanimated/apple/reanimated/apple/native/NativeProxy.mm
```diff
@@ -72,6 +72,7 @@ static inline bool getIsReducedMotion()
 
   auto reanimatedModuleProxy = std::make_shared<ReanimatedModuleProxy>(
       workletsModuleProxy, rnRuntime, jsInvoker, platformDepMethodsHolder, isBridgeless, getIsReducedMotion());
+  reanimatedModuleProxy->init(platformDepMethodsHolder);
```
tjzel, line 75: weak_from_this can only be used after the object has been instantiated. Therefore, all callbacks that use weak_from_this must be created in a subsequent init method.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -34,14 +34,21 @@ void ReanimatedCommitHook::maybeInitializeLayoutAnimations(
     // when a new surfaceId is observed we call setMountingOverrideDelegate
     // for all yet unseen surfaces
     uiManager_->getShadowTreeRegistry().enumerate(
-        [this](const ShadowTree &shadowTree, bool &stop) {
-          if (shadowTree.getSurfaceId() <= currentMaxSurfaceId_) {
+        [weakReanimatedCommitHook = weak_from_this()](
```
tjzel, line 37: I couldn't decide if I prefer the syntax

`weakThis` - `strongThis`

or

`weak<object name>` - `<object name>`

Do you have any preferences?

bartlomiejbloniarz (approved):

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.h
```diff
@@ -147,6 +152,8 @@ class ReanimatedModuleProxy : public ReanimatedModuleProxySpec {
   void unregisterCSSTransition(jsi::Runtime &rt, const jsi::Value &viewTag)
       override;
 
+  void cssLoopCallback(const double /*timestampMs*/);
```
tjzel, line 155: I moved a lambda that was handling the `cssLoop` queue to a method.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -692,37 +759,52 @@ bool ReanimatedModuleProxy::handleRawEvent(
...
-          || updatesRegistryManager_->hasPropsToRevert()
-#endif
-      ) {
-        jsi::Runtime &rt =
-            workletsModuleProxy_->getUIWorkletRuntime()->getJSIRuntime();
-        requestRender_(*cssLoop, rt);
-      } else {
-        cssLoopRunning_ = false;
-      }
-    };
```
tjzel, line 720: This pattern - a lambda holding itself via a shared pointer in the closure - is a memory leak. It will never get destroyed. I fixed this by scheduling a lambda that only has `weakThis` and then executes `cssLoopCallback` if viable.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -34,14 +34,20 @@ void ReanimatedCommitHook::maybeInitializeLayoutAnimations(
     // when a new surfaceId is observed we call setMountingOverrideDelegate
     // for all yet unseen surfaces
     uiManager_->getShadowTreeRegistry().enumerate(
-        [this](const ShadowTree &shadowTree, bool &stop) {
-          if (shadowTree.getSurfaceId() <= currentMaxSurfaceId_) {
+        [weakThis = weak_from_this()](
```
tomekzaw, line 37: Do we really need to use a weak pointer here? If I remember correctly, the callback passed to `.enumerate` is executed synchronously.

MatiPl01 (approved):
LGTM

Now it builds on iOS as well, so, for me, it's good to go.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
tjzel: @bartlomiejbloniarz Enlighten us here please 🙏

bartlomiejbloniarz: That's correct! Missed that 🙈

tjzel: I will turn this into a shared_ptr capture then.

PR #6902, opened 2025-01-15
Add unhandled exception to TabNavigatorObserver catch block

szydlovsky:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/6724

Turns out that `TabNavigatorObserver's` `onFragmentUpdate` didn't handle `NullPointerException`, which in turn sometimes crashed release builds on android.

## Test plan

I wasn't able to create a simple reproduction but I have at least one SWM project that used the exact same code to patch Reanimated and it took care of their android crashes.


szydlovsky:
@tomekzaw exactly my sentiment. I'd rather not meddle in the code risking getting something broken if a simple exception handling is enough

szydlovsky:
Update: got community confirmation that it fixes the issue: https://github.com/software-mansion/react-native-reanimated/issues/6724#issuecomment-2592904009

tomekzaw (approved):

PR #6904, opened 2025-01-15
feat(plugin): multiple worklet referencing

tjzel:
## Summary

Some time ago I added a feature of worklet referencing to the Plugin, but this was limited to only a single reference:

```ts
function foo(){
	// UI stuff
};

runOnUI(foo)(); // ok, `foo` will get workletized

function bar(){
	// UI stuff
};

const foobar = bar;

runOnUI(foobar); // error, foobar didn't get workletized
```

This PR allows for any amount of references in the reference chain. Simultaneously I discovered that Gesture Handler callbacks didn't benefit from worklet referencing, i.e. the following code would be invalid:

```ts
function foo(){
	// do something
}

const pan = Gesture.Pan().onActive(foo) // `foo` wouldn't be workletized
```

This PR adds this QoL feat for Gesture Handler callbacks. 

## Test plan

I added new unit tests for these cases.


packages/react-native-reanimated/plugin/src/autoworkletization.ts
```diff
@@ -136,7 +147,7 @@ function findWorklet(
   if (acceptObject && isWorkletizableObjectPath(arg)) {
     return arg;
   }
-  if (arg.isReferencedIdentifier() && arg.isIdentifier()) {
+  if (arg.isIdentifier() && arg.isReferencedIdentifier()) {
```
tjzel, line 150: The order of these actually doesn't matter, but it's more logical that way.

bartlomiejbloniarz (approved):

PR #6905, opened 2025-01-15
Improve animated keyboard example

szydlovsky:
## Summary

Animated keyboard example has been there quite some time, rusty, old and not understood too well. Let's give it some love.

It features displaying keyboard's height and state as well as has an "accessory bar" that sticks to the keyboard using `useAniamtedKeyboard`.

Also, interactive dismiss can now be tested here as well.

https://github.com/user-attachments/assets/1589e90a-eeab-4559-800e-b28ae7d33e04

## Test plan

open `useAnimatedKeyboard` example from example app.

### DIFF
--- a/apps/common-app/src/examples/AnimatedKeyboardExample.tsx
+++ b/apps/common-app/src/examples/AnimatedKeyboardExample.tsx
@@ -1,88 +1,166 @@
+import React from 'react';
 import Animated, {
-  KeyboardState,
   useAnimatedKeyboard,
   useAnimatedStyle,
+  useAnimatedProps,
 } from 'react-native-reanimated';
 import {
-  Button,
   Keyboard,
   ScrollView,
   StyleSheet,
   TextInput,
+  TouchableOpacity,
   View,
+  Text,
 } from 'react-native';
 
-import React from 'react';
+Animated.addWhitelistedNativeProps({ text: true });
+const AnimatedTextInput = Animated.createAnimatedComponent(TextInput);
 
-const BOX_SIZE = 50;
+const KeyboardStateToName = {
+  0: 'UNKNOWN',
+  1: 'OPENING',
+  2: 'OPEN',
+  3: 'CLOSING',
+  4: 'CLOSED',
+};
 
-function NestedView() {
-  useAnimatedKeyboard();
-  return <View style={styles.nestedView} />;
+function getRandomColor() {
+  const letters = '0123456789ABCDEF';
+  let color = '#';
+  for (let i = 0; i < 6; i++) {
+    color += letters[Math.floor(Math.random() * 16)];
+  }
+  return color;
+}
+
+function RandomView({ number }: { number: number }) {
+  const color = getRandomColor();
+  return (
+    <View style={[styles.randomView, { backgroundColor: color }]}>
+      <Text>Random view {number}</Text>
+    </View>
+  );
 }
 
 export default function AnimatedKeyboardExample() {
   const keyboard = useAnimatedKeyboard();
-  const OPENING = KeyboardState.OPENING;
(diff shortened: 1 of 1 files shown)


apps/common-app/src/examples/AnimatedKeyboardExample.tsx
```diff
@@ -1,88 +1,168 @@
...
+const AnimatedTextInput = Animated.createAnimatedComponent(TextInput);
 
-const BOX_SIZE = 50;
+const KeyboardStateToName = {
+  0: 'UNKNOWN',
+  1: 'OPENING',
+  2: 'OPEN',
+  3: 'CLOSING',
+  4: 'CLOSED',
+};
```
tjzel, line 26: Let's use `enum` here and probably change name to `KeyboardState`, i.e.:
```suggestion
enum KeyboardState = {
  0: 'UNKNOWN',
  1: 'OPENING',
  2: 'OPEN',
  3: 'CLOSING',
  4: 'CLOSED',
};
```

apps/common-app/src/examples/AnimatedKeyboardExample.tsx
```diff
@@ -1,88 +1,168 @@
...
-  useAnimatedKeyboard();
-  return <View style={styles.nestedView} />;
+function getRandomColor() {
+  const letters = '0123456789ABCDEF';
+  let color = '#';
+  for (let i = 0; i < 6; i++) {
+    color += letters[Math.floor(Math.random() * 16)];
+  }
+  return color;
+}
```
tjzel, line 35: That's a pretty interesting way to generate RGB colors 😊

apps/common-app/src/examples/AnimatedKeyboardExample.tsx
```diff
@@ -1,88 +1,168 @@
...
+
+function RandomView({ number }: { number: number }) {
```
tjzel, line 37: Using `number` as a variable name is a bad idea.

```suggestion
function RandomView({ id }: { id: number }) {
```

apps/common-app/src/examples/AnimatedKeyboardExample.tsx
```diff
@@ -1,88 +1,168 @@
...
+  const color = getRandomColor();
+  return (
+    <View style={[styles.randomView, { backgroundColor: color }]}>
+      <Text>Random view {number}</Text>
```
tjzel, line 41:
```suggestion
      <Text>Random view {id}</Text>
```

tjzel (approved):
Looks good overall, I have a few questions:

1. Does it look ok on Android too?
2. Can you change the background (or do something relevant) where the "press me!!!" text is when you press it?

apps/common-app/src/examples/AnimatedKeyboardExample.tsx
szydlovsky: cause why not 😄

apps/common-app/src/examples/AnimatedKeyboardExample.tsx
```diff
@@ -1,88 +1,168 @@
...
+        {Array.from({ length: 10 }).map((_, index) => {
+          return <RandomView key={index} number={index} />;
+        })}
+      </ScrollView>
+      <Animated.View style={[styles.accessoryBar, translateStyle]}>
+        <View style={styles.inputContainer}>
+          <TextInput
+            style={styles.textInput}
+            autoCorrect
+            defaultValue="press me!!!"
```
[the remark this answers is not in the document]
szydlovsky: that's what format document does here.. it works in its mysterious ways

szydlovsky:
@tjzel confirmed it works fine on android & addressed all the comments!

PR #6921, opened 2025-01-22
feat(Worklets): JS code duplication

tjzel:
## Summary

In this PR I'm moving the `WorkletsModule` part of the JS code to `react-native-worklets`. Reanimated would still have those files duplicated via a mechanism introduced in #6827.
Now in Reanimated all worklets code must be acquired via a special `WorkletsResolver` module. This is a temporary solution and will be dropped once we require `react-native-worklets` as a mandatory dependency for `react-native-reanimated`.

- Added a workaround so that in our monorepo `react-native-worklets` wouldn't be used in the app even though it's not a dependency.
- Added an ESLint rule that enforces imports from `WorkletsResolver`.
- I duplicated `PlatformChecker.ts` explicitly in `react-native-worklets`. This file is small and stable so I don't consider this to be an issue. We can think of some better solution once the package is more polished.
- Split some types from `commonTypes.ts` between packages.

## Test plan

Fabric example should work with & without `react-native-worklets` installed. You can check using the debugger that always the correct instance of Worklets' JavaScript code is used.


.github/workflows/example-android-build-check.yml
```diff
@@ -69,11 +69,6 @@ jobs:
         working-directory: ${{ env.COMMON_APP_DIR }}
         run: yarn add react-native-worklets@workspace:"*"
 
-      - name: Use external worklets
-        if: ${{ inputs.use-external-worklets == 'true' }}
-        working-directory: ${{ env.COMMON_APP_DIR }}
-        run: yarn add react-native-worklets@workspace:"*"
-
```
tjzel, line 76: I don't understand how things keep duplicating when I rebase 😭

bartlomiejbloniarz: :(((

bartlomiejbloniarz (approved):

PR #6923, opened 2025-01-22
chore: Clean up dependencies of apps dependent on the common-app

MatiPl01:
## Summary

This PR moves common-app dependencies to the common-app `package.json` and removes all of duplicate dependencies from `fabric-example`, `web-example` and `macos-example` apps. We can do this thanks to `react-native.config.js` which reads all dependencies from the `common-app/package.json` and adds them to the dependency list from the current app's `package.json` file (if they weren't specified in the app's `package.json`).

Some dependencies, which are used in `package.json` scripts to e.g. start the app, such as `react-native`, must be added explicitly in each app's `package.json`.

### Motivation

I think it is a bit cumbersome to repeat dependencies needed by the common-app in every app that imports application from the common-app sources. Since every app would need dependencies that common-app needs to work, we don't have to repeat them in every app dependent on the common-app.

If we will have to use a different dependency version or mock a dependency, we can always add another dep e.g. in the fabric-example `package.json` or exclude/include it in the `react-native.config.js`.

## Test plan

Build the `fabric-example` / `web-example` / `macos-example` to see that everything still works.


apps/common-app/package.json
```diff
@@ -10,41 +10,51 @@
...
-    "react-native": "*",
-    "react-native-gesture-handler": "*",
-    "react-native-pager-view": "*",
-    "react-native-reanimated": "*",
-    "react-native-safe-area-context": "*",
-    "react-native-screens": "*",
-    "react-native-svg": "*",
-    "react-native-web": "*",
-    "react-strict-dom": "*"
+    "react-native": "*"
```
MatiPl01, line 13: Only `react-native` must be installed in each app separately to make it possible to run `react-native` scripts using `yarn` (scripts from the `package.json`)

apps/common-app/src/apps/reanimated/examples/WithoutBabelPluginExample.tsx
```diff
@@ -4,16 +4,10 @@ import Animated, {
...
-import {
-  Gesture,
-  GestureDetector,
-  enableExperimentalWebImplementation,
-} from 'react-native-gesture-handler';
+import { Gesture, GestureDetector } from 'react-native-gesture-handler';
 import React, { useEffect, useState } from 'react';
 import { StyleSheet, Text, View } from 'react-native';
 
-enableExperimentalWebImplementation(true);
```
MatiPl01, line 15: This was showing gesture handler related warning because it is not needed anymore.

apps/fabric-example/package.json
```diff
@@ -6,66 +6,11 @@
     "build": "cd ios && bundle install && bundle exec pod update",
     "android": "react-native run-android",
     "ios": "react-native run-ios",
-    "lint": "eslint --max-warnings=0 --ignore-pattern 'common-app/' .",
-    "format": "prettier --write --list-different .",
-    "start": "react-native start",
-    "test": "jest"
```
MatiPl01, line 12: No need to lint, format and test `fabric-example` and other apps that depend on the `common-app` as the only thing they do is importing the app exported from the `common-app`. We can run these checks just in the `common-app` because sources of the example app are only in this directory.

apps/fabric-example/react-native.config.js
```diff
@@ -1,8 +1,20 @@
...
+
+const commonAppDir = path.resolve(__dirname, '../common-app');
+const commonAppPkg = require(path.resolve(commonAppDir, 'package.json'));
+
+const dependencies = Object.fromEntries([
+  // Include all common-app dependencies
+  ...Object.keys(commonAppPkg.devDependencies || {}),
+  ...Object.keys(commonAppPkg.dependencies || {})
+].map(name => [name, {
+  root: path.resolve(__dirname, `../../node_modules/${name}`)
```
MatiPl01, line 14: We have to point all dependencies listed in the `package.json` of the `common-app` to their sources location. It works, because all dependencies are hoisted to the root directory.

.github/workflows/example-typescript-check-and-lint.yml
```diff
@@ -47,12 +50,8 @@ jobs:
 
       - name: Check types
         working-directory: ${{ matrix.working-directory }}
-        if: ${{ matrix.working-directory != 'apps/web-example' }}
+        if: ${{ contains(env.TYPE_CHECK_DIRECTORIES, matrix.working-directory) }}
         run: yarn tsc --noEmit
-      - name: Check types
-        working-directory: ${{ matrix.working-directory }}
-        if: ${{ matrix.working-directory == 'apps/web-example' }}
```
MatiPl01, line 54: I don't think we need to run typecheck on directories that just import the entire app sources from the `common-app`. It just adds extra time on CI and, in the end, we are just running typecheck on the same typescript files for a few times.

I also had to remove this after removing `typescript` from `package.json` in apps that import `App` from the `common-app`. If we really need to use it, I can add `typescript` dependency in `package.json` in every app instead of adding it with a `react-native.config.js` script.

.github/workflows/example-typescript-check-and-lint.yml
```diff
@@ -1,6 +1,9 @@
 name: Example Typescript check and lint
 env:
   YARN_ENABLE_HARDENED_MODE: 0
+  TYPE_CHECK_DIRECTORIES: >-
+    apps/common-app
+    apps/tvos-example
```
MatiPl01, line 6: All remaining apps just import the `common-app` so there is no point in running typescript and eslint checks on them. It will just unnecessarily make the CI run longer.

apps/fabric-example/package.json
tjzel: Let's keep `format` as it formats `.js` config files.

MatiPl01:
> I like the idea but I'm not certain if this is going to work with `macos-example` (and possible `tvos-example` in the future) since they might need to be behind in versions etc.
> 
> Does it work with `macos-example`?

I updated the `macos-example` as well. I also implemented your suggestions and added a possibility to override dependencies specified in `common-app` - this was needed for the `macos-example` where we use different version of `react-native`.

The final implementation works in this way:
- if the dependency is not specified in the `package.json` of the current example app, the dependency from the `common-app` will be used instead,
- if the same dependency is specified in the current app and the `common-app`, the current app dep has higher priority and this dependency will override the one specified in the `common-app`

.github/workflows/example-macos-build-check.yml
```diff
@@ -48,7 +48,9 @@ jobs:
 
       - name: Install Pods
         working-directory: ${{ env.WORKING_DIRECTORY }}/macos
-        run: bundle install && bundle exec pod update # There's some bug on the macOS runners for pod installation and this is the workaround.
+        run: |
+          bundle install
+          bundle exec pod install
```
MatiPl01, line 53: Neither works on CI, both work locally. I don't know what is going on here.

apps/common-app/.eslintignore
```diff
@@ -0,0 +1 @@
+scripts
```
tjzel, line 1: Why do you need to ignore that?

tjzel (approved):
Looks good now!

apps/common-app/.eslintignore
MatiPl01: Fixed here: https://github.com/software-mansion/react-native-reanimated/pull/6923/commits/561576141831567ceb6d85eeda21353ea259d8c8

I also don't know why eslint paths plugin patch was still here. I remember removing it before but maybe I did something wrong.

PR #6924, opened 2025-01-22
chore: Remove paper example app

MatiPl01:
## Summary

This PR removes no longer supported `paper-example`.

Docs still contain a few mentions to `paper-example` that probably have to be updated as well. Maybe @kacperkapusciak could check if these should be removed/replaced.


tjzel (approved):
You've been a great friend, `paper-example`. It's time to rest champ 🫡

PR #6931, opened 2025-01-23
docs: update Reanimated 4 docs

kacperkapusciak:
This PR updates Reanimated 4 documentation:

- updates installation steps in Getting Started
- removes `paper-example` mentions from Contributing
- documents `textShadow` style props behavior


packages/docs-reanimated/docs/fundamentals/getting-started.mdx
```diff
@@ -20,54 +20,51 @@ With Reanimated, you can easily create smooth animations and interactions that r
 Reanimated 4.x works only with [the React Native New Architecture (Fabric)](https://reactnative.dev/architecture/landing-page).
 If your app still uses the old architecture, you can use Reanimated in version 3 which is still actively maintained.
 
-## Quick start
+Alternatively, you can dive into [our examples](https://github.com/software-mansion/react-native-reanimated/tree/3.14.0/apps/common-app/src/examples) on GitHub.
+
+## Installation
 
-If you don't have an existing project, you can create a new Expo app using a template:
+It takes two steps to add Reanimated 4 to an Expo project:
```
MatiPl01, line 27: Do we really need to mention here `Expo project`? I think that many projects still use bare react native template and it might be confusing for users because Expo is not mandatory.

Even though you mention `react-native-cli` later on this page, I feel that some people may get confused because of this explicit `Expo` mention at the very beginning of the installation instruction.

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
```diff
@@ -92,66 +89,12 @@ Add `react-native-reanimated/plugin` plugin to your `babel.config.js`.
...
+Since [Expo SDK 50](https://expo.dev/changelog/2024/01-18-sdk-50), the Expo starter template includes the Reanimated babel plugin by default.
 
-Wrapping your Metro configuration with the Reanimated Metro config wrapper will result in displaying improved reanimated errors and warnings with more accurate call stacks. Thanks to this, identifying misuses of the Reanimated API will be much easier than before.
-
-To learn more about this feature, head onto to [Accurate Call Stacks](/docs/debugging/accurate-call-stacks).
+To learn more about the plugin head onto to [Reanimated babel plugin](/docs/fundamentals/glossary#reanimated-babel-plugin) section.
 
 </details>
 
-### Step 4: Clear Metro bundler cache (recommended)
```
MatiPl01, line 125: Maybe leave it as a recommended step for `react-native-cli` users?

packages/docs-reanimated/docs/guides/contributing.mdx
```diff
@@ -215,7 +215,7 @@ We have shared code of example app, that is run on different platforms and archi
 - `packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated` &ndash; containing source code related to Android native part of Reanimated features
 - `packages/react-native-reanimated/apple` &ndash; containing source code related to iOS native part of Reanimated features
 
-Our workflow usually starts from editing `apps/common-app/src/examples/EmptyExample.tsx` with testing example for our new feature or a bug. This is the part that we often attach to PR so other developers can test the same use-case (remember not to commit your changes from `EmptyExample` file!). Next, we run the `apps/paper-example` (or `apps/fabric-example` depending on the architecture) app and start development.
+Our workflow usually starts from editing `apps/common-app/src/apps/reanimated/examples/EmptyExample.tsx` with testing example for our new feature or a bug. This is the part that we often attach to PR so other developers can test the same use-case (remember not to commit your changes from `EmptyExample` file!). Next, we run the `apps/fabric-example` app and start development.
```
MatiPl01, line 218: Don't we want to mention the `css` example app here as well? It doesn't have to be added now but I think it will be worth noting that there are 2 apps for 2 different purposes, one for new CSS features and another one for old reanimated features known from v3

packages/docs-reanimated/docs/guides/supported-properties.mdx
```diff
@@ -182,3 +182,47 @@ Yoga calculates borders in different ways for numeric values and relative (%) va
...
+
+```jsx
+<Animated.Text
+  style={{
+    // highlight-start
+    animationName: {
+      from: { textShadowColor: 'red' },
+      to: { textShadowColor: 'blue' },
+    },
+    textShadowOffset: { width: 10, height: 10 }, // 🚨 needs to be specified
```
MatiPl01, line 198: It needs to be specified only if the users wants to apply non default `shadowOffset` (different than `{ width: 0, height: 0 }`). In fact, it doesn't have to be specified if e.g. the only used props are: `shadowRadius` and `shadowColor`, even if they are animated. The `textShadow` property builder will provide default values for the remaining props.

I just wanted you to somehow convey that all shadow props that the user wants to apply to the view must be specified in animation keyframes (so, there cannot be a prop related to the `view` or `text` shadow that is specified only in the style object). It necessarily mean that `textShadowOffset` or other shadow prop is mandatory.

packages/docs-reanimated/docs/guides/supported-properties.mdx
```diff
@@ -182,3 +182,47 @@ Yoga calculates borders in different ways for numeric values and relative (%) va
...
+<Animated.Text
+  style={{
+    animationName: {
+      from: {
+        textShadowColor: 'red',
+        textShadowOffset: { width: 10, height: 10 }, // ✅
+        textShadowRadius: 8, // ✅
+      },
+      to: {
+        textShadowColor: 'blue',
```
MatiPl01, line 221: This example is fine if someone wants to animate color from `red` to `blue`, `textShadowOffset` to `{ width: 0, height: 0 }` and `textShadowRadius` to `0`. Because the `to` keyframe doesn't contain values for `textShadowOffset` and `textShadowRadius`, default values for these props will be used.

I would prefer the correct usage example to look like this:

```js
from: {
  textShadowColor: 'red',
  textShadowOffset: { width: 10, height: 10 },
},
to: {
  textShadowColor: 'blue', // Color is animated
  textShadowOffset: { width: 10, height: 10 }, // textShadowColor must be specified to preserve the same offset, even if it is not animated
}
```

Comments in the example above could be improved

packages/docs-reanimated/docs/guides/supported-properties.mdx
```diff
@@ -182,3 +182,47 @@ Yoga calculates borders in different ways for numeric values and relative (%) va
 ### flexBasis
 
 Even though changes of this property are calculated properly during the animation, they aren't applied to the view. Other flex properties, such as `flexGrow` and `flexShrink` work fine, so they could be used for animations.
+
+### textShadow
```
MatiPl01, line 186:
```suggestion
### text-shadow
```

Maybe move this section to web related docs as all these differences apply only to the web. I would also prefer to name this prop as it is named in CSS, using the kebab case. There is no `textShadow` prop in RN so it might be confusing.

MatiPl01 (changes requested):
A few remarks here, mostly related to the `text-shadow` documentation.

I would like you to add similar docs for the `box-shadow` property. Also, I would prefer to mark these docs as web-only because iOS works without any caveats and Android doesn't work at all with most of these props.

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
kacperkapusciak: I'm following the official React Native documentation here which doesn't even mention react-native-cli unless you look for it https://reactnative.dev/docs/environment-setup

MatiPl01: Ok, fine. Maybe I am just too used to RN cli. Let's leave it as it is.

packages/docs-reanimated/docs/guides/contributing.mdx
kacperkapusciak: Fixed in [77096c5](https://github.com/software-mansion/react-native-reanimated/pull/6931/commits/77096c5fdb7567e9863e2125d6d0c0919a1a886c)

packages/docs-reanimated/docs/fundamentals/getting-started.mdx
kacperkapusciak: Fixed in [f53dae6](https://github.com/software-mansion/react-native-reanimated/pull/6931/commits/f53dae61dca3b403f3a0c71cb0bc74e9c2c4985f)

packages/docs-reanimated/docs/guides/supported-properties.mdx
kacperkapusciak: Thank you! Could you look again after the changes in [65234cc](https://github.com/software-mansion/react-native-reanimated/pull/6931/commits/65234cc9ed7c99cf495c205f1ff9f82d949d49b5)?

kacperkapusciak:
Documented box shadow styles as well

packages/docs-reanimated/docs/guides/contributing.mdx
```diff
@@ -215,7 +215,19 @@ We have shared code of example app, that is run on different platforms and archi
...
+apps/common-app/src/apps/css/examples/animations/screens/testExamples/Playground.tsx
+```
+
+For other features like worklet API, Layout Animations and Shared Element Transitions we would usually start from editing
+
+```
+apps/common-app/src/apps/reanimated/examples/EmptyExample.tsx
+```
+
+This is the part that we often attach to PR so other developers can test the same use-case (remember not to commit your changes from `EmptyExample` file!). Next, we run the `apps/fabric-example` app and start development.
```
MatiPl01, line 230:
```suggestion
This is the part that we often attach to PR so other developers can test the same use-case (remember not to commit your changes from `Playground`/`EmptyExample` file!). Next, we run the `apps/fabric-example` app and start development.
```

packages/docs-reanimated/docs/guides/contributing.mdx
```diff
@@ -215,7 +215,19 @@ We have shared code of example app, that is run on different platforms and archi
 - `packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated` &ndash; containing source code related to Android native part of Reanimated features
 - `packages/react-native-reanimated/apple` &ndash; containing source code related to iOS native part of Reanimated features
 
-Our workflow usually starts from editing `apps/common-app/src/examples/EmptyExample.tsx` with testing example for our new feature or a bug. This is the part that we often attach to PR so other developers can test the same use-case (remember not to commit your changes from `EmptyExample` file!). Next, we run the `apps/paper-example` (or `apps/fabric-example` depending on the architecture) app and start development.
+We use two separate example apps depending on the feature or a bug we're currently working on. If the code relates to CSS Animations then we start from editing
+
+```
+apps/common-app/src/apps/css/examples/animations/screens/testExamples/Playground.tsx
```
MatiPl01, line 221: There is also a separate `apps/common-app/src/apps/css/examples/transitions/screens/testExamples/Playground.tsx` but I think it's fine to mention just this one for animations.

Since we can write the same code in any of the `Playground` files or even in the `EmptyExample`, we may need to remove these playground/empty examples from all apps and create just one for all of them.

I think that it's fine to leave docs in the current state for now and we may want to adjust them after modifying example apps.

packages/docs-reanimated/docs/guides/supported-properties.mdx
```diff
@@ -182,3 +182,72 @@ Yoga calculates borders in different ways for numeric values and relative (%) va
...
+    animationName: {
+      from: {
+        textShadowRadius: 8, // while the shadow radius correctly animates from 8 ✅
+      },
+      to: {
+        textShadowRadius: 16, // to 16 ✅
+      },
+    },
+    textShadowColor: 'red', // 🚨 the shadow color would be overridden in the animation keyframes
+    // that means the shadow would be black!
```
MatiPl01, line 206: Nice! I like this example

packages/docs-reanimated/docs/guides/supported-properties.mdx
```diff
@@ -182,3 +182,72 @@ Yoga calculates borders in different ways for numeric values and relative (%) va
...
+```
+
+<details>
+<summary>Why do I need this?</summary>
+
+On the web, the shadow styles aren't inherited from the style object provided to the element. Reanimated provides the default values to every keyframe definition:
+
+```json
+{
+  "textShadowColor": "none",
```
MatiPl01, line 239:
```suggestion
  "textShadowColor": "#000",
```

packages/docs-reanimated/docs/guides/supported-properties.mdx
```diff
@@ -182,3 +182,72 @@ Yoga calculates borders in different ways for numeric values and relative (%) va
...
+    },
+  }}>
+  Reanimated
+</Animated.Text>
+```
+
+<details>
+<summary>Why do I need this?</summary>
+
+On the web, the shadow styles aren't inherited from the style object provided to the element. Reanimated provides the default values to every keyframe definition:
```
MatiPl01, line 235: It sounds a little bit like Reanimated was setting these values and, because of that, the user has to override them. I don't know how detailed explanation we would like to include in docs but I think it'd better to say something about the web `text-shadow` property instead and mention that it requires all shadow properties to be specified at once because there is just a single `text-shadow` property on web, instead of separate props for color, offset and radius.

What do you think about that?

MatiPl01 (approved):
Left a few small remarks, but, in general, shadow examples are much better now. Good job!

packages/docs-reanimated/docs/guides/supported-properties.mdx
kacperkapusciak: Reanimated is the good guy, actually

Fixed in 6ad579b75e

PR #6934, opened 2025-01-24
chore: Add 0.78.0-rc.3 support

MatiPl01:
## Summary

Adds if checks for the RN 0.78 in native code, thanks to which package builds on both - 0.78 nd 0.77.

## Checklist

- [ ] Upgrade `react-native-screens` when 0.78 support is added (https://github.com/software-mansion/react-native-screens/pull/2626)
- [ ] Upgrade `react-native-svg` when 0.78 support is added (https://github.com/software-mansion/react-native-svg/pull/2619)
- [ ] Upgrade `react-native-gesture-handler` when 0.78 support is added (https://github.com/software-mansion/react-native-gesture-handler/pull/3354)
- [ ] `react-native-pager-view` auto generated schema contains invalid `StringEnumTypeAnnotation` fields and crashes build on Android

|| Fabric (iOS) | Fabric (Android) | Web | MacOS (not upgraded) | TVOS (not upgraded) |
|-|-|-|-|-|-|
| Builds | ✅ | ✅ | ✅ | ✅ | ✅ |
| Runs   | ❌ | ❌ | ❌ | ✅ | ✅ |

*Apps where RN was upgraded don't run because of incompatible dependency versions


MatiPl01:
As we discussed with @tomekzaw, we now use if/else macros to support RN 0.78 and will get rid of them when all problematic deps of our example apps can be bumped to 0.78 as well.

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsProxy.cpp
```diff
@@ -287,17 +291,24 @@ void LayoutAnimationsProxy::handleUpdatesAndEnterings(
...
-          nodeForTag_[mutation.parentShadowView.tag]->applyMutationToIndices(
-              mutation);
+
+#if REACT_NATIVE_MINOR_VERSION >= 78
+        const auto parentTag = mutation.parentTag;
+        const auto mutationParent = parentTag;
+#else
+        const auto parentTag = mutation.parentShadowView.tag;
+        const auto mutationParent = mutation.parentShadowView;
+#endif // REACT_NATIVE_MINOR_VERSION >= 78
```
MatiPl01, line 301: This `mutationParent` variable makes it possible to easily handle value of different type on different versions without needing to add many if-else declarations.

tomekzaw (approved):

PR #6944, opened 2025-01-27
feat: New style props support in CSS animations

MatiPl01:
## Summary

This PR adds support for new animation properties added in RN `0.77` and implements respective examples for all of added props.

Support was added for the following props
- `boxSizing`
- `outlineColor`
- `outlineOffset`
- `outlineStyle`
- `outlineWidth`
- `mixBlendMode`

Supported props have been also updated in [this](https://docs.google.com/spreadsheets/d/1PCYJ8e4Kk1R8rJ4J7-iuywW3hYmTcrBJsXWcgSZ9ZjY/edit?gid=0#gid=0) spreadsheet.

### DIFF
--- a/packages/react-native-reanimated/src/propsAllowlists.ts
+++ b/packages/react-native-reanimated/src/propsAllowlists.ts
@@ -29,6 +29,10 @@ export const PropsAllowlists: AllowlistsHolder = {
     scaleY: true,
     translateX: true,
     translateY: true,
+    outlineStyle: true,
+    outlineWidth: true,
+    outlineOffset: true,
+    borderStyle: true,
   },
   /**
    * Whitelist of view props that can be updated in native thread via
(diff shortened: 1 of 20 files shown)


packages/react-native-reanimated/src/propsAllowlists.ts
```diff
@@ -123,7 +130,6 @@ export const PropsAllowlists: AllowlistsHolder = {
     /* text color */
     color: true,
     tintColor: true,
-    shadowColor: true,
```
piaskowyk, line 126: Just to be sure - is it safe to do that? 🤔

piaskowyk (approved):
Looks good to me 👍 I have just one question

packages/react-native-reanimated/src/propsAllowlists.ts
MatiPl01: It works after moving to the UI props, so I think we can animate the `shadowColor` on UI:

| iOS | Android |
|-|-|
| <video src="https://github.com/user-attachments/assets/f7b053b3-ee85-49a4-873c-f53efe21f3e0" /> | <video src="https://github.com/user-attachments/assets/0a030206-f6d6-4cfc-a9a0-82c40ebf14a7" /> |

PR #6946, opened 2025-01-28
Add missing `targets` in build.gradle

tomekzaw:
## Summary

This PR potentially fixes undefined symbol linking errors happening in RN apps when Reanimated is required by other third-party libraries like Live Markdown or Vision Camera caused by incorrect order of task execution caused by incomplete dependency graph which can be fixed by adding `targets` function call in `externalNativeBuild` to specify shared libraries.


tjzel (approved):
Pinging myself to include it in https://github.com/software-mansion/react-native-reanimated/pull/6827

PR #6947, opened 2025-01-28
fix: Hot reload in web-example, improve navigation

MatiPl01:
## Summary

This PR:
- fixes hot reload on web,
- improves navigation (uses proper screen title, better back button name (adds space between words)),
- fixes web url to navigation state mapping (proper page is displayed on refresh or when navigating to a screen via url change)

### DIFF
--- a/packages/react-native-reanimated/src/css/platform/web/utils.ts
+++ b/packages/react-native-reanimated/src/css/platform/web/utils.ts
@@ -41,7 +41,7 @@ export function kebabize<T extends string>(property: T) {
 
 function easingMapper(easing: CSSTimingFunction) {
   if (typeof easing === 'string') {
-    return kebabize(easing);
+    return easing;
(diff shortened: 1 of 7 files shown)


packages/react-native-reanimated/src/css/platform/web/utils.ts
```diff
@@ -41,7 +41,7 @@ export function kebabize<T extends string>(property: T) {
 
 function easingMapper(easing: CSSTimingFunction) {
   if (typeof easing === 'string') {
-    return kebabize(easing);
+    return easing;
```
MatiPl01, line 44: I forgot to change this. It is not needed anymore, since we now use kebabized names for predefined easings.

piaskowyk (approved):
👍

tomekzaw (approved):

PR #6950, opened 2025-01-28
Use `install_modules_dependencies` by default in RNWorklets.podspec

tomekzaw:
## Summary

`respond_to?(:install_modules_dependencies, true)` returns true on RN 0.71+ and currently we support RN 0.74+ so it's high time to remove the conditional in RNWorklets.podspec and always call `install_modules_dependencies`.


piaskowyk (approved):

piaskowyk:
Finally 😍

PR #6953, opened 2025-01-29
Remove unused reference to runtime from `requestRender` method

piaskowyk:
## Summary

As the title ⬆️ 

## Test plan

Just compile application

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
@@ -519,9 +519,7 @@ void ReanimatedModuleProxy::requestAnimationFrame(
 void ReanimatedModuleProxy::maybeRequestRender() {
   if (!renderRequested_) {
     renderRequested_ = true;
-    jsi::Runtime &uiRuntime =
-        workletsModuleProxy_->getUIWorkletRuntime()->getJSIRuntime();
-    requestRender_(onRenderCallback_, uiRuntime);
+    requestRender_(onRenderCallback_);
   }
 }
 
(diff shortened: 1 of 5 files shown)


MatiPl01 (approved):
Good catch!
I wonder why I needed this runtime here if it wasn't used at all.

packages/react-native-reanimated/apple/reanimated/apple/native/PlatformDepMethodsHolderImpl.mm
```diff
@@ -85,17 +85,15 @@ SetGestureStateFunction makeSetGestureStateFunctionBridgeless(RCTModuleRegistry
...
 #if !TARGET_OS_OSX
       auto targetTimestamp = displayLink.targetTimestamp;
 #else
       // TODO macOS targetTimestamp isn't available on macOS
       auto targetTimestamp = displayLink.timestamp + displayLink.duration;
 #endif
-      double frameTimestamp =
-
-          (targetTimestamp) * 1000;
+      const double frameTimestamp = targetTimestamp * 1000;
```
tomekzaw, line 96: We purposefully left `(targetTimestamp) * 1000` here in a previous PR because we suspect there used to be a function call here that accepted `targetTimestamp` as the first argument. Most likely this was related to slow animations. I'm okay with this change, just wanted to remind about it. Perhaps we should fix it in a separate PR?

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -918,18 +900,11 @@ void ReanimatedModuleProxy::performOperations() {
...
-      [weakThis = weak_from_this()](double time) {
-        auto strongThis = weakThis.lock();
-        if (!strongThis) {
-          return;
-        }
-
-        strongThis->shouldFlushRegistry_ = true;
-      },
-      rt);
+  requestRender_([weakThis = weak_from_this()](double time) {
```
tomekzaw, line 903: This is a nit but since we're touching this part of the code, we can also unify the argument name with other occurrences.
```suggestion
  requestRender_([weakThis = weak_from_this()](const double timestamp) {
```

tomekzaw (approved):
Good catch!

packages/react-native-reanimated/apple/reanimated/apple/native/PlatformDepMethodsHolderImpl.mm
piaskowyk: I've restored those changes and I'll prepare a PR with the fix 👍

PR #6958, opened 2025-01-30
Fix `EasingFunctionFactory` type usage

bartlomiejbloniarz:
## Summary

This PR adds missing `EasingFunctionFactory` type, when it's needed (e.g. Keyframe Animations). It's needed for easings like `Easing.bezier` that return an easing factory function.


tjzel (changes requested):
Please add a minimal TypeScript test case so we know it works 🙏

tjzel (approved):

PR #6968, opened 2025-01-31
Parse `transition` shorthand

tomekzaw:

### DIFF
--- a/apps/common-app/src/apps/reanimated/examples/EmptyExample.tsx
+++ b/apps/common-app/src/apps/reanimated/examples/EmptyExample.tsx
@@ -1,11 +1,24 @@
-import { Text, StyleSheet, View } from 'react-native';
+import { StyleSheet, View, Button } from 'react-native';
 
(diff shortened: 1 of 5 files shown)


packages/react-native-reanimated/src/css/utils/conversions.ts
```diff
@@ -1,15 +1,30 @@
...
+
+export function kebabizeCamelCase<T extends string>(property: T) {
+  return property
+    .split('')
+    .map((letter, index) =>
+      letter.toUpperCase() === letter
+        ? `${index !== 0 ? '-' : ''}${letter.toLowerCase()}`
+        : letter
+    )
+    .join('');
```
tomekzaw, line 25: How about
```tsx
str.replace(/[A-Z]/g, (x) => `-${x[1].toLowerCase()}`);
```
instead for the sake of consistency with `camelizeKebabCase`?

MatiPl01 (approved):

PR #6972, opened 2025-02-03
feat: Add support for multiple transitionBehavior property settings in CSS transitions

MatiPl01:
## Summary

This PR adds support for multiple `transitionBehavior` property values in CSS transitions. Before it allowed passing just a single value `'normal'`/`'allow-discrete'`. Now it behaves properly and allows passing multiple values as an array.


packages/react-native-reanimated/Common/cpp/reanimated/CSS/core/CSSTransition.cpp
```diff
@@ -42,15 +34,57 @@ jsi::Value CSSTransition::getCurrentInterpolationStyle(jsi::Runtime &rt) const {
   return styleInterpolator_.getCurrentInterpolationStyle(rt, shadowNode_);
 }
 
+PropertyNames CSSTransition::getAllowedProperties(
```
MatiPl01, line 37: This function took a part of the previous implementation's `getChangedProps` function responsibility. 

I think it is even better and more logically structured now when transition is responsible for telling which props can be transitioned. It was also easier to implement it in this way after adding support for multiple `transitionBehavior` values.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/core/CSSTransition.cpp
```diff
@@ -42,15 +34,57 @@ jsi::Value CSSTransition::getCurrentInterpolationStyle(jsi::Runtime &rt) const {
...
 void CSSTransition::updateSettings(const PartialCSSTransitionConfig &config) {
   if (config.properties.has_value()) {
     updateTransitionProperties(config.properties.value());
   }
-  if (config.allowDiscrete.has_value()) {
-    allowDiscrete_ = config.allowDiscrete.value();
-  }
   if (config.settings.has_value()) {
-    progressProvider_.setSettings(config.settings.value());
+    settings_ = config.settings.value();
```
MatiPl01, line 87: It's better to store settings in the transition class and pass them to the progress provider when needed instead of storing them in 2 places at the same time. We cannot just simply store them only in the progress provider after adding support for multiple `transitionBehavior` props, which we have to read from the transition settings object.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/core/CSSTransition.cpp
```diff
@@ -85,10 +120,25 @@ void CSSTransition::updateTransitionProperties(
...
   progressProvider_.discardIrrelevantProgressProviders(transitionPropertyNames);
 }
 
+bool CSSTransition::isAllowedProperty(const std::string &propName) const {
+  if (!isDiscreteProperty(propName)) {
+    return true;
+  }
+
+  const auto &propertySettings =
+      getTransitionPropertySettings(settings_, propName);
```
MatiPl01, line 134: Just a helper function declared in the `CSSTransitionConfig` file that will return settings corresponding to the specific property.

tomekzaw (approved):

PR #6978, opened 2025-02-04
Add `displayName` to `forwardRef` from `AnimatedComponent`

m-bert:
## Summary

After switching to new `findNodeHandle` implementation based on `display: contents;`, we can no longer detect components from `SVG` that are wrapped using `createAnimatedComponent` (see [issue](https://github.com/software-mansion/react-native-gesture-handler/issues/3356) on Gesture Handler side). This is because our [Wrap](https://github.com/software-mansion/react-native-gesture-handler/blob/7f90209add0b761d3738882339f4949b31f6f88f/src/handlers/gestures/GestureDetector/Wrap.web.tsx#L9) receives `forwardRef` as child - that makes it impossible to detect what component is used inside. Theoretically we could do something like

```js
child.type.render().displayName
```

and use regex to get component name, but it would be very expensive.

Instead, we can add `displayName` directly to `forwardRef` on Reanimated side. This should help us detect if component passed inside `createAnimatedComponent` was from `SVG`.
## Test plan

Code used inside our [Wrap.web.tsx](https://github.com/software-mansion/react-native-gesture-handler/blob/7f90209add0b761d3738882339f4949b31f6f88f/src/handlers/gestures/GestureDetector/Wrap.web.tsx#L9)

```jsx
const child = React.Children.only(children);

console.log(child)
console.log(child.type.displayName)
```

gives the following result:

image


tomekzaw (approved):
LGTM, haven't tested it though

tjzel (approved):

piaskowyk (approved):

Issue #6839, opened 2024-12-20
__reanimatedLoggerConfig


hoantq19:
### Description

file logger.ts in node-module react-native renimate there is an error
Cannot find name '__reanimatedLoggerConfig'.t .How do i modify

### Steps to reproduce

import Carousel from 'react-native-reanimated-carousel';

 <Carousel
                                autoPlay={true}
                                loop={true}
                                autoPlayInterval={3000}
                                width={deviceWidth}
                                height={240}
                                data={[]}
                                onSnapToItem={(index) => {
                                    this.setState({ activeSlide: index });
                                    this.forceUpdate();
                                }}
                                renderItem={(item: { item: ImageHomeModel, index: Number }) => this.RenderImageHome(item.item)}
                            />

### Snack or a link to a repository

git

### Reanimated version

3.6.1

### React Native version

0.75.4

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

pixel 7a

### Acknowledgements

Yes

tlow92:
After upgrading to expo 52 I also received this error and took me a long time to figure out what caused it.

I had some configuration in `metro.config.js` (I think some leftover from past times where it was needed for react-native-skia web).

Can you check if you have one of the options activated?
Removing this piece fixed the issue for me.

```
config.transformer.getTransformOptions = async () => ({
  transform: {
    experimentalImportSupport: false,
    inlineRequires: true,
  },
});
```

noumantahir:
I was facing this exact same issue after updating my reanimated package from `3.6` to `3.16` .. 

wrapping my metro config with reanimated metro config resolved the issue for me

```
// metro.config.js
const {
  wrapWithReanimatedMetroConfig,
} = require('react-native-reanimated/metro-config');

const config = {
  // Your existing Metro configuration options
};

module.exports = wrapWithReanimatedMetroConfig(config);
```

ref: https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/getting-started#step-3-wrap-metro-config-with-reanimated-wrapper-recommended

Issue #6740, opened 2024-11-21
After updating to expo 52: ReferenceError: _WORKLET | __reanimatedLoggerConfig is not defined


abretonc7s:
### Description

After updating to Expo SDK 52, I encountered a ReferenceError: _WORKLET is not defined error on web platform. This error occurred after running expo install --fix to update dependencies. The error prevents the application from running properly on web.

Relevant dependencies:
```json
{
  "dependencies": {
    "expo": "~52.0.0",
    "react-native-reanimated": "~3.6.1",
    "@react-native-community/slider": "4.5.5",
    "@gorhom/bottom-sheet": "^5.0.5",
    "@gorhom/portal": "~1.0.14"
  }
}
```

```tsx
// packages/expo-audio-ui/src/AnimatedCandle/AnimatedCandle.tsx
import { Rect } from '@shopify/react-native-skia'
import React, { useEffect } from 'react'
import { Platform } from 'react-native'
import { useSharedValue, withTiming } from 'react-native-reanimated'

import { CANDLE_OFFCANVAS_COLOR } from '../constants'

export interface AnimatedCandleProps {
    height: number
    x: number
    y: number
    startY: number
    width: number
    color: string
    animated?: boolean
}
const AnimatedCandle: React.FC<AnimatedCandleProps> = ({
    color: targetColor,
    x: targetX,
    y: targetY,
    startY,
    height: targetHeight,
    width,
    animated = true,
}) => {
    const y = useSharedValue(startY)
    const height = useSharedValue(0)
    const x = useSharedValue(targetX)
    const color = useSharedValue(CANDLE_OFFCANVAS_COLOR)

    useEffect(() => {
        if (Platform.OS === 'web') {
            y.value = targetY
            height.value = targetHeight
            x.value = targetX
            color.value = targetColor
            return
        }

        if (animated) {
            y.value = withTiming(targetY, { duration: 500 })
            height.value = withTiming(targetHeight, { duration: 500 })
            x.value = withTiming(targetX, { duration: 500 })
            color.value = withTiming(targetColor, { duration: 500 })
        } else {
            y.value = targetY
            height.value = targetHeight
            x.value = targetX
            color.value = targetColor
        }
    }, [targetY, targetHeight, targetX, targetColor, animated])

    return <Rect x={x} y={y} width={width} height={height} color={color} />
}

export default React.memo(AnimatedCandle)
```

### Steps to reproduce

on Web it seems that using any wtihTiming creates the issue.

[...]

PR #6980, opened 2025-02-05
fix: initializer file not loaded in Expo52

tjzel:
## Summary

Turns out that `sideEffects` field was ignored and `isWeb()` check evaluated to static false, removing `initializers.ts` from the bundle.

Fixes
- #6740
- #6839

## Test plan

Tested it on RNGH web example app using Expo52


m-bert (approved):
Works for me 🚀

PR #6986, opened 2025-02-05
chore: CSS values cleanup

MatiPl01:
## Summary

This PR removes some unused code from CSS values and cleans up the `CSSValueVariant` implementation a bit

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSKeyword.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSKeyword.cpp
@@ -48,11 +48,6 @@ bool CSSKeywordBase<TValue>::canConstruct(const folly::dynamic &value) {
 }
 
 template <typename TValue>
-CSSValueType CSSKeywordBase<TValue>::type() const {
-  return CSSValueType::Keyword;
-}
-
-template <typename TValue>
 jsi::Value CSSKeywordBase<TValue>::toJSIValue(jsi::Runtime &rt) const {
   return jsi::String::createFromUtf8(rt, value);
 }
--- a/packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSValue.h
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSValue.h
@@ -28,6 +28,7 @@ struct CSSValue {
   static constexpr bool is_discrete_value = false;
 
   virtual ~CSSValue() = default;
+
   virtual jsi::Value toJSIValue(jsi::Runtime &rt) const = 0;
   virtual folly::dynamic toDynamic() const = 0;
   virtual std::string toString() const = 0;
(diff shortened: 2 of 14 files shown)


packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSValue.h
```diff
@@ -11,18 +11,6 @@ namespace reanimated {
 
 using namespace facebook;
 
-enum class CSSValueType {
```
MatiPl01, line 14: This enum wasn't used at all

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSValue.h
```diff
@@ -36,38 +24,28 @@ struct CSSResolvableValueInterpolationContext {
 };
 
 struct CSSValue {
-  virtual ~CSSValue() = default;
+  // This field should be overridden in discrete value types
+  static constexpr bool is_discrete_value = false;
```
MatiPl01, line 28: No need to repeat it twice in `CSSSimpleValue` and `CSSResolvableValue` as I did before

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSValue.h
```diff
@@ -77,23 +55,24 @@ struct CSSResolvableValue : public CSSValue {
       const CSSResolvableValueInterpolationContext &context) const = 0;
 };
 
-inline bool isDiscrete(const CSSValue &value) {
```
MatiPl01, line 80: Wasn't used because we used a static check instead of this one

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSValueVariant.h
```diff
@@ -65,6 +52,13 @@ static constexpr bool has_can_construct_jsi =
  */
 template <typename... AllowedTypes>
 class CSSValueVariant final : public CSSValue {
+  static_assert(
+      (CSSValueDerived<AllowedTypes> && ...),
+      "CSSValueVariant accepts only CSSValue-derived types");
+  static_assert(
+      (CanConstructFromJSIValue<AllowedTypes> && ...),
```
MatiPl01, line 59: Moved this check here from the `tryConstruct` for `jsi::Value` and made it more strict - now it is not trying to construct a value only if there is a `jsi::Value` constructor but this constructor is required and all classes/struct used with the `CSSValueVariant` must be constructible from the `jsi::Value`.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSKeyword.cpp
```diff
@@ -31,11 +31,6 @@ bool CSSKeywordBase<TValue>::canConstruct(
   return jsiValue.isString();
 }
 
-template <typename TValue>
```
MatiPl01, line 34: Moved to `CSSValue.h`

piaskowyk (approved):
👍

PR #6991, opened 2025-02-06
fix: move requestAnimationFrame back to Reanimated

tjzel:
## Summary

I introduced a regression in #6973 by moving JS implementation of `requestAnimationFrame` to Worklets. Native version of `requestAnimationFrame` is injected in Reanimated's `UIRuntimeDecorator.cpp` so `requestAnimationFrame` must be overwritten in Reanimated, not in Worklets.

Since all the native implementation is in Reanimated and it would be a hard task to extract it right now, I decided to keep it for the time being unless we decide where it should land eventually.

## Test plan

The only difference is the lack of batching with my regression introduced. You can `console.log(currentCallbacks.length)` to see that the batching is restored.

### DIFF
--- a/packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
+++ b/packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
@@ -18,6 +18,7 @@ import { getShadowNodeWrapperFromRef } from '../fabricUtils';
 import { checkCppVersion } from '../platform-specific/checkCppVersion';
 import { jsVersion } from '../platform-specific/jsVersion';
 import { isFabric, shouldBeUseWeb } from '../PlatformChecker';
+import { setupRequestAnimationFrame } from '../requestAnimationFrame';
 import { ReanimatedTurboModule } from '../specs';
 import type {
   IWorkletsModule,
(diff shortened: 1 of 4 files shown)


packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
```diff
@@ -81,7 +82,11 @@ See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooti
       checkCppVersion();
     }
     this.#reanimatedModuleProxy = global.__reanimatedModuleProxy;
-    executeOnUIRuntimeSync(registerReanimatedError);
```
tjzel, line 84: Forgot the second `()` previously...

packages/react-native-reanimated/src/errors.ts
```diff
@@ -4,7 +4,9 @@ import { createCustomError, registerCustomError } from './WorkletsResolver';
 
 export const ReanimatedError = createCustomError('Reanimated');
 
+const ReanimatedErrorConstructor = ReanimatedError;
```
tjzel, line 7: We must alias it as `ReanimatedError` is not a captured identifier.

piaskowyk (approved):

PR #6993, opened 2025-02-06
chore: Make CSS interpolators stateless

MatiPl01:
## Summary

This PR removes, probably unnecessary, premature optimizations which added code complexity to interpolators implementation. Interpolators kept state of the current keyframe and last interpolated value. They also were tightly coupled with progress providers which must have been passed in the constructor.

Even though the previous implementation worked fine, it was impossible to optimize creation of the animation (we had to create interpolators for all views separately, even if they used the same animation). This also resulted in higher memory consumption.

Instead of moving left or right on the keyframes sequence, I now use a binary search to find the current keyframe. It may have worse time complexity (`O(log(n))` instead of `O(1)`) but, in fact, may be even faster because of removed logic responsible for storing a previous value, iterating over keyframes, etc. This change shouldn't affect performance as long as somebody doesn't use thousands or keyframes in animations (which is rather impossible).

## How it was tested?

I just went through examples in the example app (on both, iOS and Android) to make sure that no regression was introduced in this PR.


packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSColor.cpp
```diff
@@ -26,13 +26,13 @@ CSSColor::CSSColor(const ColorChannels &colorChannels)
...
-    // iOS uses unsigned 32-bit integers for colors
-    auto color = static_cast<unsigned>(jsiValue.asNumber());
-#endif
+    double numberValue = jsiValue.asNumber();
+    uint32_t color;
+    if (numberValue < 0) {
+      color = static_cast<int32_t>(numberValue);
+    } else {
+      color = static_cast<uint32_t>(numberValue);
+    }
```
MatiPl01, line 35: This is much safer approach as it still works on both platforms and I needed to change this code after I introduced previous value reading from the updates registry. Because of that, the value was already converted to the expected format and I no longer had to change it.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/core/CSSTransition.cpp
```diff
@@ -91,22 +92,29 @@ void CSSTransition::updateSettings(const PartialCSSTransitionConfig &config) {
...
+      rt, changedProps, lastUpdateValue);
   return update(rt, timestamp);
 }
 
 jsi::Value CSSTransition::update(jsi::Runtime &rt, const double timestamp) {
   progressProvider_.update(timestamp);
-  return styleInterpolator_.update(
-      rt, shadowNode_, progressProvider_.getRemovedProperties());
+  auto result =
+      styleInterpolator_.interpolate(rt, shadowNode_, progressProvider_);
```
MatiPl01, line 110: `interpolate` method replaced previous `update` method. It no longer updates any state in the interpolator and requires passing a progress provider as a parameter, because the style interpolator is no longer dependent on the specific progress provider.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/interpolation/transforms/TransformsStyleInterpolator.cpp
```diff
@@ -338,6 +315,26 @@ TransformsStyleInterpolator::createTransformInterpolationPair(
...
+      });
+
+  // If we're at the end, return the last valid keyframe index
+  if (it == keyframes_.end()) {
+    return keyframes_.size() - 1;
+  }
+
+  return std::distance(keyframes_.begin(), it);
+}
+
```
MatiPl01, line 337: This binary search replaces a previous logic with keyframe index incrementation/decrementation.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/interpolation/transforms/TransformsStyleInterpolator.cpp
```diff
@@ -352,95 +349,6 @@ TransformsStyleInterpolator::getDefaultOperationOfType(
...
-          keyframe->toOperations});
-    } else {
-      return std::make_shared<TransformKeyframe>(TransformKeyframe{
-          keyframe->fromOffset,
-          keyframe->toOffset,
-          keyframe->fromOperations,
-          resolveTransformOperations(
-              shadowNode, unresolvedOperations.value())});
-    }
-  }
```
MatiPl01, line 394: I think that this logic was not necessary. It was intended to convert keyframes with relative values to absolute (numeric) values but we don't have to do that and can keep % values. It still works the same.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/interpolation/groups/RecordPropertiesInterpolator.cpp
```diff
@@ -6,31 +6,19 @@ namespace reanimated {
...
-        rt, jsi::PropNameID::forUtf8(rt, propName));
-
-    const auto interpolatorIt = interpolators_.find(propName);
-    if (interpolatorIt == interpolators_.end() ||
-        !interpolatorIt->second->equalsReversingAdjustedStartValue(
-            rt, propValue)) {
+    const folly::dynamic &propertyValue) const {
+  for (const auto &pair : propertyValue.items()) {
+    const auto &propName = pair.first.asString();
+    const auto &propValue = pair.second;
```
piaskowyk, line 17:
```suggestion
  for (const auto &[propName, propValue] : propertyValue.items()) {
```
What do you think?

packages/react-native-reanimated/Common/cpp/reanimated/CSS/interpolation/groups/RecordPropertiesInterpolator.cpp
```diff
@@ -108,9 +84,8 @@ folly::dynamic RecordPropertiesInterpolator::mapInterpolators(
     const {
   folly::dynamic result = folly::dynamic::object;
 
-  for (const auto &[propName, interpolator] : interpolators_) {
-    auto value = callback(*interpolator);
-    result[propName] = value;
+  for (const auto &pair : interpolators_) {
+    result[pair.first] = callback(*pair.second);
```
piaskowyk, line 88: I think unpacking it directly in the `for` loop definition would make it more readable, but the performance of both approaches is the same.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/interpolation/styles/TransitionStyleInterpolator.cpp
```diff
@@ -7,66 +7,48 @@ TransitionStyleInterpolator::TransitionStyleInterpolator(
...
+  }
 
-  for (size_t i = 0; i < propertiesCount; ++i) {
-    const auto propertyName =
-        propertyNames.getValueAtIndex(rt, i).asString(rt).utf8(rt);
-    const auto propertyValue = propertyValuesObject.getProperty(
-        rt, jsi::PropNameID::forUtf8(rt, propertyName));
+  for (const auto &pair : newPropertyValues.items()) {
+    const auto propertyName = pair.first.getString();
+    const auto &propertyValue = pair.second;
```
piaskowyk, line 21: same thing as above

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
```diff
@@ -140,9 +138,11 @@ PropsObserver CSSTransitionsRegistry::createPropsObserver(const Tag viewTag) {
     {
       std::lock_guard<std::mutex> lock{strongThis->mutex_};
 
-      const auto &initialProps =
-          transition->run(rt, changedProps, strongThis->getCurrentTimestamp_());
       const auto &shadowNode = transition->getShadowNode();
+      const auto &lastUpdates =
+          strongThis->getUpdatesFromRegistry(shadowNode->getTag());
+      const auto &initialProps = transition->run(
```
piaskowyk, line 144: Why `transition->run()` returns the `initialProps` instead new props for component?

piaskowyk (approved):
Looks good 👍 I've just left a few small comments, but nothing really important. I'm glad that these changes have significantly improved the readability of the codebase.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
MatiPl01: I just call it `initialProps` because this is what I initially (when the transition starts) put in the updates registry to prevent react from committing transitioned props without animating them.

We cannot just use new component props in here as we will get the final state right away without any transition. `initialProps` contain old style values for props that weren't transitioned when the new tranistion was triggered and last style update values for interrupred transition properties.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/interpolation/groups/RecordPropertiesInterpolator.cpp
MatiPl01: Can we use `propName` then without explicitly converting it to string? I will check that.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
MatiPl01: I will change this name to `transitionStartStyle` or something like that

PR #7006, opened 2025-02-10
Remove UI runtime usage by CSS Animations

piaskowyk:
## Summary

The goal of Reanimated 4 is to enable the use of CSS Animations without needing access to the UI runtime. This PR removes references to the UI runtime within the CSS codebase. It's important to note that in some cases, we still have access to a runtime, but this is a React Runtime used on the React thread.

There is no performance degradation after the rewrite to `folly::dynamic`. It's even a bit faster, but the improvement is within the measurement error range.


packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -869,32 +883,57 @@ void ReanimatedModuleProxy::performOperations() {
...
+  if (!hasPropsToRevert && hasLayoutUpdates.hasAny()) {
     ReanimatedSystraceSection s(
         "ReanimatedModuleProxy::synchronouslyUpdateUIProps");
-    for (const auto &[shadowNode, props] : updatesBatch) {
-      Tag tag = shadowNode->getTag();
-      synchronouslyUpdateUIPropsFunction_(rt, tag, props->asObject(rt));
+    if (!hasLayoutUpdates.cssTransition) {
+      for (const auto &[shadowNode, props] : transitionUpdatesBatch) {
+        Tag tag = shadowNode->getTag();
+        synchronouslyUpdateUIPropsByDynamicFunction_(tag, props);
```
piaskowyk, line 922: Code related to `synchronouslyUpdateUIProps` will be removed after the merge of https://github.com/software-mansion/react-native-reanimated/pull/7014

packages/react-native-reanimated/Common/cpp/reanimated/CSS/interpolation/transforms/TransformOperation.cpp
```diff
@@ -165,16 +165,68 @@ std::shared_ptr<TransformOperation> TransformOperation::fromJSIValue(
...
+  auto &obj = value;
+  if (obj.size() != 1) {
+    throw std::invalid_argument(
+        "[Reanimated] TransformOperation must have exactly one property.");
+  }
+
+  auto property = obj.items().begin()->second;
+  auto propertyName = obj.items().begin()->first.getString();
+  TransformOperationType operationType =
+      getTransformOperationType(propertyName);
```
MatiPl01, line 184: I like this approach. That's something I should also do with the constructor for the `jsi::Value`

packages/react-native-reanimated/Common/cpp/reanimated/CSS/misc/ViewStylesRepository.cpp
```diff
@@ -125,36 +106,35 @@ void ViewStylesRepository::updateCacheIfNeeded(
...
-    jsi::Runtime &rt,
+folly::dynamic ViewStylesRepository::getPropertyValue(
     const folly::dynamic &value,
     const PropertyPath &propertyPath) {
   const folly::dynamic *currentValue = &value;
 
   for (size_t i = 0; i < propertyPath.size(); ++i) {
     if (currentValue->isNull() || currentValue->empty()) {
-      return jsi::Value::undefined();
+      return folly::dynamic();
```
MatiPl01, line 116: This is null by default, right? I mean that `isNull` check will return `true` for this dynamic later on?

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.h
```diff
@@ -46,6 +46,17 @@ using namespace facebook;
 
 using UpdatesBatch =
     std::vector<std::pair<ShadowNode::Shared, std::unique_ptr<jsi::Value>>>;
+using CSSUpdatesBatch =
+    std::vector<std::pair<ShadowNode::Shared, folly::dynamic>>;
+
+struct HasLayoutupdates {
```
MatiPl01, line 52:
```suggestion
struct HasLayoutUpdates {
```

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -822,6 +834,8 @@ void ReanimatedModuleProxy::performOperations() {
       workletsModuleProxy_->getUIWorkletRuntime()->getJSIRuntime();
 
   UpdatesBatch updatesBatch;
+  CSSUpdatesBatch transitionUpdatesBatch;
+  CSSUpdatesBatch animationUpdatesBatch;
```
MatiPl01, line 838: Why do we need separate batches here? Is that because of the `jsi::Value` types in the `updatesBatch` for animated props?

Since the `flushUpdates` method called in the `performOperations` below for each registry already converts value to `folly::dynamic` and stores it in the internal `updatesRegistry_` (see UpdatesRegistry.h), I think that we can simply use the same batch with `folly::dynamic` for all 3 registries instead of creating 3 separate batches and duplicating code multiple times.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -832,22 +846,22 @@ void ReanimatedModuleProxy::performOperations() {
...
-      cssAnimationsRegistry_->update(rt, currentCssTimestamp_);
-      cssAnimationsRegistry_->flushUpdates(rt, updatesBatch, true);
+      cssAnimationsRegistry_->update(currentCssTimestamp_);
+      cssAnimationsRegistry_->flushUpdates(animationUpdatesBatch, true);
     }
 
     shouldUpdateCssAnimations_ = false;
 
-    if (updatesBatch.size() > 0 &&
+    if ((updatesBatch.size() > 0 || transitionUpdatesBatch.size() > 0) &&
```
MatiPl01, line 864: Doesn't it miss a check for animations?

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.h
```diff
@@ -46,6 +46,17 @@ using namespace facebook;
...
+  bool cssTransition = false;
+  bool cssAnimation = false;
+  bool workletAnimation = false;
```
MatiPl01, line 55: Do we really need to track layout updates separately? Why can't we have just a single flag that determines if there are layout props in the current updates batch? Is that related to having 3 separate batches now?

MatiPl01 (changes requested):
In general, it looks ok and the duplication of code in `packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values` is fine, but I don't understand why we have to duplicate code in `performOperations` and would prefer to have this simplified before merging this PR.

packages/react-native-reanimated/apple/reanimated/apple/native/REAJSIUtils.h
```diff
@@ -122,3 +122,34 @@ static id convertJSIValueToObjCObject(jsi::Runtime &runtime, const jsi::Value &v
 
   throw std::runtime_error("[Reanimated] Unsupported jsi::Value kind.");
 }
+
+static id convertDynamicToNSObject(const folly::dynamic &value)
```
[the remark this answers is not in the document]
piaskowyk: That code will be removed after the merge of https://github.com/software-mansion/react-native-reanimated/pull/7014, so we don't need to pay a lot of attention to it I think.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/misc/ViewStylesRepository.cpp
piaskowyk: Yes, exactly 👍 
image

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
piaskowyk: Good catch! It is a good idea 👍 
Solved in [#b78adbc](https://github.com/software-mansion/react-native-reanimated/pull/7006/commits/b78adbcb3fb32cddf628ff8815d53ca767ac72cb)

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.h
piaskowyk: Solved in [#b78adbc](https://github.com/software-mansion/react-native-reanimated/pull/7006/commits/b78adbcb3fb32cddf628ff8815d53ca767ac72cb)

MatiPl01 (approved):
Looks very clean now!

You got rid of multiple conversions from `folly::dynamic` to the `jsi::Value` and vice versa, so even performance might be better now.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSDimension.cpp
```diff
@@ -36,15 +36,39 @@ CSSDimension::CSSDimension(jsi::Runtime &rt, const jsi::Value &jsiValue) {
...
+}
+
 bool CSSDimension::canConstruct(const std::string &value) {
   return !value.empty() && value.back() == '%';
 }
 
+bool CSSDimension::canConstruct(const char *value) {
+  auto str = std::string(value);
+  return !str.empty() && str.back() == '%';
+}
```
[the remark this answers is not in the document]
piaskowyk: All changes ([comment 1](https://github.com/software-mansion/react-native-reanimated/pull/7006#discussion_r1954650425), [comment 2](https://github.com/software-mansion/react-native-reanimated/pull/7006#discussion_r1954651772), [comment 3](https://github.com/software-mansion/react-native-reanimated/pull/7006#discussion_r1954656400)) are related to the fact that `folly::dynamic` has a constructor with a string, which leads to ambiguous constructor issues in our variant templates. At this moment, I don't have a better solution, but I'm open to any suggestions.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSDimension.cpp
```diff
@@ -11,10 +11,10 @@ CSSDimension::CSSDimension(const double value)
 CSSDimension::CSSDimension(const double value, const bool isRelative)
     : value(value), isRelative(isRelative) {}
 
-CSSDimension::CSSDimension(const std::string &value) {
+CSSDimension::CSSDimension(const char *value) {
```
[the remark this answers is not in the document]
piaskowyk: https://github.com/software-mansion/react-native-reanimated/pull/7006#discussion_r1955936226

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSDimension.h
```diff
@@ -15,11 +15,14 @@ struct CSSDimension
   CSSDimension();
   explicit CSSDimension(double value);
   explicit CSSDimension(double value, bool isRelative);
-  explicit CSSDimension(const std::string &value);
+  explicit CSSDimension(const char *value);
```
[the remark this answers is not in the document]
piaskowyk: https://github.com/software-mansion/react-native-reanimated/pull/7006#discussion_r1955936226

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSDimension.cpp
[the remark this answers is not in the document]
tjzel: @piaskowyk 
If I understand correctly the problem is with implicit conversions. I think you can just add `explicit` keyword to the overload using `folly::dynamic`.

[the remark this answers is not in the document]
tjzel: Nevermind, explicit is already there. The problem comes from the fact that we already allow implicit conversions, that's why for e.g. `"auto"` c_str initializer list can be passed as a string to the previous constructor.

I tried to fiddle with it a bit but I failed. There's a lot of abstraction layers handling the types and I'm not that familiar with the code to know what could be changed safely and the only time it compiled for me was with infinite recursion 😵

PR #7008, opened 2025-02-10
feat: CSS animation keyframes registry to attach animations by name

MatiPl01:
## Summary

This PR adds `CSSAnimationKeyframesRegistry` to store animation configuration objects that may be used by multiple views. 

In the previous implementation we created separate style interpolators for each `CSSAnimation` object in CPP, which is created separately for every animation on different views. 

The new implementation registers animation keyframes in the native side just once for the animation, so when the user creates an animation with `css.keyframes` and reuses it on multiple views, the same `AnimationStyleInterpolator` instance will be reused for each of views where the same animation is used.

Inline animation keyframes still work in the same way as before, so even if the same keyframes are used for multiple views, separate animations will be created for these views.

## Test plan

Open the example app and see that everything still works


MatiPl01:
I used an invalid branch name here and closed this PR in favor of: #7015

PR #7012, opened 2025-02-11
Add `createLock` method in `WorkletRuntime`

tomekzaw:
## Summary

Currently, `runOnRuntime` does not implement any synchronization. This leads to crashes when worklet runtime is used directly (as `jsi::Runtime`) on another thread via `->getJSIRuntime()` and `runOnRuntime` is invoked from another thread.

This PR adds `createLock` method that returns a recursive lock. This lock is meant to be kept by each party that uses `jsi::Runtime`, including `runOnRuntime` in react-native-reanimated.

Fixes https://github.com/Expensify/react-native-live-markdown/issues/616.

### DIFF
--- a/packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.h
+++ b/packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.h
@@ -52,6 +52,7 @@ class WorkletRuntime : public jsi::HostObject,
         return;
       }
 
+      const auto lock = strongThis->createLock();
       strongThis->runGuarded(shareableWorklet);
     });
   }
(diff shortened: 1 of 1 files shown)


tomekzaw:
TODO: check if `WorkletRuntime` created by `@expensify/react-native-live-markdown` has `supportLocking` set to true, perhaps we don't need manual synchronization

tomekzaw:
Indeed, changing `supportsLocking` from `false` to `true` fixes the original issue in live-markdown.

```diff
 jsi::Value NativeReanimatedModule::createWorkletRuntime(
     jsi::Runtime &rt,
     const jsi::Value &name,
     const jsi::Value &initializer) {
   auto workletRuntime = std::make_shared<WorkletRuntime>(
       rt,
       jsQueue_,
       jsScheduler_,
       name.asString(rt).utf8(rt),
-      false /* supportsLocking */,
+      true /* supportsLocking */,
       valueUnpackerCode_);
```

Issue #6832, opened 2024-12-18
[iOS] Gestures don't work on child transformed out of the parent bounds when it originally fit inside them


j-piasecki:
### Description

Reanimated not committing non-layout changes to the shadow tree may break gestures on components outside parents (among other things). Here: https://github.com/facebook/react-native/blob/b8f3f919cc9ebbd086d9ac79c93fffd532c55b09/packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm#L620 the insets wouldn't be updated since layout will be skipped.

# Content from the original RNGH issue:

### Description

When gesture detector is attached to a view A which has a child B which may be transformed, if the child starts fully inside the bounds of its parent, the gestures will not work on the child when it's moved outside the bounds. When the child doesn't fit inside the parent's bounds, the gesture will work on it when moved further away.

Child fits fully inside parent's bounds:

https://github.com/user-attachments/assets/82dfd521-05ab-413d-808d-b2064917ebb4

Child extends out of the parent's bounds

https://github.com/user-attachments/assets/7247ae08-1782-476b-ad41-c368e06a4d70

### Steps to reproduce

See the videos, Reanimated used in repro: `3.16.1`

### Snack or a link to a repository

https://gist.github.com/j-piasecki/102a4d821059ae2bb4313207425d09f1

### Gesture Handler version

2.20.2

### React Native version

0.76.5

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

jluisrojas:
Where you able to solve this @j-piasecki ?

j-piasecki:
Unfortunately, I didn't have time to look into this 😞

hamdij0maa:
+1

Issue #6676, opened 2024-11-07
[Android][New Arch] Buttons inside Animated.View with translate are not clickable 


delphinebugner:
### Description

On my app, I have buttons inside Animated Views that translate up. 
Right after the translation finished, the pressable action of the button is not executed on click. 
If we execute another action (ex. navigating on an other tab and come back), the pressable action is executed on click. 

This is what we see on the repro below: 

![ReproButtonReanimated](https://github.com/user-attachments/assets/a36e90a1-023f-49b6-9c13-fab03aafb8fc)

Additional notes: 

- New Arch + Android only
- On **real device** only, not on emulator (😅 really weird)
- Pressable and Buttons have the same error
- Pressable and Buttons variable opacity _are_ triggered when we click the first time; only the `onPress` callback is'nt triggered
- On an animated view with only a variable _opacity_, no error; it's linked to the translation operation

### Code

Code looks like this:
```typescript
export const AnimatedButton = () => {
  const shiftAmount = useSharedValue(-100);
  const shiftDownStyle = useAnimatedStyle(() => ({
    transform: [{ translateY: shiftAmount.value }],
  }));

  if (shiftAmount.value !== 0) {
    shiftAmount.value = withTiming(0, { duration: 450 });
  }

  return (
    <Animated.View style={shiftDownStyle}>
      <Button
        title="Press me"
        onPress={() => Alert.alert("Im clicked")} // 🚨 not triggered the first time
      />
    </Animated.View>
  );
```

A full example can be found here: https://github.com/delphinebugner/repro

### Further notes

Thanks for working on the lib 🙂

If the problem is linked to a messy Reanimated usage in the component, happy to correct my code! 

### Steps to reproduce

On the [repro application](https://github.com/delphinebugner/repro):

1. Run it on a **real** device (error does not occur on emulator)
2. Just after mounting the Home, try to clic on a "Press me" button : nothing happens 🚨 
3. Click on "Explore" Tab
4. Return to "Home" Tab
5. Click on a "Press me" button: an alert shows up ✅ 
6. Click on the "Toggle animated buttons" to unmount-remount them
7. Click again a "Press me" button: nothing happens 🚨 

### Snack or a link to a repository

https://github.com/delphinebugner/repro

### Reanimated version

3.16.1

### React Native version

0.75.0

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

[...]

PR #7014, opened 2025-02-11
Remove non-layout style and prop updates path via `synchronouslyUpdatePropsOnUIThread`

tomekzaw:
## Motivation

Currently, there are two ways to update native view props and styles in Reanimated. The default path (so-called slow path) is to apply all props changes to the ShadowTree via C++ API and let React Native mount the changes. However, if all props updated in given batch are non-layout props (i.e. those that don't require layout recalculation, like background color or opacity) we use a fast path that calls `synchronouslyUpdatePropsOnUIThread` from React Native and applies the changes directly to platform views, without making changes to ShadowTree in C++. Turns out, some features like view measurement or touch detection system use C++ ShadowTree which is not consistent with what's currently on the screen. Because of that, we're removing the fast path (turns out it's not that fast, especially on iOS) to restore the correctness of view measurement and touch detection for animated components.

## Benchmarks

* Performance monitor example &rarr; Bokeh Example
* Android emulator / iPhone 14 Pro real device
* Debug mode
* Animating `transform` prop using `useAnimatedStyle`

| Platform | Before (main) | After (this PR) |
|:-:|:-:|:-:|
| Android (count=200) | 20 fps | 15 fps |
| iOS (count=500) | 22 fps | 22 fps |

<details>
<summary>App.tsx</summary>

```tsx
import { Dimensions, StyleSheet, View } from 'react-native';
import Animated, {
  Easing,
  useAnimatedStyle,
  useReducedMotion,
  useSharedValue,
  withTiming,

const dimensions = Dimensions.get('window');

function randBetween(min: number, max: number) {
  return min + Math.random() * (max - min);
}

function Circle() {
  const shouldReduceMotion = useReducedMotion();

  const [power] = useState(randBetween(0, 1));
  const [duration] = useState(randBetween(2000, 3000));

  const size = 100 + power * 250;
  const width = size;
  const height = size;
  const hue = randBetween(100, 200);
  const backgroundColor = `hsl(${hue},100%,50%)`;

  const opacity = 0.1 + (1 - power) * 0.1;
  const config = { duration, easing: Easing.linear };

  const left = useSharedValue(randBetween(0, dimensions.width) - size / 2);
  const top = useSharedValue(randBetween(0, dimensions.height) - size / 2);

  const update = () => {
    left.value = withTiming(left.value + randBetween(-100, 100), config);
    top.value = withTiming(top.value + randBetween(-100, 100), config);
  };

  React.useEffect(() => {
    update();
    if (shouldReduceMotion) {
      return;
    }
    const id = setInterval(update, duration);
    return () => clearInterval(id);
  });

  const animatedStyle = useAnimatedStyle(
    () => ({
      transform: [{ translateX: left.value }, { translateY: top.value }],
    }),
    []
  );

  return (
    <Animated.View
      style={[
        styles.circle,
        { width, height, backgroundColor, opacity },
        animatedStyle,
      ]}
    />
  );
}

interface BokehProps {
  count: number;
}

function Bokeh({ count }: BokehProps) {
  return (
    <>
      {[...Array(count)].map((_, i) => (
        <Circle key={i} />
      ))}
    </>
  );
}

  return (
      <Bokeh count={200} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    backgroundColor: 'black',
    overflow: 'hidden',
  },
  circle: {
    position: 'absolute',
    borderRadius: 999,
  },
});
```

</details>

## Summary

* Fixes https://github.com/software-mansion/react-native-reanimated/issues/6832.
* Fixes https://github.com/software-mansion/react-native-reanimated/issues/6676.

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
@@ -869,21 +869,11 @@ void ReanimatedModuleProxy::performOperations() {
     jsPropsUpdater.call(rt, viewTag, nonAnimatableProps);
   }
 
-  bool hasLayoutUpdates = false;
 #ifdef ANDROID
   bool hasPropsToRevert = updatesRegistryManager_->hasPropsToRevert();
 #else
   bool hasPropsToRevert = false;
 #endif
-
-  if (!hasPropsToRevert) {
-    for (const auto &[shadowNode, props] : updatesBatch) {
-      if (isThereAnyLayoutProp(rt, props->asObject(rt))) {
-        hasLayoutUpdates = true;
-        break;
-      }
-    }
-  }
 
   if (updatesRegistryManager_->shouldReanimatedSkipCommit()) {
     // It may happen that `performOperations` is called on the UI thread
--- a/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
@@ -885,18 +885,6 @@ void ReanimatedModuleProxy::performOperations() {
     }
   }
 
-  if (!hasLayoutUpdates && !hasPropsToRevert) {
-    // If there's no layout props to be updated, we can apply the updates
-    // directly onto the components and skip the commit.
-    ReanimatedSystraceSection s(
-        "ReanimatedModuleProxy::synchronouslyUpdateUIProps");
-    for (const auto &[shadowNode, props] : updatesBatch) {
-      Tag tag = shadowNode->getTag();
-      synchronouslyUpdateUIPropsFunction_(rt, tag, props->asObject(rt));
-    }
-    return;
-  }
-
   if (updatesRegistryManager_->shouldReanimatedSkipCommit()) {
     // It may happen that `performOperations` is called on the UI thread
     // while React Native tries to commit a new tree on the JS thread.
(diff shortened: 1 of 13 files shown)


tjzel (approved):
LGTM, should we also merge/rework props lists in TS?

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -869,34 +850,6 @@ void ReanimatedModuleProxy::performOperations() {
...
-    // If there's no layout props to be updated, we can apply the updates
-    // directly onto the components and skip the commit.
-    ReanimatedSystraceSection s(
-        "ReanimatedModuleProxy::synchronouslyUpdateUIProps");
-    for (const auto &[shadowNode, props] : updatesBatch) {
-      Tag tag = shadowNode->getTag();
-      synchronouslyUpdateUIPropsFunction_(rt, tag, props->asObject(rt));
-    }
-    return;
-  }
```
j-piasecki, line 898: What's the trick here? As far as I remember, removing this fast path had a negative impact on performance. What changed here?

tomekzaw: Turns out, fast path is not that fast, see Benchmarks in PR description.

@bartlomiejbloniarz confirms that performance of "slow" and "fast" paths are comparable on iOS.

As for Android, we'll run more benchmarks, but still, correctness is more important than performance.

tomekzaw:
> should we also merge/rework props lists in TS?

We'll need to discuss this. There's no need to keep separate sets for UI and native props (since I've merged the logic of applying UI and native props) but I think we still need to expose `addWhitelistedNativeProps` and `addWhitelistedUIProps` for backward-compatibility of third-party libraries built on top on Reanimated, to some extent.

piaskowyk (approved):

tjzel:
@tomekzaw We can keep both these methods and mark one of them as deprecated.

piaskowyk (approved):

bartlomiejbloniarz (approved):

PR #7015, opened 2025-02-11
feat: CSS animation keyframes registry to attach animations by name 

MatiPl01:
## Summary

This PR adds `CSSAnimationKeyframesRegistry` to store animation configuration objects that may be used by multiple views. 

In the previous implementation we created separate style interpolators for each `CSSAnimation` object in CPP, which is created separately for every animation on different views. 

The new implementation registers animation keyframes in the native side just once for the animation, so when the user creates an animation with `css.keyframes` and reuses it on multiple views, the same `AnimationStyleInterpolator` instance will be reused for each of views where the same animation is used.

Inline animation keyframes still work in the same way as before, so even if the same keyframes are used for multiple views, separate animations will be created for these views.

## Test plan

Open the example app and see that everything still works


packages/react-native-reanimated/src/css/registry/CSSKeyframesRegistry.ts
```diff
@@ -0,0 +1,61 @@
...
+} from '../platform/native';
+
+/**
+ * This class is responsible for managing the registry of CSS animation
+ * keyframes. It keeps track of views that use specific animations and handles
+ * native-side registration. Animation keyframes are registered on the native
+ * side only when used for the first time and unregistered when removed from the
+ * last view that uses them.
+ */
+export default class CSSKeyframesRegistry {
```
MatiPl01, line 15: I am removing this registry in the #7027

piaskowyk (commented):
Correct me if I'm wrong, but as I understand it, we currently need to call the JSI function from JS twice during the mounting render, specifically with `registerCSSKeyframes` and `registerCSSAnimations`. I think it might be more efficient to pass the `keyframesConfig` directly to `registerCSSAnimations` and have it call the `registerCSSKeyframes` method internally. This approach could reduce the number of JSI calls, which are relatively expensive. What do you think?

piaskowyk (approved):
After our offline discussion, we've decided to move forward with this approach and optimize certain parts in the future.

PR #7016, opened 2025-02-11
feat: CSS animation shorthand property

MatiPl01:

### DIFF
--- a/packages/react-native-reanimated/src/css/constants/settings.ts
+++ b/packages/react-native-reanimated/src/css/constants/settings.ts
@@ -39,3 +39,9 @@ export const VALID_PREDEFINED_TIMING_FUNCTIONS: PredefinedTimingFunction[] = [
   'step-start',
   'step-end',
 ];
+
+export const VALID_PARAMETRIZED_TIMING_FUNCTIONS: string[] = [
+  'cubic-bezier',
+  'steps',
+  'linear',
+];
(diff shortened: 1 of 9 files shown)


apps/common-app/src/apps/css/examples/animations/screens/testExamples/Playground.tsx
```diff
@@ -32,49 +71,11 @@ export default function Playground() {
...
-          backgroundColor: 'red',
-        },
-      }),
-    ],
-    animationTimingFunction: 'linear',
+    // We can pass just a keyframes object, which will be automatically converted to a string
+    // but eslint complains about that, so we need to disable the rule or use the name property
+    // explicitly
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+    animation: `${resize} 10s infinite linear, ${color.name} 10s infinite linear alternate`,
```
MatiPl01, line 78: I found this information on the [eslint rule](https://typescript-eslint.io/rules/restrict-template-expressions/) documentation page. It looks like they purposely show error for custom objects with a `toString` method.

Screenshot 2025-02-16 at 14 05 54

What we can do is to use the `.name` property explicitly without relying on the implicit `toString()` method call. Alternatively, we can suggest adding the `CSSKeyframesRule` type to the allow list via the `allow` config property (but it didn't work when I tried, because, instead of allowing just our type, it allowed everything, likely because something broke in the rule - not sure if it is a problem with this eslint plugin).

PR #7017, opened 2025-02-11
feat(worklets): AnimationFrameQueue

tjzel:
## Summary

WIP


packages/react-native-worklets/android/src/worklets/main/com/swmansion/worklets/AnimationFrameQueue.java
```diff
@@ -0,0 +1,114 @@
...
+      }
+    }
+  }
+
+  private void executeQueue(long frameTimeNanos) {
+    double currentFrameTimeMs = calculateTimestamp(frameTimeNanos);
+    if (currentFrameTimeMs <= lastFrameTimeMs) {
+      // TODO: Not sure if we should keep it.
+      // It is possible for ChoreographerCallback to be executed twice within the same frame
+      // due to frame drops. If this occurs, the additional callback execution should be ignored.
```
tjzel, line 82: I wonder if this is fixed.

packages/react-native-worklets/Common/cpp/worklets/AnimationFrameQueue/AnimationFrameBatchinator.cpp
```diff
@@ -0,0 +1,80 @@
...
+      });
+}
+
+std::vector<std::shared_ptr<const facebook::jsi::Value>>
+AnimationFrameBatchinator::pullCallbacks() {
+  std::lock_guard<std::mutex> lock(callbacksMutex_);
+  auto callbacks = std::move(callbacks_);
+  callbacks_.clear();
+  return callbacks;
+}
```
piaskowyk, line 71: You can use here the same trick what I suggested for Same think as in `AnimationFrameQueue`

packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
```diff
@@ -26,7 +28,9 @@ WorkletsModuleProxy::WorkletsModuleProxy(
     const std::shared_ptr<MessageQueueThread> &jsQueue,
     const std::shared_ptr<CallInvoker> &jsCallInvoker,
     const std::shared_ptr<JSScheduler> &jsScheduler,
-    const std::shared_ptr<UIScheduler> &uiScheduler)
+    const std::shared_ptr<UIScheduler> &uiScheduler,
+    std::weak_ptr<std::function<void(std::function<void(const double)>)>>
```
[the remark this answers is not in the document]
tjzel: We can't use a unique pointer, because we need to lock it while we actually use this function.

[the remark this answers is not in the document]
piaskowyk: I'm not sure if I understand correctly. Do you mean a lock in terms of multithreading synchronization?

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -0,0 +1,72 @@
...
+  [displayLink_ setPaused:TRUE];
+
+#if TARGET_OS_OSX
+  auto targetTimestamp = displayLink.timestamp + displayLink.duration;
+#else // TARGET_OS_OSX
+  auto targetTimestamp = displayLink.targetTimestamp;
+#endif // TARGET_OS_OSX
+  targetTimestamp = worklets::calculateTimestampWithSlowAnimations(targetTimestamp);
+
+  for (auto callback : frameCallbacks) {
```
piaskowyk, line 61:
```suggestion
  for (const auto &callback : [self pullCallbacks]) {
```

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -0,0 +1,72 @@
...
+    callback(targetTimestamp);
+  }
+}
+
+- (std::vector<std::function<void(double)>>)pullCallbacks
```
piaskowyk, line 66:
```suggestion
- (const std::vector<std::function<void(double)>>)pullCallbacks
```

packages/react-native-worklets/Common/cpp/worklets/AnimationFrameQueue/AnimationFrameBatchinator.cpp
```diff
@@ -0,0 +1,78 @@
...
+        const auto strongThis = weakThis.lock();
+        if (!strongThis) {
+          return;
+        }
+
+        auto callbacks = strongThis->pullCallbacks();
+        strongThis->flushRequested_ = false;
+
+        auto &uiRuntime = *(strongThis->uiRuntime_);
+        for (auto &callback : callbacks) {
```
piaskowyk, line 59:
```suggestion
        for (const auto &callback : strongThis->pullCallbacks()) {
```

packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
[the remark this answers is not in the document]
tjzel: I revised the approach in [4ccceac](https://github.com/software-mansion/react-native-reanimated/pull/7017/commits/4ccceac723e3caa68f9f292e6f4d5f8d51906454) after discussing this offline.

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
tjzel: In my opinion there's no reason to add `const` here as we return values with `std::move`.

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
piaskowyk: Okay, but at least let's use a reference to avoid potential copying of callbacks during iteration
```
for (const auto &callback : frameCallbacks) {
```

packages/react-native-worklets/Common/cpp/worklets/AnimationFrameQueue/AnimationFrameBatchinator.cpp
piaskowyk: ok, but lets add `const`

piaskowyk (approved):
Now it looks much better 😊 I've just left two small comments, but overall, the PR is in good shape, so I approve it

PR #7019, opened 2025-02-11
fix: CSS transition shorthand property behavior in react-native-web

MatiPl01:
## Summary

This PR fixes CSS transition properties being overridden by `react-native-web`. It happened because we didn't filter out CSS props from the style object passed to the component in the `render()` method of the `AnimatedComponent`.

I also made a few other changes to ensure that the transition shorthand is validated on web as well and preprocessed in a similar way to the native implementation. 

## Example Recordings

In the previous implementation, the `transitionTimingFunction` prop specified before the `transition` property shorthand was overriding the `transition` prop easing. Web also shown an error in the console that shorthand props shouldn't be mixed with longhand ones. We don't want this error to show up and want to support usage of both at the same time.

### Web

| Before | After |
|-|-|
| <video src="https://github.com/user-attachments/assets/59edb895-5f5d-4202-bc8c-8080166251a9" /> | <video src="https://github.com/user-attachments/assets/bfb052c0-60e7-4431-a8b8-58c85e43da4c" /> |

### iOS

This recording is just for comparison to see that the behavior on web is also correct (the same as on the recording below) after changes implemented in this PR

https://github.com/user-attachments/assets/6fb73423-28a0-4822-8a5d-5dc08e7e3bd6

<details>
  <summary>Source code</summary>

```tsx
/* eslint-disable perfectionist/sort-objects */
import React, { useReducer } from 'react';
import { createAnimatedComponent, steps } from 'react-native-reanimated';

const AnimatedView = createAnimatedComponent(View);

  const [state, toggleState] = useReducer((s) => !s, false);

  return (
      <AnimatedView
        style={{
          width: state ? 200 : 100,
          height: state ? 200 : 100,
          transform: [{ rotate: state ? '45deg' : '0deg' }],
          transitionTimingFunction: steps(2), // overridden by the shorthand
          backgroundColor: 'red',
          transition:
            '5s cubic-bezier(0, -0.54, 1, -0.5), transform 2s ease-in',
          transitionDuration: 1000, // overrides the shorthand
        }}
      />
      <Button title="Toggle width" onPress={toggleState} />
    </View>
  );
}

  container: {
    alignItems: 'center',
    flex: 1,
    justifyContent: 'center',
  },
});
```
</details>


packages/react-native-reanimated/src/css/platform/native/normalization/transition/config.ts
```diff
@@ -21,76 +21,86 @@ import {
...
-    transitionProperty: CSSTransitionProperty | undefined
+    transitionProperty: CSSTransitionProperty | undefined | string[]
   ) => `Invalid transition property "${JSON.stringify(transitionProperty)}"`,
 };
 
 function getExpandedConfigProperties(
   config: CSSTransitionProperties
-): ExpandedConfigProperties {
-  const configEntries = Object.entries(config);
-  const shorthandIndex = config.transition
```
MatiPl01, line 37: I moved this removal of transition properties listed before the `transition` shorthand property to the `filterCSSAndStyleProperties` because it simplifies the logic and removing invalid properties as soon as possible seems to be a better solution for me.

packages/react-native-reanimated/src/css/platform/web/normalization/transition.ts
```diff
@@ -0,0 +1,59 @@
...
+        result.transitionTimingFunction ?? 'ease'
+      );
+      acc.transitionDelay.push(
+        result.transitionDelay ? String(result.transitionDelay) : '0s'
+      );
+      acc.transitionBehavior.push(result.transitionBehavior ?? 'normal');
+      return acc;
+    },
+    createEmptyTransitionConfig()
+  );
```
MatiPl01, line 44: I decided to parse the transition shorthand on web as well in order to:
1. Ensure that behavior is always the same as in the mobile implementation
2. Because `parseSingleTransitionShorthand` already validates a correctness of the transition property shorthand

piaskowyk (approved):
I haven't tested it, but I didn't find anything controversial in these changes, so I can approve it 👍

PR #7020, opened 2025-02-12
fix: lack of `transform` in keyframes on web caused crash

### LINKED ISSUE #6951: [Web] Cannot read properties of undefined (reading 'map')
### Description

Hello! I added an entering animation on the web: 
```
SlideInRight.delay(0)
    .duration(400)
    .withInitialValues({
    	originX: window.width + 50,
    })
```

If I don't add transform, I get an error.:

Image

After adding transform, it helped resolve the issue: 
```
SlideInRight.delay(0)
.duration(400)
.withInitialValues({
	originX: window.width + 50,
	transform: [],
})
```

As I understand it, `addPxToTransform` expects an array: 
https://github.com/software-mansion/react-native-reanimated/blob/1c3639eb6cfa7c6925659c801d9780605d51e758/packages/react-native-reanimated/src/layoutReanimation/web/createAnimation.ts#L77-L85

https://github.com/software-mansion/react-native-reanimated/blob/1c3639eb6cfa7c6925659c801d9780605d51e758/packages/react-native-reanimated/src/layoutReanimation/web/createAnimation.ts#L27-L51

https://github.com/software-mansion/react-native-reanimated/pull/6254

Is the `transform` prop required, or am I missing something?

### Steps to reproduce

1. Add entering animation without 
2. Open web

https://snack.expo.io/

### Reanimated version

~3.16.1

### React Native version

0.76.6

### Platforms

Web

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

Real device

### Device model

_No response_

### Acknowledgements

Yes

patrycjakalinska:
## Summary

Fixes #6951

`Cannot read properties of undefined (reading 'map')` error crashed a site every time user called entering or exiting animation w/o `transform` prop. It was caused by calling `addPxToTransform` (and therefore calling `map`), before checking if `transform` is not an empty prop.

## Test plan

<details>
<summary>minimal repro to run on web</summary>

```tsx
import { View, Pressable, Text, Dimensions } from 'react-native';
import Animated, { SlideInRight } from 'react-native-reanimated';

const { width } = Dimensions.get('window');

  const [visible, setVisible] = useState(false);

  return (
    <View
      style={{
        flex: 1,
        justifyContent: 'center',
        alignItems: 'center',
        minHeight: 200,
      }}>
      {visible && (
        <Animated.View
          entering={SlideInRight.delay(0)
            .duration(400)
            .withInitialValues({ originX: width + 50 })}
          style={{
            width: 100,
            height: 100,
            backgroundColor: '#82cab2',
            borderRadius: 20,
            position: 'absolute',
          }}
        />
      )}
      <Pressable onPress={() => setVisible((prev) => !prev)}>
        <Text
          style={{
            color: 'white',
            backgroundColor: '#b58df1',
            paddingHorizontal: 16,
            paddingVertical: 8,
            borderRadius: 10,
          }}>
          Toggle
        </Text>
      </Pressable>
    </View>
  );
}


```

</details>


m-bert (changes requested):
Correct me if I'm wrong, but it is enough to move this line inside 

```js
if (transform) {
...
```

instead of adding this check 😅

patrycjakalinska:
@m-bert yes...

m-bert (approved):

PR #7021, opened 2025-02-13
feat: prevent flattening style arrays

jpudysz:
### Why

While integrating `Unistyles` with `Reanimated`, there is one clash that I would love to resolve.  
We encourage developers to merge styles using array syntax, as it makes it easier to detect the order of merging and persist the C++ state attached to styles.  

Reanimated flattens arrays passed to the `style` prop, making it impossible to distinguish which props are animated and which are provided by `Unistyles`.  

### How  

To fix this, we just need to remove ~two~ one occurrence~s~ of array flattening. I tested it locally with `Unistyles`, and I can correctly retrieve the `style` prop as-is.  

`Unistyles` have to retrieve `ref` + `style` prop while [borrowing ref](https://www.unistyl.es/v3/other/babel-plugin#3-component-factory-borrowing-ref).

Additionally, with this change, I can treat `Reanimated` styles as inline styles. Since inline styles are not processed by `Unistyles` algorithm, this can completely eliminate issues such as preventing animations when switching themes or changing device orientation.

## Test plan

I tested it locally and I'm happy with the outcome. But I do understand that you have own test suites and regression tests, so I hope this won't break anything.


jpudysz (commented):
Added one more check for objects

jeremybarbet:
I tried to apply the changes locally and had some styles not being applied anymore. Since the `flattenArray` is recursively going through the array of styles, I believe I have more than one-depth level on my array.

However, I tried with the following:

```ts
// Kept as before
const styles = flattenArray<StyleProps>(styleProp ?? []);

// Only apply this change
props[key] = processedStyle;
```

And I didn't have any warning anymore, and the regression I saw previously didn't happen in this case. Did you try that too?

jpudysz:
@jeremybarbet nevertheless you're right! It's just a one liner and I can still get Array on my side. Thanks for the verification 🙏🏼 I will update PR.

jpudysz:
I updated few tests to indicate that we may have style arrays now

packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx
```diff
@@ -53,7 +51,9 @@ export class PropsFilter implements IPropsFilter {
             return style;
           }
         });
-        props[key] = StyleSheet.flatten(processedStyle);
+        // keep styles as they were passed by the user
+        // it will help other libs to interpret styles correctly
+        props[key] = processedStyle;
```
tjzel, line 56: Looks good to me. However I'm afraid it could break some Jest flows for our users therefore I think we shouldn't backport it to Reanimated 3.

tjzel (approved):

packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx
tjzel: Also please address the CI issues 🙏

jpudysz: on it

jpudysz:
I think these warnings are not from my PR. I didn't use `any` or touch these files. 🤔

piaskowyk (approved):
I've tested it, and it looks good 👍

piaskowyk:
To fix CI run:
```
cd packages/react-native-reanimated && yarn formatter
```
then we can merge it 😄

PR #7027, opened 2025-02-14
chore: Use hostObject for css keyframes representation in CPP

MatiPl01:
## Summary

This PR removes the old JS-side animation keyframes registry and replaces it by the use of `jsi::HostObject` instance created on the CPP side. Thanks to this, we can connect the JS object with the native one and clear the keyframes object from the native-side keyframes registry once the JS keyframes object is garbage collected. This is the cleanest (in terms of implementation) and the safest approach.

I am using `jsi::HostObject` instead of `jsi::NativeState` to allow adding callable methods on the cpp object in the future (e.g. to update keyframes of already existing animation - this is possible on web).

## Test plan

Open Xcode or Android Studio and set breakpoint on the cleanup callback in the `ReanimatedModuleProxy.cpp` file. Then navigate back and forth between animation screens (a screen with **Spinners and Loaders** is the best for testing as it has multiple animations). You will see that after some time this callback is called to clean up data related to garbage collected animations.

Screenshot 2025-02-16 at 00 22 42

I tested it on my side by printing the total number of added and removed animation and a difference between the 2 wasn't changing after some time, so it looks that everything is cleaned up properly.


packages/react-native-reanimated/Common/cpp/reanimated/CSS/models/CSSKeyframesImpl.cpp
```diff
@@ -0,0 +1,35 @@
...
+      styleInterpolator_(config.styleInterpolator),
+      cleanupCallback_(std::move(cleanupCallback)) {}
+
+CSSKeyframesImpl::~CSSKeyframesImpl() {
+  cleanupCallback_(animationName_);
+}
+
+const std::string &CSSKeyframesImpl::getAnimationName() const {
+  return animationName_;
+}
```
tjzel, line 21: I'm not sure if returning a reference to a string is a good idea. Strings are optimized for copying anyway.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/models/CSSKeyframesImpl.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+const std::shared_ptr<KeyframeEasingFunctions> &
+CSSKeyframesImpl::getKeyframeEasingFunctions() const {
+  return keyframeEasingFunctions_;
+}
+
+const std::shared_ptr<AnimationStyleInterpolator> &
+CSSKeyframesImpl::getStyleInterpolator() const {
+  return styleInterpolator_;
+}
```
tjzel, line 31: Similar story goes for returning references to shared pointers. However in this case I'll let this slide since this is (unfortunately) a recurring pattern in our codebase.

tjzel, line 31: Can we make those `[[nodiscard]] inline`?

packages/react-native-reanimated/Common/cpp/reanimated/CSS/core/CSSAnimation.cpp
```diff
@@ -9,22 +9,29 @@ CSSAnimation::CSSAnimation(
     jsi::Runtime &rt,
     ShadowNode::Shared shadowNode,
     const unsigned index,
-    const CSSKeyframesConfig &keyframesConfig,
+    const std::weak_ptr<CSSKeyframes> &keyframes,
```
tjzel, line 12:
```suggestion
    const std::weak_ptr<CSSKeyframes> &weakKeyframes,
```

packages/react-native-reanimated/Common/cpp/reanimated/CSS/core/CSSAnimation.cpp
```diff
@@ -9,22 +9,29 @@ CSSAnimation::CSSAnimation(
...
-          timestamp,
-          settings.duration,
-          settings.delay,
-          settings.iterationCount,
-          settings.direction,
-          settings.easingFunction,
-          keyframesConfig.keyframeEasingFunctions)),
-      styleInterpolator_(keyframesConfig.styleInterpolator) {
+      fillMode_(settings.fillMode) {
+  auto keyframesLocked = keyframes.lock();
```
tjzel, line 19:
```suggestion
  auto keyframes = weakKeyframes.lock();
```

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSKeyframesRegistry.cpp
```diff
@@ -14,10 +14,13 @@ const CSSKeyframesConfig &CSSKeyframesRegistry::get(
   return it->second;
 }
 
-void CSSKeyframesRegistry::add(
-    const std::string &animationName,
-    CSSKeyframesConfig &&config) {
-  registry_[animationName] = std::move(config);
+void CSSKeyframesRegistry::add(const std::weak_ptr<CSSKeyframes> &rule) {
```
tjzel, line 17:
```suggestion
void CSSKeyframesRegistry::add(const std::weak_ptr<CSSKeyframes> &weakRule) {
```

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSKeyframesRegistry.cpp
```diff
@@ -14,10 +14,13 @@ const CSSKeyframesConfig &CSSKeyframesRegistry::get(
...
+  if (auto sharedRule = rule.lock()) {
```
tjzel, line 18:
```suggestion
  if (auto rule = weakRule.lock()) {
```

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -580,20 +580,26 @@ void ReanimatedModuleProxy::removeViewStyle(
...
-void ReanimatedModuleProxy::unregisterCSSKeyframes(
-    jsi::Runtime &rt,
-    const jsi::Value &animationName) {
-  cssAnimationKeyframesRegistry_->remove(animationName.asString(rt).utf8(rt));
+  const auto cleanupCallback =
+      [weakThis = weak_from_this()](const std::string &removedAnimationName) {
+        auto strongThis = weakThis.lock();
+        if (strongThis) {
+          strongThis->cssKeyframesRegistry_->remove(removedAnimationName);
+        }
```
tjzel, line 595:
```suggestion
        if (!strongThis) {
        	return;
        }
        strongThis->cssKeyframesRegistry_->remove(removedAnimationName);
```

packages/react-native-reanimated/src/css/models/CSSKeyframesRule.ts
```diff
@@ -1,17 +1,31 @@
...
 import CSSKeyframesRuleBase from './CSSKeyframesRuleBase';
 
 export default class CSSKeyframesRuleImpl<
   S extends PlainStyle = PlainStyle,
 > extends CSSKeyframesRuleBase<S> {
   private normalizedKeyframes_: NormalizedCSSAnimationKeyframesConfig;
+  // For now, this object is used only for object cleanup in cpp
+  // (we may add a possibility to modify the cpp keyframes object
+  // in the future, as we can do in the web animations api)
+  private hostObject_: CSSKeyframesHostObject;
```
tjzel, line 20: Use `#hostObject` syntax instead for private members. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

packages/react-native-reanimated/src/css/models/CSSKeyframesRule.ts
```diff
@@ -1,17 +1,31 @@
...
 import type { CSSAnimationKeyframes, PlainStyle } from '../types';
 import CSSKeyframesRuleBase from './CSSKeyframesRuleBase';
 
 export default class CSSKeyframesRuleImpl<
   S extends PlainStyle = PlainStyle,
 > extends CSSKeyframesRuleBase<S> {
   private normalizedKeyframes_: NormalizedCSSAnimationKeyframesConfig;
+  // For now, this object is used only for object cleanup in cpp
+  // (we may add a possibility to modify the cpp keyframes object
+  // in the future, as we can do in the web animations api)
```
tjzel, line 19: Change this to TSDoc so it would pop up in an editor pop ups.

packages/react-native-reanimated/src/css/platform/native/types/animation.ts
```diff
@@ -45,3 +45,7 @@ export type NormalizedSingleCSSAnimationSettings = {
   fillMode: CSSAnimationFillMode;
   playState: CSSAnimationPlayState;
 };
+
+// HOST OBJECTS
+
+export type CSSKeyframesHostObject = Record<string, never>; // Placeholder type as we don't need any properties from the host object for now
```
tjzel, line 51: Signing utility types is a good practice.

```suggestion
/**
* Placeholder type as we don't need any properties from the host object for now
*/
export type CSSKeyframesHostObject = Record<string, never> & "CSSKeyframesHostObject"
```

tjzel (changes requested):
Looks good, let's just improve overall quality of the code.

packages/react-native-reanimated/src/css/managers/__tests__/CSSAnimationsManager.test.ts
```diff
@@ -14,12 +14,11 @@ import type {
 import type { ProcessedAnimation } from '../CSSAnimationsManager';
 import CSSAnimationsManager from '../CSSAnimationsManager';
 
-jest.mock('../../platform/native/native.ts', () => ({
```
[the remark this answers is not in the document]
MatiPl01: I saw that we used this convention in other tests, e.g. `packages/react-native-reanimated/__tests__/sensors.test.ts`

```js
jest.mock('../src/core', () => {
  const originalModule = jest.requireActual('../src/core');
...
```

packages/react-native-reanimated/Common/cpp/reanimated/CSS/models/CSSKeyframesImpl.cpp
MatiPl01: What about a `std::string_view`? I was thinking about using this one but I don't know when they should be used.
Anyway, would returning a reference be slower than copying? I thought that copying is almost always slower (except primitive values like numbers).

packages/react-native-reanimated/Common/cpp/reanimated/CSS/models/CSSKeyframesImpl.cpp
MatiPl01: Yup, I can agree that in this case we don't benefit from returning a reference. We should set up clang-tidy to check for code consistency and add some rules to detect something like this.

MatiPl01: I think that we don't have to explicitly add `inline` here. As far as I know, the compiler does great job in inlining functions that can be optimized in this way and we don't have to try to help compiler do this.

I personally think that code with `[[nodiscard]]` looks ugly and I would prefer not to add this, at least for now, because we hardly ever use `[[nodiscard]]` and there are multiple places in which it could be added. I'd like to keep these changes separate from this logic and add `[[nodiscard]]` where it can be added in a single PR, ideally when we have clang-format enabled with a rule that checks for a presence of this.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/core/CSSAnimation.cpp
MatiPl01: Do we really want to include `weak` in the variable name? We don't do this for shared pointers so I assumed that the convention is the same for weak pointers.

I can prefix it with `weak` if you wish but I just wonder which convention we should follow and when.

packages/react-native-reanimated/src/css/models/CSSKeyframesRule.ts
MatiPl01: What about `normalizedKeyframes_`? Should I change this as well to `#normalizedKeyframes`?

packages/react-native-reanimated/src/css/models/CSSKeyframesRule.ts
MatiPl01: I will change that but well, we don't expose the `hostObject` anywhere and this is just a temporary comment, so I don't think it is a big deal.

packages/react-native-reanimated/src/css/platform/native/types/animation.ts
MatiPl01: Similar remark as [here](https://github.com/software-mansion/react-native-reanimated/pull/7027#discussion_r1958377618). This is not exposed anywhere, so I don't think that it must be well-documented but ok, will change that as well.

MatiPl01:
Closing this PR as it is not maintainable in its current state

PR #7043, opened 2025-02-18
refactor: RN$Bridgeless as isFabric

tjzel:
## Summary

Since Bridgeless by default `RN$Bridgeless` global property is injected before any library code is executed. We don't support "Bridgeful" mode anymore so we can depend on this variable to detect the architecture.

## Test plan

See that `console.log(globalThis.RN$Bridgeless);` yields `true` when added in top-level scope of any file.


packages/react-native-reanimated/plugin/src/globals.ts
```diff
@@ -135,7 +135,6 @@ const notCapturedIdentifiers = [
 // eslint-disable-next-line camelcase
 const notCapturedIdentifiers_DEPRECATED = [
   // Reanimated
-  '_IS_FABRIC',
```
tomekzaw, line 138: Perhaps we should leave it here for the sake of backward-compatibility?

tomekzaw (approved):

piaskowyk (approved):

packages/react-native-reanimated/plugin/src/globals.ts
tjzel: I double checked and you are right, there is some dependency:

https://github.com/software-mansion/react-native-reanimated/discussions/5118
https://github.com/search?q=repo%3Asoftware-mansion%2Freact-native-screens%20_IS_FABRIC&type=code

I made:
- https://github.com/software-mansion/react-native-screens/pull/2722

I'll restore it for the time being.

PR #7045, opened 2025-02-18
fix(worklets): compilation on iOS

tjzel:
## Summary

Turns out setting CIs is hard. Previously because of `== 'true'` condition the desired steps were falsely skipped.

## Test plan

See that CI are green now.

### DIFF
--- a/packages/react-native-worklets/RNWorklets.podspec
+++ b/packages/react-native-worklets/RNWorklets.podspec
@@ -1,7 +1,9 @@
 require "json"
+require_relative './scripts/worklets_utils'
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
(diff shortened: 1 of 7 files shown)


packages/react-native-worklets/RNWorklets.podspec
```diff
@@ -1,7 +1,11 @@
 require "json"
+require_relative './scripts/worklets_utils'
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
-folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
+$worklets_config = find_config()
```
bartlomiejbloniarz, line 5: maybe `worklets_find_config`

bartlomiejbloniarz (approved):

PR #7053, opened 2025-02-19
Fix react-freeze

bartlomiejbloniarz:
## Summary

This PR is very annoying. Until now we assumed that when `componentWillUnmount` is called, the component will never reappear. This is false when using `<Suspense>` ([docs](https://react.dev/reference/react/Suspense#caveats)), which is exactly what `react-freeze` does. To fix the issue I had to move our animation cleanup functions to be called only when we are sure that the component will not reappear. This is how the new approach works for each registry:
- `AnimatedPropsRegistry`: we remove a record from this registry only when we fail to apply it onto the ShadowTree (in the commit hook or `perfromOperations`)
- `CSSAnimationsRegistry`: we remove a record from this registry only when we fail to apply it onto the ShadowTree (in the commit hook or `perfromOperations`)
- `CSSTransitionsRegistry`: we remove a record from this registry when `componentWillUnmount` is called and there is no ongoing transition OR we failed to apply it onto the ShadowTree (in the commit hook or `perfromOperations`)
- `StaticPropsRegistry` remains the same as it doesn't store any state that needs to be persisted when freeze is used

One problem with this approach is that we are running animations on components that are freezed. This is a waste of resources. This will be addressed in a separate PR, but the main idea is to pause animations in `componentWillUnmount` and resume them (with the new timestamp) in `compnentDidMount`.

Also currently the leak check will say that `AnimatedPropsRegistry` leaked if we don't reload it. This is because an animation frame comes after the cleanup from the commit hook. Normally this would be cleaned up in `performOperations`, but because of the `sunchronouslyUpdateProps` fast-path it won't. The cleanup will still happen on the next commit hook invocation (i.e. if you hit reload on the leak check). This issue soon will be gone as we are removing the fast path in #7014.

| Before | After |
| -------- | ------- |
| <video src="https://github.com/user-attachments/assets/249acaac-f1f0-4846-a38e-4dab54cfb49b"/> | <video src="https://github.com/user-attachments/assets/5d3c578b-fb80-45e8-9955-542bf47d0573"/> |

## Test plan

Go through the app, and then visit the `FreezeExample`. Use the `check for registry leaks` button.


packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/UIRuntimeDecorator.cpp
```diff
@@ -7,9 +7,7 @@ using namespace worklets;
 
 void UIRuntimeDecorator::decorate(
     jsi::Runtime &uiRuntime,
-#ifdef RCT_NEW_ARCH_ENABLED
-    const RemoveFromPropsRegistryFunction removeFromPropsRegistry,
-#else
+#ifndef RCT_NEW_ARCH_ENABLED
```
tjzel, line 10:
```suggestion
#ifdef RCT_NEW_ARCH_ENABLED
// nothing
#else
```

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/UIRuntimeDecorator.h
```diff
@@ -15,9 +15,7 @@ class UIRuntimeDecorator {
  public:
   static void decorate(
       jsi::Runtime &uiRuntime,
-#ifdef RCT_NEW_ARCH_ENABLED
-      const RemoveFromPropsRegistryFunction removeFromPropsRegistry,
-#else
+#ifndef RCT_NEW_ARCH_ENABLED
```
tjzel, line 18:
```suggestion
#ifdef RCT_NEW_ARCH_ENABLED
// nothing
#else
```

packages/react-native-reanimated/src/privateGlobals.d.ts
```diff
@@ -54,7 +54,6 @@ declare global {
         }[]
       ) => void)
     | undefined;
-  var _removeFromPropsRegistry: (viewTags: number[]) => void | undefined;
```
tjzel, line 57: Add the leak check thing here.

apps/common-app/src/components/LeakCheck.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Button, Text, View } from 'react-native';
+
+import { DropContext } from '@/App';
+
+function registriesLeakCheck(): string {
+  if ('_registriesLeakCheck' in global) {
+    // eslint-disable-next-line no-underscore-dangle
+    return (global._registriesLeakCheck as () => string)();
+  }
+  return '';
```
tjzel, line 11: return global?._registriesLeakCheck() ?? '';

apps/common-app/src/App.tsx
```diff
@@ -15,11 +15,23 @@ import { colors, flex, radius, text } from '@/theme';
 import { IS_MACOS, IS_WEB, isFabric, noop } from '@/utils';
 
 import { CSSApp, ReanimatedApp } from './apps';
+import { LeakCheck } from './components/LeakCheck';
+
+export const DropContext = createContext<() => void>(() => '');
 
 export default function App() {
+  const [droped, setNuked] = useState(false);
```
tjzel, line 23:
```suggestion
  const [nuked, setNuked] = useState(false);
```

tjzel (approved):

tjzel:
🥶

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/UIRuntimeDecorator.cpp
```diff
@@ -8,7 +8,7 @@ using namespace worklets;
 void UIRuntimeDecorator::decorate(
     jsi::Runtime &uiRuntime,
 #ifdef RCT_NEW_ARCH_ENABLED
-    const RemoveFromPropsRegistryFunction removeFromPropsRegistry,
+// nothing
```
MatiPl01, line 11: I know that this is already merged but can't we use this instead?
```cpp
#ifndef RCT_NEW_ARCH_ENABLED
  const ScrollToFunction scrollTo,
#endif
```

packages/react-native-reanimated/src/css/component/AnimatedComponent.tsx
```diff
@@ -151,7 +151,9 @@ export default class AnimatedComponent<
     this._updateStyles(this.props);
 
     if (!IS_JEST) {
-      this._CSSManager = new CSSManager(this._getViewInfo());
+      if (!this._CSSManager) {
```
MatiPl01, line 154: As far as I understand, `componentDidMount` is called when the component becomes unfrozen and it prevents from creating `_CSSManager` if it already exists?

nit: it might be written a bit shorter, without nesting if statements
```ts
this._CSSManager ??= new CSSManager(this._getViewInfo());
```

packages/react-native-reanimated/src/css/managers/CSSManager.ts
```diff
@@ -58,7 +58,6 @@ export default class CSSManager {
   }
 
   detach(): void {
-    this.cssAnimationsManager.detach();
```
MatiPl01, line 61: why? You haven't removed `detach` method from the `CSSAnimationsManager`, so I assume this is a mistake, right?

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/UIRuntimeDecorator.h
```diff
@@ -16,7 +16,7 @@ class UIRuntimeDecorator {
   static void decorate(
       jsi::Runtime &uiRuntime,
 #ifdef RCT_NEW_ARCH_ENABLED
-      const RemoveFromPropsRegistryFunction removeFromPropsRegistry,
+// nothing
```
MatiPl01, line 19: as above

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
```diff
@@ -27,6 +27,9 @@ void CSSTransitionsRegistry::add(
 
 void CSSTransitionsRegistry::remove(const Tag viewTag) {
   std::lock_guard<std::mutex> lock{mutex_};
+  if (updatesRegistry_.contains(viewTag)) {
+    return;
+  }
```
MatiPl01, line 32: It will be almost always `true`. Since you added `removeBatch` that removes things based on presence of respective ShadowNodes in the ShadowTree I wonder if the `remove` method is still needed.

What will happen in a scenario in which the user removes transition from the view (e.g. replaces `transitionProperty` with `undefined`)?
In my opinion, it should remove the transition from the registry but, as far as I understand this implementation logic, it will just ignore the `remove` method call and the transition won't be removed until the ShadowNode is removed from the ShadowTree. Because `removeFromUpdatesRegistry(viewTag)` is not called, we will keep values calculated previously by transition and prevent react from applying its changes when a view is re-rendered with different style props.

MatiPl01: I have just tested it in the example app and it looks like my concerns were correct. I compared what happens on app one commit before this one and on this commit. You can see that after changes applied in this PR the view keeps being animated after its `transitionProperty` was set to `none`:

| at `260f873c1f` (one before) | at `b4d977f4e1` (this PR's commit) |
|-|-|
| <video src="https://github.com/user-attachments/assets/19e561e5-6fcd-4c56-8806-ea1a568f9911" /> | <video src="https://github.com/user-attachments/assets/dfae1a42-0e5a-4ea2-8809-cff232d9aa66" /> |

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/UIRuntimeDecorator.cpp
tomekzaw: We can, but this is often misleading, `ifndef` and `ifdef` look similar, we deliberately use `#ifdef` with `// nothing` inside.

packages/react-native-reanimated/src/css/component/AnimatedComponent.tsx
tjzel: Didn't know there's a nullish-coalescent assignment operator in JavaScript 🤯

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/UIRuntimeDecorator.cpp
bartlomiejbloniarz: This was on purpose. `#ifndef` can confuse people.

packages/react-native-reanimated/src/css/component/AnimatedComponent.tsx
bartlomiejbloniarz: Yes, I want to reuse the CSSManager if the component is not permanently removed from the tree.

packages/react-native-reanimated/src/css/managers/CSSManager.ts
bartlomiejbloniarz: This was on purpose. We cleanup css animations when we fail to apply them in performOperations/commit hook. I couldn't remove the detach method as it is used in `update` to remove an animation.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
bartlomiejbloniarz: This is a skill issue on my part

bartlomiejbloniarz: My logic was that I want to trigger `CSSTransition` cleanup when:
- the component freezed and there is no transition running (that's why I added this condition there - to allow cleanup triggered by `.detach()` when there is no animation)
- the component got removed - so when we fail to apply

bartlomiejbloniarz: But obviously this condition doesn't make sense when the animation is running but it was the user that tried to remove it.

bartlomiejbloniarz: One way to fix that would be to separate the user cleanup from `CSSManager.detach()` cleanup.

bartlomiejbloniarz: But maybe you have a cleaner idea?

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/UIRuntimeDecorator.cpp
MatiPl01: Ok, so if this is our convention, we should definitely get rid of all remaining `#ifndef`s that we use in the native code and use `#ifdef` in all places instead. We have currently more than 80 usages of `#ifndef`, so I assumed that we can use it as well.

Screenshot 2025-02-27 at 11 55 46

It would be very helpful if we create guidelines or set up clang-tidy to make it obvious which convention should be followed.

packages/react-native-reanimated/src/css/managers/CSSManager.ts
MatiPl01: Ok, so this just changes the cleanup behavior when the view is removed. It makes sense now.
I have another question. Do we have to keep these 2 remaining calls here:

```tsx
this.cssTransitionsManager.detach();
removeViewStyle(this.viewTag);
```

As far as I understand, they also will be cleared when the node is removed from the ShadowTree, so it doesn't seem to be necessary to clean them up manually anymore.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
MatiPl01: > One way to fix that would be to separate the user cleanup from CSSManager.detach() cleanup.

I think [this](https://github.com/software-mansion/react-native-reanimated/pull/7053#discussion_r1973366153) comment I added is a bit related. I am wondering if we still need to call `detach` manually in `componentWillUnmount`, since this cleanup was meant to run when the view is removed but you already implemented the logic that cleans up animations and transitions when there is no respective node in the ShadowTree.

MatiPl01: So maybe we can simply remove the `detach` method from the `CSSManger` at all?

packages/react-native-reanimated/Common/cpp/reanimated/RuntimeDecorators/UIRuntimeDecorator.cpp
MatiPl01: There were more other usages than just `#ifndef NDEBUG` but I had an old main and it looks like they were recently replaced with `#ifdef`. I am now good with that, I just wondered why we care about that while we are using `#ifndef` in multiple other places but this is no longer a case.

tomekzaw: `#ifndef NDEBUG` is okay, `#ifndef RCT_NEW_ARCH_ENABLED` is not okay.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
bartlomiejbloniarz: We can't because if there is no transition running, the cleanup would not happen

MatiPl01: Why? Does it happen because `cloneShadowTreeWithNewProps` receives only these nodes in the `childrenMap` which have an animation/transition running or had transition/animation applied before (and have previous updates stored in the registry)?

If I am correct, I can understand that this cleanup is only necessary for nodes which have a CSS transition attached but no updates from the previous update stored in the registry (e.g. when the transition was attached to a view but has never been triggered).

bartlomiejbloniarz: I think we also remove the updates from the registry after the transition ends? So the cleanup won't happen if there is no transition happening

MatiPl01: > I think we also remove the updates from the registry after the transition ends

Not necessarily. We remove them when the `remove` or `removeBatch` method is called, so only when a node is removed from the ShadowTree or a transition is removed from a node. We don't have to remove props when the transition ends because, when specific props are transitioned, we don't want React to be able to commit its changes without animating these properties.

Wrapping up, the body of the remove function is the same as the content of the for loop in the `removeBatch` function, so, we don't have to call `detach` from the `CSSManger`, because the `detach` method in the `CSSManger` is called only when the component unmounts and this case should be already handled by `removeBatch` as the corresponding node will be removed from the ShadowTree after the component unmounts.

RohovDmytro:
With react-native-reanimated@4.1.0 the fix is no longer working, if I navigate 2+ screens deeps with react-freeze enabled – some animation simply stop working.

Hamidrezana:
I confirm that the animation stops working when using inside tabs of `@react-navigation/bottom-tabs` with `freezeOnBlur: true`.

fr0staman:
@Hamidrezana +, also `@react-navigation/bottom-tabs` with `freezeOnBlur: true`, especially I'm using `react-native-reanimated-carousel`

RohovDmytro:
Confirm to also have `@react-navigation/bottom-tabs`

PR #7055, opened 2025-02-19
draft: prepare for react 19

patrycjakalinska:


packages/docs-reanimated/versioned_docs/version-3.x/core/createAnimatedComponent.mdx
```diff
@@ -44,7 +44,7 @@ function createAnimatedComponent<P extends object>(
 
 #### `component`
 
-The component you want to make animatable. Function components have to be wrapped with `React.forwardRef()`.
+The component you want to make animatable.
```
tjzel, line 47: Let's ship the docs in a separate PR after we release this change.

Also we should probably mark it that you need `forwardRef` for reanimated >3.17

packages/react-native-reanimated/src/component/FlatList.tsx
```diff
@@ -113,13 +113,13 @@ const FlatListForwardRefRender = function <Item = any>(
   );
 };
 
-export const ReanimatedFlatList = forwardRef(FlatListForwardRefRender) as <
+export const ReanimatedFlatList = FlatListRender as <
```
tjzel, line 116: If I understand correctly, we can't yet commit these changes, since React 19 is only available in React Native 0.78. We have to keep backwards compatibility (at the moment) for at least React Native 0.76, which still has React 18.

tjzel (changes requested):

packages/docs-reanimated/versioned_docs/version-3.x/core/createAnimatedComponent.mdx
patrycjakalinska: Yes! Was think about it

packages/react-native-reanimated/src/component/FlatList.tsx
patrycjakalinska: I talked with @piaskowyk to release this full React 19 support to Reanimated 4, and React 19 support with backward compatibility for 3.17.x, so user could use either 0.78 or for example 0.77, 0.76 with no problem - i'm currently working on it
It'll be similiar to this but without removing `forwardRef` and `.Provider` (only change from React 19 is the `useRef()` will have initial value). Further changes probably will include guard that would check if is React 19, similiary as GH did it.

packages/docs-reanimated/package.json
```diff
@@ -42,9 +42,9 @@
     "react-dom": "^17.0.2",
     "react-draggable": "^4.4.5",
     "react-native": "^0.71.4",
-    "react-native-gesture-handler": "2.18.0",
-    "react-native-reanimated": "4.0.0-nightly-20250224-c288e54a2",
-    "react-native-svg": "15.4.0",
+    "react-native-gesture-handler": "2.24.0",
+    "react-native-reanimated": "^4.0.0-beta.2",
```
tjzel, line 46: Have you tested the docs on this version? I think it would be safer to upgrade to a recent nightly.

packages/react-native-reanimated/src/reactUtils.ts
```diff
@@ -2,12 +2,16 @@
 import React from 'react';
 
 function getCurrentReactOwner() {
-  const ReactSharedInternals =
+  return (
     // @ts-expect-error React secret internals aren't typed
-    React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED ||
+    React.__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE?.A?.getOwner() ||
```
tjzel, line 7: A

codeion: Hi,
This change crashes production builds on the web.
TypeError: React.__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE?.A?.getOwner is not a function

codeion: [react-native-reanimated+4.0.0-nightly-20250513-896f6e7a7.patch](https://github.com/user-attachments/files/20207257/react-native-reanimated%2B4.0.0-nightly-20250513-896f6e7a7.patch)

patrycjakalinska: Hi @codeion, we've tested this on production on web and it seems to work - could you provide a reproduction that should yield this result?

codeion: I see you've fixed it #7543 & #7545

PR #7064, opened 2025-02-20
refactor: run `_maybeFlushUIUpdatesQueue` through `_microtaskQueueFinalizers`

tjzel:
## Summary

With current implementation of `queueMicrotask` we call `_maybeFlushUIUpdatesQueue` which executes `performOperations`. This isn't a good pattern in my opinion, since `queueMicrotask` shouldn't be tied directly to the rendering layer. Regardless, to keep the current behavior I added `_microtaskQueueFinalizers` array in `react-native-worklets` where Reanimated could register its finalizer - it has to be executed after all the tasks, to gather all the updates.

This should be considered as a temporary API and we'd need to overhaul this in the future for a consistent event loop model.

## Test plan

See that animations run just fine!


packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/UIRuntimeDecorator.h
```diff
@@ -6,7 +6,11 @@ namespace worklets {
 
 class UIRuntimeDecorator {
  public:
-  static void decorate(facebook::jsi::Runtime &uiRuntime);
+  static void decorate(
```
tomekzaw, line 9: Why `&&`? 😢

tjzel: It seems that you reviewed after the target changed but I hadn't yet merged this PR to main

packages/react-native-worklets/src/threads.ts
```diff
@@ -35,7 +36,7 @@ export function setupMicrotasks() {
         microtasksQueue[index]();
       }
       microtasksQueue = [];
-      global._maybeFlushUIUpdatesQueue();
+      global._microtaskQueueFinalizers.forEach((finalizer) => finalizer());
```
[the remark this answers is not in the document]
tjzel: I haven't measured it, but it should have virtually no impact. In this implementation `global._microtaskQueueFinalizers` has only one element. `_maybeFlushUIUpdatesQueue` was a HostFunction, same as the array's element. This bit of code is invoked only once per frame.

tomekzaw (approved):

PR #7065, opened 2025-02-20
chore: Add React Native 0.78.0 support

MatiPl01:

PR #7086, opened 2025-02-24
fix: CSS transitions glitches after upgrade to RN 0.78

MatiPl01:
## Summary

This PR fixes:

> 4. [CSS] Transitions aren't triggered / run between invalid states (it is strange as transitions on the transition routes screen work but they are triggered by the state change from the setInterval. Only transitions triggered manually don't work)

from the #7065 PR.

### What caused the issue?

(I am not sure about the exact cause but I deduced that based on my observations)

It looks like the Shadow Tree committing logic changed in RN and, when we prevented react from committing its changes to the Shadow Tree (which is what we actually do to prevent applying changes before the CSS transition starts), it was unaware of which values are were passed in the view props on the JS side. Because of that, when the next commit from React was triggered, it included old props (not props that were last passed as the view style but some previous values from the last React commit that weren't modified by us).

It worked in a different way before (in RN 0.77) and React was able to include last style values in its commit, even though we prevented it from committing them before.

### How this PR fixes the issue?

Instead of keeping only the currently transitioned style value in the CSS transitions updates registry we now store all properties that were transitioned before and are transitioned now. We only remove these props from the registry that cannot be transitioned (aren't listed in the `transitionProperty` vector). All other props are kept as long as the view exists and has CSS transition attached.

Since these props are transitioned, we don't have to remove them. Keeping it longer makes it possible to override invalid props in the RN's commit by the last values of the previous transition instead of some old invalid props that RN is trying to commit.

## Example recordings

| Before | After |
|-|-|
| <video src="https://github.com/user-attachments/assets/297c54b2-eb49-41a5-8472-81cb301f7a99" /> | <video src="https://github.com/user-attachments/assets/a8e8a655-373f-4a01-8129-9d4268597e45" /> |

## Test plan

Open transition examples, such as the **App Settings** screen and observe the difference.

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
@@ -47,8 +47,8 @@ void CSSTransitionsRegistry::updateSettings(
   // Replace style overrides with the new ones if transition properties were
   // updated (we want to keep overrides only for transitioned properties)
   if (config.properties.has_value()) {
-    const auto &currentStyle = transition->getCurrentInterpolationStyle();
-    setInUpdatesRegistry(transition->getShadowNode(), currentStyle);
+    updateInUpdatesRegistry(
+        transition, transition->getCurrentInterpolationStyle());
   }
 }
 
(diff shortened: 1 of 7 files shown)


packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
```diff
@@ -143,13 +143,44 @@ PropsObserver CSSTransitionsRegistry::createPropsObserver(const Tag viewTag) {
...
+      if (lastUpdates.count(prop)) {
+        filteredUpdates[prop] = lastUpdates[prop];
+      }
+    }
+  }
+
+  // updates object contains only allowed properties so we don't need
+  // to do additional filtering here
+  filteredUpdates.update(updates);
+  setInUpdatesRegistry(shadowNode, filteredUpdates);
```
MatiPl01, line 181: Instead of setting just these props that are transitioned in the current transition / will be transitioned in the just triggered transition, we also include props from the last update.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -852,16 +852,16 @@ void ReanimatedModuleProxy::performOperations() {
 
       // Update CSS transitions and flush updates
       cssTransitionsRegistry_->update(currentCssTimestamp_);
-      cssTransitionsRegistry_->flushUpdates(updatesBatch, false);
+      cssTransitionsRegistry_->flushUpdates(updatesBatch);
```
MatiPl01, line 855: We can no longer override props from the last transition, thus we no longer need this merging distinction logic and `flushUpdates` in the same way for all registers.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSTransitionsRegistry.cpp
```diff
@@ -143,13 +143,44 @@ PropsObserver CSSTransitionsRegistry::createPropsObserver(const Tag viewTag) {
...
+    // If transitionProperty is set to 'all' (optional has no value), we have
+    // to keep the result of the previous transition updated with the new
+    // transition starting values
+    if (!lastUpdates.empty()) {
+      filteredUpdates = lastUpdates;
+    }
+  } else if (!lastUpdates.empty()) {
+    // Otherwise, we keep only allowed properties from the last updates
+    // and update the object with the new transition starting values
+    for (const auto &prop : transitionProperties.value()) {
```
piaskowyk, line 171: Did you check what happens when you add or remove a transition property during the render?

piaskowyk: We discussed it offline, and Mateusz confirmed that this works in that scenario too

PR #7087, opened 2025-02-24
feat: make Reanimated depend on Worklets

tjzel:
## Summary

From now on `react-native-worklets` will be a required dependency for `react-native-reanimated`.

## Test plan

CI 🍏


packages/react-native-reanimated/src/index.ts
```diff
@@ -291,6 +291,6 @@ export {
   ScreenTransition,
   startScreenTransition,
 } from './screenTransition';
-export type { WorkletRuntime } from './WorkletsResolver';
-export { LogLevel as ReanimatedLogLevel } from './WorkletsResolver';
-export { isWorkletFunction } from './WorkletsResolver';
+export type { WorkletRuntime } from 'react-native-worklets';
+export { LogLevel as ReanimatedLogLevel } from 'react-native-worklets';
+export { isWorkletFunction } from 'react-native-worklets';
```
tomekzaw, line 296: Let's merge these imports

packages/react-native-reanimated/src/core.ts
```diff
@@ -3,6 +3,8 @@ import {
   controlEdgeToEdgeValues,
   isEdgeToEdge,
 } from 'react-native-is-edge-to-edge';
+import type { WorkletFunction } from 'react-native-worklets';
+import { makeShareableCloneRecursive } from 'react-native-worklets';
```
tomekzaw, line 7: Let's merge it with import below

packages/react-native-reanimated/package.json
```diff
@@ -26,7 +26,7 @@
     "type:check:plugin": "cd plugin && yarn type:check:src",
     "type:check:app": "yarn workspace common-app type:check",
     "type:check:tests:common": "./scripts/test-ts.sh __typetests__/common",
-    "build": "./scripts/duplicate-worklets-code.sh && yarn build:plugin && bob build",
+    "build": "yarn workspace react-native-worklets build && bob build",
```
tomekzaw, line 29: Is `yarn workspace react-native-worklets build` run automatically with `yarn` in project root?

packages/react-native-reanimated/package.json
```diff
@@ -96,7 +96,8 @@
   "peerDependencies": {
     "@babel/core": "^7.0.0-0",
     "react": "*",
-    "react-native": "*"
+    "react-native": "*",
+    "react-native-worklets": "0.0.1"
```
tomekzaw, line 100: `0.0.1` is okay for now, but what's the plan for the future? Note that `peerDependencies` are not respected with `npm i --force`. Maybe let's just use `*` here and add some more sophisticated checks in Reanimated?

packages/react-native-reanimated/android/build.gradle
```diff
@@ -484,7 +446,7 @@ dependencies {
         implementation "com.facebook.react:hermes-android" // version substituted by RNGP
     }
  
-    if (hasExternalWorklets) {
+    if(project != rootProject) {
```
tomekzaw, line 449:
```suggestion
    if (project != rootProject) { // skip missing dependency error during linting
```

packages/react-native-reanimated/RNReanimated.podspec
```diff
@@ -85,7 +66,7 @@ Pod::Spec.new do |s|
       '"$(PODS_ROOT)/DoubleConversion"',
       '"$(PODS_ROOT)/Headers/Private/React-Core"',
       '"$(PODS_ROOT)/Headers/Private/Yoga"',
-      external_worklets_header_path,
+      '"$(PODS_ROOT)/Headers/Public/RNWorklets"',
```
tomekzaw, line 69: Do we really need this? Can't we just declare a dependency? Does it work with use_frameworks static and dynamic?

tomekzaw (changes requested):

packages/react-native-reanimated/package.json
tjzel: Doing anything with just `yarn` (not e.g. `yarn build`) requires adding [yarn plugins](https://yarnpkg.com/advanced/plugin-tutorial#using-the-yarn-api) which would be nice to have but we've got more important things on the table right now.

packages/react-native-reanimated/RNReanimated.podspec
tjzel: I intend to drop in the future PR, as we need to change all the import paths then.

packages/react-native-reanimated/src/core.ts
tjzel: One is a type import, so can't merge.

PR #7091, opened 2025-02-25
refactor(worklets): remove nesting of some source files

tjzel:
## Summary

For the purposes of code duplication mechanism we had one more level of nesting for source files in worklets. It's no longer necessary.

TypeScript sources were in `src/worklets`. Now they are moved to just `src`.

Android Java sources were in source sets in `android/src/worklets/main`. Now they are in `android/src/main`.

## Test plan

CI 🍏


piaskowyk (approved):
If you've tested it at runtime, it seems legit

PR #7100, opened 2025-02-26
Restore `JSScheduler::scheduleOnJS` method

tomekzaw:
## Summary

This PR removes all leftovers from `JSScheduler.cpp` and `JSScheduler.h` and converts `scheduleOnJS` from `std::function` back to a regular class method.

### DIFF
--- a/packages/react-native-worklets/Common/cpp/worklets/Tools/JSScheduler.h
+++ b/packages/react-native-worklets/Common/cpp/worklets/Tools/JSScheduler.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <ReactCommon/CallInvoker.h>
-#include <ReactCommon/RuntimeExecutor.h>
 #include <jsi/jsi.h>
 
 #include <memory>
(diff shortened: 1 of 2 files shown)


piaskowyk (approved):

packages/react-native-worklets/Common/cpp/worklets/Tools/JSScheduler.h
```diff
@@ -17,13 +16,14 @@ class JSScheduler {
  public:
   explicit JSScheduler(
       jsi::Runtime &rnRuntime,
-      const std::shared_ptr<CallInvoker> &jsCallInvoker);
+      const std::shared_ptr<CallInvoker> &jsCallInvoker)
+      : rnRuntime_(rnRuntime), jsCallInvoker_(jsCallInvoker) {}
```
tjzel, line 20: I'd prefer the implementation of the constructor to be in .cpp but it's fine too.

tjzel (approved):

packages/react-native-worklets/Common/cpp/worklets/Tools/JSScheduler.h
tomekzaw: I also like it more that way but it's just a simple constructor that fills class members so maybe let's avoid code duplication in some sense and leave it here? What do you think?

tjzel: Let's leave it as you did it then 🔢

PR #7103, opened 2025-02-26
Remove `isFabric()`, `viewName`, `getScrollableNode` along with all usages

tomekzaw:
Co-authored-with: Krzysztof Piaskowy <krzysztof.piaskowy@swmansion.com>

## Summary

This PR replaces all `isFabric()`, `IS_FABRIC` and `global._IS_FABRIC` checks with `true` (as well as optimizes the expressions and removes dead code).

Turns out, `getScrollableNode` call was inside `!IS_FABRIC` so we also removed that.

We also removed `viewName` from `ViewDescriptor` as it was used only on Paper.

### DIFF
--- a/apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/TestRunner/UpdatesContainer.ts
+++ b/apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/TestRunner/UpdatesContainer.ts
@@ -27,7 +27,6 @@ export function createUpdatesContainer() {
 
   function _updateNativeSnapshot(updateInfos: JsUpdate[], jsUpdateIndex: number): void {
     'worklet';
-    const isFabric = global._IS_FABRIC;
     nativeSnapshots.modify(values => {
       'worklet';
       for (const updateInfo of updateInfos) {
@@ -35,9 +34,7 @@ export function createUpdatesContainer() {
         const updatedProps = Object.keys(updateInfo.update);
         const propsToUpdate = updatedProps.filter(propName => isValidPropName(propName));
         for (const prop of propsToUpdate) {
-          snapshot[prop] = isFabric
-            ? global._obtainPropFabric(updateInfo?.shadowNodeWrapper, prop)
-            : global._obtainPropPaper(updateInfo?.tag, prop);
+          snapshot[prop] = global._obtainPropFabric(updateInfo?.shadowNodeWrapper, prop);
         }
         values.push({
           tag: updateInfo.tag,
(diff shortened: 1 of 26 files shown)


apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/TestRunner/UpdatesContainer.ts
```diff
@@ -83,10 +80,6 @@ export function createUpdatesContainer() {
 
   function pushLayoutAnimationUpdates(tag: number, update: Record<string, unknown>) {
     'worklet';
-    if (global._IS_FABRIC) {
-      // layout animation doesn't work on Fabric yet
-      return;
-    }
```
tjzel, line 89: 🤣

packages/react-native-reanimated/src/hook/useAnimatedRef.ts
```diff
@@ -1,38 +1,31 @@
...
     uiViewClassName?: string;
   };
 }
 
 function getComponentOrScrollable(component: MaybeScrollableComponent) {
-  if (IS_FABRIC && component.getNativeScrollRef) {
+  if (component.getNativeScrollRef) {
     return component.getNativeScrollRef();
-  } else if (!IS_FABRIC && component.getScrollableNode) {
-    return component.getScrollableNode();
```
tjzel, line 35: `.getScrollableNode` is used in other places of the code, not only on Fabric. While this change here is consistent, are we sure that it was correctly implemented in the first place?

packages/react-native-reanimated/src/hook/useAnimatedRef.ts
```diff
@@ -58,42 +50,32 @@ export function useAnimatedRef<
...
 
     fun.current = null;
 
     const animatedRefShareableHandle = makeShareableCloneRecursive({
-      __init: () => {
+      __init: (): AnimatedRefOnUI => {
         'worklet';
-        const f: AnimatedRefOnUI = () => tag.value;
-        f.viewName = viewName;
-        return f;
```
tjzel, line 96: Isn't `viewName`, by any chance, used by our API consumers?

tjzel (approved):
The PR is good, I'm just worried about some bits of the code that were already there.

packages/react-native-reanimated/src/hook/useAnimatedRef.ts
piaskowyk: This is an internal property, and we've only used it on Paper, so I think we can safely remove i

packages/react-native-reanimated/src/hook/useAnimatedRef.ts
tomekzaw: Looks like `getScrollableNode` was only used on Paper since the only place where we read it was here and it's hidden behind `if (!IS_FABRIC)`

packages/react-native-reanimated/src/hook/useAnimatedRef.ts
tomekzaw: Well, technically someone could read `animatedRef.viewName` but this was never meant to be a public API so we don't need to support that. Also, this is pretty easy to replace with actual string so let's not worry about this.

packages/react-native-reanimated/src/hook/useAnimatedRef.ts
tomekzaw: I've checked `ScrollToExample` and it works correctly (after we added flash-list patch in another quite recent PR).

PR #7110, opened 2025-02-27
Remove Layout Animations and SET from iOS on Paper

piaskowyk:
## Summary

This PR removes iOS code responsible for Layout Animation and Shared Element Transition on Paper


tomekzaw (approved):

tjzel (approved):
Oof

PR #7113, opened 2025-02-27
Remove Layout Animations and SET from Android on Paper

piaskowyk:
## Summary

This PR removes Android code responsible for Layout Animation and Shared Element Transition on Paper


packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/nativeProxy/NativeProxyCommon.java
```diff
@@ -64,6 +59,9 @@ protected NativeProxyCommon(ReactApplicationContext context) {
       tempHandlerStateManager = null;
     }
     gestureHandlerStateManager = tempHandlerStateManager;
+    mNodesManager =
+        Objects.requireNonNull(mContext.get().getNativeModule(ReanimatedModule.class))
+            .getNodesManager();
```
tomekzaw, line 64: Moved here from `prepareLayoutAnimations`

tomekzaw (approved):

PR #7125, opened 2025-02-28
Remove `RCTUIView+Reanimated`

tomekzaw:
## Summary

This PR removes RCTUIView+Reanimated.h and RCTUIView+Reanimated.mm files as `center` property is no longer used.


tomekzaw:
Turns out, it's still used on macOS, at least the import.

PR #7131, opened 2025-02-28
Remove redundant includes on iOS

tomekzaw:
## Summary

This PR removes unnecessary imports on iOS as well as some unused typedefs and class member fields.

## Test plan

See if the CI is green.


piaskowyk (approved):
Once the CI goes green, let's merge it 🚀

PR #3555, opened 2022-09-07
Fixed retain cycle of `REAModule` causing memory leak

Kwasow:
## Description

A retain cycle of `REAModule` caused by some lambdas in `NativeProxy` that held a reference to it are removed.

A mechanism for easily detecting such memory leaks in the future is also introduced.

## Changes

Before:
```objective-c
auto requestRender = [reanimatedModule, &module](std::function<void(double)> onRender, jsi::Runtime &rt) {
    [reanimatedModule.nodesManager postOnAnimation:^(CADisplayLink *displayLink) {
      double frameTimestamp = calculateTimestampWithSlowAnimations(displayLink.targetTimestamp) * 1000;
      jsi::Object global = rt.global();
      jsi::String frameTimestampName = jsi::String::createFromAscii(rt, "_frameTimestamp");
      global.setProperty(rt, frameTimestampName, frameTimestamp);
      onRender(frameTimestamp);
      global.setProperty(rt, frameTimestampName, jsi::Value::undefined());
    }];
  };

auto synchronouslyUpdateUIPropsFunction = [reanimatedModule](jsi::Runtime &rt, Tag tag, const jsi::Value &props) {
    NSNumber *viewTag = @(tag);
    NSDictionary *uiProps = convertJSIObjectToNSDictionary(rt, props.asObject(rt));
    [reanimatedModule.nodesManager synchronouslyUpdateViewOnUIThread:viewTag props:uiProps];
  };
```

After:
```objective-c
auto nodesManager = reanimatedModule.nodesManager;

auto requestRender = [nodesManager, &module](std::function<void(double)> onRender, jsi::Runtime &rt) {
    [nodesManager postOnAnimation:^(CADisplayLink *displayLink) {
      double frameTimestamp = calculateTimestampWithSlowAnimations(displayLink.targetTimestamp) * 1000;
      jsi::Object global = rt.global();
      jsi::String frameTimestampName = jsi::String::createFromAscii(rt, "_frameTimestamp");
      global.setProperty(rt, frameTimestampName, frameTimestamp);
      onRender(frameTimestamp);
      global.setProperty(rt, frameTimestampName, jsi::Value::undefined());
    }];
  };

auto synchronouslyUpdateUIPropsFunction = [nodesManager](jsi::Runtime &rt, Tag tag, const jsi::Value &props) {
    NSNumber *viewTag = @(tag);
    NSDictionary *uiProps = convertJSIObjectToNSDictionary(rt, props.asObject(rt));
[...]

PR #7133, opened 2025-02-28
Remove `requiresMainQueueSetup` with `YES`

tomekzaw:
## Summary

⚠️ This is potentially a very breaking change in some setups!

This PR removes `requiresMainQueueSetup` method in `REAModule` that was responsible for marking that Reanimated module should be initialized on the main queue.

This method was added as a side effect 3 years ago in https://github.com/software-mansion/react-native-reanimated/pull/3555 in order to make the warning go away and to keep the identical initialization behavior.

Looks like this is a perfect moment to get rid of this chunk of the code and hopefully it should work just fine.

Please let us know immediately if you have any troubles with running Reanimated without `requiresMainQueueSetup`.


tomekzaw:
Waiting for https://github.com/software-mansion/react-native-reanimated/pull/7136

piaskowyk (approved):

Issue #7042, opened 2025-02-18
Animating max/min width/height properties does not work in custom animations


sofiekhullar:
### Description

Same issue as [7035](https://github.com/software-mansion/react-native-reanimated/issues/7035) but for the dimension properties:  `minWidth`, `maxWidth`, `minHeight` and `maxHeight`. Custom entering/exiting animations do not work, though these properties animate correctly when using `useAnimatedStyle`.

## Expected
´minWidth´, `maxWidth`, `minHeight`, and `maxHeight` should animate during entering/exiting transitions, just as they do when used with `useAnimatedStyle`.

## Actual
There is no animation applied to these properties during entering/exiting transitions.

## Code example
```
export default function App() {
  const [visible, setVisible] = useState(false)
  const [moved, setMoved] = useState(false)

    const entering = () => {
    "worklet";

    return {
      initialValues: { minWidth: 0, opacity: 0 },
      animations: {
        opacity: withTiming(1, { duration: 1000 }),
        minWidth: withTiming(100, { duration: 1000 })
      },
    };
  };

  const exiting = () => {
    "worklet";

    return {
      initialValues: { top: 0, opacity: 1 },
      animations: {
        opacity: withTiming(0, { duration: 1000 }),
        minWidth: withTiming(200, { duration: 1000 })
      },
    };
  };

  const animatedStyles = useAnimatedStyle(() => {
     return {
      minWidth: withTiming(moved ? 200 : 0, {duration: 1000}),
    };
  })

  return (
    <SafeAreaView style={styles.container}>
    {!visible && <Animated.View entering={entering} exiting={exiting} style={styles.box} />}
    <Animated.View style={[styles.box, styles.box2, animatedStyles]}  />
    <View style={styles.pressableContainer}>
    <Pressable style={styles.pressable} onPress={() => setVisible(!visible)}><Text style={styles.text}>Toggle</Text></Pressable>
    <Pressable style={styles.pressable} onPress={() => setMoved(!moved)}><Text style={styles.text}>Update minWidth</Text></Pressable>
    </View>
    </SafeAreaView>
  );
}
````

### Steps to reproduce

1. Create an Animated.View with a custom entering animation that includes dimension max/min properties
1. Toggle the visibility of the component
1. Observe that the dimension properties do not animate during entry
1. Compare with the same animation using useAnimatedStyle – which works correctly.

### Snack or a link to a repository

[...]

PR #7135, opened 2025-02-28
feat: Add warning for unsupported layout animation props

patrycjakalinska:
## Summary

This PR aims to help users avoid confusion by warning them when using layout animation properties that are not officially supported. It was motivated by arising amount of issues related to "X prop not working in custom layout animation" - we want to make it clear for users c: 

Related to: #7042 

## Test plan

<details>

```tsx
import { View, Button, StyleSheet, SafeAreaView } from 'react-native';
import Animated, { withTiming, withDelay } from 'react-native-reanimated';

const WIDTH = 200;

const customEntering = (targetValues) => {
  'worklet';
  const animations = {
    originX: withTiming(targetValues.targetOriginX, { duration: 3000 }),
    opacity: withTiming(1, { duration: 2000 }),
    borderRadius: withDelay(1500, withTiming(40, { duration: 3000 })),
    transform: [
      { rotate: withTiming('0deg', { duration: 4000 }) },
      { scale: withTiming(1, { duration: 3500 }) },
    ],
    maxWidth: withTiming(200, { duration: 3000 }),
  };
  const initialValues = {
    originX: -WIDTH,
    opacity: 0,
    borderRadius: 10,
    transform: [{ rotate: '90deg' }, { scale: 0.2 }],
    maxWidth: 10,
  };
  return {
    initialValues,
    animations,
  };
};

const customExiting = (values) => {
  'worklet';
  const animations = {
    originX: withTiming(2 * WIDTH, { duration: 3000 }),
    opacity: withTiming(0, { duration: 2000 }),
    transform: [{ scale: withTiming(0.2, { duration: 3500 }) }],
    minWidth: withTiming(10, { duration: 3000 }),
  };
  const initialValues = {
    originX: values.currentOriginX,
    opacity: 1,
    transform: [{ scale: 1 }],
    minWidth: 0,
  };
  return {
    initialValues,
    animations,
  };
};

export default function EnteringExample() {
  const [show, setShow] = useState(false);

  return (
    <SafeAreaView>
        <Button title="Click me" onPress={() => setShow(!show)} />
        {show && (
          <Animated.View
            style={styles.card}
            entering={customEntering}
            exiting={customExiting}
          />
        )}
      </View>
    </SafeAreaView>
  );
}

  container: {
    display: 'flex',
    padding: 128,
    alignItems: 'center',
    height: '100%',
  },
  card: {
    width: WIDTH,
    height: 300,
    backgroundColor: '#b58df1',
    justifyContent: 'center',
    alignItems: 'center',
    margin: 20,
  },
});
```
</details>


packages/react-native-reanimated/src/animation/util.ts
```diff
@@ -44,6 +44,18 @@ import {
...
+  'originY',
+  'width',
+  'height',
+  'borderRadius',
+  'globalOriginX',
+  'globalOriginY',
+  'opacity',
+  'transform',
+];
+
```
patrycjakalinska, line 58: I was not sure about placing it here - if you think supported props list belongs somewhere else - im open to suggestions!

packages/react-native-reanimated/src/animation/styleAnimation.ts
```diff
@@ -192,6 +192,14 @@ export function withStyleAnimation(
               )}`
             );
           }
+          if (
+            typeof currentEntry.path[0] === 'string' &&
+            !LAYOUT_ANIMATION_SUPPORTED_PROPS.includes(currentEntry.path[0])
+          ) {
+            logger.warn(
+              `'${currentEntry.path[0]}' property is not officially supported for layout animations. It may not work as expected.`
```
patrycjakalinska, line 200: I decided on short warning, without getting into details - if we want we can list supported props or add a link to related documentation

packages/react-native-reanimated/src/animation/util.ts
```diff
@@ -44,6 +44,18 @@ import {
...
+  'originX',
+  'originY',
+  'width',
+  'height',
+  'borderRadius',
+  'globalOriginX',
+  'globalOriginY',
+  'opacity',
+  'transform',
+];
```
MatiPl01, line 57: What about using a Set? I know that this list of props is short but set may be faster for frequent lookups.

packages/react-native-reanimated/src/animation/util.ts
MatiPl01: Maybe export a function like `isValidLayoutAnimationProp` instead of a constant? It will be cleaner for me to export a function from the `util` file with utilities instead of exporting a constant.

You can keep this constant here but don't export it. Export just a function that uses this constant.

packages/react-native-reanimated/src/animation/styleAnimation.ts
MatiPl01: It's fine for now. Let's remember to update docs and improve the warning then.

MatiPl01 (approved):
LGTM
Left a few suggestions you may consider implementing before merging this PR.

packages/react-native-reanimated/src/animation/styleAnimation.ts
piaskowyk, line 200: Looks good byt may be change it a bit to:
```suggestion
              `[Reanimated] The '${currentEntry.path[0]}' property is not supported for layout animations. It may not work as expected.`
```

packages/react-native-reanimated/src/animation/styleAnimation.ts
```diff
@@ -192,6 +192,14 @@ export function withStyleAnimation(
               )}`
             );
           }
+          if (
+            typeof currentEntry.path[0] === 'string' &&
+            !LAYOUT_ANIMATION_SUPPORTED_PROPS.includes(currentEntry.path[0])
```
piaskowyk, line 197: As Mateusz suggested, let's use a set for faster lookup

packages/react-native-reanimated/src/animation/styleAnimation.ts
MatiPl01: `[Reanimated]` prefix is added automatically by the logger, so it shouldn't be included in the log string.

packages/react-native-reanimated/src/animation/util.ts
patrycjakalinska: I couldn't use Set because it cannot be passed to UI thread (and to use Set i had to wokletize `isValidLayoutAnimationProp` function, so :/). I talked with tjzel, it is kinda possible to do that, but for now it would be best to use array (i'm also sorry to hear that)

packages/react-native-reanimated/src/animation/styleAnimation.ts
patrycjakalinska: forwarding from Mateusz comment:
I couldn't use Set because it cannot be passed to UI thread (and to use Set i had to wokletize isValidLayoutAnimationProp function, so :/). I talked with tjzel, it is kinda possible to do that, but for now it would be best to use an array

packages/react-native-reanimated/src/animation/util.ts
MatiPl01: What about an object with all values set to `true`? It will work like a set.

onst LAYOUT_ANIMATION_SUPPORTED_PROPS = {
  originX: true,
  ...
}

packages/react-native-reanimated/src/animation/styleAnimation.ts
```diff
@@ -192,6 +192,14 @@ export function withStyleAnimation(
               )}`
             );
           }
+          if (
+            typeof currentEntry.path[0] === 'string' &&
+            !isValidLayoutAnimationProp(currentEntry.path[0].trim())
+          ) {
+            logger.warn(
+              `'${currentEntry.path[0]}' property is not officially supported for layout animations. It may not work as expected.`
```
MatiPl01, line 200: nit: Maybe use a separate variable for `currentEntry.path[0]`
```suggestion
          const propName = currentEntry.path[0];
          if (
            typeof propName === 'string' &&
            !isValidLayoutAnimationProp(propName.trim())
          ) {
            logger.warn(
              `'${propName}' property is not officially supported for layout animations. It may not work as expected.`
            );
          }
```

MatiPl01 (approved):
Looks fine. Just one small suggestion.

packages/react-native-reanimated/src/animation/styleAnimation.ts
```diff
@@ -192,6 +192,15 @@ export function withStyleAnimation(
               )}`
             );
           }
+          const propName = currentEntry.path[0];
```
nandorojo, line 195: Should we only check this in dev mode?

PR #7136, opened 2025-02-28
Add asserts for queue during Reanimated initialization on iOS

tomekzaw:
## Summary

This PR adds `RCTAssertMainQueue()`, `RCTAssertUIManagerQueue()` and `REAAssertJavaScriptQueue()` calls in various initialization methods in `REAModule`, `NativeProxy` and `NodesManager` on iOS along with the implementation of `REAAssertJavaScriptQueue`. This way we can make sure that the methods are called on appropriate threads as well as track this information in the codebase.

Analogous asserts will be added in `react-native-worklets` in a separate PR: https://github.com/software-mansion/react-native-reanimated/pull/7137

## Test plan

✅ Tested in debug and release mode

### DIFF
--- a/packages/react-native-reanimated/apple/reanimated/apple/REAAssertJavaScriptQueue.h
+++ b/packages/react-native-reanimated/apple/reanimated/apple/REAAssertJavaScriptQueue.h
@@ -1,0 +1,14 @@
+#import <react/debug/react_native_assert.h>
+
+// Copied from RCTJSThreadManager.mm
+static NSString *const RCTJSThreadName = @"com.facebook.react.runtime.JavaScript";
+
+static BOOL REAIsJavaScriptQueue()
+{
+  return [NSThread.currentThread.name isEqualToString:RCTJSThreadName];
+}
+
+static void REAAssertJavaScriptQueue()
+{
+  react_native_assert(REAIsJavaScriptQueue() && "This function must be called on the JavaScript queue");
+}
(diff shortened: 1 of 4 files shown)


packages/react-native-reanimated/apple/reanimated/apple/REAAssertJavaScriptQueue.h
```diff
@@ -0,0 +1,14 @@
+#import <react/debug/react_native_assert.h>
+
+// Copied from RCTJSThreadManager.mm
+static NSString *const RCTJSThreadName = @"com.facebook.react.runtime.JavaScript";
+
+static BOOL REAIsJavaScriptQueue()
+{
+  return [NSThread.currentThread.name isEqualToString:RCTJSThreadName];
```
piaskowyk, line 8: Do we need to check it in Release mode too?

piaskowyk (commented):
What is the benefit of adding those checks to NodeManager as well?

packages/react-native-reanimated/apple/reanimated/apple/REAAssertJavaScriptQueue.h
tomekzaw: The implementation of `RCTAssertJavaScriptQueue` is analogous to `RCTAssertMainQueue` (the only difference is that `RCTAssertMainQueue` is a precompiled macro and `RCTAssertJavaScriptQueue is a static inline function which is a bit more modern) so the behavior should be identical to the one in React Native. Besides, it seems like `RCTAssert` is optimized away in release build. Finally, these asserts are not expensive and as a matter of fact, having them also in the release build doesn't sound that bad at all.

PR #7137, opened 2025-02-28
Add asserts for queue during Worklets initialization on iOS

tomekzaw:
## Summary

This PR adds `RCTAssertMainQueue()` and `REAAssertJavaScriptQueue()` calls in various initialization methods in `WorkletsModule`, `AnimationFrameQueue` and `WorkletsMessageThread` on iOS along with the implementation of `AssertJavaScriptQueue`. This way we can make sure that the methods are called on appropriate threads as well as track this information in the codebase.

I wasn't able to come up with an assert for "com.meta.react.turbomodulemanager.queue" in `invalidate` methods since `NSThread.currentThread.name` returns `nil` so I just added `// Called on com.meta.react.turbomodulemanager.queue` comments.

## Test plan

✅ Tested in debug and release mode

### DIFF
--- a/packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
+++ b/packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
@@ -1,5 +1,8 @@
 #import <worklets/apple/AnimationFrameQueue.h>
+#import <worklets/apple/AssertJavaScriptQueue.h>
 #import <worklets/apple/SlowAnimations.h>
+
+#import <React/RCTAssert.h>
 
 @implementation AnimationFrameQueue {
   /* DisplayLink is thread safe. */
@@ -14,6 +17,7 @@ typedef void (^AnimationFrameCallback)(WorkletsDisplayLink *displayLink);
 
 - (instancetype)init
 {
+  AssertJavaScriptQueue();
   displayLink_ = [WorkletsDisplayLink displayLinkWithTarget:self selector:@selector(executeQueue:)];
 #if TARGET_OS_OSX
   // nothing
@@ -27,6 +31,7 @@ typedef void (^AnimationFrameCallback)(WorkletsDisplayLink *displayLink);
 
 - (void)requestAnimationFrame:(std::function<void(double)>)callback
 {
+  RCTAssertMainQueue();
   {
     std::lock_guard<std::mutex> lock(callbacksMutex_);
     frameCallbacks_.push_back(callback);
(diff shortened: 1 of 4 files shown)


piaskowyk (approved):

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -27,6 +31,7 @@ - (instancetype)init
 
 - (void)requestAnimationFrame:(std::function<void(double)>)callback
 {
+  RCTAssertMainQueue();
```
tjzel, line 34: Please revert this change, `requestAnimationFrame` is thread safe.

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -36,18 +41,22 @@ - (void)requestAnimationFrame:(std::function<void(double)>)callback
...
 {
+  // Called on com.meta.react.turbomodulemanager.queue
   [displayLink_ invalidate];
 }
 
 #pragma mark-- private
 
 - (void)scheduleQueueExecution
 {
+  RCTAssertMainQueue();
```
tjzel, line 52: Please revert this change, `scheduleQueueExecution` is thread safe.

packages/react-native-worklets/apple/worklets/apple/WorkletsModule.mm
```diff
@@ -27,6 +28,7 @@ @implementation WorkletsModule {
 
 - (std::shared_ptr<WorkletsModuleProxy>)getWorkletsModuleProxy
 {
+  AssertJavaScriptQueue();
```
tjzel, line 31: Please revert this change, `getWorkletsModuleProxy` should be available from any thread (we'll refactor this bit in the future).

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
tomekzaw: Reverted in https://github.com/software-mansion/react-native-reanimated/pull/7161.

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
tomekzaw: Reverted in https://github.com/software-mansion/react-native-reanimated/pull/7161.

packages/react-native-worklets/apple/worklets/apple/WorkletsModule.mm
tomekzaw: This looks more like a private method of `WorkletsModule` and currently it's used only once on the JS thread. Maybe let's leave it for now and remove the assert the very first moment it misbehaves?

PR #7169, opened 2025-03-04
Remove implementation of `UIManagerListener` and `UIManagerModuleListener` interfaces in `ReanimatedModule` on Android

tomekzaw:
## Summary

Currently, `mOperations` in `ReanimatedModule` is never populated, so it can be removed together with `UIThreadOperation` interface and `willDispatchViewUpdates` interface methods from `UIManagerModuleListener` and `UIManagerListener` interfaces, along with all the remaining methods which are currently no-ops. This means we also need to remove `addUIManagerEventListener` and `removeUIManagerEventListener`. Turns out, we don't need `FabricUIManager` as well. After all these changes there will be only one runnable to execute in `invalidate` method so I also simplified `mUnsubscribe` assignment.

## Test plan

Build, launch and play with fabric-example on Android.

### DIFF
--- a/packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
+++ b/packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
@@ -10,8 +10,7 @@ import java.util.Objects;
 import javax.annotation.Nullable;
 
 @ReactModule(name = ReanimatedModule.NAME)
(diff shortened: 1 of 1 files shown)


zispidd:
I'm experiencing a crash with Android build version 0.82.1 of React Native and Reanimated 3.19.3.
I can't update to version 4 because not all dependencies support it. What can I do?

tomekzaw:
@zispidd Please report an issue.

PR #7176, opened 2025-03-05
Merge `getUIManager` into `installTurboModule` in `REAModule`

tomekzaw:
## Summary

Looks like `getUIManager` method is used only once so let's move its logic directly into `installTurboModule` method so we can optimize the code further in the future.

## Test plan

See if CI is green.


tjzel (approved):
Having a separate method looks better for me, but if there will be continuation of this it's ok 👍

piaskowyk (approved):

PR #7183, opened 2025-03-05
feat: Proper multiple CSS animations support in native

MatiPl01:
## Summary

This PR adds proper support for multiple animations.

Before, if the new animation was added, animation was removed or the order of animation was changed, all animations were created from scratch resulting in restart of animations that didn't change.

This PR changes this behavior to reuse the existing animation when possible, which is how it works on the web in CSS animations.

## Example recordings

| Before | After |
|-|-|
| <video src="https://github.com/user-attachments/assets/cc484806-d38a-4741-8ea9-77eb1a45f052" /> | <video src="https://github.com/user-attachments/assets/0f6f9ee6-aeeb-4e21-9b78-7273e48d3933" /> |

## Test plan

Open the `Miscellaneous` -> `Multiple Animations` screen and see how it works after changes.


tomekzaw (approved):
code lgtm

PR #7186, opened 2025-03-06
Create `NodesManager` instance in `ReanimatedModule` constructor instead of in getter method

tomekzaw:
## Summary

This PR simplifies the logic of creating and passing `NodesManager` instance on Android.

The conditional in `getNodesManager` has literally always been there 🤯 

Screenshot 2025-03-06 at 11 08 59

## Test plan

Build and launch fabric-example on Android.


tjzel (approved):
I wonder if this was done for lazy-loading or because previously it had to be initialized on the UI thread.

piaskowyk (approved):

PR #7193, opened 2025-03-06
Convert `WorkletsModule` to a local variable in `ReanimatedModule` constructor

tomekzaw:
## Summary

Turns out, we don't use `mWorkletsModule` in `ReanimatedModule` class so let's convert it to a local variable. This means that the lifetime of `WorkletsModule` will not longer be affected by the lifetime of `ReanimatedModule`.

## Test plan

Build, launch and reload fabric-example on Android.

### DIFF
--- a/packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
+++ b/packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
@@ -13,8 +13,8 @@ public class ReanimatedModule extends NativeReanimatedModuleSpec implements Life
   public ReanimatedModule(ReactApplicationContext reactContext) {
     super(reactContext);
     reactContext.assertOnJSQueueThread();
-    WorkletsModule mWorkletsModule = reactContext.getNativeModule(WorkletsModule.class);
-    mNodesManager = new NodesManager(reactContext, mWorkletsModule);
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):

tjzel:
I'm blocking this PR since I think we do benefit from keeping the module as a member as dogfooding of our API.

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
```diff
@@ -9,12 +9,11 @@
...
 public class ReanimatedModule extends NativeReanimatedModuleSpec implements LifecycleEventListener {
   private final NodesManager mNodesManager;
-  private final WorkletsModule mWorkletsModule;
 
   public ReanimatedModule(ReactApplicationContext reactContext) {
     super(reactContext);
     reactContext.assertOnJSQueueThread();
-    mWorkletsModule = reactContext.getNativeModule(WorkletsModule.class);
+    WorkletsModule mWorkletsModule = reactContext.getNativeModule(WorkletsModule.class);
     mNodesManager = new NodesManager(reactContext, mWorkletsModule);
```
tomekzaw:
```suggestion
    WorkletsModule workletsModule = reactContext.getNativeModule(WorkletsModule.class);
    mNodesManager = new NodesManager(reactContext, workletsModule);
```

tjzel:
#8949 supersedes this PR.

PR #7196, opened 2025-03-06
fix: crash on OTA reload

### LINKED ISSUE #7182: Getting null pointer exception in workletmodule.java after immediate auto restart after codepush bundle download/install in React Native Android app
### Description

I have a React Native Android app on RN version 0.78.0. Also, I have codepush integrated for OTA updates. On app launch, I show a splash screen, and in the background, my app downloads and installs the latest code push bundle. Once it is downloaded and installed, the app restarts, then and there I am getting the below exception. I have the latest react native reanimated version 3.17.1 installed.

### Steps to reproduce

I have a React Native Android app on RN version 0.78.0. Also, I have codepush integrated for OTA updates. On app launch, I show a splash screen, and in the background, my app downloads and installs the latest code push bundle. Once it is downloaded and installed, the app restarts, then and there I am getting the below exception. I have the latest react native reanimated version 3.17.1 installed.

Image

https://github.com/user-attachments/assets/ac265f2d-61db-4175-a0cf-b35b808e874d

### Reanimated version

3.17.1

### React Native version

0.78.0

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

### Device

Real device

### Device model

Oneplus 6T with android 11

### Acknowledgements

Yes

tjzel:
## Summary

Fixes #7182
Requires #7197

Currently on iOS React Native can go into a non-fatal race condition on a double reload. Double reload can happen during an OTA update, when an app is reloaded immediately after evaluating the bundle.

The flow goes more or less like this. I haven't explored it to the deepest details.
1. On the first reload a new JavaScript runtime is created.
2. The new runtime evaluates the bundle, starts to populate `RCTModuleRegistry`.
3. On the second reload a new JavaScript runtime is created. The old runtime remains and keeps executing. Both runtimes execute on the same thread, on the same queue.
4. `WorkletsModule.installTurboModule()` is called from the first runtime. `WorkletsModule` is fetched from the `RCTModuleRegistry`.
5. The second runtime overwrites (clears) the `RCTModuleRegistry` and starts to populate it. I don't know why both runtimes target the same instance of `RCTModuleRegistry`.
6. `WorkletsModule.installTurboModule()` finishes successfully on the first runtime.
7. `ReanimatedModule.installTurboModule()` is invoked in quick succession on the first runtime.
8. `ReanimatedModule` tries to fetch `WorkletsModule` from `RCTModuleRegistry`. However, `RCTModuleRegistry` was wiped and it doesn't have the fully installed instance of `WorkletsModule` anymore. Therefore a new instance of `WorkletsModule` is created and inserted into the `RCTModuleRegistry`.
9. In JavaScript we guarantee that `ReanimatedModule.installTurboModule()` is invoked only after `WorkletsModule.installTurboModule()`. We follow that heuristic in ObjC and therefore try operating on a non-installed instance of `WorkletsModule` resulting in a `NullPtrException`.

We can effectively detect that flow and ignore it, since that race condition is fatal only for the first runtime which is torn down soon after anyway.

See the provided screenshots which illustrate the problems.

| Screenshot 2025-03-07 at 10 34 00 | Screenshot 2025-03-07 at 12 25 23 | 
|:---|:---| 
| Setting `moduleRegistry` for `WorkletsModule` during a healthy execution of `WorkletsModule.installTurboModule()` | Setting `moduleRegistry` for `WorkletsModule` during race-condition execution. Note that the list of the modules is much shorter and that `WorkletsModule` isn't there. |

## Test plan

To reproduce this issue, reload twice on iOS in `fabric-example`. For best effect try to reload after ~0.5s after the first time, but it's easily reproducible if the App isn't slowed heavily by debugger instructions.

With these changes Reanimated doesn't crash, but the race condition continues to manifest.


tjzel:
cc @lukmccall

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
```diff
@@ -249,3 +256,38 @@ See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooti
     this.#reanimatedModuleProxy.unregisterCSSTransition(viewTag);
   }
 }
+
+class DummyReanimatedModuleProxy implements ReanimatedModuleProxy {
```
piaskowyk, line 260: I've asked my friend GPT about JSProxy that provides a default fallback implementation for undefined methods, and he suggested me something like that:
```
// Define a default method to be used when a non-existent method is called
function defaultMethod(...args) {
    console.log("Default method triggered with arguments: ", args);
}

// Create a handler object with a `get` trap
const handler = {
    get(target, prop, receiver) {
        // If the property exists, return it
        if (prop in target) {
            return target[prop];
        }
        
        // If the property does not exist, return the default method
        return function(...args) {
            defaultMethod(...args);
        };
    }
};

// Define an object with some methods
const originalObject = {
    existingMethod: function() {
        console.log("Existing method called");
    }
};

// Create a proxy for the object
const proxyObject = new Proxy(originalObject, handler);

// Test invoking existing method
proxyObject.existingMethod();  // Output: Existing method called

// Test invoking a non-existent method
proxyObject.nonExistentMethod(1, 2, 3);  // Output: Default method triggered with arguments: [ 1, 2, 3 ]
```
we could avoid listing all methods here, and it would work for methods that we add in a future

tjzel: While using proxy looks cool I think simpler code would be more maintainable. You also you need extra methods for functions that return arguments, like `subscribeForKeyboardEvent` etc.

piaskowyk (approved):

RSNara:
> I don't know why both runtimes target the same instance of RCTModuleRegistry.

RCTHost can leak a reference to the module registry to the application. We didn't want that reference to break across reloads. That's why we had the host own the module registry.

But, yeah. This is a little bit problematic. Now, it means that two different interleaving react instances can share the same module registry, leading to undefined behaviour. In this case, the first react instance is given access the modules from the second. 

The module registry should probably be owned by the react instance. But, we need to make sure we don't break the references returned from react host. Or, make sure that the host only returns weak pointers to the module registry to the application. That way, people have to check if the module registry still exists, before using it.

tjzel:
@RSNara Yeah, I encountered this problem before but didn't pay it much attention since it happened only on multiple reloads. It came back at me when it turned out that OTA updates are done via a double reload 😿 It's an easy fix on our side since we only obtain a TurboModule once, but libraries which pull them multiple times in runtime could be in more trouble.

PR #7197, opened 2025-03-07
fix: invalid Android invalidation of hybrid data

tjzel:
## Summary

When React Native is reloaded when `WorkletsModule.installTurboModule()` wasn't called yet we tried to invoke `invalidateCpp()` which resulted in a nullptr exception, because hybrid data wasn't yet initialized.

I applied similar safety mechanism to all relevant code, turns out it narrowed down only to `NativeProxy.java`.

## Test plan

See that App no longer crashes on `invalidateCpp()` on a double reload on Android.


packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -114,7 +116,12 @@ protected HybridData getHybridData() {
   }
 
   public void invalidate() {
-    invalidateCpp();
+    if (mInvalidated.getAndSet(true)) {
+      return;
```
piaskowyk, line 120: I'm not entirely sure I understand why we need to synchronize the `mInvalidated` variable. Does this mean that `invalidate` can be called multiple times from different threads for the same object? 🤔

piaskowyk: Let's add a comment in the code about that because it's not obvious, and we can merge 👍

piaskowyk (approved):

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NativeProxy.java
tomekzaw: Why can't we make it protected then?

tjzel: Changing to `protected` is a good idea, but I'd still keep synchronization as a security measure.

tjzel: @piaskowyk added the comment

tjzel: @tomekzaw changed to `protected`

tomekzaw:
Wouldn't it be enough if `invalidate` method would be just `synchronized`?

PR #7206, opened 2025-03-07
Call `NodesManager.initWithContext` and thus create `NativeProxy` in `ReanimatedModule` constructor

tomekzaw:
## Summary

Currently, we first create `ReanimatedModule` and `NodesManager` in its constructor. Then, calling `ReanimatedModule.installTurboModule` invokes `NodesManager.initWithContext` which creates `NativeProxy` and injects JSI bindings in its constructor. The part I don't like here is that `NodesManager` stays unconfigured for some time (between constructor and `initWithContext` calls). Also, there's some time period when `ReanimatedModule` exists but `NativeProxy` doesn't exist yet, even though it will be necessary, so why don't we just instantiate it earlier?

After this change, all three `ReanimatedModule`, `NodesManager` and `NativeProxy` instances will be created at the same time (but without injecting JSI bindings). Then, calling `installTurboModule` will delegate the call to already existing `NativeProxy` which will inject the bindings at the right time.

## Test plan

Build, launch and reload fabric-example on Android.

### DIFF
--- a/packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
+++ b/packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
@@ -16,6 +16,8 @@ public class ReanimatedModule extends NativeReanimatedModuleSpec implements Life
     reactContext.assertOnJSQueueThread();
     mWorkletsModule = reactContext.getNativeModule(WorkletsModule.class);
     mNodesManager = new NodesManager(reactContext, mWorkletsModule);
+    mNodesManager.initWithContext(reactContext);
+    // TODO: merge initWithContext into NodesManager constructor
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
```diff
@@ -16,6 +16,8 @@ public ReanimatedModule(ReactApplicationContext reactContext) {
     reactContext.assertOnJSQueueThread();
     mWorkletsModule = reactContext.getNativeModule(WorkletsModule.class);
     mNodesManager = new NodesManager(reactContext, mWorkletsModule);
+    mNodesManager.initWithContext(reactContext);
+    // TODO: merge initWithContext into NodesManager constructor
```
tomekzaw, line 20: This will be done in a follow-up PR for the sake of atomicity.

PR #7208, opened 2025-03-07
chore(worklets): fix android studio warnings

tjzel:
## Summary

🚀

### DIFF
--- a/packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/AndroidUIScheduler.java
+++ b/packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/AndroidUIScheduler.java
@@ -7,16 +7,11 @@ import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.UiThreadUtil;
 import java.util.concurrent.atomic.AtomicBoolean;
 
-/**
- * @noinspection JavaJniMissingFunction
- */
+@SuppressWarnings("JavaJniMissingFunction")
 public class AndroidUIScheduler {
 
-  /**
-   * @noinspection FieldCanBeLocal
-   */
   @DoNotStrip
-  @SuppressWarnings("unused")
+  @SuppressWarnings({"unused", "FieldCanBeLocal"})
   private final HybridData mHybridData;
 
   private final ReactApplicationContext mContext;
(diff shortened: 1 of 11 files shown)


packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/AndroidUIScheduler.java
```diff
@@ -18,11 +24,11 @@ public class AndroidUIScheduler {
 
   private final Runnable mUIThreadRunnable =
       () -> {
-        /**
-         * This callback is called on the UI thread, but the module is invalidated on the JS thread.
-         * Therefore we must synchronize for reloads. Without synchronization the cpp part gets torn
-         * down while the UI thread is still executing it, leading to crashes.
-         */
+        /// This callback is called on the UI thread, but the module is invalidated on the JS
```
tomekzaw, line 27:
```suggestion
        // This callback is called on the UI thread, but the module is invalidated on the JS
```

packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/AndroidUIScheduler.java
```diff
@@ -18,11 +24,11 @@ public class AndroidUIScheduler {
...
+        // thread.
+        /// Therefore we must synchronize for reloads. Without synchronization the cpp part gets
+        // torn
+        /// down while the UI thread is still executing it, leading to crashes.
```
tomekzaw, line 31:
```suggestion
        // torn down while the UI thread is still executing it, leading to crashes.
```

packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/AndroidUIScheduler.java
```diff
@@ -41,6 +47,9 @@ public AndroidUIScheduler(ReactApplicationContext context) {
 
   public native void invalidate();
 
+  /**
+   * @noinspection unused
+   */
```
tomekzaw, line 52: How about `@SupressWarnings("unused")`?

packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/AnimationFrameQueue/AnimationFrameCallback.java
```diff
@@ -1,4 +1,4 @@
-package com.swmansion.worklets;
+package com.swmansion.worklets.AnimationFrameQueue;
```
tomekzaw, line 1: Can we please avoid using `com.swmansion.worklets.AnimationFrameQueue` package in favor of `com.swmansion.worklets`? There could be a name conflict with `AnimationFrameQueue` from `com.swmansion.reanimated` package (somehow), if that's the case please resolve the conflict or rename one of them.

tomekzaw (approved):
Thanks for fixing this!

PR #7218, opened 2025-03-10
(package.json) add keywords to package.json to improve npm search result

MoOx:
## Summary

Currently, if you search on npm for "react native reanimated", you are no the top result ! You need to type "react-native-reanimated" (with dashes) to be at the top. But without dashes, you should still be visible. I would expect this tiny improvement to help about this.

Screenshot 2025-03-10 at 09 24 44

Screenshot 2025-03-10 at 09 24 52

Screenshot 2025-03-10 at 09 24 35

## Test plan

I would expect this keywords to help. We can only be sure after next release.


tjzel (approved):
@MoOx Thanks!

PR #7219, opened 2025-03-10
feat: support for cloning Set and Map

tjzel:

### DIFF
--- a/packages/react-native-worklets/src/shareables.ts
+++ b/packages/react-native-worklets/src/shareables.ts
@@ -164,6 +164,12 @@ function makeShareableCloneRecursiveNative<T>(
     // typed array (e.g. Int32Array, Uint8ClampedArray) or DataView
     return cloneArrayBufferView(value);
   }
+  if (value instanceof Set) {
+    return cloneSet(value);
+  }
+  if (value instanceof Map) {
+    return cloneMap(value);
+  }
   return inaccessibleObject(value);
 }
 
@@ -421,6 +427,36 @@ function cloneArrayBufferView<T extends ArrayBufferView>(
     },
   }) as unknown as ShareableRef<T>;
   shareableMappingCache.set(value, handle);
+
+  return handle;
+}
+
+function cloneSet<TSet extends Set<unknown>>(set: TSet): ShareableRef<TSet> {
+  const elements = Array.from(set);
+
+  const handle = makeShareableCloneRecursive({
+    __init: () => {
+      'worklet';
+      return new Set(elements);
+    },
+  }) as unknown as ShareableRef<TSet>;
+  shareableMappingCache.set(set, handle);
+
+  return handle;
+}
+
+function cloneMap<TMap extends Map<unknown, unknown>>(
+  map: TMap
+): ShareableRef<TMap> {
+  const entries = Array.from(map.entries());
+
+  const handle = makeShareableCloneRecursive({
+    __init: () => {
+      'worklet';
+      return new Map(entries);
+    },
+  }) as unknown as ShareableRef<TMap>;
+  shareableMappingCache.set(map, handle);
 
   return handle;
 }


apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/shareables/makeShareableClone.test.tsx
```diff
@@ -323,6 +323,70 @@ describe('Test makeShareableClone', () => {
...
+    // Act
+    await render(
+      <ValueComponent
+        onRunUIFunction={() => {
+          'worklet';
+          const checks = [
+            setValue.has(1),
+            setValue.has('1'),
+            setValue.has(true),
+            // setValue.has(/1/),
```
skusnierz: is it needed ?

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/shareables/makeShareableClone.test.tsx
```diff
@@ -323,6 +323,70 @@ describe('Test makeShareableClone', () => {
...
+    // Act
+    await render(
+      <ValueComponent
+        onRunUIFunction={() => {
+          'worklet';
+          const checks = [
+            mapValue.get(1) === 2,
+            mapValue.get('1') === '2',
+            mapValue.get(true) === false,
+            // mapValue.get(/1/).toString() === '/2/',
```
skusnierz: is it needed?

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/shareables/makeShareableClone.test.tsx
tjzel: Forgot to remove these 🤯

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/shareables/makeShareableClone.test.tsx
```diff
@@ -323,6 +323,68 @@ describe('Test makeShareableClone', () => {
...
+            setValue.has(true),
+            setValue.size === 3,
+            typeof setValue === 'object',
+            setValue instanceof Set,
+          ];
+          return checks.every(Boolean);
+        }}
+      />,
+    );
+    await wait(100);
```
piaskowyk, line 347: Just a nitpick—If you don't want hardcoded waiting times, you can use the pattern `await waitForNotify('notifyName');` and call `notify('notifyName');` in `ValueComponent`.

```javascript
try {
  const result = onRunUIFunction();
  sharedResult.value = result ? 'ok' : 'not_ok';
  notify('notifyName');
```

This approach would be flaky-proof and speed up test durations.

packages/react-native-worklets/src/shareables.ts
```diff
@@ -464,6 +470,39 @@ function clonePlainJSObject<T extends object>(
   return clone;
 }
 
+function cloneMap<T extends Map<unknown, unknown>>(value: T): ShareableRef<T> {
+  const clonedKeys: unknown[] = [];
+  const clonedValues: unknown[] = [];
+  for (const [key, element] of value.entries()) {
+    clonedKeys.push(makeShareableCloneRecursive(key));
+    clonedValues.push(makeShareableCloneRecursive(element));
+  }
```
piaskowyk, line 479: I see that we need to invoke quite a lot of JSI calls. I'm curious about how performant this would be with a larger dataset. Do you have any measurements?

tjzel: I don't. I figured that a good enough implementation is sufficient in the first iteration and if need be we could optimize it in the future.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/shareables/makeShareableClone.test.tsx
tjzel: We should address that in an upcoming PR since it's a recurring pattern in these tests. cc @skusnierz

PR #7223, opened 2025-03-10
refactor: lint worklets like reanimated

tjzel:
## Summary

`react-native-worklets` workspace only used a subset of linting tools that `react-native-reanimated` uses. Now they both have the exact level of linting.

## Test plan

CI green


.github/workflows/example-android-build-check.yml
```diff
@@ -29,12 +29,11 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  example-android-build-check:
```
tomekzaw, line 32: This name appears in GitHub UI in the "checks" section and because of that we don't want it to be that long. Why not just `build`?

.github/workflows/worklets-common-validation.yml
```diff
@@ -0,0 +1,52 @@
...
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  worklets-common-validation:
+    if: github.repository == 'software-mansion/react-native-reanimated'
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.13]
```
tomekzaw, line 29: Let's use env instead

.github/workflows/worklets-common-validation.yml
```diff
@@ -0,0 +1,52 @@
...
+    env:
+      WORKLETS_DIR: packages/react-native-worklets
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install python dependencies
```
tomekzaw, line 38:
```suggestion
      - name: Install Python dependencies
```

.github/workflows/worklets-common-validation.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Install monorepo node dependencies
+        run: yarn install --immutable
+
+      - name: Lint Common
+        working-directory: ${{ env.WORKLETS_DIR }}
+        run: yarn lint:common
+      - name: Disallow DEBUG macros
+        working-directory: ${{ env.WORKLETS_DIR }}
+        run: |
+          ! egrep -r '(#if DEBUG|#ifdef DEBUG)' Common/cpp/ apple/ android/src/main/cpp/
```
tomekzaw, line 52: Let's move that to a script file, like `validate-apple.sh`.

.github/workflows/worklets-static-checks.yml
```diff
@@ -49,3 +51,9 @@ jobs:
       - name: Find circular dependencies
         working-directory: ${{ env.PACKAGE_PATH }}
         run: yarn circular-dependency-check
+      - name: Check use strict
+        working-directory: ${{ env.PACKAGE_PATH }}
+        run: yarn use-strict-check
+      - name: Run jest unit tests
```
tomekzaw, line 57:
```suggestion
      - name: Run Jest unit tests
```

scripts/validate-common.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+if grep -Rn Common/cpp -e '^#include "'; then
+    echo 'Found `#include "..."` in `Common/cpp/` directory. Convert it to `#include <...>`, e.g. `#include <reanimated/Fabric/updates/UpdatesRegistryManager.h>`.'
```
tomekzaw, line 4: Let's use something else as an example
```suggestion
    echo 'Found `#include "..."` in `Common/cpp/` directory. Convert it to `#include <...>`, e.g. `#include <reanimated/NativeModules/ReanimatedModuleProxy.h>`.'
```

packages/react-native-worklets/package.json
```diff
@@ -11,21 +11,30 @@
...
     "lint:js": "eslint src && yarn prettier --check src",
     "lint:plugin": "yarn workspace babel-plugin-reanimated lint",
     "test": "jest",
-    "type:check": "yarn type:check:src && yarn type:check:plugin",
+    "type:check": "yarn type:check:src && yarn type:check:plugin && yarn type:check:app",
     "type:check:src": "yarn tsc --noEmit",
-    "type:check:plugin": "yarn workspace babel-plugin-reanimated type:check"
+    "type:check:app": "yarn workspace common-app type:check",
+    "type:check:plugin": "yarn workspace babel-plugin-reanimated type:check",
+    "use-strict-check": "node ../../scripts/validate-use-strict.js"
```
tomekzaw, line 37: This is some kind of linting so perhaps we should use `lint:strict` instead? Same fo `circular-dependency-check` and `find-unused-code:js`. We can save that for another PR, though.

packages/react-native-reanimated/package.json
tomekzaw, line 1: Just a general comment, would it be possible to centralize some of those scripts in the main `package.json` instead of having two copies, one in worklets and one in reanimated?

tomekzaw (changes requested):

packages/react-native-worklets/package.json
tjzel: Same thoughts bro 🦾

.github/workflows/example-android-build-check.yml
tjzel: Actually we want it to be that long. Sometimes `build` is the only thing that is displayed and you need extra clicks to check what CI has failed. Moreover, unique job names allow us to block merging until they pass. It's an agenda I've been pushing secretly over the months.

See that here you only see `check` and you don't actually know which CI failed.
Screenshot 2025-03-13 at 10 51 59

.github/workflows/worklets-common-validation.yml
```diff
@@ -0,0 +1,52 @@
...
+    if: github.repository == 'software-mansion/react-native-reanimated'
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.13]
+    env:
+      WORKLETS_DIR: packages/react-native-worklets
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
```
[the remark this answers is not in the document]
tjzel: Screenshot 2025-03-13 at 11 06 09

.github/workflows/example-android-build-check.yml
tomekzaw: I see your point, here's that I meant, see the duplicate names:

Screenshot 2025-03-13 at 11 35 26

tomekzaw (approved):

PR #7226, opened 2025-03-11
refactor: remove runOnUIImmediately

tjzel:
## Summary

`runOnUIImmediately` is not conforming with a well defined event loop model of the UI Runtime, so we should remove it.

I considered that we can replace it with:
1. `runOnUI`
2. `executeOnUIRuntimeSync`
3. Additional batching mechanism on the side of Reanimated, using `runOnUI`.

Of those I went for **1.** for the sake of simplicity and didn't notice any regressions. **2.** would be the second choice since it actually preserves the logic but could be potentially slowing down the initialization process.

Please submit any suggestions on how to test this change more thoroughly. 

## Test plan

Run an example and see that all animations run well 🚀

### DIFF
--- a/packages/react-native-reanimated/src/UpdateProps.ts
+++ b/packages/react-native-reanimated/src/UpdateProps.ts
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-redundant-type-constituents */
 'use strict';
 import type { MutableRefObject } from 'react';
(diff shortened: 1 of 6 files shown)


piaskowyk (approved):
This is a potentially risky change, but given that it maintains the same order of execution as before and we don't call JS functions from native code during initialization, it seems okay. I think it's a good time to take the risk and see what happens.

PR #7231, opened 2025-03-12
feat: Add support for animated props in Jest

patrycjakalinska:
## Summary

It turned out animated props when tested in Jest kept their initial value. It was caused by lack of support for animatedProps, which this PR adds. 

To-do:
- [x] inspect why svg doesn't work -> because it only accepts predefined props

## Test plan

🪨


packages/react-native-reanimated/__tests__/animatedProps.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+import React from 'react';
+import { Button, TextInput, View } from 'react-native';
+import Animated, {
+  useAnimatedProps,
+  useSharedValue,
+  withTiming,
+} from 'react-native-reanimated';
+import { Circle, Svg } from 'react-native-svg';
+
+Animated.addWhitelistedNativeProps({ text: true });
```
tjzel, line 11: Why do you have to add "text" to the whitelist but not "r" from the Circle component?

packages/react-native-reanimated/__tests__/animatedProps.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+    return {
+      text: `Box width: ${width.value}`,
+      defaultValue: `Box width: ${width.value}`,
+    };
+  });
+
+  return (
+    <View>
+      <Svg>
+        // This won't work - SVG has guard for props they pass to a component 👇
```
tjzel, line 39: What do you mean by this comment?

packages/react-native-reanimated/__tests__/animatedProps.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const r = useSharedValue(20);
+  const width = useSharedValue(20);
+
+  const handlePress = () => {
+    r.value += 10;
+    width.value += 10;
+  };
+
+  const animatedProps = useAnimatedProps(() => ({
+    r: withTiming(r.value, { duration: 100 }),
```
tjzel, line 26: Define it in the upper scope.

```suggestion
    r: withTiming(r.value, { duration: animationDuration }),
```

packages/react-native-reanimated/__tests__/animatedProps.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+  });
+  test('Custom animated component', () => {
+    const { getByTestId } = render(<AnimatedComponent />);
+    const textInput = getByTestId('text');
+    const button = getByTestId('button');
+
+    expect(textInput).toHaveAnimatedProps({ text: 'Box width: 20' });
+
+    fireEvent.press(button);
+    jest.advanceTimersByTime(100);
```
tjzel, line 80: Reuse here.

```suggestion
    jest.advanceTimersByTime(animationDuration);
```

packages/react-native-reanimated/src/createAnimatedComponent/AnimatedComponent.tsx
```diff
@@ -233,6 +236,16 @@ export default class AnimatedComponent
...
     }
 
+    if (animatedProps && IS_JEST) {
+      this.jestAnimatedProps.value = {
+        ...this.jestAnimatedProps.value,
+        ...animatedProps?.initial?.value,
+      };
+      // @ts-ignore It tells jestAnimatedValues is unknown yet it's part of animatedProps
+      // TODO: fix this somewhat in the future
+      animatedProps.jestAnimatedValues.current = this.jestAnimatedProps;
```
tjzel, line 246: Need a hand with that?

packages/react-native-reanimated/src/hook/commonTypes.ts
```diff
@@ -93,9 +94,11 @@ export interface AnimatedStyleHandle<
 }
 
 export interface JestAnimatedStyleHandle<
-  Style extends DefaultStyle = DefaultStyle,
+  Style extends DefaultStyle | AnimatedProps = DefaultStyle,
 > extends AnimatedStyleHandle<Style> {
-  jestAnimatedStyle: MutableRefObject<AnimatedStyle<Style>>;
+  jestAnimatedValues:
+    | MutableRefObject<AnimatedStyle<Style>>
+    | MutableRefObject<AnimatedProps>;
```
tjzel, line 101: Shouldn't it be an intersection?

```suggestion
  jestAnimatedValues:
     MutableRefObject<AnimatedStyle<Style>>
    & MutableRefObject<AnimatedProps>;
```

yarn.lock
tjzel, line 1: Why did the `yarn.lock` of the monorepo change? You haven't changed any dependencies. Please revert it and see if it changes again on install.

tjzel (changes requested):
Well done, but we need to polish some rough spots.

BTW we should really clean up this code at one point, it's such a mess 😢

packages/react-native-reanimated/__tests__/animatedProps.test.tsx
patrycjakalinska: Ths SVG with animatedProps cannot be tested because they filter props they recieve (they have what i;ve called the guard for props) and therefore remove our `jestAnimatedStyles` and `jestAnimatedProps` crucial for testing. I wanted to leave this test to emphasize this matter - I can change it to more descriptive like: `This won't work - SVG has a guard for props they pass to a component - therefore they remove jest props crucial for testing`

packages/react-native-reanimated/__tests__/animatedProps.test.tsx
patrycjakalinska: yes 👍

packages/react-native-reanimated/__tests__/animatedProps.test.tsx
tjzel: So, if I understand correctly, it's impossible to test animatedProps on SVG components? Then we can just say "// SVG components strip our jest props and cannot be tested"

patrycjakalinska: will do that 🪨

packages/react-native-reanimated/src/hook/commonTypes.ts
tjzel: Yeah, changing this bit would require a refactor, let's leave it as is then.

tjzel (approved):

packages/react-native-reanimated/__tests__/animatedProps.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    <View>
+      <Svg>
+        // SVG components strip our jest props and cannot be tested
+        <AnimatedCircle
+          cx="50%"
+          cy="50%"
+          fill="#b58df1"
+          testID={'circle'}
+          animatedProps={animatedProps}
+        />
```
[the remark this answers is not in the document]
patrycjakalinska: I left the test to stress the point that svg isn't supposed to work and the props shouldn't change (and maybe someday they will fail, letting us know that svg now works :"D) - but we can just put the info in docs and remove the test if you think it creates more confusion that good c:

You prefer separte PR for docs, or add disclaimer to docs in this PR?

piaskowyk (approved):

PR #7237, opened 2025-03-12
fix: CSS issues introduced by react-freeze fix

MatiPl01:
Opened this PR as a replacement for #7229

## Summary

It fixes issues introduced in the #7053 PR and keeps the necessary logic to ensure that `react-freeze` fixes work as well.

Basically, the problem is that we cannot trust `componentWillUnmount` calls because they are also called when component gets frozen (we don't want to remove animations in this case - it will be good to pause them in the future, but the original PR with react-native fix also didn't do that, so it will be addressed in the future).

To make it work, I decided to use `componentWillUnmount` only to mark nodes as removable (possible to remove). The final decision is made in the reanimated mount hook, where we check if the node was removed and clean up registries only if it no longer exits in the ShadowTree. In all other cases, we keep the node marked for removal as long as it is frozen and unmark it when it gets unfrozen.

## Test plan

1. Open the app, navigate between screens, see that everything works.
2. Open the drawer navigator and the *React Freeze* example to observe that everything behaves as expected.
3. Navigate between screens in this example and press on the *Check for registry leaks* button to observe that there are no leaks.


packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/UpdatesRegistryManager.cpp
```diff
@@ -40,6 +42,32 @@ bool UpdatesRegistryManager::shouldCommitAfterPause() {
...
+
+    const auto tag = shadowNode->getTag();
+    for (auto &registry : registries_) {
+      registry->remove(tag);
+    }
+    staticPropsRegistry_->remove(tag);
+    removableShadowNodes_.erase(shadowNode);
+  }
+}
+
```
MatiPl01, line 70: This is the most important part of this PR.

Unfortunately, when the node is frozen, we have to keep it in the `removableShadowNodes_` set as long as it exists in the ShadowTree because `componentWillUnmount` is called only when freezing happens. It won't be called again, when the respective component is finally removed, se we cannot remove this node from `removableShadowNodes_`, as long as the component exists and can be unfrozen.

packages/react-native-reanimated/src/css/managers/CSSAnimationsManager.ts
```diff
@@ -73,7 +61,19 @@ export default class CSSAnimationsManager {
     }
   }
 
-  private registerKeyframes(processedAnimations: ProcessedAnimation[]) {
+  unmountCleanup(): void {
+    this.unregisterKeyframesUsage();
```
MatiPl01, line 65: The only place where we have to clean up something from JS side. We need to do this here because CPP doesn't know when animation keyframes should be removed. Because of that, we have to remove keyframes from JS (even if component gets frozen).

For a case in which the component becomes frozen, removal of animation keyframes is safe as the animation object with keyframes shared pointer still exists in CPP, so this removal won't affect the native animation object. What is more, keyframes will be registered in CPP again when the animation with such keyframes is applied to this or a different view, so new animations will also be created properly.

packages/react-native-reanimated/src/css/managers/CSSAnimationsManager.web.ts
```diff
@@ -93,7 +90,11 @@ export default class CSSAnimationsManager {
     this.setElementAnimations(animationNames, animationSettings);
   }
 
-  detach() {
+  unmountCleanup(): void {
+    // noop
+  }
```
MatiPl01, line 95: The main cleanup is now done in CPP based on the ShadowTree changes. This `unmountCleanup` can be used for additional cleanup needed on the JS side (but it is also called when component gets frozen, so we should respect that). In most cases, no additional cleanup is needed - we need it only [here](https://github.com/software-mansion/react-native-reanimated/pull/7237/files#r1993416218).

PR #7244, opened 2025-03-13
chore(Reanimated3): fix CI

tjzel:

### DIFF
--- a/apps/macos-example/App.tsx
+++ b/apps/macos-example/App.tsx
@@ -1,3 +1,3 @@
-import App from 'common-app';
+import BokehExample from 'common-app/src/examples/BokehExample';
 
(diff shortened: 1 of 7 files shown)


tomekzaw (approved):

apps/macos-example/App.tsx
```diff
@@ -1,3 +1,3 @@
-import App from 'common-app';
+import BokehExample from 'common-app/src/examples/BokehExample';
```
tjzel, line 1: macos-example cannot use whole `common-app` because it doesn't support React 19

tomekzaw (approved):

tomekzaw (approved):

PR #7245, opened 2025-03-13
chore: amend compatibility table

tjzel:
## Summary

Reanimated 3.17 actually doesn't support 0.75.

## Test plan

🚀


piaskowyk (commented):
We should also update minimal supported version in build scripts:
- https://github.com/software-mansion/react-native-reanimated/blob/3.17-stable/packages/react-native-reanimated/scripts/reanimated_utils.rb#L56
- https://github.com/software-mansion/react-native-reanimated/blob/3.17-stable/packages/react-native-reanimated/android/build.gradle#L385

tjzel:
False alarm, it actually supports it 🤯

PR #7259, opened 2025-03-14
fix: CSS issues after RN upgrade

MatiPl01:


MatiPl01:
Closed in favor of #7296 because there were too many conflicts

PR #7270, opened 2025-03-18
fix(Worklets): crash on null context

tjzel:


tjzel:
Closing this since it doesn't cover all cases of the race condition, upstreamed to https://github.com/facebook/react-native/issues/50102

PR #7271, opened 2025-03-18
chore: enforce use of `globalThis._WORKLET`

tjzel:
## Summary

Using raw `_WORKLET` leads to "_WORKLET is not defined" error sometimes and is more error-prone than invoking `globalThis._WORKLET` directly.

In the future we might restrict implicit use of `global` even further.

## Test plan

Run FabricExample, NextExample and WebExample and see that all is well.

### DIFF
--- a/packages/react-native-reanimated/src/publicGlobals.ts
+++ b/packages/react-native-reanimated/src/publicGlobals.ts
@@ -14,7 +14,6 @@ declare global {
    * code shouldn't depend on it, we keep it here mainly for backward
    * compatibility for our users.
    */
-
   var _WORKLET: boolean | undefined;
 
   /**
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-reanimated/index.js
```diff
@@ -170,6 +170,50 @@ var require_noAnimatedStyleToNonAnimatedComponent = __commonJS({
   },
 });
 
+// public/useGlobalThis.js
+var require_useGlobalThis = __commonJS({
```
MatiPl01, line 174: This is a generated file, right? I wanted to ask why do you use `exports2`, `utils_1` with these number suffixes but, if this file is auto generated, then I have no concerns.

MatiPl01 (approved):
Looks fine

tomekzaw (approved):
Looks good to me, but what about the developers? Using `if (_WORKLET)` inside `useAnimatedStyle` worklet is a common practice.

packages/eslint-plugin-reanimated/index.js
tjzel: It is auto-generated.

tjzel:
@tomekzaw Then the user will get a meaningful stack-trace to the place where they used `_WORKLET`. This is only to make errors within Reanimated & Worklets clearer.

PR #7107, opened 2025-02-26
feat(worklets): prepare releasing pipeline

tjzel:
## Summary

To start releasing nightlies of `react-native-worklets` and make the process of releasing it similar to how it's done for `react-native-reanimated` we need to refactor some bits.

I extracted the core of determining the version (custom, nightly or fresh) to a separate script in the root of the repo.

## Test plan

Run the scripts and see that the packages are built correctly.

PR #7276, opened 2025-03-19
fix: createNPMPackage.sh not working without parameters

tjzel:
## Summary

After the changes to the releasing scripts I introduced a regression where calling `createNPMPackage.sh` without parameters would result in an error.

#7107 

## Test plan

🚀


MatiPl01 (approved):
Didn't test it out but looks good.

patrycjakalinska (approved):
it works good and looks good

PR #7280, opened 2025-03-19
feat: preworkletized target

tjzel:
## Summary

Transpilation of Reanimated's worklets takes a significant amount of time due to Worklets Babel plugin invocation of `transformSync`. See https://github.com/byCedric/expo-babel-benchmarks for more details. Adding pre-workletized target for our libraries reduces the transpilation time for Worklets Babel plugin to effectively zero, speeding up bundling.

## Test plan

Generate new targets, in Reanimated move `src` to `src.bak`, copy `lib/preworkletized` to `src`, delete `lib/module` and see that WebExample works as expected.

You should be able to specify `metroConfig.resolver.resolverMainFields = ['preworkletized']` [LINK](https://metrobundler.dev/docs/configuration/#resolvermainfields) but it doesn't work in WebExample for some reason, cc @MatiPl01 on that.

### DIFF
--- a/packages/react-native-reanimated/package.json
+++ b/packages/react-native-reanimated/package.json
@@ -30,7 +30,8 @@
     "type:check:src": "yarn tsc --noEmit",
     "type:check:app": "yarn workspace common-app type:check",
     "type:check:tests:common": "./scripts/test-ts.sh __typetests__/common",
-    "build": "yarn workspace react-native-worklets build && bob build",
+    "build": "yarn workspace react-native-worklets build && bob build && yarn build:preworkletized",
+    "build:preworkletized": "yarn babel --out-dir lib/preworkletized --config-file ../react-native-worklets/preworkletized.babel.config.js --extensions \".js,.jsx\" --source-maps true lib/module",
     "circular-dependency-check": "yarn madge --extensions js,jsx,ts,tsx --circular src lib",
     "use-strict-check": "node ../../scripts/validate-use-strict.js",
     "prepack": "cp ../../README.md ./README.md",
@@ -40,6 +41,7 @@
   "module": "lib/module/index",
   "react-native": "src/index",
   "source": "src/index",
+  "preworkletized": "lib/preworkletized/index",
   "types": "lib/typescript/index.d.ts",
   "files": [
     "Common/",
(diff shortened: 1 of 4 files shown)


EvanBacon:
`preworkletized` implies that the code that's shipped must be transformed special for a new target. But in reality, it's already transformed for JavaScript. The opposite case is probably what would be needed where a new field indicates that reanimated code is shipped as-is and must be transformed out.

tjzel:
@EvanBacon Makes sense, do you have some suggestion how to name it?

piaskowyk (commented):
Overall, it looks good. I'm just wondering if the types will resolve correctly when we export it as JavaScript code.

tjzel:
@piaskowyk TypeScript should still read types from `types` target.

piaskowyk (commented):
I've tested it like that:
```bash
cd packages/react-native-reanimated && yarn build:preworkletized
```
and it throws to me:
```
SyntaxError: /Users/piaskowyk/projects/react-native-reanimated/packages/react-native-reanimated/lib/module/component/FlatList.js: Support for the experimental syntax 'jsx' isn't currently enabled (11:12):

   9 | const createCellRendererComponent = itemLayoutAnimationRef => {
  10 |   const CellRendererComponent = props => {
> 11 |     return <AnimatedView
     |            ^
  12 |     // TODO TYPESCRIPT This is temporary cast is to get rid of .d.ts file.
  13 |     layout={itemLayoutAnimationRef?.current} onLayout={props.onLayout} style={props.style}>
  14 |         {props.children}

Add @babel/preset-react (https://github.com/babel/babel/tree/main/packages/babel-preset-react) to the 'presets' section of your Babel config to enable transformation.
If you want to leave it as-is, add @babel/plugin-syntax-jsx (https://github.com/babel/babel/tree/main/packages/babel-plugin-syntax-jsx) to the 'plugins' section to enable parsing.
```
When updated `preworkletized.babel.config.js` the problem was solved.
```diff
/** @type {import('@babel/core').TransformOptions} */
module.exports = {
+  presets: ['@babel/preset-react'],
  plugins: [['react-native-worklets/plugin', { processNestedWorklets: true }]],
};

```

PR #7281, opened 2025-03-20
Use iterative stack-based approach for flattenArray.

MegaManSec:
Previously, the _the _flattenArray() function could crash the interpreter due to infinite tail call recursion.

### DIFF
--- a/packages/react-native-reanimated/src/createAnimatedComponent/utils.ts
+++ b/packages/react-native-reanimated/src/createAnimatedComponent/utils.ts
@@ -7,19 +7,23 @@ export function flattenArray<T>(array: NestedArray<T>): T[] {
   if (!Array.isArray(array)) {
     return [array];
   }
-  const resultArr: T[] = [];
 
-  const _flattenArray = (arr: NestedArray<T>[]): void => {
-    arr.forEach((item) => {
-      if (Array.isArray(item)) {
-        _flattenArray(item);
-      } else {
-        resultArr.push(item);
+  const result: T[] = [];
+  const stack = [array];
+
+  while (stack.length > 0) {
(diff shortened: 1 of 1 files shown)


piaskowyk:
Hey, could you elaborate more on what kind of input causes this function to crash? Could you send me a small reproduction example?

MegaManSec:
Hey, sure:

```
function flattenArray(array) {
  if (!Array.isArray(array)) {
    return [array];
  }
  const resultArr = [];
  const _flattenArray = arr => {
    arr.forEach(item => {
      if (Array.isArray(item)) {
        _flattenArray(item);
      } else {
        resultArr.push(item);
      }
    });
  };
  _flattenArray(array);
  return resultArr;
}

const createDeepNestedArray = (depth) => {
  let arr = 1; // The innermost value
  for (let i = 0; i < depth; i++) {
    arr = [arr];
  }
  return arr;
};

console.log(flattenArray(createDeepNestedArray(100))); // no crash
console.log(flattenArray(createDeepNestedArray(1000))); // no crash
console.log(flattenArray(createDeepNestedArray(10000))); // crash
```

piaskowyk:
I see that this is possible with artificially generated input data, but I'm curious if you've ever experienced this crash in a real-life application use case? 🤔

MegaManSec:
Not really, this was just found during an internal audit while searching whether it would be _possible_ for an attacker to create some type of animated component (e.g. downloaded/retrieved/user-input to create a component) which would result in a crash of the interpreter.

MegaManSec:
Sorry, had to force push two fixes for two slightly incorrect implementation: reversed results, and a missing type for the stack.

PR #7286, opened 2025-03-20
feat: clangd for worklets

tjzel:
## Summary

Now we have two packages so `compile_commands.json` should be hoisted in the respective package directory.

## Test plan

🚀


patrycjakalinska (approved):
lgtm

Issue #6999, opened 2025-02-09
scrollTo performance on the New Architecture is horrible when multiple views are animated at the same time


MatiPl01:
## Problem description

I was looking into possible improvements of auto scroll performance in my [react-native-sortables](https://github.com/MatiPl01/react-native-sortables) library. Performance is fine on the Old Architecture, but on the New Architecture it is horrible. Apart from massive FPS drops and noticeably higher memory usage (might be related to RN, memory is not a big problem), there are also a some other problems:

- `scrollTo` runs synchronously, whilst style updates are asynchronous, which results in the noticeable delay between `ScrollView` position update and the view position update,
- this delay is higher when the number of animated views is increased (probably because of the style updates batching system and conflicts between Reanimated and RN commits)

## Example recording

| Old Architecture (Paper) | New Architecture (Fabric) |
|-|-|
| <video src="https://github.com/user-attachments/assets/a590fb04-0079-4544-aed9-9bf38f4e10e0" /> | <video src="https://github.com/user-attachments/assets/e3a9eaa9-0d3b-4213-a4c0-ff3d4458ac49" /> |

## What I tried

To fix the problem with delay between view update and the `ScrollView` position update I tried to read the view position with the synchronous `measure` function from Reanimated and scroll the `ScrollView` only after the view was transitioned. This approach also doesn't work as, likely, the new view position is applied somewhere between frames in the `useFrameCallback` and there is still some delay between transformation of the view and scrolling the `ScrollView`.

## Final remarks

The issue is less visible in the release build on the real device, likely because of more FPS, which result in more frequent commits and lower delays.

### Steps to reproduce

Copy this code to your app and run it on the Old and the New Architecture:

```tsx
import { useEffect, useMemo } from 'react';
import { StyleSheet, Text, View } from 'react-native';
import Animated, {
  measure,
  scrollTo,
  useAnimatedReaction,
  useAnimatedRef,
  useAnimatedStyle,
  useFrameCallback,
  useScrollViewOffset,
  useSharedValue,
  withRepeat,
  withSequence,
  withTiming
} from 'react-native-reanimated';

const ZERO_VECTOR = { x: 0, y: 0 };

import { flex, text } from '@/theme';

export default function PlaygroundExample() {
  const animatedRef = useAnimatedRef<Animated.ScrollView>();
  const itemRef = useAnimatedRef<Animated.View>();
  const containerRef = useAnimatedRef<Animated.View>();
[...]

PR #7251, opened 2025-03-13
chore: fix static framework workflow

tjzel:

PR #7292, opened 2025-03-24
Fix lost `scrollTo` updates

bartlomiejbloniarz:
## Summary

In RN 0.77 a [batching mechanism for commands was introduced](https://github.com/facebook/react-native/commit/6f1c2a512e44d25edefea53e864f688018745c07) behind a feature flag. It was enabled in RN 0.77, but mistakenly disabled in RN 0.78. Now, in RN 0.79 it is back.

The mechanism makes it so that our `scrollTo` calls are only applied from `RuntimeSecheduler_Modern::updateRendering` on event loop tick. So when there was nothing happening on the JS thread, our updates wouldn't be propagated to the platform (That's why adding a console.log to each frame would fix scrolling in the repro). To fix that we bypass batching, by calling `schedulerDidDispatchCommand` directly on the `SchedulerDelegate`.

This issue is a different thing that the one reported in #6999 by @MatiPl01. It is most probably what @hirbod reported under the same issue.

## Test plan
<details>
<summary>
Code snippet
</summary>

```jsx

import { StyleSheet, Text } from 'react-native';
import Animated, {
  scrollTo,
  useAnimatedRef,
  useAnimatedStyle,
  useFrameCallback,
  useSharedValue,
  withRepeat,
  withSequence,
  withTiming,

const COUNT = 200;

export default function PlaygroundExample() {
  const animatedRef = useAnimatedRef<Animated.ScrollView>();
  const sv = useSharedValue(0);

  useFrameCallback(() => {
    sv.value++;
    scrollTo(animatedRef, 0, sv.value, false);
  });

  return (
    <Animated.ScrollView ref={animatedRef}>
      {Array.from({ length: COUNT }, (_, index) => (
        <Item
          color={`hsl(210, 80%, ${COUNT - index}%)`}
          name={`Item ${index + 1}`}
          key={index}
        />
      ))}
    </Animated.ScrollView>
  );
}

function Item({ color, name }: { name: string; color: string }) {
    return {
      transform: [
        {
          translateX: withRepeat(
            withSequence(withTiming(100), withTiming(0)),
            -1,
            true
          ),
        },
      ],
    };
  });

  return (
    <Animated.View
      style={[styles.item, animatedStyle, { backgroundColor: color }]}>
      <Text style={styles.text}>{name}</Text>
    </Animated.View>
  );
}

  item: {
    alignItems: 'center',
    justifyContent: 'center',
    backgroundColor: 'red',
    height: 100,
  },
  text: {
    fontWeight: 'bold',
  },
});
```

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
@@ -834,10 +834,19 @@ void ReanimatedModuleProxy::dispatchCommand(
   ShadowNode::Shared shadowNode = shadowNodeFromValue(rt, shadowNodeValue);
   std::string commandName = stringFromValue(rt, commandNameValue);
   folly::dynamic args = commandArgsFromValue(rt, argsValue);
-  auto scheduler = static_cast<Scheduler *>(uiManager_->getDelegate());
-  auto shadowView = ShadowView(*shadowNode);
-  scheduler->getDelegate()->schedulerDidDispatchCommand(
-      shadowView, commandName, args);
+  const auto &scheduler = static_cast<Scheduler *>(uiManager_->getDelegate());
+
+  if (!scheduler) {
+    return;
+  }
+
+  const auto &schedulerDelegate = scheduler->getDelegate();
+
+  if (schedulerDelegate) {
+    const auto shadowView = ShadowView(*shadowNode);
+    schedulerDelegate->schedulerDidDispatchCommand(
+        shadowView, commandName, args);
+  }
 }
 
 jsi::String ReanimatedModuleProxy::obtainProp(
--- a/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
@@ -834,7 +834,9 @@ void ReanimatedModuleProxy::dispatchCommand(
   ShadowNode::Shared shadowNode = shadowNodeFromValue(rt, shadowNodeValue);
   std::string commandName = stringFromValue(rt, commandNameValue);
   folly::dynamic args = commandArgsFromValue(rt, argsValue);
-  uiManager_->dispatchCommand(shadowNode, commandName, args);
+  auto scheduler = static_cast<Scheduler*>(uiManager_->getDelegate());
+  auto shadowView = ShadowView(*shadowNode);
+  scheduler->getDelegate()->schedulerDidDispatchCommand(shadowView, commandName, args);
 }
 
 jsi::String ReanimatedModuleProxy::obtainProp(
--- a/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
@@ -834,9 +834,10 @@ void ReanimatedModuleProxy::dispatchCommand(
   ShadowNode::Shared shadowNode = shadowNodeFromValue(rt, shadowNodeValue);
   std::string commandName = stringFromValue(rt, commandNameValue);
   folly::dynamic args = commandArgsFromValue(rt, argsValue);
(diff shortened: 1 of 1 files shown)


packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -834,7 +834,10 @@ void ReanimatedModuleProxy::dispatchCommand(
   ShadowNode::Shared shadowNode = shadowNodeFromValue(rt, shadowNodeValue);
   std::string commandName = stringFromValue(rt, commandNameValue);
   folly::dynamic args = commandArgsFromValue(rt, argsValue);
-  uiManager_->dispatchCommand(shadowNode, commandName, args);
+  auto scheduler = static_cast<Scheduler *>(uiManager_->getDelegate());
+  auto shadowView = ShadowView(*shadowNode);
+  scheduler->getDelegate()->schedulerDidDispatchCommand(
+      shadowView, commandName, args);
```
MatiPl01, line 840: `schedulerDidDispatchCommand` is a little bit confusing here as it seems that this should be called after the command was dispatched, not when we want to dispatch a command, but it works, so it seems to be fine

MatiPl01 (approved):
The provided example works after changes so it LGTM

MatiPl01 (changes requested):
Removing approval as the app crashes frequently on fast refresh (when I press the `R` key while having the simulator focused).

Screenshot 2025-03-24 at 17 29 32

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -834,7 +834,10 @@ void ReanimatedModuleProxy::dispatchCommand(
   ShadowNode::Shared shadowNode = shadowNodeFromValue(rt, shadowNodeValue);
   std::string commandName = stringFromValue(rt, commandNameValue);
   folly::dynamic args = commandArgsFromValue(rt, argsValue);
-  uiManager_->dispatchCommand(shadowNode, commandName, args);
+  auto scheduler = static_cast<Scheduler *>(uiManager_->getDelegate());
+  auto shadowView = ShadowView(*shadowNode);
```
tomekzaw, line 838:
```suggestion
  const auto &scheduler = static_cast<Scheduler *>(uiManager_->getDelegate());
  const auto shadowView = ShadowView(*shadowNode);
```

tomekzaw (approved):

MatiPl01 (approved):
Works fine now

hirbod:
Lovely. Will try it tomorrow. Thanks guys!

hirbod:
Tried to use the 4.0.0-beta.3 and nightly but can't test it. Fails for me with 

The install fails though:

```› Compiling react-native-worklets Pods/RNWorklets » WorkletsModuleProxy.cpp
› Compiling react-native-worklets Pods/RNWorklets » WorkletsModuleProxySpec.cpp

❌  (../../node_modules/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxySpec.cpp:1:10)

> 1 | #include <worklets/NativeModules/WorkletsModuleProxySpec.h>
    |          ^ 'worklets/NativeModules/WorkletsModuleProxySpec.h' file not found
  2 | 
  3 | #include <utility>
  4 | 

› Compiling react-native-worklets Pods/RNWorklets » WorkletsModule.mm
```

FYI, we use dynamic linking, RN 0.77.1. 
Also using both nightly packages did not work.

bartlomiejbloniarz:
@hirbod we backported the fix to 3.17.2

hirbod:
Whoever runs into this: since main and the current v4 nightly releases dropped support for react 19, we had to downgrade the nightlies for reanimated and worklets to `20250327-509deb90b` which cointains the scrollTo fix, but also apply patches from #7251 

If anyone stumbles across this, here are some patchfiles to save you some time (even though I doubt that we have brave enough v4 users in production :D)

[react-native-reanimated@4.0.0-nightly-20250327-509deb90b.patch](https://github.com/user-attachments/files/19603228/react-native-reanimated%404.0.0-nightly-20250327-509deb90b.patch)
[react-native-worklets@0.2.0-nightly-20250327-509deb90b.patch](https://github.com/user-attachments/files/19603229/react-native-worklets%400.2.0-nightly-20250327-509deb90b.patch)

PR #7293, opened 2025-03-24
refactor: install worklet caches from cpp

tjzel:
## Summary

Injecting `__handleCache`, `__workletsCache`, and `__shareableMappingCache` through JSI on all Runtimes. This is an important step for a symmetrical worklet API - to have the same caching mechanism on all runtimes. `__handleCache` and `__workletsCache` are now unused on the RNRuntime, but I will install the ValueUnpacker on the RNRuntime soon. Similar comment for `__shareableMappingCache` on the Worklet Runtimes.

## Test plan

Tested it on FabricExample and WebExample and everything worked fine 👍

### DIFF
--- a/packages/react-native-reanimated/src/layoutReanimation/animationsManager.ts
+++ b/packages/react-native-reanimated/src/layoutReanimation/animationsManager.ts
@@ -1,7 +1,8 @@
 'use strict';
-import { runOnUI } from 'react-native-worklets';
+import { executeOnUIRuntimeSync } from 'react-native-worklets';
 
 import { withStyleAnimation } from '../animation';
+import { SHOULD_BE_USE_WEB } from '../common';
 import type {
   LayoutAnimation,
   LayoutAnimationStartFunction,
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -99,5 +99,6 @@ module.exports = {
     'jest/prefer-to-have-length': 'warn',
     'jest/valid-expect': 'error',
     'react/react-in-jsx-scope': 'off',
+    'lines-between-class-members': 'off',
```
tjzel, line 102: I don't actually need this in this PR, but this rule is annoying. It turns:

```javascript
class Foo(){
  bar(){}
  baz(){}
  bag(){
    return 1;
  }
	foobar(){}
}
```

into

```javascript
```javascript
class Foo(){
  bar(){}
  baz(){}
  
  bag(){
    return 1;
  }
  
	foobar(){}
```
Which is so inconsitent.

packages/react-native-worklets/src/threads.ts
```diff
@@ -237,10 +237,7 @@ export function runOnJS<Args extends unknown[], ReturnValue>(
       fun as
         | ((...args: Args) => ReturnValue)
         | WorkletFunction<Args, ReturnValue>,
-      args.length > 0
-        ? // TODO TYPESCRIPT this cast is terrible but will be fixed
```
tjzel, line 241: Told you!

packages/react-native-worklets/src/valueUnpacker.ts
```diff
@@ -10,34 +10,29 @@ function valueUnpacker(
   remoteFunctionName?: string
 ): unknown {
   'worklet';
-  let workletsCache = global.__workletsCache;
-  let handleCache = global.__handleCache;
-  if (workletsCache === undefined) {
-    // init
-    workletsCache = global.__workletsCache = new Map();
-    handleCache = global.__handleCache = new WeakMap();
-  }
```
tjzel, line 19: Just think about the performance gains...

PR #7296, opened 2025-03-24
fix: CSS issues after RN upgrade

MatiPl01:


MatiPl01:
Closing this attempt to fix CSS issues introduced by RN 0.78 upgrade as it turned out that this approach doesn't work properly.

PR #7303, opened 2025-03-25
fix: Misspelt height prop name in supported layout animation props

MatiPl01:
## Summary

Quick fix for incorrect `height` prop spelling which resulted in this warning being shown in the console:

```
[Worklets] 'height' property is not officially supported for layout animations. It may not work as expected.
```


patrycjakalinska (approved):
lgtm

tomekzaw (approved):

PR #7304, opened 2025-03-25
feat(Worklets): Synchronizable

tjzel:
## Summary

Implementing `Synchronizable` per its definition in the [design document](https://github.com/software-mansion/react-native-reanimated/discussions/7264)

WIP

## Test plan

TBA


apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/memory/synchronizable.test.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+describe('Test Synchronizable creation and serialization', () => {
+  test('createSynchronizable retuns Synchronizable', () => {
+    const synchronizable = createSynchronizable(0);
+
+    expect(isSynchronizable(synchronizable)).toBe(true);
+  });
+
+  test('Synchronizable serializes correctly from RN Runtime to UI Runtime', () => {
+    const synchronizable = createSynchronizable(0);
```
MatiPl01, line 23: Isn't 0 a default value for numbers in cpp? Won't it be safer to use something else than 0 then to make sure that the proper value was read from CPP (the same as the initial value passed during construction)?

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/memory/synchronizable.test.tsx
```diff
@@ -0,0 +1,320 @@
...
+      'worklet';
+      runOnJS(onJSCallback)(synchronizable);
+    })();
+    await waitForNotify(NOTIFICATION);
+
+    const value = synchronizableCopy!.getBlocking();
+    expect(value).toBe(42);
+  });
+
+  // This fail doesn't work beacuse nested `runOnJS` isn't copied properly.
```
MatiPl01: Will you fix this in the future? If yes, can you leave a TODO comment near this code? If not, can you maybe remove this test case?

packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
```diff
@@ -543,6 +559,97 @@ jsi::Value JSIWorkletsModuleProxy::get(
...
+  if (name == "createSynchronizable") {
+    return jsi::Function::createFromHostFunction(
+        rt,
+        propName,
+        1,
+        [](jsi::Runtime &rt,
+           const jsi::Value &thisValue,
+           const jsi::Value *args,
+           size_t count) {
+          auto initial = extractSerializableOrThrow<Serializable>(
```
MatiPl01:
```suggestion
          auto initial = extractSerializableOrThrow(
```
Is this explicit type needed? Why can you use the `extractSynchronizableOrThrow` function without any type below (line 616)?

packages/react-native-worklets/Common/cpp/worklets/SharedItems/SynchronizableAccess.cpp
```diff
@@ -0,0 +1,81 @@
+#include <worklets/SharedItems/SynchronizableAccess.h>
+
+#include <mutex>
+
+namespace worklets {
+void SynchronizableAccess::getBlockingBefore() {
+  std::unique_lock<std::mutex> lock(accessLock_);
+  queue_.wait(lock, [this]() {
+    return !blockingWriter_ /* && dirtyWriters_ == 0 */ &&
```
MatiPl01, line 9: should this be left commented out?

packages/react-native-worklets/src/WorkletsModule/JSWorklets.ts
```diff
@@ -19,55 +19,55 @@ const requestAnimationFrameImpl =
     : globalThis.requestAnimationFrame;
 
 class JSWorklets implements IWorkletsModule {
-  createSerializable<TValue>(): SerializableRef<TValue> {
+  createSerializable(): never {
```
MatiPl01, line 22: Nice

packages/react-native-worklets/src/shareables.ts
```diff
@@ -202,6 +204,9 @@ function createSerializableNative<T>(
   if ((isPlainJSObject(value) || isFunction) && isWorkletFunction(value)) {
     return cloneWorklet(value, shouldPersistRemote, depth);
   }
+  if (isSynchronizable(value)) {
+    return cloneSynchronizable(value) as SerializableRef<T>;
```
MatiPl01: Is that correct? Shouldn't it be `SynchronizableRef`?

packages/react-native-worklets/src/synchronizable.ts
```diff
@@ -0,0 +1,33 @@
...
+  const synchronizableRef = WorkletsModule.createSynchronizable(
+    createSerializable(initialValue)
+  );
+
+  return globalThis.__synchronizableUnpacker(
+    synchronizableRef
+  ) as unknown as Synchronizable<TValue>;
+}
+
+export interface SynchronizableRef<TValue = unknown> {
```
MatiPl01: Can use `type` instead of `interface`? Mixing the 2 is very cumbersome and `type` is more convenient to use in TS than `interface`. I can also see that the `type` specifier was used to create the `SerializableRef` type, so please keep it consistent:

```ts
export type SerializableRef<T = unknown> = {
  __serializableRef: true;
  __nativeStateSerializableJSRef: T;
};
```

packages/react-native-worklets/src/synchronizable.ts
```diff
@@ -0,0 +1,33 @@
...
+  __synchronizableRef: true;
+  __nativeStateSynchronizableJSRef: TValue;
+}
+
+export interface Synchronizable<TValue = unknown>
```
MatiPl01: same

apps/common-app/src/apps/reanimated/examples/SynchronizableExample.tsx
```diff
@@ -0,0 +1,233 @@
...
+  runOnUI,
+  createWorkletRuntime,
+  runOnRuntime,
+} from 'react-native-worklets';
+
+const initialValue = 0;
+
+const targetValue = 200000;
+
+export default function SynchronizableExample() {
```
MatiPl01: Can you include a brief explanation of how should we use this example to test the synchronizable? I would prefer to have more examples with description for easier before-release testing.

MatiPl01 (changes requested):
Getting this crash when I enter the `Synchronizable` example screen. I cleared all the project files, rebuilt project, etc. and nothing works, so seems like an issue with this implementation (or I am still doing sth wrong)

https://github.com/user-attachments/assets/e17f4fe8-c8c4-46fd-a49a-84a858918242

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/memory/synchronizable.test.tsx
tjzel: It will be fixed when Bundle Mode™ becomes the standard.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/memory/synchronizable.test.tsx
tjzel: Synchronizable can hold any JS value, not only numbers, so this is a valid test case. The value was read because it otherwise wouldn't be a number.

packages/react-native-worklets/Common/cpp/worklets/SharedItems/SynchronizableAccess.cpp
tjzel: Yes. This is commented out because `setDirty` doesn't apply yet but I didn't want to drop the implementation because it took some sweat to make it.

packages/react-native-worklets/src/shareables.ts
tjzel: It's correct. `SynchronizableRef` is already a `SerializableRef` (the type is made so) but `createSerializable` expects to return exactly `SerializableRef`, not the specialization type.

apps/common-app/src/apps/reanimated/examples/SynchronizableExample.tsx
tjzel: This example isn't for testing, it's a showcase of performance of the Synchronizable. For testing I implemented runtime tests. What do you think I should do in this case?

packages/react-native-worklets/src/synchronizable.ts
tjzel: There is (or was) a slight difference between type and interface, but it doesn't apply here. I'll change it.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/memory/synchronizable.test.tsx
MatiPl01: Ok, makes sense to me

packages/react-native-worklets/Common/cpp/worklets/SharedItems/SynchronizableAccess.cpp
MatiPl01: Ok, that's fine

packages/react-native-worklets/src/synchronizable.ts
MatiPl01: In most cases, the difference is in favor of types, which are more flexible (and seem to be more TS-like, whilst interfaces are just meant to resemble how Java's and other languages interfaces work).

apps/common-app/src/apps/reanimated/examples/SynchronizableExample.tsx
MatiPl01: In such a case, I am fine with that. Maybe just leave a short text like "Synchronizable performance testing example" or similar to make it clear that it is just for performance testing?

tjzel: I removed the `Test` button because it was redundant. I will also change the name to mention performance.

tjzel:
@MatiPl01 I merged the main so it no longer crashes.

packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
tjzel: Fixed.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/memory/synchronizable.test.tsx
tjzel: Added TODO

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.cpp
```diff
@@ -129,6 +129,10 @@ void WorkletRuntime::init(
   auto valueUnpackerBuffer =
       std::make_shared<const jsi::StringBuffer>(ValueUnpackerCode);
   rt.evaluateJavaScript(valueUnpackerBuffer, "valueUnpacker");
+
+  auto synchronizableUnpackerBuffer =
+      std::make_shared<const jsi::StringBuffer>(SynchronizableUnpackerCode);
+  rt.evaluateJavaScript(synchronizableUnpackerBuffer, "synchronizableUnpacker");
```
piaskowyk, line 135: Do you need it since you already install unpacker in `initializeRuntime` method?

packages/react-native-worklets/src/serializable.ts
```diff
@@ -202,6 +204,9 @@ function createSerializableNative<T>(
   if ((isPlainJSObject(value) || isFunction) && isWorkletFunction(value)) {
     return cloneWorklet(value, shouldPersistRemote, depth);
   }
+  if (isSynchronizable(value)) {
+    return cloneSynchronizable(value) as SerializableRef<T>;
+  }
```
piaskowyk, line 209: When can this happen? I'm wondering if it's safe to use a `synchronizable` that was created on a different runtime.

piaskowyk (approved):
Smart 😎

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.cpp
tjzel: This is not Bundle Mode so the unpacker has to be installed from C++

packages/react-native-worklets/src/serializable.ts
tjzel: This is `makeSerializable` so it's when we want to send the Synchronizable to other runtimes. Synchronizables don't need additional serialization (because they came serialized from C++) so the only thing we need to do is properly store it in our cache for quicker lookups - see line 595.

PR #7307, opened 2025-03-26
chore(plugin): enable nested worklets by default

tjzel:
## Summary

We've been dogfooding "nested worklets" for quite a while and we haven't had any issues stemming from them since. The feature is supposed to be more utilized when the new bundling method drops.

## Test plan

Our apps works just as they worked before.


packages/docs-reanimated/docs/reanimated-babel-plugin/options.md
```diff
@@ -95,34 +93,6 @@ function MyView({ taggedWidth }) {
...
-This example will result in an error. Let's quickly describe why:
-
-1. Upon creating the functions and resolving their worklet factories, the `runOnUI` function is called. This function first takes the worklet's data, loads it into the UI thread after converting it, and then schedules an execution asynchronously.
-
-2. During execution, the worklet scheduled for execution calls `runOnSomeOtherThread`. This action mirrors what `runOnUI` does, but targets SomeOtherThread.
-
-3. This process fails because the injection of `outerWorklet` into the UI thread occurred without the `innerWorklet` worklet data, therefore it's not available for SomeOtherThread.
-
-If you enable this option, the system will workletize functions depth-first, avoiding the above-mentioned scenario and ensuring things operate correctly. Keep in mind that nesting worklets like in the provided example is only useful in threading.
-
```
tomekzaw, line 125: Would be good to move that somewhere else instead of removing it.

tomekzaw (approved):

packages/docs-reanimated/docs/reanimated-babel-plugin/options.md
tjzel: It's still in the versioned docs for Reanimated 3

PR #7308, opened 2025-03-26
chore: rename plugin workspace

tjzel:
## Summary

Forgot to do it in #7066

## Test plan

CI


MatiPl01 (changes requested):
Please fix CI

MatiPl01 (approved):
CIs seems to pass now, so it should be good to merge

PR #7318, opened 2025-03-27
Performance Monitor: Guard Frame Rate

mobinni:
## Summary

I noticed that when fast-scrolling sometimes the UI frame rate would over-shoot significantly up to 200 FPS. This can happen when `const measuredRangeDuration = timestamp - droppedTimestamp;` drops below 100ms.  On a 120HZ screen frames are 8ms ~

For example if it calculated frames 100ms apart would lead to 100/20 (smoothingFrames) = 1000 / 5 => 200 FPS
The maximum should be 120 FPS / 60 FPS.

I wasn't sure if we should guard or remove the division by smoothingFrames.

### DIFF
--- a/packages/react-native-reanimated/src/component/PerformanceMonitor.tsx
+++ b/packages/react-native-reanimated/src/component/PerformanceMonitor.tsx
@@ -72,7 +72,7 @@ function getFps(renderTimeInMs: number): number {
 
 function completeBufferRoutine(
   buffer: CircularBuffer,
-  timestamp: number
+  timestamp: number,
 ): number {
   'worklet';
   timestamp = Math.round(timestamp);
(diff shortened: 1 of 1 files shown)


mobinni:
@piaskowyk Actually I wonder if we should be dividing by the amount of frames supported by the device.
120 HZ = 8 frames per second? Although this is actually determined by the display itself so sometimes it could run at 16. Maybe `smoothingFrames` should be 16 and not 20? 

```
const measuredRangeDuration = timestamp - droppedTimestamp;
const avgFrameTime = measuredRangeDuration / 16;
```

If your `measuredRangeDuration` is for example 100 MS => 100 / 16 = 6.2
1000 / 12.5 = 80 FPS on a 120hz display 🤔

Issue #7282, opened 2025-03-20
When the type of SymbolView is set to palette, the colors property cannot Updated


liixing:
### Description

I want to display a multi-color icon with animation, but I found that when the type of SymbolView is set to palette, the colors array cannot be updated dynamically.  tintColor can be animated but colors can't

I’ve tried using createAnimatedPropAdapter and addWhitelistedNativeProps, but neither worked

https://github.com/user-attachments/assets/05b987ce-4f1b-4de7-877f-11bd1cb5385b

### Steps to reproduce

1.  create AnimatedSymbolView
2. use  animatedProps for  colors
3.  tintColor can be animated but colors can't

### Snack or a link to a repository

https://github.com/liixing/animate-palette-icon

### Reanimated version

3.16.1

### React Native version

0.76.7

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

None

### Acknowledgements

Yes

liixing:
update to 3.17.1 also have the same issue

patrycjakalinska:
hi @liixing, thanks for opening this issue. I think the issue is related to how we process colors - when some prop is updated, we check if it's in our `ColorProperties` list and if so we need to process the color-related property for Native side to understand. In this case `tintColor` is declared in this list, but `colors` aren't.
I'll open a PR adjusting this list so everything should work fine c:

liixing:
@patrycjakalinska thanks for looking into this! That makes a lot of sense.

PR #7321, opened 2025-03-28
feat: Add processing for array color related props

patrycjakalinska:
## Summary

I've added processing of array and color-related props so every item would be processed.

## Test plan

Use code provided with an issue #7282


tomekzaw (commented):
The PR looks good to me, but `colors` sounds like a custom property so perhaps the developer should call `processColor` from react-native-reanimated manually.

tomekzaw (approved):

PR #7336, opened 2025-04-01
fix: Numeric colors processing

MatiPl01:
## Summary

This PR fixes 2 issues:

1. Conversion of the numeric color to hexadecimal string was invalid (we converted a decimal value to string directly without taking the hex radix into account). This was also a workaround for a real issue with numeric colors preprocessing described below.

2. Numeric colors processing was broken because of missing rgba to argb format conversion in the `processColorInitially`. There was an if statement which returned the number right away without applying the bit shift, which resulted in invalid colors in animations.

## Test plan

1. Copy the code snippet from below
2. Observe that the app crashes with the previous implementation
3. Switch to this branch and test the current implementation (the app doesn't crash)

<details>

```tsx
import React from 'react';
import {ColorValue, View} from 'react-native';

function App() {
  return (
    <View style={{flex: 1, padding: 20}}>
      <View style={{flexDirection: 'row', gap: 10, flexWrap: 'wrap'}}>
        {Object.keys(Colors).map(key => (
          <Animated.View
            key={`view-with-color-${key}`}
            style={{
              width: 50,
              height: 50,
              backgroundColor: Colors[key],
            }}
          />
        ))}
      </View>
    </View>
  );
}
export default App;

const Colors = {
  Magenta: 0xff00ffff,
  Red: 0xff0000ff,
  Green: 0xff00ff,
  Blue: 0xffff,
  Yellow: 0xffff00ff,
  Cyan: 0xffffff,
  LightGray: 0xccccccff,
} as unknown as Record<string, ColorValue>;
```
</details>

4. You can also test the code with the following CSS animation or the Shared Value based animation (both didn't work well with numbers)

```tsx
// CSS animation

animationDuration: 1000,
animationIterationCount: Infinity,
animationName: {
  from: {
    backgroundColor: 0xff00ffff,
  },
  to: {
    backgroundColor: 0xffff,
  },
},

// Shared Value animation
const sv = useSharedValue(0);

useEffect(() => {
  sv.value = withRepeat(
    withSequence(
      withTiming(0, { duration: 1000 }),
      withTiming(1, { duration: 1000 })
    ),
    -1
  );
}, []);

const astyle = useAnimatedStyle(() => ({
  backgroundColor: interpolateColor(sv.value, [0, 1], [0xff00ffff, 0xffff]),
}));
```

## Examples

In all example below the `backgroundColor` should animate between magenta and blue.

### Before applying changes to the `packages/react-native-reanimated/src/css/platform/native/style/processors/colors.ts`
(invalid conversion of number to hex string)

### Before fixing conversion of rgba to argb in `processColorInitially`

https://github.com/user-attachments/assets/32b0f2ad-4212-460f-9226-8caf357d4aa6

### After both fixes

https://github.com/user-attachments/assets/0fc6fc1d-f85a-4f9d-988c-96db89e11e9e


packages/react-native-reanimated/src/Colors.ts
```diff
@@ -632,22 +632,28 @@ export const hsvToColor = (
...
+    if (typeof normalizedColor !== 'number') {
+      return null;
+    }
+
+    colorNumber = normalizedColor;
   }
 
-  normalizedColor = ((normalizedColor << 24) | (normalizedColor >>> 8)) >>> 0; // alpha rgb
-  return normalizedColor;
+  return ((colorNumber << 24) | (colorNumber >>> 8)) >>> 0; // alpha rgb
```
MatiPl01, line 656: Bit shift must be applied to numbers as well

piaskowyk (commented):
Looks good, but did you test it with the alpha channel on both iOS and Android? I'm not sure, but there might be a different order of channels on each platform

MatiPl01:
> Looks good, but did you test it with the alpha channel on both iOS and Android? I'm not sure, but there might be a different order of channels on each platform

I think that this change is not platform specific. We had been applying this bit shift before as well after the initial processing of color (conversion from string to number).

I have just checked it again to be sure that everything works by using the following code snippet and passing different colors with alpha channel (last 2 digits) and everything worked the same on iOS and Android.

```ts
const av = useAnimatedStyle(() => ({
    sv.value,
    [0, 1],
    [0x00000000, 0xff000011]
  ),
}));
```

piaskowyk (approved):

PR #7338, opened 2025-04-02
fix: boxShadow sometimes not working for web

patrycjakalinska:
## Summary

This PR fixes boxShadow sometimes not working on web. To work boxShadow in the form of string must have `offsetX`, `offsetY`, `blurRadius` and `spreadDistance`. When one of these props was missing the boxShadow wasn't being processed ergo was staying as string - that's why it sometimes **worked** on web. This PR checks if we are on web and if no we then can process the boxShadow.

Additionally I've replaced `shadow` with `boxShadow` in ColorExample.tsx

## Test plan

Check ColorExample on web

### DIFF
--- a/apps/common-app/src/apps/reanimated/examples/ColorExample.tsx
+++ b/apps/common-app/src/apps/reanimated/examples/ColorExample.tsx
@@ -31,9 +31,9 @@ export default function ColorExample() {
   });
 
   const style4 = useAnimatedStyle(() => {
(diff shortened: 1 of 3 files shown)


MatiPl01 (approved):
Good for now.

As we talked internally, we may think of unifying the `processBoxShadow` implementation with the one for CSS in the future - it will be easier to maintain just one instead of two.

PR #7348, opened 2025-04-04
Fix conditional and BB examples on web

skusnierz:
## Summary

- Fix conditional example on web
https://github.com/user-attachments/assets/ecde6138-c7f4-4400-95bf-e5fb6256f605

- Fix BB example on web
https://github.com/user-attachments/assets/95927862-b303-4db3-ba07-cf0bf1e6d9fc
- Bump version of few dependencies to resolve warning : 
```
The following packages should be updated for best compatibility with the installed expo version:
  expo@52.0.4 - expected version: ~52.0.42
  react-dom@19.0.0 - expected version: 18.3.1
  react-native@0.78.0 - expected version: 0.76.9
  @expo/metro-runtime@4.0.0 - expected version: ~4.0.1
Your project may not work correctly until you install the expected versions of the packages.
```

### DIFF
--- a/apps/web-example/package.json
+++ b/apps/web-example/package.json
@@ -12,13 +12,13 @@
   },
   "dependencies": {
     "common-app": "workspace:*",
-    "expo": "^52.0.42",
+    "expo": "^52.0.4",
     "react-dom": "19.0.0",
     "react-native": "0.78.0",
     "react-native-web": "0.19.13"
   },
   "devDependencies": {
-    "@expo/metro-runtime": "~4.0.1",
+    "@expo/metro-runtime": "~4.0.0",
     "@stylexjs/babel-plugin": "^0.10.0",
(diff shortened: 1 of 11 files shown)


apps/web-example/package.json
```diff
@@ -12,13 +12,13 @@
   },
   "dependencies": {
     "common-app": "workspace:*",
-    "expo": "^52.0.4",
+    "expo": "^52.0.42",
```
MatiPl01, line 15: I think that it would be better to bump dependencies separately. We also don't have to care about expo warnings too much as we tend to use latest versions of libraries, which were released after the release of the Expo SDK that we use so warning can be expected here.

skusnierz: ok, done

MatiPl01 (approved):
LGTM

PR #7360, opened 2025-04-07
Remove volume example

skusnierz:
## Summary
Remove volume example 🔥


tomekzaw (approved):
cc @kacperkapusciak for the context

MatiPl01 (approved):
Nice, finally one less problematic dependency

PR #7363, opened 2025-04-08
Handle new TurboModule representation

skusnierz:
## Summary
- TurboModule has new representation : 
`The strategy we found to be most efficient and flexible longer-term is to represent the TurboModule with a plain JavaScript object and use a HostObject as its prototype.`
more info -> https://github.com/facebook/react-native/commit/20dba39dab4ef85eb28659a89b19750cec3193a4.
This PR adds proper support for the new representation.
- Added additional message `Open debugger to view warnings`

image

## Test plan
To test this scenario you can use `Freezing shareables` example, before my changes `Modified converted host object` throws error : 
https://github.com/user-attachments/assets/b5be800c-2d94-41d9-90fb-1cab952c0029
```
function tryModifyConvertedHostObject() {
  const obj = TurboModuleRegistry.get('Clipboard');
  if (!obj) {
    console.warn('No host object found.');
    return;
  }
  makeShareableCloneRecursive(obj);
  // @ts-expect-error It's ok
  obj.prop = 2; // shouldn't warn because it's not frozen
}

```


MatiPl01 (approved):
Looks good. Please also add a text component with a short explanation that 🤫 emoji is fine and shouldn't warn.

apps/common-app/src/apps/reanimated/examples/ShareableFreezingExample.tsx
```diff
@@ -11,6 +11,17 @@ import { makeShareableCloneRecursive } from 'react-native-reanimated';
 export default function FreezingShareables() {
   return (
     <View style={styles.container}>
+      <View style={styles.warning}>
+        <Text>Open debugger to view warnings.</Text>
+        <View style={{ flexDirection: 'row', alignItems: 'center' }}>
+          <Text style={styles.text}>⚠️</Text>
+          <Text>Should show warning</Text>
```
tjzel, line 18:
```suggestion
          <Text>Should show a warning.</Text>
```

apps/common-app/src/apps/reanimated/examples/ShareableFreezingExample.tsx
```diff
@@ -11,6 +11,17 @@ import { makeShareableCloneRecursive } from 'react-native-reanimated';
...
+        </View>
+        <View style={{ flexDirection: 'row', alignItems: 'center' }}>
+          <Text style={styles.text}>🤫</Text>
+          <Text>Should not show warning</Text>
```
tjzel, line 22:
```suggestion
          <Text>Should not show a warning.</Text>
```

packages/react-native-worklets/src/shareables.ts
```diff
@@ -139,6 +139,10 @@ function makeShareableCloneRecursiveNative<T>(
     return cloneRemoteFunction(value, shouldPersistRemote);
   }
   if (isHostObject(value)) {
+    // https://github.com/facebook/react-native/commit/20dba39dab4ef85eb28659a89b19750cec3193a4
+    if (isHostObject(Object.getPrototypeOf(value))) {
+      return cloneHostObject(Object.getPrototypeOf(value), shouldPersistRemote);
+    }
```
tjzel, line 145: We need to think about how to handle this situation in cpp and on other runtimes.

In this case a user can do:
```js
const hostObject = createHostObject(); // some host obj factory, returns js object with host object as prototype
hostObject.foo(); // host function invocation
hostObject.greeting = "hello"; // legal operation on plain js object

runOnUI(() => {
  console.log(hostObject.greeting); // undefined, should be "hello" for consistency
}();
```

IMO we shoul copy all the OwnProperties of the plain object, copy the host object prototype, freeze the plain object, and if we need to extract it to another runtime, we should also extract the extra properties and extract it as a plain object with host object prototype.

tjzel (changes requested):

packages/react-native-worklets/src/shareables.ts
```diff
@@ -139,6 +139,18 @@ function makeShareableCloneRecursiveNative<T>(
     return cloneRemoteFunction(value, shouldPersistRemote);
   }
   if (isHostObject(value)) {
+    // https://github.com/facebook/react-native/commit/20dba39dab4ef85eb28659a89b19750cec3193a4
+    if (isHostObject(Object.getPrototypeOf(value))) {
+      const plainJSObject = {
+        ...Object.getPrototypeOf(value),
+      };
```
tomekzaw, line 146: Let's remove the comma so it fits in one line.
```suggestion
      const plainJSObject = { ...Object.getPrototypeOf(value) };
```
Alternatively, we can use `Object.assign({}, value)` directly to emphasize the fact that we're making a plain copy.

packages/react-native-worklets/src/shareables.ts
```diff
@@ -139,6 +139,18 @@ function makeShareableCloneRecursiveNative<T>(
...
+      Object.getOwnPropertyNames(value).forEach((key) => {
+        plainJSObject[key] = value[key as keyof typeof value];
+      });
+
+      return clonePlainJSObject(plainJSObject, shouldPersistRemote, depth);
+      // return cloneHostObject(Object.getPrototypeOf(value), shouldPersistRemote);
```
tomekzaw, line 152: Let's remove irrelevant code inside a comment
```suggestion
```

packages/react-native-worklets/src/shareables.ts
```diff
@@ -355,6 +372,26 @@ function clonePlainJSObject<T extends object>(
   return clone;
 }
 
+function cloneTurboModule<T extends object>(
```
tjzel, line 375: Please also add a comment (TSDoc) what kind of structure `TurboModuleLike` is
```suggestion
function cloneTurboModuleLike<T extends object>(
```

packages/react-native-worklets/src/shareables.ts
```diff
@@ -355,6 +372,26 @@ function clonePlainJSObject<T extends object>(
...
+  value: T,
+  shouldPersistRemote: boolean,
+  depth: number
+): ShareableRef<T> {
+  const clonedProto = Object.getPrototypeOf(value);
```
tjzel, line 380: Please also add a comment why we **don't** call `makeSharableClone` on `proto`.
```suggestion
  const proto = Object.getPrototypeOf(value);
```

packages/react-native-worklets/src/shareables.ts
```diff
@@ -139,6 +139,10 @@ function makeShareableCloneRecursiveNative<T>(
     return cloneRemoteFunction(value, shouldPersistRemote);
   }
   if (isHostObject(value)) {
+    // https://github.com/facebook/react-native/blob/main/packages/react-native/ReactCommon/react/nativemodule/core/ReactCommon/TurboModuleBinding.cpp#L182
```
tjzel, line 142: Please add a summary sentence, not just a link.

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.h
```diff
@@ -204,6 +205,20 @@ class ShareableHostObject : public Shareable {
...
+ public:
+  ShareableTurboModule(
+      jsi::Runtime &rt,
+      const jsi::Object &properties,
+      const std::shared_ptr<jsi::HostObject> &__proto__);
+
+  jsi::Value toJSValue(jsi::Runtime &rt) override;
+
+ protected:
+  std::shared_ptr<ShareableHostObject> __proto__;
```
tjzel, line 218: No need for these underscores.

```suggestion
  std::shared_ptr<ShareableHostObject> proto_;
```

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.h
```diff
@@ -204,6 +205,20 @@ class ShareableHostObject : public Shareable {
   const std::shared_ptr<jsi::HostObject> hostObject_;
 };
 
+class ShareableTurboModule : public Shareable {
```
tjzel, line 208:
```suggestion
class ShareableTurboModuleLike : public Shareable {
```

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -81,7 +82,16 @@ jsi::Value makeShareableClone(
...
+      auto getPrototypeOf = rt.global()
+                                .getProperty(rt, "Object")
+                                .asObject(rt)
+                                .getPropertyAsFunction(rt, "getPrototypeOf");
+      auto objectPrototype = getPrototypeOf.call(rt, object).asObject(rt);
+
+      if (objectPrototype.isHostObject(rt)) {
+        shareable = std::make_shared<ShareableTurboModule>(
+            rt, object, objectPrototype.getHostObject(rt));
+      } else if (shouldRetainRemote.isBool() && shouldRetainRemote.getBool()) {
```
tjzel, line 94: I think, as a safety backup, we should try to check if the prototype is a `JSShareableRef` and handle this case somehow - either fail here or try to extract the underlying host object.

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -237,6 +247,34 @@ jsi::Value ShareableHostObject::toJSValue(jsi::Runtime &rt) {
   return jsi::Object::createFromHostObject(rt, hostObject_);
 }
 
+ShareableTurboModule::ShareableTurboModule(
+    jsi::Runtime &rt,
+    const jsi::Object &properties,
+    const std::shared_ptr<jsi::HostObject> &proto)
+    : Shareable(TurboModuleObjectType) {
+  // set properties prototype to empty object, we store prototype in __proto__
```
tjzel, line 255:
```suggestion
  // We must get rid of the Host Object prototype as `ShareableObject` expects the prototype to be that of a plain object.
```

apps/common-app/src/apps/reanimated/examples/ShareableFreezingExample.tsx
```diff
@@ -91,14 +96,18 @@ function tryModifyConvertedRemoteFunction() {
   obj.prop = 2; // should warn because it's frozen
 }
 
-function tryModifyConvertedHostObject() {
-  const obj = TurboModuleRegistry.get('Clipboard');
```
tjzel, line 95: Informatively:
Clipboard Turbo Module is tied somehow to the RN Runtime, and copying it to another runtime is fatal, that's why we switched to `__reanimatedModuleProxy`.

tjzel (changes requested):

apps/common-app/src/apps/reanimated/examples/ShareableFreezingExample.tsx
```diff
@@ -26,17 +31,17 @@ export default function FreezingShareables() {
...
+          onPress={tryModifyConvertedPlainObject}
         />
       </View>
       <View style={styles.textAndButton}>
-        <Text style={styles.text}>⚠️</Text>
+        <Text style={styles.text}>🤫</Text>
         <Button
-          title="Modify converted plain object"
-          onPress={tryModifyConvertedPlainObject}
+          title="Modify converted turbo module"
```
tjzel, line 43:
```suggestion
          title="Modify converted turbo module like"
```

apps/common-app/src/apps/reanimated/examples/ShareableFreezingExample.tsx
```diff
@@ -26,17 +31,17 @@ export default function FreezingShareables() {
         />
       </View>
       <View style={styles.textAndButton}>
-        <Text style={styles.text}>🤫</Text>
+        <Text style={styles.text}>⚠️</Text>
         <Button
-          title="Modify converted host object"
-          onPress={tryModifyConvertedHostObject}
+          title="Modify converted plain object"
+          onPress={tryModifyConvertedPlainObject}
```
tjzel, line 37: Let's also have an example for a Host Object (not a TurboModule-like)

tjzel (changes requested):
Looks good overall, I just want you to address some minor issues.

tjzel (commented):
Let's hold with merging this PR until we clean up in `makeShareableClone`, since then the hacky logic here could be greatly simplified.

Issue #5365, opened 2023-11-10
Variables defined after worklet functions are mistakingly undefined in worklets


gaearon:
### Description

This works:

```js
  const [x, setX] = React.useState(0)

  function someWorkletFn() {
    'worklet'
    console.log('lol')
  }

  const handler = useAnimatedScrollHandler({
    onScroll() {
      console.log(x)
      someWorkletFn()
    },
  })
```

This doesn't:

```js
  const handler = useAnimatedScrollHandler({
    onScroll() {
      console.log(x)
      someWorkletFn()
    },
  })

  const [x, setX] = React.useState(0)

  function someWorkletFn() {
    'worklet'
    console.log('lol')
  }
```

However, from the perspective of the code inside `onScroll`, this code should be equivalent — the variables have already been declared before it was called.

If the Reanimated plugin can't handle it, I think this should be a build error or something. Right now they're just `undefined`.

### Steps to reproduce

```js
import * as React from 'react'
import {View} from 'react-native'
import Animated, {useAnimatedScrollHandler} from 'react-native-reanimated'

export default function App() {
  const handler = useAnimatedScrollHandler({
    onScroll() {
      console.log(x)
      someWorkletFn()
    },
  })

  const [x, setX] = React.useState(0)

  function someWorkletFn() {
    'worklet'
    console.log('lol')
  }

  React.useEffect(() => {
    let id = setInterval(() => setX(x => x + 1), 1000)
    return () => clearInterval(id)
  }, [])

  return (
    <Animated.FlatList
      onScroll={handler}
      style={{flex: 1, backgroundColor: 'red'}}
      contentContainerStyle={{
        paddingTop: 1000,
      }}
    />
  )
}
```

Expected: it works

Actual:

```
LOG  undefined
ERROR  ReanimatedError: someWorkletFn is not a function (it is undefined), js engine: reanimated
```

This is definitely broken on Android and iOS. Not sure about web cause I can't get `onScroll` there to fire at all in this example.

### Snack or a link to a repository

https://github.com/bluesky-social/social-app/tree/reanimated-order-bug

### Reanimated version

3.4.2

### React Native version

0.72.5

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Acknowledgements

Yes

gaearon:
I did :)

Latropos:
[...]

PR #7365, opened 2025-04-08
refactor: use arguments instead of closures in Worklets plugin

tjzel:
## Summary

Currently Worklets Babel plugin uses implicit closure capturing to retain identifiers in the original worklet:

```js
// input
const mydlo = 'smiglo';

function foo() {
  'worklet';
  console.log(mydlo);
}
```

```js
// output
const mydlo = 'smiglo';
const _worklet_5876054941198_init_data = {
  // init data
};
const foo = (function () {
  const foo = function () {
    console.log(mydlo);
  };
  foo.__closure = {
    mydlo,
  };
  foo.__workletHash = 5876054941198;
  foo.__initData = _worklet_5876054941198_init_data;
  return foo;
})();
```

This PR changes that, so instead of using a closure we'd pass all closure variables as arguments to the worklet factory, the IIFE on the right of `const foo` in the output:

```
// new output
const mydlo = 'smiglo';
const _worklet_5876054941198_init_data = {
	// init data
};
const foo = function foo_fileJs1Factory(_worklet_5876054941198_init_data, mydlo) {
  const foo = function () {
    console.log(mydlo);
  };
  foo.__closure = {
    mydlo
  };
  foo.__workletHash = 5876054941198;
  foo.__initData = _worklet_5876054941198_init_data;
  return foo;
}(_worklet_5876054941198_init_data, mydlo);
```

Thanks to that worklet factories no longer have closures, which is a milestone in creating a separate worklet bundle.

As a bonus I'm fixing some bugs and making slight tweaks.

## Test plan

Ran this in our Example app and everything works fine 👍


packages/react-native-worklets/__tests__/plugin.test.ts
```diff
@@ -2355,7 +2355,9 @@ describe('babel plugin', () => {
       </script>`;
 
       const { code } = runPlugin(input);
-      expect(code).toContain('var Clazz__classFactory = function ()');
+      expect(code).toContain(
+        'var Clazz__classFactory = function Clazz__classFactory_null6Factory'
+      );
```
tjzel, line 2360: I have given names to factories, also now the factories have parameters and it's not worth hardcoding them here.

packages/react-native-worklets/plugin/bundle.js
```diff
@@ -10,5 +10,4 @@ esbuild.build({
   logLevel: 'warning',
   platform: 'node',
   sourcemap: 'linked',
-  minify: true,
```
tjzel, line 13: It was fun to have the plugin minified but the community reported it being too obscure in case of debugging/changing the code. It's also annoying for us when debugging.

packages/react-native-worklets/plugin/src/workletFactory.ts
```diff
@@ -227,7 +227,12 @@ export function makeWorkletFactory(
                     identifier(variable.name)
                   )
                 )
-              : objectProperty(identifier(variable.name), variable, false, true)
+              : objectProperty(
+                  cloneNode(variable, true),
+                  cloneNode(variable, true),
+                  false,
+                  true
+                )
```
tjzel, line 235: We always should copy the identifiers. See https://github.com/facebook/metro/issues/1481#issuecomment-2779728253.

packages/react-native-worklets/plugin/src/workletFactory.ts
```diff
@@ -294,9 +299,23 @@ export function makeWorkletFactory(
...
 
-  const newFun = functionExpression(undefined, [], blockStatement(statements));
+  const factoryParams = [
+    cloneNode(initDataId, true),
+    ...closureVariables.map((variableId) => cloneNode(variableId, true)),
+  ];
+
+  const factory = functionExpression(
+    identifier(workletName + 'Factory'),
+    factoryParams,
```
tjzel, line 309: Factory now has a name and parameters.

packages/react-native-worklets/plugin/src/workletSubstitution.ts
```diff
@@ -61,6 +61,10 @@ export function processWorklet(
   const workletFactoryCall = makeWorkletFactoryCall(path, state);
 
   substituteWorkletWithWorkletFactoryCall(path, workletFactoryCall);
+
+  // We recrawl the program parent scope to ensure that all variables are
+  // properly resolved after the substitution.
+  path.scope.getProgramParent().crawl();
```
tjzel, line 67: We should probably also crawl the scope in other places in the code...

MatiPl01:
Since all closure variables are now passed as arguments, will this possibly fix [this](https://github.com/software-mansion/react-native-reanimated/issues/5365) issue with worklets being undefined if used in a wrong order?

packages/react-native-reanimated/src/animation/util.ts
```diff
@@ -41,7 +47,15 @@ import {
...
 
-let IN_STYLE_UPDATER = false;
+/**
+ * This variable has to be an object, because it can't be changed for the
+ * worklets if it's a primitive value. We also have to bind it to a separate
+ * object to prevent from freezing it in development.
+ */
+const IN_STYLE_UPDATER = { current: false };
+const IN_STYLE_UPDATER_UI = makeShareableCloneRecursive({ current: false });
+shareableMappingCache.set(IN_STYLE_UPDATER, IN_STYLE_UPDATER_UI);
```
tjzel, line 57: This has broken semantics. In DEV we try to freeze `IN_STYLE_UPDATER` variable, but it's a primitive so we can't do it. Therefore, worklets having this in their closure would have access to the updated value. With these changes, since the primitive is passed as a parameter, the updates on it are no longer available. Changing the type to the object and preventing it from freezing restores previous semantics (but it's still silly).

tjzel:
@MatiPl01 unfortunately no, that would require topological sorting of worklets which is something we don't plan on implementing atm.

packages/react-native-worklets/__tests__/__snapshots__/plugin.test.ts.snap
```diff
@@ -21,7 +21,8 @@ exports[`babel plugin for DirectiveLiterals doesn't transform string literals 1`
   sourceMap: "\\"mock source map\\"",
   version: "x.y.z"
 };
-var foo = function () {
+var foo = function foo_null1Factory(_ref) {
```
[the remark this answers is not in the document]
tjzel: The transformed "file" is allegedly `/dev/null` - it's required for TS preset to have a filename specified.

packages/react-native-worklets/plugin/src/workletSubstitution.ts
tjzel: When you add referencing identifiers to the AST, i.e.

```js
function foo(){}
bar(foo);
```
becomes

```js
function foo(){}
bar(foo);
baz(foo);
```

The parser won't know that there is a reference to the same `foo` binding in `baz(foo)` until you recrawl the scope. If I for example tried to rename the binding of `foo` to `FOO`, I'd get 

```js
function FOO(){}
bar(FOO);
baz(foo);
```

because the added `foo` wasn't registered as a binding.

packages/react-native-reanimated/src/animation/util.ts
```diff
@@ -540,7 +554,7 @@ export function defineAnimation<
   U extends AnimationObject | StyleLayoutAnimation = T, // type that's received
 >(starting: AnimationToDecoration<T, U>, factory: () => T): T {
   'worklet';
-  if (IN_STYLE_UPDATER) {
+  if (!globalThis._WORKLET && IN_STYLE_UPDATER.current) {
```
piaskowyk, line 557: Could you explain what we need to check `globalThis._WORKLET`?

tjzel: This is more to highlight the context where it's used. `IN_STYLE_UPDATER.current` is never `true` on Worklet Runtimes.

packages/react-native-reanimated/src/animation/util.ts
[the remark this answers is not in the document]
tjzel: We only need to change the value of `IN_STYLE_UPDATER` on the RN Runtime. Using mutable here would mean reading and writing to the UI Runtime, which would be extra slow and unnecessary.

piaskowyk (approved):

PR #7366, opened 2025-04-09
Add Reanimated native view

piaskowyk:
## Summary

This PR is the first phase of rewriting `createAnimatedComponent` to a native view wrapper that uses `display: content`.

Currently, `ReanimatedNativeView` is exported as a public component, but this is just temporary for development and testing purposes. We need to remove that export once we finish rewriting `createAnimatedComponent`.

image

## Test plan

<details>
<summary>code</summary>

```jsx
import React from 'react';
import { ReanimatedNativeView } from 'react-native-reanimated';

  return (
      <Text>Hello world!</Text>
      <ReanimatedNativeView style={{width: 200, backgroundColor: 'green' }}>
        <Text>Reanimated View</Text>
        <View style={{width: 100, height: 100, backgroundColor: 'red' }} />
        <View style={{width: 100, height: 100, backgroundColor: 'blue' }} />
        <View style={{width: 100, height: 100, backgroundColor: 'yellow' }} />
      </ReanimatedNativeView>
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});

```

</details>

### DIFF
--- a/packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/view/ReanimatedView.java
+++ b/packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/view/ReanimatedView.java
@@ -1,0 +1,13 @@
+package com.swmansion.reanimated.view;
+
+import android.content.Context;
+import android.view.ViewGroup;
+
+public class ReanimatedView extends ViewGroup {
+  public ReanimatedView(Context context) {
+    super(context);
+  }
+
+  @Override
+  protected void onLayout(boolean changed, int l, int t, int r, int b) {}
+}
--- a/packages/react-native-reanimated/src/index.ts
+++ b/packages/react-native-reanimated/src/index.ts
@@ -287,9 +287,9 @@ export {
   ScreenTransition,
   startScreenTransition,
 } from './screenTransition';
+export { ReanimatedNativeView } from './specs';
 export type { WorkletRuntime } from 'react-native-worklets';
 export {
   isWorkletFunction,
   LogLevel as ReanimatedLogLevel,
 } from 'react-native-worklets';
-export { ReanimatedView } from './specs'
--- a/packages/react-native-reanimated/src/index.ts
+++ b/packages/react-native-reanimated/src/index.ts
@@ -292,3 +292,4 @@ export {
   isWorkletFunction,
   LogLevel as ReanimatedLogLevel,
 } from 'react-native-worklets';
+export { ReanimatedView } from './specs'
--- a/packages/react-native-reanimated/src/index.ts
+++ b/packages/react-native-reanimated/src/index.ts
@@ -287,7 +287,7 @@ export {
   ScreenTransition,
   startScreenTransition,
 } from './screenTransition';
-export { ReanimatedNativeView } from './specs';
+export { ReanimatedView } from './specs';
 export type { WorkletRuntime } from 'react-native-worklets';
 export {
   isWorkletFunction,
--- a/packages/react-native-reanimated/src/specs/ReanimatedNativeComponent.ts
+++ b/packages/react-native-reanimated/src/specs/ReanimatedNativeComponent.ts
@@ -4,5 +4,4 @@ import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNati
 
 interface NativeProps extends ViewProps {}
 
-// ts-prune-ignore-next-line
 export default codegenNativeComponent<NativeProps>('ReanimatedView');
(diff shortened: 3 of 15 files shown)


packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/view/ReanimatedView.java
```diff
@@ -0,0 +1,13 @@
...
+import android.content.Context;
+import android.view.ViewGroup;
+
+public class ReanimatedView extends ViewGroup {
+  public ReanimatedView(Context context) {
+    super(context);
+  }
+
+  @Override
+  protected void onLayout(boolean changed, int l, int t, int r, int b) {}
```
MatiPl01, line 12: Do we have to override this method or is it just added as an example? I don't think we will need this.

packages/react-native-reanimated/src/specs/ReanimatedNativeComponent.ts
```diff
@@ -0,0 +1,8 @@
+'use strict';
+import type { ViewProps } from 'react-native';
+import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
+
+interface NativeProps extends ViewProps {}
+
+// ts-prune-ignore-next-line
```
MatiPl01, line 7: Why do you need this?

MatiPl01 (approved):
Looks fine. I left just 2 questions.

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/view/ReanimatedView.java
piaskowyk: I've tried but `onLayout` is an abstract method and we need to implement it.
image

packages/react-native-reanimated/src/specs/ReanimatedNativeComponent.ts
piaskowyk: The `ts-prune` tool marks it as unused code because it can't resolve the import. This is due to the fact that we import that file conditionally using `require`.

```js
if (shouldBeUseWeb()) {
  // this is a workaround for the jest
  ReanimatedNativeView = {} as HostComponent<any>;
} else {
  ReanimatedNativeView = require('./ReanimatedNativeComponent').default;
}
```

packages/react-native-reanimated/src/index.ts
```diff
@@ -287,6 +287,7 @@ export {
   ScreenTransition,
   startScreenTransition,
 } from './screenTransition';
+export { ReanimatedNativeView } from './specs';
```
tomekzaw, line 290: How about `ReanimatedView` instead?

packages/react-native-reanimated/src/specs/ReanimatedViewProvider.native.ts
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+import type { HostComponent } from 'react-native';
+
+import { shouldBeUseWeb } from '../PlatformChecker';
+
+let ReanimatedNativeView: HostComponent<any>;
+if (shouldBeUseWeb()) {
+  // this is a workaround for the jest
+  ReanimatedNativeView = {} as HostComponent<any>;
```
tomekzaw, line 10: 1. Shouldn't we use `isJest()` instead?
2. Does react-native-gesture-handler need to do the same thing? Maybe we should add the workaround for Jest at a different abstraction layer?

packages/react-native-reanimated/src/specs/ReanimatedViewProvider.ts
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+import type { HostComponent } from 'react-native';
+
+export default {} as HostComponent<any>;
```
tomekzaw, line 5: What does this file do?

piaskowyk: I missed a comment. This is a workaround for Next.js 😕. Next.js can’t resolve the `react-native/Libraries/Utilities/codegenNativeComponent` import, and during SSG, we can’t detect if it is a web environment, so we can't return a mock implementation. To avoid this issue, we need two versions of `ReanimatedViewProvide`: one with a `.ts` extension for Next.js and web, and another with a `.native.ts` extension for React Native (since only Metro understands platform-dependent extensions).

packages/react-native-reanimated/src/specs/ReanimatedViewProvider.native.ts
piaskowyk: 1. Yes, we can 👍
2. Jest is awful - but I have come up with another idea

packages/react-native-reanimated/src/index.ts
piaskowyk: I suggested that name to be more explicit, but we can use just `ReanimatedView` instead 👍

PR #7368, opened 2025-04-09
fix: web-example font and circular imports issues

MatiPl01:
## Summary

This PR fixes small issues in the web example:

1. Circular imports (likely also in other examples but I noticed it in the web console) - basically, the `NukeContext` was imported from the main `App` file to the `FreezeExample` which was rendered within this `App` component.

2. Adds custom font imports - I load them via the custom style tag as I couldn't get it working right by using `expo-font`. Basically, the main problem was that `expo-font` didn't allow me to load different font weights under the same font name, which is what I want - I want to manipulate the font weight via the `fontWeight` property, not by using a different name of the font.

### Before

Screenshot 2025-04-09 at 11 54 28

Screenshot 2025-04-09 at 11 54 43

### After

Screenshot 2025-04-09 at 11 54 56


piaskowyk (approved):
🚀

PR #7370, opened 2025-04-09
fix: timestamp on physical device in s instead of ms

WoLewicki:
## Summary
During recent changes, multiplying the `currentTimestamp` value by `MILLISECONDS_IN_SECOND` on physical device was lost. Restoring the correct behavior.

## Test plan

Run any app examples on simulator and physical device and see no difference with this PR.


patrycjakalinska (approved):
🪨

tjzel (approved):

Issue #6645, opened 2024-10-28
Snapshots excessively large or failing with `Invalid string length`


BrianValente:
### Description

Since Reanimated v3.16.0, snapshots using `@testing-library/react-native` have been breaking. In some cases, the snapshot size has increased drastically, going from less than 1 KB to over 250 MB. In other cases, snapshots fail to generate, throwing an `Invalid string length` error. 

### v3.15.5
```
 PASS  ./ComponentTwo.test.tsx
 › 1 snapshot written.
 PASS  ./ComponentOne.test.tsx
 › 1 snapshot written.

Snapshot Summary
 › 2 snapshots written from 2 test suites.

Test Suites: 2 passed, 2 total
Tests:       2 passed, 2 total
Snapshots:   2 written, 2 total
Time:        1.938 s, estimated 22 s
```
```
➜ ls -lh
total 16
-rw-r--r--@ 1 brianvalente  staff   586B Oct 28 16:22 ComponentOne.test.tsx.snap
-rw-r--r--@ 1 brianvalente  staff   925B Oct 28 16:22 ComponentTwo.test.tsx.snap
```

### v3.16.1

```
 PASS  ./ComponentOne.test.tsx (6.267 s)
 › 1 snapshot written.
 FAIL  ./ComponentTwo.test.tsx (9.18 s)
  ● ComponentTwo › should pass

    RangeError: Invalid string length

       6 |     it("should pass", () => {
       7 |         const rendered = render(<ComponentTwo />).toJSON();
    >  8 |         expect(rendered).toMatchSnapshot();
         |                          ^
       9 |     });
      10 | })
      11 |

      at printElement (node_modules/jest-snapshot/node_modules/pretty-format/build/plugins/lib/markup.js:105:36)
      at Object.toMatchSnapshot (ComponentTwo.test.tsx:8:26)

Snapshot Summary
 › 1 snapshot written from 1 test suite.

Test Suites: 1 failed, 1 passed, 2 total
Tests:       1 failed, 1 passed, 2 total
Snapshots:   1 written, 1 total
Time:        9.404 s
```

```
➜ ls -lh
total 537280
-rw-r--r--@ 1 brianvalente  staff   262M Oct 28 16:23 ComponentOne.test.tsx.snap
```

### Steps to reproduce

1. Clone [demo repo](https://github.com/BrianValente/reanimated-jest-snapshot-issue-demo)
2. Install dependencies
3. Run `npm run test`

[...]

PR #7372, opened 2025-04-09
fix: Jest snapshots

tjzel:
## Summary

Adding the feature of Inline Style testing #6400 in Jest broke snapshots, since it pushed the whole `this.props.style` object to `jestInlineStyles` object. `this.props.style` contains Animated Style handles, which are recursive structures and cannot be stringified for snapshots. This pull requests filters out Animated Styles for `jestInlineStyles` and also adds snapshot tests to make sure everything can be stringized.

Fixes https://github.com/software-mansion/react-native-reanimated/issues/6645

## Test plan

Jest tests pass.


packages/react-native-reanimated/__tests__/animatedProps.test.tsx
```diff
@@ -69,7 +69,12 @@ describe('animatedProps', () => {
     const circle = getByTestId('circle');
 
     expect(circle).toHaveAnimatedProps({});
+
+    // TODO: There's still a problem with stringizing `TextInput` with `animatedProps`.
+    // const rendered = render(<AnimatedComponent />).toJSON();
+    // expect(rendered).toMatchSnapshot();
```
tjzel, line 75: I was debugging it but didn't reach any conclusions yet. I'm wondering if we should merge this PR regardless, since the problem here is with stringizing `animatedProps`, which aren't directly related to this PR.

packages/react-native-reanimated/__tests__/animatedProps.test.tsx
```diff
@@ -81,5 +86,9 @@ describe('animatedProps', () => {
     jest.advanceTimersByTime(animationDuration);
 
     expect(textInput).toHaveAnimatedProps({ text: 'Box width: 30' });
+
+    // TODO: There's still a problem with stringizing `TextInput` with `animatedProps`.
+    // const rendered = render(<AnimatedComponent />).toJSON();
+    // expect(rendered).toMatchSnapshot();
```
tjzel, line 92: I was debugging it but didn't reach any conclusions yet. I'm wondering if we should merge this PR regardless, since the problem here is with stringizing `animatedProps`, which aren't directly related to this PR.

packages/react-native-reanimated/__tests__/hooks.useSharedValue.test.tsx
```diff
@@ -20,35 +20,23 @@ describe('useSharedValue', () => {
...
+
+    component.update(<TestComponent key="box" value={updatedValue} />);
+    const updatedChild = component.getByTestId('AnimatedView');
+
+    expect(updatedChild.props.style[0].opacity).toBe(initialValue);
+
+    const rendered = render(
+      <TestComponent key="box" value={updatedValue} />
+    ).toJSON();
+    expect(rendered).toMatchSnapshot();
```
tjzel, line 40: Had to redo this test because it was outdated.

packages/react-native-reanimated/__tests__/animatedProps.test.tsx
patrycjakalinska: After giving this some thought, I think we should address the problem with stringizing `animatedProps` in another PR - `animatedProps` in jest is a new feature, and I'm not sure how complex the solution would be. 

And as you've said it's not directly related, so we should keep this PR as a single responsibility c:

packages/react-native-reanimated/src/createAnimatedComponent/AnimatedComponent.tsx
```diff
@@ -426,3 +428,19 @@ export default class AnimatedComponent
...
 }
+
+function filterOutAnimatedStyles(
+  style: NestedArray<StyleProps | AnimatedStyleHandle>
+): NestedArray<StyleProps> {
+  if (!Array.isArray(style)) {
+    return style?.viewDescriptors ? {} : style;
+  }
+  return style
+    .filter((styleElement) => !('viewDescriptors' in styleElement))
```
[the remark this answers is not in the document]
patrycjakalinska: I think that by design the styleElement won't have `undefined` values, and as you said `styleElement` being `undefined` would be indicative of another issue so the check is not necessary -> @tjzel, what do you think?

patrycjakalinska (approved):
I looked at it, ran it, inspected it and everything looks good 💯

piaskowyk (approved):

packages/react-native-reanimated/src/createAnimatedComponent/AnimatedComponent.tsx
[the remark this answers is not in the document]
tjzel: Actually it seems that our internal types aren't exactly accurate in this place - unless I missed some prefiltering - and `this.props.style` can contain `null` and `undefined`. I added special handling of these in [ff3710a](https://github.com/software-mansion/react-native-reanimated/pull/7372/commits/ff3710a342acb5265ab9ec696bdf5ad1dbad7f87)

PR #7376, opened 2025-04-10
fix: Initial animated style updater style is applied on re-render

### LINKED ISSUE #7375: [v3, Paper] Initial animated style updater style is applied on re-render
### Description

The issue was originally reported by @efstathiosntonas on the `react-native-sortables` repository ([here](https://github.com/MatiPl01/react-native-sortables/issues/344)). This regression was introduced after removing style flattening in [this](https://github.com/software-mansion/react-native-reanimated/pull/7021) PR and then cherry-picked to 3.17.2 in [this](https://github.com/software-mansion/react-native-reanimated/pull/7327) PR.

Basically, the issue can be seen if the animated view is passed more than one animated style (the additional style can be even empty). When a view with such styles re-render, the style from the initial style updater run is applied to the view instead of the last returned style. This is best visible if the `useAnimatedStyle` hook is used to create a discrete style updated based a condition (e.g. a Shared Value is a flag indicating whether to return one or the other style) and doesn't contain animations.

The issue can be fixed by just adding the style flattening in [this](https://github.com/software-mansion/react-native-reanimated/blob/df96270082846776f8ab5bc5c75dccfdf6722811/packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx#L56) line as follows:

```tsx
props[key] = StyleSheet.flatten(processedStyle);
```

### Issue recording

Observe the logs. You can see, that when re-render happens, the width of the box changes, even though the value returned within the `useAnimatedStyle` callback doesn't change.

https://github.com/user-attachments/assets/3fc19688-a5a5-4a8e-94a3-1c5fe533c94b

### Expected behavior

After bringing back the style flattening

https://github.com/user-attachments/assets/14b932c0-1e02-420f-ba86-0becc69c6b43

### Steps to reproduce

1. Copy code from the expo snack or from below
2. Paste the code to the react-native app with reanimated 3.17.2 or 3.17.3
3. Observe that the width of the view is set back to `100` after re-render

<details>
<summary>
Source code
</summary>

```tsx
import Animated, {
  LinearTransition,
  useAnimatedStyle,
  useSharedValue,

export default function PlaygroundExample() {
  const [_, setCount] = useState(0);

  const animatedWidth = useSharedValue(0);

  const animatedWidthStyle = useAnimatedStyle(() => {
    if (animatedWidth.value === 0) {
      console.log('predefined width');
      return { width: 100 };
    }

    console.log('animatedWidth.value', animatedWidth.value);
    return {
      width: animatedWidth.value,
    };
  });

  const emptyAnimatedStyle = useAnimatedStyle(() => ({}));

  useEffect(() => {
    animatedWidth.value = 10;

    setTimeout(() => {
      console.log('re-render');
      setCount((prev) => prev + 1);
    }, 1000);

    setTimeout(() => {
      animatedWidth.value = 20;
    }, 4000);
  }, [animatedWidth]);

  return (
    <Animated.View
      layout={LinearTransition}
      style={[emptyAnimatedStyle, animatedWidthStyle]}>
      <View style={styles.box} />
    </Animated.View>
  );
}

  box: {
    backgroundColor: 'red',
    height: 100,
  },
});
```
</details>

https://snack.expo.dev/vZOtho0RKCBlZNXu4uI1q

### Reanimated version

3.17.2

### React Native version

0.79.0

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

MatiPl01:
## Summary

This PR fixes #7375.

### The problem

After removing style flattening the old implementation started being invalid because the `_initialStyle` object that was created during the initial render of the component contained props from all animated styles. If a view had more than just one animated style, this object contained props from all of these styles.

Let's say we had these 3 animated styles:

```tsx
const aStyle1 = useAnimatedStyle(() => ({})); // empty

const aStyle2 = useAnimatedStyle(() => ({
  height: 100
}));

const aStyle3 = useAnimatedStyle(() => ({
  width: 100
}));

<Animated.View style={[aStyle1, aStyle2, aStyle3]} />
```

The style array on the first render looked like this:

```
[{}, { height: 100 }, { height: 100, width: 100 }] // subsequent objects contain all props from previous objects because of merging
```

On the next render, however, the `_initialStyle` was used for all animated styles, so we ended up with this array:

```
[{ height: 100, width: 100 }, { height: 100, width: 100 }, { height: 100, width: 100 }]
```

As a result, 2 of 3 styles changed. Because the implementation is different on Paper and we have no props registry (similar to what we have on Fabric), this change resulted in the view's style update, even though the animated style wasn't recalculated and didn't return a new style. This is an incorrect behavior resulting in problems described in the attached issue.

### The solution

I decided to create a map with style objects and their initial run values. As a result, instead of returning the same single style object with all style props from all animated styles, we can now return proper initial style values for a specific animated style.

## Test plan

Copy the example from #7375 issue and see that it now works.

### DIFF
--- a/packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx
+++ b/packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx
@@ -2,6 +2,7 @@
 
 import { initialUpdaterRun } from '../animation';
 import type { StyleProps } from '../commonTypes';
+import type { AnimatedStyleHandle } from '../hook/commonTypes';
 import { isSharedValue } from '../isSharedValue';
 import { isChromeDebugger } from '../PlatformChecker';
 import { WorkletEventHandler } from '../WorkletEventHandler';
@@ -21,7 +22,7 @@ function dummyListener() {
 }
 
 export class PropsFilter implements IPropsFilter {
-  private _initialStyle = {};
+  private _initialPropsMap = new Map<AnimatedStyleHandle, StyleProps>();
 
   public filterNonAnimatedProps(
     component: React.Component<unknown, unknown> & IAnimatedComponentInternal
@@ -29,22 +30,25 @@ export class PropsFilter implements IPropsFilter {
     const inputProps =
       component.props as AnimatedComponentProps<InitialComponentProps>;
     const props: Record<string, unknown> = {};
+
     for (const key in inputProps) {
       const value = inputProps[key];
       if (key === 'style') {
         const styleProp = inputProps.style;
         const styles = flattenArray<StyleProps>(styleProp ?? []);
+
         const processedStyle: StyleProps[] = styles.map((style) => {
           if (style && style.viewDescriptors) {
-            // this is how we recognize styles returned by useAnimatedStyle
+            const handle = style as AnimatedStyleHandle;
+
             if (component._isFirstRender) {
-              this._initialStyle = {
-                ...style.initial.value,
-                ...this._initialStyle,
-                ...initialUpdaterRun<StyleProps>(style.initial.updater),
-              };
+              this._initialPropsMap.set(handle, {
+                ...handle.initial.value,
+                ...initialUpdaterRun(handle.initial.updater),
+              } as StyleProps);
             }
-            return this._initialStyle;
+
+            return this._initialPropsMap.get(handle) ?? {};
           } else if (hasInlineStyles(style)) {
             return getInlineStyle(style, component._isFirstRender);
           } else {


MatiPl01:
I tested this solution on a few examples on Paper and Fabric. It seems to work on both but there was no issue on Fabric so I wonder if we want to merge it on the `main` branch as well.

tomekzaw (approved):
LGTM

efstathiosntonas:
thank you fellas!! ❤️❤️

efstathiosntonas:
I think this should be released asap, it was really tricky in my case to find out that reanimated was the cause  and posisbly many packages or daily coding could be affected!

MatiPl01:
> I think this should be released asap, it was really tricky in my case to find out that reanimated was the cause and possibly many packages or daily coding could be affected!

We will release a new version at the beginning of the next week. For now, you can patch reanimated.

PR #7379, opened 2025-04-10
Support for all transform origin values

skusnierz:
## Summary
The following PR adds support for all missing values for the `transformOrigin` prop, such as : 
- `top left`
- `50% 50%`
- `100 100 20`
- `100px 100px`
- `100px 50%`

## Test plan
The example `[100px 100px 100px]` (red color) should throw an error because the last value of `ZIndex` must always be a number. 

Note: The error message is incorrect but this is due to the errror being thrown incorrectly and not the example itself - @tjzel is working to improve error handling
image

<video src="https://github.com/user-attachments/assets/2ba7eb54-d1f0-4178-9c7f-6c803eb86e81" />


packages/react-native-reanimated/src/updateProps/processTransformOrigin.ts
```diff
@@ -0,0 +1,161 @@
...
+    throw new ReanimatedError(
+      `Transform origin x-position must be a number or percentage string. Passed value: ${x}.`
+    );
+  }
+  if (!(typeof y === 'number' || (typeof y === 'string' && y.endsWith('%')))) {
+    throw new ReanimatedError(
+      `Transform origin y-position must be a number or percentage string. Passed value: ${y}.`
+    );
+  }
+  if (!(typeof z === 'number')) {
```
MatiPl01, line 27:
```suggestion
  if (typeof z !== 'number') {
```
simplify

packages/react-native-reanimated/src/updateProps/processTransformOrigin.ts
```diff
@@ -0,0 +1,161 @@
...
+  transformOrigin: Array<string | number>
+) {
+  'worklet';
+  if (transformOrigin.length !== 3) {
+    throw new ReanimatedError('Transform origin must have exactly 3 values.');
+  }
+  const [x, y, z] = transformOrigin;
+  if (!(typeof x === 'number' || (typeof x === 'string' && x.endsWith('%')))) {
+    throw new ReanimatedError(
+      `Transform origin x-position must be a number or percentage string. Passed value: ${x}.`
```
MatiPl01, line 19:
```suggestion
      `Transform origin x-position must be a number or a percentage string. Passed value: ${x}.`
```

packages/react-native-reanimated/src/updateProps/processTransformOrigin.ts
```diff
@@ -0,0 +1,161 @@
...
+    );
+  }
+  if (!(typeof y === 'number' || (typeof y === 'string' && y.endsWith('%')))) {
+    throw new ReanimatedError(
+      `Transform origin y-position must be a number or percentage string. Passed value: ${y}.`
```
MatiPl01, line 24:
```suggestion
      `Transform origin y-position must be a number or a percentage string. Passed value: ${y}.`
```

packages/react-native-reanimated/src/updateProps/processTransformOrigin.ts
```diff
@@ -0,0 +1,161 @@
...
+            }
+            transformOriginArray[index] = numericValue;
+          }
+          break;
+        }
+      }
+
+      if (regex.lastIndex === matches.index) {
+        regex.lastIndex++;
+      }
```
MatiPl01, line 136: Is this needed? I can't see something like this in the React Native's implementation

packages/react-native-reanimated/src/updateProps/processTransformOrigin.ts
```diff
@@ -0,0 +1,161 @@
...
+const INDEX_Y = 1;
+const INDEX_Z = 2;
+
+// Implementation based on https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/StyleSheet/processTransformOrigin.js
+export function validateTransformOrigin(
+  transformOrigin: Array<string | number>
+) {
+  'worklet';
+  if (transformOrigin.length !== 3) {
+    throw new ReanimatedError('Transform origin must have exactly 3 values.');
```
MatiPl01, line 14: I think that this error will never be thrown because of how you implemented `processTransformOrigin`, which always returns an array with 3 values. Either remove this error or don't return a 3-value array from the `processTransformOrigin` and don't add default values for missing transform origin components.

packages/react-native-reanimated/src/updateProps/processTransformOrigin.ts
```diff
@@ -0,0 +1,161 @@
...
+    transformOrigin = [...transformOriginIn];
+    if (transformOrigin.length < 3) {
+      if (transformOrigin.length < 2) {
+        transformOrigin.push('50%');
+      }
+      transformOrigin.push(0);
+    }
+    if (transformOrigin.length > 3) {
+      transformOrigin = transformOrigin.slice(0, 3);
+    }
```
MatiPl01, line 153: I think that this part is not necessary, especially if you want to display an error for invalid transform origins that have a different number of components than 3. React native's implementation throws error for transform origins with missing values or with more values than expected so maybe keep it consistent.

Alternatively, if you prefer to add default values for the user, you might consider reusing the implementation I made for CSS animations and transitions (in the `packages/react-native-reanimated/src/css/platform/native/style/processors/transformOrigin.ts` file) that works mor like the CSS implementation of the transform origin on the web.

packages/react-native-reanimated/src/updateProps/updateProps.ts
```diff
@@ -33,10 +38,9 @@ if (shouldBeUseWeb()) {
   updateProps = (viewDescriptors, updates) => {
     'worklet';
     processColorsInProps(updates);
-    if (updates.transformOrigin) {
-      if (!Array.isArray(updates.transformOrigin)) {
-        throw new ReanimatedError('Please use transformOrigin in array form');
-      }
+    if ('transformOrigin' in updates) {
+      updates.transformOrigin = processTransformOrigin(updates.transformOrigin);
+      validateTransformOrigin(updates.transformOrigin);
```
MatiPl01, line 43: I would prefer to keep validation within the `processTransformOrigin`, similarly to the React Native's implementation. Also make validation dev-only not to affect prod performance (`if (__DEV__) { ... }`).

packages/react-native-reanimated/src/updateProps/updateProps.ts
```diff
@@ -33,9 +38,10 @@ if (shouldBeUseWeb()) {
   updateProps = (viewDescriptors, updates) => {
     'worklet';
     processColorsInProps(updates);
-    if (updates.transformOrigin) {
-      if (!Array.isArray(updates.transformOrigin)) {
-        throw new ReanimatedError('Please use transformOrigin in array form');
+    if ('transformOrigin' in updates) {
+      updates.transformOrigin = processTransformOrigin(updates.transformOrigin);
+      if (__DEV__) {
+        validateTransformOrigin(updates.transformOrigin);
```
MatiPl01, line 44: Don't we want to add validation in the `processTransformOrigin` in case we decide to use `processTransformOrigin` somewhere else? I think it will be safer to validate the result before returning it to ensure that we won't forget to add validation if the `processTransformOrigin` is used in other place in the code.

What do you think about that?

MatiPl01 (approved):
Good for me. I left a few small suggestions. It would be nice if you could address them but you don't have to.

Please also wait for @tomekzaw's or someone else's review.

packages/react-native-reanimated/src/updateProps/updateProps.ts
skusnierz: it makes sense

apps/common-app/src/apps/reanimated/examples/TransformOriginExample.tsx
```diff
@@ -0,0 +1,139 @@
...
+        },
+      ],
+      transformOrigin: origin,
+    };
+  });
+
+  const handlePress = (newOrigin: TransformOriginOption) => {
+    currentOrigin.value = newOrigin;
+
+    cancelAnimation(progress);
```
piaskowyk, line 34: Since you're assigning 0 directly to the progress shared value, I think calling `cancelAnimation` is not necessary

piaskowyk (approved):
Looks good to me 👍

PR #7380, opened 2025-04-10
Custom Shadow Nodes

piaskowyk:
## Summary

This PR introduces a common implementation of custom shadow nodes for both Android and iOS. It's part of the migration towards using `display: content`.

Screenshot 2025-04-11 at 13 42 07

Screenshot 2025-04-11 at 13 42 26

## Test plan

<details>

```js
import { Button, StyleSheet, Text, View } from 'react-native';
import { ReanimatedView } from 'react-native-reanimated';

  const [color, setColor] = useState('red');
  return (
      <Text>Hello world!</Text>
      <ReanimatedView style={{width: 200, backgroundColor: 'green' }}>
        <Text>Reanimated View</Text>
        <View style={{width: 100, height: 100, backgroundColor: color }} />
        <View style={{width: 100, height: 100, backgroundColor: 'blue' }} />
        <View style={{width: 100, height: 100, backgroundColor: 'yellow' }} />
      </ReanimatedView>
      <Button title='Click' onPress={() => {
        setColor(color === 'red' ? 'olive' : 'red');
      }} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});

```

</details>


packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ShadowNodes.cpp
```diff
@@ -0,0 +1,12 @@
+#include "ShadowNodes.h"
+#include <react/renderer/core/LayoutContext.h>
```
MatiPl01, line 2: Maybe store all includes in the .h file? I think it is a better and cleaner approach.

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ShadowNodes.cpp
```diff
@@ -0,0 +1,12 @@
...
+
+namespace facebook::react {
+
+extern const char ReanimatedViewComponentName[] = "ReanimatedView";
+
+void ReanimatedViewCustomShadowNode::layout(LayoutContext layoutContext) {
```
MatiPl01, line 8: Maybe use shorter name? Using this long name in multiple places will be cumbersome. I think that you can name it `ReanimatedShadowNode` and this shorter name will be self-explanatory (we don't have to make it more specific).

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ShadowNodes.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <jsi/jsi.h>
+#include <react/renderer/components/rnreanimated/EventEmitters.h>
+#include <react/renderer/components/rnreanimated/Props.h>
+#include <react/renderer/components/view/ConcreteViewShadowNode.h>
+#include <react/renderer/core/LayoutContext.h>
+#include "States.h"
```
MatiPl01, line 8: The same as before, use a full path

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/States.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#ifdef ANDROID
+#include <folly/dynamic.h>
+#endif
+
+namespace facebook::react {
+
+class ReanimatedViewState {
```
MatiPl01, line 9: I know that this class is required by the `ConcreteViewShadowNode` template param but maybe you know and could explain to me how the shadow node state works and what it can be used for? I can see that this is a minimal (empty) implementation but I wonder how we can use it in the future.

packages/react-native-reanimated/apple/reanimated/apple/view/ReanimatedView.mm
```diff
@@ -1,4 +1,5 @@
 #import <reanimated/apple/view/ReanimatedView.h>
+#import <reanimated/view/react/renderer/components/rnreanimated/ComponentDescriptors.h>
```
MatiPl01, line 2: Do we need such a long path to the native components directory? I can see that you nested these 4 directories one in another `/react/renderer/components/rnreanimated/` and I wonder if this path must be that long because of the codegen or if there is a different reason.

MatiPl01 (commented):
Looks very well. I left just a few comments.

packages/react-native-reanimated/apple/reanimated/apple/view/ReanimatedView.mm
piaskowyk: That `react/renderer/components/rnreanimated` path is required by codegen and autolinking 😕

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/States.h
piaskowyk: This is an internal state associated with the shadow node that can only be updated via `commit`. For example, the scroll view uses this state to store information about its offset. I'm not sure if we need a custom state, but let's keep it until we determine if it can help us somehow.

MatiPl01: Ok, thanks for explanation

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h
```diff
@@ -0,0 +1,15 @@
...
+#include <react/renderer/core/ConcreteComponentDescriptor.h>
+#include <react/renderer/components/rnreanimated/ReanimatedShadowNode.h>
+
+namespace facebook::react {
+
+using ReanimatedViewComponentDescriptor =
+    ConcreteComponentDescriptor<ReanimatedShadowNode>;
+
+void rnreanimated_registerComponentDescriptorsFromCodegen(
+    std::shared_ptr<const ComponentDescriptorProviderRegistry> registry);
```
tomekzaw, line 13: Can we use a const ref here?
```suggestion
    const std::shared_ptr<const ComponentDescriptorProviderRegistry> &registry);
```

packages/react-native-reanimated/apple/reanimated/apple/view/ReanimatedView.mm
```diff
@@ -1,3 +1,4 @@
+#import <react/renderer/components/rnreanimated/ComponentDescriptors.h>
 #import <reanimated/apple/view/ReanimatedView.h>
 
 #import <react/renderer/components/rnreanimated/ComponentDescriptors.h>
```
tomekzaw, line 4: Duplicate import

tomekzaw (approved):

MatiPl01 (approved):
LGTM. Please address @tomekzaw comments and should be good to go. 🚀

PR #7390, opened 2025-04-14
fix: separate implementation for custom errors

tjzel:
## Summary

Turns out that exporting `createCustomError` function is error-prone and sometimes the errors don't work properly. The solution to that is duplicating the code - it's quite small so it shouldn't be a problem.

## Test plan

Trigger `WorkletsError` and `ReanimatedError` and see they work properly now.


patrycjakalinska (changes requested):
one small adjustment, other than that tested and working great

patrycjakalinska (approved):
🪨

PR #7397, opened 2025-04-15
fix: Restore accidentally removed builder bob patch

MatiPl01:
## Summary

Patch was accidentally removed in [this](https://github.com/software-mansion/react-native-reanimated/commit/47701784d77a063e4a569f7797bc95f74bd9b7f6#r155405465) PR.

### DIFF
--- a/packages/react-native-reanimated/package.json
+++ b/packages/react-native-reanimated/package.json
@@ -148,7 +148,7 @@
     "prettier": "^3.3.3",
     "react": "19.0.0",
     "react-native": "0.79.0",
-    "react-native-builder-bob": "0.33.1",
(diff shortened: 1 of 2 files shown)


packages/react-native-reanimated/package.json
```diff
@@ -148,7 +148,7 @@
     "prettier": "^3.3.3",
     "react": "19.0.0",
     "react-native": "0.79.0",
-    "react-native-builder-bob": "0.33.1",
+    "react-native-builder-bob": "patch:react-native-builder-bob@npm%3A0.33.1#~/.yarn/patches/react-native-builder-bob-npm-0.33.1-383d9e23a5.patch",
```
tjzel, line 151: Please also use it in Worklets workspace 🙏

tjzel (approved):

packages/react-native-reanimated/package.json
MatiPl01: worklets already use the patch. It was just mistakenly changed in the reanimated package.

PR #7398, opened 2025-04-15
fix: lack of hidden hermes flags

tjzel:
## Summary

TODO:

- [ ] Android part
- [ ] Backport to v3 after merge

## Test plan

- Fabric example works in runtime:
  - [ ] debug ios
  - [ ] debug android
  - [ ] release ios
  - [ ] release android
- TVos example works in runtime:
  - [ ] debug ios
  - [ ] debug android
  - [ ] release ios
  - [ ] release android
- MacOS example works in runtime:
  - [ ] debug ios
  - [ ] debug android
  - [ ] release ios
  - [ ] release android


packages/react-native-reanimated/RNReanimated.podspec
```diff
@@ -7,14 +7,12 @@ assert_minimal_react_native_version($config)
...
 $new_arch_enabled = ENV['RCT_NEW_ARCH_ENABLED'] != '0'
 assert_new_architecture_enabled($new_arch_enabled)
-is_release = ENV['PRODUCTION'] == '1'
 
 folly_flags = "-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32"
 boost_compiler_flags = '-Wno-documentation'
 fabric_flags = $new_arch_enabled ? '-DRCT_NEW_ARCH_ENABLED' : ''
 example_flag = $config[:is_reanimated_example_app] ? '-DIS_REANIMATED_EXAMPLE_APP' : ''
 version_flags = "-DREACT_NATIVE_MINOR_VERSION=#{$config[:react_native_minor_version]} -DREANIMATED_VERSION=#{reanimated_package_json['version']}"
-debug_flag = is_release ? '-DNDEBUG' : ''
```
tomekzaw, line 17: We use `#ifndef NDEBUG` in Reanimated codebase, is it really safe to remove it from here?

packages/react-native-reanimated/RNReanimated.podspec
```diff
@@ -56,10 +54,13 @@ Pod::Spec.new do |s|
...
-  if !is_release
-    gcc_debug_definitions << " HERMES_ENABLE_DEBUGGER=1"
-  end
+  # React Native doesn't expose these flags, but not having them
+  # can lead to runtime errors due to ABI mismatches.
+  # There's also
+  #   HERMESVM_PROFILER_OPCODE
+  #   HERMESVM_PROFILER_BB
+  # which shouldn't be defined in standard setups.
+  hermes_debug_hidden_flags = '$(inherited) HERMES_ENABLE_DEBUGGER=1'
```
tomekzaw, line 63: Why don't we just inline it directly in `GCC_PREPROCESSOR_DEFINITIONS[config=Debug]` like this?
```rb
"GCC_PREPROCESSOR_DEFINITIONS[config=Debug]" => '$(inherited) HERMES_ENABLE_DEBUGGER=1',
```

packages/react-native-reanimated/RNReanimated.podspec
tjzel: It should be downstreamed by RN after this change: https://github.com/facebook/react-native/pull/41715

packages/react-native-reanimated/RNReanimated.podspec
tjzel: I wanted to keep the lengthy comment above it.

tomekzaw: You can place the length comment above.

tjzel: Doesn't look as neat.

piaskowyk (commented):
👌

macksal:
@patrycjakalinska  Is it possible to backport this for 3.17.x? 3.18.0 seems to require react-native 0.80. For expo users, that requires moving to expo@54 which only came out of beta yesterday.

patrycjakalinska:
hi @macksal, 3.18.x is compatible with RN `0.75` all up to RN `0.80` so there should not be any problem with versioning - for more info on compatibility please refer [here](https://docs.swmansion.com/react-native-reanimated/docs/guides/compatibility/).

macksal:
> hi @macksal, 3.18.x is compatible with RN `0.75` all up to RN `0.80` so there should not be any problem with versioning - for more info on compatibility please refer [here](https://docs.swmansion.com/react-native-reanimated/docs/guides/compatibility/).

thx, it seems like expo's dependency checker is overzealous about having just a single recommended version of every package. They're often slow at picking up important bugfixes on the recommended version, too... not a great combo

PR #7394, opened 2025-04-15
(cherry-pick) fix: Support optional arguments in boxShadow string (#7386)

patrycjakalinska:
Cherry pick of #7386 
This PR updates the `parseBoxShadowString` function to handle `boxShadow` strings with missing optional arguments (`blurRadius`, `spreadDistance`, `color`).

It now gracefully handles input like:

- `10px 10px white` (defaults `blurRadius` and `spreadDistance` to `0px`)

- `10px 10px 5px white` (defaults `spreadDistance` to `0px`)

- `10px 10px` (defaults remaining values, uses black as color)

The default happen on Native side where RN process it and defaults the missing values.

I've also added unit test with stripped string that make sure styles behave correctly.

I've also moved the string type check into the `processBoxShadow.js` file, since we want to support `boxShadow` values that are arrays of objects where individual properties can be strings with units. (The possibility to support this was already implemented but `typeof newValue.boxShadow === 'string'` prevented it from existing)
 
 ```
       boxShadow: [
        {
         // before that's how value had been provided
          offsetX: 10
          // now they can be provided also as strings with units
          offsetY: '4px',
          blurRadius,
          color,
        },
      ],
 ```
 
 

## Test plan

<details>
<summary>Example Code</summary>

```tsx
import React from 'react';
import { Pressable, Text, View } from 'react-native';
import Animated, {
  interpolate,
  interpolateColor,
  useAnimatedProps,
  useAnimatedStyle,
  useSharedValue,
} from 'react-native-reanimated';
import { DefaultStyle } from 'react-native-reanimated/lib/typescript/hook/commonTypes';

const AnimatedPressable = Animated.createAnimatedComponent(Pressable);

export default function EmptyExample() {
  const pressed = useSharedValue(0);

  const animatedBoxShadow = useAnimatedStyle(() => {
    const blurRadius = interpolate(pressed.value, [0, 1], [10, 0]);
    const color = interpolateColor(
      pressed.value,
      [0, 1],
      ['rgba(255, 0, 0, 1)', 'rgba(0, 0, 255, 1)']
    );

    return {
      boxShadow: `0px 4px ${blurRadius}px ${color}`,
    };
  });

  const handlePress = () => {
    pressed.value = pressed.value === 0 ? 1 : 0;
  };

  return (
    <View
      style={{
        padding: 24,
      }}>
      <AnimatedPressable
        testID={'pressable'}
        style={[
          animatedBoxShadow,
          {
            backgroundColor: 'red',
            padding: 16,
[...]

PR #7350, opened 2025-04-04
chore: Bump macos and tvos example dependencies

MatiPl01:
## Summary

This PR bumps another 2 example apps:
- macos: `0.75.4` -> `0.77.1`
- tvos: `0.77.0-0` -> `0.79.0-0rc2`

## Current state

|| MacOS | TVOS (Android) | TVOS (iOS) |
|-|-|-|-|
| Builds           | ✅ | ✅ | ✅ |
| Runs             | ✅ | ✅ | ✅ |

PR #7399, opened 2025-04-15
Release 3.17.4

MatiPl01:
## Summary

Included fixes:
- #7376
- #7395
- #7394
- #7400

## Current state

|| Fabric (iOS) | Fabric (Android) | Paper (iOS) | Paper (Android) | Web | MacOS | TVOS (iOS) | TVOS (android) |
|-|-|-|-|-|-|-|-|-|
| Builds           | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | 
| Runs             | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |

### DIFF
--- a/apps/web-example/package.json
+++ b/apps/web-example/package.json
@@ -26,6 +26,7 @@
   "devDependencies": {
     "@babel/core": "^7.20.0",
     "@expo/metro-runtime": "~4.0.0",
+    "@react-native/metro-config": "0.77.1",
     "@stylexjs/babel-plugin": "^0.10.0",
     "@types/eslint": "^8",
     "eslint": "^8.57.0",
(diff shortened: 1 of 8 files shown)


apps/web-example/package.json
```diff
@@ -26,6 +26,7 @@
   "devDependencies": {
     "@babel/core": "^7.20.0",
     "@expo/metro-runtime": "~4.0.0",
+    "@react-native/metro-config": "0.77.1",
```
MatiPl01, line 29: I added this as a temporary workaround to fix the following error displayed after running the app:

Screenshot 2025-04-16 at 01 31 22

For some reason, we cannot use metro config from RN 0.79, which crashes with the same error.

The issue started to happen after merging the #7350 PR to the `3.17-stable` branch and is likely caused by the fact that the version of the `@react-native/metro-config` in the tvos example app has changed. It seems that the web example somehow relied on the previous version used by the tvos example app.

We would have to investigate the issue in the future. For now, this is fine, because the web example works as expected.

PR #7401, opened 2025-04-16
docs: Update compatibility table in docs, add 3.17.4

MatiPl01:
## Summary

This PR adds missing `3.17.3` version and the new `3.17.4` version to compatibility tables in docs.

I decided to mark only `3.17.4` as supporting RN 0.79 because of the issue with some layout animations using rotation values without units which stopped working on RN 0.79. Here is a PR with the fix: #7400

## Example recordings

https://github.com/user-attachments/assets/8d83fccf-3413-4384-95e6-f95ec375c3d3

https://github.com/user-attachments/assets/96893a75-3f2d-4dff-b314-8f6acb02553d

### DIFF
--- a/packages/docs-reanimated/docs/guides/compatibility.mdx
+++ b/packages/docs-reanimated/docs/guides/compatibility.mdx
@@ -13,16 +13,17 @@ Reanimated 4 works only with the React Native New Architecture. If your app stil
 
 <div className="compatibility">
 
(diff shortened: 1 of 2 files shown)


patrycjakalinska (commented):
Should we also update in 4.x documentation? 
image

patrycjakalinska (approved):
nvm - it is updated, so no further comments 😅

PR #7402, opened 2025-04-16
chore: Improve bug report template

MatiPl01:
## Summary

- spelling fixes,
- proper architecture (change the sentence that says that Paper is the default architecture)
- allow selection multiple options for the `device` and the `build type`


tjzel (approved):
Can we also add a question about the host machine? Whether the app was built on Windows, Linux, MacOS machine etc.

PR #7406, opened 2025-04-17
chore: unify processBoxShadow accross Reanimated and CSS

patrycjakalinska:
## Summary

Before we had two processBoxShadow functions - in Reanimated (based on implementation from React Native) and in CSS Animations (our simplified function). For better managment and future maintance we decided to unify those two by leaving our simplified function. I've extracted common functions in `/src/common` folder and workletized them.


packages/react-native-reanimated/src/Colors.ts
```diff
@@ -697,19 +693,6 @@ export function processColorsInProps(props: StyleProps) {
...
-      const propGroupList = props[key] as StyleProps[];
-      for (const propGroup of propGroupList) {
-        const nestedPropertyName =
-          NestedColorProperties[key as keyof typeof NestedColorProperties];
-        if (propGroup[nestedPropertyName] !== undefined) {
-          propGroup[nestedPropertyName] = processColor(
-            propGroup[nestedPropertyName]
-          );
-        }
-      }
```
patrycjakalinska, line 712: it's redundant as we now process color in `processBoxShadow` func

packages/react-native-reanimated/src/common/types.ts
```diff
@@ -0,0 +1,6 @@
+'use strict';
```
MatiPl01, line 1: Same remark to this file. I think we should move all helper types that might be used in other places than CSS animations to the `common` directory as well.

packages/react-native-reanimated/src/common/utils.ts
```diff
@@ -0,0 +1,50 @@
+'use strict';
```
MatiPl01, line 1: Maybe change this file to the `common/utils` directory for consistency and create a similar file structure as in the `css/utils` with separate `guards` and `parsers` files? I would like to do this because the next thing to unify is the transform origin string processing which will require applying similar changes to this PR and it would be easier when we start with more modular file structure.

packages/react-native-reanimated/src/css/platform/native/style/processors/__tests__/shadows.test.ts
```diff
@@ -2,8 +2,8 @@
 import type { BoxShadowValue } from 'react-native';
 
 import { processColor } from '../../../../../../Colors';
-import type { ProcessedBoxShadowValue } from '../shadows';
-import { processBoxShadow } from '../shadows';
+import { processBoxShadow } from '../../../../../../common/processors';
+import type { ProcessedBoxShadowValue } from '../../../../../../common/processors/shadows';
```
MatiPl01, line 6:
```suggestion
import { processBoxShadow, type ProcessedBoxShadowValue } from '../../../../../../common';
```

Please add the `index.ts` file in the `src/common` directory to export everything that should be used outside the `src/common` directory instead of importing things directly from files nested within `src/common`.

packages/react-native-reanimated/src/css/platform/native/style/processors/colors.ts
```diff
@@ -2,9 +2,8 @@
 import type { ColorValue } from 'react-native';
 
 import { processColor as processColorInternal } from '../../../../../Colors';
+import type { Maybe, ValueProcessor } from '../../../../../common/types';
 import { ReanimatedError } from '../../../../errors';
```
MatiPl01, line 6:
```suggestion
import type { Maybe, ValueProcessor } from '../../../../../common';
import { ReanimatedError } from '../../../../errors';
```

Same here. I would also prefer to move the `ReanimatedError` to the `common` directory (similar to all other common things) but maybe don't do it now.

packages/react-native-reanimated/src/css/platform/native/style/processors/font.ts
```diff
@@ -1,7 +1,7 @@
 'use strict';
+import type { ValueProcessor } from '../../../../../common/types';
```
MatiPl01, line 2:
```suggestion
import type { ValueProcessor } from '../../../../../common';
```

packages/react-native-reanimated/src/css/platform/native/style/processors/index.ts
```diff
@@ -1,8 +1,8 @@
 'use strict';
+export { processBoxShadow } from '../../../../../common/processors';
```
MatiPl01, line 2:
```suggestion
export { processBoxShadow } from '../../../../../common';
```

packages/react-native-reanimated/src/css/platform/native/style/processors/insets.ts
```diff
@@ -1,7 +1,7 @@
 'use strict';
 import type { DimensionValue } from 'react-native';
 
-import type { ValueProcessor } from '../types';
+import type { ValueProcessor } from '../../../../../common/types';
```
MatiPl01, line 4:
```suggestion
import type { ValueProcessor } from '../../../../../common';
```

packages/react-native-reanimated/src/css/platform/native/style/processors/others.ts
```diff
@@ -1,6 +1,6 @@
 'use strict';
+import type { ValueProcessor } from '../../../../../common/types';
```
MatiPl01, line 2:
```suggestion
import type { ValueProcessor } from '../../../../../common';
```

packages/react-native-reanimated/src/css/platform/native/style/processors/transform.ts
```diff
@@ -1,4 +1,5 @@
 'use strict';
+import type { ValueProcessor } from '../../../../../common/types';
```
MatiPl01, line 2:
```suggestion
import type { ValueProcessor } from '../../../../../common';
```

packages/react-native-reanimated/src/css/platform/native/style/processors/transform.ts
```diff
@@ -161,7 +161,7 @@ function parseMatrix(values: (number | string)[]): TransformsArray {
 }
 
 export const processTransform: ValueProcessor<TransformsArray | string> = (
-  value
+  value: string | TransformsArray
```
MatiPl01, line 164:
```suggestion
  value
```

I think you don't need this type as the `ValueProcessor` type already adds this type to the function argument.

packages/react-native-reanimated/src/common/types.ts
```diff
@@ -0,0 +1,6 @@
...
+export type Maybe<T> = T | null | undefined;
+
+export type ValueProcessor<V, R = V> = (
+  value: V
+) => Maybe<R> | Record<string, R>;
```
MatiPl01, line 6:
```suggestion
) => Maybe<R>;
```

This `Record<string, R>` type is used only in CSS animations where we split property into a few separate props, e.g. `marginHorizontal` to `marginLeft` and `marginRight`.

packages/react-native-reanimated/src/css/platform/native/style/processors/transformOrigin.ts
```diff
@@ -1,7 +1,7 @@
 'use strict';
+import type { ValueProcessor } from '../../../../../common/types';
```
MatiPl01, line 2:
```suggestion
import type { ValueProcessor } from '../../../../../common';
```

packages/react-native-reanimated/src/css/platform/native/style/types.ts
```diff
@@ -1,9 +1,6 @@
 'use strict';
-import type { AnyRecord, ConfigPropertyAlias, Maybe } from '../../../types';
-
-export type ValueProcessor<V, R = V> = (
```
MatiPl01, line 4: Maybe keep CSS `ValueProcessor` type separate as it can also return a record type, whilst the value processor implemented for worklet based animations can't.

packages/react-native-reanimated/src/css/platform/native/config.ts
```diff
@@ -1,10 +1,10 @@
 'use strict';
+import { processBoxShadow } from '../../../common/processors';
```
MatiPl01, line 2:
```suggestion
import { processBoxShadow } from '../../../common';
```

packages/react-native-reanimated/src/css/platform/web/style/processors/shadows.ts
```diff
@@ -1,7 +1,7 @@
 'use strict';
 import type { BoxShadowValue, ViewStyle } from 'react-native';
 
-import { parseBoxShadowString } from '../../../../utils';
+import { parseBoxShadowString } from '../../../../../common/utils';
```
MatiPl01, line 4:
```suggestion
import { parseBoxShadowString } from '../../../../../common';
```

packages/react-native-reanimated/src/css/platform/web/style/types.ts
```diff
@@ -1,5 +1,6 @@
 'use strict';
-import type { AnyRecord, Maybe } from '../../../types';
+import type { Maybe } from '../../../../common/types';
```
MatiPl01, line 2:
```suggestion
import type { Maybe } from '../../../../common';
```

packages/react-native-reanimated/src/css/utils/__tests__/parsers.test.ts
```diff
@@ -1,5 +1,5 @@
 'use strict';
-import { parseBoxShadowString } from '../parsers';
+import { parseBoxShadowString } from '../../../common/utils';
```
MatiPl01, line 2:
```suggestion
import { parseBoxShadowString } from '../../../common';
```

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
```diff
@@ -212,11 +212,11 @@ function styleUpdater(
...
   let frameTimestamp: number | undefined;
   let hasNonAnimatedValues = false;
-  if (!SHOULD_BE_USE_WEB && newValues.boxShadow) {
-    processBoxShadow(newValues);
-  }
   for (const key in newValues) {
     const value = newValues[key];
+    if (key === 'boxShadow' && !SHOULD_BE_USE_WEB) {
+      newValues[key] = processBoxShadow(value);
+    }
```
MatiPl01, line 219: Can you move it to the `else` block? `boxShadow` cannot be used with animations (like `withTimng`, etc.) directly, so `isAnimated(value)` will always return false for this prop. This is just a small optimization.

Please also leave a comment to improve it in the future. Ideally, I would like to create something similar to `src/css/platform/native/config.ts` but a bit simplified, containing only properties that need to be processed and their processors.

MatiPl01 (changes requested):

MatiPl01 (changes requested):
Looks and works fine. I just left a bunch of comments some of which don't have to be addressed now.

packages/react-native-reanimated/src/common/constants.ts
```diff
@@ -0,0 +1,4 @@
+'use strict';
+import { Platform } from 'react-native';
+
+export const IS_ANDROID = Platform.OS === 'android';
```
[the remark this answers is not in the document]
patrycjakalinska: Yes, I was thinking about it - file structure present here was an idea how we can arrange our funcs and other stuff, after this PR my goal was to unfiy `processTranform` and adapt the file structure to what is it there and after that as you said we can move the constants, helpers etc. and use them through reanimated c:

packages/react-native-reanimated/src/common/types.ts
patrycjakalinska: Yet, as this is a common type, we won't need it to handle css?

packages/react-native-reanimated/src/css/platform/native/style/types.ts
patrycjakalinska: I was thinking maybe let this type stay at common with a explaination comment about return types, as it does no harm, and will be better to have it in one place, wdyt?

MatiPl01 (commented):
Added a few more comments.

packages/react-native-reanimated/src/common/processors/shadows.ts
```diff
@@ -1,15 +1,16 @@
...
+import { processColor } from '../../Colors';
+import { ReanimatedError } from '../../errors';
+import { IS_ANDROID } from '../constants';
 import type { ValueProcessor } from '../types';
+import { parseBoxShadowString } from '../utils';
 
 const ERROR_MESSAGES = {
   notArrayObject: (value: object) =>
-    `Box shadow value must be an array of shadow objects or a string. Received: ${JSON.stringify(value)}`,
+    `Box shadow value must be an array of shadow objects. Received: ${JSON.stringify(value)}`,
```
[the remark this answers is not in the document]
patrycjakalinska: I think it was removed by mistake when moving to common folder - will change

MatiPl01 (approved):
Good job! Thanks for addressing all comments.

Issue #7367, opened 2025-04-09
Exception in HostFunction: std::exception


VickyA371:
<img width="463" alt="Image" src="https://github.com/user-attachments/assets/ee019b4b-f559-448c-ab95-9993b5973a0a" />

### Description

Recently, I've upgraded my react native version from 0.73.6 to 0.78.1.

I'm using latest version react-native-reanimated.
I've tried to downgrade this to 3.17.1 as per compatibility table.

Although, I am facing same issue.

Due to this, I'm not able to access other components registered with AppRegistry.

This is only happening when using react-native-reanimated otherwise app is working fine and can load other components registered with AppRegistry.registerComponent function.

### Steps to reproduce

Upgrade react native version from 0.73.6 to 0.78.1. 

### Snack or a link to a repository

https://github.com/VickyA371/ve_sdk_custom_gallery

### Reanimated version

3.17.3

### React Native version

0.78.1

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Device model

iPhone 16 Pro

### Acknowledgements

Yes

WoLewicki:
Hey @VickyA371, I think I found the root of your issue. In `AppDelegate.swift`, you https://github.com/VickyA371/ve_sdk_custom_gallery/blob/257907c9c4c8747295a2be87162d463271e8a061/ios/ve_sdk_custom_gallery/AppDelegate.swift#L22 you are creating `RCTBridge` instance, and `react-native-reanimated` only supports `bridgeless` mode: https://docs.swmansion.com/react-native-reanimated/docs/guides/compatibility#supported-react-native-versions-on-the-new-architecture-fabric. After removing those lines, the app works for me. Does it solve your issue?

WoLewicki:
Closing since I believe it is fixed. Feel free to comment here if something is wrong and we can always reopen it then.

iOSAsaFinancial:
@WoLewicki I have the same issue, but the point is that I cannot use RCTAppDelegate approach. Because I need to provide native iOS SDK. And inside SDK I don't have access to subclass AppDelegate from RCTAppDelegate. So in my case I use RCTRootView in order to init react-native view. But this approach uses Bridge anyway. My react-native version is 0.77.2 and reanimated 3.17.5. Could you please provide any other solution?

WoLewicki:
[...]

PR #7407, opened 2025-04-17
feat: add check for bridgeless in both packages

WoLewicki:
## Summary

There are some issues like [this](https://github.com/software-mansion/react-native-reanimated/issues/7367) where it is not obvious what was the reason of exception. It turned out to be having `bridge` which both `Reanimated` and `Worklets` do not accept. For easier problem solving, I suggest adding an explicit check for this scenario.

I also added compliance to JSI spec for `Worklets` module to be in line with `Reanimated`.

## Test plan

Run the linked repro or manually enable bridge mode e.g. [here](https://github.com/facebook/react-native/blob/7ff5db3c462c369a8210674ab53642c3f34e9c37/packages/react-native/Libraries/AppDelegate/RCTDefaultReactNativeFactoryDelegate.mm#L121) to see that it does not work.

### DIFF
--- a/packages/react-native-reanimated/apple/reanimated/apple/ReanimatedModule.mm
+++ b/packages/react-native-reanimated/apple/reanimated/apple/ReanimatedModule.mm
@@ -132,6 +132,12 @@ RCT_EXPORT_MODULE(ReanimatedModule);
   return ![_moduleRegistry moduleIsInitialized:WorkletsModule.class];
 }
 
+- (void)checkBridgeless
+{
+  auto isBridgeless = ![self.bridge isKindOfClass:[RCTCxxBridge class]];
+  react_native_assert(isBridgeless && "[Reanimated] react-native-reanimated only supports bridgeless mode");
+}
+
 RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(installTurboModule)
 {
   REAAssertJavaScriptQueue();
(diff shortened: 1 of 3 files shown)


packages/react-native-worklets/apple/worklets/apple/WorkletsModule.mm
```diff
@@ -30,16 +31,24 @@ @implementation WorkletsModule {
   return workletsModuleProxy_;
 }
 
+- (void)checkBridgeless
+{
+  auto isBridgeless = ![self.bridge isKindOfClass:[RCTCxxBridge class]];
+  react_native_assert(isBridgeless && "[Worklets] react-native-worklets only supports bridgeless mode");
```
tjzel, line 37: Shouldn't we wrap it in `#ifndef NDEBUG`?

packages/react-native-reanimated/apple/reanimated/apple/ReanimatedModule.mm
```diff
@@ -131,6 +132,12 @@ - (BOOL)hasReactNativeFailedReload
   return ![_moduleRegistry moduleIsInitialized:WorkletsModule.class];
 }
 
+- (void)checkBridgeless
+{
+  auto isBridgeless = ![self.bridge isKindOfClass:[RCTCxxBridge class]];
+  react_native_assert(isBridgeless && "[Reanimated] react-native-reanimated only supports bridgeless mode");
```
tjzel, line 138: Shouldn't we wrap it in `#ifndef NDEBUG`?

packages/react-native-worklets/apple/worklets/apple/WorkletsModule.mm
WoLewicki: Why so? It will crash in production all the same in one of the next lines so maybe better to have it fail earlier there too?

tjzel: Makes sense, especially that it's used only in `get` and `install` methods.

tjzel (approved):

PR #7413, opened 2025-04-18
[0.80] Fix "error: method resolve in class LengthPercentage cannot be applied to given types"

tomekzaw:
## Summary

This PR fixes the following error while trying to build react-native-reanimated with react-native 0.80 for Android by removing `BorderRadiiDrawableUtils` and `ReactNativeUtils` as they are unused:

```
reanimated/BorderRadiiDrawableUtils.java:21: error: method resolve in class LengthPercentage cannot be applied to given types;
> Task :react-native-reanimated:compileDebugJavaWithJavac FAILED
    return length.resolve(bounds.width(), bounds.height()).toPixelFromDIP().getHorizontal();
                 ^
  required: float
  found:    int,int
  reason: actual and formal argument lists differ in length
```

Note that building with RN 0.80 for Android also requires https://github.com/software-mansion/react-native-reanimated/pull/7409.


tjzel:
Is it backwards compatible with RN < 0.80?

tomekzaw:
Nope, that's why it's still marked as a draft. I guess we'll need sourcesets (or reflection).

tjzel:
We were source setting this file before, we can source set it again.

piaskowyk (approved):

tomekzaw:
Turns out, we can just remove these two files as they are unused.

For Reanimated v3, we'll need to add another `reactNativeVersionPatch`.

PR #7389, opened 2025-04-14
Bump example apps to React Native 0.79.0

tomekzaw:
## Summary

This PR bumps `fabric-example` as well as other apps to `react-native@0.79.0`.

PR #7415, opened 2025-04-18
Bump React Native to 0.79.1

tomekzaw:
## Summary

This PR bumps `react-native` from `0.79.0` to `0.79.1` in example apps as well as in `devDependencies`.


apps/tvos-example/Gemfile.lock
```diff
@@ -120,4 +117,4 @@ RUBY VERSION
    ruby 3.3.6p108
 
 BUNDLED WITH
-   2.6.3
+   2.4.22
```
MatiPl01, line 120: Please don't change bundler version in the `Gemfile.lock`. If you have different version of bundler installed locally, please update it to `2.6.3` so that all of us use the same one.

Changing this version triggers regeneration of `Podfile.lock` and it is very annoying that every installation of pods ends with new `Podfile.lock` files being generated.

Please also restore the previous bundler version that you accidentally changed in the `apps/fabric-example/Gemfile.lock` file in the previous upgrade [PR](https://github.com/software-mansion/react-native-reanimated/pull/7389/files).

MatiPl01 (changes requested):
Requested small changes

apps/tvos-example/Gemfile.lock
tomekzaw: Done

MatiPl01 (approved):
Looks fine for me. Let's wait for @tjzel to confirm which version of ruby we want to use.

apps/tvos-example/Gemfile.lock
tjzel: @tomekzaw it seems that you need to setup rbenv ♦️

PR #7417, opened 2025-04-18
feat: Inject reanimated module proxy to the contextContainer

MatiPl01:
## Summary

This PR adds a possibility to access the `ReanimatedModuleProxy` from the `ReanimatedShadowNode`. It also fixes android linking issue to make it possible to access all reanimated library stuff.

## Test plan

You can open any example in the example app and see that it renders fine. 

You can also uncomment the following line in the `packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h`:

```cpp
// LOG(INFO) << "We can access the proxy: " << proxy->getCssTimestamp();
```

and see that the proxy can be accessed.

## TODO

- [ ] Fix gesture handler integration - in the current implementation gestures aren't attached to AnimatedComponents because the `ReanimatedView` is not laid out (because of `display: contents`).


packages/react-native-reanimated/src/css/component/AnimatedComponent.tsx
```diff
@@ -47,6 +50,7 @@ export default class AnimatedComponent<
 
   constructor(ChildComponent: AnyComponent, props: P) {
     super(props);
+    ReanimatedModule.initialize();
```
MatiPl01, line 53: We can think of a better solution than this in the future. I needed to add this to ensure that the `ReanimatedModule` instance is created (and our turbo module is initialized) before the `ReanimatedView`, which uses the injected reanimated module proxy, is rendered.

packages/react-native-reanimated/Common/NativeView/CMakeLists.txt
```diff
@@ -32,14 +37,19 @@ target_include_directories(
...
+target_link_libraries(react_codegen_rnreanimated log ReactAndroid::jsi fbjni jsi)
+
+target_link_libraries(react_codegen_rnreanimated react-native-reanimated::reanimated)
+target_link_libraries(react_codegen_rnreanimated react-native-worklets::worklets)
+
+if(ReactAndroid_VERSION_MINOR GREATER_EQUAL 76)
+  target_link_libraries(react_codegen_rnreanimated ReactAndroid::reactnative)
+else()
+  target_link_libraries(react_codegen_rnreanimated ReactAndroid::react_nativemodule_core)
+endif()
```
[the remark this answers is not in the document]
MatiPl01: Yeah, I forgot that we don't support old versions of RN on main

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h
```diff
@@ -1,13 +1,49 @@
...
+      return;
+    }
+
+    const auto proxy = weakPtr.lock();
+
+    // LOG(INFO) << "We can access the proxy: " << proxy->getCssTimestamp();
+  }
+
+ private:
+  std::optional<std::weak_ptr<ReanimatedModuleProxy>> reanimatedModuleProxy_;
```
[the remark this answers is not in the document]
MatiPl01: `find` method on the `contextContainer` returns optional type. I cannot use the `at` method as it crashes on reloads when there is no `ReanimatedModuleProxy` in the `contextContainer`, thus I decided to use `find` with the optional type.

I am not sure if I can do something better here. I tried returning a `nullptr` and using just `std::weak_ptr` type but I could not use `nullptr` for the `std::weak_ptr`.

Do you have an idea how to better handle this case?

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h
```diff
@@ -1,13 +1,49 @@
...
+  void adopt(ShadowNode &shadowNode) const override {
+    if (!reanimatedModuleProxy_.has_value()) {
+      return;
+    }
+    const auto &weakPtr = reanimatedModuleProxy_.value();
+    if (weakPtr.expired()) {
+      return;
+    }
+
+    const auto proxy = weakPtr.lock();
```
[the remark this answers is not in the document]
MatiPl01: Something like this should work, right?

```cpp
const auto &proxy = reanimatedModuleProxy_.value().lock();
if (!proxy) {
  return;
}
```

packages/react-native-reanimated/src/css/component/AnimatedComponent.tsx
MatiPl01: I added `initializeReanimatedModule` similar to `initializeUIRuntime` in the worklets module.

packages/react-native-reanimated/Common/NativeView/CMakeLists.txt
```diff
@@ -32,14 +37,15 @@ target_include_directories(
...
-  fbjni
-  jsi
-  reactnative
-)
+target_link_libraries(react_codegen_rnreanimated log ReactAndroid::jsi fbjni jsi)
+
+target_link_libraries(react_codegen_rnreanimated react-native-reanimated::reanimated)
+target_link_libraries(react_codegen_rnreanimated react-native-worklets::worklets)
+
+target_link_libraries(react_codegen_rnreanimated ReactAndroid::reactnative)
```
piaskowyk, line 48: I think you can merge it into one invocation:
```cmake
target_link_libraries(
  react_codegen_rnreanimated 
  log 
  ReactAndroid::jsi 
  fbjni 
  jsi
  react-native-reanimated::reanimated
  react-native-worklets::worklets
  ReactAndroid::reactnative
)
```

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h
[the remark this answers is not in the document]
piaskowyk: For me this optional make sense 👍

piaskowyk (approved):
Looks good 👍

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h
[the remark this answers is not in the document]
piaskowyk: or something like that:
```cpp
if (const auto &proxy = reanimatedModuleProxy_.value().lock()) {
  // use proxy here
}
```

packages/react-native-reanimated/Common/NativeView/CMakeLists.txt
MatiPl01: yup, I just wanted to make it consistent with remaining CMakeLists.txt files but it looks better when all `target_link_libraries` calls are grouped into a single one. I will change that.

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h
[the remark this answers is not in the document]
MatiPl01: I just wanted to avoid nesting in case we want to add more code to the `adopt` method. I think that I will keep my implementation and we could decide if we want to change it in the future.

packages/react-native-reanimated/Common/NativeView/CMakeLists.txt
```diff
@@ -32,13 +37,18 @@ target_include_directories(
   ${CODEGEN_OVERWRITTEN_COMPONENTS_DIR}
   ${CODEGEN_JNI_DIR}
   ${CODEGEN_COMPONENTS_DIR}
+  ${REACT_COMMON_DIR}
 )
 
 target_link_libraries(
   react_codegen_rnreanimated
+  log
```
WoLewicki, line 45: Do we need `log` here? I haven't seen it used anywhere.

tjzel: We use `log` for debugging

WoLewicki: Yeah I get you, but is it needed for the PR? Prolly doesn't matter, just asking.

MatiPl01: It isn't needed yet but I am sure I will be using it for debugging so I would like to leave it here.

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h
```diff
@@ -1,13 +1,47 @@
...
+  void adopt(ShadowNode &shadowNode) const override {
+    if (!reanimatedModuleProxy_.has_value()) {
+      return;
+    }
+    const auto &proxy = reanimatedModuleProxy_.value().lock();
+    if (!proxy) {
+      return;
+    }
+
+    // LOG(INFO) << "We can access the proxy: " << proxy->getCssTimestamp();
```
tjzel, line 33:
```suggestion
```

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -927,6 +927,11 @@ void ReanimatedModuleProxy::initializeFabric(
       uiManager_, updatesRegistryManager_, request);
   commitHook_ = std::make_shared<ReanimatedCommitHook>(
       uiManager_, updatesRegistryManager_, layoutAnimationsProxy_);
+
+  const auto scheduler =
+      reinterpret_cast<Scheduler *>(uiManager_->getDelegate());
```
tjzel, line 932: Why `reinterpret_cast` and not `static_cast` or some other safer cast?

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h
tjzel: Don't forget to remove that! 🚀

packages/react-native-reanimated/src/initializers.ts
```diff
@@ -0,0 +1,16 @@
+'use strict';
+import { isWeb } from './PlatformChecker';
+import type { IReanimatedModule } from './ReanimatedModule';
+
+export function initializeReanimatedModule(
+  ReanimatedModule: IReanimatedModule
+) {
+  if (isWeb()) {
```
tjzel, line 8: Is `isWeb()` necessary here? Or the contrary, should we use `shouldBeUseWeb`?

packages/react-native-reanimated/src/index.ts
```diff
@@ -3,6 +3,12 @@
 import './publicGlobals';
 
 import * as Animated from './Animated';
+import { initializeReanimatedModule } from './initializers';
+import { ReanimatedModule } from './ReanimatedModule';
+
+// TODO: Specify the initialization pipeline since now there's no
+// universal source of truth for it.
+initializeReanimatedModule(ReanimatedModule);
```
tjzel, line 11: Have you checked if Worklets still load first in this configuration? (console.log would be enough)

tjzel (approved):

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h
MatiPl01: yeah, it is just temporary. I will make many changes here in the upcoming PRs so I will remove this line.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
MatiPl01: I just copied code from `void ReanimatedModuleProxy::initializeLayoutAnimationsProxy(` where we have exactly the same line of code written by @bartlomiejbloniarz. I don't know if there is any reason why we need to use `reinterpret_cast` here but it wanted to be consistent.

packages/react-native-reanimated/src/initializers.ts
MatiPl01: I implemented this function based on the `initializeUIRuntime` implementation which performs the same check. It is your implementation so it's rather a question to you why do we need to use the `isWeb` or other check?

I can see that the `ReanimatedModule` will be always initialized, no matter which platform we use, because of this:

```ts
export const ReanimatedModule = shouldBeUseWeb()
  ? createJSReanimatedModule()
  : createNativeReanimatedModule();
```

Both functions always return an instance of the class (either `JSReanimated` or `NativeReanimatedModule`) so `ReanimatedModule` will never be falsy.

I would prefer to just leave the following implementation:

```ts
export function initializeReanimatedModule(
  ReanimatedModule: IReanimatedModule
) {
  if (!ReanimatedModule) {
    throw new ReanimatedError(
      'Tried to initialize Reanimated without a valid ReanimatedModule'
    );
  }
}
```

which will never throw this error but we somehow have to reference the `ReanimatedModule` to make sure that it gets initialized before the `AnimatedComponent` instance is created, so I think it would be fine to leave just this `if (!ReanimatedModule)` check.

piaskowyk (approved):
Looks good to me, but let's wait with merging until I figure out how to solve the problem with Gesture Handler

PR #7419, opened 2025-04-22
chore: unify processTransformOrigin accross Reanimated and CSS

patrycjakalinska:
## Summary

Further process of unification, now based on previous PR about `processBoxShadow` we move `processTransformOrigin` to common dir.


packages/react-native-reanimated/src/common/processors/transformOrigin.ts
```diff
@@ -1,7 +1,10 @@
 'use strict';
-import type { ValueProcessor } from '../../../../../common';
-import { ReanimatedError } from '../../../../errors';
-import type { NormalizedTransformOrigin, TransformOrigin } from '../../types';
+import { ReanimatedError } from '../../css/errors';
```
MatiPl01, line 2:
```suggestion
'worklet';
import { ReanimatedError } from '../../errors';
```

Add missing worklet directive.

You can use the error implementation from the `packages/react-native-reanimated/src/errors.ts` file instead of CSS ones. They are duplicated because we aimed to create separate packages for CSS and worklet-based reanimated parts but I think the `ReanimatedError` class should be moved to the common package as well and we could get rid of having 2 copies.

MatiPl01 (changes requested):
These comments are referring my code, but since you are changing this file, maybe you could change these things as well.

### 1. Validation

I think that we can make the validation of the transform origin in the `processTransformOrigin` function dev only and instead of:

```ts
return validateResult(result, components, value)
```

do this

```ts
if (__DEV__) {
  validateResult(result, components, value)
}

return result as NormalizedTransformOrigin;
```

and change the `validateResult` to a `void` function that performs only validations without returning anything.

### 2. Accepted z values

In the current implementation the `100px` can be passed as the z (third) axis value, which is not valid. You could add one more validation check (I think it must be separate from `validateResult` as the result of `parseComponent` for value with `px` is a number).

packages/react-native-reanimated/src/common/processors/transformOrigin.ts
```diff
@@ -107,11 +111,20 @@ export const processTransformOrigin: ValueProcessor<TransformOrigin> = (
     [components[0], components[1]] = [components[1], components[0]];
   }
 
+  // Check if the z component is a valid value with px unit
+  if (typeof components[2] === 'string' && !isLength(components[2])) {
+    throw new ReanimatedError(ERROR_MESSAGES.invalidZValue(components[2]));
+  }
+
```
patrycjakalinska, line 118: As of RN documentation the `z` value must be a `px` - I've added guard that makes sure it is correct.

Documentation is available [here](https://reactnative.dev/docs/transforms#three-value-syntax).

MatiPl01 (approved):
LGTM

PR #7420, opened 2025-04-22
feat(Reanimated): react-native-worklets compatibility assertions

tjzel:
## Summary

Extra QoL assertions for the user if they had installed the wrong version of `react-native-worklets` with Reanimated 4.

## Test plan

Try downgrading the version of `react-native-worklets` in its `package.json` and see that relevant error is thrown.

### DIFF
--- a/packages/eslint-plugin-reanimated/package.json
+++ b/packages/eslint-plugin-reanimated/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-reanimated",
-  "version": "1.0.0",
+  "version": "0.0.0",
   "private": true,
   "license": "MIT",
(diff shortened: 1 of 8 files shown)


packages/eslint-plugin-reanimated/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-reanimated",
-  "version": "1.0.0",
```
tjzel, line 3: Changed the version of all private packages to 0.0.0.

MatiPl01 (approved):
Very useful change!
The error indeed shows up when incorrect version of worklets is used.

PR #7424, opened 2025-04-23
chore: Process transform origin refactor

MatiPl01:
## Summary

This PR refactors `processTransformOrigin` function implementation to allow the same values and throw more similar errors to the RN implementation. The resulting function is even simpler and behavior is more similar to the one from RN than before.

I also added some more test cases and adjusted existing tests to match the new implementation.

## Test plan

Open the `apps/common-app/src/apps/reanimated/examples/TransformOriginExample.tsx` example and see that everything works.


packages/react-native-reanimated/src/common/processors/__tests__/processBoxShadow.test.ts
```diff
@@ -1,9 +1,9 @@
 'use strict';
```
MatiPl01, line 1: I just moved this test file to the new location of the function implementation.

packages/react-native-reanimated/src/common/processors/transformOrigin.ts
```diff
@@ -111,11 +110,15 @@ export const processTransformOrigin: ValueProcessor<TransformOrigin> = (
...
   }
 
-  const result = [
-    parseComponent(components[0] ?? '50%', HORIZONTAL_CONVERSIONS),
-    parseComponent(components[1] ?? '50%', VERTICAL_CONVERSIONS),
-    parseComponent(components[2] ?? 0),
+  const parsed = [
+    parse(components[0] ?? '50%', HORIZONTAL_CONVERSIONS),
+    parse(components[1] ?? '50%', VERTICAL_CONVERSIONS),
+    parse(components[2] ?? 0),
```
patrycjakalinska, line 116: Shouldn't we also check if `z` value is with `px`? I'm not sure what to do with that as RN documentation states:
> The third value (z-offset) must be a px. It always represents the Z offset.

Yet in ReactNative's `processTransformOrigin` they do not check such thing directly, rather if you use `z` without the `px` it is getting ommited and following error occurs:

>  'Transform origin must have exactly 3 values.'

packages/react-native-reanimated/src/common/processors/transformOrigin.ts
```diff
@@ -30,77 +32,74 @@ const VERTICAL_CONVERSIONS = {
...
-    return component;
+  }
+  if (typeof component === 'string' && component.endsWith('%')) {
+    // Convert 0% to 0 (for consistency)
+    return parseFloat(component) === 0 ? 0 : component;
   }
 
-  const num = parseFloat(component);
+  // Try to convert to number or fallback to the original value
+  const num = +component;
```
patrycjakalinska, line 45: in our repo we mostly use `Number()` instead of `+` - maybe we can stay with `Number()`? What do you think?

packages/react-native-reanimated/src/common/processors/transformOrigin.ts
```diff
@@ -30,77 +32,74 @@ const VERTICAL_CONVERSIONS = {
...
-    throw new ReanimatedError(ERROR_MESSAGES.invalidTransformOrigin(value));
+  let components: (string | number)[];
+  let parse: ComponentParser;
+
+  if (Array.isArray(value)) {
+    components = value;
+    parse = parseComponent;
+  } else {
+    components = value.split(/\s+/);
+    parse = parseComponentWithPx;
```
patrycjakalinska, line 102: If it's `left bottom` we'll use `parseComponentWithPx` even though there is no `px` - the naming is kinda confusing, maybe smth like `parseComponentWithUnits`? (As we also check for `%` there)

patrycjakalinska (commented):
Left some comments for explaination c:

packages/react-native-reanimated/src/common/processors/transformOrigin.ts
MatiPl01: I'm just used to to `+` conversions but I will change that for consistency

packages/react-native-reanimated/src/common/processors/transformOrigin.ts
MatiPl01: I was thinking of using `parseComponentWithUnits` but it is also confusing as the `parseComponent` (used for array transform origin type) also accepts `%` values. I will think of better names for these 2 functions.

MatiPl01:
I am closing this PR as it will be easier to start working from scratch on these changes.

PR #7431, opened 2025-04-24
chore(ci): Set `react-native-worklets` dependency version for nightly builds.

arthwood:
## Summary

Nightly build will set the `react-native-worklets` peer dependency version to its nightly value.

## Test plan
Can be tested by running `set-animated-version` script with `--nightly` option.

### DIFF
--- a/scripts/releasing.js
+++ b/scripts/releasing.js
@@ -1,70 +1,32 @@
 const { cat, exec } = require('shelljs');
 
 /**
+ * @typedef {object} Arg
+ * @property {string} value - argument value
+ * @property {boolean} handled - is argument known
+ */
+
+/**
+ * @typedef {object} Flags
+ * @property {boolean} help - whether help option was passed
+ * @property {boolean} nightly - whether nightly option was passed
+ * @property {boolean} fresh - whether fresh option was passed
+ * @property {boolean} custom - whether custom version was passed
+ */
+
+/**
  * @param {string[]} rawArgs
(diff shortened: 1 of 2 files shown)


scripts/releasing.js
tjzel, line 1: Could you sort the functions in an anti-topological order? (entry points first, helper functions last)

tjzel (commented):
Very nice job, the script is so clean right now ✨ 

The thing we should address is that now previous `peerDependency` of `react-native-worklets` won't be restored. Currently the version is restored by a cheeky `console.log` back to the shell. Do you have an idea how to do this in an elegant way? Maybe we should move the main script `createNPMPackage.sh` to Node as well.

scripts/releasing.js
arthwood: Np, done is second commit.

arthwood:
> Very nice job, the script is so clean right now ✨
> 
> The thing we should address is that now previous `peerDependency` of `react-native-worklets` won't be restored. Currently the version is restored by a cheeky `console.log` back to the shell. Do you have an idea how to do this in an elegant way? Maybe we should move the main script `createNPMPackage.sh` to Node as well.

Maybe we can just `git checkout <file>`?

tjzel:
Checking out `package.json` might remove some local changes. After second thought it's very rare that we build a nightly version locally, we can leave it as is.

tjzel (approved):

PR #7432, opened 2025-04-24
Draft: Web Worker

skusnierz:


apps/common-app/src/apps/reanimated/examples/index.ts
```diff
@@ -702,4 +703,8 @@ export const EXAMPLES: Record<string, Example> = {
     title: '[LA] View Recycling',
     screen: ViewRecyclingExample,
   },
+  WebWorkerExample: {
```
tomekzaw, line 706: Let's add `icon` prop and move example to the top.

packages/react-native-reanimated/react-native.config.js
tomekzaw, line 1: Can we introduce these changes in a separate PR if they are still relevant?

packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
```diff
@@ -120,6 +120,13 @@ jsi::Value WorkletsModuleProxy::createWorkletRuntime(
   return jsi::Object::createFromHostObject(rt, workletRuntime);
 }
 
+void WorkletsModuleProxy::terminateWorkletRuntime(
+    jsi::Runtime &rt,
+    const jsi::Value &workletRuntimeValue) {
+  auto workletRuntime = extractWorkletRuntime(rt, workletRuntimeValue);
```
tomekzaw, line 126: Can we use `const auto &` here?
```suggestion
  const auto &workletRuntime = extractWorkletRuntime(rt, workletRuntimeValue);
```

packages/react-native-worklets/Common/cpp/worklets/Tools/AsyncQueue.cpp
```diff
@@ -49,4 +49,10 @@ void AsyncQueue::push(std::function<void()> &&job) {
   state_->cv.notify_one();
 }
 
+void AsyncQueue::empty() {
```
tomekzaw, line 52: I would expect `AsyncQueue::empty` method to check whether the data structure is empty rather than making it empty. Can we rename it to `terminate` or `shutdown` instead?

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.cpp
```diff
@@ -128,6 +128,12 @@ jsi::Value WorkletRuntime::executeSync(
   return shareableResult->toJSValue(rt);
 }
 
+void WorkletRuntime::terminate() {
+  if (this->queue_) {
+    this->queue_->empty();
+  }
```
tomekzaw, line 134:
```suggestion
  if (!this->queue_) {
    // TODO: explain that queue has not been created so there's no need to call shutdown and all is good
    return;
  }
  this->queue_->shutdown();
```

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntimeCollector.h
```diff
@@ -5,7 +5,6 @@
 #include <jsi/jsi.h>
 
 #include <memory>
-
```
tomekzaw, line 8: No need for this change I guess

PR #7439, opened 2025-04-28
feat: Add a possibility to pass a CSS transition to the native view

MatiPl01:
## Summary

This PR adds a possibility to pass CSS transition config and the static JS style from the JS side to the native `ReanimatedView` component.

It also adds custom constructors to the `ReanimatedShadowNode` that will be used to properly update state (storing `CSSTransition` and more in the future) based on updated JS props.


packages/react-native-reanimated/src/specs/ReanimatedNativeComponent.ts
```diff
@@ -1,8 +1,11 @@
 'use strict';
 import type { ViewProps } from 'react-native';
+import type { UnsafeMixed } from 'react-native/Libraries/Types/CodegenTypes';
 import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 
-interface NativeProps extends ViewProps {}
+interface NativeProps extends ViewProps {
+  cssTransition?: UnsafeMixed;
```
MatiPl01, line 7: codegen is very problematic and doesn't accept types I want to pass. Because of that I decided to use the `UnsafeMixed` type to allow anything.

packages/react-native-reanimated/src/css/component/AnimatedComponent.tsx
```diff
@@ -223,7 +229,21 @@ export default class AnimatedComponent<
       return child;
     }
 
-    return <ReanimatedView style={styles.container}>{child}</ReanimatedView>;
+    // TODO - improve later
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [animationProperties, transitionProperties] =
+      filterCSSAndStyleProperties(StyleSheet.flatten(style) ?? {});
```
MatiPl01, line 235: I will get rid of css managers on the JS side and use just this function in the future. We no longer need managers as all the CSS animation/transition creation and updating logic must be handled on the CPP side.

packages/react-native-reanimated/src/specs/ReanimatedNativeComponent.ts
```diff
@@ -1,8 +1,12 @@
 'use strict';
 import type { ViewProps } from 'react-native';
+import type { UnsafeMixed } from 'react-native/Libraries/Types/CodegenTypes';
 import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 
-interface NativeProps extends ViewProps {}
+interface NativeProps extends ViewProps {
+  jsStyle: UnsafeMixed;
+  cssTransition: UnsafeMixed;
```
MatiPl01, line 8: codegen is very problematic and doesn't accept types I want to pass. Because of that I decided to use the `UnsafeMixed` type to allow anything and give me the value as the `folly::dynamic` type.

packages/react-native-reanimated/src/css/component/AnimatedComponent.tsx
```diff
@@ -223,7 +229,22 @@ export default class AnimatedComponent<
       return child;
     }
 
-    return <ReanimatedView style={styles.container}>{child}</ReanimatedView>;
+    // TODO - improve later
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [animationProperties, transitionProperties, filteredStyle] =
+      filterCSSAndStyleProperties(StyleSheet.flatten(style) ?? {});
```
MatiPl01, line 235: I will get rid of css managers on the JS side and use just this function in the future. We no longer need managers as all the CSS animation/transition creation and updating logic must be handled on the CPP side.

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ReanimatedShadowNode.h
```diff
@@ -19,7 +19,15 @@ using ReanimatedViewShadowNodeBase = ConcreteViewShadowNode<
...
  public:
-  using ConcreteViewShadowNode::ConcreteViewShadowNode;
+  ReanimatedShadowNode(
+      const ShadowNodeFragment &fragment,
+      const ShadowNodeFamily::Shared &family,
+      ShadowNodeTraits traits);
+
+  ReanimatedShadowNode(
+      const ShadowNode &sourceShadowNode,
+      const ShadowNodeFragment &fragment);
```
piaskowyk, line 29: Could you explain why we need two constructors and what the difference is between them? Specifically, when is the first one called and when is the second one used? It might be worth adding this information as a comment in the class.

piaskowyk (approved):
For me looks good 👍

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ReanimatedShadowNode.h
MatiPl01: The `ConcreteViewShadowNode` declares 2 separate constructors:

```cpp
ConcreteViewShadowNode(
    const ShadowNodeFragment& fragment,
    const ShadowNodeFamily::Shared& family,
    ShadowNodeTraits traits)
    : BaseShadowNode(fragment, family, traits) {
  initialize();
}

ConcreteViewShadowNode(
    const ShadowNode& sourceShadowNode,
    const ShadowNodeFragment& fragment)
    : BaseShadowNode(sourceShadowNode, fragment) {
  initialize();
}
```

The first one is called just once, when the view is mounted and the corresponding `ShadowNode` is created for the first time. It is never called again (e.g. when the node is cloned).

The second constructor is called when the `ShadowNode` is cloned and we want to create a copy (after each successful commit to the `ShadowTree`).

I think that we can use this logic to properly manage the state object. In the first constructor, we will initialize the `ShadowNode` state object but in the second one we will detect prop changes (if the props object instance is changed) and update the state respectively (trigger the css transition, add/remove/update css animations, etc.)

PR #7443, opened 2025-04-29
chore: Adjust CSS config parsers implementation to work with folly::dynamic

MatiPl01:
## Summary

Since we are moving implementation to a custom shadow node which gets `folly::dynamic` object in props, we no longer can create CSS transition/animation configs from a `jsi::Value`. Because of that, I needed to adjust the implementation of all config parsers and the `updateKeyframes` method in interpolators.

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/CSSAnimationConfig.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/CSSAnimationConfig.cpp
@@ -2,11 +2,11 @@
 
 namespace reanimated::css {
 
-double getIterationCount(jsi::Runtime &rt, const jsi::Object &settings) {
-  return settings.getProperty(rt, "iterationCount").asNumber();
+double getIterationCount(const folly::dynamic &settings) {
+  return settings["iterationCount"].asDouble();
 }
 
-AnimationDirection getDirection(jsi::Runtime &rt, const jsi::Object &settings) {
+AnimationDirection getDirection(const folly::dynamic &settings) {
   static const std::unordered_map<std::string, AnimationDirection>
       strToEnumMap = {
           {"normal", AnimationDirection::Normal},
@@ -14,7 +14,7 @@ AnimationDirection getDirection(jsi::Runtime &rt, const jsi::Object &settings) {
           {"alternate", AnimationDirection::Alternate},
           {"alternate-reverse", AnimationDirection::AlternateReverse}};
 
-  const auto str = settings.getProperty(rt, "direction").asString(rt).utf8(rt);
+  const auto str = settings["direction"].asString();
   auto it = strToEnumMap.find(str);
   if (it == strToEnumMap.cend()) {
     throw std::invalid_argument(
@@ -23,14 +23,14 @@ AnimationDirection getDirection(jsi::Runtime &rt, const jsi::Object &settings) {
   return it->second;
 }
 
-AnimationFillMode getFillMode(jsi::Runtime &rt, const jsi::Object &settings) {
+AnimationFillMode getFillMode(const folly::dynamic &settings) {
   static const std::unordered_map<std::string, AnimationFillMode> strToEnumMap =
       {{"none", AnimationFillMode::None},
        {"forwards", AnimationFillMode::Forwards},
        {"backwards", AnimationFillMode::Backwards},
        {"both", AnimationFillMode::Both}};
 
-  const auto str = settings.getProperty(rt, "fillMode").asString(rt).utf8(rt);
+  const auto str = settings["fillMode"].asString();
   auto it = strToEnumMap.find(str);
   if (it == strToEnumMap.cend()) {
     throw std::invalid_argument(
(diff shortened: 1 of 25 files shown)


packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/CSSAnimationConfig.cpp
```diff
@@ -54,138 +54,104 @@ AnimationPlayState getPlayState(jsi::Runtime &rt, const jsi::Object &settings) {
...
   }
 
   return result;
 }
 
 CSSAnimationSettingsMap parseNewAnimationSettings(
-    jsi::Runtime &rt,
     const std::vector<std::string> &animationNames,
-    const jsi::Value &newSettings) {
+    const folly::dynamic &newSettings) {
```
tjzel, line 128: A general question, are we supposed to use references when handling folly::dynamics?

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/CSSAnimationConfig.cpp
```diff
@@ -54,138 +54,104 @@ AnimationPlayState getPlayState(jsi::Runtime &rt, const jsi::Object &settings) {
...
+    result.animationNames = parseAnimationNames(config["animationNames"]);
   }
-
-  if (configObj.hasProperty(rt, "settingsUpdates")) {
-    result.settingsUpdates =
-        parseSettingsUpdates(rt, configObj.getProperty(rt, "settingsUpdates"));
+  if (config.count("newAnimationSettings") && result.animationNames) {
+    result.newAnimationSettings = parseNewAnimationSettings(
+        *result.animationNames, config["newAnimationSettings"]);
+  }
```
tjzel, line 152: Another general question, should we add a wrapper class that has `has` method? Using `count` is a bit weird looking.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/easing/EasingFunctions.cpp
```diff
@@ -38,26 +36,23 @@ EasingFunction getPredefinedEasingFunction(const std::string &name) {
...
+  const auto points = easingConfig["points"];
+  const auto pointsCount = points.size();
 
   for (size_t i = 0; i < pointsCount; i++) {
-    const auto pointObj = points.getValueAtIndex(rt, i).asObject(rt);
-    pointsX.push_back(pointObj.getProperty(rt, "x").asNumber());
-    pointsY.push_back(pointObj.getProperty(rt, "y").asNumber());
+    const auto pointObj = points[i];
+    pointsX.push_back(pointObj["x"].asDouble());
+    pointsY.push_back(pointObj["y"].asDouble());
```
tjzel, line 55: Can we use `foreach` loop here?

tjzel (approved):

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/CSSAnimationConfig.cpp
MatiPl01: I don't have a strong opinion on that. Maybe we can but I don't want to do this now as it doesn't affect the logic and is just a syntactic sugar.

I still think that we may want to get rid of the `folly::dynamic` in the future and create our custom type but it's rather a further future.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/CSSAnimationConfig.cpp
MatiPl01: Mhm, I am not sure if it will get copied in that case. When looking at the implementation of the `folly::dynamic`, I can see methods like this:

```cpp
dynamic dynamic::merge_diff(const dynamic& source, const dynamic& target) {
```

which also take const references to other dynamics in props so I think that the current approach with const references is valid.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/easing/EasingFunctions.cpp
MatiPl01: Sure. I didn't notice that I can change it after getting rid of the `jsi::Value`

packages/react-native-reanimated/Common/cpp/reanimated/CSS/util/keyframes.cpp
```diff
@@ -2,48 +2,36 @@
...
+    const folly::dynamic &keyframes) {
+  if (!keyframes.isArray()) {
     throw std::invalid_argument(
         "[Reanimated] Keyframes must be an array of keyframe objects");
   }
 
-  const auto keyframeArray = keyframes.asObject(rt).asArray(rt);
-  const auto keyframesCount = keyframeArray.size(rt);
+  const auto keyframesCount = keyframes.size();
+  const bool hasOffset0 = keyframes[0]["offset"].asDouble() == 0;
```
piaskowyk, line 13: Is it safe to assume that it is never empty?

piaskowyk (approved):
👍

packages/react-native-reanimated/Common/cpp/reanimated/CSS/util/keyframes.cpp
MatiPl01: I think that yes because the animation must have at least one keyframe.

We have this check in JS that ensures that the keyframes object is valid:

```ts
export const isCSSKeyframesObject = (
  value: object
): value is CSSAnimationKeyframes =>
  typeof value === 'object' && Object.keys(value).length > 0;
```

So there must be at least one keyframe in the animation keyframes object. You can see how we use `isCSSKeyframesObject` by looking at the `filterCSSAndStyleProperties` implementation where we don't return the animation config if keyframes aren't valid.

PR #7447, opened 2025-04-30
feat: Operations loop for easier update operations handling

MatiPl01:
## Summary

This PR just adds an `EventLoop` class implementation which will be useful in the future for easier CSS time based progress providers updates.

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/EventLoop.h
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/EventLoop.h
@@ -1,0 +1,25 @@
+#pragma once
+
+#include <functional>
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+namespace reanimated {
+
+class EventLoop {
+ public:
+  using Operation = std::function<bool(double timestamp)>;
+  using OperationHandle = unsigned long;
+
+  OperationHandle add(Operation operation);
+  void remove(OperationHandle handle);
+
+  void update(double timestamp);
+
+ private:
+  std::unordered_map<OperationHandle, Operation> operations_;
+  OperationHandle nextHandle_{0};
+};
+
+} // namespace reanimated
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/EventLoop.h
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/EventLoop.h
@@ -1,24 +1,33 @@
 #pragma once
 
+#include <reanimated/Tools/PlatformDepMethodsHolder.h>
+
 #include <functional>
 #include <memory>
 #include <unordered_map>
-#include <vector>
+#include <utility>
 
 namespace reanimated {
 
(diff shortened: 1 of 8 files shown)


packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
```diff
@@ -49,11 +51,15 @@ void ReanimatedCommitHook::maybeInitializeLayoutAnimations(
...
-    ShadowTree const &,
-    RootShadowNode::Shared const &,
+    ShadowTree const &shadowTree,
+    RootShadowNode::Shared const &oldRootShadowNode,
     RootShadowNode::Unshared const &newRootShadowNode) noexcept {
   ReanimatedSystraceSection s("ReanimatedCommitHook::shadowTreeWillCommit");
 
+  // Update the event loop in the commit hook to ensure that all operations
+  // will be updated in the same batch with the same timestamp
+  eventLoop_->update();
```
MatiPl01, line 61: I would like to keep this `update` method called on every `shadowTreeWillCommit` call to ensure that all operations (e.g. animation/transition time provider updates) are handled at the same time with the same timestamp.

This adds just another layer of abstraction and would help us clean up the `performOperations` implementation. The `EventLoop` object would also allow to add time providers from our custom shadow node.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.h
```diff
@@ -1,10 +1,12 @@
 #pragma once
 
+#include <reanimated/Fabric/EventLoop.h>
 #include <reanimated/Fabric/updates/UpdatesRegistryManager.h>
 #include <reanimated/LayoutAnimations/LayoutAnimationsProxy.h>
 
 #include <react/renderer/uimanager/UIManagerCommitHook.h>
 
+#include <chrono>
```
WoLewicki, line 9: Why is `chrono` used here?

MatiPl01: Good catch! I just forgot to remove it 😄.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/EventLoop.h
```diff
@@ -0,0 +1,34 @@
...
+#include <reanimated/Tools/PlatformDepMethodsHolder.h>
+
+#include <functional>
+#include <memory>
+#include <unordered_map>
+#include <utility>
+
+namespace reanimated {
+
+class EventLoop {
```
piaskowyk, line 12: Since in JavaScript the `EventLoop` name is closely related to the Runtime, I'm not sure if it's the best name in this context 🤔. Could you also elaborate more on the purpose of this EventLoop? Feel free to explain it internally on Slack, if you prefer.

Issue #7449, opened 2025-04-30
Unnecesarry internal shared value read/write


terrysahaidak:
### Description

While profiling my app using https://github.com/margelo/react-native-release-profiler i noticed too many `executeOnUIRuntimeSunc` calls when mounting my screen. I know for a fact it's used to read from shared values on UI thread and return it to JS thread so it's blocking both JS and UI thread for that time. This is why it's better to avoid reading from shared values on JS thread.

While in some cases we can just have checks such as `if (!_WORKLET) {}` to skip reading on JS thread, i've noticed that there are a few places in Reanimated source code that does it potentially unnecessarily.

### `useAniamtedStyle`

1. This check in `useAnimatedStyle` in my opinion should be done only in `_DEV_` and inside some `useEffect` otherwise it will be reading from that shared value on each rerender

https://github.com/software-mansion/react-native-reanimated/blob/3280fd0a4ee3941c589933304d933df781f3ea85/packages/react-native-reanimated/src/hook/useAnimatedStyle.ts#L586

2. I would avoid this `.value = true` running on useEffect considering that the initial value of that shared value is also `true`.

https://github.com/software-mansion/react-native-reanimated/blob/3280fd0a4ee3941c589933304d933df781f3ea85/packages/react-native-reanimated/src/hook/useAnimatedStyle.ts#L580

### `useAnimatedRef`

This `.value` read will most likely result in `-1` (initial value of the shared value) on the mount of component. can we avoid this read on mount though?

https://github.com/software-mansion/react-native-reanimated/blob/3280fd0a4ee3941c589933304d933df781f3ea85/packages/react-native-reanimated/src/hook/useAnimatedRef.ts#L74

### Unnecessary `cancelAnimation` calls

There are a few places in the code base such as the one below where I would replace this `useSharedValue` with `useState(() => makeMutable(true))` to avoid the unnecessary `cancelAnimation` call on unmount for a shared value that is not used for the animations.

https://github.com/software-mansion/react-native-reanimated/blob/3280fd0a4ee3941c589933304d933df781f3ea85/packages/react-native-reanimated/src/hook/useAnimatedStyle.ts#L496

### Steps to reproduce

1. Add a console.log or a debugger call before https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/src/mutables.ts#L159 to log all `.value` calls on JS thread 
2. Using `debugger` you can move back to the function that calls it

### Snack or a link to a repository

[...]

PR #7452, opened 2025-04-30
fix: useAnimatedRef returns wrong tag on ref attach

tjzel:
## Summary

Fixing a bug where `useAnimatedRef` would return `null` tag even though the tag was obtained, reported in #7449. The issue probably stems from Shared Value read/write refactor where this change should be included.

Also fixing unnecessary reads of `.value`.

## Test plan

Run examples that use `useAnimatedRef`, like PinExample and see that they work properly.

### DIFF
--- a/packages/react-native-reanimated/src/hook/useAnimatedRef.ts
+++ b/packages/react-native-reanimated/src/hook/useAnimatedRef.ts
@@ -49,10 +49,11 @@ export function useAnimatedRef<
   const ref = useRef<AnimatedRef<TComponent> | null>(null);
 
   if (!ref.current) {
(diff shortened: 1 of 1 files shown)


MatiPl01 (approved):
Looks fine

tomekzaw (approved):
LGTM, let's just confirm that the following still works (i.e. calling `animatedRef` as a function to obtain `tag.value` on the UI runtime):
```ts
runOnUI(() => {
  console.log(animatedRef()); // should print `{}`
})();
```
Also, let's rename `tag` to `shadowNodeWrapper` in a separate PR.

tjzel:
@tomekzaw works like a charm 👍

PR #7456, opened 2025-05-02
fix: move text color props to UI thread props

hannojg:


piaskowyk:
Hey Hanno 👋 
Since we've removed the `synchronouslyUpdatePropsOnUIThread` method, the AllowList splitting doesn't matter anymore because both lists are now merged into a single list on the native side:
https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp#L336-L345

tomekzaw:
Hi Hanno, the code modified by this PR has already been removed in https://github.com/software-mansion/react-native-reanimated/pull/7642 so I'm closing the PR.

PR #7461, opened 2025-05-03
feat: Scheduled operations chaining

MatiPl01:


packages/react-native-reanimated/Common/cpp/reanimated/Fabric/operations/AtomicQueue.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
+
+#include <atomic>
+#include <memory>
+#include <vector>
+
+namespace reanimated {
+
+template <typename TItem>
+class AtomicQueue {
```
MatiPl01, line 10: This queue aims to remove the necessity to use mutexes. It allows to call its methods from multiple threads without thinking about protecting access as the queue maintains everything internally.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/operations/DelayedItemsManager.h
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    return std::less<>()(
+        std::addressof(lhs.id), std::addressof(rhs.id));
+  }
+};
+
+// Allow specifying just TId (for cases when value is used as id)
+template <typename TId, typename TValue = TId>
+class DelayedItemsManager {
```
MatiPl01, line 35: I decided to move this `DelayedItemComparator` implementation to the `.h` file because I don't want to add all template type specifications in the `.cpp` file as I had to do before. It is cumbersome and storing the implementation within `.h` is not that bad in this case.

I also refined this implementation a bit to allow passing a custom ID and value. The previous implementation allowed just a value which was treated as a key (e.g. to remove something from the `DelayedItemsManager` I had to pass exactly the same reference to the value as I used while adding it to the manager). This was cumbersome and I adjusted the implementation to allow both - either using just a value or using a key and value.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/operations/Operation.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+Operation &Operation::doWhile(std::function<bool(double)> op) {
+  steps_.emplace_back(0, std::move(op));
+  return *this;
+}
+
+std::unique_ptr<ExecutableOperation> Operation::build() {
+  return std::unique_ptr<ExecutableOperation>(
+      new ExecutableOperation(std::move(steps_)));
+}
```
MatiPl01, line 43: This method converts the temporary `Operation` type to the `ExecutableOperation`. I need this distinction to ensure that the scheduled operation (added to the `OperationsLoop`) isn't modified after scheduling (which may break). Because `ExecutableOperation` is immutable, I can safely process operations in the `OperationsLoop`.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/operations/OperationsLoop.cpp
```diff
@@ -0,0 +1,94 @@
...
+OperationsLoop::OperationHandle OperationsLoop::schedule(
+    std::unique_ptr<ExecutableOperation> operation) {
+  OperationHandle handle = nextHandle_.fetch_add(1, std::memory_order_relaxed);
+  additionRequests_.enqueue({handle, std::move(operation)});
+  return handle;
+}
+
+void OperationsLoop::remove(OperationHandle handle) {
+  removalRequests_.enqueue(handle);
+}
```
MatiPl01, line 18: I queue all addition/removal operations to ensure that the entire batch is executed at once during a single frame. Because I use the `AtomicQueue` instances here, I can also safely call these methods from different threads, also when the `update` is being executed.

PR #7462, opened 2025-05-04
feat: Complete CSSTransitionManager implementation

MatiPl01:
## Summary

This PR add a complete implementation of the `CSSTransitionManager` class which will replace the current `CSSTransitionsRegistry` C++ class and the current `CSSTransitionManager` JS class. It is much simpler and more readable than the `CSSTransitionsRegistry`. The key difference is that the `CSSTransitionManager` is responsible for management of just one node's transition instead of managing all CSS transitions.


packages/react-native-reanimated/Common/cpp/reanimated/CSS/manager/CSSTransitionManager.cpp
```diff
@@ -8,11 +8,103 @@ CSSTransitionManager::CSSTransitionManager(
...
+            transition->run(timestamp, props, lastFrameProps);
+          })
+          .waitFor([transition = transition_](double timestamp) {
+            return transition->getMinDelay(timestamp);
+          })
+          .doWhile([transition = transition_](double timestamp) mutable {
+            transition->update(timestamp);
+            return transition->getState() == TransitionProgressState::Running;
+          })
+          .build());
```
MatiPl01, line 107: Thanks to the #7461 PR, we can now manage the transition state via a simple operations chain.

There is no more need to manually check if the transition is delayed or running before stopping it (e.g. when the transition is removed or interrupted by the new one). There is also no need to add transition to the `DelayedOperationsManager` and then pulling it from it and activating - everything will be handled by the `OperationsLop`.

packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/managers/CSSTransitionManager.cpp
```diff
@@ -0,0 +1,110 @@
...
+          })
+          .waitFor([transition = transition_](double timestamp) {
+            return transition->getMinDelay(timestamp);
+          })
+          .doWhile([transition = transition_](double timestamp) mutable {
+            transition->update(timestamp);
+            return transition->getState() == TransitionProgressState::Running;
+          })
+          .build());
+}
```
MatiPl01, line 108: Thanks to the #7461 PR, we can now manage the transition state via a simple operations chain.

There is no more need to manually check if the transition is delayed or running before stopping it (e.g. when the transition is removed or interrupted by the new one). There is also no need to add transition to the `DelayedOperationsManager` and then pulling it from it and activating - everything will be handled by the `OperationsLoop`.

PR #7467, opened 2025-05-05
feat: New CSS managers on the JS side

MatiPl01:
## Summary

This PR adds the implementation of new JS managers for CSS transition and animations. They are much simpler than previous implementation because we moved the props diffing logic to CPP as we just pass the processed config via the `ReanimatedView` props and react to all changes in CPP.

The `CSSAnimationsManager` is a bit more complex as it needs to register CSS animation keyframes (via a JSI call) before they are used. This is an optimization to allow reusing the same animation keyframes (thus the CPP style interpolator) for multiple views without having to parse keyframes and create separate interpolators multiple times.


packages/react-native-reanimated/src/css/component/AnimatedComponent.tsx
```diff
@@ -44,6 +38,7 @@ export default class AnimatedComponent<
   ChildComponent: AnyComponent;
 
   _CSSManager?: CSSManager;
+  _CSSManagerNew?: NewCSSManager; // TODO - replace the old manager in the native implementation with this one
```
MatiPl01, line 41: This is temporary and will replace the old `_CSSManager` once the CSS rewrite is ready. I need to store both because we decided to keep the old version working during the rewrite so I cannot get rid of the old one until the new implementation is ready.

PR #7469, opened 2025-05-06
feat(Worklets): runOnUIAsync

skusnierz:
## Summary
This PR adds a `runOnUIAsync` function.
A function that schedules a worklet on the UI Runtime, to be executed on the UI thread. It returns a Promise of the worklet's return value. The Promise is resolved asynchronously, not immediately after the callback execution.

Example usage :
```js
import { runOnUIAsync } from 'react-native-worklets';

// RN Runtime, JS thread

const result: Promise<number> = runOnUIAsync((): number => {
  return 42;
});

await result; // 42
```

## Test plan
`runOnUIAsync` example : 
<video src="https://github.com/user-attachments/assets/8c72548d-6d41-43d6-809c-3f27f13d664f" />


tjzel (changes requested):
Looks, very good, let's just fix types and linting.

tjzel (approved):
Congratz! 🥳

PR #7472, opened 2025-05-06
fix: use `providers.exec` to enable `configuration-cache`

lukmccall:
## Description

Fixes:
```
Starting an external process 'node --print require.resolve('react-native/package.json')' during configuration time is unsupported
``` 
when `configuration-cache` is enabled.

## Changes

Replace `execute` with `providers.exec`. You can read more about it here: https://docs.gradle.org/8.13/userguide/configuration_cache.html#config_cache:requirements:external_processes. 

- build project with `configuration-cache` flag ✅


tomekzaw (approved):
Thanks for the PR!

tjzel (approved):

PR #7481, opened 2025-05-08
refactor(Worklets)(makeShareableClone): add makeShareableString

skusnierz:
## Summary
The first in a series of PRs that are intended to refactor the logic of the `makeShareableClone` function.
The current implementation does not have a single source of truth, as we have two type checking logics for the passed value, one on the `JS` side and one on the `Cpp` side.
The idea of the refactor is to implement one source of truth on the `JS` side and call the corresponding method on the `Cpp` side e.g:
```js
if (typeof value === ‘string’) {
  return global._makeShareableString(value);
}
```

In addition, in this PR I have rewritten the `Shareables` example to make it more readable and easier to use.
## Test plan
`Shareables` example
<video src="https://github.com/user-attachments/assets/5d61009d-e31b-44fc-b4fd-c21a8c417612" />

### DIFF
--- a/apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
+++ b/apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
@@ -9,12 +9,12 @@ export default function ShareablesExample() {
     <View style={styles.container}>
       <View style={styles.statusInfoBox}>
         <View style={styles.statusInfoRow}>
-          <Text style={styles.statusInfoText}>🟢 - ok</Text>
-          <Text style={styles.statusInfoText}>🚫 - error</Text>
+          <Text style={styles.statusInfoText}>🟢 - Ok</Text>
+          <Text style={styles.statusInfoText}>🚫 - Throws</Text>
         </View>
         <View style={styles.statusInfoRow}>
-          <Text style={styles.statusInfoText}>🔴 - not_ok</Text>
-          <Text style={styles.statusInfoText}>⚪ - not_checked</Text>
+          <Text style={styles.statusInfoText}>🔴 - Failed</Text>
+          <Text style={styles.statusInfoText}>⚪ - Pending</Text>
         </View>
       </View>
 
--- a/apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
+++ b/apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
@@ -97,10 +97,10 @@ function StringDemo() {
   const expectedStatus: Status = 'ok';
 
   const handlePress = () => {
+    const testString = 'test';
     runOnUI(() => {
       'worklet';
       try {
-        const testString = 'test';
         if (testString === 'test') {
           runOnJS(isOk)();
         } else {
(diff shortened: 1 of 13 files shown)


packages/react-native-worklets/src/WorkletsModule/JSWorklets.ts
```diff
@@ -26,6 +26,12 @@ class JSWorklets implements IWorkletsModule {
     );
   }
 
+  makeShareableString(_string: string): ShareableRef<string> {
```
tjzel, line 29:
```suggestion
  makeShareableString(): ShareableRef<string> {
```

packages/react-native-worklets/src/WorkletsModule/NativeWorklets.ts
```diff
@@ -50,6 +51,10 @@ See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooti
     );
   }
 
+  makeShareableString(string: string) {
+    return this.#workletsModuleProxy.makeShareableString(string);
+  }
```
tjzel, line 56: It's better not to have the same name of the variable as a name of a type.

```suggestion
  makeShareableString(str: string) {
    return this.#workletsModuleProxy.makeShareableString(str);
  }
```

packages/react-native-worklets/src/WorkletsModule/workletsModuleProxy.ts
```diff
@@ -10,6 +10,8 @@ export interface WorkletsModuleProxy {
     nativeStateSource?: object
   ): ShareableRef<TValue>;
 
+  makeShareableString(string: string): ShareableRef<string>;
```
tjzel, line 13:
```suggestion
  makeShareableString(str: string): ShareableRef<string>;
```

tjzel (changes requested):
Excellent work, just a few minor details to improve!

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -1,10 +1,32 @@
...
   return (
     <View style={styles.container}>
+      <View style={styles.statusInfoBox}>
+        <View style={styles.statusInfoRow}>
+          <Text style={styles.statusInfoText}>🟢 - ok</Text>
+          <Text style={styles.statusInfoText}>🚫 - error</Text>
+        </View>
+        <View style={styles.statusInfoRow}>
+          <Text style={styles.statusInfoText}>🔴 - not_ok</Text>
+          <Text style={styles.statusInfoText}>⚪ - not_checked</Text>
```
tjzel, line 17: How about this?

```suggestion
          <Text style={styles.statusInfoText}>🟢 - Ok</Text>
          <Text style={styles.statusInfoText}>🚫 - Throws</Text>
        </View>
        <View style={styles.statusInfoRow}>
          <Text style={styles.statusInfoText}>🔴 - Failed</Text>
          <Text style={styles.statusInfoText}>⚪ - Pending</Text>
```

PR #7482, opened 2025-05-08
fix: Temporary fix for CSS issues on RN 0.78+

MatiPl01:
## Summary

This PR just adds one line that disables props transferring from the native to the JS shadow node. This is a temporary fix for the breaking change introduced in RN 0.78 which started causing a lot of issues with our current CSS animations and transitions implementation approach.


MatiPl01:
It was fixed properly in the #7529 PR.

Issue #7288, opened 2025-03-21
[BUILD ISSUE] RNR v4


Amurmurmur:
### Description

Had no build issues with ` "react-native-reanimated": "4.0.0-beta.2"`
However after an upgrade to beta3 and installing `react-native-worklets@0.2.0` or 0.1.0
Having these errors during build:
```
› Compiling react-native-worklets Pods/RNWorklets » WorkletsModuleProxySpec.cpp

❌  (node_modules/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxySpec.cpp:1:10)

> 1 | #include <worklets/NativeModules/WorkletsModuleProxySpec.h>
    |          ^ 'worklets/NativeModules/WorkletsModuleProxySpec.h' file not found
  2 | 
  3 | #include <utility>
  4 | 

› Compiling react-native-worklets Pods/RNWorklets » WorkletsModuleProxy.cpp
› Compiling react-native-worklets Pods/RNWorklets » WorkletsModule.mm
› Compiling react-native-worklets Pods/RNWorklets » WorkletsMessageThread.mm
› Compiling react-native-worklets Pods/RNWorklets » WorkletsJSIUtils.cpp

❌  (node_modules/react-native-worklets/Common/cpp/worklets/Tools/WorkletsJSIUtils.cpp:1:10)

> 1 | #include <worklets/Tools/WorkletsJSIUtils.h>
    |          ^ 'worklets/Tools/WorkletsJSIUtils.h' file not found
  2 | #include <vector>
  3 | 
  4 | using namespace facebook;

› Compiling react-native-worklets Pods/RNWorklets » WorkletRuntimeRegistry.cpp

❌  (node_modules/react-native-worklets/Common/cpp/worklets/Registries/WorkletRuntimeRegistry.cpp:1:10)

> 1 | #include <worklets/Registries/WorkletRuntimeRegistry.h>
    |          ^ 'worklets/Registries/WorkletRuntimeRegistry.h' file not found
  2 | 
  3 | namespace worklets {
  4 | 

› Compiling react-native-worklets Pods/RNWorklets » WorkletRuntimeDecorator.cpp

❌  (node_modules/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntimeDecorator.cpp:1:10)

> 1 | #include <worklets/SharedItems/Shareables.h>
    |          ^ 'worklets/SharedItems/Shareables.h' file not found
[...]

PR #7483, opened 2025-05-08
Fix building Reanimated on iOS with static frameworks

tomekzaw:
## Summary

Fixes https://github.com/software-mansion/react-native-reanimated/issues/7288.

This PR fixes building react-native-reanimated for iOS when `USE_FRAMEWORKS=static` or `USE_FRAMEWORKS=dynamic` is set due to missing header search path after adding `Common/NativeView` subdirectory.


piaskowyk (approved):
🫡

PR #7484, opened 2025-05-08
feat(CI): monorepo static checks

tjzel:
## Summary

QoL CI since now we keep some scripts etc. in the root. I also fixed problems found by `yarn lint` and `yarn format` and added necessary configs.

## Test plan

🪨

### DIFF
--- a/packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ReanimatedViewState.h
+++ b/packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ReanimatedViewState.h
@@ -13,10 +13,10 @@ class ReanimatedViewState {
 #ifdef ANDROID
   ReanimatedViewState(
       ReanimatedViewState const &previousState,
-      folly::dynamic data) {};
+      folly::dynamic data) {}
   folly::dynamic getDynamic() const {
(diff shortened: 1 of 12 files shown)


packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ReanimatedViewState.h
```diff
@@ -13,10 +13,10 @@ class ReanimatedViewState {
 #ifdef ANDROID
   ReanimatedViewState(
       ReanimatedViewState const &previousState,
-      folly::dynamic data) {};
+      folly::dynamic data) {}
```
patrycjakalinska, line 16: why don't we need semicolon here?

tjzel: It's the definition of a constructor, you don't need a semicolon after the function's body enclosing bracket.

patrycjakalinska: thanks!

patrycjakalinska (approved):
lgmt 🪨

PR #7485, opened 2025-05-08
fix: ensure last[propName] is defined in style updater

crjc:
## Summary

Ensure that `last[propName]` is initialised to fix an exception on web. This bug was introduced by changes in #6749.

I noticed this issue after upgrading to the latest version of react-native-reanimated. In my case, the library was trying to index `last['transform']` which didn't exist for whatever reason.

Screenshot 2025-05-08 at 16 23 04

Screenshot 2025-05-08 at 16 24 29


m-bert (approved):
LGTM

PR #7487, opened 2025-05-09
cherry-pick: fix: ensure last[propName] is defined in style updater

crjc:
## Summary

Cherry pick of #7485.

Hoping this could get into a patch release for 3.17 if possible?

### DIFF
--- a/packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
+++ b/packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
@@ -257,6 +257,9 @@ function styleUpdater(
            * states, causing abrupt transitions or 'jumps' in animation states.
            */
           if (Array.isArray(updates[propName])) {
+            if (!last[propName] || typeof last[propName] !== 'object') {
+              last[propName] = {};
+            }
             updates[propName].forEach((obj: StyleProps) => {
               for (const prop in obj) {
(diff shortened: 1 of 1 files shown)


patrycjakalinska:
Hi, as we didn't ask in initial commit - could you provide a reproduction (example code with instructions) of a problem that this fix is patching? I'm not able to reproduce a problem where I would get such error. Thanks!

crjc:
I haven't managed to create a repro with just reanimated, this example uses Moti on top:

https://github.com/crjc/reanimated-bug/blob/master/App.tsx

1. `npm i`
2. `npm run web`
3. Click "shake"
4. Observe error in console

```
Uncaught TypeError: Cannot set properties of undefined (setting 'translateX')
    at useAnimatedStyle.js:165:37
    at Array.forEach (<anonymous>)
    at frame (useAnimatedStyle.js:163:31)
```

Someone else seemingly has also experienced this bug:
https://github.com/software-mansion/react-native-reanimated/commit/bdadece7a47cf58da4a351ede5504a7e881a197b#commitcomment-156673492

patrycjakalinska:
Thanks for the repro! c: We decided to move your fix inside the loop to place emphasis on why it's there. We will cherry pick from #7506  to `3.18`, once again - thanks!

PR #7489, opened 2025-05-09
feat: inject __DEV__ to worklet runtimes

tjzel:
## Summary

In the new bundling method we won't copy implicit global identifiers. However, Reanimated & Worklets still depend on copying `__DEV__` from the global object. This PR injects `__DEV__` to Worklet Runtimes while maintaining previous behavior, it's needed for future PRs of worklets.

## Test plan

Run an app in dev & prod and see that it works properly.

### DIFF
--- a/packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
+++ b/packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
@@ -30,7 +30,7 @@ WorkletsModuleProxy::WorkletsModuleProxy(
     std::function<void(std::function<void(const double)>)>
         &&forwardedRequestAnimationFrame)
     : WorkletsModuleProxySpec(jsCallInvoker),
-      bundleFlavor_(
+      isDevBundle_(
           rnRuntime.global().getProperty(rnRuntime, "__DEV__").asBool()),
       valueUnpackerCode_(valueUnpackerCode),
       jsQueue_(jsQueue),
@@ -42,7 +42,7 @@ WorkletsModuleProxy::WorkletsModuleProxy(
           jsScheduler,
           "Reanimated UI runtime",
           true /* supportsLocking */,
-          bundleFlavor_,
+          isDevBundle_,
           valueUnpackerCode_)),
       animationFrameBatchinator_(std::make_shared<AnimationFrameBatchinator>(
           uiWorkletRuntime_->getJSIRuntime(),
@@ -116,7 +116,7 @@ jsi::Value WorkletsModuleProxy::createWorkletRuntime(
       jsScheduler_,
       name.asString(rt).utf8(rt),
       true /* supportsLocking */,
-      bundleFlavor_,
+      isDevBundle_,
       valueUnpackerCode_);
   auto initializerShareable = extractShareableOrThrow<ShareableWorklet>(
       rt, initializer, "[Worklets] Initializer must be a worklet.");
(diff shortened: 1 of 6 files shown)


packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
```diff
@@ -30,6 +30,8 @@ WorkletsModuleProxy::WorkletsModuleProxy(
     std::function<void(std::function<void(const double)>)>
         &&forwardedRequestAnimationFrame)
     : WorkletsModuleProxySpec(jsCallInvoker),
+      isDevBundle_(
+          rnRuntime.global().getProperty(rnRuntime, "__DEV__").asBool()),
```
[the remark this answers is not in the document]
tjzel: `__DEV__` is always defined, but I checked that `undefined` would throw here, I added handling of non-booleans 👍

packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
```diff
@@ -20,6 +20,8 @@ using namespace facebook;
 
 namespace worklets {
 
+auto isDevBundleFromRNRuntime(jsi::Runtime &rnRuntime) -> bool;
```
tomekzaw, line 23: Why not just ...?
```suggestion
bool isDevBundleFromRNRuntime(jsi::Runtime &rnRuntime);
```

packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
```diff
@@ -152,4 +157,12 @@ jsi::Value WorkletsModuleProxy::scheduleOnRuntime(
   return jsi::Value::undefined();
 }
 
+auto isDevBundleFromRNRuntime(jsi::Runtime &rnRuntime) -> bool {
+  auto rtDev = rnRuntime.global().getProperty(rnRuntime, "__DEV__");
+  if (rtDev.isBool()) {
+    return rtDev.asBool();
+  }
+  return false;
```
tomekzaw, line 165:
```suggestion
  return rtDev.isBool() && rtDev.asBool();
```

packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
```diff
@@ -152,4 +157,12 @@ jsi::Value WorkletsModuleProxy::scheduleOnRuntime(
   return jsi::Value::undefined();
 }
 
+auto isDevBundleFromRNRuntime(jsi::Runtime &rnRuntime) -> bool {
+  auto rtDev = rnRuntime.global().getProperty(rnRuntime, "__DEV__");
```
tomekzaw, line 161:
```suggestion
  const auto rtDev = rnRuntime.global().getProperty(rnRuntime, "__DEV__");
```

tomekzaw (approved):

packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
tjzel: It's a syntax sugar that allows for more legible functions, I can remove it if you prefer.

tjzel: i.e. you can get

```c++
auto foo() -> someType;
auto bar() -> someLongerType;
auto foobar() -> int;
```

instead of

```c++
someType foo();
someLongerType bar();
int foobar();
```

PR #7490, opened 2025-05-09
Fix ref management

piaskowyk:


piaskowyk:
I'll close it since we've stopped work on Native View

PR #7498, opened 2025-05-12
refactor(Worklets)(makeShareableClone): add makeShareableNull and makeShareableUndefined

skusnierz:
## Summary
This PR adds few new functions : 
- `makeShareableNull`
- `makeShareableUndefined`

and is part of a series of PRs that aim to refactor the `makeSharableClone` function -> https://github.com/software-mansion/react-native-reanimated/pull/7481
## Test plan
`Shareables` example

<video src="https://github.com/user-attachments/assets/1c9100ca-012c-4efe-b6e1-fdc66da08472" />


packages/react-native-worklets/src/shareables.ts
```diff
@@ -139,6 +139,14 @@ function makeShareableCloneRecursiveNative<T>(
     return cloneBigInt(value) as ShareableRef<T>;
   }
 
+  if (value === undefined) {
+    return cloneUndefined() as ShareableRef<T>;
```
tomekzaw, line 143: I believe we don't need to call `cloneUndefined()` each time – instead, we can just store only one instance of it in some const and always return the same instance. In fact, we don't even need a function, but it's okay for now.

packages/react-native-worklets/src/shareables.ts
```diff
@@ -139,6 +139,14 @@ function makeShareableCloneRecursiveNative<T>(
...
+  }
+
+  if (value === null) {
+    return cloneNull() as ShareableRef<T>;
```
tomekzaw, line 147: Same as for `cloneUndefined`.

tomekzaw (changes requested):

packages/react-native-worklets/src/shareables.ts
```diff
@@ -139,6 +142,14 @@ function makeShareableCloneRecursiveNative<T>(
     return cloneBigInt(value) as ShareableRef<T>;
   }
 
+  if (value === undefined) {
+    return CLONED_UNDEFINED_JS_REF as ShareableRef<T>;
```
tomekzaw, line 146: Can we remove `as ShareableRef<T>` part here?

packages/react-native-worklets/src/shareables.ts
```diff
@@ -104,6 +104,9 @@ const VALID_ARRAY_VIEWS_NAMES = [
   'DataView',
 ];
 
+const CLONED_UNDEFINED_JS_REF = cloneUndefined();
+const CLONED_NULL_JS_REF = cloneNull()
```
tomekzaw, line 108:
```suggestion
const CLONED_NULL_JS_REF = makeShareableNull();
```

packages/react-native-worklets/src/shareables.ts
tomekzaw, line 143: Shouldn't it be `makeShareableUndefined`?

packages/react-native-worklets/src/WorkletsModule/NativeWorklets.ts
```diff
@@ -8,14 +8,19 @@ import { WorkletsError } from '../WorkletsError';
...
+}
 
 export function createNativeWorkletsModule(): IWorkletsModule {
   return new NativeWorklets();
 }
 
 class NativeWorklets {
   #workletsModuleProxy: WorkletsModuleProxy;
+  shareableUndefined: ShareableRef<undefined>;
+  shareableNull: ShareableRef<null>;
```
tomekzaw, line 23: Let's keep them as private fields
```suggestion
  #shareableUndefined: ShareableRef<undefined>;
  #shareableNull: ShareableRef<null>;
```

tomekzaw (approved):

packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
```diff
@@ -90,6 +90,14 @@ jsi::Value WorkletsModuleProxy::makeShareableBigInt(
   return worklets::makeShareableBigInt(rt, bigint);
 }
 
+jsi::Value WorkletsModuleProxy::makeShareableUndefined(jsi::Runtime &rt) {
+  return worklets::makeShareableUndefined(rt);
+}
+
+jsi::Value WorkletsModuleProxy::makeShareableNull(jsi::Runtime &rt) {
+  return worklets::makeShareableNull(rt);
+}
```
tjzel, line 99: I wonder if we actually should cache `shareableUndefined` and `shareableNull` globally as they are runtime agnostic and we don't need multiple instances of them.

cc @tomekzaw

tjzel (approved):
Excellent work!

packages/react-native-worklets/Common/cpp/worklets/NativeModules/WorkletsModuleProxy.cpp
skusnierz: good point, I will improve this in a separate PR

PR #7501, opened 2025-05-12
feat: Set display to contents from native, fix component descriptor initialization

MatiPl01:
## Summary

- add shared and reusable `reanimatedNodeProps_` props object for all reanimated shadow nodes containing `display` property set to `contents` (since reanimated Shadow Nodes should be "transparent", we don't need any props except `display: 'contents'`. Thanks to this PR, we no longer have to pass `display: 'contents'` style prop in JS),
- fixes the initialization of the `ReanimatedViewComponentDescriptor` (it used some strange initialization logic pulling the `proxy` instance from the context container before, which didn't sometimes work. In general, the `ReanimatedModuleProxy` is usually initialized first, but it is not the case during hot reloads, when the order of initialization of `ReanimatedViewComponentDescriptor` and `ReanimatedModuleProxy` can be different. Because of that, we ended up with an uninitialized `ReanimatedModuleProxy` instance with invalid shadow nodes without properly initialized state object that depends on the `proxy`)

### DIFF
--- a/packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ReanimatedViewStateData.cpp
+++ b/packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ReanimatedViewStateData.cpp
@@ -2,16 +2,20 @@
 
 namespace facebook::react {
 
-bool ReanimatedViewStateData::isInitialized() const {
-  return initialized_;
+ReanimatedViewStateData::ReanimatedViewStateData() {
+  // We throw an error here, because we cannot create ReanimatedViewStateData
+  // without proxy. React Native requires this constructor to be present.
+  throw std::runtime_error(
+      "[Reanimated] Cannot create ReanimatedViewStateData without proxy");
+}
+
+ReanimatedViewStateData::ReanimatedViewStateData(
+    const std::shared_ptr<ReanimatedModuleProxy> &proxy) {
+  initialize(proxy);
 }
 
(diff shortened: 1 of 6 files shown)


packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ReanimatedViewStateData.cpp
```diff
@@ -2,16 +2,20 @@
 
 namespace facebook::react {
 
-bool ReanimatedViewStateData::isInitialized() const {
-  return initialized_;
+ReanimatedViewStateData::ReanimatedViewStateData() {
+  // We throw an error here, because we cannot create ReanimatedViewStateData
+  // without proxy. React Native requires this constructor to be present.
+  throw std::runtime_error(
+      "[Reanimated] Cannot create ReanimatedViewStateData without proxy");
```
MatiPl01, line 9: React Native requires this constructor because in the default implementation of the `createInitialState` method in the `ConcreteComponentDescriptor` class it constructs the state object without any params. We override the `createInitialState` method and initialize the `ReanimatedViewStateData` with the `proxy` object passed to the constructor, which is required to properly initialize CSS managers. Because of that, the empty `ReanimatedViewStateData()` constructor is invalid but we cannot get rid of it if we want CPP to compile. As a result, I decided to add this error to make sure that we never initialize state without the required proxy parameter.

MatiPl01:
Closing temporarily. Maybe I will reopen a similar PR later on.

PR #7505, opened 2025-05-13
refactor(Worklets)(makeShareableClone): add makeSharableObject

skusnierz:
## Summary
This PR adds few new functions : 
- `makeShareableArray`
- `makeShareableObject`
- `makeShareableHostObject`
- `makeShareableInitializer`

and is part of a series of PRs that aim to refactor the `makeSharableClone` function -> https://github.com/software-mansion/react-native-reanimated/pull/7481
## Test plan
`Shareables` example

<video src="https://github.com/user-attachments/assets/653aa88c-7213-482e-80f9-bbc29d895b51" />


packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -135,6 +135,67 @@ jsi::Value makeShareableBigInt(jsi::Runtime &rt, const jsi::BigInt &bigint) {
...
+  return ShareableJSRef::newHostObject(rt, shareable);
+}
+
+jsi::Value makeShareableInitializer(
+    jsi::Runtime &rt,
+    const jsi::Object &initializerObject) {
+  if (initializerObject.getProperty(rt, "__init").isUndefined()) {
+    throw std::runtime_error(
+        "[Worklets] Attempted to convert an initializer object that doesn't have __init property.");
+  }
```
tjzel, line 193:
```suggestion
    throw std::runtime_error(
        "[Worklets] Attempted to convert an initializer object that doesn't have the `__init` property.");
  }
```

tjzel (commented):
Can you split this into multiple PRs, one type for each PR? Objects, HostObjects and Initializers are quite critical so we really need these changes to be as atomic as possible.

Could you also open a PR where `makeShareableClone` would use the newly implemented functions (i.e. `makeShareableBoolean`), so we'd no longer have duplicated implementation of the logic?

PR #7507, opened 2025-05-13
docs: Bump t-rex to stable with Docusaurus 3.7

patrycjakalinska:
## Summary

- [ ] Prepare docs
- [ ] Bump `@swmansion/t-rex-ui` to `1.0.0`
- [ ] Change MDXComponents to swizzled one from `@swmansion/t-rex-ui`

### DIFF
--- a/packages/docs-reanimated/docusaurus.config.js
+++ b/packages/docs-reanimated/docusaurus.config.js
@@ -189,17 +189,17 @@ const config = {
               new webpack.DefinePlugin({
                 ...processMock,
                 __DEV__: 'false',
-                setImmediate: () => {},
               }),
             ],
             module: {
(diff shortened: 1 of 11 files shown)


kacperkapusciak (approved):
Tested on both dev and production build, it looks like it working perfectly fine 🎉

I believe with Docusaurus 3 we can now use Yarn 4 - the same we use for other parts of the monorepo. Could you double check that before merging please?

patrycjakalinska:
After upgrading to Yarn 4 there are some errors that might possibly be fixed but not straight away. We can do it sometime in another PR?

tjzel:
Let's bump Yarn in a separate PR 👍

PR #7511, opened 2025-05-14
Migrate `WorkletsModule` and `ReanimatedModule` to `TurboModuleWithJSIBindings`

tomekzaw:
## Summary

Work in progress.


packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/NativeProxy.java
```diff
@@ -84,27 +83,20 @@ public class NativeProxy {
...
 
-    CallInvokerHolderImpl callInvokerHolder = JSCallInvokerResolver.getJSCallInvokerHolder(context);
-    mHybridData =
-        initHybrid(
-            workletsModule,
-            Objects.requireNonNull(context.getJavaScriptContextHolder()).get(),
-            callInvokerHolder,
-            fabricUIManager);
+    mHybridData = initHybrid(workletsModule, fabricUIManager);
+    checkCppVersion();
```
tjzel, line 87:
```suggestion
if(BuildConfig.Debug){
    checkCppVersion();
    }
```

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
```diff
@@ -65,17 +65,10 @@ class NativeReanimatedModule implements IReanimatedModule {
...
-        // We install a DummyReanimatedModuleProxy instead.
-        this.#reanimatedModuleProxy = new DummyReanimatedModuleProxy();
-        return;
-      }
-    }
-    if (global.__reanimatedModuleProxy === undefined) {
+    if (
+      ReanimatedTurboModule === null ||
+      global.__reanimatedModuleProxy === undefined
+    ) {
```
tjzel, line 71: Without returning `false` here we might throw this error when there's a double-reload problem here - although `TurboModuleWithJSIBinding` should supposedly fix that problem. Try reloading quickly a couple of times and see if this error is thrown.

packages/react-native-worklets/android/src/main/cpp/worklets/android/WorkletsModule.cpp
```diff
@@ -13,43 +13,19 @@ using namespace react;
 
 WorkletsModule::WorkletsModule(
     jni::alias_ref<jhybridobject> jThis,
-    jsi::Runtime *rnRuntime,
     jni::alias_ref<JavaMessageQueueThread::javaobject> messageQueueThread,
-    const std::shared_ptr<facebook::react::CallInvoker> &jsCallInvoker,
-    const std::shared_ptr<worklets::JSScheduler> &jsScheduler,
-    const std::shared_ptr<UIScheduler> &uiScheduler)
+    jni::alias_ref<worklets::AndroidUIScheduler::javaobject> androidUIScheduler)
```
tjzel, line 17: Similar comment here, I think passing just the `uiScheduler` is fine here.

packages/react-native-worklets/android/src/main/cpp/worklets/android/WorkletsModule.cpp
```diff
@@ -68,14 +44,33 @@ WorkletsModule::getForwardedRequestAnimationFrame() {
...
+jni::local_ref<BindingsInstallerHolder::javaobject>
+WorkletsModule::getBindingsInstaller() {
+  return jni::make_local(BindingsInstallerHolder::newObjectCxxArgs(
+      [&](jsi::Runtime &rnRuntime,
+          const std::shared_ptr<CallInvoker> &jsCallInvoker) {
+        workletsModuleProxy_ = std::make_shared<WorkletsModuleProxy>(
+            rnRuntime,
+            messageQueueThread_,
+            jsCallInvoker,
+            uiScheduler_,
```
tjzel, line 70: I understand that now we keep `messageQueueThread_` and `uiScheduler_` here because the initialization is "deferred". How about

```suggestion
            rnRuntime,
            std::move(messageQueueThread_),
            jsCallInvoker,
            std::move(uiScheduler_),
```

since we don't need these pointers here after `WorkletsModuleProxy` is created.

packages/react-native-worklets/android/src/paper/com/swmansion/worklets/NativeWorkletsModuleSpec.java
tjzel, line 1: Isn't this needed on Paper regardless?

packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/JSCallInvokerResolver.java
tjzel, line 1: Are you sure we don't need that anymore? (including Paper)

packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/WorkletsModule.java
```diff
@@ -45,31 +45,21 @@ protected HybridData getHybridData() {
 
   @OptIn(markerClass = FrameworkAPI.class)
   private native HybridData initHybrid(
-      long jsContext,
-      MessageQueueThread messageQueueThread,
-      CallInvokerHolderImpl jsCallInvokerHolder,
-      AndroidUIScheduler androidUIScheduler);
+      MessageQueueThread messageQueueThread, AndroidUIScheduler androidUIScheduler);
 
+  @OptIn(markerClass = FrameworkAPI.class)
```
tjzel, line 50: Was this annotation unnecessary?

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
```diff
@@ -45,14 +49,10 @@ public NodesManager getNodesManager() {
...
-    getReactApplicationContext().assertOnJSQueueThread();
-    mNodesManager.getNativeProxy().installJSIBindings();
-    if (BuildConfig.DEBUG) {
-      mNodesManager.getNativeProxy().checkCppVersion();
-    }
-    return true;
+  @NonNull
+  @Override
+  public BindingsInstallerHolder getBindingsInstaller() {
+    return mNodesManager.getNativeProxy().getBindingsInstaller();
```
tjzel, line 55: Must the bindingsInstaller come from the NativeProxy?

packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
```diff
@@ -114,19 +98,25 @@ void NativeProxy::injectCppVersion() {
...
+          const std::shared_ptr<CallInvoker> &jsCallInvoker) {
+        reanimatedModuleProxy_ = std::make_shared<ReanimatedModuleProxy>(
+            workletsModuleProxy_,
+            rnRuntime,
+            jsCallInvoker,
+            getPlatformDependentMethods(),
+            getIsReducedMotion());
+        reanimatedModuleProxy_->init(getPlatformDependentMethods());
+        reanimatedModuleProxy_->initializeFabric(uiManager_);
+        WorkletRuntimeCollector::install(rnRuntime);
```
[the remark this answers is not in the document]
tomekzaw: That's correct, `WorkletRuntimeCollector` should be moved to react-native-worklets.

packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/JSCallInvokerResolver.java
tomekzaw: We no longer support Paper on current main and in Reanimated 4.

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
tomekzaw: Indeed, I'm getting the following error but only on double reloads on iOS. Single reloads on iOS are fine.

Screenshot 2025-05-15 at 08 35 49

As for Android, single reload seems to work fine, and multiple reloads fail on this assert: `assert uiManager != null;`

packages/react-native-reanimated/android/src/main/java/com/swmansion/reanimated/ReanimatedModule.java
tomekzaw: It was already done this way (`installJSIBindings` method was defined in `NativeProxy`, not in `ReanimatedModule`) so I kept it that way as this probably needs an even bigger refactor.

packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/WorkletsModule.java
tomekzaw: Honestly I don't know what it does and why it's needed

tjzel: I think Android Studio would display warnings if this annotation wasn't present, if it's no longer the case we can remove it.

packages/react-native-worklets/android/src/main/cpp/worklets/android/WorkletsModule.cpp
tjzel: I wanted not to keep them bound in `WorkletsModule.cpp` forever, and use `std::move` them to effectively reset them when constructing `WorkletsModuleProxy`.

packages/react-native-worklets/android/src/main/java/com/swmansion/worklets/JSCallInvokerResolver.java
tjzel: Last time I checked (however, it was quite a while ago) Worklets worked on Paper - it doesn't cost us much to keep that support, are you sure we want to remove it?

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
tjzel: That's what I'm worried about - this error is a false positive, since it comes from the invalidated runtime, not from the new one - or at least it was the case some time ago. I want to avoid potential issues that might be opened because of that.

tomekzaw:
TODO: check release builds

tomekzaw:
iOS release works fine

tomekzaw:
iOS debug works fine on single reload but fails on double reload

Screenshot 2025-05-19 at 13 55 45

tomekzaw:
Android debug works fine but crashes on multiple reload

```
java_vm_ext.cc:616] JNI DETECTED ERROR IN APPLICATION: JNI GetObjectRefType called with pending exception java.lang.AssertionError:
java_vm_ext.cc:616]   at void com.swmansion.reanimated.NodesManager.<init>(com.facebook.react.bridge.ReactApplicationContext, com.swmansion.worklets.WorkletsModule) (NodesManager.java:69)
java_vm_ext.cc:616]   at void com.swmansion.reanimated.ReanimatedModule.<init>(com.facebook.react.bridge.ReactApplicationContext) (ReanimatedModule.java:22)
java_vm_ext.cc:616]   at com.facebook.react.bridge.NativeModule com.swmansion.reanimated.ReanimatedPackage.getModule(java.lang.String, com.facebook.react.bridge.ReactApplicationContext) (ReanimatedPackage.java:30)
java_vm_ext.cc:616]   at com.facebook.react.bridge.NativeModule com.facebook.react.ReactPackageTurboModuleManagerDelegate.lambda$initialize$0(com.facebook.react.BaseReactPackage, com.facebook.react.bridge.ReactApplicationContext, java.lang.String) (ReactPackageTurboModuleManagerDelegate.java:67)
java_vm_ext.cc:616]   at com.facebook.react.bridge.NativeModule com.facebook.react.ReactPackageTurboModuleManagerDelegate$$ExternalSyntheticLambda0.getModule(java.lang.String) (D8$$SyntheticClass:0)
                                                                                                    
```

PR #7512, opened 2025-05-14
feat(Worklets): new ShareableImport type

tjzel:
## Summary

This pull request introduces new type of a Shareable, `ShareableImport`. It's a type that, when extracted on another runtime, tries to "pull itself" from the JavaScript bundle.

For example, imagine you have a function `foo` in a `module.js` module:

```js
// module.js
export function foo(){
    console.log('hello world');
}
```

If this function were to be serialized as `ShareableImport`, the following code:

```js
import {runOnUI} from 'react-native-worklets';
import {foo} from 'module';

runOnUI(foo)();
```

Would actually "forward" the instruction `import {foo} from 'module'` to the UI Runtime, and the UI Runtime would invoke (Metro) require to obtain it and execute it.

For the time being there are no use cases for `ShareableImport` but they will come in the upcoming PRs.

## Test plan

Nothing to test, it doesn't affect runtime at the moment.


packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.h
```diff
@@ -258,6 +264,23 @@ class ShareableWorklet : public ShareableObject {
...
+      const jsi::String &imported)
+      : Shareable(ImportType),
+        source_(source.utf8(rt)),
+        imported_(imported.utf8(rt)) {}
+
+  jsi::Value toJSValue(jsi::Runtime &rt) override;
+
+ protected:
+  const std::string source_;
+  const std::string imported_;
```
tjzel, line 281: These names were inspired by how the respective nodes are named in the AST of Babel's parser.

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -284,6 +292,21 @@ jsi::Value ShareableWorklet::toJSValue(jsi::Runtime &rt) {
       rt, obj, jsi::String::createFromAscii(rt, "Worklet"));
 }
 
+jsi::Value ShareableImport::toJSValue(jsi::Runtime &rt) {
+  const auto metroRequire = rt.global().getProperty(rt, "__r");
```
tjzel, line 296: The `require` available in Metro is called `__r`. It can be called with the name of the module only in debug, it's something to tackle on in the future.

tomekzaw: Let's add a comment in the source code about `"__r"`

bartlomiejbloniarz (approved):

packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
```diff
@@ -181,6 +183,19 @@ jsi::Value JSIWorkletsModuleProxy::get(
           return makeShareableBoolean(rt, args[0].asBool());
         });
   }
+  if (name == "makeShareableImport") {
+    return jsi::Function::createFromHostFunction(
```
mrousavy, line 187: i wish y'all would use nitro for this lol

PR #7513, opened 2025-05-14
feat(Worklets): generated worklets

tjzel:
## Summary

This pull request introduces extracting each worklet to its dedicated separate file. This feature is available behind a feature flag `experimentalBundling` - it has no purpose now, but it will have it in the upcoming PRs.

The reason to extract worklets to separate files is to enable importing arbitrary worklets on Worklet Runtimes from the bundle.

The file containing a worklet in user-space most likely has imports (side-effects) from React Native, which would break a Worklet Runtime. By extracting the worklet to a separate file, we get rid of any unwanted side-effects.

To illustrate what kind of change this brings, say we have the following file:

```ts
// file.ts
const foo = 5;

const bar = () => {
  'worklet';
  return foo;
};
```

With legacy bundling the file would be transformed to:

```js
// file.legacy.out.js
const foo = 5;
const _worklet_12379853172447_init_data = {
  code: 'function fileTs1(){const{foo}=this.__closure;return foo;}',
  location: 'location data',
  sourceMap: 'source map data',
  version: '0.4.0',
};
const bar = (function fileTs1Factory({
  _worklet_12379853172447_init_data,
  foo,
}) {
  const _e = [new global.Error(), -2, -27];
  const fileTs1 = function () {
    return foo;
  };
  fileTs1.__closure = {
    foo,
  };
  fileTs1.__workletHash = 12379853172447;
  fileTs1.__initData = _worklet_12379853172447_init_data;
  fileTs1.__stackDetails = _e;
  return fileTs1;
})({
  _worklet_12379853172447_init_data,
  foo,
});

```

With experimental bundling, the file output would be:

```js
// file.experimental.out.js
const foo = 5;
const bar =
  require('react-native-worklets/__generatedWorklets/12379853172447.js').default(
    {
      foo,
    }
  );
```

While also emitting an extra file:

```js
// react-native-worklets/__generatedWorklets/12379853172447.js
const _worklet_12379853172447_init_data = {
  code: 'function fileTs1(){const{foo}=this.__closure;return foo;}',
  location: 'location data',
  sourceMap: 'source map data',
  version: '0.4.0',
};
export default (function fileTs1Factory({ foo }) {
  const _e = [new global.Error(), -2, -27];
  const fileTs1 = function () {
    return foo;
  };
  fileTs1.__closure = {
    foo,
  };
  fileTs1.__workletHash = 12379853172447;
  fileTs1.__initData = _worklet_12379853172447_init_data;
  fileTs1.__stackDetails = _e;
  return fileTs1;
});
```

For the sake of atomicity, I'm not stripping unused `initData` in experimental bundling just yet.

## Test plan

No testing is required now, as this feature is not yet used.


packages/react-native-worklets/__tests__/__snapshots__/plugin.test.ts.snap
```diff
@@ -3750,8 +3750,10 @@ var foo = function foo_null1Factory(_ref) {
 `;
 
 exports[`babel plugin for web configuration skips initData when omitNativeOnlyData option is set to true 1`] = `
-"var foo = function foo_null1Factory(_ref) {
-  var _worklet_7087145691632_init_data = _ref._worklet_7087145691632_init_data;
+"var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault");
+var _objectDestructuringEmpty2 = _interopRequireDefault(require("@babel/runtime/helpers/objectDestructuringEmpty"));
+var foo = function foo_null1Factory(_ref) {
+  (0, _objectDestructuringEmpty2.default)(_ref);
```
tjzel, line 3756: This was actually a regression I fixed here - there should be no initData.

PR #7514, opened 2025-05-14
chore: fix prettierignore configs

tjzel:
## Summary

Turns out that prettier VSCode plugin is bugged when it comes to monorepos. When calling `yarn prettier` from the CLI, it reads the closest `.prettierignore` in the file hierarchy - which is what we want. However, the VSCode prettier plugin instead reads the `.prettierignore` from the monorepo root, effectively disabling it when the root ignore had "apps" and "packages". See https://github.com/prettier/prettier-vscode/issues/3424 for reference.

Turns out the same applies to ESLint 🤯 

## Test plan

Run `yarn format` and `yarn lint` in root and see that nothing that's ignored is formatted. Break formatting in any `apps` or `packages` file and see that VSCode plugin formats it back.


patrycjakalinska (approved):
🪨

PR #7515, opened 2025-05-14
fix: cleanup sensors in invalidate and make nativeProxy not leak

WoLewicki:
## Summary

There are essentially 2 changes in the PR. Right now when you do a reload in a simple app with reanimated installed, you will see in AS profiler that instances of all `NativeProxy`s are still in the memory. It is caused by `javaPart_` never being released  thus keeping java part after reload. 

Due to this destructor of `NativeProxy` is not called ever. After adding `javaPart_ = nullptr` in `invalidateCpp`, it works and destructor is called. But calling `reanimatedModuleProxy_->cleanupSensors();` there makes the app crash since we already reset the `reanimatedModuleProxy_` in `invalidateCpp`.

We can go other way probably and move 

```
  reanimatedModuleProxy_->cleanupSensors();
  reanimatedModuleProxy_.reset();
```
to destructor but seems like `invalidateCpp` is the cleanup method now.

Side note:
I am a bit worried that some code regarding sensors might still run after the `invalidateCpp` method is run, but maybe that would be prohibited state?

## Test plan

Make simple app with reanimated and do a couple of reloads and check in profiler if the `NativeProxy` leaks.

### DIFF
--- a/packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
+++ b/packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
@@ -50,11 +50,6 @@ NativeProxy::NativeProxy(
 NativeProxy::~NativeProxy() {
   // removed temporary, new event listener mechanism need fix on the RN side
   // reactScheduler_->removeEventListener(eventListener_);
-
-  // cleanup all animated sensors here, since NativeProxy
-  // has already been destroyed when AnimatedSensorModule's
-  // destructor is ran
-  reanimatedModuleProxy_->cleanupSensors();
 }
 
 jni::local_ref<NativeProxy::jhybriddata> NativeProxy::initHybrid(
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):
LGTM

packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
```diff
@@ -305,8 +300,11 @@ PlatformDepMethodsHolder NativeProxy::getPlatformDependentMethods() {
 }
 
 void NativeProxy::invalidateCpp() {
-  workletsModuleProxy_.reset();
```
tjzel, line 308: Have you removed this line on purpose?

I tested the app with your changes but without this line removed and I didn't get any crash related to sensors we discussed offline.

WoLewicki: Hmm I might have copied the code changes from 3.17.5 where this line does not exist yet, probably it should stay here. Good catch!

WoLewicki: Done

tomekzaw (approved):
LGTM again

tjzel (approved):

PR #7517, opened 2025-05-14
refactor: runtime-specific PlatformChecker

tjzel:


packages/react-native-worklets/src/PlatformChecker/index.ts
```diff
@@ -0,0 +1,30 @@
...
+} from './PlatformChecker';
+
+const mockedPlatformChecker = {
+  isJest: () => false,
+  isWeb: () => false,
+  shouldBeUseWeb: () => false,
+};
+
+let isJest: () => boolean;
+let isChromeDebugger: () => boolean;
```
piaskowyk: to remove 🫡

PR #7521, opened 2025-05-14
Remove `isChromeDebugger` completely

tomekzaw:
## Summary

React Native New Architecture itself doesn't support **Chrome Debugger** (not to be confused with **Chrome DevTools** which is the recommended to debug React Native apps way now) so we can finally get rid of all its traces in our codebase.


packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx
```diff
@@ -87,7 +86,7 @@ export class PropsFilter implements IPropsFilter {
         if (component._isFirstRender) {
           props[key] = value.value;
         }
-      } else if (key !== 'onGestureHandlerStateChange' || !isChromeDebugger()) {
+      } else {
```
tomekzaw, line 89: `isChromeDebugger()` always returns false, so the condition always evaluates to true.

tjzel (approved):

PR #7523, opened 2025-05-14
Move Java and C++ version checks to `NativeProxy` constructor on Android

tomekzaw:


packages/react-native-reanimated/android/src/main/cpp/reanimated/android/NativeProxy.cpp
```diff
@@ -76,7 +81,7 @@ jni::local_ref<NativeProxy::jhybriddata> NativeProxy::initHybrid(
 }
 
 #ifndef NDEBUG
-void NativeProxy::checkJavaVersion(jsi::Runtime &rnRuntime) {
+void NativeProxy::checkJavaVersion() {
```
tomekzaw, line 84: `rnRuntime` was unused.

tjzel (approved):

PR #7528, opened 2025-05-15
[0.80] Remove `folly_flags` in RNReanimated.podspec

tomekzaw:
## Summary

This PR removes `folly_flags` in RNReanimated.podspec since they break compatibility with RN 0.80.0-rc.0.

Instead, `folly-config.h` file from `RCT-Folly` dependency is used which is the default way to configure Folly.

This change is backward-compatible with at least 3 previous versions of react-native.


piaskowyk (approved):
🫡

PR #7529, opened 2025-05-15
Don't transfer reanimated updates to ReactJS

bartlomiejbloniarz:
## Summary

This PR changes the way reanimated clones `ShadowNodes`. Due to the new `runtimeShadowNodeReference` mechanism, react "sees" the updates we apply in the commit hook. This is a problem, because then we are unable to remove an animation, as no one will remember the state that would be displayed without the animation. Instead an intermediate state would be displayed. To fix that, we disable transferring of ShadowNode references for when we clone them from reanimated.


MatiPl01 (approved):
LTGM. I tested most of CSS examples on both (iOS and Android) platforms, and everything seems to work fine.

tomekzaw (approved):
You're a hero!

PR #7531, opened 2025-05-15
Add opacity and transform example

tomekzaw:
## Summary

This PR adds "opacity & transform" example which is a benchmark for animating `opacity` and `transform` styles.

<table>
<thead>
<tr>
<th>Android</th>
<th>iOS</th>
</tr>
</thead>
<tbody>
<tr>
<td><video src="https://github.com/user-attachments/assets/5f2cb9c3-463e-454f-9a06-6042429aa445" /></td>
<td><video src="https://github.com/user-attachments/assets/1d02f4f0-d796-479f-ae91-2d586bf608e8" /></td>
</tr>
</tbody>
</table>

### DIFF
--- a/apps/common-app/src/apps/reanimated/examples/OpacityTransformExample.tsx
+++ b/apps/common-app/src/apps/reanimated/examples/OpacityTransformExample.tsx
@@ -1,0 +1,50 @@
+import React, { useEffect } from 'react';
+import { StyleSheet, View } from 'react-native';
+import Animated, { useAnimatedStyle, useSharedValue, withRepeat, withTiming } from 'react-native-reanimated';
+
+const ROWS = 25;
+const COLS = 15;
+const SIZE = 25;
+
+export default function OpacityTransformExample() {
+  const sv = useSharedValue(0);
+
+  useEffect(() => {
+    sv.value = 0;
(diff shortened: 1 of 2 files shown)


apps/common-app/src/apps/reanimated/examples/OpacityTransformExample.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const ROWS = 25;
+const COLS = 15;
+const SIZE = 25;
+
+export default function OpacityTransformExample() {
+  const sv = useSharedValue(0);
+
+  useEffect(() => {
+    sv.value = 0;
```
MatiPl01, line 13:
```suggestion
```
I think that this isn't needed

tomekzaw: This is needed because if hot reload happens when `sv.value` is at let's say 0.7, after the reload shared value will oscillate between 0.7 and 1.

MatiPl01: Ok, I can see that, but I don't think that anyone will trigger hot reloads while testing this example, so it seemed to be unnecessary for me. Hot reloads would only happen if someone adds/removes code in this example and it is very unlikely.

PR #7540, opened 2025-05-18
feat: Pass CSS animation keyframes via props

MatiPl01:
## Summary

This PR adds animation keyframes and keyframe easings to the CSS animation properties. Thanks to this, we will be able to get rid of unnecessary jsi calls adding and removing CSS animations from the Keyframes registry. This will also allow us to remove or simplify the JS-side animation keyframes registry.


MatiPl01:
I will open a new PR as this one changed too much code and it would be better to implement the valid solution step by step starting from just passing keyframes via props instead of using the  JSI call and implementing the remaining logic in separate PRs.

PR #7544, opened 2025-05-19
refactor(Worklets)(makeShareableClone): add makeSharableArray

skusnierz:
## Summary
This PR adds `makeShareableArray` function and is part of a series of PRs that aim to refactor the makeSharableClone function -> https://github.com/software-mansion/react-native-reanimated/pull/7481
## Test plan

<video src="https://github.com/user-attachments/assets/10e66382-74f0-4024-a496-fd6b6dc073b5" />


apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -255,6 +256,42 @@ function NullDemo() {
   );
 }
 
+function ArrayDemo() {
+  const title = 'Array';
+  const { status, isOk, isNotOk, isError } = useStatus();
+  const expectedStatus: Status = 'ok';
+
+  const handlePress = () => {
+    const array = [1, 2, 3];
```
tjzel, line 265: Let's actually try to put all possible shareable types in this array.

packages/react-native-worklets/src/shareables.ts
```diff
@@ -580,9 +579,9 @@ export function makeShareableCloneOnUIRecursive<T>(
         return value.__remoteFunction;
       }
       if (Array.isArray(value)) {
-        return global._makeShareableClone(
+        return global._makeShareableArray(
           value.map(cloneRecursive),
-          undefined
+          false
```
tjzel, line 584: Can we actually just invoke it with 1 argument?
```suggestion
```

tjzel (commented):
Looks solid overall!

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -90,17 +85,17 @@ jsi::Value makeShareableClone(
...
+    return makeShareableNull(rt);
   } else if (value.isBool()) {
-    shareable = std::make_shared<ShareableScalar>(value.getBool());
+    return makeShareableBoolean(rt, value.getBool());
   } else if (value.isNumber()) {
-    shareable = std::make_shared<ShareableScalar>(value.getNumber());
+    return makeShareableNumber(rt, value.getNumber());
   } else if (value.isBigInt()) {
-    shareable = std::make_shared<ShareableBigInt>(rt, value.getBigInt(rt));
+    return makeShareableBigInt(rt, value.getBigInt(rt));
```
tomekzaw, line 98: Shouldn't this be done in a separate PR?

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntimeDecorator.cpp
```diff
@@ -131,6 +131,12 @@ void WorkletRuntimeDecorator::decorate(
         return makeShareableUndefined(rt);
       });
 
+  jsi_utils::installJsiFunction(
+      rt, "_makeShareableArray", [](jsi::Runtime &rt, const jsi::Value &value) {
+        return makeShareableArray(rt, value.asObject(rt).asArray(rt), false)
+            .asObject(rt);
```
tomekzaw, line 137: Why do we need `.asObject(rt);` here? Other methods here don't have it.

packages/react-native-worklets/src/shareables.ts
```diff
@@ -256,10 +256,9 @@ function cloneArray<T extends unknown[]>(
   const clonedElements = value.map((element) =>
     makeShareableCloneRecursive(element, shouldPersistRemote, depth + 1)
   );
-  const clone = WorkletsModule.makeShareableClone(
+  const clone = WorkletsModule.makeShareableArray(
     clonedElements,
-    shouldPersistRemote,
-    value
+    shouldPersistRemote
   ) as ShareableRef<T>;
```
tomekzaw, line 262: We used to pass `value` as `nativeStateSource` here. Removing it can potentially break cloning arrays with native state. Can you please check if we used to support them? If yes, we'd like to keep the support. If not, we can skip this for now.

skusnierz: image
the previous implementation didn't use `nativeStateSource` for the array

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
skusnierz: the code is the same but I can do it in a separate PR

packages/react-native-worklets/src/shareables.ts
tomekzaw: Cool, perhaps we should add a C++ check `if (object.hasNativeState(rt)) { /* throw error that is it not supported */ }`? What do you think @tjzel?

tjzel: I don't think it's a good idea. The user might have an array with a native state (which is unused, perhaps coming from a 3rd party library) and it might be confusing for the user why is the app crashing on that array.

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -255,6 +256,46 @@ function NullDemo() {
   );
 }
 
+function ArrayDemo() {
+  const title = 'Array';
+  const { status, isOk, isNotOk, isError } = useStatus();
+  const expectedStatus: Status = 'ok';
+
+  const handlePress = () => {
+    const array = [1, true, false, null, undefined, 'a', BigInt(123)];
```
tjzel, line 265: what about an object, an array, a remote function, a worklet, an initializer, an array buffer?

packages/react-native-worklets/src/shareables.ts
tomekzaw: If the library logic depends on that native state, it won't work anyway, so it's better to have some error than no error in that case.

tjzel: I was thinking more about the situation where the native state is an addition to the object, e.g. allowing for some optimizations, but not necessary for correctness.

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -255,6 +256,109 @@ function NullDemo() {
...
+      null,
+      // undefined
+      undefined,
+      // string
+      'a',
+      // bigint
+      BigInt(123),
+      // object
+      { a: 1 },
+      // remote function - not worklet
```
tjzel, line 286:
```suggestion
      // remote function - not a worklet
```

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -255,6 +256,109 @@ function NullDemo() {
...
+          // string
+          array[5] === 'a',
+          // bigint
+          typeof array[6] === 'bigint',
+          array[6] === BigInt(123),
+          // object
+          typeof array[7] === 'object',
+          array[7].a === 1,
+          // remote function - not worklet
+          typeof array[8] === 'function',
```
tjzel, line 325: We could also check for `.__remoteFunction` property when `__DEV__ === true`.

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -255,6 +256,109 @@ function NullDemo() {
...
+          // array
+          array[9].length === 1,
+          array[9][0] === 1,
+          // worklet function
+          typeof array[10] === 'function',
```
tjzel, line 330: We can also run `isWorkletFunction(array[10])`

tjzel (approved):
Me likes 👍

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -255,6 +256,111 @@ function NullDemo() {
...
+  const { status, isOk, isNotOk, isError } = useStatus();
+  const expectedStatus: Status = 'ok';
+
+  const handlePress = () => {
+    const arrayBuffer = new ArrayBuffer(3);
+    const uint8Array = new Uint8Array(arrayBuffer);
+    uint8Array[0] = 1;
+    uint8Array[1] = 2;
+    uint8Array[2] = 3;
+    const array: any[] = [
```
tjzel, line 270: Can we actually make an enum, so the code would look a bit better? i.e.

```ts
enum index {
  number = 0,
  string = 1,
  boolean = 2,
	...
}
```

I'm not sure if we would handle this well in a worklet context, but we can try.

PR #7548, opened 2025-05-20
refactor(Worklets)(makeShareableClone): add makeShareableInitializer

skusnierz:
## Summary
This PR adds `makeShareableInitializer` function and is part of a series of PRs that aim to refactor the makeSharableClone function -> https://github.com/software-mansion/react-native-reanimated/pull/7481

## Test plan

<video src="https://github.com/user-attachments/assets/8eae5fa2-ef53-4ff7-b5dd-ee6841f4d669" />


apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -255,6 +256,39 @@ function NullDemo() {
...
 
+function RegExpDemo() {
+  const title = 'RegExp';
+  const { status, isOk, isNotOk, isError } = useStatus();
+  const expectedStatus: Status = 'ok';
+
+  const handlePress = () => {
+    const regex = /test/;
+    // eslint-disable-next-line prefer-regex-literals
+    const regex2 = new RegExp('test');
```
tomekzaw, line 267:
```suggestion
    const regex1 = /test/;
    // eslint-disable-next-line prefer-regex-literals
    const regex2 = new RegExp('test');
```

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -255,6 +256,39 @@ function NullDemo() {
...
+    runOnUI(() => {
+      'worklet';
+      try {
+        const checks = [regex.test('test'), regex2.test('test')];
```
tomekzaw, line 271: Maybe there's a better way to see if two regexp are equal? Does `regex1 === regex2` work?

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -145,6 +145,18 @@ jsi::Value makeShareableNull(jsi::Runtime &rt) {
...
 }
 
+jsi::Value makeShareableInitializer(
+    jsi::Runtime &rt,
+    const jsi::Object &initializerObject) {
+  if (initializerObject.getProperty(rt, "__init").isUndefined()) {
+    throw std::runtime_error(
+        "[Worklets] Attempted to convert an initializer object that doesn't have the `__init` property.");
+  }
+  auto shareable =
```
tomekzaw, line 155:
```suggestion
  const auto shareable =
```

packages/react-native-worklets/src/shareables.ts
```diff
@@ -165,6 +165,13 @@ function makeShareableCloneRecursiveNative<T>(
   if (isHostObject(value)) {
     return cloneHostObject(value, shouldPersistRemote);
   }
+  if (isPlainJSObject(value) && value.__init) {
```
tomekzaw, line 168: Why don't we use `'__init' in value` instead? Which one is better?

packages/react-native-worklets/src/shareables.ts
```diff
@@ -332,23 +371,17 @@ function cloneWorklet<T extends WorkletFunction>(
   // that the __initData field that contains long strings represeting the
   // worklet code, source map, and location, will always be
   // serialized/deserialized once.
-  const clonedProps: Record<string, unknown> = {};
+  const clonedProps: Record<string, unknown> = cloneObjectProperties(
+    value,
+    true,
+    depth + 1
```
tomekzaw, line 377: Do we need `+1` here or not?

packages/react-native-worklets/src/shareables.ts
```diff
@@ -248,6 +255,38 @@ function cloneNull(): ShareableRef<null> {
...
 
+function cloneObjectProperties<T extends object>(
+  value: T,
+  shouldPersistRemote: boolean,
+  depth: number
+): Record<string, unknown> {
+  const clonedProps: Record<string, unknown> = {};
+  for (const [key, element] of Object.entries(value)) {
+    if (key === '__initData' && clonedProps.__initData !== undefined) {
+      continue;
```
tomekzaw, line 266: I know there was no such comment before but let's add one that explains why we continue here.

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
skusnierz: What I'm checking here is not whether `regex1 == regex2`, but whether the two different ways of defining `RegExp` are serialized well (work well om UI thread).

packages/react-native-worklets/src/shareables.ts
skusnierz: imo  it doesn't matter,  `'__init' in value` should also be ok, although it doesn't check whether there is any value under the `__init` key

tjzel (approved):
LGTM

tjzel (approved):
LGTM x2

packages/react-native-worklets/src/shareables.ts
```diff
@@ -406,7 +436,7 @@ function clonePlainJSObject<T extends object>(
 function cloneRegExp<T extends RegExp>(value: T): ShareableRef<T> {
   const pattern = value.source;
   const flags = value.flags;
-  const handle = makeShareableCloneRecursive({
+  const handle = cloneInitializer({
```
tomekzaw, line 439: Since now we have `cloneInitializer` function we can change it so that it accepts only the factory worklet instead of an object with `__init` property and also completely remove `__init` property from our logic in `ShareableInitializer` in C++.

tomekzaw, line 439: Since now we have `cloneInitializer` function we can change it so that it accepts only the factory worklet instead of an object with `__init` property and also completely remove `__init` property from our logic in `ShareableInitializer` in C++. cc @skusnierz @tjzel

tjzel: I think this is quite a significant change, since we'd need to refactor both TS, C++ and the value unpacker for that, let's hold with this change for now.

PR #7554, opened 2025-05-20
chore: move Worklets Babel Plugin deps

tjzel:
## Summary

The dependencies should have been moved when the plugin was moved from Reanimated to Worklets.

## Test plan

If you install standalone worklets now they won't fail on Babel transpilation step.


patrycjakalinska (approved):
lgtm 🪨

PR #7557, opened 2025-05-20
fix: lack of hidden hermes flags (#7398)

tjzel:
Recently Hermes made changes that made some of the methods `virtual` https://github.com/facebook/hermes/commit/91354c6aab0281d5f46a963111e47949eca2e79e. To not misalign the ABIs when reading Hermes headers we must have the exact same flags that were used in compilation. Unfortunately RN doesn't downstream these flags.

TODO:

- [x] Android part
- [x] Backport to v3 after merge

- Fabric example works in runtime:
  - [x] debug ios
  - [x] debug android
  - [x] release ios
  - [x] release android
- TVos example works in runtime:
  - [x] debug ios
  - [x] release ios
- MacOS example works in runtime:
  - [x] debug ios
  - [x] release ios


patrycjakalinska (approved):
tested 🪨

Issue #7412, opened 2025-04-18
[iOS] App crashing on launch when using custom build configuration


martinlAP:
### Description

Hello,

I'm having an issue where my app instantly crashes as soon as I use Reanimated components like: 
`<Reanimated.View></Reanimated.View>`

This only occurs when building a custom iOS build configuration on a bare React Native project. 
If I use the default "Debug" configuration it works as expected

### Steps to reproduce

1. Clone the repro repository
2. `yarn && cd ios && pod install`
3. Make sure to build the "Dev-Debug" configuration. 

### Snack or a link to a repository

https://github.com/martinlAP/keyboardcrash

### Reanimated version

3.17.4

### React Native version

0.79.1

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Other (please specify)

### Device

iOS simulator

### Host machine

macOS

### Acknowledgements

Yes

kneza23:
Yeah, the same thing is happening to me. I have lost 2 days debugging it, and finally found the culprit. Also Release mode works fine. Will try to fix it somehow and let you know if i find a solution.

martinlAP:
Hey @kneza23,
Had the same issue for the DEBUG flag, I had to compare a fresh project with my current one to find the problem.
I think I might of missed something during the upgrades 😅

Let me know if you find a solution for the crash, but I think this is something the maintainers of the library should check.
This was working perfectly fine on previous versions of React Native.

kneza23:
> Hey [@kneza23](https://github.com/kneza23), Had the same issue for the DEBUG flag, I had to compare a fresh project with my current one to find the problem. I think I might of missed something during the upgrades 😅
> 
> Let me know if you find a solution for the crash, but I think this is something the maintainers of the library should check. This was working perfectly fine on previous versions of React Native.

What did you do to fix it?

martinlAP:
> > Hey [@kneza23](https://github.com/kneza23), Had the same issue for the DEBUG flag, I had to compare a fresh project with my current one to find the problem. I think I might of missed something during the upgrades 😅
> > Let me know if you find a solution for the crash, but I think this is something the maintainers of the library should check. This was working perfectly fine on previous versions of React Native.
> 
> What did you do to fix it?

I did the same thing as you, added the DEBUG flag in the swift active compilations conditions ;)

kneza23:
[...]

PR #7428, opened 2025-04-24
fix: add hermes debugger for all debug variants

WoLewicki:
## Summary

PR analogous to https://github.com/facebook/react-native/pull/50897 fixing `HERMES_ENABLE_DEBUGGER` not being added to configs that are not strictly named `Debug`. Should fix https://github.com/software-mansion/react-native-reanimated/issues/7412. Thanks @tomekzaw and @piaskowyk for helping with it.

## Test plan

Run app from https://github.com/software-mansion/react-native-reanimated/issues/7412 and see that it doesn't crash.

PR #7558, opened 2025-05-20
chore(Reanimated3): cherry-pick #7428

tjzel:
PR analogous to https://github.com/facebook/react-native/pull/50897 fixing `HERMES_ENABLE_DEBUGGER` not being added to configs that are not strictly named `Debug`. Should fix https://github.com/software-mansion/react-native-reanimated/issues/7412. Thanks @tomekzaw and @piaskowyk for helping with it.

Run app from https://github.com/software-mansion/react-native-reanimated/issues/7412 and see that it doesn't crash.


patrycjakalinska (approved):
🪨

PR #7562, opened 2025-05-20
feat(Worklets): add Worklet Runtime bundle entry

tjzel:
## Summary

Soon we will be able to execute the bundle on Worklet Runtimes. For this purpose, we need to add an suitable entry point, that will disallow React Native from initializing. We also have to refactor various side-effects, since they tend to break Worklet Runtimes.

## Test plan

Run the example app and see that everything works properly - these changes should have no effect in runtime.

### DIFF
--- a/packages/react-native-worklets/src/bundleUnpacker.ts
+++ b/packages/react-native-worklets/src/bundleUnpacker.ts
@@ -1,0 +1,50 @@
+/* eslint-disable reanimated/use-worklets-error */
+'use strict';
+import type { WorkletFactory, WorkletFunction } from './workletTypes';
+
+export function bundleValueUnpacker(
+  objectToUnpack: ObjectToUnpack,
+  category?: string,
+  remoteFunctionName?: string
+): unknown {
+  'use strict';
+  let workletsCache = global.__workletsCache;
+  let handleCache = global.__handleCache;
+  if (workletsCache === undefined) {
+    workletsCache = global.__workletsCache = new Map();
+    handleCache = global.__handleCache = new WeakMap();
+  }
+  const workletHash = objectToUnpack.__workletHash;
+  if (workletHash !== undefined) {
+    const factory = globalThis.__r(workletHash).default as WorkletFactory;
+    const worklet = factory(objectToUnpack.__closure as never);
+    return worklet;
+  } else if (objectToUnpack.__init !== undefined) {
+    let value = handleCache.get(objectToUnpack);
+    if (value === undefined) {
+      value = objectToUnpack.__init();
+      handleCache.set(objectToUnpack, value);
+    }
+    return value;
+  } else if (category === 'RemoteFunction') {
+    const fun = () => {
+      const label = remoteFunctionName
+        ? `function \`${remoteFunctionName}\``
+        : 'anonymous function';
+      throw new Error(`[Worklets] Tried to synchronously call a non-worklet ${label} on the UI thread.
+See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooting#tried-to-synchronously-call-a-non-worklet-function-on-the-ui-thread for more details.`);
+    };
+    fun.__remoteFunction = objectToUnpack;
+    return fun;
+  } else {
+    throw new Error(
+      `[Worklets] Data type in category "${category}" not recognized by value unpacker: "${_toString(
+        objectToUnpack
+      )}".`
+    );
+  }
+}
+
+interface ObjectToUnpack extends WorkletFunction {
+  _recur: unknown;
+}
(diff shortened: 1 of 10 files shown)


packages/react-native-worklets/src/initializers.ts
```diff
@@ -18,21 +18,25 @@ import type { IWorkletsModule } from './WorkletsModule';
 const IS_JEST = isJest();
 const SHOULD_BE_USE_WEB = shouldBeUseWeb();
 
+const runtimeBoundLogToLogBoxAndConsole = logToLogBoxAndConsole;
```
tjzel, line 21: This is a tricky bit.

Since experimental bundling will allow to import functions from the bundle, we have to explicitly say that "This is a function that you must copy as a remote function to another runtime". I achieve that by reassigning a given function, so the Worklet Babel plugin wouldn't treat is as an import.

Again, it has no effect in runtime at the moment.

packages/react-native-worklets/src/initializers.ts
```diff
@@ -77,10 +81,18 @@ export function callGuardDEV<Args extends unknown[], ReturnValue>(
...
 export function setupCallGuard() {
   'worklet';
-  global.__callGuardDEV = callGuardDEV;
-  global.__ErrorUtils = {
+  if (!globalThis.__callGuardDEV) {
+    globalThis.__callGuardDEV = callGuardDEV;
+  }
+}
+
+const runtimeBoundReportFatalErrorOnJS = reportFatalErrorOnJS;
```
tjzel, line 89: See the comment about `runtimeBoundLogToLogBoxAndConsole`.

packages/react-native-worklets/src/initializers.ts
```diff
@@ -55,7 +59,7 @@ if (SHOULD_BE_USE_WEB) {
   // are applied before any async operations are executed on the UI runtime)
   executeOnUIRuntimeSync(registerWorkletsError)();
   executeOnUIRuntimeSync(registerLoggerConfig)(DEFAULT_LOGGER_CONFIG);
-  executeOnUIRuntimeSync(overrideLogFunctionImplementation)();
+  // executeOnUIRuntimeSync(overrideLogFunctionImplementation)();
```
tjzel, line 62: This function is the bane of my existence. It uses imports from React Native and is written in such a way that it keeps breaking the bundle no matter how I try to fix this.

However, it seems to actually have no effect since the introduction of Chrome Dev Tools, cc @MatiPl01 if it's safe to remove it.

packages/react-native-worklets/src/initializers.ts
```diff
@@ -77,10 +81,18 @@ export function callGuardDEV<Args extends unknown[], ReturnValue>(
...
-  global.__callGuardDEV = callGuardDEV;
-  global.__ErrorUtils = {
+  if (!globalThis.__callGuardDEV) {
+    globalThis.__callGuardDEV = callGuardDEV;
+  }
+}
+
+const runtimeBoundReportFatalErrorOnJS = reportFatalErrorOnJS;
+
+export function setupErrorUtils() {
```
tjzel, line 91: I split these functions, since with experimentalBundling we can install `setupCallGuard` directly from the bundle - but `setupErrorUtils` needs that RN Runtime binding.

packages/react-native-worklets/src/initializers.ts
tjzel: I actually managed to handle this problem in [0bcfe5d](https://github.com/software-mansion/react-native-reanimated/pull/7562/commits/0bcfe5da83dbbbe5da67103831e750f2b231c51b), so if we want to remove it we can do it in a separate PR.

packages/react-native-worklets/src/bundleUnpacker.ts
```diff
@@ -0,0 +1,50 @@
...
+  category?: string,
+  remoteFunctionName?: string
+): unknown {
+  'use strict';
+  let workletsCache = global.__workletsCache;
+  let handleCache = global.__handleCache;
+  if (workletsCache === undefined) {
+    workletsCache = global.__workletsCache = new Map();
+    handleCache = global.__handleCache = new WeakMap();
+  }
```
piaskowyk: To remove

packages/react-native-worklets/src/logger/LogBox.ts
```diff
@@ -45,7 +44,11 @@ interface LogBoxExtended extends LogBoxStatic {
   addLog(data: LogData): void;
 }
 
-const LogBox = RNLogBox as LogBoxExtended;
+const LogBox = globalThis._WORKLET
+  ? null
+  : // We cannot import LogBox on Worklet Runtimes.
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    (require('react-native').LogBox as LogBoxExtended);
```
piaskowyk: import maybe?

PR #5862, opened 2024-04-03
Split `useAnimatefRef` into two implementations

tjzel:
## Summary

Making this super dense code less dense.

## Test plan

:shipit:

PR #7566, opened 2025-05-21
fix: scrollTo not working on web

MatiPl01:
## Summary

It turns out that the #7452 PR broke `animatedRef()` calls on web. After changes introduced in the mentioned PR, this call always returned `null` instead of a correct reference to the component.

This was caused by the fact that the function was called without the `component` argument and the entire `if` statement that assigned a value to the `initialTag` was not executed. As a result, we always returned `initialTag` with the `null` value.


MatiPl01:
Closed in favor of: #5862

PR #7570, opened 2025-05-21
feat: Bump eslint to v9

skusnierz:
## Summary
- Bump eslint to v9
- Adjust eslint rules
## Test plan
🚀


.github/workflows/babel-plugin-static-check.yml
```diff
@@ -54,9 +55,6 @@ jobs:
       - name: Check types
         working-directory: ${{ env.PLUGIN_DIR }}
         run: yarn type:check
-      - name: Lint and format
```
skusnierz: we don't need it anymore because `yarn run lint:root` will scan reanimated plugin files

package.json
```diff
@@ -18,23 +18,42 @@
...
     "lint": "yarn lint:root && yarn lint:workspaces",
     "lint:workspaces": "yarn workspaces foreach --all --exclude \"react-native-reanimated-monorepo\" --parallel --topological-dev run lint",
-    "lint:root": "eslint --ignore-path .gitignore --ignore-path .eslintignore --ignore-path .prettiereslintignore . && yarn prettier --check --ignore-path .gitignore --ignore-path .prettierignore --ignore-path .prettiereslintignore ."
+    "lint:root": "eslint --flag unstable_config_lookup_from_file . && yarn prettier --check --ignore-path .gitignore --ignore-path .prettierignore --ignore-path .prettiereslintignore ."
   },
   "devDependencies": {
+    "@eslint/compat": "^1.2.9",
     "@types/shelljs": "^0.8.15",
-    "eslint": "^8.57.0",
+    "@typescript-eslint/eslint-plugin": "^8.34.0",
```
MatiPl01: I think that this plugin should be replaced by `typescript-eslint`. The official documentation suggests installing it for the new flat config: https://typescript-eslint.io/getting-started/

`@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin` had to be used separately in eslint v8. `typescript-eslint` combines both and adds support for the v9 flat config (at least I found this piece of information on the Internet and chat GPT is saying the same thing)

package.json
```diff
@@ -18,23 +18,42 @@
...
+    "eslint": "^9.29.0",
+    "eslint-config-prettier": "^10.1.5",
     "eslint-config-standard": "^17.1.0",
```
MatiPl01: Do we still need `eslint-config-standard`? It wasn't updated for about 2 years and it doesn't seem to be used anywhere in our eslint configuration.

package.json
```diff
@@ -18,23 +18,42 @@
...
-    "eslint-plugin-jsdoc": "^50.2.2",
+    "eslint-plugin-import": "^2.31.0",
+    "eslint-plugin-jest": "^28.14.0",
+    "eslint-plugin-jsdoc": "^50.7.1",
+    "eslint-plugin-n": "^17.20.0",
+    "eslint-plugin-promise": "^7.2.1",
```
MatiPl01, line 39: `eslint-plugin-promise` is not used anywhere

package.json
```diff
@@ -18,23 +18,42 @@
...
     "husky": "^9.0.11",
     "lint-staged": "^15.2.2",
     "node-fetch": "^3.3.2",
+    "prettier": "^3.5.3",
+    "prettier-plugin-jsdoc": "^1.3.2",
+    "typescript": "~5.3.0",
+    "typescript-eslint": "^8.34.0"
+  },
+  "resolutions": {
+    "@react-native/community-cli-plugin@npm:0.80.0-rc.4": "patch:@react-native/community-cli-plugin@npm%3A0.80.0-rc.4#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-rc.4-af2762c07e.patch",
```
MatiPl01: Should this change be a part of this PR?

eslint.tmp.js
```diff
@@ -0,0 +1,167 @@
+const { defineConfig, globalIgnores } = require('eslint/config');
```
MatiPl01: Is this a temporary file that you forgot to remove?

eslint.config.mjs
```diff
@@ -0,0 +1,237 @@
...
+import reactNative from 'eslint-plugin-react-native';
+import jest from 'eslint-plugin-jest';
+import globals from 'globals';
+import reanimated from 'eslint-plugin-reanimated';
+import importPlugin from 'eslint-plugin-import';
+import reactHooks from 'eslint-plugin-react-hooks';
+import react from 'eslint-plugin-react';
+import nodePlugin from 'eslint-plugin-n';
+
+/** @type {import('eslint').Linter.RulesRecord} */
```
MatiPl01: Cool, I like this typing here. I should adapt it to my config :)

eslint.config.mjs
```diff
@@ -0,0 +1,237 @@
...
+const tsDocRules = {
+  'tsdoc/syntax': 'error',
+};
+
+/** @type {import('eslint').Linter.Plugins} */
+const tsPlugins = {
+  tsdoc,
+  'simple-import-sort': simpleImportSort,
+};
```
MatiPl01: Just wondering if `simple-import-sort` should be under `tsPlugins`. It is not important but the name `tsPlugins` looks weird to me compared with the import sorting plugin, which isn't necessarily TS specific.
Maybe just put these plugins directly in the common ts config (`tsCommonConfig`)?

eslint.config.mjs
```diff
@@ -0,0 +1,237 @@
...
+
+const tsCommonExtends = [
+  tsEslint.configs.recommendedTypeChecked,
+  react.configs.flat.recommended,
+  react.configs.flat['jsx-runtime'],
+  reactHooks.configs['recommended-latest'],
+  importPlugin.flatConfigs.recommended,
+  importPlugin.flatConfigs.typescript,
+  nodePlugin.configs['flat/recommended-script'],
+];
```
MatiPl01, line 108: I think that you are missing prettier plugins here to prevent eslint conflicts with prettier configuration. `eslint-config-prettier` doesn't seem to be used anywhere.

eslint.config.mjs
```diff
@@ -0,0 +1,237 @@
...
+    ...tsPlugins,
+    reanimated,
+  },
+  rules: {
+    ...tsRules,
+    ...nodeRules,
+    'reanimated/use-reanimated-error': 'error',
+    'reanimated/use-global-this': 'error',
+  },
+  ignores: ['**/node_modules/**'],
```
MatiPl01: As far as I know, `node_modules` are ignored by default and you don't have to ignore them manually.

eslint.config.mjs
```diff
@@ -0,0 +1,237 @@
...
+};
+
+/** @type {import('eslint').Linter.Config[]} */
+const reanimatedConfig = [
+  {
+    files: ['**/react-native-reanimated/**/*.ts'],
```
MatiPl01: Why only `.ts` files? What about `.tsx`?

I think that you can omit the extension and it will still work as plugins are already configured to run on proper file types. At least you can try if it works without specifying any extensions.

eslint.config.mjs
```diff
@@ -0,0 +1,237 @@
...
+    },
+    rules: {
+      'reanimated/use-reanimated-error': 'error',
+      'reanimated/use-global-this': 'error',
+    },
+  },
+];
+
+/** @type {import('eslint').Linter.Config[]} */
+const workletsConfig = [
```
MatiPl01: Do we need to have separate a `workletsConfig` whilst it looks almost the same as `reanimatedConfig`?

I think that you could just use the same config for all files (so go without specifying `files` field explicilty) and then override eslint configuration locally in packages that need a different config (likely docs would need a bit different configuration).

If that is not possible, I think that you could still merge `workletsConfig` and `reanimatedConfig` into a single config as they are the same and just put both paths in the `files` array.

eslint.config.mjs
```diff
@@ -0,0 +1,237 @@
...
+      },
+      globals: {
+        React: true,
+        ...reactNative.environments['react-native']['react-native'],
+        ...jest.environments.globals.globals,
+        ...globals.node,
+      },
+    },
+    plugins: {
+      'react-native': reactNative,
```
MatiPl01: I feel that you would need to use `fixupPluginRules` as the react-native plugin doesn't support v9 and v8-style rules have to be adjusted to match the v9 style.

```js
'react-native': fixupPluginRules(reactNative)
```

eslint.config.mjs
```diff
@@ -0,0 +1,237 @@
...
+      jest,
+    },
+    settings: {
+      'import/resolver': {
+        'babel-module': { extensions: ['.js', '.jsx', '.ts', '.tsx'] },
+      },
```
MatiPl01, line 142: Shoudln't you use typescript resolver here? I am not sure if this is still needed as the example in the import plugin is still in eslintrc v8 style but I think that it would be safer to manually specify the ts resolver here instead of babel one. See: https://github.com/import-js/eslint-plugin-import?tab=readme-ov-file#typescript

I am also not sure if you should add parser here like this?

```js
'import/parsers': {
    '@typescript-eslint/parser': ['.ts', '.tsx']
},
```

Maybe it is added automatically now, but I don't know.

packages/react-native-reanimated/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "test": "jest",
     "lint": "yarn lint:js && yarn lint:common && yarn lint:android && yarn lint:apple",
-    "lint:js": "eslint src __tests__ __typetests__ && yarn prettier --check src __tests__ __typetests__",
+    "lint:js": "eslint src && yarn prettier --check src __tests__ __typetests__",
```
MatiPl01: Is this intended to remove eslint check on these directories?

packages/react-native-reanimated/src/Colors.ts
```diff
@@ -657,7 +657,7 @@ export function isColor(value: unknown): boolean {
   if (typeof value !== 'string') {
     return false;
   }
-  return processColorInitially(value) != null;
+  return processColorInitially(value) !== null;
```
MatiPl01: Not sure if it won't change behavior. Please ensure that everything still works.

packages/react-native-reanimated/src/css/component/createAnimatedComponent.tsx
```diff
@@ -5,7 +5,6 @@ import type {
   FunctionComponent,
   Ref,
 } from 'react';
-import React from 'react';
```
MatiPl01: I am not sure if it wasn't needed by next/web. In theory, we don't need to import React, but I remember having problem with this before.

packages/eslint-plugin-reanimated/index.js
```diff
@@ -241,7 +241,7 @@ var require_useReanimatedError = __commonJS({
       },
       meta: {
         docs: {
-          recommended: 'recommended',
+          // recommended: 'recommended',
```
MatiPl01: Mhm, should this be removed or was temporarily commented out?

packages/eslint-plugin-reanimated/index.js
```diff
@@ -286,7 +286,7 @@ var require_useWorkletsError = __commonJS({
...
```
MatiPl01: Here as well

packages/eslint-plugin-reanimated/src/useWorkletsError.ts
```diff
@@ -25,7 +25,7 @@ const rule: TSESLint.RuleModule<'useWorkletsError', []> = {
   },
   meta: {
     docs: {
-      recommended: 'recommended',
+      // recommended: 'recommended',
```
MatiPl01: Should this be removed?

I don't want to add this comment in each file, so you can check others as well, where you commented out this field.

apps/next-example/pages/_app.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable jsdoc/require-jsdoc */
+/* eslint-disable no-unused-vars */
```
MatiPl01: It is strange that this plugin reports unused vars in files like that. Have you tried to search on the Internet why that happens and if it can be fixed in other way than by just disabling the plugin?

apps/common-app/src/apps/reanimated/.eslintrc.js
```diff
@@ -5,20 +5,29 @@ const rootDir = path.resolve(__dirname, '../../../../../');
 
 /** @type {import('eslint').ESLint.ConfigData} */
```
MatiPl01: Does this v8 eslint config style still work in this example app? It seems that you bumped eslint to v9 for the common-app as well, so this config should no longer work.

.github/workflows/monorepo-static-checks.yml
```diff
@@ -33,11 +33,18 @@ jobs:
         uses: actions/setup-node@v4
         with:
           cache: 'yarn'
+          node-version: 22
 
       # Setup repo
       - name: Install dependencies
         run: yarn install --immutable
 
+      - name: Build worklets
```
MatiPl01: Should this PR change workflows? Is that needed after eslint upgrade or can it be done in a separate PR?

MatiPl01 (changes requested):
I left some comments.

This PR also introduces the following circular dependencies:

1) lib/typescript/commonTypes.d.ts > lib/typescript/Easing.d.ts
2) lib/typescript/hook/commonTypes.d.ts > lib/typescript/ViewDescriptorsSet.d.ts

And I can see TS complaining about the types in the `eslint.config.mjs`.

Please address these issues and I will do a second review then.

packages/react-native-reanimated/src/css/component/createAnimatedComponent.tsx
skusnierz: yeah, I forgot to add `"react/jsx-uses-react": "error",` rule

.github/workflows/monorepo-static-checks.yml
skusnierz: Without these changes, CI pipeline will fail

apps/common-app/src/apps/reanimated/eslint.config.mjs
```diff
@@ -0,0 +1,86 @@
+import jsEslint from '@eslint/js';
+import tsEslint from 'typescript-eslint';
+// @ts-expect-error
```
MatiPl01: Maybe create the `eslint.d.ts` file somewhere in the root directory with type declarations for all plugins that don't supply their own type declarations?

It should contain these declarations for each plugin that doesn't include its own types:

```
declare module '<plugin-name>' {
  import type { ESLint } from 'eslint';

  const plugin: ESLint.Plugin;
  export default plugin;
}
```

packages/eslint-plugin-reanimated/package.json
```diff
@@ -14,11 +14,11 @@
   "devDependencies": {
```
MatiPl01, line 14: Please change `"types": "lib"` to `"types": "types"` here to fix error shown by TS near reaniamted eslint plugin imports.

packages/eslint-plugin-reanimated/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -1,10 +1,9 @@
 import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
```
MatiPl01, line 1: I can see that `recommended: 'recommended'` is not valid anymore in the `meta.docs` field in the rule config and TS complains about this field. I know that I asked you why it is commented out before but it seems that this field is no longer valid in v9.

Can you maybe remove this filed in such a case to prevent TS from showing errors?

eslint.config.mjs
```diff
@@ -0,0 +1,247 @@
...
+
+/** @type {import('typescript-eslint').ConfigWithExtends[]} */
+const reanimatedConfig = [
+  {
+    files: [
+      '**/react-native-reanimated/**/*.ts',
+      '**/react-native-reanimated/**/*.tsx',
+      '**/react-native-worklets/**/*.ts',
+      '**/react-native-worklets/**/*.tsx',
+    ],
```
MatiPl01: Do we really have to specify these files manually? I would prefer to either get rid of these paths matches completely or, if this eslitn config is conflicting with other configs (for example app or docs) and you add these paths for purpose, I would like to just match all files under `react-native-reanimated` and `react-native-worklet` packages without specifying extensions manually. Plugins should work for proper file types without additional configuration and I think we may want to lint js/json files as well (if we have any other types than ts/tsx).

packages/eslint-plugin-reanimated/src/noAnimatedStyleToNonAnimatedComponent.ts
```diff
@@ -53,7 +52,11 @@ const rule: TSESLint.RuleModule<'animatedStyle' | 'sharedValue'> = {
           }
 
           // @ts-expect-error TODO: FIX ME
-          const styleExpression = styleValue.expression;
+          const styleExpression = styleValue.expression as
+            | TSESTree.Identifier
+            | TSESTree.ArrayExpression
+            | TSESTree.ObjectExpression
+            | TSESTree.MemberExpression;
```
MatiPl01: Is this type cast required here? Since there is still `@ts-expect-error`, this type override seems to have no effect.

eslint.config.mjs
```diff
@@ -0,0 +1,247 @@
+import jsEslint from '@eslint/js';
+import tsEslint from 'typescript-eslint';
+import { globalIgnores } from 'eslint/config';
+import jsdoc from 'eslint-plugin-jsdoc';
+import tsdoc from 'eslint-plugin-tsdoc';
+import simpleImportSort from 'eslint-plugin-simple-import-sort';
+// @ts-ignore no types for this plugin
```
MatiPl01: I think that you can get rid of these all ts ignores or ts expect errors by creating a top-level declarations file like `eslint.d.ts`. I created this one myself and it seems to fix all of these type issues and, at least to me, is a cleaner than just ignoring types:

```ts
type Globals = Record<string, boolean | 'readonly' | 'writable'>;

declare module 'eslint-plugin-react-native' {
  const plugin: {
    environments: {
      'react-native': Record<string, Globals>;
    };
  };

  export default plugin;
}

declare module 'eslint-plugin-no-inline-styles' {
  import type { plugin } from 'typescript-eslint';

  export default plugin;
}

declare module 'eslint-plugin-import' {
  import type { plugin } from 'typescript-eslint';

  const modifiedPlugin: typeof plugin & {
    flatConfigs: {
      recommended: ConfigWithExtends;
      typescript: ConfigWithExtends;
    };
  };

  export default modifiedPlugin;
}

declare module 'eslint-plugin-promise' {
  import type { plugin } from 'typescript-eslint';

  const modifiedPlugin: typeof plugin & {
    configs: {
      'flat/recommended': ConfigWithExtends;
    };
  };

  export default modifiedPlugin;
}

declare module 'eslint-plugin-jsdoc' {
  import type { plugin } from 'typescript-eslint';

  const modifiedPlugin: typeof plugin & {
    configs: {
      'flat/recommended': ConfigWithExtends;
    };
  };

  export default modifiedPlugin;
}

declare module 'globals' {
  const globals: {
    node: Globals;
  };
  export default globals;
}
```

You may improve these type declaration if you want but I think that this one should be sufficient.

apps/common-app/src/apps/reanimated/examples/CubesExample.tsx
```diff
@@ -247,7 +246,6 @@ function CubeWithQuaternions() {
   });
 
   const sidesStyles = sidesRotations.map((rotation, i) =>
-    // eslint-disable-next-line react-hooks/rules-of-hooks
```
MatiPl01: I think that this plugin doesn't work anymore and thus it doesn't report any issues (at least in the example app as it seems to work in the package sources). Can you make sure that the plugin works?

.github/workflows/reanimated-static-checks.yml
```diff
@@ -29,6 +29,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           cache: 'yarn'
+          node-version: 22
```
tjzel: Why?

.github/workflows/babel-plugin-static-check.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           cache: 'yarn'
+          node-version: 22
```
tjzel: Why?

apps/common-app/eslint.config.mjs
```diff
@@ -0,0 +1,337 @@
+import jsEslint from '@eslint/js';
+import tsEslint from 'typescript-eslint';
+import perfectionist from 'eslint-plugin-perfectionist';
+// @ts-ignore
+import reactNative from 'eslint-plugin-react-native';
+import react from 'eslint-plugin-react';
+// @ts-ignore
+import globals from 'globals';
+// @ts-ignore
+import importPlugin from 'eslint-plugin-import';
```
tjzel: Why ignores?

apps/common-app/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "build": "yarn patch-package",
-    "lint": "eslint .",
+    "lint": "yarn eslint . --flag unstable_config_lookup_from_file && yarn prettier --check .",
```
MatiPl01: Should we run prettier at the same time as eslint?

apps/common-app/scripts/dependencies.js
```diff
@@ -41,6 +26,21 @@ function getDependencies(currentAppDir = '.', exclude = []) {
...
+  return Object.fromEntries(
+    Object.keys(dependencies)
+      .filter((name) => !exclude.has(name))
+      .map((name) => [
+        name,
+        { root: path.resolve(__dirname, `../../../node_modules/${name}`) },
+      ])
+  );
+}
+
```
tjzel: Why?

apps/common-app/src/apps/reanimated/App.tsx
```diff
@@ -1,5 +1,6 @@
 import './types';
 
+import { useEffect } from 'react';
```
tjzel:
```suggestion
```

apps/common-app/src/apps/reanimated/eslint.config.mjs
tjzel: Add a reason for 'ts-expect-error'

apps/common-app/src/apps/reanimated/eslint.config.mjs
```diff
@@ -0,0 +1,86 @@
...
+import { fixupPluginRules } from '@eslint/compat';
+import { globalIgnores } from 'eslint/config';
+import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended';
+// @ts-expect-error
+import reanimated from 'eslint-plugin-reanimated';
+// @ts-expect-error
+import noInlineStyles from 'eslint-plugin-no-inline-styles';
+import react from 'eslint-plugin-react';
+
+/** @type {import('typescript-eslint').ConfigWithExtends[]} */
```
tjzel, line 12: Is there no way to extend root's config? Or is it already applied?

apps/common-app/src/apps/reanimated/examples/CubesExample.tsx
```diff
@@ -207,7 +207,6 @@ function CubeWithEulerAngles() {
...
```
tjzel: Why do you disable that? This should be picked up as an error.

apps/common-app/src/apps/reanimated/examples/OldMeasureExample.tsx
```diff
@@ -27,13 +27,11 @@ function createSharedVariables() {
   const contentHeights = indices.map(() => useSharedValue(0));
 
   const contentHeightsCopy = contentHeights;
-  // eslint-disable-next-line react-hooks/rules-of-hooks
```
tjzel: This should be picked up as an error.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/.prettierrc.js
```diff
@@ -12,4 +9,7 @@ module.exports = {
       },
     },
   ],
+  printWidth: 120, // Increase line width to make test cases more compact
+  singleQuote: true,
+  trailingComma: 'all',
```
tjzel: Why?

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/Presets.ts
```diff
@@ -54,10 +54,7 @@ const NOT_NUMBERS = [Infinity, -Infinity, NaN];
 // #region strings
 const TYPICAL_STRINGS = ['Aaaaaaa\n \t\t \v aaaaaa', 'Super long'.repeat(10000000), '', 'A string primitive'];
 
-const STRING_OBJECTS = [
-  // eslint-disable-next-line no-new-wrappers
```
tjzel, line 58: This should be picked up as an error.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/Presets.ts
```diff
@@ -157,7 +153,6 @@ const DATES = [
   new Date(1999, 120, 17, 33, 54, 12),
 ];
 
-// eslint-disable-next-line prefer-regex-literals
```
tjzel: This should be picked up as an error.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/TestRunner/TestRunner.ts
```diff
@@ -79,7 +79,6 @@ export class TestRunner {
   }
 
   public useTestRef(name: string): MutableRefObject<Component | null> {
-    // eslint-disable-next-line react-hooks/rules-of-hooks
```
tjzel: This should be picked up as an error.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/matchers/Matchers.ts
```diff
@@ -14,7 +14,7 @@ import { compareSnapshots } from './snapshotMatchers';
 
 export class Matchers {
   private _negation = false;
-  // eslint-disable-next-line no-useless-constructor
```
tjzel, line 17: Why?

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/utils/stringFormatUtils.ts
```diff
@@ -77,11 +77,11 @@ export function getColorSquare(color: unknown) {
     return '??';
   }
   const colorNumber = processColor(color) as number;
-  /* eslint-disable no-bitwise */
```
tjzel: This should be picked up as an error.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/utils/util.ts
```diff
@@ -6,7 +6,6 @@ export function convertDecimalColor(color: unknown): string {
     return `${color?.toString()}`;
   }
 
-  // eslint-disable-next-line no-bitwise
```
tjzel: This should be picked up as an error.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/core/useSharedValue/assigningObjects.test.tsx
```diff
@@ -53,7 +53,7 @@ describe('Test setting different values as sharedValue', () => {
     async testedValue => {
       await expect(async () => {
         await render(<SharedValueComponent initialValue={testedValue} progress={0} />);
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
tjzel: This should be picked up as an error.

apps/common-app/src/apps/reanimated/examples/ShareableFreezingExample.tsx
```diff
@@ -118,7 +118,6 @@ function tryModifyConvertedRegExpLiteral() {
 }
 
 function tryModifyConvertedRegExpInstance() {
-  // eslint-disable-next-line prefer-regex-literals
```
tjzel: This should be picked up as an error.

tjzel (changes requested):
Why are so many errors now not registered as errors?

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/matchers/Matchers.ts
skusnierz: looks like the set of recommended rules has changed... 
Therefore, in several places when I executed `eslint --fix` the `//eslint-error` comments were deleted.
I have added them manually, but this `no-useless-constructor` rule doesn't work, I have tried various ways but it doesn't detect that there is an error, imo this rule doesn't work correctly with eslint v9 but I have not found any issue on gh 😕

apps/common-app/src/apps/reanimated/eslint.config.mjs
skusnierz: these configs are not common because they have different plugins and rules, if I extend these rules with the rules from root I get a lot of errors

.github/workflows/babel-plugin-static-check.yml
tjzel: Please restore 🙏

.github/workflows/monorepo-static-checks.yml
tjzel: Please restore this CI to it's previous functionality

apps/common-app/scripts/dependencies.js
```diff
@@ -29,9 +29,9 @@ function getDependencies(currentAppDir = '.', exclude = []) {
   const currentAppPkg = require(path.resolve(currentAppDir, 'package.json'));
 
   const excludedDependencies = new Set([
-    ...Object.keys(currentAppPkg.devDependencies),
-    ...Object.keys(currentAppPkg.dependencies),
     ...exclude,
+    ...Object.keys(currentAppPkg.dependencies),
+    ...Object.keys(currentAppPkg.devDependencies),
```
tjzel: Please restore

apps/common-app/src/apps/css/components/misc/ConfigSelector.tsx
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
tjzel: Please restore this rule

apps/common-app/src/apps/css/utils/code.ts
```diff
@@ -1,6 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
-/* eslint-disable @typescript-eslint/no-unsafe-call */
```
tjzel: Please restore these rules

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/Presets.ts
```diff
@@ -140,7 +138,6 @@ const BUFFER_ARRAYS = [new ArrayBuffer(8), new ArrayBuffer(0)];
 
 const EMPTIES = [[], null, undefined, {}, [[]], [{}], [null]];
 
-// eslint-disable-next-line symbol-description
```
tjzel: Please restore

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/RuntimeTestsRunner.tsx
```diff
@@ -69,7 +69,7 @@ export default function RuntimeTestsRunner({ tests }: RuntimeTestRunnerProps) {
   function handleStartClick() {
     testSelectionCallbacks.current.forEach(callback => callback());
     setStarted(true);
-    // eslint-disable-next-line no-void
```
tjzel: Please restore

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/types.ts
```diff
@@ -120,7 +120,6 @@ export type Mismatch = {
   capturedSnapshot: OperationUpdate;
 };
 
-/* eslint-disable no-var */
```
tjzel, line 123: Please restore

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/utils/stringFormatUtils.ts
tjzel: Please restore original state

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -439,7 +441,6 @@ function CyclicObjectDemo() {
       runOnUI(() => {
         'worklet';
         try {
-          // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
tjzel: Something is yes!!!

apps/common-app/src/apps/reanimated/examples/WorkletFactoryCrashExample.tsx
```diff
@@ -7,7 +7,7 @@ export default function App() {
     function badWorklet() {
       'worklet';
       // @ts-expect-error This is an expected error
-      // eslint-disable-next-line no-unused-expressions
```
tjzel, line 10: Please restore 🙏

apps/next-example/pages/_document.js
```diff
@@ -1,4 +1,4 @@
-import { Children } from 'react';
+ import { Children } from 'react';
```
tjzel: 😠 

```suggestion
import { Children } from 'react';
```

apps/tvos-example/Gemfile.lock
tjzel, line 1: Gemfile.lock shouldn't be affected by these changes.

package.json
```diff
@@ -18,25 +18,42 @@
...
+    "build:reanimated": "yarn workspace react-native-reanimated run build",
     "build": "husky && yarn workspaces foreach --all --parallel --topological-dev run build",
     "format": "yarn format:root && yarn format:workspaces",
     "format:root": "prettier --write --list-different --ignore-path .gitignore --ignore-path .prettierignore --ignore-path .prettiereslintignore .",
     "format:workspaces": "yarn workspaces foreach --all --exclude \"react-native-reanimated-monorepo\" --parallel --topological-dev run format",
     "lint": "yarn lint:root && yarn lint:workspaces",
     "lint:workspaces": "yarn workspaces foreach --all --exclude \"react-native-reanimated-monorepo\" --parallel --topological-dev run lint",
-    "lint:root": "eslint --ignore-path .gitignore --ignore-path .eslintignore --ignore-path .prettiereslintignore . && yarn prettier --check --ignore-path .gitignore --ignore-path .prettierignore --ignore-path .prettiereslintignore ."
+    "lint:root": "eslint --flag unstable_config_lookup_from_file . && yarn prettier --check --ignore-path .gitignore --ignore-path .prettierignore --ignore-path .prettiereslintignore ."
```
tjzel: Please make it so this command lints all directories in the root but doesn't recurse into apps/ or packages/

package.json
```diff
@@ -18,25 +18,42 @@
   },
   "private": true,
   "scripts": {
+    "build:worklets": "yarn workspace react-native-worklets run build",
+    "build:reanimated": "yarn workspace react-native-reanimated run build",
```
tjzel:
```suggestion
```

package.json
```diff
@@ -18,25 +18,42 @@
...
     "eslint-plugin-simple-import-sort": "^12.1.1",
     "husky": "^9.0.11",
     "lint-staged": "^15.2.2",
     "node-fetch": "^3.3.2",
+    "prettier": "^3.5.3",
+    "prettier-plugin-jsdoc": "^1.3.2",
+    "typescript": "~5.3.0",
+    "typescript-eslint": "^8.34.0"
+  },
+  "resolutions": {
```
tjzel: Please remove `resolutions`

packages/eslint-plugin-reanimated/tsconfig.json
```diff
@@ -15,5 +15,5 @@
     "declaration": true,
     "declarationDir": "types"
   },
-  "include": ["src"]
+  "include": ["src", "types"]
```
tjzel: Why?

packages/react-native-reanimated/__typetests__/common/FlatListTest.tsx
```diff
@@ -1,8 +1,8 @@
-/* eslint-disable no-unused-expressions */
-/* eslint-disable @typescript-eslint/no-empty-function */
```
tjzel, line 2: Please restore if still applicable.

packages/react-native-reanimated/__typetests__/common/useAnimatedPropsTest.tsx
```diff
@@ -1,7 +1,6 @@
-/* eslint-disable no-unused-expressions */
-/* eslint-disable @typescript-eslint/no-empty-function */
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 /* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable react/jsx-no-undef */
```
tjzel: Why?

packages/react-native-reanimated/__typetests__/common/useComposedEventHandlerTest.tsx
```diff
@@ -1,6 +1,3 @@
-/* eslint-disable no-unused-expressions */
-/* eslint-disable @typescript-eslint/no-empty-function */
-/* eslint-disable @typescript-eslint/ban-ts-comment */
```
tjzel, line 3: Please restore.

packages/react-native-reanimated/src/Easing.ts
```diff
@@ -1,6 +1,6 @@
 'use strict';
 import { Bezier } from './Bezier';
-import type { EasingFunction } from './commonTypes';
+import type { EasingFunction, EasingFunctionFactory } from './commonTypes';
```
tjzel: Please restore it and make a separate PR that restricts circular checks only to generated `.js` code to allow circular type imports.

packages/react-native-reanimated/src/ViewDescriptorsSet.ts
```diff
@@ -1,13 +1,9 @@
...
-import type { SharedValue } from './commonTypes';
 import { makeMutable } from './core';
-import type { Descriptor } from './hook/commonTypes';
-
-export interface ViewDescriptorsSet {
-  shareableViewDescriptors: SharedValue<Descriptor[]>;
-  add: (item: Descriptor) => void;
-  remove: (viewTag: number) => void;
-}
+import type {
```
tjzel: Please restore.

packages/react-native-reanimated/src/animation/timing.ts
```diff
@@ -4,10 +4,10 @@ import type {
   Animation,
   AnimationCallback,
   EasingFunction,
+  EasingFunctionFactory,
   ReduceMotion,
   Timestamp,
 } from '../commonTypes';
-import type { EasingFunctionFactory } from '../Easing';
```
tjzel: Please restore.

packages/react-native-reanimated/src/core.ts
```diff
@@ -65,7 +65,7 @@ export function getViewProp<T>(
       component,
       (result: T) => {
         if (typeof result === 'string' && result.substr(0, 6) === 'error:') {
-          reject(result);
+          reject(new ReanimatedError(result));
```
tjzel: Does it work in runtime? Maybe it'd better to leave as is and add an eslint ignore comment.

packages/react-native-reanimated/src/createAnimatedComponent/commonTypes.ts
```diff
@@ -10,7 +10,18 @@ import type {
...
+export interface Descriptor {
+  tag: number | ReanimatedHTMLElement;
+  shadowNodeWrapper: ShadowNodeWrapper;
+}
+
+export interface ViewDescriptorsSet {
+  shareableViewDescriptors: SharedValue<Descriptor[]>;
+  add: (item: Descriptor) => void;
+  remove: (viewTag: number) => void;
+}
```
tjzel: Please restore.

packages/react-native-reanimated/src/css/types/helpers.ts
```diff
@@ -13,7 +13,6 @@ export type Repeat<
 
 export type Simplify<T> = {
   [K in keyof T]: T[K];
-  // eslint-disable-next-line @typescript-eslint/ban-types
 } & {};
```
tjzel, line 16:
```suggestion
} & object;
```

packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
```diff
@@ -576,7 +580,6 @@ For more, see the docs: \`https://docs.swmansion.com/react-native-reanimated/doc
     return () => {
       stopMapper(mapperId);
     };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tjzel: Please restore.

packages/react-native-reanimated/src/platform-specific/findHostInstance.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable camelcase */
```
tjzel: Please restore.

packages/react-native-reanimated/src/platformFunctions/setNativeProps.ts
```diff
@@ -41,7 +41,7 @@ function setNativePropsNative(
   }
   const shadowNodeWrapper = animatedRef() as ShadowNodeWrapper;
   processColorsInProps(updates);
-  global._updateProps!([{ shadowNodeWrapper, updates }]);
+  global._updateProps?.([{ shadowNodeWrapper, updates }]);
```
tjzel: Using `!` is an optimization to avoid unnecessary runtime checks, please restore.

packages/react-native-reanimated/src/updateProps/updateProps.ts
```diff
@@ -86,7 +86,7 @@ function createUpdatePropsManager() {
       });
     },
     flush(this: void) {
-      global._updateProps!(operations);
+      global._updateProps?.(operations);
```
tjzel: Same here

packages/react-native-worklets/src/WorkletsError.ts
```diff
@@ -1,9 +1,10 @@
-/* eslint-disable reanimated/use-worklets-error */
 'use strict';
 
 function WorkletsErrorConstructor(message?: string): WorkletsError {
   'worklet';
   const prefix = '[Worklets]';
+
+  // eslint-disable-next-line reanimated/use-reanimated-error
```
tjzel: Please restore.

scripts/format-java.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
```
tjzel, line 1: Please restore.

packages/react-native-reanimated/src/css/types/helpers.ts
MatiPl01: Unfortunately, I won't agree here. This won't produce the same result.

| with `{}` | with `object` |
|-|-|
| Screenshot 2025-06-24 at 10 28 14 | Screenshot 2025-06-24 at 10 28 04 |

MatiPl01:
```ts
type C = Simplify<A & B>;

const test = {} as C;
```

tjzel: Why do you need to use the `{}` type though? It's really obscure.

MatiPl01: I found this helper type somewhere on the Internet. I don't really know how `{}` works but I know that it somehow works 😆. It is getting the job done, so I would like to keep using it for more user-friendly types.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/types.ts
skusnierz: I don't know why, but eslint says it's correct : 
image

packages/react-native-reanimated/__typetests__/common/useAnimatedPropsTest.tsx
skusnierz: image

packages/react-native-reanimated/__typetests__/common/useComposedEventHandlerTest.tsx
skusnierz: these rules are not needed here

packages/react-native-reanimated/__typetests__/common/FlatListTest.tsx
skusnierz: not applicable

scripts/format-java.js
skusnierz: it is no longer needed

packages/react-native-reanimated/src/css/types/helpers.ts
tjzel: 🤯

apps/common-app/src/apps/css/utils/code.ts
```diff
@@ -95,7 +95,6 @@ const stringifyConfigObject = <T extends AnyRecord>(
 
   return `{\n${keys
     .map((key) => {
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
skusnierz, line 98: It is no longer needed, `@typescript-eslint/no-unsafe-assignment` rule is enabled

packages/react-native-reanimated/__typetests__/common/useAnimatedPropsTest.tsx
```diff
@@ -1,5 +1,4 @@
-/* eslint-disable no-unused-expressions */
-/* eslint-disable @typescript-eslint/no-empty-function */
+/* eslint-disable react/jsx-no-undef */
```
skusnierz: image
```
/* eslint-disable no-unused-expressions */
/* eslint-disable @typescript-eslint/no-empty-function */
```
not needed anymore

packages/react-native-reanimated/__typetests__/common/useEventTest.tsx
```diff
@@ -1,7 +1,6 @@
-/* eslint-disable no-unused-expressions */
-/* eslint-disable @typescript-eslint/no-empty-function */
-/* eslint-disable @typescript-eslint/ban-ts-comment */
```
skusnierz, line 3: not needed

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/RuntimeTestsRunner.tsx
```diff
@@ -51,6 +51,7 @@ export default function RuntimeTestsRunner({ tests }: RuntimeTestRunnerProps) {
 
   useEffect(() => {
     tests.forEach(testData => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
tjzel: Please set `allowShortCircuit` to true for this rule and don't use `eslint-disable-next-line` for it 🙏

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/types.ts
tjzel: Seems like they can finally detect correct use of `var` 👍

apps/fabric-example/ios/Podfile.lock
tjzel, line 1: Podfile.locks shouldn't change with this PR, please revert.

apps/tvos-example/Gemfile.lock
tjzel, line 1: Gemfile.locks shouldn't change with this PR, please restore 🙏

package.json
```diff
@@ -20,17 +20,26 @@
...
     "format": "yarn format:root && yarn format:workspaces",
-    "format:root": "prettier --write --list-different --ignore-path .gitignore --ignore-path .prettierignore --ignore-path .prettiereslintignore .",
+    "format:root": "prettier --cache --write --list-different --ignore-path .gitignore --ignore-path .prettierignore --ignore-path .prettiereslintignore .",
     "format:workspaces": "yarn workspaces foreach --all --exclude \"react-native-reanimated-monorepo\" --parallel --topological-dev run format",
     "lint": "yarn lint:root && yarn lint:workspaces",
     "lint:workspaces": "yarn workspaces foreach --all --exclude \"react-native-reanimated-monorepo\" --parallel --topological-dev run lint",
-    "lint:root": "eslint --ignore-path .gitignore --ignore-path .eslintignore --ignore-path .prettiereslintignore . && yarn prettier --check --ignore-path .gitignore --ignore-path .prettierignore --ignore-path .prettiereslintignore ."
+    "lint:root": "eslint . --ignore-pattern 'packages/**' --ignore-pattern 'apps/**' && yarn prettier --check --ignore-path .gitignore --ignore-path .prettierignore --ignore-path .prettiereslintignore ."
```
tjzel, line 27: Can we pull ignore patterns from a file?

packages/docs-reanimated/.eslintrc.js
tjzel, line 1: This PR should have no effect on `docs-reanimated`, please revert.

packages/docs-reanimated/yarn.lock
tjzel, line 1: This PR should have no effect on `docs-reanimated`, please revert.

packages/react-native-worklets/plugin/src/class.ts
```diff
@@ -294,6 +294,7 @@ function recursiveTopoSort(
   stack: Set<string>
 ) {
   if (stack.has(current.name)) {
+    // eslint-disable-next-line reanimated/use-worklets-error
```
tjzel: `react-native-worklets/plugin` shouldn't have the rule for `WorkletsError` enabled. Perhaps it needs a separate eslint config.

apps/common-app/src/apps/css/utils/code.ts
MatiPl01: It is not needed here as I disabled this rule for the entire file and forgot to remove it from here

apps/common-app/src/apps/reanimated/examples/LayoutAnimations/ViewRecyclingExample.tsx
```diff
@@ -49,6 +49,7 @@ const OptionInput = ({ onChange }: OptionProps) => {
 
   const handleSelect = (option: Option) => {
     setActiveOption(option);
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
MatiPl01: I think it shouldn't work in this way here. The issue seems to be caused by:

```ts
type OptionProps = {
  onChange: (option: Option) => void;
};
```

because it shows that `option` param is unused, which is strange, as it is just a type declaration. Something must be wrong with the current plugin setup.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/matchers/Matchers.ts
```diff
@@ -14,7 +14,7 @@ import { compareSnapshots } from './snapshotMatchers';
...
+
   constructor(
     private _currentValue: TestValue,
     private _testCase: TestCase,
```
MatiPl01, line 20: It shows 

```
'_testCase' is defined but never used.eslintno-unused-vars`
```

but I think that vars with `_` at the beginning should be ignored. Seems that this rule is not configured properly for the `common-app`

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/utils/stringFormatUtils.ts
```diff
@@ -82,6 +82,7 @@ export function getColorSquare(color: unknown) {
   const green = (colorNumber >> 8) & 255;
   const blue = colorNumber & 255;
   /* eslint-enable no-bitwise */
+
```
MatiPl01: unnecessary line break

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/RuntimeTestsRunner.tsx
MatiPl01: You can also do a little trick here and use this instead:

```ts
tests.forEach(testData => !testData.skipByDefault && testSelectionCallbacks.current.add(testData.importTest));
```

as one-liner. This expression will be considered as used because it is returned from the `forEach` function.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/types.ts
MatiPl01: Don't know why but it shows me warnings like this:

```
'callback' is defined but never used.eslintno-unused-vars
```

near every variable in every callback function type declaration

apps/common-app/src/apps/reanimated/examples/ChatHeadsExample.tsx
```diff
@@ -92,6 +92,7 @@ function ChatHeads({
     <>
       {childrenArray.length > 1 && (
         <Followers
+          // eslint-disable-next-line react/no-children-prop
```
MatiPl01: Can it be changed to:

```
<Followers transX={transX} transY={transY}>
    {childrenArray.slice(1)}
</Followers>
```

?

apps/common-app/src/apps/reanimated/examples/PerfomanceMonitorExample.tsx
```diff
@@ -17,10 +17,13 @@ enum Examples {
 export default function PerformanceMonitorExample() {
   const exampleElements = useRef(
     new Map<Examples, JSX.Element>([
-      [Examples.Empty, <EmptyExample />],
-      [Examples.Bokeh, <BokehExample />],
-      [Examples.Planets, <PlanetsExample />],
-      [Examples.Emojis, <EmojiWaterfallExample />],
+      [Examples.Empty, <EmptyExample key="empty-example" />],
```
MatiPl01, line 20: I think it would be safe to disable this eslint rule that checks for key existence here but adding keys is also correct, so that's up to you which approach you decide to leave.

apps/common-app/src/apps/reanimated/examples/WorkletFactoryCrashExample.tsx
```diff
@@ -7,7 +7,7 @@ export default function App() {
...
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
MatiPl01, line 10: Why do you use `no-unused-expressions` instead of `@typescript-eslint/no-unused-expressions`? I think that it causes problems in some files such as this one: https://github.com/software-mansion/react-native-reanimated/pull/7570#discussion_r2166918794

apps/common-app/src/apps/reanimated/examples/WorkletRuntimeExample.tsx
```diff
@@ -76,6 +76,7 @@ function CreateWorkletRuntimeDemo() {
     console.log(runtime.name);
     // eslint-disable-next-line @typescript-eslint/no-base-to-string, @typescript-eslint/restrict-template-expressions
     console.log(`${runtime}`);
+    // eslint-disable-next-line @typescript-eslint/no-base-to-string, @typescript-eslint/restrict-template-expressions
```
MatiPl01: Are these really needed here? It didn't show any warning on my side when I removed the comment.

apps/common-app/src/apps/reanimated/eslint.config.mjs
```diff
@@ -0,0 +1,95 @@
...
+import react from 'eslint-plugin-react';
+import reactHooks from 'eslint-plugin-react-hooks';
+
+/** @type {import('typescript-eslint').ConfigWithExtends[]} */
+export default tsEslint.config(
+  jsEslint.configs.recommended,
+  react.configs.flat.recommended,
+  eslintPluginPrettierRecommended,
+  reactHooks.configs['recommended-latest'],
+  {
```
MatiPl01, line 17: Can this and the following configuration object be combined together? I don't understand why there are used 2 separate configs, first for all files and the second one with `.ts` and `.tsx` file extension specifiers.

apps/common-app/src/components/LeakCheck.tsx
```diff
@@ -4,7 +4,6 @@ import { Button, Text, View } from 'react-native';
 export const NukeContext = createContext<() => void>(() => '');
 
 function registriesLeakCheck(): string {
-  // eslint-disable-next-line no-underscore-dangle
```
MatiPl01: Please restore, it shows a warning without this

apps/common-app/eslint.config.mjs
```diff
@@ -0,0 +1,333 @@
...
+          project: './tsconfig.json',
+        },
+      },
+      react: {
+        version: 'detect',
+      },
+    },
+  },
+  {
+    files: ['**/*.ts', '**/*.tsx'],
```
MatiPl01, line 279: Again, do we really need to declare some rules separately for `ts` and `tsx` files? If yes, please explain why.

apps/common-app/tsconfig.json
```diff
@@ -3,8 +3,9 @@
   "compilerOptions": {
     "baseUrl": ".",
     "paths": {
-      "@/*": ["src/*"]
+      "@/*": ["./src/*"]
     }
   },
-  "include": ["src", ".eslintrc.js", "index.ts"]
+  "include": ["src", "index.ts"],
+  "exclude": ["**/*eslint.config.mjs"]
```
MatiPl01:
```suggestion
  "exclude": ["**/eslint.config.mjs"]
```

apps/next-example/pages/test.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable jsdoc/require-jsdoc */
+
```
MatiPl01: unnecessary change

packages/react-native-reanimated/__tests__/bezier.test.ts
```diff
@@ -252,5 +252,6 @@ describe('Test `Bezier` function', () => {
     });
   });
 
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
   describe('Bezier(a,b,(1-a),(1-b)) should have point symmetry at point (0.5, 0.5)', () => {});
 });
```
MatiPl01, line 258: Maybe remove this test if it is empty or add a TODO comment to implement it?

packages/react-native-reanimated/src/ReanimatedModule/js-reanimated/JSReanimated.ts
```diff
@@ -37,6 +37,7 @@ class JSReanimated implements IReanimatedModule {
    * We keep the instance of `WorkletsModule` here to keep correct coupling of
    * the modules and initialization order.
    */
+  // eslint-disable-next-line no-unused-private-class-members
   #workletsModule: IWorkletsModule = WorkletsModule;
```
MatiPl01, line 41: I'd suggest using `private` modifier from TS and removing `#` from the variable name, but I know that @tjzel likes to use truly private JS class members and might not agree with me.

packages/react-native-reanimated/src/ReanimatedModule/NativeReanimated.ts
```diff
@@ -55,9 +56,9 @@ class NativeReanimatedModule implements IReanimatedModule {
    * We keep the instance of `WorkletsModule` here to keep correct coupling of
    * the modules and initialization order.
    */
+  // eslint-disable-next-line no-unused-private-class-members
```
MatiPl01, line 59: Similar here, I would just use `private`, since this `NativeReanimatedModule` is never exposed outside of the reanimated package and we use TS, so marking field as private via TS modifier would be enough

packages/react-native-reanimated/src/css/platform/native/config.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
MatiPl01: Maybe disable just for a single line? I think it would be better, since only a single prop uses invalid casing.

packages/react-native-reanimated/src/css/platform/web/config.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
MatiPl01: Same here, disable just for one line

packages/react-native-worklets/plugin/src/class.ts
MatiPl01: Why not? Wasn't it added specifically for the worklets package? Reanimated package uses its own `useReanimatedError` eslint rule.

MatiPl01 (commented):
I left some comments.

I can also see that TS complains about types in the eslint.config.mjs and this PR introduced 2 circular dependencies in the reanimated package:

1) lib/typescript/commonTypes.d.ts > lib/typescript/Easing.d.ts
2) lib/typescript/hook/commonTypes.d.ts > lib/typescript/ViewDescriptorsSet.d.ts

Please address these issues and I will do a second review

MatiPl01 (commented):
Left a few more comments

packages/react-native-reanimated/__typetests__/common/useAnimatedPropsTest.tsx
tjzel: Please fix the function that forgot to define this JSX element instead of the `react/jsx-no-undef` 🙏

package.json
skusnierz: unfortunately, but `--ignore-path` is no longer available

```
Ignore Files:
  --no-ignore                      Disable use of ignore files and patterns
  --ignore-pattern [String]        Patterns of files to ignore
```

packages/react-native-worklets/plugin/src/class.ts
tjzel: This is the plugin, it doesn't work in runtime and it has it's own decorated error try-catch

https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-worklets/plugin/src/plugin.ts#L25

packages/react-native-reanimated/src/ReanimatedModule/js-reanimated/JSReanimated.ts
tjzel: I like truly private 😎

packages/react-native-worklets/plugin/src/class.ts
skusnierz: ok, I'll add a separate config

apps/common-app/src/apps/reanimated/examples/LayoutAnimations/ViewRecyclingExample.tsx
skusnierz: yeah, this additional options to `@typescript-eslint/no-unused-expressions ` resolve this problem : 
```
      '@typescript-eslint/no-unused-expressions': [
        'error',
        {
          allowShortCircuit: true,
          allowTernary: true,
        },
      ],
```

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/matchers/Matchers.ts
skusnierz: My IDE doesn't complaint about it :  
image

apps/common-app/eslint.config.mjs
skusnierz: just wanted to not change the old config, combining this together causes new eslint errors, but I will try to correct them

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/matchers/Matchers.ts
MatiPl01: let me check again. I will reinstall all deps and make sure that this is not just an issue on my side

MatiPl01: Well, on my side it still shows this warning. Maybe let someone else confirm that this issue is related to my local setup and something is wrong on my end

Screenshot 2025-06-25 at 20 05 57

apps/common-app/src/apps/reanimated/examples/WorkletFactoryCrashExample.tsx
MatiPl01: Ok, never mind, I thought that it was the opposite and you changed `@typescript-eslint/no-unused-expressions` to `no-unused-expressions`.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/ReJest/matchers/Matchers.ts
skusnierz: ok, so I cloned the repo and set up everything from scratch, and now I have this warning : 
image
But if it's a warning, should we be concerned about this ?

apps/common-app/src/components/LeakCheck.tsx
skusnierz: same here, it's just a warning

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/TestsOfTestingFramework.test.tsx
```diff
@@ -25,6 +25,34 @@ import {
 import { ComparisonMode } from '../ReJest/types';
 import { Snapshots } from './TestsOfTestingFramework.snapshot';
 
+const styles = StyleSheet.create({
```
MatiPl01: Maybe move it below the component implementation to align with the styling convention?

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/TestsOfTestingFramework.test.tsx
```diff
@@ -25,6 +25,34 @@ import {
...
+  },
+  layoutAnimationComponent: {
+    top: 41,
+    left: 42,
+    width: 50,
+    height: 50,
+    backgroundColor: 'chocolate',
+    margin: 10,
+  },
+});
```
tjzel: Please move it to the bottom of the file 🙏

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/core/useSharedValue/components.tsx
```diff
@@ -9,14 +9,20 @@ export enum MutableAPI {
...
-          backgroundColor: 'darkorange',
-        }}
-      />
+    <View style={styles.container}>
+      <View style={[styles.progressBar, { width: `${100 * progress}%` }]} />
     </View>
   );
 };
+
+export const styles = StyleSheet.create({
```
tjzel:
```suggestion
const styles = StyleSheet.create({
```

apps/common-app/src/apps/reanimated/examples/WorkletFactoryCrashExample.tsx
```diff
@@ -7,7 +7,7 @@ export default function App() {
     function badWorklet() {
       'worklet';
       // @ts-expect-error This is an expected error
-      // eslint-disable-next-line no-unused-expressions
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions, no-unused-expressions
```
tjzel: Seems like one too many

```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```

packages/react-native-reanimated/__typetests__/common/useEventTest.tsx
```diff
@@ -1,7 +1,6 @@
...
+/* eslint-disable @typescript-eslint/no-unused-expressions */
 /* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable no-unused-expressions */
```
tjzel, line 3: Seems like one too many of those `no-unused-expressions`.

apps/common-app/src/apps/reanimated/examples/WorkletFactoryCrashExample.tsx
MatiPl01: Do we really need to use both `@typescript-eslint/no-unused-expressions` and `no-unused-expressions` at the same time? Do they differ in behavior? I think that `no-unused-expressions` might sometimes not work as intedned with TS so inly the one from `@typescript-eslint` should be used.

packages/react-native-worklets/plugin/src/plugin.ts
```diff
@@ -58,6 +58,7 @@ module.exports = function WorkletsBabelPlugin(): PluginItem {
           state: ReanimatedPluginPass
         ) {
           runWithTaggedExceptions(() => {
+            // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
tjzel: Please disable this rule for this package 🙏
```suggestion
```

tjzel (commented):
Almost ready!

packages/react-native-worklets/plugin/src/plugin.ts
```diff
@@ -58,6 +58,7 @@ module.exports = function WorkletsBabelPlugin(): PluginItem {
...
             processIfWithWorkletDirective(path, state) ||
               processIfAutoworkletizableCallback(path, state);
           });
```
MatiPl01:
```suggestion
          runWithTaggedExceptions(() =>
          processIfWithWorkletDirective(path, state) ||
              processIfAutoworkletizableCallback(path, state)
          );
```
Maybe just return the result innstead of adding this eslint-disable? Seems cleaner to me.

packages/react-native-worklets/plugin/src/plugin.ts
MatiPl01: I would prefer to keep it but instead [change](https://github.com/software-mansion/react-native-reanimated/pull/7570#discussion_r2168662951) the implementation in such a way that it doesn't complain. What do you think about that?

MatiPl01 (approved):
Let it go!

I think that it works fine now. Some of issues that I reported turned out to be related to my local node version (`20.10.0`) which seems to be uncompatible. Bumping to `v22` fixed these issues.

PR #7571, opened 2025-05-21
fix: Remaining CSS issues fixes

MatiPl01:
## Summary

This PR fixes remaining CSS issues I noticed while testing example apps

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSAnimationsRegistry.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSAnimationsRegistry.cpp
@@ -274,8 +274,7 @@ void CSSAnimationsRegistry::applyViewAnimationsStyle(
 
     folly::dynamic style;
     const auto &currentState = animation->getState(timestamp);
-    if (startTimestamp == timestamp ||
-        (startTimestamp > timestamp && animation->hasBackwardsFillMode())) {
+    if (startTimestamp > timestamp && animation->hasBackwardsFillMode()) {
       style = animation->getBackwardsFillStyle();
     } else if (
         currentState == AnimationProgressState::Running ||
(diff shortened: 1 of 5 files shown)


packages/react-native-reanimated/Common/cpp/reanimated/CSS/registry/CSSAnimationsRegistry.cpp
```diff
@@ -274,8 +274,7 @@ void CSSAnimationsRegistry::applyViewAnimationsStyle(
 
     folly::dynamic style;
     const auto &currentState = animation->getState(timestamp);
-    if (startTimestamp == timestamp ||
```
MatiPl01, line 277: This was breaking animations with `0s` duration, for which we automatically returned the `getBackwardsFillStyle` which is not a valid behavior.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -697,7 +697,7 @@ void ReanimatedModuleProxy::performOperations() {
 
     if (shouldUpdateCssAnimations_) {
       currentCssTimestamp_ = getAnimationTimestamp_();
-      cssAnimationsRegistry_->lock();
+      cssTransitionsRegistry_->lock();
```
MatiPl01, line 700: I made a stupid mistake some time ago when I was moving locks outside of registers. I put the transitions lock in place of the animations one and animations lock where the transitions lock should be placed.

Somehow iOS didn't crash but I noticed multiple crashes on Android.

packages/react-native-reanimated/src/css/managers/CSSTransitionsManager.web.ts
```diff
@@ -11,6 +11,7 @@ import { kebabizeCamelCase } from '../utils';
 
 export default class CSSTransitionsManager implements ICSSTransitionsManager {
   private readonly element: ReanimatedHTMLElement;
+  private isAttached = false;
```
tjzel, line 14: How about the more private kind of private?
```suggestion
  #isAttached = false;
```

tjzel (approved):

PR #7573, opened 2025-05-22
feat: Inject reanimated module proxy to reanimated shadow node

MatiPl01:
## Summary

This PR adds a possibility to access the `ReanimatedModuleProxy` from the `ReanimatedShadowNode`. It also fixes android linking issue to make it possible to access all reanimated library stuff.

## Test plan

You can open any example in the example app and see that it renders fine. 

You can also uncomment the following line in the `packages/react-native-reanimated/Common/NativeView/react/renderer/components/rnreanimated/ComponentDescriptors.h`:

```cpp
// LOG(INFO) << "We can access the proxy: " << proxy->getCssTimestamp();
```

and see that the proxy can be accessed.

## TODO

- [ ] Fix gesture handler integration - in the current implementation gestures aren't attached to AnimatedComponents because the `ReanimatedView` is not laid out (because of `display: contents`).

### DIFF
--- a/packages/react-native-reanimated/Common/NativeView/CMakeLists.txt
+++ b/packages/react-native-reanimated/Common/NativeView/CMakeLists.txt
@@ -18,6 +18,11 @@ file(
   ${CODEGEN_OVERWRITTEN_COMPONENTS_DIR}/*.cpp
 )
 
+find_package(fbjni REQUIRED CONFIG)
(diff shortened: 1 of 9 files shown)


MatiPl01:
This is the same as #7417, which was accidentally closed when I removed the base branch.

PR #7575, opened 2025-05-22
feat(Worklets): add exp. bundling native apis and flags

tjzel:
## Summary

This pull requests adds relevant APIs for iOS `WorkletsModule.mm` and Android `WorkletsModule.cpp` - they now can receive the bundle and the source url and forward it to Worklet Runtimes. There's no change in behavior if the feature flag for experimental bundling is not enabled.

Do not enable the experimental bundling yet, as PRs that bring more required changes will be coming shortly.

## Test plan

No runtime behavior difference.

### DIFF
--- a/packages/react-native-worklets/apple/worklets/apple/WorkletsModule.mm
+++ b/packages/react-native-worklets/apple/worklets/apple/WorkletsModule.mm
@@ -8,8 +8,10 @@
 #import <worklets/apple/WorkletsMessageThread.h>
 #import <worklets/apple/WorkletsModule.h>
 
+#import <React/NSDataBigString.h>
 #import <React/RCTBridge+Private.h>
 #import <React/RCTCallInvoker.h>
+#import <jsireact/JSIExecutor.h>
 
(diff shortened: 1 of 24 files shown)


packages/react-native-worklets/apple/worklets/apple/WorkletsModule.mm
```diff
@@ -64,7 +77,13 @@ - (void)checkBridgeless
         [animationFrameQueue requestAnimationFrame:callback];
       });
   workletsModuleProxy_ = std::make_shared<WorkletsModuleProxy>(
-      rnRuntime, jsQueue, jsCallInvoker, uiScheduler, std::move(forwardedRequestAnimationFrame));
+      rnRuntime,
+      jsQueue,
+      jsCallInvoker,
+      uiScheduler,
+      std::move(forwardedRequestAnimationFrame),
+      std::move(script),
```
piaskowyk: don't move 👮🏻‍♂️

PR #7579, opened 2025-05-23
fix: Crashes by adding missing worklet directive to plain object check

MatiPl01:
## Summary

Quick fix for the issue introduced in the #7548 PR.

## Test plan

Open the drawer navigator in the fabric example app and see this crash when there is no `'worklet'` directive.

Screenshot 2025-05-23 at 17 08 53


tjzel (approved):
LGTM

PR #7582, opened 2025-05-26
fix: CSS crashes caused by invalid locks usage

MatiPl01:
## Summary

I thought I fixed this issue in the #7571 PR but it turns out that I wasn't using locks properly and they were released immediately after being acquired because I never assigned them to the local scope variable.

## Example recordings

Before the app crashed after a few drag and drops. After these changes, it works as shown:

https://github.com/user-attachments/assets/751b1e09-4021-4fa3-890f-73e014847d96

<details>
<summary>Source code</summary>

```tsx
import { useEffect, useState } from 'react';
import { Dimensions, Image, StyleSheet, Text, View } from 'react-native';
import { Gesture, GestureDetector } from 'react-native-gesture-handler';
import Animated, {
  CSSAnimationKeyframes,
  Easing,
  LinearTransition,
  runOnJS,
  useAnimatedStyle,
  useSharedValue,
  withTiming,
} from 'react-native-reanimated';
import { useSafeAreaInsets } from 'react-native-safe-area-context';

const ITEMS_IN_ROW_COUNT = 4;
const GAP = 16;

const TILE_SIZE =
  (Dimensions.get('screen').width - (ITEMS_IN_ROW_COUNT + 1) * GAP) /
  ITEMS_IN_ROW_COUNT;

const data = new Array(20).fill(0).map((_, i) => ({
  id: `App ${i}`,
  name: `App ${i}`,
  image: 'https://picsum.photos/64/64',
}));

const shake: CSSAnimationKeyframes = {
  from: {
    transform: [{ rotateZ: '2deg' }],
  },
  '15%': {
    transform: [{ rotateZ: '-2deg' }],
  },
  '30%': {
    transform: [{ rotateZ: '2deg' }],
  },
  '45%': {
    transform: [{ rotateZ: '-2deg' }],
  },
  '60%': {
    transform: [{ rotateZ: '2deg' }],
  },
  '75%': {
    transform: [{ rotateZ: '-2deg' }],
  },
  to: {
    transform: [{ rotateZ: '2deg' }],
  },
};

export default function ReorderIconsLesson() {
  const insets = useSafeAreaInsets();
  const [items, setItems] = useState(data);
  const activeItemId = useSharedValue<string | null>(null);
  const [placeholderIndex, setPlaceholderIndex] = useState<number | null>(null);
  const [isReordering, setIsReordering] = useState(false);

  const reorderItems = () => {
    if (placeholderIndex === null || activeItemId.value === null) {
      return;
    }
[...]

PR #7588, opened 2025-05-27
release: Reanimated 4.0.0-beta.5

MatiPl01:
## Summary

This PR is a release of the next beta version with a hot fix for CSS animation crashes in `4.0.0-beta.4`.

This version is still compatible with react-native-worklets `0.3.0`.

I did a quick test, mainly mobile (iOS, Android) and quick tests for web. I also tested the example from the #7582 PR which was used as a repro for CSS crashes before.

### DIFF
--- a/apps/common-app/package.json
+++ b/apps/common-app/package.json
@@ -35,7 +35,7 @@
     "react-native-safe-area-context": "5.3.0",
     "react-native-screens": "patch:react-native-screens@npm%3A4.10.0#~/.yarn/patches/react-native-screens-npm-4.10.0-a9c3a4e55e.patch",
     "react-native-svg": "15.12.0",
-    "react-native-worklets": "workspace:*",
+    "react-native-worklets": "0.3.0",
     "react-strict-dom": "0.0.27"
   },
   "devDependencies": {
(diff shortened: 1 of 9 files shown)


apps/common-app/package.json
```diff
@@ -35,7 +35,7 @@
     "react-native-safe-area-context": "5.3.0",
     "react-native-screens": "patch:react-native-screens@npm%3A4.10.0#~/.yarn/patches/react-native-screens-npm-4.10.0-a9c3a4e55e.patch",
     "react-native-svg": "15.12.0",
-    "react-native-worklets": "workspace:*",
+    "react-native-worklets": "0.3.0",
```
tomekzaw, line 38: I guess we should still use `react-native-worklets` from our workspace here, right?

apps/next-example/package.json
```diff
@@ -21,7 +21,7 @@
     "react-dom": "19.1.0",
     "react-native-reanimated": "workspace:*",
     "react-native-web": "0.20.0",
-    "react-native-worklets": "workspace:*"
+    "react-native-worklets": "0.3.0"
```
tomekzaw, line 24: Same

apps/tvos-example/package.json
```diff
@@ -15,7 +15,7 @@
     "react": "19.0.0",
     "react-native": "npm:react-native-tvos@0.79.1-0",
     "react-native-reanimated": "workspace:*",
-    "react-native-worklets": "workspace:*"
+    "react-native-worklets": "0.3.0"
```
tomekzaw, line 18: Same

apps/common-app/package.json
tomekzaw: Cool, haven't noticed the commit title, only scanned through the PR description. Thanks for explaining!

PR #7591, opened 2025-05-27
refactor(Worklets)(makeShareableClone): makeShareableFunction

skusnierz:
## Summary
This PR adds `makeShareableFunction` function and is part of a series of PRs that aim to refactor the makeSharableClone function -> https://github.com/software-mansion/react-native-reanimated/pull/7481
## Test plan
`Shareables` example

### DIFF
--- a/packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
+++ b/packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
@@ -297,6 +297,7 @@ jsi::Value JSIWorkletsModuleProxy::get(
               rt, args[0].asObject(rt).getHostObject(rt));
         });
   }
+
   if (name == "makeShareableFunction") {
     return jsi::Function::createFromHostFunction(
         rt,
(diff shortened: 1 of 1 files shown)


packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -153,6 +153,18 @@ jsi::Value makeShareableInitializer(
   return ShareableJSRef::newHostObject(rt, shareable);
 }
 
+jsi::Value makeShareableFunction(
+    jsi::Runtime &rt,
+    jsi::Function function ) {
+  std::shared_ptr<Shareable> shareable;
+  if (function.isHostFunction(rt)) {
```
skusnierz, line 160: I wanted to split this into two separate functions, but I don't know how I can check if the function is a host function on the JS side ? I did a little research but couldn't find the answer, @tjzel @tomekzaw do you have any idea ?

packages/react-native-worklets/src/shareables.ts
```diff
@@ -321,13 +321,10 @@ function cloneArray<T extends unknown[]>(
 
 function cloneRemoteFunction<T extends object>(
   value: T,
-  shouldPersistRemote: boolean
 ): ShareableRef<T> {
-  const clone = WorkletsModule.makeShareableClone(
-    value,
-    shouldPersistRemote,
```
skusnierz, line 328: We don't use it on the Cpp side so I'm removing it

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
tjzel: I don't think it's possible either. For the time being check in C++ is good.

What I think would be the next step here is to inject two host functions, one that checks if an object is a host object, and another one which checks if a function is a host function. That would reduce the amount of hacks we do and performance should stay the same.

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -153,6 +153,18 @@ jsi::Value makeShareableInitializer(
   return ShareableJSRef::newHostObject(rt, shareable);
 }
 
+jsi::Value makeShareableFunction(jsi::Runtime &rt, jsi::Function function) {
```
tjzel, line 162: Please use this function in `makeShareable` above.

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -646,6 +647,44 @@ function ArrayBufferDemo() {
   );
 }
 
+function RemoteFunctionDemo() {
```
tjzel, line 820: What about host function demo?

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -47,23 +47,11 @@ jsi::Value makeShareableClone(
   if (value.isObject()) {
     auto object = value.asObject(rt);
     if (!object.getProperty(rt, "__workletHash").isUndefined()) {
-      if (shouldRetainRemote.isBool() && shouldRetainRemote.getBool()) {
-        shareable =
-            std::make_shared<RetainingShareable<ShareableWorklet>>(rt, object);
-      } else {
-        shareable = std::make_shared<ShareableWorklet>(rt, object);
-      }
+      return makeShareableWorklet(rt, object, shouldRetainRemote.getBool());
```
tjzel: Now this is likely to crash, because `shouldRetainRemote` would be of `undefined` type here - I think we can pass just pass `false` here since `makeShareableCloneOnUIRecursive` never passes this parameter.

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -681,6 +683,83 @@ function WorkletDemo() {
...
+    />
+  );
+}
+
+function HostFunctionDemo() {
+  const title = 'Host function';
+  const { status, isOk, isNotOk, isError } = useStatus();
+
+  const handlePress = () => {
+    // @ts-ignore _toString function is registered for UI runtime
```
tjzel:
```suggestion
    // @ts-expect-error _toString function is registered for UI runtime
```

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -681,6 +683,83 @@ function WorkletDemo() {
...
+    const hostFunction = globalThis._toString;
+    runOnUI(() => {
+      'worklet';
+      try {
+        const checks = [
+          // @ts-ignore _toString function is registered for UI runtime
```
tjzel:
```suggestion
          // @ts-expect-error _toString function is registered for UI runtime
```

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -681,6 +683,83 @@ function WorkletDemo() {
...
+          typeof hostFunction === 'function',
+          // @ts-ignore _toString function is registered for UI runtime
```
tjzel:
```suggestion
          // @ts-expect-error _toString function is registered for UI runtime
```

apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
```diff
@@ -770,6 +772,82 @@ function WorkletDemo() {
...
+  );
+}
+
+function HostFunctionDemo() {
+  const title = 'Host function';
+  const { status, isOk, isNotOk, isError } = useStatus();
+
+  const handlePress = () => {
+    // @ts-expect-error _toString function is registered for UI runtime
+    const hostFunction = globalThis._toString;
```
tjzel: I think `globalThis._toString` is not available on RN Runtime, no? Maybe use something from `workletsModuleProxy`, i.e. `globalThis.__workletsModuleProxy.makeShareableFunction`?

packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
```diff
@@ -293,6 +295,18 @@ jsi::Value JSIWorkletsModuleProxy::get(
...
+  if (name == "makeShareableFunction") {
+    return jsi::Function::createFromHostFunction(
+        rt,
+        propName,
+        1,
+        [](jsi::Runtime &rt,
+           const jsi::Value &thisValue,
+           const jsi::Value *args,
+           size_t count) {
+          return makeShareableFunction(rt, args[0].asObject(rt).asFunction(rt));
```
tjzel, line 310: Would it work?
```suggestion
          return makeShareableFunction(rt, args[0].asFunction(rt));
```

packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.cpp
```diff
@@ -47,23 +47,11 @@ jsi::Value makeShareableClone(
   if (value.isObject()) {
     auto object = value.asObject(rt);
     if (!object.getProperty(rt, "__workletHash").isUndefined()) {
-      if (shouldRetainRemote.isBool() && shouldRetainRemote.getBool()) {
-        shareable =
-            std::make_shared<RetainingShareable<ShareableWorklet>>(rt, object);
-      } else {
-        shareable = std::make_shared<ShareableWorklet>(rt, object);
-      }
+      return makeShareableWorklet(rt, object, false);
```
tjzel, line 54:
```suggestion
			// We pass `false` because this function is invoked only
			// by `makeShareableCloneOnUIRecursive` which doesn't
			// make Retaining Shareables.
      return makeShareableWorklet(rt, object, false);
```

packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
skusnierz: nope 😞

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntimeDecorator.cpp
```diff
@@ -173,6 +173,13 @@ void WorkletRuntimeDecorator::decorate(
         return makeShareableInitializer(rt, value.asObject(rt));
       });
 
+  jsi_utils::installJsiFunction(
+      rt,
+      "_makeShareableFunction",
+      [](jsi::Runtime &rt, const jsi::Value &value) {
+        return makeShareableFunction(rt, value.getObject(rt).getFunction(rt));
```
tjzel: Why isn't it `value.asObject(rt).asFunction(rt)`?

```suggestion
        return makeShareableFunction(rt, value.getObject(rt).getFunction(rt));
```

packages/react-native-worklets/src/WorkletsModule/NativeWorklets.ts
```diff
@@ -105,6 +105,10 @@ See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooti
     return this.#workletsModuleProxy.makeShareableInitializer(obj);
   }
 
+  makeShareableFunction(func: object) {
+    return this.#workletsModuleProxy.makeShareableFunction(func);
+  }
```
tjzel: I think this should be a properly generic type.

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntimeDecorator.cpp
skusnierz: `getFunction` does an extra check to see if the value is a function, but as I'm thinking now it doesn't make sense since we're checking that on the JS side

packages/react-native-worklets/src/WorkletsModule/NativeWorklets.ts
```diff
@@ -105,6 +105,10 @@ See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooti
     return this.#workletsModuleProxy.makeShareableInitializer(obj);
   }
 
+  makeShareableFunction<TFunction extends object>(func: TFunction) {
+    return this.#workletsModuleProxy.makeShareableFunction(func);
+  }
```
tjzel: I was actually thinking about something like

```suggestion
  makeShareableFunction<TArgs extends unknown[], TRet>(func: ((...args: TArgs) => TReturn) {
    return this.#workletsModuleProxy.makeShareableFunction(func);
  }
```

packages/react-native-worklets/src/WorkletsModule/workletsModuleProxy.ts
```diff
@@ -42,6 +42,8 @@ export interface WorkletsModuleProxy {
 
   makeShareableInitializer(obj: object): ShareableRef<object>;
 
+  makeShareableFunction(func: object): ShareableRef<object>;
```
tjzel: Fix this type and we gucci 👌

tjzel (approved):

PR #7596, opened 2025-05-28
Add `makeShareableOnUI` tests

skusnierz:
## Summary
This pr adds tests for `makeShareableOnUI`. Currently all tests are commented out, and we only have a test for `String`, because I want to know if my approach makes sense ? what do you think @tjzel @tomekzaw  ?

```js
const checkShareableCloneOnUI = (x: string) => {
  try {
    if (x === 'test') {
      setStatusShareableCloneOnUI('ok');
    } else {
      setStatusShareableCloneOnUI('not_ok');
    }
  } catch (e) {
    setStatusShareableCloneOnUI('error');
  }
};

runOnUI(() => {
  'worklet';
  const testStringUI = 'test';
  runOnJS(checkShareableCloneOnUI)(testStringUI);
})();
```

## Test plan
`Shareable` Example
image

### DIFF
--- a/apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
+++ b/apps/common-app/src/apps/reanimated/examples/ShareablesExample.tsx
@@ -7,7 +7,11 @@ import {
   TouchableOpacity,
   View,
 } from 'react-native';
-import { runOnJS, runOnUI } from 'react-native-reanimated';
+import {
+  executeOnUIRuntimeSync,
+  runOnJS,
+  runOnUI,
+} from 'react-native-reanimated';
 
 type Status = 'ok' | 'not_ok' | 'error' | undefined;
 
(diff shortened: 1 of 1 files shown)


tjzel:
I think it would make more sense to use `executeOnUIRuntimeSync` with its return value.

```js
const result = executeOnUIRuntimeSync(/*...*/)(/*...*/);
// assert result
```

When a type is not supported, then we should catch an error instead.

skusnierz:
> I think it would make more sense to use `executeOnUIRuntimeSync` with its return value.
> 
> ```js
> const result = executeOnUIRuntimeSync(/*...*/)(/*...*/);
> // assert result
> ```
> 
> When a type is not supported, then we should catch an error instead.

@tjzel corrected could you take a look now ?

PR #7597, opened 2025-05-29
feat: Rename useScrollViewOffset to useScrollOffset and allow any scrollable

MatiPl01:
## Summary

This PR changes the `useScrollViewOffset` hook name to `useScrollOffset` and adjusts types to accept any scrollable component (tested with `ScrollView, `FlatList` and `FlashList`).

The previous implementation also assumed that the `useScrollViewHook` works only with the `Animated.ScrollView`, which is not true and it unnecessarily limited the usage of the hook to the `Animated.ScrollView`. I changed types so that plain `ScrollView`, `FlatList` and `FlashList` can be used as well, instead of just animated ones.

## Test plan

Copy and paste this example to the `EmptyExample` in the example app and see how it works. You can also replace plain scrollabe components with 

<details>

https://github.com/user-attachments/assets/5a63baa1-7b98-45ba-af0e-c0143c9c34d5

```tsx
import { FlashList } from '@shopify/flash-list';
import React from 'react';
import { FlatList, ScrollView, StyleSheet, Text, View } from 'react-native';
import {
  useAnimatedReaction,
  useAnimatedRef,
  useScrollOffset,

const COLORS = [
  '#FF6B6B', // Coral Red
  '#4ECDC4', // Turquoise
  '#45B7D1', // Sky Blue
  '#96CEB4', // Sage Green
  '#FFEEAD', // Cream
  '#D4A5A5', // Dusty Rose
  '#9B59B6', // Purple
  '#3498DB', // Blue
  '#E67E22', // Orange
  '#2ECC71', // Emerald
];

function ColorItem({ color }: { color: string }) {
  return (
    <View style={[styles.colorItem, { backgroundColor: color }]}>
      <Text style={styles.colorText}>{color}</Text>
    </View>
  );
}

function ScrollViewExample() {
  const aref = useAnimatedRef<ScrollView>();
  const offset = useScrollOffset(aref);

  useAnimatedReaction(
    () => offset.value,
    (value) => {
      console.log('ScrollView offset', value);
    }
  );

  return (
    <ScrollView ref={aref} style={styles.listContainer}>
      {COLORS.map((color, index) => (
        <ColorItem color={color} key={index} />
      ))}
    </ScrollView>
  );
}

function FlatListExample() {
  const aref = useAnimatedRef<FlatList<string>>();
  const offset = useScrollOffset(aref);

  useAnimatedReaction(
    () => offset.value,
    (value) => {
      console.log('FlatList offset', value);
    }
  );
  return (
    <FlatList
      ref={aref}
      data={COLORS}
      keyExtractor={(_, index) => index.toString()}
      renderItem={({ item }) => <ColorItem color={item} />}
      style={styles.listContainer}
    />
  );
}

function FlashListExample() {
  const aref = useAnimatedRef<FlashList<string>>();
  const offset = useScrollOffset(aref);

  useAnimatedReaction(
    () => offset.value,
    (value) => {
      console.log('FlashList offset', value);
    }
  );
  return (
    <FlashList
      ref={aref}
      data={COLORS}
      estimatedItemSize={100}
      keyExtractor={(_, index) => index.toString()}
      renderItem={({ item }) => <ColorItem color={item} />}
      style={styles.listContainer}
    />
  );
}

  return (
      <View style={styles.section}>
        <Text style={styles.title}>ScrollView</Text>
        <ScrollViewExample />
      </View>
      <View style={styles.section}>
        <Text style={styles.title}>FlatList</Text>
        <FlatListExample />
      </View>
      <View style={styles.section}>
        <Text style={styles.title}>FlashList</Text>
        <FlashListExample />
      </View>
    </View>
  );
}

  section: {
    flexBasis: '30%',
  },
  colorItem: {
    alignItems: 'center',
    borderRadius: 8,
    elevation: 5,
    height: 100,
    justifyContent: 'center',
    marginVertical: 4,
    shadowColor: '#000',
    shadowOffset: {
      height: 2,
      width: 0,
    },
    shadowOpacity: 0.25,
    shadowRadius: 3.84,
  },
  colorText: {
    color: '#fff',
    fontSize: 16,
    fontWeight: 'bold',
    textShadowColor: 'rgba(0, 0, 0, 0.75)',
    textShadowOffset: { height: 1, width: -1 },
    textShadowRadius: 10,
  },
  container: {
    backgroundColor: '#fff',
    flex: 1,
    padding: 16,
  },
  listContainer: {
    marginBottom: 16,
  },
  title: {
    fontSize: 20,
    fontWeight: 'bold',
    marginVertical: 8,
  },
});
```
</details>

### DIFF
--- a/packages/react-native-reanimated/src/hook/useScrollOffset.ts
+++ b/packages/react-native-reanimated/src/hook/useScrollOffset.ts
@@ -60,8 +60,7 @@ function useScrollOffsetWeb<C extends ScrollableComponent>(
       offset.value =
         element.scrollLeft === 0 ? element.scrollTop : element.scrollLeft;
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [animatedRef, animatedRef?.current]);
+  }, [animatedRef, offset]);
 
   useEffect(() => {
     if (!animatedRef) {


packages/react-native-reanimated/src/component/FlatList.tsx
```diff
@@ -123,5 +123,6 @@ export const ReanimatedFlatList = FlatListRender as <
   }
 ) => React.ReactElement;
 
-export type ReanimatedFlatList<T> = typeof AnimatedFlatList &
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ReanimatedFlatList<T = any> = typeof AnimatedFlatList &
```
MatiPl01, line 127: I added this to simplify usage in the animated ref without having to specify the type of the data and make the `Animated.FlatList` type similar to the RN's `FlatList` type.

MatiPl01:
> How about restoring `useScrollViewOffset` with a deprecation warning? I'm afraid this will break some apps and libraries. I understand we're free to change any of the APIs in the v4 but let's at least try to provide a developer-friendly error message.

So maybe we could just throw an error when the old implementation is used instead of deprecating it? I'm ok with a deprecation as well, but the release of reanimated 4 seemed to me to be a good moment for breaking changes like this one.

packages/react-native-reanimated/src/index.ts
```diff
@@ -139,7 +139,7 @@ export {
   useFrameCallback,
   useHandler,
   useReducedMotion,
-  useScrollViewOffset,
+  useScrollOffset,
```
tjzel, line 142: Like @tomekzaw and @piaskowyk said, let's keep that export as `@deprecated` - it doesn't cost us anything and enables the users to gradually migrate.

packages/react-native-reanimated/src/hook/useScrollOffset.ts
```diff
@@ -13,21 +14,35 @@ import type { EventHandlerInternal } from './useEvent';
...
+  'onScrollBeginDrag',
+  'onScrollEndDrag',
+  'onMomentumScrollBegin',
+  'onMomentumScrollEnd',
+] as const;
+
+type ScrollableComponent = Component<
+  Pick<ScrollViewProps, (typeof NATIVE_SCROLL_EVENT_NAMES)[number]>
+> &
+  Pick<ScrollView, 'getScrollableNode'>;
```
tjzel, line 32: Are these props optional?

packages/react-native-reanimated/src/hook/useScrollOffset.ts
```diff
@@ -13,21 +14,35 @@ import type { EventHandlerInternal } from './useEvent';
 import { useEvent } from './useEvent';
 import { useSharedValue } from './useSharedValue';
 
+const NATIVE_SCROLL_EVENT_NAMES = [
+  'onScroll',
+  'onScrollBeginDrag',
+  'onScrollEndDrag',
+  'onMomentumScrollBegin',
+  'onMomentumScrollEnd',
+] as const;
```
tjzel, line 27: Is there a reason to keep this at the top of the file? If not I'd rather keep it at the end.

MatiPl01: What's the difference? I prefer to keep all global (in the file scope) constants at the top of the file - it is way more natural to me.

packages/react-native-reanimated/src/hook/useScrollOffset.ts
MatiPl01: Yes, they are. I wasn't sure if I should include them in that case. They are also declared optional on the `ScrollViewProps` type so maybe it is not necessary to include these props in here and only the `getScrollableNode` method is everything what we need in this type.

I am open to suggestions for better typings if you have some.

packages/react-native-reanimated/src/index.ts
MatiPl01: done

MatiPl01:
@piaskowyk 

> We should also update the `useScrollViewOffset` documentation page.

I have a draft in progress with docs changes #7603. I will finish it so that we can merge docs right after the next release in which these changes will become available.

packages/docs-reanimated/src/theme/NotFound/mapOldDocsToNewUrl.ts
```diff
@@ -22,7 +22,7 @@ const oldDocsToNewDocsMap = {
   "/docs/api/hooks/useEvent": "/docs/advanced/useEvent",
   "/docs/api/hooks/useFrameCallback": "/docs/advanced/useFrameCallback",
   "/docs/api/hooks/useHandler": "/docs/advanced/useHandler",
-  "/docs/api/hooks/useScrollViewOffset": "/docs/scroll/useScrollViewOffset",
+  "/docs/api/hooks/useScrollOffset": "/docs/scroll/useScrollOffset",
```
piaskowyk: If you plan to change the documentation in another PR, I think we should move the documentation changes from this PR to https://github.com/software-mansion/react-native-reanimated/pull/7603

packages/react-native-reanimated/src/mock.ts
```diff
@@ -118,7 +118,7 @@ const hook = {
   }),
   // useFrameCallback: ADD ME IF NEEDED
   useAnimatedKeyboard: () => ({ height: 0, state: 0 }),
-  useScrollViewOffset: () => ({ value: 0 }),
+  useScrollOffset: () => ({ value: 0 }),
```
piaskowyk, line 122: I think we should keep both names in the mocks.

packages/react-native-reanimated/src/hook/useScrollOffset.ts
tjzel: This more of a programming philosophy - you read files top to bottom, so at the top should be the most important bits (entry points) and less relevant stuff below. Basically it's kind-of anti-topological sorting of stuff.

MatiPl01: Well, I always preferred to declare things before I use them. This is more natural to me.

Also, I think that the array of event names to which we are listening to can be considered important as it is the core part of the `useScrollOffset` implementation logic.

packages/react-native-reanimated/src/mock.ts
MatiPl01: Do we really need that if we are no longer using `useScrollViewOffset` anywhere in tests? I know that I can leave both but it won't give us any benefit, since the `useScrollViewOffset` is essentially just a name alias for the `useScrollOffset`, so we should `useScrollOffset` in tests.

packages/docs-reanimated/src/theme/NotFound/mapOldDocsToNewUrl.ts
MatiPl01: Yup, this must have been changed by mistake, when I was replacing all usages of `useScrollViewOffset` to `useScrollOffset`. I must have mistakenly replace it in here.

packages/react-native-reanimated/src/hook/useScrollOffset.ts
MatiPl01: In general, the best approach is to keep things closes to where they are used, so I would normally put this constant just above the `useScrollOffsetNative` implementation but I would prefer to leave it on the top of the file as it is used in the `ScrollableComponent` type. Thanks to this, it is clear what this type consist of and we don't have to jump between the top and the bottom of the file.

packages/react-native-reanimated/src/hook/useScrollOffset.ts
MatiPl01: In general, the best approach is to keep things close to where they are used, so I would normally put this constant just above the `useScrollOffsetNative` implementation but I would prefer to leave it on the top of the file as it is used in the `ScrollableComponent` type. Thanks to this, it is clear what this type consist of and we don't have to jump between the top and the bottom of the file.

packages/react-native-reanimated/src/mock.ts
tjzel: This is for our users that need these mocks.

packages/react-native-reanimated/src/hook/useScrollOffset.ts
tjzel: These are personal preferences, so I won't enforce my school of thought, feel free to leave it the way you want ❤️

packages/react-native-reanimated/src/hook/useScrollOffset.ts
tjzel: I think we can leave it as is.

packages/react-native-reanimated/src/hook/useScrollOffset.ts
```diff
@@ -82,32 +96,22 @@ function useScrollViewOffsetNative(
...
-    if (!animatedRef) {
-      return;
-    }
-
-    if (!animatedRef.getTag) {
-      logger.warn(
-        'animatedRef is not initialized. Please make sure to pass the animated ref to the scrollable component if you want to use useScrollViewOffset.'
-      );
-      return;
-    }
```
tjzel: Why did you remove this warning?

packages/react-native-reanimated/src/mock.ts
MatiPl01: Ok, makes sense. Thanks for explanation.

packages/react-native-reanimated/src/hook/useScrollOffset.ts
MatiPl01: We can introduce some guidelines regarding the style of code we want to use in the project. I would adapt to them if we introduce them and it won't be just a personal code style preference of a person that is doing a review.

Nevertheless, thanks for your suggestions and discussion on this topic.

packages/react-native-reanimated/src/hook/useScrollOffset.ts
MatiPl01: I must have changed that mistakenly during rebase. I merged a PR that adds these lines #7606 recently and must have resolved conflicts improperly.

Thanks a lot for spotting this on.

piaskowyk (approved):

PR #7600, opened 2025-05-29
(v3) fix: TVOS example crashing right after launch

patrycjakalinska:
## Summary

TVOS example was crashing because of the `resolutions` field used in the top-level `package.json`, which was overriding version of the `@react-native/gradle-plugin` dependency in all example apps to `0.80-rc.3`, whilst TVOS example still uses RN 0.79.

I tested all example apps (except the MacOS example that we don't support on fabric yet) and everything seems to work fine.

### Error log

```
import com.facebook.react.common.annotations.internal.LegacyArchitectureLogger;
                                                     ^
symbol:   class LegacyArchitectureLogger
location: package com.facebook.react.common.annotations.internal
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,5 @@
     "prettier": "^3.3.3",
     "prettier-plugin-jsdoc": "^1.3.0",
     "typescript": "~5.3.0"
-  },
-  "resolutions": {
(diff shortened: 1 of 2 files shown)


MatiPl01 (approved):
LGTM

Issue #7480, opened 2025-05-07
Performance on Fabric is getting worse over time


Szymon20000:
### Description

The main Issue I and @hannojg found in one of our projects is that commit hook that injects reanimated animated props in shadow tree (on RN flush) is getting really slow over time. It sometimes get's multiple frames to execute.
The main problem we noticed is that Props Registry contains all reanimated props/views that were ever animated and not unmounted yet. Imagine you have an animated list with useAnimatedStyle for each item. Most of the time the items are not animated but because they have useAnimatedStyle and they call updateProps at least once we will include them in Commit hook action. 

We would like to propose a few solutions to that that reduced the number of junk frames by a lot.

The simplest approach is to get rid of views/props from props registry once the props we get from RN match those from reanimated. If they do we evict them from reanimated registry. Of course normally the props won't magically update and that's why we have the commit hook. However, we can make few things to synchronize them. Here are few ways how to do it:
1. We keep state in every view wrapped by createReanimatedComponent. We check on each frame on UI side which props were not updated in the last 2 frames if they were not it means their animations are finished and we can schedule a set state. As setStates are batched on Fabric and we only call them once animations are over they actually end up not adding much overhead. Then on next React Native flush to Shadow Tree we compare if props of the animated node are already the same and if that's the case we don't need to update the node and can evict the props from the registry.
2. We can make use of new feature flag on iOS that pushes latest yoga nodes to js.
3. We can patch cloneNodeWithChildren to always take latest nodes for children from the shadow tree and inject reanimated animated props in clone method. (Even though we of course may not clone all animated nodes it doesn't matter as we would take latest children nodes that already contains animated changes or will contain on next render which will lead to their eviction in commit hook. 

@tomekzaw @kmagiera It would be great to hear your thoughts on it. We are also happy to share exact numbers that we got and exact implementation. 

### Steps to reproduce

1. Render screen with few animated views
2. wrap reanimated commit hook in time measurement.

### Snack or a link to a repository

/

### Reanimated version

latest

### React Native version

0.78

### Platforms

iOS, Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Host machine

None

### Acknowledgements

Yes

tomekzaw:
[...]

PR #7601, opened 2025-05-29
[PoC] perf: Sync state back to JS + clear prop registry

hannojg:
## Summary

Addresses: https://github.com/software-mansion/react-native-reanimated/issues/7480

This PR shows a PoC of how we can sync our state from the UI thread back to JS thread / react reconciler.

Consider the following example:
- In your app you have a complicated navigation structure and so far across multiple screen you've animated 100 views that aren't removed yet
- On a screen you want to animate one button
- in the cloneShadowTree method we still have the previous 100 animated views in the registry 
- We now need to clone 101 nodes, instead of just one for running the animation

The more nodes we keep and clone the slower the function gets our performance profilings have shown, which is a major performance bottleneck.

We keep all props in the props registry so that when react is performing a commit from the JS thread we don't "drop" the changes we made on the UI thread.
This mechanism now allows us to update the views/nodes on JS, so we don't need to keep all props.

As updates in fabric are batched, even if we update a lot of nodes it should result only in one react commit.

To provide a performance measurement. We have a huge list where each item is animated in, so they are added to the prop registry. Scrolling became slow over time.
For this screen we had a 37.43% jank frame rate. With this change the jank frame rate dropped to 19.99%.
As a reference, when having completely disabling the commit/mount hooks the jank frame rate is 16.70%.

(I don't want to dilute the scope of this description too much, but we also added back [the fast update path where we update none layout props directly instead of running through the commit pipeline](https://github.com/software-mansion/react-native-reanimated/pull/7014), and landed on a jank frame rate of 12.86%)

### Things to improve in this PR

- [ ] Currently for detecting if an animation is done in updateProps we check if the last update is more than 20 ms ago (~ 2 frames). This is really wonky and it would be better to detect this through some other mechanisms
- [ ] I created `ComponentRegistry.ts` but there might be existing registries / mechanisms that I could use?
- [ ] Some C++ cleanups

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedCommitHook.cpp
@@ -76,10 +76,16 @@ RootShadowNode::Unshared ReanimatedCommitHook::shadowTreeWillCommit(
   {
     auto lock = updatesRegistryManager_->lock();
 
-    PropsMap propsMap = updatesRegistryManager_->collectProps();
+    PropsMap propsMap =
+        updatesRegistryManager_->collectProps(); // Reanimated.View {...props}
     updatesRegistryManager_->cancelCommitAfterPause();
 
-    rootNode = cloneShadowTreeWithNewProps(*rootNode, propsMap);
+    std::vector<Tag> tagsToRemove;
+    rootNode = cloneShadowTreeWithNewProps(*rootNode, propsMap, tagsToRemove);
+    for (const auto& tag : tagsToRemove) {
+        updatesRegistryManager_->markNodeAsImmediateRemovable(tag);
+    }
+
     // If the commit comes from React Native then pause commits from
     // Reanimated since the ShadowTree to be committed by Reanimated may not
     // include the new changes from React Native yet and all changes of animated
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedMountHook.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ReanimatedMountHook.cpp
@@ -23,6 +23,13 @@ void ReanimatedMountHook::shadowTreeDidMount(
     double) noexcept {
   ReanimatedSystraceSection s("ReanimatedMountHook::shadowTreeDidMount");
 
+  // Nodes marked earlier in ReanimatedCommitHook for removal (if there props were the same) are removed here
+  // because here we know that the commit is finished and we can safely remove
+  {
+    auto lock = updatesRegistryManager_->lock();
+    updatesRegistryManager_->removeImmediateRemovableNodes();
+  }
+
   auto reaShadowNode =
       std::reinterpret_pointer_cast<ReanimatedCommitShadowNode>(
           std::const_pointer_cast<RootShadowNode>(rootShadowNode));
--- a/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
@@ -6,6 +6,29 @@
 
 namespace reanimated {
 
+/**
+ * Checks it the props of a are identical to the props in b.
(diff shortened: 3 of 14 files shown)


tomekzaw:
@hannojg I'd like to revisit this draft PR. I've merged current main but wasn't able to push commits to this PR. Could you please merge https://github.com/software-mansion/react-native-reanimated/commits/sync-state/ branch?

Also, looks like this draft PR works only on Android because it's not possible to obtain `rawProps` from `Props` on iOS. Do you have any idea how to achieve the same thing on iOS?

tomekzaw:
@hannojg I've managed to revisit this draft PR. Here are my main concerns regarding the proposed changes:
1. **This draft PR doesn't work on iOS.** The `rawProps` field is available only on Android (see `Props.h`: https://github.com/facebook/react-native/blob/60471cb268a70546d6ea555d25c90ca47ad3d839/packages/react-native/ReactCommon/react/renderer/core/Props.h#L63). It's hidden behind `RN_SERIALIZABLE_STATE` compilation macro which is supposed to be enabled only on Android. This draft PR doesn't provide any counterpart mechanism for iOS. https://github.com/facebook/react-native/blob/60471cb268a70546d6ea555d25c90ca47ad3d839/packages/react-native/ReactCommon/react/renderer/core/Props.h#L63
2. **It doesn't work for `<Animated.View>` inside `<GestureDetector>` (pretty common scenario).** Turns out, `GestureDetector` is an animated component itself. In such case, both `Animated.View` and `GestureDetector` have the same `viewTag` (because they point to the same host view). Because of this, `ComponentRegistry.register(viewTag, this);` will first register `Animated.View` under some `viewTag` which will be then overwritten by registering `GestureDetector` which is one level higher in the React tree. So the style updates will be sent to `GestureDetector` instead of `Animated.View`. It can be fixed by skipping the map update if there's already a map item with the same `viewTag`. (Thanks to @j-piasecki and @bartlomiejbloniarz for helping me and figuring that out.)
3. **It requires additional memory to store accumulated animated styles.** The `global.__lastUpdateByTag` map holds all style updates for all animated components on the JavaScript side. This duplicates the current data structures that we have on the C++ side and will increase the memory usage of the app.
4. **It is not safe to remove the accumulated styles in `ReanimatedCommitHook` because the commit may fail.** If the commit fails, React Native will retry the commit. However, during the second pass, `ReanimatedCommitHook` won't apply the animated changes because they have already been removed from the registry. This would lead to flickering issues.
5. **The `checkPropsEqual` mechanism can be flaky in some edge cases.** I can imagine a case that some animated component is updated as a result of React render with exactly the same values that the `checkPropsEqual` condition is met, leading to removing the animated styles from the registry on the C++ side before the synchronization of animated styles back to React is complete.

This means we can't merge the PR in its current state. ⚠️ We also discourage from using the changes from this draft PR in their apps, even if it seems to improve the performance, as it can lead to much more serious issues described above. 

However, I'd like to confirm that the core idea of this draft PR (which is how to synchronize the animated styles back to React once the animations have settled using `this.setState` method in `AnimatedComponent` class) **is correct**. Inspired by this PR, I managed to submit another draft PR which uses the same technique to synchronize styles of settled animations back to React:
* https://github.com/software-mansion/react-native-reanimated/pull/8529

We already tested this draft PR on a large production app and managed to increase JS FPS from ~17 fps to ~56 fps while keeping stable 60 fps on the UI thread so we have really high hopes for this one. We also have some draft ideas on how to achieve the same effect without requiring a full React commit using the ShadowNode runtime reference update mechanism but this is yet to come as we want to land the draft PR in its current form as soon as possible.

tomekzaw:
Hey, thanks again @hannojg for this draft PR! The approach you came up with proved to be quite successful in terms of increasing JS fps in RN apps with multiple animated components across various screens. Based on your idea, we implemented similar optimization and gated it behind `FORCE_RENDER_FOR_SETTLED_ANIMATIONS` feature flag (https://github.com/software-mansion/react-native-reanimated/pull/8529) which is now enabled by default in Reanimated 4.3.0. Also, @bartlomiejbloniarz and @coado have co-authored the Shared Animation Backend for React Native which uses RSNRU mechanism to achieve similar effect and we're currently adopting the animation backend in https://github.com/software-mansion/react-native-reanimated/pull/8875 so I'm going to close this draft PR.

PR #7604, opened 2025-05-30
fix: useHandler updating on every render, even if dependencies didn't change

MatiPl01:

### DIFF
--- a/packages/react-native-reanimated/src/hook/useHandler.ts
+++ b/packages/react-native-reanimated/src/hook/useHandler.ts
@@ -3,7 +3,7 @@ import { useEffect, useRef } from 'react';
 import type { WorkletFunction } from 'react-native-worklets';
 import { makeShareable } from 'react-native-worklets';
 
-import { IS_JEST, IS_WEB } from '../common';
+import { IS_JEST, IS_WEB, ReanimatedError } from '../common';
 import type { DependencyList, ReanimatedEvent } from './commonTypes';
 import { areDependenciesEqual, buildDependencies } from './utils';
 
@@ -82,6 +82,14 @@ export function useHandler<
   }, []);
 
   const { context, savedDependencies } = initRef.current;
+
+  for (const handlerName in handlers) {
+    if (!handlers[handlerName]!.__workletHash) {
+      throw new ReanimatedError(
+        'Passed function is not a worklet. Please provide a worklet function.'
+      );
+    }
+  }
 
   dependencies = buildDependencies(
     dependencies,
--- a/packages/react-native-reanimated/src/hook/utils.ts
+++ b/packages/react-native-reanimated/src/hook/utils.ts
@@ -27,6 +27,7 @@ export function buildDependencies(
   const handlersList = Object.values(handlers).filter(
     (handler) => handler !== undefined
   ) as NonNullable<Handler>[];
+
   if (!dependencies) {
     dependencies = handlersList.map((handler) => {
       return {
--- a/apps/common-app/src/apps/reanimated/examples/EmptyExample.tsx
+++ b/apps/common-app/src/apps/reanimated/examples/EmptyExample.tsx
@@ -1,7 +1,32 @@
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { StyleSheet, Text, View } from 'react-native';
+import { useAnimatedGestureHandler } from 'react-native-reanimated';
 
 export default function EmptyExample() {
+  const [ticker, setTicker] = useState(0);
+  useEffect(() => {
+    const interval = setInterval(() => setTicker((prev) => prev + 1), 1000);
+    return () => clearInterval(interval);
+  }, []);
+  const onStart = useCallback(() => {
+    'worklet';
+  }, []);
+  const onActive = useCallback(() => {
+    'worklet';
+  }, []);
+  const onFinish = useCallback(() => {
+    'worklet';
+  }, []);
+
+  const gestureHandler = useAnimatedGestureHandler(
+    {
+      onStart,
+      onActive,
+      onFinish,
+    },
+    []
+  );
+
   return (
     <View style={styles.container}>
       <Text>Hello world!</Text>
--- a/packages/react-native-reanimated/src/hook/useAnimatedGestureHandler.ts
+++ b/packages/react-native-reanimated/src/hook/useAnimatedGestureHandler.ts
@@ -88,6 +88,7 @@ export function useAnimatedGestureHandler<
     handlers,
     dependencies
   );
+
   const handler = (e: WebOrNativeEvent) => {
     'worklet';
     const event = useWeb


packages/react-native-reanimated/src/hook/useHandler.ts
```diff
@@ -83,6 +83,14 @@ export function useHandler<
 
   const { context, savedDependencies } = initRef.current;
 
+  for (const handlerName in handlers) {
+    if (!handlers[handlerName]!.__workletHash) {
+      throw new ReanimatedError(
+        'Passed function is not a worklet. Please provide a worklet function.'
+      );
+    }
+  }
```
MatiPl01, line 92: Without this check, passing non-worklet functions with an empty dependencies array resulted in the following error:



The following code resulted in this error:

```tsx
const onStart = useCallback(() => {}, []);
const onActive = useCallback(() => {}, []);
const onFinish = useCallback(() => {}, []);

  {
    onStart,
    onActive,
    onFinish,
  },
  []
);
```

packages/react-native-reanimated/src/hook/utils.ts
```diff
@@ -41,6 +41,17 @@ export function buildDependencies(
...
 }
 
+function isWorkletHandler(
+  dep: unknown
+): dep is { workletHash: unknown; closure: unknown } {
+  return (
+    typeof dep === 'object' &&
+    dep !== null &&
+    'workletHash' in dep &&
+    'closure' in dep
```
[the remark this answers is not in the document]
MatiPl01: I added it just for safety to make sure that the object is the same as what is returned from this function:

```ts
// Builds dependencies array for useEvent handlers.
export function buildDependencies(
  dependencies: DependencyList,
  handlers: Record<string, WorkletFunction | undefined>
) {
  type Handler = (typeof handlers)[keyof typeof handlers];
  const handlersList = Object.values(handlers).filter(
    (handler) => handler !== undefined
  ) as NonNullable<Handler>[];
  if (!dependencies) {
    dependencies = handlersList.map((handler) => {
      return {
        workletHash: handler.__workletHash,
        closure: handler.__closure,
      };
    });
  } else {
    dependencies.push(buildWorkletsHash(handlersList));
  }

  return dependencies;
}
```

All worklet handlers are converted to the following object:

```ts
{
  workletHash: handler.__workletHash,
  closure: handler.__closure,
}
```

It might not be necessary, but I added it to be extra safe here. I have no strong opinion on that, though.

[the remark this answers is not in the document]
headfire94: > All worklet handlers are converted to the following object:

yes actually my question about this deps builder, `closure` prop not used from it as well

packages/react-native-reanimated/src/hook/useHandler.ts
```diff
@@ -83,6 +83,14 @@ export function useHandler<
 
   const { context, savedDependencies } = initRef.current;
 
+  for (const handlerName in handlers) {
+    if (!isWorkletFunction(handlers[handlerName])) {
```
piaskowyk, line 87: Why not something like this? 🤔 
```suggestion
  for (const handler of handlers) {
    if (!isWorkletFunction(handler)) {
```

tjzel: You could also do `Object.values(handlers).some(...)`

packages/react-native-reanimated/src/hook/utils.ts
```diff
@@ -41,6 +41,17 @@ export function buildDependencies(
   return dependencies;
 }
 
+function isWorkletHandler(
+  dep: unknown
+): dep is { workletHash: unknown; closure: unknown } {
```
tjzel: I think we should explicitly define this type, use it `buildDependencies` functions, here and in other appropriate places.

packages/react-native-reanimated/src/hook/utils.ts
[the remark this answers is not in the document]
tjzel: I wonder if this should be a `__DEV__` only check.

packages/react-native-reanimated/src/hook/utils.ts
```diff
@@ -62,9 +73,17 @@ export function areDependenciesEqual(
...
-        return false;
+      const nextDep = nextDeps[i];
+      const prevDep = prevDeps[i];
+      if (objectIs(nextDep, prevDep)) {
+        continue;
+      }
+      if (isWorkletHandler(nextDep) && isWorkletHandler(prevDep)) {
+        if (nextDep.workletHash !== prevDep.workletHash) {
+          return false;
+        }
```
tjzel: This makes me a bit worried. The hash might stay the same, but the closures could have changed:

```js
const rnd = Math.random();
const handlers = {
  onStart: () => {
    'worklet';
		return rnd;
	}
}

const { context, doDependenciesDiffer, useWeb } = useHandler(
    handlers,
    dependencies
  );
```

It's tricky to decide what should happen in this situation. I believe, that current behavior works, since it actually enforces a change. To avoid unnecessary updates the code should probably be something among these lines:

```js
const rnd = Math.random();
const handlers = useMemo({
  onStart: () => {
    'worklet';
		return rnd;
	}
}, [rnd]);


const { context, doDependenciesDiffer, useWeb } = useHandler(
    handlers,
    dependencies
  );
```

packages/react-native-reanimated/src/hook/useHandler.ts
MatiPl01: For loop is faster than getting an array of object values only to iterate over them. I also feel that it is cleaner.

packages/react-native-reanimated/src/hook/utils.ts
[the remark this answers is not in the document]
MatiPl01: I don't think so. If we change it to a dev only check, then it would be very cumbersome as comparison will work properly in dev but break after building the app in prod. I don't think we want to ship code that behaves in a different way in prod, especially if will result in less performant code in prod, like in this case, when we don't compare worklet handler objects properly and thus, return `true` for `doDependenciesDiffer` in each render.

packages/react-native-reanimated/src/hook/utils.ts
MatiPl01: I decided to get rid of this `isWorkletHandler` check and remove conversion of the worklet function to the object with `workletHash` and `closure` fields. I prefer to just return a worklet itself as it will be easier to compare later on in the `areDependenciesEqual` (see the new `areWorkletsEqual` function).

packages/react-native-reanimated/src/hook/utils.ts
MatiPl01: As we talked on site, I added closure values comparison. I also updated the example code snippet in this PR description to include test checking what happens if the worklet closure changes.

The updated implementation handles all cases properly (closure change, passed worklet function -> wotklet hash change)

packages/react-native-reanimated/src/hook/utils.ts
```diff
@@ -62,8 +80,17 @@ export function areDependenciesEqual(
...
     for (let i = 0; i < prevDeps.length; ++i) {
-      if (!objectIs(nextDeps[i], prevDeps[i])) {
+      const nextDep = nextDeps[i];
+      const prevDep = prevDeps[i];
+      if (objectIs(nextDep, prevDep)) {
+        continue;
+      }
+      if (!isWorkletFunction(nextDep) || !isWorkletFunction(prevDep)) {
+        return false;
+      }
```
MatiPl01, line 90: Since we are doing a shallow comparison and `objectIs` returned false, we can safely assume that everything not being a worklet is not equal. Worklets need to be compared separately (via `workletHash` and `closure` comparison).

packages/react-native-reanimated/src/hook/utils.ts
```diff
@@ -28,19 +29,32 @@ export function buildDependencies(
...
   return dependencies;
 }
 
+function areWorkletsEqual(
+  worklet1: WorkletFunction,
+  worklet2: WorkletFunction
+) {
+  if (worklet1.__workletHash === worklet2.__workletHash) {
+    const closure1Keys = Object.keys(worklet1.__closure);
+    const closure2Keys = Object.keys(worklet2.__closure);
```
MatiPl01, line 45: I checked that `__closure` exists on web but is an empty object so it is safe to call `Object.keys` without checking for `__closure` existence.

packages/react-native-reanimated/src/hook/useHandler.ts
```diff
@@ -83,6 +83,14 @@ export function useHandler<
...
+      throw new ReanimatedError(
+        'Passed a function is not a worklet. Please provide a worklet function.'
+      );
+    }
+  }
```
tjzel, line 92: This will break when using Reanimated without the Babel plugin, which seems to be broken already though.

packages/react-native-reanimated/src/hook/utils.ts
```diff
@@ -28,19 +29,32 @@ export function buildDependencies(
...
+
+    return (
+      closure1Keys.length === closure2Keys.length &&
+      closure1Keys.every(
+        (key) => worklet1.__closure[key] === worklet2.__closure[key]
+      )
```
tjzel: This will fail if you have two worklets, one with closure

```js
{
  a: undefined,
  b: undefined,
  c: undefined,
}
```

and the second

```js
{
  d: undefined,
  e: undefined,
  f: undefined,
}
```

We must also check if both objects have the same keys.

packages/react-native-reanimated/src/hook/useHandler.ts
MatiPl01: So how we can check if the worklet function was passed when this function accepts only worklets?

packages/react-native-reanimated/src/hook/utils.ts
MatiPl01: yup, you are right. I forgot that we will get `undefined` if there is no prop with a given key.

packages/react-native-reanimated/src/hook/useHandler.ts
tjzel: Keep it as is, I don't think Reanimated without the plugin works anymore.

PR #7607, opened 2025-05-30
fix: bump to C++20

okwasniewski:


tjzel:
@okwasniewski Thanks for your contribution!

PR #7610, opened 2025-06-02
Update NativeEventsManager.ts

ovieokeh:
## Summary
Ensures that crashes don't happen in cases where `componentAnimatedRef` doesn't exist.

An example case is wrapping [AgendaList from react-native-calendars](https://github.com/wix/react-native-calendars/blob/master/src/expandableCalendar/AgendaList/agendaList.tsx). This causes a crash because `componentAnimatedRef` seems to not exist initially, but eventually resolves, and then it's fine.

```
const AnimatedAgendaList = Animated.createAnimatedComponent(AgendaList);
```

This is on iOS so I'm not sure if it would happen on an Android build as well (I think it would)

## Test plan
- Clone and `npm install` [this repo](https://github.com/ovieokeh/create-animated-component-bug-repro)
- Run the app and notice the following crash

https://github.com/user-attachments/assets/1d0998c6-9ffa-4c88-8bf4-79f084089a5f

- Implement the fix and refresh the app
- Notice it now works like below

https://github.com/user-attachments/assets/aec2775c-b1b0-4ac7-bd2e-b9398e19d059


piaskowyk (approved):
Thank you for that PR! 🙌

PR #7611, opened 2025-06-02
Remove HostInstancePaper

piaskowyk:
## Summary

This PR removes leftovers from the migration from Paper to Fabric and eliminates unnecessary old backward compatibility (the minimum supported version is now 0.75). In the next step, I'll try to get rid of unnecessary occurrences of the view tag.

## Test plan

Smoke test - Check if animation works


tjzel (approved):
I remember there had been some comments that claimed something is available on Paper but it was on Fabric too. I think we already handled this bits, but make sure to double check it.

piaskowyk:
I checked again and found one leftover that's somewhat related to my changes, so I decided to remove it in this PR as well 🫡

PR #7612, opened 2025-06-02
fix: useScrollViewOffset not working after delayed ref initialization

MatiPl01:
## Summary

This PR fixes `useScrollViewOffset` hook not updating the returned SharedValue if the scrollable to which the ref is attached is rendered with some delay.

The previous implementation didn't work because the `animatedRef` was uninitialized when the `useEffect` hook in the `useScrollViewOffset` was called if the ref wasn't passed to any component. The `useEffect` was not triggered after the ref was passed to the component later on, even though (at least as I understand) the `animatedRef?.current` in the `useEffect` dependencies changed, because this change happened after hook dependencies were recalculated.

Since the code od the component executes from the top to the bottom during render, the `useEffect` hook dependencies array was created before the `animatedRef.current` was set in the function created within the `useAnimatedRef` hook (it is set as soon as the ref function is called with the exiting component argument, which happened after `useEffect` hook dependencies were recalculated).

### Mobile

| Before | After |
|-|-|
| <video src="https://github.com/user-attachments/assets/16b2d7dd-ec29-45be-ab7e-44007e922204" /> | <video src="https://github.com/user-attachments/assets/9e6f4475-4f66-4b46-a42b-2322a3cdb301" /> |

### Web

| Before | After |
|-|-|
| <video src="https://github.com/user-attachments/assets/4e8f02af-5aee-498a-a738-a674db062bbc" /> | <video src="https://github.com/user-attachments/assets/cd0965da-7e2c-4281-ac3b-f3f58c97e153" /> |

## Test plan

Copy-paste the following code snippet and see how it works.

<details>

```tsx
import { Button, ScrollView, StyleSheet, Text, View } from 'react-native';
import type Animated from 'react-native-reanimated';
import {
  useAnimatedReaction,
  useAnimatedRef,
  useScrollViewOffset,

  const [shouldRender, setShouldRender] = useState(false);
  const scrollRef = useAnimatedRef<Animated.ScrollView>();
  const scrollOffset = useScrollViewOffset(scrollRef);

  useAnimatedReaction(
    () => scrollOffset.value,
    (value) => {
      console.log('scrollOffset', value);
    }
  );

  return (
      <Text style={styles.text}>
        ScrollView is {shouldRender ? 'rendered' : 'not rendered'}
      </Text>
      <Button
        title={shouldRender ? 'Remove ScrollView' : 'Render ScrollView'}
        onPress={() => setShouldRender(!shouldRender)}
      />

      {shouldRender && (
        <ScrollView
          ref={scrollRef}
          scrollEventThrottle={1000 / 60}
          style={styles.scrollView}>
          {Array.from({ length: 20 }).map((_, index) => (
            <View key={index} style={styles.item}>
              <Text style={styles.itemText}>Scroll Item {index + 1}</Text>
            </View>
          ))}
        </ScrollView>
      )}
    </View>
  );
}

  container: {
    flex: 1,
    backgroundColor: '#fff',
    padding: 20,
  },
  text: {
    fontSize: 16,
    marginBottom: 10,
  },
  scrollView: {
    flex: 1,
    marginTop: 20,
  },
  item: {
    height: 100,
    backgroundColor: '#f0f0f0',
    marginVertical: 8,
    borderRadius: 8,
    justifyContent: 'center',
    alignItems: 'center',
  },
  itemText: {
    fontSize: 16,
    color: '#333',
  },
});
```
</details>

I also checked if the following examples in the example app still work (on mobile and web):

- useAnimatedScrollHandler
- scrollTo
- useScrollViewOffset

### DIFF
--- a/packages/react-native-reanimated/src/hook/commonTypes.ts
+++ b/packages/react-native-reanimated/src/hook/commonTypes.ts
@@ -25,13 +25,16 @@ export interface Descriptor {
   shadowNodeWrapper: ShadowNodeWrapper;
 }
 
+export type AnimatedRefObserver = (tag: number | null) => void;
+
 export type AnimatedRef<T extends Component> = {
   (component?: T):
     | number // Paper
     | ShadowNodeWrapper // Fabric
     | HTMLElement; // web
   current: T | null;
-  getTag: () => number;
+  observe: (observer: AnimatedRefObserver) => void;
+  getTag?: () => number | null;
 };
 
 // Might make that type generic if it's ever needed.
--- a/packages/react-native-reanimated/src/hook/commonTypes.ts
+++ b/packages/react-native-reanimated/src/hook/commonTypes.ts
@@ -25,7 +25,9 @@ export interface Descriptor {
   shadowNodeWrapper: ShadowNodeWrapper;
 }
 
-export type AnimatedRefObserver = (tag: number | null) => void;
+export type MaybeObserverCleanup = (() => void) | undefined;
+
+export type AnimatedRefObserver = (tag: number | null) => MaybeObserverCleanup;
 
 export type AnimatedRef<T extends Component> = {
   (component?: T):
--- a/packages/react-native-reanimated/src/hook/useAnimatedRef.ts
+++ b/packages/react-native-reanimated/src/hook/useAnimatedRef.ts
@@ -36,52 +36,39 @@ function getComponentOrScrollable(component: MaybeScrollableComponent) {
   return component;
 }
 
-function useAnimatedRefNative<
-  TComponent extends Component,
->(): AnimatedRef<TComponent> {
-  const [wrapper] = useState(() => makeMutable<ShadowNodeWrapper | null>(null));
+function useAnimatedRefBase<TComponent extends Component>(
+  getWrapper: (component: TComponent) => ShadowNodeWrapper
+): AnimatedRef<TComponent> {
   const [observers] = useState<Set<AnimatedRefObserver>>(() => new Set());
   const wrapperRef = useRef<ShadowNodeWrapper | null>(null);
 
   const ref = useRef<AnimatedRef<TComponent> | null>(null);
 
   if (!ref.current) {
-    /** Called by React when ref is attached to a component. */
     const fun: AnimatedRef<TComponent> = <AnimatedRef<TComponent>>((
       component
     ) => {
-      let currentWrapper: ShadowNodeWrapper | null = null;
       if (component) {
-        const getTagOrShadowNodeWrapper = () => {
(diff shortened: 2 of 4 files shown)


packages/react-native-reanimated/src/hook/commonTypes.ts
```diff
@@ -25,13 +25,16 @@ export interface Descriptor {
...
+
 export type AnimatedRef<T extends Component> = {
   (component?: T):
     | number // Paper
     | ShadowNodeWrapper // Fabric
     | HTMLElement; // web
   current: T | null;
-  getTag: () => number;
+  observe: (observer: AnimatedRefObserver) => void;
+  getTag?: () => number | null;
```
MatiPl01, line 39: I changed this to the correct type as `getTag` might not exist if the ref is not initialized (wasn't passed to any component, thus the ref function wasn't called with the component argument) or might return `null` if there is no component - ref was initialized before but component was removed later on.

packages/react-native-reanimated/src/hook/useAnimatedRef.ts
tjzel, line 1: Can you refactor this file in a separate PR? This code is super dense and this diff is unreadable like this.

MatiPl01: I can separate changes related to the `useAnimatedRef` hook from this PR but I don't see how this will improve the diff of this file. Can you elaborate on your suggestion?

tjzel: I think that you can duplicate your changes in both `useAnimatefRefNative` and `useAnimatedRefWeb` in this pull request, and merge them with `useAnimatedRefBase` in the follow-up PR.

MatiPl01: Ok, it makes sense to me. I will do that.

MatiPl01: done 😃
the deduplication PR is here: #7618

packages/react-native-reanimated/src/hook/useAnimatedRef.ts
```diff
@@ -36,6 +41,9 @@ function useAnimatedRefNative<
   TComponent extends Component,
 >(): AnimatedRef<TComponent> {
   const [tag] = useState(() => makeMutable<ShadowNodeWrapper | null>(null));
+  const [observers] = useState<Map<AnimatedRefObserver, MaybeObserverCleanup>>(
+    () => new Map()
+  );
```
tjzel: Let's turn it into a ref, since state should be only mutated with `setState`

PR #7616, opened 2025-06-03
chore: Add missing CSS properties helper warnings

MatiPl01:
## Summary

This PR adds helper warnings for missing CSS animation/transition properties.

### Animations

Screenshot 2025-06-03 at 17 42 50

### Transitions

Screenshot 2025-06-03 at 17 43 44


MatiPl01:
The reason why I opened is PR is the confusion I saw sometimes during the Animations workshops at app.js conf. Some people were wondering why the animation doesn't work when they forgot to specify the duration. I think it's fine to keep all props optional, to align with how CSS works on web but also add some helper warning messages for easier development.

PR #7620, opened 2025-06-04
feat: POC of useSharedArray

skusnierz:
## Summary
This PR introduces a new hook `useSharedArray`, which allows you to subscribe to only modify a specific array element or many.

Previously, modifying any element of the array created with `useSharedValue` updated all animated styles, regardless of whether the animated style used only one element of that array or all of them:

```js
  const sharedValue = useSharedValue([100, 200]);

  const animatedStyleElement1 = useAnimatedStyle(() => {
    console.log('AnimatedStyleElement1 updated...');
    return {
      width: sharedValue.value[0],
    };
  });

  const animatedStyleElement2 = useAnimatedStyle(() => {
    console.log('AnimatedStyleElement2 updated...');
    return {
      width: sharedValue.value[1],
    };
  });

  const onUpdateElement1 = () => {
    sharedValue.modify((array) => {
      array[0] = randomIntNumber(100, 200);
      return array;
    });
  };

  onUpdateElement1();
  // Will update both elements
  // AnimatedStyleElement1 updated...
  // AnimatedStyleElement2 updated...
```

In the new approach, we let the user define when the update should occur : 

```js
  const sharedValue = useSharedArray([100, 200]);

  const animatedStyleElement1 = useAnimatedStyle((registry) => {
    console.log('AnimatedStyleElement1 updated...'); 
    registry?.registerForUpdates(sharedValue, [0]); // Register for updates on element 1
    return {
      width: sharedValue.value[0],
    };
  });

  const animatedStyleElement2 = useAnimatedStyle((registry) => {
    console.log('AnimatedStyleElement2 updated...');
    registry?.registerForUpdates(sharedValue, [1]); // Register for updates on element 2
    return {
      width: sharedValue.value[1],
    };
  });

  const onUpdateElement1 = () => {
    sharedValue.modifyValue(0, randomIntNumber(100, 200));
    // To modify the value, we need to use the `modifyValue` method
    // old `modify` method will update all animated styles that are using this shared array
  };
```

## Test plan
`UseSharedArrayExample`
<video src="https://github.com/user-attachments/assets/fd116fb4-ee35-485f-b3ae-5e86edf0b665" />


apps/common-app/src/apps/reanimated/examples/UseSharedArrayExample.tsx
```diff
@@ -0,0 +1,61 @@
...
+export default function UseSharedArrayExample() {
+  const sharedValue = useSharedArray([100, 200]);
+
+  const animatedStyleElement1 = useAnimatedStyle((registry) => {
+    console.log('AnimatedStyleElement1 updated...');
+    registry?.registerForUpdates(sharedValue, [0]);
+    return {
+      width: sharedValue.value[0],
+    };
+  });
```
tjzel, line 20: I wonder if better API would be something like this, but wait till we discuss it with the team:

```suggestion
  const animatedStyleElement1 = useAnimatedStyle((styleRef) => {
    console.log('AnimatedStyleElement1 updated...');
    sharedValue?.registerForUpdates(styleRef, [0]);
    return {
      width: sharedValue.value[0],
    };
  });
```

PR #7622, opened 2025-06-04
@matipl01/custom node lifecycle hooks

MatiPl01:


MatiPl01:
We won't implement custom `ShadowNode` for now so I decided to close this PR.

PR #7624, opened 2025-06-04
iOS ProMotion stutters

piaskowyk:
## Summary

Main goal: Improve the visual perception of animations by dynamic decreasing the screen frame rate to avoid frame drops.

The problem is visible on screens with higher frame rates (more than 60FPS), where we can't keep up with the 8ms frame size. This can result in a situation where we physically compute more than 60FPS, but due to dropped frames, stutters become very noticeable and the display looks significantly worse than a stable 60FPS.

Screenshot 2025-06-05 at 08 43 37

On ProMotion screens, we have the ability to change the frame rate dynamically. This PR adds logic to detect whether we can maintain 120FPS. If not, it decreases the frame rate to 60FPS to achieve a better visual effect for heavy animations, effectively eliminating stutters.

Screenshot 2025-06-05 at 08 43 52

When possible, we restore 120FPS again.

ProMotion supports a range of refresh rates, as described here: [ProMotion Refresh Rates Guide](https://developer.apple.com/documentation/quartzcore/optimizing-promotion-refresh-rates-for-iphone-13-pro-and-ipad-pro?language=objc). I suggested adding more granularity by implementing 4 levels of frame rates. ProMotion actually supports more than 4, but my predictions are based on only 3 measurements to minimize delays in switching. Adding more levels of granularity may result in frequent switching between different frame rates, which could produce poor visual effects, as only 3 measurements can lead to high variance for such granularity levels.

Remarks:
- I didn't find any information indicating that changing the frame rate is an expensive operation.
- I've also looked for a system solution to obtain animation loop frame rate information, so I needed to compute it on my own.
- I tried to attach a video, but screen recording on iOS works only at 60FPS.

I replaced [deprecated property](https://developer.apple.com/documentation/quartzcore/cadisplaylink/preferredframespersecond?language=objc) `preferredFramesPerSecond` with [newer one](https://developer.apple.com/documentation/quartzcore/cadisplaylink/preferredframeraterange?language=objc) `preferredFrameRateRange`


packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -2,14 +2,34 @@
...
+};
+
 @implementation AnimationFrameQueue {
   /* DisplayLink is thread safe. */
   WorkletsDisplayLink *displayLink_;
   std::vector<std::function<void(double)>> frameCallbacks_;
   std::mutex callbacksMutex_;
+  std::chrono::duration<double, std::milli> timeDeltas_[TIME_SAMPLES_AMOUNT];
+  int timeDeltaIndex_;
+  FrameRateRangeEnum curentFrameRate;
```
tomekzaw:
```suggestion
  FrameRateRangeEnum currentFrameRate_;
```

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -19,7 +39,10 @@ @implementation AnimationFrameQueue {
 - (instancetype)init
 {
   AssertJavaScriptQueue();
-  displayLink_ = [WorkletsDisplayLink displayLinkWithTarget:self selector:@selector(executeQueue:)];
+  bool supportsProMotion = [UIScreen mainScreen].maximumFramesPerSecond > 60;
+  SEL frameCallback = supportsProMotion ? @selector(executeQueueForProMotion:) : @selector(executeQueue:);
+  curentFrameRate = supportsProMotion ? BEST : STANDARD;
```
tomekzaw:
```suggestion
  currentFrameRate_ = supportsProMotion ? BEST : STANDARD;
```

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -65,12 +88,35 @@ - (void)executeQueue:(WorkletsDisplayLink *)displayLink
...
+  averageFrameDuration = averageFrameDuration / TIME_SAMPLES_AMOUNT;
+  if (averageFrameDuration < 8 && curentFrameRate != BEST) {
+    displayLink_.preferredFrameRateRange = FrameRateRange::BEST;
+  } else if (averageFrameDuration < 16 && curentFrameRate != STANDARD) {
+    displayLink_.preferredFrameRateRange = FrameRateRange::STANDARD;
+  } else if (averageFrameDuration < 33 && curentFrameRate != LOW) {
+    displayLink_.preferredFrameRateRange = FrameRateRange::LOW;
+  } else if (curentFrameRate != POOR) {
+    displayLink_.preferredFrameRateRange = FrameRateRange::POOR;
+  }
```
tomekzaw:
```suggestion
  FrameRageRange frameRageRange;
  if (averageFrameDuration < 8 && curentFrameRate != BEST) {
    preferredFrameRageRange = FrameRateRange::BEST;
  } else if (averageFrameDuration < 16 && curentFrameRate != STANDARD) {
    preferredFrameRageRange = FrameRateRange::STANDARD;
  } else if (averageFrameDuration < 33 && curentFrameRate != LOW) {
    preferredFrameRageRange = FrameRateRange::LOW;
  } else if (curentFrameRate != POOR) {
    preferredFrameRageRange = FrameRateRange::POOR;
  }
  displayLink_.preferredFrameRageRange = preferredFrameRageRange;
```

tomekzaw: btw why do we need `&& currentFrameRate != BEST` conditions?

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -2,14 +2,34 @@
...
 
 #import <React/RCTAssert.h>
 
+#define TIME_SAMPLES_AMOUNT 3
+
+namespace FrameRateRange {
+const CAFrameRateRange BEST = CAFrameRateRangeMake(60, 120, 120);
+const CAFrameRateRange STANDARD = CAFrameRateRangeMake(30, 60, 60);
+const CAFrameRateRange LOW = CAFrameRateRangeMake(24, 30, 30);
+const CAFrameRateRange POOR = CAFrameRateRangeMake(10, 24, 24);
```
tomekzaw:
```suggestion
constexpr auto BEST = CAFrameRateRangeMake(60, 120, 120);
constexpr auto STANDARD = CAFrameRateRangeMake(30, 60, 60);
constexpr auto LOW = CAFrameRateRangeMake(24, 30, 30);
constexpr auto POOR = CAFrameRateRangeMake(10, 24, 24);
```

piaskowyk: I've already tried but it doesn't works :/ 
image

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
piaskowyk: Setting `preferredFrameRateRange` shouldn't be expensive. but since we can't be sure about the overhead of the `preferredFrameRateRange` property (as its implementation is private). I found some speculation on how it might be implemented, and I think that it's safer to modify it only when we actually want to change it, rather than on every frame. Better safe than sorry.

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
piaskowyk: Solved - https://github.com/software-mansion/react-native-reanimated/pull/7624/commits/79bcedbfea376fc00d285060b66c8907d71ecafd

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
piaskowyk: Solved - https://github.com/software-mansion/react-native-reanimated/pull/7624/commits/79bcedbfea376fc00d285060b66c8907d71ecafd

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -2,14 +2,34 @@
...
+  LOW,
+  POOR,
+};
+
 @implementation AnimationFrameQueue {
   /* DisplayLink is thread safe. */
   WorkletsDisplayLink *displayLink_;
   std::vector<std::function<void(double)>> frameCallbacks_;
   std::mutex callbacksMutex_;
+  std::chrono::duration<double, std::milli> timeDeltas_[TIME_SAMPLES_AMOUNT];
```
tjzel, line 31: Where's `TIME_SAMPLES_AMOUNT` coming from?

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -65,12 +88,35 @@ - (void)executeQueue:(WorkletsDisplayLink *)displayLink
...
+  auto start = std::chrono::high_resolution_clock::now();
+
+  [self executeQueue:displayLink];
+
+  timeDeltaIndex_ = (timeDeltaIndex_ + 1) % TIME_SAMPLES_AMOUNT;
+  timeDeltas_[timeDeltaIndex_] = std::chrono::high_resolution_clock::now() - start;
+  float averageFrameDuration = 0;
+  for (int i = 0; i < TIME_SAMPLES_AMOUNT; i++) {
+    averageFrameDuration += timeDeltas_[i].count();
+  }
```
tjzel: How did you come up with this algorithm?

Also, do we really want to call it on each frame?

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
piaskowyk: It is defined at the beginning of the file - https://github.com/software-mansion/react-native-reanimated/blob/79bcedbfea376fc00d285060b66c8907d71ecafd/packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm#L9

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
piaskowyk: 1. This is just a circular buffer to compute the average computation time. As for how I came up with the sample amount - it was determined empirically 🙃
2. It's a trade-off between how often we want to compute it and how quickly we want to react.

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
tomekzaw: Okay, how about `const auto` instead?

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
tomekzaw: Is 3 enough for this mechanism to behave in a stable manner? How about 4 or 5 instead?

tjzel: I think 3.14 is the sweet spot

tomekzaw: 22/7

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
piaskowyk: done - https://github.com/software-mansion/react-native-reanimated/pull/7624/commits/a3fc88b4959a20021a7dc6434925bdc7defdac2f

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
piaskowyk: 3.14 ≈ 4 🫡 
https://github.com/software-mansion/react-native-reanimated/pull/7624/commits/87ad2f4d9109d278b8c295fe626489d5982343fb

packages/react-native-worklets/apple/worklets/apple/AnimationFrameQueue.mm
```diff
@@ -2,14 +2,34 @@
 #import <worklets/apple/AssertJavaScriptQueue.h>
 #import <worklets/apple/AssertTurboModuleManagerQueue.h>
 #import <worklets/apple/SlowAnimations.h>
+#import <chrono>
 
 #import <React/RCTAssert.h>
 
+#define TIME_SAMPLES_AMOUNT 4
```
tjzel:
```suggestion
constexpr auto TIME_SAMPLES_AMOUNT = 4;
```

tjzel (approved):

PR #7626, opened 2025-06-05
docs: Add neccessary CodeBlocks in TabItems

patrycjakalinska:
## Summary

They're needed around the fragments that we want to interpret as a code.

Before:
image

After:
image


packages/docs-reanimated/.prettierignore
```diff
@@ -0,0 +1 @@
+*.mdx
```
piaskowyk: Add empty line to the end please 🙏

piaskowyk (approved):

kacperkapusciak (approved):

PR #7631, opened 2025-06-05
docs: add Radon IDE banner to TOCItems

kacperkapusciak:


packages/docs-reanimated/src/components/RadonBanner/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+    button: 'Try Radon IDE for free',
+  },
+  {
+    text: 'Reduce context switching & streamline React Native and Expo development.',
+    button: 'Install Radon IDE',
+  },
+];
+
+export default function RadonBanner() {
+  const item = items[Math.floor(Math.random() * items.length)];
```
piaskowyk, line 49: Actually, there's a small chance that `Math.random()` will return 1, causing you to try to access an index outside the array 🫣

image

kacperkapusciak: Nope. `Math.random()` [never returns 1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random#return_value:~:text=between%200%20(inclusive)%20and%201%20(exclusive).).

PR #7632, opened 2025-06-05
feat(Worklets): no copy bundle on Android

tjzel:
## Summary

Making Android on par with iOS implementation of experimental bundling, by not copying the bundle at all.

As a bonus, Experimental Bundling works properly in production on Android.

Unfortunately I had to source set `WorkletsModule.java` because I needed to conditionally include a package from patched React Native.

## Test plan

I checked the following configurations:

The following configurations works:
- iOS debug
    - [x] legacy bundling
    - [x] experimental bundling
- iOS prod
    - [x] legacy bundling
    - [x] experimental bundling
- android debug
    - [x] legacy bundling
    - [x] experimental bundling
- android prod
    - [x] legacy bundling
    - [x] experimental bundling


piaskowyk (approved):
👍

PR #7638, opened 2025-06-09
chore(Worklets): detect runtime initialization failures with experimental bundling

tjzel:
## Summary

Throw a QoL error when something went wrong and we received a different error than expected during Worklet Runtime initialization when using experimental bundling.

I had to move `reportFatalErrorOnJS` to C++ to be able to invoke it on a wrong error. The changes are backwards compatible with legacy bundling.

## Test plan

Enable experimental bundling as per instructions in `packages/react-native-worklets/experimentalBundling.md` and see that there are no crashes.

If a Worklet Runtime failed to initialize, you'll get the following error in JS now (I modified `workletRuntimeEntry.ts` to achieve those configurations of errors):

| Initialization failure of UI Runtime (sync invocation) | Initialization failure of Worklet Runtime (sync invocation) | Initialization failure of Worklet Runtime (spawned from the UI Thread, async invocation) |
|:--------:|:--------:|:--------:|
| Screenshot 2025-06-16 at 19 48 33 | Screenshot 2025-06-16 at 19 50 05 | Screenshot 2025-06-16 at 20 00 53 |

For legacy bundling, throw an error in EmptyExample and see that it works just as before.


packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.cpp
```diff
@@ -124,7 +124,12 @@ WorkletRuntime::WorkletRuntime(
   try {
     rt.evaluateJavaScript(script, sourceUrl);
   } catch (facebook::jsi::JSIException ex) {
-    // Nothing
+    const auto exStr = std::string(ex.what());
+    if (!exStr.starts_with("Worklets initialized successfully")) {
+      throw std::runtime_error(
```
piaskowyk: Is it possible to throw jsi::Error on React Runtime from here? 🤔

tjzel: [2f07328](https://github.com/software-mansion/react-native-reanimated/pull/7638/commits/2f0732868e250d660d28d6663f3911bb18cfaf6c)

packages/react-native-worklets/src/errors.ts
```diff
@@ -42,22 +45,31 @@ function processStack(stack: string): string {
...
-  // @ts-ignore React Native's ErrorUtils implementation extends the Error type with jsEngine field
-  error.jsEngine = moduleName;
-  // @ts-ignore the reportFatalError method is an internal method of ErrorUtils not exposed in the type definitions
-  global.ErrorUtils.reportFatalError(error);
+  error.stack = processStack(stack);
+  error.name = name;
+  error.jsEngine = jsEngine;
+  if (force) {
+    throw error;
+  } else {
```
tjzel, line 68: When the error is thrown during initial evaluation, `reportFatalError` doesn't actually stop the evaluation which leads to a further crash in C++. That's why we must throw an error here instead for a meaningful error.

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.cpp
```diff
@@ -123,8 +124,23 @@ WorkletRuntime::WorkletRuntime(
 
   try {
     rt.evaluateJavaScript(script, sourceUrl);
-  } catch (facebook::jsi::JSIException ex) {
-    // Nothing
+  } catch (facebook::jsi::JSError error) {
+    const auto &message = error.getMessage();
+    const auto &stack = error.getStack();
+    if (!message.starts_with("Worklets initialized successfully")) {
+      jsScheduler->invokeSync_UNSAFE(
```
piaskowyk: How unsafe can `invokeSync_UNSAFE` be? 🤔

packages/react-native-worklets/src/WorkletsModule/NativeWorklets.ts
```diff
@@ -148,4 +148,18 @@ See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooti
       shareableWorklet
     );
   }
+
+  reportFatalErrorOnJS(
```
piaskowyk, line 152: Do we call it somewhere? What is the benefit of changing the implementation from pure JS (with runOnJS) to a C++ implementation?

packages/react-native-worklets/Common/cpp/worklets/Tools/JSLogger.h
```diff
@@ -7,12 +7,29 @@
...
+
 class JSLogger {
  public:
   explicit JSLogger(const std::shared_ptr<JSScheduler> &jsScheduler)
       : jsScheduler_(jsScheduler) {}
   void warnOnJS(const std::string &warning) const;
 
+  static void reportFatalErrorOnJS(
+      const std::shared_ptr<JSScheduler> &jsScheduler,
+      JSErrorData jsErrorData,
```
piaskowyk: probably this can be a `const JSErrorData &jsErrorData`

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.cpp
tjzel: Not unsafe anymore with #7663

packages/react-native-worklets/src/WorkletsModule/NativeWorklets.ts
tjzel: `NativeWorklets` is available on Worklet Runtimes and I use it in `callGuard.ts`, although not directly since I can't call `NativeWorklets` in `callGuardDev` **yet**.

packages/react-native-worklets/Common/cpp/worklets/Tools/JSLogger.h
tjzel: I cleaned up the implementation in [eb51c8a](https://github.com/software-mansion/react-native-reanimated/pull/7638/commits/eb51c8a72fba5efb7e9932fe911576e7dd15e272)[eb51c8a](https://github.com/software-mansion/react-native-reanimated/pull/7638/commits/eb51c8a72fba5efb7e9932fe911576e7dd15e272)

PR #7640, opened 2025-06-09
feat: remove all legacy arch code from libraries

tjzel:
## Summary

Making `react-native-worklets` New Arch - only and removing some Paper artifacts from the code.

## Test plan

CI

### DIFF
--- a/packages/react-native-reanimated/android/build.gradle
+++ b/packages/react-native-reanimated/android/build.gradle
@@ -70,6 +70,7 @@ if (isNewArchitectureEnabled()) {
 def reactNativeRootDir = resolveReactNativeDirectory()
 def REACT_NATIVE_MINOR_VERSION = getReactNativeMinorVersion()
 def REANIMATED_VERSION = getReanimatedVersion()
+def IS_NEW_ARCHITECTURE_ENABLED = isNewArchitectureEnabled()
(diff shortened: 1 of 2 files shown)


packages/react-native-reanimated/android/build.gradle
```diff
@@ -236,7 +234,7 @@ task assertMinimalReactNativeVersionTask {
 preBuild.dependsOn(assertMinimalReactNativeVersionTask)
 
 task assertNewArchitectureEnabledTask {
-    onlyIf { !IS_NEW_ARCHITECTURE_ENABLED }
+    onlyIf { !isNewArchitectureEnabled() }
```
piaskowyk: I think we had some issues in the past with calling methods within the task, but I can't remember exactly what the issue was - it might have been related to the Gradle cache, but I'm not sure. However, we can merge it like this, but we need to be prepared for potential fix.

tomekzaw: +1, let's avoid using `isNewArchitectureEnabled` inside `onlyIf`

tomekzaw (approved):

packages/react-native-reanimated/android/build.gradle
tjzel: I'll keep the variable then 👍

PR #7641, opened 2025-06-09
feat(Worklets): replace `_WORKLET` with `getRuntimeKind` and `__RUNTIME_KIND`

tjzel:
## Summary

Replacing the pattern of using `_WORKLET` global variable to differentiate between runtimes with programmable API: `getRuntimeKind` and `RuntimeKind` enum.

RuntimeKind differentiates between React Native, UI, and extra Worklet Runtimes.

To allow tight optimization we also add a global variable `__RUNTIME_KIND` that can be used to avoid an extra function call and potential serialization/deserialization of `getRuntimeKind`.

Keeping `_WORKLET` for backwards compatibility with a deprecation message.

## Test plan

Runtime tests.


tomekzaw:
I think we'd like to keep `_WORKLET` as deprecated for backwards-compatibility reasons.

tjzel:
@tomekzaw It's kept, I didn't remove it.

packages/react-native-worklets/src/runtimeKind.ts
```diff
@@ -0,0 +1,37 @@
...
+export function getRuntimeKind(): RuntimeKind {
+  'worklet';
+  return globalThis.__RUNTIME_KIND;
+}
+
+if (globalThis.__RUNTIME_KIND === undefined) {
+  // In Jest environments eager imports make this file to evaluate before
+  // `initializers.ts` file, therefore we have to set the RuntimeKind here,
+  // just to be safe.
+  globalThis.__RUNTIME_KIND = RuntimeKind.ReactNative;
```
piaskowyk, line 36: Do we need the same code in the `initializers.ts` file?

packages/react-native-worklets/src/specs/index.ts
```diff
@@ -1,8 +1,10 @@
...
 import type { Spec } from './NativeWorkletsModule';
 import RNWorkletsTurboModule from './NativeWorkletsModule';
 
-export const WorkletsTurboModule: Spec | null = globalThis._WORKLET
-  ? null
-  : RNWorkletsTurboModule;
+export const WorkletsTurboModule: Spec | null =
+  globalThis.__RUNTIME_KIND === RuntimeKind.ReactNative
+    ? RNWorkletsTurboModule
+    : null;
```
piaskowyk: I know this is outside the scope of this PR, but could you add a comment to explain that condition? It can be done in another PR, but I think it's worth it.

packages/react-native-worklets/src/runtimeKind.ts
tjzel: Yes. With lazy imports (that's how Metro does it) `init` function from `initializers.ts` gets invoked first. However in Jest, which does eager imports, `runtimeKind.ts` gets evaluated first. This is why for safety I need to do it in two locations.

packages/react-native-worklets/src/specs/index.ts
tjzel: Sure, I'll add it.

PR #7642, opened 2025-06-09
Refactor props allowlists

tomekzaw:
## Summary

Work in progress.


tomekzaw:
TODO: fix class name for `RCTView`/`View` and `RCTSinglelineTextInputView`/`TextInput`

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/AnimatedPropsRegistry.h
```diff
@@ -6,12 +6,13 @@
...
 #include <string>
-#include <utility>
+#include <tuple>
 #include <vector>
 
 namespace reanimated {
 
-using JSIUpdates = std::vector<std::pair<Tag, std::unique_ptr<jsi::Value>>>;
+using JSIUpdates =
+    std::vector<std::tuple<Tag, std::string, std::unique_ptr<jsi::Value>>>;
```
MatiPl01: Won't it be more readable if we used a `struct` instead of a `tuple` here? I think that having named fields is better than having a tuple with multiple values.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -559,27 +562,33 @@ void ReanimatedModuleProxy::unregisterCSSTransition(
...
+  const auto found = it != nativePropNamesForComponentNames_.end();
+  react_native_assert(
+      found && "Native props not found for given component name");
+  const auto &nativePropNamesForComponentName = it->second;
   for (size_t i = 0; i < propNames.size(rt); ++i) {
     const std::string &propName =
         propNames.getValueAtIndex(rt, i).asString(rt).utf8(rt);
-    if (!collection::contains(animatablePropNames_, propName)) {
-      hasAnyNonAnimatableProp = true;
+    if (!collection::contains(nativePropNamesForComponentName, propName)) {
```
MatiPl01:
```suggestion
    if (!nativePropNamesForComponentName.contains(propName)) {
```

Can't we just call the `.contains` method on the `nativePropNamesForComponentName` without using `collection::contains`?

packages/react-native-reanimated/src/ConfigHelper.ts
```diff
@@ -2,57 +2,23 @@
...
+import { SHOULD_BE_USE_WEB } from './common';
+import type { ShadowNodeWrapper } from './commonTypes';
+import {
+  executeOnUIRuntimeSync,
+  jsiRegisterNativePropNamesForComponentName,
+} from './core';
+
+export function addWhitelistedNativeProps(): void {
+  console.warn(
+    '[Reanimated] `addWhitelistedNativeProps` is no longer necessary in Reanimated 4 and will be removed in next version. Please remove this call from your code.'
```
MatiPl01:
```suggestion
  logger.warn(
    '`addWhitelistedNativeProps` is no longer necessary in Reanimated 4 and will be removed in next version. Please remove this call from your code.'
```

I know that we now export `logger` from the `react-native-worklets` package which is not valid as it adds the `[Worklets]` to all logs but I think we should refactor it later, since all logs in reanimated are now handled in this way and it would be easier to keep things consistent in order not to forget to change these logs in the future.

packages/react-native-reanimated/src/ConfigHelper.ts
```diff
@@ -2,57 +2,23 @@
...
-    ...props,
-  };
-  if (
-    oldSize !== Object.keys(PropsAllowlists.UI_THREAD_PROPS_WHITELIST).length
-  ) {
-    configureProps();
-  }
+export function addWhitelistedUIProps(): void {
+  console.warn(
+    '[Reanimated] `addWhitelistedUIProps` is no longer necessary in Reanimated 4 and will be removed in next version. Please remove this call from your code.'
```
MatiPl01:
```suggestion
  logger.warn(
    '`addWhitelistedUIProps` is no longer necessary in Reanimated 4 and will be removed in next version. Please remove this call from your code.'
```

packages/react-native-reanimated/src/PropAdapters.ts
```diff
@@ -13,12 +12,10 @@ export function createAnimatedPropAdapter(
...
+  _nativeProps?: string[]
 ): AnimatedPropsAdapterWorklet {
-  const nativePropsToAdd: { [key: string]: boolean } = {};
-  nativeProps?.forEach((prop) => {
-    nativePropsToAdd[prop] = true;
-  });
-  addWhitelistedNativeProps(nativePropsToAdd);
+  console.warn(
+    '[Reanimated] `createAnimatedPropAdapter` is no longer necessary in Reanimated 4 and will be removed in next version. Please remove this call from your code and pass the adapter function directly.'
+  );
```
MatiPl01:
```suggestion
  logger.warn(
    '`createAnimatedPropAdapter` is no longer necessary in Reanimated 4 and will be removed in next version. Please remove this call from your code and pass the adapter function directly.'
  );
```

MatiPl01 (changes requested):
In general, it looks quite well, except some minor suggestions and one problem I noticed when I navigated between `Animations -> Transitions` tabs in the CSS example app

Screenshot 2025-06-11 at 11 18 43

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/AnimatedPropsRegistry.h
tjzel: structs in modern C++ are really convenient to use, +1 on that

packages/react-native-reanimated/src/ConfigHelper.ts
tjzel: Where will we provide instructions on how to migrate from `addWhitlelistedUIProps`?

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -329,20 +329,23 @@ jsi::Value ReanimatedModuleProxy::enableLayoutAnimations(
...
-    animatablePropNames_.insert(name);
-  }
-  auto nativePropsArray = nativeProps.asObject(rt).asArray(rt);
-  for (size_t i = 0; i < nativePropsArray.size(rt); ++i) {
-    auto name = nativePropsArray.getValueAtIndex(rt, i).asString(rt).utf8(rt);
-    animatablePropNames_.insert(name);
+    const jsi::Value &shadowNodeWrapper,
+    const jsi::Value &nativePropNames) {
+  const auto shadowNode = shadowNodeFromValue(rt, shadowNodeWrapper);
+  const auto componentNameStr = shadowNode->getComponentName();
```
tjzel: Just to make sure, are components names guaranteed to be unique?

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
tomekzaw: I wanted to make it analogous to the previous code which had to be compatible with C++17 but since we already have C++20 enabled I can change that.

tomekzaw:
TODO: instead of detecting native props automatically, let's manually register JS props for a given component

edit: Done in e8ab843f9139abd81cbb875d6d5f85e4e5ac0264.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/AnimatedPropsRegistry.h
tomekzaw: Done in 983ce196f3

packages/react-native-reanimated/src/ConfigHelper.ts
tomekzaw: Done in b9d2d95f8e

packages/react-native-reanimated/src/ConfigHelper.ts
tomekzaw: @MatiPl01 Done in b9d2d95f8e

tomekzaw: @tjzel Most likely in the migration guide in the docs, once the PR is merged I plan to submit a follow-up PR with changes to docs and add a link to migration guide here.

packages/react-native-reanimated/src/PropAdapters.ts
tomekzaw: Done in b9d2d95f8e

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
tomekzaw: Not really, for instance `<TextInput />` has a `uiViewClassName` of `"RCTSinglelineTextInput"` and `<TextInput multiline />` has`"RCTMultilineTextInput"`, but they both translate to `ViewComponentView` which has `componentName` of `"Text"`.

tomekzaw:
TODO: add page in docs describing how to animate native and JS props

edit: This will be done in a separate PR.

apps/common-app/src/apps/reanimated/examples/ThirdPartyComponentsExample.tsx
```diff
@@ -0,0 +1,179 @@
...
+          animatedProps={animatedProps}
+        />
+      </Svg>
+    </View>
+  );
+}
+
+const AnimatedG = Animated.createAnimatedComponent(G);
+
+registerJSProps('RNSVGGroup', ['x', 'y']);
```
MatiPl01, line 112: These names of components are always unique, right?

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/AnimatedPropsRegistry.cpp
```diff
@@ -1,5 +1,7 @@
 #include <reanimated/Fabric/updates/AnimatedPropsRegistry.h>
 
+#include <utility>
```
MatiPl01, line 3: Can you include headers in the `.h` file? It just my preference to keep all headers in the `.h` file and have just a single include in the `.cpp` file.

This is not mandatory, so you can leave it as it is.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/AnimatedPropsRegistry.cpp
```diff
@@ -19,8 +21,11 @@ SurfaceId AnimatedPropsRegistry::update(
 
     const jsi::Value &updates = item.getProperty(rt, "updates");
     addUpdatesToBatch(shadowNode, jsi::dynamicFromValue(rt, updates));
-    jsiUpdates_.emplace_back(
-        shadowNode->getTag(), std::make_unique<jsi::Value>(rt, updates));
+    JSIUpdate update{
+        /* .tag = */ shadowNode->getTag(),
+        /* .componentName = */ shadowNode->getComponentName(),
+        /* .props = */ std::make_unique<jsi::Value>(rt, updates)};
```
MatiPl01:
```suggestion
        .tag = shadowNode->getTag(),
        .componentName = shadowNode->getComponentName(),
        .props = std::make_unique<jsi::Value>(rt, updates)};
```

I think that you can list prop names explicitly without using comments, which seems to be cleaner to me.

apps/common-app/src/apps/reanimated/examples/ThirdPartyComponentsExample.tsx
```diff
@@ -0,0 +1,179 @@
...
+      <SvgRectDemo sv={sv} />
+      <SvgPathDemo sv={sv} />
+      <SvgGDemo sv={sv} />
+    </ScrollView>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
```
MatiPl01:
```suggestion
    paddingVertical: 40,
```
Don't set `flex: 1` on the content container as it makes it impossible to scroll.
Rather add some padding to make it look better and not render content below the bottom nav bar.

https://github.com/user-attachments/assets/0bfdb6f8-b4ac-4451-94e6-7cbc00d83a76

tomekzaw: I forgot to remove `flex: 1` when I changed `<View>` to `<ScrollView>`. Done in d06bef39db. Thanks!

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/AnimatedPropsRegistry.cpp
tomekzaw: I remember times when some compilers didn't support `.foo =` syntax so wrapping those in comments was a common pattern. I even remember seeing this in react-native codebase itself. Now that I've checked they no longer use this pattern so I think we're safe to use it too. https://github.com/facebook/react-native/blob/ffe928ce508950d91f6a5ac5fde782402f2c5b6a/packages/react-native/ReactCommon/react/renderer/uimanager/UIManagerBinding.cpp#L460

Done in 27c70fc96c.

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/AnimatedPropsRegistry.cpp
tomekzaw: I always try to apply to the following rule – if some header is not "used" in the header file and only used in the `.cpp` file, I place it in the `.cpp` file instead of in the `.h` file since it's an implementation detail. Then, if you remove its usage in the `.cpp` file, the `.h` file doesn't change and other compilation units don't need to be recompiled. In this case, I need `#include <utility>` for the `std::move` call in the `.cpp` file.

apps/common-app/src/apps/reanimated/examples/ThirdPartyComponentsExample.tsx
tomekzaw: I don't think they are guaranteed to be unique but I'm pretty sure things will go wild if two libraries decide to use the same component name.

tomekzaw:
> The app freezes when I press R to reload on the ThirdPartyComponentsExample screen. Please fix that and other smaller issues and will be good to merge as the rest looks well.

@MatiPl01 I was able to reproduce the same crash on current main (81730630fa) so this PR doesn't introduce a regression.

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -328,20 +328,20 @@ jsi::Value ReanimatedModuleProxy::enableLayoutAnimations(
...
+  const auto componentNameStr = componentName.asString(rt).utf8(rt);
+  const auto jsPropsNamesArray = jsPropsNames.asObject(rt).asArray(rt);
+  const auto size = jsPropsNamesArray.size(rt);
+  std::unordered_set<std::string> jsPropsNamesSet;
+  for (size_t i = 0; i < size; i++) {
+    jsPropsNamesSet.insert(
+        jsPropsNamesArray.getValueAtIndex(rt, i).asString(rt).utf8(rt));
   }
+  auto lock = std::unique_lock<std::mutex>(jsPropsNamesForComponentNamesMutex_);
+  jsPropsNamesForComponentNames_[componentNameStr] = std::move(jsPropsNamesSet);
```
bartlomiejbloniarz, line 344: Should calling `registerJSProps` override previous calls?

bartlomiejbloniarz (approved):

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
MatiPl01: Good question. I think that nobody will need to call it multiple times but maybe there are cases in which someone would register some props in one call and rest in another one.

For me, the current implementation is fine. We can always add props merging in the future if it turns out that we need it
WDYT?

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.h
```diff
@@ -76,10 +77,10 @@ class ReanimatedModuleProxy
 
   jsi::Value enableLayoutAnimations(jsi::Runtime &rt, const jsi::Value &config)
       override;
-  jsi::Value configureProps(
+  jsi::Value registerJSProps(
       jsi::Runtime &rt,
-      const jsi::Value &uiProps,
-      const jsi::Value &nativeProps) override;
+      const jsi::Value &componentNames,
```
piaskowyk:
```suggestion
      const jsi::Value &componentName,
```

piaskowyk (approved):

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.h
tomekzaw: Good catch! Fixed in 629c510a1b

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
tomekzaw: @bartlomiejbloniarz Assuming we want to rewrite JS props handling logic from C++ to TS later on, is it okay if we leave it as it is for now?

bartlomiejbloniarz: I think we can leave it as it is

tomekzaw: Good, otherwise we would have to revert the PR 😢

PR #7645, opened 2025-06-10
Use `react_native_assert` instead of `assert`

tomekzaw:
## Summary

This PR converts all usages of `assert()` into `react_native_assert()` which offers better logging as well as adds missing error messages.


packages/react-native-worklets/Common/cpp/worklets/SharedItems/Shareables.h
```diff
@@ -258,9 +258,7 @@ class ShareableHostFunction : public Shareable {
  public:
   ShareableHostFunction(jsi::Runtime &rt, jsi::Function function)
       : Shareable(HostFunctionType),
-        hostFunction_(
-            (assert(function.isHostFunction(rt)),
-             function.getHostFunction(rt))),
+        hostFunction_(function.getHostFunction(rt)),
```
tomekzaw, line 261: `getHostFunction` already calls `assert(function.isHostFunction(rt))` on its own.

piaskowyk (approved):

PR #7646, opened 2025-06-10
feat: Better inline css animation keyframes handling

MatiPl01:


MatiPl01:
Superseded by #7733 - inline keyframes now reuse a registered rule through the `cssText` -> name index in `CSSKeyframesRegistry`. The C++ file this branch rewrites (`Common/NativeView/.../CSSAnimationsManager.cpp`) no longer exists on `main`.

PR #7648, opened 2025-06-11
chore: Separate reanimated logger from worklets logger

MatiPl01:
## Summary

This PR adds a separate logger for the reanimated package, because the previous logger was moved to the `react-native-worklets` package and logs started being prefixed with `[Worklets]`.

The new separate logger prefixes all messages with a proper `[Reanimated]` text.

This PR also removes the usage of `LogBox` which no longer works for warnings after the new devtools were introduced.

### [Worklets] Tried to modify key...

<details>

```tsx
import { useAnimatedReaction } from 'react-native-reanimated';

const TEST = {
  a: 1,
  b: 2,
};

  useAnimatedReaction(
    () => {},
    () => {
      console.log(TEST);
    }
  );

  useEffect(() => {
    TEST.a = 0;
  }, []);

  return (
      <Text>Hello world!</Text>
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```
</details>

Screenshot 2025-06-11 at 15 17 48

### [Reanimated] Invalid read/write during render

Just open the `Invalid read/write during render` example in the example app

Screenshot 2025-06-11 at 15 21 45

### [Reanimated]

<details>

```tsx
import React from 'react';
import { StyleSheet, Text } from 'react-native';
import Animated, { FadeOut, useAnimatedStyle } from 'react-native-reanimated';

    return {
      opacity: 0,
    };
  });

  return (
    <Animated.View style={[styles.container, animatedStyle]} exiting={FadeOut}>
      <Text>Hello world!</Text>
    </Animated.View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```
</details>

Screenshot 2025-06-11 at 15 38 21


packages/react-native-reanimated/src/component/ReducedMotionConfig.tsx
```diff
@@ -22,7 +22,7 @@ export function ReducedMotionConfig({ mode }: { mode: ReduceMotion }) {
       return;
     }
     logger.warn(`Reduced motion setting is overwritten with mode '${mode}'.`);
-  }, []);
+  }, [mode]);
```
MatiPl01: I think this one was missing here. At least eslint complained about this.

tjzel: I assume it wasn't lister here for a reason. cc @bartlomiejbloniarz

packages/react-native-worklets/src/logger.ts
tjzel, line 1: Actually `react-native-worklets` uses the logger only in `shareables.ts` on the RN Runtime. For this purpose I think we can remove `__workletsLoggerConfig` and all worklets here entirely, only leaving `export logger`.

MatiPl01: done

packages/react-native-reanimated/src/index.ts
```diff
@@ -31,7 +32,7 @@ export {
 } from './animation';
 export type { ParsedColorArray } from './Colors';
 export { convertToRGBA, isColor } from './Colors';
-export { processColor } from './common';
+export { processColor, LogLevel as ReanimatedLogLevel } from './common';
```
tjzel: Why do we need to rename it here?

MatiPl01: We don't need it but:
1. It is more explicit and prevents potential conflicts
2. It was implemented in such a way before and is described in this way in [docs](https://docs.swmansion.com/react-native-reanimated/docs/debugging/logger-configuration/#reference)

tjzel: Can we then rename it inside the library too, for the sake of transparency?

packages/react-native-reanimated/src/component/ReducedMotionConfig.tsx
bartlomiejbloniarz: I think we just wanted to warn here once?

bartlomiejbloniarz: It was @piaskowyk I think that added this component. I think it was better the way it was before, we don't need to repeat the message

MatiPl01: Ok, I will restore the previous implementation. This PR didn't mean to change the behavior so I think it would be safer to keep the previous one.

PR #7650, opened 2025-06-11
Remove integration with `react-native-v8` and `rnRuntime` argument from `WorkletRuntime` constructor

tomekzaw:
## Summary

This PR removes integration with `react-native-v8` and thus the possibility to use V8 engine for Reanimated UI runtime. We no longer actively support V8 and the project itself seems to be abandoned as there hasn't been any new commits since 10 months ago (as of June 2025).

If you are affected by this change, please to leave a comment under this PR and explain your motivation to use V8 engine with Reanimated. We might consider restoring support for V8 engine at some point in the future.

## Test plan

Build fabric-example app for Android.


packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
```diff
@@ -69,7 +69,6 @@ inline jsi::Value createWorkletRuntime(
     const jsi::Value &name,
     const jsi::Value &initializer) {
   auto workletRuntime = std::make_shared<WorkletRuntime>(
-      rt,
```
tjzel, line 72: Actually I need that in parameter in #7638 to throw an error on RN Runtime synchronously.

tjzel (approved):
LGTM, but I actually need that parameter in `WorkletRuntime` 😁

packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
tomekzaw: Do you really really need it? I've always wanted to remove it 😢

tjzel: I can possibly try `invokeSync` with JSScheduler, assuming it won't be crashing 🤔

tomekzaw: `invokeSync` has been broken for ages.

PR #7652, opened 2025-06-11
refactor(Worklets): define JS initialization pipeline

tjzel:
## Summary

Initialization in `react-native-worklets` is spread out through several lines of code without a proper pipeline. This pull requests organizes these bits for more transparency on why and when do we call some side-effects.

## Test plan

I tested WebExample and FabricExample with `experimentalBundling` off and on and noticed no regressions.


packages/react-native-worklets/src/index.ts
```diff
@@ -34,3 +29,10 @@ export type {
   WorkletRuntime,
   WorkletStackDetails,
 } from './workletTypes';
+
+// @ts-expect-error We must trick the bundler to include
+// the `workletRuntimeEntry` file the way it cannot optimize it out.
+if (globalThis._ALWAYS_FALSE) {
+  // Experimental bundling.
+  experimentalBundlingInit();
+}
```
tjzel, line 37: This unfortunately has to be in the `index.ts` file to avoid circular dependencies, it calls `init()` from `initializers.ts`.

packages/react-native-worklets/src/initializers.ts
```diff
@@ -5,49 +5,13 @@ import { setupRequestAnimationFrame } from './animationFrameQueue/requestAnimati
...
-  // Experimental bundling.
-  initializeLibraryOnWorkletRuntime();
-}
-
-// this is for web implementation
-if (SHOULD_BE_USE_WEB) {
-  globalThis._WORKLET = false;
-  globalThis._log = console.log;
-  globalThis._getAnimationTimestamp = () => performance.now();
-} else if (!globalThis._WORKLET) {
```
tjzel, line 32: Moved to `initializeRuntimeOnWeb`.

packages/react-native-worklets/src/initializers.ts
```diff
@@ -5,49 +5,13 @@ import { setupRequestAnimationFrame } from './animationFrameQueue/requestAnimati
...
-        `Failed to create a worklet. See https://docs.swmansion.com/react-native-reanimated/docs/guides/troubleshooting#failed-to-create-a-worklet for more details.`
-      );
-    }
-  }
-
-  // Register WorkletsError and logger config in the UI runtime global scope.
-  // (we are using `executeOnUIRuntimeSync` here to make sure that the changes
-  // are applied before any async operations are executed on the UI runtime)
-  executeOnUIRuntimeSync(registerWorkletsError)();
-}
```
tjzel, line 49: Moved to `initializeRNRuntime` and `installRNBindingsOnUIRuntime`.

packages/react-native-worklets/package.json
```diff
@@ -83,6 +83,7 @@
     "react": "19.1.0",
     "react-native": "patch:react-native@npm%3A0.80.0#~/.yarn/patches/react-native-npm-0.80.0-dababd395b.patch",
     "react-native-builder-bob": "patch:react-native-builder-bob@npm%3A0.33.1#~/.yarn/patches/react-native-builder-bob-npm-0.33.1-383d9e23a5.patch",
+    "ts-prune": "^0.10.3",
```
MatiPl01, line 85: `ts-prune` hasn't been updated for 4 years. Maybe we could replace it by [knip](https://knip.dev/)? Not necessarily now but we can do this in the future.

packages/react-native-worklets/src/initializers.ts
```diff
@@ -110,15 +74,65 @@ export function setupConsole(boundCapturableConsole: typeof console) {
...
+    initializeWorkletRuntime();
+  } else {
+    initializeRNRuntime();
+    if (!SHOULD_BE_USE_WEB) {
+      installRNBindingsOnUIRuntime();
+    }
   }
+}
+
+/** A function that should be ran on any kind of runtime. */
```
MatiPl01:
```suggestion
/** A function that should be run on any kind of runtime. */
```

packages/react-native-worklets/package.json
tjzel: Sure, we can do it soon<tm>

PR #7653, opened 2025-06-11
chore(Worklets): enable ts-prune

tjzel:
## Summary

Turns out `ts-prune` wasn't installed properly in Worklets' workspace and it didn't work. I fixed all errors it found.

## Test plan

🚀

### DIFF
--- a/packages/react-native-worklets/package.json
+++ b/packages/react-native-worklets/package.json
@@ -83,6 +83,7 @@
     "react": "19.1.0",
     "react-native": "patch:react-native@npm%3A0.80.0-rc.4#~/.yarn/patches/react-native-npm-0.80.0-rc.4-ad01aea617.patch",
     "react-native-builder-bob": "patch:react-native-builder-bob@npm%3A0.33.1#~/.yarn/patches/react-native-builder-bob-npm-0.33.1-383d9e23a5.patch",
+    "ts-prune": "*",
     "typescript": "~5.3.0"
   },
   "main": "./lib/module/index",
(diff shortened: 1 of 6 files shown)


packages/react-native-worklets/package.json
```diff
@@ -83,6 +83,7 @@
     "react": "19.1.0",
     "react-native": "patch:react-native@npm%3A0.80.0-rc.4#~/.yarn/patches/react-native-npm-0.80.0-rc.4-ad01aea617.patch",
     "react-native-builder-bob": "patch:react-native-builder-bob@npm%3A0.33.1#~/.yarn/patches/react-native-builder-bob-npm-0.33.1-383d9e23a5.patch",
+    "ts-prune": "*",
```
MatiPl01, line 86: Why `"ts-prune": "*"` instead of a valid version? We have this in the reanimated package: `"ts-prune": "^0.10.3"`, so shouldn't we use an explicit version in here as well?

tjzel: This is a monorepo thing that we should've done a long time ago. Putting "*" means use a version available in the monorepo, so you don't have to duplicate the versions in each `package.json`

MatiPl01: Yeah, I know, but I don't like treating the version from `packages/react-native-reanimated` as the main version and putting `*` in all other places. I think that if you want to apply this approach, I would suggest adding `ts-prune` package in the top-level `package.json` file with a proper version and setting `*` in other packages (or omitting them at all, because you shouldn't need to specify `ts-prune` in every package if it is installed from the root one).

tjzel: That's what I want to do in the future 🙏

MatiPl01: Ok, so I'd prefer to keep a correct version in this package for consistency with other dependencies and change it in the future but this is just my suggestion, so you can merge this PR without changing it.

tjzel: I'll put a proper version instead then 👍

PR #7657, opened 2025-06-13
Bump fabric-example to 0.80.0

tomekzaw:
## Summary

This PR bumps `react-native` in fabric-example to 0.80.0 and `react-native-gesture-handler` to 2.26.0.


.yarn/patches/react-native-npm-0.80.0-dababd395b.patch
```diff
@@ -1,3 +1,39 @@
+diff --git a/Libraries/AppDelegate/RCTAppSetupUtils.mm b/Libraries/AppDelegate/RCTAppSetupUtils.mm
```
MatiPl01, line 1: Why do you need to patch `react-native`? Is this needed because of not yet released changes we need in worklets?

package.json
```diff
@@ -40,6 +40,9 @@
     "typescript": "~5.3.0"
   },
   "resolutions": {
-    "@react-native/community-cli-plugin@npm:0.80.0-rc.4": "patch:@react-native/community-cli-plugin@npm%3A0.80.0-rc.4#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-rc.4-af2762c07e.patch"
+    "@react-native/community-cli-plugin": "0.80.0",
+    "@react-native/community-cli-plugin@npm:0.75.4": "patch:@react-native/community-cli-plugin@npm%3A0.80.0#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-4137b3f35c.patch",
+    "@react-native/community-cli-plugin@npm:0.79.1": "patch:@react-native/community-cli-plugin@npm%3A0.80.0#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-4137b3f35c.patch",
+    "@react-native/community-cli-plugin@npm:0.80.0": "patch:@react-native/community-cli-plugin@npm%3A0.80.0#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-4137b3f35c.patch"
```
MatiPl01: What happened here? Why all `@react-native/community-cli-plugin` point to `0.80.0` patch?

I also don't like using resolutions `"@react-native/community-cli-plugin": "0.80.0"`. It usually seems that the problem is located somewhere else in dependencies and this may just break some of our example apps again, as during previous react-native upgrade.

.yarn/patches/react-native-npm-0.80.0-dababd395b.patch
tomekzaw: Yes, and I have just updated the patch added by @tjzel.

package.json
tomekzaw: > What happened here?

This is what I got after running `yarn patch @react-native/community-cli-plugin`, I haven't noticed that yarn has patched all versions in the repo. Fixed in 230f7a919d

tomekzaw: > I also don't like using resolutions

Removed `resolutions` in e2706874a9

PR #7660, opened 2025-06-16
Remove CollectionUtils.h

tomekzaw:
## Summary

This PR replaces `collection::contains` function calls with `.contains()` method calls (available from C++20) as well as removes unused `collection::contains` function.


MatiPl01 (approved):
LGTM

PR #7663, opened 2025-06-16
feat(Worklets): add invokeSync in JSScheduler

tjzel:
## Summary

Motivated by #7638, it would be handy to invoke something synchronously and safely on the RN Runtime. Since `jsCallInvoker` deadlocks on `invokeSync`, I added `invokeSync` to `JSScheduler` that also checks whether such call is safe.

## Test plan

Nothing so far, there's no use of `invokeSyncOnJS`.


tomekzaw:
I don't think `canInvokeSyncOnJS` needs to be public. How about making it private? I'm afraid it can be misused sometime in the future if we keep it public.

tjzel:
@tomekzaw Could you elaborate on that? I added it specifically to be able to know if I should make a sync or an async call.

PR #7664, opened 2025-06-16
fix(Worklets): re-apply yarn patch for experimental bundling

tjzel:
## Summary

Resolution for `@react-native/community-cli-plugin` was forgotten in #7657.

## Test plan

Experimental bundling no longer fails on `_fbBatchedBridge` missing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,5 +38,8 @@
     "prettier": "^3.3.3",
     "prettier-plugin-jsdoc": "^1.3.0",
     "typescript": "~5.3.0"
+  },
+  "resolutions": {
+    "@react-native/community-cli-plugin@npm:0.80.0": "patch:@react-native/community-cli-plugin@npm%3A0.80.0#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-4137b3f35c.patch"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -5617,6 +5617,27 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@react-native/community-cli-plugin@patch:@react-native/community-cli-plugin@npm%3A0.80.0#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-4137b3f35c.patch":
+  version: 0.80.0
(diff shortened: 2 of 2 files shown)


tomekzaw:
I've removed `resolutions` field on purpose (after updating `yarn.lock`) according to @MatiPl01's suggestion in order to avoid breaking other apps like tvos-example (which required @patrycjakalinska's fix https://github.com/software-mansion/react-native-reanimated/pull/7600).

tjzel:
This resolution is harmless for other apps, it only allows `getModulesRunBeforeMainModule` metro config field to not be overridden, we only use that field in experimental bundling in `FabricExample`.

MatiPl01:
> I've removed `resolutions` field on purpose (after updating `yarn.lock`) according to @MatiPl01's suggestion in order to avoid breaking other apps like tvos-example (which required @patrycjakalinska's fix #7600).

I was mainy concerned about [this](https://github.com/software-mansion/react-native-reanimated/pull/7657#discussion_r2147364915), where resolution of `@react-native/community-cli-plugin` was set to `"@react-native/community-cli-plugin": "0.80.0"` in the top level `package.json`, which results in overriding `@react-native/community-cli-plugin` version in every subpackage.

Also all of these resolutions were pointing to the 0.80 patch, which doesn't seem to be right:
```json
"@react-native/community-cli-plugin@npm:0.75.4": "patch:@react-native/community-cli-plugin@npm%3A0.80.0#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-4137b3f35c.patch",
"@react-native/community-cli-plugin@npm:0.79.1": "patch:@react-native/community-cli-plugin@npm%3A0.80.0#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-4137b3f35c.patch",
"@react-native/community-cli-plugin@npm:0.80.0": "patch:@react-native/community-cli-plugin@npm%3A0.80.0#~/.yarn/patches/@react-native-community-cli-plugin-npm-0.80.0-4137b3f35c.patch"
```

tjzel:
This is overeagerness of yarn when patching, but you can set the resolution only for one given version as I did in this PR.

tjzel:
I just tested `tvos-example` with this setup and it works @MatiPl01 👍

MatiPl01:
> I just tested `tvos-example` with this setup and it works @MatiPl01 👍

Yup, resolutions just for the specific version `@react-native/community-cli-plugin@npm:0.80.0` work fine. The previous approach in which all versions were pointing to the 0.80 patch was concerning, but this one is correct.

MatiPl01 (approved):

PR #7667, opened 2025-06-17
Don't run exiting animations on modal removal

bartlomiejbloniarz:
## Summary

We don't run exiting animations when a screen is removed. This PR extends this behavior to modals.

This should resolve the [issue from rn-srceens](https://github.com/software-mansion/react-native-screens/issues/2545).

## Test plan
Check layout animations examples for regressions. Check if the new `[LA] Modal exiting example` example works properly.

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
@@ -135,13 +135,10 @@ static inline void updateLayoutMetrics(
 }
 
 static inline bool isRNSScreen(std::shared_ptr<MutationNode> node) {
-  return !std::strcmp(
-             node->mutation.oldChildShadowView.componentName,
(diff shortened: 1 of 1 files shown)


piaskowyk (approved):
🫡

packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
```diff
@@ -139,7 +139,9 @@ static inline bool isRNSScreen(std::shared_ptr<MutationNode> node) {
              node->mutation.oldChildShadowView.componentName,
```
tomekzaw: How about this?
```cpp
static inline bool isRNSScreen(std::shared_ptr<MutationNode> node) {
  const auto &componentName = node->mutation.oldChildShadowView.componentName;
  return !std::strcmp(componentName, "RNSScreenStack") ||
      !std::strcmp(componentName, "RNSScreen") ||
      !std::strcmp(componentName, "RNSModalScreen");
}
```

PR #7680, opened 2025-06-23
fix: `skipExiting` not working on Android

MatiPl01:
### Issue description

For some reason, `componentWillUnmount` was called too late on android and the `ShadowNode` removal mutation triggered the exiting layout animation before the `componentWillUnmount` was called. Since we used `componentWillUnmount` to disable exiting animations for `LayoutAnimationConfig`s on which the `skipExiting` prop was used, the disabled flag was set in CPP layout animations config object after the exiting animation started.

### How it was fixed?

The fix is not perfect, but I think that it is sufficient, at least for most cases, because people don't usually modify `skipExiting` prop at all.

Instead of waiting until the component unmounts, I mark exiting animations as disabled just after the component mounts (only if `skipExiting` is set to `true`) and then update it every time that the `skipExiting` value changes. In most cases it doesn't change at all, so in effect I set the `skipExiting` value in CPP only once.

## Example recordings

| Before | After |
|-|-|
| <video src="https://github.com/user-attachments/assets/432ee396-5b57-430a-bba9-e4ece460c733" /> | <video src="https://github.com/user-attachments/assets/5fb96cc2-044c-4d55-b239-410b06aa2165" /> |

## Test plan

Copy this code and paste to the `EmptyExample` to see how it works.

<details>

```tsx
import { StyleSheet } from 'react-native';
import Animated, {
  FadeOut,
  LayoutAnimationConfig,

  useEffect(() => {
    setTimeout(() => {
      setShow(false);
    }, 1000);
  }, []);

  return (
    show && (
      <LayoutAnimationConfig skipEntering skipExiting>
        <Animated.View exiting={FadeOut} style={styles.box} />
      </LayoutAnimationConfig>
    )
  );
}

  box: {
    width: 200,
    height: 200,
    backgroundColor: 'red',
  },
});
```
</details>

### DIFF
--- a/apps/common-app/src/App.tsx
+++ b/apps/common-app/src/App.tsx
@@ -1,169 +1,32 @@
-import { PortalProvider } from '@gorhom/portal';
-import AsyncStorage from '@react-native-async-storage/async-storage';
-import { createDrawerNavigator } from '@react-navigation/drawer';
-import type { NavigationState } from '@react-navigation/native';
-import {
-  getPathFromState,
-  NavigationContainer,
-} from '@react-navigation/native';
-import { useCallback, useEffect, useState } from 'react';
-import { ActivityIndicator, Linking, View } from 'react-native';
-import { GestureHandlerRootView } from 'react-native-gesture-handler';
-import { SafeAreaProvider } from 'react-native-safe-area-context';
-
-import { colors, flex, radius, text } from '@/theme';
-import { IS_MACOS, IS_WEB, noop } from '@/utils';
-
-import { CSSApp, ReanimatedApp } from './apps';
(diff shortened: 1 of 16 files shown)


MatiPl01:
After consulting with @bartlomiejbloniarz, it turned out that this approach also breaks this case when the exiting animation should be triggered:

```tsx
<LayoutAnimationConfig skipEntering skipExiting>
  {show && <Animated.View exiting={FadeOut} style={styles.box} />}
</LayoutAnimationConfig>
```

I will have to work on a different solution then.

MatiPl01:
Closed as RN 0.82 fixes the issue

PR #7681, opened 2025-06-23
fix: 

MatiPl01:
## Test plan

Paste the following code in the `apps/common-app/src/App.tsx` file to make sure that it is the only one component rendered on the screen and Reanimated won't be initialized because of other calls to `ReanimatedModule` methods.

<details>

```tsx

```
</details>


tjzel:
Can we simply call an init method that will instantiate `ReanimatedModule` in `index.ts` of `react-native-reanimated` instead? This is how it's done in Worklets.

MatiPl01:
> Can we simply call an init method that will instantiate `ReanimatedModule` in `index.ts` of `react-native-reanimated` instead? This is how it's done in Worklets.

Yup, it seems to work as well. I wonder if I should leave these changes in CPP code as well just for safety to make sure that the window is measured before trying to start the layout animation.

This might not be necessary, since it works just after adding early initialization in the `index.ts` file, so you'd likely prefer to remove these CPP changes.

MatiPl01:
I decided myself to remove cpp changes and use just a proper js-side initialization.

PR #7684, opened 2025-06-24
chore: Remove JS props updating via C++

MatiPl01:
## Summary

This PR removes the entire CPP logic of JS props handling and keeps this logic only on the JS side. In the new approach, instead of marking props as JS props for all components with the same native component name, we can easily decide, which props are JS props, separately for each component instance, which is more flexible and safer behavior than before.

The new approach allows these 2 possible ways to declare js props:

1. Via the static `getJSProps` method declared on the component class or function component declaration

```ts
static get jsProps() {
  return ['x', 'y'];
}
```

or

```ts
function FunctionComponent() { ... }

FunctionComponent.jsProps = ['x', 'y']
```

2. By passing prop names array using the `options` parameter on the `createAnimatedComponent` function:

```ts
const AnimatedG = Animated.createAnimatedComponent(G, {
  jsProps: ['x', 'y'],
});
```

If the component has `jsProps` static getter or `jsProps` public field already declared and `jsProps` were provided in `options` at the same time, `jsProps` from the `options` argument take precedence over ones declared on the component.

### DIFF
--- a/apps/common-app/src/apps/reanimated/examples/ThirdPartyComponentsExample.tsx
+++ b/apps/common-app/src/apps/reanimated/examples/ThirdPartyComponentsExample.tsx
@@ -106,11 +106,11 @@ function SvgPathDemo({ sv }: { sv: SharedValue<number> }) {
   );
 }
 
-const AnimatedPolygonNoJSProps = Animated.createAnimatedComponent(Polygon);
-
 const AnimatedPolygonJSProps = Animated.createAnimatedComponent(Polygon, {
   jsProps: ['points'],
 });
+
+const AnimatedPolygonNoJSProps = Animated.createAnimatedComponent(Polygon);
 
 function SvgPolygonsDemo({ sv }: { sv: SharedValue<number> }) {
   const animatedProps = useAnimatedProps(() => {
@@ -127,6 +127,11 @@ function SvgPolygonsDemo({ sv }: { sv: SharedValue<number> }) {
   return (
     <View style={styles.demo}>
       <Text style={styles.text}>With vs without JS props</Text>
+      <Text>
+        Both components use the same animated props, but only the shape of the
+        left one is animated, because the 'points' property is a JS prop and
+        only this component specified 'points' as the JS prop.
+      </Text>
       <Svg height="200" width="200">
         <AnimatedPolygonJSProps
           fill="lime"
@@ -180,6 +185,9 @@ const styles = StyleSheet.create({
   },
   demo: {
     borderWidth: 1,
+    padding: 20,
+    width: '100%',
+    alignItems: 'center',
   },
   text: {
     textAlign: 'center',
--- a/packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
+++ b/packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
@@ -18,7 +18,7 @@ import type {
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type AnimatableComponent<C extends ComponentType<any>> = C & {
-  getJSProps?(): string[];
+  jsProps?: string[];
 };
 
 /**
@@ -69,7 +69,7 @@ export function createAnimatedComponent(
 
     constructor(props: AnimatedComponentProps<InitialComponentProps>) {
       // User can override component-defined jsProps via options
-      const jsProps = options?.jsProps || Component.getJSProps?.();
+      const jsProps = options?.jsProps ?? Component.jsProps;
       const modifiedOptions = jsProps?.length
         ? { ...options, jsProps }
         : options;


packages/react-native-reanimated/src/createAnimatedComponent/createAnimatedComponent.tsx
```diff
@@ -16,6 +16,11 @@ import type {
   InitialComponentProps,
 } from './commonTypes';
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type AnimatableComponent<C extends ComponentType<any>> = C & {
```
MatiPl01, line 20: Not sure if I should look for a better name

apps/common-app/src/apps/reanimated/examples/JSPropsExample.tsx
```diff
@@ -13,11 +12,11 @@ import Animated, {
 import Svg, { Circle, G, Path } from 'react-native-svg';
 
 const AnimatedPath = Animated.createAnimatedComponent(Path);
-const AnimatedG = Animated.createAnimatedComponent(G);
+const AnimatedG = Animated.createAnimatedComponent(G, {
+  jsPropNames: ['x', 'y'],
```
tomekzaw:
```suggestion
  jsProps: ['x', 'y'],
```

MatiPl01:
### Update

I adjusted the old invalid `JS Props` example name and implementation. I also added the proper JS props example to the `ThirdPartyComponentsExample` screen.

| old JS Props | new JS Props (last example) |
|-|-|
| <video src="https://github.com/user-attachments/assets/287ff222-e8ac-4a48-b11d-f585a4c08c94" /> | <video src="https://github.com/user-attachments/assets/c89af6a7-f95c-4994-9998-98e94406a253" /> |

apps/common-app/src/apps/reanimated/examples/index.ts
```diff
@@ -173,10 +173,10 @@ export const EXAMPLES: Record<string, Example> = {
     title: 'Modify',
     screen: ModifyExample,
   },
-  JSPropsExample: {
-    icon: '🟨',
-    title: 'JS props',
-    screen: JSPropsExample,
+  CircularSliderExample: {
+    icon: '🔘',
+    title: 'Circular Slider Example',
```
tomekzaw:
```suggestion
    title: 'Circular slider',
```

apps/common-app/src/apps/reanimated/examples/ThirdPartyComponentsExample.tsx
```diff
@@ -107,25 +106,39 @@ function SvgPathDemo({ sv }: { sv: SharedValue<number> }) {
   );
 }
 
-const AnimatedG = Animated.createAnimatedComponent(G);
-
-registerJSProps('RNSVGGroup', ['x', 'y']);
+const AnimatedPolygonNoJSProps = Animated.createAnimatedComponent(Polygon);
```
tomekzaw: How about a newline here?
```suggestion
const AnimatedPolygonNoJSProps = Animated.createAnimatedComponent(Polygon);

```

apps/common-app/src/apps/reanimated/examples/ThirdPartyComponentsExample.tsx
```diff
@@ -107,25 +106,39 @@ function SvgPathDemo({ sv }: { sv: SharedValue<number> }) {
...
+function SvgPolygonsDemo({ sv }: { sv: SharedValue<number> }) {
   const animatedProps = useAnimatedProps(() => {
+    const topY = 75 - sv.value * 50;
+    const leftX = 25 + sv.value * 15;
+    const rightX = 75 - sv.value * 15;
+
     return {
-      x: sv.value * 200,
-      y: Math.sin(sv.value * Math.PI) * 200,
+      points: `50 ${topY}, ${leftX} 125, ${rightX} 125`, // this is JS prop
```
tomekzaw:
```suggestion
      points: `50 ${topY}, ${leftX} 125, ${rightX} 125`, // this is a JS prop
```

MatiPl01:
> Overall looks good to me. I've noticed that sometimes it's jsProp and sometimes it's jsProps, perhaps we should unify the identifiers to plural everywhere, i.e. jsPropsNames instead of jsPropNames.

I think it's quite common to add `s` at the end of the word in english without adding it in twice, so the `jsPropNames` is more correct in this case. Nevertheless, I will rename all usages of `jsPropNames` to `jsProps` as it is shorter and understandable.

apps/common-app/src/apps/reanimated/examples/ThirdPartyComponentsExample.tsx
MatiPl01: I would normally keep them next to each other without a blank separation line, but I don't care much about that, so I will change as you suggest.

apps/common-app/src/apps/reanimated/examples/JSPropsExample.tsx
MatiPl01: Should I also change `getJSPropNames` to `getJSProps` for consistency?

edit: I already added this change as well

PR #7685, opened 2025-06-24
Remove unused `enableLayoutAnimations`

piaskowyk:
## Summary

We no longer use that flag, so we can remove it. For backward compatibility purposes, I've left a deprecated mock of that function.


packages/react-native-reanimated/src/index.ts
```diff
@@ -82,7 +82,6 @@ export type { AnimatedScrollViewProps } from './component/ScrollView';
 export { configureReanimatedLogger } from './ConfigHelper';
 export {
   createWorkletRuntime,
-  enableLayoutAnimations,
```
tomekzaw: Perhaps we should keep the export here?

piaskowyk: yes, you're right - my bad 🫣
fixed - https://github.com/software-mansion/react-native-reanimated/pull/7685/commits/b124845de826fb1c98c490d20f4fbb91ffeff843

tomekzaw (approved):

packages/react-native-reanimated/src/core.ts
```diff
@@ -171,33 +171,18 @@ export function unregisterSensor(sensorId: number): void {
...
-  } else if (
-    !featuresConfig.setByUser &&
-    featuresConfig.enableLayoutAnimations !== flag
-  ) {
-    featuresConfig.enableLayoutAnimations = flag;
-    ReanimatedModule.enableLayoutAnimations(flag);
-  }
+  console.warn(
+    '[Reanimated] `enableLayoutAnimations` is deprecated and will be removed in the future.'
+  );
```
MatiPl01, line 185: Meh, please use `logger.warn` in the future for consistency 🙏🏻

PR #7686, opened 2025-06-24
Remove unnecessary `registerJSProps` for `G` component from react-native-svg in ThirdPartyComponentsExample

tomekzaw:
## Summary

Turns out `x` and `y` props of `G` component from `react-native-svg` can be animated as native props.

Thanks @jakex7 for finding that out!


MatiPl01 (approved):
`x` and `y` are now native props on the `RNSVGGroup` component?

tomekzaw:
@MatiPl01 Looks like they are no different than other props of other SVG components, I've removed `registerJSProps` call and it works without it.

tomekzaw:
Turns out I'm dumb and I forgot to remove `registerJSProps` call in JSPropsExample.tsx. Closing the PR.

PR #7687, opened 2025-06-24
chore: update circular dependency check

skusnierz:
## Summary
Together with @tjzel we agree that the circular dependency check is too restrictive, moreover having circular dependency in typescript is not forbidden, sometimes it is useful to simplify the code.
This PR restricts the circular dependency check only to `js` files in the lib folder for `reanimated` and `worklets` packages.

## Test plan
🚀


packages/react-native-reanimated/package.json
```diff
@@ -31,7 +31,7 @@
     "type:check:app": "yarn workspace common-app type:check",
     "type:check:tests:common": "./scripts/test-ts.sh __typetests__/common",
     "build": "yarn workspace react-native-worklets build && bob build",
-    "circular-dependency-check": "yarn madge --extensions js,jsx,ts,tsx --circular src lib",
+    "circular-dependency-check": "yarn madge --extensions js,jsx,ts,tsx --circular lib",
```
tjzel:
```suggestion
    "circular-dependency-check": "yarn madge --extensions js,jsx --circular lib",
```

packages/react-native-worklets/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "build": "yarn workspace babel-plugin-worklets build && bob build && yarn build:valueUnpacker",
     "build:valueUnpacker": "node ./scripts/export-value-unpacker.js",
-    "circular-dependency-check": "yarn madge --extensions js,jsx,ts,tsx --circular src lib",
+    "circular-dependency-check": "yarn madge --extensions js,jsx,ts,tsx --circular lib",
```
tjzel:
```suggestion
    "circular-dependency-check": "yarn madge --extensions js,jsx --circular lib",
```

tjzel (approved):

PR #7690, opened 2025-06-24
feat(Worklets): rename experimental bundling to Bundle Mode

tjzel:
## Summary

`Experimental Bundling` doesn't make a good outside name.

## Test plan

I tested bundle mode iOS and Android and their worked 👌

### DIFF
--- a/packages/docs-reanimated/docs/utilities/getRelativeCoords.mdx
+++ b/packages/docs-reanimated/docs/utilities/getRelativeCoords.mdx
@@ -26,7 +26,7 @@ const Comp = () => {
   });
 
   return (
-    <View ref={aref}>
+    <View ref={animatedRef}>
       <PanGestureHandler onGestureEvent={gestureHandler}>
         <Animated.View style={[styles.box]} />
       </PanGestureHandler>
--- a/packages/docs-reanimated/versioned_docs/version-3.x/utilities/getRelativeCoords.mdx
+++ b/packages/docs-reanimated/versioned_docs/version-3.x/utilities/getRelativeCoords.mdx
@@ -26,7 +26,7 @@ const Comp = () => {
   });
 
   return (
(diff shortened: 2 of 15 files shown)


apps/fabric-example/metro.config.js
```diff
@@ -17,45 +20,23 @@ const root = path.resolve(__dirname, '../..');
...
-  //         const id = Number(base);
-  //         idFileMap.set(moduleName, id);
-  //         return id;
-  //       }
-  //       idFileMap.set(moduleName, nextId++);
-  //       return idFileMap.get(moduleName);
-  //     };
-  //   },
-  // },
-  // --------------------------------------------------------
```
tjzel, line 51: Moved it to `react-native-worklets/bundleMode`

packages/react-native-worklets/bundleMode/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+module.exports = {
+  bundleModeMetroConfig: {
+    serializer: {
+      getModulesRunBeforeMainModule() {
+        return [
+          require.resolve('react-native-worklets/src/workletRuntimeEntry.ts'),
+          require.resolve(
+            'react-native-worklets/lib/module/workletRuntimeEntry.js'
+          ),
```
piaskowyk, line 11: Why do we need them twice?

piaskowyk (approved):

packages/react-native-worklets/bundleMode/index.js
tjzel: The way it works is that metro will actually run these files if they are included in the bundle. Therefore, we have to add both here since we don't know if the user uses source from `src/` or generated js from `lib/module/`.

It's designed that way so if you'd have different entry points per platform only one of them would be run.

PR #7691, opened 2025-06-24
feat: Eslint rules to ensure logger usage instead of console

MatiPl01:


packages/eslint-plugin-reanimated/src/index.ts
```diff
@@ -1,13 +1,19 @@
...
+import useLogger from './useLogger';
 import useReanimatedError from './useReanimatedError';
 import useWorkletsError from './useWorkletsError';
+import wrongLoggerImport from './wrongLoggerImport';
 
 export const rules = {
   'animated-style-non-animated-component':
     noAnimatedStyleToNonAnimatedComponent,
+  'use-logger': useLogger,
+  'wrong-logger-import': wrongLoggerImport,
```
MatiPl01: I am not sure if we want to export the logger from `react-native-worklets` or this export can be now removed and the corresponding rule can be removed as well.

tjzel: We don't want to export the loggers 👍

packages/eslint-plugin-reanimated/src/useLogger.ts
```diff
@@ -0,0 +1,49 @@
...
+      CallExpression(node: TSESTree.CallExpression) {
+        const { callee } = node;
+
+        if (
+          callee.type === AST_NODE_TYPES.MemberExpression &&
+          !callee.computed &&
+          callee.object.type === AST_NODE_TYPES.Identifier &&
+          callee.object.name === 'console' &&
+          callee.property.type === AST_NODE_TYPES.Identifier &&
+          (callee.property.name === 'warn' || callee.property.name === 'error')
```
tjzel, line 17: What about `log`?

tjzel (approved):
Let's remove the `import` rule and remove the logger exports altogether. LGTM but please wait until @skusnierz bumps ESLint in the repo first 🙏

packages/eslint-plugin-reanimated/src/useLogger.ts
MatiPl01: Our logger doesn't contain the `log` method as we shouldn't use `console.log` in the package for purposes different than just debugging before shipping the code. That's why I don't check for it and I would prefer just to enable the `no-console` eslint rule showing warnings near `console.log` statements that we may leave in code.

tjzel: 👍

PR #7697, opened 2025-06-25
Add feature flags

piaskowyk:
## Summary

This PR introduces the concept of feature flags, providing a tool to control the failure impact of experimental changes by allowing users control over which experimental features are enabled. I propose two types of feature flags: static and dynamic. The main idea is to have one source of truth accessible from both the JS and Native sides.

### Static Feature Flags

Static flags are intended to be resolved during code compilation and cannot be changed during application runtime. Changing these flags requires recompilation of the app and, for iOS, running `pod install` again. They are not exposed to the user via JS; we only access them internally in our reanimated code (from both JS and C++). In C++, we can use these flags as constant expressions to switch code branches during compilation without runtime overhead.

C++ usage:
```cpp
if constexpr (StaticFeatureFlags::getFlag("EXAMPLE_STATIC_FLAG")) {
  // do something
} else {
  // do something else
}
```
it works the same as:
```cpp
#ifdef EXAMPLE_STATIC_FLAG
  // do something
#elseif
  // do something else
#endif
```
but constexpr checks types compatibility in both branches.

In Reanimated, the source of truth is `src/featureFlags/staticFlags.json`. During compilation, we parse this file and send it as a compilation flag to C++, where we use this flag information in constant expressions. Users can control these flags by overriding them in their application's `package.json` file:

```json
"reanimated": {
  "staticFeatureFlags": {
    "EXAMPLE_STATIC_FLAG": false
  }
}
```

### Dynamic Feature Flags

Dynamic flags can be modified by users at runtime. The source of truth is `src/featureFlags/dynamicFlags.ts`. Changes to these flags are synchronized with the flag registry on the native side. The use of dynamic flags in our internal code requires considering that their values can change at any moment.

Example:
```javascript
import { setDynamicFeatureFlag } from 'react-native-reanimated';
<Button title="Press me" onPress={() => {
  setDynamicFeatureFlag('EXAMPLE_DYNAMIC_FLAG', true);
}} />
```

### Summary

|                      | Static Flags | Dynamic Flags |
|----------------------|--------------|---------------|
| Can change in runtime | ❌           | ✅            |
| Known value during compilation | ✅  | ❌           |
| Recompilation required on change | ✅  | ❌           |
| Changeable via JS function | ❌     | ✅            |
| Changeable via config file | ✅     | ❌           |
| Accessible from JS | ✅     | ✅            |
| Accessible from C++ | ✅     | ✅            |
| Available in public JS API | ❌     | ✅            |


packages/react-native-reanimated/src/featureFlags/dynamicFlags.ts
```diff
@@ -0,0 +1,47 @@
...
+// Public API function to update a feature flag
+export function setDynamicFeatureFlag(
+  name: DynamicFlagName,
+  value: boolean
+): void {
+  DynamicFlags.setFlag(name, value);
+}
+
+// ts-prune-ignore-next
+export default DynamicFlags;
```
tjzel:
```suggestion
```

packages/react-native-reanimated/src/featureFlags/dynamicFlags.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+  },
+  setFlag(name, value) {
+    if (name in DynamicFlags) {
+      DynamicFlags[name] = value;
+      ReanimatedModule.setDynamicFeatureFlag(name, value);
+    } else {
+      console.warn(
+        `[Reanimated] The feature flag: '${name}' no longer exists, you can safely remove invocation of \`setDynamicFeatureFlag('${name}')\` from your code.`
+      );
```
tjzel:
```suggestion
      logger.warn(
        `The feature flag: '${name}' no longer exists, you can safely remove invocation of \`setDynamicFeatureFlag('${name}')\` from your code.`
      );
```

packages/react-native-reanimated/src/featureFlags/dynamicFlags.ts
```diff
@@ -0,0 +1,47 @@
...
+
+type DynamicFlagsType = {
+  TEST_DYNAMIC_FLAG: boolean;
+  init(): void;
+  setFlag(name: DynamicFlagName, value: boolean): void;
+};
+type DynamicFlagName = keyof Omit<Omit<DynamicFlagsType, 'setFlag'>, 'init'>;
+
+// ts-prune-ignore-next
+export const DynamicFlags: DynamicFlagsType = {
```
tjzel, line 13: Can we make an enum instead?

packages/react-native-reanimated/scripts/reanimated_utils.rb
```diff
@@ -70,3 +70,30 @@ def assert_new_architecture_enabled(new_arch_enabled)
     raise "[Reanimated] Reanimated requires the New Architecture to be enabled. If you have `RCT_NEW_ARCH_ENABLED=0` set in your environment you should remove it."
   end
 end
+
+def get_feature_flags()
+  feature_flags = {}
+
+  static_feature_flags_path = File.path('./src/featureFlags/staticFlags.ts')
```
tjzel: How about we store the flags schema in JSON and require it both here and in TS?

packages/react-native-reanimated/android/build.gradle
```diff
@@ -63,6 +63,31 @@ def toPlatformFileString(String path) {
   return path
 }
 
+def getReanimatedFeatureFlags() {
```
tjzel:
```suggestion
def getReanimatedStaticFeatureFlags() {
```

packages/react-native-reanimated/scripts/reanimated_utils.rb
```diff
@@ -70,3 +70,30 @@ def assert_new_architecture_enabled(new_arch_enabled)
     raise "[Reanimated] Reanimated requires the New Architecture to be enabled. If you have `RCT_NEW_ARCH_ENABLED=0` set in your environment you should remove it."
   end
 end
+
+def get_feature_flags()
```
tjzel:
```suggestion
def get_static_feature_flags()
```

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.h
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+#endif
+};
+
+class DynamicFeatureFlags {
+  static std::unordered_map<std::string, bool> flags_;
+
+ public:
+  static bool getFlag(const char *key);
```
tjzel: Why `const char *`?

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.cpp
```diff
@@ -0,0 +1,15 @@
+#include <reanimated/Tools/FeatureFlags.h>
+
+namespace reanimated {
+
+std::unordered_map<std::string, bool> DynamicFeatureFlags::flags_;
+
+bool DynamicFeatureFlags::getFlag(const char *key) {
+  return flags_.find(key) != flags_.end() && flags_[key];
```
tomekzaw:
```suggestion
  return flags_.contains(key) && flags_[key];
```

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.cpp
```diff
@@ -0,0 +1,15 @@
...
+}
+
+void DynamicFeatureFlags::setFlag(const std::string &key, bool value) {
```
tomekzaw:
```suggestion
void DynamicFeatureFlags::setFlag(const std::string &key, const bool value) {
```

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.cpp
```diff
@@ -0,0 +1,15 @@
+#include <reanimated/Tools/FeatureFlags.h>
+
+namespace reanimated {
+
+std::unordered_map<std::string, bool> DynamicFeatureFlags::flags_;
+
+bool DynamicFeatureFlags::getFlag(const char *key) {
```
tomekzaw: Why not `const std::string &` or `const std::string_view &`?

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.cpp
```diff
@@ -0,0 +1,15 @@
...
+  return flags_.find(key) != flags_.end() && flags_[key];
+}
+
+void DynamicFeatureFlags::setFlag(const std::string &key, bool value) {
+  flags_[key] = value;
```
tomekzaw: How about using `name` instead of `key`?
```
suggestion  flags_[name] = value;
```

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.h
```diff
@@ -0,0 +1,42 @@
...
+#include <unordered_map>
+
+namespace reanimated {
+
+class StaticFeatureFlags {
+ public:
+#ifdef REANIMATED_FEATURE_FLAGS
+
+#define STRINGIFY(x) #x
+#define TOSTRING(x) STRINGIFY(x)
```
tomekzaw: What's the difference between `TOSTRING` and `STRINGIFY` then?

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.h
```diff
@@ -0,0 +1,42 @@
...
+#define REANIMATED_FEATURE_FLAGS_STRING TOSTRING(REANIMATED_FEATURE_FLAGS)
+
+  static constexpr bool getFlag(const char *key) {
```
tomekzaw:
```suggestion
  static constexpr bool getFlag(const std::string_view &name) {
```

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.h
```diff
@@ -0,0 +1,42 @@
...
+    const std::string keyStr = key;
+    std::string featureFlags = REANIMATED_FEATURE_FLAGS_STRING;
+    if (featureFlags.find(key) == std::string::npos) {
```
tomekzaw: I don't think this will work properly if we have two flags where one is contained within another, e.g. `cde` and `abcdefg`. We should look for `[cde:` instead of only `cde`.

packages/react-native-reanimated/src/ReanimatedModule/js-reanimated/JSReanimated.ts
```diff
@@ -250,6 +250,10 @@ class JSReanimated implements IReanimatedModule {
     throw new ReanimatedError('getViewProp is not available in JSReanimated.');
   }
 
+  setDynamicFeatureFlag(_: string, __: boolean): void {
```
tomekzaw:
```suggestion
  setDynamicFeatureFlag(_name: string, _value: boolean): void {
```

packages/react-native-reanimated/src/featureFlags/staticFlags.ts
```diff
@@ -0,0 +1,5 @@
+'use strict';
+// ts-prune-ignore-next
+export const StaticFeatureFlags = {
+  TEST_STATIC_FLAG: true,
```
tomekzaw: Let's avoid using "test" for things that are not tests. Instead, let's use "example" or "demo".
```suggestion
  EXAMPLE_STATIC_FLAG: true,
```

packages/react-native-reanimated/src/featureFlags/dynamicFlags.ts
piaskowyk: We discussed it offline that we don't need them 🫡

packages/react-native-reanimated/scripts/reanimated_utils.rb
piaskowyk: Done 🫡 https://github.com/software-mansion/react-native-reanimated/pull/7697/commits/1eb75d882d1695c229b5c889caccbeee2d83367e

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.h
piaskowyk: I added a comment with an explanation to that part of the code 🫡

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.cpp
```diff
@@ -0,0 +1,15 @@
+#include <reanimated/Tools/FeatureFlags.h>
+
+namespace reanimated {
+
+std::unordered_map<std::string, bool> DynamicFeatureFlags::flags_;
```
tjzel, line 5:
```suggestion
std::unordered_map<const std::string, bool> DynamicFeatureFlags::flags_;
```

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.h
```diff
@@ -0,0 +1,44 @@
...
+#define XTOSTRING(x) #x
+// Evaluate the flag value; without this step, it would stringify the flag name
+// itself instead of the flag value
+#define TOSTRING(x) XTOSTRING(x)
+
+  static constexpr bool getFlag(const std::string_view &name) {
+    std::string nameStr = name.data();
+    std::string featureFlags = TOSTRING(REANIMATED_FEATURE_FLAGS);
+    if (featureFlags.find("[" + nameStr + ":") == std::string::npos) {
+      // this will cause compilation error not runtime error
```
tjzel: I think we can safely remove that comment since it's inside a `constexpr`.

```suggestion
```

tjzel (approved):
LGTM, but can you provide a more comprehensive example of static flags usage?

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.cpp
piaskowyk: Unfortunately, I can't do that :/ A string key can't be `const` in a map

tjzel: 🤯

msluszniak: `std::unoreder_map` under the hood uses `std::allocator` which expects cv-unqualified object type. There is even a clang tidy check for this type of mistakes: https://clang.llvm.org/extra/clang-tidy/checks/portability/std-allocator-const.html.

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+bool DynamicFeatureFlags::getFlag(const std::string &name) {
+  return flags_.contains(name) && flags_[name];
+}
+
+void DynamicFeatureFlags::setFlag(const std::string &name, const bool value) {
```
msluszniak:
```suggestion
void DynamicFeatureFlags::setFlag(const std::string &name, bool value) {
```
Top const qualifiers are dropped automatically by compiler, remove const and use `bool value`. Check this: https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.h
```diff
@@ -0,0 +1,43 @@
...
+#endif
+};
+
+class DynamicFeatureFlags {
+  static std::unordered_map<std::string, bool> flags_;
+
+ public:
+  static bool getFlag(const std::string &name);
+  static void setFlag(const std::string &name, bool value);
+};
```
msluszniak: Just a suggestion, so feel free to ignore this, but this way seems to have two benefits:
* You do not increase cognitive effort of reader to control that default access specifier in c++ classes is id `private` and you explicitly mark that this private
* Usually, code is read from top to bottom, so having private members below public sounds reasonable
```suggestion
class DynamicFeatureFlags {
 public:
  static bool getFlag(const std::string &name);
  static void setFlag(const std::string &name, bool value);

 private:
   static std::unordered_map<std::string, bool> flags_;
};
```

packages/react-native-reanimated/Common/cpp/reanimated/Tools/FeatureFlags.h
```diff
@@ -0,0 +1,43 @@
...
+    std::string featureFlags = TOSTRING(REANIMATED_FEATURE_FLAGS);
+    if (featureFlags.find("[" + nameStr + ":") == std::string::npos) {
+      throw std::logic_error("Unable to recognize flag: " + nameStr);
+    }
+    return featureFlags.find("[" + nameStr + ":true]") != std::string::npos;
+  }
+
+#else
+
+  static constexpr bool getFlag(const char *) {
```
msluszniak: If you use c++ at least c++20, force this to be compile-time evaluated. If you do not support c++20, ignore this comment.
```suggestion
  static consteval bool getFlag(const char *) {
```

PR #7699, opened 2025-06-25
feat(Worklets): improve errors in Worklets Babel Plugin

tjzel:
## Summary

QoL change that makes it so we don't replace the stack traces and get an actually informative error.

Before:
Screenshot 2025-06-25 at 18 09 03

After:
Screenshot 2025-06-25 at 18 07 08

## Test plan

CI


MatiPl01 (approved):
LGTM

tomekzaw (approved):

PR #7702, opened 2025-06-26
chore: use same versions of dependencies in monorepo where applicable

tjzel:
## Summary

WIP


yarn.lock
```diff
@@ -75,88 +79,48 @@ __metadata:
...
+  version: 7.27.5
+  resolution: "@babel/compat-data@npm:7.27.5"
+  checksum: 10/04c343b8a25955bbbe1569564c63ac481a74710eb2e7989b97bd10baf2f0f3b1aa1b6c6122749806e92d70cfc22c10c757ff62336eb10a28ea98ab2b82bc0c2c
   languageName: node
   linkType: hard
 
-"@babel/core@npm:^7.25.8":
-  version: 7.26.0
-  resolution: "@babel/core@npm:7.26.0"
+"@babel/core@npm:^7.11.6, @babel/core@npm:^7.12.3, @babel/core@npm:^7.12.9, @babel/core@npm:^7.13.16, @babel/core@npm:^7.20.0, @babel/core@npm:^7.23.9, @babel/core@npm:^7.25.2, @babel/core@npm:^7.25.8, @babel/core@npm:^7.27.4":
```
MatiPl01, line 100: Should it look like this now?

.lintstagedrc-common.js
```diff
@@ -1,7 +1,7 @@
 /** @type {import('lint-staged').Config} */
 module.exports = {
   '*.(js|jsx|ts|tsx)': [
-    'yarn eslint --flag unstable_config_lookup_from_file',
+    'yarn eslint --flag v10_config_lookup_from_file',
```
tjzel, line 4: `unstable_config_lookup_from_file` was deprecated and replaced by `v10_config_lookup_from_file`.

apps/common-app/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "build": "yarn patch-package",
```
tjzel, line 6: We no longer use `patch-package`.

apps/common-app/src/apps/css/navigation/Navigator.tsx
```diff
@@ -85,9 +85,7 @@ function createRoutesScreen(
     const navigation = useNavigation();
     const localNavigationRef = useLocalNavigationRef();
 
-    if (!localNavigationRef.current) {
-      localNavigationRef.current = navigation;
-    }
+    localNavigationRef.current ??= navigation;
```
tjzel, line 88: Required by bumped eslint. Cool operator though!

apps/common-app/src/apps/css/utils/code.ts
```diff
@@ -146,9 +146,7 @@ export const getCodeWithOverrides = <C extends AnyRecord, O extends AnyRecord>(
   for (const item of overrides) {
     for (const key in item) {
       if (!excludeSet.has(key)) {
-        if (!propertyOverrides[key]) {
-          propertyOverrides[key] = [];
-        }
+        propertyOverrides[key] ??= [];
```
tjzel, line 149: Required by bumped eslint.

packages/react-native-reanimated/__tests__/__snapshots__/Animation.test.tsx.snap
```diff
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
```
tjzel, line 1: Required by bumped eslint.

packages/react-native-worklets/package.json
```diff
@@ -50,37 +50,37 @@
   },
   "homepage": "https://docs.swmansion.com/react-native-worklets",
   "peerDependencies": {
-    "@babel/core": "^7.0.0-0",
+    "@babel/core": "*",
```
tjzel, line 53: Caret doesn't scale well.

tjzel:
Failing Next CI is a flake - it works for me locally, I guess some node_modules caches on github are stale.

docs/docs-worklets/package.json
```diff
@@ -32,8 +33,9 @@
     "@mdx-js/react": "3.0.0",
     "@mui/material": "7.1.0",
     "@react-native/assets-registry": "0.79.1",
-    "@shopify/flash-list": "1.6.3",
+    "@shopify/flash-list": "2.1.0",
```
MatiPl01, line 36: Have you confirmed that the new version works in docs? v2 has some breaking changes compared to the v1

apps/next-example/package.json
```diff
@@ -15,23 +15,26 @@
...
+    "next": "15.5.4",
     "react": "19.1.1",
     "react-dom": "19.1.1",
+    "react-native": "0.82.0",
     "react-native-reanimated": "workspace:*",
-    "react-native-web": "0.20.0",
-    "react-native-worklets": "workspace:*"
+    "react-native-web": "0.21.1",
+    "react-native-worklets": "workspace:*",
+    "webpack": "5.102.1"
```
MatiPl01, line 26: Why the explicit webpack dependency wasn't needed before but is listed here now? Was it installed implicitly and you added it just to have more control over which version is installed?

apps/next-example/package.json
```diff
@@ -15,23 +15,26 @@
     "format": "prettier --write --list-different ."
   },
   "dependencies": {
-    "expo": "~48.0.10",
-    "next": "latest",
+    "expo": "54.0.13",
```
MatiPl01, line 18: Should this PR change versions of dependencies like `expo`, etc.? You said that it updates versions of these:

> Bumped all common dependencies (like eslint, prettier, typescript) to the same exact versions.

But I wonder if `expo`, `@shopify/flash-list`, etc. needs to be updated here as well.

apps/fabric-example/package.json
```diff
@@ -12,10 +12,13 @@
...
   "dependencies": {
     "common-app": "workspace:*",
+    "react": "19.1.1",
     "react-native": "0.82.0"
   },
   "devDependencies": {
-    "prettier": "3.6.2"
+    "@types/react": "19.2.2",
+    "prettier": "3.6.2",
+    "typescript": "5.8.3"
```
MatiPl01, line 21: Do you have to add all of these deps to the fabric-example `package.json`? I added a script some time ago that pulls all dependencies from the common-app's `package.json` file so you shouldn't have to specify them separately in each app's `package.json`.

There might be some exceptions to this rule, if we run scripts like `prettier`, etc. from the separate app's directories (have scripts defined in their `package.json` files), but I believe that some of the dependencies you added can be safely removed.

apps/macos-example/package.json
```diff
@@ -13,18 +13,20 @@
...
+    "@babel/preset-env": "7.28.3",
+    "@babel/runtime": "7.28.4",
     "@react-native/babel-preset": "0.75.4",
     "@react-native/eslint-config": "0.75.4",
     "@react-native/metro-config": "0.75.4",
     "@react-native/typescript-config": "0.75.4",
-    "prettier": "3.6.2"
+    "eslint": "9.37.0",
+    "prettier": "3.6.2",
+    "typescript": "5.8.3"
```
MatiPl01, line 32: Same here, why can't we pull these deps from the common-app's `package.json`?

apps/next-example/package.json
```diff
@@ -15,23 +15,26 @@
...
+    "@babel/core": "7.28.4",
+    "@expo/next-adapter": "6.0.0",
+    "@next/bundle-analyzer": "15.5.4",
+    "cypress": "15.4.0",
+    "eslint": "9.37.0",
+    "next-compose-plugins": "2.2.1",
     "prettier": "3.6.2",
-    "start-server-and-test": "^2.0.1"
+    "start-server-and-test": "2.1.2",
+    "typescript": "5.8.3"
```
MatiPl01, line 37: Same here (`react-native`, `typescript`, etc.), why can't we pull these deps from the common-app's `package.json`?

apps/web-example/package.json
```diff
@@ -13,16 +13,19 @@
...
   "devDependencies": {
-    "@expo/metro-runtime": "~6.1.1",
-    "@stylexjs/babel-plugin": "^0.10.0",
+    "@expo/metro-runtime": "6.1.2",
+    "@stylexjs/babel-plugin": "0.16.1",
+    "@types/react": "19.2.2",
     "prettier": "3.6.2",
-    "serve": "^14.2.3"
+    "serve": "14.2.5",
+    "typescript": "5.8.3"
```
MatiPl01, line 28: Same here (`react`, `typescript`, etc.), why can't we pull these deps from the common-app's `package.json`?

docs/docs-reanimated/package.json
```diff
@@ -32,8 +33,9 @@
     "@mdx-js/react": "3.0.0",
     "@mui/material": "7.1.0",
     "@react-native/assets-registry": "0.79.1",
-    "@shopify/flash-list": "1.6.3",
+    "@shopify/flash-list": "2.1.0",
```
MatiPl01, line 36: Have you confirmed that the new version works in docs? v2 has some breaking changes compared to the v1

MatiPl01 (changes requested):
I think you can remove some dependencies from particular example app's `package.json` files as they will be automatically pulled from the common-app that already has them specified. The rest looks fine, so just address my comments and I will approve.

apps/next-example/package.json
tjzel: Peer dependency warning required installing webpack.

apps/fabric-example/package.json
tjzel: @types/react is a solution to a peer-dependency warning. explicit prettier and ts are to make sure that local node_modules has respective binaries.

apps/macos-example/package.json
tjzel: Explicit prettier and ts are to make sure that local node_modules has respective binaries. If only the root had eslint you wouldn't be able to call `yarn eslint` in the workspace since there's no hierarchical lookup for binaries.

apps/next-example/package.json
tjzel: Yes, they work.

docs/docs-reanimated/package.json
tjzel: Yes it works.

apps/next-example/package.json
tjzel: Explicit prettier and ts are to make sure that local node_modules has respective binaries. If only the root had eslint you wouldn't be able to call yarn eslint in the workspace since there's no hierarchical lookup for binaries.

apps/web-example/package.json
tjzel: Explicit prettier and ts are to make sure that local node_modules has respective binaries. If only the root had eslint you wouldn't be able to call yarn eslint in the workspace since there's no hierarchical lookup for binaries.

apps/macos-example/package.json
MatiPl01: Ok, so it allows us to run linter e.g. from the `apps/fabric-example` directory, which didn't work before? I always ran linting from the root so I didn't notice that. If you feel that the ability to run linting from each workspace is necessary, then I am ok with leaving the explicit eslint dependency.

MatiPl01: What about the `typescript` dependency? Is the issue the same as with `eslint`?

apps/next-example/package.json
MatiPl01: Ok, if they work, then I'm ok with leaving the update in this PR. It seemed to be a bit out of the scope for me, but it is not a big deal and you can leave this version bump here.

PR #7710, opened 2025-06-26
fix: Not working window size dependent layout animations

MatiPl01:
## Summary

Similar fix to #7681 that fixes the same issue. I just moved reanimated initialization to the `index.ts` file to ensure that it is initialized before anything from reanimated is used.


tjzel (approved):
Please make sure to test it properly beforehand 🙏

MatiPl01:
> Please make sure to test it properly beforehand 🙏

Already tested :)

PR #7721, opened 2025-06-30
chore: Migrate from ts-prune to knip

MatiPl01:
## Summary

This PR replaces the old `ts-prune` unused code checking library by the more modern equivalent - `knip`. `ts-prune` is not maintained for 4 years and I decided to use `knip`, because it is more up-to-date and offers better customization. `knip` also marks code as unused more properly and spots more unused code parts that can be removed, compared to the `ts-prune`.


eslint.config.mjs
```diff
@@ -126,7 +126,6 @@ const config = tsEslint.config(
       parserOptions: {
         ecmaFeatures: { jsx: true },
         projectService: true,
-        tsconfigRootDir: import.meta.dirname,
```
MatiPl01, line 129: was resolving to the root `tsconfig.json` in our subpackages that extended this config (in `packages/react-native-reanimated` and `packages/react-native-worklets`)

packages/react-native-reanimated/eslint.config.mjs
```diff
@@ -20,6 +20,11 @@ const config = tsEslint.config(
     },
   },
   {
+    languageOptions: {
+      parserOptions: {
+        tsconfigRootDir: import.meta.dirname,
```
MatiPl01, line 25: Needed to add this locally in the package config to resolve the local `tsconfig.json` file.

packages/react-native-reanimated/tsdoc.json
```diff
@@ -0,0 +1,10 @@
+{
```
MatiPl01, line 1: `tsdoc.json` files must be at the same level as the eslint resolved `tsconfig.json` file, thus every package needs its own `tsdoc.json` file.

packages/react-native-reanimated/src/__mocks__/ReanimatedEventEmitter.js
```diff
@@ -3,7 +3,6 @@ const NOOP = () => {
   // noop
 };
 
-// ts-prune-ignore-next Is this even used?
```
tjzel, line 6: Why do these no longer need to be ignored?

packages/react-native-reanimated/src/__mocks__/ReanimatedModule.native.js
```diff
@@ -3,7 +3,6 @@ const NOOP = () => {
   // noop
 };
 
-// ts-prune-ignore-next Is this even used?
```
tjzel, line 6: Why do these no longer need to be ignored?

packages/react-native-reanimated/src/animation/springUtils.ts
```diff
@@ -126,7 +126,6 @@ export function checkIfConfigIsValid(config: DefaultSpringConfig): boolean {
   return errorMessage === '';
 }
 
-// ts-prune-ignore-next This function is exported to be tested
```
tjzel, line 129: Why do these no longer need to be ignored?

packages/react-native-reanimated/src/animation/transformationMatrix/matrixUtils.tsx
```diff
@@ -38,7 +38,6 @@ export function isAffineMatrixFlat(x: unknown): x is AffineMatrixFlat {
   );
 }
 
-// ts-prune-ignore-next This function is exported to be tested
```
tjzel, line 41: Why do these no longer need to be ignored?

packages/react-native-reanimated/src/core.ts
```diff
@@ -176,7 +176,6 @@ export function unregisterSensor(sensorId: number): void {
  * @deprecated This function no longer has any effect in Reanimated and will be
  *   removed in the future.
  */
-// ts-prune-ignore-next This function is needed for backward compatibility
```
tjzel, line 187: Why do these no longer need to be ignored?

packages/react-native-reanimated/src/platform-specific/checkCppVersion.ts
```diff
@@ -20,8 +20,6 @@ export function checkCppVersion() {
   }
 }
 
-// This is used only in test files, therefore it is reported by ts-prune (which is desired)
-// ts-prune-ignore-next
```
tjzel, line 24: Why do these no longer need to be ignored?

packages/react-native-reanimated/src/specs/ReanimatedViewProvider.native.ts
```diff
@@ -1,5 +1,4 @@
 'use strict';
 import ReanimatedView from './ReanimatedNativeComponent';
 
-// ts-prune-ignore-next-line
```
tjzel, line 4: Why do these no longer need to be ignored?

packages/react-native-reanimated/tsdoc.json
tjzel: Can you double check if there's no way to hoist them to the root of monorepo?

packages/react-native-worklets/src/specs/NativeWorkletsModule.ts
```diff
@@ -2,7 +2,7 @@
 import type { TurboModule } from 'react-native';
 import { TurboModuleRegistry } from 'react-native';
 
-// ts-prune-ignore-next False positive. Required for TypeScript compilation.
+/** @knipIgnore False positive. Required for TypeScript compilation. */
```
tjzel: Is it a false positive even with knip? 😿

packages/react-native-reanimated/src/__mocks__/ReanimatedEventEmitter.js
MatiPl01: It looks that `knip` automatically ignores this directory when `jest` is installed in the project. When I changed the name from `__mocks__` to something else, like random letters or removed `jest` from the project, an error was reported for this file.

This is a short fragment of the output for when I removed `jest` - all jest-related files are treated as unused in this case, so `knip` is kinda smart at checking for the file usefulness.

Screenshot 2025-07-02 at 13 57 38

packages/react-native-reanimated/src/__mocks__/ReanimatedModule.native.js
MatiPl01: same as above

packages/react-native-reanimated/src/animation/springUtils.ts
MatiPl01: It is imported in `packages/react-native-reanimated/__tests__/spring.test.tsx`, so `ts-prune` reported an invalid error.

packages/react-native-reanimated/src/animation/transformationMatrix/matrixUtils.tsx
MatiPl01: used in: `packages/react-native-reanimated/__tests__/matrixUtils.test.tsx`

packages/react-native-reanimated/src/core.ts
MatiPl01: it is exported from `index.ts` file and `index.ts` file is treated as the library entry point, so everything exported from this file is treated as used, thanks to the following setting:

```
"entry": ["src/index.ts", "mock.js"],
```

I needed to add mock as well as it can be used in the user's project to mock things and is not used by us.

packages/react-native-reanimated/src/platform-specific/checkCppVersion.ts
MatiPl01: Because, as the comment, which I removed, already said, it is used in tests. `ts-prune` didn't work with test files well and assumed that this function wasn't used, whilst it is actually imported in the `packages/react-native-reanimated/__tests__/checkVersion.test.ts`

packages/react-native-reanimated/src/specs/ReanimatedViewProvider.native.ts
MatiPl01: Because it is imported in the `packages/react-native-reanimated/src/specs/index.ts` file where I added this TSDoc comment to ignore the unused export:

```ts
export {
  ReanimatedTurboModule,
  /** @knipIgnore */
  ReanimatedView,
};
```

packages/react-native-reanimated/tsdoc.json
MatiPl01: tsdoc eslint plugin [docs](https://tsdoc.org/pages/packages/tsdoc-config/) say this

> The tsdoc.json file is optional. When used, it is expected to be found in the same folder as the tsconfig.json file for a project. The loader looks for it by walking upwards in the directory tree until it finds a folder containing tsconfig.json or package.json, and then it attempts to load tsdoc.json from that location.

But I read below that I can use this `tsdoc.json` file in the package local dir to import shared config from another place:

```
{
  "$schema": "https://developer.microsoft.com/json-schemas/tsdoc/v0/tsdoc.schema.json",
  "extends": ["../../tsdoc.json"]
}
```

packages/react-native-worklets/src/specs/NativeWorkletsModule.ts
MatiPl01: It is not imported anywhere, so that's why `knip` consider this export useless. I know that it is required to build the worklets package and without it the `tsc` command fails.

If you don't want this ignore, I can import the `Spec` type in the `packages/react-native-worklets/src/specs/index.ts` file and use it to explicitly type the `WorkletsTurboModule` like this:

```ts
import type { Spec } from './NativeWorkletsModule';
import RNWorkletsTurboModule from './NativeWorkletsModule';

export const WorkletsTurboModule: Spec | null = globalThis._WORKLET
  ? null
  : RNWorkletsTurboModule;
```

or this

```ts
import type { Spec } from './NativeWorkletsModule';
import RNWorkletsTurboModule from './NativeWorkletsModule';

export const WorkletsTurboModule = globalThis._WORKLET
  ? null
  : (RNWorkletsTurboModule as Spec);
```

What do you think about that?

tjzel: We can keep it as is, I was just curious.

tjzel (approved):
LGTM

packages/react-native-worklets/src/specs/NativeWorkletsModule.ts
MatiPl01: I already changed it to the first solution of solutions I proposed in the comment above. Should I revert it back or can I keep it as it is now?

PR #7724, opened 2025-07-01
chore: Add reanimatedMajorVersion variable instead of isReanimated3

MatiPl01:
## Summary

This PR adds more future-proof check for the reanimated version and deprecates the old `isReanimated3` check.

### DIFF
--- a/packages/react-native-reanimated/src/core.ts
+++ b/packages/react-native-reanimated/src/core.ts
@@ -33,8 +33,20 @@ export {
 
 const EDGE_TO_EDGE = isEdgeToEdge();
 
-/** @returns `true` in Reanimated 3, doesn't exist in Reanimated 2 or 1 */
-export const isReanimated3 = () => true;
+/**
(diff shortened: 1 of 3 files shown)


packages/react-native-reanimated/src/core.ts
```diff
@@ -33,8 +33,20 @@ export {
...
+ */
+export const isReanimated3 = () => {
+  logger.warn(
+    'The `isReanimated3` function is deprecated. Please use the exported variable `reanimatedMajorVersion` instead.'
+  );
+  return false;
+};
+
+/** @returns Reanimated major version, doesn't exist before Reanimated 4 */
+export const reanimatedMajorVersion = 4;
```
piaskowyk: We can use value from here - https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/src/platform-specific/jsVersion.ts

tomekzaw (approved):

piaskowyk (approved):

PR #7727, opened 2025-07-01
poc: `TSTyche` type tests

skusnierz:

### DIFF
--- a/packages/react-native-reanimated/__typetests__/common/__typetests__.tst.ts
+++ b/packages/react-native-reanimated/__typetests__/common/__typetests__.tst.ts
@@ -1,0 +1,126 @@
+import { useRef } from 'react';
+import { describe, expect, test } from 'tstyche';
+
+import Animated, {
+  makeMutable,
+  measure,
+  useAnimatedRef,
+  useAnimatedStyle,
+  useDerivedValue,
+  useSharedValue,
+  withDecay,
+  withSpring,
+} from '../..';
+
+describe('makeMutable', () => {
+  test('makeMutable<number> should be callable with number', () => {
+    expect(makeMutable<number>).type.toBeCallableWith(0);
+  });
+
+  test('makeMutable<number> should not be callable with string', () => {
+    expect(makeMutable<number>).type.not.toBeCallableWith('a');
+  });
+});
+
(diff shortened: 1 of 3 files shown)


packages/react-native-reanimated/__typetests__/common/__typetests__.tst.ts
```diff
@@ -0,0 +1,126 @@
...
+  });
+});
+
+describe('useDerivedValue', () => {
+  test('should not be able to set value', () => {
+    const progress = useSharedValue(0);
+    const width = useDerivedValue(() => {
+      return progress.value * 250;
+    });
+    expect(() => (width.value = 10)).type.toRaiseError();
```
mrazauskas, line 114: Avoid `.toRaiseError()`. Better import `pick` from `'tstyche'` and rewrite the assertion like this:

```suggestion
    expect(pick(width, "value")).type.toBe<{ readonly value: number }>();
```

I should make this startegy more visible. Currently documented here: https://tstyche.org/reference/expect-api#pick

mrazauskas (commented):
Thanks for considering TSTyche. I am its author, for those who do not know yet (;

Very important! Always run the `tstyche` command. Preferably in CI. Otherwise the assertions do nothing at all.

---

By the way, I see you already have a bunch of type test files. That is great! Just pass them through TSTyche (use the [`testFileMatch`](https://tstyche.org/reference/config-file#test-file-match) option). It will type-check them for you using the installed `typescript`. Or the specified version of TypeScript, or a [range of versions](https://tstyche.org/guides/typescript-versions) if that is important.

If you go for that, remember to enable the [`checkSuppressedErrors`](https://tstyche.org/reference/config-file#check-suppressed-errors) option. It will check errors hidden by `// @ts-expect-error`. So you can be sure you got that exact error, not just “some errors must be there”. (Behind an option, because that is a very new thing. Will be enable by default in the future.)

Please ping me if there are any questions.

skusnierz:
@mrazauskas thanks for the review and for your help ❤️  so far I'm just playing with the library to see what it can do, but it looks promising, if I have any questions I'll ping you 👍

MatiPl01:
Superseded by #9557

PR #7733, opened 2025-07-02
chore: Optimize css keyframes registration in the keyframes registry

MatiPl01:
## Summary

This implementation brings a few optimizations related to the CSS keyframes registration:

1. Allows passing `cssText` (stringified keyframes) to the `CSSKeyframesRule` constructor to prevent converting keyframes object to string twice
2. Improves the JS side `CSSKeyframesRegistry` to allow keyframes selection by the `cssText` (useful for inline CSS keyframes - we can just reuse existing keyframes object from the registry if `cssText` matches)
3. Modifies the `CSSAnimationsManager` to check for inline keyframes existence in the `CSSKeyframesRegistry` by checking if the keyframes with the particular `cssText` exist in this registry instead of just comparing the current keyframes object with the previous one at the same index (the one in the previous CSS animation keyframes array passed to the view).

### DIFF
--- a/packages/react-native-reanimated/src/css/managers/CSSAnimationsManager.ts
+++ b/packages/react-native-reanimated/src/css/managers/CSSAnimationsManager.ts
@@ -115,7 +115,7 @@ export default class CSSAnimationsManager implements ICSSAnimationsManager {
       createSingleCSSAnimationProperties(animationProperties);
 
     const processedAnimations = singleAnimationPropertiesArray.map(
-      (properties, i) => {
+      (properties) => {
         const keyframes = properties.animationName;
         let keyframesRule: CSSKeyframesRuleImpl;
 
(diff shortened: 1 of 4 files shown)


packages/react-native-reanimated/src/css/registry/CSSKeyframesRegistry.ts
```diff
@@ -52,10 +67,20 @@ export default class CSSKeyframesRegistry {
...
     }
   }
+
+  clear() {
+    this.nameToKeyframes_.clear();
+    this.cssTextToNameMap_.clear();
+  }
 }
+
+const cssKeyframesRegistry = new CSSKeyframesRegistry();
```
MatiPl01, line 84: Since there is always a single keyframes registry in CPP, it is safer to ensure that there is only a single instance in JS as well. Instead of storing it as a static prop of the animations manager, it is better to export just a single object.

It is also easier to use, e.g. in tests, where I need to clean it up between unit tests.

packages/react-native-reanimated/src/css/managers/CSSAnimationsManager.ts
```diff
@@ -125,22 +115,20 @@ export default class CSSAnimationsManager implements ICSSAnimationsManager {
...
-          this.attachedAnimations[i]?.keyframesRule.cssText !==
-          JSON.stringify(keyframes)
-        ) {
+        } else {
           // If the keyframes are not an instance of the CSSKeyframesRule class (e.g. someone
           // passes a keyframes object inline in the component's style without using css.keyframes()
           // function), we don't want to restart the animation on every component re-render.
-          // In this case, we need to compare the stringified keyframes of the old and the new
-          // animation configuration object to determine if the animation has changed.
-          keyframesRule = new CSSKeyframesRuleImpl(
```
MatiPl01, line 137: This created a new keyframes object with new animation name if the inline animation at the specific index changed, which was not valid, because the relative order of animations in the `animationName` array may change and it shouldn't start the new animation.

Consider this example:

- Before:

```ts
{ 
  animationName: [
    { from: { opacity: 0 } }, 
    { to: { transform: [{ rotate: '180deg' }] } }
  ]
}
```

- After:

```ts
{ 
  animationName: [
    { to: { transform: [{ rotate: '180deg' }] } },
    { from: { opacity: 0 } }
  ]
}
```

Videos show that before this change, the 2 inline animations were considered different, even tough they should be treated as the same animations.

| Old implementation | New implementation |
|-|-|
| <video src="https://github.com/user-attachments/assets/f8e29d30-1375-4abd-aff5-a504654902a1" /> | <video src="https://github.com/user-attachments/assets/a159724d-ba5b-4a59-a243-b1e0ead7a718" /> |

<details>
<summary>Full code</summary>

```tsx
import { Button } from 'react-native';
import type { CSSAnimationKeyframes } from 'react-native-reanimated';
import Animated, { css } from 'react-native-reanimated';

import { Screen } from '@/apps/css/components';
import { flex } from '@/theme';

export default function Playground() {
  const [animations, setAnimations] = useState<Array<CSSAnimationKeyframes>>([
    { from: { opacity: 0 } },
    { to: { transform: [{ rotate: '180deg' }] } },
  ]);

  const reorder = () => setAnimations((prev) => [...prev].reverse());

  return (
    <Screen style={flex.center}>
      <Button title="Reorder" onPress={reorder} />
      <Animated.View
        style={[
          styles.box,
          {
            animationName: animations,
            animationDuration: '5s',
            animationIterationCount: Infinity,
          },
        ]}
      />
    </Screen>
  );
}

const styles = css.create({
  box: {
    width: 100,
    height: 100,
    backgroundColor: 'red',
  },
});
```
</details>

PR #7736, opened 2025-07-02
feat: Support for separate CSS interpolators for different components

MatiPl01:
## Summary

This PR splits one uber interpolators config into smaller, separate configs for different component types. I followed the separation of `ViewStyle`, `TextStyle`, `ImageStyle`, etc. types in react native and recreated corresponding interpolator maps in CPP separated into different files and maps.

This PR is needed to add support for different interpolated properties, like props of `react-native-svg` components.

## Test plan

I went through all examples under `Animations`/`Animated Properties` to verify that all supported CSS props still work properly.


packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/registry.cpp
```diff
@@ -0,0 +1,36 @@
...
+    {"Paragraph", TEXT_INTERPOLATORS},
+    {"Image", IMAGE_INTERPOLATORS},
+    // svg
+    // TODO
+};
+
+} // namespace
+
+const bool hasInterpolators(const std::string &componentName) {
+  return registry_.find(componentName) != registry_.end();
```
msluszniak: CAFrameRateRangeMake is an `std::unordered_map` so to check existence use `contains` instead

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/registry.cpp
```diff
@@ -0,0 +1,36 @@
...
+    {"View", VIEW_INTERPOLATORS},
+    {"Paragraph", TEXT_INTERPOLATORS},
+    {"Image", IMAGE_INTERPOLATORS},
+    // svg
+    // TODO
+};
+
+} // namespace
+
+const bool hasInterpolators(const std::string &componentName) {
```
msluszniak: Rather remove const from return type here: https://clang.llvm.org/extra/clang-tidy/checks/readability/const-return-type.html

msluszniak:
@MatiPl01 You removed one const too much ;p. When you return const reference, const here is not the top-level const, and it's absolutely fine. Just the const next to bool was incorrect ;)

If this change was because you really want to return the mutable reference, please ignore this comment ;)

MatiPl01:
> @MatiPl01 You removed one const too much ;p. When you return const reference, const here is not the top-level const, and it's absolutely fine. Just the const next to bool was incorrect ;)
> 
> If this change was because you really want to return the mutable reference, please ignore this comment ;)

Yup, thanks for noticing that and clarification. I just automatically removed both, but yeah, I want to return a const reference, so the one of two that I removed should be brought back.

apps/common-app/src/apps/css/examples/animations/screens/animatedProperties/appearance/colors/ColorProperties.tsx
```diff
@@ -317,7 +317,7 @@ function OtherColors() {
           }}
         />
         <ImageExample
-          labelTypes={['iOS', 'Android']}
+          labelTypes={['Android']}
```
tjzel: where ios?

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/base/common.h
```diff
@@ -0,0 +1,139 @@
...
+const InterpolatorFactoriesRecord TRANSFORMS_INTERPOLATORS = {
+    {"transformOrigin",
+     array(
+         {value<CSSDimension>(RelativeTo::Self, "width", "50%"),
+          value<CSSDimension>(RelativeTo::Self, "height", "50%"),
+          value<CSSDouble>(0)})},
+    {"transform",
+     transforms(
+         {{"perspective",
+           transformOp<PerspectiveOperation>(0)}, // 0 - no perspective
```
tjzel, line 121: Is this comment necessary?
```suggestion
           transformOp<PerspectiveOperation>(0)},
```

packages/react-native-reanimated/Common/cpp/reanimated/CSS/util/props.cpp
```diff
@@ -2,9 +2,12 @@
...
 
-bool isDiscreteProperty(const std::string &propName) {
-  const auto it = PROPERTY_INTERPOLATORS_CONFIG.find(propName);
-  if (it == PROPERTY_INTERPOLATORS_CONFIG.end()) {
+bool isDiscreteProperty(
+    const std::string &propName,
+    const std::string &componentName) {
+  const auto &interpolators = getInterpolators(componentName);
+  const auto it = interpolators.find(propName);
+  if (it == interpolators.end()) {
```
tjzel, line 10: Can we use `contains` here?

apps/common-app/src/apps/css/examples/animations/screens/animatedProperties/appearance/colors/ColorProperties.tsx
MatiPl01: Seems that it doesn't work but I will double check if it broke recently or this prop just doesn't work on iOS at all.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/base/common.h
MatiPl01: Mhm. I added it because we cannot animate perspective from `0` to anything else, so `0` is something like `'none'` and the user must explicitly animate from non-zero perspective. I am not sure if we should somehow gracefully handle this case and allow animating from `0` but it is as it is for now.

I can make the comment more explicit if you wish or get rid of it completely.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/util/props.cpp
MatiPl01: I don't know if this change pays off, since I am using the iterator variable `it` in one of the lines below.

I think that we would save just a single line by applying the suggested change:

```cpp
if (!interpolators.contains(propName)) {
  return false;
}
return interpolators[propName]->isDiscreteProperty();
```

Not a big deal for me, but I can change this code if you want me to do so.

tjzel: Then we can leave as is 👍

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/base/common.h
tjzel: You can decide, just for me this comments means nothing as `0` is usually the default value of everything.

apps/common-app/src/apps/css/examples/animations/screens/animatedProperties/appearance/colors/ColorProperties.tsx
MatiPl01: It turns out that on iOS, the `tintColor` prop must be passed in component's prop when it renders for the first time. I added a note and fixed this example in this PR: https://github.com/software-mansion/react-native-reanimated/pull/7860

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/base/common.h
MatiPl01: Well, usually, but not always. e.g. opacity has default value of 1, colors aren't transparent but are black, so it is `#000000ff` instead of `#00000000`, default scale is 1, etc.

PR #7738, opened 2025-07-02
feat(Worklets): implement cancelAnimationFrame

tjzel:
## Summary

No idea why wasn't it implemented before.

I aimed for simplicity, not performance.

## Test plan

🚀

### DIFF
--- a/packages/react-native-worklets/src/animationFrameQueue/requestAnimationFrame.ts
+++ b/packages/react-native-worklets/src/animationFrameQueue/requestAnimationFrame.ts
@@ -4,9 +4,6 @@ import { callMicrotasks } from '../threads';
 
 export function setupRequestAnimationFrame() {
   'worklet';
-
-  // Jest mocks requestAnimationFrame API and it does not like if that mock gets overridden
-  // so we avoid doing requestAnimationFrame batching in Jest environment.
   const nativeRequestAnimationFrame = global.requestAnimationFrame;
 
   let animationFrameCallbacks: Array<(timestamp: number) => void> = [];
@@ -32,10 +29,13 @@ export function setupRequestAnimationFrame() {
         global.__frameTimestamp = undefined;
       });
     }
-    // Reanimated currently does not support cancelling callbacks requested with
-    // requestAnimationFrame. We return -1 as identifier which isn't in line
-    // with the spec but it should give users better clue in case they actually
-    // attempt to store the value returned from rAF and use it for cancelling.
-    return -1;
+    return animationFrameCallbacks.length - 1;
+  };
+
+  global.cancelAnimationFrame = (id: number) => {
+    if (id < 0 || id >= animationFrameCallbacks.length) {
+      return;
+    }
+    animationFrameCallbacks[id] = () => {};
   };
 }


apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/runLoop/requestAnimationFrame.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    await render(
+      <TestComponent
+        worklet={() => {
+          'worklet';
+          requestAnimationFrame(() => notify(notification));
+        }}
+      />,
+    );
+
+    await waitForNotify(notification);
```
bartlomiejbloniarz, line 48: If this test doesn't pass will the test suite just run forever? I guess that's the case for most of them?

packages/react-native-worklets/src/animationFrameQueue/requestAnimationFrame.ts
```diff
@@ -4,38 +4,64 @@ import { callMicrotasks } from '../threads';
...
-      });
+
+      nativeRequestAnimationFrame(globalThis.__flushAnimationFrame);
+    }
+    return handle;
+  };
+
+  globalThis.cancelAnimationFrame = (handle: number) => {
+    if (handle < flushedCallbacksBegin || handle >= callbacksEnd) {
+      return;
```
bartlomiejbloniarz, line 58: Since we have this check, could we have an array instead of the map - just index it with `handle - flushedCallbacksBegin`?

bartlomiejbloniarz: I think we could do that for both the current queue and the "next" one, as we know when they both start?

tjzel: I used a map because the spec said to use a map but now I read more it seems that they only want something that works like a map on a mental level. I will try to move back to an array.

apps/common-app/src/apps/reanimated/examples/RuntimeTests/tests/runLoop/requestAnimationFrame.test.tsx
tjzel: Yes, unfortunately we don't have good promise handling in the framework yet.

piaskowyk:
Could you add some description to the PR or provide a graph? It's a bit hard to understand what's happening (or at least it is for me 😭)

tjzel:
@piaskowyk

When you schedule an animation callback, it lands into the array of `queuedCallbacks` and a request is sent to the native side to execute function `__flushAnimationFrame`. Scheduled callbacks always land in `queuedCallbacks` array.

Once `__flushAnimationFrame` starts, the callbacks from `queuedCallbacks` are moved into `flushedCallbacks` and are executed sequentially. `queuedCallbacks` is now empty. However, the user may schedule more animation callbacks during the execution of `queuedCallbacks`. In this case both `flushedCallbacks` and `queuedCallback` aren't empty.

Therefore, if a user calls `cancelAnimationFrame` we must check whether the callbacks is in `queuedCallbacks` or in `flushedCallbacks`. We check it by keeping the start and end handles (ids) of `queuedCallbacks` and `flushedCallbacks`, namely `queuedCallbacksBegin`, `queuedCallbacksEnd`, `flushedCallbacksBegin`, `flushedCallbacksEnd`. When the handle to cancel falls into the specific range, we replace the callback the user wants to cancel with an empty lambda.

packages/react-native-worklets/src/animationFrameQueue/requestAnimationFrame.ts
```diff
@@ -4,38 +4,64 @@ import { callMicrotasks } from '../threads';
...
+    flushedCallbacksEnd = callbacksEnd;
+    callbacksBegin = callbacksEnd;
+
+    flushRequested = false;
+
+    for (const callback of flushedCallbacks) {
+      callback(timestamp);
+    }
+
+    flushedCallbacksBegin = callbacksEnd;
```
piaskowyk: It would be more intuitive if you use the variable family related to the same callback set.

```suggestion
    flushedCallbacksBegin = flushedCallbacksEnd;
```

packages/react-native-worklets/src/animationFrameQueue/requestAnimationFrame.ts
```diff
@@ -4,38 +4,64 @@ import { callMicrotasks } from '../threads';
...
+
+  globalThis.cancelAnimationFrame = (handle: number) => {
+    if (handle < flushedCallbacksBegin || handle >= callbacksEnd) {
+      return;
+    }
+
+    if (handle < flushedCallbacksEnd) {
+      flushedCallbacks[handle - flushedCallbacksBegin] = () => {};
+    } else {
+      animationFrameCallbacks[handle - callbacksBegin] = () => {};
```
piaskowyk: It would be super helpful if you used the `flushedCallbacks` and `queuedCallback` naming convention that you mentioned in the comment above.

piaskowyk (commented):
Thank you for the explanation - it was super useful for me. I don't have any more suggestions. It looks good 👍

packages/react-native-worklets/src/animationFrameQueue/requestAnimationFrame.ts
tjzel: makes sense, renamed 👍

bartlomiejbloniarz (approved):

PR #7739, opened 2025-07-02
feat(Worklets): RuntimeManager

tjzel:
## Summary

Introducing a new class, `RuntimeManager`, responsible for creating new Worklet Runtimes and storing information about them, but not owning them. A runtime is still owned by its creator.

I have added `getAllRuntimes` method for the purpose of hot-reload in Bundle Mode. During hot-reload I need to forward the update to all runtimes, that's why I need a class that allows that. In the future we might handle hot-reloads some other way, but this is the way to go now.

- https://github.com/software-mansion/react-native-reanimated/pull/7617 will make the code less redundant.

## Test plan

Run the app, see that spawning a new runtime/UI runtime works properly. Reload the app.


packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/RuntimeManager.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+auto RuntimeManager::getAllRuntimes()
+    -> std::vector<std::shared_ptr<WorkletRuntime>> {
+  std::shared_lock lock(weakRuntimesMutex_);
+
+  auto runtimes = std::vector<std::shared_ptr<WorkletRuntime>>{};
+  runtimes.reserve(weakRuntimes_.size());
+
+  for (const auto &pair : weakRuntimes_) {
+    if (auto runtime = pair.second.lock()) {
```
msluszniak: It's just a suggestion, I don't know the accurate names for these variables, but this way you make the code more self-describing and remove using explicitly `second` argument. If you do not support structured bindings aka c++17, resolve this comment.
```cpp
  for (const auto &[first_arg, second_arg] : weakRuntimes_) {
    if (auto runtime = second_arg.lock()) {
```

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/RuntimeManager.h
```diff
@@ -0,0 +1,69 @@
...
+  auto getAllRuntimes() -> std::vector<std::shared_ptr<WorkletRuntime>>;
+
+  auto getUIRuntime() -> std::shared_ptr<WorkletRuntime>;
+
+  auto createWorkletRuntime(
+      std::shared_ptr<JSIWorkletsModuleProxy> jsiWorkletsModuleProxy,
+      const std::shared_ptr<MessageQueueThread> &jsQueue,
+      const std::shared_ptr<JSScheduler> &jsScheduler,
+      const bool isDevBundle,
+      const bool supportsLocking,
```
msluszniak: Top const qualifiers in function declaration, remove them: https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/RuntimeManager.cpp
msluszniak: However, when compiled with `-Wunused-variable` you'll probably got warning about unused variable `first_arg`. To suppress this error, you will be either forced to stay with your implementation or use `[[maybe_unsused]]` for the whole structured binding since there is no support in c++ to mark only parts of structured bindings with `[[maybe_unsused]]`. This will make mentioned approach less attractive :/

packages/react-native-worklets/apple/worklets/apple/WorkletsModule.mm
```diff
@@ -87,8 +88,8 @@ - (void)checkBridgeless
       script,
       sourceURL);
   auto jsiWorkletsModuleProxy = workletsModuleProxy_->createJSIWorkletsModuleProxy();
-  auto optimizedJsiWorkletsModuleProxy =
-      worklets::jsi_utils::optimizedFromHostObject(rnRuntime, std::move(jsiWorkletsModuleProxy));
+  auto optimizedJsiWorkletsModuleProxy = worklets::jsi_utils::optimizedFromHostObject(
```
tomekzaw, line 91: What's the reason why now we need `std::dynamic_pointer_cast` here?

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.h
```diff
@@ -69,7 +70,16 @@ class WorkletRuntime : public jsi::HostObject,
...
+  [[nodiscard]] auto getRuntimeId() const -> uint64_t {
+    return runtimeId_;
+  }
+
+  [[nodiscard]] auto getRuntimeName() const -> std::string {
+    return name_;
+  }
+
  private:
+  const uint64_t runtimeId_;
```
tomekzaw, line 82: What's the reason why we need an integer runtime ID now?

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.h
```diff
@@ -69,7 +70,16 @@ class WorkletRuntime : public jsi::HostObject,
 
   std::vector<jsi::PropNameID> getPropertyNames(jsi::Runtime &rt) override;
 
+  [[nodiscard]] auto getRuntimeId() const -> uint64_t {
```
tomekzaw, line 73: As said previously, I'm not a fan of using `auto getRuntimeId() const -> uint64_t { ... }` in a codebase where we mostly use `uint64_t getRuntimeId() const { ... }`. Let's discuss it with the team.

packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
```diff
@@ -60,27 +60,28 @@ inline jsi::Value executeOnUIRuntimeSync(
 }
 
 inline jsi::Value createWorkletRuntime(
+    std::shared_ptr<JSIWorkletsModuleProxy> jsiWorkletsModuleProxy,
     const std::shared_ptr<MessageQueueThread> &jsQueue,
     const std::shared_ptr<JSScheduler> &jsScheduler,
-    std::shared_ptr<JSIWorkletsModuleProxy> jsiWorkletsModuleProxy,
```
tomekzaw, line 65: I don't think this change was necessary, it makes the diff more difficult to compare and the order of arguments is not critical here. Feel free to leave it as it is now though.

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/RuntimeManager.cpp
```diff
@@ -0,0 +1,107 @@
+#include <worklets/NativeModules/JSIWorkletsModuleProxy.h>
+#include <worklets/WorkletRuntime/RuntimeManager.h>
+
+#include <utility>
+
+namespace worklets {
+
+const std::string uiRuntimeName{"UI"};
```
tomekzaw, line 8: Do we really need to have both runtime ID and runtime name? Also, it looks like runtime names should be unique (as they are used as map keys) so it seems they serve the same purpose as runtime IDs.

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/RuntimeManager.cpp
```diff
@@ -0,0 +1,107 @@
...
+    return getRuntime(nameToRuntimeId_.at(name));
+  }
+  return nullptr;
+}
+
+auto RuntimeManager::getAllRuntimes()
+    -> std::vector<std::shared_ptr<WorkletRuntime>> {
+  std::shared_lock lock(weakRuntimesMutex_);
+
+  auto runtimes = std::vector<std::shared_ptr<WorkletRuntime>>{};
```
tomekzaw:
```suggestion
  std::vector<std::shared_ptr<WorkletRuntime>> runtimes;
```

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.h
msluszniak: Yeah, trailing return type is generally not prefered even though there is clang tidy check to modernize it: https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-trailing-return-type.html. Generally, it should be used when there is a non-trivial template when type is deduced like: template with two diffrent elements, and return type is type of multiplication of these types, so you use `auto fun(A a, B b) -> decltype(A*B)` etc.

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/RuntimeManager.h
tjzel: That's a good catch but I honor the rule that if something (minor) is not enforced on CI it shouldn't be enforced in review, since it adds more labor with very small returns. We use not very powerful `cpplint` for linting and I believe it doesn't have this check available. We also use `clang-tidy` on a separate CI and we plan to switch over to it, but for the time being it's extremely slow and we need to think how to configure it to be actually usable.

packages/react-native-worklets/apple/worklets/apple/WorkletsModule.mm
tjzel: The function signature changed. Previously `createJSIWorkletsModuleProxy()` would return a `jsi::HostObject`, now it actually returns `JSIWorkletsModuleProxy` and `optimizedFromHostObject` requires a `shared_ptr` to a host object. Perhaps we should instead change it that `optimizedFromHostObject` takes an actual HO instead of a pointer.

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.h
tjzel: Initially I thought we'd ideally get a UUID but that seemed to be an overkill for the moment.

While strings are nice I think they are not the best choice for map keys in C++. Also I'm not sure how we'd handle conflicting runtime names, so I decided to use an id internally and retain the string as more user-friendly external way of identifying runtimes.

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/WorkletRuntime.h
tjzel: I believe that in c++20 (correct me if I'm wrong) trailing return types don't have limitations.

I like trailing return types because they allow for slightly more readable code, function names are aligned and the type is on the right, just like in TS, Kotlin, Rust etc.

I figured that for an entirely new file I can use trailing return types because it wouldn't be internally inconsistent. However, if you prefer I can rewrite it to non-trailing return types.

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/RuntimeManager.cpp
tjzel: I couldn't really make my mind about this so I left both. More on that in another comment.

packages/react-native-worklets/Common/cpp/worklets/NativeModules/JSIWorkletsModuleProxy.cpp
tjzel: You are right, I shouldn't have done this. I think some OCD argument importance sorting kicked in on my end.

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/RuntimeManager.cpp
tjzel: I like destructuring assignments 🫡

packages/react-native-worklets/Common/cpp/worklets/WorkletRuntime/RuntimeManager.h
tjzel: Also for a constructor this long I believe we should instead make a struct parameter instead for better readibility but I don't want to bloat this PR.

tjzel:
@piaskowyk I removed trailing return types and changed `dynamic_pointer_cast` to compilation `static_pointer_cast`.

PR #7741, opened 2025-07-03
feat: Support for first `react-native-svg` properties in CSS

MatiPl01:


packages/react-native-reanimated/src/css/platform/native/configs/svg/circle.ts
```diff
@@ -0,0 +1,7 @@
+'use strict';
+export const SVG_CIRCLE_PROPERTIES_CONFIG = {
```
MatiPl01, line 2: TODO: Find a way to use types from `react-native-svg` without requiring lib installation from the user. I don't know if we can do this somehow but I would like to use strict typing here.

cc @tjzel

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/registry.h
```diff
@@ -6,6 +6,8 @@
 #include <reanimated/CSS/config/interpolators/base/image.h>
 #include <reanimated/CSS/config/interpolators/base/text.h>
 #include <reanimated/CSS/config/interpolators/base/view.h>
+// react-native-svg style props interpolators
+#include <reanimated/CSS/config/interpolators/svg/circle.h>
```
tjzel:
```suggestion
#include <reanimated/CSS/config/interpolators/svg/circle.h>
```

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/registry.cpp
```diff
@@ -9,8 +9,8 @@ ComponentInterpolatorsMap registry_ = {
     {"View", VIEW_INTERPOLATORS},
     {"Paragraph", TEXT_INTERPOLATORS},
     {"Image", IMAGE_INTERPOLATORS},
-    // svg
-    // TODO
+    // react-native-svg
+    {"RNSVGCircle", SVG_CIRCLE_INTERPOLATORS},
```
tjzel:
```suggestion
    {"RNSVGCircle", SVG_CIRCLE_INTERPOLATORS},
```

packages/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp
```diff
@@ -433,16 +433,26 @@ void ReanimatedModuleProxy::unmarkNodeAsRemovable(
...
 void ReanimatedModuleProxy::unregisterCSSKeyframes(
     jsi::Runtime &rt,
-    const jsi::Value &animationName) {
-  cssAnimationKeyframesRegistry_->remove(animationName.asString(rt).utf8(rt));
+    const jsi::Value &animationName,
+    const jsi::Value &viewName) {
+  cssAnimationKeyframesRegistry_->remove(
+      animationName.asString(rt).utf8(rt),
+      // Convert react view name to Fabric component name
+      componentNameByReactViewName(viewName.asString(rt).utf8(rt)));
```
tjzel:
```suggestion
      componentNameByReactViewName(viewName.asString(rt).utf8(rt)));
```

packages/react-native-reanimated/src/css/platform/native/configs/svg/circle.ts
tjzel: I think installing `react-native-svg` as a dev-dependency is the best bet here, as we previously discussed offline.

packages/react-native-reanimated/src/css/platform/native/configs/svg/circle.ts
```diff
@@ -0,0 +1,7 @@
...
+  // TODO - add more properties
```
tjzel, line 3: Do you have another PR that resolves this TODO?

packages/react-native-reanimated/src/css/platform/native/configs/svg/circle.ts
MatiPl01: Ok, so I will leave it as it is

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/registry.h
MatiPl01: Don't you like this logical separation of includes? I know that the `svg` part in the path might be enough but I wanted to make it explicit that these interpolators are needed for RN SVG integration.

packages/react-native-reanimated/src/css/platform/native/configs/svg/circle.ts
MatiPl01: This one added support for all common SVG properties, as well as the remaining circle props: https://github.com/software-mansion/react-native-reanimated/pull/7750/files.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/registry.h
tjzel: I think the code is self explanatory here and adding include separatio increases the mental capacity you need to read it instead of reducing it.

PR #7742, opened 2025-07-03
Fix race-condition during component initialization

piaskowyk:
## Problem Overview

When a child receives an animated style as a prop from a parent component and the animation is already running, it's possible for heavy components that the view descriptor registration will be called after the animation has finished. As a result, no effect of the animation will be applied, or the component will end up with a state from the middle of the animation.

## Solution

This PR modifies the view descriptor's `add` method such that after the registration of the descriptor, it triggers a force update of the component on the UI. This ensures that the component will have all potentially already computed animation styles applied.

## Demo

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/26d82ed1-a7b2-4875-8260-de7fd394e9e7" /> | <video src="https://github.com/user-attachments/assets/fc2b02f3-8bda-46a0-b3b1-1e9f04f48ab6" /> |

## Example

<details>
<summary>code</summary>

```js
import { Button, StyleSheet, View, Text } from 'react-native';
import Animated, { useAnimatedStyle, useSharedValue, withTiming } from 'react-native-reanimated';

function InnerComponent({index, aniamtedStyle}: {index: number, aniamtedStyle: any}) {

  useEffect(() => {
    console.log(Date.now(), 'loop')
    let a = 0;
    for (let i = 0; i < 9999999; i++) {
      a = a * a + 6;
    }
  })

  return <Animated.View 
    style={[
      {width: 100, height: 100, backgroundColor: 'olive', borderColor: 'blue', borderWidth: 3},
      aniamtedStyle
    ]} 
  >
    <Text>{index}</Text>
  </Animated.View>
}

function Wrapper({animation, sv}: {animation: any, sv: any}) {
  const [index, setIndex] = useState(0);
  return <View style={styles.container}>
    <Button title='click' onPress={() => {
      if (sv.value > 150) {
        sv.value = withTiming(100, {duration: 60});
      } else { 
        sv.value = withTiming(300, {duration: 60});
      }
      setIndex(index + 1);
    }} />
    <InnerComponent index={index} key={index} aniamtedStyle={animation} />
    <InnerComponent index={index + 1} key={index + 1} aniamtedStyle={animation} />
  </View>
}

  const width = useSharedValue(100);
    return { width: width.value }
  });
  return (<View style={{marginTop: 60}}>
    <Wrapper animation={style} sv={width}/>
  </View>);
}

  container: {},
});

```

</details>


packages/react-native-reanimated/src/ViewDescriptorsSet.ts
```diff
@@ -1,19 +1,20 @@
 'use strict';
-import type { SharedValue } from './commonTypes';
+import type { ForceUpdateContainer, SharedValue } from './commonTypes';
 import { makeMutable } from './core';
 import type { Descriptor } from './hook/commonTypes';
 
 export interface ViewDescriptorsSet {
   shareableViewDescriptors: SharedValue<Descriptor[]>;
-  add: (item: Descriptor) => void;
+  add: (item: Descriptor, forceUpdateCounter?: ForceUpdateContainer) => void;
```
bartlomiejbloniarz: forceUpdateCounter?

PR #7744, opened 2025-07-03
feat(Reanimated): add troubleshooting when Worklets are not installed

tjzel:
## Summary

WIP


packages/docs-reanimated/docs/guides/troubleshooting.mdx
```diff
@@ -278,3 +278,21 @@ In this example, `callee` cannot be called from a worklet ran on UI thread becau
...
 
 Check [this page](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/glossary#worklet) to learn more about Reanimated and its worklets.
+
+## Dependency issues
+
+### Unable to find a specification for `RNWorklets` depended upon by `RNReanimated`
+
+**Problem:** This happens when you have installed Reanimated 4 but don't have `react-native-worklets` installed as a dependency.
+
+**Platform:** iOS
```
MatiPl01, line 288: Is this related to the fact that you weren't able to add an error thrown in runtime or during the build process?

I think that this is a bit misleading as it looks at the first glance that the error is related only to the iOS platform and missing worklets issue cannot happen on other platforms (at least it is misleading to me).

packages/react-native-reanimated/scripts/worklets-version.json
```diff
@@ -0,0 +1,4 @@
+{
+  "min": "0.4.0",
+  "max": "1000.0.0"
```
MatiPl01, line 3: This is a placeholder, right?

MatiPl01 (approved):
Looks good. I left just to not important comments.

packages/react-native-reanimated/scripts/reanimated_utils.rb
```diff
@@ -16,6 +16,7 @@ def find_config()
     :react_native_common_dir => nil,
     :dynamic_frameworks_reanimated_dir => nil,
     :dynamic_frameworks_worklets_dir => nil,
+    :has_compatible_worklets => nil,
```
piaskowyk: Do we use it somewhere?

packages/react-native-reanimated/scripts/worklets-version.json
piaskowyk: I believe we should have a symmetrical mechanism for worklets. Instead of storing information only in Reanimated regarding supported versions, we should also store information in worklets about which version of Reanimated they are compatible with. It is essential to maintain this information in both libraries.

Consider this scenario: I am using Reanimated version X, which supports worklets ranging from versions 1 to max. If worklets introduce a breaking change and I upgrade the worklets version, I won't receive any notification about the incompatibility unless I also upgrade my Reanimated version.

This situation could be improved if compatibility information were stored directly in the worklets. That way, when I upgrade the worklets version, I can easily check if my current Reanimated version is compatible. By implementing symmetrical mechanisms and ensuring that both sides meet their requirements, we can be sure that user has compatible version.

tjzel:
@piaskowyk the "1000.0.0" is a temporary placeholder. Once we release stable Reanimated 4 we'll limit it to a given version range of worklets which will have to retain backwards compatibility within that range. I.e. if Reanimated 4.0.0 releases and the last version of Worklets is 0.5.0, we'll make it require `react-native-worklets@~0.5.0`. By design version 0.6.0 could not be compatible with 4.0.0 but with a higher version, i.e. 4.1.0.

packages/docs-reanimated/docs/guides/troubleshooting.mdx
tjzel: The idea is that this specific error only happens on that platform. Other platforms have different error messages in this kind of situations and their error messages already include troubleshooting.

packages/react-native-reanimated/scripts/worklets-version.json
tjzel: The "1000.0.0" is a temporary placeholder. Once we release stable Reanimated 4 we'll limit it to a given version range of worklets which will have to retain backwards compatibility within that range. I.e. if Reanimated 4.0.0 releases and the last version of Worklets is 0.5.0, we'll make it require react-native-worklets@~0.5.0. By design version 0.6.0 could not be compatible with 4.0.0 but with a higher version, i.e. 4.1.0.

I wanted to use JSON with comments to add the comment about it being temporary but the framework refuses to parse it properly.

piaskowyk: ok, that convinces me 👍

packages/react-native-reanimated/scripts/reanimated_utils.rb
tjzel: Actually not, good catch!

packages/react-native-reanimated/scripts/worklets-version.json
MatiPl01: Can we use the `jsonc` file? This is a format that supports comments in json.

PR #7747, opened 2025-07-04
fix: add missing handling of `box-shadow` prop

skusnierz:
## Summary
This PR adds missing handling of `box-shadow` prop 

Before : 
<video src="https://github.com/user-attachments/assets/625abd2b-cf5e-477d-87e6-4914fcda7a55" />

After :
<video src="https://github.com/user-attachments/assets/9ca117bd-1e6f-46b5-9af9-aa6f2d2ad97a" />

## Test plan
<details>
<summary>Empty Example - Code snippet</summary>

```js
import React, { useEffect } from 'react';
import { StyleSheet, Text, View } from 'react-native';
import Animated, {
  AnimatableValue,
  useAnimatedStyle,
  useSharedValue,
  withSpring,
} from 'react-native-reanimated';
import { DefaultStyle } from 'react-native-reanimated/lib/typescript/hook/commonTypes';

export default function EmptyExample() {
  const boxShadowActiveSV = useSharedValue({
    blurRadius: 7,
    color: 'blue',
    offsetX: 20,
    offsetY: 20,
    spreadDistance: 10,
  });

  const styleActive = useAnimatedStyle(() => {
    return {
      boxShadow: [
        withSpring(boxShadowActiveSV.value as unknown as AnimatableValue, {
          duration: 700,
        }),
      ],
    } as DefaultStyle;
  });

  useEffect(() => {
    setTimeout(() => {
      boxShadowActiveSV.value = {
        blurRadius: 10,
        color: 'red',
        offsetX: 30,
        offsetY: 30,
        spreadDistance: 20,
      };
    }, 1000);
  }, []);

  return (
    <View style={styles.container}>
      <Animated.View style={[styles.box, styleActive]} />
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: {
    width: 100,
    height: 100,
    backgroundColor: 'palevioletred',
    boxShadow: '20px 20px 5px 0px red',
  },
});
```
</details>

PR #7748, opened 2025-07-04
fix: remove unnecessary process `box-shadow` call

skusnierz:
## Summary
After merging this [PR]( https://github.com/software-mansion/react-native-reanimated/pull/7747), I noticed that I introduced a regression using `useAnimatedStyle` with a passive `box-shadow` style value : 

```js
  const stylePassive = useAnimatedStyle(() => {
    return {
      boxShadow: [boxShadowPassiveSV.value],
    } as DefaultStyle;
  });
```

this was due to the fact that in `useAnimatedStyle` we had already processed the `box-shadow` style, so in the previous [PR]( https://github.com/software-mansion/react-native-reanimated/pull/7747) I added another call of `processBoxShadow` which caused the color to be processed twice : 
```js
((colorNumber << 24) | (colorNumber >>> 8)) >>> 0; // alpha rgb
```
and the value was incorrect 😞 
This PRs removes unnecessary `processBoxShadow` call in `useAnimatedStyle`.

Before : 
<video src="https://github.com/user-attachments/assets/967bd859-d979-4b8c-8488-baa4b638d2b7" />

After : 
<video src="https://github.com/user-attachments/assets/3d609f04-3716-47e9-85aa-2acf3e4faf54" />

## Test plan

<details>
<summary>Empty Example - Code snippet</summary>

```js
import Animated, {
  AnimatableValue,
  useAnimatedStyle,
  useSharedValue,
  withSpring,
import { DefaultStyle } from 'react-native-reanimated/lib/typescript/hook/commonTypes';

const finalBoxShadow = {
  blurRadius: 10,
  color: '#ff0000ff',
  offsetX: 30,
  offsetY: 30,
  spreadDistance: 10,
};

const startBoxShadow = {
  blurRadius: 10,
  color: '#ff0000fe',
  offsetX: 20,
  offsetY: 20,
  spreadDistance: 10,
};

  const boxShadowActiveSV = useSharedValue(startBoxShadow);
  const boxShadowPassiveSV = useSharedValue(startBoxShadow);

  const styleActive = useAnimatedStyle(() => {
    return {
      boxShadow: [
        withSpring(boxShadowActiveSV.value as unknown as AnimatableValue, {
          duration: 5000,
        }),
      ],
    } as DefaultStyle;
  });

  const stylePassive = useAnimatedStyle(() => {
    return {
      boxShadow: [boxShadowPassiveSV.value],
    } as DefaultStyle;
  });

  useEffect(() => {
    const timeout = setTimeout(() => {
      boxShadowActiveSV.value = finalBoxShadow;
      boxShadowPassiveSV.value = finalBoxShadow;
    }, 2000);

    return () => clearTimeout(timeout);
  }, [boxShadowActiveSV, boxShadowPassiveSV]);

  return (
      <Animated.View style={[styles.animatedBox, styleActive]} />
      <Animated.View style={[styles.animatedBox, stylePassive]} />
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    gap: 40,
  },
  animatedBox: {
    width: 100,
    height: 100,
    backgroundColor: 'blue',
  },
});

```
</details>


packages/react-native-reanimated/src/hook/useAnimatedStyle.ts
```diff
@@ -221,13 +216,6 @@ function styleUpdater(
       animations[key] = value;
       hasAnimations = true;
     } else {
-      /* TODO: Improve this config structure in the future
```
MatiPl01, line 224: Maybe add this comment near the if statement in the `updateProps`. I think that we should clean up this a little bit in the future instead of adding more `if ... else` statements, especially when we add support for `backgroundImage` (which is still experimental) and other props that would require parsing.

MatiPl01 (approved):
LGTM

skusnierz:
> Does it work in Web like this?

@tjzel it doesn't work, but it didn't work before my changes either 😞 I'll check it

tjzel (commented):
Can this fix be verified with any kind of tests?

packages/react-native-reanimated/src/common/processors/shadows.web.ts
```diff
@@ -37,5 +37,10 @@ export const processBoxShadow: ValueProcessor<
...
 
-export const processShadowOffset: ValueProcessor<ShadowOffset> = (value) =>
-  `${value.width}px ${value.height}px`;
+export const processShadowOffset: ValueProcessor<
+  ShadowOffset,
+  Record<string, string> | string
+> = (value) => ({
+  width: `${value.width}px`,
+  height: `${value.height}px`,
+});
```
MatiPl01: Let's move it to `packages/react-native-reanimated/src/css/platform/web/style/builders/shadows.ts`

packages/react-native-reanimated/src/common/utils/suffix.ts
```diff
@@ -0,0 +1,7 @@
+export function hasSuffix(value: unknown): value is string {
```
MatiPl01, line 3: Please also remove these implementations from `packages/react-native-reanimated/src/css/platform/web/utils.ts` and import them in there from the `common` dir.

MatiPl01:
As we decided on Slack, we will wait with this change until we find a better solution to maintain proper props preprocessing. Let's wait for our Wednesday's meeting to discuss about this and maybe we would have to do a bigger refactor here for easier maintenance and better consistency.

packages/react-native-reanimated/src/common/processors/shadows.web.ts
```diff
@@ -1,12 +1,12 @@
...
 import type { BoxShadowValue, ViewStyle } from 'react-native';
 
-import { parseBoxShadowString } from '../../../../../common';
-import { maybeAddSuffix } from '../../utils';
 import type { ValueProcessor } from '../types';
+import { maybeAddSuffix, parseBoxShadowString } from '../utils';
 
-export const processBoxShadow: ValueProcessor<
-  string | ReadonlyArray<BoxShadowValue>
+export const processBoxShadowWeb: ValueProcessor<
```
MatiPl01: I know that I suggested using `shadows.web.ts`, but since the return type of the value processor for web differs from the return type for the native one and the web config (`packages/react-native-reanimated/src/css/platform/web/config.ts`) uses strict typing that doesn't allow to pass a processor that returns a type different than `string`, I think that it would be better to move this implementation from the `shadows.web.ts` to the `shadows.ts`.

In general, the use of files with `.web.ts` extension usually makes when we are exporting the same functions (having the same name) but with different implementations. In this case, I think that it will be better do do something like this:

<details>

```tsx
'use strict';
'worklet';
import type { BoxShadowValue } from 'react-native';

import { IS_ANDROID } from '../constants';
import { ReanimatedError } from '../errors';
import type { ValueProcessor } from '../types';
import { maybeAddSuffix, parseBoxShadowString } from '../utils';
import { processColor } from './colors';

const ERROR_MESSAGES = {
  notArrayObject: (value: object) =>
    `Box shadow value must be a string or an array of shadow objects (e.g. [{ offsetX, offsetY, color }]). Received: ${JSON.stringify(value)}.`,

  invalidColor: (color: string, boxShadow: string) =>
    `Invalid color "${color}" in box shadow "${boxShadow}".`,
};

export type ProcessedBoxShadowValue = {
  offsetX: number;
  offsetY: number;
  blurRadius?: number;
  color?: number;
  spreadDistance?: number;
  inset?: boolean;
};

const parseBlurRadius = (value: string) => {
  if (IS_ANDROID) {
    // Android crashes when blurRadius is smaller than 1
    return Math.max(parseFloat(value), 1);
  }
  return parseFloat(value);
};

export const processBoxShadowNative: ValueProcessor<
  ReadonlyArray<BoxShadowValue> | string,
  ProcessedBoxShadowValue[]
> = (value) => {
  if (value === 'none') {
    return;
  }

  const parsedShadow =
    typeof value === 'string' ? parseBoxShadowString(value) : value;

  if (!Array.isArray(parsedShadow)) {
    throw new ReanimatedError(ERROR_MESSAGES.notArrayObject(parsedShadow));
  }

  return parsedShadow.map<ProcessedBoxShadowValue>((shadow) => {
    const {
      color = '#000',
      offsetX = 0,
      offsetY = 0,
      spreadDistance = 0,
      blurRadius = 0,
      ...rest
    } = shadow;
    const processedColor = processColor(color);

    if (processedColor === null) {
      throw new ReanimatedError(
        ERROR_MESSAGES.invalidColor(color, JSON.stringify(shadow))
      );
    }

    return {
      ...rest,
      blurRadius: parseBlurRadius(blurRadius as string),
      color: processedColor,
      offsetX: parseFloat(offsetX as string),
      offsetY: parseFloat(offsetY as string),
      spreadDistance: parseFloat(spreadDistance as string),
    };
  });
};

export const processBoxShadowWeb: ValueProcessor<
  string | ReadonlyArray<BoxShadowValue>,
  string
> = (value) => {
  const parsedShadow =
    typeof value === 'string' ? parseBoxShadowString(value) : value;

  return parsedShadow
    .map(
      ({
        offsetX,
        offsetY,
        color = '#000',
        blurRadius = '',
        spreadDistance = '',
        inset = '',
      }) =>
        [
          maybeAddSuffix(offsetX, 'px'),
          maybeAddSuffix(offsetY, 'px'),
          maybeAddSuffix(blurRadius, 'px'),
          maybeAddSuffix(spreadDistance, 'px'),
          color,
          inset ? 'inset' : '',
        ]
          .filter(Boolean)
          .join(' ')
    )
    .join(', ');
};
```
</details>

And then, import them separately, depending on which implementation is needed. I think it will be cleaner.

MatiPl01 (approved):

PR #7749, opened 2025-07-04
feat: Accept multiple values and CSS props in `animatedProps`

MatiPl01:
## Summary

This PR uses a similar approach to the `style` property and allows passing css properties via the `animatedProps` prop.

Passing CSS style object to the `style` and `animatedProps` prop at the same time is disallowed, because we cannot determine how the result should look like. Since, for all `react-native` components, CSS properties should be passed via the `style` prop but for all `react-native-svg` components via the `animatedProps` prop (as these components don't have the `style` prop), I think that trying to add support for css props passed via both, `style` and `animatedProps`, is pointless and we can just show a warning saying that it ignores one of the provided CSS configs in this case.

## Test plan

Set the following feature flag to `true`:

```json
"UNSTABLE_CSS_ANIMATIONS_FOR_SVG_COMPONENTS": true
```

in the `packages/react-native-reanimated/src/featureFlags/staticFlags.json` file and rebuild the project before running the example.

<details>

```tsx
import Animated, {
  interpolate,
  useAnimatedProps,
  useSharedValue,
  withRepeat,
  withTiming,
import Svg, { Circle } from 'react-native-svg';

export default function SvgPolygonsDemo() {
  const sv = useSharedValue(0);

  useEffect(() => {
    sv.value = withRepeat(withTiming(1), -1, true);
  }, [sv]);

  const animatedProps = useAnimatedProps(
    () => ({
      r: interpolate(sv.value, [0, 1], [10, 20]),
    }),
    []
  );

  return (
    <Svg height="400" width="400">
      <AnimatedCircle
        fill="blue"
        animatedProps={[
          {
            animationName: {
              from: {
                cy: 400,
              },
              to: {
                cx: 400,
              },
            },
            animationDuration: '5s',
            animationIterationCount: 'infinite',
            animationDirection: 'alternate',
          },
          animatedProps,
        ]}
      />
    </Svg>
  );
}
```
</details>

https://github.com/user-attachments/assets/8205a39f-b432-4a55-8b57-780ae9a90ecc

### DIFF
--- a/packages/react-native-reanimated/src/createAnimatedComponent/AnimatedComponent.tsx
+++ b/packages/react-native-reanimated/src/createAnimatedComponent/AnimatedComponent.tsx
@@ -307,7 +307,7 @@ export default class AnimatedComponent
     this._prevAnimatedProps = this._animatedProps;
     this._animatedProps = filteredAnimatedProps.animatedStyles;
 
-    if (filteredStyles.cssStyle && filteredAnimatedProps.cssStyle) {
+    if (__DEV__ && filteredStyles.cssStyle && filteredAnimatedProps.cssStyle) {
       logger.warn(
         'AnimatedComponent: CSS properties cannot be used in style and animatedProps at the same time. Using properties from animatedProps.'
       );


packages/react-native-reanimated/src/createAnimatedComponent/AnimatedComponent.tsx
```diff
@@ -202,46 +203,33 @@ export default class AnimatedComponent
     }
   }
 
-  _attachAnimatedStyles() {
-    const animatedProps = this.props.animatedProps;
-    const prevAnimatedProps = this._animatedProps;
-    this._animatedProps = animatedProps;
-
+  _handleAnimatedStylesUpdate(
```
MatiPl01: Since animated props work in the same way as animated styles, I just decided to reuse the same logic as used for animated styles. I think that we should somehow clean up this `AnimatedComponent` implementation in the future and fix up types (most of types used resolve to `any` and are invalid), but it is not a subject of this PR.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/interpolation/groups/RecordPropertiesInterpolator.cpp
```diff
@@ -23,7 +23,7 @@ void RecordPropertiesInterpolator::updateKeyframes(
     jsi::Runtime &rt,
     const jsi::Value &keyframes) {
   // TODO - maybe add a possibility to remove interpolators that are no longer
-  // used (for now, for simplicity, we only add new ones)
+  // used  (for now, for simplicity, we only add new ones)
```
tjzel:
```suggestion
  // used (for now, for simplicity, we only add new ones)
```

packages/react-native-reanimated/__typetests__/common/withSpringTest.tsx
tjzel, line 1: ?

packages/react-native-reanimated/src/helperTypes.ts
```diff
@@ -99,7 +101,9 @@ export type AnimatedProps<Props extends object> = RestProps<Props> &
      *
      * @see https://docs.swmansion.com/react-native-reanimated/docs/core/useAnimatedProps
      */
-    animatedProps?: Partial<AnimatedPropsProp<Props>>;
+    animatedProps?: AddArrayPropertyType<
+      Partial<AnimatedPropsProp<Props>> | CSSStyle<Props>
+    >;
```
tjzel, line 105: Please add some type tests for that, the `animatedProps` and `style` props are notoriously prone to type errors.

packages/react-native-reanimated/src/css/managers/CSSAnimationsManager.ts
tjzel, line 1: Please restore, no need to change the order here.

packages/react-native-reanimated/src/createAnimatedComponent/utils.ts
```diff
@@ -36,18 +36,17 @@ export const has = <K extends string>(
...
+type FilteredStyles = {
+  cssStyle: CSSStyle | null;
   animatedStyles: StyleProps[];
-} {
+};
+
+export function filterStyles(styles: StyleProps[] | undefined): FilteredStyles {
   if (!styles) {
-    return { animatedStyles: [], cssStyle: {} };
+    return { animatedStyles: [], cssStyle: null };
```
tjzel, line 46: Why null instead of an empty object?

packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx
```diff
@@ -58,15 +58,23 @@ export class PropsFilter implements IPropsFilter {
...
+        animatedPropsArray.forEach((animatedProps) => {
+          if (animatedProps?.viewDescriptors) {
+            if (animatedProps.initial !== undefined) {
+              Object.keys(animatedProps.initial.value).forEach(
+                (initialValueKey) => {
+                  props[initialValueKey] =
+                    animatedProps.initial?.value[initialValueKey];
+                }
+              );
+            }
```
tjzel:
```suggestion
          if (animatedProps?.viewDescriptors && animatedProps.initial) {
              Object.keys(animatedProps.initial.value).forEach(
                (initialValueKey) => {
                  props[initialValueKey] =
                    animatedProps.initial?.value[initialValueKey];
                }
              );
```

packages/react-native-reanimated/__typetests__/common/withSpringTest.tsx
MatiPl01: Were auto removed when I saved the file. It seems that we no longer have to ignore these rules here.

Maybe I should revert these changes, as they aren't related to the main PR's topic.

packages/react-native-reanimated/src/createAnimatedComponent/utils.ts
MatiPl01: Because of a faster check of emptiness

```ts
if (filteredStyles.cssStyle && filteredAnimatedProps.cssStyle) {}
```

I don't have to call `Object.keys()`, which has `O(n)` time complexity and `cssStyle` objects usually won't be empty, so it will add an unnecessary cost.

apps/common-app/src/apps/reanimated/examples/EmptyExample.tsx
piaskowyk, line 1: Let's restore changes in EmptyExample before merge

MatiPl01: Tomek asked me to separate animated props changes from css ones so I decided to create this [PR](https://github.com/software-mansion/react-native-reanimated/pull/7890) first and will update the current one later on to contain only CSS-specific changes.

MatiPl01: Here is a separate PR: https://github.com/software-mansion/react-native-reanimated/pull/7890

packages/react-native-reanimated/src/createAnimatedComponent/PropsFilter.tsx
```diff
@@ -58,21 +58,15 @@ export class PropsFilter implements IPropsFilter {
...
-        });
+        const animatedProp = inputProps.animatedProps as Partial<
+          AnimatedComponentProps<AnimatedProps>
+        >;
+        if (animatedProp.initial !== undefined) {
+          Object.keys(animatedProp.initial.value).forEach((initialValueKey) => {
+            props[initialValueKey] =
+              animatedProp.initial?.value[initialValueKey];
+          });
+        }
```
piaskowyk: I think I'm missing something because I don't understand why in the base branch you're handling `animatedProp` as an array to allow passing multiple animated props objects, but here you're handling `animatedProp` as an object instead of an array 🤔.

piaskowyk: or maybe this is just because of rebase

MatiPl01: It was changed because of a mistake during rebase. Thanks for noticing that

packages/react-native-reanimated/src/helperTypes.ts
MatiPl01: Added some tests as you requested

packages/react-native-reanimated/src/createAnimatedComponent/AnimatedComponent.tsx
```diff
@@ -307,9 +307,15 @@ export default class AnimatedComponent
     this._prevAnimatedProps = this._animatedProps;
     this._animatedProps = filteredAnimatedProps.animatedStyles;
 
-    this._cssStyle = filteredStyles.cssStyle;
-  }
+    if (filteredStyles.cssStyle && filteredAnimatedProps.cssStyle) {
+      logger.warn(
+        'AnimatedComponent: CSS properties cannot be used in style and animatedProps at the same time. Using properties from animatedProps.'
+      );
+    }
```
tjzel:
```suggestion
    if (__DEV__ && filteredStyles.cssStyle && filteredAnimatedProps.cssStyle) {
      logger.warn(
        'AnimatedComponent: CSS properties cannot be used in style and animatedProps at the same time. Using properties from animatedProps.'
      );
    }
```

tjzel (approved):
LGTM

PR #7750, opened 2025-07-04
feat: Support for SVG circle base props

MatiPl01:
## Summary

This PR adds support for base SVG circle properties interpolation and an example with just SVG-specific props. More example screens for different svg props will be added in separate PRs soon.

### DIFF
--- a/packages/react-native-reanimated/src/css/platform/native/configs/common.ts
+++ b/packages/react-native-reanimated/src/css/platform/native/configs/common.ts
@@ -1,3 +1,4 @@
+'use strict';
 import { processColor } from '../style';
 
 export const colorAttributes = { process: processColor };


packages/react-native-reanimated/src/css/platform/native/configs/svg/common.ts
```diff
@@ -0,0 +1,117 @@
+'use strict';
```
MatiPl01, line 1: I will adjust default values in this file in upcoming PRs when I implement examples for all of these props.

packages/react-native-reanimated/Common/cpp/reanimated/CSS/config/interpolators/svg/common.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
```
MatiPl01, line 1: I will adjust default values in this file in upcoming PRs when I implement examples for all of these props.

piaskowyk (approved):
👍

PR #7758, opened 2025-07-07
feat: Add CSS app examples search bar

MatiPl01:
## Summary

This PR adds powerful search bar to easily filter examples in the CSS example app, which is huge and will be even bigger after all SVG prop animation examples are added. To make the navigation process simpler, I decided to add a search bar powered by the `fuse.js` library to allow easy search by best matching results (not exact matches).

The search bar can be open on any routes screen (any screen that shows a list of routes to select from) by pulling the screen down.

Depending on which screen you start to search, the applied search filter will differ (by default, only results from the current screen and its subscreens - screens to which you can navigate to by selecting routes on the current screen, will be displayed in search results). Filters can be changed at any depth level by selecting the proper category/subcategory from the dropdown.

## Example recording

https://github.com/user-attachments/assets/d48a5504-5b60-44f1-b85f-f5fb02e05233

## TODO

- [ ] Improve web in a separate PR (I think that swipe gesture for search is not good on web and, since we have more space on the screen, it should be replaced by the search icon; it also doesn't work well now).


apps/common-app/package.json
```diff
@@ -28,6 +28,7 @@
     "@shopify/flash-list": "patch:@shopify/flash-list@npm%3A1.8.0#~/.yarn/patches/@shopify-flash-list-npm-1.8.0-54e02d8f74.patch",
     "d3-shape": "^3.2.0",
     "eslint-plugin-reanimated": "workspace:*",
+    "fuse.js": "patch:fuse.js@npm%3A7.1.0#~/.yarn/patches/fuse.js-npm-7.1.0-5dcae892a6.patch",
```
MatiPl01, line 31: Had to patch this because of the TS issue: https://github.com/krisk/Fuse/issues/780

PR #7767, opened 2025-07-10
Bump example apps to React Native 0.81

tomekzaw:
## Summary

Work in progress.

If you're getting `TypeError: (0 , _util.styleText) is not a function` during `yarn start`, please use Node.js 20 instead by running `nvm use 20`.

`RN_SERIALIZABLE_STATE`: https://github.com/facebook/react-native/pulls?q=sort%3Aupdated-desc+is%3Apr+rn_serializable_state+is%3Aclosed
* Merged separately in https://github.com/software-mansion/react-native-reanimated/pull/7776

## Test plan

|| Fabric (iOS) | Fabric (Android) | Web | MacOS | TVOS (iOS) | TVOS (Android) |
|-|-|-|-|-|-|-|
| Builds           | ✅ | ✅ | ✅ | not upgraded | ✅ (not upgraded) | ✅ (not upgraded) 
| Runs             | ✅ | ✅ | ✅ | not upgraded | ✅ (not upgraded) | ✅ (not upgraded)

### DIFF
--- a/apps/common-app/src/apps/css/components/examples/ExamplesScreen.tsx
+++ b/apps/common-app/src/apps/css/components/examples/ExamplesScreen.tsx
@@ -40,30 +40,35 @@ export default function ExamplesScreen<
   S extends AnyRecord = PlainStyle,
 >(props: ExamplesScreenProps<P, S>) {
   if ('tabs' in props) {
+    const renderTab = (
+      name: string,
+      { buildAnimation, renderExample, ...rest }: ExamplesListProps<P, S>
+    ) => {
+      return (
+        <TabView.Tab name={name}>
+          <ExamplesList
+            {...props}
+            {...rest}
+            buildAnimation={
+              buildAnimation ??
+              ('buildAnimation' in props
+                ? props.buildAnimation
+                : () => ({ animationName: {} }))
+            }
+            renderExample={
+              renderExample ??
+              ('renderExample' in props ? props.renderExample : () => <></>)
+            }
+          />
+        </TabView.Tab>
+      );
+    };
+
     return (
       <Screen>
         <TabView>
-          {props.tabs.map(
-            ({ buildAnimation, name, renderExample, ...rest }) => (
-              <TabView.Tab key={name} name={name}>
-                <ExamplesList
-                  {...props}
-                  {...rest}
-                  buildAnimation={
-                    buildAnimation ??
-                    ('buildAnimation' in props
-                      ? props.buildAnimation
-                      : () => ({ animationName: {} }))
-                  }
-                  renderExample={
-                    renderExample ??
-                    ('renderExample' in props
-                      ? props.renderExample
-                      : () => <></>)
-                  }
-                />
-              </TabView.Tab>
-            )
+          {props.tabs.map(({ name, ...rest }) =>
+            renderTab(name, rest as ExamplesListProps<P, S>)
           )}
         </TabView>
       </Screen>


apps/fabric-example/tsconfig.json
```diff
@@ -6,5 +6,13 @@
       "@/*": ["../common-app/src/*"]
     }
   },
-  "exclude": ["metro.config.js", "android", "ios", ".bundle", "node_modules"]
+  "include": ["**/*.ts", "**/*.tsx"],
```
MatiPl01: Do we really want to limit type checks to `ts` and `tsx` files? We have type comments in files like `babel.config.js` and `metro.config.js`, so I think it would be useful to keep type checking these files as well.

I think that these errors are useful. I can get the following error when the `includes` field is removed.

Screenshot 2025-07-16 at 10 54 06

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSValue.h
```diff
@@ -16,7 +16,7 @@ enum class RelativeTo {
 };
 
 struct CSSResolvableValueInterpolationContext {
-  const ShadowNode::Shared &node;
+  const std::shared_ptr<const ShadowNode> &node;
```
tjzel: Are these C++ changes backwards compatible with 0.80?

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
```diff
@@ -43,7 +43,7 @@ Props::Shared mergeProps(
   return newProps;
 }
 
-ShadowNode::Unshared cloneShadowTreeWithNewPropsRecursive(
+std::shared_ptr<ShadowNode> cloneShadowTreeWithNewPropsRecursive(
```
tjzel: In this PR you usually change `ShadowNode::Shared` to `shared_ptr`. Is the change from `ShadowNode::Unshared` intentional here?

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/AnimatedPropsRegistry.cpp
```diff
@@ -4,6 +4,15 @@
...
 
+#if REACT_NATIVE_MINOR_VERSION >= 81
+static inline std::shared_ptr<const ShadowNode> shadowNodeFromValue(
+    jsi::Runtime &rt,
+    const jsi::Value &shadowNodeWrapper) {
+  return Bridging<std::shared_ptr<const ShadowNode>>::fromJs(
+      rt, shadowNodeWrapper);
+}
+#endif
+
```
tjzel: Seems like it's unnecessary here.

tjzel (commented):
🧢

packages/react-native-reanimated/Common/cpp/reanimated/CSS/common/values/CSSValue.h
tomekzaw: Yes, I'm pretty sure this is backwards compatible

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/ShadowTreeCloner.cpp
tomekzaw: Yes, `ShadowNode::Unshared` is also deprecated in favor of `std::shared_ptr<ShadowNode>` (note the lack of `const` qualifier before `ShadowNode`).

packages/react-native-reanimated/Common/cpp/reanimated/Fabric/updates/AnimatedPropsRegistry.cpp
tomekzaw: Not really, `shadowNodeFromValue` is used in this file couple of lines below

apps/tvos-example/package.json
```diff
@@ -35,7 +35,7 @@
     "typescript": "~5.3.0"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=20.13.0"
```
MatiPl01, line 38: I changed the node version because I was unable to run metro with `node` versions `<= 20.12.0`. The first working version was `20.13.0`, so I set this version in `package.json` files.

With node `20.10.0` I was getting this error:

```
yarn start --reset-cache --client-logs

error (0 , _util.styleText) is not a function.
TypeError: (0 , _util.styleText) is not a function
    at Object.runServer [as func] (/Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native/community-cli-plugin/dist/commands/start/runServer.js:74:25)
    at async Command.handleAction (/Users/mateu/Work/projects/react-native-reanimated/node_modules/@react-native-community/cli/build/index.js:139:9)
```

It seems that the node module, from which the `styleText` function is imported, doesn't exist in older node versions.

tjzel (approved):

PR #7768, opened 2025-07-10
Fix zombie view when parent is moved

bartlomiejbloniarz:
## Summary

This PR fixes how we handle the case when a view is supposed to be removed but its parent is being moved due to view flattening, and has an exiting animation configured. Currently this scenario would lead to a state where the children are not removed (because there is an exiting animation configured), but the animation would not start (because the view is not deleted- it's only moved). This would lead to the children being left on the screen forever.

## Test plan

Test the `[LA] Move with exiting` example.


tomekzaw (approved):
Thanks for fixing this!

PR #7770, opened 2025-07-10
fix: make web work with bundlers that don't allow "require" in ES modules

### LINKED ISSUE #6775: Building a for web with vite causes an error when doing a production build
### Description

When running in dev with vite everything works fine, when I run the build for production for example with `npm run build && npm run preview` in the repro then I get this error

Initially I ran into this error when working on a new vite framework for storybook but I found the same error present in a users example on the discussions here where it works on dev but fails when serving the static files.

```
index-DvzMhtGV.js:53 Uncaught TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at dl (index-DvzMhtGV.js:53:6327)
    at index-DvzMhtGV.js:53:48672
    at Array.forEach (<anonymous>)
    at Dc (index-DvzMhtGV.js:53:48646)
    at Object.a [as worklet] (index-DvzMhtGV.js:53:50781)
    at o (index-DvzMhtGV.js:53:13912)
```

image

### Steps to reproduce

in the repro:
1. click the terminal
2. ctrl + c to end the dev server
3. use the command `npm run build && npm run preview` to run a preview of the prod build
4. note the animation fails and open the browser console to see the error

in general: 
1. setup react native web with vite
2. add reanimated and add the babel plugin to the react plugin's babel config
3. add an animated component and attempt to animate it
3. build for prod

https://stackblitz.com/edit/vitejs-vite-wcfpnl?file=package.json

same repo from the stack blitz: 
https://github.com/dannyhw/vite-reanimated-web-prod-repro

to run here just do
`npm install`
`npm run build && npm run preview`

The example code is actually mostly taken from here: https://github.com/software-mansion/react-native-reanimated/discussions/5814

In which disabling strict mode fixes one problem but then in the build theres this new error

### Reanimated version

~3.16.3

### React Native version

^0.76.3

### Platforms

Web

### JavaScript runtime

V8

### Workflow

None

### Architecture

None

### Build type

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

matthieugicquel:
## Summary

Fixes #6775

When building for production, vite doesn't like having `export` and `require` in the same file: it considers the file an ES module, and `require` is undefined in it.

We want reanimated to be buildable with vite because it's widely used, eg for storybook (see #6775).

## Test plan

Without this patch, most reanimated features (eg a simple `useAnimatedStyle`) were broken in a vite storybook prod build.

I applied an identical patch on an app I'm working on and verified that everything builds and works with:

- [x] vite dev and prod (via storybook)
- [x] webpack (via nextjs)
- [x] metro for native iOS/Android
- [  ] metro for web (still need to check that there's no regression on this one)

I don't see how I can add an automated test for this fix, short of introducing a whole storybook or vite build pipeline, which you probably don't want.

### DIFF
--- a/apps/next-example/next.config.js
+++ b/apps/next-example/next.config.js
@@ -15,7 +15,12 @@ module.exports = withPlugins([withBundleAnalyzer, withExpo], {
   eslint: {
     ignoreDuringBuilds: true,
   },
-  transpilePackages: ['react-native-reanimated', 'react-native', 'expo'],
+  transpilePackages: [
+    'react-native-reanimated',
+    'react-native',
+    'react-native-web',
+    'expo',
+  ],
   webpack(config) {
     if (disableMinification) {
       config.optimization.minimizer = [];


tjzel:
Hi @matthieugicquel, thanks for the contribution. The requires here are to optionally resolve dependencies - we aren't granted that these imports are available and as you can see on CI https://github.com/software-mansion/react-native-reanimated/actions/runs/16196012959/job/45722744583?pr=7770 Next build fails because of unavailable import.

Do you know how to handle optional/dynamic imports in that case?

matthieugicquel:
> Hi @matthieugicquel, thanks for the contribution. The requires here are to optionally resolve dependencies - we aren't granted that these imports are available and as you can see on CI https://github.com/software-mansion/react-native-reanimated/actions/runs/16196012959/job/45722744583?pr=7770 Next build fails because of unavailable import.
> 
> Do you know how to handle optional/dynamic imports in that case?

I compared the setup in the example app with an app I'm working on, and it turns out the MODULE_NOT_FOUND issue gets solved by adding `react-native-web` to `transpilePackages` in the next config ([which is something solito does](https://github.com/nandorojo/solito/blob/be03583de99426d9ac8f8504e9ef90924e79b264/example-monorepos/blank/apps/next/next.config.js#L73))

I'm not sure why yet...

As for making the import conditional:
- as far as I know, the only way that is spec-compliant (= will make most bundlers happy) is with `await import`, which would probably introduce other issues.
- and it seems safe to assume react-native-web should be available when `.web` files are being used by the bundler.
- also, since most reanimated features are broken without these imports, I would even say it's a good thing to break the build when these imports are not found

→ I wouldn't try to find a way to keep the conditional imports

I'll add a commit to modify the config in the example app, tell me what you think :)

tjzel:
I think you convinced me. There're still more files with conditional imports, were the ones you changed the only troublesome ones for you?

alexandrermo:
HI @tjzel, thanks for the awesome library!
Do you think it is possible to accept this PR? We are experiencing this bug using reanimated with Vite on our project too.

tjzel:
There are more uses of `require` in other places in the code and I'm afraid this PR in its current shape won't fix all troublesome bits.

Perhaps dynamic ESM `import` would work better as replacement for all `require` uses but that has to be verified first.

alexandrermo:
@tjszel Hello 🙌 
We've successfully tested these changes as a patch in our Storybook + Vite setup and they're working. This resolves the production build issues we were experiencing, but we had to freeze the Reanimated version to ensure the patch applies correctly.

PR #7772, opened 2025-07-10
[LA] Don't clone props when not necessary

bartlomiejbloniarz:

### DIFF
--- a/packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
+++ b/packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
@@ -19,18 +19,23 @@ struct Rect {
 
 struct Frame {
   std::optional<double> x, y, width, height;
+  int count = 0;
   Frame(jsi::Runtime &runtime, const jsi::Object &newStyle) {
     if (newStyle.hasProperty(runtime, "originX")) {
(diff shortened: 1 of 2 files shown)


packages/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsUtils.h
```diff
@@ -19,18 +19,23 @@ struct Rect {
 
 struct Frame {
   std::optional<double> x, y, width, height;
+  int count = 0;
   Frame(jsi::Runtime &runtime, const jsi::Object &newStyle) {
     if (newStyle.hasProperty(runtime, "originX")) {
       x = newStyle.getProperty(runtime, "originX").asNumber();
+      count++;
     }
     if (newStyle.hasProperty(runtime, "originY")) {
```
tomekzaw, line 28: We should add early returns or use `else if` instead. Once this is done, we can just a boolean to store the information whether there is some custom style or not.
```suggestion
    else if (newStyle.hasProperty(runtime, "originY")) {
```

bartlomiejbloniarz: We can't, the purpose of this class is to aggregate non-props updates in layout animations. We need to parse all of them, and we need to count them all, so that we can calculate how many props updates are there in a given layout animation.

PR #7777, opened 2025-07-11
fix(Reanimated): Worklets version validation scripts failing in runtime

tjzel:
## Summary

`process.exit` function can be unavailable in runtime. Therefore runtime script must not use it.

## Test plan

@tomekzaw 🚊 🚋 🚌

### DIFF
--- a/packages/react-native-reanimated/scripts/validate-worklets-build.js
+++ b/packages/react-native-reanimated/scripts/validate-worklets-build.js
@@ -7,6 +7,6 @@ const validateVersion = require('./validate-worklets-version');
 const result = validateVersion();
 if (!result.ok) {
   // eslint-disable-next-line reanimated/use-logger
-  console.error('[Reanimated]' + result.message);
+  console.error('[Reanimated] ' + result.message);
   process.exit(1);
 }
--- a/packages/react-native-reanimated/scripts/validate-worklets-build.js
+++ b/packages/react-native-reanimated/scripts/validate-worklets-build.js
@@ -1,0 +1,12 @@
+'use strict';
+
+// We don't use this script in runtime since `process` might not be available.
+
+const validateVersion = require('./validate-worklets-version');
+
+const result = validateVersion();
+if (!result.ok) {
+  // eslint-disable-next-line reanimated/use-logger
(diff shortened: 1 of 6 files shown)


packages/react-native-reanimated/src/platform-specific/workletsVersion.ts
```diff
@@ -1,6 +1,15 @@
...
+/* eslint-disable @typescript-eslint/no-require-imports */
 'use strict';
 
+import { ReanimatedError } from '../common/errors';
+
 export function assertWorkletsVersion() {
-  // eslint-disable-next-line @typescript-eslint/no-require-imports, n/no-missing-require
-  require('react-native-reanimated/scripts/validate-worklets-version');
+  const result =
+    require('react-native-reanimated/scripts/validate-worklets-version')();
```
piaskowyk: do we need to use require since we have `module.exports` in `validate-worklets-version`?

tjzel: Changed it to `import` 👍

packages/react-native-reanimated/src/platform-specific/workletsVersion.ts
```diff
@@ -1,6 +1,14 @@
 'use strict';
 
+//@ts-expect-error There's no types for this script.
```
tomekzaw:
```suggestion
// @ts-expect-error There's no types for this script.
```

packages/react-native-reanimated/scripts/validate-worklets-build.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+// We don't use this script in runtime since `process` might not be available.
+
+const validateVersion = require('./validate-worklets-version');
+
+const result = validateVersion();
+if (!result.ok) {
+  // eslint-disable-next-line reanimated/use-logger
+  console.error('[Reanimated]' + result.message);
```
tomekzaw:
```suggestion
  console.error('[Reanimated] ' + result.message);
```

tomekzaw (approved):

